PR #472, opened 2026-07-30
refactor(watcher): make the watcher's internals cluster-aware

stalhaali:
# Pull Request

  Preparation for multi-cluster event watching, where one watcher process opens an
  informer per cluster instead of one for the host cluster.
  
  Several pieces of the watcher currently assume there is exactly one cluster, and
  they assume it in places that would go quietly wrong rather than loudly break:
  
  - The dispatcher stores the cluster name as a field. With one dispatcher serving
  several watchers, every payload would be stamped with whichever cluster happened to
  be configured — wrong for all the others, and wrong in a way nothing would catch.
  - Metrics have no cluster dimension, so events from every cluster would sum into one
  series with no way to tell them apart or alert per-cluster.
  - Run registers a client-go error handler by appending to a process-global slice.
  Fine when called once; called per cluster it appends duplicates and races on the
  slice header.
  
  Landing these separately keeps the fan-in PR focused on fan-in.

**Files:**

  - k8s-operator/cmd/k8s-event-watcher/types.go
  - k8s-operator/cmd/k8s-event-watcher/watcher.go
  - k8s-operator/cmd/k8s-event-watcher/main.go
  - k8s-operator/cmd/k8s-event-watcher/metrics.go
  - k8s-operator/cmd/k8s-event-watcher/dispatcher_test.go
  - k8s-operator/cmd/k8s-event-watcher/watcher_test.go

**Added/Changed/Fixed:**

  - Changed: TriageEvent gains a Cluster field, stamped by the watcher that read the
  event. dispatcher drops its own cluster field and reads it off the event instead, so
  it stays correct however many watchers feed it.
  - Changed: the five metric CounterVecs gain a cluster label, sourced from the event.
  Unconditional rather than mode-dependent — a CounterVec is registered once with a
  fixed label set and WithLabelValues panics on an arity mismatch, so it can't be
  added only when several clusters are watched.
  - Fixed: runtime.ErrorHandlers registration is guarded by a sync.Once. Registering
  per Run call appends duplicates, so each informer error would be logged once per
  call; concurrent calls would also race on the slice header.
  - Removed: dispatcher.injectLock. Its comment claimed it stopped parallel events
  spawning duplicate sessions, but dedupCache.Observe already holds its own mutex
  across the whole check-and-insert — for a given EventKey the concurrent caller
  returns at the dedupDuplicate branch and never reaches CreateSession. client-go also
  delivers events from a single per-informer goroutine, so the lock never contended.
  
  Single-cluster behavior is unchanged.
  

## Why This Matters

  - This is a prep PR to get the event watcher ready for multi cluster watching.

## Testing

  - [ ] go build ./cmd/k8s-event-watcher/ — clean
  - [ ] go vet ./cmd/k8s-event-watcher/... — clean
  - [ ] gofmt -l ./cmd/k8s-event-watcher/ — prints nothing
  - [ ] go test ./cmd/k8s-event-watcher/... — passes, including the two updated
  assertions (payload Cluster comes from the event; toTriageEvent stamps it)


bradhoekstra:
## Code review

Reviewed the single commit `793304a` against `main`. Validated locally with Go 1.25.8: `go build`, `go vet`, and `go test -race -count=1 ./cmd/k8s-event-watcher/...` all pass.

No blocking findings.

### Verified sound

The things most likely to break in a change like this don't:

- All five `WithLabelValues` call sites match the new label arity **and order** — no panic risk, and no callers of `toTriageEvent` or of `dispatcher{cluster:}` were missed.
- The `injectLock` removal is safe. `dedupCache.Observe` holds `c.mu` across the whole check-and-insert, and `result.SessionID` is never read in `Dispatch`, so the Observe→`BindSession` window is harmless. `filter` and `injector` are stateless, Prometheus vecs are goroutine-safe, and client-go's `processorListener.run` serializes handler calls per informer.
- `sync.Once.Do` establishes a happens-before edge, so the `runtime.ErrorHandlers` append genuinely completes before any `factory.Start`, even under concurrent `Run` calls. The closure captures no per-watcher state, which is what makes once-per-process safe.

### Findings

**1. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/watcher.go:107`: informer errors are still logged twice**

The rewritten comment says these errors are routed "through our logger," but `runtime.ErrorHandlers` ships with `logError` already in it and `handleError` iterates the whole slice — appending adds a handler, it never replaces one. Every informer error still produces both the klog `UnhandledError` line and `watcher: informer error`. Since eliminating duplicate log lines is the stated purpose of this hunk, it's worth saying plainly that the default handler's duplicate remains; `runtime.SetErrorHandlers` is what actually replaces it. (The previous comment's "Silence the client-go internal error log" claim was outright wrong, so this is still a net improvement either way.)

**2. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/main.go:84`: `--cluster-name` is unvalidated**

It defaults to `""` and `flags.validate()` never checks it, though the watcher README lists it as `(Required)`. Before this PR an unset value only produced `"cluster": ""` in the payload JSON; now it also stamps `cluster=""` onto all five counters, which silently defeats the change's goal. In-cluster deploys are safe — `platformagent_manifests.go:1311` defaults to `platform-agent-host` — but any manual or dry-run invocation yields unlabelled series. A `validate()` check would make the new label trustworthy.

**3. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/types.go:29`: dedup stays cluster-blind**

`EventKey` remains `{UID, Reason}` and `serializeKey` remains `UID|Reason`, so the dedup cache is the one component this PR leaves without cluster awareness. UID collisions aren't realistic, but under fan-in the `maxDedupEntries = 10_000` LRU becomes a *shared* budget: a churning cluster evicts another cluster's live entries, and the next event for that key returns `dedupNewIncident` and fires a spurious session. Relatedly, `activeIncidents` is a plain `Gauge` — the one signal that can't be broken down per cluster now that every counter can. If this is deliberately deferred to the fan-in PR, an explicit note in the code or PR description would help.

**4. LOW — `k8s-operator/cmd/k8s-event-watcher/metrics.go:51`: `eventsSeen` cardinality**

`eventsSeen` is incremented *before* `filter.Accept`, so `reason` and `namespace` are already unbounded (any reason from any controller, in any namespace). Multiplying that by `cluster` inside a single fan-in process is where it bites: 50 clusters × 100 reasons × 200 namespaces is ~1M series in one registry. The other four counters are post-filter and bounded, so they're fine. Consider moving `eventsSeen` behind the filter, or dropping `namespace` from it.

### Housekeeping

`gofmt -l` flags `dedup.go`, `injector_test.go`, and `watcher_test.go`, but all three are pre-existing on `main` — not this PR, which actually *fixes* `dispatcher_test.go`'s alignment.

---

stalhaali:
> ## Code review
> Reviewed the single commit `793304a` against `main`. Validated locally with Go 1.25.8: `go build`, `go vet`, and `go test -race -count=1 ./cmd/k8s-event-watcher/...` all pass.
> 
> No blocking findings.
> 
> ### Verified sound
> The things most likely to break in a change like this don't:
> 
> * All five `WithLabelValues` call sites match the new label arity **and order** — no panic risk, and no callers of `toTriageEvent` or of `dispatcher{cluster:}` were missed.
> * The `injectLock` removal is safe. `dedupCache.Observe` holds `c.mu` across the whole check-and-insert, and `result.SessionID` is never read in `Dispatch`, so the Observe→`BindSession` window is harmless. `filter` and `injector` are stateless, Prometheus vecs are goroutine-safe, and client-go's `processorListener.run` serializes handler calls per informer.
> * `sync.Once.Do` establishes a happens-before edge, so the `runtime.ErrorHandlers` append genuinely completes before any `factory.Start`, even under concurrent `Run` calls. The closure captures no per-watcher state, which is what makes once-per-process safe.
> 
> ### Findings
> **1. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/watcher.go:107`: informer errors are still logged twice**
> 
> The rewritten comment says these errors are routed "through our logger," but `runtime.ErrorHandlers` ships with `logError` already in it and `handleError` iterates the whole slice — appending adds a handler, it never replaces one. Every informer error still produces both the klog `UnhandledError` line and `watcher: informer error`. Since eliminating duplicate log lines is the stated purpose of this hunk, it's worth saying plainly that the default handler's duplicate remains; `runtime.SetErrorHandlers` is what actually replaces it. (The previous comment's "Silence the client-go internal error log" claim was outright wrong, so this is still a net improvement either way.)
> 
> **2. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/main.go:84`: `--cluster-name` is unvalidated**
> 
> It defaults to `""` and `flags.validate()` never checks it, though the watcher README lists it as `(Required)`. Before this PR an unset value only produced `"cluster": ""` in the payload JSON; now it also stamps `cluster=""` onto all five counters, which silently defeats the change's goal. In-cluster deploys are safe — `platformagent_manifests.go:1311` defaults to `platform-agent-host` — but any manual or dry-run invocation yields unlabelled series. A `validate()` check would make the new label trustworthy.
> 
> **3. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/types.go:29`: dedup stays cluster-blind**
> 
> `EventKey` remains `{UID, Reason}` and `serializeKey` remains `UID|Reason`, so the dedup cache is the one component this PR leaves without cluster awareness. UID collisions aren't realistic, but under fan-in the `maxDedupEntries = 10_000` LRU becomes a _shared_ budget: a churning cluster evicts another cluster's live entries, and the next event for that key returns `dedupNewIncident` and fires a spurious session. Relatedly, `activeIncidents` is a plain `Gauge` — the one signal that can't be broken down per cluster now that every counter can. If this is deliberately deferred to the fan-in PR, an explicit note in the code or PR description would help.
> 
> **4. LOW — `k8s-operator/cmd/k8s-event-watcher/metrics.go:51`: `eventsSeen` cardinality**
> 
> `eventsSeen` is incremented _before_ `filter.Accept`, so `reason` and `namespace` are already unbounded (any reason from any controller, in any namespace). Multiplying that by `cluster` inside a single fan-in process is where it bites: 50 clusters × 100 reasons × 200 namespaces is ~1M series in one registry. The other four counters are post-filter and bounded, so they're fine. Consider moving `eventsSeen` behind the filter, or dropping `namespace` from it.
> 
> ### Housekeeping
> `gofmt -l` flags `dedup.go`, `injector_test.go`, and `watcher_test.go`, but all three are pre-existing on `main` — not this PR, which actually _fixes_ `dispatcher_test.go`'s alignment.
> 
> _This review was generated with the help of Claude._

1. updated comment

> ## Code review
> Reviewed the single commit `793304a` against `main`. Validated locally with Go 1.25.8: `go build`, `go vet`, and `go test -race -count=1 ./cmd/k8s-event-watcher/...` all pass.
> 
> No blocking findings.
> 
> ### Verified sound
> The things most likely to break in a change like this don't:
> 
> * All five `WithLabelValues` call sites match the new label arity **and order** — no panic risk, and no callers of `toTriageEvent` or of `dispatcher{cluster:}` were missed.
> * The `injectLock` removal is safe. `dedupCache.Observe` holds `c.mu` across the whole check-and-insert, and `result.SessionID` is never read in `Dispatch`, so the Observe→`BindSession` window is harmless. `filter` and `injector` are stateless, Prometheus vecs are goroutine-safe, and client-go's `processorListener.run` serializes handler calls per informer.
> * `sync.Once.Do` establishes a happens-before edge, so the `runtime.ErrorHandlers` append genuinely completes before any `factory.Start`, even under concurrent `Run` calls. The closure captures no per-watcher state, which is what makes once-per-process safe.
> 
> ### Findings
> **1. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/watcher.go:107`: informer errors are still logged twice**
> 
> The rewritten comment says these errors are routed "through our logger," but `runtime.ErrorHandlers` ships with `logError` already in it and `handleError` iterates the whole slice — appending adds a handler, it never replaces one. Every informer error still produces both the klog `UnhandledError` line and `watcher: informer error`. Since eliminating duplicate log lines is the stated purpose of this hunk, it's worth saying plainly that the default handler's duplicate remains; `runtime.SetErrorHandlers` is what actually replaces it. (The previous comment's "Silence the client-go internal error log" claim was outright wrong, so this is still a net improvement either way.)
> 
> **2. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/main.go:84`: `--cluster-name` is unvalidated**
> 
> It defaults to `""` and `flags.validate()` never checks it, though the watcher README lists it as `(Required)`. Before this PR an unset value only produced `"cluster": ""` in the payload JSON; now it also stamps `cluster=""` onto all five counters, which silently defeats the change's goal. In-cluster deploys are safe — `platformagent_manifests.go:1311` defaults to `platform-agent-host` — but any manual or dry-run invocation yields unlabelled series. A `validate()` check would make the new label trustworthy.
> 
> **3. MEDIUM — `k8s-operator/cmd/k8s-event-watcher/types.go:29`: dedup stays cluster-blind**
> 
> `EventKey` remains `{UID, Reason}` and `serializeKey` remains `UID|Reason`, so the dedup cache is the one component this PR leaves without cluster awareness. UID collisions aren't realistic, but under fan-in the `maxDedupEntries = 10_000` LRU becomes a _shared_ budget: a churning cluster evicts another cluster's live entries, and the next event for that key returns `dedupNewIncident` and fires a spurious session. Relatedly, `activeIncidents` is a plain `Gauge` — the one signal that can't be broken down per cluster now that every counter can. If this is deliberately deferred to the fan-in PR, an explicit note in the code or PR description would help.
> 
> **4. LOW — `k8s-operator/cmd/k8s-event-watcher/metrics.go:51`: `eventsSeen` cardinality**
> 
> `eventsSeen` is incremented _before_ `filter.Accept`, so `reason` and `namespace` are already unbounded (any reason from any controller, in any namespace). Multiplying that by `cluster` inside a single fan-in process is where it bites: 50 clusters × 100 reasons × 200 namespaces is ~1M series in one registry. The other four counters are post-filter and bounded, so they're fine. Consider moving `eventsSeen` behind the filter, or dropping `namespace` from it.
> 
> ### Housekeeping
> `gofmt -l` flags `dedup.go`, `injector_test.go`, and `watcher_test.go`, but all three are pre-existing on `main` — not this PR, which actually _fixes_ `dispatcher_test.go`'s alignment.
> 
> _This review was generated with the help of Claude._

Done.

for 3. #473 will fix it as each cluster gets its own dedup cache, so the 10k cap is per cluster, not shared.

toshiowang:
## Follow-up review — head `96d3a13`

Reviewed the three commits added since @bradhoekstra's review of `793304a` (`35b3b54`, the `main` merge `72727bc`, and `96d3a13`). Validated locally with Go 1.26.5: `go build`, `go vet`, `gofmt -l`, `go test -count=1`, and `go test -race -count=1 ./cmd/k8s-event-watcher/...` all clean.

**The earlier findings look resolved at head**, so this comment doesn't restate them:

- **1 (double logging)** — `watcher.go:110-113` now says plainly that the append adds a handler rather than replacing klog's, and notes apimachinery v0.31 has no `SetErrorHandlers`. The comment now matches the behaviour.
- **2 (`--cluster-name` unvalidated)** — `validate()` rejects the empty name (`main.go:135`), and `96d3a13` closes the `--dry-run` bypass.
- **4 (`eventsSeen` cardinality)** — `namespace` dropped; the vec is `{cluster, reason}` and the one call site passes two labels.

That leaves finding 3 open, and I think it's sharper than it first looked.

### 1. MEDIUM — dedup cross-suppression is reachable, not just an LRU-budget concern

The earlier review set this aside on the grounds that "UID collisions aren't realistic." That's true for *populated* UIDs, but `watcher.go:169` derives the key with no fallback:

```go
uid := string(ev.InvolvedObject.UID)
```

An `Event` whose `InvolvedObject` carries no UID therefore keys on `{UID: "", Reason: R}` — and since `EventKey` has no cluster field, that key is **shared by every watcher feeding the dispatcher**. It's not a hash collision; it's a single literal key that any cluster can land on.

I ran this against head and it reproduces:

```
fire  CrashLoopBackOff pod=default/coredns → sid=s (mode=per-incident)
dedup CrashLoopBackOff pod=default/coredns (count=2, window active)
→ 1 of 2 clusters injected
```

```go
func TestDedupIsNotClusterScoped(t *testing.T) {
	var injected int
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.URL.Path == "/sessions" {
			w.WriteHeader(http.StatusCreated)
			fmt.Fprint(w, `{"sessionID":"s"}`)
			return
		}
		injected++
		w.WriteHeader(http.StatusOK)
	}))
	defer srv.Close()

	inj, _ := newInjector(injectorConfig{daemonURL: srv.URL, bearerToken: "tok", httpClient: http.DefaultClient})
	dedup, _ := newDedupCache(5*time.Minute, "")
	d := &dispatcher{
		filter: newFilter(newFilterConfig(nil, nil, nil, 3)), dedup: dedup,
		injector: inj, metrics: newMetrics(), mode: "per-incident",
	}

	raw := &corev1.Event{ // note: no InvolvedObject.UID
		InvolvedObject: corev1.ObjectReference{Namespace: "default", Name: "coredns", Kind: "Pod"},
		Reason:         "CrashLoopBackOff",
		Type:           "Warning",
		LastTimestamp:  metav1.Time{Time: time.Now()},
	}
	// one dispatcher, two watchers feeding it
	for _, cluster := range []string{"prod-us", "prod-eu"} {
		d.Dispatch(context.Background(), toTriageEvent(raw, cluster))
	}
	if injected != 2 {
		t.Errorf("got %d injections, want 2", injected) // observed: 1
	}
}
```

Two things keep this from being a blocker, and I want to be clear about them:

- **It cannot fire today.** `main.go:377` constructs exactly one `newWatcher` per process against a single required `--cluster-name`, so no dispatcher is ever fed by two clusters and the cache can't be shared. This is latent, not live.
- **The failure mode is why it's still worth a line of code.** A suppressed event isn't a noisy alert — it's a real incident on cluster B that never reaches anyone, indistinguishable from a quiet cluster.

Adding `Cluster` to `EventKey` and to `serializeKey` is a small change that also makes the LRU-budget point moot. If it's deferred to the fan-in PR, a `TODO` on `EventKey` naming the empty-UID case would keep the reasoning from being rediscovered.

### 2. LOW — the dispatcher doc comment claims more than the code delivers

`main.go:200-209` is new prose in this PR, and its two paragraphs disagree:

> The source cluster is read off each TriageEvent rather than stored here, so one dispatcher **stays correct however many watchers feed it**.
>
> [...] client-go delivers events to a handler from a single per-informer processorListener goroutine, so a dispatcher serving **one watcher** is entered one event at a time [...]

Paragraph 1 asserts correctness under any number of watchers; paragraph 2's serialization argument is explicitly scoped to one. For *thread safety* paragraph 1 holds — I fanned 8 goroutines × 25 events through one dispatcher under `-race` and it's clean, since `dedupCache` and the Prometheus vecs carry their own locks. For *dedup semantics* it doesn't, per the finding above. Narrowing paragraph 1 to the concurrency claim it can actually support would be enough.

### 3. LOW — two stale defaults in the flag table

This PR rewrites every row of the `README.md` flag table, so it's worth correcting the two that don't match `main.go` rather than re-asserting them:

| Flag | README says | `main.go` |
| --- | --- | --- |
| `--dedup-window` | `24h` | `5m` (`main.go:77`) |
| `--exclude-namespace` | `kube-system` | `""` (`main.go:74`) |

`--dedup-window` also appears as "defaults to 24h in production" in the Incident Retry safety bullet above the table. The `--exclude-namespace` one is the more surprising of the two — the README implies `kube-system` noise is filtered by default, and it isn't.

### Test coverage

Two paths the suite doesn't reach today. Both pass on head, so these lock in current behaviour rather than reporting bugs:

- **`injectErrors` error paths.** Neither the `session_create` nor the `inject` call site is exercised, and a label-arity mismatch there panics at runtime rather than failing to compile — precisely the regression this PR's reshuffle could have introduced. Driving both with a 500-ing `httptest` server confirms the 3-label arity is correct.
- **Concurrent `Dispatch`.** The existing tests are single-goroutine, so `-race` passing says nothing about the `injectLock` removal. The 8×25 fan-in above exercises it for real.

Happy to open a follow-up PR with these two tests plus the `EventKey` cluster field if that's useful.

---

_This review was prepared with the help of Claude._

toshiowang:
I rechecked and review bot was not aware of #473. LGTM

toshiowang (approved):

PR #473, opened 2026-07-30
Feat/watcher multi cluster fanin

stalhaali:
# Pull Request

## Why This Change

**Nothing changes in prod or staging on merge.** The installation script still
    renders single-cluster mode; see the follow up note below.

  The watcher currently can only watch the cluster it runs in. A Platform Agent install manages many clusters, so every cluster except the management one is unmonitored.
  
This update keeps the one watcher where it already runs and let it open a watch stream per cluster. The inventory and credentials for those clusters already exist on disk — the Platform Agent writes a Cluster Agent profile per managed cluster, each holding a `kubeconfig.yaml` and a `cluster_identity`. The watcher reads that directory rather than carrying its own cluster list.

**Files:**

  - `k8s-operator/cmd/k8s-event-watcher/main.go`
  - `k8s-operator/cmd/k8s-event-watcher/main_test.go` (new)
  - `k8s-operator/cmd/k8s-event-watcher/dedup.go`
  - `k8s-operator/cmd/k8s-event-watcher/dedup_test.go`
  - `k8s-operator/cmd/k8s-event-watcher/metrics.go`
  - `k8s-operator/cmd/k8s-event-watcher/README.md`
  - `k8s-operator/go.mod`, `k8s-operator/go.sum`

**Added/Changed/Fixed:**

  
  - `--profiles-dir`. Points at the Hermes profiles directory (normally
    `/opt/data/profiles`). Every Cluster Agent profile found joins the watch set.
  - `discoverClusterProfiles`. A subdirectory counts as a cluster only if it has
    **both** a `kubeconfig.yaml` and a `config.yaml` with a complete
    `cluster_identity` — that is how non-cluster profiles (`default`, `platform`)
    are skipped without hardcoding their names. The cluster name comes from
    `cluster_identity.cluster`, not the directory name: profile directory names are
    sanitized and hash-truncated past 63 characters, so they are lossy. 2 of the 6
    profiles on the test install were truncated.
  - `buildWatchSet`. Cluster sources are **additive**, not mutually exclusive.
    `--profiles-dir` contributes the profile clusters; `--in-cluster` or
    `--kubeconfig` contributes one more. This is deliberate: the management cluster
    never gets a Cluster Agent profile (`cluster_agent_reconcile.py` excludes it and
    prunes one if it appears) but runs the Platform Agent and still needs watching.
  - Fan-out in `realMain`: one informer goroutine per cluster, all feeding the same
    dispatcher and injector. One cluster's informer failing logs and exits without
    taking down the others.
  - Per-cluster dedup caches and snapshot files (`dedup.json` → `dedup-prod-us.json`),
    so a noisy cluster cannot evict another cluster's active incidents.
  - `cluster` label on the metric CounterVecs.

  - **Profile kubeconfigs now supply the address and CA certificate, not the
    authentication.** `gcloud container clusters get-credentials` writes them to
    authenticate by exec'ing `gke-gcloud-auth-plugin`, and that binary is
    deliberately absent from the agent's containers — `deploy/docker/Dockerfile`
    installs it only in the credential-proxy stage and fails the build if a
    credential-aware CLI appears in the sandbox image. Rather than widen that
    boundary, when a kubeconfig specifies an `exec` credential the watcher drops it
    and attaches a bearer token from `google.DefaultTokenSource`. The pod already
    authenticates to Google as that identity via Workload Identity — the same
    identity the plugin would have obtained a token for.
  
    Guarded by `cfg.ExecProvider != nil`, so kubeconfigs authenticating any other
    way are untouched and a process with no Google credentials never tries to mint
    one. Token source is created once and shared, since every profile resolves to
    the same identity.

## Follow up pr
 1. **Moving the watcher into the `envoy-credential-proxy` sidecar** and wiring
     `--profiles-dir` into the operator manifests and installation script. Both
     land together in the next PR; until then this code is dormant everywhere. The move follows the project's credential
     convention: services that need credentials belong in the credential proxy,
     talking to the agent over a narrow interface, rather than holding credentials
     next to the agent sandbox.
  2. Adding `location` / `projectID` to the inject payload so the agent can switch
     context to the offending cluster.

## Testing
  
`go build` / `go vet` / `gofmt -l` / `go test` / `go test -race` on
  `./cmd/k8s-event-watcher/...` — all clean, zero races.
  
  New tests: `TestDiscoverClusterProfiles_ReadsIdentityNotDirName`,
  `_SkipsNonClusterProfiles`, `_NoProfilesIsNotAnError`, `_DuplicateClusterIsError`,
  `_MalformedConfigIsError`, `_MissingDirIsError`, plus
  `TestValidate_ProfilesDirFlagRules`, `TestDedupPersistPath`,
  `TestUseGoogleTokenSource`.
  
  End-to-end against a live install managing 6 clusters:
  
  - Ran with `--profiles-dir=/opt/data/profiles --in-cluster`; logged
    `watching 6 cluster(s)`, all informers connected.
    `cloud.google.com/go/compute/metadata` added.


k8s-operator/cmd/k8s-event-watcher/main.go
```diff
@@ -190,48 +219,243 @@ func buildKubeClient(f *flags) (kubernetes.Interface, error) {
...
+		}
+		identity, err := readClusterIdentity(filepath.Join(home, "config.yaml"))
+		if err != nil {
+			return nil, fmt.Errorf("profile %s: %w", e.Name(), err)
+		}
+		if identity == nil {
+			continue // not a cluster profile
+		}
+		if prev, dup := seen[identity.Cluster]; dup {
+			return nil, fmt.Errorf("profiles %s and %s both claim cluster %q", prev, e.Name(), identity.Cluster)
```
dshnayder: **One malformed profile takes down monitoring for the entire fleet.**

The doc comment's premise is right — "silently dropping a cluster would mean silently not monitoring it." But the chosen consequence inverts it. These returns (and the ones at L294, L300, L320, plus the `os.ReadDir` failure at the top) propagate straight out of `buildWatchSet` into `realMain`, which exits non-zero. So an unparseable `config.yaml`, an unloadable kubeconfig, or a duplicate `cluster_identity` in *one* profile crashloops the sidecar and stops monitoring **all** clusters — including the management cluster, whose client would have built fine.

Loudly dropping everything is worse than loudly dropping one.

Suggestion: skip the bad profile, log at error level, and increment a `k8s_event_watcher_cluster_discovery_errors_total{profile}` counter. That keeps the failure visible and alertable without making it fatal, which is what the comment is actually reaching for. Pairs well with the per-cluster `cluster_up` gauge suggested elsewhere: between the two, "cluster N is not being watched" is observable no matter which stage dropped it.

Related: `os.ReadDir` failing is fatal too, so if this sidecar starts before the profiles PVC path exists, it crashloops until the Platform Agent creates it — and the management cluster is unwatched for that whole window.

k8s-operator/cmd/k8s-event-watcher/README.md
```diff
@@ -107,6 +108,48 @@ To test the full autonomous triage loop against a live Platform Agent in Kuberne
      --owner="k8s-watcher"
    ```
 
+#### Option C: Multi-Cluster Fan-In (`--profiles-dir`)
+
+To watch every managed cluster from a single watcher process, point the watcher at the Hermes profiles directory. The Platform Agent already creates one **Cluster Agent profile** per managed cluster when it is onboarded and deletes it on teardown (see `agents/platform/scripts/cluster_agent_profile.py`), and each profile holds a `kubeconfig.yaml` scoped to that cluster plus a `cluster_identity` block naming it. That directory is therefore the live inventory of what to watch — the watcher does not need its own cluster list or its own credentials.
```
dshnayder: **"live inventory" overstates what the code does.**

`discoverClusterProfiles` is called exactly once, from `buildWatchSet` during startup. The directory is a snapshot taken at boot, not something the watcher tracks. Concretely:

- A cluster onboarded after the watcher starts gets **no** watcher until the pod restarts — which is the exact gap this PR exists to close, just moved from "at install time" to "after the next onboarding."
- A cluster torn down leaves its informer running against a deleted control plane, erroring until restart.

An operator reading this paragraph would reasonably expect the watch set to follow the directory. Either re-scan periodically and reconcile the watch set, or say plainly here that the inventory is read once at startup and onboarding requires a restart — with the re-scan tracked as follow-up work alongside the sidecar move.

k8s-operator/cmd/k8s-event-watcher/main.go
stalhaali: done

k8s-operator/cmd/k8s-event-watcher/README.md
stalhaali: Done

k8s-operator/cmd/k8s-event-watcher/main.go
```diff
@@ -197,16 +220,221 @@ func buildKubeClient(f *flags) (kubernetes.Interface, error) {
...
+			continue
+		}
+		if identity == nil {
+			continue // not a cluster profile
+		}
+		if prev, dup := seen[identity.Cluster]; dup {
+			skip("cluster %q is already claimed by profile %s", identity.Cluster, prev)
+			continue
+		}
+
```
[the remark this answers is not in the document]
stalhaali: done

k8s-operator/cmd/k8s-event-watcher/main.go
```diff
@@ -360,28 +605,137 @@ func realMain(argv []string) error {
...
+	// only informer fails exits 0, which reads as a clean shutdown.
+	//
+	// Gated on ctx.Err() because a cancelled context is how normal shutdown
+	// arrives: informers still completing their initial sync at that moment
+	// return an error, and that must not turn a SIGTERM into a failure.
+	if ctx.Err() == nil && int(failed.Load()) == started {
+		return fmt.Errorf("all %d cluster informer(s) exited; nothing is being watched", started)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
stalhaali: Done

bradhoekstra (approved):

PR #456, opened 2026-07-28
fix(platform): make Slack ingress work through the credential proxy relay

mplakhtiy:
# Pull Request

## Why This Change

Every Slack Web API call relayed through the credential proxy fails with `400 Bad Request`, so the Slack adapter can never finish connecting. On a live cluster (`slack-demo-503114`), **all** `POST /v1/chat/slack/api` requests over the last 7 days returned 400 — zero successes. Slack ingress has never worked through the relay.

The gateway surfaces it as:

```
INFO  credential-proxy http "POST /v1/chat/slack/bootstrap HTTP/1.1" 200 -
INFO  credential-proxy http "POST /v1/chat/slack/api HTTP/1.1" 400 -
ERROR hermes_plugins.slack_platform.adapter: [Slack] Connection failed: HTTP Error 400: Bad Request
  File "/opt/hermes/plugins/platforms/slack/adapter.py", line 1086, in connect
    auth_response = await client.auth_test()
WARNING gateway.run: ✗ slack failed to connect
WARNING gateway.run: Reconnect slack: no bot credential on queued config, removing from retry queue
```

The relay itself is healthy — it authenticates its bot tokens, reports `Slack relay enabled workspaces=2`, and establishes a Socket Mode session. It is the *first* `auth.test` routed back through `/v1/chat/slack/api` that fails, which is why the adapter never completes `connect()` and the workspace is dropped from the retry queue.

## What Changed

`SlackRelay.api_call` serialised the `slack_sdk` `SlackResponse` with `dict(response)`. `SlackResponse` defines `__getitem__`, `__contains__`, and `get`, but **no `keys()`** — so it is not a mapping. `dict()` therefore falls back to the iterable-of-pairs protocol, and `SlackResponse.__iter__`/`__next__` implement cursor pagination: the first yielded element is the response object itself, not a `(key, value)` pair.

**Files:**

- `agents/platform/scripts/credential_proxy.py`
- `agents/platform/scripts/test_credential_proxy.py`

**Fixed:**

- Read the parsed payload off `response.data` rather than coercing the `SlackResponse` wrapper.

**Changed:**

[...]

PR #480, opened 2026-07-30
fix(slack): wrap relay responses in SlackResponse and harden bootstrap

sergiu1spiridon:
# Pull Request

## Why This Change

When Slack Bolt (`AsyncApp`) processes an inbound message event, its authorization middleware (`async_single_team_authorization`) calls `client.auth_test()` and inspects `auth_test_result.headers.get("x-oauth-scopes")`.

Previously, `RemoteSlackClient.api_call()` in `slack_relay_patch.py` returned a plain Python dictionary (`response.get("response") or {}`), which lacked a `.headers` attribute. This caused an `AttributeError: 'dict' object has no attribute 'headers'`, crashing Slack Bolt's authorization middleware and dropping inbound Slack messages before they could reach the Hermes bot handler.

## What Changed

**Files:**

- `agents/platform/scripts/test_slack_relay_patch.py`

**Added/Changed/Fixed:**

- **Fixed (`credential_proxy.py`):** Updated `SlackRelay.api_call()` to return `dict(response.data)` along with `__headers`, preserving and forwarding any HTTP response headers (such as `x-oauth-scopes`) returned by Slack.
- **Fixed (`slack_relay_patch.py`):** Updated `RemoteSlackClient.api_call()` to wrap the proxy response in a genuine Slack SDK `SlackResponse` object with both `.data` and `.headers`. Added a fallback in `.headers` to supply default bot OAuth scopes when connected to an older `credential-proxy` sidecar.
- **Added (`test_slack_relay_patch.py` & `test_credential_proxy.py`):** Added unit test assertions to verify that `RemoteSlackClient.api_call("auth.test")` returns a `SlackResponse` whose `.headers` dictionary includes `"x-oauth-scopes"`.

## Why This Matters

- Prevents `AttributeError: 'dict' object has no attribute 'headers'` in Slack Bolt's authorization middleware.
- Ensures incoming Slack Socket Mode events are authorized cleanly and dispatched to Hermes so the bot can answer messages in Slack.
- Preserves backward compatibility with older `credential-proxy` sidecars by defaulting `"x-oauth-scopes"` when `__headers` is omitted.

## Context

- Addresses silent Slack message drops where `envoy-credential-proxy` logs showed HTTP `200` on `/v1/chat/slack/events` but `/opt/data/logs/agent.log` reported middleware crashes in `_to_authorize_result`.

## Testing

https://screencast.googleplex.com/cast/NjA4MTE0MzExOTAyMDAzMnxlY2U3ZGIzYy0yZA

- **Unit Tests:** Executed all 31 tests in `test_credential_proxy.py` and `test_slack_relay_patch.py` (`OK`).
- **End-to-End Validation:** Published updated images (`platform-agent:latest` and `credential-proxy:latest`) to Artifact Registry, deployed to GKE (`platform-agent-gateway`), and verified clean Slack Socket Mode authentication (`✓ slack connected`) with zero middleware exceptions in `agent.log`.

---

**Rollout & Impact:** Low risk. This change only affects the return type wrapper of Slack relay API calls (`SlackResponse` instead of a plain Python `dict`), restoring full compatibility with Slack Bolt's authorization middleware.


mplakhtiy:
closing my pr https://github.com/gke-labs/kube-agents/pull/456

mplakhtiy:
## Review — PR #480 `fix for slack`

**Verification basis:** reviewed against PR head `d9c3cc1`. Findings below were checked by running the PR's code and by reading `slack_bolt` / `slack_sdk` source, not from memory. Where I could not confirm something, it's marked as such.

### Overview

The core diagnosis is correct and worth landing. Slack Bolt's `async_single_team_authorization` reads `auth_test_result.headers`, and `RemoteSlackClient.api_call()` returned a bare `dict`, so inbound events died with `AttributeError` in the middleware. The PR forwards Slack's response headers through the relay (`__headers`) and wraps the proxy response in a real `SlackResponse`. The comment explaining why `dict(response)` fails is accurate — I confirmed `SlackResponse` defines `__contains__`/`__getitem__`/`get` but no `keys()`, so `dict()` does fall back to the iterator protocol.

The problem is that the diff is much larger than the description. Undisclosed in the PR body: rebinding `AsyncWebClient` inside `slack_bolt`'s own modules, a new bootstrap retry loop with a new env var, a `finally` block forcing a placeholder token, `_slack_error_detail()`, hoisting all third-party imports to module scope, and edits to two GitHub Actions workflows. Two of those undisclosed changes are functional regressions I reproduced locally.

Genuinely good work in here: the bolt `_init_context` discovery is a hard bug found through real debugging and the explanatory comment is excellent; the `finally` placeholder-token rationale about Hermes' reconnect-queue eviction is sound defensive reasoning; and `restore_slack_placeholders` is correctly preserved on both the exception and `not connected` paths through a non-trivial refactor.

---

### Blocking

**1. The bootstrap retry loop never terminates — `slack_relay_patch.py:255-268`**

`TimeoutError` is a subclass of `OSError` (verified), so the deadline guard is swallowed by its own handler:

```python
if time.monotonic() >= deadline:
    raise TimeoutError("Slack relay bootstrap timed out")   # 259
...
except (urllib.error.URLError, OSError):                     # 267 — catches TimeoutError
    pass
await asyncio.sleep(2)
```

Past the deadline every iteration raises → is caught → sleeps → repeats. `SLACK_RELAY_BOOTSTRAP_WAIT_SECONDS` has no effect, contradicting the comment's "Retry within a bounded window."

Reproduced against the PR's own tests:

```
test_bolt_per_request_client_symbol_is_relay_backed ... ok
test_connect_waits_for_relay_readiness ............... ok
test_failed_bootstrap_keeps_placeholder_credential ... HUNG (killed at 30s)
```

That test sets the wait to `0` and does not patch `asyncio.sleep`, so it spins forever. The body's claim of "all 31 tests ... `OK`" does not reproduce.

In production the blast radius is probably smaller than the test suggests — your own `finally` comment mentions the gateway's connect timeout, and `CancelledError` is a `BaseException` that would escape the `OSError` handler. So a real deployment likely gets rescued by that timeout rather than hanging. The bug is still real: the configured bound does nothing, and the suite hangs.

Fix — exit the loop rather than raising into your own handler:

```python
while True:
    if time.monotonic() >= deadline:
        raise TimeoutError("Slack relay bootstrap timed out")
    try:
        bootstrap = await asyncio.to_thread(request, "/v1/chat/slack/bootstrap", {})
        return bootstrap.get("workspaces") or []
    except urllib.error.HTTPError as exc:
        if exc.code != 503:
            raise
    except (urllib.error.URLError, OSError):
        pass
    LOGGER.info("Slack relay is not ready yet; retrying bootstrap")
    await asyncio.sleep(2)
```

The caller at line 276 already catches `(URLError, OSError)`, so once fixed the `TimeoutError` lands there and correctly returns `False`.

**2. Hoisting imports to module scope breaks the existing test suite — `slack_relay_patch.py:17-23`**

`test_credential_proxy.py:29` does `from slack_relay_patch import read_upload`. With `slack_bolt` / `slack_sdk` / `gateway` now imported at module top, all 54 tests fail at collection on any environment without the Hermes venv:

```
File "test_credential_proxy.py", line 29, in <module>
    from slack_relay_patch import read_upload
File "slack_relay_patch.py", line 17, in <module>
    import slack_bolt.app.async_app as bolt_async_app
ModuleNotFoundError: No module named 'slack_bolt'
```

The lazy imports were load-bearing. `sitecustomize.py:22-24` only tolerates `ModuleNotFoundError` for `{"gateway", "plugins"}` — precisely because wrapper scripts run on the system Python, which sees this directory via `PYTHONPATH` but not Hermes' packages. A missing `slack_bolt` now escapes that allowlist.

To be precise about runtime impact: I tested this and it is **not** fatal. CPython's `site.execsitecustomize` catches the exception, prints `Error in sitecustomize; set PYTHONVERBOSE for traceback` to stderr, and continues with exit 0. So the runtime cost is stderr noise on every system-Python invocation in the image with `SLACK_RELAY_URL` set. The confirmed hard breakage is the test suite.

Simplest fix is to revert the import move. If you want them at the top for readability, they belong inside `install()` (which only runs under Hermes) and `sitecustomize.py`'s allowlist needs `slack_bolt`/`slack_sdk` added — but reverting preserves the existing contract with less to get wrong.

---

### Significant

**3. The hardcoded OAuth scope list is unnecessary and actively misleading — `slack_relay_patch.py:189`**

I read bolt's `middleware/authorization/internals.py`. `_to_authorize_result` does:

```python
oauth_scopes: Optional[str] = auth_test_result.headers.get("x-oauth-scopes")
```

Plain `.get()`, plain string, no subscripting, and `None` flows into `AuthorizeResult` harmlessly. So `headers = {}` is *provably sufficient* to fix the `AttributeError` — the 200-character inline default buys nothing and asserts a scope set the app may not hold, which can mask a genuine misconfiguration. Drop it, or if you want an old-sidecar fallback, hoist it to a module constant beside `DEFAULT_MAX_FILE_BYTES` rather than burying it mid-function.

**4. Unfiltered header pass-through across the credential boundary — `credential_proxy.py:400-413`**

```python
if hasattr(response, "headers") and response.headers:
    result["__headers"] = dict(response.headers)
```

Bolt needs exactly one header, and the relay has no method allowlist (`api_call` rejects only `_`-prefixed names), so this forwards every header from any Slack method to the agent side.

Being fair about severity: this is defense-in-depth, not a demonstrated leak. The existing precedent in the same file (`AgentAPIProxyHandler`, lines 133-144) uses a hop-by-hop *denylist* plus CRLF sanitisation rather than an allowlist, so the new code isn't out of step with the file. But given the formal credential-isolation requirement in `docs/security-requirements.md` §6, narrowing costs nothing:

```python
WANTED = ("x-oauth-scopes", "x-accepted-oauth-scopes")
headers = {k: v for k, v in (response.headers or {}).items() if k.lower() in WANTED}
if headers:
    result["__headers"] = headers
```

**5. `_slack_error_detail` dumps whole API payloads into logs — `credential_proxy.py:251-277`**

`json.dumps(payload)` on `response.data`, untruncated, invoked from the generic `except Exception` at line 1294 that covers every relayed method. For `auth.test` that's `{"ok": false, "error": "invalid_auth"}`; for an arbitrary method it's whatever Slack returned. Suggest extracting the diagnostic fields only:

```python
return json.dumps({k: payload[k] for k in ("ok", "error", "needed", "provided") if k in payload})
```

**6. Workflow version comments regress a convention landed four commits ago**

`AGENTS.md` requires a 40-char SHA *"with the human-readable version in a trailing comment (`uses: actions/checkout@3d3c42e… # v7.0.1`)"*. This PR rewrites 8 comments to `# v7` / `# v3` in two files, while the rest of `.github/workflows/` keeps full semver (25× `v7.0.1`, 13× `v3.0.0`, 11× `v7.0.0`). Commit `db64962` established that convention. Please drop these hunks.

`persist-credentials: false` is a real improvement — I confirmed neither deploy workflow nor `provision_08` uses `git push` / `gh` / `GITHUB_TOKEN` after checkout, so it's safe — but it belongs in its own PR, applied to all 25 checkout sites rather than 2.

---

### Minor

**7. Sync response class returned from an async client — `slack_relay_patch.py:169, 197`**

`AsyncWebClient.api_call` returns `AsyncSlackResponse`; the PR imports the sync `SlackResponse`. `SlackResponse.__next__` calls `self._client._request_for_pagination(...)` synchronously, which `RemoteSlackClient` (an `AsyncWebClient` subclass) doesn't have — so pagination would `AttributeError`. Latent only: it bites if something ever paginates through this client. Importing `AsyncSlackResponse` is the cheap fix.

**8. Test-file cleanups**

- `self._saved_modules = {}` (`test_slack_relay_patch.py:241`) is never populated, so `tearDown`'s restore loop is a no-op. `setUp` mutates `slack_bolt.app.async_app.AsyncWebClient` and `PlatformRegistry.create_adapter` and never restores them — cross-suite pollution if these ever share a process.
- Dead duplicate `FakeSlackResponse` at line 159; only the one defined at line 26 is wired into `sys.modules`.
- The `__headers` branch in `credential_proxy.SlackRelay.api_call` — the actual fix — is untested: `FakeResponse` (`test_credential_proxy.py:685`) has no `.headers`, so `hasattr` is `False` and the branch never runs.

**9. PR hygiene**

- Title `fix for slack` isn't a Conventional Commit. Suggest `fix(slack): wrap relay responses in SlackResponse and harden bootstrap`.
- The body states `test_credential_proxy.py` gained assertions that `api_call("auth.test")` returns a `SlackResponse` with `x-oauth-scopes`. It didn't — the only new test there is `test_slack_error_detail_serializes_response_to_json`.
- New env var `SLACK_RELAY_BOOTSTRAP_WAIT_SECONDS` is undocumented; per `AGENTS.md` please run the `review-docs-drift` skill. It also inlines its `"20"` default instead of following the `DEFAULT_MAX_FILE_BYTES` constant pattern next to it.
- Please describe the bolt module rebinding in the body — reviewers can't evaluate a process-wide monkeypatch that isn't mentioned.

**10. No CI coverage for these tests.** `validate.yml` runs `make validate`, which only checks for skills in invalid paths; nothing invokes `unittest` on `agents/platform/scripts/`. That's why the hanging test reached review. Worth a follow-up PR adding a Python unit-test job.

---

### Recommendation

**Request changes**, on items 1 and 2 — both reproduced locally. The core `SlackResponse` fix is sound and should land once the retry loop terminates and the imports move back inside `install()`. Item 6 should be split into its own PR.

---

mplakhtiy (approved):

PR #496, opened 2026-07-31
feat(security): default the GKE Sandbox (gVisor) runtime class on (RUNTIME-002)

AntonTyb:
# Pull Request

## Why This Change

Autonomous AI agent workloads in Kubernetes execute dynamic tasks driven by Large Language Models and user prompts. The Platform Agent pod runs today on the standard Linux kernel with no sandbox, so a container breakout following prompt injection or malicious code execution reaches the node directly. GKE Sandbox (gVisor) interposes a user-space kernel; the operator already supports `runtimeClassName`, but nothing sets it, so defense-in-depth is opt-in and in practice unused.

This addresses security finding **RUNTIME-002** (Buganizer Task 537148050).

**Scope narrowed.** This PR originally carried all three findings from that task. The other two have since been split out:

- **RUNTIME-007** (read-only root filesystem) landed separately in **#674**, rebuilt on current `main`. It was removed from this branch rather than left duplicated: this branch's copy predates a review finding that the `tmp-scratch` `emptyDir` needs a `sizeLimit`, so keeping both would let merge order decide whether the volume ends up bounded.
- **RUNTIME-006** (administrative CLIs confined to the `credential-proxy` stage) was a verification of existing behaviour with no code change, so there is nothing here to merge for it.

## What Changed

**Files:**

- `.agents/skills/install-kube-agents/SKILL.md`
- `.github/workflows/agent-startup-test.yml`
- `README.md`
- `charts/kube-agents/README.md`
- `docs/site/src/content/docs/install/quickstart-gke.mdx`
- `docs/site/src/content/docs/install/uninstall.md`
- `docs/site/src/content/docs/operator/provisioning-scripts.md`
- `docs/site/src/content/docs/overview/architecture.mdx`
- `install.sh`
- `k8s-operator/scripts/platform-agent.yaml.template`
- `k8s-operator/scripts/provision_02_gvisor_nodepool.sh`
- `k8s-operator/scripts/provision_08_deploy_platform_agent.sh`
- `k8s-operator/scripts/teardown_02_gvisor_nodepool.sh`
- `tests/test_gvisor_default.py`

**Added/Changed/Fixed:**

- `DEFAULT_ENABLE_GVISOR="true"` in `common.sh` is now the single home for the default. `provision_02`, `provision_08` and `install.sh` all read it rather than spelling it themselves.
- `platform-agent.yaml.template` ships `runtimeClassName: gvisor` enabled; `provision_08` comments the field out only on an explicit `ENABLE_GVISOR=false`.
- `provision_08` asks the API server whether the cluster actually has the `gvisor` RuntimeClass before applying, because the operator refuses to reconcile a CR naming one it cannot find. An explicit `ENABLE_GVISOR=true` fails loudly; a merely defaulted one falls back to deploying unsandboxed with a warning; a probe that cannot be read at all (expired credential, RBAC denial, throttling, 5xx, no connection) stops the step rather than guessing.
- The defaulted value is deliberately **not** persisted to `vars.sh`, so a fallback on one run cannot come back as an explicit request on the next.
- `provision_02` skips itself on Autopilot, which rejects user-created node pools and needs none for GKE Sandbox.
- **Teardown keeps the opposite default** (`ENABLE_GVISOR:-false`): a pool nobody asked for is cheap to delete, a pool somebody wanted is not cheap to lose.
- `tests/test_gvisor_default.py` (new) pins all of the above. There is no shell test harness in this repository and the Go suite does not reach `k8s-operator/scripts/`, so without it a revert here is silent.
- Docs: the quickstart, `architecture.mdx`, `uninstall.md`, `provisioning-scripts.md`, the install skill and the two READMEs.

**The declarative install path deliberately does not change.** `charts/kube-agents/values.yaml` still defaults `platformAgent.deployment.availability.runtimeClassName` to `""`, because the sandbox node pool `provision_02` creates has no Terraform or chart-side equivalent. Every claim about the default in this PR therefore names the script install path, and `charts/kube-agents/README.md` now lists `runtimeClassName` alongside `dashboardEnabled` as a field the two paths disagree on. `make iac-parity-check` passes; this is the structural kind of drift it does not cover.

## Why This Matters

Eliminates unsandboxed default execution on the shared host kernel for the workload most exposed to untrusted input, while leaving `ENABLE_GVISOR=false` as the opt-out for clusters that cannot run a sandbox node pool.

## Context

Buganizer Task 537148050.

Two review rounds have run on this branch; see **Self-Review** below for what each found and where it landed.

## Self-Review

This branch has been read adversarially twice, and both passes found real defects. Angles covered across the two: idempotency of a re-run, fail-open vs fail-closed error handling, cluster-shape coverage (Standard vs Autopilot), install-surface parity (scripts vs Helm/Terraform), dead code, teardown asymmetry, and whether the docs claim more than the code delivers.

**Round 1 — eight findings (six inline, two outside the diff), all closed in `8986eb47`.**

- 🔴 The redeploy workflows run `provision_08` alone and would have requested a RuntimeClass their clusters lack, producing a green pipeline that deployed nothing. Closed by the RuntimeClass probe.
- 🔴 `install.sh` — the front door the quickstart sends people to — still defaulted gVisor off, and its answer is written into `vars.sh`, which `load_state` sources ahead of every script default. It actively defeated the flip. Closed by resolving `--gvisor` from `DEFAULT_ENABLE_GVISOR`, reordering the interactive prompt so the enter-key answer enables the sandbox, and updating `install-kube-agents/SKILL.md`.
- 🟠 The quickstart flowchart still labelled step 02 "(opt-in)", and three further doc surfaces called it optional — one of them generated from a script header this PR had not touched. Closed by editing the sources and regenerating.
- 🟠 The Helm chart had no way to express `runtimeClassName` at all. The chart has since gained `platformAgent.deployment.availability.runtimeClassName`, so the field is now reachable there; it still defaults to `""`, which is the deliberate divergence documented above and in the chart README.
- 🟡 A dead `if` arm that logged as though it had acted, and nothing pinning the template default. Both closed; `tests/test_gvisor_default.py` now runs the script's own `sed` so the opt-out cannot drift from the template.
- **One finding was rejected with a reason rather than fixed:** teardown deliberately keeps `ENABLE_GVISOR:-false`. Sharing the provisioning default would let `teardown.sh --no-confirm`, run from a clone with no `vars.sh`, delete a `gvisor-pool` it does not know it owns. Provision and teardown want opposite defaults, and the asymmetry is now asserted by a test so it reads as deliberate rather than as an oversight.

**Round 2 — four findings, all fixed in the two commits above.**

- 🔴 `provision_08` persisted the value it defaulted, so a second run of the identical command read its own fallback back as an explicit request and exited 1 — on a cluster and a command that had not changed. Reproduced against the real `common.sh`, then fixed by resolving the default without persisting it.
- 🟠 The RuntimeClass probe failed open on every `kubectl` error, not just NotFound.
- 🟠 Step 02 aborted the pipeline at step 2 of 13 on Autopilot.
- 🟠 `README.md` and `architecture.mdx` promised the default on the Helm/Terraform path, which does not have it, and the chart README's "the one field where the two install paths disagree" became false.

**Still open, and not addressed here:** whether the autopush and staging clusters have a gVisor node pool. With no pool they now redeploy unsandboxed behind a warning on a green build, which is a fleet question rather than a code one. Also unresolved: `pull-kube-agents-smoke-test` is failing with an unreadable log (403 on the GCS object, the Prow `/log` endpoint 404s). It is `Required: false` and it fails on other PRs too, so it is not obviously specific to this diff, but nobody has read it.

## Testing

- `python3 -m unittest discover -s tests -p 'test_*.py'` — 143 tests pass, including the 19 in `tests/test_gvisor_default.py`. The new ones run the scripts' own code rather than a copy of it: the resolution block against the real `common.sh`, the probe against a stubbed `kubectl` returning NotFound / Forbidden / Unauthorized / TooManyRequests / InternalError / unreachable, and the Autopilot predicate against a stubbed `gcloud`. Each was confirmed to fail against the pre-fix scripts before being kept.
- `go build ./...` in `k8s-operator/` is clean. The Go surface is untouched by this branch, so the operator suite was not re-run for the round-2 commits.
- `make docs-check` passes — generated regions current, links resolve, terminology matches, doc map coverage complete.
- `make iac-parity-check` — 15 checks pass across scripts, Terraform, and the Helm chart.
- `bash -n` clean on every changed script. `shellcheck` was not run: it is not installed in the environment this was written in.

### Live validation

**Not live-tested.** Exercising this needs a GKE cluster in both shapes it now branches on — Standard without a gVisor node pool, to see the probe fall back and the second run *not* hard-fail, and Autopilot, to see step 02 skip itself instead of dying — and neither was available here. The round-2 High was reproduced instead against the real `common.sh` with a temporary `VARS_FILE` and `CI=true`, which exercises the actual defect (`vars.sh` state surviving between runs) end to end:

```
run 1:   ENABLE_GVISOR=true GVISOR_EXPLICIT=0
vars.sh: export ENABLE_GVISOR=true
run 2:   ENABLE_GVISOR=true GVISOR_EXPLICIT=1     <- before the fix
run 2:   ENABLE_GVISOR=true GVISOR_EXPLICIT=0     <- after
```

What that does **not** cover: whether the operator reconciles a real CR carrying `runtimeClassName: gvisor` onto a real sandbox pool, and whether Autopilot accepts the CR the skip leaves behind. Both need a cluster.

---

_This PR was updated with the help of Claude._


mplakhtiy:
## Findings

### 1. Agent-only redeploys will request a RuntimeClass their cluster may not have, and the pipeline reports success while deploying nothing — High

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:36`,
`k8s-operator/scripts/platform-agent.yaml.template:28`

`.github/workflows/reusable-deploy-agent.yml:103-112` redeploys the agent by running **only**
`provision_08` — it never runs `provision_02`, which is what creates the gVisor node pool.
`autopush-redeploy-agent.yml` and `staging-redeploy-agent.yml` both call it.

The runner starts from a fresh checkout, and `k8s-operator/scripts/vars.sh` is not tracked, so
`load_state` (`common.sh:211-217`) creates an empty state file. `ENABLE_GVISOR` is not in the
workflow's env either, so `init_var "ENABLE_GVISOR" "true"` falls through to
`common.sh:138-139` — `is_ci_pipeline` is true under GitHub Actions — and takes this PR's new
default, `true`. The template ships `runtimeClassName: gvisor` uncommented, so the CR that gets
applied requests the gVisor RuntimeClass.

The operator then refuses to reconcile the workload
(`k8s-operator/internal/controller/platformagent_controller.go:138-150`):

```go
if err := r.validateRuntimeClass(ctx, instance); err != nil {
    if errors.IsNotFound(err) {
        …
        return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
    }
```

`reconcileWorkload` is at line 153, below that early return, so the Deployment is never touched —
it keeps the previously deployed image.

**Failure scenario.** This merges; the next push to `main` triggers `autopush-redeploy-agent`. The
autopush cluster was provisioned while `ENABLE_GVISOR` defaulted to `false`, so it has no
`gvisor-pool` and GKE has therefore never created the `gvisor` RuntimeClass. `provision_08` applies
the CR successfully, the operator flips the `PlatformAgent` to `Degraded/RuntimeClassNotFound` and
requeues every 30s, and the workflow's next step, `kubectl rollout status
deployment/platform-agent-gateway --timeout=180s`, inspects a Deployment that was never modified and
is already healthy — so it returns success immediately. The pipeline is green, the new image is not
running, and the only signal is a status condition on a CR nobody is watching.

The condition for the failure is that the target cluster has no gVisor node pool. Before this PR the
default was `false`, so that is the state of every environment where nobody opted in. The fix is
either to pin `ENABLE_GVISOR` explicitly in the agent-redeploy workflows, or to have `provision_08`
verify the `gvisor` RuntimeClass exists before it renders `runtimeClassName` (it already knows the
cluster — it connects to it in step 1).

**Verdict:** Confirmed on mechanism; the trigger depends on whether the autopush and staging
clusters currently have a gVisor pool, which is worth checking before merge either way.

### 2. The quickstart flowchart still labels step 02 "(opt-in)", contradicting the flag list eighteen lines below — Medium

`docs/site/src/content/docs/install/quickstart-gke.mdx:65`

```
    A --> C["02: gVisor node pool (opt-in)"]
```

Line 82 in the same file now reads `` `ENABLE_GVISOR=true` (default) — provisions the gVisor sandbox
node pool in step 2 ``. The two statements are directly contradictory, and the flowchart is the
false one: `provision_02_gvisor_nodepool.sh:29` defaults `ENABLE_GVISOR` to `"true"` and
`provision.sh:23` runs that script unconditionally, so a plain `./provision.sh` now creates the pool.
The sibling page was updated consistently — both gVisor mentions in `architecture.mdx` were
corrected — which makes this a missed occurrence rather than a distinction.

**Failure scenario.** A reader following the quickstart diagram concludes no sandbox node pool will
be created unless they opt in, runs `./provision.sh`, and gets a billable regional node pool
(`--machine-type=e2-standard-4 --num-nodes=1` against a regional cluster is one node per zone) they
did not plan for.

AGENTS.md makes this a pre-merge gate rather than a matter of taste: under Pull Request Hygiene it
requires a "**Docs-drift review before opening a PR:** run the `review-docs-drift` skill
(`.agents/skills/review-docs-drift/SKILL.md`) against your branch diff and address its Blocking
findings", and that skill classifies "a doc now states something false" as Blocking.

**Verdict:** Confirmed.

### 3. The new `/tmp` emptyDir is unbounded, unlike every other scratch emptyDir the operator creates — Medium

`k8s-operator/internal/controller/platformagent_manifests.go:1582-1587`

```go
{
    Name: "tmp-scratch",
    VolumeSource: corev1.VolumeSource{
        EmptyDir: &corev1.EmptyDirVolumeSource{},
    },
},
```

Every other scratch volume this same function family creates carries a `SizeLimit`
(`platformagent_manifests.go:1231-1234`): `credential-proxy-tmp` 2Gi, `credential-proxy-state` 5Gi,
`credential-proxy-runtime` 16Mi, `event-watcher-kubeconfig` 1Mi. The one this PR adds does not, and
it is now the **only** writable path outside the PVC for the two containers that host the agent
process, so all scratch traffic that used to be spread across the container's writable layer
concentrates here.

**Failure scenario.** An LLM-driven task writes a large artifact to `/tmp` — a pulled chart bundle, a
`kubectl logs` dump, a downloaded archive. Nothing bounds it, so it consumes node ephemeral storage
until the kubelet evicts the pod under disk pressure. That is a whole-pod restart, including the
credential proxy and the log shipper, where a `sizeLimit` would have produced an `ENOSPC` in the one
container that overreached. A limit in the low single-digit Gi, matching `credential-proxy-tmp`,
would keep the blast radius local.

**Verdict:** Confirmed. Not a regression — writes to `/tmp` were equally unbounded when they landed
in the container layer — but the PR is introducing the volume, and the repo's own convention for
exactly this kind of volume is to bound it.

### 4. The `ENABLE_GVISOR=true` branch is now a no-op that logs as though it did something — Low

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:131-133`

```sh
if is_truthy "$ENABLE_GVISOR"; then
    print_info "Enabling gVisor runtimeClassName in '$CR_MANIFEST'..."
    sed -i.bak 's/# runtimeClassName: gvisor/runtimeClassName: gvisor/g' "$CR_MANIFEST" && rm -f "${CR_MANIFEST}.bak"
```

The template line this pattern used to match is gone. It now reads
`      runtimeClassName: gvisor # Enabled by default for GKE Sandbox (gVisor) kernel isolation`,
which contains no `# runtimeClassName: gvisor` substring. I ran the substitution against the new
template and it changes nothing.

**Failure scenario.** The enable path only works because the template happens to ship the line
uncommented; the `sed` contributes nothing. Anyone editing the template back toward a commented
default, or reindenting that block, gets a silent no-op with a log line that positively asserts
"Enabling gVisor runtimeClassName" — the operator reading the provisioning output has no way to tell
the manifest was not modified. Dropping the `if` branch and keeping only the `else` (disable) arm
says what actually happens.

**Verdict:** Confirmed.

### 5. Nothing exercises the rendered CR, so the disable path is one template edit away from silently breaking — Low

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:134-136`

The Go tests this PR adds are good on the operator side — they assert
`ReadOnlyRootFilesystem` on all three containers and the `tmp-scratch` mount and volume, and the
golden manifest is regenerated. But the other half of the change, the rendered `PlatformAgent` CR, is
covered by nothing: there is no shell test harness in this repository (no `*.bats`, no
`scripts/tests`), and no Go test renders the template.

The new opt-out arm is coupled to the exact shape of one template line:

```sh
sed -i.bak 's/^[[:space:]]*runtimeClassName: gvisor/# runtimeClassName: gvisor/g' "$CR_MANIFEST"
```

**Failure scenario.** Someone renames the field, moves it under a different key, or appends a
`${...}` substitution to that line. The regex stops matching, `ENABLE_GVISOR=false` stops disabling
anything, and the CR goes out requesting gVisor on a cluster that was deliberately configured not to
have it — landing in exactly the `Degraded/RuntimeClassNotFound` state of Finding 1. Nothing in CI
would catch it. The cheapest coverage is a Go test that runs `envsubst` plus both `sed` arms over the
template and asserts the two expected outputs.

**Verdict:** Confirmed — the behaviour is untested; the regression would ship silently.

### 6. The PR title describes a design proposal and is not a Conventional Commit — Low

Title: `Design Proposal: Workload Hardening & Sandbox Default Enforcement (Task 537148050)`

The branch history shows the design document was added and then deliberately removed —
`9393357c docs(design): remove design doc for task 537148050 before implementation` — and what
remains is implementation: script defaults, operator container security contexts, a new volume, and
docs. The title still says "Design Proposal".

AGENTS.md, Pull Request Hygiene: "Use Conventional Commits for commit messages."

**Failure scenario.** A squash merge takes the PR title as the commit subject, so `main`'s history
gets a non-conforming subject that describes a proposal for a commit that changes the default
runtime isolation of every new install. Anyone later bisecting for when gVisor became the default
skips right past it. `feat(security): default the gVisor sandbox on and enforce a read-only root
filesystem` would say what landed.

**Verdict:** Confirmed.


---

AntonTyb:
Narrowed the scope of this PR: the read-only root filesystem work (RUNTIME-007) has been removed from the branch and landed separately in #674, rebuilt on current `main`.

It was removed rather than left in place because a duplicate is worse than a rebase here. This branch's copy of the `tmp-scratch` `emptyDir` predates the review finding on this PR that it needs a `sizeLimit`; #674 has that fix. With both open, merge order would have decided whether the volume ends up bounded, silently.

RUNTIME-006 was a verification with no code change, so nothing here corresponds to it.

What is left is RUNTIME-002 alone — the gVisor default flip — which is the half the open High finding is against. Title and description updated to match. Still not ready to merge: the redeploy workflow gap and `install.sh` (#519, which landed after this branch was written) both need addressing first.

_Posted with the help of Claude._

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
```diff
@@ -38,7 +38,7 @@ DEFAULT_PROJECT_ID="${ACTIVE_PROJECT:-$(whoami 2>/dev/null || echo "user")}"
 init_var "PROJECT_ID" "$DEFAULT_PROJECT_ID" "Enter Target GCP Project ID"
 init_var "REGION" "$DEFAULT_REGION" "Enter GKE GCP Region"
 init_var "CLUSTER_NAME" "$DEFAULT_CLUSTER_NAME" "Enter GKE Cluster Name"
-init_var "ENABLE_GVISOR" "false" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
+init_var "ENABLE_GVISOR" "true" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
```
mplakhtiy: **HIGH — the autopush and staging redeploy workflows will fail on every push to `main`.**

`.github/workflows/reusable-deploy-agent.yml:103-112` redeploys the agent by running **only** `provision_08` — it never runs `provision_02`, the script that creates the gVisor node pool — and it passes no `ENABLE_GVISOR` in either the job `env` or the step `env`. `autopush-redeploy-agent.yml:31` and `staging-redeploy-agent.yml:19` both call it.

The runner starts from a fresh checkout and `k8s-operator/scripts/vars.sh` is untracked, so `load_state` (`common.sh:417-424`) creates an empty state file. With nothing to source and nothing in the environment, this line takes the new `true` default — `is_non_interactive` is true under Actions via `is_ci_pipeline` (`common.sh:389-390`). The template now ships the field uncommented, so the applied CR requests the `gvisor` RuntimeClass.

The operator refuses to reconcile the workload (`platformagent_controller.go:239-250`) because `validateRuntimeClass` (`:1083-1095`) does a plain `Get` on the RuntimeClass and returns `IsNotFound`; the reconcile early-returns with a 30s requeue, above `reconcileWorkload`. The Deployment is never touched.

`provision_08` then catches this, loudly. Since #533 it records both generations before applying and fails when the CR generation moved but the Deployment generation did not (`provision_08:197-215`):

```sh
    if [ "$current_generation" = "$prev_deploy_generation" ] || [ -z "$current_generation" ]; then
      print_error "Operator did not reconcile the changed PlatformAgent spec into deployment/${deploy_name} within ${AGENT_READY_TIMEOUT}."
```

`run_step` turns that `return 1` into `exit 1`.

**Failure scenario.** This merges; the next push to `main` triggers `autopush-redeploy-agent`. The autopush cluster was provisioned while `ENABLE_GVISOR` defaulted to `false`, so it has no `gvisor-pool` and GKE has therefore never created the `gvisor` RuntimeClass. `provision_08` applies the CR, the operator flips the `PlatformAgent` to `Degraded/RuntimeClassNotFound`, the script spins for the full `AGENT_READY_TIMEOUT` (600s default, `common.sh:699`) and exits 1. Continuous deployment to autopush and staging is dead until someone adds a gVisor node pool to both clusters or pins the variable — and every push burns ten minutes of runner time rediscovering it.

Note that `rc-deploy-environment.yml:71,111` *does* pass `ENABLE_GVISOR: ${{ vars.ENABLE_GVISOR }}`, and its `provision_rc_environment.sh` runs the full `provision.sh` so `provision_02` runs first. The redeploy workflows are the only path that applies the CR without ever having a chance to create the pool.

**Fix.** Set `ENABLE_GVISOR` explicitly in the redeploy workflows' env, or have `provision_08` probe for the `gvisor` RuntimeClass before rendering the field — it is already connected to the cluster by step 1.

**Worth answering in the PR body either way:** do the autopush and staging clusters currently have a gVisor node pool? One `gcloud container node-pools list` against each decides whether this fires on merge or lies dormant.

docs/site/src/content/docs/install/quickstart-gke.mdx
```diff
@@ -108,7 +108,7 @@ See [Provisioning scripts](/kube-agents/operator/provisioning-scripts/) for the
 ## Common flags and toggles
 
 - `--dry-run` — prints what would be done without applying anything.
-- `ENABLE_GVISOR=true` — provisions the gVisor sandbox node pool in step 2.
+- `ENABLE_GVISOR=true` (default) — provisions the gVisor sandbox node pool in step 2 (set `ENABLE_GVISOR=false` to opt out).
```
mplakhtiy, line 111: **MEDIUM — the flowchart seventeen lines above still labels step 02 "(opt-in)", directly contradicting this line.**

`quickstart-gke.mdx:94`:

```
    A --> C["02: gVisor node pool (opt-in)"]
```

The flowchart is the false one: `provision_02_gvisor_nodepool.sh:29` now defaults `ENABLE_GVISOR` to `"true"` and `provision.sh:28` runs that script unconditionally, so a plain `./provision.sh` creates the pool. The sibling page was updated consistently — both gVisor mentions in `architecture.mdx` were corrected — which makes this a missed occurrence rather than a deliberate distinction.

**Failure scenario.** A reader following the diagram concludes no sandbox node pool will be created unless they opt in, runs `./provision.sh`, and gets a billable node pool they did not plan for: `provision_02:55-64` creates `e2-standard-4 --num-nodes=1` against a cluster that `provision_01:157-158` created with `--location "$REGION"` — i.e. regional, so one node per zone, three `e2-standard-4` nodes.

AGENTS.md makes this a pre-merge gate rather than a matter of taste: under Pull Request Hygiene it requires a "**Docs-drift review before opening a PR:** run the `review-docs-drift` skill … and address its Blocking findings", and that skill classifies "a doc now states something false" as Blocking. `kube-agents-bot` raised this on 2026-07-31 and it is still open.

k8s-operator/scripts/provision_02_gvisor_nodepool.sh
```diff
@@ -26,7 +26,7 @@ check_prereqs "gcloud" "kubectl"
 print_step "Setting up Configuration State"
 load_state
 
-init_var "ENABLE_GVISOR" "false" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
+init_var "ENABLE_GVISOR" "true" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
```
mplakhtiy: **MEDIUM — three more doc surfaces still call this step optional, one of them generated from a header this PR did not touch.**

The script header comments are unchanged:

```sh
# 🤖 Step 2: Optional Dedicated gVisor Node Pool Initialization      # provision_02:3
# 🧹 Step 2: Optional Teardown of Dedicated gVisor Node Pool         # teardown_02:3
```

`scripts/generate_docs.py` lifts those headers verbatim into the `provisioning-steps` generated region, so `k8s-operator/scripts/README.md:109` and `:128` now describe the default step as "**Optional Dedicated gVisor Node Pool Initialization**". `make docs-check` passes precisely because the generated text still matches its source — the check verifies currency, not truth.

Two hand-written pages say the same thing:

- `docs/site/src/content/docs/operator/provisioning-scripts.md:8` — "a set of idempotent step scripts (plus their teardown mirrors and an optional gVisor step)"; `:14` — "skipping opt-in steps unless enabled"; `:21` — "opt-in steps no-op unless their flag is set".
- `docs/site/src/content/docs/install/uninstall.md:73` — "gVisor node pool only (optional)".

**Failure scenario.** An operator reading `k8s-operator/scripts/README.md` — which AGENTS.md names as the canonical home for "what each provisioning script does" — plans a run on the understanding that step 02 no-ops by default, and is billed for a three-node regional pool.

Per AGENTS.md, edit the two script headers and then run `make docs-generate`; hand-editing the generated region will fail CI.

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
```diff
@@ -154,6 +154,9 @@ execute_custom_resource() {
   if is_truthy "$ENABLE_GVISOR"; then
     print_info "Enabling gVisor runtimeClassName in '$CR_MANIFEST'..."
     sed -i.bak 's/# runtimeClassName: gvisor/runtimeClassName: gvisor/g' "$CR_MANIFEST" && rm -f "${CR_MANIFEST}.bak"
+  else
+    print_info "Disabling gVisor runtimeClassName in '$CR_MANIFEST' (ENABLE_GVISOR=false)..."
+    sed -i.bak 's/^[[:space:]]*runtimeClassName: gvisor/# runtimeClassName: gvisor/g' "$CR_MANIFEST" && rm -f "${CR_MANIFEST}.bak"
```
mplakhtiy: **LOW — this `if` arm is a no-op that logs as though it acted.**

The template line this pattern used to match is gone. It now reads `      runtimeClassName: gvisor # Enabled by default for GKE Sandbox (gVisor) kernel isolation`, which contains no `# runtimeClassName: gvisor` substring. I ran the substitution against the new template: the file is byte-identical afterwards.

**Failure scenario.** The enable path works only because the template happens to ship the line uncommented; the `sed` contributes nothing. Anyone editing the template back toward a commented default, or reindenting that block, gets a silent no-op behind a log line that positively asserts "Enabling gVisor runtimeClassName" — the operator reading the provisioning output has no way to tell the manifest was not modified.

Keeping only the disable arm says what actually happens:

```suggestion
  if ! is_truthy "$ENABLE_GVISOR"; then
    print_info "Disabling gVisor runtimeClassName in '$CR_MANIFEST' (ENABLE_GVISOR=false)..."
    sed -i.bak 's/^[[:space:]]*runtimeClassName: gvisor/# runtimeClassName: gvisor/g' "$CR_MANIFEST" && rm -f "${CR_MANIFEST}.bak"
```

k8s-operator/scripts/teardown_02_gvisor_nodepool.sh
```diff
@@ -17,8 +17,8 @@ source "${SCRIPT_DIR}/common.sh" "$@"
 # ─── Configuration State Restoration ──────────────────────────────────────────
 ensure_teardown_state
 
-if ! is_truthy "${ENABLE_GVISOR:-false}"; then
-  print_info "Skipping gVisor node pool teardown (ENABLE_GVISOR=${ENABLE_GVISOR:-false})."
+if ! is_truthy "${ENABLE_GVISOR:-true}"; then
+  print_info "Skipping gVisor node pool teardown (ENABLE_GVISOR=${ENABLE_GVISOR:-true})."
```
mplakhtiy: **LOW — a stateless uninstall now deletes the gVisor node pool without asking.**

`ensure_teardown_state` only populates `ENABLE_GVISOR` when `vars.sh` exists; its no-state branch never sets it. So with no state file this guard now falls through where it previously exited 0. Execution reaches `:26-34`, which defaults `GVISOR_POOL_NAME` to `gvisor-pool`, and then `:47-52`:

```sh
    if [ "$NO_CONFIRM" -ne 1 ]; then
      ...read -r -n 1 REMOVE_GVISOR...
    else
      REMOVE_GVISOR="y"
```

**Failure scenario.** Someone runs `teardown.sh --no-confirm` — which is what `provision_rc_environment.sh` and the uninstall skill pass — from a clone without `vars.sh`, targeting a cluster that happens to have a node pool named `gvisor-pool`. Before this PR the script printed "Skipping" and exited. After it, the pool is deleted with no prompt.

The default flip is defensible on the provision side, where the cost of being wrong is a node pool you did not want. On the teardown side the cost of being wrong is a node pool you did want. These two arguably want opposite defaults, and the PR gives them the same one:

```suggestion
if ! is_truthy "${ENABLE_GVISOR:-false}"; then
  print_info "Skipping gVisor node pool teardown (ENABLE_GVISOR=${ENABLE_GVISOR:-false})."
```

k8s-operator/scripts/platform-agent.yaml.template
```diff
@@ -25,7 +25,7 @@ spec:
     imagePullPolicy: Always
     availability:
       replicas: 1
-      # runtimeClassName: gvisor # Uncomment to enable GKE Sandbox (gVisor) kernel isolation
+      runtimeClassName: gvisor # Enabled by default for GKE Sandbox (gVisor) kernel isolation
```
mplakhtiy, line 28: **LOW — nothing pins this default, so a silent revert would ship.**

No test renders the CR template or exercises either `sed` arm in `provision_08:154-159`. There is no shell test harness in this repository (no `*.bats`, no `scripts/tests`), the Go suite does not touch `k8s-operator/scripts/`, and the operator-side tests only cover a `RuntimeClassName` the test fixture sets itself.

The opt-out arm is coupled to the exact shape of this one line:

```sh
    sed -i.bak 's/^[[:space:]]*runtimeClassName: gvisor/# runtimeClassName: gvisor/g' "$CR_MANIFEST"
```

**Failure scenario.** Someone renames the field, moves it under a different key, or appends a `${...}` substitution to this line. The regex stops matching, `ENABLE_GVISOR=false` stops disabling anything, and the CR goes out requesting gVisor on a cluster deliberately configured not to have it — landing in exactly the `Degraded/RuntimeClassNotFound` state described on `provision_08:41`. Nothing in CI would catch it.

The cheapest coverage is a Go test that runs `envsubst` plus both arms over this template and asserts the two expected outputs. It would have caught the dead `if` arm as well.

mplakhtiy (commented):
Review of `5273d08`. Six findings inline; two more below that anchor to files outside this diff.

## Findings outside this diff

### HIGH — `install.sh` still defaults gVisor off, and its saved state outranks the new script default

`install.sh:72`, `:109`, `:803`, `:1392`, `:1445-1451`, `:1546`

The quickstart puts the one-liner installer first — "Run the Zero-Friction One-Liner Installer", `quickstart-gke.mdx:22-30` — ahead of the manual `./provision.sh` workflow. That installer landed in #519, after this branch was written, and it was not updated:

```sh
PARAM_ENABLE_GVISOR="${ENABLE_GVISOR:-false}"            # :72
  --gvisor=true|false           Enable GKE Sandbox (gVisor) runtime isolation (default: false)   # :109
  local enable_gvisor="${PARAM_ENABLE_GVISOR:-false}"    # :1392
```

The interactive prompt agrees — `prompt_menu "Enable GKE Sandbox (gVisor) Runtime Isolation for Agent Workloads?" "No - Standard Container Runtime (Default)" ...` (`:1445-1447`), with `enable_gvisor` set to `true` only on choice 2.

This is not merely a second default that disagrees; it actively defeats the one this PR changes. `install.sh:1546` writes the answer into the state file with `write_state_var "$vars_file" ENABLE_GVISOR "$enable_gvisor"` (`:299-304`, emitting `export ENABLE_GVISOR=false`) at `${repo_dir}/k8s-operator/scripts/vars.sh` (`:1521`). `load_state` sources that file, so by the time `provision_02` and `provision_08` reach `init_var "ENABLE_GVISOR" "true"`, `current_val` is already `false` and the new default is never consulted. Saved state wins, as `load_state`'s own comment says for every knob.

**Failure scenario.** A user follows the quickstart, runs the one-liner, and reads `quickstart-gke.mdx:111` — "`ENABLE_GVISOR=true` (default)", a line this PR adds. They get an install with no sandbox node pool and a Platform Agent on the host kernel, and the page told them otherwise. RUNTIME-002 is recorded as fixed while the most common install path is unchanged.

`.agents/skills/install-kube-agents/SKILL.md:120` documents the same `false` default and needs the same edit.

### MEDIUM — the Helm chart cannot express this default at all, so the Terraform path never gets the sandbox

`charts/kube-agents/templates/platform-agent-cr.yaml:19-22`, `charts/kube-agents/values.yaml`

The chart renders `spec.deployment` as image, tag and pull policy only. There is no `availability` block and no `runtimeClassName`, and `values.yaml` has no key that could set one — `grep -rn "runtimeClassName" charts/` matches exactly one line, the CRD copy at `charts/kube-agents/crds/kubeagents.x-k8s.io_platformagents.yaml:1030`. The field is in the schema the chart ships and absent from the CR the chart renders.

`quickstart-gke.mdx:10` presents Terraform + Helm as a peer of the script pipeline — "both create equivalent GCP resources" — and `terraform/examples/full-install/` is the single-apply composition that installs the chart. That path is now strictly less hardened than the script path, with nothing in the chart a user could set to close the gap.

This is an altitude question as well as a parity one. The default is expressed in three places at the lowest layer (two `init_var` calls and a template line), in none at the chart layer, and in none at the CRD layer — `k8s-operator/api/v1alpha1/common_types.go:404-406` declares `RuntimeClassName *string` with `+optional` and no `+kubebuilder:default`. A CRD default would reach every surface at once, including the chart, and would remove the `sed` machinery entirely.

**Failure scenario.** A team follows the declarative path, reads `architecture.mdx:110` ("enabled by default" — a line this PR adds), and ships an unsandboxed agent. Worth noting the Terraform module builds an Autopilot cluster (`terraform/modules/gke-cluster/main.tf:36 enable_autopilot = true`), where the operator's own diagnostic says gVisor needs no node pool — "In GKE Autopilot, gVisor is supported automatically" (`platformagent_controller.go:243`) — so this is the one surface where the default would have been free to turn on.

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
AntonTyb: Fixed in 8986eb47.

You had the mechanism exactly right, so I fixed it a layer up from the workflows rather than in them. `provision_08` now asks the API server the same question `validateRuntimeClass` will, before it applies anything:

```sh
if is_truthy "$ENABLE_GVISOR" && ! kubectl get runtimeclass gvisor >/dev/null 2>&1; then
```

Setting `ENABLE_GVISOR=false` in `autopush-deploy.yml` and `staging-deploy.yml` would have unblocked those two, but every other caller of step 08 against a cluster with no sandbox pool has the same problem — including anyone redeploying an install that predates this default. The probe repairs all of them.

What it does with the answer depends on who chose the value, which is the part worth reviewing. `GVISOR_EXPLICIT` is set before `init_var` runs, so it records whether the value came from the environment or `vars.sh` (deliberate) or from `DEFAULT_ENABLE_GVISOR` (not). An explicit `ENABLE_GVISOR=true` against a cluster with no RuntimeClass is a requirement that cannot be met, so it fails with the remediation; a defaulted one is a preference, so it warns and deploys without the sandbox. Silently downgrading an explicit request would be the worst possible response to it.

`load_state` runs before this, so `-n "${ENABLE_GVISOR:-}"` also treats an empty value as not-explicit — which is what `ENABLE_GVISOR: ${{ vars.ENABLE_GVISOR }}` in `rc-deploy-environment.yml` expands to when the repo variable is unset. That matches `init_var`, which uses `[ -z "$current_val" ]` and so takes the default in the same case.

docs/site/src/content/docs/install/quickstart-gke.mdx
AntonTyb: Fixed in 8986eb47 — the flowchart node is now `A --> C["02: gVisor node pool (on by default)"]`, so it agrees with the line below it.

k8s-operator/scripts/provision_02_gvisor_nodepool.sh
AntonTyb: Fixed in 8986eb47. All four surfaces, and the generated one at the source:

- `provision_02_gvisor_nodepool.sh` — header is now `Step 2: Dedicated gVisor Node Pool Initialization (on by default)`, with a body line `Runs by default; set ENABLE_GVISOR=false to skip it.`
- `k8s-operator/scripts/README.md` — regenerated with `make docs-generate`, since `scripts/generate_docs.py` lifts that header verbatim into the `provisioning-steps` region. Editing the region directly would have gone stale on the next run.
- `docs/site/src/content/docs/operator/provisioning-scripts.md:8` — "(plus their teardown mirrors)" instead of calling the mirrors optional.
- `docs/site/src/content/docs/install/uninstall.md:73` — the teardown row now reads "skipped unless `vars.sh` records `ENABLE_GVISOR=true`", which is what that script actually does (see the teardown thread — I kept its default at `false`).
- `README.md:86` and `.agents/skills/install-kube-agents/SKILL.md:120` while I was there.

`make docs-check` is clean.

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
AntonTyb: Removed in 8986eb47. The arm is gone entirely — the template ships `runtimeClassName: gvisor` uncommented, so `s/# runtimeClassName: gvisor/.../` could never match, and only the opt-out needs an edit now:

```sh
# The template ships the field enabled, so only the opt-out needs an edit.
if ! is_truthy "$ENABLE_GVISOR"; then
  print_info "Disabling gVisor runtimeClassName in '$CR_MANIFEST' (ENABLE_GVISOR=${ENABLE_GVISOR})..."
  sed -i.bak 's/^\([[:space:]]*\)runtimeClassName: gvisor/\1# runtimeClassName: gvisor/' "$CR_MANIFEST" && rm -f "${CR_MANIFEST}.bak"
fi
```

The remaining arm had a second, quieter problem you did not flag: `s/^[[:space:]]*runtimeClassName/# runtimeClassName/` consumed the leading whitespace, so the commented-out line came back flush against the left margin and lost its place in the block. It now captures the indentation and puts it back.

`tests/test_gvisor_default.py::test_there_is_no_dead_enable_arm` asserts the enable arm stays gone, and `test_the_opt_out_comments_the_field_out` extracts this exact `sed` from the script and runs it over an `envsubst`-rendered template, so the substitution cannot drift from the line it edits.

k8s-operator/scripts/teardown_02_gvisor_nodepool.sh
AntonTyb: Agreed, and reverted in 8986eb47 — this half of the flip should not have been made at all.

Teardown keeps `${ENABLE_GVISOR:-false}` and deliberately does *not* read the new `DEFAULT_ENABLE_GVISOR`, with the reason written next to it so nobody "fixes" the inconsistency later:

```sh
# Deliberately NOT DEFAULT_ENABLE_GVISOR. Provisioning and teardown want
# opposite defaults: getting it wrong on the way in costs a node pool nobody
# asked for, getting it wrong on the way out deletes one somebody wanted.
# ensure_teardown_state only sets ENABLE_GVISOR when vars.sh exists, so an
# unset value here means "no state to say this install owns a gVisor pool" —
# and `teardown.sh --no-confirm` from a stateless clone would otherwise delete
# any pool named gvisor-pool without asking.
```

The asymmetry is not a symmetry bug: an unset value on the way out is not "the user wants a sandbox", it is "there is no state saying this install owns one", and a pool named `gvisor-pool` on that cluster may well belong to something else. The script header goes back to being accurate as written, so it needed no change.

`tests/test_gvisor_default.py::test_teardown_keeps_the_opposite_default` pins it, asserting `DEFAULT_ENABLE_GVISOR` appears nowhere in the script outside comments.

k8s-operator/scripts/platform-agent.yaml.template
AntonTyb: Fixed in 8986eb47, though as Python rather than Go — `tests/test_gvisor_default.py`, 9 tests.

Go would have meant a new package under `k8s-operator/` reaching up into `scripts/` to shell out, whereas `tests/` already runs `envsubst`-adjacent shell-facing assertions and is wired into `agent-startup-test.yml` with PyYAML available. Same coverage, no new harness.

It does what you asked for and a bit more:

- renders the template through the real `envsubst` and parses the YAML, asserting `spec.deployment.availability.runtimeClassName == "gvisor"`;
- runs the opt-out arm — the `sed` **read out of `provision_08` at test time**, not restated — over that rendered manifest, and asserts the field is commented out, the document still parses, and the `.bak` is cleaned up. Restating the `sed` would have let the two drift, which is the failure mode the test exists for;
- asserts the dead enable arm stays gone, which as you say this would have caught;
- pins `DEFAULT_ENABLE_GVISOR="true"` as the single home, that both provision steps read it rather than a literal, that `install.sh` resolves from it and has no `:-false` fallback left on the parameter path, and that the interactive prompt lists the sandbox first (`prompt_menu` always defaults to option 1, so the ordering *is* the default);
- asserts teardown keeps the opposite default on purpose;
- asserts the RuntimeClass probe exists and that the explicit-request `return 1` precedes the fallback, so an explicit `ENABLE_GVISOR=true` cannot be silently downgraded.

One thing that would otherwise have made the whole test useless: `agent-startup-test.yml` gates on a `dorny/paths-filter` allow-list that did not include the provisioning scripts, the template or `install.sh`. A change to exactly the files this guards would have skipped the only job that runs it. The filter is widened, with a comment saying why.

AntonTyb:
@mplakhtiy — the two findings outside the diff, since they have no thread to answer in. Both fixed or answered in `8986eb47`.

## HIGH — `install.sh` still defaults gVisor off — **fixed**

Correct, and the "actively defeats" part is the bit that mattered: `write_state_var` puts the answer in `vars.sh`, `load_state` sources it, so `init_var` never sees an unset `current_val` and the script default is dead on arrival. Patching the four `false`s would have left the same trap for the next default.

So the default has one home now, in the block in `common.sh` that already says exactly this:

```sh
# The values the per-step provision scripts and the zero-friction installer must
# agree on. install.sh sources this file rather than restating them, so each
# default has exactly one home and the two entry points cannot drift apart.
DEFAULT_CLUSTER_NAME="platform-agent-host"
DEFAULT_REGION="us-central1"
DEFAULT_MODEL_PROVIDER="gemini"
DEFAULT_ENABLE_GVISOR="true"
```

`install.sh:72` is now empty-by-default and `resolve_shared_defaults()` fills it from there, alongside `PARAM_MODEL_PROVIDER` and `PARAM_REGISTRY_PREFIX`; `:109`, `:803` and `:1392` follow. `.agents/skills/install-kube-agents/SKILL.md:120` now points at `common.sh` `DEFAULT_ENABLE_GVISOR` rather than restating a value.

The prompt at `:1445` had a second bug the reordering exposes. It only ever assigned on choice 2:

```sh
if [ "$gvisor_choice" = "2" ]; then
  enable_gvisor="true"
fi
```

With "No" first that was harmless. With "Yes" first, a fall-through would have left whatever `--gvisor=` had already put in the variable, so an explicit `--gvisor=false` plus an interactive "No" would still have enabled the sandbox. Both branches assign now.

Two `${...:-false}` occurrences survive on purpose and are not on the parameter path: `:744` is a JSON literal in the pre-flight report, which must stay valid YAML/JSON even if it renders before `resolve_shared_defaults` runs, and the closing summary reads back what was actually written.

## MEDIUM — the chart cannot express this — **agreed on the altitude, deferring the change**

You are right that a `+kubebuilder:default` on `RuntimeClassName` is the correct altitude, and right that Autopilot is where it would have been free. I do not think it belongs in this PR:

- It changes the rendered spec for **every** existing chart and Terraform user on the next upgrade, not just new installs. That is a different blast radius from flipping a provisioning-script default, and it wants its own live validation against an existing install.
- It does not compose with the `sed`, it replaces it. Commenting the field out means "unset", which is precisely what a CRD default fills back in — so `ENABLE_GVISOR=false` would stop working the moment the default lands. Removing the `sed` machinery is the right end state, but it means the opt-out has to become an explicit empty value or a separate knob, and that is a design decision rather than a follow-up edit.
- The chart gap is real but orthogonal: `spec.deployment.availability` is absent from `platform-agent-cr.yaml` entirely, so the chart could not express this default even if the scripts and the CRD agreed. Adding the block and a `values.yaml` key is chart work that stands on its own.

Suggest a follow-up issue covering all three — CRD default, chart `availability` block, and retiring the `sed` — rather than widening this one. Happy to file it if you agree; say the word and I will, or push back and I will do it here.

---

_This response and the fixes in `8986eb47` were generated with the help of Claude._

AntonTyb:
Flagging an interaction with an open bug that I do not think has been considered on this PR, because it argues about sequencing rather than about the change itself.

## #610 says gVisor corrupts the agent's databases, and this PR makes gVisor the default

[#610](https://github.com/gke-labs/kube-agents/issues/610) — "SQLite WAL on gVisor 9p mount corrupts `state.db` and `kanban.db`" — is **open, unassigned, and untouched since 2026-08-10** (zero comments). It reports two independent database corruptions in three days on the same volume: `kanban.db` on 2026-08-07, `state.db` on 2026-08-09.

The mechanism is specific to gVisor rather than incidental to it. The PD is ext4, but the sandbox does not see it as ext4 — inside the sandbox the mount is a 9p gofer:

```
none /opt/data 9p rw,trans=fd,...,cache=remote_revalidating,directfs 0 0
```

SQLite WAL needs mmap'd `-shm` coherence and fcntl byte-range locks that a 9p gofer mount does not provide. The reported pod was running `runtimeClassName: gvisor` — i.e. the configuration this PR makes the default for every new install.

Worse, the existing safety net does not catch it. #610's analysis is that `hermes_state.py` falls back to `journal_mode=DELETE` **reactively, by matching error strings** (`"locking protocol"`, etc.), and gVisor raises none of them — so the fallback never fires and the corruption is silent.

## Verifiable from this repository, not only from the issue

The upstream `hermes_state.py` is not in this tree, so I cannot say whether its marker list has since been widened — that is worth checking against the image this PR deploys. But the same unguarded pattern is in this repository, on `main`, today:

`agents/chat/defaults/plugins/session_store/store.py:174-183`

```python
conn = sqlite3.connect(db_path, timeout=5.0, check_same_thread=False)
conn.execute("PRAGMA journal_mode=WAL")
```

No filesystem probe, no fallback, no gVisor special case. Whatever the state of the upstream fix, this connection opens in WAL on whatever filesystem it is handed.

## Why this is a question for this PR

Nothing here says the default is wrong — RUNTIME-002 is a real hardening win and the reasoning in this PR holds. The issue is blast radius and ordering. Before this change, gVisor was opt-in, so #610 affected the installs that had deliberately chosen the sandbox. After it, it affects every new install by default, and #745 would widen it to the Terraform and Helm path as well. That turns an open data-corruption bug from something a handful of hardened installs are exposed to into the default posture, while the detection gap that makes it silent is still open.

Three ways to sequence it, roughly in order of how much I like them:

1. **Fix the detection first.** Make the WAL fallback proactive — probe the filesystem (`/proc/mounts` shows `9p` plainly, as the mount line above does) rather than waiting for an error string that gVisor never emits. That closes #610 for both the upstream store and `session_store/store.py`, and this PR then merges without carrying the risk.
2. **Merge this, and treat #610 as a release blocker** rather than an open issue — the default is not shipped to users until the corruption path is closed.
3. **Merge this as is** and accept the exposure, with the interaction written into the PR body so the decision is on the record rather than discovered later.

I have no strong view between 1 and 2; 3 only seems reasonable if someone has evidence the corruption is rarer than two-in-three-days suggests. Either way #610 should probably get an assignee before this lands.

Happy to be wrong about any of this — in particular, if the upstream `hermes_state.py` fallback has been fixed since 2026-08-10, most of the above collapses and this is just a note to close #610.

---

_Analysis produced with the help of Claude._

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
```diff
@@ -39,7 +39,16 @@ DEFAULT_PROJECT_ID="${ACTIVE_PROJECT:-$(whoami 2>/dev/null || echo "user")}"
...
+# Whether gVisor was asked for or merely defaulted. load_state has already
+# sourced vars.sh, so anything set by now came from the environment or from the
+# choice the installer saved — either way deliberate. execute_custom_resource
+# uses the distinction to decide whether a cluster with no gvisor RuntimeClass
+# is an error or a fallback.
+GVISOR_EXPLICIT=0
+if [ -n "${ENABLE_GVISOR:-}" ]; then
+  GVISOR_EXPLICIT=1
+fi
+init_var "ENABLE_GVISOR" "$DEFAULT_ENABLE_GVISOR" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
```
mplakhtiy: **🔴 High — this step persists the value it defaulted, so the second run reads it back as explicit and hard-fails.**

The comment says anything set by now "came from the environment or from the choice the installer saved — either way deliberate." There is a third source it does not account for: `init_var` itself. When `init_var` takes the default it calls `save_var` (`common.sh:205`), and `save_var` appends `export ENABLE_GVISOR=true` to `vars.sh` (`common.sh:122`). Nothing in `execute_custom_resource` unwinds that — the fallback at line 185 sets `ENABLE_GVISOR="false"` in memory only, so `vars.sh` is left recording `true`. On the next invocation `load_state` sources that file before line 47, `ENABLE_GVISOR` is non-empty, and `GVISOR_EXPLICIT=1`.

Reproduced against the real `common.sh` with a temp `VARS_FILE` and `CI=true`:

```
=== run 1 ===  RESULT: ENABLE_GVISOR=true GVISOR_EXPLICIT=0
=== vars.sh === export ENABLE_GVISOR=true
=== run 2 ===  RESULT: ENABLE_GVISOR=true GVISOR_EXPLICIT=1
```

**Failure scenario.** An operator redeploys the agent to an existing cluster that has no gVisor node pool — a cluster provisioned before this became the default, which is every cluster today. Run 1 prints "Cluster 'X' has no 'gvisor' RuntimeClass; deploying WITHOUT sandbox isolation" and succeeds. They change an image tag and run the identical command again. Run 2 stops at lines 179-181 with `✗ ENABLE_GVISOR=true, but cluster 'X' has no 'gvisor' RuntimeClass`, and `run_step` turns that `return 1` into `exit 1`. Nothing about the cluster, the environment or the command changed between the two runs; the only difference is a line the first run wrote. The script's own header (`:5`) advertises "Idempotent script", and the `--menu` → "Save & Apply" path (`install.sh:962-1008`) reaches the same code.

The CI redeploy workflows escape it only by accident: they start from a fresh checkout, so `vars.sh` never survives to the next run. A self-hosted runner, a Cloud Shell session, or a laptop does not have that property. This defeats the mechanism `8986eb47` added to close round-1 finding 1.

Fix options, all your call: compute `GVISOR_EXPLICIT` from the environment *before* `load_state` and separately honour an installer-written value; record the provenance (`ENABLE_GVISOR_SOURCE`) in `vars.sh` alongside the value; or leave the resolved value unpersisted in this step, the way `IMAGE_TAG` already is for the reason spelled out at `common.sh:585-593`.

*(For the record, `GVISOR_EXPLICIT` on the full `provision.sh` path is correct, not a bug: `provision_02` runs first and saves the value, so a missing RuntimeClass at step 08 really is an error there. This is about the step-08-alone path.)*

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
```diff
@@ -152,9 +161,34 @@ execute_custom_resource() {
...
+  # whole AGENT_READY_TIMEOUT discovering it. kubectl is already connected by
+  # step 1, so ask the API server the same question the operator will.
+  #
+  # The answer means different things depending on who chose the value. An
+  # explicit ENABLE_GVISOR=true is a requirement, and quietly dropping the
+  # sandbox would be the worst possible response to it. The default is a
+  # preference, and clusters provisioned before this became the default have no
+  # gvisor pool — the redeploy workflows (reusable-deploy-agent.yml runs only
+  # this step, never provision_02) hit exactly that.
+  if is_truthy "$ENABLE_GVISOR" && ! kubectl get runtimeclass gvisor >/dev/null 2>&1; then
```
mplakhtiy: **🟠 Medium — the probe fails open on every kubectl error, not just NotFound.**

`2>&1` to `/dev/null` collapses every non-zero exit into one answer. An expired credential, an RBAC denial on `node.k8s.io/runtimeclasses`, API-server throttling, a transient 5xx, or a `kubectl` that is not connected at all all read as "the cluster has no gvisor RuntimeClass". The operator's `validateRuntimeClass` — which the comment above says this probe asks the same question as — does the opposite: `platformagent_controller.go:241` branches on `errors.IsNotFound(err)` and returns a hard error for anything else.

**Failure scenario.** A redeploy to a cluster that *does* have the sandbox pool hits a transient API error during the probe. `GVISOR_EXPLICIT` is 0 (fresh CI checkout), so the script warns, comments the field out, applies a CR with no `runtimeClassName`, and the pipeline goes green. The agent moves off `gvisor-pool` onto the default pool and runs on the host kernel. Nothing downstream notices: with no `runtimeClassName` the operator's `validateRuntimeClass` returns `nil` and the Deployment reconciles healthy. The only record is a `print_warning` in a log nobody reads on a passing build.

That is a fail-open on the one control this PR exists to turn on. Capturing the output and falling back only when the error actually says `NotFound` is a couple of lines.

k8s-operator/scripts/provision_02_gvisor_nodepool.sh
```diff
@@ -26,7 +27,7 @@ check_prereqs "gcloud" "kubectl"
 print_step "Setting up Configuration State"
 load_state
 
-init_var "ENABLE_GVISOR" "false" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
+init_var "ENABLE_GVISOR" "$DEFAULT_ENABLE_GVISOR" "Enable GKE Sandbox (gVisor) runtime isolation? (true/false)"
```
mplakhtiy, line 31: **🟠 Medium — this now aborts the whole pipeline on an Autopilot cluster, where gVisor needs no pool at all.**

Before the flip, `ENABLE_GVISOR` defaulted to `false` and this script exited 0 without touching the cluster. Now it defaults to `true`, falls through the guard below, and runs `gcloud container node-pools create --sandbox=type=gvisor`, which Autopilot rejects — Autopilot does not accept user-created node pools. `run_step` (`common.sh:749-758`) turns that into `exit 1`, and `provision.sh` stops at step 2 of 13.

Autopilot is a reachable target for this pipeline, not a hypothetical:

- `provision_01`'s `verify_cluster` (`:117-129`) only checks that the named cluster exists with an acceptable `databaseEncryption.state`, so an existing Autopilot cluster passes and creation is skipped.
- `provision_03:132-152` detects Autopilot explicitly and adapts cert-manager to it — the pipeline is written to run there.
- `terraform/modules/gke-cluster` builds an Autopilot cluster, and the quickstart presents Terraform + Helm as a peer of the script path.

The irony is that Autopilot is the one place the default is free — the operator's own diagnostic says so: "In GKE Autopilot, gVisor is supported automatically" (`platformagent_controller.go:243`). The right behaviour there is to skip step 02 and keep `runtimeClassName: gvisor`, which is close to what already happens elsewhere: probe the cluster the way `provision_08` now does, or check `autopilot.enabled` before attempting the pool.

**Failure scenario.** Someone points `install.sh` at an existing Autopilot cluster. The install dies at step 2 with a `gcloud` error about node pools, on a cluster where the sandbox would have worked with no node pool at all.

README.md
```diff
@@ -83,7 +83,7 @@ The runtime is built on the Hermes agent framework and wires in MCP servers for
 - **Least-privilege RBAC** — the agent's Kubernetes identity is read-only and cannot read Secrets.
 - **Credential isolation** — the agent sandbox container never receives API keys or tokens; an Envoy credential-proxy sidecar injects them at the network boundary.
 - **At-rest database encryption & state security** — GKE etcd database encryption (CMEK) via Cloud KMS, strict state file permissions (`umask 077`), and mandatory encryption pre-flight gates.
-- **Kernel-level sandboxing** — agent workloads can run under a gVisor RuntimeClass (GKE Sandbox).
+- **Kernel-level sandboxing** — agent workloads run under a gVisor RuntimeClass (GKE Sandbox) by default; opt out with `ENABLE_GVISOR=false`.
```
mplakhtiy: **🟠 Medium — this promises the default on a surface that does not have it, and `charts/kube-agents/README.md:268` now states something false.**

`ENABLE_GVISOR` does not exist on the Helm/Terraform path, and that path still ships the agent unsandboxed — `charts/kube-agents/values.yaml:231` defaults `runtimeClassName` to `""`. This bullet is unqualified, in the project's security-features list. `architecture.mdx:110,118` carry the same unqualified "by default" on a page that documents both install surfaces.

The chart README (unchanged by this PR, so no line to anchor to) is the sharper half. It currently reads:

> - `harness.hermes.dashboardEnabled` is the one field where the two install paths disagree by default: the CRD defaults it to `true`, the script path to `false`.

After this PR, `runtimeClassName` is a second such field: the script path defaults it to `gvisor`, the chart to `""`. "the one field" is no longer true, and that sentence is exactly where a chart user would look to find that out.

`make iac-parity-check` passes, correctly — it enforces the scalar subset and this is a structural divergence, the kind `AGENTS.md` sends to the `review-iac-parity` skill.

**Failure scenario.** A team installs via `terraform/examples/full-install/`, reads this security list, and records RUNTIME-002 as mitigated. Their agent is on the host kernel.

Cheapest honest fix is a clause on both claims naming the script path; the more complete one is `runtimeClassName: gvisor` in `values.yaml` behind the warning the chart README already carries. Either way it belongs in the PR body, because nothing there currently says the declarative path opted out.

mplakhtiy (commented):
Adversarial review pass against `b19d4184` merged onto `main` @ `76a074b8` (merged cleanly, no conflicts). Medium-and-above findings only, anchored inline — three Lows are held back.

Re-verified first: `8986eb47` closes **all eight** round-1 findings, and the fixes are real rather than claimed. Every finding below is new, and all but one are in code that commit added.

- 🟠 **Medium — the PR body is stale.** The **Context** section still says "This is not ready to merge as it stands" and describes both round-1 High findings as open; `8986eb47` closed both. The **What Changed** file list is also missing nine of the fifteen changed files, including `install.sh`, `common.sh` and the new test.
- ❓ **Open question — do the autopush and staging clusters have a gVisor node pool?** Round 1 asked; the body still does not say. It no longer decides whether CI breaks, since the probe handles that, but it now decides something quieter: with no pool, those two environments deploy the agent *unsandboxed* on every push, forever, behind a warning on a green build — RUNTIME-002 closed while the project's own clusters do not have it. One `gcloud container node-pools list --cluster=<c>` against each settles it.
- ❓ **Open question — `pull-kube-agents-smoke-test` is still failing and its log is still unreadable** (403 on the GCS object; the Prow `/log` endpoint 404s, pod gone). It is `Required: false`. It passes on #494 and #504 and fails on #546 and #674, so it is neither universally broken nor obviously specific to this diff — but if it provisions or deploys against a real cluster it is the most likely first place the three findings above would surface, and someone with bucket access should read it rather than retest past it.

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
AntonTyb: Fixed in `f3e43a5e`. You were right, and it reproduced exactly as you described — I ran your harness against the real `common.sh` before changing anything and watched `GVISOR_EXPLICIT` go 0 → 1 across two identical invocations.

Took the third of your options: the step resolves the default without persisting it.

```bash
GVISOR_EXPLICIT=0
if [ -n "${ENABLE_GVISOR:-}" ]; then
  GVISOR_EXPLICIT=1
elif is_non_interactive; then
  export ENABLE_GVISOR="$DEFAULT_ENABLE_GVISOR"
else
  init_var "ENABLE_GVISOR" "$DEFAULT_ENABLE_GVISOR" "..."
  GVISOR_EXPLICIT=1
fi
```

`init_var` is still used on the interactive path, because an answer typed at the prompt _is_ a choice and is worth saving — pressing enter there accepts a default the user was shown, which is not the same as a default nobody saw. Only the non-interactive branch bypasses persistence, and that is the branch your failure scenario runs through. `IMAGE_TAG` was the precedent, as you suggested.

Two things your parenthetical prompted beyond the fix:

- The comment above the block now names `provision_02` as the third legitimate source, since it saves the value and then creates the pool. That was the part of the original comment that was actually true, and it was not written down.
- `init_var`'s own docstring in `common.sh` now says it persists the default too, and the comment in `init_var_registry_prefix` claiming init_var "only saves values it prompted for" is gone — it was false in the same way, and it is what made this easy to get wrong.

`tests/test_gvisor_default.py` pins it with three tests that execute the real resolution block against the real `common.sh` with a temp `VARS_FILE`: the default is not written to `vars.sh` and does not latch across runs, an exported value is explicit, and a `vars.sh`-saved value is explicit. All three fail against the pre-fix script.

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
AntonTyb: Fixed in `2ec5a66b`. The probe now captures the output and gives a three-way answer instead of a boolean:

```bash
gvisor_runtimeclass_state() {
  local out
  if out=$(kubectl get runtimeclass gvisor 2>&1); then
    echo "present"; return 0
  fi
  case "$out" in
    *NotFound*) echo "absent" ;;
    *) echo "unknown: ${out}" ;;
  esac
  return 1
}
```

Only `absent` reaches the fallback. `unknown` prints what kubectl actually said and returns 1, so `run_step` stops the step rather than deploying unsandboxed on a cluster that may well have the pool. That matches `validateRuntimeClass` branching on `errors.IsNotFound` — which is what the comment claimed the probe did, and now does.

Matching on the message text rather than a status code is the part worth a second opinion. `kubectl get` does not give the shell a distinguishable exit code, and the string it prints is `Error from server (NotFound): runtimeclasses.node.k8s.io "gvisor" not found`. If you would rather this went through `-o json` or a `--raw` call for a structural answer, say so and I will change it.

Six error shapes are covered by tests against a stubbed `kubectl`: exit 0, NotFound, Forbidden, Unauthorized, TooManyRequests, InternalError, and an unreachable server. There is also a static assertion that `ENABLE_GVISOR="false"` appears in the `absent)` arm and **not** in the `*)` arm, so a future edit cannot quietly reintroduce the fail-open.

k8s-operator/scripts/provision_02_gvisor_nodepool.sh
AntonTyb: Fixed in `2ec5a66b`. `provision_02` now checks `autopilot.enabled` and exits 0 with an explanation before `run_step` can reach `node-pools create`:

```bash
cluster_is_autopilot() {
  local autopilot
  autopilot="$(gcloud container clusters describe "$CLUSTER_NAME" \
      --location="$REGION" --project="$PROJECT_ID" \
      --format='value(autopilot.enabled)' 2>/dev/null || echo "")"
  is_truthy "$autopilot"
}
```

I went with the `gcloud describe` form rather than the `kubectl … gk3-` node check `provision_03:129` uses, because at that point in `provision_02` kubectl is not connected yet — `connect_cluster` is step 2 of this script, after the pool creation this needs to guard. Same question, the only source that can answer it there.

A `describe` that cannot answer reads as "not Autopilot", so a Standard cluster with a transient gcloud failure behaves exactly as before and the `create` below reports the real error. That is the opposite default from the probe in finding 2, deliberately: there, guessing wrong drops the sandbox silently; here, guessing wrong skips a pool a Standard cluster needs, so the safe direction is to keep going.

Nothing downstream is turned off. `provision_08` still deploys a CR asking for `runtimeClassName: gvisor`, which is right — the operator's "In GKE Autopilot, gVisor is supported automatically" is advisory, and `validateRuntimeClass` still requires the RuntimeClass **object** to exist before it will reconcile. I checked that rather than assuming it, since "Autopilot needs no pool" and "Autopilot needs nothing" are not the same claim and only the first is true.

The script's header banner now says it skips on Autopilot, and `make docs-generate` was re-run so the generated table in `k8s-operator/scripts/README.md` picked that up. Tests: the predicate against a stubbed `gcloud` for `True` / `False` / empty, plus a static check that the guard and its `exit 0` precede `run_step "1. Provision gVisor Node Pool"`.

README.md
AntonTyb: Fixed in `2ec5a66b`. I verified both halves against the tree before changing anything: `values.yaml:231` is `""`, and the chart README did say "the one field".

Took the cheaper honest fix, on all three claims:

- `README.md:86` now says "on the script install path", and adds that a Helm or Terraform install leaves `platformAgent.deployment.availability.runtimeClassName` empty and runs on the host kernel until you set it to `gvisor`.
- `architecture.mdx:110` and `:118` are both qualified the same way. `:110` also picks up the Autopilot skip from finding 3.
- `charts/kube-agents/README.md` no longer calls `dashboardEnabled` the one divergence. Both are listed now, under a sentence that does not carry a count — a hard-coded "two" would be the same bug again the next time one is added.

I did **not** take the more complete option of defaulting `runtimeClassName: gvisor` in `values.yaml`, and I think that is right rather than merely cheaper: the sandbox node pool `provision_02` creates has no Terraform or chart-side equivalent, so a chart default of `gvisor` would put every chart install into Degraded/`RuntimeClassNotFound` unless the user had built the pool by hand first. That is a bigger change than this PR — it needs the pool in `terraform/modules/gke-cluster` before the chart can honestly default it. Happy to file it as a follow-up if you agree that is the right end state.

You were right that it belonged in the PR body too. The body is being replaced in full; the new version states the declarative path deliberately does not change and why, and it also adds the **Self-Review** section this PR has never had.

PR #497, opened 2026-07-31
Feat/watcher in credential proxy

stalhaali:
# Pull Request

  ## Why This Change
     
  Move the multi-cluster event watcher into the credential proxy and turn fan-in on
  by default.
  
  The watcher needs credentials for every cluster it reaches. This project keeps
  credentials in `envoy-credential-proxy` and away from the agent sandbox, so
  rather than grow a second credential holder next to the sandbox, the watcher
  moves into the proxy. That follows the convention directly: services that need
  credentials live in the credential proxy and talk to the agent over a narrow
  interface.
  
  Based on `main` — #473 has merged.
  
  ## What Changed
  
  **Files:**
  
  - `k8s-operator/internal/controller/platformagent_manifests.go`
  - `k8s-operator/internal/controller/platformagent_manifests_test.go`
  - `k8s-operator/internal/controller/platformagent_controller_test.go`
  - `deploy/shared/start-services.sh` (new)
  - `deploy/shared/envoy-credential-sidecar.sh` (deleted)
  - `deploy/shared/docker-entrypoint.sh`
  - `deploy/docker/Dockerfile`
  - `INSTALL.md`
  - `k8s-operator/cmd/k8s-event-watcher/README.md`
  - `docs/credential-isolation-design.md`
  - `docs/site/src/content/docs/reference/credential-isolation.md`
  - `docs/site/src/content/docs/reference/glossary.md`
  - `docs/site/src/content/docs/deploy/kustomize.md`
  - `docs/site/src/content/docs/concepts/observability.md`
  
  **Added/Changed/Fixed:**
  
  - **The standalone `event-watcher` container is deleted.** Pod container count
    drops from five to four. 
  - **One entrypoint starts three peer services.** `envoy-credential-sidecar.sh`
    becomes `start-services.sh`, which launches the credential runtime, Envoy, and
    the watcher as siblings. They share a container because all three need
    credentials, not because any belongs to another — the old name and structure
    implied the watcher was part of the Envoy sidecar.
  - **Only two of the three are fatal.** Envoy and the credential runtime are the
    container's reason to exist: if either dies the agent loses every credentialed
    command, so `wait -n` tracks them and their exit ends the container. The
    watcher is best-effort observability, so it is supervised and restarted in
    place on a 10s loop instead. Shutdown stops the supervisor before the watcher
    so it is not restarted on the way down.
  - **The container passes no arguments.** The watcher's flags — `--in-cluster`,
    the loopback `--daemon-url`, `--token-env`, `--owner`, `--reason`,
    `--profiles-dir` — live in the entrypoint. They describe how processes inside
    one container reach each other, not how the deployment is configured, and
    passing them as container `Args` made the Pod spec look like it was configuring
    the entrypoint while `"$@"` forwarded them to a different binary. `--reason`
    loses nothing by moving: it was already a literal in the operator with no CRD
    field behind it.
  - **`EVENT_WATCHER_CLUSTER_NAME` is the one per-install value, set from
    `resolveHarnessClusterName`.** Deliberately not read from `GKE_CLUSTER_NAME` in
    the shell: that variable is only set when `projectID`, `location` and
    `clusterName` are all present, while the name itself needs only `clusterName`,
    and all three are `omitempty`. A CR naming its cluster but omitting the project
    would have fallen through to the default and mislabelled every payload and
    metric — silently, since nothing would fail.
  - **Fan-in is on by default.** `--profiles-dir` and `--in-cluster` are set
    unconditionally; there is no CR field to opt in. An install managing N clusters
    wants all N watched, and a flag would leave clusters silently unmonitored.
  - **`event-watcher-ksa-token` moves to the credential proxy**, at the
    conventional `/var/run/secrets/kubernetes.io/serviceaccount`. The container now
    holds two differently audienced tokens: its own, scoped to
    `kubeagents-credential-proxy`, which the Kubernetes API rejects; and the
    watcher's, with the default audience, which it accepts. The second is what lets
    `--in-cluster` still cover the management cluster — that cluster never gets a
    Cluster Agent profile, so profile discovery alone would miss it.
  - **`API_SERVER_KEY` is declared inside `buildCredentialProxyEnv`**, not appended
    after it. The watcher reads it via `--token-env`; it is a non-secret loopback
    sentinel, not a credential (the real secret is `API_SERVER_EXTERNAL_KEY`).
    Declaring it in the managed set means `mergeCredentialProxyEnv` reserves the
    name structurally — appending after the merge left it protected only by
    happening to appear in `SensitiveEnvVars`.
  - **Memory request 256Mi → 384Mi, CPU 100m → 150m.** The watcher keeps an
    informer cache and a dedup cache per cluster.
  - **Tests now look containers up by name.** Removing a container shifted every
    index after it, and positional assertions had already broken once before on a
    reorder — silently reading the wrong container, then nil-dereferencing
    somewhere unrelated. A `containerByName` helper replaces the indices; the
    blocks that assert container *order* keep theirs, since order is what they test.
  - Seven documents corrected where they described the watcher as its own
    container, named the old entrypoint, or gave the old container count.
  
  **Note:** `event-watcher-kubeconfig` and the `KUBECONFIG` env var on the
  credential proxy are *not* the watcher's, despite the name — they are the
  proxy's own scratch kubeconfig, written by `CREDENTIAL_PROXY_BOOTSTRAP_COMMAND`.
  Left untouched.
  
  ## Testing
  
  `go build`, `go vet`, `gofmt -l`, and `go test -race -count=1` across
  `./api/...`, `./cmd/k8s-event-watcher/...`, `./internal/...` — clean. Plus
  `bash -n` on both entrypoints, `make docs-check`, and `prettier`.
  
  Manifest tests updated for the removed container: container counts 6→5, 5→4, and
  4→3 across the three deployment shapes, with assertions that no container is
  named `event-watcher`, that the credential proxy takes **no** arguments, and that
  it carries `EVENT_WATCHER_CLUSTER_NAME`, the `API_SERVER_KEY` sentinel, and the
  default-audience token at the in-cluster path.
  
  Both golden manifests regenerated. `platformagent-tagged.yaml` is worth a look:
  it arrived from `main` after this branch was cut, merged without conflict because
  nothing here had touched it, and still described the old layout. A clean merge
  says nothing about whether a rendered-output snapshot still matches the code, and
  the golden test was the only thing that caught it.
    wants all N watched, and a flag would leave clusters silently unmonitored.
  - **`event-watcher-ksa-token` moves to the credential proxy**, at the
    conventional `/var/run/secrets/kubernetes.io/serviceaccount`. The container now
    holds two differently audienced tokens: its own, scoped to
    `kubeagents-credential-proxy`, which the Kubernetes API rejects; and the
    watcher's, with the default audience, which it accepts. The second is what lets
    `--in-cluster` still cover the management cluster — that cluster never gets a
    Cluster Agent profile, so profile discovery alone would miss it.

  Four things I'd point out about the changes:
  
    Declaring it in the managed set means `mergeCredentialProxyEnv` reserves the
    name structurally — appending after the merge left it protected only by
    happening to appear in `SensitiveEnvVars`.
  - **Memory request 256Mi → 384Mi, CPU 100m → 150m.** The watcher keeps an
    informer cache and a dedup cache per cluster. 
  - **Tests now look containers up by name.** Removing a container shifted every
    index after it, and positional assertions had already broken once before on a
    reorder — silently reading the wrong container, then nil-dereferencing
    somewhere unrelated. A `containerByName` helper replaces the indices; the 
    blocks that assert container *order* keep theirs, since order is what they test.
  - Seven documents corrected where they described the watcher as its own
    container, named the old entrypoint, or gave the old container count.
  
  **Note:** `event-watcher-kubeconfig` and the `KUBECONFIG` env var on the
  credential proxy are *not* the watcher's, despite the name — they are the
  proxy's own scratch kubeconfig, written by `CREDENTIAL_PROXY_BOOTSTRAP_COMMAND`.
  Left untouched.

  ## Testing

  `go build`, `go vet`, `gofmt -l`, and `go test -race -count=1` across
  `./api/...`, `./cmd/k8s-event-watcher/...`, `./internal/...` — clean. Plus
  `bash -n` on both entrypoints, `make docs-check`, and `prettier`.          

  Manifest tests updated for the removed container: container counts 6→5, 5→4, and
  4→3 across the three deployment shapes, with assertions that no container is
  named `event-watcher`, that the credential proxy takes **no** arguments, and that
  it carries `EVENT_WATCHER_CLUSTER_NAME`, the `API_SERVER_KEY` sentinel, and the
  default-audience token at the in-cluster path.

  Both golden manifests regenerated. `platformagent-tagged.yaml` is worth a look:
  it arrived from `main` after this branch was cut, merged without conflict because
  nothing here had touched it, and still described the old layout. A clean merge
  says nothing about whether a rendered-output snapshot still matches the code, and
  the golden test was the only thing that caught it.

  ---

  **Rollout:** this needs a **rebuilt `credential-proxy` image** — the entrypoint is
  baked in at build time, so redeploying the operator alone yields the new manifest
  and an image with no `start-services`. The Pod is recreated regardless: a
  container is removed, mounts and resource requests change. This is also the change
  that switches multi-cluster watching on, so alert volume scales with fleet size
  and triage sessions will fire for clusters that were previously unmonitored.
  Rollback is a revert plus the prior image.

  Four things I'd point out about the changes:

  - The first bullet was actively wrong now — it said arguments "move onto
  envoy-credential-proxy, which passes them through," which is exactly the pattern Haoxu
  objected to and that's now gone. Replaced with the no-arguments bullet.
  - resolveHarnessClusterName dropped from the list. It's no longer a bare refactor note; it's
  load-bearing for EVENT_WATCHER_CLUSTER_NAME, so it's folded into that bullet where the
  reasoning lives.
  - "Four docs" → "Seven documents", covering the entrypoint rename in the glossary and
  kustomize tree, plus the INSTALL.md container count.
  - The rollout note now leads with the image rebuild. That's the operational trap here — new
  manifest against an old image gives you a container whose Command doesn't exist, and it's not
  obvious from the diff.


stalhaali:
> Found 1 issue.
> 
> ### Findings outside this diff
> GitHub only accepts an inline comment on a line the diff actually shows. These point somewhere it does not — a file this pull request leaves alone, or a line too far from a changed one to appear in the diff's context.
> 
> #### 🟡 Low — INSTALL.md still says the agent Pod runs five containers
> Removing the `event-watcher` container drops the default Pod from five managed containers to four, but the troubleshooting step in `INSTALL.md` still tells the reader to expect five.
> 
> The dashboard defaults to enabled (`k8s-operator/examples/platformagent.yaml` sets `dashboardEnabled: true`), so the documented install renders `platform-agent`, `platform-agent-dashboard`, `fluent-bit`, `event-watcher`, `envoy-credential-proxy` — the five this line was written for. The PR's own test updates confirm the new count: `TestBuildDeployment_DashboardEnabled` moves from `!= 5` to `!= 4`, and the regenerated golden `k8s-operator/internal/testing/testdata/platform/expected/platformagent.yaml` now contains exactly four container entries.
> 
> `AGENTS.md` requires that "Docs describe the current state of `main`" and makes a docs-drift pass a required pre-PR step, noting that `make docs-check` only enforces the mechanical subset (generated regions, links, terminology, map coverage) and so would not catch a stale count in prose. The PR corrected four documents that described the watcher as its own container; this is a fifth. Suggested fix: "runs four containers".
> 
> https://github.com/gke-labs/kube-agents/blob/e4fd240e64bb44e91b29ed96e8957af08b5564a3/INSTALL.md#L400-L405
> 
> Automated review by [kube-agents-bot](https://github.com/bradhoekstra/kube-agents-bot).
> 
> React 👍 if this was useful, 👎 if it was not. Add the `agent:ignore` label to opt this pull request out.

done

deploy/shared/envoy-credential-sidecar.sh
```diff
@@ -16,4 +23,20 @@ runtime_pid=$!
...
 
+# Any arguments to this script are the k8s-event-watcher's. It lives here because
+# it authenticates to cluster API servers and credentials belong in this container,
+# not beside the agent sandbox. It is supervised rather than waited on: event
+# watching is best-effort observability, so its failure must not take down Envoy
+# or the credential server and cut the agent off from every credentialed command.
+if [[ $# -gt 0 ]]; then
+  (
+    while true; do
+      /usr/local/bin/k8s-event-watcher "$@" || true
```
haoxuw: 2 concerns

1. assuming arguments forwarded to an inner binary is a confusing convention, would be confusing for readability
2. I think we should not invoke k8s-event-watcher in this script. We can ether create a new .sh and run it independently. Or we create a overall script and invoke envoy/watcher ( + any other future services ). In anyway, watcher and envoy should stay as sister-level.

haoxuw: Right now the script is kicked off by
	return corev1.Container{
		Name:            "envoy-credential-proxy",
		Image:           image,
		ImagePullPolicy: pullPolicy,
		Command:         []string{"/usr/local/bin/envoy-credential-sidecar"},

I think we can create a combined entry point, something like:
Command:         []string{"/usr/local/bin/start-services"}

docs/site/src/content/docs/concepts/observability.md
```diff
@@ -14,7 +14,7 @@ The Platform Agent (Hermes) Deployment exports OpenTelemetry traces, and LiteLLM
...
 
-The Platform Agent (Hermes) Deployment does **not** expose a Prometheus `/metrics` endpoint — it serves only the API (`8642`) and Dashboard (`9119`) ports. Its runtime signals surface as OpenTelemetry traces (below) and `tool_call_audit` log records; pod-level CPU/memory is available through the Kubernetes metrics API (`kubectl top`). The `event-watcher` sidecar can expose watcher metrics (`k8s_event_watcher_*`) via a `--metrics-addr` flag, but this is disabled by default in the shipping deploy.
+The Platform Agent (Hermes) Deployment does **not** expose a Prometheus `/metrics` endpoint — it serves only the API (`8642`) and Dashboard (`9119`) ports. Its runtime signals surface as OpenTelemetry traces (below) and `tool_call_audit` log records; pod-level CPU/memory is available through the Kubernetes metrics API (`kubectl top`). The event watcher, which runs inside the `envoy-credential-proxy` sidecar, can expose watcher metrics (`k8s_event_watcher_*`) via a `--metrics-addr` flag, but this is disabled by default in the shipping deploy.
```
haoxuw, line 17: I don't think there any immediate security concerns. However, could we add documentation that mandate: Watcher should never return logs that could contain security credential or tokens.

Just to be future proof for changes

haoxuw (changes requested):

docs/site/src/content/docs/concepts/observability.md
stalhaali: done

deploy/shared/envoy-credential-sidecar.sh
stalhaali: done

haoxuw (approved):
LGTM

deploy/shared/start-services.sh
```diff
@@ -0,0 +1,77 @@
...
+    while true; do
+      /usr/local/bin/k8s-event-watcher \
+        --cluster-name="${EVENT_WATCHER_CLUSTER_NAME:-}" \
+        --profiles-dir="${CREDENTIAL_PROXY_WORKSPACE_ROOT:-/opt/data}/profiles" \
+        --in-cluster \
+        --daemon-url=http://127.0.0.1:8699 \
+        --token-env=API_SERVER_KEY \
+        --owner=platform \
+        --reason=Failed,FailedToDrainNode,CrashLoopBackOff,BackOff,ImagePullBackOff,ErrImagePull,OOMKilled || true
+      echo "start-services: k8s-event-watcher exited, restarting in 10s" >&2
```
jayantid: This loop makes a permanently broken watcher indistinguishable from a healthy one.

  As a standalone container, a watcher that couldn't start showed up as CrashLoopBackOff.
  Here it retries every 10s forever, and the readiness probe only curls Envoy's /healthz —
  nothing checks the watcher. The Pod stays Ready and the only symptom is that triage
  sessions quietly stop firing, which looks exactly like "no incidents happened."

  One way in: if /opt/data/profiles doesn't exist, the watcher treats that as fatal and
  exits before it ever sets up the --in-cluster watch (main.go:319, 769-773) — so we lose
  the management cluster too, not just the fleet. That dir is created by the platform-agent
  entrypoint, which is allowed to fail with just a warning (docker-entrypoint.sh:64). The
  10s retry handles the normal startup race between the two containers, but not that.
  And with no backoff, an RBAC gap or API throttling gets hammered every 10s forever.

  Suggested fix:

  1. Surface the failure — bump a restart counter / set an up gauge on the existing metrics
     registry, or have the supervisor touch a file the readiness probe checks after N
     consecutive failures. Anything visible outside the log.
  2. Exponential backoff, capped at a minute or two, instead of flat 10s.
  3. Don't let a missing profiles dir cost us the in-cluster watch — either mkdir -p it here
     before launching, or have buildWatchSet fall back to the --in-cluster cluster when
     discovery fails. Watching one cluster beats watching none.

  (3) removes the likeliest trigger, but (1) is the one I wouldn't ship without: the current
  failure mode is silence, and silence is the one signal we can't act on.

jayantid (changes requested):

deploy/shared/start-services.sh
stalhaali: Applied 2. 

1 and 3 will be follow ups.

For 1. The watcher already tracks whether each cluster is being watched, cluster_up metric. It's just switched off: nothing passes --metrics-addr, so the watcher never opens the endpoint and nothing can read it.

 So the solution is to turn it on and point Prometheus at it. Then:
  
  - A cluster stops being watched → cluster_up for that cluster goes to 0.
  - The watcher dies entirely → the endpoint disappears with it, and Prometheus reports the
  target as down.


For 3. Agreed that watching one cluster beats watching none. Not falling back immediately though: discovery runs once, so on the normal startup race that would silently leave us on the management cluster forever. Instead we'll retry the missing profiles dir for a bounded window, then fall back to the in-cluster watch with an `ALERT`.

jayantid (approved):

jayantid (approved):

Issue #500, opened 2026-07-31
Private-registry overrides for third-party integration images (LiteLLM, GitHub token minter)


mastersingh24:
## Problem

The custom-registry work covers the images this repo publishes (k8s-operator, platform-agent, credential-proxy, replay-proxy) plus the fluent-bit sidecar, but the optional integrations still hardcode third-party images that behind-the-firewall clusters cannot pull:

- `k8s-operator/config/integrations/litellm/base/deployment.yaml` — `ghcr.io/berriai/litellm:v1.92.0`
- `k8s-operator/config/integrations/github/deployment.yaml.template` — `us-docker.pkg.dev/abcxyz-artifacts/docker-images/github-token-minter-server:v2.7.1-amd64`

## Proposal

Parameterize these the same way the replay proxy already is (`REPLAY_IMAGE` via envsubst in `make deploy-inference-replay`): e.g. `LITELLM_IMAGE` and `TOKEN_MINTER_IMAGE` variables consumed by the deploy targets and provisioning steps (09, 10), defaulting to the current upstream references. Document them in the "Private / custom registry" section of `docs/site/src/content/docs/deploy/docker-images.md`.

## Context

Follow-up from the custom-registry work; these integrations are optional, so the base install works in a private-registry environment without this, but LiteLLM in particular is on the default provisioning path when a model gateway is used.

Issue #499, opened 2026-07-31
Support imagePullSecrets for agent workloads pulling from authenticated private registries


mastersingh24:
## Problem

PR support for private/custom registries (REGISTRY_PREFIX, PLATFORM_AGENT_IMAGE / CREDENTIAL_PROXY_IMAGE / FLUENT_BIT_IMAGE overrides) makes image *names* configurable, but the operator does not manage `imagePullSecrets` anywhere: neither the `PlatformAgent` CRD (`spec.deployment`) nor the generated Deployment/StatefulSet pod specs expose them.

Behind-the-firewall customers whose private registry requires authentication currently have to rely on node-level pull credentials (e.g. GKE service-account scopes / Artifact Registry IAM) or a pull-through cache. Registries like Harbor or Artifactory with token auth cannot be used directly.

## Proposal

- Add `spec.deployment.imagePullSecrets` ([]LocalObjectReference) to the `PlatformAgent` CRD, wired into the generated pod specs (main containers, init containers, and sidecars all run from the same pull identity).
- Optionally an operator-wide default (env var or flag on the controller manager) so fleet installs don't have to set it on every CR, mirroring the image-override env vars.

## Context

Follow-up from the custom-registry work; the limitation is documented in `docs/site/src/content/docs/deploy/docker-images.md` ("Private / custom registry").

bradhoekstra:
Implemented in #806 (05a04fd, "feat(operator): support imagePullSecrets for agent workloads"), which landed both halves this issue asked for:

- `spec.deployment.imagePullSecrets` on the CRD — `k8s-operator/api/v1alpha1/common_types.go:343`, `[]corev1.LocalObjectReference`, wired into the generated pod specs.
- An install-wide default — `charts/kube-agents/values.yaml:49`, `imagePullSecrets: []`, with `spec.deployment.imagePullSecrets` replacing rather than adding to it (noted at `values.yaml:43`).

Closing as fixed. @mastersingh24 — if the Harbor/Artifactory path still does not work in practice, please reopen with what failed.

PR #501, opened 2026-07-31
feat(install): make container image registry configurable

mastersingh24:
# Pull Request

## Why This Change

Today the install assumes all kube-agents containers come from the static public registry (`ghcr.io/gke-labs/kube-agents`). Behind-the-firewall customers need to mirror the images into their own registry and point the whole install at it. Two images had no override path at all: the operator's default agent image was a compile-time Go constant, and the fluent-bit logging sidecar was hardcoded to Docker Hub.

## What Changed

**Files:**

- `k8s-operator/internal/controller/manifest_helpers.go`, `platformagent_manifests.go` — the operator reads three optional env vars: `PLATFORM_AGENT_IMAGE` (default agent image when a CR omits `spec.deployment.image`), `CREDENTIAL_PROXY_IMAGE` (explicit sidecar override; derivation from the agent image is unchanged when unset), and `FLUENT_BIT_IMAGE`. Defaults are unchanged when unset.
- `k8s-operator/config/manager/manager.yaml` — documents the env vars for kustomize patching.
- `k8s-operator/scripts/common.sh`, `provision_03/08/11` — new `REGISTRY_PREFIX` variable derives the defaults for `OPERATOR_IMAGE`, `AGENT_IMAGE`, and `REPLAY_IMAGE`; the individual variables still win. `provision_03` forwards the image overrides to the operator Deployment.
- `docs/site/.../deploy/docker-images.md` — new "Private / custom registry" section (canonical home: which images to mirror, the three override layers, precedence).
- `INSTALL.md`, `k8s-operator/scripts/README.md`, `docs/site/.../operator/platformagent-crd.md`, `docs/README.md` — pointers to the canonical section; map updated.
- `platformagent_manifests_test.go` — unit tests for the env overrides and precedence.

**Added/Changed/Fixed:**

- One knob (`REGISTRY_PREFIX`) for scripted installs; per-image variables and per-CR `spec.deployment.image`/`tag` still take precedence.
- Fluent-bit sidecar image is now overridable (previously impossible).
- No CRD/API changes; no behavior change for existing installs.

## Why This Matters

- Unblocks air-gapped / behind-the-firewall deployments that cannot pull from `ghcr.io` or Docker Hub.

## Context

- Complementary to #493 (versioned releases): that PR injects a version tag at build time and adds a Helm chart; this one adds runtime registry overrides. They touch the same lines in `manifest_helpers.go`, so whichever merges second needs a trivial rebase.
- Follow-ups filed: #499 (`imagePullSecrets` for authenticated private registries), #500 (overrides for third-party integration images: LiteLLM, GitHub token minter).

## Testing

- `make -C k8s-operator test` (includes new unit tests for override precedence)
- `make docs-check`, `make validate`, prettier check on changed files
- `docs/site` `npm ci && npm run build`
- `docker build -f deploy/docker/Dockerfile --target platform`
- `kustomize build config/default`
- `review-docs-drift` skill run against the branch; no blocking findings, advisories addressed

---

No functional impact when the new variables are unset: all defaults resolve to today's values, verified by the existing golden-manifest tests.

---

> **Post-merge note (upgrade impact):** `resolveCredentialProxyImage` now derives the sidecar image from the resolved agent image, so on operator upgrade, existing PlatformAgents with a tagged `spec.deployment.image` (e.g. `...:v5`) roll their credential-proxy sidecar from `:latest` to the agent's tag. This deliberately fixes a pre-existing version skew across the credential path (review issue 6) — "no behavior change for existing installs" above does not hold for that case. Remaining review items are addressed in follow-up #509.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -110,6 +110,20 @@ make gcp-provision
 > `cd k8s-operator && make help` for the complete, always-current list of provisioning and teardown
 > targets.
 
+- **Private container registry**: If your clusters cannot pull from `ghcr.io`, mirror the
+  kube-agents images into your own registry and export `REGISTRY_PREFIX` before provisioning:
+
+  ```bash
+  export REGISTRY_PREFIX=registry.example.com/kube-agents
+  make gcp-provision
+  ```
+
+  The prefix replaces `ghcr.io/gke-labs/kube-agents` as the default for the operator, agent, and
+  replay-proxy images (the individual `OPERATOR_IMAGE`, `AGENT_IMAGE`, and `REPLAY_IMAGE`
+  variables still win). See the
+  [Docker images guide](docs/site/src/content/docs/deploy/docker-images.md) for the full list of
+  images to mirror and the operator-level override env vars.
+
 #### Step 3: Verify Running Components
 
 Verify that the operator, LiteLLM gateway, and custom resources are healthy:
@@ -242,6 +256,18 @@ Install the Custom Resource Definitions (CRDs) and deploy the controller manager
 make install
 make deploy IMG=$IMG
 ```
+
+If the agent images are mirrored into a private registry as well, tell the operator where to
+find them (used whenever a `PlatformAgent` CR does not set `spec.deployment.image`):
+
+```bash
+kubectl set env deployment/kubeagents-controller-manager -n kubeagents-system \
+  PLATFORM_AGENT_IMAGE=registry.example.com/kube-agents/platform-agent:latest \
+  FLUENT_BIT_IMAGE=registry.example.com/mirror/fluent-bit:5.0.7
+```
+
+See the [Docker images guide](docs/site/src/content/docs/deploy/docker-images.md) for all
+override env vars and their precedence.
 
 Verify controller readiness:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -109,17 +109,19 @@ PR that touches one of these files as a change to documented identifiers and
 uses this table to find what to re-verify; when a new category of documented
 identifier appears, add its source here.
 
-| Identifier                                                | Source of truth                                                                        |
-| --------------------------------------------------------- | -------------------------------------------------------------------------------------- |
-| Service-account names, namespace, permission-set defaults | `k8s-operator/scripts/common.sh`                                                       |
-| Go toolchain version                                      | `k8s-operator/go.mod`                                                                  |
-| Toolsets, plugins, and MCP servers of an agent profile    | that profile's `config.yaml` (`agents/platform/`, `agents/chat/`, `agents/cluster/`)   |
-| Cron job rosters and schedules                            | the profile's cron `jobs.json` (`agents/platform/cron/`, `agents/chat/defaults/cron/`) |
-| Persona rules and `§N` section numbering                  | the profile's `SOUL.md`                                                                |
-| RBAC bindings and KSA defaults laid down per agent        | `k8s-operator/internal/controller/platformagent_manifests.go`                          |
-| Controller permissions                                    | `k8s-operator/config/rbac/`                                                            |
-| `make` targets                                            | the root `Makefile` and `k8s-operator/Makefile`                                        |
-| Paths baked into the agent image (`/opt/defaults/...`)    | `deploy/docker/Dockerfile`                                                             |
+| Identifier                                                           | Source of truth                                                                        |
+| -------------------------------------------------------------------- | -------------------------------------------------------------------------------------- |
+| Service-account names, namespace, permission-set defaults            | `k8s-operator/scripts/common.sh`                                                       |
+| Go toolchain version                                                 | `k8s-operator/go.mod`                                                                  |
+| Toolsets, plugins, and MCP servers of an agent profile               | that profile's `config.yaml` (`agents/platform/`, `agents/chat/`, `agents/cluster/`)   |
+| Cron job rosters and schedules                                       | the profile's cron `jobs.json` (`agents/platform/cron/`, `agents/chat/defaults/cron/`) |
+| Persona rules and `§N` section numbering                             | the profile's `SOUL.md`                                                                |
+| RBAC bindings and KSA defaults laid down per agent                   | `k8s-operator/internal/controller/platformagent_manifests.go`                          |
+| Controller permissions                                               | `k8s-operator/config/rbac/`                                                            |
+| `make` targets                                                       | the root `Makefile` and `k8s-operator/Makefile`                                        |
+| Paths baked into the agent image (`/opt/defaults/...`)               | `deploy/docker/Dockerfile`                                                             |
+| Image defaults and override env vars (`PLATFORM_AGENT_IMAGE` et al.) | `k8s-operator/internal/controller/manifest_helpers.go`                                 |
+| Registry prefix default (`REGISTRY_PREFIX`)                          | `k8s-operator/scripts/common.sh`                                                       |
 
 ## 3. Documentation eras and status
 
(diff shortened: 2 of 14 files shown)


mastersingh24:
Both review findings addressed:

- **Credential-proxy tag skew (Medium)** — fixed in 2bf6352. `resolveCredentialProxyImage` now derives the sidecar from the *resolved* agent image (`resolveAgentImage(deployment, defaultPlatformAgentImage())`), so it always shares the agent container's registry and tag; the `deployment.Tag` override branch that beat the derived tag is gone. The reviewed scenario (`PLATFORM_AGENT_IMAGE=…:v0.5.0` + `deployment: {replicas: 2}` with the CRD-defaulted `tag: latest`) now yields `credential-proxy:v0.5.0`, and there's a regression test for exactly that case, plus cases for an embedded tag beating the defaulted tag field and for digest-pinned agent images (the digest can't name the proxy image, so the tag field / `latest` is used, as before).
- **Unrelated `package-lock.json` churn (Low)** — reverted to main in c8147a0. It was a byproduct of a local npm run during pre-submit validation.

mplakhtiy:
## Review — `feat(install): make container image registry configurable`

### Issues

#### 1. `PLATFORM_AGENT_IMAGE` is unreachable on every install path this repo ships

The env var only applies when a CR omits `spec.deployment.image` (`platformagent_manifests.go:636`). Every CR the repo ships sets it explicitly:

| Path | Sets `spec.deployment.image`? |
|---|---|
| `k8s-operator/scripts/platform-agent.yaml.template:23` | `"${AGENT_IMAGE}"` |
| `k8s-operator/examples/platformagent.yaml:41` | `"ghcr.io/gke-labs/kube-agents/platform-agent"` — hardcoded |
| `docs/site/.../operator/index.md:62` | `ghcr.io/gke-labs/kube-agents/platform-agent` — hardcoded |

So the `kubectl set env` block added to `provision_03:106-125` sets a variable that nothing on the scripted path consults — `provision_08`'s template always overrides it. And a behind-the-firewall user who copy-pastes the example CR or the docs' CR gets `ghcr.io` regardless of every knob this PR adds. Neither file is updated.

This is the core gap: the feature's headline use case (an air-gapped install) still fails on the two most likely copy-paste entry points. Suggest parameterising `examples/platformagent.yaml` (or at minimum a comment pointing at the override), and either dropping the hardcoded image from `operator/index.md`'s example or noting the override there.

#### 2. `REGISTRY_PREFIX` isn't persisted, so a second run applies it only halfway

Every other knob goes through `init_var` → `save_var` → `vars.sh`. `REGISTRY_PREFIX` is read as a bare `${REGISTRY_PREFIX:-}` and never saved. Because `load_state` sources `vars.sh` *before* `init_var` runs, adding the prefix after a first install is silently a no-op for the derived defaults:

```bash
make gcp-provision                                # run 1: vars.sh gets OPERATOR_IMAGE=ghcr.io/...
export REGISTRY_PREFIX=registry.corp/kube-agents  # decide to move to a mirror
make gcp-provision                                # run 2
```

Run 2: `init_var "OPERATOR_IMAGE"` sees the saved `ghcr.io` value and skips the prefix entirely — same for `AGENT_IMAGE` and `REPLAY_IMAGE`. But the new block in `provision_03` reads `REGISTRY_PREFIX` *directly*, so it fires and sets `PLATFORM_AGENT_IMAGE=registry.corp/kube-agents/platform-agent:$IMAGE_TAG`. You land on a half-migrated install — operator and agent still pulling from `ghcr.io`, one env var pointing at the mirror — with no warning. `scripts/README.md` (which this PR edits) promises the opposite: *"running the script again will reuse the same options selected on the first run."*

Fix: route it through `init_var` like everything else.

Related: `kubectl set env` defaults to field manager `kubectl-set` (confirmed in `kubectl set env --help`, client v1.36.3), while `make deploy` uses `kubectl apply --server-side` under manager `kubectl`. Since `env` is an associative list keyed on `name`, the apply won't prune the `kubectl-set`-owned entry — so the value is sticky and can only be removed by hand. Worth knowing; it also means uncommenting the `manager.yaml` block later would hit an SSA conflict, and `make deploy` has no `--force-conflicts`.

#### 3. `AGENT_IMAGE` alone doesn't reach the operator

`provision_03:110-114` derives `PLATFORM_AGENT_IMAGE` from `REGISTRY_PREFIX` only, never from `AGENT_IMAGE`. The docs spell this out (*"`AGENT_IMAGE` itself only feeds the CR rendered in `provision_08`"*), which documents a footgun rather than removing it. Given issue #1 this is currently latent, but it becomes live the moment a hand-written CR omits the image. Cheap to fix by adding an `elif [ -n "${AGENT_IMAGE:-}" ]` branch (guarding against `AGENT_IMAGE` already carrying a tag).

#### 4. The new docs section overstates coverage

> "point every layer of the install at it"

Three images in the install path have no override and aren't listed among the images to mirror:

- **cert-manager** — `provision_03:70` does `kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.14.4/cert-manager.yaml`, pulling quay.io images, in the very script this PR edits. Behind a firewall this step fails before the operator is ever deployed.
- `ghcr.io/berriai/litellm:v1.92.0` (`config/integrations/litellm/base/deployment.yaml:22`)
- `us-docker.pkg.dev/abcxyz-artifacts/.../github-token-minter-server:v2.7.1-amd64` (`config/integrations/github/deployment.yaml.template:31`)

The last two are acknowledged in follow-up #500, but a reader of this page has no way to know the install still won't complete. cert-manager isn't covered by #499 or #500 at all. A short "not yet covered" list would make the page honest without expanding scope.

#### 5. Digest-pinned agent images silently fall back to a mutable `:latest` sidecar

`platformagent_manifests.go:1274-1281`: for `image: example/platform-agent@sha256:…` with no explicit `tag`, the proxy resolves to `example/credential-proxy:latest`. The old behaviour (`credential-proxy@sha256:<agent digest>`) was also wrong but failed loudly at pull time. The new path is a silent supply-chain downgrade — an unpinned sidecar that shares the agent's credential path — with nothing in the CR status, events, or logs to say the pin was dropped. A `log.Info` at minimum; a webhook rejection of digest-without-tag would be better.

#### 6. "No behavior change for existing installs" isn't accurate — and the real change is bigger than it looks

`spec.deployment.tag` is defaulted to `latest` in **two** places: `config/crd/bases/…platformagents.yaml:8130` (`default: latest`) and `internal/webhook/platformagent_webhook.go:78-79`. So essentially every CR that reaches the API server has a non-empty `Tag`.

Under the old code, `if deployment.Tag != nil && *deployment.Tag != ""` returned early and won unconditionally. That means any CR written as `image: myregistry/platform-agent:v5` (tag left to default) ran the **agent** at `v5` and the **credential proxy** at `latest` — a pre-existing version skew across a security-sensitive sidecar boundary. The new code fixes it, and the test at `platformagent_manifests_test.go:999` pins the behaviour.

This is a good fix and arguably the most valuable change in the PR — but it will roll the sidecar for every existing agent with a tagged `image:` on upgrade, and it deserves to be called out in the PR body rather than described as no-change. The golden manifest doesn't catch it (`internal/testing/testdata/platform/expected/platformagent.yaml` leaves `deployment.image` unset), so the new unit tests are the only coverage — worth a golden case with a tagged image.

### Nits

- **`REGISTRY_PREFIX` name collision.** `scripts/release/verify_candidate_images.sh:15` already defines `REGISTRY_PREFIX` with different semantics. It's CI-only so a local export can't reach it, but the new `docs/README.md:124` row claiming `common.sh` as *the* source of truth for this identifier is now only half true. Either cross-reference both or rename one.
- **`fluent-bit:5.0.7` duplication.** Pre-PR the pin lived in 4 places (constant, 2 test lines, golden). The PR adds it to 3 more prose/comment locations: `INSTALL.md:266`, `manager.yaml:60`, `docker-images.md:74`. `AGENTS.md` explicitly warns against restating machine-readable values — the docs need to *name* the image but not pin the version; `fluent/fluent-bit` (version per `manifest_helpers.go`) reads the same.
- **`os.Getenv` inside `resolveCredentialProxyImage`** makes a previously pure function impure, which is why the new tests need `t.Setenv`. Reading the three vars once at manager startup into a small config struct would keep these as plain table tests and match the kubebuilder convention.
- **`imagePullSecrets` guidance.** The docs recommend node-level pull credentials, but `provision_03:65` branches on GKE Autopilot, where that isn't available. Better to say plainly that authenticated registries aren't supported yet (#499).
- **`registry_prefix()`** strips a trailing `/` but doesn't reject a scheme — `REGISTRY_PREFIX=https://registry.corp/x` produces an unpullable ref with no diagnostic.
- **Precedence list, layer 3** says the CR "overrides the defaults above," but `FLUENT_BIT_IMAGE` has no CR-level equivalent.
- **Tag-without-image is silently ignored.** `resolveAgentImage:79` only applies `Tag` when `Image != ""`, so a CR setting only `tag: v9` gets `PLATFORM_AGENT_IMAGE`'s tag. Pre-existing, but this PR makes it more likely to be hit.
- **No script-level test coverage.** `make validate` (root `Makefile:70`) only checks skill directory placement; the shell branch where issues 2 and 3 live isn't exercised by anything.

### Recommendation

Issue #1 is the blocker — as it stands, the shipped install paths can't actually use the operator-level override the PR adds. Issues #2 and #3 are small shell fixes. #4 and #6 are documentation. #5 is a log line.

---

*This review was generated with the help of Claude.*

eLeontev (approved):

mplakhtiy (approved):

mplakhtiy:
Followup pr with fixes https://github.com/gke-labs/kube-agents/pull/509

PR #505, opened 2026-08-02
fix(platform): repair the credential-free Slack path

adamparco:
# Pull Request

## Why This Change

Slack was dead on the credential-free path. A message sent to the Platform Agent produced no reply and no obvious error — the pod looked healthy, the relay reported an authenticated workspace, and the bootstrap call returned 200. Nothing surfaced because the failures happen after startup, on the inbound request, and the gateway swallows them.

Four distinct defects were stacked, each one hidden behind the one ahead of it. Fixing any single one just moved the silence a few seconds later.

## What Changed

**Files:**

**Added/Changed/Fixed:**

1. **`SlackRelay.api_call` mangled every response.** It built its reply with `dict(response)`. A `SlackResponse` defines `__iter__` but no `keys()`, so `dict()` treated it as a sequence of pairs and choked on the first key — every relayed call came back HTTP 400. Now reads the payload off `.data`, falling back to the response itself so a plain mapping still works.

2. **The adapter bootstrapped before the sidecar existed.** Regular sidecars have no start-ordering guarantee, and the agent container was observed winning by 5–49s. The resulting connection error was fatal rather than merely slow: the bot token lives in the relay, so the gateway's reconnect found no bot credential on the queued config and dropped Slack from the retry queue **for the life of the pod**. New `wait_for_relay` absorbs it.

3. **Surviving the refused connection wasn't enough.** `credential_proxy.py` binds its HTTP port before Slack authentication completes — relay setup runs on a daemon thread so a Slack outage can't wedge the whole proxy — and answers `503 Slack relay disabled` until it finishes. `wait_for_relay` treats 503 as the readiness signal it is and retries; every other status still raises, because the relay answering is not a readiness problem. Bounded by `SLACK_RELAY_READY_TIMEOUT_SECONDS` (default 120s).

4. **Bolt never used the relay client.** `AsyncApp._init_context` does not reuse the client the app was built with — it constructs a fresh `AsyncWebClient` per request from the name bound in `slack_bolt.app.async_app`. Rebinding only the adapter module's copy left every request calling `slack.com` directly holding the `"relay:"` placeholder for a token, which failed authorization with `invalid_auth`. The patch now rebinds the name Bolt actually reads. It has to stay a class, not a factory — `AsyncApp` isinstance-checks the client it is handed.

5. **And then the response type.** With requests finally routed through the relay, Bolt's authorization middleware died with `'dict' object has no attribute 'headers'`: `_to_authorize_result` reads `x-oauth-scopes` off the response. `RemoteSlackClient.api_call` now returns a real `AsyncSlackResponse`. The relay does not forward Slack's response headers, so scope introspection sees none.

Also: `RemoteSlackClient` now takes the `team_id` Bolt resolved from the inbound event, rather than splitting it out of the token. Across several workspaces that token is a comma-joined list, so splitting yields every team at once instead of the one the request belongs to.

## Why This Matters

- Slack is one of two chat front doors. On this path it was completely non-functional, and it failed silently — no listener ever ran, and the only trace was a `Reconnect slack: no bot credential on queued config` line at startup.
- Defects 3 and 4 only fire when a message actually arrives, which is why the pod looks healthy and the smoke tests pass while every message vanishes.
- Defect 2 is unrecoverable without a pod restart. Once the adapter leaves the retry queue it never rejoins.

## Context

Found while testing the fleet-audit cron work end-to-end on a live GKE cluster; the Slack breakage is independent of that feature and pre-exists it on `main`, so it lands on its own branch.

`k8s-operator/internal/controller/platformagent_manifests.go` already injects `SLACK_RELAY_URL=http://127.0.0.1:8765`; no operator or manifest change is needed. The new timeout knob follows the precedent of its sibling `SLACK_RELAY_MAX_FILE_BYTES` — an env-tunable default, not part of the documented surface.

## Testing

- Platform suite: 207 tests, green. Fleet-audit suite: 346 tests, green.
- Two new test classes, `RelayReadinessTest` (6) and `BoltClientPatchTest` (6). The Bolt-client binding had **zero** coverage despite carrying four fatal bugs. `BoltClientPatchTest` stubs `slack_bolt`/`slack_sdk` into `sys.modules`, so it runs without either package installed — matching what CI has available.
- Confirmed the new tests actually catch the regression rather than just passing: replacing the `bolt_async_app.AsyncWebClient = RemoteSlackClient` line with `pass` fails 3 of 4, and restoring it re-passes.
- Hardened `SlackRelayTest`'s fake, which was a plain `dict` — precisely what let defect 1 ship.
- Verified end-to-end against a live GKE cluster (`platform-agent-host`): relay authenticated, bootstrap 200, and a Slack message now gets a reply, with zero `invalid_auth`, middleware, or traceback errors in the pod log.

---

Behavior change is confined to the Slack path, which was previously non-functional, so the blast radius is bounded by definition. The startup wait is the one thing worth a second look: it is bounded, it only retries the two specific failures that mean "not up yet," and it runs on a thread that was already doing blocking I/O.

### DIFF
--- a/agents/platform/scripts/credential_proxy.py
+++ b/agents/platform/scripts/credential_proxy.py
@@ -374,7 +374,11 @@ class SlackRelay:
         response = self._client(team_id).api_call(
             method, **self._decode_argument(arguments)
         )
-        return dict(response)
+        # WebClient.api_call returns a SlackResponse, which defines __iter__ but
+        # no keys(), so dict() takes it for a sequence of pairs and dies on the
+        # first key. The payload lives on .data; fall back to the response itself
+        # so a plain mapping still works.
+        return dict(getattr(response, "data", response))
 
     def download(self, team_id: str, url: str) -> bytes:
         def is_slack_url(value: str) -> bool:
--- a/agents/platform/scripts/slack_relay_patch.py
+++ b/agents/platform/scripts/slack_relay_patch.py
@@ -8,13 +8,22 @@ import json
 import logging
 import os
 import sys
+import time
+import urllib.error
 import urllib.request
 from pathlib import Path
-from typing import Any
+from typing import Any, Callable
 
 
 LOGGER = logging.getLogger("slack-relay-patch")
 DEFAULT_MAX_FILE_BYTES = 20 * 1024 * 1024
+DEFAULT_RELAY_READY_TIMEOUT = 120.0
+RELAY_READY_POLL_SECONDS = 2.0
+# credential_proxy binds its port before the Slack relay finishes
+# authenticating — relay setup runs on a background thread so a Slack outage
+# cannot wedge the whole proxy — and answers 503 "Slack relay disabled" until
+# then. That is a readiness signal, unlike every other status it can return.
+RELAY_NOT_READY_STATUS = 503
 
 
 def read_upload(path: Path, max_file_bytes: int) -> bytes:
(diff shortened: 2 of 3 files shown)


bradhoekstra (approved):

bradhoekstra:
Superseding with the smaller #515 , since #480 already contained a lot of the changes in this PR.

PR #530, opened 2026-08-04
fix(scripts): give Cloud KMS its own location variable

### LINKED ISSUE #518: GitHub token minter: REGION is used as both the cluster location and the KMS location, so zonal clusters can't provision it
## Context

`provision_10_deploy_github_minter.sh` uses `$REGION` for two things that don't take the same values.  On a zonal cluster they can't both be satisfied, so the token minter can't be provisioned at all.

The rest of the pipeline is fine with a zonal cluster.  `gcloud container clusters ...` accepts a zone wherever the scripts pass `--region`, so `REGION=us-central1-c` works from stage 01 through stage 09.  KMS is where it falls over.

## The conflict

In `config/integrations/github/deployment.yaml.template`:

```
line 50:  .../v1/projects/${PROJECT_ID}/locations/${REGION}/clusters/${CLUSTER_NAME}   # ISSUER_ALLOWLIST
line 57:  projects/${PROJECT_ID}/locations/${REGION}/keyRings/${KMS_KEYRING}/...       # KMS_KEY_NAME
```

Line 50 wants the cluster's location, which for a zonal cluster is `us-central1-c`.  Line 57 wants a KMS location, and there is no such KMS location:

```
$ gcloud kms locations list --filter="locationId~us-central1" --format="value(locationId)"
us-central1
```

`provision_10` has the same problem in the keyring and key calls.  Lines 87, 88, 92, 94, 107, 116 and 165 all pass `--location="${REGION}"`.

With `REGION=us-central1-c` the keyring creation fails.  Set `REGION=us-central1` instead and KMS is happy, but then the `ISSUER_ALLOWLIST` issuer URL names a cluster that doesn't exist, so the minter won't authorize the agent's tokens.  `connect_cluster` would also stop finding the cluster.

## Repro

- zonal cluster, `us-central1-c`, `REGION` set to the zone
- stages 01 through 09 pass
- stage 10 fails creating the keyring

## Suggested fix

Give KMS its own variable defaulted from `REGION`, so the regional case is unchanged.  Something like `KMS_LOCATION`, used in place of `REGION` in the seven `gcloud kms` calls and on line 57 of the template.  Deriving it by stripping the zone suffix (`us-central1-c` -> `us-central1`) would probably cover most cases, but an explicit var is easier to reason about.

I haven't tried either fix yet - I worked around it by deferring the minter.

Related: `provision_01` creates regional clusters, so this only shows up when pointing the pipeline at a pre-existing zonal cluster.

bnaylor:
## Why This Change

`provision_10_deploy_github_minter.sh` uses `$REGION` for two things that don't accept the same values, so the GitHub token minter can't be provisioned against a zonal cluster at all.

`REGION` is the GKE cluster location, which for a zonal cluster has to stay the zone.  It's also passed to every `gcloud kms` call and into `KMS_KEY_NAME`, and Cloud KMS has no zonal locations:

```
$ gcloud kms keyrings list --location=us-central1-c
ERROR: NOT_FOUND: The request concerns location 'us-central1-c' ... Cloud KMS is not available in 'us-central1-c'

$ gcloud kms keyrings list --location=us-central1
Listed 0 items.
```

So with `REGION=us-central1-c` the keyring creation fails.  Set `REGION=us-central1` instead and KMS is satisfied, but `ISSUER_ALLOWLIST` then names a cluster that doesn't exist, the minter won't authorize the agent's tokens, and `connect_cluster` stops finding the cluster.

Fixes #518.

## What Changed

Adds a `KMS_LOCATION` variable, defaulted by stripping the zone suffix from `REGION`, and uses it everywhere Cloud KMS is addressed.  The cluster location carries on using `REGION`.

**Files:**

- `k8s-operator/scripts/provision_10_deploy_github_minter.sh`
- `k8s-operator/scripts/teardown_10_deploy_github_minter.sh`
- `k8s-operator/config/integrations/github/deployment.yaml.template`
- `k8s-operator/Makefile`

**Added/Changed/Fixed:**

- `derive_kms_location()` and `init_var_kms_location()` in `common.sh`.  `us-central1-c` becomes `us-central1`, and a value that is already a region comes back unchanged.
- The seven `gcloud kms` calls in `provision_10` now pass `--location="${KMS_LOCATION}"`, as do the minty `import-pk` invocation and the manual-import hint it prints on failure.
- `KMS_KEY_NAME` in `deployment.yaml.template` uses `${KMS_LOCATION}`.  `ISSUER_ALLOWLIST` still uses `${REGION}`, since that one really is the cluster location.
- `KMS_LOCATION` added to the `envsubst` allowlists for `deploy-github` and `undeploy-github`, and to the exports in both stage 10 scripts.
- `teardown_10` derives the value when it is missing.  A `vars.sh` written before this change won't have it, and that script runs under `set -u`.

## Why This Matters

Regional clusters are unaffected.  The derived value equals `REGION`, so nothing changes for an install that used `provision_01` to build its own cluster.  This only matters when pointing the pipeline at a pre-existing zonal cluster, which is a hard failure today.

## Context

Found while installing against an existing zonal GKE cluster.  Filed as #518.  Wider write-up of that install: https://gist.github.com/bnaylor/b8e7b7f9dda75c6f159815b36aec3d49

## Testing

Verified:

- `bash -n` on all three changed scripts
- `derive_kms_location` across zonal and regional inputs.  `us-central1-c -> us-central1`, `europe-west1-b -> europe-west1`, `us-east4` and `asia-northeast1` unchanged
- the rendered manifest carries both locations correctly at the same time, which wasn't possible before:
  ```
  locations/us-central1-c/clusters
  locations/us-central1/keyRings
  ```
- Cloud KMS rejects a zone and accepts the enclosing region, per the output above

Not verified:

- an end-to-end stage 10 run.  That needs a GitHub App (org, repo, app ID, private key PEM) and a live zonal cluster, and I had neither by the time I got here.  The PEM import path and the minter's runtime behaviour aren't touched by this change, but they haven't been exercised with it either.

---

Low risk: no behaviour change for regional clusters, and stage 10 is skipped entirely unless `GITHUB_ORG`, `GITHUB_REPO` and `GITHUB_APP_ID` are all set.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -290,6 +290,7 @@ make deploy-litellm
 export PROJECT_ID="your-gcp-project-id"
 export REGION="your-gcp-region"
 export CLUSTER_NAME="your-gke-cluster-name"
+export KMS_LOCATION="your-kms-region" # a region; Cloud KMS has no zonal locations
 export KMS_KEYRING="your-kms-keyring"
 export KMS_KEY="your-kms-key"
 export KMS_KEY_VERSION="your-kms-key-version"
--- a/k8s-operator/Makefile
+++ b/k8s-operator/Makefile
@@ -124,11 +124,11 @@ undeploy-inference-replay: kustomize ## Undeploy Inference Replay proxy integrat
 
 .PHONY: deploy-github
 deploy-github: kustomize ## Deploy GitHub integration.
-	$(KUSTOMIZE) build config/integrations/github | envsubst '$$NAMESPACE $$PROJECT_ID $$REGION $$CLUSTER_NAME $$KMS_KEYRING $$KMS_KEY $$KMS_KEY_VERSION $$GITHUB_ORG $$GITHUB_REPO $$GITHUB_MINTER_KSA_NAME $$GITHUB_MINTER_GSA_NAME $$PLATFORM_AGENT_GSA_NAME' | kubectl apply -n $${NAMESPACE:-kubeagents-system} -f -
+	$(KUSTOMIZE) build config/integrations/github | envsubst '$$NAMESPACE $$PROJECT_ID $$REGION $$CLUSTER_NAME $$KMS_KEYRING $$KMS_KEY $$KMS_KEY_VERSION $$KMS_LOCATION $$GITHUB_ORG $$GITHUB_REPO $$GITHUB_MINTER_KSA_NAME $$GITHUB_MINTER_GSA_NAME $$PLATFORM_AGENT_GSA_NAME' | kubectl apply -n $${NAMESPACE:-kubeagents-system} -f -
 
 .PHONY: undeploy-github
 undeploy-github: kustomize ## Undeploy GitHub integration.
-	$(KUSTOMIZE) build config/integrations/github | envsubst '$$NAMESPACE $$PROJECT_ID $$REGION $$CLUSTER_NAME $$KMS_KEYRING $$KMS_KEY $$KMS_KEY_VERSION $$GITHUB_ORG $$GITHUB_REPO $$GITHUB_MINTER_KSA_NAME $$GITHUB_MINTER_GSA_NAME $$PLATFORM_AGENT_GSA_NAME' | kubectl delete --ignore-not-found=$(ignore-not-found) -n $${NAMESPACE:-kubeagents-system} -f -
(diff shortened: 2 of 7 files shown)


bnaylor:
Confirmed and fixed in 290c600.

Reproduced it first, with `KMS_LOCATION` unset and everything else from the documented list:

```
value: projects/proj/locations//keyRings/kr/cryptoKeys/k/cryptoKeyVersions/1
```

So the finding is accurate, including the consequence.  `envsubst` substitutes the empty string rather than leaving the placeholder, `kubectl apply` succeeds, and Minty then signs against a malformed resource name.  `REGION` used to fill that slot and was already documented, so this was a regression I introduced.

Added `export KMS_LOCATION` to both lists:

- `INSTALL.md` Step 5
- `k8s-operator/README.md` step-by-step deployment

Also added a line to the README explaining that `KMS_LOCATION` is the Cloud KMS location and differs from `REGION` only for zonal clusters, since the two being separate is the whole point of the change.

Following the documented list now renders both correctly:

```
locations/us-central1-c/clusters
locations/us-central1/keyRings
```

`make docs-check` passes.

bradhoekstra (approved):

PR #533, opened 2026-08-05
fix(scripts): fail the step when provisioning actually failed

### LINKED ISSUE #522: Provisioning steps report success after the underlying command failed
## Stage 04

`gcloud` printed a hard error and the step still passed:

```
ERROR: (gcloud.iam.service-accounts.add-iam-policy-binding) INVALID_ARGUMENT: Identity Pool does not exist (PROJECT.svc.id.goog).
  i Removing legacy sandbox Workload Identity binding from platform-agent-sandbox...
  v Successfully executed.
```

The pipeline carried on green and finished with "Controller & Agent GCP Permissions Configured Successfully!".  The agent had no working identity.

Failing to detect a missing prerequisite is one problem.  Reporting a failed step as successful is a worse one, because it takes away the last chance to notice.

## Stage 08

Quieter version of the same thing.  `provision_08` applies the `PlatformAgent` CR, prints "PlatformAgent Custom Resource applied successfully to GKE!" and exits 0.  It never waits for the agent to become Ready.

The CR's own status was accurate the whole time:

```
phase: Degraded
conditions: [{type: Ready, status: "False", reason: PodInitializing}]
```

So a completely broken agent looks like a clean install at the CLI.

## Suggested fix

Stage 04 should fail the step when the binding command fails.  Stage 08 should wait for the CR to go Ready before declaring success, or at minimum print `status.phase`.  Stage 03 already does the right thing with `wait_for_k8s_resource` on the controller deployment.

bnaylor:
## Why This Change

Two stages report success after the work underneath them failed.

### Stage 04

`execute_platform_agent` calls `execute_agent_iam` and discards the result:

```bash
execute_agent_iam "Platform Agent" "${PLATFORM_AGENT_KSA_NAME}" "${PLATFORM_AGENT_GSA_NAME}" "${roles[@]}"
```

The legacy sandbox cleanup that follows ends in `|| true`, so that trailing command decides the function's exit status.  `execute_agent_iam` already returns 1 on every failure path - creating the GSA, adding a role binding, binding Workload Identity - and all of it was being thrown away.

Running the real function body with a stubbed `execute_agent_iam` that fails:

```
before: execute_platform_agent -> 0     <- the bug
after:  execute_platform_agent -> 1
```

What that looked like in practice: `gcloud` printed `INVALID_ARGUMENT: Identity Pool does not exist`, and the stage carried on to print

```
  v Successfully executed.
  >>> Controller & Agent GCP Permissions Configured Successfully!
```

The agent had no working identity.  Nothing surfaced it until the credential sidecar crashlooped four stages later.

Not detecting a missing prerequisite is one problem.  Reporting a failed step as successful is a worse one, because it removes the last chance to notice.

### Stage 08

`execute_custom_resource` applies the CR, prints "PlatformAgent Custom Resource applied successfully to GKE!" and exits 0.  It never waits for anything.  `kubectl apply` succeeding only means the operator accepted the resource.

The CR's own status was accurate the whole time:

```
phase: Degraded
conditions: [{type: Ready, status: "False", reason: PodInitializing}]
```

So a completely broken agent - crashlooping sidecar, unschedulable Pod, unpullable image - looks like a clean install at the CLI.

Fixes #522.

## What Changed

**Files:**

- `k8s-operator/scripts/provision_04_gcp_iam.sh`
- `k8s-operator/scripts/provision_08_deploy_platform_agent.sh`
- `k8s-operator/scripts/provision_13_verify_agent_rollout.sh` (new)
- `k8s-operator/scripts/provision.sh`
- `k8s-operator/Makefile`
- `k8s-operator/README.md`
- `hack/ci-deploy.sh`

**Added/Changed/Fixed:**

- Stage 04 checks the return of `execute_agent_iam`.  `run_step` already turns a non-zero `execute_func` into `Failed to execute step` and `exit 1`, so nothing else was needed.
- Stage 08 now fails unless the operator actually reconciled the apply: the Deployment must exist, and if the apply changed the CR spec, the Deployment generation must move too.  (The CRD has the status subresource, so `kubectl apply` bumps the CR's `metadata.generation` iff the spec changed.)  It deliberately does not wait for the rollout - the agent's model backend, the `litellm` Service, is deployed by stage 09, so on a fresh install the workload cannot become Ready during stage 08.
- New stage 13 waits for the `platform-agent-gateway` rollout at the end of the pipeline, once everything the agent depends on exists.  On timeout it prints `status.phase`, every condition with its reason and message, per-container ready/restart counts, and recent container logs, then fails.
- `hack/ci-deploy.sh` calls stage 13 instead of hand-rolling the same creation wait and rollout gate.
- Both waits derive their budget from `AGENT_READY_TIMEOUT` (default 600s), which replaces the earlier 60s cap on Deployment creation.  A bare integer gets an implied `s` suffix rather than surfacing kubectl's duration parse error as a rollout failure.
- `kubectl apply` is now `|| return 1` rather than relying on it being the last statement.
- The stage banners record the behavior; `make docs-generate` propagated them into the `provisioning-steps` region in `scripts/README.md`.

## Why This Matters

The whole value of a staged, idempotent installer is that a green run means the cluster is in the expected state.  Both of these break that, and they break it in the direction that costs the most: you find out much later, somewhere unrelated.

Stage 13 is now the last gate the pipeline runs, so its success message is what a first-time installer reads as "done" - and it only prints after the workload is actually rolled out.

## Context

Filed as #522.  Both were hit during a single install against an existing cluster.  Wider write-up: https://gist.github.com/bnaylor/b8e7b7f9dda75c6f159815b36aec3d49

The review findings on this PR drove the current shape: the rollout gate originally sat in stage 08, one stage before the agent's model backend exists, which on a fresh install could burn the whole timeout and abort stages 09-11 under `provision.sh`'s `set -e`.  #534 tracks the operator-side fix (`observedGeneration` on `AgentStatus`) that would let the CR's own condition carry this information directly.

## Testing

- Stubbed `kubectl` and `ensure_k8s_resource_exists`, then ran the extracted `execute_custom_resource` and `execute_agent_rollout` bodies through 14 scenarios: fresh install, no-op re-apply, spec change with and without operator reconcile, apply failure, missing Deployment, healthy and stuck rollout, and the `AGENT_READY_TIMEOUT` parse cases (`600`, `10m`, `1h`, `12x`, unset).  The false-success case - spec changed, operator never reconciled - now returns 1.
- Confirmed from `common.sh` that `run_step` maps a non-zero `execute_func` to `exit 1`, and that `DRY_RUN` returns before `execute_func` runs at all (which is also why the dry-run guards inside the function were dead code, now removed).
- Stage 13 runs clean under `--dry-run --non-interactive`.
- `bash -n` clean on all changed scripts.
- `make docs-check` passes, including the regenerated `provisioning-steps` region.

---

Stage 04 is a one-line correctness fix.  Stages 08 and 13 turn a silent pass into a real check, which is a behaviour change worth reviewing on its own terms.


bnaylor:
Confirmed and fixed in 72af9ad.  Verified both halves of the claim against source before changing anything:

- `platformagent_controller.go` sets `newPhase = "Ready"` when `newDeploymentStatusReadyReplicas > 0`, so the condition is derived purely from the live replica count.
- `PlatformAgentStatus` has no `observedGeneration`.  `AgentPluginStatus` does have one, which makes the omission look accidental rather than deliberate.

So the condition genuinely carries no indication of which spec it describes, and since `verify_custom_resource` returns 1 unconditionally, the re-apply path is the normal one rather than an edge case.  My gate only closed the fresh-install case.

Now gating on `kubectl rollout status` for `deployment/platform-agent-gateway`, which compares the Deployment's own `observedGeneration` against `metadata.generation` and so cannot be satisfied by a stale success.

One gap that suggestion alone leaves: immediately after `kubectl apply` of the CR, the operator may not have translated it into a Deployment change yet, so `rollout status` could report the *previous* rollout complete.  To close that, the script records the Deployment generation before the apply and waits up to 30s for it to be bumped before gating.  A re-apply that changes nothing never bumps it and falls through to confirm the running Deployment is still healthy, so a genuine no-op does not stall.

Residual limitation, which I do not think belongs in this PR: the 30s window is a heuristic.  The robust fix is `observedGeneration` on `PlatformAgentStatus` so the CR's own condition becomes trustworthy — an operator change rather than a scripts one.  Happy to file that as a follow-up if you agree it is worth doing.

The banner and its generated row in `scripts/README.md` were updated to describe the rollout wait rather than the Ready condition.

bnaylor:
Filed the follow-up as #534 — adding `observedGeneration` to `AgentStatus` so the CR's own condition becomes trustworthy and this PR's generation heuristic can be dropped.

Correction to my earlier comment and to the review that prompted it: the type is `AgentStatus` in `common_types.go`, not `PlatformAgentStatus`. It is used only by `PlatformAgent`, so the change is contained to one CRD.

mplakhtiy:
## Findings

### 1. HIGH — Stage 08 now blocks for up to 600s before LiteLLM exists, and a timeout aborts the rest of the pipeline

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:180-192`

```bash
kubectl rollout status "deployment/${deploy_name}" -n "${NAMESPACE}" \
    --timeout="${AGENT_READY_TIMEOUT:-600s}" || {
```

`provision.sh` runs with `set -e` (line 9) and invokes the stages in order — stage 08 at line 34,
LiteLLM at line 35, GitHub minter at 36, inference replay at 37. `run_step` turns a non-zero
`execute_func` into `exit 1` (common.sh:383). So the new gate makes stage 08 the point at which a
slow or unhealthy agent stops the whole install: stages 09-11 never run. That is a change of blast
radius, not just of reporting, and it applies to `make gcp-provision`, `hack/ci-deploy.sh`, and the
one-liner `install.sh`, all of which drive `provision.sh`.

The worst case is a deadlock. The agent's model backend is
`http://litellm.<ns>.svc.cluster.local/v1` (`deploy/shared/defaults/config.yaml:5`,
`platformagent_manifests.go:320`), and that Service is created by stage 09 — one stage *after* the
new wait. The repo's own e2e acknowledges the ordering: `hack/ci-deploy.sh` runs 08, then 09, and
only then waits for `deployment/platform-agent-gateway` to roll out (lines 78-100). Moving the wait
into stage 08 inverts that.

**Failure scenario:** a first-time `make gcp-provision` against an empty project. If
`hermes gateway run` exits when its configured model endpoint does not resolve, the agent container
never reaches Running, the Pod never becomes Ready, `kubectl rollout status` burns the full 600s,
and `provision.sh` aborts before LiteLLM is deployed. Re-running does not escape it — every run
stops at the same stage, and the dependency that would fix it is always one stage away.

**Verdict: PLAUSIBLE.** The ordering, the `set -e` abort, and the loss of stages 09-11 are all
confirmed from source. What is not confirmed is whether the agent container tolerates a missing
model backend: the Pod's only readiness probe is on the credential-proxy sidecar and targets
`http://127.0.0.1:8765/healthz` (`platformagent_manifests.go:1277-1281`), which is Envoy inside the
same Pod and has no external dependency, so a merely *unreachable* backend would not block Ready.
The question is whether the agent process stays up. One fresh install against an empty project
settles it; so does `kubectl get pod -l app=platform-agent-gateway` on a cluster where the CR is
applied and no `litellm` Service exists.

If it does tolerate it, the finding reduces to the blast-radius half, which still stands: the
readiness gate sits at the stage that creates the workload rather than at the end of the pipeline
where the workload's dependencies exist, so any stage-08 slowness now costs stages 09-11.

### 2. MEDIUM — The 60s cap on Deployment creation is five times tighter than the repo's own budget, and `AGENT_READY_TIMEOUT` does not cover it

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:163`

```bash
ensure_k8s_resource_exists "deployment/${deploy_name}" "${NAMESPACE}" 30 || return 1
```

The third argument is a retry *count*, not seconds: `ensure_k8s_resource_exists` forwards it to
`retry "$retries" 2` (common.sh:439), giving 30 attempts with 2s between them — about 58s in total.
`AGENT_READY_TIMEOUT` is wired only into `--timeout` on `kubectl rollout status` (line 183), so
there is no override for this window.

`hack/ci-deploy.sh:78-84` waits **300 seconds** for exactly the same event, under the comment
"Waiting for deployment/platform-agent-gateway to be created by operator". That is the repo's own
empirical budget for the operator's first reconcile, and the new gate is one fifth of it.

**Failure scenario:** `hack/ci-deploy.sh` runs stage 03 (operator install), stage 07, then stage 08
back to back. If the controller-manager is still pulling its image, waiting on leader election, or
requeueing behind cert-manager when the CR lands, the Deployment does not appear within 58s. Stage
08 exits 1 with "Timeout waiting for deployment/platform-agent-gateway to be created in
'kubeagents-system'" on a healthy install, and — per Finding 1 — takes stages 09-11 down with it.
The operator has no environment variable to widen the window.

**Verdict: CONFIRMED.** Raise the count to match the 300s precedent (`150`), or better, derive both
waits from `AGENT_READY_TIMEOUT` so the documented escape hatch actually covers the whole gate.

### 3. MEDIUM — The 30s generation heuristic narrows the false-success window rather than closing it

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:169-178`

```bash
while [ "$waited" -lt 30 ]; do
  current_generation=$(kubectl get "deployment/${deploy_name}" …)
  [ -n "$current_generation" ] && [ "$current_generation" != "$prev_generation" ] && break
  sleep 3
  waited=$((waited + 3))
done
```

The loop cannot distinguish "the operator reconciled and nothing changed" from "the operator has
not reconciled yet". Both leave the Deployment generation untouched, and both fall through to
`kubectl rollout status`, which then reports the *previous*, already-complete rollout as successful.

**Failure scenario:** an operator upgrade or a redeploy runs while the controller-manager is itself
mid-rollout or backing off. Stage 08 applies a CR carrying a new `IMAGE_TAG`, the operator does not
translate it into a Deployment change inside 30s, `rollout status` returns 0 against the old
ReplicaSet, and the stage prints "PlatformAgent Custom Resource applied successfully to GKE!" for an
image that was never deployed. That is the same class of false success the PR exists to remove,
narrowed to a smaller window rather than eliminated.

**Verdict: CONFIRMED.** The author raised this themselves and filed #534 to add `observedGeneration`
to `AgentStatus`, which is the right long-term fix. There is a cheaper tightening available inside
this PR, though, and it removes the need to guess: record the *CR's* `metadata.generation` before
the apply as well. `kubectl apply` bumps it if and only if the spec actually changed. If the CR
generation changed but the Deployment generation has not moved when the window closes, the operator
has demonstrably not reconciled and the step should fail rather than gate on a stale rollout; if the
CR generation did not change, the apply was a genuine no-op and falling through is correct.

### 4. LOW — The `DRY_RUN` guards added to `execute_custom_resource` are unreachable

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:144`, `:169`, `:180`

`run_step` returns before it ever calls `$execute_func` under dry-run (common.sh:370-373), and
`execute_custom_resource` is invoked only through `run_step` (line 197). All three
`[ "${DRY_RUN:-0}" -ne 1 ]` conditionals are therefore dead. The tell is line 163, which the author
left unguarded because `ensure_k8s_resource_exists` no-ops internally (common.sh:431) — a
distinction that only matters if the guards were load-bearing, and they are not.

**Cost:** three conditionals that read as dry-run support, will never be exercised, and would be
trusted without ever having run if the invocation changes. Drop them; the surrounding stages do not
carry equivalents.

### 5. LOW — `AGENT_READY_TIMEOUT` without a unit suffix is reported as a rollout failure

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:183`

The comment banner and the generated README row both say "default 600s", which invites a user hitting
a slow image pull to set `AGENT_READY_TIMEOUT=1200`. `kubectl` rejects a bare integer
(`invalid argument "1200" for "--timeout" flag: time: missing unit in duration`) and exits non-zero
immediately, so the `|| { … }` block fires and the step prints "PlatformAgent workload
'platform-agent-gateway' did not roll out successfully" plus a pile of healthy diagnostics.

**Failure scenario:** the operator reaches for the documented escape hatch on a constrained cluster
and is told the agent failed to roll out, within a second of applying the CR, while the rollout is
in fact proceeding normally. Validating the value against `^[0-9]+[smh]$` (or appending `s` when the
value is bare) turns a misdiagnosis into a clear message.

### 6. LOW — Three copies of the rollout gate now exist, and the richest diagnostics became unreachable

`hack/ci-deploy.sh:78-100`, `.github/workflows/reusable-deploy-agent.yml:117-127`

Both already do what stage 08 now does. In the workflow the duplication is worse than redundant:
"Perform Agent Redeployment" (line 104) runs the script, so on a stuck rollout the script's own 600s
wait elapses first and fails the step, and the following "Verify Agent Deployment Rollout" step —
default `if: success()` — never runs. That step is where `kubectl describe deployment`,
`kubectl describe pods`, and `kubectl logs --all-containers --tail=50` live. The script's
replacement diagnostics print CR conditions and per-container ready/restart counts, but no logs and
no events.

**Cost:** the CI failure path loses its container logs, and the workflow's 180s budget no longer
bounds anything because the script's 600s wait runs ahead of it. Either delete the now-duplicated
blocks in favour of the script's gate, or move the workflow's diagnostics into an
`if: failure()` step so they still fire.

### 7. LOW — The PR body describes an implementation the diff does not contain

The body states: *"Stage 08 waits for the `PlatformAgent` to report `Ready` using the existing
`wait_for_k8s_resource` helper, which no-ops under `--dry-run`."* The code at head does neither —
it gates on `kubectl rollout status` against the Deployment and never calls `wait_for_k8s_resource`.
The generated README row and the comment banner were updated to match the code; the body was not.
The author corrected this in a follow-up comment, but the body is what becomes the squash-merge
commit message, so the correction does not travel with it. Fold the comment's explanation into the
body before merging.

### 8. LOW — Neither fix has regression coverage

There is no shell test harness in the repository — no `.bats` files, no `tests/` under
`k8s-operator/scripts/`, and CI's only shellcheck job covers `install.sh`, `uninstall.sh`, and
`upgrade.sh` (`installer-matrix-test.yml:44-48`). The author verified both fixes by hand with
stubbed functions, which is exactly the right method; nothing captures it.

**Regressions that would ship silently:** someone re-adds a trailing `|| true` to
`execute_platform_agent`, or reorders `execute_custom_resource` so `kubectl rollout status` is no
longer the statement whose status the function returns. Both are the precise mistakes this PR
fixes, and both would go green. The author's stubbing approach — source the script with
`kubectl`/`gcloud` shims on `PATH` and assert the function's exit status — is small enough to check
in alongside the fix.


---

bnaylor:
All eight verified against source before changing anything.  b512bcc reworks the gate along the lines finding 1 suggests.  The branch is also rebased onto main (the conflicting `scripts/README.md` rows are generated, so the resolution was `make docs-generate`).

**1 - gate placement:** Confirmed.  `ci-deploy.sh` already had the right shape - 08, then 09, then wait - and this PR inverted it.  Stage 08 now gates only on the operator having reconciled the apply: the Deployment exists, and its generation moves when the CR generation moved.  The rollout wait lives in a new stage 13 at the end of the pipeline, where the `litellm` Service exists.  I didn't chase down whether `hermes gateway run` survives a missing backend, since the redesign makes it moot - stage 08 no longer waits on readiness at all.

**2 - 60s creation cap:** Fixed.  Both waits now derive from `AGENT_READY_TIMEOUT`, so the default creation budget is 600s and the documented escape hatch covers the whole gate.

**3 - the 30s heuristic:** Took your suggestion.  Stage 08 records the CR's `metadata.generation` before the apply.  If it moved and the Deployment generation doesn't move within the budget, the step fails; if it didn't move, the apply was a genuine no-op and stage 13 verifies the running Deployment.  One residual: a CR spec change that legitimately touches nothing in the Deployment would false-fail.  The operator stamps config hashes into the pod template annotations, so config-only changes do roll the Deployment, which makes that window narrow.  #534 remains the real fix.

**4 - dead DRY_RUN guards:** Dropped.

**5 - bare AGENT_READY_TIMEOUT:** Fixed.  A bare integer gets an implied `s`, and anything else that doesn't match `^[0-9]+[smh]$` fails immediately with a message naming the variable.

**6 - three copies of the gate:** `ci-deploy.sh` now calls stage 13 instead of its hand-rolled wait.  The workflow I left alone: with the rollout wait out of stage 08, its "Verify Agent Deployment Rollout" step is reachable again.  That step's 180s budget bounds the redeploy path as before, with its log-dumping diagnostics firing on rollout failure.  If stage 08's reconcile gate fails there, pod-level diagnostics wouldn't help anyway - the Deployment never changed.

**7 - stale body:** Rewritten to match the code.

**8 - regression coverage:** Deferring.  The stub harness now covers 14 scenarios including both regressions you name, but checking it in means introducing a shell test harness to a repo that has none, and I'd rather not fold that decision into a bugfix PR.  Happy to file it as a follow-up.

mplakhtiy (approved):

PR #536, opened 2026-08-05
fix(onboarding): run bootstrap exactly once

dshnayder:
# Pull Request

## Why This Change

Users report that first-run bootstrap runs several times: the fleet gets scanned repeatedly, and the greeting and the inventory report can be delivered more than once.

The flow's only "already ran" record is `.bootstrap_completed`, and it is written at the very end of the chain — after a report has been both produced *and* delivered to a human. Both onboarding cron jobs tick every 60 seconds while the work they guard takes minutes, so every stage re-enters that window dozens of times. Each stage was asking "has onboarding *completed*?" to decide whether to *start*, and the answer is "no" for the entire time it is being re-run.

**What tipped it over was the move to subagent delegation.** The card the scan gate files now completes almost immediately: the Platform Agent worker's job is to fan discovery out to per-cluster child cards, not to scan, so it finishes as soon as it has delegated. `INVENTORY.md` arrives minutes later from the aggregation card. For that whole window the board says "done" and the disk says "no report" — indistinguishable from "never scanned" — so the gate filed a fresh fleet-wide sweep every minute. The card's fixed `idempotency_key` did not stop it: it dedupes against non-archived rows in one board's database, which is not a guarantee the flow can lean on.

Two related repeats had the same shape:

- The plugin greeted on any first turn without `.bootstrap_completed`. Every new session starts with `is_first_turn=True`, so a second user, a new thread, or a pruned session re-greeted, re-marked human presence, and re-pointed the delivery job at whoever spoke last.
- Delivery checked `.bootstrap_completed` and wrote it *after* emitting, leaving a window in which a scheduled tick and the plugin's `trigger_job` run could both post the full report.

## What Changed

Each once-only step now writes a durable marker at the moment it acts, rather than inferring from state that only appears at the end.

**Files:**

- `agents/chat/scripts/bootstrap_scan_gate.py`
- `agents/chat/scripts/bootstrap_delivery.py`
- `agents/chat/defaults/plugins/bootstrap_onboarding/plugin.py`
- `agents/chat/defaults/plugins/bootstrap_onboarding/README.md`
- `agents/chat/scripts/test_bootstrap_onboarding_scripts.py`
- `agents/chat/defaults/plugins/bootstrap_onboarding/test_plugin.py`
- `docs/site/src/content/docs/concepts/chatops.md`

**Added/Changed/Fixed:**

- **Scan gate files once.** The filed card's id is recorded in `/opt/data/.bootstrap_scan_filed` and the job skips on it. The card is created with `--json` so the id can be confirmed; a create that fails writes no marker and retries next tick, so a board hiccup cannot silence discovery permanently. The idempotency key stays as a backstop, not as the guarantee, and the same keys are now propagated to the aggregation and per-cluster cards the worker fans out to.
- **Greeting happens once.** The plugin writes `.bootstrap_greeted` and gates on it as well as `.bootstrap_completed`. It also binds the delivery origin *first* and primes nothing at all when no chat origin can be bound — previously a CLI/API session could mark a human present in a session that cannot receive anything, letting the single-use report fire while still set to `deliver: local`.
- **Delivery is claimed atomically.** `.bootstrap_completed` is created with `O_CREAT | O_EXCL` before the first byte reaches stdout, so of two overlapping runs exactly one delivers and the other exits silently.
- **The report survives delivery.** `INVENTORY.md` is renamed to `INVENTORY.delivered.md` instead of being deleted, so a lost chat message costs a re-send rather than a full re-scan of the fleet.
- **Docs:** the plugin README gains the marker table, the new sequence steps, and Rule 8 ("every once-only step writes its own marker, at the moment it acts") recording this failure mode; `chatops.md` no longer credits the idempotency key with preventing duplicates.

## Why This Matters

- A fleet-wide GKE discovery sweep launched once a minute is expensive in tokens and in API calls against every cluster in the project, and it lands on the user as a stream of duplicate work.
- Being greeted again, or receiving the same long inventory report twice, is the first impression the harness makes on a new install.
- The markers also make the flow debuggable and re-runnable: `.bootstrap_scan_filed` names the card to inspect when onboarding stalls, and deleting a marker is the supported way to re-arm just that stage.

## Context

- No behavior change once onboarding has completed: `.bootstrap_completed` still suppresses everything, so deployments that already finished (and therefore have no new markers) stay quiet.
- Deliberately unchanged: the 1-minute job intervals, and the two-job split between scan and delivery — the scheduler snapshots a job's delivery target at run start, so merging them would lose the report for a user who connects mid-scan.
- Generated with the help of Claude.

## Testing

- `python3 -m unittest agents.chat.scripts.test_bootstrap_onboarding_scripts` — 23 tests, passing. New coverage: filing only once across repeated ticks, no marker when the board refuses the card, the atomic claim rejecting a racing run, archive-instead-of-delete, and an unreadable report leaving state untouched.
- `python3 -m unittest discover -s agents/chat/defaults/plugins/bootstrap_onboarding -p 'test_*.py'` — 13 tests, passing. New coverage: greeting once across sessions, the greeted marker, and an unbindable session priming nothing while leaving the flow armed.
- `make docs-check` — passing. `npx prettier --check` on changed Markdown — clean. `cd docs/site && npm run build` — succeeds.
- Docs-drift review run against the branch diff: every doc stating a fact about these markers was updated (`chatops.md`, the plugin README); `reference/cron-jobs.md` needs no change.

---

Each once-only stage is covered by two tests: one for acting, one for refusing to act again. Rollout risk is low — the markers are additive, and the worst case of a marker that fails to write is the previous behavior (a repeat), never a silenced onboarding.


bradhoekstra (changes requested):
@dshnayder Need to sync your branch to main to get the Actions Scan to pass.

bradhoekstra (approved):

PR #544, opened 2026-08-05
feat(operator): stamp the recommended app.kubernetes.io labels on every installed object

bradhoekstra:
# Pull Request

## Why This Change

Nothing marks the Kubernetes objects this project installs as belonging to kube-agents:

- the operator's per-CR output carries only `app: <name>-gateway` — a selector, not an identity;
- the operator's own install manifests carry an inconsistent partial set (`app.kubernetes.io/name` + `managed-by: kustomize` on some files, nothing on `config/rbac/role.yaml`, `config/webhook/manifests.yaml`, or `config/crd/bases/*`);
- the LiteLLM, GitHub-minter and inference-replay integrations and the provisioned Secrets carry nothing at all.

So there is no way to answer "what in this cluster did kube-agents put here?" — which matters for fleet audits, cost attribution, cleanup, and for a human staring at an unfamiliar cluster. Cluster-scoped RBAC is the sharpest edge: a namespaced `PlatformAgent` cannot own a cluster-scoped resource through an owner reference, so orphaned ClusterRoles and ClusterRoleBindings were previously unfindable.

This adopts the [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) across everything the project installs, so `-l app.kubernetes.io/part-of=kube-agents` selects the whole footprint.

## What Changed

**Files:**

- `k8s-operator/internal/controller/manifest_helpers.go` — label constants, `instanceLabel()`, `commonLabels()`, `withCommonLabels()`; `ReconcileServiceAccount` takes a `labels` parameter
- `k8s-operator/internal/controller/platformagent_controller.go` — new `applyManaged()` apply choke point; PVC labelling
- `k8s-operator/internal/controller/platformagent_manifests.go` — recommended labels on the pod template
- `k8s-operator/internal/controller/manifest_helpers_test.go` — `TestInstanceLabel`, `TestWithCommonLabels`
- `k8s-operator/internal/testing/testdata/platform/expected/platformagent{,-tagged}.yaml` — regenerated goldens
- `k8s-operator/config/{default,crd}/kustomization.yaml` and the four integration kustomizations — `labels:` stanzas
- `deploy/kustomize/platform/service.yaml` — hand-labelled (no kustomization to hang a stanza on)
- `charts/kube-agents/templates/_helpers.tpl` — `part-of` added to the shared `kube-agents.labels` define
- `k8s-operator/scripts/provision_07_gcp_k8s_secrets.sh` — labels the two generated Secrets
- `docs/site/src/content/docs/reference/resource-labels.md` (new), `reference/attribution.md`, `deploy/kustomize.md`, `docs/README.md`

**Added/Changed/Fixed:**

- **Four recommended labels on every install path** — `name`, `instance`, `part-of`, `managed-by`.

  | Source                                                  | `name`                 | `instance`                    | `part-of`     | `managed-by`               |
  | ------------------------------------------------------- | ---------------------- | ----------------------------- | ------------- | -------------------------- |
  | PlatformAgent controller output (per CR)                | `platform-agent`       | `<namespace>-<agent name>`    | `kube-agents` | `platformagent-controller` |
  | Operator install (`k8s-operator/config/`)               | `kube-agents-operator` | `kube-agents-operator`        | `kube-agents` | `kustomize`                |
| Helm chart (`charts/kube-agents/`)                      | `kube-agents`, `kube-agents-operator`, or `litellm` | `<release name>` | `kube-agents` | `Helm`          |
  | LiteLLM integration                                     | `litellm`              | `litellm`                     | `kube-agents` | `kustomize`                |
  | GitHub token minter                                     | `github-token-minter`  | `github-token-minter`         | `kube-agents` | `kustomize`                |
  | Inference replay                                        | `inference-replay`     | `inference-replay`            | `kube-agents` | `kustomize`                |
  | Provisioned Secrets (`provision_07_gcp_k8s_secrets.sh`) | `platform-agent`       | `${NAMESPACE}-platform-agent` | `kube-agents` | `provisioner`              |

- **`component` is set by nothing, `version` only by the chart.** The object's own `name` already says what it is, and a second key that has to stay consistent with the first eventually disagrees with it. `version` is a different case: the chart fills it from `Chart.AppVersion`, and a chart release maps to exactly one application release, so there is a correct value to write. The controller cannot — its objects run whatever image the CR named, and an image reference may carry a digest whose `@` and `:` are not legal in a label value — and the kustomizations and the provisioner have no build identity at all.
- **The Helm chart is in scope.** It renders 13 objects the operator does not (the operator Deployment and its RBAC, the LiteLLM stack, the `PlatformAgent` CR), and none of them carried `part-of` — the documented footprint query missed every object a Helm install created. `part-of` now lives in the shared `kube-agents.labels` define, which every template already includes.
- **One documented gap.** Helm applies `crds/` verbatim and never templates it, and `make chart-check` enforces those files as byte-for-byte copies of `config/crd/bases/`, so chart-installed CRDs cannot carry the labels. Installing CRDs through `k8s-operator/config/crd/` does label them. The reference page says so rather than leaving it to be discovered.
- **A single apply choke point in the controller.** The ten duplicated `r.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner("platformagent-controller"))` calls now route through `applyManaged`, which stamps the labels first. A future object cannot reach the cluster unlabelled, and the repeated field-owner string literal is now one constant. PVCs are labelled at their `Create` call, since they do not go through the choke point.
- **`instance` carries the namespace** for controller output, because the controller also writes cluster-scoped RBAC where a bare CR name is ambiguous between two agents of the same name in different namespaces. Nothing bounds a `PlatformAgent` name to a length that fits a label value, so the joined value is truncated to 63 characters and any trailing separator trimmed — `TestInstanceLabel` asserts the result against `validation.IsValidLabelValue`.
- **`kubectl create secret` has no `--label` flag**, so `provision_07` pipes the generated manifest through `kubectl label --local -f - -o yaml` before applying. The manifest stays on the pipe — never written to disk or echoed — so the credentials it carries are not exposed. Each pipeline runs in a `set -o pipefail` subshell so a failure in the generate or label stage is not masked by the apply at the end of the pipe.
- **New reference page** `reference/resource-labels.md` with the contract table, the immutability constraints as an upgrade note, and query recipes. `reference/attribution.md` gains a sentence distinguishing per-requester attribution *annotations* from these project-identity *labels*.

## Why This Matters

- One query now answers "what did kube-agents install here?", including the cluster-scoped RBAC that no owner reference can reach.
- `managed-by` distinguishes an object the operator reconciles from one a human applied with kustomize or the provisioner created — which is what you need when deciding whether it is safe to delete something.
- The choke point makes the invariant structural rather than a convention someone has to remember.

## Context

**Deliberately out of scope:** objects the Platform Agent authors at runtime through its skills. Those are user workloads, and `reference/attribution.md` already reserves the `kubeagents.x-k8s.io/requested-by` **annotation** for them.

**Two constraints that would silently break upgrades, and are honoured:**

1. **Selectors are never touched.** Deployment and StatefulSet `spec.selector` is immutable, so the API server would reject the update on every existing install. The controller keeps `app: <name>-gateway` as the sole selector key, and every kustomization sets `includeSelectors: false`.
2. **StatefulSet `volumeClaimTemplates` are never touched.** Every StatefulSet spec field other than `replicas`/`template`/`updateStrategy` is immutable, so labelling them would break the SSA patch for existing installs. `buildRWOVolumeClaimTemplates` is left alone.

**Two accepted limitations:**

- Pod-template labels change the template hash, so upgrading causes **one** rollout of the agent Deployment. Expected, happens once.
- PVCs are create-only in the controller, so claims that predate this change stay unlabelled until recreated and will not appear in the `part-of` query.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` — green, including the two new unit tests.
- Goldens regenerated with `-update`. Filtering `app.kubernetes.io/*` lines and new `labels:` keys out of `git diff` on `internal/testing/testdata` leaves **zero** lines — the diff is purely additive, with no selector or `volumeClaimTemplates` churn.
- All six kustomize roots built before and after and diffed; every root is additive only, again with no selector drift:

  ```
  default -> 15 objects labelled          integrations/litellm/base -> 6
  crd -> 2                                integrations/litellm/overlays/chatgpt -> 7
  integrations/github -> 6                integrations/inference-replay/base -> 6
  ```

- `provision_07_gcp_k8s_secrets.sh`: `bash -n` and `shellcheck -x` introduce no new warnings; the label pipeline was run against a real `kubectl` and verified to add the four labels while preserving the Secret's `data`.
- `helm template` rendered with LiteLLM enabled: all **14** objects carry `app.kubernetes.io/part-of: kube-agents`, asserted by parsing the rendered stream rather than grepping it. `helm lint` and `make chart-check` clean.
- `make docs-check` green (generated regions, links, terminology, map coverage); `npx prettier --check` clean on all changed Markdown and YAML; `cd docs/site && npm run build` renders `/reference/resource-labels/` with no new warnings.
- `review-docs-drift` run against the branch diff — one finding, addressed: the label values are a new category of documented identifier, so `docs/README.md` §2 gains an identifier-sources row pointing at `manifest_helpers.go` and the kustomizations.

---

Additive across the board. The one behavioural effect on an existing install is a single rollout of the agent Deployment from the pod-template label change; no immutable field is written, so there is no path to a rejected apply.


bradhoekstra:
> 🟡 Low — `deploy/kustomize.md` embeds a now-stale copy of `service.yaml`

Confirmed and fixed in f96c321. The page frames the block as the current state of the file ("the Kustomize surface today is one file"), so leaving it unlabelled was a genuine copy-paste trap, not just cosmetic drift.

The embedded YAML now matches `deploy/kustomize/platform/service.yaml` exactly — verified mechanically rather than by eye:

```
$ awk '/^```yaml$/{f=1;next} /^```$/{f=0} f' docs/site/src/content/docs/deploy/kustomize.md > /tmp/snippet.yaml
$ diff -u deploy/kustomize/platform/service.yaml /tmp/snippet.yaml && echo MATCHES
MATCHES
```

Rather than restate the label semantics a second time, the page now points at `reference/resource-labels` as canonical for what the keys mean and why `component`/`version` are absent.

I checked whether the same drift class hit anywhere else in this PR — `kustomize.md` was the only page embedding an affected manifest. `make docs-check`, prettier, and the site build all pass.

Not fixed here: the duplication itself remains a drift magnet that none of the four `docs-check` gates can catch, exactly as the finding notes. Worth a follow-up to either drop the snippet or generate it, but that is wider than this PR.

adamparco (approved):

PR #545, opened 2026-08-05
docs(contributing): document the kube-agents-bot review in the PR workflow

bradhoekstra:
# Pull Request

## Why This Change

`kube-agents-bot` reviews every pull request in this repository automatically, but nothing in the repository said so. The result is a predictable failure mode for AI agents working here: they open a PR, report the task as finished, and walk away roughly a minute before the review lands. The findings then sit unread, or the user has to relay them by hand.

Agents also had no way to know the parts of the bot's contract that are not guessable — most importantly that **pushing more commits does not trigger another review**. An agent that pushes a fix and waits will wait forever.

## What Changed

**Files:**

- `AGENTS.md`
- `docs/site/src/content/docs/contributing.md`

**Added/Changed/Fixed:**

- Added an **"Automated Review After Opening a Pull Request"** section to `AGENTS.md` covering:
  - **When it runs** — `opened`, `reopened`, draft-marked-ready; never on push. `/review` on its own line forces a fresh pass; `agent:ignore` opts out and outranks it.
  - **How to read it** — 👀 means started, a posted review means finished; severity badges (🔴/🟠/🟡); unanchored findings land in the summary body under "Findings outside this diff"; "no findings" is a result, not silence; a 👀 with nothing following it is a bug in the bot.
  - **What agents must do** — after creating a PR, tell the user the review is coming and *offer to wait* rather than declaring the work done; then triage findings **with** the user (fix / push back / defer) instead of acting unilaterally, answering disagreements in the thread rather than dropping them. After pushing fixes, ask whether to `/review` again.
  - Copy-pasteable `gh` commands to poll for the review, list inline findings with their comment ids, and reply in-thread.
- Mirrored the same contract for humans under **Code review → Automated review** on the site's contributing page, following that page's existing "PR hygiene (from `AGENTS.md`)" pattern and linking back to `AGENTS.md` as canonical.

## Why This Matters

The bot only comments — it never pushes and never merges — so every finding needs a human or an agent to act on it. Writing the contract down turns "the PR is open" into "the PR is open and the review is inbound", and makes the agent's default behaviour to wait and collaborate rather than to disengage.

The guidance is deliberately non-autonomous: the agent surfaces and recommends, the user decides. A review bot's findings are advice from a reviewer, not instructions, and an agent silently rewriting code to satisfy one is the outcome this is written to avoid.

## Context

Docs-only; no behaviour change to any agent, workflow, or the operator.

One detail worth flagging for reviewers, since it is easy to get wrong: `gh pr view --json reviews` reports the login as `kube-agents-bot`, while the REST API (`/pulls/N/comments`) reports `kube-agents-bot[bot]`. Both spellings appear in the documented commands, and each is correct for its own endpoint.

## Testing

- `make docs-check` — all four gates pass (generated regions current, 170 files' relative links resolve, terminology check, map inventory covers 147 tracked docs).
- `npx prettier --check` on both changed files — clean.
- `cd docs/site && npm run build` — 41 pages built, no errors.
- Ran the `review-docs-drift` skill against the branch diff: no competing canonical home for this fact, no map staleness backlog, no new/renamed/deleted docs requiring a `docs/README.md` update.
- The documented `gh` commands were verified against real reviews on PRs #536 and #539 rather than written from memory.

---

Docs-only, no rollout risk. The main thing to sanity-check is whether the split is right: the full contract in `AGENTS.md` and a shorter mirror on the site, matching how PR hygiene is already handled.


toshiowang:
Reviewed the diff and verified the claims against the live API rather than reading them off the PR body. The two-spellings detail (`kube-agents-bot` from `gh pr view --json reviews`, `kube-agents-bot[bot]` from REST) checks out on #536, `make docs-check` passes all four gates on the head, and prettier is clean on both files. The AGENTS.md-full / site-summary split is the right call and matches the existing "PR hygiene (from `AGENTS.md`)" precedent — I'd keep it.

Four non-blocking findings, all in the documented `gh` commands or their surroundings.

### 🟡 The poll recipe false-positives on a re-review

`select(.author.login == "kube-agents-bot") | .body` prints *every* bot review on the PR. The section later tells agents to comment `/review` after pushing fixes; an agent that reuses this command then sees the **previous** review's body immediately and concludes the new one has landed. Taking the latest and surfacing the timestamp makes the staleness visible:

```bash
gh pr view <number> --json reviews \
  --jq '[.reviews[] | select(.author.login == "kube-agents-bot")] | last | "\(.submittedAt)\n\(.body)"'
```

### 🟡 The REST comments call is unpaginated

`gh api repos/.../pulls/<n>/comments` defaults to 30 per page, so a review with more findings is silently truncated — the worst failure mode for this particular command, because the output still looks complete. Add `--paginate`. Latent today (the busiest recent PR had 3 comments), but the doc is what an agent will copy verbatim a year from now.

While in there: `.line` is `null` on outdated comments, which renders as `path:null`. `(.line // .original_line)` is more robust.

### 🟡 Neither doc links to the bot's own source of truth

The contract originates in `bradhoekstra/kube-agents-bot` — the bot's intro comment states it almost verbatim — and is now restated in two in-repo files. `AGENTS.md`'s own rule is "every fact has one home… link rather than summarise". The mirror-with-canonical-pointer shape is right, but without a link to the bot repo from the `AGENTS.md` section, drift between upstream behaviour and these two copies is undetectable.

### 🟡 `docs/README.md` map descriptions are now slightly stale

Line 166 enumerates `AGENTS.md`'s contents ("repo layout, skills guidelines, the canonical-home documentation rules, generated-regions rule, and PR hygiene") and line 223 describes `contributing.md` ("CLA, PR hygiene, and the local checks CI enforces"). Neither mentions the automated review. `make docs-check` only enforces presence so this passes CI — it's the kind of enumeration staleness `review-docs-drift` is meant to catch.

---

### Worth keeping as-is

- Hardcoding `repos/gke-labs/kube-agents` in the REST call is correct, not an oversight — `AGENTS.md` mandates fork branches, and `{owner}/{repo}` substitution would resolve to the contributor's fork, where the review comments don't live.
- "A 👀 with nothing following it is a bug in the bot, not a verdict" and "no findings is a result, not silence" are exactly the ambiguities that make an agent hang or guess.
- The non-autonomous framing — surface and recommend, user decides, disagreements answered in-thread rather than silently dropped — is the right default for a reviewer that says of itself "I can be wrong."

Two minor things: `gh pr view <number>` in the first snippet omits `--repo` while the second hardcodes it (it resolved correctly from my fork checkout, but the asymmetry invites confusion), and the "Pull Request Hygiene" section doesn't cross-reference the new one — a one-line pointer next to the existing docs-drift bullet would help agents reading top-down.

Of the four, `--paginate` and `| last` are the two I'd want in before merge, since both fail silently in a way that looks like success.

docs/site/src/content/docs/contributing.md
```diff
@@ -73,6 +73,17 @@ Before pushing, run the checks CI enforces:
 
 All submissions, including from project members, require review through GitHub pull requests. See [GitHub Help — About pull requests](https://help.github.com/articles/about-pull-requests/).
 
+### Automated review
+
+Every pull request is also reviewed by [`kube-agents-bot`](https://github.com/bradhoekstra/kube-agents-bot), a GitHub App that runs a coding agent over the branch diff. It only comments — it never pushes commits and never merges, and it does not replace the human review above. Its own repository is the source of truth for how it behaves.
```
toshiowang: FYI I can't actually access https://github.com/bradhoekstra/kube-agents-bot, it's a broken 404 page

bradhoekstra: Confirmed, and fixed in 7e70894 — the repository is private, so that link 404s for everyone but me. Rather than swap it for another pointer nobody can open, both docs now defer to the bot's intro comment, which it posts on every pull request it picks up and which states the same contract:

> The bot introduces itself in a comment on every pull request it picks up, and that comment states its current contract; if it disagrees with what follows, believe the comment and fix this section.

That keeps the drift check you were after in the 🟡 finding below — there is still a live source to compare these two copies against — without depending on repository access. The bot name itself stays as plain text, since it is the login the documented `gh --jq` filters match on.

bradhoekstra:
Thanks for verifying the claims against the live API rather than off the PR body — the two-spellings detail was the one I most wanted a second pair of eyes on.

All four findings are addressed, plus both minors. Two commits: `f760051` for the command hardening, `7e70894` for the bot-repo link.

### 🟡 Poll recipe false-positives on a re-review — fixed

Agreed, and this is the failure mode the section itself creates by telling agents to `/review` after pushing fixes. Took your shape and added `select(.)` so a no-review-yet case prints nothing rather than `null`:

```bash
gh pr view <number> --repo gke-labs/kube-agents --json reviews \
  --jq '[.reviews[] | select(.author.login == "kube-agents-bot")] | last | select(.)
        | "\(.submittedAt)\n\(.body)"'
```

The comment above it now says why the timestamp comes first.

### 🟡 Unpaginated REST call — fixed

`--paginate` added, and `(.line // .original_line)` for outdated comments. Your framing of it as the worst place for silent truncation is right — the output looks complete either way, and this is a snippet written to be copied verbatim.

### 🟡 No link to the bot's source of truth — addressed differently

I linked the bot repository first, then you found it 404s (thread [above](https://github.com/gke-labs/kube-agents/pull/545#discussion_r3730222953)) — it is private. Both docs now point at the bot's intro comment instead, which is posted on every pull request it reviews and states the same contract, with an explicit instruction to believe the comment over the doc when they disagree. Same drift check, no access dependency.

### 🟡 `docs/README.md` enumerations stale — fixed

Both rows updated: line 166 for `AGENTS.md` and line 223 for `contributing.md` now name the automated review, in the summary and the key-topics cells.

### Minors

- `--repo gke-labs/kube-agents` added to the first snippet, so both are explicit. A shared comment above them explains why upstream is hardcoded — worth stating outright given the fork rule three sections up, rather than leaving it to be rediscovered.
- One-line pointer to the new section added under "Pull Request Hygiene", next to the docs-drift bullet.

On the split: good, keeping it as-is.

**Verification:** `make docs-check` passes all four gates (155 tracked docs, relative links resolve across 178 files), prettier clean on all three files, `npm run build` in `docs/site` builds 43 pages.

Ready for another look. Since a push does not re-trigger the bot — the very thing this PR documents — I will leave `/review` to you.

toshiowang (approved):

PR #519, opened 2026-08-04
feat(installer): add zero-friction one-liner install script with Cloud Shell auto-discovery

fkc1e100:
Resolves #618

### 🤖 Summary & Purpose

This Pull Request introduces End-to-End Lifecycle Management (Zero-Friction Installation, Day-2 Control Panel, Hot-Reload Upgrades, & Teardown) for `kube-agents`:

```bash
# One-Liner Install
curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash

# Day-2 Upgrade Engine (Skills / Harness / Operator / Full)
curl -fsSL https://gke-labs.github.io/kube-agents/upgrade.sh | bash -s -- --upgrade-mode=skills
```

---

### 🎬 PR Visual Walkthrough & Video Demos
A standalone markdown document with human-speed terminal animations for code reviewers is available directly in this PR branch:
👉 **[Zero-Friction Installer & Control Panel Video Demos Walkthrough](https://github.com/fkc1e100/kube-agents/blob/feat/zero-friction-installer/docs/designs/zero-friction-installer-demos.md)**

- **Demo 0: One-Liner Quickstart (`curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash`)**
- **Demo 1: Interactive Setup Wizard with Step-Back (`b`) Navigation**
- **Demo 2: Day-2 Control Panel (`raspi-config` style) via `./install.sh --menu`**
- **Demo 3: Non-Interactive Automated CI/CD Setup (`./install.sh -y`)**
- **Demo 4: Day-2 Updating Chat Integrations & Allowed User Permissions**
- **Demo 5: Day-2 Disabling Web UI for Instant Security Hardening**

---

### 🌟 Key Enhancements Included

1. **Zero-Friction One-Liner Installer (`install.sh`)**:
   - Cloud Shell auto-discovery (`CLOUD_SHELL=true` / `DEVSHELL_PROJECT_ID`).
   - Interactive numbered menus for GCP Project & GKE Cluster selection.
   - **Step-Back (`b` / `back`) Navigation**: Type `b` at any menu prompt to step backward without starting over.
   - Non-interactive AI Agent / CI/CD automation mode (`-y`, `--non-interactive`).
   - Dry-run validation mode (`--dry-run`) with JSON report output (`/tmp/kube-agents-install-report.json`).

2. **Day-2 Control Panel Menu (`./install.sh --menu` / `./install.sh --config`)**:
[...]

PR #546, opened 2026-08-05
feat(provision): label kube-agents host clusters

haoxuw:
# Pull Request

## Why This Change

The admin console discovers the GKE cluster hosting kube-agents through a project-level resource label. Both supported full-install paths need to apply that marker consistently, while a label update failure must not prevent the functional agent stack from being provisioned.

## What Changed

**Files:**

- `k8s-operator/scripts/{common.sh,provision_08_deploy_platform_agent.sh,teardown_08_deploy_platform_agent.sh}`
- `terraform/examples/full-install/{main.tf,README.md}` and `terraform/modules/gke-cluster/`
- Provisioning, uninstall, generated step, and documentation-map references
- `scripts/test_host_label_provisioning.py` and `hack/check-docs-terminology.sh`

**Added/Changed/Fixed:**

- Define the admin console discovery contract as `kube-agents-host=true` and centralize shell reads/updates in `common.sh`.
- Register the label after applying the `PlatformAgent` CR, retry transient update failures, and treat registration as advisory so later provisioning steps still run.
- Remove the script-managed label during shell teardown only after local manifest cleanup; report a deferred nonzero status when label cleanup fails.
- Pass the label through the Terraform full-install composition while keeping the reusable module input optional for standalone callers.
- Document that Terraform installations must use their Terraform teardown path to avoid out-of-band state drift.
- Add CI-discovered tests for advisory registration and teardown continuation, plus a terminology guard that keeps shell, Terraform, and docs aligned.

## Why This Matters

- The admin console can automatically select the single labeled host and can surface zero or multiple labeled hosts for manual selection.
- A missing permission or in-flight GKE operation cannot abort provisioning before LiteLLM and the remaining functional components are installed.
- Shell and Terraform ownership remain explicit, avoiding unexplained Terraform plan drift.
- The existing provisioning `-y` behavior and current-main cluster discovery helpers are unchanged.

## Context

The label is consumed by the admin console. The agent runtime itself continues to receive its host identity from the configured `PlatformAgent` cluster name. The reusable GKE module does not assume every standalone cluster hosts kube-agents; the full-install composition opts in explicitly.

## Testing

- `python3 -m unittest scripts.test_host_label_provisioning -v`
- `bash -n k8s-operator/scripts/*.sh k8s-operator/scripts/dev/*.sh hack/check-docs-terminology.sh`
- `go build ./...` in `k8s-operator`
- Terraform 1.15.8 `fmt -check -recursive`, `init -backend=false`, and `validate` for every module/example
- `make docs-generate` followed by a clean generated-file check
- `make docs-check`
- Prettier check for changed Markdown/MDX files
- `npm run build` in `docs/site`
- `git diff --check origin/main...HEAD`

---

Both full-install paths now expose the host to the admin console without making discovery-label availability a prerequisite for a working deployment.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -143,6 +143,7 @@ identifier appears, add its source here.
 | Image defaults and override env vars (`PLATFORM_AGENT_IMAGE` et al.) | `k8s-operator/internal/controller/manifest_helpers.go` |
 | OTLP endpoint default, discovery candidates, and `otlpEndpointSource` values | `k8s-operator/internal/controller/telemetry.go` |
 | Registry prefix default (`REGISTRY_PREFIX`) | `k8s-operator/scripts/common.sh` |
+| GKE host-discovery label | `k8s-operator/scripts/common.sh` |
 | GitOps clone layout (`/opt/data/gitops/...`) and leases | `agents/platform/scripts/gitops_workspace.py` |
 | fleet-audit finding-id pattern and rendering caps | `agents/platform/skills/fleet-audit/scripts/audit_report.py` |
 | Helm chart value defaults (KSA/secret names, image repos, tag rules) | `charts/kube-agents/values.yaml` |
--- a/hack/check-docs-terminology.sh
+++ b/hack/check-docs-terminology.sh
@@ -63,6 +63,27 @@ forbid 'platform-agent-ksa' \
 forbid 'platform-agent-system' \
   "Stale namespace. The namespace is kubeagents-system."
 
+# --- GKE host-discovery label -------------------------------------------
+# Ground truth: k8s-operator/scripts/common.sh. The Terraform full-install
+# composition must mirror the stable key because it cannot source Bash.
+HOST_LABEL=$(awk -F'"' '/^KUBE_AGENTS_HOST_LABEL=/{print $2; exit}' k8s-operator/scripts/common.sh)
+if [ -z "$HOST_LABEL" ]; then
+  echo "ERROR: could not read KUBE_AGENTS_HOST_LABEL from k8s-operator/scripts/common.sh." >&2
+  exit 1
+fi
+
+WRONG_HOST_LABEL=$(search 'kube-?agents-host=true' | grep -vF "${HOST_LABEL}=true" || true)
+if [ -n "$WRONG_HOST_LABEL" ]; then
+  echo "::error::Documented host-discovery label does not match KUBE_AGENTS_HOST_LABEL=${HOST_LABEL}."
+  printf '%s\n\n' "$WRONG_HOST_LABEL" | sed 's/^/    /'
+  FAILED=1
+fi
+
+if ! grep -qE "\"${HOST_LABEL}\"[[:space:]]*=[[:space:]]*\"true\"" terraform/examples/full-install/main.tf; then
+  echo "::error::Terraform full-install composition does not apply ${HOST_LABEL}=true."
+  FAILED=1
+fi
+
 # --- Go toolchain ---------------------------------------------------------
 # Ground truth: k8s-operator/go.mod
 GO_MOD_VERSION=$(awk '/^go /{print $2; exit}' k8s-operator/go.mod)
--- a/docs/site/src/content/docs/install/quickstart-gke.mdx
+++ b/docs/site/src/content/docs/install/quickstart-gke.mdx
@@ -97,7 +97,7 @@ flowchart TB
     A --> F["05: Google Chat Pub/Sub (opt-in)"]
     A --> G["06: Slack (opt-in)"]
     A --> H["07: LLM API key Secret"]
-    A --> I["08: PlatformAgent CR"]
+    A --> I["08: PlatformAgent CR<br/>host-discovery label"]
     A --> J["09: LiteLLM Deployment"]
     A --> K["10: GitHub token minter + Cloud KMS (opt-in)"]
     A --> L["11: inference-replay (opt-in)"]
--- a/docs/site/src/content/docs/install/uninstall.md
+++ b/docs/site/src/content/docs/install/uninstall.md
@@ -47,6 +47,13 @@ Once the CR is gone, the operator's finalizer first removes the cluster-scoped R
 
 ## Full teardown
 
+The shell teardown is for environments created by `k8s-operator/scripts/provision.sh`. For the
+Terraform full-install example, follow its
+[Terraform teardown procedure](https://github.com/gke-labs/kube-agents/tree/main/terraform/examples/full-install#teardown)
+instead. In particular, do not run `teardown_08_deploy_platform_agent.sh` against a
+Terraform-managed cluster: it removes the `kube-agents-host` label that Terraform owns and creates
+plan drift.
+
 ```bash
 cd k8s-operator/scripts
 ./teardown.sh
@@ -64,7 +71,7 @@ You can also run individual `teardown_NN_*.sh` scripts to remove one layer at a 
 | `teardown_11_deploy_inference_replay.sh` | Inference-replay proxy + PVC; restores original LiteLLM Service                            |
 | `teardown_10_deploy_github_minter.sh`    | Minty deployment, GSAs, KMS resources                                                      |
 | `teardown_09_deploy_litellm.sh`          | LiteLLM Gateway                                                                            |
-| `teardown_08_deploy_platform_agent.sh`   | `PlatformAgent` CR and rendered manifests                                                  |
+| `teardown_08_deploy_platform_agent.sh`   | `PlatformAgent` CR, script-managed host-discovery label, and rendered manifest             |
 | `teardown_07_gcp_k8s_secrets.sh`         | Kubernetes secrets in the target namespace                                                 |
 | `teardown_06_slack.sh`                   | Slack tokens and state                                                                     |
 | `teardown_05_gcp_gchat.sh`               | Google Chat Pub/Sub topic + subscription                                                   |
--- a/k8s-operator/scripts/README.md
+++ b/k8s-operator/scripts/README.md
@@ -112,7 +112,7 @@ Generated from each script's own comment banner.
 | 5 | [`provision_05_gcp_gchat.sh`](provision_05_gcp_gchat.sh) | **Google Chat & Pub/Sub Setup** — Configures the Google Chat backend: Pub/Sub routing, the Agent's Service Account, and grants the Service Account permission to read incoming chat messages. Also enables the Workspace Add-ons and Chat APIs and provisions their service identities — without the Chat API identity, Google Chat fails silently. |
 | 6 | [`provision_06_slack.sh`](provision_06_slack.sh) | **Slack Integration Setup** — Configures Slack bot tokens, app tokens, and home channel settings. |
 | 7 | [`provision_07_gcp_k8s_secrets.sh`](provision_07_gcp_k8s_secrets.sh) | **GKE Kubernetes Secrets Setup** — Idempotent setup script to validate and configure Kubernetes secrets directly. Requires CMEK database encryption unless ALLOW_UNENCRYPTED_SECRETS=true is set. |
-| 8 | [`provision_08_deploy_platform_agent.sh`](provision_08_deploy_platform_agent.sh) | **Deploy PlatformAgent Custom Resource Manifest** — Idempotent script that connects to GKE, renders the platform-agent.yaml template, deploys it to the cluster, and fails unless the operator reconciles the change into the agent Deployment (override the wait budget with AGENT_READY_TIMEOUT, default 600s). Whether the Deployment then rolls out is verified by step 13, after the agent's dependencies exist. |
+| 8 | [`provision_08_deploy_platform_agent.sh`](provision_08_deploy_platform_agent.sh) | **Deploy PlatformAgent Custom Resource Manifest** — Idempotent script that connects to GKE, renders the platform-agent.yaml template, deploys it to the cluster, labels the host cluster for discovery, and fails unless the operator reconciles the change into the agent Deployment (override the wait budget with AGENT_READY_TIMEOUT, default 600s). Whether the Deployment then rolls out is verified by step 13, after the agent's dependencies exist. |
 | 9 | [`provision_09_deploy_litellm.sh`](provision_09_deploy_litellm.sh) | **Deploy LiteLLM Gateway** — Idempotent script that connects to GKE and deploys the LiteLLM Gateway. |
 | 10 | [`provision_10_deploy_github_minter.sh`](provision_10_deploy_github_minter.sh) | **Deploy GitHub Token Minter** — Idempotent script that deploys the GitHub Token Minter. Runs only when GITHUB_ORG, GITHUB_REPO, and GITHUB_APP_ID are all set; skipped otherwise. |
 | 11 | [`provision_11_deploy_inference_replay.sh`](provision_11_deploy_inference_replay.sh) | **Deploy Inference Replay Proxy (optional)** — Idempotent script that deploys the Inference Replay proxy in front of the LiteLLM gateway. Skipped unless INFERENCE_REPLAY_ENABLED=true. The proxy intercepts the `litellm` Service so agents need no configuration changes. With REPLAY_MODE=off (default) it is a pure pass-through; flip the `inference-replay-config` ConfigMap to `on` to start recording/replaying. |
@@ -131,7 +131,7 @@ Generated from each script's own comment banner.
 | 5 | [`teardown_05_gcp_gchat.sh`](teardown_05_gcp_gchat.sh) | **Teardown Google Chat & Pub/Sub Setup** — Idempotent script to clean up GChat Pub/Sub Topic/Subscription and the Bot GSA. |
 | 6 | [`teardown_06_slack.sh`](teardown_06_slack.sh) | **Teardown Slack Integration Setup** — Idempotent script to clean up Slack integration state and tokens. |
 | 7 | [`teardown_07_gcp_k8s_secrets.sh`](teardown_07_gcp_k8s_secrets.sh) | **Teardown GKE Secrets** — Idempotent script to clean up Kubernetes secrets and sanitize local state. |
-| 8 | [`teardown_08_deploy_platform_agent.sh`](teardown_08_deploy_platform_agent.sh) | **Teardown PlatformAgent Custom Resource** — Idempotent script to clean up the applied PlatformAgent Custom Resource (CR) and delete the local generated manifest file. |
+| 8 | [`teardown_08_deploy_platform_agent.sh`](teardown_08_deploy_platform_agent.sh) | **Teardown PlatformAgent Custom Resource** — Idempotent script to clean up the applied PlatformAgent Custom Resource (CR), remove the host-discovery label, and delete the local generated manifest file. |
 | 9 | [`teardown_09_deploy_litellm.sh`](teardown_09_deploy_litellm.sh) | **Teardown LiteLLM Gateway** — Idempotent script to undeploy the LiteLLM gateway. |
 | 10 | [`teardown_10_deploy_github_minter.sh`](teardown_10_deploy_github_minter.sh) | **Teardown GitHub Token Minter** — Idempotent script to clean up the GitHub Token Minter. |
 | 11 | [`teardown_11_deploy_inference_replay.sh`](teardown_11_deploy_inference_replay.sh) | **Teardown Inference Replay Proxy** — Idempotent script to undeploy the Inference Replay proxy and restore the original LiteLLM Service. Safe to run even when the proxy was never deployed. |
@@ -166,6 +166,12 @@ To run a dry-run check (simulates commands without modifying cloud resources):
 ```bash
 ./provision.sh --dry-run
 ```
+
+After the `PlatformAgent` resource is applied, provisioning labels its GKE cluster
+`kube-agents-host=true` so the admin portal can discover the host. Label registration is advisory:
+if it fails, provisioning continues and the portal can still connect through manual cluster
+selection. The in-repository runtime receives its host identity from the configured
+`PlatformAgent` cluster name rather than discovering it by label.
 
 ### Run Teardown Pipeline
 
--- a/k8s-operator/scripts/common.sh
+++ b/k8s-operator/scripts/common.sh
@@ -37,6 +37,10 @@ else
   C_WHITE='\033[97m'
 fi
 
+# Stable project-level discovery marker for the GKE cluster hosting kube-agents.
+# Keep this value aligned with the Terraform full-install composition and admin portal.
+KUBE_AGENTS_HOST_LABEL="kube-agents-host"
+
 # ─── UI Helpers ───────────────────────────────────────────────────────────────
 print_step() { echo -e "\n${C_MAGENTA}${C_BOLD}>>>  $1  <<<${C_RESET}"; }
 print_success() { echo -e "  ${C_GREEN}✓ $1${C_RESET}"; }
@@ -686,6 +690,29 @@ cluster_exists() {
   gcloud container clusters list --filter="name=${CLUSTER_NAME} AND location=${REGION}" --format="value(name)" --project="${PROJECT_ID}" 2>/dev/null || echo ""
 }
 
+host_label_value() {
+  gcloud container clusters describe "$CLUSTER_NAME" \
+    --location="$REGION" \
+    --project="$PROJECT_ID" \
+    --format="value(resourceLabels.${KUBE_AGENTS_HOST_LABEL})" 2>/dev/null
+}
+
+verify_host_label() {
+  [ "$(host_label_value)" = "true" ]
+}
+
+update_host_label() {
+  gcloud container clusters update "$CLUSTER_NAME" \
+    --location="$REGION" \
+    --project="$PROJECT_ID" \
+    --update-labels="${KUBE_AGENTS_HOST_LABEL}=true" \
+    --quiet
+}
+
+execute_host_label() {
+  retry 3 5 update_host_label
+}
+
 connect_cluster() {
   print_info "Fetching cluster credentials..."
   gcloud container clusters get-credentials "$CLUSTER_NAME" --location "$REGION" --project "$PROJECT_ID" --quiet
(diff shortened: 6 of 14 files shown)


mplakhtiy:
## Review notes

Reviewed at `81100f6a`. Findings only — claims below were verified against the source or reproduced with a stubbed `gcloud`.

### Blocking

**1. A failed label removal skips the rest of teardown** — `teardown_08_deploy_platform_agent.sh:77-81`

The file runs under `set -euo pipefail` and the new `gcloud container clusters update --remove-labels` is unguarded, unlike every other destructive action here (`|| true`, `--ignore-not-found`, the finalizer fallback). Reproduced with a stubbed `gcloud` that fails the update:

```
INFO: Removing the label...
ERROR: PERMISSION_DENIED
EXIT=1          # ">>> STEP 4 REACHED" never printed
```

Step 4 (deleting the local `platform-agent.yaml`) never runs. `teardown.sh:35` appends `|| true` at the call site, which hides the failure from the pipeline but does not un-skip Step 4.

```bash
gcloud container clusters update "$CLUSTER_NAME" ... --quiet \
  || print_warning "Could not remove the '${KUBE_AGENTS_HOST_LABEL}' label; remove it manually."
```

**2. The confirmation prompt cannot be bypassed, and fails silently without a TTY** — `provision_08_deploy_platform_agent.sh:84-94`

Two symptoms of one defect — the guard omits the non-interactive checks that `common.sh:255` already centralises in `is_non_interactive()`:

- **`-y` doesn't reach the step.** `common.sh:94` sets `NO_CONFIRM=0` *without `export`*, and `provision.sh:34` forwards only `$DRY_RUN_ARG`. (Contrast `teardown.sh:32-45`, which passes `--no-confirm` to every step.) So `./provision.sh -y` still blocks on the prompt; the `NO_CONFIRM` branch is only reachable when the step script is invoked directly.
- **No `[ ! -t 0 ]` check.** Under `set -e`, `read` returning 1 at EOF exits the script *at the `read` itself* — the `is_truthy` branch and the `print_error "Provisioning stopped…"` message never execute. Confirmed with a harness: it printed the prompt, then exited 1 with no diagnostic.

The non-TTY path is narrower than a blanket regression, since `init_var`'s `read` dies the same way — but it is reachable on the common re-run: `vars.sh` already populated (so `init_var` never prompts), stdin not a TTY, `CI` unset, and a duplicate host present. The operator gets a bare exit 1.

Both are fixed by one substitution:

```bash
if is_non_interactive; then return 0; fi   # DRY_RUN, NO_CONFIRM, CI, and !tty
```

**3. Reimplements `confirm_action`** — `common.sh:477`

That helper already performs the identical guard → prompt → `is_truthy` sequence. One divergence to preserve deliberately or drop: `confirm_action` calls `exit 0` on decline, the new code returns 1 → `set -e` → exit 1. Non-zero is the better choice for a provisioning abort, but it reads as incidental rather than intended.

### Moderate

**4. The Terraform install path never sets the label** — `terraform/modules/gke-cluster/main.tf`

`google_container_cluster.autopilot` declares `name`, `location`, `project`, `enable_autopilot`, `deletion_protection`, `workload_identity_config`, and `release_channel` — no `resource_labels`. A Helm chart cannot label a GCP resource either. The one-command IaC install landed in the immediately preceding commit (#493), so the two supported install paths now diverge: script-installed hosts are discoverable, Terraform-installed hosts are not. Add `resource_labels = { "kube-agents-host" = "true" }` (or a variable) to the module — otherwise the README's "project-level tooling can discover the host" claim only holds for the script path and should be scoped accordingly.

**5. The teardown banner wasn't updated → docs drift**

The provision banner was updated correctly; I confirmed against `parse_script_banner` (`scripts/generate_docs.py:226`) that its two lines join to exactly the committed table row, so `make docs-check` passes. But `teardown_08`'s banner (lines 3-6) is unchanged and still describes only CR + manifest cleanup, while the script now also removes a cloud resource label. `make docs-check` will **not** catch this — the generated region is consistent with the unchanged banner. This is what the `review-docs-drift` pre-PR step exists for.

**6. `verify_host_label` lists every cluster in the project to check one** — `provision_08_deploy_platform_agent.sh:103-111`

Teardown already uses the cheaper, direct `describe`. Mirror it:

```bash
[ "$(gcloud container clusters describe "$CLUSTER_NAME" --location="$REGION" \
     --project="$PROJECT_ID" --format="value(resourceLabels.${KUBE_AGENTS_HOST_LABEL})" 2>/dev/null)" = "true" ]
```

**7. `--location` (provision) vs `--region` (teardown)** — `provision_08_deploy_platform_agent.sh:116` / `teardown_08_deploy_platform_agent.sh:69,78`

Both work for regional clusters; `--location` is the general flag — pick one. There's a latent leak in teardown specifically: `connect_cluster || true` (line 36) lets the script continue past a location mismatch, then `describe --region` fails, `2>/dev/null || echo ""` swallows it, `HOST_LABEL` is empty, and the label is silently left behind. `cluster_exists()` (`common.sh:433`) uses a prefix match `location:${REGION}*`, so the codebase does contemplate location ≠ REGION.

### Nits

- **Label name breaks the repo convention.** Everything else is `kubeagents-*` — namespace `kubeagents-system`, GSAs `kubeagents-platform-gsa`, CRD group `kubeagents.x-k8s.io` (`common.sh:295-302`). `kube-agents-host` is the outlier, and a GCP resource label is awkward to rename once deployed.
- **Imperative call inside the function-definition block** (`provision_08_deploy_platform_agent.sh:122`). Every other script keeps execution under `─── Execution Pipeline ───`. Move it there (still before `run_step 1`).
- **No consumer for the label.** `git grep kube-agents-host` matches only the two scripts and the README; the host is still identified by hardcoded name / `GKE_CLUSTER_NAME` (`agents/platform/scripts/session_kv_server.py:252`). Fine as groundwork — worth saying so in the PR description.
- **Consider `retry` for the label update.** `common.sh:49` already provides it; `retry 3 5 execute_host_label` would absorb a transient "cluster is currently being updated" `FAILED_PRECONDITION`. As written that exits 1 with the CR deployed but the host unlabeled (a rerun fixes it).
- **`warn_if_duplicate_host` calls gcloud under `--dry-run`** (line 62, before any DRY_RUN check). Read-only and the failure path is handled, but with a placeholder `PROJECT_ID` it emits a spurious warning.

### Checked and clear

I simulated the parsing loop against empty output, unlabeled clusters (trailing empty tab field), a missing trailing newline, and a `gcloud` failure: all correct, and `set -e` does not misfire on the `[ … ] && return 0` guards.

---

<sub>Review generated with the help of Claude.</sub>

mplakhtiy:
## Findings

### 1. HIGH — `-y` silently grows from "skip one prompt" to "approve everything", including a live control-plane update

`k8s-operator/scripts/provision.sh:19-21`, consumed at
`k8s-operator/scripts/provision_01_gcp_cluster.sh:141`

```bash
if [ "$NO_CONFIRM" -eq 1 ]; then
  STEP_ARGS+=("--no-confirm")
fi
```

`NO_CONFIRM` is a plain (unexported) assignment in `common.sh:98`, so before this PR `-y` stopped at
`provision.sh` and never reached a child step. Forwarding it in `STEP_ARGS` hands it to all twelve
steps plus both instruction scripts.

**As authored**, that means `provision_01_gcp_cluster.sh:141` no longer asks:

```bash
confirm_action "Existing GKE Cluster '$CLUSTER_NAME' database encryption is not enabled. Updating it will modify the live cluster control plane." \
```

`confirm_action` (`common.sh:566`) returns 0 immediately when `NO_CONFIRM=1`, so `./provision.sh -y`
against a pre-existing unencrypted cluster now issues
`gcloud container clusters update --database-encryption-key` against a running control plane with no
prompt. That is the one confirmation in the provisioning path that guards a mutation of a live
cluster, and this PR removes the operator from that decision.

**After merging current `main`**, the blast radius grows again. #519 changed `init_var` to key off
`is_non_interactive` rather than `[ DRY_RUN ] || is_ci_pipeline`:

```bash
# main @ b63075c5, k8s-operator/scripts/common.sh:175
if is_non_interactive; then
  final_val="$default_val"
```

and `is_non_interactive` (`common.sh:312`) is true whenever `NO_CONFIRM=1`. Both halves are
verifiable in source today; neither is touched by the conflicting hunk. The merged result is that
`./provision.sh -y` on a fresh checkout stops prompting for every unset variable in every step and
takes the default instead — `ENABLE_GVISOR=false`, `GOOGLE_CHAT_ENABLED=false`,
`SLACK_ENABLED=false`, the KMS keyring and key names, and most importantly
`ALLOWED_USERS` (`provision_05_gcp_gchat.sh:64`) and `SLACK_ALLOWED_USERS`
(`provision_06_slack.sh:126`), whose own prompt text reads *"Leaving it empty will allow all
users."* An operator who exports `GOOGLE_CHAT_ENABLED=true` and runs `provision.sh -y` gets a chat
integration open to every user in the workspace and is never asked.

The documentation added in this same PR describes none of this.
`docs/site/src/content/docs/install/quickstart-gke.mdx:84` says:

```
- `-y`, `--no-confirm` — bypasses confirmation when another labeled host already exists.
```

**Failure scenario:** an operator reads the new quickstart line, understands `-y` to suppress a
single duplicate-host warning, and runs `./provision.sh -y` on a project with an existing
unencrypted cluster. The control plane is reconfigured for CMEK without a prompt; post-merge, the
Google Chat allowlist is also silently set to "everyone".

**Fix:** either scope the forwarding to what the PR needs — export a dedicated variable for the
duplicate-host check, or forward `--no-confirm` only to `provision_08` — or accept the wider meaning
deliberately and say so in the quickstart, in which case `provision_01:141` needs its own opt-out
that `-y` does not satisfy.

### 2. MEDIUM — A label with no consumer hard-fails the provisioning pipeline before LiteLLM is deployed

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:212`

```bash
run_step "3. Register kube-agents host" verify_host_label execute_host_label 0
```

`run_step` (`common.sh:440`) calls `exit 1` when the execute function fails:

```bash
else
  print_error "Failed to execute step: $name"
  exit 1
fi
```

`provision.sh` runs under `set -e` with no `|| true` at the call site, so a failed
`gcloud container clusters update --update-labels` aborts the pipeline at step 8 of 12. Steps 09
(LiteLLM Gateway), 10 (GitHub token minter), 11 (inference replay) and 12 (backup plan) never run —
and step 09 is what the agent actually talks to. The install ends with a `PlatformAgent` CR applied
and no gateway behind it.

The PR description states the case plainly: *"There is intentionally no in-repository consumer yet."*
A marker that nothing reads should not be able to stop the components that do.

`execute_host_label` wraps the update in `retry 3 5`, which covers roughly fifteen seconds. GKE
`clusters update` rejects concurrent work with `FAILED_PRECONDITION` for as long as the in-flight
operation runs, which is minutes, so the retry does not cover the failure mode it was added for.

**Failure scenario:** a provisioning principal without `container.clusters.update`, or a cluster with
another operation in flight, leaves the operator with an aborted install and no LiteLLM gateway,
recoverable only by re-running the pipeline.

**Fix:** treat step 3 as advisory — call it outside `run_step`, or wrap it so a failure emits
`print_warning` and returns 0. The label is idempotent and a later run reapplies it.

### 3. MEDIUM — The duplicate-host warning fires too late to prevent what it warns about

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:209`

```bash
# ─── Execution Pipeline ───────────────────────────────────────────────────────
warn_if_duplicate_host
```

The check sits at the top of step 08's pipeline, which means steps 01 through 07 have already run:
the GKE cluster is created (5–8 minutes and billing), the gVisor node pool exists, the operator is
deployed, the GSAs and IAM bindings are in place, the Pub/Sub topic and subscription are created,
and the Kubernetes secrets are written. Declining the prompt returns 1, `set -e` exits step 08, and
`provision.sh` aborts — leaving all of that behind.

The message is technically accurate — `"Provisioning stopped before the PlatformAgent was changed."`
— but the PR's stated goal is that *"Operators see project-level duplication before deployment
continues without automatically changing another deployment."* By step 08 the deployment has
substantially happened; only the CR is missing. An operator who answers "no" now has to run
`teardown.sh` to undo the seven steps the warning did not prevent.

The project-level lookup needs nothing from steps 01–07 — it takes `PROJECT_ID` and `CLUSTER_NAME`,
both available at `load_state` time.

**Failure scenario:** an operator points a second install at a project that already hosts
kube-agents, sees the duplicate warning for the first time nine minutes in, declines, and is left
with an orphaned Autopilot cluster and a full set of IAM bindings to clean up by hand.

**Fix:** run `warn_if_duplicate_host` from `provision.sh` before step 01, or from
`provision_01_gcp_cluster.sh`, so declining costs nothing.

### 4. MEDIUM — The `cluster_exists` rewrite is not needed for this PR and is the sole cause of the merge conflict

`k8s-operator/scripts/common.sh:500`

```bash
cluster_exists() {
  gcloud container clusters describe "$CLUSTER_NAME" \
    --location "$REGION" \
    --project "$PROJECT_ID" \
    --format="value(name)" 2>/dev/null || echo ""
}
```

`main` fixed the same line independently in #519, keeping `list` and tightening the filter from a
prefix match to an equality match:

```bash
# main @ b63075c5
gcloud container clusters list --filter="name=${CLUSTER_NAME} AND location=${REGION}" ...
```

The two are behaviourally equivalent for every caller — all eight consumers are teardown scripts
that only test the result for emptiness — and `main`'s version already achieves the
location-precision this PR was after. `connect_cluster`'s `--region` → `--location` change in this
PR is likewise already on `main`, character for character.

AGENTS.md, *Pull Request Hygiene*: "Keep changes scoped to the request." The request is host
labeling; `cluster_exists` is a discovery helper that no part of the labeling feature calls.

**Failure scenario:** the PR is unmergeable today solely because of this hunk. Resolving it also
costs a reviewer a decision between two equivalent implementations that nothing in the feature
depends on.

**Fix:** drop the `cluster_exists` and `connect_cluster` hunks and take `main`'s versions. The
conflict disappears and the diff shrinks to the feature.

### 5. LOW — The host-label read is duplicated across the provision and teardown scripts

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:97` and
`k8s-operator/scripts/teardown_08_deploy_platform_agent.sh:77`

Both run the identical query:

```bash
gcloud container clusters describe "$CLUSTER_NAME" \
  --location="$REGION" --project="$PROJECT_ID" \
  --format="value(resourceLabels.${KUBE_AGENTS_HOST_LABEL})"
```

`KUBE_AGENTS_HOST_LABEL` already lives in `common.sh` precisely so the two paths cannot drift; the
query that reads it belongs next to it. A single `host_label_value()` in `common.sh` would let
`verify_host_label` become `[ "$(host_label_value)" = "true" ]` and let teardown reuse it.

`teardown_08` also pays for the same call twice: `cluster_exists()` at line 34 now issues a
`clusters describe`, and line 77 issues another against the same cluster moments later. One
`describe` requesting both `name` and `resourceLabels.<label>` would serve both.

**Cost:** two copies of a projection string that must stay in sync by hand, plus a redundant GCP API
round trip per teardown.

### 6. LOW — Teardown removes a label that Terraform declares, creating plan drift for Terraform-installed hosts

`k8s-operator/scripts/teardown_08_deploy_platform_agent.sh:91`

```bash
--remove-labels="$KUBE_AGENTS_HOST_LABEL" \
```

`terraform/modules/gke-cluster/main.tf:38-40` declares the same label as managed state:

```hcl
resource_labels = merge(var.resource_labels, {
  "kubeagents-host" = "true"
})
```

The PR's premise is that the two install paths apply the same marker, but only one of them owns it.
A Terraform install followed by `teardown_08_deploy_platform_agent.sh` — a plausible sequence, since
the shell teardown is the documented way to remove the agent layer while keeping the cluster —
strips a Terraform-managed attribute out of band.

**Failure scenario:** the next `terraform plan` on a Terraform-installed host reports an unexplained
`resource_labels` diff and silently re-adds the label on apply, so teardown does not stick.

**Fix:** scope the label removal to the shell-provisioned path, or document in
`docs/site/src/content/docs/install/uninstall.md` that Terraform installs must be torn down with
`terraform destroy`.

### 7. LOW — None of the new shell logic is exercised by CI

`k8s-operator/scripts/provision_08_deploy_platform_agent.sh:52-118`,
`k8s-operator/scripts/teardown_08_deploy_platform_agent.sh:34-115`

The PR adds a tab-delimited `gcloud` output parser, a self-exclusion comparison against
`CLUSTER_NAME`/`REGION`, a four-way teardown state machine (`cluster_present`, `kubectl_connected`,
`cleanup_failed`, and the deferred non-zero exit), and a new argument-forwarding path — roughly 120
lines of branching shell. `make validate` checks skill placement, and the repository has no shell
test harness at all, so CI verifies none of it. The PR's testing notes list stubbed reproductions
run by hand, which do not persist.

**Untested behaviours that would regress silently:** a `gcloud` format change that reorders the
`value(name,location,resourceLabels.…)` fields would make every cluster look unlabeled and the
duplicate warning would never fire again; and the `cleanup_failed` accumulator could stop
propagating, so teardown would report success after failing to remove the label.

Adding a harness is out of proportion for this PR. Naming the gap here is the point — the
duplicate-host parser in particular is the kind of code that fails quietly.


---

haoxuw:
hi @mplakhtiy thanks for the review, addressed them. Please have a look thanks

haoxuw:
@mplakhtiy Thanks for the thorough reviews — addressed in `db411ae`:

- Removed the global `-y` forwarding and the late duplicate-host prompt; retained current `main` cluster helpers.
- Made label registration advisory, centralized label reads, and moved teardown label failure after local cleanup.
- Scoped Terraform ownership to the full-install composition and documented its teardown path.
- Kept `kube-agents-host` because the admin console now consumes that exact label.
- Added CI-discovered shell tests and regenerated/verified the docs.

All repository CI checks pass; only the external smoke test is still pending.

haoxuw:
Regarding the -y flag, I removed the duplicate-host prompt entirely, so no new --no-confirm forwarding is needed. The original -y behavior remains unchanged

docs/README.md
```diff
@@ -140,6 +140,7 @@ identifier appears, add its source here.
 | Image defaults and override env vars (`PLATFORM_AGENT_IMAGE` et al.) | `k8s-operator/internal/controller/manifest_helpers.go` |
```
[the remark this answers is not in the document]
haoxuw: invalid

hack/check-docs-terminology.sh
```diff
@@ -63,6 +63,27 @@ forbid 'platform-agent-ksa' \
...
+fi
+
+WRONG_HOST_LABEL=$(search 'kube-?agents-host=true' | grep -vF "${HOST_LABEL}=true" || true)
+if [ -n "$WRONG_HOST_LABEL" ]; then
+  echo "::error::Documented host-discovery label does not match KUBE_AGENTS_HOST_LABEL=${HOST_LABEL}."
+  printf '%s\n\n' "$WRONG_HOST_LABEL" | sed 's/^/    /'
+  FAILED=1
+fi
+
+if ! grep -qF "\"${HOST_LABEL}\" = \"true\"" terraform/examples/full-install/main.tf; then
```
mplakhtiy: **MEDIUM — this fixed-string match is whitespace-brittle, and `terraform fmt` will break it.**

`grep -F` pins exactly one space on either side of the `=`. `terraform fmt` aligns the `=` columns of adjacent entries in an object literal, and `.github/workflows/validate.yml:72` runs `terraform fmt -check -recursive terraform/` — so the formatting this guard depends on is not the author's to keep.

**Failure scenario:** someone adds a second label to the composition, e.g. `"kube-agents-install-path" = "terraform"`. `terraform fmt` rewrites the block so the existing entry becomes `"kube-agents-host"         = "true"`, and `make docs-check-terminology` then fails with `::error::Terraform full-install composition does not apply kube-agents-host=true` — about a file that applies it on the line above. The two gates contradict each other and no formatting satisfies both. Reproduced end to end locally.

A whitespace-tolerant ERE fixes it:

```suggestion
if ! grep -qE "\"${HOST_LABEL}\"[[:space:]]*=[[:space:]]*\"true\"" terraform/examples/full-install/main.tf; then
```

hack/check-docs-terminology.sh
```diff
@@ -63,6 +63,27 @@ forbid 'platform-agent-ksa' \
...
+  echo "::error::Terraform full-install composition does not apply ${HOST_LABEL}=true."
+  FAILED=1
+fi
```
mplakhtiy, line 85: **MEDIUM — the guard covers the docs and the Terraform copy, but not the only code that reads the label.**

The rationale in the comment above ("must mirror the stable key because it cannot source Bash") applies just as much to the consumer, and the consumer is unchecked:

```python
# admin_console/connections.py:413
(item.get("resourceLabels") or {}).get("kube-agents-host", "")
```

**Failure scenario:** `KUBE_AGENTS_HOST_LABEL` changes in `common.sh`. The guard forces the Markdown and Terraform copies to follow and goes green; `connections.py` still filters on the old key, so every cluster reads as unlabeled, the console reports "no cluster is labeled kube-agents-host=true", and auto-selection — the behaviour this PR exists to enable — is silently gone on every install, with no failing check anywhere. Note the doc probe on line 25 only scans `git ls-files '*.md' '*.mdx'`, so none of the ~10 occurrences under `admin_console/*.py` are reachable by it. #681 adds more.

One more assertion in the same style as the Terraform one:

```suggestion
fi

if ! grep -qF "\"${HOST_LABEL}\"" admin_console/connections.py; then
  echo "::error::admin_console/connections.py does not read ${HOST_LABEL}."
  FAILED=1
fi
```

Alternatively, have `connections.py` read the key from a single named constant the guard can check.

Two further copies also sit outside the guard's reach, if you want them in the same pass: `terraform/modules/gke-cluster/variables.tf:28` spells the label in its description (no `.tf` file outside the one hard-coded path is scanned), and the HCL block at `terraform/modules/gke-cluster/README.md:23` writes it as `"kube-agents-host" = "true"`, which the `kube-?agents-host=true` probe does not match.

mplakhtiy (commented):
Two findings on the new host-discovery guard in `hack/check-docs-terminology.sh`. Both are about the guard itself, not the shipped install behaviour — the label registration and teardown paths read correctly.

mplakhtiy (approved):
Approving. The two inline comments are non-blocking guard-quality items — the install behaviour itself reads correctly: registration is genuinely advisory, and teardown removes the label only after local cleanup.

mplakhtiy (approved):

hack/check-docs-terminology.sh
```diff
@@ -63,6 +63,27 @@ forbid 'platform-agent-ksa' \
...
+if [ -n "$WRONG_HOST_LABEL" ]; then
+  echo "::error::Documented host-discovery label does not match KUBE_AGENTS_HOST_LABEL=${HOST_LABEL}."
+  printf '%s\n\n' "$WRONG_HOST_LABEL" | sed 's/^/    /'
+  FAILED=1
+fi
+
+if ! grep -qE "\"${HOST_LABEL}\"[[:space:]]*=[[:space:]]*\"true\"" terraform/examples/full-install/main.tf; then
+  echo "::error::Terraform full-install composition does not apply ${HOST_LABEL}=true."
+  FAILED=1
+fi
```
mplakhtiy, line 85: **MEDIUM — the drift guard covers the docs and the Terraform copy, but not the only code that reads the label.**

The guard derives `HOST_LABEL` from `common.sh`, greps `git ls-files '*.md' '*.mdx'` (line 25) for stale spellings, and asserts the literal in `terraform/examples/full-install/main.tf`. Its own comment states the rationale: *"The Terraform full-install composition must mirror the stable key because it cannot source Bash."* Exactly the same is true of the consumer, and the consumer is not checked:

```python
# admin_console/connections.py:449
(item.get("resourceLabels") or {}).get("kube-agents-host", "")
```

**Failure scenario.** `KUBE_AGENTS_HOST_LABEL` is changed in `common.sh`. The guard forces the Markdown and Terraform copies to follow and goes green; `scripts/test_host_label_provisioning.py` follows too, because it asserts on the shell's own output. `connections.py` still filters on the old key, so every cluster reads as unlabeled, `host_clusters` is empty, and the console degrades to `"No cluster has the kube-agents-host=true label."` — auto-selection, the behaviour this PR exists to enable, is silently gone on every install with no failing check anywhere.

Discovery does at least fail *closed* rather than returning every cluster, so this is a silent loss of function, not a mis-targeting.

`admin_console/tests/test_connections.py:46` does not help: its fixture hardcodes the same `"kube-agents-host"` string, so both sides of the consumer drift together and the test stays green. The exposure has grown since the last round — the literal now appears at ten sites under `admin_console/`, up from one.

Two further copies also sit outside the guard's reach. `terraform/modules/gke-cluster/variables.tf:28` spells the label in its description, but the probe only ever runs over `git ls-files '*.md' '*.mdx'`, so no `.tf` file outside the one hard-coded path is looked at. The HCL block at `terraform/modules/gke-cluster/README.md:23` *is* in a scanned file, but writes the label as `"kube-agents-host" = "true"`, which the probe `kube-?agents-host=true` does not match.

**Fix.** Extend the guard with a `grep -qF "\"${HOST_LABEL}\"" admin_console/connections.py` assertion in the same style as the Terraform one, or have `connections.py` read the key from a single named constant the guard can check.

Also worth noting: `make iac-parity-check` does not exist in this tree — no `scripts/check_iac_parity.py`, no `parity` rule in the Makefile or any workflow, despite AGENTS.md describing both. That is a pre-existing gap in `main` rather than this PR's doing, but it means no automated parity gate backstops this.

mplakhtiy (commented):
Re-review of `d3802e6` (force-pushed over `935765d2`). One Medium inline; a second below that has no line to anchor to.

The force-push resolved the previous round's Finding 1 — the Terraform assertion is now a whitespace-tolerant ERE, and I re-ran the reproduction: added `"kube-agents-install-path" = "terraform"`, let `terraform fmt` realign to `"kube-agents-host"         = "true"`, and the guard exited 0. The other six findings still stand.

## Findings outside this diff

### MEDIUM — the pull-request body omits the required Live validation section for a change that mutates a live cluster

AGENTS.md, *Pull Request Hygiene*, names itself canonical:

> **Live-test the change before opening a PR, and describe it in the PR body.** Every pull request fills in the template's **Testing → Live validation** section with how the change was exercised against a real, running kube-agents installation … Green unit tests and a clean `make docs-check` are necessary, not sufficient. … **If the change cannot reach a running installation** — docs-only, a CI workflow, a code path that needs infrastructure you do not have — write "Not live-tested" and say why. An empty section is not an answer.

The body's `## Testing` list is nine static checks — `unittest`, `bash -n`, `go build`, `terraform validate`, `make docs-check`, prettier, `npm run build`, `git diff --check`. There is no `### Live validation` heading at all. `kube-agents-bot` raised this (comment `3767951457`) and the reply was the single word "invalid" (comment `3776221425`), with no reasoning. The force-push did not add the section.

The exemption does not apply here. This change issues `gcloud container clusters update --update-labels=kube-agents-host=true` against a running GKE cluster from step 08, and `--remove-labels` from teardown 08. It is precisely the class of change the rule exists for, and the two behaviours a reviewer cannot verify from the diff are the two that matter: that a real `clusters update` on an Autopilot cluster succeeds and is picked up by `gcloud container clusters describe --format="value(resourceLabels.kube-agents-host)"`, and that a denied update leaves provisioning running through steps 09-13.

**Failure scenario.** The projection string in `host_label_value` is the single expression both scripts depend on, and no test exercises it against real `gcloud` output — the unit tests stub `gcloud` and ignore `--format`. If it returns empty for a labelled cluster, provisioning re-issues a control-plane update on every run and teardown silently leaves the label behind: a pair of regressions only a live run can rule out, and that CI is structurally unable to see.

**Fix.** Run step 08 and teardown 08 against a real cluster, record the `describe` output before and after, and fill in the section. If a cluster is not available, say "Not live-tested" and why.

PR #550, opened 2026-08-05
feat(admin): add Kube Agents Console

haoxuw:
# Pull Request

## Why This Change

Operators need a first-party local surface for connecting to a kube-agents deployment, reviewing retained conversations and agent work, chatting without depending on Google Chat or Slack, and investigating live telemetry with explicit attribution and completeness boundaries.

## What Changed

**Files:**

- `admin_console/`
- `.streamlit/config.toml`
- `scripts/admin_portal.sh`
- `docs/designs/admin-console.md`
- `docs/README.md`
- `agents/chat/defaults/plugins/bootstrap_onboarding/`

**Added/Changed/Fixed:**

- Added a loopback-only Streamlit console with launcher-verified gcloud identity and bounded read paths.
- Added one editable, source-labeled project selector plus shared cluster connection controls, labeled-host discovery, manual fallback, diagnostics, secure target persistence, ten-minute revalidation, and disabled animated working states.
- Added a full-width, filterable session table with URL-stable selection and transcripts below it.
- Added portal-native agent chat, bounded approval handling, retained cross-user history, and linked Task Kanban progress.
- Added Overview, Activity Explorer, Task Kanban, and Scheduled Cron views backed by live Cloud Logging, Cloud Trace, and agent-runtime read models.
- Added progressive Cloud Trace pagination, evidence redaction, result-limit reporting, and reusable connection gates.
- Bounded conversation and Kanban run queries before loading metadata and details.
- Made first-turn onboarding silent on request/response surfaces that cannot receive asynchronous delivery.
- Pinned Streamlit's native theme to the console's dark palette so tables and controls render consistently across browsers and headless clients.
- Kept deterministic telemetry fixtures under `admin_console/tests/`; production pages use only live providers.
- Added unit and Streamlit functional coverage plus the console design and operating documentation.

## Why This Matters

- Gives platform teams an always-available local chat and inspection surface independent of third-party chat integrations.
- Makes human-triggered and autonomous agent activity inspectable without fabricating missing causal links.
- Keeps Google credentials out of browser state and persisted portal state while still making local connections durable.
- Provides actionable connection diagnostics instead of failing provider-backed pages implicitly.

## Context

- Related host-label provisioning work: #546.
- The current launcher is deliberately local-only. A remotely served console still requires application-level authentication and authorization, a narrow chat/history API, and production retention and redaction policy.
- Google Chat and Slack conversations remain read-only in the portal; portal follow-ups use separate attributed sessions.

## Testing

- `.venv/bin/python -m unittest discover -s admin_console/tests -v` — 63 tests passed.
- `.venv/bin/python agents/chat/defaults/plugins/bootstrap_onboarding/test_plugin.py` — 14 tests passed.
- `.venv/bin/python -m compileall -q admin_console` — passed.
- `make docs-check` — passed.
- `make docs-generate` — generated regions remained current.
- `npx prettier --check <changed Markdown/YAML/JSON>` — passed.
- `git diff --check` — passed.
- Local Streamlit health endpoint returned `ok` on `127.0.0.1:8501`.

---


admin_console/charts.py
```diff
@@ -0,0 +1,192 @@
...
+    return figure
+
+
+def activity_over_time(events: list[ActivityEvent]) -> go.Figure:
+    grouped: dict[tuple[object, str], int] = defaultdict(int)
+    for event in events:
+        bucket = event.occurred_at.replace(minute=(event.occurred_at.minute // 15) * 15)
+        grouped[(bucket, event.trigger_kind.value)] += 1
+
+    figure = go.Figure()
```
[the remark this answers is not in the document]
haoxuw: activity_over_time added second=0

admin_console/ui.py
```diff
@@ -0,0 +1,285 @@
...
+        page = 1
+    page = min(max(page, 1), total_pages)
+
+    start = (page - 1) * page_size
+    end = min(start + page_size, len(rows))
+    page_rows = rows[start:end]
+    page_ids = row_ids[start:end]
+    selected = selected_id if selected_id in page_ids else page_ids[0]
+    st.query_params[page_query] = str(page)
+
```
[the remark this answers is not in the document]
haoxuw: done

bradhoekstra:
I would like @fkc1e100 or @mastersingh24 to review the concept and approve committing this to the main repo.

haoxuw:
## UI validation evidence

Validated PR head `2459afb` by running the actual Streamlit app locally against `haoxuw-gkedemos / provision-ux-e2e-0727`.

- Connected successfully: **8 passed, 2 warnings, 0 failed**
- All six routes loaded with **0 Streamlit exceptions**
- Screenshots were captured at 1440×1100; the authenticated email is masked
- Images live in evidence-only commit [`ca54f0e`](https://github.com/haoxuw/kube-agents/commit/ca54f0e), outside this PR's code diff

| Connection | Chat |
| --- | --- |
| Connected admin console showing 8 passed, 2 warnings, and 0 failed checks | Chat sessions page backed by live agent history |

| Overview | Activity Explorer |
| --- | --- |
| Overview page showing live activity metrics and charts | Activity Explorer showing live causal flow |

| Task Kanban | Scheduled Cron |
| --- | --- |
| Task Kanban showing live task state | Scheduled Cron showing live execution history and diagnostics |

mastersingh24 (approved):
My review is solely based on the current functionality and not a deep dive code review, but overall the concept is good and we should get folks trying this out.

@haoxuw @bradhoekstra

admin_console/api/streamlit_proxy.py
```diff
@@ -0,0 +1,209 @@
...
+        requested_protocols = [
+            item.strip()
+            for item in websocket.headers.get("sec-websocket-protocol", "").split(",")
+            if item.strip()
+        ]
+        try:
+            async with websockets.connect(
+                upstream_url,
+                additional_headers=headers,
+                origin=upstream_http(),
```
bradhoekstra, line 184: **Security: WebSocket proxy strips the browser Origin, defeating Streamlit's cross-origin check**

The proxy strips the browser's real `Origin` header (`_request_headers`, line 115) and substitutes the upstream's own URL, so Streamlit's cross-origin WebSocket check can never reject anything despite `--server.enableXsrfProtection=true`. FastAPI's `@app.websocket("/{path:path}")` does no origin check of its own.

**Failure scenario:** The portal is running on 127.0.0.1:8501. The operator visits any web page in the same browser; that page opens `new WebSocket("ws://127.0.0.1:8501/_stcore/stream")`. WebSocket handshakes are exempt from CORS preflight, so `--server.enableCORS=true` does not stop it, the attacker's `Origin` is dropped, and `websockets.connect` presents Streamlit its own `http://127.0.0.1:8502`, which always passes. The remote page then speaks Streamlit's protocol as a fully privileged session and can drive every widget on Connections/Chat/Kanban, each of which shells out to `kubectl exec` and `gcloud` against the operator's cluster. This is the mechanism that converts every other loopback-only assumption in this PR into a remotely reachable one.

admin_console/chat/service.py
```diff
@@ -0,0 +1,501 @@
...
+        agent_id: str,
+        input_text: str,
+        profile: str = "default",
+        session_id: str = "",
+        history: Sequence[dict[str, str]] = (),
+        user_email: str = "",
+    ) -> Interaction:
+        now = utc_now()
+        interaction_id = f"ix_{uuid.uuid4().hex}"
+        session_id = session_id or f"portal_{uuid.uuid4().hex}"
```
bradhoekstra, line 79: **Security: "portal-owned sessions only" is enforced in the widget, not the API**

The rule is enforced solely in the Streamlit widget (`pages/chat.py:360` `PORTAL_SESSION` + `:655` `disabled=`). `api/models.py:26` declares `sessionId` with `max_length=256` and no pattern, this line only supplies a default, and `agent_chat.py:482`'s regex accepts any id — so the API will post a turn into any session id the operator can read back from `GET /api/v1/agents/{id}/sessions`.

**Failure scenario:** `POST /api/v1/interactions {"agentId":"platform-agent","profile":"default","sessionId":"gchat-11ab","input":{"text":"..."}}` appends a turn to a live Google Chat or Slack thread that the design doc (`admin-console.md:734-737`) declares read-only. Worse, `agent_chat.py:75-76` gates only the attribution metadata write on the `portal_` prefix, not the post itself, so the injected message lands in the foreign thread with no `admin_portal` attribution at all — indistinguishable from the real participant. Combined with the `streamlit_proxy.py:139` origin bypass, this is reachable from any web page the operator visits.

admin_console/agent_runtime.py
```diff
@@ -0,0 +1,1266 @@
...
+                       ) AS previous_error
+                FROM tasks t
+                WHERE t.session_id = ?
+                ORDER BY t.created_at, t.id
+                LIMIT ?
+                """,
+                (session_id, limit),
+            )
+        ]
+    return {"tasks": rows, "truncated": len(rows) == limit}
```
bradhoekstra: **Regression: off-by-one makes `truncated` a false positive, pinning the settle loop open**

The in-pod tasks read binds the exact `limit` rather than `limit + 1`, so `truncated = len(rows) == limit` reports truncation whenever the session has exactly `limit` tasks and nothing was actually dropped. The messages read has the same defect at line 167. Two other queries in this same file (lines 116-124, 258-261) do it correctly with `limit + 1` / `rows[:limit]` / `>`.

**Failure scenario:** `chat/service.py:340` calls `get_task_updates(limit=100)`. A session with exactly 100 linked kanban rows — all `done` — returns `truncated=True` on every poll. The new settle precondition at `service.py:368-372` begins `not result.truncated`, so `settled` is never True, `quiet` is reset to 0 every iteration, and the loop runs the full 900 s `task_timeout` before `service.py:415-423` writes `TIMED_OUT`. `pages/chat.py:98-99` then renders "Run failed: ..." and discards the stored answer for a run that completed successfully 15 minutes earlier. This is a regression: `HEAD~1`'s settle test had no `truncated` term, so the same session used to settle as `COMPLETED`. The PR's own `test_interaction_api.py:293` pins the never-settle-when-truncated behaviour, so the fix belongs here in `agent_runtime.py`, not in `service.py`.

admin_console/chat/service.py
```diff
@@ -0,0 +1,501 @@
...
+                    {"tasks": [task.to_dict() for task in tasks]},
+                )
+                previous = tasks
+                quiet = 0
+
+            statuses = {task.status.strip().lower() for task in tasks}
+            settled = (
+                not result.truncated
+                and not statuses & ACTIVE_TASK_STATUSES
+                and statuses <= SUCCESSFUL_TASK_STATUSES | FAILED_TASK_STATUSES
```
bradhoekstra: **Regression: closed-world status subset test pins the settle loop open on any unknown status**

`settled` requires `statuses <= SUCCESSFUL_TASK_STATUSES | FAILED_TASK_STATUSES` — a closed-world subset test over nine hardcoded strings, with no escape hatch and no logging. `agent_runtime.py:1049` also normalises a missing status to the literal `"unknown"`, which is in neither set.

**Failure scenario:** `archived` is a real, documented board state (`platform_cron_dispatch.py:172` `KNOWN_STATUSES`; the kanban `VALID_STATUSES` vocabulary is triage/todo/scheduled/ready/running/blocked/review/done/archived), and the tasks query at `agent_runtime.py:210-211` applies no status filter, so archived cards are returned to the settle loop. One archived card in the session means `settled` is permanently False, the loop burns the full 900 s deadline, and the interaction is written `TIMED_OUT` even though every task finished. Same for any future status string. Regression: `HEAD~1` used `if any(task.status in ACTIVE_TASK_STATUSES ...)`, an open-world test that settled correctly. `platform_cron_dispatch.py:161-163` warns explicitly that "the board's vocabulary is not schema-constrained".

admin_console/chat/service.py
```diff
@@ -0,0 +1,501 @@
...
+    def _settle_tasks(self, interaction_id: str) -> None:
+        deadline = time.monotonic() + self._task_timeout
+        quiet = 0
+        previous: tuple[TaskProjection, ...] | None = None
+        while time.monotonic() < deadline:
+            interaction = self._required(interaction_id)
+            if interaction.terminal:
+                return
+            try:
+                result = self._backend_factory().get_task_updates(
```
bradhoekstra, line 347: **One transient read failure discards an already-successful run**

This call sits inside the settle loop with zero retry tolerance; the `except` arm at 345-354 calls `self._fail(...)` and returns. A single transient failure out of the up-to-450 reads issued over a 900 s window is enough.

**Failure scenario:** The root run has already completed and its answer is stored (`service.py:315-321` writes `output=result.output` before `_settle_tasks` is reached). One `kubectl exec` failure during settling — a gateway pod rollover (`agent_runtime.py:941-945` "No running gateway pod was found."), the 25 s exec timeout, or an operator clicking Disconnect (`connection_sidebar.py:394` → `delete_connection` → `backend.py:75-81` raises) — flips the interaction to `FAILED`. `_fail` does not clear `output`, so the API returns the correct answer while `pages/chat.py:90-99` throws it away and prints "Run failed". The Streamlit-side poller for this identical read tolerates three consecutive errors (`chat.py:39` `TASK_ERROR_POLL_LIMIT = 3`); this server-side one tolerates none.

admin_console/telemetry.py
```diff
@@ -0,0 +1,779 @@
...
+        r"(?i)(authorization\s*[:=]\s*(?:bearer\s+)?)[^\s\"',}]+"
+    ),
+    re.compile(
+        r"(?i)((?:api[_-]?key|access[_-]?token|password|secret)"
+        r"\s*[:=]\s*[\"']?)[^\"'\s,}]+"
+    ),
+)
+_DETAIL_LIMIT = 8_000
+MAX_TRACE_PAGES = 10
+_SECRET_KEY = re.compile(
```
bradhoekstra: **Security: redaction regex misses `token`, `refresh_token`, `private_key`, `credential`, `ssh_key`**

The `_SECRET_KEY` key-name regex covers only `authorization|api[_-]?key|access[_-]?token|password|secret`. Verified by running the real `redact_evidence`: `token`, `id_token`, `refresh_token`, `private_key`, `credential`, and `ssh_key` all pass through verbatim, nested or top-level. The value-level pass cannot rescue them, and because dicts are re-serialised to JSON at line 117 it cannot even match its own covered keys in dict form.

**Failure scenario:** Any tool-call evidence carrying `{'token': 'eyJhbGciOi...'}`, `{'refresh_token': '1//0...'}`, `{'private_key': '-----BEGIN PRIVATE KEY-----...'}`, or a kubeconfig blob is rendered verbatim by `pages/activity.py:174` `st.json({... 'evidence': event.details})` and written to Cloud Logging-derived details at `telemetry.py:289`/`:459`. This helper is the sole redaction chokepoint for all 24 call sites in `agent_runtime.py` and all 6 in `agent_chat.py`, so the same gap applies to every Kanban, conversation, and chat projection in the portal. Control cases confirm the function works correctly for `api_key`/`access_token`/`password`, which is exactly what makes the omissions easy to miss.

admin_console/clients/portal_api.py
```diff
@@ -0,0 +1,281 @@
...
+            return
+        base_url = base_url or os.environ.get(
+            "KUBE_AGENTS_PORTAL_API_URL", ""
+        ).strip()
+        if base_url:
+            self._transport = httpx.Client(
+                base_url=f"{base_url.rstrip('/')}/",
+                timeout=httpx.Timeout(30, connect=5),
+            )
+            return
```
bradhoekstra, line 93: **`target` is silently discarded in the httpx branch — two tabs run prompts against the wrong cluster**

In the production httpx branch the `target` argument is discarded: an early `return` fires before it is ever read, and no method sends it as a header or query param. The server re-derives scope per request from one shared file keyed only on the account — `api/app.py:55` reads `KUBE_AGENTS_ADMIN_USER` once, and every handler calls `load_connection(account)` against a single `connection_state_path()`.

**Failure scenario:** Tab A is connected to project-a; the operator opens tab B and connects to project-b, whose `_persist_connection` overwrites the shared file. Tab A still renders `st.caption(f"{target.project_id} · {target.cluster_name} · {target.namespace}")` for project-a (`chat.py:183`) while `list_agents`, `get_messages`, `get_task_updates`, and `POST /api/v1/interactions` all execute against project-b's cluster — the prompt runs on the wrong cluster with the UI asserting the right one. It is self-inflicted and bidirectional: `connection_sidebar.py:276-283` re-verifies every 10 minutes and writes the file back, so two tabs flip the shared scope against each other with no user action. No test covers this branch: the AppTest suite never sets `KUBE_AGENTS_PORTAL_API_URL`, so it always takes the in-process branch where `target` **is** honoured.

admin_console/pages/chat.py
```diff
@@ -0,0 +1,677 @@
...
+            for diagnostic in last_run.diagnostics:
+                st.caption(diagnostic)
+            st.page_link("pages/activity.py", label="Open Activity Explorer")
+
+    prompt = st.chat_input(
+        "Message the agent",
+        disabled=not portal_owned or bool(active_id),
+    )
+    if prompt:
+        prior_history = [*context, *messages]
```
bradhoekstra, line 681: **Unsliced history exceeds the API's `max_length=100`, hard-bricking a session at ~turn 51**

`prior_history = [*context, *messages]` is sent unsliced, but `api/models.py:28` caps history at `max_length=100`, while the buffer is seeded from up to 500 persisted messages (`chat.py:366`/`:384`, limit 500) and grows by 2 per turn. Reproduced against the real app: 100 items → 202, 101 → 422.

**Failure scenario:** At roughly turn 51 every send 422s permanently. The user-facing message is useless because FastAPI's 422 `detail` is a **list**, and `portal_api.py:111-113` (`detail.get(...) if isinstance(detail, dict) else {}`) collapses it to the generic "Portal API request failed.", so `chat.py:676` renders "Run failed: Portal API request failed." with no hint about history size. Each failed attempt appends 2 more entries, and the Refresh button (`chat.py:413`) only rebuilds the buffer from the ≥101-message persisted transcript, so the session is unrecoverable. No test covers the cap.

admin_console/telemetry.py
```diff
@@ -0,0 +1,779 @@
...
+        or "platform-agent",
+        platform=_first(payload, "platform"),
+        user_id=_first(payload, "user_id", "user.id"),
+        session_id=session_id,
+        task_id=task_id,
+        tool_name=tool_name,
+        cluster=_first(payload, "cluster") or _first(labels, "cluster_name"),
+        namespace=_first(payload, "namespace") or _first(labels, "namespace_name"),
+        resource=_first(payload, "resource"),
+        duration_ms=int(payload.get("duration_ms") or 0),
```
bradhoekstra: **Unguarded `int()` on a log field takes down the whole Activity and Overview pages**

`duration_ms=int(payload.get("duration_ms") or 0)` is the only unguarded coercion on the logging row — every sibling field (`_parse_time`, `_first`, `_status`, `_logging_payload`) already has a fallback. A string like `"12.5"`, `"1.2s"`, or `"123ms"`, or a dict/list value, raises `ValueError`/`TypeError`.

**Failure scenario:** One malformed log line takes down two whole pages, not one row. The exception propagates from here → the bare list comprehension at `:605-610` → `logging_future.result()` at `:749` → `get_snapshot` → `list_activity`, and there is no try/except anywhere on that path (`pages/activity.py:21` and `pages/overview.py:23` are bare module-level statements, `admin_console/app.py` has zero try/except, and `.streamlit/config.toml` does not set `client.showErrorDetails`), so Streamlit renders a raw traceback. The in-repo emitter is a plausible source: `tool_call_audit/audit.py:47` types `duration_ms` as `Optional[float]` and serialises with `default=str`, which turns a `Decimal` or `timedelta` into the string `"12.5"` or `"0:00:00.012500"`. The log query at `:542-553` is scoped only by `resource.type` and namespace, so any container in `kubeagents-system` can poison the snapshot. Per-source error handling exists (`:582-596`) but nothing at row granularity.

admin_console/api/streamlit_proxy.py
```diff
@@ -0,0 +1,209 @@
...
+            f"--server.port={streamlit_port()}",
+            "--server.headless=true",
+            "--server.enableXsrfProtection=true",
+            "--server.enableCORS=true",
+            "--browser.gatherUsageStats=false",
+        ],
+        cwd=PACKAGE_ROOT.parent,
+        env=environment,
+    )
+    client = httpx.AsyncClient(base_url=upstream_http(), follow_redirects=False)
```
bradhoekstra: **No timeout and no exception handling — a dead upstream returns 500, not the intended 503**

The proxy client is built with no timeout (httpx default is 5 s, verified on the pinned 0.28.1) and `proxy_http` (181-209) has no try/except; there are no exception handlers or middleware anywhere in `admin_console/api/`. `process.poll()` is also never re-checked after startup, so `streamlit_ready` stays True for the life of the app — it is only reset in the lifespan `finally`.

**Failure scenario:** Reproduced against the real module: if Streamlit dies mid-session, every subsequent request raises `httpx.ConnectError` and the operator gets `HTTP 500 Internal Server Error` instantly — exactly where the 503 "Streamlit is starting." guard at line 182 was meant to fire. A >5 s time-to-first-byte on a cold or CPU-contended upstream produces the same 500 after 5.0 s. Note httpx's read timeout is per socket read, not total elapsed, so large bundles that drip continuously over loopback are fine; the trigger is a stalled or dead upstream, not a merely slow one. No test file imports `streamlit_proxy` at all.

admin_console/api/app.py
```diff
@@ -0,0 +1,240 @@
...
+        wait_seconds: float = Query(default=30.0, alias="waitSeconds", ge=0, le=60),
+    ) -> StreamingResponse:
+        interaction = service.get(interaction_id)
+        if interaction is None:
+            raise HTTPException(
+                status_code=404,
+                detail=_error("interaction_not_found", "Interaction was not found."),
+            )
+
+        def stream() -> Iterator[str]:
```
bradhoekstra: **Sync SSE route parks anyio worker threads — 45 concurrent waiters take down the whole portal**

`interaction_events` is a sync `def` returning a `StreamingResponse` over a sync generator, so Starlette wraps it in `iterate_in_threadpool` and the blocking `store.wait_for_change(..., timeout=wait_seconds)` (default 30 s, max 60) parks an anyio worker thread. All 11 routes in this file are sync `def` and share one 40-token limiter, under `--workers=1`.

**Failure scenario:** Reproduced against the real app under uvicorn: 45 concurrent `GET /api/v1/interactions/{id}/events?waitSeconds=60` starve the pool and an unrelated `GET /healthz` fails with `TimeoutError` after 25 s — the entire portal, UI proxy included, stops responding. Made worse by `store.py:446-454`, where the wait predicate re-enters the public API and opens three fresh SQLite connections per waiter per `notify_all`, and the notify is store-wide rather than per-interaction (measured 1 + 3N connections per `append_event`; 50 waiters on unrelated interactions cost 1.21 s CPU over 3.4 s wall). That predicate also calls `self.get()` twice without binding, so a concurrent `prune()` raises `AttributeError` out of `Condition.wait_for`.

admin_console/chat/store.py
```diff
@@ -0,0 +1,506 @@
...
+    return root / "kube-agents" / "admin-portal-interactions.db"
+
+
+def _encode_interaction(interaction: Interaction) -> str:
+    payload = {
+        "interaction_id": interaction.interaction_id,
+        "agent_id": interaction.agent_id,
+        "profile": interaction.profile,
+        "session_id": interaction.session_id,
+        "input_text": interaction.input_text,
```
bradhoekstra: **Prompts are persisted unredacted, contradicting the PR's own design doc**

`_encode_interaction` persists `interaction.input_text` verbatim. The entire `chat/` package calls the redaction helper zero times (grep for `redact|scrub|mask|sanitiz|secret` across `chat/` returns no matches), even though `agent_chat.py:445-462` redacts the agent's output, error, approval, and event strings on the way in.

**Failure scenario:** Round-tripped through the real `SQLiteInteractionStore`: a prompt containing a kubeconfig token, `api_key=AIzaSyD-...`, and `password=hunter2` is stored byte-for-byte in `~/.local/state/kube-agents/admin-portal-interactions.db` and returned raw by `chat/models.py:114` through both `POST` and `GET /api/v1/interactions`. `api_key=` and `password=` are forms `redact_evidence` would have caught, so this is a missed chokepoint rather than an unredactable value. It directly contradicts the design doc added by this PR: `admin-console.md:454-457` — "Redaction happens before persistence and again before response serialization. Raw secrets, bearer tokens ... are never stored in the interaction database."

admin_console/pages/chat.py
```diff
@@ -0,0 +1,677 @@
...
+        task_snapshots[state_key] = refreshed
+
+        if has_active_tasks(refreshed):
+            task_poll_grace[state_key] = 0
+            if changed:
+                st.rerun(scope="app")
+            return
+
+        grace = int(task_poll_grace.get(state_key, 0))
+        if grace > 1 and not changed:
```
bradhoekstra: **Dead branch plus status drift: the page stops polling and never resumes**

`if grace > 1 and not changed:` is unreachable — every write to `task_poll_grace` sets it to 0 (lines 516, 528) or decrements it (523), and the decrement is the only thing that could make it non-zero, so it can never bootstrap. `TASK_EMPTY_POLL_LIMIT` (line 38) is defined and never read anywhere in the repo. Compounding this, `pages/chat.py:36` `ACTIVE_TASK_STATUSES = {todo, ready, running}` is a strict subset of `chat/service.py:23`, missing `triage`, `scheduled`, and `review`.

**Failure scenario:** The first poll that sees no active task falls straight through to lines 527-530: polling is disabled, "Agent work is up to date" is printed, and `st.rerun` fires. After that rerun the fragment's `run_every` is None and the only True-setter (line 460) is inside `if state_key not in task_snapshots:`, which line 513 just populated — so auto-refresh is off until a manual Refresh or session switch, and a task delegated seconds later is never picked up. The status drift makes it fire early: the moment the last task moves `running` → `review`, the page declares "up to date" and stops polling while `service.py:370` still counts that task as active and keeps the interaction in `WAITING_FOR_TASKS`. A page opened on a session whose task sits in `review` or `scheduled` never starts polling at all.

admin_console/clients/portal_api.py
```diff
@@ -0,0 +1,281 @@
...
+        if self._in_process:
+            deadline = time.monotonic() + 1
+            while view.status in {"queued", "running", "waiting_for_tasks"}:
+                if time.monotonic() >= deadline:
+                    break
+                time.sleep(0.002)
+                view = self.get_interaction(view.interaction_id)
+        return view
+
+    def get_interaction(self, interaction_id: str) -> InteractionView:
```
bradhoekstra, line 269: **The cancel path is fully implemented and completely unwired — no way to stop a runaway run**

`PortalApiClient` has no `cancel` method (`dir()` confirms 9 public/private members, none of them cancel), and the fully implemented `POST /api/v1/interactions/{id}/cancel` → `service.cancel` → `backend.stop` chain has no caller anywhere in the repo. The SSE endpoint at `api/app.py:183-210` plus its sequence/resume machinery is likewise dead — the UI polls `get_interaction` every 2 s instead, and its only exercise is one test with `waitSeconds=0`.

**Failure scenario:** `chat.py:655` disables the chat input while `active_id` is set, and `active_id` is cleared only by `finish_interaction` on a terminal result. There is no Stop control in the UI (the only interaction buttons are Approve/Deny), and the Refresh button pops six session-state dicts but deliberately not `portal_active_interactions`, so it does not help. With production defaults (root run timeout 600 s bounded by kubectl at 620 s, then `task_timeout` 900 s) an operator who sends a runaway prompt sits with a greyed-out chat box for up to ~25 minutes with no way to stop the agent, while a working cancel path sits unwired one function call away. A browser reload re-enables the input but neither stops the run nor is offered anywhere in the UI.

bradhoekstra (changes requested):
## Code review: 15 findings (2 security, 8 correctness, 5 quality)

Automated deep review of this PR, run at maximum effort: 10 independent finder angles over the diff, each candidate then independently verified by a skeptic prompted to refute it. Two candidates were refuted and dropped. Fourteen findings are posted inline; the fifteenth is below because it has no anchor line in this diff.

Requesting changes rather than commenting because findings 1 and 2 are remotely exploitable and findings 3 and 4 are regressions against `HEAD~1` that silently discard successful runs.

### The two that matter most

**`api/streamlit_proxy.py:139` — the WebSocket proxy defeats Streamlit's cross-origin check.** `_request_headers` (line 115) strips the browser's real `Origin` and `websockets.connect` substitutes the upstream's own URL, so the check can never reject anything despite `--server.enableXsrfProtection=true`. WebSocket handshakes are exempt from CORS preflight, so `--server.enableCORS=true` does not help either, and FastAPI's `@app.websocket("/{path:path}")` does no origin check of its own. Any page the operator visits in the same browser can open `ws://127.0.0.1:8501/_stcore/stream` and drive every widget on Connections/Chat/Kanban — each of which shells out to `kubectl exec` and `gcloud` against the operator's cluster. This is the mechanism that converts every other loopback-only assumption in this PR into a remotely reachable one.

**`chat/service.py:77` — session ownership is enforced only in the widget.** The API will post a turn into any session id the operator can read back from `GET /api/v1/agents/{id}/sessions`, including live Google Chat and Slack threads that the design doc declares read-only. `agent_chat.py:75-76` gates only the *attribution metadata* on the `portal_` prefix, not the post itself, so an injected message lands in a foreign thread indistinguishable from the real participant.

### Finding 15 — no CI covers any of this (`Makefile:98`)

Posted here because `Makefile` is not in this diff, which is precisely the problem: `git diff main...HEAD -- .github/ Makefile requirements-test.txt` is empty, and nothing under `.github/` references `admin_console` or `bench`.

Of the 14,403 lines this PR adds, the only thing `make test-python` exercises is the 6-line change to `bootstrap_onboarding/test_plugin.py`. `PYTHON_TEST_DIRS` does not glob `admin_console/tests/` (9 modules, 3,369 lines) or `bench/tests/test_cuj.py` (302 lines). Run by hand the suites are healthy — 82 + 6 tests passing in ~19 s — so real coverage exists, it just never gates a merge, and every regression listed above ships green.

Three separate blockers, all of which need fixing together:
1. The globs at `Makefile:98` don't match either directory.
2. `requirements-test.txt` has no `streamlit`, `plotly`, `httpx`, `pydantic`, or `websockets`.
3. `Makefile:148` uses `(cd $$dir && python3 -m unittest discover ...)`, which raises `ModuleNotFoundError` on all 9 modules — they use absolute `from admin_console...` imports and need a `-t` top-level dir.

The comment at `Makefile:86-89` is self-indicting about this exact failure mode.

### Also flagged, cut by the 15-finding cap

Not verified to the same depth, listed so they aren't lost:

- `agent_runtime.py:110` — cutoff and `LIMIT` applied outside the `candidates` CTE: three correlated subqueries per session before the time window, so Conversations permanently 25 s-timeouts on the busiest agents.
- `chat/store.py:456` — `recover_incomplete()` marks every non-terminal row `FAILED`, so a second portal instance kills the first's in-flight runs on the shared per-user DB.
- `agent_chat.py:438` — the non-streaming `_invoke` branch always raises: the pod emits a checkpoint line before the terminal line, so `json.loads` sees `Extra data`.
- `agent_chat.py:113` — unguarded SQLite write plus a permanent `journal_mode=WAL` conversion of the agent's own `session_kv.db`.
- `telemetry.py:748` — `_load_trace` gets an empty `session_users`, and the compensating `replace()` unconditionally rewrites attribution to INHERITED/HUMAN.
- `tests/test_admin_portal_ui.py:520` — `patch.dict` without `clear=True` lets an inherited `KUBE_AGENTS_PORTAL_API_URL` make eight tests drive a live portal and real `kubectl exec`.
- `pages/autonomous.py:169` — measured 2.8 s at 10 cron jobs, 5.8 s at 20, of uncached per-rerun CPU.
- `pages/kanban.py:86` — 5 `kubectl` subprocesses per rerun, 10 per row click due to a double `st.rerun`.
- `bench/kube_agents_bench/cuj.py:259` — approval `continue` skips both the deadline check and the sleep; `:535` — `expandvars` over raw JSON text corrupts on any value containing a quote; `:180` — dropped `ProxyHandler({})` relative to `harness.py:311-314`.
- `api/streamlit_proxy.py:192` — `headers.items()` collapses duplicate `Set-Cookie`. Latent, since shipped Streamlit emits at most one.

admin_console/api/streamlit_proxy.py
```diff
@@ -0,0 +1,209 @@
...
+        ]
+        try:
+            async with websockets.connect(
+                upstream_url,
+                additional_headers=headers,
+                origin=upstream_http(),
+                subprotocols=requested_protocols or None,
+                max_size=None,
+            ) as upstream:
+                await websocket.accept(subprotocol=upstream.subprotocol)
```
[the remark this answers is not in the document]
haoxuw: it's currently unnecessary, because the code only runs on the admin's local machine. But addressed the change anyway

admin_console/pages/chat.py
```diff
@@ -0,0 +1,677 @@
...
+        messages.append({"role": "user", "content": prompt})
+        try:
+            with st.spinner("Starting agent interaction…"):
+                result = runtime_provider.start_interaction(
+                    selected_agent,
+                    prompt=prompt,
+                    session_id=session_id,
+                    history=prior_history,
+                    profile=profile,
+                )
```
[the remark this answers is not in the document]
haoxuw: valid

admin_console/pages/chat.py
haoxuw: Fixed in 5f9ac08. PortalApiClient.start_interaction now sends only the last MAX_HISTORY_MESSAGES entries; StartInteractionRequest imports that same limit, and _payload formats FastAPI validation-detail lists. Regression tests cover a 125-message history being reduced to the last 100 and an actual 422 history error including both `history` and `100`.

admin_console/agent_runtime.py
haoxuw: Fixed in 2fc475a. The embedded messages and tasks queries now fetch limit plus one, return only the requested limit, and set truncated only when an extra row exists. Behavioral SQLite tests cover exactly 100 rows and 101 rows for both queries.

admin_console/chat/service.py
haoxuw: Fixed in 2fc475a. StartInteractionRequest and ChatService now reject every supplied session ID that is not portal-owned. The API regression test proves gchat-11ab returns 422 and a direct service call with a Slack session raises ValueError.

admin_console/chat/service.py
haoxuw: Fixed in 2fc475a. Task settlement now tolerates two consecutive read errors and fails only on the third, resetting the counter after a successful read. Tests prove one transient error preserves the successful root output and completes, while three failures stop boundedly.

admin_console/clients/portal_api.py
haoxuw: Fixed in 2fc475a. The HTTP PortalApiClient now retains its DeploymentTarget and sends all target coordinates on every request. FastAPI compares them with the current verified connection and returns 409 to a stale tab. Tests cover both emitted headers and stale-target rejection.

admin_console/telemetry.py
haoxuw: Fixed in 2fc475a. duration_ms now uses a bounded coercion helper that accepts numeric strings and falls back to zero for malformed shapes instead of raising. Tests exercise both 12.5 and a dictionary value through normalize_logging_row.

admin_console/api/app.py
haoxuw: Fixed in 2fc475a and refined in d544bda. The SSE route and generator are async, and long-poll waits now use cooperative asyncio polling rather than parking either the AnyIO route pool or another thread pool. SQLite notifications are also per interaction and the terminal predicate reads the interaction once. The deterministic event-stream lifecycle test continues to cover ordered delivery and resume behavior; the environment-sensitive 45-client timing test was removed.

admin_console/api/streamlit_proxy.py
haoxuw: Fixed in 2fc475a. The private HTTP client now has explicit connect/read timeouts, the proxy rechecks the Streamlit process, and httpx failures map to 503. Tests prove both a ConnectError and an exited process return 503 without reaching a dead upstream.

admin_console/telemetry.py
haoxuw: Fixed in 2fc475a. Evidence redaction now covers token, id_token, refresh_token, private_key, credential, and ssh_key in addition to the existing names. The regression test passes every key through the real redact_evidence function and verifies no value remains.

admin_console/chat/store.py
haoxuw: Fixed in 2fc475a. ChatService keeps the raw prompt only in memory for the backend while persisting a redacted projection; the SQLite encoder and API serializer redact again. A real SQLite round-trip test proves the backend receives the original while the database and response omit both test secrets.

admin_console/chat/service.py
haoxuw: Fixed in 2fc475a. Settlement is now open-world: only known active states keep it open, while known failures still fail the interaction. Regression tests prove archived, an unknown future terminal state, and an empty status can settle.

admin_console/clients/portal_api.py
haoxuw: Fixed in 2fc475a. PortalApiClient now calls the cancel endpoint and the Chat page displays a Stop action for every nonterminal interaction, updating local state from the cancellation result. Client endpoint coverage complements the existing queued and running cancellation service tests.

admin_console/pages/chat.py
haoxuw: Fixed in 2fc475a. The page now imports the server active-status vocabulary, initializes and consumes the three-poll grace period, resets it while work is active, and starts task polling with every new interaction. Streamlit tests cover review status and delayed polling shutdown.

bradhoekstra (approved):

bradhoekstra (approved):

PR #551, opened 2026-08-05
fix(fleet-audit): align the five audit SOPs and guard the contract

adamparco:
# Pull Request

## Why This Change

`cc11068` shipped five cron-driven fleet audits — five SOPs in
`agents/platform/governance/`, five prompts in `agents/platform/cron/jobs.json`, one shared
harness in `audit_report.py`. The starting suspicion was that the five SOPs share a lot of text
and should be factored into one file.

**They don't.** They share an *outline*, not text: measured across the five, 98% of substantive
lines appear in exactly one file, and 0 of 798 distinct normalised lines appear in all five. A
template would template ~13 lines and carry ~785 lines of per-stream fragments, and its source
would have to live under `agents/platform/governance/`, which ships wholesale into every profile
home (the scaffolder never prunes).

What *is* shared is a set of rules the harness enforces, and those had diverged — not by drift
(`cc11068` was still the last commit to touch any SOP) but **at birth**. Seven rules were stated
in some SOPs and missing from others, one term was defined three incompatible ways, and nothing
checked any of it. So this PR consolidates **the guarantee, not the artifact**: repair the
divergences by hand, then add the guards that make the next one fail in CI instead of at 06:20
in production.

Two live bugs get fixed on the way. `obtainability_audit_sop.md` told the worker to write a
kubeconfig to a `wra_…` path that exists nowhere in the source, and two published site pages
quoted a cron prompt whose line numbers were ~90 lines stale.

## What Changed

Three commits, deliberately separate — only the second changes what the audits report.

**Files:**

- `agents/platform/governance/*_sop.md` — all five
- `agents/platform/cron/jobs.json`
- `agents/platform/skills/fleet-audit/scripts/audit_report.py` (one stale comment)
- `hack/check-docs-terminology.sh`
- `docs/README.md`, `docs/site/.../concepts/governance-sops.md`,
  `.../concepts/autonomous-watchdogs.md`, `.../reference/cron-jobs.md`

**Added/Changed/Fixed:**

1. **`fix(fleet-audit): align the five audit SOPs on the rules the harness enforces`** — six
   prose repairs, each copying the best existing wording rather than inventing new phrasing, and
   adapted to the stream's own nouns (the drift audit says "facets", not "checks"):

   | Rule                                                              | Source of truth                                             | Was missing from                                          |
   | ----------------------------------------------------------------- | ----------------------------------------------------------- | ---------------------------------------------------------- |
   | `checks_run[].command` names an inspection binary and is ≥8 characters | `INSPECTION_BINARIES` / `MIN_CHECK_COMMAND_CHARS`           | four SOPs                                                  |
   | Never print raw credentials — as a **Red Line**, not body prose   | prose only; harness redaction is a backstop                 | cost, drift                                                |
   | `grep` is kind-blind; open the hit before trusting it             | prose only, but it picks `remediation.kind`                 | security-patch, cost                                       |
   | `/remediate all` is accepted, not just `/remediate <id>`          | `finish` expands `all` against the run's manifest findings  | cost, drift                                                |
   | No unstable identity — name the durable object, never the Pod     | `derive_finding_id`                                         | security-patch, cost, drift                                |
   | Per-cluster kubeconfig isolation                                  | `_thread_kubeconfig_path` in `platform_mcp_server.py`       | obtainability (wrong path), cost (no isolation at all)     |

   All five cron prompts were then re-measured in the same commit, because the anti-skim prompts
   state the SOP's line count and check-section span and those move when the file grows.

2. **`fix(fleet-audit): reconcile the system-namespace definition across three SOPs`** —
   compliance's regex becomes canonical, rendered into obtainability and cost in each file's own
   notation, plus `gmp-public` (only obtainability had it; a real GMP namespace).

3. **`test(fleet-audit): guard the SOP, cron, and rendering-cap contract`** — the guards. In
   `test_audit_report.py`: a set-equality assertion between `AUDITS` and the `fleet-audit` cron
   jobs, replacing three skip-guards that let a stream with no watchdog ship green; four
   hardenings of the cron-geography test; and a `SHARED_RULES` roll-call over all five SOPs. In
   `hack/check-docs-terminology.sh`, following the `FINDING_ID_RE` block already there: the seven
   rendering caps are read out of `audit_report.py` and matched against every doc line quoting
   one, and any doc line quoting a cron prompt must be a verbatim substring of `jobs.json`.

## Why This Matters

- **The kubeconfig path was a live bug.** `obtainability_audit_sop.md` named
  `/opt/data/.kubeconfigs/wra_<project>_<cluster>_<location>.yaml`. The `wra_` prefix appears
  nowhere in the source; the real shape is `kubeconfig_…` under `$HERMES_HOME`. The cost SOP had
  a bare `get-credentials` with no isolation at all, so two clusters could bleed into each other
  through a shared current-context — and a cost audit that reads cluster B while it believes it
  is reading cluster A files findings against the wrong cluster.
- **The two site pages were teaching the wrong thing.** They quote the compliance prompt as the
  worked example of an anti-skim prompt, and said "all 348 lines of it … lines 56-270" against
  an actual 406 and 102-314. That example is the one readers copy when adding a watchdog.
- **These are birth defects, so a guard is the whole point.** Nothing had drifted; the files were
  born unequal. Repairing them by hand without a conformance test just resets the clock.

## Context

- Follows up `cc11068` (#517).
- **Deliberately out of scope**, and worth stating so nobody re-derives it: generating the SOPs
  from a template (the 98% figure above); teaching `AuditSpec` to own schedule/cadence and
  generating `jobs.json`; a `new_audit_stream` scaffolding CLI; the 23 non-audit skills and six
  non-audit cron jobs. The original ask also covered "make it easier to add a cron job" — that
  half is untouched here by scoping agreement, and the harness's own registry
  (`AUDITS` + `jobs.json` + one SOP) is now at least set-equality-checked, so a half-added stream
  fails loudly.
- **Follow-up left in place, not fixed:** `compliance_audit_sop.md:182` keeps a narrower inline
  `jq` system-namespace set inside check 2.4. It answers a different question — which
  ServiceAccount namespaces are system-owned for a `cluster-admin` binding — from the `$SYS`
  regex, which decides which namespaces the audit skips. Reconciling the two needs a security
  judgement about `cluster-admin` scope that does not belong in a docs-alignment PR.
- **Residual gap:** the cron prompt's check count is now pinned to the roster, but nothing pins a
  count written in narrative prose. `governance-sops.md` said "Ten checks" for an eleven-check
  audit; it is corrected by hand here and stays unguarded.

## Testing

- `test_audit_report.py` — **449 tests, green** on Python 3.9 and 3.14.
- **Every new guard proven red before it was proven green**, since a guard that has never failed
  is untested:
  - `SHARED_RULES` on the pre-repair SOPs → **12 failures**, one per rule genuinely missing,
    across all four rows.
  - Each rendering cap perturbed at the source (`MAX_EXCERPT_LINES`, `BODY_BUDGET`,
    `AUTO_PROMOTION_CAP`, `MIN_NA_REASON_CHARS`) → each guard and its floor fired.
  - The cron-prompt guard reverted against the pre-fix page → red on that page.
- `make docs-check` — all four checks pass. `make docs-generate` leaves the tree clean.
- `review-docs-drift` run against the branch diff: no doc outside the two fixed pages states a
  SOP line count, no doc lists the system namespaces (so commit 2's behaviour change has no doc
  surface), and both stated check counts now match the roster. Its findings on `docs/README.md`
  — a missing identifier-source row and an understated `docs-check-terminology` summary — are
  fixed in commit 3.
- **Not run locally:** `npx prettier --check` and the site build. No JavaScript runtime is
  installed in this environment. Prettier's default is `proseWrap: preserve` and the repo has no
  `.prettierrc`, so the long single-line paragraphs will not rewrap; the one structural change (a
  fenced block added inside a numbered list in the cost SOP) matches the file's existing
  three-space continuation indent, and the new `docs/README.md` table row is padded to the
  existing column widths by hand. CI is the real verdict on both.
- Commit 2 changes what the cost audit reports: it stops flagging `gatekeeper-system`,
  `asm-system`, `configconnector-operator-system`, and `composer-system`, and starts flagging
  `anthos-`-prefixed namespaces other than `anthos-identity-service` (its old `anthos-*` glob was
  broader than anything the other two SOPs suppress and would hide real findings in a user
  namespace). `security-patch-orchestrator` and `fleet-consistency-drift` are untouched — both
  are control-plane audits with no namespace concept.

---

Prose-and-tests only; no harness behaviour changes. The one behavioural effect is commit 2's
namespace reconciliation, isolated in its own commit so it can be reverted alone. First run after
merge will re-derive finding ids as usual and the ledger deltas absorb it.

## Review Response (commit 4)

Review mutation-tested the new guards and found four that stay green on a document that has lost
the thing they check. All four are fixed, and each fix is proven green unmutated and red on the
exact mutation that slipped past the previous version.

| Guard | Was | Mutation that passed before, fails now |
| ----- | --- | -------------------------------------- |
| credentials Red Line | anchored on `credential` alone, which compliance's read-only bullet already satisfies with `gcloud container clusters get-credentials` two bullets above | replacing compliance's `- **Never print raw credentials.**` line with an equal-length placeholder → 1 clean `FAIL` |
| `MIN_CHECK_COMMAND_CHARS` | absent from `check-docs-terminology.sh` entirely — the one new constant with no source-derived guard | `8` → `10` → mismatch error naming all seven doc lines, plus the copies floor |
| excerpt-clip `JOIN` | second alternative spelled `and \*\* `, which matches nothing; the slash form still matched six files so the copies floor could not see it | `40` → `41` in `SKILL.md:605` → excerpt-clip error |
| system-namespace set | unguarded — the only behavioural change in the PR | adding a namespace to `$SYS` alone, dropping one from `SYSTEM_NS`, or widening an S1 glob → a named `FAIL` per SOP |

The `MIN_CHECK_COMMAND_CHARS` probe carries both spellings the docs use — five SOPs and the skill
write "anything under eight characters", `fleet-audit-issue-ledger.md:943` writes "shorter than
eight characters" — so all seven lines are held, not six.

`test_one_system_namespace_set_spelled_three_ways` re-derives the set from all three notations and
compares them as sets: the `SYS='^(…)$'` alternation, the `SYSTEM_NS` backtick list and S1's
backtick list, with globs normalised to shell form. Each parse yields the same 17 members today.
The prose parser stops at the first token whose preceding gap is not a list connector, so the cost
SOP's following sentence — "Note it is `anthos-identity-service` and not `anthos-*`" — cannot leak
`anthos-*` into the set. **This is the hunk worth reading carefully:** it is a hand-rolled parser
over prose, and its failure mode if the prose is reworded is a confusing error rather than a
silent pass. Compliance check 2.4's narrower inline `jq` set is still deliberately excluded, for
the reason given above.

Four smaller items from the same review:

- `cron_jobs()` asserts its key set before returning. Dropping a cron job used to give three
  `KeyError` tracebacks around one real failure; it now gives four clean `FAIL`s.
- The body-budget probe's `targets` arm requires a four-or-more digit figure, so a future
  "the sweep targets 3 clusters" is not read as a stale copy of the cap. Same six lines matched
  before and after.
- The cron-prompt guard gets a comment stating the constraint it imposes: `all N lines of it` is
  treated as a claim to be quoting a prompt, so paraphrase fails the check even when nothing has
  drifted.
- The cost SOP's Step 2 commands carry the `KUBECONFIG=$KC` its Step 1.3 promises, and
  compliance's kubeconfig export gets the `${HERMES_HOME:-/opt/data}` fallback the other two SOPs
  use. Both edits are in-place, leaving the files at 261 and 406 lines so the anti-skim cron
  prompts still cite real geography (`git diff --numstat` → 3/3 and 1/1).

Verified on the fix commit: **449 tests green**, `make docs-check` all four checks pass,
`make chart-check` passes, `npx prettier --check` clean on both changed SOPs.


toshiowang:
# Review

Three well-separated commits, and the thesis — "consolidate the guarantee, not the artifact" — is the right call. The analysis behind not templatizing is convincing.

I checked out head `d0dd4f5` and ran everything the description said it couldn't, then mutation-tested each new guard. Three findings survived; the patches below are verified green unmutated and **red** on the exact mutation that slipped past the current version.

## What I verified

- **448 tests green** (Python 3.14); **`make docs-check`** all four checks green.
- **`npx prettier --check` clean** on all seven changed Markdown/JSON files — this resolves the "not run locally" caveat, including the fenced block added inside the numbered list in the cost SOP.
- **Kubeconfig fix correct**: `${HERMES_HOME:-/opt/data}/.kubeconfigs/kubeconfig_<slug>.yaml` matches `_thread_kubeconfig_path` (`agents/platform/scripts/platform_mcp_server.py:234-259`) exactly.
- **Namespace sets genuinely reconcile** across compliance `$SYS`, cost `SYSTEM_NS`, obtainability `S1` — diffed by hand.
- **The new cap guards aren't vacuous**: they match 7/2/1/7/3/11/5/2 doc lines respectively.

---

## 1. The credentials Red Line guard is vacuous for `compliance_audit_sop.md`

`agents/platform/skills/fleet-audit/scripts/test_audit_report.py:84`

`SHARED_RULES` anchors that rule on `r"credential"` scoped to the Red Lines section. But compliance's Red Lines already contains, two bullets above, in an *unrelated* rule:

> `gcloud container clusters get-credentials` is the sole exception and touches only a local kubeconfig.

**Mutation:** replace compliance's `- **Never print raw credentials.**` line with an unrelated placeholder of identical length — so the geography test can't catch it by line count — and **all 448 tests stay green**. The guard cannot see the deletion of the rule it exists to protect, on the one SOP where the redaction backstop matters most.

The comment above `SHARED_RULES` states the principle correctly ("anchor on the distinctive words of the rule itself and nothing else"); the pattern just doesn't follow it. All five phrasings put both words on one line:

```diff
--- a/agents/platform/skills/fleet-audit/scripts/test_audit_report.py
+++ b/agents/platform/skills/fleet-audit/scripts/test_audit_report.py
@@ -81,7 +81,10 @@ SHARED_RULES = (
     (
         "credentials never reach an excerpt",
         "red-lines",
-        r"credential",
+        # Both words, one line. A bare `credential` is satisfied by the
+        # read-only rule's `gcloud container clusters get-credentials`, which
+        # sits in compliance's Red Lines two bullets above this one.
+        r"credential[^\n]*excerpt|excerpt[^\n]*credential",
         "the harness redacts high-confidence shapes as a backstop, not as the "
         "primary control; the primary control is this line",
     ),
```

With this applied the same mutation fails cleanly:

```
FAIL: test_every_sop_states_the_rules_that_hold_on_every_stream
      (audit='compliance-audit', rule='credentials never reach an excerpt')
```

## 2. `MIN_CHECK_COMMAND_CHARS` is the one new rule with no source-derived guard

`hack/check-docs-terminology.sh:184` · `test_audit_report.py:79`

The eight-character floor is the flagship repair — added to four SOPs (`compliance:51`, `cost:49`, `drift:37`, `obtainability:46`, `security-patch:39`). But "eight" is a hardcoded literal in `SHARED_RULES`, and the constant is absent from the `for CONSTANT in …` loop, so nothing ties the prose to `audit_report.py:757`.

**Mutation:** `MIN_CHECK_COMMAND_CHARS = 8` → `10`. **Terminology check passed, 448 tests passed**, all five SOPs still telling workers "eight characters." That is precisely the drift class this PR exists to prevent, on the constant its own new prose introduces.

```diff
--- a/hack/check-docs-terminology.sh
+++ b/hack/check-docs-terminology.sh
@@ -181,7 +181,8 @@ cap_guard() {
 }
 
 for CONSTANT in MAX_EXCERPT_LINES MAX_EXCERPT_CHARS MAX_COMMAND_CHARS \
-  MAX_BODY_CHARS BODY_BUDGET MAX_SCOPE_ROWS AUTO_PROMOTION_CAP MIN_NA_REASON_CHARS; do
+  MAX_BODY_CHARS BODY_BUDGET MAX_SCOPE_ROWS AUTO_PROMOTION_CAP MIN_NA_REASON_CHARS \
+  MIN_CHECK_COMMAND_CHARS; do
   if ! spellings "$CONSTANT" > /dev/null; then
     echo "ERROR: could not read ${CONSTANT} from ${AUDIT_SCRIPT}." >&2
     exit 1
@@ -232,6 +233,14 @@ cap_guard \
   "reason under $(spellings MIN_NA_REASON_CHARS) characters" \
   "Documented not-applicable reason floor does not match MIN_NA_REASON_CHARS in ${AUDIT_SCRIPT}."
 
+# The check-command floor, which all five SOPs state as "as is anything under
+# eight characters". Distinct probe from the one above so the two floors cannot
+# satisfy each other.
+cap_guard \
+  'anything under [a-z0-9]+ characters' \
+  "anything under $(spellings MIN_CHECK_COMMAND_CHARS) characters" \
+  "Documented check-command floor does not match MIN_CHECK_COMMAND_CHARS in ${AUDIT_SCRIPT}."
+
 # --- fleet-audit cron prompts ---------------------------------------------
```

The probe is deliberately distinct from `'reason under [a-z0-9]+ characters'` so the two floors cannot satisfy each other, and `spellings` already yields `(8|eight)`. With this applied, the 8→10 mutation fires both the mismatch error and the "no document states this cap any more" floor.

## 3. Commit 2's namespace reconciliation is the only repair left unguarded

`grep -c` for `SYSTEM_NS`, `gmp-public`, or `composer-system` across both `test_audit_report.py` and `check-docs-terminology.sh` returns **0**. The PR's own argument is that hand-repair without a conformance test "just resets the clock," and the residual-gap section calls out narrative check counts — but not this, the one behavioural change in the PR.

No patch proposed: three notations (a regex at `compliance_audit_sop.md:41`, two prose backtick-lists at `cost:110` and `obtainability:163`) is awkward but not intractable — parse the `SYS='^(…)$'` alternation, parse the two lists plus their glob tails, assert set equality. If that is judged too brittle to be worth it, it belongs in the residual-gap list rather than unmentioned.

---

## Minor

**KeyError instead of a clean failure** — `test_audit_report.py:1460`. Three callers now index `jobs[audit_id]` directly. The commit's rationale ("assert the equality once, here, so the rest can index the catalogue directly") assumes the set-equality test runs first, but unittest orders alphabetically and `test_cron_prompts…` / `test_check_rosters…` sort ahead of `test_every_stream…`. Dropping a cron job yields **3 KeyError tracebacks around 1 clean failure**. Verified fix — the same mutation then gives 4 clean `FAIL`s:

```diff
--- a/agents/platform/skills/fleet-audit/scripts/test_audit_report.py
+++ b/agents/platform/skills/fleet-audit/scripts/test_audit_report.py
@@ -1457,8 +1460,18 @@ class TestAuditCatalogue(unittest.TestCase):
         )
         if not jobs_file.is_file():  # not shipped alongside the skill at runtime
             self.skipTest(f"{jobs_file} not present")
-        jobs = json.loads(jobs_file.read_text(encoding="utf-8"))["jobs"]
-        return {job["id"]: job for job in jobs}
+        jobs = {job["id"]: job for job in json.loads(
+            jobs_file.read_text(encoding="utf-8")
+        )["jobs"]}
+        # Callers index this by audit id. The set equality has its own test,
+        # but unittest orders alphabetically and two callers sort ahead of it,
+        # so a stream that lost its watchdog would otherwise surface as three
+        # KeyError tracebacks around one real failure.
+        missing = sorted(set(audit_report.AUDITS) - set(jobs))
+        self.assertFalse(
+            missing, f"no cron watchdog for {', '.join(missing)}"
+        )
+        return jobs
```

**The description mislabels a table row.** The ≥8-char / inspection-binary rule is enforced on `checks_run[].command` (`validate_check_command`, `audit_report.py:760-808`), not `evidence.command` — which at `audit_report.py:1354` only requires a non-empty string. The SOP prose is placed correctly in the `checks_run` paragraphs; only the table is wrong. Worth fixing, since that table is what a reviewer audits against.

**The cost SOP's new promise isn't reflected in its own examples.** `fleet_wide_cost_analysis_sop.md:67` now promises "Every `kubectl` in Step 2 runs with that cluster's `KUBECONFIG=$KC`", but §2's collect bullet is still bare. Obtainability's equivalent (`obtainability_audit_sop.md:75`) shows it inline:

```diff
--- a/agents/platform/governance/fleet_wide_cost_analysis_sop.md
+++ b/agents/platform/governance/fleet_wide_cost_analysis_sop.md
@@ -79 +79 @@
-- Per cluster: `kubectl get nodes,pods,pvc,pv,svc,jobs,cronjobs,pdb,ns,resourcequota -A -o json`, plus `gcloud container node-pools list --cluster=<c> --location=<l> --format=json`.
+- Per cluster: `KUBECONFIG=$KC kubectl get nodes,pods,pvc,pv,svc,jobs,cronjobs,pdb,ns,resourcequota -A -o json`, plus `gcloud container node-pools list --cluster=<c> --location=<l> --format=json`.
```

**Three spellings of one rule.** `compliance_audit_sop.md:35` uses `export KUBECONFIG="$HERMES_HOME/.kubeconfigs/…"` — no fallback, so it yields `/.kubeconfigs/…` if `HERMES_HOME` is unset — while cost and obtainability now use `export KC="${HERMES_HOME:-/opt/data}/.kubeconfigs/…"`. Pre-existing in compliance, but this is the PR that reconciled the other two.

**Two probes broad enough to false-positive later.** `'(body at |targets \*{0,2})[0-9][0-9,]*'` will fire on any future "targets 3 clusters"; the cron-prompt guard flags any line matching `all [0-9]+ lines of it` that is not a full verbatim quote, so narrative prose paraphrasing the prompt becomes a CI failure. Both are correct today; a one-line comment noting the constraint would save the next author the debugging.

## Style and convention

Good. Comments explain *why* in the register the rest of this codebase uses — the `_outside_fences` docstring on why a `### ` inside a fence shifts every derived span "silently, in the direction that makes a stale citation look correct" is the standout. Conventional Commits, three commits cleanly separated with the behavioural one isolated for revert, and `cap_guard`'s "a guard that passes because every copy disappeared is not a passing guard" floor mirrors the existing `ID_COPIES` block. No unrelated formatting churn. The strongest new assertion is the one requiring the cited line span to actually contain the full roster — that closes the hole where a prompt could cite a real section that happens to be the preamble.

## Verdict

LGTM after 1 and 2 — both patches above are verified green-then-red. 3 needs a decision (guard it, or list it as a known gap), not necessarily code.

---

Review performed with the help of Claude.

toshiowang (approved):

PR #552, opened 2026-08-05
fix(operator): stop regenerating and rewriting tracked files during deploy

### LINKED ISSUE #526: make deploy regenerates tracked files and shells out to npx prettier
## What happens

Running the install from a clean checkout leaves the tree dirty:

```
 config/crd/bases/..._agentplugins.yaml    |   554 +-
 config/crd/bases/..._platformagents.yaml  | 15675 +++++-----
 config/manager/kustomization.yaml         |     8 +-
 config/rbac/role.yaml                     |   274 +-
 config/webhook/manifests.yaml             |    82 +-
 5 files changed, 8090 insertions(+), 8503 deletions(-)
```

`deploy` depends on `manifests`, which runs `controller-gen` and then:

```make
npx prettier --write "config/crd/bases/*.yaml" "config/rbac/*.yaml" "config/webhook/*.yaml"
```

Behind a private npm mirror that fails with `E401`.  The Makefile ignores it (`make: [manifests] Error 1 (ignored)`), so the regenerated YAML is left unformatted, and that accounts for most of those 8,000 lines.  `kustomize edit set image` rewrites `config/manager/kustomization.yaml` on top.

## Why it matters

A deploy shouldn't need npm, or reachability to npmjs.  It arguably shouldn't regenerate committed manifests either.  If the generated files are in the repo, deploying should apply them rather than rebuild them.

As it stands, anyone installing from a clone ends up with an 8,000 line diff they didn't write, and has to work out that discarding it is safe.

## Suggested fix

Drop `manifests` from `deploy`'s prerequisites, or make the `prettier` call conditional on npm being reachable.  Regenerating and reformatting tracked files is a development task, not a deployment one.

bnaylor:
## Why This Change

Installing from a clean clone leaves the tree dirty:

```
 config/crd/bases/..._agentplugins.yaml    |   554 +-
 config/crd/bases/..._platformagents.yaml  | 15675 +++++++++-----
 config/manager/kustomization.yaml         |     8 +-
 config/rbac/role.yaml                     |   274 +-
 config/webhook/manifests.yaml             |    82 +-
 5 files changed, 8090 insertions(+), 8503 deletions(-)
```

Two separate causes:

- `deploy` (and `install`, `uninstall`) depend on `manifests`, which runs `controller-gen` and then prettier over `config/`.  Where prettier fails to run (see the comment thread for what that actually was on this machine), the Makefile ignores the error, and the regenerated YAML is left unformatted -- which is most of those 8,000 lines.
- `deploy` runs `kustomize edit set image` against the tracked `config/manager/kustomization.yaml`.

Underneath the churn there is a correctness point.  Regenerating during a deploy means applying **whatever the local tree produces**, which is not necessarily what is committed.  A deploy should ship what is in git.

Fixes #526.

## What Changed

**Files:**

- `Makefile` (root -- prettier install guidance)
- `AGENTS.md` (same)
- `k8s-operator/Makefile`
- `k8s-operator/README.md`
- `.github/workflows/k8s-operator-test.yml`
- `.github/workflows/prettier.yml`
- `.github/workflows/agent-startup-test.yml`
- `docs/site/src/content/docs/operator/development.md`
- `tests/test_operator_makefile.py` (new)

**Added/Changed/Fixed:**

- `install`, `uninstall` and `deploy` no longer depend on `manifests`.  They apply the committed manifests as-is.
- `deploy` copies `config/` to a temp directory and pins the image there, so `config/manager/kustomization.yaml` is untouched.
- `manifests` (and `prettier-check`/`prettier-write`) use the root Makefile's `PRETTIER` helper: an installed binary when there is one, `npx` as the fallback.  The prettier step is still non-fatal, but a failure now says what it costs (on stderr) rather than scrolling past as `Error 1 (ignored)`.
- Operator CI regenerates and fails on a diff.  The gate installs a pinned `prettier@3.9.6` first -- it asserts byte-equality against formatted output, so an unpinned version would fail every operator PR whenever a prettier release changes YAML formatting.  The same pin now lives in `prettier.yml` so the two jobs cannot disagree.
- The gate stages `config/` with `git add --intent-to-add` before diffing, so a newly generated CRD for a new API type (an untracked file, invisible to plain `git diff`) fails it too.
- The gate also checks committed deepcopy: `make generate` plus a diff of `k8s-operator/api` with `-I'Copyright [0-9]{4}\.'`, which ignores the licence-header year that `generate` stamps from the current date.  Nothing else covers this -- `make test` regenerates before compiling.
- `tests/test_operator_makefile.py` pins the deploy contract with `make -n`: the deploy targets invoke no `controller-gen`, and `deploy` does not `cd` into the tracked `config/manager`.
- `operator/development.md` records that the deploy targets deliberately do not regenerate, and that `make manifests` is now the developer's job.  The README note in `k8s-operator/` states the local fact and links there.

`build`, `run` and `test` keep their `manifests` dependency -- regenerating is appropriate there.

## Why This Matters

Beyond the tidiness: anyone installing from a clone currently ends up with thousands of lines of modified files they did not write, and has to work out that discarding them is safe.  Worse, the manifests actually applied to the cluster may differ from the ones in the repository, which makes a deploy unreproducible.

## Context

Found during a first install; the churn also obscured a real one-line change I was making at the time.  Wider write-up: https://gist.github.com/bnaylor/b8e7b7f9dda75c6f159815b36aec3d49

The review thread reshaped the CI gate considerably: the prettier pin, the untracked-file staging, and the deepcopy check all came out of it.

## Testing

Against the live cluster from a clean tree:

- `make install IMG=…` -- CRDs applied server-side.
- `make deploy IMG=…` -- full manifest set applied, controller rolled out on `ghcr.io/gke-labs/kube-agents/k8s-operator:latest`.
- `git status` afterwards showed **only** the intentional edits.  `config/manager/kustomization.yaml` verified unchanged.

On the CI gate:

- `make manifests` produces zero diff against `config/` locally, so the committed manifests are current and the gate passes for the right reason.
- The `-I` year immunity reproduced directly: `YEAR=2031 make generate` fails a plain diff of `api/` and passes with `-I'Copyright [0-9]{4}\.'`.
- The untracked-file case reproduced both ways: a fake new CRD passes the gate without `--intent-to-add` and fails it with.
- The new tests bite: against main's Makefile the `make -n deploy` recipe hits `cd config/manager` once and `controller-gen` twice.  Full `tests/` suite passes.

`make docs-check` passes.

---


bnaylor:
Correcting the scale of the problem in the description, now that I have a working prettier locally.

With npm reachable, `make manifests generate` produces **zero diff** — the committed manifests are current, and controller-gen plus prettier reproduces them exactly. So the ~8,000 lines in the description are entirely an artifact of prettier not running; they are what you get behind a registry you cannot reach, not what a typical contributor sees.

On a machine where npm works, `make deploy` dirties **one** file, `config/manager/kustomization.yaml`, by 8 lines.

That is a much smaller problem than the description implies, and I would rather say so than let the diff stat carry an argument it cannot support. What I do not think it changes:

- `kustomize edit set image` writing to a tracked file during a deploy is still wrong, and is the part that dirties the tree even in the good case.
- Applying regenerated rather than committed manifests is still the correctness point. Today they happen to be identical; nothing guaranteed that, which is what the new CI check is for.
- The bad case is not exotic. Any restricted or mirrored registry reproduces it, and the failure is silent because the Makefile ignores it.

The zero-diff result is also the first real evidence that the CI guard in this PR passes for the right reason rather than by luck — I had only been able to infer that from CI going green.

bnaylor:
Both findings fixed in 637e116. Verified each before changing anything.

**The year bomb is real.** `YEAR ?= $(shell date +%Y)` is stamped into the deepcopy licence header by `generate`, and the committed header reads `Copyright 2026.`. Reproduced it directly:

```
$ YEAR=2027 make -C k8s-operator generate
$ git diff --stat -- k8s-operator/api
 k8s-operator/api/v1alpha1/zz_generated.deepcopy.go | 2 +-
$ git diff --quiet -- k8s-operator/config && echo "config/ still clean"
config/ still clean
```

So the gate would have failed every operator PR from 1 January on output that is not stale. Took the first of the two suggested fixes: the step now runs `manifests` alone and diffs `k8s-operator/config` only. That is also the better-scoped check on its own terms — `config/` is what the deploy targets apply, whereas deepcopy is compiled in and its staleness surfaces as a build or test failure anyway.

**The README was wrong.** `k8s-operator/README.md` still said `make install` uses controller-gen to generate the CRDs from Go structs, which stopped being true when `install` lost its `manifests` dependency. Replaced with what it now does, plus the consequence a developer actually needs: edits to the API types do not reach the cluster until `make manifests` is run.

Worth recording how I missed it, since the finding is precisely right that `make docs-check` cannot catch prose drift. I did run a docs-drift grep, and it did cover `k8s-operator/README.md` — but I piped it through `head -8` and the hits fell past the cutoff. Re-running it untruncated surfaces this file immediately. Truncating the search is a good way to make a docs-drift review look clean when it is not.

Also checked the remaining `make install` / `make deploy` mentions untruncated (INSTALL.md 282, 352, 405; README 215, 260; kustomize.md; security-and-iam.md) — none of them assert regeneration, so nothing else needed changing.

The gate passes locally now that npm is reachable here: `make manifests` produces zero diff against config/.

bnaylor:
One more correction, on the cause rather than the scale.

I attributed the prettier failure to an unreachable or private npm mirror. That was wrong. On this machine the failure was an expired corp SSO ticket — the npm error is preceded by `gpkg: ERROR: failed to get OAuth2 token ... Try running gcert`, and it recurs roughly daily. After re-authenticating, `npx prettier` works and `make manifests` produces zero diff. So it was a transient auth failure that I generalised into an environmental constraint on the strength of one invocation.

That does not change the change itself, but it does change how strongly one of its arguments should be read:

- "A deploy should not need npm or reachability to npmjs" still stands on its own — a deploy step that shells out to a package manager can fail for many reasons, and this Makefile swallows the error either way.
- "Behind a registry you cannot reach" is a weaker claim than I made it. Restricted mirrors do exist, but I have not demonstrated one here; what I demonstrated was an expired ticket.

The parts of this PR that do not depend on npm at all — `kustomize edit` writing to a tracked file during deploy, and deploying regenerated rather than committed manifests — are unaffected.

mplakhtiy:
## Findings

### 1. `manifests` hard-codes `npx prettier`, so the new gate misreports an npx failure as stale manifests — High

`k8s-operator/Makefile:40-41`

```make
	@npx prettier --write "config/crd/bases/*.yaml" "config/rbac/*.yaml" "config/webhook/*.yaml" \
		|| echo "WARNING: prettier did not run (no npm registry reachable?). …"
```

The `|| echo` swallows the exit status, which is deliberate and fine on its own. What is new is
that `.github/workflows/k8s-operator-test.yml:53` now asserts on the resulting tree. When prettier
does not run, `manifests` leaves `config/` unformatted, `git diff --exit-code` fails, and the step
prints:

> `::error::Committed manifests are out of date. Run 'make -C k8s-operator manifests' and commit the result.`

That is the wrong diagnosis, and following it is actively harmful: a contributor whose npx is also
broken regenerates the same unformatted YAML and commits 8,000 lines of it, which then fails the
`prettier` job instead.

Reproduced on this machine — `npx` hits a 403 from an authenticated Artifact Registry npm mirror,
`make manifests` exits 0 with the warning, and the tree gains +8321 / −8734 across four files.
Running `prettier --write` over the identical globs with the installed 3.9.6 binary brings the diff
back to zero, which confirms the manifests were never stale.

`main` already solves this. The root `Makefile:60-71` carries the pattern and the reasoning:

```make
PRETTIER := $(shell command -v prettier 2>/dev/null || echo npx prettier)
```

with a comment describing precisely this failure ("on a machine whose registry is an authenticated
mirror these targets failed with an auth error even though prettier was installed and on PATH").
`AGENTS.md:104` repeats it: "Prefer the installed binary over `npx prettier`". Since this PR is
already rewriting that line and is built on the argument that a swallowed prettier failure is a
problem, adopting the existing helper here — and in `k8s-operator/Makefile:354-358`, which has the
same two `npx prettier` invocations — is the fix. Send the warning to stderr while you are there;
it currently goes to stdout, interleaved with prettier's own file list.

**Verdict: CONFIRMED** — reproduced locally, both the failure and the fix.

### 2. The gate cannot see a newly generated CRD — Medium

`.github/workflows/k8s-operator-test.yml:53`

```bash
if ! git diff --exit-code -- k8s-operator/config; then
```

`git diff` reports changes to tracked files only. `controller-gen` writes one file per CRD into
`config/crd/bases/`, so a contributor who adds a kubebuilder root type and forgets `make manifests`
produces an **untracked** `config/crd/bases/<group>_<plural>.yaml`, and the gate passes. Verified:
creating `kubeagents.x-k8s.io_newtypes.yaml` and running the exact command from the workflow exits 0.

The consequence is the one the gate exists to prevent. `make install` and `make deploy` now apply
only what is committed, so the new CRD never reaches the cluster and the operator's reconcile for
that type fails at runtime against a missing resource — with CI green. The mirror case leaks too: a
removed API type leaves its committed CRD behind, since `controller-gen` does not delete files, and
the gate sees nothing.

Either stage first, or check status rather than diff:

```bash
git add --intent-to-add -A -- k8s-operator/config
git diff --exit-code -- k8s-operator/config
```

**Verdict: CONFIRMED.**

### 3. The gate's pass/fail depends on an unpinned prettier version — Medium

`.github/workflows/k8s-operator-test.yml:52`

`make manifests` runs bare `npx prettier`, which resolves the newest matching release at run time.
Nothing in the repository pins a version: there is no `package.json` and no `.prettierrc`, and
`.github/workflows/prettier.yml:28` installs a bare `npm install prettier`. So the gate asserts
byte-equality between freshly formatted output and files formatted by whichever prettier version the
last contributor happened to resolve.

The first prettier release that changes YAML output fails this step on every operator PR, on files
the contributor did not touch, until someone commits a reformat. That is the same shape as the
`YEAR` bomb fixed in `637e116` — a gate whose input drifts on a calendar rather than on the code —
and it wants the same remedy of pinning the input. The pre-existing `prettier` job has the same
exposure, but it is bounded to files the PR changed; this one asserts across all of `config/`.

Pin the version where `manifests` invokes it (`npx prettier@3`, or a `package.json` devDependency
that both workflows install from).

**Verdict: CONFIRMED** — the absence of a pin is verifiable; the trigger is a future release, which
is exactly how the year bomb read before it was fixed.

### 4. Deepcopy staleness is now guarded by nothing — Low

`.github/workflows/k8s-operator-test.yml:44-48`

Scoping the gate to `manifests` and `config/` was the right call for the year bomb, but it leaves
`k8s-operator/api/v1alpha1/zz_generated.deepcopy.go` with no check anywhere. `make test`
(`k8s-operator/Makefile:56`) runs `generate` _before_ compiling, so CI always builds against freshly
regenerated deepcopy and never compiles the committed file. The Docker `build` job compiles
committed sources, but that only catches staleness severe enough to break the build — a brand-new
type missing `DeepCopyObject`.

The case it misses: a new pointer, slice or map field added to a spec without `make generate`. The
stale `DeepCopyInto` compiles cleanly and shallow-copies the field, so a controller that mutates it
aliases the shared informer cache. Nothing in CI notices.

`git diff` can ignore the drifting line directly, which keeps the year bomb shut:

```bash
make -C k8s-operator generate
git diff --exit-code -I'Copyright [0-9]{4}\.' -- k8s-operator/api
```

I confirmed `-I` is accepted by the git on the runner image.

**Verdict: CONFIRMED as a coverage gap** — the mechanism by which `make test` hides it is in
`Makefile:56`; whether it is worth closing in this PR is the author's call.

### 5. The deploy targets' new contract has no test — Low

`k8s-operator/Makefile:94-111`

The PR's two behavioural claims — the deploy targets do not regenerate, and `deploy` does not write
to the tracked `config/manager/kustomization.yaml` — are enforced only by a comment. Both revert
silently: re-adding `manifests` to `deploy:` restores deploy-time regeneration, and someone
"simplifying" the temp-directory recipe back to `cd config/manager && kustomize edit` restores the
dirty tree. Neither fails any check, and the churn only shows up the next time a contributor
notices their working tree is dirty after an install — which is how #526 was found in the first
place.

This is cheap to pin down without a cluster; `make -n deploy IMG=x` prints the recipe, and the
repository already has a Python suite for this kind of assertion (`tests/`). Asserting that the
printed recipe does not `cd` into `config/manager`, and that `make -n install` invokes no
`controller-gen`, covers both claims.

**Verdict: CONFIRMED** — no test in the diff or the repository exercises either claim.

### 6. "CI fails if the committed manifests are stale" now lives in two docs — Low

`k8s-operator/README.md:176` and `docs/site/src/content/docs/operator/development.md:31`

Both files now assert that the deploy targets do not regenerate and that CI catches staleness.
`AGENTS.md` Documentation Guidelines: "Every fact has one home. Duplicating documentation across
files is how it goes stale" and "**Link rather than summarise** when another page already owns the
topic." The README note needed rewriting — it was asserting the opposite of what the code does —
but the second half of it restates what `development.md` already owns, and the two sentences are
already worded differently ("manifests" vs "output"), which is how this kind of pair drifts.

Keeping the README to the local fact (`make install` applies the committed CRDs; run `make
manifests` first) and linking out for the CI behaviour costs nothing.

**Verdict: CONFIRMED** against the quoted rule.


---

bnaylor:
All six land in 1dc37f3.  The branch is also rebased onto main, which is where the root `PRETTIER` helper lives -- the branch predated it.

**1 - npx in the gate's path:** Adopted the root helper at all three call sites in `k8s-operator/Makefile`, warning to stderr.  In CI the gate installs the pinned binary first, so the npx fallback only matters off-runner.

**2 - untracked CRDs:** Took the `--intent-to-add` staging as suggested.  Reproduced both ways: a fake new CRD in `config/crd/bases/` passes the old gate and fails the new one.

**3 - unpinned prettier:** Pinned `3.9.6` -- the version the committed files are formatted with -- in both workflows, with cross-referencing comments.  Exact pin rather than `@3`, since prettier does change formatting in minors.  I stopped short of a root `package.json`; introducing a root npm manifest felt like a bigger decision than this PR should carry.  Happy to be overruled on that.

**4 - deepcopy gap:** Closed with your `-I` suggestion.  The year immunity reproduces here: `YEAR=2031 make generate` fails a plain diff of `api/` and passes with `-I'Copyright [0-9]{4}\.'`.

**5 - deploy contract test:** Added `tests/test_operator_makefile.py` -- stdlib unittest, so the existing discover run in `agent-startup-test.yml` picks it up.  Four assertions over `make -n` output.  Confirmed they bite: against main's Makefile the deploy recipe hits `cd config/manager` once and `controller-gen` twice.

**6 - duplicated CI fact:** The README note now states the local fact and links the development guide for the rest.

`make manifests` produces zero diff against `config/` here, and the full `tests/` suite passes with the new file.  PR body updated to match.

bnaylor:
Right -- the filter list was the one place I didn't look when wiring the test in.  Fixed in ccdab5b: `k8s-operator/Makefile` is now in `agent-startup-test.yml`'s paths filter, with a comment saying why it is there, so a Makefile-only change runs the suite.  Went with the filter entry over running the suite from the operator job, which would have added a Python setup to a Go workflow.

bnaylor:
Fixed in fc61474.  The root Makefile comment and AGENTS.md now point at the pin in `prettier.yml` rather than restating a version, so the literal lives only in the two workflows that execute it.  The gate's error message also names version skew as a cause, since "run make manifests and commit" is a loop when the skew is the problem.

mplakhtiy (approved):

PR #554, opened 2026-08-05
fix(scripts): let provisioning run unattended

### LINKED ISSUE #525: Provisioning can't run unattended: CI=true is undocumented and stage 06 always prompts
## `CI=true` is load-bearing and undocumented

`init_var` only takes defaults when `DRY_RUN=1` or `is_ci_pipeline` is true.  Otherwise it does a bare `read`:

```bash
echo -ne "  ${prompt_msg} [${default_val}]: "
read -r input_val
```

Under `set -e` in a non-TTY that `read` hits EOF, returns non-zero and can abort the run.  `CI=true` is what makes the pipeline usable unattended, and it doesn't appear in INSTALL.md.

`init_var_image_tag` then hard-errors in non-interactive mode unless `IMAGE_TAG` is exported.  Also undocumented.

## Stage 06 prompts unconditionally

`provision_06_slack.sh` prompts even when the values are already set.  With `SLACK_BOT_TOKEN` present it asks add/replace/keep:

```bash
echo -ne "  Do you want to (a)dd additional tokens, (r)eplace existing tokens, or (k)eep ... [a/r/K]: "
read -r action_bot
```

Those reads aren't guarded by `DRY_RUN` or `CI`, so the stage can't be automated at all.

It makes no cloud changes.  Every path through it is a `save_var` call.  So it's a pure prompt barrier sitting in the middle of an otherwise scriptable pipeline.  I skipped the stage and wrote the variables into `vars.sh` directly, which worked fine.

## Suggested fix

Document `CI=true` and `IMAGE_TAG` for unattended runs.  Guard stage 06's prompts the way `init_var` does, and skip them when the values are already present.

bnaylor:
## Why This Change

The pipeline cannot be driven without a human at the keyboard, for two separate reasons.

### `init_var` prompts even when nothing can answer

```bash
if [ "${DRY_RUN:-0}" -eq 1 ] || is_ci_pipeline; then
  final_val="$default_val"
else
  echo -ne "  ${prompt_msg} [${default_val}]: "
  read -r input_val
```

With no terminal, `read` hits EOF and returns non-zero, and under `set -e` that aborts the run.  `CI=true` is the workaround, and it is not documented anywhere — you have to read `common.sh` to find it.

`common.sh` already has the right predicate, and `init_var` simply was not using it:

```bash
is_non_interactive() {
  [ ! -t 0 ] || [ "${NO_CONFIRM:-0}" -eq 1 ] || [ "${DRY_RUN:-0}" -eq 1 ] || is_ci_pipeline
}
```

### Stage 06 prompts unconditionally

`provision_06_slack.sh` guarded only on `DRY_RUN`, so with `SLACK_ENABLED=true` it asks add/replace/keep **even when both tokens are already configured**, then asks again for the app token.  There is no way to get through it unattended.

It makes no cloud calls at all — every path is a `save_var` — so it was purely a prompt barrier sitting in the middle of an otherwise scriptable pipeline.

Fixes #525.

## What Changed

**Files:**

- `k8s-operator/scripts/provision_06_slack.sh`
- `INSTALL.md`

**Added/Changed/Fixed:**

- `init_var` uses `is_non_interactive()`, so a missing TTY is enough.  `DRY_RUN` and `CI` still work — they are subsumed by the same predicate — and `NO_CONFIRM=1` now works too.
- Stage 06's two prompt blocks follow the same rule.  Unattended keeps the existing configuration, which is exactly what the interactive prompt defaults to (`[a/r/K]`).
- INSTALL.md documents unattended runs, and that `IMAGE_TAG` is deliberately the one value with no default, so an unattended run cannot silently pick a tag nobody chose.

## Why This Matters

This is the difference between the pipeline being scriptable and not.  Any CI job, cron redeploy, or agent driving the install currently either hangs or dies at the first prompt, and the escape hatch is an undocumented environment variable.

It also removes an inconsistency rather than adding a special case: `is_non_interactive()` already existed and already meant this; one caller had drifted.

## Context

Hit while installing with an agent driving.  I worked around it by exporting `CI=true` and skipping stage 06 entirely, writing its variables into `vars.sh` by hand.  Wider write-up: https://gist.github.com/bnaylor/b8e7b7f9dda75c6f159815b36aec3d49

## Testing

With **no TTY and no `CI=true`**, which previously aborted:

- `./provision_06_slack.sh < /dev/null` exits 0, prints the instructions, and leaves both tokens intact in `vars.sh`.
- `init_var` on an unset variable returns its default and survives `set -euo pipefail` rather than aborting on EOF.

Under a **real pty** (`script -q /dev/null …`), to confirm the human path is untouched:

- stage 06 still shows `(a)dd additional tokens … keep current configuration`.
- `init_var` still shows its prompt.

Also: `bash -n` clean on both scripts, `make docs-check` passes, `make docs-generate` produces no diff — the change is below each script's comment banner, so the `provisioning-steps` generated region is unaffected.

---

No change to interactive behaviour. The only difference is what happens when there is nobody there to type.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -174,6 +174,7 @@ make gcp-provision
 > Because the provisioning scripts persist configuration state in `scripts/vars.sh`, running the script again will reuse the same options selected on the first run. If you want to change configuration variables, manually edit `scripts/vars.sh` or perform a teardown first.
 
 - **Dry-run check**: To preview actions without modifying cloud infrastructure:
+
   ```bash
   make gcp-provision ARGS="--dry-run"
   ```
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -181,17 +181,22 @@ make gcp-provision
 
 - **Unattended runs** (CI, cron, or an agent driving the install): the provisioning stages detect
   that there is no terminal and take the saved or default value for every prompt instead of
-  waiting for input. Force the same behaviour on a terminal with `NO_CONFIRM=1` (or `CI=true`).
-  Teardown still asks for confirmation before destroying anything unless you pass `--no-confirm`.
-  `IMAGE_TAG` is the one value with no default — export it explicitly, so an unattended run
-  cannot silently pick a tag you did not choose:
+  waiting for input. Nothing needs to be set — a piped or CI run works as-is. To get the same
+  behaviour from a terminal, close stdin:
 
   ```bash
-  IMAGE_TAG=latest NO_CONFIRM=1 make gcp-provision
+  IMAGE_TAG=latest make gcp-provision < /dev/null
   ```
 
-  Anything the run must not prompt for — API keys, Slack tokens — should be written to
-  `scripts/vars.sh` (or exported) beforehand. Values already present are reused as-is.
+  `IMAGE_TAG` is the one value with no default — export it explicitly, so an unattended run
+  cannot silently pick a tag you did not choose. Anything else the run must not prompt for —
+  API keys, Slack tokens — should be written to `scripts/vars.sh` (or exported) beforehand;
+  values already present are reused as-is.
+
+  > [!WARNING]
+  > `NO_CONFIRM=1` and `CI=true` also suppress prompts, but they are read by `confirm_action`
+  > as a standing "yes" — a teardown run in the same environment will delete the cluster and
+  > its service accounts without asking. Prefer closing stdin for provisioning.
 
 #### Security & CMEK Encryption
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -179,9 +179,10 @@ make gcp-provision
   make gcp-provision ARGS="--dry-run"
   ```
 
-- **Unattended runs** (CI, cron, or an agent driving the install): the pipeline detects that
-  there is no terminal and takes the saved or default value for every prompt instead of
+- **Unattended runs** (CI, cron, or an agent driving the install): the provisioning stages detect
+  that there is no terminal and take the saved or default value for every prompt instead of
   waiting for input. Force the same behaviour on a terminal with `NO_CONFIRM=1` (or `CI=true`).
+  Teardown still asks for confirmation before destroying anything unless you pass `--no-confirm`.
   `IMAGE_TAG` is the one value with no default — export it explicitly, so an unattended run
   cannot silently pick a tag you did not choose:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -178,6 +178,19 @@ make gcp-provision
   make gcp-provision ARGS="--dry-run"
   ```
 
+- **Unattended runs** (CI, cron, or an agent driving the install): the pipeline detects that
+  there is no terminal and takes the saved or default value for every prompt instead of
+  waiting for input. Force the same behaviour on a terminal with `NO_CONFIRM=1` (or `CI=true`).
+  `IMAGE_TAG` is the one value with no default — export it explicitly, so an unattended run
+  cannot silently pick a tag you did not choose:
+
+  ```bash
+  IMAGE_TAG=latest NO_CONFIRM=1 make gcp-provision
+  ```
+
+  Anything the run must not prompt for — API keys, Slack tokens — should be written to
+  `scripts/vars.sh` (or exported) beforehand. Values already present are reused as-is.
+
 #### Security & CMEK Encryption
 
 The automated installer includes local state hardening and Cloud KMS (CMEK) etcd database encryption:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -194,9 +194,10 @@ make gcp-provision
   values already present are reused as-is.
 
   > [!WARNING]
-  > `NO_CONFIRM=1` and `CI=true` also suppress prompts, but they are read by `confirm_action`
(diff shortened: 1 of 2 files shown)


bnaylor:
Both fixed in 44d7c33. Verified each first; both were right, and the first one matters more than its severity label suggests.

**Stage 07 — and the regression is real.** Three `DRY_RUN || is_ci_pipeline` guards left unconverted, each followed by `read -s -r INPUT_KEY`. Confirmed the abort by replaying the old guard with stdin closed under `set -e`: it prints "about to read", then never reaches the next line.

The important part is the one the finding makes about blast radius. Before this branch a no-TTY run died at the first `init_var` in stage 01, before any cloud call. With only the partial fix, stages 01-06 succeed and the silent exit lands after the cluster, node pool, operator and IAM bindings exist. So the half-done change made the failure later and more expensive, which is worse than not making it. All three now use `is_non_interactive`, falling back to the same `placeholder` the CI path already used.

Verified end to end with `GEMINI_API_KEY` removed from `vars.sh` and no TTY: exit 0, warns that the key is a placeholder, reaches "Secrets Configured".

**`NO_CONFIRM` from the environment.** Confirmed — `NO_CONFIRM=0` is assigned unconditionally at source time, before argument parsing:

```
$ NO_CONFIRM=1 bash -c 'source common.sh; echo $NO_CONFIRM'
0
```

So the example I put in INSTALL.md was a no-op. Both `DRY_RUN` and `NO_CONFIRM` now seed from the environment, and it reads 1 with `is_non_interactive` true.

**`provision_05`.** Same shape on the project-number fallback. I did not give it a default: making up a number would configure the Pub/Sub IAM bindings against the wrong project and fail further away. The unattended path leaves it empty and lets the existing required-value check at line 55 report it properly.

**On the INSTALL.md claim.** Rather than only narrowing the promise, I audited every `read` in `provision_*.sh` — all of them now sit inside an `is_non_interactive`-guarded branch (the two `IFS=… read -a <<<` are here-strings and never touch stdin), so the claim is now true rather than scoped around. I did narrow it in one respect: it says *provisioning stages*, and notes teardown still confirms before destroying.

Two things I deliberately left alone: `confirm_action` still requires an explicit `NO_CONFIRM`/`CI` rather than treating a missing TTY as consent, because auto-approving a destructive teardown on the basis of no terminal is not an improvement. And `teardown_02_gvisor_nodepool.sh` has the same guard shape but already honours `NO_CONFIRM` and `CI`, so only the bare no-TTY case is affected; it is teardown rather than provisioning, so I left it out of this PR.

bnaylor:
Correct, High is the right severity, and the hazard was mine — fixed in af77c90.

Verified it before changing anything:

```
$ NO_CONFIRM=1 bash -c 'source common.sh; confirm_action "DESTROY EVERYTHING" ...'
DISARMED - returned 0 without prompting
```

So the sentence I added was not merely optimistic, it was false because of the change it sat next to. And the sequence is a realistic one: export `NO_CONFIRM=1` once for the unattended provisioning the docs recommend, run `make gcp-teardown` later in the same shell, and the cluster, the GSAs, the KMS key versions and `platform-agent-secrets` go without a prompt — with `teardown_01` removing `vars.sh`, so the record of what was destroyed goes too.

**Reverted the environment seeding**, with a comment on the assignment explaining why it has to stay that way, so the next person to notice that `NO_CONFIRM=1 make gcp-provision` looks like it should work does not re-add it.

Of the two options offered, I took neither exactly. Gating `confirm_action` on a flag-only variable would have worked, but it widens this PR into changing destructive-confirmation semantics; and simply documenting that `NO_CONFIRM=1` disarms teardown leaves a foot-gun the docs actively recommend stepping on.

The better answer is that unattended provisioning never needed the variable. `is_non_interactive` already detects a missing TTY, so closing stdin gets the prompt suppression without touching the confirmation:

```bash
IMAGE_TAG=latest make gcp-provision < /dev/null
```

Confirmed the redirection reaches the stages through `make` — stage 06 runs to completion with zero interactive prompts and exit 0 — that an exported `NO_CONFIRM=1` no longer reaches `confirm_action`, and that `--no-confirm` still does.

The INSTALL.md paragraph now recommends that form and carries a warning that `NO_CONFIRM=1` and `CI=true` disarm the teardown confirmation. Worth noting the `CI=true` half is pre-existing and unchanged by this PR: any CI environment setting `CI=true` — which is most of them — has always had teardown confirmations short-circuited. That seems worth its own issue rather than a quiet fix here; happy to file it.

bnaylor:
Right, and it is a leftover from my own revert rather than a fresh mistake: 44d7c33 seeded `NO_CONFIRM` from the environment, af77c90 reverted that, and the warning I had written to describe the hazard was not updated to match. So the code comment said the hazard was prevented while INSTALL.md said it was current.

Verified the corrected claim in both directions rather than just deleting the word:

- `NO_CONFIRM=1` exported: `confirm_action` prints its prompt and aborts, so the variable is genuinely ignored.
- `--no-confirm`: short-circuits.
- `CI=true`: short-circuits, which is the half that was always true.

Warning now names only `CI=true`, and adds that an exported `NO_CONFIRM` is ignored and settable per-invocation only — since that is the thing a reader is most likely to try after seeing the variable mentioned in the code.

bnaylor:
This is obsolete.

Issue #538, opened 2026-08-05
Slack is silently dead on a clean install: the agent's ConfigMap config is overwritten on first boot


bnaylor:
## Symptom

On a clean install with `SLACK_ENABLED=true`, the Slack bot connects and then ignores everything.  No error appears anywhere:

- the Pod is `5/5 Running`, `restartCount=0`
- the `PlatformAgent` CR reports `phase: Ready`, `Ready=True (Reconciled)`
- the sidecar logs `A new session has been established` and `Slack relay enabled workspaces=1`

Every health check a person would run says the install is fine.  Messages to the bot get no reply.

Restarting the agent container fixes it permanently, and any queued messages are then replayed.

## Root cause

The Chat Agent's live config never gets the Slack platform.

`renderConfigYAML` produces the right thing, and it lands in the `platform-agent-config` ConfigMap:

```yaml
platforms:
  google_chat:
    enabled: false
  slack:
    enabled: true
```

But inside the Pod, `/opt/data/config.yaml` has no `slack` key at all.  With no Slack platform configured, Hermes starts no Slack consumer, so nothing ever calls the relay's `/v1/chat/slack/events`.  Events pile up unread in the sidecar's queue.

Relay traffic on a broken Pod — note the total absence of the Slack endpoints:

```
48  "GET  /healthz
44  "POST /v1/exec
 0  ... /v1/chat/slack/events
```

and on a working one:

```
35  "POST /v1/chat/slack/api
16  "GET  /v1/chat/slack/events
 3  "POST /v1/chat/slack/events/ack     <- the queued messages, replayed
 1  "POST /v1/chat/slack/bootstrap
```

## Reproduction

Two experiments, differing only in whether the ConfigMap already existed when the Pod started.

**A — first-install path.**  Delete the `PlatformAgent` CR (its ConfigMap is garbage-collected with it), then re-run `provision_08`, so the CR, ConfigMap, Deployment and Pod are all created together:

```
ConfigMap created: 2026-08-05T02:19:06Z
Pod created:       2026-08-05T02:19:06Z
Pod started:       2026-08-05T02:19:08Z
restartCount:      0

slack mentions in /opt/data/config.yaml: 0
ConfigMap 92 lines vs live file 72 lines -> DIFFERENT
slack event polls: 0
```

**B — Pod recreated against an existing ConfigMap.**  Delete only the Pod, leaving the ConfigMap in place:

```
slack mentions in /opt/data/config.yaml: 1
[...]

PR #558, opened 2026-08-05
fix(agent): take the default profile's config.yaml from the directory mount

bnaylor:
## Why This Change

On a clean install, Slack connects and then ignores everything, with no error anywhere: the Pod is `5/5 Running` with `restartCount=0`, the CR reports `phase: Ready`, and the sidecar logs `A new session has been established` / `Slack relay enabled workspaces=1`.

The agent's live `config.yaml` has no `platforms.slack` entry, so Hermes starts no Slack consumer and nothing ever polls `/v1/chat/slack/events`. Messages queue in the sidecar unread. Restarting the Pod fixes it permanently and the queued messages replay.

The operator's ConfigMap is correct throughout — it contains `slack: enabled: true`. It just is not what the agent reads.

Fixes the root cause behind #538.

## Root cause

The operator mounts the same ConfigMap two ways:

- a `subPath` mount straight onto `$HERMES_HOME/config.yaml`
- the whole ConfigMap as a directory at `/opt/agent-config`, outside `$HERMES_HOME`

On a first boot against a brand-new PVC, kubelet does **not** establish the subPath. From `/proc/mounts` in a failing Pod:

```
/dev/sdc  /opt/data                   ext4 rw     <- the PVC
/dev/sda1 /opt/data/leader_elect.py   ext4 ro     <- subPath, mounted
/dev/sda1 /opt/data/SETTINGS.md       ext4 ro     <- subPath, mounted
                                                  <- config.yaml absent
```

`leader_elect.py` is a subPath from the *same* volume and it mounts. Both are declared in the Pod spec, both keys exist in the ConfigMap, and there are no `FailedMount` events.

With no mount there, step 2a's unconditional `cp -f /opt/defaults/config.yaml` wins, and the image default — which has no `slack` key — is what the agent loads.

## Isolating it

Changing one thing at a time, three runs:

| PVC | Deployment | ConfigMap | `config.yaml` mounted | Slack |
| **fresh** | fresh | fresh | no | dead |
| populated | **fresh** | same | yes | works |
| populated | same | same | yes | works |

Deleting the `PlatformAgent` CR also deletes the PVC, which is what makes a first install the exposed case. Deployment recreation and ConfigMap age are both ruled out.

I also ruled out the entrypoint being what removes the mount. `cp -f` cannot replace a live subPath — tested against `SETTINGS.md` in a running Pod:

```
cp: cannot remove '/opt/data/SETTINGS.md': Device or resource busy
```

and the mount survived. So on a populated PVC the force-copy fails harmlessly and the ConfigMap content wins; on a fresh PVC there is no mount for it to lose against.

**What this does not explain** is why kubelet skips one file subPath from a volume and not its sibling, deterministically, only when the PVC is new. That needs kubelet logs I could not get at on GKE. The fix does not depend on knowing.

## What Changed

**Files:** `deploy/shared/docker-entrypoint.sh`

One copy, after the image sync, taking `config.yaml` from the directory mount that is always present:

```bash
if [ -f "/opt/agent-config/config.yaml" ]; then
    cp -f "/opt/agent-config/config.yaml" "$TARGET_DIR/config.yaml" 2>/dev/null || true
fi
```

When the subPath *did* mount, this fails with EBUSY against the mount point and the already-correct content stays — the `|| true` is a no-op on that path, not a fallback. The `-f` guard also keeps it inert against an operator too old to create `/opt/agent-config` (added in #514).

Only the default profile's `config.yaml` is touched; cluster profiles are identity-stamped at scaffold time and synced by the separate loop further down.

## Testing

Verified against a live cluster on **current `main`**, after force-pulling today's operator image (the first reproduction ran against an operator that predated #514, so I redid it):

- Reproduced the failure twice on a fresh PVC: `config.yaml` unmounted, `slack` absent from the live config, `leader_elect.py` mounted.
- Confirmed `/opt/agent-config/config.yaml` is the correct source: 1598 bytes, byte-identical to the ConfigMap, contains the `slack` entry. The live file was the 1241-byte image default.
- Ran the exact line this PR adds, inside a failing Pod: the live config went from 1241 bytes with `slack=0` to 1598 bytes with `slack=1`, byte-identical to the ConfigMap.

**Not verified in-cluster:** the entrypoint executing it at the right point in the sequence, and Slack coming up as a result. That needs a rebuilt agent image, which I could not produce on this machine. The end state it produces is the same one a Pod restart produces today, and that is known to bring Slack up — but the sequencing itself is unverified, so it is worth a smoke test on a fresh PVC before this is trusted.

---

One line of behaviour change, on the path where the config is currently wrong.

### DIFF
--- a/deploy/shared/docker-entrypoint.sh
+++ b/deploy/shared/docker-entrypoint.sh
@@ -39,6 +39,24 @@ if [ -d "/opt/defaults" ]; then
     done
 fi
 
+# 2a-bis. The operator owns the default profile's config.yaml, and delivers it two
+# ways: as a subPath mount straight onto $TARGET_DIR/config.yaml, and as part of the
+# whole-ConfigMap directory mount at /opt/agent-config. The subPath is not reliable —
+# on a first boot against a brand-new PVC kubelet does not establish it (its sibling
+# subPath from the same volume, leader_elect.py, mounts fine), and step 2a above then
+# leaves the image default live. The agent starts with no `platforms.slack` entry, no
+# Slack consumer runs, and chat is silently dead while every health check passes.
+#
+# So take config.yaml from the directory mount, which is a plain projected volume and
+# always present. When the subPath *did* mount, this copy fails with EBUSY against the
+# mount point and the already-correct content stays — hence the `|| true`, which is a
+# no-op rather than a fallback. This touches only the default profile's config.yaml;
+# the cluster profiles are identity-stamped at scaffold time and are synced separately
+# below, so they are unaffected.
+if [ -f "/opt/agent-config/config.yaml" ]; then
+    cp -f "/opt/agent-config/config.yaml" "$TARGET_DIR/config.yaml" 2>/dev/null || true
+fi
+
 # 2b. Force-sync the shared scripts, for the reason step 2a gives for the default
 # profile's files: they are image-owned, never runtime state, and `cp -ru` above can skip
 # them. It skips whenever the destination looks newer, which covers both a rollback to an


bnaylor:
Now verified end to end on a live cluster, against the exact failure condition. The PR description said the sequencing was unverified; it no longer is.

A patched `platform-agent` image was built from this branch, pushed to Artifact Registry, and the `PlatformAgent` CR deleted so the PVC was recreated — which is what triggers the bug.

**Result on a first boot, fresh PVC, `restartCount=0`, no restart at any point:**

```
config.yaml mounted?  0          <- the subPath still does not establish
live config           1598 bytes, platforms.slack present
Slack relay enabled workspaces=1
A new session has been established
1x /v1/chat/slack/bootstrap, 4x /v1/chat/slack/events, 2x /v1/chat/slack/api
```

The same condition before the fix gives 1241 bytes, no `platforms.slack`, and zero relay traffic. So the mount failure is unchanged — the fix does not paper over it, it just stops the config depending on it — and Slack now comes up on the first boot instead of needing a Pod restart.

I confirmed the fix was actually in the artifact rather than trusting the build, by extracting `/usr/local/bin/agent-entrypoint` from the image layers and grepping it:

```
if [ -f "/opt/agent-config/config.yaml" ]; then
    cp -f "/opt/agent-config/config.yaml" "$TARGET_DIR/config.yaml" 2>/dev/null || true
```

Two unrelated things surfaced during the test, neither caused by this change:

- Pointing `spec.deployment.image` at another registry also redirects the credential-proxy image, which I had not built. `CREDENTIAL_PROXY_IMAGE` on the operator is the intended override and worked.
- The test image was cross-built on arm64, and `k8s-event-watcher` came out aarch64 inside an amd64 image (`exec format error`), because the `watcher-builder` stage runs `go build` with no `GOARCH`/`TARGETARCH`. Filed separately. It kept the CR at `Degraded` during the run but left `platform-agent` and the sidecar healthy, so it does not affect the result above.

The cluster has been returned to the published images and is `Ready` with Slack polling.

dshnayder:
**Stale invariant in the operator, outside this diff**

Two comments in `k8s-operator/` assert the invariant this PR disproves. I'd amend them here rather than leave them — they're a good part of why the failure was invisible.

`platformagent_manifests.go:2031-2035`:

> In the gateway this exact path is a ConfigMap mount, and ConfigMap volumes are always read-only, so the entrypoint's copy from `/opt/defaults` cannot land a `config.yaml` on the volume underneath it (hence step 3's `[ -w ]` guard).

`platformagent_manifests_test.go:2122-2126`:

> In the gateway container `$HOME/config.yaml` is a ConfigMap mount, and ConfigMap volumes are always read-only, so the entrypoint's copy from `/opt/defaults` can never land a `config.yaml` on the PVC underneath it.

The premise is true; the conclusion is not. Read-only makes the copy fail *where the mount exists* — and the finding in this PR is that on a fresh PVC it sometimes does not, at which point step 2a's copy lands squarely on the volume and is exactly what goes live. That inference is what makes step 2a look harmless, so anyone reading either comment re-derives the assumption that hid this bug, quite possibly while editing step 2a itself.

Suggested shape for the operator comment:

```go
// ... In the gateway this path is normally a ConfigMap subPath mount, and ConfigMap
// volumes are read-only, so the entrypoint's copy from /opt/defaults fails against it
// (hence step 3's `[ -w ]` guard). That is not guaranteed: on a fresh PVC the kubelet
// can skip the subPath, and the copy then DOES land on the volume — see step 2a-bis of
// deploy/shared/docker-entrypoint.sh, which is what makes the rendered config win
// either way.
```

plus the equivalent qualification on `TestDashboardReadsTheRenderedConfig`, whose doc comment currently states the unqualified version as the reason the test exists.

dshnayder (approved):

PR #559, opened 2026-08-06
chore(cron): drop the five retired platform watchdogs from the roster

toshiowang:
# Pull Request

## Why This Change

`agents/platform/cron/jobs.json` carried eleven jobs, five of them shipping `enabled: false` for several releases: `blueprint-sync`, `policy-propagation`, `global-capacity-orchestrator`, `standardization-validator`, and `lifecycle-deprecation-manager`.

None can produce a finding on a stock install, which is why they were switched off in the first place:

- `blueprint-sync` and `standardization-validator` compare clusters against a "master blueprint" / "corporate architectural patterns" document no install provides. `fleet-consistency-drift` already covers the intent by deriving its baseline from the live fleet.
- `policy-propagation` reads policy templates from `/opt/defaults/templates/`, which is not shipped.
- `global-capacity-orchestrator` ran hourly with no defined output artifact.
- `lifecycle-deprecation-manager` overlaps `security-patch-orchestrator`.

Keeping them listed costs review attention on every pass over the roster and invites someone to flip a flag on a job whose SOP cannot succeed. The schedule was never the problem; the SOP is.

## What Changed

**Files:**

- `agents/platform/cron/jobs.json`
- `agents/platform/scripts/profile_scaffold.py`
- `agents/platform/scripts/test_profile_scaffold.py`
- `docs/site/src/content/docs/concepts/autonomous-watchdogs.md`
- `docs/site/src/content/docs/concepts/governance-sops.md`
- `docs/site/src/content/docs/overview/proactive-autonomy.md`
- `docs/site/src/content/docs/reference/cron-jobs.md`
- `docs/README.md`

**Added/Changed/Fixed:**

- Removed the five disabled entries. The roster is now six jobs, all enabled: the five fleet audits plus `github-issue-resolver`.
- Reworded `merge_cron_store`'s docstring, which cited these five as its standing example of "retire with `enabled: false`, do not delete". It now states the two-step retirement rule instead.
- Corrected the same docstring's account of what a straight `copytree` broke — see Context.
- Updated the matching rationale comment on `test_a_job_withdrawn_from_the_image_is_not_pruned`, which cited the five as entries still present in `agents/platform/cron/`. The assertion is unchanged.
- Renamed the docs sections that existed only to explain the five ("The disabled jobs" → "The retired jobs"; "The disabled SOPs" → "The unscheduled SOPs") and updated every inbound anchor.
- Regenerated the `reference/cron-jobs.md` table via `make docs-generate`, and fixed two hand-written field-table rows that referenced "the disabled governance jobs".
- The SOPs themselves stay on disk under `agents/platform/governance/`. Reviving one means rewriting the SOP and re-adding a job, not just re-adding a job.

## Why This Matters

**Deleting an entry is not removal on a cluster that already has it.** `merge_cron_store` (`agents/platform/scripts/profile_scaffold.py:191`) keeps volume jobs the image is silent about, so an id dropped from the shipped roster stays on the PVC in whatever state the last release left it — and the image loses its ability to hold it off.

That makes this deletion safe only because the disabled form was merged everywhere first: every provisioned cluster already has `enabled: false` recorded on its volume, so the five stay off with no help from the image. It would **not** be safe for an entry some cluster has enabled. The docstring, its test's rationale comment, and `concepts/autonomous-watchdogs.md` now all spell out that ordering — ship `enabled: false`, let it merge, then delete — rather than leaving it as tribal knowledge.

No behavior change on any existing cluster.

## Context

While verifying the above I checked the `merge_cron_store` docstring's claim that erasing run history meant "a daily audit could fire twice in a morning". It does not hold against the scheduler in the image, so the docstring is corrected here:

- A wiped `next_run_at` is recomputed from **now** (`cron/jobs.py:1976-1977` → `compute_next_run` line 775), always landing on a future occurrence, and the due check requires `next_run_dt <= now`. A recurring job cannot replay.
- `advance_next_run` runs up front, right after `get_due_jobs` (`cron/scheduler.py:3946`, `:3961`), before the job executes — so even a restart mid-run does not re-trigger it.
- What state loss *does* re-fire is one-shots: `_recoverable_oneshot_run_at` returns `None` once `last_run_at` is set (`cron/jobs.py:670-671`), so that field **is** the already-ran guard. Erase it and the job is eligible again — recognizably the bug behind #536.
- For recurring jobs the failure is the opposite: destroying the stale past `next_run_at` throws away the input to the catch-up window, so a merely-late audit is silently skipped rather than caught up.

Related: #517 introduced `merge_cron_store`; #439 added `profile_scaffold.py`.

Follow-up, not in scope here: the `platform` profile has no cron ticker at all today — ticking is a property of a running gateway and only the `default` profile has one — so none of the six remaining jobs currently fire. That is being worked separately.

## Testing

- `python3 -m unittest test_profile_scaffold` — 21 tests, all pass.
- `make docs-generate` — regenerated; re-running leaves the tree clean.
- `make docs-check` — all four checks pass (generated regions, relative links, terminology, map inventory).
- `npx prettier --check` on every changed Markdown and JSON file — clean.
- `python3 -c "import json; json.load(...)"` on `jobs.json` — parses; six jobs, all `enabled: true`.
- `review-docs-drift` skill run against the branch diff. One Blocking finding, fixed in `e7fad17`: `docs/README.md` still described the five SOPs as "disabled".
- kube-agents-bot review finding, fixed in `ca79bdc`: the stale rationale comment in `test_profile_scaffold.py`. Neither of my own sweeps reached it — the comment names none of the five ids, and the test calls the function through a helper, so `merge_cron_store` appears nowhere in the file.

---

Documentation-and-configuration only; no operator or runtime code is touched. Existing clusters are unaffected — the five entries persist on their volumes in the disabled state they are already in. Fresh installs get a roster of six jobs that all do something.


toshiowang:
Good catch — fixed in `ca79bdc`.

The comment on `test_a_job_withdrawn_from_the_image_is_not_pruned` now describes the two-step retirement instead of pointing at entries this PR deleted:

> Shipping it with `enabled: false` does, so retiring a watchdog takes two steps: ship it disabled, let every live cluster merge that state, and only then drop the entry — from that point the volume's own copy keeps it off with no help from the image. That is the path the five retired watchdogs took out of `agents/platform/cron/`.

The assertion is untouched. 21 tests still pass.

Worth recording why my own sweeps missed it, since the same blind spot would recur: I grepped for the five job ids, and the comment names none of them; I also grepped for `merge_cron_store`, which returns only the source file because the test reaches it through a `self.overlay` helper. A rationale comment that refers to its subject by description rather than identifier is invisible to both.

toshiowang:
Both fixed in `b7065f2`, including the entrypoint copy you pointed at.

**1. The retracted re-fire claim.** You're right that the module and its test suite were asserting opposite semantics. `CronStoreMergeTest`'s class docstring and `deploy/shared/docker-entrypoint.sh:146-147` now describe the same failure modes as `merge_cron_store`: an erased `last_run_at` re-fires a one-shot because that field *is* the already-ran guard, while a recurring job's wiped `next_run_at` is recomputed from now, so a late audit is skipped rather than caught up.

**2. `governance-sops.md`.** The closing note now covers `github-issue-resolver` preloading its namesake skill and says an SOP needing no preload can omit the key — no reference to a disabled job or an empty array.

That's three findings from you on this PR, all of them prose that restated something the diff invalidated, and all three invisible to my sweeps because they name their subject by description rather than by identifier. So I stopped grepping ids and grepped the claim instead:

```
grep -rniE "fire[ds]? (a )?(second time|again|twice)|twice in a (morning|day)|fired again|double.?fire|ran twice|run twice"
```

That returns no further copies, so I believe the retraction is complete.

Gates: 21 tests pass, `bash -n` clean on the entrypoint, `make docs-check` green, prettier clean.

bradhoekstra (approved):

PR #561, opened 2026-08-06
fix(docker): cross-compile k8s-event-watcher for the target platform

### LINKED ISSUE #560: Dockerfile cross-builds produce an unusable k8s-event-watcher: the Go stage ignores TARGETARCH
## What happens

The `watcher-builder` stage compiles for whatever the builder happens to be:

```dockerfile
FROM golang:1.25-alpine AS watcher-builder
...
RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /workspace/k8s-event-watcher ./cmd/k8s-event-watcher/
```

No `GOARCH`, no `TARGETARCH`, and nothing else in the Dockerfile sets them. Building on an arm64 machine for an amd64 cluster therefore yields an image whose manifest says `amd64` while `/usr/local/bin/k8s-event-watcher` inside it is `aarch64`.

The container then crashloops:

```
exec /usr/local/bin/k8s-event-watcher: exec format error
```

Confirmed by reading the ELF header straight out of the image layers: `e_machine=0xB7` (aarch64), in an image whose config reports `architecture: amd64`.

## Why it is worth fixing rather than documenting

The failure is quiet in the ways that matter:

- `crane config` and `docker inspect` both report `amd64`, so the image looks correct.
- Every other container in the Pod starts normally — only the Go binary is wrong, so it reads as "the event watcher is broken" rather than "the image is mis-built".
- The Pod reaches 4/5 and the `PlatformAgent` sits at `Degraded`, which invites debugging the watcher rather than the build.

CI is unaffected because it builds natively on amd64 runners, so this only bites local and dev builds — including `make dev-rebuild-agent --local` and anyone on an Apple Silicon machine following the Docker build in AGENTS.md.

## Suggested fix

Make the builder stage cross-compile properly, which is the standard Docker pattern:

```dockerfile
FROM --platform=$BUILDPLATFORM golang:1.25-alpine AS watcher-builder
ARG TARGETARCH
ARG TARGETOS
...
RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} \
    go build -ldflags="-w -s" -o /workspace/k8s-event-watcher ./cmd/k8s-event-watcher/
```

This is also faster than the alternative: the Go stage runs natively on the builder and cross-compiles, instead of running the whole toolchain under emulation.

Worth a matching check somewhere, since the symptom is so indirect — even a `file`/ELF assertion on the built binary in the docker-build workflow would turn this into a build failure rather than a runtime one.

Found while building a patched agent image to verify #558 on a live cluster.

bnaylor:
## Why This Change

The `watcher-builder` stage compiled the event watcher for whatever architecture the build host happened to be, so building on an arm64 workstation put an **aarch64** binary inside an image that ships as **amd64**.  At runtime:

```
exec /usr/local/bin/k8s-event-watcher: exec format error
```

Fixes #560.

## Root cause

Nothing in the repo ever says what platform these images are for.  The bases (`nousresearch/hermes-agent`, `envoyproxy/envoy`) are multi-arch indexes, no build invocation passes `--platform`, and deployment targets are always amd64 GKE nodes.  So each stage quietly resolves to the build host, or - when a pin interferes - to something else entirely.

The branch's original pin (`v2026.7.20@sha256:a6ce64e...`) muddied the diagnosis.  That digest is the amd64 *child* of a multi-arch index, so on this branch the base was effectively single-arch and the image always shipped amd64 while the watcher followed the host.  An earlier revision of this PR concluded "the base is single-arch amd64" from exactly that measurement and hardcoded `GOARCH=amd64` accordingly.  main has since moved `tags.env` to the v2026.8.3 *index* digest, which platform-selects again - so the hardcoded pin breaks on rebase: an arm64 host builds an arm64 image with an amd64 watcher inside.  Same failure, inverted.  envoy has the same exposure with no digest at all.

## What Changed

**Files:** `deploy/docker/Dockerfile`, `Makefile`, `deploy/docker/cloudbuild.yaml`, `k8s-operator/scripts/dev/dev_rebuild_agent.sh`, `.github/workflows/docker-build.yml`, `.github/workflows/docker-publish-ghcr.yml`, `AGENTS.md`

The platform is a fact about deployment, not about any one base image, so it is now pinned at every build invocation: `--platform linux/amd64` in the Makefile, cloudbuild, the dev rebuild script and the documented bare `docker build`, and `platforms: linux/amd64` in both workflows.  That covers `envoy-bin` too - a `FROM` without a flag resolves to the target platform.

With that in place the Dockerfile goes back to the standard cross-compile shape:

```dockerfile
FROM --platform=$BUILDPLATFORM golang:1.25-alpine AS watcher-builder
ARG TARGETOS
ARG TARGETARCH
RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} \
    go build -ldflags="-w -s" -o /workspace/k8s-event-watcher ./cmd/k8s-event-watcher/
```

This is correct under either `tags.env` pin, and under any future `--platform`.  `$BUILDPLATFORM` keeps the Go toolchain running natively rather than emulated.  The `WATCHER_GOARCH` arg from the previous revision is gone.

main is merged into the branch, mostly for `tags.env`: CI on this PR now builds against the multi-arch index pin, which is the case the earlier revisions got wrong.

The three earlier revisions are left in the branch history with the review comments explaining why each was wrong or a no-op.  Short version: `GOARCH=${TARGETARCH}` alone changes nothing when no invocation passes `--platform`, a stage-scoped `ARG` default silently expands to empty, and `GOARCH=amd64` only worked while the base pin was a single-arch child digest.

## Testing

Still no Docker on this machine (which is how the bug surfaced in the first place).  What was checked:

- Registry manifests, verified directly this time: `a6ce64e...` is a bare image manifest, while the `v2026.7.20` tag, main's `16788311...` pin, and `envoyproxy/envoy:v1.38.0` are all multi-arch indexes (amd64 + arm64).
- The exact cross-compile the stage now runs, executed on this arm64 host: `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/k8s-event-watcher/` produces an `ELF 64-bit ... x86-64, statically linked` binary.
- `hadolint` reports no new findings.
- `make docs-check` and prettier pass on the changed files.

The Docker Build job runs on amd64 runners, so it proves syntax and wiring rather than the cross-build.  Demonstrating the fix end-to-end still needs an arm64 host with Docker: build, then check that `k8s-event-watcher` inside the image is `x86-64`.

---

Build-time only.  No change to what an amd64 build produces, which is every build CI performs today.


bnaylor:
This is right, and the first version of the fix did nothing. Verified both premises before changing it:

```
$ crane manifest nousresearch/hermes-agent@sha256:a6ce64e...
  mediaType: application/vnd.oci.image.manifest.v1+json
  SINGLE-ARCH manifest (no platform list)
$ crane config  ... -> arch: amd64  os: linux

$ grep -rn -- "--platform|platforms:" Makefile deploy/docker/cloudbuild.yaml \
    dev_rebuild_agent.sh docker-build.yml docker-publish-ghcr.yml
  (no matches)
```

So the chain is exactly as described: with no `--platform` anywhere, `TARGETPLATFORM` is the host, an unpinned `FROM golang:...` already resolved to it, and the original build was producing host-arch output via emulation. Pinning to `$BUILDPLATFORM` and setting `GOARCH=$TARGETARCH` selects the same architecture natively — faster, identical output. On the arm64 workstation, `TARGETARCH=arm64` both before and after, and the watcher stays aarch64.

The part I had backwards is where the amd64 in the image comes from. It is not `TARGETARCH`; it is inherited from the single-arch base. That divergence between "arch the image ships as" and "arch `TARGETARCH` names" is the bug, and deriving `GOARCH` from `TARGETARCH` reproduces it rather than fixing it. My comment above the stage asserted the opposite mechanism, which would have sent the next person the wrong way — rewritten.

Corrected in 8b6ab98: `GOARCH` is pinned to `amd64` behind a `WATCHER_GOARCH` arg, because the single-arch base means the final image can only be amd64. `$BUILDPLATFORM` stays so the toolchain still runs natively rather than emulated.

On the other option offered — pinning the target platform at the build invocations — I did not take it, for two reasons. It touches five call sites across the Makefile, the dev script, cloudbuild and two workflows, which is a wider change than the bug needs; and it would encode `linux/amd64` in five places instead of one, next to a base image that already dictates it. If you would rather have it there, the `WATCHER_GOARCH` arg is the seam and I am happy to move it.

Still not built locally — no Docker on that machine, which is how this arose. `hadolint` is clean on the changed lines. The **Docker Build** job is the check for syntax and arg wiring; the cross-build behaviour needs an arm64 host, which CI does not have, so that half remains unverified by CI either way.

bnaylor:
Correct, and this one would also have regressed a path that worked. Fixed in the latest commit.

The scoping is the part I got wrong: `ARG WATCHER_GOARCH=amd64` sat after `FROM envoyproxy/... AS envoy-bin`, so it belonged to that stage, and the valueless re-declaration in `watcher-builder` had no global to inherit from. `GOARCH=` empty is silent — Go treats it as unset and uses the toolchain's own architecture, which with `--platform=$BUILDPLATFORM` is the build host. Same aarch64 binary, no error, no lint.

The regression point stands too: `docker build --platform linux/amd64` on an arm64 host previously ran the golang stage emulated as amd64, so an unset `GOARCH` defaulted correctly. Pinning to `$BUILDPLATFORM` moved it to a native arm64 toolchain while `GOARCH` stayed empty, breaking the one invocation that used to produce a good image.

Both suggestions applied rather than one: the defaulted `ARG` is hoisted above the first `FROM` so it is a genuine global and `--build-arg WATCHER_GOARCH=` works, and the `RUN` uses `${WATCHER_GOARCH:-amd64}` so the expansion stays correct if the declaration is ever moved again.

The contrast drawn with `k8s-operator/Dockerfile` is the useful bit and I have noted it in the commit: `TARGETARCH` is predefined and populated by BuildKit on declaration, so its valueless `ARG` behaves differently from a user-defined one. Citing it as precedent hid that difference.

`hadolint` clean. Still unbuilt locally; CI's Docker Build job covers syntax and arg wiring on amd64, and the arm64 behaviour needs a host neither of us has in CI.

bnaylor:
Right — and worse than stale, since the two `TARGETARCH` claims were inverted relative to what the code now does, and a squash merge would have made that body the permanent commit message for a pin someone will need to find later.

Description rewritten to match the head commit. It now leads with the actual root cause (single-arch base fixes the image at amd64; no invocation passes `--platform`, so the build platform is the host), documents `WATCHER_GOARCH`, and states plainly that `TARGETARCH` is deliberately not used because it reproduces the bug here.

I also corrected the `k8s-operator/Dockerfile` comparison rather than dropping it. The reason that file can rely on `TARGETARCH` is that its runtime base, `gcr.io/distroless/static`, is multi-arch, so image and target agree. Citing it as flat precedent was what led me to the wrong fix twice.

The two dead-end revisions are left in the branch history with the review comments explaining why each was a no-op, since the next person to touch this will likely reach for `TARGETARCH` first.

jayantid (changes requested):
Some findings from claude -

**1. The amd64 guarantee comes from the pin, not the base — and it breaks on rebase.**
  
  Your check of `a6ce64e…` is right for this branch, but not for the stated reason. The
  `v2026.7.20` tag is a multi-arch index; `a6ce64e…` is just its amd64 child. Pinning
  `tag@child-digest` means the digest wins and no platform selection happens — so amd64 comes from
  the pin, not from the base being single-arch.

  Main has already moved on. `tags.env:1` on main pins `v2026.8.3@sha256:16788311…`, which is the
  **index**:

  ```bash
  crane manifest nousresearch/hermes-agent@sha256:16788311… | jq '.manifests[].platform'
  # -> linux/amd64 AND linux/arm64
  ```

  After rebase the base platform-selects again: an arm64 host builds an arm64 image, and the
  hardcoded `GOARCH=amd64` puts an amd64 binary inside it. Same exec format error, inverted — with
  no error, no lint, and no CI signal.

  Fix: `--platform=linux/amd64` at the build invocations plus `GOARCH=${TARGETARCH}` — correct
  under both pins. Also bump `tags.env` on this branch to match main.

**2. `envoy-bin` has the same bug, with no pin protecting it.**

  Line 8 is `FROM envoyproxy/envoy:${ENVOY_VERSION}` — tag only, no digest — and that image is
  multi-arch, so it platform-selects to the build host today. Line 279 copies the binary into the
  credential-proxy image, so envoy matches the builder rather than the final image.

  Needs `--platform` too.

bnaylor:
Both findings held up under verification, and the premise of the previous fix was just wrong.  `crane manifest` on the `v2026.7.20` *tag* returns a multi-arch index (amd64 + arm64), so the "single-arch base" I measured was an artifact of the child-digest pin, not a property of the image.  main's `16788311...` pin is the index, and `envoyproxy/envoy:v1.38.0` is multi-arch too - all three re-checked against the registry before touching anything.

Fixed in 91f07cc, taking the shape you suggested:

- `--platform linux/amd64` at every build invocation: Makefile, cloudbuild.yaml, dev_rebuild_agent.sh, both workflows, and the documented bare `docker build` in AGENTS.md.  That covers `envoy-bin` as well, since a `FROM` without a flag resolves to the target platform.
- The Dockerfile goes back to `GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64}` with the golang stage on `$BUILDPLATFORM`.  `WATCHER_GOARCH` is gone.
- main is merged into the branch rather than cherry-picking the `tags.env` bump, so CI on this PR now builds against the index pin - the case the previous revision got wrong.

Verified on this arm64 machine that the exact cross-compile the stage runs emits a static x86-64 ELF.  Still no local Docker here, so the in-image check (the watcher inside a `--platform linux/amd64` build reports `x86-64`) remains the one thing only an arm64 host with Docker can confirm.

PR description rewritten to match.

rune42808:
## Experimental verification — amd64 host with Docker ✅

Built on an amd64 machine with Docker:

```bash
docker build --platform linux/amd64 \
  --build-arg HERMES_AGENT_TAG="v2026.8.3@sha256:16788311..." \
  --target platform -t kube-agents-test:amd64 \
  -f deploy/docker/Dockerfile .
```

Extracted the watcher binary from the image and checked architecture:

```
$ file k8s-event-watcher
k8s-event-watcher: ELF 64-bit LSB executable, x86-64, version 1 (SYSV),
statically linked, BuildID[sha1]=08f37d..., stripped
```

`x86-64` ✅ — the cross-compile in `watcher-builder` is working correctly.
The `--platform linux/amd64` flag propagates through the build chain:
golang:1.25-alpine runs natively on the amd64 builder, `GOARCH=${TARGETARCH:-amd64}`
picks up `TARGETARCH=amd64` from the target platform, and the resulting binary
is a static x86-64 ELF.

Also confirmed via ELF header: `xxd | head -2` shows `7f45 4c46 ... 3e00 ...`
(EM_X86_64 at offset 0x12).

The `:-amd64` fallback in `GOARCH=${TARGETARCH:-amd64}` is the safety net for
builders that don't populate TARGETARCH, which is the correct place for a default:
the last-layer pin, not a stage-scoped ARG that can go empty.

bnaylor:
## Experimental verification — arm64 host with Docker ✅

Built on an arm64 machine (Apple Silicon Mac, Docker Desktop) with the explicit platform flag:

```bash
docker build --platform linux/amd64 \
  --build-arg HERMES_AGENT_TAG="v2026.8.3@sha256:16788311e2fa3035456bdc1bafb8ec2b1777db64ebf020af9bb7eb73c3712c9e" \
  --target watcher-builder \
  -f deploy/docker/Dockerfile .
```

ELF header check on the binary inside the image:

```
$ docker run --rm --platform linux/amd64 --entrypoint sh watcher-check \
    -c 'head -c 20 /workspace/k8s-event-watcher | od -An -tx1'
 7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
 02 00 3e 00
```

Byte 18 (offset 0x12) is `3e` — `EM_X86_64` ✅. Not `b7` (aarch64).

The golang stage ran natively on the arm64 build host (`$BUILDPLATFORM`), `TARGETARCH=amd64` was injected by `--platform linux/amd64`, and `GOARCH=${TARGETARCH:-amd64}` cross-compiled correctly to x86-64. This is the case that every previous revision of this fix got wrong.

jayantid (approved):

PR #563, opened 2026-08-06
ci(gcloud): pin Google Cloud SDK to version 579.0.0 in release workflows

eLeontev:
### Summary

Pins the Google Cloud SDK version to `579.0.0` in the release workflows.

### UPDATE

Following discussion and investigation of the RC pipeline failure (https://github.com/gke-labs/kube-agents/actions/runs/31110976882/job/92648454225):
- The `--managed-otel-scope=COLLECTION_AND_INSTRUMENTATION_COMPONENTS` flag is supported natively in GA for `gcloud container clusters create` in modern Google Cloud SDK releases (`579.0.0+`).
- The GitHub Actions runner VM previously fell back to an older pre-installed Google Cloud SDK release on the runner image where the flag was not yet in GA for cluster creation.
- The correct and robust fix is to pin the recommended Google Cloud SDK version (`version: "579.0.0"`) in `setup-gcloud` across the workflows (`rc-deploy-environment.yml` and `rc-release-pipeline.yml`), keeping `provision_01_gcp_cluster.sh` on clean GA `gcloud container clusters create` without relying on beta release tracks or unpinned floating versions.

### Verification
- `make docs-check` passes.
- `make validate` passes.
- `go test ./...` in `k8s-operator` passes.

### DIFF
--- a/.github/workflows/rc-deploy-environment.yml
+++ b/.github/workflows/rc-deploy-environment.yml
@@ -84,6 +84,7 @@ jobs:
       - name: Set up Google Cloud SDK
         uses: google-github-actions/setup-gcloud@aa5489c8933f4cc7a4f7d45035b3b1440c9c10db # v3.0.1
         with:
+          version: "579.0.0"
           install_components: "beta,gke-gcloud-auth-plugin"
 
       - name: Set up Go
--- a/.github/workflows/rc-release-pipeline.yml
+++ b/.github/workflows/rc-release-pipeline.yml
@@ -77,6 +77,7 @@ jobs:
       - name: Set up Google Cloud SDK
         uses: google-github-actions/setup-gcloud@aa5489c8933f4cc7a4f7d45035b3b1440c9c10db # v3.0.1
         with:
+          version: "579.0.0"
           install_components: "beta,gke-gcloud-auth-plugin"
 
       - name: Connect to GKE & Wait for Pod Readiness
--- a/k8s-operator/scripts/provision_01_gcp_cluster.sh
+++ b/k8s-operator/scripts/provision_01_gcp_cluster.sh
@@ -53,7 +53,7 @@ verify_cluster() {
 }
 execute_cluster() {
   print_info "Creating GKE Standard Cluster with Workload Identity. This takes approximately 5-8 minutes in Google Cloud..."
-  gcloud beta container clusters create "$CLUSTER_NAME" \
+  gcloud container clusters create "$CLUSTER_NAME" \
       --region "$REGION" \
       --machine-type="e2-standard-4" \
       --num-nodes=1 \
--- a/k8s-operator/scripts/provision_01_gcp_cluster.sh
+++ b/k8s-operator/scripts/provision_01_gcp_cluster.sh
@@ -53,7 +53,7 @@ verify_cluster() {
 }
 execute_cluster() {
   print_info "Creating GKE Standard Cluster with Workload Identity. This takes approximately 5-8 minutes in Google Cloud..."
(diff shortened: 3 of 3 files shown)


bradhoekstra:
I would like @bnaylor's review since the 'beta' was dropped since not all clients will have `gcloud beta` available by default.

bradhoekstra:
Can we try update to the latest `gcloud` in the deployment pipeline instead?

eLeontev:
> Can we try update to the latest `gcloud` in the deployment pipeline instead?

It's the right way to solve the issue. The pipeline is updated to use minimal recommended version to support used flags.

PR is ready to review.

bradhoekstra (approved):

bnaylor (commented):
Thanks for chasing this down.  The diagnosis matches what I see, and the fix unblocks the pipeline,
so I'm happy for it to go in.  Context for anyone reading later: the trigger was #539, which dropped
`gcloud beta` from cluster creation on the grounds that `--managed-otel-scope` is GA.  That is true
on a current SDK.  It is not true on the older gcloud preinstalled in the runner image, which is
what broke.  My change, so my apologies for the detour.

Four things, one of which I think wants a follow-up.

**1. This protects CI but not the path users actually run.**

`provision_01_gcp_cluster.sh` is the same script people run locally via `make gcp-provision` and
INSTALL.md.  #539 removed the beta fallback for everyone; this PR restores a floor only for the
runner.  There is no gcloud version preflight anywhere in `k8s-operator/scripts/`, and `INSTALL.md`
line 57 lists the requirement as "Latest", which is nominally what CI had when it failed.  So a
contributor on an older SDK hits exactly this error with nothing to catch it, and the message they
get is a raw `gcloud` usage error rather than "your SDK is too old".

Not a blocker for this PR - the pipeline being red is the urgent half.  But the user-facing half is
still open, and a version check in `common.sh` would close it for both.

**2. `579.0.0` is not the minimum, and that number will get copied.**

Local SDK **576.0.0** already has the flag on the GA surface:

```
$ gcloud version | head -1
Google Cloud SDK 576.0.0

$ gcloud container clusters create --help | grep -A2 managed-otel-scope
       --managed-otel-scope=MANAGED_OTEL_SCOPE
          Specifies the scope of Managed OpenTelemetry within the cluster.
```

So the real floor is somewhere at or below 576, not 579.  Pinning to 579.0.0 is harmless in itself
and I would not change the pin.  The wording is what I would adjust: this is a known-good pin, not
the release where the flag went GA.  It matters because that is the number someone will lift into a
documented minimum gcloud version for the local install path, where it would be both wrong and
needlessly strict.

**3. The pin will rot silently.**

Dependabot's `github-actions` ecosystem updates `uses:` refs, not `with:` inputs, so nothing in
`.github/dependabot.yml` will ever bump `version: "579.0.0"`.  Worth a short inline comment saying
why this version is pinned and what would let it move, in the same spirit as the trailing version
comments we already require on action SHAs.  Otherwise this is a number nobody dares touch in a
year's time.

**4. Two notes, no change requested.**

`install_components: "beta"` is still load-bearing even though cluster creation is now GA -
`provision_05_gcp_gchat.sh` lines 104 and 109 use `gcloud beta services identity create`.  Flagging
it so it does not get removed later as leftover from this change.

`docker-publish-gcp.yml` line 39 also uses `setup-gcloud` and stays unpinned.  It only drives Cloud
Build with no cluster create, so I do not think it needs this, but two of three workflows now pin
and one does not, with no stated reason.  Fine either way, just noting the inconsistency.

Reviewed with the help of Claude.

PR #567, opened 2026-08-06
fix(deploy): set hermes ownership on patched python files in Dockerfile

bradhoekstra:
# Pull Request

## Why This Change

Patched python files (`kanban_handoff_clip.py` and `cron_run_scope.py`) in `/opt/hermes/gateway` and `/opt/hermes/tools` were copied into the Docker container image without explicit `hermes:hermes` ownership, leaving potential permission issues for non-root runtime processes.

This is the **ownership** half of the problem. The **mode** half landed separately in #580, which gave the same two files `--chmod=0644` and added `/opt/hermes/gateway` and `/opt/hermes/tools` to the final `chmod` normalisation loop. The two changes are independent — a file can be world-readable and still owned by `root` — so this branch has been rebased onto #580 and now carries only what #580 did not cover.

## What Changed

**Files:**

- `deploy/docker/Dockerfile`

**Added/Changed/Fixed:**

- Added `--chown=hermes:hermes` alongside the existing `--chmod=0644` on the `COPY` commands for `kanban_handoff_clip.py` and `cron_run_scope.py`.
- Added `/opt/hermes/gateway` and `/opt/hermes/tools` to the `chown -R hermes:hermes` line in the final normalisation block, so the two directory trees match the `chmod` loop #580 already extended. Both path lists are kept in the same order so they stay readable side by side.

Net diff against `main` is three lines.

## Why This Matters

Ensures the non-root `hermes` user owns, not merely reads, all patched modules and their directory trees — consistent with `/opt/hermes/skills` and `/opt/hermes/plugins`, which the same block has always chowned.

## Context

- Rebased onto `main` after #580 merged; the two pull requests edited the same three lines and the resolution takes the union rather than either side.
- The `COPY` instructions live in the `agent-base` stage and the normalisation block in `platform`, which is `FROM agent-base` — so the recursive chown does reach the copied files.

## Testing

- `docker build --check --build-arg HERMES_AGENT_TAG=… --target platform` — no warnings.
- Full `docker build --target platform` against the `HERMES_AGENT_TAG` pinned in `tags.env` — succeeded.
- Inspected the resulting image: both patched files are `-rw-r--r-- hermes hermes`, both directories are `drwxr-xr-x hermes hermes`, and `find /opt/hermes/gateway /opt/hermes/tools ! -user hermes` returns nothing.
- `make docs-check` passed cleanly.

_This PR was generated with the help of Claude._

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -66,7 +66,7 @@ RUN /opt/hermes/.venv/bin/python3 -c "import pathlib; p = pathlib.Path('/opt/her
 # ledger link into ".../is" in chat. Swap both slices for a whitespace-boundary
 # clip that also stops discarding every line after the first. See the module
 # docstring in deploy/docker/patches/kanban_handoff_clip.py.
-COPY --chmod=0644 deploy/docker/patches/kanban_handoff_clip.py /opt/hermes/gateway/kanban_handoff_clip.py
+COPY --chown=hermes:hermes --chmod=0644 deploy/docker/patches/kanban_handoff_clip.py /opt/hermes/gateway/kanban_handoff_clip.py
 RUN /opt/hermes/.venv/bin/python3 -c "import pathlib; p = pathlib.Path('/opt/hermes/gateway/kanban_watchers.py'); c = p.read_text(); t1 = 'h = lines[0][:200] if lines else payload_summary[:200]'; t2 = 'r = lines[0][:160] if lines else task.result[:160]'; exit(1) if (t1 not in c or t2 not in c) else None; c = c.replace(t1, 'h = _clip_handoff(payload_summary)').replace(t2, 'r = _clip_handoff(task.result)'); p.write_text(c + '\n\n# kube-agents patch: see gateway/kanban_handoff_clip.py\nfrom gateway.kanban_handoff_clip import clip_handoff as _clip_handoff\n')" && \
     grep -q "h = _clip_handoff(payload_summary)" /opt/hermes/gateway/kanban_watchers.py && \
     grep -q "r = _clip_handoff(task.result)" /opt/hermes/gateway/kanban_watchers.py && \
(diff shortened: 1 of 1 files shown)


deploy/docker/Dockerfile
```diff
@@ -66,7 +66,7 @@ RUN /opt/hermes/.venv/bin/python3 -c "import pathlib; p = pathlib.Path('/opt/her
 # ledger link into ".../is" in chat. Swap both slices for a whitespace-boundary
 # clip that also stops discarding every line after the first. See the module
 # docstring in deploy/docker/patches/kanban_handoff_clip.py.
-COPY --chmod=0644 deploy/docker/patches/kanban_handoff_clip.py /opt/hermes/gateway/kanban_handoff_clip.py
+COPY --chown=hermes:hermes --chmod=0644 deploy/docker/patches/kanban_handoff_clip.py /opt/hermes/gateway/kanban_handoff_clip.py
```
eLeontev, line 69: I'm not an expert, but looks like there is potential security issue that hermes can override the script which can lead to the prompt injection issue.

Potentially we need to split the right to 2 categories:
- readonly (skills and other runtime tools which hermes can use)
- owner (temporary files, etc.)

If it's not relevant, please ignore.

bradhoekstra: You're right, and this is a better catch than "not an expert" suggests. I checked the specifics and the concern holds:

- `hermes` is uid 10000, which is exactly the `runAsUser` the operator pins — the agent process runs *as* `hermes`.
- The platform and cluster profiles carry the `terminal` and `code_execution` toolsets (`agents/chat/config.yaml:119,122`), so the agent can write anything `hermes` owns.
- `chown -R hermes:hermes` with mode 0644/755 means the owner can write. This PR would turn `/opt/hermes/gateway` and `/opt/hermes/tools` from a tree the agent can only read into one it can rewrite — including the very `kanban_watchers.py` and cron scheduler the patches above modify. Injection reaching code-exec would get pod-lifetime persistence inside the gateway's own import path.

And the write access buys nothing. Grepping `/opt/hermes/gateway|/opt/hermes/tools` across every script and template turns up only two docstring mentions; nothing writes there at runtime, and runtime scaffolding targets `/opt/data` (`deploy/shared/docker-entrypoint.sh:4`). Python needs read on the file and r-x on the directory to import — which is exactly what your #580 already provides with `--chmod=0644` and the 755/644 normalisation loop, whoever owns the files.

So the motivation I opened this on ("potential permission issues for non-root processes") was speculative, and #580 resolved it properly. What survived the rebase onto #580 is only the part that widens write access. Both remaining pieces — the `--chown` on the two `COPY`s and the `chown -R` — share the defect, so there's no narrower version worth keeping. Closing this as superseded by #580.

Your readonly/owner split is the right model, and it points at something larger: the same block on `main` already chowns `/opt/hermes/skills`, `/opt/hermes/plugins`, `/opt/defaults`, and both templates to `hermes`. I found nothing writing to those at runtime either, but I haven't traced it far enough to call it a bug. Filing that as a separate issue to apply your split across the whole block rather than widening this PR's scope.

bradhoekstra:
Closing as superseded by #580, which already delivered the read access these patched files needed (`--chmod=0644` plus the 755/644 normalisation loop). After rebasing onto it, all that remained here was the ownership change — and as discussed in [this thread](https://github.com/gke-labs/kube-agents/pull/567#discussion_r3736298005), that would make `/opt/hermes/gateway` and `/opt/hermes/tools` writable by the uid the agent runs as, for no runtime benefit. Follow-up issue to apply the readonly/owner split across the whole normalisation block to come.

PR #574, opened 2026-08-06
feat(cron): fire the Platform Agent's watchdogs from the Chat Agent's ticker

toshiowang:
# Pull Request

## Why This Change

The Platform Agent's cron jobs were not firing. Cron ticking is a property of a running **gateway**, and gateways are per profile — only the `default` (Chat Agent) profile has one. The Platform Agent is reached through the kanban dispatcher, which spawns a worker per card and exits, so a schedule sitting in `agents/platform/cron/jobs.json` had nothing to advance it.

The first version of this branch bridged the gap: it left the jobs where they were and added six `dispatch-<id>` trigger entries on the Chat Agent, each filing a card that asked the Platform Agent to call `cronjob(action='run', job_id='<id>')` against its own roster. That worked, but it described every job twice — a schedule on one profile and a prompt on the other, in two files that could disagree with each other and had no mechanical reason to stay in step.

This PR now does the simpler thing: **move the jobs**. Each keeps its id, name, schedule, prompt and skills, and lives in the one roster that ticks.

## What Changed

`agents/chat/defaults/cron/jobs.json` is the only cron roster. `agents/platform/cron/jobs.json` is `{"jobs": []}`.

Every entry is a `no_agent` script job — a plain subprocess, no LLM turn. The six governance entries run `platform_cron_dispatch.py`, which reads the prompt off the entry that scheduled it and copies it onto a kanban card assigned to `platform`. A Platform Agent worker does the work with the full platform toolset, which the Chat Agent profile deliberately lacks (`agents/chat/config.yaml` strips it to `mcp-router`, `kanban` and `memory`).

Editing `jobs.json` is now the whole of editing what a watchdog does.

**Files:**

- `agents/chat/defaults/cron/jobs.json` — the roster: 3 pre-existing script jobs + the 6 governance jobs
- `agents/platform/cron/jobs.json` — emptied, with a `_comment` block recording why an entry here would never fire
- `agents/chat/scripts/platform_cron_dispatch.py` — the script behind every governance job; its module docstring is the reference
- `agents/chat/scripts/dispatch_<id>.py` × 6 — one-line wrappers, one per job
- `agents/chat/scripts/test_platform_cron_dispatch.py` — 110 tests
- `agents/platform/AGENTS.md`, `agents/platform/skills/fleet-audit/SKILL.md`, `agents/platform/governance/*_sop.md` — the on-demand route
- `scripts/generate_docs.py`, `hack/check-docs-terminology.sh` — both read the chat roster
- `docs/` — 15 pages repointed; `concepts/autonomous-watchdogs.md` and `reference/cron-jobs.md` document the new job shape

**Added/Changed/Fixed:**

- **The prompt has one home.** It lives on the roster entry and is copied to the card verbatim. Nothing restates it.
- **`deliver: "all"` on all six.** `local` resolves to no delivery target, so the watchdog alert a non-zero exit raises was being dropped on the floor. The failure this PR guards against was itself silent.
- **The card force-loads the job's `skills`.** Each name is passed to `kanban create` as `--skill`, persisted on the task, and handed to the worker as `--skills` on spawn — the same preload the cron scheduler performed by prepending skill content to the prompt. The body names them as well, as the board's record of what the job expected.
- **A job with an empty prompt is an error, not an empty card.**
- **`cronjob(action='run')` is gone** from the Platform Agent's `AGENTS.md`, the `fleet-audit` skill and the five governance SOPs. It executes a job synchronously in the session that calls it, so on a cluster upgraded from an older image — where the merge's never-prune rule leaves the old entries in place — it would re-enact a frozen copy of the prompt outside the in-flight guard, concurrently with the card for the same audit. The on-demand route is `platform_cron_dispatch.py <job-id>` followed by `kanban_notify_propagate.py --to <task-id>`, so the card a person asked for is subscribed to their thread and can speak.
- **The six wrappers stay.** The scheduler runs a script with no arguments and sets no variable naming the job, so the wrapper is the only place the id can live. They now duplicate an id and nothing else.
- **Cards filed under the old `Run the <name> cron job [<id>]` title are still swept**, so the in-flight and retention guards do not lose track across this change.
- Carried forward from the earlier design and unchanged: stdout stays empty on every path, two independent dedup layers (a minute-bucketed idempotency key and an open-card check), finished-card retention at three per job, cards keyed on job id, and `created_by == "cron"` required before a card is claimed.

**From review** (each has its own thread):

- **A create the board refuses now alerts.** `hermes_cli.kanban.run_slash` never raises and discards the return code, so every refusal — argparse, `_cmd_create`'s guards, a locked database — arrived as an ordinary string. The tick logged that it had filed a card and exited 0. Refusal is now read off the return value, and alerts when the listing had just succeeded (the board is up and said no) rather than when it had not (weather).
- **The board's JSON is located, not assumed.** `run_slash` redirects `sys.stderr` into the buffer it returns, so one `DeprecationWarning` made `json.loads` fail identically on every future tick and both dedup layers fail open. A `raw_decode` scan finds the payload past noise that carries brackets of its own.
- **The roster fallback searches for the job, not for a readable file.** Taking the first readable path returned the stale live copy and never opened the image's, which is the one case the fallback exists for.
- **A job that blocks every run has a floor.** `blocked` is still not a lock, but past five blocked cards for one job the tick stops filing and exits non-zero, which turns a permanently-failing job into the page it should have been instead of 48 cards a day.

## Why This Matters

- The autonomous watchdogs — the feature the product is built around — did not run. This is what makes them run.
- One file, one schedule, one prompt. The bridge design had two rosters that had to agree; nothing enforced it, and `generate_docs.py`, the terminology guard and the `fleet-audit` name assertion each read only one of them.
- `deliver: "local"` meant a trigger could break and say nothing. Six jobs were one silent failure away from stopping without anyone noticing.

## Context

- The mechanism is documented in `concepts/autonomous-watchdogs.md` under "How a watchdog fires"; `architecture.mdx`, `proactive-autonomy.md` and `reference/cron-jobs.md` link to it rather than restating it.
- **This lands as a pair with #528.** `docker-entrypoint.sh` syncs `/opt/defaults` onto the PVC with `cp -ru`, which copies only when the *source* is newer; the scheduler rewrites `jobs.json` every tick, so the volume's copy is permanently newer and image changes to `cron/` never land. On its own, this roster therefore reaches fresh installations and no others. #528 adds entrypoint step 2c, which reconciles `/opt/defaults/cron/jobs.json` into the live store — image keys winning, scheduler run state preserved — and is what carries these nine jobs onto a cluster that already exists. Separately, `profile_scaffold.merge_cron_store` merges per key and never prunes, so emptying the platform roster does not empty a volume that already has it; that is addressed in the `agents/platform/AGENTS.md` guidance rather than by shipping the six disabled, and the reasoning is on the review thread.
- Once #528 is on `main`, one commit here rewrites `concepts/autonomous-watchdogs.md`'s "the shipped roster reaches a fresh volume and no other" — true today, false the moment that merges. Docs describe `main`, so it cannot be corrected ahead of the merge.
- **Follow-up, not in scope:** gate the `github-issue-resolver` job on a `resolver.py poll` so no card is filed when there is nothing to resolve; delete the five long-retired disabled entries from the platform roster; subscribe the home channel to cron cards so a scheduled run is not wholly silent.
- Several docs show more changed lines than they have edits — prettier realigns a Markdown table's columns when one cell grows.

## Testing

- `test_platform_cron_dispatch.py` — 110 tests. Coverage includes stdout silence on every path, both dedup layers, the retention sweep's ordering and exclusions, roster lookup and `enabled: false`, the empty-prompt guard, the dispatch-era card title, and a wiring test asserting the platform roster is empty, that every governance job is present with `no_agent`/`deliver: "all"`, and that no wrapper is orphaned and no job wrapper-less. The review fixes added: refusal strings rather than raised exceptions on `create`, an id quoted back inside an error not being read as a filed card, a noise table run against both the listing and the create, the wrong-shape scanner case, roster fallback by job id in four arrangements, the blocked-card cap including its per-job counting and its fail-open on an unreadable board, and that every skill a job names is one the worker can load.
- `test_audit_report.py` — 450 tests, including the `AUDITS` name assertion now made against the chat roster.
- `make docs-check` passes all four gates: generated regions, links, terminology, doc-map coverage.
- `npx prettier --check` passes on every changed Markdown, JSON and YAML file.
- Ran the `review-docs-drift` skill against the branch diff; it found three claims the move falsified (`overview/architecture.mdx`'s walk through `cronjob(action='run')`, `install/manual.md` telling a manual installer the schedules were in the workspace it had just copied, and `architecture/04-workflow-model.md`'s "10 governance jobs" listing five that were retired). All three are fixed here.
- Merged up to `main` with no conflicts; the full changed-file set is unchanged by the merge, and `make test-python`'s local failures reproduce identically on a clean `main` worktree (missing `yaml`/`fastapi`, `/opt/data` permissions).
- Deployed to a live GKE cluster and verified: the chat roster holds nine jobs with run state carried across the id change so nothing re-fires as never-run, the platform roster holds only the five disabled retirees, and the dispatcher renders each card from the live roster. The scheduler has ticked against it without error.

---

Rollout is a move, not an addition: the six jobs keep their ids and schedules, and each stays independently disableable via its own `enabled` flag. On an existing cluster the roster arrives with #528; see Context.


agents/platform/cron/jobs.json
```diff
@@ -1,82 +1,3 @@
...
-        "expr": "*/30 * * * *",
-        "display": "*/30 * * * *"
-      },
-      "prompt": "Run the github-issue-resolver skill to poll, triage, investigate, and resolve unaddressed open issues on our target repository strictly according to the skill's safety red lines, scope constraints, and turn completion checklist.",
-      "skills": ["github-issue-resolver"],
-      "enabled": true,
-      "deliver": "all"
-    }
-  ]
+  "jobs": []
```
bradhoekstra, line 28: A comment that any jobs listed here won't actually run might be useful in the future.

bradhoekstra (changes requested):

agents/platform/cron/jobs.json
toshiowang: Added

agents/chat/scripts/platform_cron_dispatch.py
```diff
@@ -0,0 +1,469 @@
...
+        f"create --json --assignee {shlex.quote(ASSIGNEE)} "
+        f"--created-by {shlex.quote(CREATED_BY)} "
+        f"--max-runtime {shlex.quote(MAX_RUNTIME.get(job_id, DEFAULT_MAX_RUNTIME))} "
+        f"--idempotency-key {shlex.quote(idempotency_key(job_id, now))} "
+        f"--body {shlex.quote(card_body(job))} "
+        f"{shlex.quote(title)}"
+    )
+    try:
+        out = _run_slash(cmd)
+    except Exception as e:  # noqa: BLE001 - the board decides whether this alerts
```
bradhoekstra, line 531: 🔴 **High — the "board refused the create → alert" path is dead code; a refused create exits 0 silently.**

`hermes_cli.kanban.run_slash` never raises and never surfaces an exit code. It wraps the command in `except SystemExit: pass` / `except Exception as exc: print(f"error: {exc}", file=sys.stderr)` and returns the captured text:

```python
# hermes_cli/kanban.py (run_slash)
with contextlib.redirect_stdout(buf_out), contextlib.redirect_stderr(buf_err):
    try:
        kanban_command(args)
    except SystemExit:
        pass
    except Exception as exc:
        print(f"error: {exc}", file=sys.stderr)
```

So every way the board can say no returns a *string*, not an exception:

- an argparse rejection → `"⚠ /kanban usage error\n…"`
- `_cmd_create` validation (`kanban: --max-runtime: …`, `--max-retries must be >= 1`) → return code `2`, which `run_slash` discards
- a DB error out of `kb.create_task` → `"error: database is locked"`

In all of those the `except Exception` here never fires. Control falls through to `_parse_task_id(out)`, which returns `None`, and the tick logs `filed <job-id> but could not read a task id from: …` and **returns `True` → exit 0**. No card is on the board, no watchdog alert is raised, and the log line asserts the card was filed when it was not — the exact silent failure this PR exists to end, now applying to all six audits at once.

`test_a_create_the_board_refuses_alerts` encodes the wrong model: it injects `create_raises = RuntimeError("unrecognized arguments: --max-runtime")`, and that message is precisely the case the real `run_slash` returns as a string rather than raising. The test passes and production does not.

Failure has to be read off the returned text — a successful `create --json` is a JSON object, and the non-JSON form is `Created <id>  (…)`. Anything else is a refusal and should alert when `reachable` is true. `bootstrap_scan_gate.file_scan_task` already takes the conservative branch here (unparseable id → treat as not filed, retry next tick).

agents/chat/scripts/platform_cron_dispatch.py
```diff
@@ -0,0 +1,469 @@
...
+    stopped because a listing error read as "already running" is worse.
+
+    `reachable` is what stops that leniency swallowing a real defect. It lets
+    the caller tell "the board is down" — transient, retry next tick, stay
+    quiet — from "the board is up and rejected our create", which is a bad
+    request that will fail identically forever.
+    """
+    try:
+        raw = _run_slash(f"list --json --assignee {shlex.quote(ASSIGNEE)}")
+        tasks = json.loads(raw) if raw else []
```
bradhoekstra: 🟠 **Medium — strict `json.loads` on the board listing makes both dedup layers fail open on any stderr noise.**

`run_slash` returns `f"{out}\n{err}"` whenever *either* buffer has content, and it redirects `sys.stderr` for the duration of the call. So a single `DeprecationWarning`, sqlite adapter warning, or `kanban:` diagnostic emitted while `list --json` runs corrupts the payload, `json.loads` raises, and `survey_board` returns `(False, [], False)`.

That is not a one-tick blip — the condition is deterministic, so **every** tick thereafter files anyway: concurrent audit runs writing the same ledger issue, `github-issue-resolver` firing twice per 30 minutes, and the retention sweep permanently disabled at 48 cards/day.

The sibling script in this same directory documents exactly this hazard, and `_parse_task_id` below already does the tolerant thing for `create`:

> ``--json`` is asked for, but the board's own stderr can share the buffer, so locate the JSON object rather than assuming the whole string is one. — `bootstrap_scan_gate._parse_task_id`

Suggest locating `[` … `]` here the same way `_parse_task_id` locates `{` … `}`.

agents/chat/scripts/platform_cron_dispatch.py
```diff
@@ -0,0 +1,469 @@
...
+    no identically on every future tick, so a quiet exit would retire the audit
+    permanently and leave one stderr line as the only trace.
+    """
+    job_name = str(job.get("name") or job_id)
+    title = card_title(job_id, job_name)
+    in_flight, stale, reachable = survey_board(job_id, job_name)
+    if in_flight:
+        return True
+
+    cmd = (
```
bradhoekstra, line 518: 🟠 **Medium — the entry's `skills` are no longer force-loaded into the worker, and `kanban create --skill` would restore exact parity.**

On the old cron path the scheduler expanded `job["skills"]` through `skill_view` and prepended the skill content to the prompt (`cron/scheduler.py:_build_job_prompt`). The card body now only names them in prose, so nothing loads `fleet-audit` — the worker has to decide to go and find it.

This is one flag away from parity, not a constraint of the kanban path:

```python
# hermes_cli/kanban.py — create parser
p_create.add_argument("--skill", action="append", default=[], dest="skills", ...)

# hermes_cli/kanban_db.py — worker spawn argv
if task.skills:
    for sk in task.skills:
        if sk:
            cmd.extend(["--skills", sk])
```

So `--skill fleet-audit` on the create would be persisted on the card and passed to the worker as `--skills fleet-audit`, which is the same force-load the cron entry had.

The module docstring's "A kanban worker is not a cron run, so … does not inherit the entry's `skills`" reads as unavoidable but is a choice, and `concepts/autonomous-watchdogs.md:84` goes further — "the card body names them instead, **and the worker loads them**" — which nothing guarantees.

Concrete failure: a worker that reads the prompt but never invokes `fleet-audit` runs the audit without `audit_report.py open`/`finish`, so no ledger issue is opened or closed and the findings are lost, while the card still completes "successfully". That is the 2026-08-03 failure mode the SKILL.md rewrite cites, minus the concurrency.

bradhoekstra (commented):
## Review — correctness pass over the diff

Read every hunk plus the surrounding runtime this change depends on: `cron/scheduler.py` (`_run_job_script`, the `no_agent` short-circuit, `_deliver_result`), `hermes_cli/kanban.py` (`run_slash`, `_cmd_create`, `_cmd_list`, `_cmd_archive`), `hermes_cli/kanban_db.py` (`create_task`, the worker spawn argv), `deploy/shared/docker-entrypoint.sh`, and `agents/platform/scripts/profile_scaffold.py`.

Verified green on my side: the 55 new tests pass, `scripts/generate_docs.py --check` passes, `hack/check-docs-terminology.sh` passes, `deliver: "all"` does resolve through `_expand_routing_tokens` (so the `local`-drops-the-alert reasoning is right), the `create` flags used all exist, `created_at` really is an epoch int so `_created_at`'s coercion is sound, `finished[:-KEEP_FINISHED]` keeps the newest three as intended, `"Archived <id>"` is the real confirmation line, and `kanban_home()` collapses profiles onto one board so `--assignee platform` from the chat profile lands where the dispatcher looks.

Eight findings below — one High inline, one High that has no line in this diff, and the rest Medium/Low. The two High ones both undercut the PR's stated goal, so they're worth resolving before merge; the rest are smaller.

🔴 **High — the `default` profile's cron store is never merged, so the six jobs never land on an existing cluster and `enabled: false` stops working.** (`deploy/shared/docker-entrypoint.sh:151`)

The module docstring discloses half of this, but the consequence is bigger than "an upgraded cluster needs a hand".

`merge_cron_store` / `overlay_template` only run for scaffolded **named** profiles. The entrypoint calls them for `platform` (step 2.6, `--items "config.yaml SOUL.md AGENTS.md CAPABILITIES.md cron skills governance"`) and for the cluster profiles below it. The `default` profile is not one of those — it lives at `$TARGET_DIR` itself and receives only:

```sh
# step 2
cp -ru /opt/defaults/. "$TARGET_DIR/"
# step 2a — force-sync list
for f in config.yaml SOUL.md AGENTS.md CAPABILITIES.md; do ...
# step 2b — force-sync
cp -rf /opt/defaults/scripts/. "$TARGET_DIR/scripts/"
```

`cron/` is in neither force-sync list, and the scheduler rewrites `last_run_at` into the volume's `jobs.json` on every tick, so the destination is permanently newer and `cp -u` skips it forever.

Two consequences, and only the first is documented:

1. On every existing install the six governance jobs never appear in `/opt/data/cron/jobs.json`, so the fix this PR ships no-ops on upgrade.
2. **The documented kill switch is gone.** While these jobs lived on the platform roster, `enabled: false` in the image took effect on a live cluster via the per-key merge. On the chat roster nothing merges, so flipping the flag disables nothing until the volume is recreated. `concepts/autonomous-watchdogs.md` still promises the old behaviour (inline comment on that file).

Adding `cron/jobs.json` to the default profile's sync through the same `merge_cron_store` path would close both, and would also make the rollout story here "redeploy" rather than "redeploy and then hand-edit two files on the PVC".

---

*The move itself is clearly the right call, and the reasoning in the docstrings is unusually good — most of what follows is about the seam between the new location and the deploy machinery that was built around the old one.*

Reviewed with the help of Claude.

agents/chat/scripts/platform_cron_dispatch.py
```diff
@@ -0,0 +1,469 @@
...
+# second one would run the same audit concurrently with itself — two runs
+# writing the same ledger issue, or two `github-issue-resolver` passes racing
+# for the same issue.
+#
+# `blocked` is deliberately absent. A card blocks when its worker tripped the
+# failure breaker or escalated for input, and it then stays blocked until a
+# human clears it. Counting that as in-flight would let one bad run switch the
+# audit off indefinitely, silently, which is the failure mode this whole change
+# exists to end. A blocked card is a thing to go and look at, not a lock.
+IN_FLIGHT = frozenset({"triage", "todo", "ready", "scheduled", "running", "review"})
```
bradhoekstra, line 132: 🟡 **Low — `blocked` is excluded from both `IN_FLIGHT` and `FINISHED`, so a persistently-blocking job accumulates cards without bound and re-spawns a worker every tick.**

Each half is well argued on its own (blocked must not be a lock; blocked must never be archived), but together they have no floor. If `github-issue-resolver` hits a condition that trips the failure breaker, every 30-minute tick files a new card, the dispatcher spawns a worker with a *fresh* `consecutive_failures` counter, it fails, and it blocks. That is 48 blocked cards a day plus 48× the breaker's retry budget in worker spawns, none of them swept, indefinitely — and the run never exits non-zero, so nothing alerts.

A cap would keep the "blocked is not a lock" property without the runaway: stop filing once N blocked cards exist for this job and exit non-zero saying so, which turns the condition into the page it should be.

agents/chat/scripts/platform_cron_dispatch.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+def log(msg: str) -> None:
+    """Write to stderr — stdout is the delivery channel and must stay empty."""
+    sys.stderr.write(f"platform_cron_dispatch: {msg}\n")
+
+
+def load_roster(paths=ROSTER_PATHS) -> dict:
+    """Return {job_id: job} from the first readable copy of this cron roster."""
+    for path in paths:
```
bradhoekstra, line 216: 🟡 **Low — `load_roster` takes the first *readable* path, not the first path that has the job, which defeats the `/opt/defaults` fallback in exactly the case it exists for.**

On an upgraded cluster `$HERMES_HOME/cron/jobs.json` exists but is the stale pre-upgrade copy (see the `cp -ru` finding in the review summary), so the loop returns it and `/opt/defaults/cron/jobs.json` — which *does* carry the entry — is never consulted.

The on-demand route the new `agents/platform/AGENTS.md:24` and `fleet-audit/SKILL.md:56` tell the agent to use is `platform_cron_dispatch.py <job-id>`. It will exit 1 with `'compliance-audit' is not in the cron roster` on precisely the clusters where a human is trying to work around the missing schedule.

Falling through to the next path when the id is absent (rather than when the file is unreadable) would fix it, and costs nothing on a healthy pod.

docs/site/src/content/docs/concepts/autonomous-watchdogs.md
```diff
@@ -56,21 +71,24 @@ Each job in `jobs.json` follows this schema:
...
-- **`deliver`** (optional) — controls chat delivery. `"all"` means every run reports back. It is set on all six enabled jobs, which is safe because each returns `[SILENT]` when it has nothing to say.
+- **`deliver`** — where a tick's stdout goes. A successful tick prints nothing and is delivered as a silent run, so this only matters on failure: `"all"` sends the watchdog alert to the configured target, while `"local"` resolves to no target at all and would drop it. All six governance jobs use `"all"`, so a bridge that stops filing cards is visible rather than indistinguishable from a quiet fleet.
 
 ## Disabling a watchdog
 
-Edit `cron/jobs.json`, flip `enabled` to `false`, and redeploy the workspace (`provision_08_deploy_platform_agent.sh` or `dev/dev_rebuild_agent.sh`). The change is picked up on the next agent restart.
+Edit `agents/chat/defaults/cron/jobs.json`, flip `enabled` to `false`, and redeploy the workspace (`provision_08_deploy_platform_agent.sh` or `dev/dev_rebuild_agent.sh`). The change is picked up on the next agent restart. One flag, one file: the scheduler stops ticking the job, so nothing is filed and nothing runs.
```
bradhoekstra: 🟠 **Medium — "Disabling a watchdog" now documents a mechanism that does not apply to the moved roster.**

The paragraph two lines below this one says:

> `cron/jobs.json` is image-owned configuration and live scheduler state in the same file, so start-up merges the two rather than replacing one with the other (`profile_scaffold.py`). The image wins every key it ships — which is what makes `enabled: false` take effect…

That was true of `agents/platform/cron/jobs.json`: the entrypoint runs `profile_scaffold.py --name platform --items "… cron …"`, which routes through `merge_cron_store`. The `default` profile is **not** a scaffolded named profile — it lives at `$TARGET_DIR` itself and gets only step 2's `cp -ru /opt/defaults/. "$TARGET_DIR/"`. There is no merge, and `cron/` is in neither force-sync list (2a is `config.yaml SOUL.md AGENTS.md CAPABILITIES.md`, 2b is `scripts/`).

Since the scheduler rewrites `last_run_at` into the volume's copy every tick, the destination is permanently newer and `cp -u` skips it forever. So an operator who follows this page to switch off a noisy audit on a running cluster ships `enabled: false`, redeploys, and the job keeps firing with no indication why. The kill switch this page documents does not exist for these six jobs any more.

Either wire the default profile's cron store through `merge_cron_store` too (which would also fix the jobs never landing on upgrade), or rewrite this section to say the flag only takes effect on a fresh volume.

agents/platform/cron/jobs.json
bradhoekstra, line 28: 🟠 **Medium — emptying this roster violates the retirement protocol this repo documents, so the six entries stay live and *enabled* on every existing cluster.**

From `agents/platform/scripts/profile_scaffold.py:merge_cron_store` — the code that actually syncs this file:

> That last rule is also the limit: nothing here can tell an operator's own job from one this release deleted, so *removing* an entry from the shipped roster does not stop it firing on a cluster that already has it — it only ends the image's ability to hold it off. Retire a watchdog by shipping `enabled: false` and leaving the entry in place; an id is safe to drop from the roster only once every live cluster has merged that disabled form… **The five unrunnable watchdogs were retired that way before being deleted here.**

This PR deletes all six outright instead. On an upgraded cluster `profiles/platform/cron/jobs.json` therefore still holds six `enabled: true` entries that this image can no longer switch off — and the PR body confirms the live cluster already carries leftovers of exactly this kind ("the platform roster holds only the five disabled retirees").

Suggest shipping the six as `enabled: false` here for one release, then deleting the ids in a follow-up.

agents/platform/AGENTS.md
```diff
@@ -18,13 +18,15 @@ You wake up fresh each session. Maintain continuity through:
...
-- **More than one job in one request ("run all the fleet audits") → one child card per job.** A card produces exactly one chat message, when it completes. Dispatch five jobs from one card and the user gets one message for all five: four reports are invisible and the fifth is whatever fitted. So do not dispatch them here. Resolve the list with `cronjob(action='list')`, then for **each** job:
+- **A governance job arrives as a card.** Every cron job lives in the Chat Agent profile's roster (`/opt/data/cron/jobs.json`), because that profile owns the only ticking gateway — yours has none. Its scheduler files one card per due job, carrying that job's own prompt, and you are the worker. Do exactly what the card says, then `kanban_complete`. No person is waiting on a scheduled card: it carries no chat subscription, so your summary posts nowhere and is written for the board's record and for whoever comes looking when a schedule appears to have stopped.
+- **"Run the `<x>` cron job now" → file its card, do not re-enact it.** `cronjob` is not the route: your profile's roster is empty and `cronjob(action='run')` has nothing to run. Instead, for **each** job the request names:
```
bradhoekstra: 🟠 **Medium — "your profile's roster is empty and `cronjob(action='run')` has nothing to run" is false on every existing cluster.**

`merge_cron_store` adds and overwrites but never prunes, so deleting the six entries from `agents/platform/cron/jobs.json` leaves them in place — and enabled — on any volume that already merged them. The Platform Agent reading this line will be told its roster is empty while `cronjob(action='list')` shows six jobs.

The sharper edge is the interaction: if anyone does call `cronjob(action='run', job_id='compliance-audit')` on such a cluster, the old entry *will* fire, with the old prompt, and `platform_cron_dispatch.py`'s in-flight guard cannot see it — so it can run concurrently with a chat-side card for the same audit, which is the double-ledger-write this PR's `IN_FLIGHT` set exists to prevent.

This resolves itself if the six are shipped `enabled: false` rather than deleted (see the comment on `agents/platform/cron/jobs.json`).

agents/chat/scripts/platform_cron_dispatch.py
toshiowang: Fixed in 558bde9. You were right on both halves, and the second half — that the test encoded the wrong model — is the part that had kept it invisible: `create_raises = RuntimeError("unrecognized arguments: --max-runtime")` asserts a raise for the one case `run_slash` is guaranteed to return as a string.

Refusal is now read off the return value. `_parse_task_id(out)` returning `None` is the refusal signal, not an exception:

```python
task_id = _parse_task_id(out)
if not task_id:
    log(f"the board refused the create for {job_id}: {out.strip()[:300]}")
    return not reachable
```

Three things about that, in case any of them is the wrong call:

- **`not reachable`, not a flat `False`.** `survey_board` now reports whether the listing itself answered. A board that is down refuses the create too, and alerting on that turns every transient outage into a page for six jobs at once. Reachable means the listing succeeded seconds earlier, so the board is up and talking and it still said no — which it will say identically on every future tick.
- **The log carries what the board said**, clipped to 300 characters. The old line (`could not read a task id from: …`) sent the last reader looking for a card that had never been created. `test_the_refusal_is_logged_with_what_the_board_said` pins it.
- **No `t_[0-9a-f]+` fallback.** It is the tempting reading of "locate the id rather than parse the payload", and it is wrong: `error: duplicate of t_abc12345` would match and the tick would report someone else's card as the one it filed. `test_an_id_quoted_back_inside_an_error_is_not_a_filed_card` fails on that regex.

`_parse_task_id` locates the JSON object rather than accepting the whole of stdout, since `--json` is always passed and `_cmd_create` prints either one `json.dumps` of the task or nothing — so there is no "the card is probably there, only its id came back unreadable" state to be lenient about. See the sibling comment on `survey_board` for why locating rather than slicing.

The `except Exception` branch stays. It is now unreachable through `run_slash` as it stands today, but it reaches the same verdict as the parse path, so a future `run_slash` that does propagate is handled rather than crashing the tick.

Tests: `test_a_create_the_board_refuses_alerts` now iterates the three real refusal *strings* — an argparse rejection, `kanban: --max-runtime: malformed duration 'zzz'`, `error: database is locked` — and asserts `rc == 1` with empty stdout. Alongside it, `test_output_with_no_task_object_is_read_as_a_refusal`, `test_a_refusal_on_an_unreachable_board_stays_quiet`, and `test_a_refused_create_still_prunes` (the stale-card sweep is in a `finally` and does not depend on today's create landing).

agents/chat/scripts/platform_cron_dispatch.py
toshiowang: Fixed in 0206734, and the fix went further than locating `[` … `]` because that would not have held.

The noise brings brackets of its own. A `[ROUTER-MCP] cannot list profiles for …` line in front of the payload defeats a first-opener-to-last-closer slice exactly as thoroughly as `json.loads` defeats the whole string — it just fails on a different input. So both call sites now share a scanner:

```python
def _find_json(out: str, opener: str, kind: type):
    decoder = json.JSONDecoder()
    idx = out.find(opener)
    while idx != -1:
        try:
            value = decoder.raw_decode(out, idx)[0]
        except ValueError:
            value = None
        if isinstance(value, kind):
            return value
        idx = out.find(opener, idx + 1)
    return None
```

`raw_decode` from each candidate offset, stepping over anything that does not parse or parses as the wrong shape. `survey_board` asks for a `list`, `_parse_task_id` for a `dict`, and the `kind` check is what stops a stray `{…}` in a warning being read as the created task.

Two related changes fell out of it:

- **A listing that holds no array now says so.** It still fails open — a board nobody can read must not retire the audits — but it logs `could not parse the board listing` with the payload. Silence was the actual defect here: the finding's "48 cards a day, permanently" only stays invisible because nothing recorded why. `test_a_listing_that_holds_no_array_fails_open_and_says_so`.
- **An empty listing is an empty board**, not a parse failure. `--json` on a board with no matching cards can return nothing at all, and logging a parse error on the ordinary first-tick case would train whoever reads the log to ignore the line. `test_an_empty_listing_is_an_empty_board_not_a_parse_failure`.

Tests: `test_noise_around_the_listing_does_not_blind_the_dedup` runs the dedup against a table of real prefixes/suffixes/wrappers as subtests, `test_noise_around_the_created_task_is_not_read_as_a_refusal` covers the same hazard on `create` (where strictness costs more — the card exists, so the alert is false *and* the next tick sees it in flight and never corrects the record), and `test_the_scanner_steps_over_candidates_that_are_the_wrong_shape` pins the scanner directly.

agents/chat/scripts/platform_cron_dispatch.py
toshiowang: Fixed in 5747378 — you found the flag and it does restore exact parity, so the docstring that called this an unavoidable property of the kanban path was wrong, not just pessimistic.

The create now carries one `--skill` per entry:

```python
# Before `--body`, because the title is positional and has to stay last.
skill_flags = "".join(f"--skill {shlex.quote(s)} " for s in job_skills(job))
```

which `kanban_db` turns into `--skills <name>` on the worker spawn — the same force-load the scheduler used to perform by prepending skill content to the prompt. The card body still names them, but as the board's record of what the job expected rather than as the loading mechanism.

The two prose claims are corrected in the same commit rather than left to be true by accident:

- the module docstring now names the mechanism (`kanban create --skill` → persisted on the task → `--skills` on the worker) instead of saying skills are not inherited;
- `concepts/autonomous-watchdogs.md`'s "and the worker loads them" is now a guarantee with the chain written out, since it was the sentence with nothing behind it.

Tests: `test_the_create_force_loads_the_jobs_skills`, `test_a_job_with_no_skills_passes_no_skill_flag` (no stray empty flag), and `test_the_flags_and_the_body_name_the_same_skills` — the last one because the body and the flags are two renderings of `job_skills(job)` and a future edit to one of them should not be able to drift from the other. `test_every_skill_a_job_names_is_one_the_worker_can_load` checks each name against the shipped skill directories, so a typo in a roster entry fails here rather than as a worker that silently loads nothing.

The PR body still carries the old "a kanban worker is not a cron run" bullet this commit falsified; I am fixing that in the description now.

agents/chat/scripts/platform_cron_dispatch.py
toshiowang: Fixed in 08aecd4, and taken as filed: a floor, not a reclassification of `blocked` as in-flight.

`survey_board` now also counts this job's blocked cards, and `file_card` stops past the cap and exits non-zero:

```python
MAX_BLOCKED = 5
...
if blocked >= MAX_BLOCKED:
    log(f"{blocked} blocked card(s) for {job_id} (cap {MAX_BLOCKED}) — not "
        f"filing another; clear or archive them to resume the schedule")
    archive_cards(stale)
    return False
```

Five rather than one or two, because the property in the comment above `IN_FLIGHT` has to survive: one bad run must not be able to switch an audit off. A genuinely stuck job accumulates two or three before somebody looks; by the fifth, nobody is being deprived of a run that would have worked, and the condition has earned the page.

Two details worth checking:

- **It fails open.** An unreadable listing reports zero blocked, so a board that will not answer cannot trip the cap and retire the audits — the same reasoning as `in_flight` and `stale`. `test_an_unreadable_board_does_not_look_like_a_blocked_backlog`.
- **The sweep still runs.** `archive_cards(stale)` happens before the return, so a job that has hit the cap keeps pruning its own finished-card backlog. Blocked cards are still never archived. `test_a_blocked_backlog_is_still_swept_of_finished_cards`, `test_blocked_cards_are_never_archived`.

Also `test_a_backlog_under_the_cap_still_files`, `test_a_backlog_at_the_cap_stops_filing_and_alerts`, and `test_the_cap_counts_only_this_jobs_blocked_cards` — the count is per job, since one wedged audit should not hold the other five off the board.

agents/chat/scripts/platform_cron_dispatch.py
toshiowang: Fixed in 0206734. `load_roster` takes the job id and returns the first roster that *has* it:

```python
first = None
for path in paths:
    ...
    if job_id is None or job_id in roster:
        return roster
    if first is None:
        first = roster
return first if first is not None else {}
```

The `first` bookkeeping is there for the error path you named. Falling straight through on a miss would make `'compliance-audit' is not in the cron roster` report against `/opt/defaults`, which is not the roster the operator's cluster is running — so the message would describe a file the operator cannot edit and would not explain what they are looking at. It now names the live one.

Tests: `test_a_stale_first_roster_does_not_hide_the_job_in_a_later_one` is the regression; `test_the_stale_roster_still_wins_when_it_has_the_job` keeps the live copy authoritative when it does carry the entry (an operator's local edit to a prompt must not be silently overridden by the image's); `test_a_job_in_no_roster_reports_the_one_the_cluster_runs`; `test_a_job_found_only_in_the_fallback_is_dispatchable`; `test_first_readable_roster_wins`.

Worth saying where the underlying gap goes: the `cp -ru` skip this rests on is fixed in #528, whose entrypoint step 2c reconciles `/opt/defaults/cron/jobs.json` onto the live volume on every start. The two pull requests are a pair — without #528 this roster reaches fresh volumes only, and this fallback is the difference between a hand-run working and not working on every cluster that already exists. The fallback stays after #528 lands regardless: it costs one file read on a healthy pod, and it covers the window before that image is deployed anywhere.

docs/site/src/content/docs/concepts/autonomous-watchdogs.md
toshiowang: Rewritten in 0206734, with a correction in 3192b2f. You offered two options and this branch takes the second; the first is being done, but in a different pull request and by a different mechanism.

**Why not `merge_cron_store` here.** That function is reached through `profile_scaffold.py --name <profile>`, which scaffolds a *named* profile under `profiles/<name>/`. The `default` profile is not one — it is `$HERMES_HOME` itself, and nothing scaffolds it. Wiring it in would have meant teaching the scaffolder about a profile that has no profile directory. #528 does the same job from the entrypoint instead: step 2c runs `cron_jobs_sync.py`, which merges `/opt/defaults/cron/jobs.json` into the live store, image keys winning, live scheduler state (`last_run_at`, the run ledger) preserved. So the kill switch this page documents will exist for these six jobs — one image away, not never.

**What the section says now.** The whole "On an existing cluster the edit does not travel by itself" paragraph is new, and it names the specific reason rather than gesturing at it: no merge on this path, `cron/` in neither force-sync list (2a is the four persona files, 2b is `scripts/`), the scheduler's per-tick `last_run_at` write making the destination permanently newer so `cp -u` skips it for good. Then the two consequences that cut in opposite directions — the same gap is what protects a hand-added entry from being pruned on upgrade — and the actual procedure: edit `$HERMES_HOME/cron/jobs.json` on the PVC and restart the gateway.

Two things the section gained that the finding did not ask for but the same reader needs:

- `platform_cron_dispatch.py` honours `enabled` a second time (`main`, before it files anything), so a hand-run of a `dispatch_*.py` wrapper cannot resurrect an audit somebody switched off. The flag is worth something on the live roster even where the image cannot deliver it.
- The `cronjob` tool is explicitly *not* a route to this: it is denied to the Chat Agent in `agents/chat/config.yaml`, and the Platform Agent's copy addresses that profile's store, which has no ticker.

**One thing I owe this thread.** The sentence "the shipped roster reaches a **fresh** volume and no other" is true of `main` today and false the moment #528 merges. Since docs here describe `main`, it cannot be fixed ahead of time — I will land the rewrite as a commit on this branch after #528 goes in and before this one merges, rather than leaving the page to go stale on the merge that fixes it.

agents/platform/cron/jobs.json
toshiowang: This one is a deliberate deviation rather than a fix, so let me put the reasoning on the record instead of quietly leaving it.

The mechanism you describe is correct and I am not disputing any of it: `merge_cron_store` adds and overwrites but never prunes, so an upgraded volume keeps all six entries, `enabled: true`, and this image can no longer switch them off.

Where I part company is on what that costs here. The protocol's own sentence — the one quoted in the finding — says removing an entry "only ends the image's ability to hold it off." That is a real loss on a roster whose entries fire. `profiles/platform/cron/jobs.json` is not one. The cron ticker lives in the gateway, and the `default` profile owns the only gateway in the deployment; the Platform Agent is reached through the kanban dispatcher, which spawns a worker per card and exits. The evidence is on the volume itself: that directory has `jobs.json`, `executions.db` and `.jobs.lock`, but none of the ticker's own files — no `.tick.lock`, no `ticker_heartbeat`, no `ticker_last_success` — which exist only under the Chat Agent's `HERMES_HOME/cron/`. Nothing has ever ticked it, which is how six governance watchdogs went a release without firing and is the entire reason this pull request exists.

So shipping `enabled: false` for a release would write a flag into a store that no scheduler reads, to hold off jobs that no scheduler runs, and would then need a follow-up nobody is holding to delete the ids. It also puts the image back to asserting a six-job schedule for a profile that has no clock — the confusion the empty roster and its `_comment` block are there to end.

**The residual is real and I have not pretended otherwise.** A leftover entry is still reachable by hand through `cronjob(action='run', job_id=…)`, which is your sharper point on the sibling `AGENTS.md` thread, and I could not verify from this repository whether that path honours `enabled` — the `cronjob` tool is upstream Hermes and is not vendored here — so I cannot claim the suggested remedy would have closed it either. What I did instead was close it where it is verifiable, in the instructions the agent actually reads: `AGENTS.md` now says that `cronjob(action='list')` may still show the governance jobs on an upgraded cluster, that they are leftovers with prompts frozen at the release that shipped them, and that the route is `platform_cron_dispatch.py <job-id>` — which files a card, so the in-flight guard applies and the concurrent double-ledger-write cannot happen.

Written down in 97e50c0 (the `_comment` block in the emptied roster: why an entry here is inert, that emptying this file does not empty an existing cluster's copy, and why that is harmless) and 3192b2f (`AGENTS.md`, plus the retired-jobs section, which had been naming the wrong file for the leftovers). If you would still rather have the disabled-for-one-release form, say so and I will ship it — it is a small change and yours is the call, I would just rather it be a decision than a default.

agents/platform/AGENTS.md
toshiowang: Fixed in 3192b2f. The line said something false to the one reader who cannot check it.

It now reads:

> **"Run the `<x>` cron job now" → file its card, do not re-enact it.** `cronjob` is not the route. The image ships your profile an empty roster, but a cluster upgraded from an older one keeps what it was given before — the start-up merge adds and overwrites, it never prunes — so `cronjob(action='list')` may still show the governance jobs. Those are leftovers: nothing ticks them (your profile has no gateway) and their prompts are frozen at the release that shipped them. `cronjob(action='run')` executes a job synchronously in the session that calls it, which is the re-enactment this bullet exists to prevent.

Three changes in that, each answering a different part of the finding: the agent is told to expect a non-empty `list`, told what those entries are, and given the reason not to `run` one that does not depend on the roster being empty. The old sentence made the whole instruction contingent on a fact that is false on every existing cluster — an agent that ran `cronjob(action='list')`, saw six jobs, and concluded the guidance was stale would have been reasoning correctly from what it was given.

The concurrency edge you name is why "nothing to run" was the wrong argument to lean on: it collapses the moment the roster is not empty. What holds instead is that the two paths produce different things — `cronjob(action='run')` executes in the calling session with a frozen prompt and no card, so `platform_cron_dispatch.py`'s in-flight guard cannot see it, while step 1 of this bullet files a card that the guard can. That is now stated as the reason rather than implied by an absence.

The same commit fixed the retired-jobs section, which named `cron/jobs.json` for the leftovers when they sit on `profiles/platform/cron/jobs.json`, and covered only the five retired in #559 rather than those plus the six that just moved.

See the sibling thread on `agents/platform/cron/jobs.json` for why the six were deleted outright rather than shipped disabled — that is a deliberate deviation from the retirement protocol, and this bullet is a load-bearing part of it.

docs/site/src/content/docs/concepts/autonomous-watchdogs.md
toshiowang: Addendum, because checking #528 against this page turned up a second thing the follow-up commit has to fix, and it is not the one I expected.

#528's reconcile treats `enabled` as **runtime-owned**: `RUNTIME_OWNED_FIELDS = ("last_run", "enabled", "deliver", "origin")`, on the reasoning that an operator who disables a job on a live pod must not have an image roll switch it back on. And a job the image stops declaring is not pruned either — "retirement does not propagate."

So after #528 the image gains the power to deliver a job's *definition* to a live cluster and still has no kill switch for it: shipping `enabled: false` is ignored, and deleting the entry leaves it running. Which makes the instruction at the top of this section — "Flip `enabled` to `false` in `agents/chat/defaults/cron/jobs.json`", and the rule under it that an id may be dropped "only once no live cluster still needs the image to hold the job off" — resting on a capability that will not exist. Today it is saved only by the `cp -ru` gap and by the paragraph below telling the operator to edit the PVC copy; once the definitions start travelling and the flag does not, "flip the flag" reads as the kill switch and is not one.

The two conventions are also in direct opposition as sentences on `main`: this page says "flip the flag, do not delete the entry", `cron_jobs_sync.py`'s docstring says "to retire a job, remove it from the image file rather than disabling it there". Both are defensible for the roster each was written about; they cannot both be the instruction for this one.

I have flagged it on #528 as the one thing there for a reviewer to decide rather than picking a side unilaterally, since it changes what that PR's contract is *for* and not just how this page words it. Either resolution is a small edit here:

- keep `enabled` runtime-owned and rewrite this section so the flag is documented as a PVC-side operation with the repo entry as the declaration of record; or
- make `enabled` image-owned for jobs the image declares, which restores the kill switch and makes this section true as written, at the cost of an image roll being able to re-enable a job somebody switched off by hand.

Either way it lands in the same commit as the "reaches a fresh volume and no other" rewrite, after #528 and before this merges.

toshiowang: Resolved, and the follow-up this page needs is now smaller than the addendum above promised.

#528's reviewer landed on the same problem from the other side — its `merge_job` kept a closed allowlist of "runtime-owned" field names, which silently deleted every scheduler field not on the list. The fix was to drop the list and adopt `profile_scaffold.merge_cron_store`'s rule instead: **the image wins every key it ships, and every key it does not ship is left as the volume had it.** `enabled` is shipped, so it is image-owned.

That is the second bullet of the two I offered, arrived at for an independent reason, and it means this page's [Disabling a watchdog](#disabling-a-watchdog) section is now true as written for **both** rosters rather than only the platform one — the sentence already there, "the image wins every key it ships, which is what makes `enabled: false` take effect", is literally the rule the Chat Agent's reconcile now uses. So does "deleting an entry does not stop it firing on a cluster that already has it", and so does the two-step retirement path the five retired watchdogs took. Nothing in that section changes.

What remains for the follow-up commit here is only the delivery half: with #528 merged, a job added to this roster reaches an *existing* volume and not just a fresh one, so any sentence on this page implying otherwise needs rewriting. That is a smaller edit than the contract question I was holding it for.

bradhoekstra (approved):

adamparco:
This moved platform agent tasks to the "chat" agent. This seems wrong. The chat agent is not supposed to have any tools or permissions.

PR #578, opened 2026-08-07
feat(scripts): support an opt-in Cloud Build private worker pool

adamparco:
# Pull Request

## Why This Change

Cloud Build runs every build on the project's default pool (`e2-standard-2`, 2 vCPU) unless the build explicitly names a private pool. Creating a pool does not route builds to it — the pool has to be passed per submit.

Image builds dominate both the dev rebuild loop and the Prow deploy. Recent default-pool builds in a personal project measured 2m49s–6m01s for `platform-agent` and 3m51s–10m25s for `credential-proxy`, roughly 12–20 min for a full three-image deploy. A private pool with more CPU is a straightforward win for anyone who has one, but there was no way to select it short of editing shared config.

## What Changed

**Files:**

- `hack/ci-deploy.sh`
- `k8s-operator/scripts/dev/dev_rebuild_agent.sh`
- `docs/site/src/content/docs/operator/development.md`

**Added/Changed/Fixed:**

- Added `CLOUD_BUILD_WORKER_POOL`, an opt-in env var taking a full pool resource name (`projects/PROJECT/locations/REGION/workerPools/POOL`).
- Both scripts pass `--worker-pool` to every `gcloud builds submit` when it is set — five call sites total, including the operator's `--tag` build, which has no config file and can only take the flag.
- The pool's region is parsed out of the resource name rather than taken as a second variable. `gcloud builds submit` otherwise falls back to the `global` region, which cannot reach a regional pool — the failure mode anyone setting only the pool would have hit.
- A malformed value exits 1 with a usage message instead of silently building `--region=`.
- Documented under "Fast agent iteration (dev only)".

## Why This Matters

- **Default is unchanged.** When `CLOUD_BUILD_WORKER_POOL` is unset the scripts emit byte-identical `gcloud` commands, so contributors and CI projects without a pool are unaffected.
- **Nothing is pinned.** No project, region, or pool name is committed anywhere.
- `deploy/docker/cloudbuild.yaml` is deliberately left alone — an `options.pool` entry there would pin every consumer of the shared config to a single project.

## Context

- `.github/workflows/docker-publish-gcp.yml` is intentionally out of scope. It is gated on `github.repository == 'gke-labs/kube-agents'` and runs in `kube-agents-prow`; wiring a pool there is cross-project and needs a repo variable plus a `WorkerPool User` grant. Happy to follow up if maintainers want it.
- Empty-array expansion uses the `${arr[@]+"${arr[@]}"}` form rather than a plain splat: `ci-deploy.sh` runs `set -euo pipefail`, and expanding an empty array under `set -u` is an error in bash 3.2, which is what macOS ships.

## Testing

- `bash -n` clean on both scripts.
- Behavior verified under `set -euo pipefail` on bash 3.2 across all three paths: unset → flags absent; set → `--worker-pool=... --region=us-east4`; malformed → exit 1.
- Verified end to end against a real private pool (`c3-standard-22`, us-east4) with a throwaway `--no-source` build: SUCCESS, 22 vCPU / 86 GB / 1.1 TB visible to the build, public egress working.
- `make docs-check` passes.
- `npx prettier --check` passes on the changed Markdown.
- `cd docs/site && npm run build` passes (43 pages).
- The `review-docs-drift` skill procedure was run against this diff: no generated-region sources touched, no doc added/moved/renamed/deleted so the map needs no update, and the other pages referencing these scripts describe them at a level this additive change does not invalidate.

Not yet exercised through a full `make dev-rebuild-agent` or a Prow run — the flag composition is verified, but a maintainer may want one real deploy on a pool before relying on it.

---

Low risk and fully inert unless the new variable is set. This PR was generated with the help of Claude.

### DIFF
--- a/docs/site/src/content/docs/operator/development.md
+++ b/docs/site/src/content/docs/operator/development.md
@@ -65,6 +65,16 @@ make dev-rebuild-agent ARGS="platform"
 
 This builds the Platform Agent image, pushes to Artifact Registry, and restarts the Deployment. First run creates a dev Artifact Registry repo; clean it up later with `make gcp-teardown-dev-artifact-registry`.
 
+### Building on a private worker pool
+
+Cloud Build runs on the project's default pool (2 vCPU) unless you point it elsewhere. To use a [private pool](https://cloud.google.com/build/docs/private-pools/private-pools-overview) with more CPU, export its full resource name:
+
+```bash
+export CLOUD_BUILD_WORKER_POOL=projects/PROJECT/locations/REGION/workerPools/POOL
+```
+
+`dev_rebuild_agent.sh` and `hack/ci-deploy.sh` both read this variable and pass `--worker-pool` (plus the pool's region, parsed from the name) to every `gcloud builds submit`. Leave it unset to keep the default pool. The pool must allow public egress, or builds fail pulling base images and packages.
+
 ## Integrations (Kustomize)
 
 Integrations have dedicated deploy/undeploy targets:
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -43,6 +43,27 @@ export USER_PROFILE_ENABLED="false"
 export GOOGLE_CHAT_ENABLED="false"
 export SLACK_ENABLED="false"
 
+# Optional Cloud Build private worker pool. Unset by default, so builds keep
+# going to the project's default pool. Opt in by exporting a full resource
+# name: projects/PROJECT/locations/REGION/workerPools/POOL
+# The region is read back out of that name because `gcloud builds submit`
+# otherwise falls back to the `global` region, which cannot reach a regional
+# pool.
+BUILD_POOL_ARGS=()
+if [ -n "${CLOUD_BUILD_WORKER_POOL:-}" ]; then
+  case "$CLOUD_BUILD_WORKER_POOL" in
+    projects/*/locations/*/workerPools/*) ;;
+    *)
+      echo "ERROR: CLOUD_BUILD_WORKER_POOL must be a full resource name: projects/PROJECT/locations/REGION/workerPools/POOL"
+      exit 1
+      ;;
+  esac
+  BUILD_POOL_ARGS=(
+    --worker-pool="$CLOUD_BUILD_WORKER_POOL"
+    --region="$(echo "$CLOUD_BUILD_WORKER_POOL" | cut -d'/' -f4)"
+  )
+fi
+
 START_TIME=$SECONDS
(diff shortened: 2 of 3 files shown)


adamparco:
### Exercised through a full `make dev-rebuild-agent`

The PR description flags that the flag composition was verified but never run through a real deploy. It has been now — one end-to-end `dev-rebuild-agent` against a live GKE dev cluster, both build targets, on a private regional pool.

```
export CLOUD_BUILD_WORKER_POOL="projects/adamparco-kage/locations/us-east4/workerPools/more-workers"
make dev-rebuild-agent ARGS="platform"
```

Both builds landed on the pool rather than the shared default:

| Build | Target | `options.pool.name` | Status | Build time |
| `5381e61c-3c67-4181-80c4-11fe294278be` | `platform` | `projects/adamparco-kage/locations/us-east4/workerPools/more-workers` | SUCCESS | 2m31s |
| `aa080ac2-3005-4a4f-a628-bac9aafc934b` | `credential-proxy` | `projects/adamparco-kage/locations/us-east4/workerPools/more-workers` | SUCCESS | 3m12s |

The negative control — an earlier run of the same target with `CLOUD_BUILD_WORKER_POOL` unset — reports an empty `options.pool.name` (`8dbb8ec2-d192-4678-8fe9-7390ad577ebe`), so the field is genuinely coming from the new flag and not from a project default.

The rest of the script's work completed normally on top of that: both images pushed to Artifact Registry at `dev-20260806-235141`, the `PlatformAgent` CR patched to that tag, and the rolling update finished clean.

```
$ kubectl -n kubeagents-system get platformagents.kubeagents.x-k8s.io platform-agent \
    -o jsonpath='{.spec.deployment.tag}'
dev-20260806-235141

$ kubectl -n kubeagents-system rollout status deploy/platform-agent-gateway
deployment "platform-agent-gateway" successfully rolled out

$ kubectl -n kubeagents-system get pods -l app.kubernetes.io/name=platform-agent
platform-agent-gateway-75b5f6ddf6-7dkd7   5/5 Ready   0 restarts   Running
```

I also spot-checked the resulting image rather than trusting the tag alone — the in-image source patches the Dockerfile applies are present and behave correctly in the running pod, so a pool-built image is byte-for-byte as functional as a default-pool one.

One note for anyone adopting this: `--region` is load-bearing. A regional worker pool is unreachable from the `global` build region, and `gcloud builds submit` defaults to `global`, so omitting the region fails with a pool-not-found rather than silently falling back. This PR already passes them together; worth keeping that way.

Not covered here: a Prow run. This was a local `make dev-rebuild-agent` only.

_Testing and this comment were produced with the help of Claude._

bradhoekstra (approved):

PR #581, opened 2026-08-07
feat(governance): implement fleet stockout prevention and capacity au…

### LINKED ISSUE #547: (feat) cron-based stockout prevention
Following up on the work in #517 and also using skills from #541, a cron-based stockout prevention mechanism can be built. In the first iteration, for each cluster:

* go through each workload, gathering resource limits and requests, along with machine types
* verify node pools if autopilot is not used
* verify existing CCC configurations using CCC skills
* verify historical spot capacity to get a grasp on resource obtainability for those configurations
* emit PRs with suggested improvements 

This way, Hermes can proactively reduce the likelyhood of stockouts if the configuration can be made more resilient.

sergiu1spiridon:
# Pull Request

## Why This Change

Following up on the reactive stockout investigator (#517) and GKE ComputeClass architecture skills (#541), this PR introduces a proactive, cron-based **Fleet Stockout Prevention & Capacity Audit** stream (`stockout-prevention`).

Previously, capacity stockouts and fragile compute class configurations were only diagnosed *reactively* after a workload had already failed to scale. This audit sweeps every managed GKE cluster and GCP region daily to detect single-zone affinity pins, missing fallback machine families, Spot-only priority lists lacking On-Demand floors, and regional quota exhaustion risks before they cause scheduling outages.

## What Changed

**Files:**

- `agents/platform/governance/stockout_prevention_sop.md`: Canonical SOP playbook defining the 10 proactive capacity and compute class resilience checks (§3.1–§3.10), data sources (`kubectl`, `gcloud compute advice capacity`, quota APIs), and findings schema.
- `agents/platform/cron/jobs.json`: Registered the `stockout-prevention` watchdog (schedule `20 9 * * *`, daily 09:20 UTC) and associated prompt.
- `agents/platform/skills/fleet-audit/SKILL.md`: Added `stockout-prevention` to the list of active fleet audit streams.
- `agents/platform/skills/fleet-audit/scripts/audit_report.py`: Registered the `stockout-prevention` stream definition and check slugs into the `AUDITS` catalog and validation schemas.
- `agents/platform/skills/fleet-audit/scripts/test_audit_report.py`: Added comprehensive unit test coverage for the `stockout-prevention` stream, check schema validations, and false-positive secret scanner mitigations.
- `agents/platform/scripts/platform_mcp_server.py`: Fixed `send_notification` target formatting to pass `google_chat:${GOOGLE_CHAT_HOME_CHANNEL}` directly to `hermes send`.
- `agents/platform/config.yaml`: Explicitly injected `GOOGLE_CHAT_HOME_CHANNEL`, `GOOGLE_CLOUD_PROJECT`, and `GCP_PROJECT_ID` into the `platform_control` MCP server environment.
- `agentplugins/pubsub-platform/install.sh` & `adapter.py`: Added fallback support for `GCP_PROJECT_ID` and `GKE_PROJECT_ID`.
- `agentplugins/gke-stockout-investigator/install.sh`: Added project ID flag propagation during builds.
- `docs/site/src/content/docs/reference/cron-jobs.md`: Updated generated cron schedules table.
- `docs/README.md`: Updated documentation map inventory to include the 12th governance SOP.

**Added/Changed/Fixed:**

- **Added 10 Proactive Capacity Checks:**
  1. `ccc-missing-fallbacks`: Detects ComputeClasses pinned to a single machine family or zone without fallback priorities.
  2. `ccc-no-ondemand-floor`: Flags Spot-only priority lists that lack an On-Demand floor.
  3. `ccc-granular-type-bloat`: Identifies configurations with >10 granular machine types triggering autoscaler backoff starvation loops.
  4. `spot-scarcity-risk`: Evaluates high-preemption Spot configurations using GCP capacity advice APIs.
  5. `accelerator-type-pin`: Flags hard GPU/TPU accelerator pins lacking multi-zone/multi-family fallbacks.
  6. `large-shape-unobtainable`: Flags workloads requesting >32 vCPUs without smaller-replica horizontal scaling or family fallbacks.
  7. `unsupported-hyperdisk-fallback`: Identifies Hyperdisk-attached workloads falling back to incompatible machine generations.
  8. `regional-quota-saturation`: Identifies workloads exceeding regional CPU/GPU quota limits.
  9. `single-zone-nodepool`: Flags standard GKE clusters with single-zone node pools.
  10. `rigid-scheduling-pin`: Detects critical workloads with hard nodeSelector / nodeAffinity to a single zone.
- **Fixed Chat Notification Delivery:** Ensured MCP child processes inherit `GOOGLE_CHAT_HOME_CHANNEL` and properly dispatch Google Chat start and remediation alerts.

## Why This Matters

- **Proactive Reliability:** Prevents production scheduling outages by catching fragile infrastructure configurations before demand spikes or zonal hardware shortages occur.
- **Unified Governance:** Integrates capacity governance into the existing `fleet-audit` single-living-issue ledger and automated remediation PR workflow.
- **Zero Configuration Drift:** Reuses the verified ComputeClass rules and GitOps leasing infrastructure established in #517 and #541.

## Context

- Follows up on #517 (Reactive GKE Stockout Investigator).
- Leverages domain rules from #541 (GKE ComputeClass Architecture & Optimization Skill).
- Implements #547 (Proactive Fleet Stockout Prevention Audit).

Closes #547 

## Testing

1. **Unit Tests:**
   - Ran `python3 -m unittest agents/platform/skills/fleet-audit/scripts/test_audit_report.py` — all 449 tests passed.
2. **Docs & Link Validation:**
   - Ran `make docs-check` — all 185 Markdown files, terminology, generated regions, and doc map inventory verified.
3. **End-to-End Reactive Verification:**
   - Ran `./agentplugins/gke-stockout-investigator/scenarios/04-missing-zone-fallback.sh` against live cluster `platform-agent-host`.
   - Verified that `send_notification` dispatched start and completion messages to Google Chat space `spaces/wuMQxKAAAAE`.
4. **End-to-End Proactive Audit Verification:**
   - Deployed faulty test workload and verified execution of `audit_report.py start` and `audit_report.py finish`.
   - Verified automated publication and updating of the GitHub Issue Ledger ([Issue #5](https://github.com/test-org-kube-agent/agents-repo/issues/5)).


agents/platform/scripts/platform_mcp_server.py
```diff
@@ -521,7 +521,15 @@ def get_active_platform() -> str:
...
-    target = active_platform # default fallback
+    home_channel = (
+        os.environ.get("GOOGLE_CHAT_HOME_CHANNEL")
+        or os.environ.get("SLACK_HOME_CHANNEL")
+        or ""
+    ).strip()
+    if home_channel:
+        target = f"{active_platform}:{home_channel}"
+    else:
+        target = active_platform
```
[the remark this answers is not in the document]
tlipski: +1

agents/platform/governance/stockout_prevention_sop.md
```diff
@@ -0,0 +1,215 @@
...
+
+### 2. Collect capacity and workload state
+
+Collect the live cluster definitions and GCP regional capacity metrics:
+
+```bash
+# 1. Dump ComputeClasses, Workloads, and StorageClasses
+KUBECONFIG=$KC kubectl get customcomputeclasses,computeclasses,deployments,statefulsets,storageclasses,nodepools -A -o json > /opt/data/scratch/stockout_state_<cluster>.json
+
+# 2. Inspect GCP Regional Quotas for the cluster's region (e.g. us-central1)
```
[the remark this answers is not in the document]
tlipski: yes, it that the names are incorrect:
```tomeklipski-mac:kube-agents tomeklipski$ kubectl get customcomputeclasses,computeclasses,deployments,statefulsets,storageclasses,nodepools -A -o json
error: the server doesn't have a resource type "customcomputeclasses"
tomeklipski-mac:kube-agents tomeklipski$ kubectl get computeclasses,deployments,statefulsets,storageclasses,nodepools -A -o json
error: the server doesn't have a resource type "nodepools"
tomeklipski-mac:kube-agents tomeklipski$ kubectl get computeclasses,deployments,statefulsets,storageclasses -A -o json|wc -l
    4991
```

agentplugins/gke-stockout-investigator/install.sh
```diff
@@ -155,7 +155,7 @@ IMAGE="${PLUGIN_IMAGE:-gcr.io/${PROJECT_ID}/gke-stockout-investigator:latest}"
 if [ -n "${PLUGIN_IMAGE:-}" ]; then
     echo "Using the prebuilt image ${IMAGE} (PLUGIN_IMAGE set); skipping the build."
 else
-    gcloud builds submit --tag "$IMAGE" "$SCRIPT_DIR"
+    gcloud builds submit --project="$PROJECT_ID" --tag "$IMAGE" "$SCRIPT_DIR"
```
tlipski, line 158: good catch!

agents/chat/defaults/cron/jobs.json
```diff
@@ -119,6 +119,21 @@
       "enabled": true,
       "deliver": "all"
     },
+    {
+      "id": "stockout-prevention",
+      "name": "Fleet Stockout Prevention & Capacity Audit",
+      "schedule": {
+        "kind": "cron",
+        "expr": "20 9 * * *",
```
tlipski, line 127: that's a good cadence, it allows the operator to review the outcome on the morning. 

nitpick: One note is that it might be prudent to skill weekends? but it's configuration, so not a blocker

agents/platform/config.yaml
```diff
@@ -13,6 +13,9 @@ mcp_servers:
       HERMES_HOME: "${HERMES_HOME}"
       GOOGLE_CHAT_PROJECT_ID: "${GOOGLE_CHAT_PROJECT_ID}"
       GOOGLE_CHAT_SUBSCRIPTION_NAME: "${GOOGLE_CHAT_SUBSCRIPTION_NAME}"
+      GOOGLE_CHAT_HOME_CHANNEL: "${GOOGLE_CHAT_HOME_CHANNEL}"
```
tlipski: yeah, that's a thing needed for stockout investigator as well to make it work with notifications.

agents/platform/governance/stockout_prevention_sop.md
```diff
@@ -0,0 +1,215 @@
+# SOP: Fleet Stockout Prevention & Capacity Audit (Daily Governance)
```
tlipski, line 1: One thing that I wonder about is how will this be tested automatically? Is it gonna be under devops-bench?

agents/platform/governance/stockout_prevention_sop.md
```diff
@@ -0,0 +1,215 @@
...
+
+```bash
+# 1. Dump ComputeClasses, Workloads, and StorageClasses
+KUBECONFIG=$KC kubectl get customcomputeclasses,computeclasses,deployments,statefulsets,storageclasses,nodepools -A -o json > /opt/data/scratch/stockout_state_<cluster>.json
+
+# 2. Inspect GCP Regional Quotas for the cluster's region (e.g. us-central1)
+gcloud compute regions describe <region> --format="json(quotas.filter(metric=CPUS),quotas.filter(metric=NVIDIA_L4_GPUS),quotas.filter(metric=NVIDIA_A100_GPUS))"
+
+# 3. Check Spot Capacity & Preemption Advice
+gcloud beta compute advice capacity --provisioning-model=SPOT --instance-selection-machine-types="g2-standard-4,n4-standard-4,c3-standard-4" --target-distribution-shape=ANY --size=1 --region=<region> --format="json"
```
tlipski: I asked coding agent to review what other skills/tools can be used and why, please have a look (or feed it to your coding agent ;)

# Review: `stockout_prevention_sop.md` (PR #581) vs. the stockout investigator skill

Comparing `agents/platform/governance/stockout_prevention_sop.md` (PR #581) against
`agentplugins/gke-stockout-investigator/files/skills/gke-stockout-investigator/SKILL.md`.

The SOP is a **prevention** audit, not an incident responder — the investigator skill already owns
the reactive path. That difference is the lens for everything below: a check earns its place here
only if it does one of three things.

- **P1 — Finds a config that will fail at the next capacity event.** The classic preventive finding.
- **P2 — Proves a proposed remediation will actually work.** A fix that cannot provision is not
  prevention; it is a closed ledger entry over a still-broken cluster.
- **P3 — Suppresses a false positive.** A daily fleet audit that flags every single-family
  ComputeClass trains operators to ignore it, and the real finding dies with the noise.

Each item below is tagged with which one it serves.

## Coverage map

The SOP is a clean superset of the skill's rule set on the _static config_ axis:

| Skill rule                                                            | SOP check                                                          |
| --------------------------------------------------------------------- | ------------------------------------------------------------------ |
| A — zone/family fallbacks                                             | 3.1 `ccc-missing-fallbacks`                                        |
| B — large VM shapes >32 vCPU                                          | 3.3 `ccc-large-vm-scarcity`                                        |
| C — stateful disk-generation mix                                      | 3.5 `ccc-mixed-disk-generations`                                   |
| D — missing on-demand floor                                           | 3.2 `ccc-no-ondemand-floor`                                        |
| **E — region-wide scarcity of specialized hardware → migrate region** | **absent**                                                         |
| F — regional quota exceeded                                           | 3.7 `quota-exhaustion-risk`                                        |
| G — priority starvation                                               | 3.4 `ccc-priority-starvation`                                      |
| H — Hyperdisk incompatibility                                         | 3.6 `ccc-hyperdisk-incompatible`                                   |
| —                                                                     | 3.9 `single-zone-nodepool`, 3.10 `rigid-scheduling-pin` (SOP-only) |

Only Rule E is dropped. The real divergence is in **tools**.

## Tools the skill runs that the SOP never does

### 1. `gcloud compute reservations list` (skill §4B) — **P1 + P2**

The SOP has no reservation awareness at all, and this is the most consequential omission.

**Prevents:** a reservation is the _only_ capacity guarantee GCP sells — it is literally the
strongest stockout prevention available, and the SOP never checks whether the fleet uses one. Worse,
the repo's own reference documents that `reservations.affinity: AnyBestEffort`/`Automatic`
**silently bypasses ComputeClass priorities** and falls back to On-Demand at the GCE layer
(`references/compute-class-gotchas-and-cuds.md:7`, `compute-class-debug.md:100`). On such a class,
every fallback 3.1/3.2/3.8 recommends is dead code — the audit would open a PR, close the finding,
and leave the cluster exactly as exposed as before (**P2**). Three checks fall out:

- `AnyBestEffort`/`Automatic` affinity on any ComputeClass → the whole fallback chain is inert.
- `Specific` affinity naming a reservation that does not exist, or is in a zone the priority cannot
  reach → that priority can never fire.
- A reservation with `inUseCount` far below `count` while a workload in the same region is running
  unreserved → guaranteed capacity is sitting idle next to something exposed to a stockout.

Note the reference's distinction from CUDs (`compute-class-gotchas-and-cuds.md:37`): CUDs are
consumed **automatically** by family and require no ComputeClass config, whereas reservations must
be explicitly targeted. So a CUD is not a capacity guarantee and must not be counted as one — worth
saying in the SOP, because "we're covered, we have a commitment" is the standard misconception.

### 2. `gcloud beta compute advice capacity-history --types=PREEMPTION,PRICE` (skill §4D.2) — **P1**

**Prevents:** obtainability is a property of the region and shape _right now_, not of the YAML. A
ComputeClass that reads as perfectly diversified on paper can be pinned to three families that are
all scarce this month. This is the only data source in either document that describes the world
rather than the config, and it is what makes "fragile" a measured claim instead of a heuristic.

The SOP names the command under "Data sources" but never issues it — and 3.8 flags on ">20%
historical preemption rates", a number only that command produces. As written, **3.8 cannot be
evidenced**, which under the SOP's own red line ("a finding you cannot reproduce is dropped") means
3.8 never ships a finding. Separately, 3.8's `Command:` line drops
`--instance-selection-machine-types` and `--size`, which Step 2's version of the same command
carries.

### 3. `kubectl top node` / `top pod` / metrics API (skill §4C) — **P3**

**Prevents a bad PR, not a stockout.** 3.3 proposes shrinking >32-vCPU shapes with no evidence the
workload needs less. The prioritization reference is explicit that this remediation is conditional:
node auto-creation sizes nodes to Pod _requests_, so a single pod requesting >32 vCPU **cannot** land
on a smaller node, and only horizontally-scalable workloads benefit — for a genuinely large single
pod you vary zone/family instead, never cores
(`references/compute-class-prioritization.md:36`). Without utilization data the SOP cannot tell those
two cases apart, and half its 3.3 PRs will propose a shape the workload cannot fit in.
`roles/monitoring.viewer` is already granted, so this is free.

### 4. Live pending-pod state (skill §2B) — **P3**

**Prevents:** the SOP reads only declarations, so it cannot distinguish a ComputeClass that is
fragile from one that is fragile _and already failing_ — nor from one whose pods are Pending for a
reason that has nothing to do with capacity (a missing GPU toleration, an unschedulable resource
request). No `kubectl get pods -A --field-selector=status.phase=Pending`, no
`FailedScheduling`/`NotTriggerScaleUp` events, no `cluster-autoscaler-status` ConfigMap. This is
also the input that lets severity mean something — see the next section.

## Checks worth adding, ranked

### 1. Autoscaler visibility logs — the biggest single win

**Prevents (P1 + P3), and this is the part worth arguing.** Reading logs sounds like detection, not
prevention. It is prevention here for two reasons:

- **They are a leading indicator, not an incident record.** `no.scale.up.*` and MIG-cooldown entries
  fire when the autoscaler tries a priority, fails, and succeeds on a fallback. Nobody is paged, no
  pod is Pending, and the operator has no idea it happened — but the class just spent its safety
  margin. That is precisely the pre-outage state a prevention audit exists to catch, and it is
  invisible to a config read.
- **They are the only way to rank the findings.** A fleet audit that flags every single-family
  ComputeClass emits hundreds of `critical`s that are all theoretically true and mostly never going
  to fire. One that can say "these three already hit `out.of.resources` twice last week" gives the
  operator a work queue instead of a wall. Without this input, the SOP's severities are assigned by
  rule, not by risk.

`roles/logging.viewer` is already granted, and `compute-class-debug.md:32` already tabulates the
messageIds. It also surfaces **`scale.up.error.ip.space.exhausted`** — subnet/pod-CIDR exhaustion,
which presents identically to a stockout and has zero coverage in either document (**P3**: without
it, the audit will diagnose a full subnet as a capacity problem and propose fallback families that
cannot help).

The SOP must specify _how_ to read these logs, not just that it should. Measured on
`tomeklipski-izrhgv` (14 RUNNING clusters, mostly idle):

#### Volume of the naive read

`gcloud logging read 'log_id("container.googleapis.com/cluster-autoscaler-visibility")'` over 24h,
fleet-wide:

| entry kind         | count (6.6h sample) | median size         | share     |
| ------------------ | ------------------- | ------------------- | --------- |
| `status`           | 1,771               | 523 B               | **88.6%** |
| `noDecisionStatus` | 177                 | 894 B (max 4.7 KB)  | 8.9%      |
| `decision`         | 26                  | 1.0 KB (max 3.5 KB) | 1.3%      |
| `resultInfo`       | 26                  | 539 B               | 1.3%      |

2,000 entries = 1.78 MB in 6.6 hours → **~7,300 entries / ~6.4 MB per 24h**, roughly 1.7M tokens.
Nearly nine-tenths is the `status` heartbeat, which carries `autoscaledNodesCount` and nothing
diagnostic. Another ~350 B of every entry is envelope (`insertId`, `logName`, `receiveTimestamp`,
`resource`).

#### Filtering collapses it

| query                                        | entries/24h | bytes      | ≈tokens   |
| -------------------------------------------- | ----------- | ---------- | --------- |
| bare `log_id(...)`                           | ~7,300      | 6.4 MB     | ~1.7M     |
| `+ jsonPayload.noDecisionStatus.noScaleUp:*` | 293         | 755 KB     | ~190k     |
| same, `--format="value(cluster,messageId)"`  | 293         | **4.4 KB** | **~1.1k** |

The real signal is smaller still. Fleet-wide over **7 days**, on a project where stockout scenarios
were deliberately run:

```
scale.up.error.out.of.resources     : 83
scale.up.error.quota.exceeded       : 36
scale.up.error.ip.space.exhausted   :  0
```

Those 83 are not 83 problems — the autoscaler re-emits the same rejection every pass, so they dedup
by `(cluster, nodepool, messageId, podGroup.controller)` down to a handful of distinct workloads.

#### The trap: entry size scales with the failure

The largest entry observed was 4.7 KB — one pod group against three rejected MIGs. The payload is
`unhandledPodGroups[] × rejectedMigs[]`, so an entry is
**O(pending pod groups × node pools × zones)**, and it re-emits every autoscaler pass (~20s) for as
long as pods stay Pending. A cluster with 8 stuck pod groups and 20 MIGs produces ~100 KB per entry;
an hour of that is ~18 MB from a single cluster. Volume peaks exactly when the check has something to
say, so an unbounded read is fine on a healthy fleet and catastrophic on a sick one.

#### Prescribe a two-stage read

**Stage 1 — triage, projected and cheap.** Per cluster, filter server-side on the stockout
messageIds and project with `--format="value(...)"`, never `--format=json`. Costs ~1 KB per cluster:

```bash
gcloud logging read \
  'log_id("container.googleapis.com/cluster-autoscaler-visibility")
   AND resource.labels.cluster_name="<cluster>"
   AND (jsonPayload.noDecisionStatus.noScaleUp:* OR jsonPayload.resultInfo.results.errorMsg:*)' \
  --project=<project> --freshness=24h --limit=1000 \
  --format="value(timestamp,resource.labels.cluster_name,jsonPayload.noDecisionStatus.noScaleUp.unhandledPodGroups[0].napFailureReasons[0].messageId)"
```

**Stage 2 — full JSON only for clusters that hit**, with an explicit `--limit`:

```bash
gcloud logging read \
  'log_id("container.googleapis.com/cluster-autoscaler-visibility")
   AND resource.labels.cluster_name="<cluster>"
   AND jsonPayload.noDecisionStatus.noScaleUp:*' \
  --project=<project> --freshness=24h --limit=50 --format=json
```

Two constraints worth writing into the SOP text:

- **Filter on messageId server-side.** Cloud Logging returns nothing for entries the filter excludes,
  so the agent never pays for them. A client-side `grep` over `--format=json` pays the full 6.4 MB.
- **A `--limit` truncation is a `coverage_gap`.** The SOP's own "no silent caps" red line applies: a
  capped log read is indistinguishable from a clean one, so it must set `partial: true` rather than
  be read as "no stockouts in the last 24h".

Note that the existing reference command at `compute-class-debug.md:151` uses
`--freshness=1h --limit=50`, which is right for interactive debugging and wrong as a fleet audit's
basis.

### 2. Remediation feasibility gate before Step 4 writes a manifest — **P2**

**Prevents the audit from shipping a fix that cannot provision.** Every one of the SOP's
`kind: manifest` remediations adds machine families or Spot tiers to a ComputeClass, and none of them
checks that the added capacity is reachable. A merged PR then closes the ledger entry — so the
finding stops being reported while the cluster stays exposed. That is worse than not having run the
check. Gate on:

- **Per-family quota.** `N4_CPUS`, `C4_CPUS` and friends are commonly **0** on a project that has
  never used the family. "Add an `n4` fallback" against zero quota is a no-op.
- **Preemptible quota.** `PREEMPTIBLE_CPUS` and preemptible GPU metrics are separate from on-demand,
  so the Spot tiers that 3.2 and 3.8 add can fail on quota independently.
- **Zonal offering.** The family or accelerator must actually be sold in the cluster's zones —
  `gcloud compute machine-types list --zones=…`, `gcloud compute accelerator-types list`. A fallback
  to a family the zone does not offer is not a fallback.

3.7's `gcloud compute regions describe` returns only a handful of aggregate metrics and won't catch
any of these.

### 3. Obtainability comes from varying **dimensions**, not counting families — **P1**

This is the deepest gap, and it is stated plainly in the repo's own reference
(`references/compute-class-prioritization.md:30`, "Key Rules"):

> **No repetition:** Doesn't improve obtainability. **Vary dimensions:** Zone, Family, Capacity
> (Spot/OD), **machine size (cores)**.

3.1 passes any ComputeClass with "2+ machine families". But a class with `c3-standard-88` and
`n4-standard-80`, both on-demand, both in one zone, varies exactly **one** of the four dimensions and
is nearly as stockout-prone as a single-family class — while sailing through 3.1 and 3.3 both.

**Suggested check (`ccc-low-dimension-diversity`):** score each ComputeClass on how many of the four
dimensions its `priorities[]` actually vary; flag anything below two. This subsumes 3.1 and is
strictly more accurate. Reference: `compute-class-prioritization.md` "Key Rules" — no need to restate
the rationale in the SOP, just cite it.

Two corollaries from the same section that the SOP contradicts or misses:

- **Spot is not a universal fallback.** "For CPU, if OD is out, Spot usually is too. For
  Accelerators, Spot often has capacity when OD doesn't"
  (`compute-class-prioritization.md:56`). So 3.8's remediation — expand Spot instance selection — is
  sound for GPU classes and close to useless for CPU ones. The SOP should make 3.2/3.8 accelerator-aware
  rather than applying one rule to both.
- **The documented fallback patterns are per-workload-type**, not universal:
  Inference `Res → OD → DWS → Spot`, Prod Training `Res → DWS → OD → Spot`, Dev `Spot → OD`
  (`compute-class-prioritization.md:22`, table). 3.2 checks only "is there an OD floor", which passes
  a latency-sensitive inference class that is Spot-first — the exact anti-pattern the skill calls out
  as a `CRITICAL AI/ML RULE`.

### 4. ComputeClass fields the SOP never reads — **P1**

All from the CRD reference (`references/compute-class-crd-fields.md`); cite it rather than restating.

- **`status.conditions`** — the cluster has already evaluated the class and reports invalid config
  (e.g. _"location config with specific reservations enabled"_, unsupported `sysctls`). **Prevents:**
  a rejected ComputeClass provides no fallback at all, and this is a zero-false-positive finding for
  one free read.
- **`whenUnsatisfiable`** (default `DoNotScaleUp`, `crd-fields.md:104`) — **Prevents:** `DoNotScaleUp`
  plus a narrow priority list is a guaranteed `Pending` at the first shortage; `ScaleUpAnyway`
  silently lands **E2** nodes, which for a Hyperdisk or GPU workload means the scale-up "succeeds"
  into nodes the workload cannot use. Both are stockout outcomes the config read can see coming.
- **`nodePoolAutoCreation.enabled`** (default **`false`**) — **Prevents:** priorities with no matching
  existing pool and no auto-creation can never be satisfied, which makes 3.1's entire remediation a
  no-op on those classes (**P2** as well).
- **`location.locationPolicy`** — `ANY` is the **default** and "packs for utilization, tends to fill
  one zone"; `BALANCED` spreads nodes across zones (`crd-fields.md:71`). **Prevents:** a class that
  lists three zones but leaves the default policy concentrates its nodes in one of them, so a single
  zonal shortage takes the whole class down — a multi-zone class that is single-zone in practice.
  3.1 inspects the zone list and would call this healthy.
- **`priorityScore` validity** — "If one rule has a score, **all** must. Max 3 rules per score"
  (`compute-class-prioritization.md:9`). **Prevents:** a partially-scored priority list is invalid,
  so the fallback chain does not traverse as written.
- **Manual-pool binding** — a `priorities[].nodepools` entry only fires if the pool carries both the
  `cloud.google.com/compute-class` **label and taint**
  (`references/compute-class-provisioning-methods.md:64`). **Prevents:** a dangling or unlabelled pool
  reference is a priority that silently never fires — the fallback exists in YAML and not in reality.
- **ARM/x86 mixing** — allowed, but "**Must use multi-platform image builds**"
  (`compute-class-prioritization.md:53`). **Prevents:** an `n4a` fallback under a single-arch image is
  a fallback that provisions a node the pod cannot run on.
- **Control-plane version gating** — many fields need 1.33+/1.34+, and unsupported keys are
  ignored or surface in `status.conditions` (`compute-class-gotchas-and-cuds.md:55`,
  `compute-class-debug.md:3`). **Prevents:** a fallback written against a newer schema than the
  cluster runs is not enforced at all.
- **3.4's threshold is wrong against the repo's own reference.** `compute-class-prioritization.md:5`
  and `compute-class-debug.md:127` both put the cooldown/backoff loop at **>10 priorities of any
  kind**; 3.4 counts only granular `machineType` rules, so a family-compressed list of 12 still loops
  and won't be flagged.

### 5. 3.5 and 3.6's remediations increase stockout risk on modern clusters — **P2**

The SOP's fix for both disk checks is to **narrow** `priorities[]` to one machine generation. On a
prevention audit that is backwards: narrowing the fallback list is the thing 3.1 exists to prevent.
The reference gives the correct modern fix
(`references/compute-class-gotchas-and-cuds.md:18`, `compute-class-provisioning-methods.md:38`): on
GKE **1.35.3-gke.1290000+**, back the data PVs with the built-in **`dynamic-rwo`** StorageClass
(`type: dynamic`, `use-allowed-disk-topology: "true"`), which makes the autoscaler disk-topology-aware
so it skips incompatible-generation priorities instead of attach-failing — letting the class **keep**
a broad cross-generation fallback list.

So 3.5/3.6 should branch on cluster version, and the remediation should prefer the StorageClass
change over trimming priorities. One caveat the SOP must carry: `dynamic-rwo` governs only **newly
provisioned** PVs; an existing fixed PD or Hyperdisk does not convert in place
(`compute-class-gotchas-and-cuds.md:23`), so for existing stateful workloads the finding is
`kind: manual`, not `kind: manifest`.

### 6. GPU/TPU specifics — **P1 + P3**

- **Missing `nvidia.com/gpu` toleration** (`compute-class-debug.md:43`). **P3:** GKE auto-taints GPU
  nodes, so a pod without the toleration sits `Pending` with `noScaleUp` logs and looks exactly like
  a stockout. Without this check the audit will misdiagnose it and propose capacity fallbacks that
  change nothing.
- **No DWS / `flexStart` check anywhere.** `priorities[].flexStart.enabled` is a real CRD field
  (`crd-fields.md:77`) and DWS is the documented answer to accelerator scarcity, sitting in both
  recommended AI/ML fallback chains. **Prevents:** a GPU class with no reservation and no flex-start
  in a scarce region is the canonical stockout setup, and no SOP check looks for it. Cite
  `compute-class-prioritization.md` "Fallback Patterns" rather than restating the chains. Carry one
  caveat from `compute-class-gotchas-and-cuds.md:31`: DWS is queued (~3 min) and
  `maxRunDurationSeconds` does not improve obtainability.
- **TPU gang-scheduling** — `priorities[].tpu` carries count/type/**topology** (`crd-fields.md:78`).
  **Prevents:** TPU slices are all-or-nothing, which makes them the most stockout-prone thing in GKE;
  3.7 mentions TPU quota while no check ever reads a TPU workload.
- **Kueue-managed batch** (`compute-class-provisioning-methods.md:87`). **Prevents:** a
  `ResourceFlavor` whose `nodeLabels` name a ComputeClass that does not exist means admitted jobs get
  a `nodeSelector` nothing can satisfy — permanent Pending that reads as a stockout.
- **Rule E belongs here too** — a fleet audit is exactly where "100% of your L4 demand is in one
  region" should be said, and it is the one skill rule the SOP dropped entirely.

### 7. Free signals from data 3.9 already fetches — **P1 + P3**

The node-pools JSON that 3.9 pulls also contains:

- **`autoscaling.maxNodeCount`** — **P3:** a pool sitting at its ceiling produces symptoms identical
  to a stockout and needs a completely different fix. **P1:** a pool within a node or two of its max
  is a scale-up failure scheduled for the next traffic spike.
- **Cluster-level NAP `resourceLimits`** — one more `clusters describe`. Same failure, fleet-scoped:
  the autoscaler refuses to provision GPU/CPU past a limit nobody remembers setting.

### 8. Dangling ComputeClass references — **P1**

Cheap, and the SOP already has the data: Step 2 dumps ComputeClasses and workloads in the same
`kubectl get -o json`. A workload whose `nodeSelector: cloud.google.com/compute-class: <name>` names
a class that does not exist in the cluster — or a namespace labelled
`cloud.google.com/default-compute-class=<name>` pointing at a missing class
(`compute-class-provisioning-methods.md:78`) — is permanently unschedulable.
**Prevents:** this is the cheapest possible finding to compute and produces pods that are Pending
forever while looking like a capacity problem.

### 9. 3.7 under-counts on multi-cluster projects — **P1**

Quota is per project+region, but Step 2 dumps state per cluster and nothing instructs aggregation
across clusters sharing a project and region. Two clusters in `us-central1` each at 60% of quota both
pass, while together they are at 120%. **Prevents:** the quota wall is a fleet property; a per-cluster
check cannot see it.

## Two things to resolve before merge

- **3.10 collides with an existing daily audit.**
  `agents/platform/governance/obtainability_audit_sop.md:169` (`rigid-scheduling`, cron 06:50) already
  flags the same objects with the same conditions and the same critical/major split — and deliberately
  sets `kind: manual`, reasoning that "blindly stripping a scheduling constraint is how an audit causes
  an incident". The new 3.10 sets `kind: manifest`, so at 09:20 the fleet gets an auto-PR that the
  06:50 audit refuses to write, plus a duplicate ledger entry. Either drop 3.10 or align it to
  `manual`.

- **`roles/compute.viewer` is not granted.** It is absent from both the read-only and gke-admin sets
  (`k8s-operator/scripts/provision_04_gcp_iam.sh:158`,
  `terraform/modules/kube-agents-iam/variables.tf:42`); only
  `agentplugins/gke-stockout-investigator/install.sh:139` adds it. As a core governance cron, this
  SOP's `gcloud compute regions describe` and `gcloud beta compute advice capacity` calls will fail on
  a stock install — the audit would report `partial` every run, or worse, silently record checks it
  never completed.

## A note on sourcing

Most of section 3–6 above is already written down in
`agents/platform/skills/gke-compute-classes/` and its `references/`. The SOP should **cite those
files rather than restate them** — the skill is the canonical home for ComputeClass semantics, and a
governance SOP that paraphrases it will drift from it. Concretely, the SOP's §3 checks want a
`Reference:` line next to `Command:`/`Flag when:`, pointing at the reference that defines the failure
mode. That also makes the checks maintainable: when GKE changes a default, the reference is updated
once and every check that cites it inherits the fix.

agents/platform/governance/stockout_prevention_sop.md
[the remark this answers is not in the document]
sergiu1spiridon: Removed customcomputeclasses and nodepools

agents/platform/scripts/platform_mcp_server.py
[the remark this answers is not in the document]
sergiu1spiridon: Updated. It will broadcast to both google chat and slack now

tlipski (approved):
lgtm, the tests comment is something worth chatting over with @mplakhtiy and @AntonTyb

agents/platform/governance/stockout_prevention_sop.md
sergiu1spiridon: We could also use kube-agents-autopush-infra to define some scenarios and test the audits in the e2e tests meant for release promotion

mplakhtiy (approved):

PR #466, opened 2026-07-29
feat(bench): run devops-bench evals via a library dependency on kubernetes-sigs/devops-bench

eugeneng04:
# Pull Request

## Why This Change

Prow CI (`hack/ci-eval-pr.sh`) runs evals through the legacy evaluator baked into the eval container image (`/app/pkg/evaluator/evaluate.py`, `AGENT_TARGET=kubeagents`), which couples this repo to the devops-bench source tree: an eval change means an image rebuild, and a devops-bench change can break us without warning.

This PR consumes [kubernetes-sigs/devops-bench](https://github.com/kubernetes-sigs/devops-bench) as a pip-installed library (pinned git SHA — no PyPI release yet) and keeps the agent transport and tasks here, so both projects ship independently.

## How to run it

Evals run through the **stock `devops-bench` CLI** — there is no wrapper command and nothing imports `kube_agents_bench` by name:

```bash
cd bench
uv sync
export PLATFORM_AGENT_TOKEN="$(kubectl get secret platform-agent-secrets -n kubeagents-system \
  -o jsonpath='{.data.API_SERVER_KEY}' | base64 --decode)"
export GEMINI_API_KEY=...   # the judge
uv run devops-bench ./tasks --no-infra --agent-type kubeagents
```

`--agent-type kubeagents` resolves through the `devops_bench.agents` entry point declared in `bench/pyproject.toml` — the sole registration path. devops-bench scans the group lazily on the first unresolved agent lookup (upstream #48), so the stock CLI works from a clean venv with no devops-bench checkout and no kube-agents import anywhere in the invocation.

If the agent's port is not already reachable locally, the harness spawns `kubectl port-forward svc/platform-agent -n kubeagents-system` itself; set `AGENT_LOCAL_PORT` to target an already-forwarded (or local) agent. For tasks that provision infrastructure, drop `--no-infra` and set `BENCH_TF_ROOT=./tf` (upstream #49) so stacks resolve inside this repo.

## What Changed

A new self-contained `bench/` directory. No existing behaviour is modified; `hack/ci-eval-pr.sh` is untouched and continues to run the legacy path until the team chooses to switch.

[...]

PR #582, opened 2026-08-07
ci(eval): run devops-bench evaluations via bench/ package as an extenal library dependency

lapis2002:
# Pull Request

## Why This Change
Follow-up to #466. Now that `kube-agents` consumes `devops-bench` as an external Python library dependency via `bench/`, we need to update the Prow CI evaluation script (`hack/ci-eval-pr.sh`) to execute `uv run devops- bench` from `./bench` instead of invoking the legacy evaluator script baked into the container image.

Additionally, in the multi-agent setup, the top-level Router agent delegates investigations to a subagent and returns the summarized diagnosis in its response. Because the evaluation judge inspects the final response text but cannot inspect tool calls or file artifacts created across container boundaries inside the subagent workspace, we simplify the `gpu-stress-test-diagnosis` task rubric by removing the markdown report file requirement (`analysis_report.md`). This allows all findings to be returned directly in the response text for accurate scoring.

## What Changed
**Files:**
- `hack/ci-eval-pr.sh`
- `bench/tasks/gpu-stress-test-diagnosis/task.yaml`

**Added/Changed/Fixed:**
- Updated `hack/ci-eval-pr.sh` to run devops-bench evaluations via new bench/ package
- Updated `gpu-stress-test-diagnosis` task spec to use the pinned `prebuilt/gpu-stress-test` OpenTofu stack.
- Simplified task prompt and rubric to output post-mortem findings directly in the response text rather than
requiring a file creation in the workspace.

## Why This Matters
- Completes the decoupling of `kube-agents` smoke test from custom evaluation container images.
- Resolves evaluation score degradation where the judge was unable to verify files created in subagent
workspaces.

## Context
- Accompanies [`oss-test-infra` PR](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2648) updating the Prow presubmit job image to `kubekins-e2e:latest-1.32` 

## Testing
- Manually verified the task with new bench locally.


hack/ci-eval-pr.sh
```diff
@@ -69,15 +70,16 @@ for TASK in "${TASKS[@]}"; do
...
 
-  # Snapshot existing result directories before running evaluate.py to prevent stale score leakage
-  PRE_RUNS="$(ls -d /app/results/run_* 2>/dev/null | sort || true)"
+  BENCH_DIR="${SCRIPT_DIR}/../bench"
+  # Snapshot existing result directories before running to prevent stale score leakage
+  PRE_RUNS="$(ls -d "${BENCH_DIR}/results/run_"* 2>/dev/null | sort || true)"
   EVAL_LOG="/tmp/eval_${TASK_NAME}.log"
 
-  (cd /app && python3 /app/pkg/evaluator/evaluate.py "${TASK}" 2>&1 | tee "${EVAL_LOG}") || true
+  (cd "${BENCH_DIR}" && uv run devops-bench "${TASK}" --agent-type kubeagents 2>&1 | tee "${EVAL_LOG}") || true
```
jayantid, line 86: Holding on this until the Prow side lands.

  oss-test-infra#2648 is still open and in **draft**, and today's job image (`devops-bench-harness:latest`) has no `uv`. If this
  merges first, `uv run` fails on every PR, gets swallowed by `|| true`, and the smoke test reports `[RESOURCE_PREPARATION_FAILED]
  Infrastructure setup/teardown error` — pointing at OpenTofu rather than the missing binary.

  One thing to fix in #2648 before it lands: the OpenTofu install line uses single-dash `-proto` / `-tlsv1.2`. curl reads those as
  bundled short options (`-p`, `-r oto`, `-t lsv1.2`), treats `=https` as a second URL, and binds `-o install-opentofu.sh` to *that*
  one — so the script body goes to stdout, the file is never created, and curl still exits 0. The job then dies on `chmod +x
  install-opentofu.sh`. Needs `--proto '=https' --tlsv1.2`.

  Adding `/hold` — happy to drop it once #2648 is merged and rolled out.

jayantid, line 86: Can we guard on `uv` before the task loop? Right now `|| true` swallows a missing binary, and the run falls through to the
  `[RESOURCE_PREPARATION_FAILED]` branch — so "uv isn't installed" gets reported as an infra failure and sends whoever's debugging
  to the wrong place.

  Something like this before the loop (around line 58):

  ```bash
  if ! command -v uv >/dev/null 2>&1; then
    echo "ERROR: uv is not installed. The eval harness requires it to run devops-bench."
    echo "Check that the Prow job image provides uv (see oss-test-infra#2648)."
    exit 1
  fi
  ```

  Failing fast here names the actual problem, and it keeps the `RESOURCE_PREPARATION_FAILED` label meaning what it says.

jayantid (changes requested):

hack/ci-eval-pr.sh
lapis2002: Updated.

lapis2002: `uv` is pre-installed in `devops-bench-harness:latest` image ([code](https://github.com/gke-labs/devops-bench/blob/main/Dockerfile#L77)), and I also did several test runs with prow (e.g: [this run](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/582/pull-kube-agents-smoke-test/2086673353103380480)) so this PR can can be merged independently in either order. I've updated oss-test-infra#2648 with the `--proto '=https' --tlsv1.2` as suggested.

jayantid (approved):

PR #586, opened 2026-08-07
fix(platform): give each Google Chat relay call its own transport

bradhoekstra:
# Pull Request

## Why This Change

The `envoy-credential-proxy` sidecar in `platform-agent-gateway` answered `502 Bad Gateway` to
`POST /v1/chat/api` in bursts:

```
WARNING credential-proxy chat relay operation failed path=/v1/chat/api type=SSLError
INFO    credential-proxy http "POST /v1/chat/api HTTP/1.1" 502 -
INFO    credential-proxy http "POST /v1/chat/api HTTP/1.1" 200 -
```

Each 502 is an outbound Google Chat call the agent lost — in practice a reply that never reached the
user. The immediate recovery reads like a transient network blip, but the code produces exactly this
signature on its own.

`GoogleChatRelay.__init__` builds one discovery service. `build()` resolves credentials through
`googleapiclient._auth.authorized_http()`, which creates a **single** `AuthorizedHttp` wrapping a
**single** `httplib2.Http` — one TLS socket, no locking, and httplib2 is documented as not thread
safe. The proxy serving those requests runs a thread per connection (`ThreadingUnixHTTPServer`), so
concurrent `/v1/chat/api` posts — an agent reply, the kanban notifier, a second session — interleave
records on that one socket. OpenSSL reports that as `ssl.SSLError`; the next call re-handshakes and
succeeds.

Two smaller defects compounded it. `api_call` executed with the default `num_retries=0`, opting out
of googleapiclient's own retry over `ssl.SSLError`, socket timeouts and 5xx. And the handler logged
only `type(exc).__name__`, so a transport fault, a `404` for an unknown space and a `403` for a
missing scope were one indistinguishable `"Google Chat operation failed"` — which is why diagnosing
this from the logs took a guess.

## What Changed

**Files:**

**Added/Changed/Fixed:**

- **Fixed:** each `api_call` checks out its own `AuthorizedHttp` from a bounded LIFO pool
  (`_checkout_http`). Pooled rather than thread-local because request threads are per-connection and
  the agent-side client opens a connection per call, so thread-locals would mean a fresh TLS
  handshake to `chat.googleapis.com` for every message the agent sends. A transport whose call
  raised is dropped rather than returned — a socket that failed mid-record would otherwise spread
  one failure across every call after it.
- **Changed:** `execute(http=…, num_retries=…)` opts into googleapiclient's jittered backoff.
  `messages.create` retries too: a duplicate message is a better outcome than a reply the user never
  sees, and the duplicate window is narrow (request reached Google, failure landed on the response).
- **Added:** `_chat_error_fields`, modelled on the existing `_slack_error_fields`, whitelists the
  status line onto both the log (`status=404`) and the 502 body (`{"chat": {"status": 404,
  "reason": "Not Found"}}`). Only the status crosses the boundary: an `HttpError` stringifies to a
  message embedding the request URI, and that URI names the space and the credentialed query.
- **Added:** eight tests in `GoogleChatRelayTest`, including a `threading.Barrier` test that holds
  two calls genuinely in flight and asserts they get different transports.

Two new knobs, `GOOGLE_CHAT_HTTP_POOL_SIZE` (default 8) and `GOOGLE_CHAT_API_NUM_RETRIES`
(default 3), are read with defaults in the same style as `SLACK_RELAY_MAX_REQUEST_BYTES`, so the
operator manifests and their golden testdata are untouched.

## Why This Matters

- Dropped Chat replies are silent from the user's side — the agent simply never answers.
- The shared-socket race scales with load: the more concurrent chat traffic, the more 502s.
- Any 502 that does remain now names its cause in the log line, so the next occurrence needs no
  guesswork.

## Context

The client side (`agents/platform/scripts/google_chat_relay_patch.py`) is deliberately unchanged —
the retry belongs behind the credential boundary, where the transport actually lives.

## Testing

- `make test-python` — 336 tests. The 4 errors present are pre-existing on `main` (verified by
  stashing): local-environment failures in `test_platform_mcp_server`, `test_session_kv_server` and
  `test_slack_relay_patch`, none touched here.
- Mutation-checked: reverting `execute(http=…, num_retries=…)` to bare `execute()` fails 5 of the
  new tests, so they genuinely pin the fix.
- Smoke-tested `_build_http` against the real `google_auth_httplib2.AuthorizedHttp` — builds, pools,
  and reuses.
- `make docs-check` passes; `review-docs-drift` run against the branch diff found nothing. The
  Chat section of `docs/credential-isolation-design.md` describes the relay at a level this change
  does not touch, and no page documents these env knobs.

---

Behaviour change is confined to the Google Chat relay path inside the credential sidecar. Rollout
risk is low: worst case is a duplicate Chat message where a retry lands after the request already
reached Google, against dropped replies today.

### DIFF
--- a/agents/platform/scripts/credential_proxy.py
+++ b/agents/platform/scripts/credential_proxy.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import argparse
 import base64
+import contextlib
 import hmac
 import http.client
 import io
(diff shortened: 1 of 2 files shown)


haoxuw (approved):
Lgtm

PR #599, opened 2026-08-09
fix(runtime): make the agent pod boot deterministically

adamparco:
# Pull Request

**Stacked PR 3 of 5.** Base: `ap-image-patch-surface`.

> **How this stack works upstream.** GitHub cannot base a cross-fork pull request on a branch that lives in the fork, so the five base branches (`ap-build-ci`, `ap-image-patch-surface`, `ap-agent-runtime`, `ap-governance-cron-2`, `ap-agent-prose`) have been pushed to this repository solely to serve as targets. `ap-governance-cron-2` mirrors the head of the fourth pull request; it replaced an earlier `ap-governance-cron` that branch protection would not let us update in place, so that leftover should be deleted too. These branches carry no work beyond the pull request that proposes them and should all be deleted once the stack lands. **As each pull request merges, retarget the next one at `main`.**

## Why This Change

Multi-replica agent pods raced on the shared volume at startup. Every replica believed it owned bootstrap, so profile scaffolding, config merges, and the image-version marker were written concurrently by several processes, and the result depended on which replica won. A second defect sat in the same boot path: an image upgrade was discarded when the volume already existed, because the version marker was only written on first boot.

The default profile's `config.yaml` keeps the delivery it has today — rendered by the operator into the `<agent>-config` ConfigMap and subPath-mounted read-only over `$HERMES_HOME/config.yaml`. An earlier revision of this branch replaced that with a three-way merge into a writable file, so that `/sethome` and `monitoring.install_id` could be written at runtime; that work has been **removed from this PR** (commit `2651658`) to keep the change to boot ordering and the change to config ownership separable. It can return on its own branch, on top of the `IS_BOOTSTRAP_PRIMARY` election this PR introduces.

## What Changed

**Files:**

- `deploy/shared/docker-entrypoint.sh`, `deploy/shared/defaults/config.yaml`
- `deploy/docker/Dockerfile` (image-layout and boot hunks), `deploy/docker/test_image_startup_contract.py` (new), `deploy/docker/test_merge_configs.py` (new)
- `agents/{chat,cluster,platform}/config.yaml`, `agents/platform/scripts/*` and their suites
- `k8s-operator/` (controller, CRD, testdata, LiteLLM integration), `charts/kube-agents/`, `agentplugins/`
- `tests/test_profile_overlay.py`, `tests/e2e/operator/`
- `docs/designs/gchat-session-metadata-data-flow.md`, `docs/site/.../operator/platformagent-crd.md`, `docs/site/.../reference/config.md`

**Added/Changed/Fixed:**

- Elected a single bootstrap primary in the entrypoint and gated the one-time work on it; other replicas wait for the marker.
- Serialised the writers to the shared volume with a `flock`, and gave each config writer its own staging file so two of them cannot clobber one another mid-write.
- Fixed the version marker so an image upgrade is applied on a pre-existing volume.
- Started the agent process inside the shared workspace so relative paths resolve, and so credentialed CLIs are not refused as out-of-workspace by the credential proxy.
- Resolved `platform_mcp_server.py` from `${HERMES_HOME}` rather than `/opt/data`.
- Ticked every named profile's cron store, not only the default.
- Dropped the unreachable `call_agent` tool surface from the profile configs.
- Enabled `incident_context` on the profile that actually receives chat ingress.
- Pinned the API server's advertised model to the one LiteLLM serves.
- Gave the LiteLLM gateway startup, readiness, and liveness probes.
- Precompiled the vendored Python tree at build time, made the remote MCP servers `lazy`, and removed the import-time `kubectl` shell-out from `agent_common_server` — together a large fixed cost off every worker cold start.
- Rendered agent replies as Slack Block Kit instead of flat mrkdwn; authored the incident-triage templates in standard markdown.
- Woke the front door only on kanban failure events, not on `completed`, whose summary the notifier has already delivered.
- Raised the agent container's CPU/memory ceiling for kanban fan-out, and corrected the `maxInProgress` guidance, which was stated against figures that had never been measured.
- Made `USER.md`'s kubeconfig line readable by the cluster-agent preflight.
- Covered the build-time platform-profile config merge with `test_merge_configs.py`, and the entrypoint's image-path contract with `test_image_startup_contract.py`.

## Why This Matters

- A nondeterministic boot is the worst class of bug to operate: it reproduces on some replicas, some of the time, and looks like a flake.
- Discarded image upgrades mean a rollout can appear to succeed while every pod keeps running the old code.
- The cold-start work is paid by every kanban worker, so it compounds across a fan-out rather than being a one-off at pod start.

## Context

Squashes: `dc7e68d`, `9a80aab`, `d8d4e25`, `2d84508`, `7d1d21c`, `a3a465e`, `f97c9ed`, `11bd246`, `feb123c`, `636a9fa`, `66c1df6`, `f6e98a3`, `02f17ee`, `3227848`, `cf9ce4d`, `43abf9c`, `ffd1482`, `6f6f325`, `8bbd2c0`, `d8c7661`, `446deff`, `9ffde26`, `4a4cd6f` (runtime portions). `8375e8f`, the writable-config commit, was squashed in and has since been reverted out again by `2651658`.

This PR completes `deploy/docker/Dockerfile`, whose patch-wiring hunks landed in PR 2.

Stack: `ap-build-ci` → `ap-image-patch-surface` → **`ap-agent-runtime`** → `ap-governance-cron-2` → `ap-agent-prose`.

## Testing

- `make test-python` — pass, including the new `test_image_startup_contract.py` and `test_merge_configs.py`.
- `python -m unittest discover -s tests -p 'test_*.py'` — pass (49 tests), plus `sh -n deploy/shared/docker-entrypoint.sh`.
- `go build ./...` and `go test ./...` in `k8s-operator/` — pass. The golden manifests were regenerated with `-update` and then prettier-formatted.
- `make docs-check` — pass.
- `prettier --check` on the 16 changed Markdown/JSON/YAML files — pass.
- Not run: a live multi-replica rollout. The bootstrap election is covered by the startup-contract test rather than by an in-cluster run.

---

Behaviour-changing, and the highest-risk PR in the stack: it alters pod startup ordering and the operator's rendered manifests. Rollout is via the agent image plus an operator upgrade; the boot changes are idempotent on an existing volume, and the version-marker fix means an existing volume now correctly picks up the new image.


deploy/shared/docker-entrypoint.sh
```diff
@@ -145,20 +159,104 @@ fi
...
+# DELETED the name the agent had just written — the observed
+# `resource_attributes: {}`). Shared-state writes are serialised by the lock
+# below instead, not skipped, so a non-primary owner still leaves the volume
+# usable on its own.
+if [ "$PLATFORM_AGENT_ROLE" = "sidecar" ]; then
+    IS_BOOTSTRAP_PRIMARY=0
+else
+    IS_BOOTSTRAP_PRIMARY=1
+fi
+
```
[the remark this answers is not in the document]
adamparco: Confirmed on the fact, corrected on the consequence. Fixed in `0f4c1a0`.

**The fact is right, and it is stronger than you put it.** An unrestricted search returns seven occurrences of `PLATFORM_AGENT_ROLE`, and the only producer is the one you named — `platformagent_manifests.go:2179`, inside `if isDashboardEnabled(agent)`, on `platform-agent-dashboard`. There is no `envFrom` anywhere in `internal/controller/`, no downward-API env, and the sole operator init container has no `Env` block at all. `safeSandboxEnvOverrides` (:1956-1972) admits only four `OTEL_*` names, so a hand-written CR cannot deliver it either.

It is also worse than "no producer sets it": per-replica stamping is **structurally impossible**. `buildDeployment` (:1631) and `buildStatefulSet` (:1662) both derive every pod from a single `buildPodTemplateSpec` call, and leadership above one replica is a dynamic Lease held by `leader_elect.py`. No env var can name "the non-primary replica". And the one container that does carry `sidecar` also carries `AGENT_SHARED_STATE_SETUP=skip`, so the step-1.5 gate `exec`s it away 36 lines earlier — the `sidecar` branch has zero live readers in anything the operator emits. `git blame` puts the false sentence on `d935212`, this PR's own commit, so there is no pre-existing-on-main defence: it is ours to fix.

**Where the finding overreaches is the harm.** Step 2d (`:343`) sits *inside* the `flock` region acquired at `:231-236` and released at `:684-686`, so replicas are serialised, not concurrent. I ran the real `default_profile_config.py` to check the rest: after simulating a live agent's `/sethome`, `monitoring.install_id` and a `/model` override, a second replica's rebuild reported `3 top-level key(s) carry runtime state` and left all three intact — the three-way merge rules at `:150-172` are what save them. A peer's boot is an idempotent no-op, not a clobber.

More to the point, **"always primary" is the required answer for the other two consumers, not a bug.** Each pod's event-watcher posts to its *own* `127.0.0.1:8699` (`start-services.sh:70`), so every pod must run a session KV server; stamping `sidecar` on "non-primary replicas" would leave those pods' watchers with no daemon. The OTel service name comes from the same pod template, so every replica writes identical content. So the defect was the *account* of the flag, not its behaviour — which is why the fix is the comment, and why "add the missing env var to the operator" would have been the wrong repair.

**What `0f4c1a0` changes:**

1. The comment above the election now states the per-container contract it actually implements, says explicitly that it is not a per-replica election and why one is impossible here, and records that across replicas step 2d is serialised by the step-1.6 lock and idempotent rather than skipped. The step-2d comment gets the matching clarification, so it and `platformagent_manifests.go:2174-2175` no longer describe the same variable in contradictory terms.
2. A test pins the property the entrypoint now leans on: `test_a_peer_replicas_boot_keeps_the_live_leaders_runtime_edits`. You were right that there was no coverage here — a search for `IS_BOOTSTRAP_PRIMARY` or `BOOTSTRAP_LOCK` in tests returned nothing repo-wide.

**One real defect your finding walked past, now also fixed.** `_atomic_write_yaml` staged through a fixed `config.yaml.tmp`. That is atomic only while exactly one writer exists, and the step-1.6 lock has three documented ways to let two through — no `flock` binary (`:221`), an unwritable `$TARGET_DIR` (`:230`), and the `flock -w 300` timeout that says outright it is "proceeding concurrently with the peer container". Running eight rebuilds with the lock bypassed reproduced exactly what the docstring warns about: one writer died with `FileNotFoundError: 'config.yaml.tmp' -> 'config.yaml'`, and concurrent readers saw `config.yaml` at **zero bytes** 641 times. Each writer now stages through a uuid-suffixed name, matching what `multiuser_memory/__init__.py:145` already does, and cleans up on failure. `test_writers_do_not_share_a_staging_path` fails against the old code (4 writes, 2 distinct paths) and passes now.

**Known residual, deliberately not fixed here.** The lock is released at `:686` before `exec "$@"` at `:739`, so a live leader holds nothing. A follower booting during a rolling update still does a read-merge-`os.replace` against the leader's live `config.yaml` (`default_profile_config.py:259` → `:268`), a window measured at ~2.6 ms. Worst case is one lost `/sethome` or `/model` edit — a lost update, not corruption, since `os.replace` is atomic within the directory. A working primary election would not close it either, because the pod designated primary can restart while a different pod holds the lease; closing it properly means leadership-gating step 2d or having the running agent take the same lock, which is a design change rather than a comment fix. Flagging it rather than smuggling it in.

Full suite green: 1,770 Python tests across the stack, plus `make docs-check`.

toshiowang:
+@dshnayder 

I recall read-only config was intentional to prevent agent breaking itself and getting stuck in an unrecoverable state.

adamparco:
@toshiowang Your recollection is right, and thank you for raising it — it is worth being precise
about what was intentional and what actually changed here, because the scope is narrower than
"the agent's config is writable now".

**What changed is one file: the default (Chat) profile's `$HERMES_HOME/config.yaml`.** The agent's
read-only ceiling on the Kubernetes and cloud APIs is untouched — `docs/architecture/03-security-model.md`
§3 still holds ("Every agent is **read-only** on the cluster and cloud APIs"). The platform profile's config stays entirely image-owned and force-synced, and cluster
profile configs stay identity-stamped at scaffold time. Nothing else on the volume became writable.

**Why it had to change.** The operator subPath-mounted its rendering over that path, and a subPath
mount is a read-only mount *point*, so:

- `/hermes sethome` failed. `os.replace` onto a mount point gives EBUSY, Hermes' `shutil.copyfile`
  fallback then gives EACCES, and the Slack egress sanitiser scrubbed the path out of the message —
  which is why the user saw `Failed to save home channel: [Errno 13] Permission denied: ''`.
- `monitoring.install_id` could never be minted. `ensure_install_id` is fail-open on a read-only
  home, so it silently re-rolled on every start.
- No saved slash-command preference survived a restart.

A second defect hid behind the first: entrypoint step 2a force-copied the image's config over that
same path on every boot, so the operator's rendering never reached the agent at all. On the live
deployment, 12 keys the CR asked for — including `platforms.slack.enabled` and the model endpoint —
were absent from one side or the other of the mount. So the read-only mount was not holding a line
in practice; it was breaking `/sethome` while also silently dropping what the CR asked for.

**Why I think the unrecoverable-state risk is smaller than it looks** — with a real caveat below,
because you may still be right:

- The file is rebuilt from scratch on every start (`docker-entrypoint.sh` step 2d — unconditional,
  never mtime-gated) from the image template plus the operator's `profile-default.overlay.yaml`.
- Any key the image or the CR declares is reconciled at that point. If the runtime changed a key
  *and* the baseline also moved, the operator wins — a stale local edit cannot veto a CR. The
  per-key rules are in `deploy/shared/default_profile_config.py`'s module docstring.
- So the recovery lever is the CR itself: change the key in the `PlatformAgent`, roll, and the
  agent's own value is overwritten. The blunt lever is deleting `config.yaml` and
  `.default-config-baseline.yaml` from the PVC — step 2d re-seeds from `/opt/chat-template/config.yaml`
  and adopts.

**Where you could still be right.** One case does survive a restart: the runtime changed a key that
the image or the operator declares, and neither of them has changed it since — then the runtime's
value stands. That is deliberate, since it is what makes `/model` and `/sethome` stick, but it is
exactly the shape you are describing. An agent that writes a bad value into a key nobody else is
reconciling keeps it until someone touches that key in the CR or wipes the file. Recoverable, but
not self-correcting, and the operator has to know to do it.

If you want a belt-and-braces guard on top, two cheap options: (a) validate the rebuilt config at
boot and fall back to image + overlay if it does not parse or fails a sanity check, or (b) keep the
last known-good config beside the baseline so a boot can roll back to it. I would default to (a) if
you think the residual case is worth closing — happy to add it to this PR or as a follow-up,
whichever you prefer.

One note on the alternative we rejected, since it bears on the same worry: a preserve-allowlist of
runtime keys to rescue across the force-copy. Hermes has 60-odd `save_config` call sites and gains
more with every release, so the 61st is dropped silently — the same silent-no-op failure this step
exists to remove. The three-way merge was chosen because it fails visibly instead.

deploy/shared/default_profile_config.py
```diff
@@ -0,0 +1,314 @@
...
+        if isinstance(ours_v, list) and isinstance(theirs_v, list):
+            out[key] = _reconcile_list(base_v if isinstance(base_v, list) else [], ours_v, theirs_v)
+            continue
+
+        # Scalar (or a type change, which is the same decision). The runtime only wins
+        # where the baseline has not moved; if both changed, the operator is mid-
+        # reconcile and wins.
+        if base_v is _ABSENT or theirs_v != base_v:
+            continue
+        out[key] = copy.deepcopy(ours_v)
```
toshiowang: **[SECURITY RISK]** — the mechanism is definitive; the trigger is any agent write to `config.yaml`.

`buildDefaultVolumeMounts` drops the `config.yaml` subPath mount, so the effective config is now a PVC file the agent can write — and this rule pins whatever it writes. The runtime wins whenever the operator's rendering *held still* since the last boot, which is the steady state on every reconcile, so a runtime edit to an operator-declared key survives indefinitely.

Running this file directly:

| step | endpoint | `slack.enabled` | `approvals.cron_mode` |
| --- | --- | --- | --- |
| healthy | `litellm:4000` | `True` | `approve` |
| agent writes `config.yaml` | `typo-i-made-up:4000` | `False` | `auto` |
| `kubectl rollout restart` | unchanged | unchanged | unchanged |
| re-apply the *same* CR | unchanged | unchanged | unchanged |
| *change* the CR's endpoint | healed | still `False` | still `auto` |
| delete `.default-config-baseline.yaml`, restart | healed | healed | healed |

Restart does not heal, and neither does re-applying the CR: "held still" is measured against the recorded baseline, not the live file, so healing needs a genuinely different value and heals only that key. The only full reset is deleting the baseline file by hand, which needs `kubectl exec` or PVC access.

Two consequences worth separating:

- **Self-lockout.** `platforms.slack.enabled: false` is unrecoverable in-band — Slack is the front door, so nobody can tell the agent to undo it, and no restart or CR re-apply does it for them. This is the litellm-ConfigMap incident's shape, moved inside the pod: the operator still reconciles the ConfigMap, but the ConfigMap is no longer what the agent runs from.
- **Durable escalation.** The rule is unscoped, so `approvals.cron_mode`, `approvals.cron_scan`, `plugins.enabled` (which carries `tool_call_audit`) and `tool_loop_guardrails.loop_caps` are all operator-declared and all runtime-overridable, with the override surviving every restart.

**Fix:** `adopt()` already has the safe semantic — runtime wins only for keys *neither* declaration mentions (`platforms.*.home_channel`, `monitoring.install_id`) — and the extra power `reconcile()` adds over it is exactly what buys the lockout. `adopt` everywhere is too blunt, since `_reconcile_list`'s `_save_platform_tools` case is real and `platform_toolsets` is operator-declared, so scope this to an allowlist of runtime-carryable declared paths rather than "any key where the declaration held still". At minimum make `approvals.*`, `plugins.enabled` and `platforms.*.enabled` operator-wins-always.

If the merge rules are out of scope here, please at least have `rebuild()`'s status line name the carried keys. It currently returns `3 top-level key(s) carry runtime state`, which is the only signal that anything is pinned and it does not say what.

deploy/shared/docker-entrypoint.sh
```diff
@@ -181,6 +289,141 @@ if [ -d "/opt/defaults/scripts" ]; then
...
+# Add an id to --cron-jobs only when the image genuinely owns that job's
+# definition on a live volume.
+if [ -f "/opt/defaults/cron/jobs.json" ] && [ -f "$SCAFFOLD" ]; then
+    HOME=/tmp HERMES_HOME="$TARGET_DIR" "$INSTALL_DIR/.venv/bin/python3" \
+        "$SCAFFOLD" \
+        --home "$TARGET_DIR" \
+        --template /opt/defaults \
+        --items "cron" \
+        --cron-jobs "profile-cron-tick" \
+        --cron-retire "compliance-audit obtainability-audit security-patch-orchestrator fleet-wide-cost-analysis fleet-consistency-drift ai-security-audit github-issue-resolver" \
```
toshiowang, line 336: **[CRITICAL BUG]** — definitive, and it fires unconditionally on every upgraded cluster.

The comment above justifies this with "now that `profile-cron-tick` gives it one, they are back on the Platform Agent's own roster." Neither premise holds on this branch:

- `profile-cron-tick` exists in no cron roster — `grep -rn 'profile-cron-tick'` matches only docstrings under `deploy/docker/patches/`.
- `agents/chat/scripts/profile_cron_tick.py` does not exist.
- `agents/platform/cron/jobs.json` is `"jobs": []`, and its own comment says the six governance jobs live in the Chat Agent's roster precisely because the platform profile has no ticker.
- Neither cron file is modified by this PR, so `/opt/defaults/cron/jobs.json` still ships all six enabled with their `dispatch_*.py` scripts.

Running this exact argv through the branch's `profile_scaffold.py` against the shipped roster:

| volume | resulting ids |
| --- | --- |
| fresh (`prior` empty, so `_merge_after_overlay` no-ops) | all 9, including the six |
| existing PVC | `cluster-agent-reconcile`, `bootstrap-inventory-scan`, `bootstrap-inventory-delivery` |

Every upgraded cluster silently loses `compliance-audit`, `obtainability-audit`, `security-patch-orchestrator`, `fleet-wide-cost-analysis`, `fleet-consistency-drift` and `github-issue-resolver` on the next pod restart, with nothing anywhere to run them; a fresh install keeps firing them, so the two diverge permanently.

Separately, `--cron-jobs "profile-cron-tick"` names an id no roster carries, so `merge_cron_store`'s `only_ids` filter reduces `image_jobs` to `[]`. Step 2c force-merges nothing onto an existing volume — the problem its comment describes ("a newly shipped job simply never arrives") is not fixed either. As written the step only deletes.

**Fix:** land this together with the change that adds `profile-cron-tick` and moves the six ids onto `agents/platform/cron/jobs.json`, so the retire is a move rather than a deletion; or drop the `--cron-retire` line from this PR and keep `--cron-jobs` as the only filter until the destination roster exists. `ai-security-audit` is not shipped by any roster on this branch and can go either way.

toshiowang (changes requested):
Two blocking findings, one tentative risk.

**1. The default profile's config has left the operator's control loop.** Dropping the `config.yaml` subPath mount fixes the `/sethome` EACCES, but `_reconcile_dict`'s scalar rule then pins *any* key the agent writes: the runtime wins whenever the operator's rendering held still since the last boot, which is the steady state on every reconcile. Running the PR's own `default_profile_config.py`: after the agent writes `model.endpoint`, `platforms.slack.enabled: false` and `approvals.cron_mode: auto`, a `kubectl rollout restart` changes nothing, re-applying the same CR changes nothing, changing the CR heals only the one key you changed, and the only full reset is deleting `.default-config-baseline.yaml` by hand. The rule is unscoped, so `approvals.*`, `plugins.enabled` (which carries `tool_call_audit`) and `tool_loop_guardrails.loop_caps` are durably self-overridable, and `platforms.slack.enabled: false` is an unrecoverable in-band lockout — you cannot reach the agent to undo it. The operator still reconciles the ConfigMap, but the ConfigMap is no longer what the agent runs from.

**2. Step 2c deletes six governance cron jobs with nowhere for them to go.** Its comment says they moved to the Platform Agent's roster now that `profile-cron-tick` gives it a ticker, but `profile-cron-tick` appears in no roster on this branch, `agents/chat/scripts/profile_cron_tick.py` does not exist, and `agents/platform/cron/jobs.json` is `"jobs": []` and documents itself as inert for want of a ticker. Neither cron file is touched by this PR. Running the branch's `profile_scaffold.py` with this argv leaves a fresh volume with all nine jobs and an upgraded volume with three: `compliance-audit`, `obtainability-audit`, `security-patch-orchestrator`, `fleet-wide-cost-analysis`, `fleet-consistency-drift` and `github-issue-resolver` stop firing on upgraded clusters and keep firing on fresh ones. `--cron-jobs "profile-cron-tick"` also names an id no roster carries, so `only_ids` filters the image side to `[]` and the step force-merges nothing — it is purely destructive as written.

Everything else checked out: the flock fd-9 lifecycle, `adopt`'s handling of `/sethome` and `monitoring.install_id`, the `args` list-union, the dashboard's overlay-as-config subPath, and the LiteLLM probes. Go build/vet/test and both Python suites pass.

*Tentative risk* — `agents/platform/scripts/profile_scaffold.py:296`: ``scratch = destination.with_name(destination.name + ".tmp")``. This PR adds a second scaffolder write through this path (step 2c, on `$TARGET_DIR/cron/jobs.json`) while `deploy/shared/default_profile_config.py:250` moves its own writer to a uuid-suffixed staging name for exactly this reason, with `tests/test_default_profile_config.py::test_writers_do_not_share_a_staging_path` asserting it and naming the three documented ways two writers get past the step-1.6 lock. Under any of those, writer B truncates the shared `.tmp` that A then `os.replace`s into place, and the profile boots with a torn `jobs.json`. Same one-line fix as the config writer.

<sub>Review produced with the help of Claude.</sub>

deploy/shared/docker-entrypoint.sh
toshiowang: Retracted. This is covered in the followup #600. 

Small, self-contained PRs are ideal, but given how fast repo is changing, I don't want the PR staying open for much longer and creating more churn. 

It just needs to be merged to main around the same time, which it should.

toshiowang:
The race condition fixes are great, but writable config.yaml needs more discussion. 

If the writable config.yaml can be refactored out for a separate PR, I think this PR is in a good state.

dshnayder:
Concern on making the default profile's `config.yaml` writable.

We've already hit this in practice: an agent reacting to intermittent quota errors from the LLM gateway decided it would do better talking to a provider directly, rewrote its own `model` block, and left a placeholder where the API key should have been. It then had no working model, so it couldn't reason its way back out — recovery was a manual edit of `config.yaml`.

The three-way merge doesn't prevent a repeat. `_reconcile_dict` hands the key to the operator only when the *rendered value changed*; a CR nobody edits renders identically every boot, so `theirs == base` and the runtime edit wins indefinitely (`test_an_unchanged_overlay_leaves_a_runtime_edit_alone` asserts exactly this). A restart doesn't clear it and neither does re-applying the same CR. The only recoveries I can find are changing the CR to a *different* value, or deleting `.default-config-baseline.yaml` so `adopt()` runs — neither is documented.

Worth flagging that the writer needn't be the Chat Agent. It has no file/terminal tools and can't write its own config. But the platform and cluster profiles both grant `hermes-cli` (which carries terminal/file), neither sets `disabled_toolsets`, and `HERMES_HOME=$TARGET_DIR=/opt/data` puts the Chat Agent's config at the root of the shared volume those workers run in. So another agent can rewrite the front door's config, and the write sticks across restarts.

Suggestion: instead of the allow-list the docstring rejects (fairly), a small deny-list applied after `reconcile`/`adopt` — always take `theirs` for `model.base_url` / `model.provider` / `model.api_key`, `mcp_servers`, `agent.disabled_toolsets`, `platform_toolsets`. `model.default` stays mutable so `/model` keeps working, and anything not listed keeps today's behaviour. Alongside that: log the carried key *paths* rather than just a count, and document the break-glass.

dshnayder:
**Follow-up idea, not blocking this PR** — an agent-side liveness probe, so the model-config reset becomes self-healing instead of "someone notices and bounces the pod."

With the reset in place the recovery is a restart, but nothing triggers one; the agent sits dead until a human spots it. A probe closes that loop. Two of them, both zero-token:

**Liveness — config drift, no network, no tokens.** Compare the live `$HERMES_HOME/config.yaml` `model` block against the declaration:

```python
live = load_yaml(pathlib.Path(os.environ.get("PLATFORM_AGENT_HOME", "/opt/data")) / "config.yaml")
declared = build_baseline("/opt/chat-template/config.yaml",
                          ["/opt/agent-config/profile-default.overlay.yaml"])
drift = [k for k in ("provider", "base_url", "api_key", "default", "model")
         if (live.get("model") or {}).get(k) != (declared.get("model") or {}).get(k)]
sys.exit(1 if drift else 0)
```

Importing `build_baseline` rather than reimplementing it is the point: the probe's notion of "correct" becomes literally the function that runs on restart, so it can't restart-loop a pod the entrypoint is happy with, or pass one it isn't. Reading the two read-only inputs rather than `.default-config-baseline.yaml` matters for the same reason — that file sits on the writable PVC beside the file being policed.

Deliberately **not** a call to LiteLLM. A liveness probe that dials the gateway also fails when the gateway is down, and restarting the agent can't fix that — it converts a LiteLLM outage into a CrashLoopBackOff on the one component that could still tell you about it.

**Readiness — reachability.** `GET {base_url}/models` with the configured key. Metadata listing, not an inference call, so still zero tokens. Asserting that `model.default` appears in the returned ids also catches the `400 Invalid model name` failure from the other direction, including a rename on the LiteLLM side.

**Wiring.** The `platform-agent` container has no probes today; these would be rendered by the operator alongside the container spec, the same shape as the LiteLLM probes this PR adds:

```yaml
startupProbe:
  exec:
    command: ["/opt/hermes/.venv/bin/python3", "/opt/defaults/scripts/model_alive.py"]
  periodSeconds: 10
  failureThreshold: 30      # ~5 min: step 2d, vendored-tree import, hermes boot
livenessProbe:
  exec:
    command: ["/opt/hermes/.venv/bin/python3", "/opt/defaults/scripts/model_alive.py"]
  periodSeconds: 60
  timeoutSeconds: 10
  failureThreshold: 2
readinessProbe:
  exec:
    command: ["/opt/hermes/.venv/bin/python3", "/opt/defaults/scripts/model_ready.py"]
  periodSeconds: 30
  timeoutSeconds: 10
  failureThreshold: 3
```

The `startupProbe` isn't optional: before step 2d runs, `config.yaml` is absent or unmerged, so the drift check correctly fails and would kill the container mid-boot. It suppresses liveness until it first passes.

One caveat on readiness — it won't gate chat ingress. Slack is socket-mode and Google Chat is Pub/Sub pull, both outbound, so a not-ready pod still receives messages. Its value here is signal (`kubectl get pod`, alerting), not traffic control; liveness is what does the self-healing.

Two short scripts, sitting alongside `default_profile_config.py`. Scoped at the failure we've actually seen — an agent breaking its own config while trying to fix something — rather than at an agent working to defeat the check.

dshnayder:
Updating this — my "the sidecar reads a partial file" claim was wrong, and the rewritten comment on the mount is right. I rendered `profile-default.overlay.yaml` out of the expected ConfigMap and diffed it against `merge(agents/chat/config.yaml, overlay)`: 36 leaf paths on both sides, none missing, none differing. The operator emits a whole-file rendering, so the sidecar does see a complete config.

What's actually wrong is narrower, and it's been confirmed on a live pod: the sidecar sees a config with **no runtime state in it**. `/hermes sethome` writes the home channel into the default profile's `config.yaml` on the PVC; the dashboard is reading the ConfigMap render, which by construction can never contain it. So the Slack settings don't show up there, and never will. That drift is new with this PR — before it there was no runtime state in that file to miss, so the shadow mount was harmless.

It also leaves the mount's stated justification stale. The comment says the subPath exists because "nothing puts one on the PVC for this container to find." As of step 2d something does: `/opt/data/config.yaml` is a real merged file on the volume the dashboard already mounts at `homeDir`. The subPath now shadows it.

And this isn't only a viewing problem. `hermes dashboard` is `hermes_cli/web_server.py`, which has `PUT /api/config`, `PUT /api/config/raw` and ~15 `save_config()` call sites — a settings editor, not a viewer. Every one of those writes lands on a ConfigMap-backed volume, so they all fail. Not silently, at least: the handler raises `HTTPException(500, "Internal server error")`, so a user sees a failure — with nothing to suggest the cause is a read-only mount.

### What it would take to make it a real settings editor

Four things, and the first two are close to free because this PR already did the hard part.

**1. Drop the subPath mount.** One line in `dashboardVolumeMounts`. The dashboard then reads and writes `/opt/data/config.yaml` through the `platform-agent-data-vol` mount it already has.

**2. Durability — already solved here.** Step 2d's three-way merge doesn't care which container wrote a key. A dashboard save is `ours` and gets carried across restarts by the same rules as `/sethome`, while operator-declared keys still reconcile when the CR moves. Worth stating explicitly, because it's what makes this worth doing at all: before #599, a dashboard save would have been wiped by the next boot regardless of the mount.

**3. Boot ordering — needs the initContainer.** Nothing sequences containers within a pod, and the entrypoint already names the hazard: the dashboard "must not come up against a missing config, fall back to Hermes' built-in defaults, and save those over the top." Today the shadow mount is the only thing preventing that. Remove it and the hazard goes live — on a fresh PVC the dashboard could overwrite the real config with Hermes defaults before step 2d ever runs.

The fix is the one already recorded as the KNOWN LIMIT at step 1.5: move step 2d into an operator-managed initContainer. There's a template for it a few hundred lines away — `buildSandboxCredentialCleanup` is the same shape (agent image, data-vol mount, runs to completion, `readOnlyRootFilesystem: true`); it would additionally need `platform-agent-config-vol` and would run `default_profile_config.py`. That converts the race into an ordering for both containers at once, and it retires the `IS_BOOTSTRAP_PRIMARY` election for this step as a side effect.

**4. Concurrency — serialize the read-modify-write.** Both sides do RMW on one file with no lock between them:

```python
existing = read_raw_config()
incoming = _denormalize_config_from_web(body.config)
save_config(_deep_merge(existing, incoming))
```

The agent's slash commands are the same shape. `_atomic_write_yaml`'s `os.replace` means you'll never see a torn file, but a dashboard save concurrent with a `/sethome` silently loses one of the two.

I would not do this in the same change. Both writers are human-driven and infrequent, so the window is genuinely small — worth knowing about, not worth blocking on. When it does start to matter, the existing patch mechanism fits: a `deploy/docker/patches/` triple (`config_write_lock.py` plus `apply_`/`verify_`) wrapping `save_config` and the two web_server config handlers so read, merge and write all happen under an exclusive `flock` on a lockfile beside `config.yaml` — the same pattern as `cron_tick_lock_scope.py`.

**One scope question worth settling first.** `PUT /api/config` takes a profile — `_profile_scope(body.profile or profile)` — so a "real settings editor" can write *any* profile's config, including `platform`, which grants terminal/file/exec. There's no auth on 9119. It's ClusterIP-only so this isn't internet-reachable, but it's worth deciding deliberately whether the dashboard edits the default profile or becomes a full admin plane for every profile in the pod.

### For #599

None of the above blocks this PR. The only thing I'd change here is the code comment, which still says the mount exists because nothing puts a config on the PVC — step 2d does.

The entrypoint comment at step 3 is also still inaccurate: *"platform-agent-dashboard does not mount platform-agent-config-vol at all (the operator's dashboardVolumeMounts)."* It does, at `$HOME/config.yaml`. The guard it defends (`[ -d "$OVERLAY_DIR" ]`) is still correct, since `/opt/agent-config` genuinely isn't mounted there — a wrong reason for a right answer rather than a live bug.

dshnayder (approved):

PR #601, opened 2026-08-09
docs(agents): align the personas, skills, and site prose with the kanban result contract

adamparco:
# Pull Request

**Stacked PR 5 of 5.** Base: `ap-governance-cron-2` (the upstream mirror of PR 4's head). Merging this completes the stack; the tree then matches `deploy-test` byte for byte.

> **How this stack works upstream.** GitHub cannot base a cross-fork pull request on a branch that lives in the fork, so the five base branches (`ap-build-ci`, `ap-image-patch-surface`, `ap-agent-runtime`, `ap-governance-cron-2`, `ap-agent-prose`) have been pushed to this repository solely to serve as targets. `ap-governance-cron-2` mirrors the head of the fourth pull request; it replaced an earlier `ap-governance-cron` that branch protection would not let us update in place, so that leftover should be deleted too. These branches carry no work beyond the pull request that proposes them and should all be deleted once the stack lands. **As each pull request merges, retarget the next one at `main`.**

## Why This Change

The persona files and the site's concept pages still described pre-kanban behaviour: an agent that answers in whatever channel addressed it, and a dependency edge whose direction was never actually stated. After the earlier PRs in this stack, a worker must record a result on the card, and the card is what carries the answer back.

For a repository whose prose _is_ the agent's instructions, this is not cosmetic drift — an agent reading these files was being told to do something the harness no longer permits.

## What Changed

**Documentation and persona files:**

- `agents/chat/{AGENTS,SOUL}.md`, `agents/cluster/{AGENTS,SOUL}.md`, `agents/platform/SOUL.md`, `agents/platform/docs/glossary.md`
- `agents/cluster/skills/gke-workload-troubleshooting/SKILL.md`, `agents/platform/skills/cluster-agent-lifecycle/SKILL.md`, `agents/platform/skills/workload-rebalancing/SKILL.md`
- `docs/designs/agent-communication.md`
- `docs/site/.../concepts/{chatops,cluster-agents,platform-agent}.md`, `.../overview/architecture.mdx`, `.../deploy/telemetry.md`, `.../reference/security-and-iam.md`, `.../index.mdx`

**Code (7 files, +120/−26):** these are the non-prose remainder of the squashed commits listed under Context, carried here because this is the last PR in the stack.

- `deploy/shared/docker-entrypoint.sh` — profile scaffolding at pod startup.
- `deploy/shared/default_profile_config.py` and `tests/test_default_profile_config.py` — the default profile's generated config, with its unit tests.
- `deploy/docker/patches/cron_tick_lock_scope.py` and `deploy/docker/patches/test_cron_tick_lock_scope.py` — cron tick lock scoping, with its unit tests.
- `deploy/docker/patches/verify_kanban_notifier.py`, `deploy/docker/patches/verify_kanban_result.py` — build-gate assertions.

**Added/Changed/Fixed:**

- Stated the kanban result contract in the Chat, Cluster, and Platform personas and in the three affected skills.
- Corrected the same prose in the site's ChatOps, cluster-agent, platform-agent, and architecture pages.
- Said which way a kanban dependency edge points, in the glossary and the communication design.
- Resolved the GCP project at runtime in example commands; never wildcard it.
- Recorded where token spend is actually captured, on the telemetry page.

## Why This Matters

- These files are read by the agents at runtime. Stale instructions here produce wrong behaviour, not just a confusing read.
- It closes the stack: with this merged, the documented contract and the implemented contract agree.

## Context

Squashes: `e38360d`, `7c0d971`, `4a4cd6f`, `5efffd9`, `a1baac6` (prose portions), plus the persona and site-page portions of `633c380`, `c64a9cf`, `a3a465e`, `7d1d21c`, `f642c2a`, `a33b249`, `2fc3a6c`, `4abfa6c`.

Deliberately last in the stack: the prose describes behaviour that PRs 2–4 introduce, so landing it earlier would have documented something that did not yet exist.

Stack: `ap-build-ci` → `ap-image-patch-surface` → `ap-agent-runtime` → `ap-governance-cron` → **`ap-agent-prose`**.

## Testing

- `make docs-check` — pass (generated regions, relative links, terminology, documentation-map coverage).
- `make test-python` — pass.
- `prettier --check` on the changed Markdown/MDX files — pass.
- Verified the stack tip's tree hash is byte-identical to `deploy-test`, confirming the five-PR split loses nothing.

---


agents/platform/SOUL.md
```diff
@@ -12,36 +12,48 @@ The Chat Agent delegates to you **exclusively through the Kanban board** — it
...
 
-Crucial detail: a child card you create **while running as a worker is not automatically subscribed to the user's chat** (only the Chat Agent's original card is). So immediately after each `kanban_create`, propagate the subscription onto the new child:
+- **Want them to run now? Create them with no `parents` at all**, or with `parents` listing only cards that are already done. Then `kanban_block(kind="dependency")` and the dispatcher will bring you back when they finish.
+- **Never `kanban_block` waiting on cards that list your card in their `parents`.** They cannot start until you finish and you are not finishing until they start. On 2026-08-07 this stalled the fleet security baseline assessment for fifteen minutes across two attempts. The image now repairs this shape automatically when it can (`deploy/docker/patches/kanban_scheduling.py`), but the repair is a backstop for a mistake, not a supported pattern — and it deliberately declines to touch graphs it cannot prove are broken.
+
```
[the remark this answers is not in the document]
adamparco: Confirmed, and fixed in `fb2b112`.

I verified this against the source rather than taking the finding as read, and the strongest evidence is the patch module's own docstring: `block_task(kind="dependency")` routes the card to `todo` rather than `blocked`, "which skips the `block_recurrences` loop breaker entirely", so the orchestrator is "re-promoted and re-spawned every few seconds, burning a full agent run each cycle". And `find_deadlocked_children` selects `WHERE l.parent_id = ?`, so with no edge to invert it returns `[]` and `repair_inverted_dependencies` writes nothing — the backstop cannot rescue this shape.

You are also right that §6 of the same file already documents the shape that works, which made this an internal contradiction rather than just a wrong sentence. The bullet now points at it: parentless helpers, one fan-in card assigned to yourself with `parents=[<those ids>]`, then **complete** your current card — and it states explicitly why a dependency block is the wrong instrument here.

agents/platform/docs/glossary.md
```diff
@@ -62,6 +62,6 @@ This glossary defines key terms and concepts related to the Kubernetes Agentic H
...
 - **Exception — the Chat Agent:** The [Chat Agent](#chat-agent-agentschat-the-default-profile) is deliberately exempt from the pointer-only rule. As the conversational relay it passes full context to a specialist (in the `kanban_create` task `body`) and relays the specialist's real responses back to the user as they stream into the thread. The pointer-only rule still governs all specialist-to-specialist coordination.
 - **Note:** This replaced the earlier bespoke `worklog.py` shared-file store.
```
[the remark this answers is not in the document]
adamparco: Confirmed, and fixed in `fb2b112`.

Verified the contradiction directly: `agents/chat/SOUL.md:103` says the front door "must not try to relay, re-post, summarize, or acknowledge" a completion, and `agents/chat/AGENTS.md:22` says "Completions are not yours to repeat." Your escalation is also correct — `deploy/docker/Dockerfile:629` copies this file to `/opt/defaults/docs/`, and `agents/cluster/AGENTS.md:10` points workers at exactly that path, so this is runtime instruction, not site prose.

Both bullets now scope the exemption to the **outbound** direction only: the Chat Agent passes full context in the `kanban_create` body, and the gateway carries the return path without waking it.

One addition — the same claim had two more instances that this finding did not cover, and both are in #600 rather than here:

- `docs/site/src/content/docs/reference/glossary.md:16` — "relays progress and results back into the thread"
- `docs/README.md:184` — described `agents/chat/SOUL.md` as "relay results with attribution"

Fixed in the same terms on that branch, so the stack is internally consistent rather than half-corrected. Worth noting #600 came back "no findings"; its review body did caveat that it sampled rather than read the change exhaustively.

agents/chat/SOUL.md
```diff
@@ -35,9 +35,17 @@ Besides filing work, you are the user's window into the shared Kanban board. Whe
...
+
+A comment on a card that has already finished is worse than useless: it does not restart the work, the thread's subscription was deleted when the card closed, and nothing will ever read what you wrote. If `kanban_show` reports a terminal state (`done`, `blocked`, `gave_up`, or failed), do not promise delivery — answer from the card's `result`, which is already there.
+
 **Hard boundary.** Reading and these two updates (`kanban_comment`, `kanban_unblock`) plus delegation (`kanban_create`) are the ONLY kanban actions you take. Never call `kanban_complete`, `kanban_block`, `kanban_heartbeat`, or `kanban_link` — those belong to the specialist actually doing the work, not the front door. And never use board reads to _answer an infrastructure question yourself_ (cluster state, fleet data, best practices): those still go to a specialist via `kanban_create` per §1. Reading the board tells the user about their **tasks**; it does not turn you into a specialist.
```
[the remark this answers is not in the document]
adamparco: Confirmed and fixed in `ea1b9ce`.

Verified against source rather than taken as read: `TERMINAL_STATUSES` in `deploy/docker/patches/kanban_comment_status.py:136` is `frozenset({"done", "archived"})`, and `blocked` is not in it — so on a blocked card `comment_reaches` names the next worker and no terminal `note` is attached, exactly as you describe. `gave_up` and failure are event kinds the gateway raises, not statuses `kanban_show` hands back.

The internal contradiction is the strongest part of the finding and the reason this needed more than a word deleted. §1.5's **Unblock** bullet has the front door comment the user's answer onto a blocked card and then call `kanban_unblock` — so the paragraph as written forbade the one comment in the section that is guaranteed to reach someone, and redirected the agent to a `result` that a card parked on `needs_input` does not have.

The list is now `done` or `archived`. Rather than leave `blocked`'s absence to inference, the paragraph states that a blocked card is still open and still subscribed, and points back at the Unblock step — the silence there is what let the error in.

docs/site/src/content/docs/reference/security-and-iam.md
```diff
@@ -157,6 +157,7 @@ The agent never has direct write access to running infrastructure — see [Decla
...
 - **Bounded recovery.** The agent retries a blocker through its recovery ladder (roughly five iterations or ~10 minutes) before escalating to a human instead of looping indefinitely.
```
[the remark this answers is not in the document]
adamparco: Confirmed on every claim, and fixed in `06898d7`.

I checked each leg against source rather than taking the finding as read:

- `concepts/autonomous-watchdogs.md` does own the paragraph, and it does end with the sentence the copy dropped — "It is not total coverage: a pure-ASCII lookalike TLD and terminal escape injection get through both layers."
- `deploy/docker/patches/cron_tirith_scan.py:108-114`, under "Out of scope, recorded so it is not mistaken for covered": "`check_execute_code_guard` still auto-approves `execute_code` under `cron_mode: approve` … A cron run can therefore still reach `subprocess` through `execute_code` without a content scan."
- The reachability half holds too — `code_execution` is in `disabled_toolsets` in `agents/chat/config.yaml:182`, and the Platform Agent's config disables nothing of the sort. So the tool is live on exactly the profile the watchdogs run as.

The framing is the part worth agreeing with out loud: this is the page `AGENTS.md` makes canonical for what the agent may and may not do, so a bullet whose heading asserts coverage it does not have is worse here than the same sentence would be anywhere else. It read as a guarantee.

The rewrite does two things. It stops competing with the watchdogs page — that page is now named as canonical for how the scan behaves, and the mechanism (denial string, `Cron Tirith block: …` logging, ledger omission, `github-issue-resolver` reading attacker-controlled text) is no longer restated here at all. And it keeps only what this page is canonical for: the checks that survive `cron_mode: approve`, and the three ways coverage falls short — the lookalike TLD, terminal escape injection, and the unscanned `execute_code` path, with the Chat Agent's exemption noted so the gap is not read as fleet-wide.

`make docs-check` passes and prettier reports the file unchanged.

toshiowang:
## Review

**Range reviewed:** `ap-governance-cron-2...ap-agent-prose` (25 files, +237/−92).

**What I verified rather than took on faith:** ran `deploy/docker/patches/test_cron_tick_lock_scope.py` (15 pass) and `tests/test_default_profile_config.py` (32 pass, including both new cases), ran `make docs-check` (all four checks pass), and checked the load-bearing prose claims against source — `file`/`code_execution` in `agents/chat/config.yaml:178-186`, `wake_on_events` excluding `completed` at `agents/chat/config.yaml:163-167`, the 400-char first-line summary clip in `kanban_result_required.py:37`, `task_status`/`comment_reaches` in `kanban_comment_status.py:211-220`, `PLATFORM_AGENT_ROLE=sidecar` set only on the dashboard container (`platformagent_manifests.go:2179`), `GKE_PROJECT_ID`/`GCP_PROJECT_ID` injection (`:1383`, `:1401`, `:1892`), the 4-job chat roster and the 6 audit jobs + `github-issue-resolver`, and `mcp-agent_common`'s removal (asserted in `test_mcp_env_contract.py:163`). All of those check out.

The two code changes are correct. `claim()`'s new `except BaseException` is sound: `_lock_handle` closes the handle on its refusal path and hands it to `AdvisoryLock` on success, so a raise genuinely means the descriptor is still the caller's. The uuid staging name in `_atomic_write_yaml` removes a real zero-byte-read window and its cleanup path is right.

---

### [REVIEW RISK — not a code defect]

- **Location:** PR description ("Prose only — no code, no behaviour change, no rollout considerations.")
- **Problem:** The diff contains three functional changes and two test files that the description's file list omits entirely: `deploy/docker/patches/cron_tick_lock_scope.py` (`claim()` exception path), `deploy/shared/default_profile_config.py` (`_atomic_write_yaml` staging-name change), `deploy/docker/patches/verify_kanban_{notifier,result}.py`, plus `tests/test_default_profile_config.py` and `deploy/docker/patches/test_cron_tick_lock_scope.py`. Commits `4d0a5bc` and `423fd5c` are runtime fixes, not prose. A reviewer who reads "prose only" will skip exactly the files where a mistake would be expensive, and the "tree hash is byte-identical to `deploy-test`" claim predates those commits.
- **Fix:** Update the description's **Files** list and drop the "prose only" line, or split the code portions into a sixth PR.

---

### [DOC ACCURACY — instruction the agent will act on]

- **Location:** `agents/platform/SOUL.md` (§0, the `result`/`summary` rule), also `docs/site/.../concepts/chatops.md` and `overview/architecture.mdx`
- **Problem:** The new prose says `result` reaches the user "verbatim" and "in full" six times, and this PR simultaneously deletes the only length guidance the persona had (the 1200-character summary budget). `result` is in fact clipped at `RESULT_LIMIT = 30000` (`deploy/docker/patches/kanban_notifier.py:145`, applied in `result_block` at `:188-190`). The failure: a worker that reads "in full — not summarised away" and dumps a large fleet audit into `result` gets `[Result clipped at 30000 characters — ask for the full card to see the rest.]` and the tail is lost. The 400-character `summary` correction is right and well-sourced; the `result` side is the one now missing its number.
- **Fix:** Add the ceiling to the `result` paragraph, e.g. after "Whatever the card asked for goes there **in full**": `result` is delivered whole up to 30,000 characters, past which the notifier clips it with a marker — that is the threshold at which the "publish it and link it" guidance below stops being optional.

---

### [INTERNAL CONTRADICTION — persona instructions]

- **Location:** `agents/chat/SOUL.md` — §1.5's new rule vs. §2 step 4's mandated template
- **Problem:** §1.5 states, absolutely: "Before you comment on a card, unblock it, **or tell the user anything at all about its state** — including 'it's still running' or 'the results will post here shortly' — call `kanban_show(task_id)`. **Every time.**" §2 step 4 then mandates a reply template that does exactly that without a `kanban_show`: "The answer will post into this thread as soon as it's ready." The model resolves this one of two ways, both bad — it burns a `kanban_show` round-trip on every single delegation, or it learns the "every time" rule is soft, which is precisely the erosion the 2026-08-08 anecdote in the same section exists to prevent.
- **Fix:** Carve the acknowledgement out explicitly in §1.5, e.g. "The one exception is the acknowledgement in §2 step 4, spoken on the `kanban_create` return itself; from your next turn onward the rule is absolute."

A softer version of the same tension sits between §1.5's "answer from the card's `result`, which is already there" and §2 step 5 / `AGENTS.md`'s "must not relay, re-post, summarize, or acknowledge one." The scopes differ (user asked vs. woken), but a sentence saying so would remove the ambiguity.

---

### [MINOR — orphan accumulation, non-blocking]

- **Location:** `deploy/shared/default_profile_config.py:250`
- **Problem:** The uuid staging name is the right fix for the concurrent-writer window, but it changes the crash behaviour: a `SIGKILL` between `write_text` and `os.replace` (OOM kill, eviction, node preemption mid-boot) now leaves `config.yaml.<uuid>.tmp` on the PVC permanently, where the fixed name was reused by the next boot. Under a crashloop this accumulates one file per attempt. Nothing globs `$TARGET_DIR/*.yaml` — the two loops in `docker-entrypoint.sh:551,638` iterate `profiles/*` directories — so this is disk litter, not a correctness problem.
- **Fix:** Sweep stale `config.yaml.*.tmp` at the top of `rebuild()` (they are unreachable by construction once the uuid is gone), or leave it and note the tradeoff in the docstring.

---

Nothing here is a breaking bug — the code is correct and tested. The blocking items are the description/diff mismatch and the `result` ceiling, both of which are cheap to fix.

bradhoekstra (commented):
### Code review

Found 4 issues:

1. `agents/platform/SOUL.md` still opens by telling the Platform Agent that the Chat Agent relays its result — the pre-kanban contract this PR removes everywhere else, including further down the same file (outside this diff; see below).
2. `docs/site/src/content/docs/overview/architecture.mdx:92` keeps "Other SOPs post to the configured Chat channel instead" on a line this PR edited, contradicting the canonical `overview/proactive-autonomy.md` and the now-empty set of non-audit SOPs.
3. `agents/cluster/SOUL.md:79` was rewritten for the `result` contract but kept "relays results to the user", which the next paragraph in the same file contradicts.
4. The description says "Prose only — no code, no behaviour change", but the diff changes runtime behaviour in `deploy/shared/default_profile_config.py` and `deploy/docker/patches/cron_tick_lock_scope.py` under a `docs(agents):` title.

#### `agents/platform/SOUL.md` line 3 still describes the Chat Agent as the relay

The rewritten §0 in this same file now says the gateway posts `result` "into the user's chat thread verbatim, so it is the only thing that reaches the person who asked", and `agents/chat/SOUL.md` §2 step 5 plus the glossary entry both say the Chat Agent is not woken for completions and must not relay one. The untouched opening paragraph still says the Chat Agent "routes work to you with full context, and you return your result for it to relay to the user". Before this PR the two agreed; after it they do not, in a file loaded verbatim as the Platform Agent's system prompt.

https://github.com/gke-labs/kube-agents/blob/c6e0a335617b60f65b5598fbae11fc033849dbc2/agents/platform/SOUL.md#L3-L3

Suggested rewrite of the clause:

```text
... the front-door Chat Agent (the `default` profile) routes work to you with full
context on a kanban card, and your card's `result` is what reaches the user.
```

(`kube-agents-bot` raised this same point in its 2026-08-10 review and it is still open at `c6e0a33`.)

docs/site/src/content/docs/overview/architecture.mdx
```diff
@@ -82,16 +82,16 @@ flowchart TB
...
-4. The five fleet audits hand their findings to the `fleet-audit` skill, which renders and maintains one GitHub **issue** per audit stream — the ledger — and promotes findings with a mergeable manifest into narrow per-finding remediation PRs linked back to it. A run with no findings closes the ledger as completed and returns `[SILENT]` — unless it could not read the whole fleet, in which case the ledger stays open and the gaps are reported, or it resolved findings on the way, in which case it says what closed. Other SOPs post to the configured Chat channel instead.
+4. The six fleet audits hand their findings to the `fleet-audit` skill, which renders and maintains one GitHub **issue** per audit stream — the ledger — and promotes findings with a mergeable manifest into narrow per-finding remediation PRs linked back to it. A run with no findings closes the ledger as completed and returns `[SILENT]` — unless it could not read the whole fleet, in which case the ledger stays open and the gaps are reported, or it resolved findings on the way, in which case it says what closed. Other SOPs post to the configured Chat channel instead.
```
bradhoekstra: 🟠 **Medium** — The trailing sentence "Other SOPs post to the configured Chat channel instead" survived the `five` → `six` edit, and it is now wrong twice over.

- `agents/platform/cron/jobs.json` ships seven jobs: the six audits named here plus `github-issue-resolver`, which is a skill run, not an SOP. The five non-audit SOPs (`blueprint-sync`, `policy-propagation`, `global-capacity-orchestrator`, `standardization-validator`, `lifecycle-deprecation-manager`) are "now gone from the roster entirely" per `concepts/autonomous-watchdogs.md`, so there are no "other SOPs" left to post anywhere.
- It also contradicts the canonical page for this topic: `overview/proactive-autonomy.md` states "The loop ends at the repo, not in chat … the report itself is never posted to a chat channel" — which is the same correction this PR makes in `concepts/chatops.md` ("A governance watchdog's findings are not on this path").

AGENTS.md: "Link rather than summarise when another page already owns the topic." Dropping the sentence is enough.

```suggestion
4. The six fleet audits hand their findings to the `fleet-audit` skill, which renders and maintains one GitHub **issue** per audit stream — the ledger — and promotes findings with a mergeable manifest into narrow per-finding remediation PRs linked back to it. A run with no findings closes the ledger as completed and returns `[SILENT]` — unless it could not read the whole fleet, in which case the ledger stays open and the gaps are reported, or it resolved findings on the way, in which case it says what closed.
```

agents/cluster/SOUL.md
```diff
@@ -72,9 +73,11 @@ Your loop:
...
 5. **Acknowledge only:** your final chat reply is a brief ack. Do not put the RCA or patch in the reply — the card is the channel.
 
-The Platform Agent reads your completed card (its `summary`/`metadata`), relays results to the user, and owns any remediation (Pull Requests via `submit-suggestion`).
+The Platform Agent reads your completed card (its `result`, plus `metadata` for the patch), relays results to the user, and owns any remediation (Pull Requests via `submit-suggestion`).
```
bradhoekstra: 🟠 **Medium** — This line was updated for the `result` contract but kept `relays results to the user`, which is the pre-kanban relay language the PR removes elsewhere (`agents/chat/AGENTS.md`: "Completions are not yours to repeat"; `agents/chat/SOUL.md` §2 step 5).

It also contradicts the paragraph immediately below it, which this PR leaves in place: "Your own task's completion already reaches the user's chat thread (the Platform Agent subscribed your card when it delegated to you)." With `kanban_notify_propagate.py` having copied the subscription, the gateway posts this card's `result` into the thread verbatim — there is nothing for the Platform Agent to relay, and a relay would be the duplication `agents/platform/SOUL.md` §0 now warns against.

```suggestion
The Platform Agent reads your completed card (its `result`, plus `metadata` for the patch) and owns any remediation (Pull Requests via `submit-suggestion`). It does not re-post your RCA — your completion already delivered `result` into the user's thread verbatim.
```

deploy/shared/default_profile_config.py
```diff
@@ -233,11 +234,28 @@ def build_baseline(image_path, overlay_paths):
...
+    Sharing one staging path across those writers is what turns a survivable repeat into
+    the failure this docstring warns about: the loser's `os.replace` fires on a name the
+    winner already renamed away, and a reader between the two sees config.yaml at zero
+    bytes. Per-writer names cost nothing and make the repeat harmless.
+    """
     path.parent.mkdir(parents=True, exist_ok=True)
-    tmp = path.with_name(path.name + ".tmp")
-    tmp.write_text(yaml.safe_dump(data, sort_keys=True))
-    os.replace(tmp, path)
+    tmp = path.with_name(f"{path.name}.{uuid.uuid4().hex}.tmp")
```
bradhoekstra, line 250: 🟠 **Medium** — Scope/description mismatch. The PR body says "Prose only — no code, no behaviour change, no rollout considerations" and its **Files** list names only personas, skills, and site pages. This hunk is a real runtime behaviour change to how the front door's `config.yaml` is written (per-writer uuid staging name + unlink-on-failure), and it is not the only one: `deploy/docker/patches/cron_tick_lock_scope.py` now closes the handle when `_lock_handle` raises, and `tests/test_default_profile_config.py` gains 44 lines of coverage for both.

AGENTS.md asks to "Keep changes scoped to the request" and to use the PR template for "level of detail"; a `docs(agents):` Conventional Commit carrying an atomic-write change means a reviewer scanning the title and the "prose only" footer will not look at it. Either move these into the runtime PR in the stack, or list them in **What Changed** and drop the "no code, no behaviour change" line so the change is reviewed on its merits.

(The change itself reads correct — `os.replace` onto a shared `config.yaml` stays atomic, and `missing_ok=True` keeps the cleanup safe.)

docs/site/src/content/docs/overview/architecture.mdx
adamparco: Confirmed, and fixed in `9354d76` — the trailing sentence is gone.

Checked it three ways rather than taking the diff's word for it, because "other SOPs" is the kind of claim that is easy to leave standing:

- `agents/platform/defaults/cron/jobs.json` — the roster is six audits plus `github-issue-resolver`. There is no job on it that posts to a Chat channel, so the sentence describes an empty set.
- `docs/site/.../concepts/autonomous-watchdogs.md:54` says the channel-posting SOP is "now gone from the roster entirely".
- `docs/site/.../concepts/proactive-autonomy.md:18` says results are "never posted to a chat channel".

So the page was contradicting two other pages as well as the roster.

agents/cluster/SOUL.md
adamparco: Confirmed, and fixed in `9354d76`.

You are right that it contradicts itself — and it does so against the paragraph directly beneath it, which is what makes it worth catching. The line now reads:

> The Platform Agent reads your completed card (its `result`, plus `metadata` for the patch) and owns any remediation (Pull Requests via `submit-suggestion`). It does not relay your `result` onward — the gateway has already posted it into the user's thread verbatim, which is why step 4 insists the whole RCA lives there and reads well as written.

One thing worth flagging, since it affects how you read the rest of the stack: the correct wording already existed on `ap-result-delivery`, the branch downstream of this one. The fix had landed on the wrong PR in the stack, so this branch was the only place still carrying the old text. I back-ported the downstream wording rather than inventing a third phrasing, so the two branches now say the same thing.

That also means the same sweep was worth doing properly, and it turned up two more of the same class that neither of us had flagged:

- `agents/platform/SOUL.md` §132 (single-cluster delegation, step 4) said "then relay/act on it" — the relay half being exactly the instruction this stack removes.
- `agents/chat/AGENTS.md:23` pointed at "the relay format in `SOUL.md` §2", which is now the attribution format.

Both are fixed in the same commit.

deploy/shared/default_profile_config.py
adamparco: Confirmed — the PR body was wrong, and I have corrected it rather than the diff.

The measurement, for the record: against `upstream/ap-governance-cron-2`, this branch changes 7 non-Markdown files, +120/−26.

```
deploy/docker/patches/cron_tick_lock_scope.py      | 12 ++++-
deploy/docker/patches/test_cron_tick_lock_scope.py |  5 ++
deploy/docker/patches/verify_kanban_notifier.py    |  3 +-
deploy/docker/patches/verify_kanban_result.py      |  3 +-
deploy/shared/default_profile_config.py            | 26 +++++++++--
deploy/shared/docker-entrypoint.sh                 | 53 ++++++++++++++--------
tests/test_default_profile_config.py               | 44 ++++++++++++++++++
```

These are the non-prose remainder of the commits the body already listed under **Context**; being the last PR in the stack, this is where anything not claimed by PRs 1–4 came to rest. The body now lists them explicitly under **What Changed** with a line on what each one is, and the "Prose only — no code, no behaviour change" trailer is gone.

I would rather not move the code out into a sixth PR at this point: the stack's stated invariant is that the tip's tree hash matches `deploy-test` byte for byte, and re-splitting now would churn four base branches to fix a sentence. But the sentence was the actual problem — a reviewer who trusted it would have skipped a `docker-entrypoint.sh` change on the strength of the description, which is precisely the failure you are pointing at.

adamparco:
Thanks — all four confirmed, all four addressed in `9354d76` (pushed). Replies to 2, 3 and 4 are in their own threads; this covers finding 1, which had no inline anchor.

**1. `agents/platform/SOUL.md` line 3 still described the Chat Agent as the relay — confirmed.**

You are right that the file contradicted itself, and right that it is the worst place for it: this file is loaded verbatim as the Platform Agent's system prompt, so the two readings were live instructions, not documentation.

I did not take the suggested rewrite, for one reason worth surfacing. The wording you wanted already existed — on `ap-result-delivery`, the branch immediately downstream of this one. The fix had simply landed on the wrong PR in the stack, so this branch was the last place still carrying the old clause. Rather than introduce a third phrasing, I back-ported the downstream text so the branches converge:

> ... routes work to you with full context, and your card's `result` posts into the user's thread verbatim. Nothing relays, re-renders or summarises it on the way out, so it has to arrive complete and readable exactly as you wrote it (§0).

It says the same thing as your suggestion and additionally states the consequence — "so it has to arrive complete and readable" — which is the part the agent actually needs to act on.

Since the point was that the file disagreed with itself, I swept the rest of the personas for the same class rather than fixing only the reported line. Two more, both fixed in the same commit:

- `agents/platform/SOUL.md` §132, single-cluster delegation step 4: "then relay/act on it". The relay half is the instruction this stack removes. It now says step 3 has already put the worker's `result` in the user's thread, so the agent must not repeat it back.
- `agents/chat/AGENTS.md:23`: "see the relay format in `SOUL.md` §2" → "the attribution format".

Validation: `make docs-check` passes, and `prettier --check` passes on all four changed files.

toshiowang (approved):

PR #602, opened 2026-08-09
fix(agents,runtime,cron): land the ap-* stack — agent runtime, governance cron, prose, and result + notification delivery

adamparco:
# Pull Request

**Retargeted at `main`. This pull request now carries the whole `ap-*` stack.**

The stack was reviewed and merged one pull request at a time, but after #598 landed on `main`
the "retarget the next one at `main`" step was missed. Each subsequent pull request merged
into the *mirror branch* below it instead, and those branches are dead ends — none has a path
to `main`. Four approved-and-merged pull requests therefore delivered no content:

| Pull request                  | Merged into                                                        | Reached `main`? |
| ----------------------------- | ------------------------------------------------------------------ | --------------- |
| #597 `ap-build-ci`            | `main`                                                               | yes             |
| #598 `ap-image-patch-surface` | `main`                                                               | yes             |
| #599 `ap-agent-runtime`       | `ap-image-patch-surface`, a day after that branch had already merged | **no**          |
| #600 `ap-governance-cron`     | `ap-agent-runtime`                                                   | **no**          |
| #601 `ap-agent-prose`         | `ap-governance-cron-2`                                               | **no**          |
| #604 `ap-notification-delivery` | `ap-result-delivery` (the upstream mirror, not this branch's head) | **no**          |

This pull request is now the only carrier for #599, #600, #601, #604 and its own work. Its
diff against `main` is correspondingly large. **The content has already been reviewed and
approved in the five pull requests named above and is unchanged here**, apart from the
conflict resolutions listed under [Merge resolution](#merge-resolution) — which is where
review attention is best spent.

`ap-build-ci`, `ap-image-patch-surface`, `ap-agent-runtime`, `ap-governance-cron`,
`ap-governance-cron-2`, `ap-agent-prose` and `ap-result-delivery` carry no work beyond the
pull requests that proposed them and should all be deleted once this lands.

## Why This Change

Each constituent pull request states its own rationale in full; in one line each:

- **#599 — the agent pod did not boot deterministically.** Every replica believed it owned
  bootstrap, so profile scaffolding and config merges raced on the shared volume, and an
  image upgrade was silently discarded whenever the volume already existed.
- **#600 — the governance watchdogs ran from the wrong agent.** Fleet governance hung off the
  Chat Agent through a 600-line dispatcher and one hand-written shim per job, and the
  documented schedule was prose that nothing checked against the roster. The fleet audit also
  had no coverage of AI workloads.
- **#601 — the personas described pre-kanban behaviour.** In a repository whose prose *is* the
  agent's instructions, that is not cosmetic: agents were being told to do something the
  harness no longer permits.
- **#602 (this one) — a card's `result` had no rendering contract.** Nothing told a worker to
  write Markdown, so reports arrived as ASCII art that Block Kit cannot structure; identical
  questions came back in four different shapes; the credential proxy was returning ~9,500
  failed Slack calls a day from the directory refresh; and no scheduled job on a named
  profile had ever posted its report.
- **#604 — the system produced answers nobody saw.** Cron briefs could not reach Slack in the
  credential-proxy posture this repository recommends, briefs that did send buried themselves
  in a stale thread, and kanban notification was at-most-once — a real report (`t_a18254ca`,
  7,124 bytes) was lost to a crash 162 ms after its cursor advanced.

Landing them together is not a new decision about scope; it is the only remaining way to get
the approved work onto `main` without re-reviewing it five times.

## What Changed

**Files:** 137 changed, +13,305 / −2,623. By constituent pull request:

- **#599** — `deploy/shared/docker-entrypoint.sh`, `deploy/shared/defaults/config.yaml`,
  `deploy/docker/Dockerfile`, `agents/{chat,cluster,platform}/config.yaml`, `k8s-operator/`
  (controller, CRD, goldens, LiteLLM), `charts/kube-agents/`, `agentplugins/`, plus new
  `test_image_startup_contract.py` and `test_merge_configs.py`.
- **#600** — `agents/chat/scripts/profile_cron_tick.py` and `kanban_board_health.py` (new,
  replacing `platform_cron_dispatch.py` and its six shims), `agents/platform/cron/`,
  `agents/platform/governance/ai_security_audit_sop.md` (new) and the five existing SOPs,
  `agents/platform/skills/fleet-audit/`, `scripts/generate_docs.py`.
- **#601** — the Chat/Cluster/Platform personas and three skills, `docs/designs/`,
  `docs/site/` concept and reference pages, plus `default_profile_config.py` and
  `cron_tick_lock_scope.py` with their suites.
- **#602** — `deploy/docker/patches/kanban_report_format.py` and
  `channel_directory_threads.py` (each with its `apply_`/`verify_`/`test_` triple),
  `kanban_result_required.py`, `kanban_notifier.py`, `profile_cron_tick.py`, the four persona
  files.
- **#604** — `agents/platform/scripts/slack_relay_patch.py`,
  `deploy/docker/patches/kanban_notify_delivery.py` (with its triple),
  `profile_cron_tick.py`, `docs/site/.../concepts/chatops.md`.

**Added/Changed/Fixed:** see each pull request above — the descriptions are accurate and
nothing has been added to them here.

### Merge resolution

`main` moved while the stack sat on its mirror branches, so this branch merges `main` in.
Three files conflicted in the #604 recovery and 45 in the `main` merge, 78 hunks in total.
Most were whitespace or adjacent-line edits. These are the resolutions that are **not** a
straight take-one-side, and they are the parts of this pull request that have not been
reviewed before:

- **`stockout-prevention` would have been silently deleted.** #581 added the job to `main`,
  as a `dispatch_*.py` shim on the Chat Agent's dispatcher — the exact file #600 removes. The
  mechanical union therefore left the job on *neither* roster and the audit would simply have
  stopped firing, with no error anywhere. It is restored to
  `agents/platform/cron/jobs.json` in schedule order (`20 9 * * *`, between
  `ai-security-audit` and `github-issue-resolver`), without `main`'s `no_agent`/`script` shim
  keys, which the roster format has no use for. Its line-number references were re-verified
  against `stockout_prevention_sop.md` as it stands on `main`.
- **`platform_cron_dispatch.py` and its test resolve as delete** (modify/delete: the stack
  removes them, `main` edited them). `dispatch_stockout_prevention.py` is deleted with them —
  it is a two-line shim into the deleted module, referenced nowhere else.
- **`docker-entrypoint.sh` `--cron-retire`.** `stockout-prevention` is retired from the
  default profile's roster and shipped on the Platform Agent's instead — the same move the
  other seven ids made one release earlier, applied late because the job landed after the
  move was written. `merge_cron_store` never prunes, so an upgraded PVC would otherwise keep
  ticking the old entry.
- **`kanban_report_format.py` and its two test modules resolve to the stack's side.** The
  three conflicts are comments describing the shape *gate*, which `9d0ce49` replaced with a
  warn-only detector, so `main`'s prose describes behaviour this branch no longer has.
  `main`'s side also carried a call to `gate_defects`, renamed here to `serious_defects`,
  which git auto-merged into a latent `NameError`; the affected test class is taken from the
  stack's side whole.
- **`SLACK_HOME_CHANNEL` is added to the `platform_control` `env:` block — a new fix, and the
  only behaviour change in this merge that is not in any constituent pull request.** `main`
  taught `send_notification` to target Slack, reading that variable for the channel, but
  Hermes hands a stdio MCP child only the keys its own block names, so the read always
  yielded `""` and every Slack notification fell back to a channel-less `slack` target — the
  same "No home channel set" failure the Google Chat entry three lines above already
  documents. The operator does set the value on the container
  (`platformagent_manifests.go:1534`); only the allowlist entry was missing. This is a latent
  bug on `main`, and the first thing the stack's new `test_mcp_env_contract.py` has caught
  that it was not written for.
- **`platformagent-telemetry.yaml`** regenerated with `go test -update` and then
  prettier-formatted, matching how the other goldens on `main` are maintained. The 18/11-line
  change is #599's: the overlay ConfigMap key, the router's `${HERMES_HOME}` path,
  wake-on-events, and the raised CPU/memory.

#### Second `main` merge (#651, #657)

`main` moved again after this pull request was retargeted, bringing lazy MCP registration
(#651) and the restructured documentation map (#657). Four files conflicted:

- **`lazy: true` was added independently on both sides**, so every server was a near-miss
  duplicate YAML key — a duplicate would have been accepted by the parser and silently kept
  the last one. Resolved to a single occurrence per server; the build-time merged config is
  verified to carry exactly three servers, each lazy.
- **`agent_common` stays deleted.** #651 added `lazy: true` to that server in the shared
  defaults, and #599 removes the server outright. Taking `main`'s side would have resurrected
  it — the same shape as the `stockout-prevention` near-miss above.
  `AgentCommonIsNotMountedTest.test_the_server_is_not_declared` is what pins this.
- **#651's measurements predate that removal.** "Four servers and 44 tools" is three and 43
  here — `agent_common_server.py` exposes exactly one tool, `call_agent`. Both copies of the
  prose keep the measured figures and say what they were measured against, rather than
  restating a count that is no longer true.
- **`connect_timeout` / `timeout` on `developer_knowledge` and `gke`** are #599's mcp-remote
  backstops and survive the merge; `main`'s side had neither. The config reference's timeout
  paragraph folds `main`'s lazy caveat into this branch's version: `connect_timeout: 120` now
  bounds the first tool call rather than startup.
- **`docs/README.md` takes #657's restructured map wholesale** — that is the point of #657 —
  with this branch's four semantic edits re-applied on top (both cron rosters as the generated
  region's source, the same pair on the identifier row, and new rows for image-patch module
  names and the shared agent defaults), plus the inventory row for
  `agents/platform/cron/README.md`, which is new on this branch.
- **The config reference's `env:` snippet had drifted four keys short** of the block it claims
  to mirror; synced, which is also where this branch's `SLACK_HOME_CHANNEL` entry belongs.
  `agent_common` is dropped from the sentence above it: `main` rewrote a line this branch had
  left at its base version, so the merge quietly took a claim that the server still ships from
  the shared defaults.

## Why This Matters

- Four approved pull requests are currently sitting in branches with no path to `main`. Until
  this merges, none of that work exists for anyone who deploys from `main`.
- The `stockout-prevention` resolution is the reason a mechanical union was not good enough
  here: the failure mode is a watchdog that stops running and reports nothing, which is
  indistinguishable from a clean fleet.
- The `SLACK_HOME_CHANNEL` gap and the `gate_defects` `NameError` are both defects that only
  exist *because* the two sides evolved apart. Neither would have been found by reviewing
  either side alone.

## Context

- Supersedes the stacked arrangement described in #597–#604. **This pull request is the only
  one of the six still open**; #604's content is recovered here by merging
  `gke-labs:ap-result-delivery` (`2fd107c8`) rather than by rebase, because #604
  squash-merged at a parent three commits behind this branch's head and a fast-forward would
  have dropped it.
- Merges `upstream/main` twice: once at `edcd6b5` (where the conflict work was done) and once
  at `dc695ce` to pick up #650, which auto-merged clean.
- The seven `ap-*` branches should be deleted once this lands.

## Testing

- `make test-python` — **16 suites, 2,075 tests, 0 failures** (2 skips).
- `go build ./...` and `go test ./...` in `k8s-operator/` — pass, all five packages. Goldens
  regenerated with `-update`, then prettier-formatted.
- `make docs-check` — pass: generated regions current, 235 Markdown files' relative links all
  resolve, terminology check passes, documentation map covers all 208 tracked docs.
- `make docs-generate` — no drift.
- `prettier --check` on all 70 changed Markdown/JSON/YAML files — clean. (Repository-wide
  `make prettier-check` reports 19 pre-existing failures under `.agents/skills/`; this branch
  does not touch those files and, per AGENTS.md, they are not reformatted here.)
- The merge was validated by re-running the whole suite after resolution, not only the files
  that conflicted — which is how the `SLACK_HOME_CHANNEL` gap surfaced.

### Live validation

**The merge itself has not been exercised against a running installation.** No image was
built from this tip and no cluster was rolled onto it. What follows is what the constituent
pull requests validated live, plus what remains unproven — stated plainly rather than
inherited by implication.

Validated live on-cluster, before the merge:

- **#604, Slack delivery half** — image `dev-317006e-0809-120553`. The profile's platforms
  went `['api_server']` → `['api_server', 'slack']` with `enabled: True` and
  `token set: False`, proving the relay path rather than a local token; `deliver=all`
  resolved to a real DM target; a send returned
  `{'success': True, 'message_id': '1786292815.485629'}` and arrived flat in the DM rather
  than threaded. Reverting to the pre-patch image reproduced
  `SLACK_BOT_TOKEN not configured`, so the mechanism is proven and not a coincidence.
- **#602's three defects were measured on the live cluster, not inferred** — card
  `t_3ba2166a`'s 2,213-character ASCII report rendering to three Block Kit blocks against
  nine for the same content in Markdown; the four-shape fan-out of 2026-08-08; and the
  ~9,500/day `channel_not_found` bursts from the directory refresh.
- **#604's kanban incident is a real trace** — task `t_a18254ca`, event id 971,
  cursor 958 → 971, gateway fatal 162 ms later, the 7,124-byte report intact in
  `tasks.result` and never delivered.

Not live-tested, and why:

- **#599's multi-replica boot election** — needs a deliberate multi-replica rollout on a
  fresh and then a pre-existing volume. Covered by `test_image_startup_contract.py` only.
  This is the highest-risk item in the pull request.
- **#600's watchdogs against a real fleet** — a full audit run costs a day of schedule to
  observe. Covered by unit tests over the SOP, cron and rendering-cap contract.
- **#604's kanban at-least-once half** — was never in a deployed image; covered by tests plus
  `verify_kanban_notify_delivery.py`, which replays the `t_a18254ca` incident against a real
  board.
- **The new `SLACK_HOME_CHANNEL` declaration** — proving it needs a Slack-configured install.
  To verify after rollout: `hermes` MCP child env for `platform_control` should now contain
  `SLACK_HOME_CHANNEL`, and `send_notification` on a Slack-enabled profile should produce a
  `slack:<channel>` target instead of a bare `slack`. Setting the channel to a distinctly
  wrong id should route there, confirming the value is read rather than defaulted.
- **A `docker build` of this tip.** #602's tip built clean (92 steps, both new modules'
  `apply_`/`verify_` running inside the build); the merge adds no new patch module, but the
  build has not been re-run since.

No test artifacts were created, so none needed cleaning up.

---

Functional impact is the union of five reviewed pull requests, and the risk profile is
theirs: #599 changes pod startup ordering and the operator's rendered manifests and is the
riskiest; #604's kanban half converts a single-ownership claim into at-least-once delivery
and is safe only while the notifier is the single writer of those cursors (`replicas: 1`,
`strategy: Recreate` — raising `availability.replicas` becomes a correctness change, and that
constraint is recorded in the module docstring). Rollout is an agent image plus an operator
upgrade. The merge-specific changes are small and additive: one restored cron entry, one env
allowlist key, three deletions of code nothing references.


deploy/docker/patches/kanban_report_format.py
```diff
@@ -0,0 +1,337 @@
...
+        return ()
+    if len(body) < min_chars:
+        return ()
+
+    defects: list[str] = []
+    if _H1.search(body):
+        defects.append("top-level-heading")
+    if not _BLOCK_MARKDOWN.search(body) and _ASCII_STRUCTURE.search(body):
+        defects.append("ascii-substitute")
+    if _has_heading_without_prose(body):
```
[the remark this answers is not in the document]
adamparco: Confirmed and fixed in `2427d69`, with one deliberate difference from the suggested edit.

Reproduced first: a report with an `##` heading, a paragraph of prose, and a fenced `# list the pods that moved` comes back `('top-level-heading',)` from `result_shape_defects` and is refused by `gate_defects`. Your read of the cost is right too — `DEFECT_ADVICE["top-level-heading"]` tells the worker to use `##` instead, which is unfollowable when the hash is inside a fence Slack renders verbatim, so the one-nudge budget is spent on an edit that cannot be made.

The difference: the fix is `_H1.search(_FENCE.sub("", body))`, not `_strip_code(body)`. Stripping inline code as well introduces the mirror-image bug. An H1 has to open a line, so it can never hide inside a backtick pair — but removing one lets a line such as

    `--dry-run` # prints the plan and exits

collapse onto its hash and match `^ {0,3}# +\S`, inventing the very defect this check exists to catch. `_has_unquoted_numerics` needs both strippers because a backticked value is precisely what it must not count; the H1 scan needs only the fence.

Both directions are covered by new tests in `test_kanban_report_format.py` — the fenced-comment one fails against the old expression with `'top-level-heading' unexpectedly found in ('top-level-heading',)`. The 674-test patch suite and `make test-python` are green.

Confirming your closing note as well: `ascii-substitute` is unaffected, since a fence is part of `_BLOCK_MARKDOWN` and a body containing one never reaches the `_ASCII_STRUCTURE` check.

bradhoekstra (commented):
Review of `ap-result-delivery`, verified against a worktree at head `3b354d9`. All existing suites pass (486 patch tests, 47 cron-tick tests); the nine inline findings below are things the tests do not cover.

**Summary:** one High (an unguarded config read that can fail every cron tick), four Medium (a warning that fires on cosmetic defects, a gate that refuses raw manifests, a refusal that discards a complete report, and a Slack-only home-target map that leaves the reported bug open on Google Chat), and four Low.

**Checked and clear:** `deploy/shared/docker-entrypoint.sh` is comment-only; the `_atomic_write_yaml` uuid staging plus `unlink(missing_ok=True)` cleanup is correct; the `cron_tick_lock_scope` `BaseException` handle-close is correct; the Dockerfile layer accounting (-2 COPY in the kanban block, +2 for channel-directory) nets to zero, and the `apply_kanban_report_format` -> `apply_kanban_result_required` ordering requirement is satisfied by copying both modules into `/opt/hermes/tools/` before the RUN.

agents/chat/scripts/profile_cron_tick.py
```diff
@@ -214,7 +214,69 @@ def _rotate(log_path: Path) -> None:
...
+        import yaml
+
+        config = yaml.safe_load((root_home / "config.yaml").read_text()) or {}
+    except Exception:
+        # Never the reason a tick does not happen: a job with an explicit
+        # target still delivers, and `deliver=all` degrades to today's
+        # behaviour rather than to a crash.
+        return {}
+
+    platforms = config.get("platforms") or {}
```
bradhoekstra: 🔴 **High — `home_target_env` only guards the YAML parse, so a structurally-odd config crashes every tick.**

The `try` ends at line 257; lines 263-272 run unguarded. If `platforms`, `platforms.slack`, or `home_channel` is any non-mapping, `.get()` raises `AttributeError`, which propagates out of `main()` (nothing catches it at line 345) and the `profile-cron-tick` job fails — so *no* profile ticks at all, every minute. That is strictly worse than the delivery bug being fixed, and it contradicts the docstring's promise that "a config that is missing, empty, or corrupt returns no keys rather than raising".

```
platforms:\n  slack:\n    home_channel: C0FFEE   -> AttributeError: 'str' object has no attribute 'get'
platforms:\n  - slack                            -> AttributeError: 'list' object has no attribute 'get'
```

Note that `tests/test_default_profile_config.py:341`, added in this same PR, writes exactly the first shape (`home_channel = "C0FFEE"`). Widen the `try` to cover the whole function body, or `isinstance`-check each level.

deploy/docker/patches/kanban_notifier.py
```diff
@@ -191,6 +197,98 @@ def result_block(
...
+    second package on the delivery path's import graph for the sake of a log
+    line. When the import fails we still report the one defect this module can
+    detect on its own.
+    """
+    try:
+        body = str(result).strip() if result is not None else ""
+        try:
+            from tools.kanban_report_format import result_shape_defects
+
+            defects = result_shape_defects(result, min_chars=UNSTRUCTURED_MIN_CHARS)
```
bradhoekstra, line 287: 🟠 **Medium — in the image the delivery-path warning fires on cosmetic, warn-only defects.**

`_warn_if_unstructured` calls `result_shape_defects`, which returns all four defects — including `unquoted-numerics` and `heading-without-prose` — and then logs at WARNING that the result "will not render well in chat". Reproduced with `tools/` on the path:

```
WARNING:gateway.run:[kanban] card t_x completed with a 189-character result whose formatting
will not render well in chat: heading-without-prose, unquoted-numerics.
```

That report renders fine; it just has bare floats. Any result >=150 chars with two 6-digit numbers outside backticks now warns on every completion. The module docstring says the note fires only when "a long result carries no block-level Markdown but does use an ASCII section marker … All three conditions are required", which is true only of the fallback path. Suggest passing `GATED_DEFECTS` (or a dedicated warn-set) rather than the full list, or dropping the log to INFO/DEBUG for non-gated defects.

deploy/docker/patches/kanban_report_format.py
```diff
@@ -0,0 +1,343 @@
...
+        return ()
+
+    defects: list[str] = []
+    # Fenced code is removed first: a shell comment inside a code block is not a
+    # heading, and this defect is gated, so reading one as an H1 refuses a
+    # well-shaped report for an edit the worker cannot make. Inline code stays —
+    # an H1 has to open a line, so it can never hide inside a backtick pair, and
+    # removing one would let ``\`--dry-run\` # note`` collapse onto its hash and
+    # read as a heading nobody wrote.
+    if _H1.search(_FENCE.sub("", body)):
```
bradhoekstra, line 321: 🟠 **Medium — `top-level-heading` is gated and reads a `#` comment at column 0 as an H1.**

Only fenced code is stripped. A `result` that is a raw manifest or a diff — which is exactly what a card saying `Output format: raw YAML, no prose` asks for, and what `has_format_directive` deliberately suppresses the stanza for — is refused:

```
YAML with "# the replica count was raised from 2 to 4"   -> gate_defects(...) == ('top-level-heading',)
diff with "# bump the memory request so the OOMKills stop"  -> ('top-level-heading',)
```

The refusal then tells the worker to "Use `##` or `###` instead", which is wrong advice for a manifest and unactionable — the worker's only real fix (wrap it in a fence) is not mentioned. Consider requiring that the H1 be the first non-blank block-level line, or skipping the gate when the body carried a format directive.

deploy/docker/patches/kanban_result_required.py
```diff
@@ -195,16 +230,33 @@ def require_result(
...
+
+    if not nudge_spent:
+        # Never let a shape check throw on the completion path: a card that
+        # cannot close because a regex raised would be a far worse bug than the
+        # formatting it was inspecting.
+        try:
+            defects = _gate_defects(result)
+        except Exception:  # pragma: no cover - defensive
+            defects = ()
+        if defects:
```
bradhoekstra: 🟠 **Medium — a shape refusal withholds a complete answer, and the retry is not guaranteed.**

Unlike the empty-result gate, this one returns `(error, None)` for a `result` that already contains the whole report. `kanban_complete` is typically a worker's last action; if it has exhausted `max_turns`, or crashes, or the run is reaped between the refusal and the retry, the card stays open with no `result` and the requester gets nothing — where before they got an ugly-but-complete report. Worth considering storing the result and returning the advice as a non-fatal note, or restricting the gate to defects where the delivered text is genuinely unreadable.

agents/chat/scripts/profile_cron_tick.py
```diff
@@ -214,7 +214,69 @@ def _rotate(log_path: Path) -> None:
...
 
 
-def spawn_tick(binary: Path, profile_home: Path) -> subprocess.Popen:
+# Home-target routing keys, per platform: (chat id, thread id).
+#
+# ``deliver=all`` resolves through ``cron/scheduler.py``'s
+# ``_get_home_target_chat_id``, which reads ``os.getenv`` and nothing else — it
+# never consults ``config.yaml``. So the child needs these in its environment
+# or it has nowhere to post.
+HOME_TARGET_ENV_KEYS = {
```
bradhoekstra, line 231: 🟠 **Medium — `HOME_TARGET_ENV_KEYS` covers Slack only.**

`agents/chat/config.yaml:93` pins Google Chat as "the real ingress key", and `agents/platform/scripts/platform_mcp_server.py:564` already reads `GOOGLE_CHAT_HOME_CHANNEL`. On a Google Chat deployment `home_target_env` returns `{}` and `deliver=all` still resolves to nowhere — the bug this PR reports as fixed. The dict is clearly built for extension; adding the `google_chat` entry (and reading `platforms.google_chat.home_channel`) would close it.

deploy/docker/patches/apply_channel_directory_threads.py
```diff
@@ -0,0 +1,161 @@
...
+        client = next(iter(team_clients.values()))
+        for _channel_id, _entries in unresolved:
+            try:
+                resp = await client.conversations_info(channel=_channel_id)
+                if not resp.get("ok"):
+                    _note_unresolvable(_channel_id, resp.get("error", "not ok"))
+                    continue
+                ch_info = resp.get("channel", {})
+                if ch_info.get("is_im"):
+                    peer_user = ch_info.get("user", "")
```
bradhoekstra, line 123: 🟡 **Low — the `is_im` branch can record neither outcome, so the channel is re-probed forever.**

If `conversations.info` returns `ok` with `is_im: true` but no `user`, or `users.info` fails (`ratelimited`, missing scope), neither `_note_resolved` nor `_note_unresolvable` runs. The channel stays raw and is re-probed on every five-minute refresh indefinitely — the same unbounded-repetition class the miss cache was added to bound. An `else: _note_unresolvable(_channel_id, "no peer user")` on both inner branches would cover it.

deploy/docker/patches/channel_directory_threads.py
```diff
@@ -0,0 +1,221 @@
...
+            MISS_TTL_SECONDS,
+        )
+    if len(items) > MAX_PROBES_PER_REFRESH:
+        logger.warning(
+            "Channel directory: %d channels need resolving, probing %d this "
+            "refresh; the rest keep their raw ids until the next one",
+            len(items),
+            MAX_PROBES_PER_REFRESH,
+        )
+        items = items[:MAX_PROBES_PER_REFRESH]
```
bradhoekstra, line 220: 🟡 **Low — the probe cap can starve channels permanently.**

`items[:MAX_PROBES_PER_REFRESH]` truncates a list built in deterministic first-seen order, and successful resolutions are not cached (`note_resolved` clears the miss, and session-derived names are rebuilt raw each refresh). With more than 25 distinct raw-id channels, the same first 25 are probed every refresh and channels 26+ never resolve, while the WARNING ("a refresh wanting more than this many *distinct* channels is a new bug") repeats every five minutes. Rotating the window, or skipping channels resolved in the previous refresh, would let it converge.

agents/chat/SOUL.md
```diff
@@ -44,7 +44,7 @@ Your context for a card ends at `kanban_create → subscribed: true`. Completion
...
 
-A comment on a card that has already finished is worse than useless: it does not restart the work, the thread's subscription was deleted when the card closed, and nothing will ever read what you wrote. If `kanban_show` reports a terminal state (`done`, `blocked`, `gave_up`, or failed), do not promise delivery — answer from the card's `result`, which is already there.
+A comment on a card that has already finished is worse than useless: it does not restart the work, the thread's subscription was deleted when the card closed, and nothing will ever read what you wrote. If `kanban_show` reports a terminal state (`done` or `archived`), do not promise delivery — answer from the card's `result`, which is already there. A `blocked` card is not one of these: it is still open, still subscribed, and a comment on it is exactly how §1.5's **Unblock** step feeds the specialist its answer.
```
bradhoekstra, line 47: 🟡 **Low — narrowing the terminal list to `done`/`archived` re-opens the bug this paragraph documents.**

The fix for `blocked` is right, but `gave_up`, `crashed` and `timed_out` — all listed at line 103 as states that wake the front door — are terminal too: the subscription is deleted and a comment on one is read by nobody. Under the new wording the agent can once again tell a user "you'll see the results post here" on a card that has already given up. Suggest `(done, archived, gave_up, crashed, timed_out)` with the `blocked` carve-out kept as written.

deploy/docker/patches/test_kanban_notifier.py
```diff
@@ -210,6 +212,134 @@ def test_the_summary_branch_keeps_the_status_line_and_adds_the_report(self):
...
+* Total Wall-Clock Delay: 125.798794 seconds (approx 2 minutes, 5.8 seconds)
+
+3. ACTIVE EXECUTION TIME VS. SCHEDULING LATENCY
+* Total Active Execution Time (Container Runtimes): 66.638123 seconds
+  - Concurrent Sleep Phase: 24.000000 seconds
+  - Synthesis Phase: 42.638123 seconds
+"""
+
+#: The same report, written the way the persona contract now asks for.
+STRUCTURED_RESULT = """# Wall-Clock Delay Synthesis Report
```
bradhoekstra: 🟡 **Low — `STRUCTURED_RESULT` opens with a `#` H1 and the test around it is only green by accident.**

The constant is documented as "the same report, written the way the persona contract now asks for", but every persona file in this PR now says never `#`, and the gate refuses it. In the image, `result_shape_defects(STRUCTURED_RESULT)` returns `('top-level-heading', 'unquoted-numerics')`, so `test_a_structured_report_logs_nothing` (line 329) would fail — it passes only because `tools/` is not importable in the unit-test environment. Demote the heading to `##`, backtick the epochs, and ideally add a `verify_` check that exercises the notifier warning with the shared module present, since nothing at build time covers that seam.

agents/chat/scripts/profile_cron_tick.py
adamparco: Confirmed, reproduced, and fixed in `e243c31`. This was the right call to mark High — if anything the blast radius is worse than the comment says.

**Reproduced.** `platforms: "slack"` in a profile's `config.yaml` gives:

```
AttributeError: 'str' object has no attribute 'get'
```

**Blast radius.** `home_target_env` is called unguarded from `main()`, so the exception escapes the ticker itself. It is not "one profile loses its home target" — it is *no profile ticks at all*, every minute, for as long as the file is that shape. And the file is one the running agent writes: `/sethome` is a chat command, so a half-written or hand-edited `config.yaml` takes down the whole cron schedule for the pod.

**Fix.** Every read now goes through one helper:

```python
def _mapping(value: object) -> dict:
    """``value`` if it is a mapping, else an empty one."""
    return value if isinstance(value, dict) else {}
```

so a bad shape costs one un-restored home target instead of the schedule. The docstring records why this function in particular must never raise.

**Tests.** `test_no_shape_of_config_can_stop_the_tick` covers six shapes as subTests — scalar document, list document, `platforms` as a string, `platforms` as a list, the platform as a string, `home_channel` as a string. Plus `test_a_good_home_still_survives_a_junk_sibling`, which pins that the guard degrades one unreadable branch rather than the whole read.

**Regression-proof.** Stashing only `profile_cron_tick.py` and re-running fails with 6 errors, all `AttributeError: 'str' object has no attribute 'get'` — so the tests fail without the fix rather than passing vacuously.

`make test-python` is green.

deploy/docker/patches/kanban_notifier.py
adamparco: I want to push back on this one, though I think the underlying observation is right and it is worth saying why the split exists.

The two functions are deliberately different sets, not a strict/lenient pair:

```python
GATED_DEFECTS = ("top-level-heading", "ascii-substitute")
```

`result_shape_defects` returns all four; `gate_defects` filters to the two that can *refuse* a completion. The module docstring marks the other two explicitly:

```
``heading-without-prose`` (warn only)
``unquoted-numerics``     (warn only)
```

"Warn only" is their entire purpose. This notifier warning is the only consumer they have — nothing else calls `result_shape_defects`. So switching this call site to `gate_defects` would not make the warning stricter or looser; it would delete the two warn-only defects from the system, leaving code that detects them and nobody who looks. If that is the intent it should be a deletion, not a narrowing, so the next reader is not left wondering why they are computed.

Where I think you are right: this fires at WARNING, on the delivery path, for things that are genuinely cosmetic, and it is a log line nobody has asked for. If the complaint is noise, the honest fixes are to drop the level to INFO, or to retire `heading-without-prose` and `unquoted-numerics` outright. I have not done either here because it is a judgement call about a different PR's design and I would rather not fold it in silently — happy to do whichever you prefer, in this PR or a follow-up.

Worth noting the related finding on `top-level-heading` (your comment on `kanban_report_format.py:321`) *did* stick, and I fixed the advice there — so this is not a blanket disagreement with the pair.

deploy/docker/patches/kanban_report_format.py
adamparco: Confirmed as a real hazard, and fixed in `e243c31` — but I kept the gate and fixed the advice instead, which is a deliberate difference from the suggested change.

You are right that a `#` at column 0 inside an unfenced manifest, diff or script is read as an H1. Where I landed differently is on what the agent should then do. The old advice said, in effect, "that `#` is a heading, demote it" — so an agent that hit this on a Kubernetes manifest would be told to edit its own YAML comment to satisfy a formatter. That is the actively harmful outcome: it corrupts the payload to silence the check.

But the check is not wrong to have fired. An unfenced manifest in a chat message renders as a wall of text whichever way you slice it, and its first comment really does render as a heading. The defect is real; the *diagnosis* was wrong.

So the advice now distinguishes the two cases:

> `result` starts a line with `#`. If that is a heading, use `##` or `###` instead — the chat message already carries this card's title, so an H1 renders as a second banner saying the same thing. If it is a comment in a manifest, diff or script, put that block in a ``` fence: unfenced, it renders as a wall of text and its first comment reads as a heading. **Do not delete the `#` in that case.**

Both readings now lead to a correct action, and the fence fixes the rendering problem the gate was pointing at rather than hiding it.

I also added a comment block above the `_H1` check recording why it deliberately does not try to parse fences — a regex that tracks fence state is a second, worse Markdown parser, and getting it wrong fails open on the case the gate exists for.

Pinned with `test_the_advice_for_an_h1_covers_the_unfenced_manifest_reading`.

deploy/docker/patches/kanban_result_required.py
adamparco: You are right, and I traced the mechanism to be sure rather than argue from the diff.

**The window is real.** On a refusal the injected gate does:

```python
_result_err, result = _require_result(tid, summary, result)
if _result_err:
    return tool_error(_result_err)
```

That returns before `kb.complete_task` is ever reached, so nothing is written: the card stays open with no `result`. Your reading is exactly right.

**What limits it.** The refusal is a one-shot nudge, not a standing rule. `_refused_at[key]` is set on the refusal and popped on the next call, and while `nudge_spent` is true the shape check is skipped entirely — `return None, result` stores whatever comes back, ugly or not, for `NUDGE_TTL_SECONDS` (15 minutes). So the steady state is "one extra round trip", not "the card can never close".

**What that leaves.** Death in the gap, and of your three cases `max_turns` is the one that actually worries me: if `kanban_complete` was the worker's last permitted turn, the refusal consumes it and an expensive, complete report is gone. Crash and reaping are the same shape but rarer.

**And the trade is worse than it looks**, because of what the gate fires on. Both gated defects are rendering faults on reports whose content is complete:

- `top-level-heading` — an H1 that duplicates the card title.
- `ascii-substitute` — `=== Title ===` or ALL-CAPS sections, structure Slack cannot see, so it renders flat.

Neither means the answer is missing. So the current behaviour risks losing a complete report to protect against an ugly one, in a stack whose whole premise is that the report reaching the user's thread *is* the deliverable. Put that way I do not think the trade survives.

**Why I have not just changed it.** Both of your suggestions fix it, and they are meaningfully different products:

1. *Store and advise* — complete the card, return success, carry the formatting advice in the response. Nothing can ever be lost; the nudge gets much weaker, because agents rarely act after a successful completion.
2. *Narrow the gate* — but as above, both gated defects are already cosmetic-by-construction, so narrowing on "genuinely unreadable" would empty the gate rather than shrink it.

A third option keeps both properties: persist the result before refusing, so a death loses nothing and the retry overwrites. It costs a signature change to `_require_result`, and the Dockerfile greps for that call string as its anchor, so it is a slightly larger change than it sounds.

My recommendation is (1) — guaranteed delivery beats a formatting nudge, and the warn-only log line already covers the diagnostics. But it is your PR's design and I would rather not fold it in silently the way I did the mechanical fixes. Say which you want and I will make it, here or in a follow-up.

agents/chat/scripts/profile_cron_tick.py
adamparco: Not a bug, but the code gave you no way to know that — so I have fixed the thing that was actually wrong, which is that the entry criterion was never written down. `e243c31`.

The map is not "platforms we support". It is "platforms whose home-channel variable Hermes strips from the child environment". Checked against the running image:

```
$ grep -n HOME_CHANNEL /opt/hermes/tools/environments/local.py
279:  "TELEGRAM_HOME_CHANNEL"
280:  "TELEGRAM_HOME_CHANNEL_NAME"
281:  "DISCORD_HOME_CHANNEL"
282:  "DISCORD_HOME_CHANNEL_NAME"
286:  "SLACK_HOME_CHANNEL"
287:  "SLACK_HOME_CHANNEL_NAME"
296:  "SIGNAL_HOME_CHANNEL"
297:  "SIGNAL_HOME_CHANNEL_NAME"

$ grep -c GOOGLE_CHAT /opt/hermes/tools/environments/local.py
0
```

`GOOGLE_CHAT_HOME_CHANNEL` is not on the blocklist, so it already survives `build_subprocess_env` and the child inherits it. Adding it here would restore a variable that was never removed — at best a no-op, at worst it would overwrite a live inherited value with whatever `config.yaml` last recorded.

Telegram, Discord and Signal *are* on the blocklist, so by the real criterion they would qualify. They are out of scope for this PR, which is fixing an observed Slack failure, but that is a defensible follow-up and the comment now says so.

The map now carries the rule so the next person does not have to re-derive it:

```python
# Slack-only because the entry criterion is "the platform's home-channel var is
# on the provider-env blocklist", not "the harness supports the platform". Of
# the platforms shipped here, ``tools/environments/local.py`` blocks Slack's
# (and Telegram's, Discord's and Signal's); ``GOOGLE_CHAT_HOME_CHANNEL`` is
# absent from that list, so it survives ``build_subprocess_env`` and the child
# inherits it with nothing to restore. Add a platform here when its var joins
# the blocklist, not when the platform is enabled.
```

Your instinct that a bare `"slack"` in a dict named `HOME_TARGET_ENV_KEYS` looks like an oversight was the right one to follow — it looked like one because nothing said otherwise.

deploy/docker/patches/apply_channel_directory_threads.py
adamparco: Confirmed and fixed in `e243c31`.

You are right that the `is_im` branch had paths that recorded neither outcome. Concretely, an IM whose `conversations.info` succeeded but returned no `user`, and one whose follow-up `users.info` came back not-ok, both fell out of the branch having called neither `_note_resolved` nor `_note_unresolvable`. The channel stays unresolved *and* unsuppressed, so it is re-probed on every refresh forever — and each of those probes holds a slot under `MAX_PROBES_PER_REFRESH` that a resolvable channel could have used. That is the starvation half of your other finding, arriving through this door.

Restructured both dead ends to early-return style so each one records the miss before it leaves:

```python
peer_user = ch_info.get("user", "")
if not peer_user:
    _note_unresolvable(_channel_id, "im with no peer user")
    continue
user_resp = await client.users_info(user=peer_user)
if not user_resp.get("ok"):
    _note_unresolvable(
        _channel_id,
        "users.info: " + str(user_resp.get("error", "not ok")),
    )
    continue
```

The nesting was what hid it — with the success path inside two `if`s, the paths that fall through are invisible. Flattened, every exit is on the page. There is now a comment above the branch saying that both dead ends must record the miss and why, so the next edit does not reintroduce a silent third path.

deploy/docker/patches/channel_directory_threads.py
adamparco: Half right, and the half that was right is fixed — though via your other finding rather than here.

Taking the two parts separately:

**Permanent starvation: no, once the `is_im` hole is closed.** The cap is not the whole story, because the queue it truncates is not static. A channel that gets probed and misses is recorded by `_note_unresolvable` and suppressed for `MISS_TTL_SECONDS` (6 hours), so it leaves the queue and the ones behind it move up. `items[:MAX_PROBES_PER_REFRESH]` is therefore a rotating window, not a fixed prefix, and every channel reaches the front within a bounded number of refreshes.

**But that argument depended entirely on every probe recording an outcome** — which is exactly what your `apply_channel_directory_threads.py:123` finding showed was not true. A channel that recorded neither outcome never gets suppressed, so it re-enters the queue at the same position on every refresh and does hold a slot indefinitely. With enough of them the window stops rotating and the tail genuinely starves. That is a real permanent-starvation vector, and it is fixed in `e243c31` — both `is_im` dead ends now call `_note_unresolvable` before they `continue`.

So I would frame it as: the cap is safe *given* the invariant "every probe records an outcome", and the bug was that the invariant did not hold. Fixing the invariant is the correct fix; raising or removing the cap would have papered over it.

**Visibility: already there.** Hitting the cap logs at WARNING with the count and the limit (`channel_directory_threads.py:214`), so a refresh that is truncating says so rather than silently dropping the tail.

Nothing changed on this line as a result, but the finding was worth making — it is what made me check whether the rotation argument actually held, and it did not.

agents/chat/SOUL.md
adamparco: I have to disagree with this one, and I would rather explain why than just decline it — because making the suggested change would reintroduce the exact bug the paragraph exists to prevent.

The finding conflates two vocabularies that Hermes keeps separate: card **statuses** and **run outcomes / event kinds**. Checked against the running image:

```
$ grep -n VALID_STATUSES /opt/hermes/hermes_cli/kanban_db.py
102: VALID_STATUSES = {"triage", "todo", "scheduled", "ready", "running",
                      "blocked", "review", "done", "archived"}
```

`gave_up`, `crashed` and `timed_out` are not statuses at all — `kanban_db` raises `ValueError` if you try to set one. They are event kinds. What the *status* does on a give-up is spelled out in the same file:

```
7725: # the task transitions ready → blocked with a ``gave_up`` event
      # on top of the event we already emitted
```

and, on the distinction itself:

```
4114: This emits ``"gave_up"``, *not* ``"blocked"``, and is meant to recover
      automatically once the underlying conditions change
```

— that "not blocked" is about the **event row**, not the status. The card lands in `blocked`.

So `kanban_show` can never report `gave_up` as a state. Adding it to the terminal list would add a value that cannot occur, and — the part that actually matters — a card that gave up *is* `blocked`, so widening the terminal list to include it would tell the front door not to comment on a card that is still open and still subscribed. That is precisely the "you cannot reach this card" failure the next sentence is there to prevent:

> A `blocked` card is not one of these: it is still open, still subscribed, and a comment on it is exactly how §1.5's **Unblock** step feeds the specialist its answer.

`done` and `archived` really are the only two terminal statuses, so the list is correct as written.

Worth adding that the file already uses your three names correctly, forty lines further down, in the register where they belong — as *wake* events rather than states:

> What _does_ wake you is a card that **blocked or failed** (`blocked`, `crashed`, `timed_out`, `gave_up`)

Both passages are right; they are answering different questions. I have left line 47 alone.

deploy/docker/patches/test_kanban_notifier.py
adamparco: Confirmed and fixed in `e243c31`. You were right that it was green by accident, and the accident is worth naming: the fixture was a *well-shaped result* fixture that would have been refused by the gate it shares a module with.

Two changes:

1. `STRUCTURED_RESULT` now opens `## Wall-Clock Delay Synthesis Report` instead of `#`, with a five-line comment recording that the heading level is load-bearing — an H1 here is a gated defect, so a future edit that "tidies" it back to `#` is changing the fixture's meaning, not its formatting.

2. Added a `WellShapedFixtureTest` that asserts it directly:

```python
self.assertEqual((), gate_defects(STRUCTURED_RESULT))
```

That is the part that stops it recurring. Before, nothing connected the fixture to the gate, so the fixture could drift into a shape the gate refuses and every test around it would still pass — which is exactly the state you found it in.

adamparco:
Thanks Brad — all nine answered in their own threads, with the evidence in each. Roll-up here, plus **two I did not decide unilaterally and would like your call on**.

**Fixed** (`e243c31`, `aa45d7a`):

| # | Finding | Outcome |
| - | ------- | ------- |
| 1 | Unguarded config read can fail every cron tick | **Confirmed.** Highest-value finding in the review — a mis-shaped `config.yaml` took the whole schedule down, not just the one profile. Guarded, with a test. |
| 3 | Gate refuses raw manifests | **Confirmed in part.** The gate is right to refuse them, but the advice it printed sent the author the wrong way. Gate kept, advice rewritten. |
| 6 | (Low) | **Confirmed and fixed.** |
| 9 | Fixture green by accident | **Confirmed and fixed**, plus a `WellShapedFixtureTest` so the fixture cannot drift back into a shape the gate refuses. |

**Pushed back, with reasoning in-thread:**

- **Finding 5** — not a bug. The map is keyed on a criterion that happens to be Slack-only today; the Google Chat case you describe does not reach it. But nothing said so, so the criterion is now documented rather than inferred.
- **Finding 7** — half right. The cap is safe once finding 6's invariant holds, and the WARNING you asked for already exists.
- **Finding 8** — this one I think would be actively harmful, and it is worth being explicit about why: card **statuses** (`VALID_STATUSES`) and run **outcomes** (`TERMINAL_KINDS`) are two separate vocabularies that happen to share some spellings. Merging them would make a card's status settable to an event kind. Details and the two enum sites in the thread.

**Open — your call, not mine:**

- **Finding 2 (the warning on cosmetic defects).** You are right that it fires on things nobody needs woken for. But the two warn-only defects have no other consumer: drop the warning and they become dead code. So it is a choice between demoting the notifier to INFO and retiring those two defects, and I did not want to make it for you.
- **Finding 4 (a refusal discarding a complete report).** You were right and my first read was wrong — I traced it again and the gate does `return tool_error(...)` **before** `kb.complete_task`, so the card stays open with no result and the report is gone. The loss window is real. Three ways out, and they are a design choice rather than a bug fix: store-then-advise (my recommendation — keep the refusal, but persist the report first so nothing is lost), persist-before-refusing, or accept the loss and say so. Say which and I will implement it.

**Verification.** The disputed findings were checked against the running image rather than reasoned about. Full suite green on the branch: 1,722 tests, `make docs-check` clean.

_This reply was prepared with the help of Claude._

deploy/docker/patches/kanban_result_required.py
adamparco: Resolved in `9d0ce49`, and not the way I first proposed — the gate is gone rather than reordered.

**Correcting my own option list first.** I offered three ways out and two of them ("store-then-advise" and "persist-before-refusing") were the same thing described twice. The genuinely distinct choices were: reorder so the report is stored and the advice returned anyway; keep the refusal but persist first; remove the shape check; or accept the loss and document it. I recommended the first without noticing the second was its twin.

**What settled it was measuring the retry, not the refusal.** On the running image, with a 367-character H1 report:

```
FIRST  attempt -> refused = True  | result_to_store = None
RETRY  attempt -> refused = False | result_to_store = <the identical report>
```

The retry skips the check — that is the one-nudge budget working as designed. So the mechanism could **lose a complete report** but could **not guarantee a well-formed one**. It bought exactly one nudge, and the thing it risked to buy it was the deliverable. In a change whose whole premise is that `result` reaching the user's thread is the answer, I do not think that trade survives. Storing first and advising anyway would have fixed the loss, but it would have kept a round trip that, measurably, does not reliably produce a better report.

**So:** `require_result` now refuses exactly one thing, a blank `result`. No length, quality or formatting floor. `SHAPELESS_RESULT_ERROR` and `shapeless_result_error()` are deleted with their only caller, and `kanban_result_required` no longer imports the report-format module at all.

The contract has not gone anywhere — it is still stated in the report-format stanza stapled to every card body and in the `result` schema description, which is where a worker actually reads it. What changed is that it is an instruction rather than a gate. The measurement moved to the delivery path at two log levels, which is your finding #2; details in that thread.

Verified in the image: `verify_kanban_result.py` (24 checks) and `verify_kanban_report_format.py` (33 checks) both green, now including "the H1 report is stored on its first attempt", "the ASCII-structured report is stored too", and "a mis-shaped report does not spend the empty-result nudge". The negative control matters here more than the assertions do, so I ran it: reinstating a shape refusal fails three tests in `ShapeIsNotRefusedTest`.

Six prose sites claimed the refusal and no longer do — both `SOUL.md` files, both `AGENTS.md` files, the Dockerfile patch-block comment, and the platform-agent concepts page.

_This reply was prepared with the help of Claude._

deploy/docker/patches/kanban_notifier.py
adamparco: Taking this one after all, in `9d0ce49` — your observation was right and my pushback rested on a constraint that no longer exists.

What I argued was that dropping the warning would leave `heading-without-prose` and `unquoted-numerics` with no consumer at all, since the gate only acted on the other two. Removing the shape gate entirely (your finding #4) changes that: the delivery-path log is now the *only* consumer of any of them, so the question stops being "warn or retire" and becomes "at what level".

`_warn_if_unstructured` is now `_log_result_shape` and splits the four defects two ways:

- **WARNING** for `top-level-heading` and `ascii-substitute` — the two that always render wrongly — with the `DEFECT_ADVICE` text interpolated, so the line names the edit rather than just the rule.
- **INFO** for `heading-without-prose` and `unquoted-numerics`, which is where a bad report read back later still shows why, without waking anyone.

Live on the running image, both reports delivered unchanged:

```
--- SERIOUS (H1) ---
WARNING [kanban] card t_probe completed with a 240-character result whose formatting will not
render well in chat: top-level-heading. `result` starts a line with `#`. If that is a heading,
use `##` or `###` instead — the chat message already carries this card's title... The contract
is in the card body's report-format stanza and in agents/platform/SOUL.md §0 / ...
delivered chars: 254 | report reached the message: True

--- COSMETIC (bare list) ---
INFO [kanban] card t_probe completed with a 189-character result with cosmetic formatting
defects: heading-without-prose, unquoted-numerics.
delivered chars: 203 | report reached the message: True
```

Two details worth flagging because they are easy to get wrong:

**The import-failure fallback still warns.** `gateway` imports `tools` lazily and optionally, and when that import fails the only defect this module can detect unaided is `ascii-substitute` — which is in the serious tier. Demoting the fallback to INFO would silently lose the one finding it can make on its own, so it stays a WARNING. `test_the_fallback_treats_its_one_defect_as_serious` pins that.

**`STRUCTURED_RESULT` is a good demonstration of the split.** The notifier suite's own well-shaped fixture still carries `unquoted-numerics` under the shared detector — raw epochs in its table. Warning on it would be exactly the noise you objected to, and `test_the_structured_fixture_never_warns` now says so out loud.

`GATED_DEFECTS`/`gate_defects()` are renamed `SERIOUS_DEFECTS`/`serious_defects()`, since they pick a log level now rather than a refusal. Negative-controlled: warning on every defect fails three tests in `ResultShapeLogLevelTest`.

_This reply was prepared with the help of Claude._

deploy/docker/patches/kanban_report_format.py
adamparco: Correction to my reply above: the gate I said I was keeping is gone as of `9d0ce49`.

The fix I described still stands and is unchanged — fenced code is stripped before the H1 scan, so a `# list the pods` comment in a ``` block is no longer read as a heading, and `DEFECT_ADVICE["top-level-heading"]` now names the fence as the other valid edit instead of telling the author to demote a comment marker.

What changed is the consequence. When I wrote "the gate is right to refuse them", the detector's verdict could cost a worker a round trip and, as your finding #4 showed, its complete report. Removing that gate makes this finding *less* severe than either of us treated it: an unfenced manifest whose first comment reads as an H1 now produces a WARNING in the gateway log naming both possible edits, and the report is delivered either way. So the reason the fence-stripping still matters is smaller but real — a log line that tells a reader to demote a YAML comment is a log line that wastes their time.

The check `verify_kanban_report_format.py` runs for this is unchanged in substance, only relabelled: "the warning for `top-level-heading` names the edit to make".

_This reply was prepared with the help of Claude._

adamparco:
Closing out the two I left open for you, Brad — both are now decided and implemented in [`9d0ce49`](https://github.com/gke-labs/kube-agents/pull/602/commits/9d0ce49), and one of them changes an answer I gave you earlier.

| # | Finding | Outcome |
| - | ------- | ------- |
| 4 | A refusal discarding a complete report | **Taken, and further than I proposed.** The shape refusal is removed rather than reordered. |
| 2 | The warning fires on cosmetic defects | **Taken.** WARNING for the two that always render wrongly, INFO for the two that are taste. |
| 3 | Gate refuses raw manifests | **Correction to my earlier reply.** I said "gate kept, advice rewritten"; the gate is now gone. The fence-stripping fix and the rewritten advice both stand. |

**What settled Finding 4 was measuring the retry, not the refusal.** On the running image, with a 367-character H1 report: the first attempt returned `refused = True, result_to_store = None`, and the retry returned `refused = False` storing the *identical* badly-shaped report, because the retry skips the check. The mechanism could lose a complete report but could not guarantee a well-formed one — it bought one nudge, and what it risked to buy it was the deliverable. In a change whose whole premise is that `result` reaching the user's thread *is* the answer, that trade does not survive.

I also owe you a correction on how I framed the options: I offered three and two of them were the same thing described twice. The real choices were reorder-and-advise, persist-then-refuse (its twin), remove the check, or accept the loss and document it.

**The contract has not gone anywhere** — it is still in the report-format stanza on every card body and in the `result` schema description, which is where a worker actually reads it. It is now an instruction rather than a gate, and the measurement moved to the delivery path at the two levels your Finding 2 asked for. That is also what keeps `heading-without-prose` and `unquoted-numerics` from becoming the dead code I was worried about.

**Verification.** Both in-image verifies re-run against the real patched `kanban_tools.py` on the live cluster: `verify_kanban_result.py` at 24 checks and `verify_kanban_report_format.py` at 33, both green. A probe on the real delivery path shows both an H1 report and a bare-list report reaching the message, the first with a WARNING naming the edit, the second with an INFO. Full suite green at 1,726 tests; `make docs-check` clean; `prettier --check` clean.

Both new behaviours were negative-controlled rather than just asserted: reinstating a shape refusal fails three tests in `ShapeIsNotRefusedTest`, and warning on every defect fails three in `ResultShapeLogLevelTest`.

Six prose sites claimed the refusal and no longer do — both `SOUL.md` files, both `AGENTS.md` files, the Dockerfile patch-block comment, and the platform-agent concepts page. The PR description is updated too.

_This reply was prepared with the help of Claude._

deploy/docker/patches/kanban_report_format.py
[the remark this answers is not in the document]
adamparco: Fixed, in two independent ways — both already on this branch.

**The fence scan.** `2427d69` ("stop a shell comment in a fence reading as an H1") changed the check to `_H1.search(_FENCE.sub("", body))`, so a `# namespace: …` inside a fenced block no longer matches.

One deliberate deviation from the suggested fix: it strips the fence only, not `_strip_code`. An H1 has to open a line, so it can never hide inside a backtick pair, and removing inline code would let `` `--dry-run` # note `` collapse onto its hash and read as a heading nobody wrote. The reasoning is recorded in the comment above the call.

The residual case you would still have — an *unfenced* manifest whose first line is `# Managed by kube-agents` — is left matching on purpose: that report is defective either way, since unfenced YAML renders as a wall of text in Slack. `DEFECT_ADVICE["top-level-heading"]` was rewritten to name the fence as the edit for that case and to say explicitly "Do not delete the `#`", which was the actively-wrong advice you flagged.

**The gate.** Separately, `9d0ce49` removed the shape gate entirely (review finding #4 on the PR): `GATED_DEFECTS` is now `SERIOUS_DEFECTS` and the check is warn-only, so this can no longer refuse a completion. The one-nudge budget you noted as bounding the cost is no longer the thing bounding it — nothing is refused.

Both changes are in this PR, which has since been retargeted at `main` and now carries the whole `ap-*` stack.

dshnayder (approved):

bradhoekstra (approved):
Quick approve to make sure this merges.

PR #604, opened 2026-08-09
fix(delivery): make cron briefs and kanban notifications actually reach the user

adamparco:
# Pull Request

**Stacked on #602.** Base: `ap-result-delivery` (#602's head, mirrored into this repository
solely to serve as a target).

**This PR's own work is the three commits below — 10 files.** The diff GitHub renders is
larger (currently 26 commits / 27 files) because the mirrored base branch in this repository
is behind #602's head, so #602's carry-over files appear here too: `cron_tick_lock_scope.py`,
`kanban_report_format.py`, `default_profile_config.py`, `docker-entrypoint.sh`,
`docs/README.md`, `tests/test_default_profile_config.py`. Those belong to #602 and should be
reviewed there. Refreshing the mirrored base needs upstream push rights; the gap closes on
its own when #602 merges and this is retargeted.

- `fix(agents): deliver cron briefs to Slack without a local bot token`
- `fix(agents): post scheduled Slack briefs flat, not in a stale thread`
- `fix(image): deliver kanban notifications at least once`

> This pull request was originally opened against `ap-agent-prose` — #602's own base — and so
> showed #602's four commits alongside these three. It has since been retargeted at
> `ap-result-delivery`, which now exists here. **When #602 merges, retarget this at
> `ap-agent-prose` (or at `main` once the rest of the stack has landed) and rebase** — this
> repository squash-merges, so #602's four commits collapse into one new commit and will not
> be recognised as already applied. `ap-result-delivery` carries no work beyond #602 and
> should be deleted along with the stack's other base branches once everything lands.

## Why This Change

Three defects with one symptom in common: **the system produced an answer and the human
never saw it.**

**Cron briefs never reached Slack.** A named-profile cron tick runs in a child interpreter
spawned by `profile_cron_tick.py` with `adapters=None, loop=None`, so every Slack surface it
can reach hangs off the `PlatformEntry` in the registry rather than off a live adapter. Two
independent gates on that entry both asked for a local bot token that, in this deployment,
only the credential proxy holds:

- `is_connected` reported Slack unconfigured, which surfaced in the dashboard as
  `delivery: platform 'slack' not configured/enabled`, and earlier as
  `delivery: no delivery target resolved for deliver=all`.
- Behind it, `_standalone_send` resolved its own token and returned
  `SLACK_BOT_TOKEN not configured`, so even a resolved target would not have sent.

**Briefs that did send buried themselves in a thread.** `/sethome` records whichever thread
it happened to be typed in, and cron then threaded every scheduled report under that one
ageing thread — where only the first is visible in the channel.

**Kanban notifications were at-most-once.** `claim_unseen_events_for_sub` committed the
subscription's cursor and only then, ~130 lines of control flow later, tried to send.
On 2026-08-09 13:00:29 task `t_a18254ca`'s `completed` event (id 971) was claimed, the
cursor moved 958 → 971, and the gateway took a fatal signal 162 ms later at the `get_task`
on the very next line. The 7,124-byte report was intact in `tasks.result`; no tick ever
looked at the event again and the answer had to be asked for by hand. The crash is not the
defect — the per-subscription handler catches bare `Exception` and does **not** rewind, so
an ordinary exception in that window loses the message just as completely.

## What Changed

**Files:**

- `agents/platform/scripts/test_slack_relay_patch.py`
- `agents/chat/scripts/profile_cron_tick.py`
- `agents/chat/scripts/test_profile_cron_tick.py`
- `deploy/docker/patches/kanban_notify_delivery.py`
- `deploy/docker/patches/apply_kanban_notify_delivery.py`
- `deploy/docker/patches/verify_kanban_notify_delivery.py`
- `deploy/docker/patches/test_kanban_notify_delivery.py`
- `deploy/docker/Dockerfile`
- `docs/site/src/content/docs/concepts/chatops.md`

**Added/Changed/Fixed:**

- A `PlatformRegistry.register` wrapper patches the Slack entry as it is registered,
  replacing both gates at once: `is_connected` returns true when Slack is reachable only
  through the relay, and a relay-backed `standalone_sender_fn` sends via
  `chat.postMessage` over the credential proxy. Both are marked
  `_credential_proxy_relay_patched` so the state is observable.
- The sender resolves a DM through `conversations.open`, formats mrkdwn using the adapter's
  own `format_message`, and delegates untouched to the original sender whenever a local
  token *is* present — so token-holding deployments are unaffected.
- `home_target_env` now passes the home channel to the cron child with the thread id
  explicitly cleared, so briefs post flat.
- The kanban claim becomes a pure read (`unseen_events_for_sub`, upstream's own read-only
  sibling), the cursor is written once *after* delivery, and the now-vestigial
  `_kanban_rewind` is neutered. An in-process high-water map bounds a permanently failing
  cursor write to one duplicate per process, and `advance_after_delivery` logs rather than
  unwinding the tick.

## Why This Matters

- Any deployment where Slack credentials live in the credential proxy rather than the agent
  container got **zero** scheduled briefs. That is the credential-isolation posture this
  repository recommends, so the default secure configuration was the broken one.
- The fix adds no token to the agent container and mutates no environment — it routes
  through the relay that already exists.
- At-least-once trades duplicates for recoverability. That trade depends on there being
  exactly one notifier, which holds today (`replicas: 1`, `strategy: Recreate`); raising
  `availability.replicas` becomes a correctness change rather than a scaling knob, and that
  constraint is recorded in the module docstring.

## Context

- Deliberately **not** included, each rejected on evidence: no `get_task` reorder (hoisting
  it splits `complete_task`'s single transaction and makes a stale, result-less task row the
  deterministic outcome under contention), no `mmap_size` change (already 0 — the gateway
  maps no `.db` file), and no durable claim lease (the upgrade path if single-writer ever
  has to go).
- `files_upload_v2` is an SDK helper rather than a Web API method, so it is unreachable
  through the relay; the sender returns an explicit error for media instead of failing
  opaquely.
- Roughly 73% of cron runs legitimately return the `[SILENT]` sentinel and never attempt
  delivery. This PR fixes the runs that had something to say; it does not change that ratio.

## Testing

- `make test-python` — 9 suites, 1,768 tests, all passing.
- `make docs-check` — all four checks pass.
- 26 tests on the relay patch (up from 10), covering the token gate, delegation when a local
  token exists, DM resolution, thread handling, media rejection, blank-message handling, and
  both error shapes. One of these caught a real bug during development: a whitespace-only
  message became non-blank once mrkdwn-formatted and reached the relay, so the guard now
  checks the raw message as well as the formatted one.
- `verify_kanban_notify_delivery.py` replays the `t_a18254ca` incident against a real board
  and asserts upstream's `unseen_events_for_sub` is still a pure read — the fix silently
  reverts to at-most-once if it ever grows a write.
- **Verified live on-cluster** for the Slack half (image `dev-317006e-0809-120553`): profile
  platforms went `['api_server']` → `['api_server', 'slack']` with `enabled: True` and
  `token set: False`; `deliver=all` resolved to a real DM target; and a send returned
  `{'success': True, 'message_id': '1786292815.485629'}`, arriving flat in the DM.
- The kanban half is covered by tests and the verify script but is **not yet in a deployed
  image**.

---

Risk is contained to deployments without a local Slack token: with a token present the
sender delegates to the original path unchanged. The kanban change is the riskier half — it
converts a single-ownership claim into at-least-once delivery, so it is safe only while the
notifier is the single writer of these cursors.

This pull request was generated with the help of Claude.


deploy/docker/patches/kanban_notify_delivery.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    try:
+        kb.advance_notify_cursor(
+            conn,
+            task_id=sub["task_id"],
+            platform=sub["platform"],
+            chat_id=sub["chat_id"],
+            thread_id=sub.get("thread_id") or "",
+            new_cursor=int(cursor),
+        )
```
toshiowang, line 253: **Correctness risk — the only cursor write here that is not in a transaction.**

Every other mutator in this codebase is wrapped: `verify_kanban_notify_delivery.py` writes `with K.write_txn(conn): K._append_event(...)`, the module docstring above says `claim_unseen_events_for_sub` advances the cursor "inside its own `BEGIN IMMEDIATE`", and `_kanban_advance` opens its own connection and commits (section 6 of the verify proves that cross-connection). This call is bare.

If `kanban_db.connect()` does not use `isolation_level=None`, this UPDATE sits in an uncommitted implicit transaction and is discarded when the collect connection closes — so after a failed advance the cursor is **never** repaired, `logger.info("repaired the cursor…")` lies on every tick, and a gateway restart re-delivers the notification. Even under autocommit, skipping `BEGIN IMMEDIATE` means this loses to a concurrent writer with `database is locked` where the wrapped path would have waited out the busy timeout.

The build gate cannot catch either case: section 7 asserts `cursor_now() == BEFORE_FAILURE + 1` through the **same** `conn` that issued the write, so an uncommitted transaction reads back as a success. If you apply the fix, the verify should re-read on a second connection so the assertion means something.

```suggestion
    try:
        with kb.write_txn(conn):
            kb.advance_notify_cursor(
                conn,
                task_id=sub["task_id"],
                platform=sub["platform"],
                chat_id=sub["chat_id"],
                thread_id=sub.get("thread_id") or "",
                new_cursor=int(cursor),
            )
```

toshiowang (commented):
Three findings, one of which should be settled before merge.

**Scope note:** the description says "exactly the three commits below — 10 files", but the PR diff is 9 commits / 21 files — the mirrored base `ap-result-delivery` is behind #602, so the carry-over files (`cron_tick_lock_scope.py`, `kanban_report_format.py`, `default_profile_config.py`, `docker-entrypoint.sh`, `docs/README.md`, `tests/test_default_profile_config.py`) are in the diff too. Those all look correct.

The blocking item is `_repair_cursor` writing the notify cursor without a transaction wrapper — it is the only cursor write in the patch that does not, and the build gate cannot detect it because the verify reads back through the same connection. The other two are risks worth an answer rather than confirmed defects.

Checked and correct: the three-tuple shape and `min(new_cursor, seen)` in `read_unclaimed`; `MAX_TRACKED` eviction; the `ADVANCE_ANCHOR` uniqueness argument and its matching `grep -c … = "1"` guards; patch ordering against `apply_kanban_wake_nudge.py` and `apply_kanban_notifier.py`; `cd /opt/hermes` reaching the new verify; delegation-on-local-token in the relay sender; and the SOUL.md terminal-state correction, which matches `TERMINAL_STATUSES = {"done", "archived"}` in `kanban_comment_status.py:136`.

Minor, no reply needed: a successful `_repair_cursor` does not drop the high-water entry the way `advance_after_delivery` does, so a subscription that ever hit a failed write keeps a map entry for the life of the process. Bounded by `MAX_TRACKED`, but it contradicts "in normal operation the map is empty" in the docstring.

_Review produced with the help of Claude._

agents/platform/scripts/slack_relay_patch.py
```diff
@@ -472,6 +486,198 @@ async def download_bytes(self: Any, url: str, team_id: str = "") -> bytes:
...
+                    target = str(resolved)
+
+                body: dict[str, Any] = {
+                    "channel": target,
+                    "text": formatted,
+                    "mrkdwn": True,
+                }
+                if thread_id:
+                    body["thread_ts"] = thread_id
+                data = await relay_api_call("chat.postMessage", {"json": body})
```
toshiowang, line 643: **Risk (tentative): no length guard, where the sender being replaced is length-aware.**

This posts `formatted` as a single `chat.postMessage`. The registry entry carries `max_message_length` — your own fake entry in `test_slack_relay_patch.py` declares it at `40000` — and `test_kanban_notifier.py:147` records that "Slack's adapter chunks at MAX_MESSAGE_LENGTH = 39000".

Concrete failure: a brief over Slack's 40,000-character `text` limit comes back `msg_too_long`, this returns `{"error": "Slack API error: msg_too_long"}`, and the entire report is lost — the exact failure class this PR exists to close, on the deployments the PR exists to fix.

This is tentative because I cannot see upstream's `_standalone_send`. If the chunking lives in the caller (`gateway/delivery.py`) rather than in the sender, this is a non-issue and worth a one-line note in the docstring saying so. If it lives in `_standalone_send`, the relay path dropped it silently.

agents/platform/scripts/slack_relay_patch.py
```diff
@@ -483,3 +689,42 @@ def create_adapter(self: Any, name: str, config: Any) -> Any:
...
+        # this patch is deliberately deferring.
+        try:
+            registry_module = sys.modules.get("gateway.platform_registry")
+            entries = getattr(
+                getattr(registry_module, "platform_registry", None), "_entries", None
+            )
+            if isinstance(entries, dict) and "slack" in entries:
+                patch_slack_entry(entries["slack"])
+        except Exception:
+            LOGGER.debug("No pre-registered Slack entry to patch", exc_info=True)
```
toshiowang, line 756: **Risk: this is probably the primary path, and it is both untested and silent on failure.**

`install()` imports `gateway.platform_registry` at the top, which is very likely what populates `_entries` — so by the time the `register` wrapper is installed eight lines above, the Slack entry already exists and no future `register()` call will reach it. In a fresh cron child this fallback is not the corner case, it is the path that does the work.

It depends on three guessed private names (the `platform_registry` singleton attribute, `_entries`, and the `"slack"` key), and every way any of them can be wrong lands in a `LOGGER.debug`. The resulting symptom is indistinguishable from the bug being fixed: `delivery: platform 'slack' not configured/enabled`, with nothing above DEBUG to say why.

It is also uncovered: `SlackStandaloneRelaySendTest` builds a `registry_module` with no `platform_registry` attribute, so this block short-circuits in every test and only the wrapper path is exercised. Two suggestions: log at WARNING when the module is present but neither the singleton nor `_entries` resolves, and add a test that registers the Slack entry **before** calling `install()`.

deploy/docker/patches/kanban_notify_delivery.py
adamparco: Thank you for this one — the reasoning is right, the conditional is the important part, and the last paragraph found a genuine hole. Taking it in three pieces.

**The premise does not hold, so the suggested edit breaks the repair.** All three checks, against the running image:

```
$ grep -n isolation_level /opt/hermes/hermes_cli/kanban_db.py
1439:        isolation_level=None,

$ sed -n '10025,10039p' /opt/hermes/hermes_cli/kanban_db.py
def advance_notify_cursor(conn, *, task_id, platform, chat_id, thread_id=None, new_cursor):
    with write_txn(conn):                       # <-- already wrapped
        conn.execute("UPDATE kanban_notify_subs SET last_event_id = ? ...")

$ sed -n '2801,2813p' /opt/hermes/hermes_cli/kanban_db.py
def write_txn(conn):
    """Context manager for an IMMEDIATE write transaction."""
    _execute_boundary_with_retry(conn, "BEGIN IMMEDIATE")
```

So `connect()` *is* autocommit, and `advance_notify_cursor` opens its own `BEGIN IMMEDIATE` internally — via the same `write_txn` your suggestion would wrap it in a second time. That also answers the concurrency half: the bare call already gets `BEGIN IMMEDIATE` with the busy-timeout retry, so it does not lose to a concurrent writer.

**I ran your suggestion rather than reasoning about it.** Applied verbatim in the image, against a scratch database:

```
kanban notifier: cursor repair for t_9a888866 to 3 failed: cannot start a transaction within a transaction
kanban notifier: cursor repair for t_9a888866 to 3 failed: cannot start a transaction within a transaction
kanban notifier: cursor repair for t_9a888866 to 3 failed: cannot start a transaction within a transaction
kanban notifier: cursor repair for t_9a888866 to 3 failed: cannot start a transaction within a transaction
  FAIL the durable cursor was repaired without re-sending: cursor is 2, expected 3
  FAIL the repair is durable — a second connection sees it: another connection reads 2, expected 3
```

`OperationalError` on every tick, swallowed by the surrounding `except` into a warning — which produces exactly the outcome you were trying to prevent: the cursor never repaired and a re-delivery on restart. The bare call is deliberate, and that is now recorded in `_repair_cursor`'s docstring, because "this one is not wrapped and every other one is" is a reasonable thing to flag and the code gave no reason not to.

**Your last paragraph was right, and it is fixed.** Section 7 really did read back through the same `conn` that issued the write, so an uncommitted transaction would have satisfied it — the assertion could not distinguish "committed" from "pending on this handle", which is the only distinction that matters, since a restart is a new connection. The verify now re-reads on a second one:

```python
def cursor_on_a_fresh_connection():
    other = K.connect(DB)
    try:
        row = other.execute(
            "SELECT last_event_id FROM kanban_notify_subs WHERE ...", ...
        ).fetchone()
        return None if row is None else int(row["last_event_id"])
    finally:
        other.close()

check("the repair is durable — a second connection sees it",
      cursor_on_a_fresh_connection() == BEFORE_FAILURE + 1, ...)
```

That check is the third FAIL line in the run above, so it demonstrably bites rather than passing along with everything else. Committed in `0798da6`; the full verify is green in the image (40 checks).

So: premise falsified, but the review found a real defect in the build gate that would have hidden the exact failure you described. Worth the flag.

agents/platform/scripts/slack_relay_patch.py
adamparco: Your tentative branch is the one that holds: the chunking lives in the caller, so this is a non-issue — and I have added the docstring note you asked for. `0798da6`.

Resolved against the running image rather than inferred:

```
$ grep -rn max_message_length /opt/hermes --include=*.py | grep -i slack
/opt/hermes/plugins/platforms/slack/adapter.py:9084:        max_message_length=39000,

$ sed -n '843,849p' /opt/hermes/tools/send_message_tool.py
    # Smart-chunk the message to fit within platform limits.
    chunks = BasePlatformAdapter.truncate_message(message, max_len, len_fn=_len_fn)

$ sed -n '903,911p' /opt/hermes/tools/send_message_tool.py
        for i, chunk in enumerate(chunks):
            is_last = (i == len(chunks) - 1)
            result = await entry.standalone_sender_fn(
                pconfig, chat_id, chunk, thread_id=thread_id,
                media_files=media_files if is_last else [],
            )
```

`send_message_tool` splits against the registry entry's `max_message_length` and then calls `standalone_sender_fn` — this function — **once per chunk**. So `message` is already bounded on arrival, and the relay never sees the full brief.

Two details worth recording, since both bear on your reasoning:

- The live Slack entry declares **39000**, not 40000. The `40000` in `test_slack_relay_patch.py` is a fake entry in a test fixture, so it does not constrain anything; the real value sits under Slack's 40,000-character `text` limit with room for the API's own accounting. `test_kanban_notifier.py:147` was the accurate one.
- Adding a guard here would be actively wrong, not merely redundant: a second chunker would re-split an already-split chunk and interleave the pieces across the thread.

The docstring now says so, including who owns the split if that ever changes:

> Deliberately does not chunk. `message` arrives already split: `tools/send_message_tool.py` smart-chunks against the registry entry's `max_message_length` and then calls `standalone_sender_fn` — this function — once per chunk. The Slack entry declares 39000, under Slack's 40000-character `text` limit, so a chunk is bounded before it gets here. Adding a second chunker would re-split an already-split message and interleave the pieces across the thread. If this ever grows a caller that is not `send_message_tool`, that caller owns the splitting.

Flagging it was the right call — the asymmetry you spotted (a length-aware sender being replaced by one that is not) is exactly the shape of a real regression, and nothing in the file said why it was safe.

agents/platform/scripts/slack_relay_patch.py
adamparco: The premise turns out to be inverted — the wrapper is the primary path and this sweep is the corner case — but **both of your suggestions were right anyway, and I have taken both**. `0798da6`.

**Why the ordering is the other way round.** From a fresh interpreter in the image:

```
$ python3 -c 'import gateway.platform_registry as m; r = m.platform_registry; ...'
register.__module__   = slack_relay_patch
register.__qualname__ = install.<locals>.register
wrapper installed?    True
_entries at this point: []
```

Two things fall out of that. `sitecustomize` runs `install()` at interpreter startup, so `PlatformRegistry.register` is already the patched wrapper before anything else executes — and importing `gateway.platform_registry` populates **nothing**. `_entries` is empty.

The reason is that platforms are registered lazily. `hermes_cli/plugins.py:1751` hands the registry a *loader*, not an entry:

```python
platform_registry.register_deferred(platform_name, _loader)
```

and the entry object is only constructed when something first asks for the platform — `_resolve` pops the loader, runs it, and the loader's eager-load path calls `platform_registry.register(entry)`. That is the wrapped call. So in a cron child the sequence is: `install()` wraps `register` → the plugin manager registers deferred loaders → the first `get("slack")` fires the loader → the wrapper patches the entry. An earlier probe of mine confirmed the last step directly (`WRAPPER FIRED = True`).

So the sweep only matters if something resolves Slack *before* `sitecustomize` gets to `install()`, which is the narrow case rather than the common one.

**That does not rescue it, and your two suggestions stand.** A path that is rarely taken and silent when it breaks is worse than one that is often taken, not better — and you were right that nothing tested it. Both fixed:

1. **WARNING instead of silence.** The block now distinguishes "no registry yet" (normal, and the wrapper covers it — no log) from "the registry is up but its entries are not where this reads them" (the rot case):

```python
elif singleton is not None:
    LOGGER.warning(
        "Slack relay: platform registry is loaded but its entries are not "
        "introspectable (_entries is %s); a Slack entry registered before "
        "this patch installed will not be patched", type(entries).__name__,
    )
```

`_entries` is a private name, so a base-image bump that renames it is a live risk, and your point that the symptom is indistinguishable from the original bug is exactly right.

2. **Tests for the ordering.** New `SlackRegisteredBeforeInstallTest` — four cases: a Slack entry pre-registered before `install()` gets patched, a pre-registered non-Slack entry is left alone, an unreadable entries mapping warns, and an absent registry does *not* warn (that one would otherwise fire on every boot).

Both are negative-controlled rather than assumed green: disabling the sweep fails the first test, and disabling the warning fails the third. 30 tests in the file, all passing.

Your reading of the consequence was right even though the ordering was not — this was the least-covered code in the patch.

adamparco:
Thanks Toshio — all three findings answered in their own threads, and the two items from this summary are handled here.

**The minor one you said needed no reply: fixed, because it was load-bearing.** You were right that a successful `_repair_cursor` left the high-water entry behind, contradicting the docstring's "in normal operation the map is empty". `_repair_cursor` now returns whether the write landed, and the caller applies the same invariant `advance_after_delivery` already uses:

```python
repaired_to = min(int(new_cursor), seen)
if _repair_cursor(kb, conn, sub, repaired_to) and repaired_to >= seen:
    tracked.pop(sub_key(sub), None)
```

It is deliberately *not* dropped when the write landed short of `seen` — that remainder is still only in this process's memory.

Catching this also improved the verify. The existing check asserted the entry is "kept while the write is outstanding", but with the drop in place that is only true for one tick, so reading it after the loop would have tested the wrong instant. The verify now samples the map inside the loop and asserts both facts: kept while outstanding, gone once the repair makes it redundant.

**Scope note: you were right, and the description is corrected.** It claimed "exactly the three commits below — 10 files"; the rendered diff is 26 commits / 27 files. The body now states the real numbers, names the six carry-over files as #602's, and says the gap closes when #602 merges and this is retargeted. Refreshing the mirrored base branch needs upstream push rights, so I have not done it — same structural artifact you identified, just no longer misdescribed.

**Verification.** Everything above was checked against the running image rather than reasoned about, including running your suggested transaction wrapper to see what it does (it fails: `cannot start a transaction within a transaction` — details in that thread). The full verify passes in the image at 40 checks, and both new checks were negative-controlled so they demonstrably fail when the behaviour they assert is removed.

Pushed as `0798da6`.

toshiowang (approved):

PR #606, opened 2026-08-10
feat(watcher): persist the dedup cache across restarts

stalhaali:
# Pull Request

## Why This Change

  The dedup cache starts empty on every watcher restart, and an empty cache is not a neutral
  state.
  
  When the watcher starts, each cluster's informer performs an initial LIST before switching to
  a watch stream. That LIST returns every Event still held by the API server — an hour's worth
  on GKE by default. With a populated cache those are recognised and suppressed. With an empty
  one, every incident inside that window is re-reported as new: duplicate Chat alerts, duplicate
  triage sessions, and duplicate LLM spend on incidents the operator has already seen.
  
  The watcher has supported --dedup-persist since it was written, and it already does the hard
  parts — per-cluster snapshot files, atomic tmp-and-rename writes, graceful handling of a
  missing or corrupt snapshot. Nothing ever passed the flag, so none of it ran.
  
  This matters more now than it used to. Since the watcher moved into envoy-credential-proxy, it
  is supervised and restarted in place rather than being allowed to end the container, so
  restarts are a routine occurrence rather than a rare one. Pod rollouts restart it too. Every
  one of those replayed the last hour of events.
## What Changed

**Files:**

  - deploy/shared/start-services.sh
  - k8s-operator/cmd/k8s-event-watcher/README.md

**Added/Changed/Fixed:**

  - The entrypoint now creates /opt/data/event-watcher and passes
  --dedup-persist=<dir>/dedup.json. The watcher suffixes that base path with each cluster's
  profile directory name, so every cluster keeps its own snapshot file.
  - Chose the data volume over the container's own state directory. /var/lib/credential-proxy is
  an emptyDir with medium: Memory and a 16Mi cap — it would survive a watcher restart but not a
  pod restart, which is the case that actually bites.
  - If the directory cannot be created, the entrypoint logs a warning and passes an empty path.
  Empty cleanly disables persistence (dedupPersistPath returns "" when the base is empty),
  leaving in-memory dedup working. Creating the directory in the entrypoint rather than the
  watcher also keeps the failure at startup, logged once, instead of recurring on every
  30-second snapshot tick.
  - README: added --dedup-persist to the flags table and recorded in the snapshots section that
  the deployed install now enables it. That file already states what the operator passes or
  omits for --dedup-window, --reason, and --exclude-namespace, so leaving this one silent would
  have been drift.
  

## Testing

- Tested manually locally. Ran event watcher with --dedup-persist and on redploys event watcher did not realert on recent events.


deploy/shared/start-services.sh
```diff
@@ -66,6 +92,7 @@ start_event_watcher() {
       /usr/local/bin/k8s-event-watcher \
         --cluster-name="${EVENT_WATCHER_CLUSTER_NAME:-}" \
         --profiles-dir="${CREDENTIAL_PROXY_WORKSPACE_ROOT:-/opt/data}/profiles" \
+        --dedup-persist="${dedup_persist}" \
```
jayantid, line 95: This flag makes a previously-dead error path live. `newDedupCache` errors on any snapshot read failure that isn't "file not
  found" (`dedup.go:86-90`), and `main.go:663-671` treats that as fatal for the cluster — log, `cluster_up=0`, `continue`.
  `--dedup-persist` was always empty before, so nothing could reach that branch.

  **If this merges as-is:** healthy PVCs are fine, so you won't see it in testing. But on an unreadable snapshot — EIO on an
  NFS-backed RWX volume, a restored PVC with different ownership, a UID change on an image bump — that cluster silently drops out.
  With multiple clusters the survivors keep the process alive, so there's no short exit and no supervisor alert; the container
  stays Ready and one startup log line is the only signal. And since `restore()` only runs at process start, a transient error
  becomes permanent until someone restarts the pod.

  **Fix:** treat it like the corrupt-snapshot case already at `dedup.go:267-270` — log and start empty.

  ```go
  data, err := os.ReadFile(c.persistPath)
  if err != nil {
      if !errors.Is(err, os.ErrNotExist) {
          log.Printf("dedup: read %s (starting fresh): %v", c.persistPath, err)
      }
      return nil
  }

  Worst case becomes "re-alerts on the last hour once" instead of "never watched again" — the same trade-off your comment at line
  76 already makes.
  ```

jayantid (changes requested):

deploy/shared/start-services.sh
stalhaali: done

lapis2002:

Retesting as the prev test was aborted because of new commit.

jayantid (approved):

PR #607, opened 2026-08-10
test(operator): assert the shape of config-hash annotations, not their value

bradhoekstra:
# Pull Request

## Why This Change

The operator's golden files pinned the exact SHA-256 digests of four pod-template annotations:

```yaml
kubeagents.x-k8s.io/config-hash: 16bea89758dd8443112ac8fb2b6d98b7b5084693d9b54978e65556daad07b31a
kubeagents.x-k8s.io/fluent-bit-config-hash: 50923ab88e1741b0729c37837bc1c3869161e3161402494a4ee64cda3a2cfab3
kubeagents.x-k8s.io/proxy-policy-hash: db64678068cfed7481ee3b503145c688a2b49257e8e534565333c172185a750d
kubeagents.x-k8s.io/settings-config-hash: 9db5b5cb5fb8ec46b1a0572ef34ef27e0ed44dc1f787a4a76bdd75f43934c8b9
```

Those digests are derived from the rendered ConfigMaps, settings, and proxy policy — content the same golden files already capture in full. Any change to any of those inputs invalidates the digests, so the pinned values generate churn on unrelated PRs without catching a defect the rest of the diff would miss. A digest is also not something a reviewer can verify by reading; the only way to "check" it is to re-run the test that produced it.

What the test actually needs to guarantee is that the annotation is present and well formed, so the pod template rolls when the config changes.

## What Changed

**Files:**

- `k8s-operator/internal/testing/testutil/testutil.go`

**Added/Changed/Fixed:**

- Added `redactHashAnnotations`, which walks any annotation whose key ends in `-hash`, asserts the value matches `^[0-9a-f]{64}$`, and replaces it with `<SHA256_FOR_TESTS>` before the object is marshalled. A missing or malformed digest still fails the test, and it fails with a message naming the annotation rather than a 64-character diff.
- Applied it in `CleanAndMarshalResources` to object-level annotations plus the pod template of both `Deployment` and `StatefulSet` — the two workload shapes `buildPodTemplateSpec` feeds. The key-suffix match means a fifth hash annotation is covered the day it is added, without touching this file.
- Folded the pre-existing `leader_elect.py` omission into the same type switch, since it was already a `*corev1.ConfigMap` type assertion sitting in the same loop.
- Regenerated the two golden files to record the placeholder.

## Why This Matters

- Config-affecting PRs no longer carry an unexplained golden-file diff that a reviewer has to take on faith.
- The assertion is stronger where it counts: previously a digest that silently became the empty string would have been "fixed" by a routine `-update` run. Now that is a test failure.

## Context

Test-only. No operator behaviour changes — `buildPodTemplateSpec` still emits the real digests; only the test's view of them is redacted.

One thing reviewers should know: I hand-edited the eight golden lines rather than running `go test -update`. The `-update` path writes raw `sigs.k8s.io/yaml` output, whose list indentation differs from the prettier-formatted files on disk, so a regeneration rewrites all 1160 lines of each file and fails CI's prettier check. That trap predates this PR and is worth fixing separately — either by running the golden writer through prettier or by adding the expected files to `.prettierignore`.

I left `platformagent_manifests_test.go:451-460` alone. It passes literal `"abcd1234"` into `buildDeployment` and asserts it comes back out, which is a propagation test on a value the test itself chose — not a pinned digest, so it is not brittle in the same way.

## Testing

- `go test ./...` in `k8s-operator/` — passes.
- `go build ./...` — passes.
- Verified the new assertion actually fires by temporarily tightening the pattern to `{65}`: both golden cases fail with `annotation kubeagents.x-k8s.io/config-hash = "16bea..." , want a 64-character hex SHA-256 digest`, confirming the check runs against the real rendered value and is not vacuous.

---

Test-only change; no runtime or deployment impact. Generated with the help of Claude.

### DIFF
--- a/k8s-operator/internal/testing/testdata/platform/expected/platformagent-tagged.yaml
+++ b/k8s-operator/internal/testing/testdata/platform/expected/platformagent-tagged.yaml
@@ -584,10 +584,10 @@ spec:
   template:
     metadata:
       annotations:
-        kubeagents.x-k8s.io/config-hash: 550310a3ab2c9e35816fa1a8be64e68cc87377afc9609c0aa8a604e2b778c341
-        kubeagents.x-k8s.io/fluent-bit-config-hash: 50923ab88e1741b0729c37837bc1c3869161e3161402494a4ee64cda3a2cfab3
-        kubeagents.x-k8s.io/proxy-policy-hash: db64678068cfed7481ee3b503145c688a2b49257e8e534565333c172185a750d
-        kubeagents.x-k8s.io/settings-config-hash: 9db5b5cb5fb8ec46b1a0572ef34ef27e0ed44dc1f787a4a76bdd75f43934c8b9
+        kubeagents.x-k8s.io/config-hash: <SHA256_FOR_TESTS>
+        kubeagents.x-k8s.io/fluent-bit-config-hash: <SHA256_FOR_TESTS>
+        kubeagents.x-k8s.io/proxy-policy-hash: <SHA256_FOR_TESTS>
+        kubeagents.x-k8s.io/settings-config-hash: <SHA256_FOR_TESTS>
       creationTimestamp: null
       labels:
         app: platformagent-gateway
--- a/k8s-operator/internal/testing/testdata/platform/expected/platformagent-telemetry.yaml
+++ b/k8s-operator/internal/testing/testdata/platform/expected/platformagent-telemetry.yaml
@@ -555,10 +555,10 @@ spec:
   template:
     metadata:
       annotations:
-        kubeagents.x-k8s.io/config-hash: fff685f46be77b2e136806220db6dff17c663aee1077db530ad287ff9a158c14
-        kubeagents.x-k8s.io/fluent-bit-config-hash: 50923ab88e1741b0729c37837bc1c3869161e3161402494a4ee64cda3a2cfab3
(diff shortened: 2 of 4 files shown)


adamparco:
LTGTM, but I have no perms.

bradhoekstra:
> LTGTM, but I have no perms.

You do, just reply with /lgtm.

adamparco (approved):
LGTM. Test-only change that swaps pinned SHA-256 digests in the golden files for a shape assertion (`^[0-9a-f]{64}$`) plus a placeholder — exactly the right trade for killing golden-file churn, and the suffix match covers future `-hash` annotations for free. Verified locally: `go build` and the operator test suite pass with `main` merged in, and the redaction is non-vacuous (a tightened pattern fails the test).

One non-blocking suggestion: all four annotations now collapse to the same placeholder, so the golden files no longer show that the four digests are distinct — a future reconcile-path wiring slip (e.g. the same hash passed twice into `reconcileWorkload`'s four same-typed string params) would keep every check green. Identity-preserving placeholders (`<SHA256-1>`, `<SHA256-2>`, …) or a single pairwise-distinctness test would close that without reintroducing churn. Fine as a follow-up.

(Also worth a quick refresh: the PR body lists 3 files/8 lines but the rebase picked up the telemetry golden file, making it 4 files/12 lines.)

PR #609, opened 2026-08-10
fix(stockout investigator) take left over feedback from #541 into account

tlipski:
Dedup entries expire on the window they were written under, so a short-window route no longer prunes a daily route's records and reopens an incident it filed. Sink names come from config, not a guess; the check is off pending a dependency. Bound the delivery map, and make verify.sh assert what the adapter did within this run's log window. Memoise scenario resolvers; skip bare-metal shapes.

# Pull Request

## Why This Change

Follow-up to the review on #541. Each item is a case where the alert path looked healthy while silently doing the wrong thing — the failure mode that makes an investigation untrustworthy rather than merely broken.

All routes share one dedup registry, and pruning used the *calling* route's window, so a five-minute route expired a daily route's entries on its way past them and the daily route re-opened an incident it had already filed — with nothing in either route's logs naming the culprit. The log-sink check derived a name (`hermes-pubsub-<route>-sink`) that nothing creates, so the one check meant to spot a broken ingress path cried wolf on every healthy start. `verify.sh` grepped for the route name, and since every line carries it — including the ones reporting the alert being thrown away — a filtered or deduplicated alert reported SUCCESS. Two scenarios resolved their machine shape inside hook subshells that run up to three times, so the ComputeClass could pin one shape while the alert named another.

## What Changed

**Files:**

- `agentplugins/pubsub-platform/files/platforms/pubsub/adapter.py` (+138/-24) — record `window_seconds` on each registry entry and add `_entry_window()` so pruning respects the window an entry was written under; add `_remember_delivery()` with a 256-entry cap, called only on the gateway-hosted path and no longer storing the payload; add `_warn_on_inert_config()` to warn at startup when `require_skills` is paired with `dispatch: kanban`; take the sink name from config instead of deriving it; leave `_check_log_sink_exists()` defined but uncalled, with the restore snippet inline.
- `agentplugins/pubsub-platform/tests/test_dedup.py` (+43) — three cases: a short-window route must not expire a long-window route's entry, an entry records the window it was written under, and a legacy entry without one still expires.
- `agentplugins/pubsub-platform/README.md` (+20/-5) — `require_skills` documented as `dispatch: api` only; new `sink` row and a paragraph stating the presence check is disabled and `sink` is recorded but unread.
- `agentplugins/pubsub-platform/files/platforms/pubsub/README.md` (+23/-11) — `sink` added to the parameter table, `query` and the "Resource Presence Verification" bullet corrected so neither claims log sinks are checked.
- `agentplugins/gke-stockout-investigator/verify.sh` (+93/-13) — match on what the adapter *did* rather than on the route name, across eight outcomes; scope the log read to this run with `--since` computed from a pre-publish timestamp; fresh per-run controller name plus `kind: ReplicaSet`; match the `require_skills` refusal on its own clause so the new startup warning cannot be mistaken for it; unscoped tail only on failure; real exit code.
- `agentplugins/gke-stockout-investigator/scenarios/lib/common.sh` (+48) — new `scenario_memo` helper (run-scoped `mktemp -d`, `%q`-quoted values, `EXIT` trap) plus a note in the hook contract that hooks run in subshells and `scenario_manifest` runs more than once.
- `agentplugins/gke-stockout-investigator/scenarios/03-large-vm-shape-scarcity.sh` (+27/-3) — wrap the shape resolver in `scenario_memo`; exclude `-lssd` anywhere in the name and `-metal` outright.
- `agentplugins/gke-stockout-investigator/scenarios/04-missing-zone-fallback.sh` (+18/-2) — wrap the family resolver in `scenario_memo`.
- `agentplugins/gke-stockout-investigator/templates/agentplugin.yaml` (+40/-6) — drop `approvals: mode: off` and `require_skills`, each with the reasoning left in place; add `sink`.
- `agentplugins/gke-stockout-investigator/values.yaml` (+4) — add `pubsub.sink`, documented as currently unread.
- `agentplugins/gke-stockout-investigator/install.sh` (+11/-4) — pass `--set pubsub.sink="$SINK_NAME"` so the chart gets the name only the installer knows.
- `agentplugins/gke-stockout-investigator/README.md` (+4) — note that `verify.sh` exits non-zero, names the suppression that swallowed the alert, and varies the workload per run.

**Added/Changed/Fixed:**

- Registry entries expire on their own terms whichever route prunes. Legacy entries fall back to the caller's window only for that route's own records, and to the 24h default otherwise, so guessing short cannot expire them early.
- The sink name is configured and flows installer → values → chart. **The presence check itself is disabled**: it needs `google-cloud-logging`, which the agent image does not carry, and supplying it — image dependency or Hermes lazy installs — is a pending decision.
- `_delivery_info` is bounded and no longer written on the kanban and API paths, which never read it back through `send()`.
- `approvals: mode: off` removed — it lands on the whole `platform` profile, and the kanban path needs no bypass. `require_skills` removed because it cannot work under `dispatch: kanban`; the adapter now warns instead of implying a guarantee.
- Scenario 03 no longer selects bare-metal shapes: `endswith("-lssd")` missed `c4-highmem-288-lssd-metal`, and since the rule picks the largest shape, metal won every time and Autopilot cannot provision it — so that scenario staged no stockout at all.

## Why This Matters

- A suppressed alert and a dispatched one are now distinguishable, in the logs and in `verify.sh`'s exit code. Previously both read as success.
- One incident stays one investigation across routes with different windows — the case that produced duplicate remediation PRs.
- The plugin no longer grants itself profile-wide approval bypass, or advertises a `require_skills` guarantee nothing enforces.
- Scenario 03 actually stages the failure it claims to, and 03/04 can no longer describe a different shape in the manifest than in the alert.

## Context

Follows up review findings on #541. The disabled sink check is deliberately reversible: `_check_log_sink_exists` is kept with the exact call site to restore, and `sink` stays plumbed end to end, so re-enabling is a one-line adapter change with no reinstall once the dependency question is settled.

## Testing

- 30/30 `tests/test_dedup.py`. `prettier --check` and `make docs-check` clean. Chart renders and its `spec.config` parses.
- Verified by AST against the built OCI image that the sink check has zero runtime call sites.
- **e2e on the ka-demo fleet:** scenarios 01, 03 and 04 end to end. 01 and 03 produced correct remediation PRs; 04 correctly stood down against an existing open PR. Live registry confirmed `window_seconds: 86400`.
- **e2e on ka-dev-mgmt → ka-production:** reinstalled both plugins, then `verify.sh` PASSED (exit 0). A negative run with a deliberately wrong cluster name, seconds after that success, correctly FAILED with "the adapter filtered the alert out" (exit 1) — the pre-fix `--tail=200` code would have matched the previous run's line and passed. Scenario 04 wedged 4 pods, dispatched an investigation, and produced a correct multi-family ComputeClass remediation PR; a concurrent repeat deduplicated against it, leaving exactly one PR.

---

Behavioural risk is low and mostly in the direction of *less* suppression: the dedup change only extends entry lifetimes, and the delivery-map change touches paths whose entries were never read. The one deliberate regression is the sink presence check, which is off until the `google-cloud-logging` dependency is decided — it was reporting a false alarm on every start beforehand, so nothing observable is lost.


mplakhtiy:
## Findings

### 1. Scenario 05 still resolves inside three separate hooks, which is the bug this PR fixes for 03 and 04 — Medium

`agentplugins/gke-stockout-investigator/scenarios/05-missing-ondemand-floor.sh:69`, `:130`, `:159`

`_resolve_spot_only_shapes` is called bare from `scenario_manifest`, `scenario_reasons` and
`scenario_notes`, none of them wrapped in the `scenario_memo` helper this PR introduces. The
resolver is exactly the kind the PR's own contract note now forbids leaving unmemoised — it probes
live capacity with `gcloud beta compute advice capacity` per candidate shape, so its answer is a
property of the region at the moment of the call:

```
lib/common.sh:35-38  A hook that resolves something at run time must wrap the resolver in
                     scenario_memo (below) or it will resolve repeatedly, and may not resolve
                     the same way twice.
```

`scenario_manifest` runs twice per invocation (`_manifest_part infra`, `_manifest_part workload`),
`scenario_reasons` once from `_inner_payload`, `scenario_notes` once from `_report` — four
resolutions, each up to four `gcloud` invocations, so up to sixteen API calls per run where one
would do.

**Failure scenario.** `h4d-standard-192` has no Spot capacity in the region when
`scenario_manifest` runs, so the ComputeClass `batch-spot-only-class` pins it. Spot for that shape
becomes obtainable in the seconds before `publish_alert` calls `scenario_reasons`, so the resolver
now returns only `h4d-highmem-192` and the alert's `rejectedMigs` MIG name and
`napFailureReasons` describe a shape the ComputeClass does not name. `scenario_notes` then prints
a third answer. The agent is asked to diagnose a workload whose alert and manifest disagree, which
reads as a malformed scenario rather than the Rule D case being staged.

### 2. `--tail=200` can truncate the very lines the new verify.sh checks depend on — Low

`agentplugins/gke-stockout-investigator/verify.sh:117-118`

`--since` already scopes the read to this run, which is the fix. The `--tail=200` alongside it does
not add safety — it caps the read from the *wrong end*. Every line the check matches
("Received message on route", "filed alert on route", "filtered out by expression", "Duplicate
message detected") is emitted by the adapter within milliseconds of the publish, i.e. at the
*start* of the window, and `--tail` keeps the *last* N lines.

**Failure scenario.** `dispatch: kanban` files the task and the board's dispatcher spawns the
worker into the same `platform-agent` container, whose output lands in the same log stream. If
that worker emits more than 200 lines during the remaining seconds of the 10-second window, the
adapter's own lines are dropped from `LOGS`, `ROUTE_LOGS` comes back empty, and the script prints
"✗ FAILED: no sign the adapter saw the message at all" and exits 1 on a completely healthy
install — the smoke test's new exit code turning into a false red. I did not reproduce the line
volume (it needs a live gateway mid-turn), so the trigger is a busy gateway rather than a
certainty; dropping `--tail` entirely, or raising it well past any plausible burst, removes the
question.

### 3. `_check_resources`'s docstring still advertises the log-sink check the same function just disabled — Low

`agentplugins/pubsub-platform/files/platforms/pubsub/adapter.py:182`

```python
    def _check_resources(self, name: str, sub_cfg: dict, project_id: str) -> str:
        """Check presence of GCP Pub/Sub topics, subscriptions, and log sinks without creating them."""
```

The body twelve lines below now explains at length that the sink check is deliberately not called.
Both READMEs were corrected on precisely this point in this PR —
`files/platforms/pubsub/README.md:39` now reads "The equivalent log-sink check is currently
disabled" — so the docstring is the one place the old claim survives, and it is the place a reader
of the adapter meets first.

**Failure scenario.** Someone debugging why a broken sink went unnoticed reads the docstring,
concludes the sink is verified at connect, and looks for the missing warning in a log filter rather
than discovering that no check runs at all.

### 4. Nothing tests that a gateway-hosted route still delivers to its configured platform — Low

`agentplugins/pubsub-platform/files/platforms/pubsub/adapter.py:866-871`, `agentplugins/pubsub-platform/tests/test_dedup.py`

The three new tests cover the dedup half of the change well. The delivery half is the part that
moved: `_remember_delivery` is now called from exactly one place, below the `if target_profile:`
branch, and `send()` at `:383` silently falls back to `deliver: log` when the map has no entry for
the chat id. That fallback is unobservable from the outside — the answer is still written, just to
the log rather than to the configured destination — and no test pins the placement. The test module
already stubs the whole gateway, so a case asserting `send()` returns the configured `deliver` type
for a route with no `agent_profile` costs a few lines.

**Failure scenario.** A later refactor moves the `_remember_delivery(...)` call below
`await self.handle_message(event)` — a natural-looking tidy-up, since that is where the dispatch
happens. Every route with `deliver: google_chat` then writes its answer to the container log
instead of to chat, the suite stays green, and the regression surfaces as "the agent stopped
replying" with nothing in the diff that looks related.

### 5. The scenarios README's "Adding a scenario" section does not mention the new memoisation requirement — Low

`agentplugins/gke-stockout-investigator/scenarios/README.md:113-119`

The PR adds a real contract for scenario authors — hooks run in subshells, `scenario_manifest` runs
more than once, and a run-time resolver must be wrapped in `scenario_memo` — but writes it only
into `lib/common.sh`'s header comment. `AGENTS.md` puts "How to develop a specific directory" in
that directory's `README.md`, and this section is the canonical instruction set: "Copy the shortest
one, `10-false-signal.sh` … Define `scenario_manifest` to echo the YAML that wedges,
`scenario_reasons` to echo the `rejectedMigs` … fragment". A sentence and a pointer to
`scenario_memo` would close it.

**Failure scenario.** The next scenario author follows the README, copies `10-false-signal.sh`
(which has no resolver and therefore no example of the pattern), adds a `gcloud`-backed resolver
called from both hooks, and reproduces finding 1 in a sixth file.


---

mplakhtiy (approved):

PR #611, opened 2026-08-10
feat(ci): add 3-hour cron schedule and smart commit resolution to RC pipeline

eLeontev:
# Pull Request

## Why This Change

- Automates the end-to-end Release Candidate (RC) testing and validation pipeline on a 3-hour cron schedule.
- Prevents spurious CI failures when commits are rapidly merged to `main` by dynamically resolving the latest commit that already has prebuilt images in GHCR, rather than blindly taking `HEAD`.
- Automatically skips redundant 30-minute deployment and E2E test runs during cron triggers if the latest commit on `main` is already marked with a `*_validated` tag.
- Preserves mandatory `commit_sha` input for manual `workflow_dispatch` runs.
- Eliminates duplicate image names and registry URL construction by centralizing release helpers in `scripts/release/common.sh`.

## What Changed

**Files:**

- `.github/workflows/rc-release-pipeline.yml`
- `.github/workflows/rc-create-tag.yml`
- `scripts/release/common.sh`
- `scripts/release/resolve_rc_tag.sh`
- `scripts/release/verify_candidate_images.sh`
- `scripts/release/README.md`

**Added/Changed/Fixed:**

- Added `schedule: - cron: "0 */3 * * *"` to `.github/workflows/rc-release-pipeline.yml`.
- Added auto-skip logic (`is_commit_already_validated()` and `skip_rc` output) so recurring cron runs skip deployment and E2E tests if the latest candidate commit was already validated in a previous RC run.
- Added `if: needs.step-1-create-tag.outputs.skip_rc != 'true'` to steps 2, 3, and 4 in `rc-release-pipeline.yml` and tagging steps in `rc-create-tag.yml`.
- Kept `commit_sha` mandatory (`required: true`) for manual runs; set `required: false` on `workflow_call` for automated schedule runs.
- Centralized `DEFAULT_REGISTRY_PREFIX`, `DEFAULT_RELEASE_REPO`, `REQUIRED_RELEASE_IMAGES`, and helper functions in `scripts/release/common.sh`.
- Added `find_latest_built_commit()` to scan recent commits on `main` (`FETCH_HEAD`) and return the newest commit with verified container images (`k8s-operator`, `platform-agent`).
- Refactored `verify_candidate_images.sh` and `resolve_rc_tag.sh` to source `common.sh`.
- Documented 3-hour cron cadence, smart resolution, and redundant run skip behavior in `scripts/release/README.md`.

## Why This Matters

- Ensures automated, continuous RC validation without manual triggering.
- Saves significant cloud resources and execution time by skipping redundant test runs on unchanged/already validated commits.
- Avoids race conditions with active image build workflows.
- Keeps release scripts clean, modular, and DRY.

## Context

Follow-up to Release Candidate CI/CD automation pipeline.

## Testing

Locally validated across all key scenarios:
- **Cron Auto-resolve & Skip Validation**: Successfully scans fetched upstream `main`, resolves newest commit with published GHCR images (`ebc956b8`), and correctly flags `skip_rc: true` when marked with `*_validated`.
- **Manual Runs**: Successfully accepts explicit valid commit SHAs (`HEAD`, full SHA) and fails with clear error messages on missing or invalid inputs.
- **Image Verification**: Validated manifest checks against GHCR and custom Google Artifact Registry (GAR) endpoints.
- **Repository Health**: `make docs-check` passes cleanly.

---

### DIFF
--- a/.github/workflows/rc-create-tag.yml
+++ b/.github/workflows/rc-create-tag.yml
@@ -30,6 +30,8 @@ on:
         value: ${{ jobs.create-tag.outputs.commit_sha }}
       rc_tag:
         value: ${{ jobs.create-tag.outputs.rc_tag }}
+      skip_rc:
+        value: ${{ jobs.create-tag.outputs.skip_rc }}
 
 jobs:
   create-tag:
@@ -40,6 +42,7 @@ jobs:
     outputs:
       commit_sha: ${{ steps.resolve_target.outputs.commit_sha }}
       rc_tag: ${{ steps.resolve_target.outputs.rc_tag }}
+      skip_rc: ${{ steps.resolve_target.outputs.skip_rc }}
 
     steps:
       - name: Checkout repository
@@ -59,12 +62,14 @@ jobs:
         run: ./scripts/release/resolve_rc_tag.sh
 
       - name: Verify Prebuilt Container Images Exist in Registry
+        if: steps.resolve_target.outputs.skip_rc != 'true'
         env:
           COMMIT_SHA: ${{ steps.resolve_target.outputs.commit_sha }}
           REGISTRY_PREFIX: ${{ vars.REGISTRY_PREFIX }}
         run: ./scripts/release/verify_candidate_images.sh
 
       - name: Create Git Tag & Push
+        if: steps.resolve_target.outputs.skip_rc != 'true'
         env:
           COMMIT_SHA: ${{ steps.resolve_target.outputs.commit_sha }}
           RC_TAG: ${{ steps.resolve_target.outputs.rc_tag }}
--- a/.github/workflows/rc-create-tag.yml
+++ b/.github/workflows/rc-create-tag.yml
@@ -21,6 +21,11 @@ on:
         description: "Release Candidate Tag (Optional, Default: Auto-generated rc_YYMMDDHHMM)"
         required: false
         type: string
+      is_scheduled:
+        description: "Flag indicating whether run was triggered by automated cron schedule"
+        required: false
+        type: boolean
+        default: false
     secrets:
       RELEASE_BOT_TOKEN:
         description: "Release Bot PAT Token"
@@ -56,6 +61,7 @@ jobs:
         env:
           COMMIT_SHA: ${{ inputs.commit_sha }}
           RC_TAG: ${{ inputs.rc_tag }}
+          IS_SCHEDULED: ${{ inputs.is_scheduled }}
           GH_ORG: ${{ vars.GH_ORG }}
           GH_REPO: ${{ vars.GH_REPO }}
           REGISTRY_PREFIX: ${{ vars.REGISTRY_PREFIX }}
--- a/.github/workflows/rc-create-tag.yml
+++ b/.github/workflows/rc-create-tag.yml
@@ -14,8 +14,8 @@ on:
   workflow_call:
     inputs:
       commit_sha:
-        description: "Target Commit SHA to tag (Mandatory)"
-        required: true
+        description: "Target Commit SHA to tag (Optional for automated schedule, Mandatory for manual runs)"
+        required: false
         type: string
       rc_tag:
         description: "Release Candidate Tag (Optional, Default: Auto-generated rc_YYMMDDHHMM)"
@@ -53,6 +53,9 @@ jobs:
         env:
           COMMIT_SHA: ${{ inputs.commit_sha }}
           RC_TAG: ${{ inputs.rc_tag }}
+          GH_ORG: ${{ vars.GH_ORG }}
+          GH_REPO: ${{ vars.GH_REPO }}
+          REGISTRY_PREFIX: ${{ vars.REGISTRY_PREFIX }}
         run: ./scripts/release/resolve_rc_tag.sh
 
       - name: Verify Prebuilt Container Images Exist in Registry
--- a/.github/workflows/rc-release-pipeline.yml
+++ b/.github/workflows/rc-release-pipeline.yml
@@ -27,6 +27,7 @@ jobs:
     with:
       commit_sha: ${{ inputs.commit_sha }}
       rc_tag: ${{ inputs.rc_tag }}
+      is_scheduled: ${{ github.event_name == 'schedule' }}
     secrets:
       RELEASE_BOT_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
 
--- a/.github/workflows/rc-release-pipeline.yml
+++ b/.github/workflows/rc-release-pipeline.yml
@@ -1,6 +1,8 @@
 name: RC Release E2E Pipeline
 
 on:
+  schedule:
+    - cron: "0 */3 * * *"
   workflow_dispatch:
     inputs:
       commit_sha:
--- a/.github/workflows/rc-release-pipeline.yml
+++ b/.github/workflows/rc-release-pipeline.yml
@@ -33,6 +33,7 @@ jobs:
   step-2-deploy-env:
     name: Step 2 - Provision RC Environment
     needs: step-1-create-tag
+    if: needs.step-1-create-tag.outputs.skip_rc != 'true'
     uses: ./.github/workflows/rc-deploy-environment.yml
     permissions:
       contents: read
@@ -45,6 +46,7 @@ jobs:
   step-3-run-e2e-tests:
     name: Step 3 - Run E2E Tests
     needs: [step-1-create-tag, step-2-deploy-env]
+    if: needs.step-1-create-tag.outputs.skip_rc != 'true'
     runs-on: ubuntu-latest
     environment: rc
     concurrency:
@@ -103,6 +105,7 @@ jobs:
   step-4-tag-validated:
     name: Step 4 - Tag Validated Release
     needs: [step-1-create-tag, step-3-run-e2e-tests]
+    if: needs.step-1-create-tag.outputs.skip_rc != 'true'
     uses: ./.github/workflows/rc-tag-validated.yml
     permissions:
       contents: write
--- a/scripts/release/README.md
+++ b/scripts/release/README.md
@@ -4,13 +4,23 @@ This directory contains executable scripts supporting the Release Candidate (RC)
 
 ## Overview of Scripts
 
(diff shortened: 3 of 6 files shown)


mplakhtiy (approved):
Reviewed at `628b112cc29f472cf333d951bc9f3275273ab980`, with `main` (`117313ee`) merged in locally. Findings below — none blocking.

### 1. A commit that fails E2E is re-tested by every cron tick, with no backoff — Medium

`scripts/release/resolve_rc_tag.sh:39-44`, `.github/workflows/rc-release-pipeline.yml:4-5`

The only thing that suppresses a run is a `*_validated` tag on the resolved commit, and that tag is written exclusively by `step-4-tag-validated`, which needs `step-3-run-e2e-tests` to have passed. A commit that fails E2E therefore never becomes skippable, and the next tick resolves the same commit and starts over. There is no failure memo, no attempt counter, and no exponential backoff.

Each non-skipped run reaches `step-2-deploy-env`, whose `scripts/release/provision_rc_environment.sh` is two lines: `k8s-operator/scripts/teardown.sh --no-confirm` followed by `k8s-operator/scripts/provision.sh --no-confirm`.

**Failure scenario.** A regression that breaks `execute_e2e_tests.sh` merges to `main` on Friday at 18:00 and nothing else merges over the weekend. The scan picks the same tip commit at 21:00, 00:00, 03:00 … — 16 ticks by Sunday evening. Each one destroys and rebuilds the `platform-agent-host` GKE cluster in `kube-agents-rc` and then fails E2E, producing 16 red workflow runs and 16 full cluster rebuilds for a defect that was already known after the first one.

### 2. The schedule removes the only human gate in front of an unattended teardown of `kube-agents-rc` — Medium

`.github/workflows/rc-release-pipeline.yml:4-5`

Before this change the pipeline was `workflow_dispatch`-only: a person chose the commit and the moment. With the cron, the newest commit on `main` is deployed automatically. The `rc` environment that gates `step-2-deploy-env` and `step-3-run-e2e-tests` has an empty `protection_rules` list, so there is no reviewer approval to intercept a run. `step-2-deploy-env` checks out the candidate commit and executes `teardown.sh --no-confirm` and `provision.sh --no-confirm` **from that commit**, authenticated as `github-deploy-sa@kube-agents-rc.iam.gserviceaccount.com` via WIF — and `117313ee`, the current tip of `main`, added an `--admin` flag to the WIF setup script explicitly for this pipeline.

**Failure scenario.** A PR that widens a `gcloud` delete scope in `k8s-operator/scripts/teardown.sh` is merged with the usual review. Nobody schedules an RC run, because nobody needs one. Within three hours the cron picks that commit up and runs the modified teardown against `kube-agents-rc` with a privileged service account. Under the previous trigger model the same commit sits inert until an engineer chooses to exercise it.

This is arguably the point of the PR, but it deserves an explicit decision — a required reviewer on the `rc` environment, or an allowlist of paths whose changes may auto-deploy, would preserve the cadence without giving up the gate.

### 3. Every non-skipped run pushes a fresh `rc_*` tag, so tags grow without bound — Low

`scripts/release/resolve_rc_tag.sh:51-54`, `scripts/release/create_release_tag.sh:22`

The auto-generated tag is `rc_$(date -u +'%y%m%d%H%M')_${SHORT_SHA}`. The timestamp defeats `ensure_git_tag`'s idempotency check: for the same commit at a different tick, the name differs, so `git rev-parse "${rc_tag}"` misses and a new annotated tag is created and pushed. Nothing prunes them; the repo has six `*_validated` tags accumulated over roughly six weeks of manual runs, so this is a step change in tag volume.

**Failure scenario.** The weekend from finding 1 leaves 16 distinct `rc_2608xxxxxx_<sha>` tags all pointing at one broken commit, none of them validated. At the intended steady state — a merge between most ticks — the pipeline adds up to 16 tags a day (`rc_*` plus `rc_*_validated`), so `git fetch --tags` and the releases page accumulate thousands of refs a year with no retention policy. No workflow triggers on `rc_*` (the tag-triggered workflows match `v*.*.*` and `staging/**`), so this is clutter rather than a build storm.

### 4. `--depth=30` shallows whatever clone it runs in, which matters when the script is run locally — Low

`scripts/release/common.sh:71`

`find_latest_built_commit` fetches with `git fetch <url> main +refs/tags/*:refs/tags/* --depth=30`. Against a repository that is not already shallow, `--depth` converts it: I reproduced this on a scratch full clone, where `git rev-parse --is-shallow-repository` flipped from `false` to `true` and `git log` collapsed to the fetched depth. The `+` on the tag refspec also force-overwrites local tags. In CI this is harmless — creating and pushing an annotated tag from the resulting shallow repository still works, so `create_release_tag.sh` is unaffected — and the `fetch-depth: 0` checkout in `rc-create-tag.yml` makes the depth limit pointless there anyway.

**Failure scenario.** A maintainer reproduces the scan the way the PR description describes doing it (“resolves newest commit with published GHCR images (`ebc956b8`)”) by running `IS_SCHEDULED=true ./scripts/release/resolve_rc_tag.sh` in their working clone. The clone is silently truncated to 30 commits and their local tags are force-updated; `git log`, `git blame` on anything older, and `git bisect` stop working until they notice and run `git fetch --unshallow`.

### 5. The hard-coded upstream fallback in `get_target_repo` turns a loud local failure into a push to `gke-labs/kube-agents` — Low

`scripts/release/common.sh:16-17`, `scripts/release/common.sh:143`

`get_target_repo()` ends with `echo "${DEFAULT_RELEASE_REPO}"`. The code it replaced in `ensure_git_tag` used `target_repo="${GITHUB_REPOSITORY:-}"`, which outside Actions left the value empty and made the push URL `https://github.com/.git` — a guaranteed, obvious failure. The new fallback makes the same invocation target the real upstream repository.

**Failure scenario.** A maintainer runs `COMMIT_SHA=<sha> RC_TAG=rc_local_test ./scripts/release/create_release_tag.sh` on their laptop to try the script. With `GH_ORG`/`GH_REPO`/`GITHUB_REPOSITORY` all unset and a git credential helper holding a token with write access (the `gh` helper is the common setup for people who work on this repo), the script pushes an `rc_local_test` tag to `github.com/gke-labs/kube-agents` instead of erroring out.

### 6. `0 */3 * * *` sits in GitHub's most contended scheduling slot — Low

`.github/workflows/rc-release-pipeline.yml:5`

GitHub delays scheduled workflows during periods of high load, and the top of the hour is the heaviest slot; runs can start tens of minutes late or be dropped entirely.

**Failure scenario.** The 12:00 tick is delayed past 12:40 or skipped; an engineer reading `scripts/release/README.md`, which promises a “**Scheduled Cadence (every 3 hours…)**”, concludes the pipeline is broken and dispatches a manual run against the same commit, duplicating a full cluster rebuild. Moving the minute off the hour (`17 */3 * * *`) removes the contention, and the README should say the cadence is best-effort.

### 7. The new shell logic ships with no automated coverage — Low

`scripts/release/common.sh`, `scripts/release/resolve_rc_tag.sh`

The PR adds roughly 95 lines of branching shell — registry-prefix precedence, repo resolution, a 30-commit registry scan with three fallback revisions, and the skip decision — and nothing in the repository exercises it. There is no test file anywhere that references `common.sh`, `resolve_rc_tag.sh`, or `verify_candidate_images.sh`; `scripts/release/` holds only scripts and a README. `shellcheck` is the sole gate, and it only catches syntax and quoting.

**Failure scenario.** A later edit inverts the precedence in `get_registry_prefix()` so `REGISTRY_PREFIX` no longer wins over the derived default. `shellcheck` stays clean and every PR check stays green; the first signal is a 3-hourly cron that tears down the RC cluster and then fails image verification. The paths I exercised by hand for this review (scheduled resolve with a stubbed `docker`, the skip branch, the empty-scan branch, the missing-input branch) are exactly the four cases a small harness could pin.

### Not findings, for the record

- **The resolve step and the verify step derive the registry prefix from different environment.** `rc-create-tag.yml:61-67` passes `GH_ORG`/`GH_REPO` to `Resolve Commit SHA and Tag Name` but `rc-create-tag.yml:72-74` does not pass them to `Verify Prebuilt Container Images Exist`. With the environment as configured (`GH_ORG=gke-labs`, `GH_REPO=kube-agents`, `REGISTRY_PREFIX` unset) both paths land on `ghcr.io/gke-labs/kube-agents`, so the two steps agree. It only diverges if someone repoints the vars at a different repo than the one running the workflow.
- **`docker manifest inspect` without a registry login.** The GHCR packages are anonymously pullable — a token from `ghcr.io/token` with no credentials returns HTTP 200 on the manifests for the last six commits on `main`, for both required images.
- **`git tag --points-at <sha> "*_validated"` against annotated tags.** The `*_validated` tags in the repo are annotated; `--points-at` peels them. The tags are present locally because the `rc-create-tag.yml` checkout uses `fetch-depth: 0` and `find_latest_built_commit` refetches `+refs/tags/*` before the check.
- **`if: needs.step-1-create-tag.outputs.skip_rc != 'true'` on jobs 2-4 does not run them when step 1 fails.** A job-level `if` without a status-check function still carries the implicit `success()` on `needs`.
- **`if [ -z "${COMMIT_SHA}" ]` after `COMMIT_SHA=$(find_latest_built_commit)` (`resolve_rc_tag.sh:35-37`) is unreachable.** `set -e` aborts on the failing command substitution first. Likewise the `else` arms guarded by `[ -n "${target_repo}" ]` in `find_latest_built_commit` and `ensure_git_tag` are dead, because `get_target_repo()` never returns an empty string. Harmless, just noise.
- **Cron and manual dispatch contending for `concurrency: group: rc-release-pipeline`.** The pipeline's own timeouts cap a run at roughly 95 minutes, well under the 180-minute interval, so a cron tick cannot arrive while an earlier one still occupies the group.
- **Docs placement.** `docs/site/src/content/docs/deploy/release-versioning.md:12-13` already names `scripts/release/README.md` as the canonical reference for how RC builds are created and tagged, and that is the file this PR extends. `reference/cron-jobs.md` is the generated catalogue of in-cluster agent jobs, not of GitHub Actions schedules.
- **Scan window versus merge commits.** `git log -n 30` is not `--first-parent`, but recent `main` is linear (squash merges) and every recent commit has images, so the loop matches on its first iteration.

---

eLeontev:
The follow up PR will be created based on the PR review. @mplakhtiy thanks for the feedback.

PR #615, opened 2026-08-10
feat(terraform): add drift-pubsub module for audit-log ingress

stalhaali:
## Why This Change

  Provisions the Cloud Logging → Pub/Sub delivery path the drift detector (CUJ 3) consumes: a project-scoped Log Router sink over GKE audit logs, the drift-audit topic and pull subscription, and the IAM bindings that let the sink publish and the
  detector subscribe.

  GKE audit logs are not readable from the Kubernetes API. The control plane is managed, so the API server's audit backend is not the operator's to configure and the stream surfaces only in Cloud Logging. A sink is therefore the only ingress,
  which is why this is infrastructure rather than an informer inside the detector.

  ## What Changed

  New reusable Terraform module provisioning six resources: the Log Router sink, the drift-audit topic and pull subscription, and three IAM bindings.

  **Files:**

  - `terraform/modules/drift-pubsub/{main,variables,outputs,versions}.tf`
  - `terraform/modules/drift-pubsub/README.md`
  - `docs/README.md` (inventory row)
  - `AGENTS.md`, `docs/designs/semver-deployment-versioning.md`, `examples/gitops-repo/clusters/cluster-a/provisioning/README.md` — module inventories that enumerated four modules

  **Added/Changed/Fixed:**

  - Sink filter scoped to mutating calls on `k8s_cluster`, optionally narrowed by `cluster_names`.
  - `exclude_machine_lease_heartbeats` (default `true`) — drops `coordination.k8s.io` Lease writes by machine identities.
  - `message_retention_duration` defaults to `2678400s` (31 days), the subscription maximum. Pub/Sub's own default is 7 days; retention that lapses drops drift events with nothing downstream to record the gap.
  - `expiration_policy.ttl = ""` so the subscription does not self-delete after 31 idle days, and a `retry_policy` so parse failures back off rather than hot-loop.
  - `roles/pubsub.viewer` alongside `subscriber`: `subscriber` grants `subscriptions.consume` but not `subscriptions.get`, which clients calling `Subscription.Exists` need.

  ## Why This Matters

  - **The lease exclusion is measured, not guessed.** On a live two-cluster project over a 15-minute window, `leases.update` + `leases.create` were 9,558 of 10,000 sampled mutating calls — 95.6%. The 10,000 is the query's row cap rather than the
  window's true total, so it fixes the ratio but not the volume; a separate untruncated count put the surviving stream at 623 calls per 15 minutes, roughly 60k/day against ~1.35M/day unfiltered. A Lease is created at runtime by the controller
  holding it and never applied from a manifest, so it has no Git-side object and cannot be drift.
  - **The exclusion is scoped by principal, not by kind.** A human `kubectl patch lease` still reaches the detector. Both principal clauses are load-bearing: `^system:` alone leaves `container-engine-robot` behind, which made 287 Lease writes in
  the same window and would have inflated the surviving stream by 65%.
  - **Nothing else is filtered at the sink.** ConfigMap heartbeats are 445 of the 623 surviving calls (71%), but ConfigMaps genuinely are GitOps-managed — that is a classification problem for the detector, where it is a tunable allowlist, not a
  Log Router filter you cannot see past. They are all `system:`-tier and cost nothing to drop downstream.
  - **The sink's publisher grant fails silently.** Without it Log Router raises no error and the topic receives nothing, which is indistinguishable from "no drift happened." It is the most likely cause of a freshly applied pipeline delivering
  zero messages.

  ## Context

  Prerequisite (T0/T1 infrastructure) for CUJ 3 noise-filtered drift triage. The consumer that pulls this subscription is the next change.

  Deliberately **not** wired into `terraform/examples/full-install/` yet: `pubsub.googleapis.com` is currently enabled there only when `enable_google_chat = true`, so composing this module would break a Chat-less install. Moving Pub/Sub out of
  that conditional is a separate follow-up, noted in the module README.

  ## Testing

  CI covers `terraform fmt -check -recursive` and per-module `init`/`validate` (`.github/workflows/validate.yml`). Neither proves the sink publishes, so the module was applied against `stalhaali-gkedemos` and exercised end to end.

  The module was sourced by git ref, not by local path, so the artifacts below can only reflect this branch's code:
  
  ```
  source = "git::https://github.com/stalhaali/kube-agents.git//terraform/modules/drift-pubsub?ref=feat/drift-pubsub-terraform"
  ```
  
  **Rendered sink filter** (`terraform output` after apply):
  
  ```
  logName="projects/stalhaali-gkedemos/logs/cloudaudit.googleapis.com%2Factivity"
  resource.type="k8s_cluster"
  protoPayload.methodName=~"create|patch|update|delete"
  NOT (protoPayload.methodName=~"coordination\.v1\.leases" AND (protoPayload.authenticationInfo.principalEmail=~"^system:" OR protoPayload.authenticationInfo.principalEmail=~"\.iam\.gserviceaccount\.com$"))
  ```

  The `NOT (... AND (... OR ...))` group stays on one line deliberately: Cloud Logging treats a newline as an implicit AND, which would break the OR grouping if it wrapped.

  **Post-apply verification** — `Apply complete! Resources: 6 added, 0 changed, 0 destroyed.`

  ```
  $ gcloud pubsub topics get-iam-policy platform-agent-drift-audit
  - members:
    - serviceAccount:service-<project-number>@gcp-sa-logging.iam.gserviceaccount.com
    role: roles/pubsub.publisher

  $ gcloud pubsub subscriptions describe platform-agent-drift-audit-sub --format=yaml
  ackDeadlineSeconds: 60
  expirationPolicy: {}
  labels:
    goog-terraform-provisioned: 'true'
  messageRetentionDuration: 2678400s
  retryPolicy:
    maximumBackoff: 600s
    minimumBackoff: 10s
  state: ACTIVE
  topic: projects/stalhaali-gkedemos/topics/platform-agent-drift-audit
  ```

  `messageRetentionDuration: 2678400s` is the API accepting the 31-day maximum — distinct from the `604800s` this module previously defaulted to, so the value proves the mechanism rather than coinciding with a default. `expirationPolicy: {}`
  confirms `ttl = ""` applied; the plan hides that attribute behind "1 unchanged attribute hidden", so it needs reading back from the API.

  **End-to-end drift event.** Created and patched a ConfigMap as a human principal, then pulled from the subscription:

  ```
  $ kubectl create configmap drift-test --from-literal=color=blue -n default
  $ kubectl patch configmap drift-test -n default --type merge -p '{"data":{"color":"red"}}'

  $ gcloud pubsub subscriptions pull platform-agent-drift-audit-sub --limit 1000 --auto-ack --format=json \
      | jq -r '.[].message.data | @base64d | fromjson
         | "\(.protoPayload.authenticationInfo.principalEmail)  \(.protoPayload.methodName)  \(.protoPayload.resourceName)"'

  <redacted>@google.com  io.k8s.core.v1.configmaps.create  core/v1/namespaces/default/configmaps/drift-test
  <redacted>@google.com  io.k8s.core.v1.configmaps.patch   core/v1/namespaces/default/configmaps/drift-test
  ```

  **Across ~330 real messages pulled:**

  - Zero `leases.*` entries — the exclusion holds in the live Log Router, not just in `gcloud logging read`.
  - `cluster-autoscaler-status`, `gke-common-webhook-heartbeat`, and balloon-pod creates all present — confirming the sink did not over-filter and the tier denominators survive for the detector to measure.
  - `principalEmail` present on every entry, which is the field the whole design depends on.

  **Not covered:** the end-to-end pull above is from an earlier apply of the same module, before the retention and comment commits; those two changes touch a subscription property and code comments, neither of which affects message delivery, so
  it was not re-run. Every resource was torn down with `terraform destroy` after the final run; `gcloud logging sinks list` and `gcloud pubsub subscriptions list` confirm nothing remains in the project.

  `make docs-check` passes.


terraform/modules/drift-pubsub/main.tf
```diff
@@ -0,0 +1,155 @@
+# Cloud Logging -> Pub/Sub delivery path for the drift detector.
+#
+# GKE audit logs cannot be read from the Kubernetes API. The control plane is
+# managed, so the API server's audit backend is not ours to configure, and the
+# stream surfaces only in Cloud Logging. This module builds the route out of
+# Cloud Logging and into a subscription the detector pulls from.
+#
+# The design and the Phase 0 spike that produced the sink filter live in
+# agents/platform/docs/drift-detection/.
+
```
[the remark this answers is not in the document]
stalhaali: done

terraform/modules/drift-pubsub/variables.tf
```diff
@@ -0,0 +1,73 @@
...
+  default     = 60
+
+  validation {
+    condition     = var.ack_deadline_seconds >= 10 && var.ack_deadline_seconds <= 600
+    error_message = "ack_deadline_seconds must be between 10 and 600."
+  }
+}
+
+variable "message_retention_duration" {
+  description = "How long unacked messages are retained, as a duration string. The 7-day default is Pub/Sub's maximum, chosen so a detector outage over a weekend does not lose the drift events it was down for."
```
jayantid: "The 7-day default is Pub/Sub's maximum" isn't right — 7 days is the *default*; the max for `messageRetentionDuration` is **31 days** ([subscription properties](https://docs.cloud.google.com/pubsub/docs/subscription-properties)).

Worth fixing because of what the wording forecloses: the stated goal here is surviving a detector outage, and the next person reading this won't go looking for a higher value when the description says they're already at the ceiling. Detector down over a holiday shutdown, back on day nine, and the first eight days of drift are gone with nothing to say anything was dropped.

Either default to `"2678400s"`, or keep 7 days and say it's a volume/cost choice rather than a limit.

jayantid:
Terraform looks solid — resource references, the newline-AND filter assembly, `ttl = ""`, and the viewer-alongside-subscriber grant all check out against the provider and the `chat-pubsub` precedent. Lease carve-out scoped by principal is the right call.

One thing on the evidence in the description, since those numbers are the whole justification for the carve-out and I'd like to be able to cite them:

- The rendered sink filter has no `\.` escapes (`coordination.v1.leases`, `.iam.gserviceaccount.com$`) where `main.tf:47` has them. Behaviourally equivalent — unescaped `.` is a superset — but it means the plan output shown isn't the plan this branch produces.
- The daily volumes disagree with the code: the description says ~1.35M/day cut to ~60k/day, while `main.tf:31` and the README derive ~918k/day and ~42k/day from the same 9,558/442 sample.

Worth refreshing both.

terraform/modules/drift-pubsub/variables.tf
stalhaali: done

stalhaali:
> Terraform looks solid — resource references, the newline-AND filter assembly, `ttl = ""`, and the viewer-alongside-subscriber grant all check out against the provider and the `chat-pubsub` precedent. Lease carve-out scoped by principal is the right call.
> 
> One thing on the evidence in the description, since those numbers are the whole justification for the carve-out and I'd like to be able to cite them:
> 
> * The rendered sink filter has no `\.` escapes (`coordination.v1.leases`, `.iam.gserviceaccount.com$`) where `main.tf:47` has them. Behaviourally equivalent — unescaped `.` is a superset — but it means the plan output shown isn't the plan this branch produces.
> * The daily volumes disagree with the code: the description says ~1.35M/day cut to ~60k/day, while `main.tf:31` and the README derive ~918k/day and ~42k/day from the same 9,558/442 sample.
> 
> Worth refreshing both.

Replaced with real `terraform plan` output, with the module sourced by git ref (`?ref=feat/drift-pubsub-terraform`) so it can only reflect this branch. Escapes intact.

On the volumes, the code was the stale bit, not the description. `main.tf:31` scaled the 10,000-row sample out to a day, but 10,000 was just the `--limit` on my query, not what the window actually held. That's where ~960k and ~42k came from (10000×96 and 442×96). The 9,558/442 split isn't affected by the truncation so it stands, but the daily numbers don't. `5e1b7ff` redoes them off a real untruncated count: 623 non-lease calls in 15 minutes, so ~60k/day, and ~1.35M/day before the filter. The README had the sample counts but never a daily number, so I added it.

jayantid (approved):

PR #616, opened 2026-08-11
fix(security): authenticate the Session KV server and pseudonymise chat identities

AntonTyb:
# Pull Request

## Why This Change

The Session KV server on port 8699 bound `0.0.0.0` and accepted every request. It serves session
metadata and incident reports keyed by chat identity, and that identity was stored, logged, and
stamped onto OTel spans in plaintext — on Google Chat the "user id" *is* the sender's e-mail
address.

## What Changed

**Files:**

- `agents/platform/scripts/session_kv_server.py` — bearer-token auth, loopback bind, schema pages off, legacy-row migration
- `agents/chat/defaults/plugins/common/redactor.py` — new shared HMAC pseudonymiser (+ `__init__.py`)
- `agents/chat/defaults/plugins/{session_store,session_otel_bridge,tool_call_audit}/` — hash identities before SQLite, spans, and audit records
- `agents/chat/defaults/hooks/chat_message_audit/handler.py`
- `agents/platform/scripts/{session_manager.py,platform_mcp_server.py}`, `agents/platform/plugins/incident_context/__init__.py` — in-pod callers present the key
- `k8s-operator/api/v1alpha1/common_types.go`, `k8s-operator/internal/controller/platformagent_manifests.go` — two new optional `SecretKeySelector` CRD fields
- `charts/kube-agents/`, `k8s-operator/scripts/provision_07_gcp_k8s_secrets.sh`, `terraform/examples/full-install/` — key generation across all three install paths
- Docs: `INSTALL.md`, `docs/credential-isolation-design.md`, `docs/designs/gchat-session-metadata-data-flow.md`, `docs/site/src/content/docs/reference/{attribution,credential-isolation}.md`, `docs/site/src/content/docs/operator/platformagent-crd.md`
- New tests: `test_redactor.py`, `test_bridge.py`, `test_store.py`, `test_audit.py`, `test_handler.py`

**Added/Changed/Fixed:**

- **Authentication.** Every route except `/healthz` requires a bearer token from the new
  `SESSION_KV_API_KEY` Secret key, compared with `hmac.compare_digest`, accepted as either
  `Authorization: Bearer` or `X-Api-Key`. Deliberately *not* `API_SERVER_KEY`: that value is the
  loopback sentinel `cluster-internal-trusted`, a marker rather than a secret, so reusing it would
  authenticate nothing. An unset key fails closed with `503` and says why. The bind narrows to
  `127.0.0.1` — every caller shares the Pod's network namespace — and `/docs`, `/redoc`, and
  `/openapi.json` are switched off, since an interactive schema page on a port carrying chat
  identifiers is a browsable index of them.
- **Pseudonymisation.** A shared `redactor.py` HMAC-SHA256 hashes any address-shaped identity with
  `SESSION_KV_SALT` before it reaches SQLite, an audit record, or a span, and strips credentials and
  addresses out of logged text. Slack member ids contain no `@`, are already pseudonyms, and stay
  readable. `init_db()` strips plaintext identity keys out of pre-existing rows, rewriting rather
  than deleting them so `chat_id`/`thread_id` survive and threaded replies keep routing. Redaction
  never raises — its callers are `pre_gateway_dispatch` and `start_span`, so a regex false positive
  would otherwise take out message dispatch or every span the agent opens.
- **Hardened** the OTel bridge's attribute lookup with a `try`/`except` (and the `logger` that
  handler needs), so a failure to resolve session metadata degrades to the original attributes
  rather than propagating into every span. `bridge.py` on `main` has no error path here at all —
  this adds one; it does not fix an existing defect.
- **Fixed** a Makefile glob for hook tests — `agents/*/defaults/hooks/*/test_*.py` matched nothing,
  so the chat audit hook was untestable by CI however many tests it grew.

## Why This Matters

- Anyone able to reach the Pod's network namespace could read every stored chat identity and
  incident report without presenting a credential.
- On Google Chat those identities are e-mail addresses, which were being written to disk, logs, and
  a telemetry backend in plaintext.
- Rotating the salt re-anonymises every user, severing their past sessions from their future ones —
  so the chart preserves it across upgrades rather than regenerating.

## Context

Both keys are optional and degrade rather than wedge the pod: absent, the server refuses
authenticated requests and hashing falls back to a per-process random salt with one warning.

They are also the sandbox's only Secret-backed environment variables. The credential-isolation
design now documents this as an explicit loopback-only exception rather than an erosion — both are
pod-scoped, and the sandbox is the *server* for one and the point of hashing for the other. The
operator test enumerates the two allowed names, so a third cannot appear unnoticed.

`X-Hermes-User-Email` was removed rather than hashed: it existed so a downstream agent could show an
operator who asked, and a hash cannot do that. The pseudonymous `X-Hermes-Sender-Id` already
correlates turns from the same person.

## Testing

- `make docs-check` — passes.
- `make test-python` — passes for every directory whose third-party imports were installable in the
  authoring environment. `agents/platform/scripts/` could not be run locally (`fastapi`, `mcp`,
  `dotenv` unavailable, no `pip`); relying on CI for that directory, which is where
  `test_session_kv_server.py` and `test_session_manager.py` live.
- Operator build and `go test ./internal/controller/...` could not be run locally (no Go toolchain);
  relying on CI.
- `prettier --check` could not be run locally (no Node); relying on CI.
- New unit tests cover the redactor, the store, the OTel bridge, the tool-call audit plugin, and the
  chat-message audit hook.

---

Security fix with a migration step: existing session rows are rewritten in place on first startup
after the upgrade. Deployments that do not set the two new keys keep working, with the Session KV
server returning `503` and identity hashing unstable across restarts.

*This PR was generated with the help of Claude.*


mplakhtiy:
## Findings

### 1. BLOCKER — the Platform MCP server never receives the key it is supposed to present

**`agents/platform/config.yaml:10-37`** (the `mcp_servers.platform_control.env` map), against
**`agents/platform/scripts/platform_mcp_server.py:37`**.

`_session_kv_headers()` reads the token from the process environment:

```python
token = (os.environ.get("SESSION_KV_API_KEY") or "").strip()
if token:
    headers["Authorization"] = f"Bearer {token}"
```

But `platform_control` is a stdio MCP server, and its environment is an explicit allowlist. The
file's own comment states the contract:

> Hermes hands a stdio MCP server only a safe baseline (PATH, HOME, …) plus the keys named here
> (`_build_safe_env` in hermes `tools/mcp_tool.py`)

`API_SERVER_KEY` is listed at line 37 for exactly this reason. `SESSION_KV_API_KEY` is not listed
anywhere in `config.yaml`, and no operator overlay adds it — `grep -rn 'SESSION_KV_API_KEY'` over the
whole tree returns the operator manifests, the chart, the provisioning script, Terraform, and the
three Python readers, but never `config.yaml`. The operator injects the variable into the sandbox
*container*, which is not the same thing as injecting it into the MCP subprocess.

**Failure scenario.** `token` is `""`, no `Authorization` header is set, and both call sites 401:

- `send_notification(message, session_id)` — `GET /v1/sessions/{id}/metadata` raises `HTTPError`,
  caught by the bare `except Exception` at `platform_mcp_server.py:569`, printing `Failed to resolve
  session metadata for threading`. `chat_id` and `thread_id` stay `None`, so `targets` falls back to
  the home-channel broadcast: **every alert-driven incident report loses its chat thread.**
- Because `chat_id and thread_id` is then falsy, the `POST /v1/incidents` block at line 599 never
  runs at all. **No incident report is ever stored**, so `incident_context`'s two-way reply lookup
  returns nothing for the life of the deployment.

Both failures are silent — a `print` to stderr and a skipped branch. The operator test added in this
PR asserts the sandbox and the credential-proxy container agree on the Secret key, which catches the
event-watcher half of the same class of bug; nothing covers the MCP subprocess half.

**Fix.** Add `SESSION_KV_API_KEY: "${SESSION_KV_API_KEY}"` beside `API_SERVER_KEY` in
`agents/platform/config.yaml`, and add a test asserting `_session_kv_headers()` produces an
`Authorization` header so the regression cannot recur.

**Verdict: CONFIRMED.**

### 2. MEDIUM — legacy plaintext `user_id` is trusted while legacy `user_email` beside it is hashed

**`agents/chat/defaults/plugins/session_otel_bridge/bridge.py:102-106`** and
**`agents/platform/scripts/session_manager.py:93`**.

```python
sender_id = (
    metadata.get("user_id")
    or metadata.get("user_email_hash")
    or AuditRedactor.pseudonymise_identity(metadata.get("user_email"))
)
```

The expression goes to the trouble of hashing a legacy `user_email` in its third branch, but takes
`user_id` verbatim in its first. On Google Chat those two fields held the *same* address — the design
doc's own example at `docs/designs/gchat-session-metadata-data-flow.md:151` shows both. So the branch
that is guarded is the redundant one, and the branch that is unguarded is the one that actually
carries the address on a legacy row.

The only thing standing between that and a plaintext address on a span is
`_purge_plaintext_identities`, which runs in a different process on a different startup path. That
purge is best-effort by construction: `session_kv_server.py:112-114` does `except Exception:
continue` on a row whose metadata will not JSON-parse, and `session_kv_server.py:135-136` logs and
moves on when the `UPDATE` fails. Any row it skips reaches these two readers intact.

**Failure scenario.** A pre-upgrade Google Chat row whose metadata blob is unparseable (or which is
read before the KV server has completed `init_db()`) yields
`hermes.sender.id = "alice@example.com"` and `user.id = "google_chat:alice@example.com"` on the span,
and `X-Hermes-Sender-Id: alice@example.com` on the delegation headers — the exact plaintext-to-
telemetry leak the PR exists to close. `session_manager.py`'s own new test
`test_current_context_ignores_plaintext_email` covers the `user_email`-only row shape and explicitly
notes "the raw row is returned verbatim"; it does not cover the `user_id` shape.

**Fix.** Wrap both reads: `AuditRedactor.pseudonymise_identity(metadata.get("user_id"))`.
`pseudonymise_identity` is a pass-through for any value without `@`, so an already-hashed digest and
a Slack member id both survive unchanged. One call in each file closes the gap without depending on
the purge.

**Verdict: CONFIRMED.**

### 3. MEDIUM — the audit log loses every GCP service-account identity

**`agents/chat/defaults/plugins/common/redactor.py:88`**

```python
EMAIL_PATTERN = re.compile(r"[a-zA-Z0-9._%+\-]+@[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}")
```

`redact_text` is now applied to every tool argument and every tool result through
`tool_call_audit._serialize` (`audit.py:63` and `audit.py:83`). GCP service-account addresses match
this pattern:

```
$ AuditRedactor.redact_text('binding kube-agents-platform@my-proj.iam.gserviceaccount.com to roles/container.admin')
binding [REDACTED_EMAIL] to roles/container.admin
```

A service-account address is not personal data; in this repository it is the primary identifier an
operator greps the audit log for. The redactor's own test file states the standard this violates: "a
value that is not a credential survives unchanged — an over-eager redactor makes the audit log
useless, which is the failure mode that gets redaction switched off."

**Failure scenario.** The agent runs `gcloud projects add-iam-policy-binding` or
`kubectl annotate serviceaccount`. The tool-call audit record shows the role and the resource but
`[REDACTED_EMAIL]` where the principal should be, so the record cannot answer *which identity* was
granted the role — the one question an IAM audit record exists to answer.

**Fix.** Exempt the Google service-account domains, e.g. a negative lookahead for
`@[^\s]*\.gserviceaccount\.com` before the general pattern, and add a case to
`test_ordinary_text_is_untouched`.

**Verdict: CONFIRMED.**

### 4. LOW — a non-ASCII byte in the API-key header returns 500 instead of 401

**`agents/platform/scripts/session_kv_server.py:89`**

```python
if not presented or not hmac.compare_digest(presented, expected):
```

Starlette decodes header values as latin-1, so any byte in `0x80`–`0xFF` becomes a non-ASCII `str`.
`hmac.compare_digest` refuses that:

```
$ hmac.compare_digest('café', 'secret')
TypeError: comparing strings with non-ASCII characters is not supported
```

The `TypeError` is not an `HTTPException`, so it escapes the dependency and FastAPI returns 500 with
a traceback in the log.

**Failure scenario.** An unauthenticated caller inside the pod's network namespace sends
`X-Api-Key: <byte 0xC3>` and gets a 500 and a stack trace rather than the 401 the route is specified
to return. No data leaks, and the loopback bind narrows who can reach it, but it is a wrong status
code on the security-critical path and a noisy log line a fuzzer will find.

**Fix.** Compare bytes: `hmac.compare_digest(presented.encode("utf-8"), expected.encode("utf-8"))`.

**Verdict: CONFIRMED.**

### 5. LOW — `values.yaml` credits the wrong Terraform component with generating the keys

**`charts/kube-agents/values.yaml:103-104`**

> With create=false (the production path) they come from whatever created the Secret —
> `provision_07_gcp_k8s_secrets.sh` and the **kube-agents-iam Terraform module** both generate them.

`terraform/modules/kube-agents-iam` provisions the GSA, the Workload Identity binding, and project
IAM roles. It creates no Kubernetes Secret and contains no reference to either key — `grep -rn
'SESSION_KV' terraform/` hits only `terraform/examples/full-install/`, where the two
`random_password` resources actually live. `charts/kube-agents/README.md` gets this right ("the
Terraform example"); `values.yaml` does not.

AGENTS.md, Documentation Guidelines: *"Verify identifiers against source, not against other docs."*

**Failure scenario.** An operator composing `kube-agents-iam` into their own Terraform assumes the
keys are handled, installs with `create=false`, and ships a deployment where the Session KV server
503s every request and salts are per-pod random.

**Verdict: CONFIRMED.**

### 6. LOW — the PR description claims a `NameError` fix that has no counterpart on main

The body lists:

> **Fixed** a guaranteed `NameError` in the OTel bridge's error path (it logged through an undefined
> `logger`).

`git show upstream/main:agents/chat/defaults/plugins/session_otel_bridge/bridge.py | grep -n logger`
returns nothing — main's `bridge.py` contains no `logger` reference at all, and `plugin.py` in the
same package defines one correctly. What the diff actually does is *add* a new `try`/`except` around
`_span_attributes_for_session` and add the `logger` that new handler needs. That is a worthwhile
hardening change; it is not a fix for an existing defect.

**Failure scenario.** The claim becomes the release note. A future reader looking for the bug this
supposedly fixed finds nothing, and the changelog overstates what shipped.

**Verdict: CONFIRMED.**

### 7. LOW — the `sys.path` bootstrap is copy-pasted into four files with a different depth in each

**`session_store/store.py:15`**, **`session_otel_bridge/bridge.py:13`**,
**`tool_call_audit/audit.py:8`** (all `parents[1]`), and
**`hooks/chat_message_audit/handler.py:11`** (`parents[2] / "plugins"`).

Four near-identical five-line blocks whose one varying element — the parent depth — is the thing
easiest to get wrong, and whose failure mode is an `ImportError` at plugin-load time rather than a
test failure.

**Failure scenario.** The fifth consumer copies the block from whichever neighbour it was reading,
gets the depth wrong for its own directory, and the plugin silently fails to load.

The `common` package is the natural home for this: a small `common/_bootstrap.py` (or documenting the
depth rule once in `common/__init__.py`, which already carries the "why this is not a plugin" note)
would leave each call site with one line. Worth doing while the pattern is only four copies old.

**Verdict: CONFIRMED.**


---

AntonTyb:
Thanks — all seven reproduce. Six are fixed in 224259c; one I'd like to push back on.

### 1. BLOCKER — `SESSION_KV_API_KEY` never reaches the MCP subprocess — **fixed**

Confirmed: `grep -n SESSION_KV_API_KEY agents/platform/config.yaml` returned nothing, and the file's
own comment on `GOOGLE_CHAT_HOME_CHANNEL` documents exactly this failure shape for exactly this
reason. Added the entry beside `API_SERVER_KEY`, with a comment naming both silent consequences
(thread lost to the home-channel fallback, incident report never POSTed) so the next person to prune
the allowlist knows what it costs.

Three tests in `test_platform_mcp_server.py`: the token becomes a `Bearer` header, an unset token
sets none, and — the one that actually guards the regression — `config.yaml`'s `platform_control.env`
carries `SESSION_KV_API_KEY`. A unit test on `_session_kv_headers()` alone would have passed on the
broken code, since the bug was in the config, not the function.

### 2. MEDIUM — legacy plaintext `user_id` trusted — **fixed, two ways**

Right, and the asymmetry was exactly as you describe: the guarded branch was the redundant one.

- `bridge.py`: `AuditRedactor.pseudonymise_identity(metadata.get("user_id"))`, as suggested.
- `session_manager.py`: **dropped rather than hashed**. It has no access to `common/redactor.py`
  (different tree, different container), and dropping is what `_purge_plaintext_identities` does to
  the same field — the hash reappears on the user's next message either way. Hashing here would also
  need the salt, which the comment in `_purge_plaintext_identities` explicitly says does not live in
  that process.

Both keep a Slack member id: no `@`, no change. Tests added for the plaintext-`user_id` row shape in
both files, plus one asserting the opaque-id path still survives.

### 3. MEDIUM — service-account addresses redacted out of the audit log — **fixed**

Agreed, and the standard you quoted is the right one. Negative lookahead, slightly tighter than the
suggested form:

```python
r"[a-zA-Z0-9._%+\-]+@(?!(?:[a-zA-Z0-9\-]+\.)*gserviceaccount\.com\b)[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}"
```

`@[^\s]*\.gserviceaccount\.com` would also exempt `foo@notgserviceaccount.com`; requiring whole
labels does not. Verified `*.iam.`, `appspot.`, and `cloudservices.` all survive, that a real address
in the same string is still redacted, and that `…gserviceaccount.community` is not exempted. Two
cases added to `test_ordinary_text_is_untouched`.

### 4. LOW — non-ASCII header byte returns 500 — **fixed**

Comparing `.encode("utf-8")` on both sides, with a comment on the latin-1 decode so it does not get
"simplified" back. Test asserts `X-Api-Key: café` is a 401.

### 5. LOW — `values.yaml` credits the wrong Terraform component — **fixed**

Correct, and the README had it right, which is the tell. Now reads
`terraform/examples/full-install` — the composition that actually holds the two `random_password`
resources.

### 6. LOW — the `NameError` claim has no counterpart on `main` — **fixed in the PR body**

Confirmed: `main`'s `bridge.py` contains no `logger` at all. The diff adds the `try`/`except` *and*
the logger it needs. The bullet now says that.

One thing your finding doesn't cover: the same sentence is in the commit message of 15a8aa4, and if
this lands as a squash the commit body is what becomes the release note, not the PR body. Fixing
that means amending and force-pushing the branch. Happy to do it — say the word and I will;
otherwise the PR body is the accurate copy.

### 7. LOW — the `sys.path` bootstrap copy-pasted four times — **pushing back**

The duplication is real, but I don't think `common/_bootstrap.py` can remove it: importing the
bootstrap helper requires the `sys.path` entry the helper exists to add. The call site would end up
as `importlib.util.spec_from_file_location` against a hand-computed path — the same depth constant,
spelled less legibly, with a worse failure mode. Documenting the rule in `common/__init__.py`
doesn't shrink the blocks either; it moves prose that is already there (each copy after the first is
one line, `# See the matching note in plugins/session_store/store.py`).

I'd also read the varying depth as load-bearing rather than accidental: `handler.py` is at
`hooks/<name>/`, not `plugins/<name>/`, and its comment says so. A shared helper would have to
compute or be told the depth anyway.

What would genuinely catch the failure you describe is a test, since the failure mode is an
`ImportError` at plugin-load time. If you want one, a test that imports every module under
`plugins/*/` and `hooks/*/` from a clean `sys.path` would fail loudly on a bad depth, and would cover
the fifth consumer without requiring it to know the rule. Happy to add that instead — but I'd rather
not add indirection that cannot actually be imported.

---

`make docs-check` passes. Python tests pass for `plugins/common`, `session_otel_bridge`, and
`agents/platform/scripts` except `test_session_kv_server.py`, which still needs `fastapi` —
unavailable here, relying on CI for that one (the byte-comparison test is in it).

*This reply was generated with the help of Claude.*

deploy/shared/start-services.sh
```diff
@@ -95,7 +100,7 @@ start_event_watcher() {
         --dedup-persist="${dedup_persist}" \
         --in-cluster \
         --daemon-url=http://127.0.0.1:8699 \
-        --token-env=API_SERVER_KEY \
+        --token-env=SESSION_KV_API_KEY \
```
mplakhtiy, line 112: **BLOCKER — the documented upgrade path never delivers the keys, and the event watcher stops**

The watcher treats this variable as mandatory (`k8s-operator/cmd/k8s-event-watcher/main.go:642-644`):

    token = os.Getenv(f.tokenEnv)
    if token == "" {
        return fmt.Errorf("bearer token env var %s is empty", f.tokenEnv)
    }

`API_SERVER_KEY`, the name used before, is set on every existing deployment. `SESSION_KV_API_KEY` is set on none of them, and nothing in the upgrade path adds it: `upgrade.sh:337-355` runs `provision_03_gcp_gke_operator.sh` and `provision_08_deploy_platform_agent.sh` in every mode, and neither touches `platform-agent-secrets`. `provision_07_gcp_k8s_secrets.sh`, the script this PR correctly extends, runs only during a fresh `make gcp-provision`. The operator marks both variables `optional: true`, so a Secret without the keys yields containers without the variables rather than a failed mount.

**Failure scenario.** An operator runs the documented one-liner (`curl ... upgrade.sh | bash -s -- --upgrade-mode=full ...`) against a running fleet. The rollout succeeds and `upgrade.sh`'s own health verification passes, because both Deployments become Ready. Inside the pod:

- the credential-proxy container's watcher exits immediately on every attempt; `start-services.sh` retries with capped backoff and, after three short exits, prints `ALERT k8s-event-watcher has failed to start ... NO cluster events are being watched`. **No GKE warning is observed on any cluster in the fleet from that moment on.**
- the sandbox's Session KV server answers `503` to all six data routes, so `send_notification` loses thread resolution and `incident_context` loses two-way reply context — moot, since nothing reaches them any more.
- `SESSION_KV_SALT` is absent too, so pseudonyms are per-pod random.

Nothing outside the credential-proxy container log says so: the container stays Ready by design, the CR `.status` is unaffected, and no probe covers the watcher.

The PR body says "Deployments that do not set the two new keys keep working, with the Session KV server returning `503` and identity hashing unstable across restarts." `INSTALL.md:322-328`, `charts/kube-agents/values.yaml:118-128`, `charts/kube-agents/README.md:28-38` and `docs/credential-isolation-design.md:155-160` each enumerate the same two consequences. None of the five mentions that event watching stops outright, which is the one an operator would act on.

**Fix.** Either add the key generation/recovery block to a script the upgrade path runs, or add a release note and an `INSTALL.md` upgrade step telling operators to patch the existing Secret before upgrading — and correct the five "it degrades gracefully" passages to name the watcher. A `--token-env` fallback that keeps the watcher alive when the key is absent would defeat the point of the change, so it is not the answer here.

mplakhtiy (commented):
Review of `7b188673`. Findings only.

The three new commits close six of the seven findings from the previous round — `SESSION_KV_API_KEY` in `platform_control.env` with a test that reads `config.yaml` back, the legacy plaintext `user_id` hashed in `bridge.py` and dropped in `session_manager.py`, the `*.gserviceaccount.com` exemption, the byte comparison, the `values.yaml` credit, and the `NameError` claim out of the body. The `sys.path` bootstrap push-back is sound and is dropped.

**MEDIUM — no Live validation section, on the change most in need of one**

`AGENTS.md`, Pull Request Hygiene: *"Live-test the change before opening a PR, and describe it in the PR body. Every pull request fills in the template's Testing -> Live validation section with how the change was exercised against a real, running kube-agents installation ... Green unit tests and a clean `make docs-check` are necessary, not sufficient."*

`.github/PULL_REQUEST_TEMPLATE.md:33` carries the section; the body has no equivalent and no "Not live-tested" statement either. What it has instead is a list of what could not be run locally plus a note that CI covers it — precisely the substitution the rule rejects.

This is not a paperwork finding: the blocker below is invisible to every check that was run and visible within thirty seconds of upgrading a live install and reading the credential-proxy log. So are two other things nothing in CI covers — whether the Hermes plugin loader tolerates the new non-plugin `plugins/common/` directory, and whether `cp -ru` in `docker-entrypoint.sh` lands `plugins/common/redactor.py` onto a long-lived PVC before the plugins that import it load. An `ImportError` there takes out `session_store`, `session_otel_bridge` and `tool_call_audit` at once.

The rule postdates this PR being opened (it landed on `main` in #642/#643 and arrived here with the `9666c79a` merge), so it is a "before it merges" item rather than a "should have been" one. Exercising an *upgrade* of an existing deployment, rather than a fresh install, is what closes the blocker honestly.

agents/chat/defaults/plugins/common/redactor.py
```diff
@@ -0,0 +1,194 @@
...
+        ``tokenizer`` and ``author`` must not — hence whole-word matching
+        against :attr:`SENSITIVE_KEYS` rather than a substring test.
+        """
+        text = re.sub(r"([a-z0-9])([A-Z])", r"\1_\2", str(key)).lower()
+        words = set(re.split(r"[^a-z0-9]+", text))
+        words.add(text)
+        return {word for word in words if word}
+
+    @classmethod
+    def redact_text(cls, text: str) -> str:
```
mplakhtiy, line 188: **MEDIUM — a second credential redactor, weaker than the one already in the tree**

`_serialize` now runs `AuditRedactor.redact` over every tool argument and every tool result (`tool_call_audit/audit.py:38`), and the PR describes this module as stripping "credentials and addresses out of logged text". This repository already has a credential redactor built for exactly that job — `redact_secrets` in `agents/platform/skills/fleet-audit/scripts/audit_report.py:490-586` — and this is a narrower reimplementation of it. Measured against the same inputs, `redact_text` returns these unchanged:

| Input | `redact_secrets` | `AuditRedactor.redact_text` |
| --- | --- | --- |
| a `data:` / `stringData:` block with arbitrary key names | blanked by `_redact_secret_blocks` | verbatim |
| `xoxb-...` Slack bot token | redacted | verbatim |
| `eyJ....eyJ....sig` JWT (a projected SA token) | redacted | verbatim |
| `github_pat_...` | redacted | verbatim |
| `Authorization: Basic <b64>` | redacted | verbatim |
| PEM body with header/footer preserved | preserved shape | whole block replaced |

The `data:` row matters most, because the dictionary-key heuristic does not reach a string result and the key-name regex does not either — `SECRET_KV_PATTERN` requires `\b` before `api_key`, which `SESSION_KV_API_KEY` and `ANTHROPIC_API_KEY` do not offer.

**Failure scenario.** The agent runs `kubectl get secret platform-agent-secrets -n kubeagents-system -o yaml`. `log_post_tool_call` serialises the result, redaction leaves it untouched, and the whole `data:` block — both Slack tokens, the model-provider key, `API_SERVER_KEY`, and the `SESSION_KV_API_KEY` and `SESSION_KV_SALT` this PR introduces — is written to stdout in base64 and shipped to Cloud Logging by GKE's log agent. Not a regression (`_serialize` redacted nothing before), but the change is what makes the audit path claim coverage it does not have, and the implementation that closes the gap is already written and tested here.

**Fix.** Port `_redact_secret_blocks` and `_TOKEN_SHAPE_RE` into `AuditRedactor`, or — better, given both trees land in the same image under `/opt/defaults` — make one of the two the shared implementation and have the other import it.

agents/platform/scripts/session_kv_server.py
```diff
@@ -35,11 +36,115 @@
...
+# than dropped outright — a Slack member id is opaque and stays.
+_PLAINTEXT_IDENTITY_KEYS = ("user_email",)
+
+
+def _purge_plaintext_identities(conn: sqlite3.Connection) -> None:
+    """Strip plaintext identities left in rows written before this change.
+
+    Stripping rather than deleting: the row also carries `chat_id`/`thread_id`,
+    and dropping it would break threaded replies for conversations that are
+    still open. The hash is not recomputed here — the salt lives in the sandbox
```
mplakhtiy: **LOW — the purge declines to rehash on a false premise about which container it runs in**

`session_kv_server.py` runs in the sandbox container. `docker-entrypoint.sh:688` starts it, that script is the `platform` image's `ENTRYPOINT` (`/usr/local/bin/agent-entrypoint`), the sandbox container sets `Args` and never `Command` so the entrypoint runs (`platformagent_manifests.go:1986-1996`), and this PR's own `buildPodTemplateSpec` change puts `SESSION_KV_SALT` in that container's environment (`platformagent_manifests.go:1248-1251`, visible in all three golden manifests). The credential proxy is the container that does *not* get the salt, and it does not run this server.

The premise is genuinely true for one caller — the fallback server that `platform_mcp_server.start_session_kv_server()` spawns inherits the stdio MCP allowlist, which carries `SESSION_KV_API_KEY` but not `SESSION_KV_SALT` — but that is the secondary path, and the comment does not say so. The same reasoning is repeated at `session_manager.py:96-100`.

**Failure scenario.** A legacy Google Chat row is purged of its `user_id` without gaining a `user_email_hash`, so that session yields no `X-Hermes-Sender-Id` and no `user.id` span attribute until the user's next message — a small, self-healing loss the purge could have avoided. The durable cost is the comment: it is the kind of container-topology claim the next reader will take on trust while reasoning about where a secret is allowed to live.

agents/chat/defaults/plugins/common/redactor.py
```diff
@@ -0,0 +1,194 @@
...
+        r"(?i)\b(password|passwd|secret|token|api_key|apikey|access_token|client_secret)\b"
+        r"([\"']?\s*[:=]\s*)([\"']?)([^\"'\s,}{\]]+)\3"
+    )
+    # The negative lookahead exempts GCP service-account addresses. They are not
+    # personal data, and in this repository the principal is the one thing an
+    # operator greps an IAM audit record for — redacting it leaves a record that
+    # says which role was granted on which resource but not to whom, which is
+    # the over-eager-redactor failure mode that gets redaction switched off.
+    EMAIL_PATTERN = re.compile(
+        r"[a-zA-Z0-9._%+\-]+@(?!(?:[a-zA-Z0-9\-]+\.)*gserviceaccount\.com\b)"
```
mplakhtiy: **LOW — the service-account exemption is unanchored on its right edge**

The lookahead ends at `\b`, not at the end of the host, so it fires on any domain that merely *contains* `gserviceaccount.com` as a label sequence:

    >>> AuditRedactor.redact_text('see victim@corp.gserviceaccount.com.attacker.io end')
    'see victim@corp.gserviceaccount.com.attacker.io end'

The left edge is correctly anchored — `a@notgserviceaccount.com` is redacted, and `b@foo.gserviceaccount.company.com` is too — so the tightening chosen over the suggested `@[^\s]*\.gserviceaccount\.com` did its job in one direction and not the other.

**Failure scenario.** An address under an attacker-registered domain of that shape reaches the audit log unredacted. The trigger is contrived — it needs a real domain ending in something after `gserviceaccount.com` — so this is an anchoring defect rather than a live exposure. `test_ordinary_text_is_untouched` already has the positive cases and needs the negative one.

```suggestion
        r"[a-zA-Z0-9._%+\-]+@(?!(?:[a-zA-Z0-9\-]+\.)*gserviceaccount\.com(?![\w\-.]))"
```

deploy/shared/start-services.sh
mplakhtiy, line 112: **LOW — the shell half of the token wiring has no test, and it is the same class as round 1's blocker**

The previous round's blocker was a variable name that existed in Go and Python but not in the allowlist that feeds the process. That was closed well: `test_config_yaml_passes_the_key_into_this_subprocess` reads `config.yaml` back and asserts the entry. The symmetric wiring — this line naming `SESSION_KV_API_KEY` to `--token-env`, and `buildCredentialProxyEnv` injecting a variable of that name — has no such test. The Go test (`platformagent_manifests_test.go:691-706`) proves the two containers resolve the same `SecretKeySelector`, but nothing reads this shell script.

**Failure scenario.** A future rename of the operator-side variable passes `go test` (the Go test hardcodes the new name along with the code) while `start-services.sh` keeps the old one. The watcher then exits on every start with `bearer token env var ... is empty`, and the only symptom is a log line in a container that stays Ready.

A four-line test that greps `start-services.sh` for the `--token-env` value and compares it against the constant in `platformagent_manifests.go` — or the import-every-module test offered in your earlier reply, extended to this — would catch it.

AntonTyb: **Fixed in a7e4750b.**

Reproduced by reading the path: `upgrade.sh` runs `provision_03` and `provision_08` in every mode, `provision_07` runs only on a fresh install, and both refs are `optional: true`, so the containers come up without the variables rather than failing to mount. The watcher then exits on every start.

Took the first of the two options you offered, since a release note only helps the operator who reads it:

- `upgrade.sh` grows `backfill_session_kv_keys`, called as a new step 3 between the `get-credentials` and the provisioning steps. It patches `platform-agent-secrets` with `stringData` for whichever of the two keys is absent and leaves an existing value alone — rotating `SESSION_KV_SALT` re-anonymises every user, so preservation is the point rather than a nicety, same as `provision_07`. A Secret that does not exist (Helm with `credentials.create`, Terraform, an external secret store) is a warning and a skip, not a failure.
- `--upgrade-mode=operator` rolls only the controller manager, so a Secret patched there would sit unread until some later harness upgrade with the watcher dead in between. When the backfill patched something in that mode the script now also restarts `platform-agent-gateway` and waits for the rollout. The other two modes re-render the Deployment and pick the keys up on their own.
- The dry-run preview names the backfill, so `--dry-run` still describes everything the real run would do.

Not a `--token-env` fallback, agreed: that would authenticate nothing, which is what the change exists to stop.

Second half of your fix, since the docs were wrong independently of the script. All five passages now name the watcher — `INSTALL.md:322`, `charts/kube-agents/values.yaml`, `charts/kube-agents/README.md`, `charts/kube-agents/templates/NOTES.txt`, `docs/credential-isolation-design.md`, plus the two site pages (`reference/credential-isolation.md` and the CRD reference, where it goes in prose under the table rather than in the cell, so prettier does not have to re-pad every row). `INSTALL.md` also carries the manual `kubectl patch` + `rollout restart` for anyone not using `upgrade.sh`.

One thing your finding did not ask for, but is the same defect seen from inside the pod: `start-services.sh` now checks the variable before entering the retry loop and prints the ALERT once, up front, naming the consequence and the remediation. The watcher's own error names a variable, not what stopped working, and the existing ALERT only fires after three short exits.

agents/chat/defaults/plugins/common/redactor.py
AntonTyb: **Fixed in 47eedb17.**

Confirmed the `data:` row first, which is the one that matters: `_serialize` hands `redact_text` a string, the dictionary-key heuristic never sees the keys, and `\bapi_key` matches neither `SESSION_KV_API_KEY` nor `ANTHROPIC_API_KEY` because `_` is a word character. So `kubectl get secret platform-agent-secrets -o yaml` went through untouched.

Ported into `AuditRedactor` rather than shared:

- `_redact_secret_blocks` as `SECRET_BLOCK_PATTERN` / `INDENTED_PAIR_PATTERN` plus the same line scan, with an early return when neither `data:` nor `stringData:` appears so the common case stays one substring test.
- The three missing token shapes — `xox[baprs]-`, `github_pat_`, and the JWT triple. The last is the one most likely to land in a tool result here, since that is what a projected ServiceAccount token looks like.
- `Basic` alongside `Bearer`, with `+/` added to the value class and the scheme preserved through the substitution rather than hardcoded to `Bearer`.
- The key-name prefix: `\b([\w.\-]*?(?:password|…|api[_-]?key|…))\b`, so `SESSION_KV_API_KEY=…` matches and the full key survives into the record. The trailing `\b` still does the work that matters — `TOKENIZER_PATH` does not match.

Two deliberate differences from `redact_secrets`:

- **PEM.** Left as a whole-block `[REDACTED_PRIVATE_KEY]`. The replacement already says what was there, and the shape-preserving form is a nicety in evidence prose rather than in a tool-call audit line. Happy to switch if you would rather the two agree exactly.
- **Shared implementation.** Went with the port over the import. `audit_report.py` is a skill script loaded from a different root, so a plugin reaching it needs a hardcoded `/opt/defaults` path — the class of thing the round-1 `sys.path` discussion already argued against, and the failure mode is an `ImportError` at plugin-load time. If you would rather have one copy, the direction I would take is moving the shared regexes into `plugins/common/` and having the skill import *up*, but that is a bigger change than this PR should carry.

One cost worth naming: a ConfigMap's `data:` is blanked too. The line scan cannot tell the two apart in a fragment, `redact_secrets` has always had the same behaviour, and it is the safe direction to err in — noted in a comment so the next reader does not think it accidental.

Tests: a Secret manifest whose values disappear while the keys and the surrounding document stay, the three token shapes, `Basic`, and both prefixed key names.

agents/platform/scripts/session_kv_server.py
AntonTyb: **Fixed in 47eedb17** — the comment, not the behaviour.

You are right on the topology: `docker-entrypoint.sh:688` starts this server, the sandbox container runs the entrypoint, and this PR puts `SESSION_KV_SALT` in that container. The comment claimed the opposite and would have been read as a statement about where the salt is allowed to live.

Both comments now give the real reason, which is the secondary path you identified: the fallback instance `start_session_kv_server()` spawns inherits the stdio MCP allowlist, which carries `SESSION_KV_API_KEY` and not the salt. Kept the drop rather than rehashing on the primary path, because the write is durable and the read is not: a digest computed on the fallback path under some other salt would sit in the row permanently uncorrelated with every hash the Chat Agent plugins produce, which is worse than the absent value — the plugins rewrite the hash on the user's next turn, so dropping costs one message and no more.

`session_manager.py:93` carried the same claim and now says the same thing in its own terms: its main caller is the stdio MCP server, whose environment is the `config.yaml` allowlist, so a hash computed there would not match the one the plugins produce for the same person.

agents/chat/defaults/plugins/common/redactor.py
AntonTyb: **Fixed in 47eedb17**, with one character different from your suggestion.

Reproduced: the left edge was anchored and the right edge was not, so `victim@corp.gserviceaccount.com.attacker.io` survived.

Took `(?!\.?[\w\-])` rather than `(?![\w\-.])`. Both redact the attacker domain, but `(?![\w\-.])` also loses the exemption for an address that ends a sentence — `granted to gcp-sa@my-proj.iam.gserviceaccount.com.` has a `.` in the lookahead position and would be redacted, which is the over-eager failure your MEDIUM #3 was about. Allowing an optional dot only when no word character follows it keeps both cases right:

- `…gserviceaccount.com.attacker.io` → redacted
- `…gserviceaccount.com.` at end of sentence → exempt
- `a@notgserviceaccount.com`, `b@foo.gserviceaccount.company.com` → still redacted

Three negative cases added in `test_the_service_account_exemption_is_anchored_at_both_edges`, and the sentence-ending address added to `test_ordinary_text_is_untouched`.

deploy/shared/start-services.sh
AntonTyb: **Fixed in 47eedb17.**

Agreed on the class — the Go test hardcodes the name alongside the code, so a rename passes it while the script keeps the old value.

`TestEventWatcherTokenEnvMatchesStartServices` in `platformagent_manifests_test.go` reads `deploy/shared/start-services.sh`, pulls the name out of `--token-env=(…)`, and asserts the credential-proxy container carries a variable of *that* name with a value or a `ValueFrom`. Deriving the expected name from the script rather than hardcoding it is what makes it bite: both sides have to move together, and a `--token-env` flag that disappears entirely fails too.

Checked it fails for the right reason before trusting it — renaming the flag value in the script gives:

```
--- FAIL: TestEventWatcherTokenEnvMatchesStartServices
    ../../../deploy/shared/start-services.sh passes --token-env=BOGUS_KEY, but the
    credential proxy container has no such variable; the watcher will exit on every start
```

Went with a Go test over the import-every-module one I offered earlier: this failure is a cross-language wiring mismatch, and the test that catches it should live next to the constant whoever renames it is editing. The import-every-module test is still worth having for the `sys.path` depth question — separate concern, and I would rather not fold it into this thread.

AntonTyb:
## Round 3 — all five findings addressed, live validation still open

Pushed as [`a7e4750b`](https://github.com/gke-labs/kube-agents/commit/a7e4750b) and [`47eedb17`](https://github.com/gke-labs/kube-agents/commit/47eedb17); each finding has a reply in its own thread with the reasoning and the two places I deviated from the suggested fix.

| Finding | Status |
| --- | --- |
| BLOCKER — upgrade path never delivers the keys, watcher stops | Fixed: `upgrade.sh` backfills both keys before the rollout (additive, never rewrites an existing salt), operator-mode also restarts the agent, `start-services.sh` alerts up front, and all five "it degrades gracefully" passages now name the watcher |
| MEDIUM — second, weaker credential redactor | Fixed: `data:`/`stringData:` block scan, Slack / `github_pat_` / JWT shapes, `Basic`, and prefixed key names ported into `AuditRedactor` |
| LOW — purge comment's container claim | Fixed: the comment was wrong; corrected in both files, behaviour unchanged and the real reason (the MCP-spawned fallback server's allowlist) now stated |
| LOW — service-account exemption unanchored right | Fixed, with `(?!\.?[\w\-])` instead of `(?![\w\-.])` so an address ending a sentence keeps the exemption |
| LOW — no test for the shell half of the token wiring | Fixed: a Go test that derives the expected name from `start-services.sh` and fails when the two diverge |

### MEDIUM — no Live validation section

**Not closed, and I am not going to paper over it.** This has not been exercised against a running installation, and the section stays empty until it has been. You are right that it is the finding that matters most here: the blocker was invisible to `make docs-check`, `go test`, and the Python suite, and thirty seconds of credential-proxy log after an upgrade would have shown it.

What still needs a live *upgrade* (not a fresh install) to confirm, in priority order:

1. `platform-agent-secrets` gains both keys, the agent restarts, and the credential-proxy log shows the watcher running rather than the new ALERT — then the same upgrade re-run leaves the existing salt byte-for-byte unchanged.
2. The Hermes plugin loader walks past the non-plugin `plugins/common/` directory.
3. `cp -ru` in `docker-entrypoint.sh` lands `plugins/common/redactor.py` on a long-lived PVC before `session_store`, `session_otel_bridge` and `tool_call_audit` import it.

Local checks that did run: `make docs-check` passes; `go build ./...` and `go test ./internal/controller/...` pass; the Python suite passes except `test_session_kv_server.py`, which still needs `fastapi` (unavailable here, and only its module docstring changed in this round). `prettier` is not installed locally — the CRD reference change went into prose under the table rather than into a cell, so no table needs re-padding.

*This reply was generated with the help of Claude.*

mplakhtiy (approved):

PR #617, opened 2026-08-11
feat(tooling): share the pr-review-batch PR-review command

mplakhtiy:
# Pull Request

## Why This Change

The batched PR-review command that several of us have been running against this
repo's pull requests lived only in a local, untracked `.claude/` directory, and
it was written for exactly one person's checkout. It hardcoded remote names —
`origin` as the fork, `upstream` as `gke-labs` — and absolute `/Users/...`
paths, so nobody else could run it without editing it first. It also refused to
review any PR that conflicted with `main`, which is a large fraction of the
queue at any given moment and precisely the set most likely to need a look.

This checks the command in and makes it portable.

## What Changed

**Files:**

- `.claude/commands/pr-review-batch.md`
- `.claude/.gitignore`

**Added/Changed/Fixed:**

- **Remote-naming agnostic.** `gke-labs/kube-agents` stays fixed — PRs always
  live there — but the local remote pointing at it is now discovered by URL
  match across `git remote`, so it works whether you call it `upstream`,
  `origin`, or nothing at all. When no remote matches, it fetches from
  `git@github.com:gke-labs/kube-agents.git` directly rather than guessing a
  name. GitHub access goes through `gh --repo`, git access through SSH.
- **Base branch is read from the PR** (`gh pr view --json baseRefName`) instead
  of assuming `main`, so stacked PRs diff against the right thing.
- **No absolute paths.** `MAIN_ROOT` comes from
  `git rev-parse --path-format=absolute --git-common-dir`, which resolves to the
  primary checkout even when the reviewing agent is standing inside a worktree.
  Saved reviews therefore land in one shared place and the "already reviewed
  this head SHA" skip check keeps working across runs.
- **Conflicting PRs are reviewed rather than skipped.** The command attempts the
  sync, and on conflict records the conflicting paths, runs `git merge --abort`,
  and reviews `git diff "$BASE_REF"...pr<N>` — the PR's own diff, as GitHub
  renders it. It is told explicitly never to resolve a conflict: the author owns
  the rebase, and a resolution invented by a reviewer would make every finding
  downstream of it fiction. The output has to say that the review covers the PR
  as authored, not as merged, so interaction bugs with newer base commits are
  never silently claimed as checked. `status: conflict` is gone from the report
  contract; conflicts are now a `conflicts:` field on an ordinary `reviewed`
  result.
- **Review process folded in from Claude Code's built-in `/code-review`.** The
  find phase is now eight named angles — line-by-line diff scan,
  removed-behavior auditor, cross-file tracer, operations and security, reuse,
  simplification and efficiency, altitude, and conventions/docs — and the verify
  phase adopts its CONFIRMED / PLAUSIBLE / REFUTED ladder, including the
  "PLAUSIBLE by default" rule that stops a reviewer from refuting a real race or
  a rare-but-reachable nil path for being "speculative". The
  operations-and-security angle is adapted to this repo: IAM and RBAC scope,
  credential handling and redaction, NetworkPolicy reach, and pinned Action
  SHAs.
- **The review is anchored on the PR's stated intent.** A new Phase 2b reads
  the description and any linked issue and records, in one sentence, what the
  PR claims to do; a ninth angle then holds the diff against that sentence,
  covering scope creep — unrelated refactors and drive-by reformatting riding
  along — and test coverage, asking for each claimed behaviour which test would
  fail if it regressed. The saved review carries the intent sentence in its
  header so the next reader knows what the findings were measured against.
  Guardrails run both ways: a "known limitation, follow-up PR" in the body does
  not make a dropped guard correct, and where the description and the diff
  disagree, the diff is what merges.
- **`.claude/.gitignore`** keeps the per-developer scratch the command produces —
  review worktrees, saved review files, `settings.local.json` — out of the
  repository, so checking in `commands/` doesn't drag sixteen worktrees along
  with it.

## Why This Matters

- The review workflow becomes a team asset instead of one laptop's shell
  history, and improvements to it are reviewable like any other change.
- Conflicting PRs are the ones most likely to have been sitting a while. Being
  able to review them without touching the conflict means feedback arrives
  before the rebase, not after.
- Anchoring the find and verify phases to the built-in `/code-review` structure
  means findings arrive with a stated verdict and a concrete failure scenario
  rather than a list of impressions.

## Context

`.claude/` is a root-level dot-directory, which `docs/README.md` §1 and
`scripts/check_docs_map.py` both place explicitly outside the documentation
map's scope, so no map entry is needed; the check's exempt-file count moves from
22 to 23 and the stated document total is unaffected.

## Testing

- `make docs-check` — passes at HEAD. Generated regions current, all relative
  links resolve across 178 Markdown files, terminology check passes, map
  inventory covers all 154 tracked docs with the stated total of 155 matching.
- `npx prettier --check .claude/commands/pr-review-batch.md` — passes.
- `git check-ignore -v` confirms `.claude/worktrees/`, `.claude/pr-reviews/`,
  and `.claude/settings.local.json` are ignored while
  `.claude/commands/pr-review-batch.md` stays tracked.

---

Documentation and tooling only — no runtime, chart, or operator code is touched,
so there is nothing to roll out. The risk is confined to the review command
itself, and it is visible on the first run: if the remote discovery or the
conflict path misbehaves, the subagent's structured report says so rather than
producing a wrong review.


mplakhtiy:
@bradhoekstra WDYT about this skill?

bradhoekstra (commented):
### Code review

Found 2 issues:

1. `.claude/.gitignore` is inert: the root `.gitignore:43` still excludes `.claude/` as a whole, so git never descends into it. The nested file's rules are never consulted, and any *new* file under `.claude/commands/` is invisible to `git status` and needs `git add -f`.
2. `.claude/commands/pr-review-batch.md:68-69` — `2>/dev/null || true` on `git worktree add` followed by an unchecked `cd` lets a subagent silently continue in the shared checkout and run `git checkout -B` / `git merge` there.

.claude/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Shared: commands/ is checked in so the team gets the same slash commands.
+# Everything below is per-developer scratch produced by running them.
+worktrees/
+pr-reviews/
+settings.local.json
```
bradhoekstra: The root `.gitignore` still excludes the whole directory (line 43, `.claude/`), and git does not descend into an excluded directory — "It is not possible to re-include a file if a parent directory of that file is excluded." So this nested `.gitignore` is never consulted, and it is itself ignored.

Reproduced on a scratch repo with the same two files:

```
$ git check-ignore -v .claude/commands/pr-review-batch.md
.gitignore:2:.claude/   .claude/commands/pr-review-batch.md
$ git add .claude/
The following paths are ignored by one of your .gitignore files:
.claude
hint: Use -f if you really want to add them.
$ echo new > .claude/commands/another.md && git status --porcelain -uall
            # nothing - the new command is invisible
```

The two files in this PR are tracked (they were force-added), so nothing breaks today. What breaks is the next command: a teammate adding `.claude/commands/foo.md` will not see it in `git status` and will not be able to `git add` it without `-f`. The PR body cites `git check-ignore -v` as evidence the split works, but as shown above `check-ignore` reports the *command* file as ignored too — by the root rule, not by this file.

Fix in the root `.gitignore` (outside this diff), using `.claude/*` so git traverses the directory. That also corrects the comment there, which now says the opposite of what this PR does:

```diff
-# Per-developer Claude Code state (session settings, local skills). The skills
-# this repo ships live in .agents/skills/ and are committed.
-.claude/
+# Per-developer Claude Code state (session settings, local skills). The shared
+# slash commands in .claude/commands/ and the review skills in .agents/skills/
+# are committed. `.claude/*` rather than `.claude/` so git descends into the
+# directory and .claude/.gitignore is honoured.
+.claude/*
+!.claude/commands/
```

.claude/commands/pr-review-batch.md
```diff
@@ -0,0 +1,346 @@
...
+`$MAIN_ROOT` is the top of the primary checkout even when you are standing inside a worktree; it is
+where saved reviews live so that every run — and every teammate — sees the same history.
+
+### Phase 1 — Worktree
+
+Create your own worktree so you never contend with the other agents:
+
+```bash
+git -C "$MAIN_ROOT" worktree add "$MAIN_ROOT/.claude/worktrees/pr-<N>" -b pr-<N>-review 2>/dev/null || true
+cd "$MAIN_ROOT/.claude/worktrees/pr-<N>"
```
bradhoekstra: `2>/dev/null || true` swallows the failure and the `cd` is unchecked, so a failed `worktree add` leaves the subagent running every later phase in the shared checkout instead of aborting.

The trigger is routine, because nothing in this command ever removes the worktree or the branch: run the command once for PR 500, then `rm -rf .claude/worktrees` to clean up the "sixteen worktrees" the PR body mentions. Branch `pr-500-review` survives. On the next run `worktree add -b pr-500-review` fails with `fatal: a branch named 'pr-500-review' already exists`, the message is discarded, `|| true` keeps the script going, `cd` fails into a bare error, and Phase 1's `git checkout -B pr-<N>-review "pr<N>"` plus Phase 1b's `git merge` then execute against the developer's primary checkout — with several subagents doing it concurrently, since they all fell back to the same directory. The section's own next paragraph ("Never use `git -C` pointing at the shared checkout after this point") is the invariant this silently breaks.

`worktree prune` clears the stale registration, `-B` reuses the leftover branch instead of failing on it, and the explicit `|| exit 1` turns a lost worktree into a reported `status: skipped` rather than a write to the wrong tree.

```suggestion
git -C "$MAIN_ROOT" worktree prune
git -C "$MAIN_ROOT" worktree add --force -B pr-<N>-review "$MAIN_ROOT/.claude/worktrees/pr-<N>" || exit 1
cd "$MAIN_ROOT/.claude/worktrees/pr-<N>" || exit 1
```

.claude/commands/pr-review-batch.md
```diff
@@ -0,0 +1,346 @@
...
+write down, in one sentence, **what this PR claims to do**. Keep that sentence; it is the yardstick
+for Angle I and it is what tells you whether a given change belongs here at all.
+
+Two failure modes to avoid. Do not let the description talk you out of a defect — "known
+limitation, follow-up PR" in the body does not make a dropped guard correct, though it does change
+how you phrase the finding. And do not treat the description as a description of the diff: where
+the two disagree, the diff is what merges. A body that promises a behaviour the diff does not
+implement, or omits a behaviour the diff does, is itself a finding.
+
+### Phase 3 — Find candidates (nine angles)
```
bradhoekstra: The angle proposed below would make this ten. Flagging the count here so the heading and the list stay in sync.

```suggestion
### Phase 3 — Find candidates (ten angles)
```

bradhoekstra (commented):
### Review command — design suggestions

Three additive suggestions, from running an equivalent multi-PR review by hand across six PRs today (#548, #593, #595, #601, #617, #623).

The structure here is a better review engine than what I ran ad hoc — nine angles against my five, hostile CONFIRMED/PLAUSIBLE/REFUTED verification against a 0-100 confidence cutoff that conflates "is it real" with "does it matter", and saved reviews with a head-SHA skip check that makes re-running affordable. The notes below are about coverage and delivery, not the core design.

1. Phase 3 has no cross-PR awareness, which is a structural miss against this repo's pattern of one author opening several variations of a change.
2. Posting is correctly gated on an explicit ask, but has no mechanics behind it, so the GitHub round-trip gets improvised each time.
3. A heading count that would need to follow (1).

Separate from these: the worktree fallback at lines 68-69, raised as an inline finding on my earlier review, is the one I would treat as blocking — it is what currently makes the isolation this command depends on unsafe.

.claude/commands/pr-review-batch.md
```diff
@@ -0,0 +1,346 @@
...
+home per fact, generated `<!-- BEGIN GENERATED -->` regions regenerated rather than hand-edited,
+identifiers verified against source rather than against other docs.
+
+**Angle I — scope and test coverage.** Hold the diff against the intent sentence from Phase 2b.
+Flag changes that do not serve it: an unrelated refactor riding along, a dependency bump nobody
+asked for, a behaviour change buried in a PR described as a rename, reformatting that inflates the
+diff and hides the real hunks. Repo convention is scoped changes and no unrelated formatting, so
+cite the rule when it applies. Judge by whether a change serves the stated intent, not by how large
+it is — a big diff that does one thing is in scope, and a three-line change that does a second
+thing is not.
```
bradhoekstra, line 248: Phase 2 reads this PR's conversation, but nothing in the command looks across PRs — so an author who opens several variations of one change gets the same defect found independently each time, and divergence between the copies is never noticed.

This is not hypothetical in this repo. In a six-PR batch run today, the two highest-value findings came only from reading siblings: #548 turned out to be a superset of #620/#621/#623, with #620 already carrying accepted fixes that #548 lacks, and the fabricated-evidence defect in #595/#593 is a family-wide pattern across #591-#594 rather than four coincidences.

```suggestion
thing is not.

**Angle J — sibling pull requests.** List the author's other open PRs touching adjacent paths
(`gh pr list --repo "$REPO" --author <login> --state open --json number,title,files`), and read the
review comments on them. A finding accepted on a sibling usually applies here unchanged. A
near-identical PR that has already diverged is itself a finding — name which copy carries the fix
and which does not, since merge order then decides whether it survives.
```

.claude/commands/pr-review-batch.md
```diff
@@ -0,0 +1,346 @@
...
+
+## After the subagents return
+
+Print one compact table across all PRs — number, title, verdict, block reason, blocker count,
+review file path — then the blocker one-liners grouped by PR. Call out explicitly any PR that was
+skipped and why, and any that was reviewed against a conflicting base (those reviews cover the PR
+as authored, not as merged).
+
+Do not post to GitHub unless I ask. When I do, post findings only — no verdict, no CI summary, no
+closing section — and tell me whether you posted it as an issue comment or a formal review.
```
bradhoekstra, line 373: Keeping posting behind an explicit ask is the right default. The gap is that when the ask comes, there are no mechanics here — so the GitHub round-trip gets improvised each time, and the usual result is one summary comment rather than findings anchored where the author is reading.

Worth specifying:

````
cat > /tmp/review-<N>.json <<'JSON'
{"event":"COMMENT","body":"<summary>","comments":[
  {"path":"<path>","line":<n>,"side":"RIGHT","body":"<finding>\n\n```suggestion\n<replacement>\n```"}
]}
JSON
python3 -m json.tool /tmp/review-<N>.json   # a malformed payload 422s; nothing partial is posted
gh api "repos/$REPO/pulls/<N>/reviews" --input /tmp/review-<N>.json
````

Rules worth stating alongside it: `line` must be a RIGHT-side line present in the diff, with `start_line` + `line` for a range; a `suggestion` block replaces exactly the commented range, so the replacement must be the complete new text for those lines with correct indentation; findings that anchor to no changed line belong in the summary body under a **Findings outside this diff** heading rather than being forced onto an unrelated line; and `event=COMMENT` only — never APPROVE or REQUEST_CHANGES on the human's behalf.

One caveat from running this today: a `suggestion` block cannot itself contain a fenced code block, so findings whose fix includes one need prose instead.

File: .claude/.gitignore (hunk 1 above)
Comment (mplakhtiy) [reply .claude/.gitignore]: Confirmed and fixed in 5f2d4a3b — you are right, and my earlier reasoning against the bot's version of this finding was wrong because I checked a stale local `main` that predated the rule.

Root rule is now `.claude/*` with `!.claude/commands/`, and the comment is rewritten to say what the directory actually does. I dropped the nested `.claude/.gitignore` rather than keeping it: once the root rule stops excluding the directory, `.claude/*` already covers `worktrees/`, `pr-reviews/`, and `settings.local.json`, so the nested file was redundant — and it matched `.claude/*` itself, which is a confusing thing to leave lying around.

Verified rather than assumed this time:

```
$ echo test > .claude/commands/_probe.md && git status --porcelain -uall
?? .claude/commands/_probe.md
$ git check-ignore -v .claude/worktrees/x .claude/pr-reviews/y.md .claude/settings.local.json
.gitignore:46:.claude/*  .claude/worktrees/x
.gitignore:46:.claude/*  .claude/pr-reviews/y.md
.gitignore:46:.claude/*  .claude/settings.local.json
```

The next command file is visible to `git add`; the scratch stays ignored.

File: .claude/commands/pr-review-batch.md (hunk 2 above)
Comment (mplakhtiy) [reply .claude/commands/pr-review-batch.md]: Fixed in 5f2d4a3b. The trigger you describe is the normal case rather than an edge case, since nothing in the command ever removes a worktree or its branch — I had sixteen of them lying around while writing this.

Took your suggestion with one addition, a `[ -d "$WT" ]` guard: `worktree prune` only clears registrations whose directory is *gone*, so an intact leftover worktree still fails on the occupied path.

```bash
WT="$MAIN_ROOT/.claude/worktrees/pr-<N>"
git -C "$MAIN_ROOT" worktree prune
[ -d "$WT" ] || git -C "$MAIN_ROOT" worktree add --force -B pr-<N>-review "$WT" || exit 1
cd "$WT" || exit 1
```

Also stated the consequence explicitly, since the failure mode is a redirect rather than a degradation: if the worktree cannot be entered, report `status: skipped` and stop — never continue into the later phases from the shared checkout.

.claude/commands/pr-review-batch.md
mplakhtiy: Done — heading and lead-in both say ten now, and Angle J is in.

.claude/commands/pr-review-batch.md
mplakhtiy: Added in 5f2d4a3b, close to your wording, with the superset case spelled out as a third thing to report — the #548 / #620 example is the one I would otherwise have missed, and "this PR is a subset of another" is worth knowing before spending a review on the subset.

The angle now names three outcomes: a finding accepted on a sibling applies here unchanged, a diverged near-duplicate is itself a finding (naming which copy carries the fix, since merge order decides whether it survives), and a superset relationship gets called out.

.claude/commands/pr-review-batch.md
mplakhtiy: Added as a "Posting, when I ask for it" section in 5f2d4a3b, with your payload and all four rules — `event=COMMENT` only, RIGHT-side lines that the diff shows with `start_line` for ranges, suggestions carrying the complete replacement text at the right indentation, and unanchorable findings going under **Findings outside this diff** in the summary body rather than being forced onto a nearby line.

Kept your `python3 -m json.tool` step and the reason for it: the API rejects the whole review on one bad anchor, so an unvalidated payload usually means posting nothing while believing you posted everything.

Your caveat about a `suggestion` not being able to contain a fenced code block is in there too — that one is not guessable, and this command's own findings are frequently about fenced blocks.

AntonTyb (approved):

PR #624, opened 2026-08-11
fix(agentplugins): build plugin images locally, push to Artifact Regi…

tlipski:
# Pull Request

## Why This Change

Neither plugin installer could run as written on a developer machine, and the workaround —
build the image somewhere else, then pass `PLUGIN_IMAGE=<ref>` — was documented as a
feature rather than treated as the gap it was.

Three separate problems:

- **`gcr.io` is deprecated.** Both installers derived
  `gcr.io/$PROJECT/<plugin>:latest`. Container Registry hosts are being turned down, so a
  fresh install stood a new deployment up on a registry with an end date.
- **`gcloud builds submit` is not available where the installers are run.** Cloud Build
  needs its API enabled, billing wired up, and a credential it will accept for quota —
  corporate SSO credentials are rejected with `ACCESS_TOKEN_TYPE_UNSUPPORTED` on
  `serviceusage.services.use`. On a laptop the build step simply failed.
- **`:latest` made re-installing a silent no-op.** A second install rendered a
  byte-identical `AgentPlugin`, so the operator saw no change, never rolled the pod, and
  the edited skill or adapter stayed unpublished on a deployment that reported healthy.

## What Changed

The image build moves out of Cloud Build and into a small shared library that both
installers source. Because every plugin image is `FROM scratch` plus a COPY of one
directory, the whole build is a single tar layer — which `docker build` or `crane append`
can produce locally, with no daemon required in the crane case.

**Files:**

- `agentplugins/lib/plugin_image.sh` (new)
- `agentplugins/pubsub-platform/install.sh`
- `agentplugins/gke-stockout-investigator/install.sh`
- `agentplugins/pubsub-platform/.dockerignore` (new)
- `agentplugins/gke-stockout-investigator/.dockerignore` (new)
- `agentplugins/pubsub-platform/tests/dedup_e2e_test.py`
- `agentplugins/README.md`, both plugin `README.md`s,
  `agentplugins/pubsub-platform/values.yaml`

**Added/Changed/Fixed:**

- **Artifact Registry instead of `gcr.io`:**
  `<location>-docker.pkg.dev/<project>/<repository>/<plugin>:<tag>`. Which registry is
  **read off the agent's own image** on the `PlatformAgent`, not guessed from a region:
  nothing in this repository grants `artifactregistry.reader`, so whether the kubelet can
  pull a plugin image depends on how the fleet scoped that role, and the agent's image is
  the one reference it is already known to pull. `AR_LOCATION`/`AR_REPOSITORY` pin it;
  `$REGION` then `us-central1` are fallbacks for a first install. A repository is created
  if missing, with a warning that a new one has no reader binding of its own.
- **Local builds:** `IMAGE_BUILDER=auto` picks `docker` when a daemon answers and `crane`
  otherwise; `CRANE_BIN` and `TARGET_PLATFORM` override. Registry auth is per builder
  (`gcloud auth configure-docker` / `crane auth login`). Both builders stamp
  `linux/amd64`: `crane append` on an empty base writes `os: ""`, so a
  `crane mutate --set-platform` follows it.
- **Content-addressed tags:** a 12-character digest of the source tree, the target
  platform and a build-recipe version, so the CR changes exactly when the image changes
  and an installer re-run with nothing changed republishes nothing. The recipe is in the
  digest because a published tag is never rebuilt: were the build alone to change, every
  image already published would keep the old behaviour forever.
- **`PLUGIN_IMAGE_TAG` opts out of the naming, not the safety.** A tag you chose is no
  evidence of what is behind it, so it is rebuilt and overwritten rather than skipped —
  otherwise `PLUGIN_IMAGE_TAG=dev` reintroduces the exact `latest` failure above.
- **Everything that can fail, fails before anything is provisioned.** The project, the
  source tree, the registry and the builder are resolved up front; a missing builder used
  to surface at step 5, after the topic, subscription, log sink and four IAM bindings
  already existed.
- **`PLUGIN_IMAGE` still skips the build**, for pipelines that build once and install many
  times. `demo/provision.sh` sets it per installer on purpose, so a demo stays on the tag
  it rehearsed with.
- **The stockout installer verifies the subscription is attached to its topic.** A
  subscription's topic is fixed at creation, so overriding `STOCKOUT_TOPIC` alone left the
  default subscription on the default topic: the sink published to one queue and the
  adapter pulled another, and not one alert was ever seen without an error anywhere.
- **And the sink at the other end of the same pipe.** Its `create` and `update` are both
  tolerated so a race cannot abort the run, which is precisely why the outcome has to be
  read back: an `update` that did not take leaves the sink exporting to a previous
  `STOCKOUT_TOPIC`, or filtering for a previous `TARGET_CLUSTER_NAME`, while `describe`
  still returns a `writerIdentity` and the install reports done. The destination and the
  cluster-name predicate are now checked afterwards. A sink already exporting somewhere
  else is refused rather than repointed: a sink is project-wide, so an `update` would
  quietly commandeer whatever other deployment depends on it, and the only symptom over
  there is that alerts stop arriving.
- **The agent is named once, and read rather than assumed.** `AGENT_REF` (default
  `platform-agent`) and `HERMES_NAMESPACE` are honoured by both installers — a namespace or
  an agent respected by only one of them puts the route and the adapter that serves it on
  different agents. `AGENT_REF` drives every place the agent is named: the CR's `agentRef`,
  now passed to `helm` rather than left to `values.yaml` because the operator silently
  ignores a plugin whose `agentRef` matches no `PlatformAgent`; the registry the image
  reference is copied from; the service account the roles land on; and the tuning patch.
  That service account is read from `spec.security.serviceAccountName` on the
  `PlatformAgent` instead of hardcoded to `kubeagents-platform-agent` — on a fleet with a
  second agent the fixed name finds nothing, or finds the other agent's identity and grants
  this plugin's roles to it, and the failure surfaces only when the skill's first `gcloud`
  call is denied. A service account carrying no Workload Identity annotation now warns that
  the GSA that follows is a guess.
- **The plugin e2e installs what it tests.** `dedup_e2e_test.py` runs both `install.sh`
  scripts, waits for the gateway rollout and polls for the route before asserting;
  `SKIP_INSTALL=true` reuses an existing deployment. Prebuilt images are named per plugin
  (`PUBSUB_PLUGIN_IMAGE`, `STOCKOUT_PLUGIN_IMAGE`) and an inherited `PLUGIN_IMAGE` is
  dropped — one reference shared by both installers publishes the adapter's files as the
  investigator's image, and `require_skills: true` then refuses every dispatch.
- **Smaller fixes:** `.dockerignore` files keep the docker and crane paths byte-equal by
  excluding `__pycache__`, `*.pyc` and `.DS_Store`; an installer that overruns
  `INSTALL_TIMEOUT` reports as a test failure rather than a traceback.

macOS specifics are handled where they bite: numeric `--uid/--gid` (plain `tar` aborts with
"Numeric user ID too large"), `COPYFILE_DISABLE` against AppleDouble entries, and a member
list that avoids a `./` root entry — `crane export` rejects a layer that has one.

## Why This Matters

- `./install.sh` now works from a checkout with `gcloud`, `kubectl`, `helm`, one image
  builder and an agent to attach to. That is the difference between a plugin somebody can
  install and a plugin somebody has to be told how to install.
- Shipping an edited skill no longer needs a hand-built image and a `kubectl patch` on the
  CR; re-running the installer does it, and does nothing when there is nothing to do.
- New deployments stop being created on a registry that is going away, and land where the
  kubelet is already known to have pull access.
- The e2e now covers the installers themselves, so the charts it exercises are the ones in
  this repository rather than whatever happened to be deployed on the cluster.

## Context

- Container Registry deprecation is the forcing function for the registry move.
- The operator AgentPlugins e2e (`tests/e2e/operator/agentplugins_e2e_test.py`) is
  unchanged on purpose: it builds a synthetic plugin to drive operator behaviour it cannot
  reach with a real one (missing images, duplicate names, ghost profiles), so `install.sh`
  has nothing to offer it.
- `demo/provision.sh` (already on `main`) pins `PLUGIN_IMAGE` for the demo fleet — a demo
  that moves between the rehearsal and the room is the reason — and is unaffected here. It
  sets the variable separately per installer, which is the shape the e2e now follows.
- Follow-up, not addressed here: the `pubsub-platform` image deployed on the demo fleet was
  found to be **ahead of this repository** (a lazy `google.cloud.logging` install, a
  configurable sink name, a delivery-entry cap) with no matching commit on any branch.
  Re-installing rolls those back. A copy of the deployed `adapter.py` was kept so the work
  can be recovered and landed properly.

## Testing

Run against `ka-dev-mgmt` (project `tomeklipski-izrhgv`, `europe-west1`) on a host with no
Docker daemon and no usable Cloud Build — the environment the old script could not run in
at all. Clean install both times: `helm uninstall` of both releases, zero `AgentPlugin`s
and a completed gateway rollout before the test started.

- **`dedup_e2e_test.py`, clean install, first run: all five assertions passed** (incident
  filed, identical alert suppressed, retry variants suppressed, second workload filed
  separately, workload-less summary dropped by the filter). Cleanup archived its tasks and
  restored the dedup registry.
- That run also caught a real defect: the images were already published from an earlier
  build and reported `already published`, but carried `os: ""` — the digest covered the
  source only, so the new platform stamp could never reach them. Fixed by putting the
  build recipe and target platform in the digest, which is what the second run verifies.
- **Second clean install, after the fix: all five assertions passed again**, this time with
  both images genuinely rebuilt under new tags (`pubsub-platform:ved0044c39253`,
  `gke-stockout-investigator:v2b35c968b9b6`), both reporting `os=linux arch=amd64`, both
  `AgentPlugin`s `Ready` and the image volumes mounted by the kubelet.
- **Third install: republished nothing** (`already published`, no rollout) — idempotent.
- **Registry discovery exercised for real:** this fleet's agent runs from `gcr.io`, so
  discovery correctly declined to infer anything and the fallback applied. Also unit-tested
  against synthetic agent images (Artifact Registry, custom repository, `ghcr.io`, no agent
  deployed, `REGION` set, explicit pins).
- **Failure paths, against stubbed `gcloud`/`kubectl`/`helm`/`crane`:** a subscription
  attached to another topic aborts at step 2b before the log sink is touched; no image
  builder aborts before step 1 in both installers.
- **Library behaviour against a local `crane registry`:** content tag publishes then skips,
  a pinned tag rebuilds every time, an empty source tree is refused, and the pushed layer
  has uid/gid 0, world-readable modes, no `__pycache__` and no `./` root entry.
- `python3 -m unittest` over `agentplugins/pubsub-platform/tests` (27 tests),
  `make docs-check`, `prettier --check agentplugins/`, `bash -n` on all three shell files,
  and the library exercised under macOS `bash` 3.2.

Separately, on the demo fleet (`ka-demo-mgmt` / `ka-demo-cluster-1`), where both plugins
now run the images these installers built — `pubsub-platform:v3bb8daa32322` and
`gke-stockout-investigator:vf60463d110d6`, content tags in
`europe-west1-docker.pkg.dev/tomeklipski-izrhgv/kube-agents`, both `AgentPlugin`s `Ready`:

- **A real stockout scenario, end to end.** `scenarios/01-gpu-regional-scarcity.sh` wedged
  three GPU pods on the target cluster and published the matching alert. It passed the
  route's cluster-name filter, the skill resolved in the platform profile, the agent
  notified Chat and filed a task, and the investigation ran the full Rule E path — pod
  events, `gcloud compute regions describe`, `gcloud beta compute advice capacity`, the
  GitOps clone — before standing down on its own duplicate-PR check, an open PR from the
  previous run of the same scenario. 130s from alert to task done.
- That is also the first live confirmation that the rolled-back `pubsub-platform` adapter
  noted under Context serves the alert path correctly: this fleet had been running the
  unversioned image that was ahead of this repository, and re-installing replaced it.

---

Default behaviour changes for anyone re-running an installer: the image moves from
`gcr.io/<project>/<plugin>:latest` to an Artifact Registry reference with a content tag,
and the first run creates the repository if it is absent. Existing deployments are
untouched until their installer is re-run, and `PLUGIN_IMAGE=<ref>` reproduces the old
behaviour exactly.


tlipski:
need to resolve merge conflicts

tlipski:
tested the installation with plugins also with a cloud shell environment to be extra sure it works e2e

agentplugins/gke-stockout-investigator/install.sh
```diff
@@ -111,12 +218,32 @@ gcloud pubsub topics add-iam-policy-binding "$TOPIC" \
...
+#
+# The KSA is read off the agent rather than hardcoded. Its name is spec.security.
+# serviceAccountName, which is per-agent, so a fleet with a second PlatformAgent runs under
+# a KSA of its own — and looking up a fixed `kubeagents-platform-agent` there finds either
+# nothing or, worse, the OTHER agent's identity, and grants this plugin's roles to it.
+# Nothing errors: the alerts arrive, the skill runs, and its gcloud calls are denied.
+KSA_NAME="$(kubectl --context="$CONTEXT" -n "$NAMESPACE" get platformagent "$AGENT_REF" \
+    -o jsonpath='{.spec.security.serviceAccountName}' 2>/dev/null || echo "")"
+# The operator's own default when the field is unset (k8s-operator/scripts/common.sh).
+KSA_NAME="${KSA_NAME:-kubeagents-platform-agent}"
```
mplakhtiy: **MEDIUM — the KSA fallback is not the operator's default, so the plugin's IAM can land on the wrong identity**

The operator's default is not `kubeagents-platform-agent`. When `spec.security.serviceAccountName` is empty, both `platformagent_manifests.go:1158` (`saName := agent.Name`) and `platformagent_controller.go:298` fall back to the **PlatformAgent's own name** — that is, to `$AGENT_REF`. `kubeagents-platform-agent` is what `k8s-operator/scripts/common.sh:361` exports for the provisioning scripts and what `charts/kube-agents/values.yaml:105` sets explicitly; it is a convention of those two callers, not a defaulting rule in the operator.

This is the one remaining hardcode in a step the PR otherwise went to some length to make per-agent, and the comment states the wrong source of truth for it.

**Failure scenario.** A `PlatformAgent` created without `security.serviceAccountName` — a hand-written CR, a second agent, or `--set platformAgent.security.serviceAccountName=""`. The operator creates and runs the pod as KSA `$AGENT_REF`. The installer looks up `kubeagents-platform-agent`, does not find it, warns, and guesses `kubeagents-platform-gsa@<project>` — which on any project provisioned by `k8s-operator/scripts/` **does exist** (`common.sh:363`). The bindings therefore succeed quietly: `roles/pubsub.subscriber` on the topic and subscription and project-level `roles/compute.viewer` land on an identity this agent does not use, the agent's real GSA gets nothing, and the failure surfaces only when the skill's first `gcloud` call is denied — exactly the outcome the surrounding comment says the change was made to prevent. The warning does fire, but it names a ServiceAccount the operator never configured, so it reads as a bug in the installer rather than as a pointer at the real cause.

**Fix.** Use `$AGENT_REF` as the fallback, and cite `k8s-operator/internal/controller/platformagent_manifests.go:1158` rather than `common.sh`.

```suggestion
# The operator's own default when the field is unset
# (k8s-operator/internal/controller/platformagent_manifests.go:1158).
KSA_NAME="${KSA_NAME:-$AGENT_REF}"
```

agentplugins/lib/plugin_image.sh
```diff
@@ -0,0 +1,901 @@
...
+# The files the image ships, relative to the current directory (which must be src_dir).
+#
+# One definition, used by both the content tag and the crane layer, so the two cannot
+# drift: a file that changes the tag must be a file that reaches the image. `prefix` is
+# what plugin_image_src_prefix returned, so the exclusions are applied to the same path
+# `docker build` would apply them to.
+plugin_image_source_files() {
+    local prefix="$1" file
+
+    find . -type f | while IFS= read -r file; do
```
mplakhtiy, line 298: **MEDIUM — a symlink in the source tree is shipped in the layer but invisible to the content tag**

`plugin_image_source_files` is the single definition of "the files the image ships", and its comment says so: *"a file that changes the tag must be a file that reaches the image."* `find . -type f` excludes symlinks. `plugin_image_stage` does not — `cp -R` preserves them, and the crane tar writes them out as symlink entries.

Verified directly — retargeting a symlink left the tag unchanged:

    tag before=v4e3a8b34a676   after retargeting symlink=v4e3a8b34a676

and the resulting layer carried it:

    -rw-r--r--  0 0  0   4 real.txt
    lrwxr-xr-x  0 0  0   0 link.txt -> real.txt

**Failure scenario.** Re-point a symlink inside `files/` (or add one) and the content tag does not move. `plugin_image_publish` then reports `already published — the source has not changed since it was built` and republishes nothing, so the edit never reaches a cluster on a deployment that reports healthy. That is precisely the `latest` failure the content tag exists to eliminate, reintroduced through a file type the digest cannot see. It compounds: whether `docker build`'s `COPY` preserves a symlink or dereferences it is builder-dependent, so the two builders can put different bytes under one tag — the silent divergence `plugin_image_ignore_load`'s refusal discipline was written to prevent, arriving by a route that discipline does not cover.

Neither shipped plugin contains a symlink today, so this is latent rather than live.

**Fix.** Mirror the treatment the library already gives an empty tree and an unsupported `.dockerignore` pattern — refuse rather than guess. A `find . -type l` check in `plugin_image_content_tag` that dies with "symlinks are not supported in a plugin source tree" is one line and fails closed. Hashing the link target instead would also work, but only if `plugin_image_stage` and both builders are then known to agree on what a symlink becomes.

agentplugins/lib/plugin_image.sh
```diff
@@ -0,0 +1,901 @@
+#!/usr/bin/env bash
```
mplakhtiy, line 1: **MEDIUM — 901 lines of new shell with no automated test, and the body describes tests that are not in the diff**

This finding is about the file as a whole.

The new library carries a bespoke `.dockerignore` parser and segment-wise glob matcher, a content-digest scheme, Artifact Registry discovery and parsing, a staging step and two publish paths. CI exercises none of it: `.github/workflows/agentplugins-test.yml:60` discovers `agentplugins/*/tests/test_*.py` (there is no `agentplugins/lib/tests/`), and the only other step that touches these files is `bash -n`, a syntax check.

The PR body reports "unit-tested against synthetic agent images (Artifact Registry, custom repository, `ghcr.io`, no agent deployed, `REGION` set, explicit pins)" and "Library behaviour against a local `crane registry`". Those runs happened — I reproduced their results by hand — but none of them are committed, so they protect nothing after this merges.

**Failure scenario.** The regression that ships silently is a change to `plugin_image_glob_segments` or `plugin_image_ignore_load` that lets a pattern match differently from Docker — say `*` crossing a `/`. The tag still mints, the publish still succeeds, and the two builders ship different file sets under one tag. Because a published tag is never rebuilt, whichever builder reaches it first defines that image permanently. The library's own header calls this out as the worst case it is designed against; nothing would catch it. Second on the list: forgetting to bump `PLUGIN_IMAGE_RECIPE` after a build change, which the file documents as a manual step with no check behind it.

**Fix.** The workflow loop already globs `agentplugins/*/tests`, so `agentplugins/lib/tests/test_plugin_image.py` — stdlib `unittest` shelling out to `bash -c '. lib/plugin_image.sh; ...'` — would be picked up with no workflow change. The highest-value cases are the ones I ran by hand: the matcher against root-level, nested and directory paths; the tag moving for a content change and holding still for a no-op; registry discovery over the synthetic agent images; and the refusals (empty tree, negation, `src_dir` outside `plugin_dir`).

mplakhtiy (commented):
Three findings, all MEDIUM. No correctness blockers: I executed the `.dockerignore` matcher, the content-tag digest, registry discovery over six synthetic agent images, tar layer construction, and all three refusal paths, and both content tags I computed match the tags the PR body reports from its live runs.

One note so nobody re-litigates it: an early pass appeared to show a matcher bug under `zsh`. That was a zsh artifact (no `GLOB_SUBST`); under `/bin/bash` 3.2 the matcher is correct.

mplakhtiy (approved):

tlipski:
> Three findings, all MEDIUM. No correctness blockers: I executed the `.dockerignore` matcher, the content-tag digest, registry discovery over six synthetic agent images, tar layer construction, and all three refusal paths, and both content tags I computed match the tags the PR body reports from its live runs.
> 
> One note so nobody re-litigates it: an early pass appeared to show a matcher bug under `zsh`. That was a zsh artifact (no `GLOB_SUBST`); under `/bin/bash` 3.2 the matcher is correct.

Addressed in #675 (it's mostly adding a regression test)

PR #631, opened 2026-08-11
fix(installer): repair terminal output and make provision scripts the source of truth

mplakhtiy:
# Pull Request

## Why This Change

Running `./install.sh` on macOS printed raw escape codes instead of colour — every status line
read `\e[96mℹ Environment Detected…` — and the banner rules were glued to the ASCII art. Chasing
that surfaced a security-relevant bug and a structural one.

**The permission-set menu defaulted to `gke-admin`.** Option 1 was "SRE GitOps & Remediations", so
pressing Enter escalated the agent GSA to `roles/container.admin`, while `common.sh`, the
`--permission-set` flag, and the docs all default to `read-only`. The same menu advertised the
GitOps pull-request path as exclusive to `gke-admin`; that path works under every permission set,
because what makes the agent propose rather than apply is Kubernetes RBAC — read-only in all of
them. It also omitted the `custom` set entirely.

**That bug was a symptom.** `install.sh` restated defaults and validation that
`k8s-operator/scripts/` already owns, and the two copies had drifted: region (`us-central1` vs
`us-east4`), cluster name (`kube-agents-platform` vs `platform-agent-host`), the accepted model
providers (`chatgpt` missing), the model-name table in six places, and a verbatim second copy of
`derive_kms_location`.

## What Changed

**Files:**

- `k8s-operator/scripts/common.sh` — new shared constants and validators
- `k8s-operator/scripts/provision_0*.sh`, `provision_1*.sh`, `dev/dev_rebuild_agent.sh` — use them
- `install.sh` — sources `common.sh`; terminal, project-picker, image-tag and allowlist fixes
- `k8s-operator/scripts/README.md`, `INSTALL.md`, `.agents/skills/install-kube-agents/SKILL.md`,
  `docs/site/src/content/docs/reference/security-and-iam.md`,
  `docs/site/src/content/docs/install/quickstart-gke.mdx` — docs parity

**Added/Changed/Fixed:**

- `common.sh` gains `DEFAULT_CLUSTER_NAME`, `DEFAULT_REGION`, `DEFAULT_MODEL_PROVIDER`,
  `default_model_for_provider()`, `is_valid_model_provider()`, `is_valid_permission_set()`. The
  provision scripts reference the constants instead of repeating literals.
- `install.sh` acquires the provisioning sources at **step 2** (was step 9), sources `common.sh`,
  and deletes its duplicated defaults, validators, and `derive_kms_location`.
- Permission-set menu rewritten: `read-only` first and default, `gke-admin` second, `custom` third,
  with a new `--custom-roles` flag persisted as `PLATFORM_AGENT_CUSTOM_ROLES`. Labels no longer
  claim GitOps is gated on the set.
- `--allow-unverified-source` (alias `--allow-dirty`, env `ALLOW_UNVERIFIED_SOURCE`) opts out of
  the source/image check for local iteration; every strict failure now prints how to override.
- Colours use `\033` rather than `\e` (bash 3.2's `echo -e` does not expand `\e`);
  `draw_separator` ends its own line.
- Project picker accepts a typed project ID, leads with the active project, and reports truncation
  — a project outside the first ten used to be unreachable in practice.
- Image tag defaults to the checkout's `HEAD`, displayed as `local HEAD checkout <short-sha>` while
  storing the full SHA.
- Google Chat allowlist defaults to an empty list rather than the invoking gcloud account, shown as
  `[default: empty list]`.
- Docs: `scripts/README.md` documents the shared defaults and the installer's delegating role; the
  skill's flag table went from 10 to 21 entries and stops calling the permission set an "RBAC
  scope"; `INSTALL.md` and the security reference name the installer alongside the provisioner.

## Why This Matters

- A default that escalates IAM on Enter is the wrong way round; `read-only` is the documented,
  fail-closed posture and is now what the installer picks.
- One home per default means the installer and the pipeline cannot disagree again — the class of
  bug above becomes impossible rather than merely fixed.
- The installer was unusable-looking on any macOS system bash, which is most laptops running it.
- Failing the source check before the interview instead of after it saves a dozen re-answered
  prompts per attempt.

## Context

`install.sh` never duplicated provisioning itself — it writes `vars.sh` and runs
`make gcp-provision`. What it duplicated was policy, which is what this PR removes.

⚠️ **Behaviour change:** the installer's default region is now `us-east4` (the pipeline's value),
not `us-central1`. Explicit `--region` is unaffected; docs and examples are updated.

## Testing

- Full non-interactive dry run against a real project: steps renumbered 1–13, `Permission
  Boundary: read-only`, `AI Model Provider: gemini (gemini-3.5-flash)` resolved through the shared
  function, `KMS_LOCATION` derived by `common.sh`.
- Interactive paths exercised under a pty on bash 3.2.57: escapes render as real `ESC[`, the
  permission menu defaults to `read-only`, the project picker rejects out-of-range numbers and
  malformed IDs while accepting an unlisted project ID, and the image-tag prompt stores the full
  SHA behind the short label.
- Source verification: strict mode aborts with the override hint; `--allow-unverified-source`
  warns and continues; a clean checkout still reports verified; the second call is memoised silent.
- `bash -n` on `install.sh` and every changed provisioning script; `make docs-check` passes;
  `prettier --write` applied to changed Markdown. `shellcheck` could not be run locally (the
  Homebrew binary is killed on this machine even for `--version`), so CI's run is the check.

---

Low risk for the provisioning pipeline: those scripts changed only by replacing two literals with
the constants that hold the same values. The installer's flow changed more substantially — the
region default and the earlier source check are the two behaviours to look at first.

### DIFF
--- a/.agents/skills/install-kube-agents/SKILL.md
+++ b/.agents/skills/install-kube-agents/SKILL.md
@@ -7,6 +7,19 @@ description: Provision and install the Kubernetes Agentic Harness (kube-agents) 
 
 This skill provides step-by-step instructions for AI Agents to non-interactively provision Google Cloud GKE infrastructure and deploy the `kube-agents` Platform Agent.
 
+## What `install.sh` actually does
+
+It is a front-end, not a second provisioner. It collects configuration, writes
+`k8s-operator/scripts/vars.sh`, and then runs `make gcp-provision` — the pipeline in
+[`k8s-operator/scripts/`](../../../k8s-operator/scripts/README.md) does every GCP and GKE
+operation. The installer sources `k8s-operator/scripts/common.sh` before its first prompt, so its
+defaults and accepted values are the ones defined there; that file is where a default changes.
+
+Order of operations: resolve the image/source ref → check CLI prerequisites → put the provisioning
+scripts on disk and verify them against that ref → interview → write `vars.sh` → run the pipeline.
+The source check happens **before** the interview, so a bad ref fails in seconds rather than after
+a dozen answers.
+
 ## Quick Execution for AI Agents
 
 To run the installer non-interactively in automated subagent execution, pass `--non-interactive` along with explicit configuration flags:
@@ -15,12 +28,16 @@ To run the installer non-interactively in automated subagent execution, pass `--
 curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/main/install.sh | bash -s -- \
   --non-interactive \
   --project-id="YOUR_GCP_PROJECT_ID" \
-  --cluster-name="kube-agents-platform" \
-  --region="us-central1" \
+  --cluster-name="platform-agent-host" \
+  --region="us-east4" \
   --image-tag="<SEMVER_TAG_OR_FULL_COMMIT_SHA>" \
   --model-provider="gemini" \
   --permission-set="read-only"
 ```
+
+`--image-tag` accepts a SemVer release tag or a full 40-character commit SHA; mutable refs
+(`latest`, `main`, `master`, `HEAD`) are rejected. When the installer runs from a Git checkout it
+defaults to that checkout's `HEAD`, so the flag is only required on the `curl | bash` path.
 
 ## Dry-Run Inspection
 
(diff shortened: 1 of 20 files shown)


eLeontev (approved):
I will validate it tomorrow with proper scope, right now the wizard is still considered as beta.

PR #633, opened 2026-08-11
fix(operator): serve admission webhooks on 10250 for private GKE

bradhoekstra:
# Pull Request

## Why This Change

A production cluster wedged: the GKE control plane could not reach the operator's webhook server on
`9443`.

GKE creates a single control-plane-to-node firewall rule, and it permits only `tcp:443` and
`tcp:10250`. The API server dials the endpoint **pod IP** on the webhook Service's `targetPort`, so
on a private cluster a webhook on any other port is unreachable until someone adds a VPC firewall
rule by hand, per cluster. Both webhook configurations carry `failurePolicy: Fail`, so the
unreachable webhook blocked every `PlatformAgent` create, update, and delete — including the edits
needed to recover.

`9443` was never a deliberate choice: `cmd/main.go` built `webhook.Options{}` with no `Port` and
inherited controller-runtime's default, which is why `9443` appeared nowhere in the Go source.

## What Changed

The webhook server binds `10250`, which lands inside the firewall rule GKE already made — no
per-cluster VPC rule, no action on existing clusters beyond a rollout. It does not collide with the
kubelet, which binds `10250` on the *node* IP in a different network namespace. `443` was the other
in-rule option, but it is privileged and would mean adding `NET_BIND_SERVICE` back to the manager's
`drop: ["ALL"]` securityContext; `10250` needs no capability.

**Files:**

- `k8s-operator/internal/webhook/platformagent_webhook.go` — new `DefaultPort = 10250` with the
  rationale; the single source the flag, the manifests, and the test key off.
- `k8s-operator/cmd/main.go` — `--webhook-port` flag defaulting to `DefaultPort`, wired into
  `webhook.Options{Port: ...}`.
- `k8s-operator/config/manager/manager.yaml` — `containerPort: 10250`.
- `k8s-operator/config/webhook/service.yaml` — `targetPort: 10250`; `port: 443` unchanged, since
  that is what the `*WebhookConfiguration` `clientConfig` resolves to rather than what crosses the
  network.
- `k8s-operator/internal/testing/webhook_port_test.go` — new `TestWebhookPortsMatchDefault`.
- `charts/kube-agents/templates/operator-deployment.yaml` — port in the `ENABLE_WEBHOOKS=false`
  comment.
- Docs: `docs/site/.../operator/index.md` (new **Admission webhooks** section, canonical),
  `install/prerequisites.md`, `deploy/kustomize.md`, `k8s-operator/README.md` link to it;
  `docs/README.md` gains the identifier-source row.

**Added/Changed/Fixed:**

- Fixed: private GKE clusters can reach the admission webhooks without a hand-added firewall rule.
- Added: `--webhook-port`, so clusters where `10250` is not the reachable port are not stuck with a
  GKE-shaped default.
- Added: a drift guard. The port lives in three places now, and a silent mismatch reproduces the
  outage — `TestWebhookPortsMatchDefault` parses both manifests and fails the build if either
  disagrees with `DefaultPort`.

## Why This Matters

- `failurePolicy: Fail` plus an unreachable webhook is a wedged cluster, not a degraded one. The
  operator's own CRD is the tool you would reach for to fix it.
- The trap was silent and per-cluster: nothing in the repo or the install path warned that a private
  cluster needed a firewall rule, so every new private cluster inherited it.

## Context

**This is a workaround, and hopefully a temporary one.** It is worth being explicit about what it
does not fix, and about the way it is load-bearing:

- It depends on a GKE firewall rule this repo does not own. Google's documented fix for this failure
  is "add a firewall rule for your port"; reusing the kubelet's port is not a supported contract, and
  a future narrowing of that rule to node IPs would break every cluster at once with the same
  fail-closed symptom.
- The manifests become GKE-shaped. The port means nothing on EKS, kind, or bare clusters — hence the
  `--webhook-port` escape hatch.
- Security tooling reads `:10250` as kubelet and will flag it.
- It closes the firewall gap only. cert-manager stays a prerequisite, `failurePolicy: Fail` still
  couples all `PlatformAgent` writes to operator availability, and Helm chart installs still run
  `ENABLE_WEBHOOKS=false` and get no validation at all.

**Proposed follow-up:** replace the webhooks with CRD CEL validation plus a
`ValidatingAdmissionPolicy`, which removes all four problems — the API server evaluates CEL
in-process, so there is no port, no certificate, and no network dependency. Most of the migration is
already half-done: the mutating webhook's two defaults are already covered by `+kubebuilder:default`
markers (`common_types.go:190,172`), and `gitRepo` validation is already duplicated in the controller
as a `Degraded`/`InvalidGitRepoURL` condition. What genuinely needs new homes is the cross-object
singleton check (→ controller status condition) and the `prevent-deletion` annotation (→ VAP on
`DELETE`, which would also give chart installs delete-protection for the first time). I'm happy to
open that as a follow-up if reviewers agree with the direction.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt` — clean.
- `go test ./...` in `k8s-operator/` — all packages pass, including the existing webhook suite
  (validation logic untouched).
- Drift guard verified to actually bite: flipping `targetPort` back to `9443` fails with
  `webhook Service targetPort = 9443, want 10250 (agentwebhook.DefaultPort)`.
- `kustomize build config/default` renders `containerPort: 10250` and `443 → 10250`.
- `make docs-check`, `prettier --check` on changed files, and `cd docs/site && npm run build` — pass.

After the review round (`b5683aa8`):

- `--webhook-port` range guard verified against the built binary: `./manager --webhook-port=0` exits
  `1` with `invalid --webhook-port: must be between 1 and 65535 {"port": 0}`, instead of
  controller-runtime silently substituting its own `9443` default (`pkg/webhook/server.go:139-140`).
- The documented Kustomize patch verified with `kustomize build` on an overlay over
  `config/default`: renders `--webhook-port=8443`, `containerPort: 8443`, `targetPort: 8443`, and
  Service `port: 443` unchanged.
- New heading anchors (`#serving-on-a-different-port`,
  `#upgrading-from-an-operator-that-served-9443`) confirmed present in the built site output.

**Not yet run — needs a cluster, and the reason for the change:** an apply against a private GKE
cluster proving the API server now reaches the webhook (a policy *rejection* rather than a timeout),
kubelet health on the node running the operator pod, and a kind run with `--webhook-port=8443` plus
the matching `containerPort`/`targetPort` patch to confirm the override still works where `10250` is
not special.

---

Rollout: **re-apply the manifests; do not bump the image alone.** `targetPort` is a manifest field,
so a `kubectl set image` — or any pipeline that rolls the tag without re-applying
`config/webhook/` — leaves the Service on `9443` while the new pod listens on `10250`, which is this
same outage arriving on a routine version bump. `make deploy` applies both. Even applied together
there is a brief fail-closed window between the Service change landing and the new pod becoming
Ready. (An earlier version of this paragraph said "nothing to do on the cluster itself"; that was
wrong, and [#633 review](https://github.com/gke-labs/kube-agents/pull/633#discussion_r3784715970)
caught it. The operator overview now documents the upgrade.)

Rollback is the same two commands as the incident recovery — delete the two
`*WebhookConfiguration`s and set `ENABLE_WEBHOOKS=false` — now documented in the operator overview.


toshiowang (commented):
The port change itself checks out: the API server does resolve `clientConfig.service` through Endpoints to pod IP + `targetPort`, no manifest in the repo sets `hostNetwork` so there is no kubelet collision, all three `9443` occurrences repo-wide are updated with none missed, the recovery commands match the `namePrefix: kubeagents-` resource names, and the new test compiles into the existing `package testing` with both YAML deps already direct in `go.mod`. The findings are about the escape hatch as documented.

Three pages tell the reader that a cluster where 10250 is not reachable "can override the flag" — singular. Following that literally binds the new port while `config/webhook/service.yaml` still targets 10250, so the API server dials a port nothing is listening on and `failurePolicy: Fail` wedges every `PlatformAgent` write: the exact outage this PR fixes, now reachable by following its own instructions. The new drift guard sharpens this, since the correct three-place edit fails the build unless `DefaultPort` moves too. Two tentative items follow: `--webhook-port=0` silently falls back to controller-runtime's 9443 (`pkg/webhook/server.go:139-140`) with no validation, and the rollout note ("nothing to do on the cluster itself") does not hold for existing Kustomize installs, where an image-only bump leaves the Service on 9443.

docs/site/src/content/docs/operator/index.md
```diff
@@ -81,6 +81,40 @@ proxy only bootstraps a kubectl context when it has the complete triple; leave a
...
+on any other port is unreachable until someone adds a VPC firewall rule for it — per cluster, by
+hand. Serving on 10250 lands inside the rule GKE already made. It does not collide with the kubelet,
+which binds 10250 on the node IP in a different network namespace.
+
+The port is set in three places that must agree, and `TestWebhookPortsMatchDefault` fails the build
+if they drift: the `--webhook-port` flag default (`DefaultPort` in
+`internal/webhook/platformagent_webhook.go`), the manager `containerPort`, and the Service
+`targetPort`. Clusters where 10250 is not the reachable port can override the flag; the Service
+`port` stays `443` either way, since that is what the `*WebhookConfiguration` `clientConfig` resolves
+to rather than what crosses the network.
```
toshiowang: **[CRITICAL BUG]** The sentence above correctly says the port lives in three places "that must agree," then tells the reader a cluster where 10250 is not reachable "can override the flag" — singular.

An operator who follows that sets `--webhook-port=8443`; the pod binds 8443, `config/webhook/service.yaml` still says `targetPort: 10250`, so the API server dials 10250 on the pod IP with nothing listening. `failurePolicy: Fail` then blocks every `PlatformAgent` create, update, and delete — the exact wedge this PR fixes, reached by following this PR's own instructions.

`TestWebhookPortsMatchDefault` compounds it: the *correct* three-place edit now fails the build unless `DefaultPort` moves too, which is worth stating here.

```suggestion
`targetPort`. Overriding `--webhook-port` alone is not enough: the Service `targetPort` and the
manager `containerPort` must move with it, or the API server dials a port nothing is listening on.
The Service `port` stays `443` either way — that is what the `clientConfig` resolves to, not what crosses the network.
```

docs/site/src/content/docs/install/prerequisites.md
```diff
@@ -21,6 +21,12 @@ The shipping install path targets GKE. You'll need one working GCP project plus
 
 The provisioner will enable APIs and create all resources itself; you don't need to pre-provision the cluster.
 
+**No extra firewall rule is needed on private clusters.** The operator's webhook server listens on
+`10250`, one of the two ports GKE's automatic control-plane-to-node rule already permits — see
+[Admission webhooks](/kube-agents/operator/#admission-webhooks). A cluster that hardens `10250`
+beyond the GKE default (scoping it to node IPs, say) still needs a rule for the webhook, or the
+`--webhook-port` flag pointed at a port it does allow.
```
toshiowang: **[CRITICAL BUG]** Same one-knob claim as the operator page, reached from the install path: a hardened cluster that sets only `--webhook-port` leaves the Service targeting 10250 and wedges all `PlatformAgent` writes behind `failurePolicy: Fail`.

There is also no shipped way to pass the flag — `config/manager/manager.yaml` hardcodes `args` and chart installs run `ENABLE_WEBHOOKS=false` — so what this actually requires is a kustomize patch across three files, not a flag.

```suggestion
`--webhook-port` flag, the manager `containerPort`, and the Service `targetPort` moved together to a port it does allow.
```

k8s-operator/README.md
```diff
@@ -193,6 +193,8 @@ ENABLE_WEBHOOKS=false go run ./cmd/main.go
 > [!TIP]
 > This compiles and runs the entry point [main.go](cmd/main.go) with webhooks disabled. The process runs in the foreground, prints reconciliation logs, and watches for custom resource events in the cluster.
 
+When webhooks are enabled, the server binds `10250` rather than Kubebuilder's usual `9443`: it is one of only two ports GKE's automatic control-plane-to-node firewall rule permits, so a private cluster reaches the webhook without a hand-added VPC rule. `--webhook-port` overrides it where 10250 is not the reachable port. The rationale, the drift guard across the three places the port is written, and the recovery steps for an unreachable webhook are in [Admission webhooks](../docs/site/src/content/docs/operator/index.md#admission-webhooks).
```
toshiowang: **[CRITICAL BUG]** Third copy of the one-knob claim. `--webhook-port` alone does not move the Service `targetPort`, so overriding it as described makes the webhook unreachable and fail-closed admission blocks all `PlatformAgent` writes.

Suggest changing "`--webhook-port` overrides it where 10250 is not the reachable port" to "`--webhook-port`, the manager `containerPort`, and the Service `targetPort` must be changed together where 10250 is not the reachable port", so all three pages agree with the section they link to.

k8s-operator/cmd/main.go
```diff
@@ -83,6 +84,13 @@ func main() {
 	flag.StringVar(&webhookCertPath, "webhook-cert-path", "", "The directory that contains the webhook certificate.")
 	flag.StringVar(&webhookCertName, "webhook-cert-name", "tls.crt", "The name of the webhook certificate file.")
 	flag.StringVar(&webhookCertKey, "webhook-cert-key", "tls.key", "The name of the webhook key file.")
+	flag.IntVar(&webhookPort, "webhook-port", agentwebhook.DefaultPort,
+		"The port the webhook server binds to. Defaults to 10250 because GKE's automatic "+
+			"control-plane-to-node firewall rule only permits tcp:443 and tcp:10250; on a private "+
+			"cluster the API server cannot reach a webhook on any other port without a manual VPC "+
+			"firewall rule, and failurePolicy=Fail then blocks all PlatformAgent writes. The "+
+			"kubelet's 10250 is bound on the node IP in a different network namespace, so there is "+
+			"no conflict. Override this on clusters where 10250 is not the reachable port.")
```
toshiowang: **[TENTATIVE RISK]** The flag is unvalidated, and controller-runtime v0.19.0 does `if o.Port <= 0 { o.Port = DefaultPort }` (`pkg/webhook/server.go:139-140`) where its `DefaultPort` is 9443.

So `--webhook-port=0` silently binds 9443 — the one port this whole change exists to avoid. That spelling is not far-fetched: `--metrics-bind-address=0` in this very Deployment uses `0` to mean "off." The only signal is an info-level `Serving webhook server … "port": 9443`, and fail-closed admission then blocks all `PlatformAgent` writes.

Not a drop-in edit to these lines; suggest after `flag.Parse()`:

```go
if webhookPort < 1 || webhookPort > 65535 {
    setupLog.Error(nil, "invalid --webhook-port", "port", webhookPort)
    os.Exit(1)
}
```

docs/site/src/content/docs/operator/index.md
```diff
@@ -81,6 +81,40 @@ proxy only bootstraps a kubectl context when it has the complete triple; leave a
...
+webhook deployment:
+
+```bash
+kubectl delete validatingwebhookconfiguration kubeagents-validating-webhook-configuration
+kubectl delete mutatingwebhookconfiguration kubeagents-mutating-webhook-configuration
+kubectl -n kubeagents-system set env deploy/kubeagents-controller-manager ENABLE_WEBHOOKS=false
+```
+
+That leaves the cluster with the same validation coverage a chart install has. Re-apply with
+`make deploy` once the cause is fixed.
```
toshiowang, line 163: **[TENTATIVE RISK]** The PR body says "the port ships in the operator image, so a cluster picks it up on its next rollout; nothing to do on the cluster itself." That does not hold for existing Kustomize installs: `targetPort` is a manifest field, so rolling the image alone (`kubectl set image`, or any pipeline that bumps the tag without re-applying `config/webhook/service.yaml`) leaves the Service on 9443 while the new pod listens on 10250 — unreachable webhook, wedged cluster, on a routine version bump.

Worth a sentence here that the upgrade requires re-applying the manifests and not just the image, and that there is a brief unreachable window between the Service change landing and the new pod becoming Ready.

docs/site/src/content/docs/operator/index.md
bradhoekstra: Agreed, and fixed in b5683aa8 — this was the real bug in the change.

I went further than the suggestion because the one-knob claim is worse than it reads: `config/manager/manager.yaml` hardcodes `args`, so there was no shipped way to pass `--webhook-port` in the first place. The override is a Kustomize patch or nothing. The section now says that and carries a worked patch:

```yaml
- target: {kind: Deployment, name: controller-manager}
  patch: |
    - op: add
      path: /spec/template/spec/containers/0/args/-
      value: --webhook-port=8443
    - op: replace
      path: /spec/template/spec/containers/0/ports/1/containerPort
      value: 8443
- target: {kind: Service, name: webhook-service}
  patch: |
    - op: replace
      path: /spec/ports/0/targetPort
      value: 8443
```

Verified with `kustomize build`: renders `--webhook-port=8443`, `containerPort: 8443`, `targetPort: 8443`, and `port: 443` unchanged.

Your point about the drift guard is in there too — changing `DefaultPort` instead of patching means `TestWebhookPortsMatchDefault` fails the build unless both manifests move with it.

docs/site/src/content/docs/install/prerequisites.md
bradhoekstra: Fixed in b5683aa8, and your second paragraph is the sharper half of this — there is no shipped way to pass the flag, so "or the `--webhook-port` flag pointed at a port it does allow" was describing something an operator could not actually do. The prerequisites page now sends them to the patch:

> ... or a move to a port it does allow — which is a Kustomize patch across the `--webhook-port` flag, the manager `containerPort`, and the Service `targetPort` together, not a single flag. Changing one of the three leaves the API server dialing a port nothing is listening on; see [Serving on a different port](/kube-agents/operator/#serving-on-a-different-port).

The worked patch lives on the operator page rather than here, since that page owns the topic.

k8s-operator/README.md
bradhoekstra: Fixed in b5683aa8, with your wording. The README line now reads:

> Where 10250 is not the reachable port, `--webhook-port`, the manager `containerPort`, and the Service `targetPort` have to be changed together — the flag alone moves the listener and leaves the Service dialing a dead port, which fail-closed admission turns into a wedged cluster.

All three pages now agree with the operator section, which grew a **Serving on a different port** subsection with the patch that moves all three.

k8s-operator/cmd/main.go
bradhoekstra: Confirmed and fixed in b5683aa8. `sigs.k8s.io/controller-runtime@v0.19.0/pkg/webhook/server.go:139-140` is exactly what you quoted, and the `--metrics-bind-address=0` precedent is sitting three lines above in the same `args` block, so this was reachable.

Took your suggested guard, placed after `ctrl.SetLogger` so it logs through the configured logger rather than the deferred one:

```go
if webhookPort < 1 || webhookPort > 65535 {
    setupLog.Error(nil, "invalid --webhook-port: must be between 1 and 65535", "port", webhookPort)
    os.Exit(1)
}
```

Verified against the built binary:

```
$ ./manager --webhook-port=0
ERROR   setup   invalid --webhook-port: must be between 1 and 65535   {"port": 0}
$ echo $?
1
```

The flag help also no longer says "Override this on clusters where 10250 is not the reachable port" — same one-knob claim as the docs.

docs/site/src/content/docs/operator/index.md
bradhoekstra: You are right and the PR body was wrong. `targetPort` is a manifest field, so "nothing to do on the cluster itself" only holds for a `make deploy`, not for a `kubectl set image` or a tag-bump pipeline — and that failure mode is nastier than the original bug because it arrives on a routine version bump.

b5683aa8 adds an **Upgrading from an operator that served 9443** subsection covering both halves, including the window you flagged:

> Re-apply the manifests; do not bump the image alone. [...] `make deploy` applies both.
>
> Applying both together still leaves a short window: the Service starts sending traffic to 10250 the moment it is applied, and the old pod does not answer there. Any `PlatformAgent` write in the gap between the Service change and the new pod becoming Ready fails closed. It is seconds on a healthy rollout, but schedule the upgrade accordingly rather than alongside a `PlatformAgent` change.

I have corrected the rollout paragraph in the PR description to match.

toshiowang (approved):

PR #634, opened 2026-08-11
feat(memory): replace the file-based memory provider with Hindsight

dshnayder:
# Pull Request

## Why This Change

The Chat Agent's memory was a flat file. `multiuser_memory` kept shared knowledge in a single
`MEMORY.md` and per-user facts in `users/<id>.md`, and injected the whole thing into the system
prompt on every turn. That works at a dozen facts and stops working well before an organisation's
real backlog of decisions, ownership, and gotchas.

We measured where it stops. A synthetic 500-cluster fleet — 1,664 records, 1,414 of them shared —
put in front of the same agent twice, scored on the same 26 probes at five corpus sizes:

| Provider   | Rung | Gold recall | Contamination | Current ranked first | Context tokens |
| ---------- | ---: | ----------: | ------------: | -------------------: | -------------: |
| Hindsight  |  100 |       0.718 |         0.407 |                0.833 |          4,588 |
| Hindsight  | 1414 |       0.702 |         0.407 |                0.833 |      **4,264** |
| File-based |  100 |       1.000 |         0.722 |                0.429 |         13,780 |
| File-based | 1414 |       1.000 |         0.722 |                0.429 |    **110,907** |

The file provider's 1.000 gold recall is not a retrieval result — it is what "the whole store is
already in the window" scores by definition, and the cost of it is the last column: 110,907 tokens,
**1.66 windows of memory alone** against Opus 5's 200k, before a single skill body or tool schema is
loaded. Hindsight's cost is flat in corpus size because the store holds everything and the window
carries only what the turn needs. It also ranks the *current* version of a fact first in 83% of
probes where the file provider manages 43%, and contaminates answers with superseded records at
roughly half the rate.

Full argument, including what the 0.702 actually loses and why that is the right trade:
[`docs/designs/memory.md`](https://github.com/gke-labs/kube-agents/blob/main/docs/designs/memory.md).

## What Changed

**Files:**

- `agents/chat/plugins/memory/kube_agents_memory/` — the new provider (wrapper, not a fork)
- `agents/chat/defaults/hindsight/config.json`, `agents/platform/hindsight/config.json`
- `agents/platform/plugins/memory/multiuser_memory/` — removed
- `k8s-operator/config/integrations/hindsight/` — API, Postgres, PodMonitoring, NetworkPolicy
- `k8s-operator/scripts/provision_13_deploy_hindsight.sh`, `teardown_13_deploy_hindsight.sh`
- `deploy/shared/docker-entrypoint.sh`, `deploy/docker/Dockerfile`
- `agents/chat/scripts/memory_file_import.py`, `memory_ttl_curator.py`
- `docs/designs/memory.md`

**Added/Changed/Fixed:**

- **One bank, tag-scoped.** A slim wrapper around the stock Hindsight plugin pins a single bank
  (`kube-agents-memory`) and injects `user:<id>` on retain, `[user:<id>, scope:shared]` on recall.
  The ~92 KB upstream plugin is not forked. A hand-edited `bank_id` on the PVC is now logged and
  ignored rather than obeyed.
- **Hindsight installs like the other integrations** — kustomize base under
  `k8s-operator/config/integrations/`, a numbered provisioning step, a matching teardown, and a
  `PodMonitoring` so managed Prometheus scrapes it.
- **Connection config is image-owned.** `hindsight/config.json` is force-synced from the image by
  the entrypoint on every start, joining the files that already work that way. It was previously
  hand-written on the PVC and survived rollouts carrying whatever design was current when it was
  last touched — which is how a stale `bank_id` silently outlived the two-bank design it came from.
- **Existing file-based memory migrates at startup**, one record per retain call, leaving a receipt
  under `$HERMES_HOME/hindsight/imported/`.
- **Specialists get shared-scope memory, read-only.** They are spawned with no human identity, so
  they get `scope:shared` recall and no write path. Without it they build their own caches: one
  specialist had appended 79,815 bytes of corpus into its own `SKILL.md`, a stale unreviewable fork
  of the corpus that reloads into its context every turn.
- **A read reports its outcome** instead of letting "I could not reach it" become "it does not
  exist".

## Why This Matters

- Memory stops competing with the agent's actual work for context. 4,264 tokens instead of 110,907
  at the same corpus size, and flat as the corpus grows rather than linear.
- Superseded facts stop winning. Ranking the current version first went 43% → 83%.
- The memory backend becomes reviewable infrastructure — a kustomize base and a provisioning step —
  instead of a file somebody wrote onto a volume by hand.

## Context

Merges `main` (through 9396c01, 18 commits). Seven files conflicted; the merge commit records each
resolution. One semantic conflict had no textual overlap: the new zero-friction installer
(`install.sh`, #519) seeded `MEMORY_PROVIDER=multiuser_memory` into the state file, which would have
overridden the correct default in `provision_08_deploy_platform_agent.sh` and named a provider this
branch deletes.

**The experiment is not in this branch.** Per review feedback this pull request carries only design
and code. The 48 files and ~1 MB of corpus fixtures, job manifests and raw scorer output that
decided the provider live on
[`experiment/memory-scale-ab`](https://github.com/dshnayder/kube-agents/tree/experiment/memory-scale-ab/tests/memory-scale),
cut from the same commit as this change so the harness runs against exactly the code described here:

- [`tests/memory-scale/README.md`](https://github.com/dshnayder/kube-agents/blob/experiment/memory-scale-ab/tests/memory-scale/README.md)
  — the apparatus: corpus, rungs, how to re-run it
- [`transcript/README.md`](https://github.com/dshnayder/kube-agents/blob/experiment/memory-scale-ab/tests/memory-scale/transcript/README.md)
  — every probe verbatim, per-answer scoring, and the run's own corrections
- [`results/`](https://github.com/dshnayder/kube-agents/tree/experiment/memory-scale-ab/tests/memory-scale/results)
  — raw scorer output, one file per provider per rung; the source of every number in the table above
- [`harness/gen_fleet_corpus.py`](https://github.com/dshnayder/kube-agents/blob/experiment/memory-scale-ab/tests/memory-scale/harness/gen_fleet_corpus.py),
  [`queries.json`](https://github.com/dshnayder/kube-agents/blob/experiment/memory-scale-ab/tests/memory-scale/queries.json)
  — the fixed-seed generator and the 26 scored probes

`docs/designs/memory.md` stays in this branch and keeps every citation, now as full links to that
branch. The unit tests under `tests/memory/` also stay — they test this branch's code and do not
depend on the experiment.

Deliberately **not** in scope, listed so they are not mistaken for oversights:

- `charts/kube-agents/` ships no Hindsight template, so the Helm and Terraform
  `examples/full-install` paths install LiteLLM but no memory backend. The provisioning-script path
  is complete. Worth its own PR and its own decision about whether the chart should own a database.
- `hindsight-api` stays at one replica. Horizontal scaling is untested here.
- References to `multiuser_memory` that remain in `memory_file_import.py` and in the historical
  section of `docs/designs/memory.md` are deliberate: they name the on-disk format the importer
  reads and the provider this replaced.

## Testing

- **Fresh install on a brand-new GKE cluster**, provisioned from scratch with the branch's
  manifests: all six pods Running with zero restarts (`hindsight-api`, `hindsight-postgresql-0`,
  `kubeagents-controller-manager`, `litellm` ×2, `platform-agent-gateway` 4/4), `PlatformAgent`
  reaching `phase: Ready`. Images pinned to the merge-base tag so the install differed from a stock
  one only by this branch's manifests.
- **Managed Prometheus verified end to end**, not by manifest: Cloud Monitoring returns
  `up{job="hindsight-monitoring"} = 1` and live `hindsight_http_requests_total` series, against
  exactly one target — the API pod, not Postgres, which is what the `component: api` selector is
  for.
- **The A/B itself** is reproducible from the archive branch (below): fixed-seed corpus generator
  (`20260731`, recorded in `manifest.json`), the 26 scored probes, the Jobs that seeded and audited
  against a live cluster, and per-provider-per-rung scorer output. The transcript records the run's
  own corrections, including a seeding defect that inflated a bank to 1,959 units and the reseed
  that fixed it.
- `go build`, `go vet`, `go test ./...` pass in `k8s-operator/`; golden manifests regenerated.
- `make docs-check` passes; `prettier --check` clean over every Markdown/JSON/YAML file the branch
  changes; the docs site builds.
- The `review-docs-drift` skill was run against the branch diff. It found four blocking items — the
  installer's provider default, the map's stated document total, an unregistered identifier-source
  category, and the stale golden hashes — all fixed in this branch.

---

Rollout is not automatic: the memory backend only appears once provisioning step 13 runs, and
existing installs keep their file store until the importer runs at startup and leaves its receipt.
The riskiest single change is the entrypoint force-sync, because it makes the image authoritative
over a file operators may have hand-edited; that is the intent, and the previous behaviour is the
defect it fixes.

This pull request was generated with the help of Claude.


agents/chat/defaults/hindsight/config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "mode": "local_external",
+  "api_url": "http://hindsight-api.kubeagents-system.svc.cluster.local:8888",
```
fatoshoti: IMO this should be configurable and passed from the controller

agents/platform/config.yaml
```diff
@@ -83,14 +102,38 @@ toolsets:
...
+#
+# `memory_enabled: false` is not a contradiction of the provider being on: it
+# gates only the built-in file store, and the provider loads off `provider`
+# alone. See the note on the toolset gate above.
+#
+# See docs/designs/memory.md and agents/chat/config.yaml (the front door, where
+# writes do happen).
 memory:
   memory_enabled: false
+  provider: kube_agents_memory
```
fatoshoti: should this read from env variables?

fatoshoti (commented):
Is the intention for this option to be the only one? Should we keep the current option as the "lightweight" option

agents/chat/defaults/hindsight/config.json
dshnayder: Agreed — done in fd8f086.

The operator now derives it from the agent's namespace and passes it as an environment variable, right beside where it already derives the LiteLLM endpoint the same way:

```go
cfg.Model.BaseURL = fmt.Sprintf("http://litellm.%s.svc.cluster.local/v1", agent.Namespace)
...
Value: fmt.Sprintf("http://hindsight-api.%s.svc.cluster.local:8888", agent.Namespace)
```

One subtlety worth recording, because it makes the obvious version of this fix a no-op: the plugin resolves the endpoint as `config.get("api_url") or os.environ.get("HINDSIGHT_API_URL", default)`, so the file outranks the environment. Injecting the variable while leaving `api_url` in the shipped config would have been silently ignored — the same class of failure as the stale `bank_id` the entrypoint force-sync exists to prevent. So the key is now absent from both `hindsight/config.json` files, which keep only what is identical in every install (`mode`, `memory_mode`, `recall_budget`). An operator who writes `api_url` into the file still wins, deliberately.

`memory_file_import.py` and `memory_ttl_curator.py` read it with the same precedence. Design doc updated (`docs/designs/memory.md`, "Where the connection settings come from"), goldens regenerated.

agents/platform/config.yaml
dshnayder: The endpoint now does — see the thread on `hindsight/config.json` (fd8f086). The provider name here deliberately does not, for two reasons.

This file is the platform specialist's profile config, and unlike the default profile's it is not operator-rendered: `renderProfileOverlayYAML` carries only `plugins.enabled` and the allowlisted subtrees, so nothing merges a `memory` block into it. Making the name env-driven would mean the profile silently loads whatever provider the environment names, with no manifest showing which.

And the name is not an install-time variable — it is the provider bundled in this image at `agents/chat/plugins/memory/kube_agents_memory/`. The value that genuinely varies per install was the endpoint, which is exactly what moved.

The default profile does take it from the CR: `spec.harness.memory.provider` overrides the operator's default in `renderConfigYAML`. Happy to extend that to the specialist profiles if you think it should be settable there too — it just needs a real channel rather than an env var.

agents/chat/plugins/memory/kube_agents_memory/__init__.py
mastersingh24, line 1: Is there a reason you did not break this into a few smaller files?  Seems like that's the pattern the built-in plugins use?

mastersingh24 (commented):
**TL;DR:** I’m pushing back on committing to Hindsight for our memory substrate without a proper comparative benchmark against Honcho. While the engineering is clean, deploying a Postgres StatefulSet, a dedicated API pod, and an 882-line wrapper is an unnecessary architectural risk. This isn't about finding a lighter operational footprint—it's a deliberate trade-off of native tenancy correctness and long-term code maintainability versus service complexity.

### Core Issues
* **False Dilemma:** The current design doc (`docs/designs/memory.md`) compares Hindsight only to the legacy flat-file `multiuser_memory`. Flat-file was a temporary stopgap; beating it on token cost is a low bar that doesn't justify our provider choice.
* **Artificial Complexity:** The 882-line wrapper exists solely to emulate multi-tenancy on Hindsight. Because Hindsight's recall is hard-scoped to a single `bank_id`, private facts live in the same bank and are separated only by filters rather than true tenancy boundaries.
* **Single-Replica Risk:** `hindsight-api` is single-replica, and horizontal scaling is untested. Operating an unproven database as core infra is an operational liability we shouldn't accept blindly.

### Architectural & Operational Trade-Offs

| Dimension | Hindsight (Current PR) | Honcho (Alternative) |
|---|---|---|
| **Tenancy Boundary** | **Emulated:** Filters inside a shared bank | **Native:** Workspaces (org) & Peers (per-user) |
| **Wrapper Complexity** | ~882 lines of load-bearing custom wrapper | **Zero wrapper:** Native multi-peer sessions |
| **Recall Model** | Knowledge-graph / entity recall | **Dialectic:** Per-peer representations & cards |
| **Infra Footprint** | **Lighter:** Postgres + API pod (single-replica) | **Heavier:** Postgres/pgvector + Redis + Deriver + LLM |

### Actionable Next Steps

1. **Run the Comparative Benchmark:** Put Honcho through our exact 26-probe testing harness.
   * **If Hindsight wins on our fleet:** Merge the PR. We’ll have the hard data to justify the wrapper complexity and operational footprint.
   * **If Honcho wins:** Adopt Honcho and delete the ~882-line wrapper (trading wrapper complexity for a larger service count).
2. **Immediate Code Hygiene:** **Regardless of the provider we choose, keeping 882 lines of code in `\__init\__.py` is a bad pattern. Refactor it to keep `\__init\__.py` purely as the entry point, and split the implementation into `client.py`, `session.py`, and `config_schema.py` (matching Honcho's bundled plugin pattern).**

agents/platform/config.yaml
dshnayder: Good question — it would work mechanically, but it would be the wrong value here.

Hermes does expand `${VAR}` in config values (`_expand_env_vars` in `hermes_cli/config.py`), and the pod does carry `MEMORY_PROVIDER`. Two reasons not to use it in this particular file:

**1. The specialists must not get the install's provider verbatim.** `MEMORY_PROVIDER` holds the Chat Agent's choice. This profile needs that choice *filtered*: a specialist only gets a provider that can be made read-only and scoped by tag. A per-user file provider like `multiuser_memory` keys its store off the gateway identity, and a kanban-spawned specialist has none, so every write would land in one anonymous bucket. `memoryOverlay` in `k8s-operator/internal/controller/platformagent_manifests.go` blanks the key for anything that isn't Hindsight-backed. An `${MEMORY_PROVIDER}` here would silently defeat that — it is a different value, not the same value by another route.

**2. Unresolved references are kept verbatim.** With no operator in play, `${MEMORY_PROVIDER}` would be passed through as a literal provider name and fail obscurely. The literal in the file is deliberately the no-operator fallback — `docker run` and local dev get a working default rather than a broken string.

Worth saying that this line isn't the live value on a deployed pod either way: the operator overwrites the whole `memory` subtree in the rendered ConfigMap from `spec.harness.memory.provider`, so the CR is already the single source of truth. Env expansion would add a second path to the same setting, and the two would disagree in exactly the case above.

The env var exists for a narrower consumer: the entrypoint gates the one-way file-store import on it (`memory_import_wanted` in `deploy/shared/docker-entrypoint.sh`), and that runs before `config.yaml` is in play.

I've expanded the comment above the block to say which of these values the operator owns and which is the fallback. Happy to reconsider if you'd rather the fallback be expressed some other way.

dshnayder: To be precise about what changed, since the paragraph I pointed at was not in the commit you reviewed:

No code change here — the answer above is "the literal is deliberate". What did change is the explanatory comment above the block, which landed in [`868caf6`](https://github.com/gke-labs/kube-agents/commit/868caf6) (pushed after `8a1656c`, so it is not in the diff you were reading). It now states which of these keys the operator owns and which is the no-operator fallback:

```yaml
# `provider` here is a default, not the decision. The operator overwrites it from
# spec.harness.memory.provider through this profile's overlay (memoryOverlay in
# k8s-operator/internal/controller/platformagent_manifests.go), because the choice
# belongs to the install and this file is baked into the image. It writes an empty
# value for any provider that cannot be made read-only and scoped by tag — a
# per-user file provider has no identity to key on here. The value below is what a
# run without the operator gets.
```

That commit also carries a separate change worth a look while you are in this file: memory is now **on** by default (`--memory=hindsight`), and `MEMORY_ENABLED` has been split from the CR's `memoryEnabled` — they were one variable answering two questions, and provisioning step 8 now derives the narrow one from the broad one.

agents/chat/plugins/memory/kube_agents_memory/__init__.py
dshnayder: I went and measured the built-in plugins before answering, and the pattern is more mixed than it looks — five of the eight are a single `__init__.py`, including the one this wraps:

| Plugin | `.py` files | `__init__.py` |
| --- | --- | --- |
| `openviking` | 1 | 3,725 |
| `hindsight` | 2 | 1,968 (+76-line `config_schema.py`) |
| `supermemory` | 1 | 1,044 |
| `retaindb` | 1 | 771 |
| `byterover` | 1 | 449 |
| `honcho` | 7 | 1,561 |
| `mem0` | 4 | 627 |
| `holographic` | 4 | 419 |
| **`kube_agents_memory`** | **1** | **882** |

The ones that split did it around a separable concern rather than around size — `honcho` has an OAuth flow and a CLI, `mem0` has a setup wizard, `holographic` has its own store and retrieval engines. Ours has no equivalent: it is one `MemoryProvider` subclass whose job is to resolve an identity and pin four upstream settings, and every method is on the isolation path. Splitting the class across files would put the tag construction in one file and the tag matcher that consumes it in another, which is where I'd expect a leak to eventually hide.

The one seam that does exist is the ~200 lines of prompt and notice constants at the top (`RECALL_PREAMBLE`, `NO_MATCH_GUIDANCE`, `SYSTEM_PROMPT_*`, the retain missions) — those are content, not logic, and would sit fine in a `prompts.py`. **Happy to do that if you'd like it**; I left them inline because each one is read together with the method that emits it, but it is a genuine 23% of the file and I don't feel strongly.

You did surface a real gap in the same breath, though: every built-in plugin ships a `README.md` and this one didn't — even `multiuser_memory` grew one in this branch. Added in [`d203584`](https://github.com/gke-labs/kube-agents/commit/d203584). It is deliberately short, since the four pinned settings and the upstream code that forces each of them are already in the module docstring, and the design lives in `docs/designs/memory.md`.

k8s-operator/scripts/provision_08_deploy_platform_agent.sh
```diff
@@ -44,7 +44,7 @@ DEFAULT_AGENT_IMAGE="$(registry_prefix)/platform-agent"
 init_var "AGENT_IMAGE" "$DEFAULT_AGENT_IMAGE" "Enter Platform Agent Image Path"
 warn_on_registry_prefix_mismatch "AGENT_IMAGE"
 init_var "MEMORY_ENABLED" "false" "Enable agent memory persistence? (true/false)"
-init_var "MEMORY_PROVIDER" "multiuser_memory" "Enter agent memory provider"
+init_var "MEMORY_PROVIDER" "kube_agents_memory" "Enter agent memory provider"
 init_var "USER_PROFILE_ENABLED" "false" "Enable per-user memory profiling? (true/false)"
```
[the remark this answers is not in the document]
dshnayder: Two halves to this, and the first premise no longer holds.

**`multiuser_memory` is not deleted.** It moved from `agents/platform/plugins/memory/` to `agents/chat/plugins/memory/`, which is where the profile that uses it lives. `git diff upstream/main --stat` shows `__init__.py` and `plugin.yaml` at `| 0` — pure renames; only the README was rewritten, which is why GitHub renders the pair as a delete plus an add rather than a move. The Dockerfile copies both `agents/platform/plugins/` and `agents/chat/plugins/` into `/opt/hermes/plugins/`, so a persisted `MEMORY_PROVIDER="multiuser_memory"` resolves exactly as it did before. That was settled before this review; it is restored as of `9fa85ee`.

**The `init_var` observation was right about something else, though, and it caught a live regression.** You are correct that `load_state` sources the saved `vars.sh` before `init_var` runs, so an existing install keeps whatever it persisted. Following that through: every install this repository has ever written persisted `MEMORY_ENABLED="false"` *and* `MEMORY_PROVIDER="multiuser_memory"`, because the provider loads off `provider` alone and the built-in-store flag was never part of that decision. The commit under review had started deriving the provider from `MEMORY_ENABLED` — forcing it to `none` when the flag was false. On every existing install that would have read a working file store as "this install wants no memory" and switched it off, stranding each user's `USER.md`.

Fixed in `722ee80` by removing the derivation rather than patching around it:

- `MEMORY_ENABLED` goes back to meaning only what it meant on `main` — Hermes' built-in `MEMORY.md`/`USER.md` store. Step 8 copies it into `spec.harness.memory.memoryEnabled` unchanged, and `platform-agent.yaml.template` is now byte-identical to `main`.
- Step 13 gates on `MEMORY_PROVIDER` alone: `kube_agents_memory` or `hindsight` deploys the API and Postgres, anything else exits 0. Nothing reads `MEMORY_ENABLED` to decide whether memory exists, so a false left in an old `vars.sh` cannot be misread on upgrade. "No memory at all" is spelled `MEMORY_PROVIDER=none`.

**On the entrypoint importer** — also correct, and also already fixed, in `fd8f086`. Step 5.6 no longer keys off `hindsight/config.json` existing (an image-owned file, therefore always present, so an install that had deliberately kept the file store had it taken away on the next roll). It reads `MEMORY_PROVIDER` from the pod and runs only for a Hindsight-backed one. An *unset* variable — an operator too old to send it — still takes the old file-presence path, which is the safe reading for an install predating the choice; `${VAR+x}` rather than `-z` is what keeps "set to empty" and "unset" apart there.

What remains true from your finding and is deliberate: nothing rewrites a stale `MEMORY_PROVIDER` or a stale CR field on upgrade. An install that chose the file store keeps it, which is the intended behaviour — the new default applies to new installs, and switching an existing one is `--memory=` on a re-run.

agents/chat/plugins/memory/kube_agents_memory/__init__.py
```diff
@@ -0,0 +1,882 @@
...
+
+    Mirrors Hindsight's own ``_sanitize_bank_segment``: alphanumerics, dash and
+    underscore survive, everything else collapses to a single dash. Applied for
+    the same reason it is applied to bank names — the value is attacker-adjacent
+    (it comes from the chat platform) and ends up in a query filter.
+    """
+    cleaned = re.sub(r"[^A-Za-z0-9_-]+", "-", str(user_id or ""))
+    return re.sub(r"-{2,}", "-", cleaned).strip("-_")
+
+
```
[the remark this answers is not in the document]
dshnayder: Agreed, and fixed in `722ee80`. This was a real two-way leak: colliding identities would read each other's private facts on recall and retain into each other's scope at session end, and the tag is the only thing narrowing a single shared bank.

`_sanitize_user_id` now produces `<readable>_<sha256(raw)[:12]>` — the same guard `multiuser_memory` applies to its filenames, which is exactly the precision the importer goes to the trouble of recovering:

```python
raw = str(user_id or "").strip()
if not raw:
    return ""
cleaned = re.sub(r"[^A-Za-z0-9_-]+", "-", raw)
cleaned = re.sub(r"-{2,}", "-", cleaned).strip("-_")
digest = hashlib.sha256(raw.encode("utf-8")).hexdigest()[:12]
return f"{cleaned}_{digest}" if cleaned else digest
```

Three details worth naming:

- **Empty stays empty**, deliberately not a hash of the empty string. `initialize` reads a falsy result as "no identity" and sets `NO_IDENTITY_NOTICE`, refusing personal memory entirely; a digest there would have turned "nobody" into a valid-looking user and quietly re-opened the hole from the other side.
- **The readable half stays first**, so a tag is still recognisable when reading the bank. An identity that sanitizes to nothing at all (`@@@`) falls back to the bare digest rather than a leading underscore.
- **`.strip()` before hashing**, so a padded copy of an id is the same person rather than a second one.

`memory_file_import.sanitize_user_id` is a mirror of this function — it files migrated entries under the tag the provider later reads them back with — so it moved in the same commit. `tests/memory/test_user_tag_isolation.py` locks down the guard, the empty-identity path through `initialize`, and the parity between the two copies, since a drift between them would strand every migrated memory silently.

One consequence to flag rather than bury: the tag format changed, so anything already stored under the old lossy form is unreachable to its owner. `kube_agents_memory` is new in this PR and no release ships it, so the only affected store is my own test bank on `kage-management`, which I will reseed.

docs/site/src/content/docs/reference/config.md
```diff
@@ -101,7 +101,9 @@ A second, top-level gate distinct from `platform_toolsets`: listing `kanban` her
...
+No memory provider is configured either. Per-user memory scopes its store by the sender's gateway identity, and the Platform Agent is reached through the kanban dispatcher, which spawns workers with no human identity attached — so per-user memory only makes sense on the [Chat Agent](/kube-agents/concepts/chatops/), the profile that actually receives chat ingress. The Chat Agent records each user's durable facts and resolves them into concrete values before delegating, so the Platform Agent gets what it needs inline in the card body.
+
+How that profile keeps every user's memory apart — one Hindsight bank, scope tags, and the upstream defaults it has to override — is documented in [`docs/designs/memory.md`](https://github.com/gke-labs/kube-agents/blob/main/docs/designs/memory.md), along with the retention mechanism and the scale test that chose a document store over a flat file.
```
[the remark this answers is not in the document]
dshnayder: Correct at `8a1656c`, and fixed in `868caf6` — which landed after this review was cut, so the page you read was genuinely stale at the time.

`docs/site/src/content/docs/reference/config.md` on the current branch:

- The **`## Full file`** block lists `memory` under both `platform_toolsets.cli` and `platform_toolsets.api_server`, and the `memory:` stanza now shows all four keys — `memory_enabled: false`, `provider: kube_agents_memory`, `read_only: true`, `user_profile_enabled: false` — with the comment that `provider` here is only the no-operator default.
- The **`### memory`** section no longer says a provider is absent. It opens by separating the two mechanisms (`memory_enabled` / `user_profile_enabled` gate Hermes' built-in file store, not the provider), then states that personal memory is impossible on this profile and always will be — it keys off the gateway identity and the Platform Agent is reached through the kanban dispatcher, which spawns workers with no human attached — and that shared memory *is* readable, with `read_only: true` keeping it one-way across all three enforcement points.
- **`### platform_toolsets`** explains why `memory` has to appear on both keys: it is a gate rather than a tool grant, and a specialist reached as a kanban worker resolves from `cli` while the API-server path uses `api_server`.

Rationale is not duplicated here — the page links to `docs/designs/memory.md`, which owns it.

k8s-operator/config/integrations/hindsight/README.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+That URL is baked into `agents/chat/defaults/hindsight/config.json`, which the
+container entrypoint force-syncs onto the agent's PVC on every start. The
+manifests themselves are namespace-agnostic — every in-cluster address in them is
+a short service name, and the provisioning step applies them with `-n $NAMESPACE`
+— but the agent image is not. Installing into a different namespace means editing
+that file and rebuilding the agent image; a hand-edit on the PVC is overwritten
+on the next roll.
+
```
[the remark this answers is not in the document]
dshnayder: Fixed in 9b5f619 — you are right on both counts, and the README paragraph was the older of the two errors.

`agents/chat/defaults/hindsight/config.json` carries `mode`, `memory_mode` and `recall_budget` and no `api_url`, and the operator passes `HINDSIGHT_API_URL` derived from the agent's own namespace. The README now says so, and adds the reason the key is deliberately absent rather than merely unset: the plugin prefers the file over the environment, so a value left there would silently outrank the operator's. The kustomization comment now makes the same point about the agent side instead of implying a rebuild.

docs/designs/memory.md
```diff
@@ -0,0 +1,1463 @@
...
+
+Three things then read the choice rather than assuming it:
+
+- **Provisioning step 13 deploys nothing** unless memory is enabled _and_ the
+  provider is Hindsight-backed. Both halves matter: the provider keeps its default
+  name when memory is switched off, so the name alone would have stood up a
+  Postgres database for a memory nobody turned on. It exits 0 — for those settings
+  "nothing to deploy" is the step succeeding, and turning memory on later is a
+  matter of re-running it.
+- **The specialist profiles get a provider only if it is Hindsight-backed**, via
```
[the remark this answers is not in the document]
dshnayder: Fixed in 9b5f619. The bullet now reads "Provisioning step 13 deploys nothing unless the provider is Hindsight-backed. That is the whole gate — `multiuser_memory` and `none` stand up no database."

Your second point is the one I had missed: the justification was not just describing a gate that no longer exists, it was false on its own terms. `install.sh` maps `off` to the provider name `none`, so the provider does not keep its default name when memory is switched off — which is exactly why the provider alone is sufficient. Dropped the sentence rather than rewording it. The trailing "gated the same way" bullet now inherits the corrected statement, and `memory_import_wanted()` is provider-only, so it is accurate.

agents/chat/scripts/memory_ttl_curator.py
```diff
@@ -0,0 +1,500 @@
...
+Checkpoints are written under the `checkpoint` retain strategy, which
+`kube_agents_memory` provisions on the bank and which pins extraction to
+`verbatim`. That matters: Hindsight's default extraction would re-summarise the
+observation, and re-summarising a summary every cycle is a game of telephone that
+walks the bank away from what was actually said. In `verbatim` mode
+`_collapse_to_verbatim` overwrites the fact text with the raw chunk, so one
+checkpoint in is one fact out, unchanged. The LLM still runs — it attaches
+entities and dates — so this is not a free operation and the run is paced
+accordingly.
+
```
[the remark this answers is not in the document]
dshnayder: Fixed in 9b5f619. `CHECKPOINT_STRATEGY` pins `retain_extraction_mode: "chunks"` and the prose now matches it in all three places: the strategy table, the distill-then-retire paragraph, and the curator's module docstring.

The docstring needed more than a word swap, because the paragraph after it explained why the LLM still runs under `verbatim` — and under `chunks` no LLM runs at all. It now says why `verbatim` was the obvious choice and the wrong one (it re-emits the observation inside a JSON response schema, which failed on 5% of observations at scale and aborted every pass at the safety check), and what `chunks` costs instead: no extracted entities, so the graph retriever cannot see checkpoints. The one-item-per-call comment lost its extraction-LLM rationale too, since the reason that survives is the endpoint being neither atomic nor idempotent. The findings themselves were already recorded correctly further down the design doc — it was only the forward-looking prose that had not caught up.

dshnayder:
@mastersingh24 

**On the false dilemma: you're right, and it's fixed.** The design doc compared Hindsight only to `multiuser_memory`, and Honcho was one table row that said "Hosted" — which is simply wrong, since the bundled Hermes plugin takes a `baseUrl` and Honcho self-hosts as a container set. It was the only candidate the comparison never actually argued against. `0cf1902` and `afce506` rewrite that section.

The elimination stands, but on different grounds than the doc gave. Two of them are hard, and one of them shapes the benchmark itself.

### 1. No organisation-wide read — and this is where the wrapper comes back

This is the point that matters most, because it is the problem the PR exists to solve: a specialist agent spawned with **no human identity** must read the org corpus — fleet topology, ownership, incident history — while never seeing a private fact.

In Honcho every read is perspective-scoped. `session.py` searches with `filters={"peer_perspective": peer_id}`, and `_conclusions_scope` is dyadic, `observer.conclusions_of(target_peer_id)`. Peers and workspaces are genuinely first-class and the plugin resolves the gateway identity to a peer natively — I'll concede that half of your table outright.

But "zero wrapper" only holds for the per-user half. **Per-user memory _and_ an org-wide corpus is the requirement, and the second half has no native form in a peer-perspective model.** There is no workspace-level corpus a peer-less reader can query. Getting one means inventing a synthetic organisation peer that every real peer observes and that specialists read _as_, plus the routing code to keep each write landing on the right one of the two. That is a wrapper by any definition — sitting on top of a plugin that is already 7,096 lines — and a stranger one than tags, because it encodes org knowledge as one fictional peer's beliefs inside a model built for beliefs _about_ peers. The choice isn't wrapper vs. no wrapper; it's which wrapper, and ours is the smaller and more literal of the two.

### 2. The embedder is a cost for us and a constraint on the customer

`src/config.py`: `EmbeddingTransport = Literal["openai", "gemini"]`. No in-process embedder, and `EMBED_MESSAGES=false` doesn't degrade gracefully — `src/utils/search.py` skips the semantic branch entirely and leaves keyword matching.

For us that means adding a second model, an embedding model, to LiteLLM. Today LiteLLM carries only the model the installer defines, so that's a new install-time input, a new key to hold, and a new thing to keep in sync per environment — and it hard-wires the provider choice, since only two are on offer.

For customers it's the heavier half, and it's the part I'd weigh most. The transport list admits an OpenAI-compatible `base_url` override but nothing outside those two wire formats. **A customer standardised on Anthropic has no embedding API to point at at all.** One on Bedrock, Azure, Vertex-native or an in-house embedder has to stand up a compatible shim before Honcho will index anything. We would be telling every such install that our memory layer requires an OpenAI- or Google-compatible embedding endpoint specifically. Hindsight embeds and reranks in-pod from models baked into the image (`HF_HUB_OFFLINE=1`) and asks LiteLLM for exactly one model — the installer's. The customer's choice of provider is a question the memory store never asks.

### On the benchmark

I'd rather agree the shape before running it than after. Standing Honcho up to score 26 probes means paying for both of the above first: an embedding model wired in somewhere, and the org-peer shim from point 1 — because without the shim the shared-scope half of the probe set scores zero by construction and the run measures nothing. So the benchmark can't adjudicate between the two on its own; it presupposes that the shim exists, which is the thing in dispute.

Scoped in full, in case you want the number regardless:

| Step                                      | Cost                                                          | Note                                                                                                                                                        |
| ----------------------------------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Honcho self-hosted in a scratch namespace | 4 workloads — api, deriver, `pgvector/pgvector:pg15`, `redis:8.2` | No manifests exist; the repo has no Honcho chart                                                                                                            |
| Embedding model                           | new install-time dependency                                   | A `text-embedding-*` entry in LiteLLM plus a key, or a direct provider call bypassing it and losing cost attribution. Only OpenAI- or Gemini-compatible endpoints qualify |
| Org-peer shim                             | unknown, and the real risk                                    | Without it the shared-scope probes score 0 by construction                                                                                                  |
| Re-seed the 300-fact corpus               | LLM-per-message via the deriver                               | Hindsight's `chunks` seed was under 100s; the LLM-backed path was 55 minutes                                                                                |
| 26 probes × 2 rounds + scoring            | roughly the Round A/B effort already spent                    | The harness is reusable; the scoring rubric is not Honcho-shaped                                                                                            |

Happy to run it — I'd just want the embedding dependency and the shim to be a deliberate call rather than a side effect of starting.

### 3. On code volume

The bundled Honcho plugin is 7,096 lines across seven files; `__init__.py` alone is 1,561 — larger than our entire wrapper. The honest framing is that it's 7,096 lines _we don't maintain_, which is worth something real. But it isn't a code-volume win, and the shim in point 1 lands on top of it.

### 4. Licence and footprint

Self-hosted Honcho is four workloads against Hindsight's two — your table has this right. Of those four, the two that are Honcho's own code — **the API and the background deriver, both built from the `plastic-labs/honcho` tree — are AGPL-3.0**; the other two are the stock `pgvector/pgvector:pg15` and `redis:8.2` images. Hindsight is MIT. A copyleft licence on a service the customer runs inside their own boundary is a question their legal review will ask, and it lands on the two components that are the product.

### On the single-replica risk

Conceded, no argument. `hindsight-api` is one replica, horizontal scaling is untested, and the PR body says so rather than claiming otherwise. It's a known gap, not a resolved one.

### On code hygiene

Agreed, and I'd already offered it upthread. The file is 907 lines, of which roughly 502 are code and 308 are docstrings and comments. I'll split it, following the `client.py` / `session.py` / `config_schema.py` shape you named, and report back on this thread when it lands.

dshnayder:
Reporting back on the code-hygiene point, as promised: the split has landed in `0bc49e2`.

`__init__.py` is now the entry point and nothing else — the design docstring, the re-exports, `__all__`, and `register()`. The implementation is four modules, split along the lines of what would break if each one changed, following the `client.py` / `session.py` / `config_schema.py` shape you named:

| Module             | Lines | Holds                                                                    | A change here is                     |
| ------------------ | ----: | ------------------------------------------------------------------------ | ------------------------------------ |
| `config_schema.py` |   157 | Bank name, the two scope tags, retain strategies and missions, config reads | a data-migration question            |
| `prompts.py`       |   215 | System prompt variants, tool schemas, the guidance strings a result carries | an agent-behaviour question          |
| `client.py`        |   249 | The only code that touches the stock provider or its generated client    | where a Hermes base-image bump lands |
| `session.py`       |   354 | The provider class: identity, the three session states, tool dispatch    | the wrapper's own logic              |
| `__init__.py`      |   160 | Docstring, re-exports, `register()`                                      | —                                    |

No behaviour changes. The one thing that is not a pure move: the `read_only` and no-identity branches of `apply_scoping` were three branches producing two outcomes, and are now one `if`/`else`. Nine helpers became public because they cross a module boundary now — `sanitize_user_id`, `memory_is_read_only`, `thread_sessions_are_per_user`, `SCOPES`, `VALID_BUDGETS`, and `client.load_hindsight` / `apply_scoping` / `ensure_bank` / `apply_budget`.

One thing the split introduced that a single file could not have: relative imports, and Hermes does not `import` a plugin. It loads it by path — `spec_from_file_location(name, __init__.py, submodule_search_locations=[dir])` — then walks `dir.glob("*.py")` and execs each submodule before the package itself, catching and debug-logging anything that raises. So load order is filesystem order, and a broken import would not surface as a failed load; it would surface as a provider that registered and then misbehaved. `tests/memory/test_plugin_loads_as_a_package.py` replays that exact mechanism under all 24 permutations of the four submodules and asserts `register()` still returns a working provider.

Everything else passes unchanged: `test_user_tag_isolation.py`, `test_recall_reporting.py` (its source scan now walks every module, not just the entry point), `test_read_only_profile.py`, and `make docs-check`.

dshnayder:
@mastersingh24 — on the third core issue, **"single-replica, and horizontal scaling is untested"**: fair as written, so I tested it. It scales.

Measured on the live cluster against the real 4,427-unit bank at `budget: mid`, driven from an in-cluster Job so kube-proxy is actually in the path — a `kubectl port-forward` binds one endpoint for the life of the tunnel and would have measured a single pod twice. Twelve recalls per concurrency level, 72 requests per configuration, zero errors:

| Concurrent recalls | 1 replica | 2 replicas    | p50: 1 → 2 replicas |
| ------------------ | --------- | ------------- | ------------------- |
| 1                  | 0.073 rps | 0.076 rps     | 13.6s → 13.2s       |
| 2                  | 0.087 rps | 0.096 rps     | 22.7s → 20.5s       |
| 4                  | 0.089 rps | **0.162 rps** | 44.1s → **22.8s**   |

Same twelve requests at concurrency 4: 134.8s → 73.9s wall. Aggregate CPU across the pods went 1,830m → 3,600m, so the second node's cores are doing work rather than idling behind the scheduler.

The baseline row is the more interesting half. At one replica, throughput is flat at ~0.09 rps from concurrency 2 upward while latency rises linearly — one pod is saturated by roughly *one* concurrent recall, because a recall is 300 cross-encoder forward passes on CPU inside the request. So the single-replica risk you named was real, and larger than "no redundancy": it was a hard concurrency ceiling of about one user at a time.

**Two replicas is also the fix for a null result already in the doc.** Raising the pod's CPU limit from 500m/2 to 2/4 had previously changed nothing, because an `e2-standard-4` is two physical cores behind four hyperthreads and the vector units are what saturate. A pod on a *different node* is the only way to add cores the reranker can use. That makes required pod anti-affinity load-bearing rather than hygiene — two co-scheduled replicas reproduce the null result exactly.

**The topology is supported, not tolerated.** The API runs the worker in-process (`WORKER_ENABLED` defaults true), workers claim tasks with `SELECT … SKIP LOCKED` (`worker/poller.py`), startup migrations take a `pg_advisory_lock` (`migrations.py`), worker identity is the hostname, and the pod mounts no volumes. Replicas neither double-run consolidation nor race Alembic.

**One real gotcha the test surfaced.** `ankane/pgvector` ships stock `max_connections = 100`, while `HINDSIGHT_API_DB_POOL_MAX_SIZE` defaults to 100 *per pod* — one replica already sits at parity with the server, so a naive `kubectl scale` can exhaust it under load. The runs above pinned it to 40. Worth knowing before anyone scales this in anger.

Three honest limits on the claim:

- **It buys concurrency, not speed.** Single-user latency is unchanged, 13.2s vs 13.6s. Someone waiting on one recall waits exactly as long.
- **The concurrency-2 row is not evidence of much.** kube-proxy balances per connection at random, so with two requests in flight both hit the same pod about half the time; that predicts ~20.3s against the 20.5s measured. Only the concurrency-4 row is outside that noise. And it's one run per configuration — read 82% as the size of the effect, not a figure to three digits.
- **This tests the API tier, not the database tier.** Postgres is still a single StatefulSet and still a single point of failure. Your "unproven database as core infra" concern is half-answered, not answered: the stateless tier scales horizontally and is now measured; Postgres HA is a separate piece of work this PR doesn't do.

Written up in `docs/designs/memory.md` under "A second pod buys throughput, not speed" (`4b8843d`), including the parts that don't flatter it.

`api.yaml` still ships `replicas: 1` deliberately — two replicas double an install's baseline CPU request to four cores and require two schedulable nodes, which a small install won't have. The point is that scaling is now a known-good one-line change with a measured payoff and a documented prerequisite, rather than an open question.

dshnayder:
Correcting my own last bullet — "Postgres is still a single point of failure" was too generous a concession, and more to the point it doesn't differentiate anything.

**Honcho has the same single Postgres.** Its `docker-compose.yml.example` runs one `database: pgvector/pgvector:pg15` behind one `pgdata` volume. Switching providers buys exactly zero database availability. If anything it's the wrong direction: Honcho also runs `redis:8.2` with its own `redis-data` volume, so that's two stateful services to keep up rather than one.

**And it wasn't a bottleneck.** Postgres sat at 12–19m of CPU against a 1 CPU limit through both runs, while the API pod pinned 1.8–2.7 cores. The load is entirely CPU-bound cross-encoder inference in the API tier — which is precisely the tier that scales horizontally, and the reason the second replica returned 82% of a theoretical doubling.

So what actually remains is **availability of a single RDBMS instance**, which is generic to any Postgres-backed system and is answered with ordinary Postgres HA — CloudNativePG, Cloud SQL with a standby, or a streaming replica with failover. That is orthogonal to the memory provider and not an argument for or against either candidate.

One detail worth having, since it cuts both ways: Honcho's own reference compose ships `command: ["postgres", "-c", "max_connections=200"]`. They already raise the stock default. That confirms the connection ceiling I flagged is a config line every Postgres-backed deployment of this shape needs, not an architectural property of Hindsight.

I've added this to the Honcho section of `docs/designs/memory.md` rather than leaving it only in this thread.

dshnayder:
@mastersingh24 — upthread I said I'd want the embedding dependency and the org-peer shim agreed before running the benchmark, rather than have them become side effects of starting. Fair enough to make that a deliberate call and then get on with it: I built both and ran the thing. This also answers @bradhoekstra's `/hold for Honcho comparison`.

Honcho, self-hosted, through the exact 26-probe harness — same 1,664-document corpus, same generator seed, same scoring code, all five rungs (100 → 1,414). Context budget matched by construction, since Honcho returns whole messages where Hindsight returns extracted observations; the evaluator spends an 18k-character budget in rank order, sized to Hindsight's measured 4,264–4,588 tok/turn. Matching on result count instead would have handed Honcho several times the context and made every other number meaningless.

- **[Results](https://github.com/dshnayder/kube-agents/blob/experiment/memory-honcho-ab/tests/memory-scale/honcho/RESULTS.md)** — three-way ladder, per-class breakdown, and what the run does *not* settle
- **[Deployment](https://github.com/dshnayder/kube-agents/blob/experiment/memory-honcho-ab/tests/memory-scale/honcho/README.md)** — manifests and config
- Backend, seeder and drain check are in the same branch's `harness/`, so it re-runs

Your decision rule was: if Hindsight wins on our fleet, merge. **It wins** — on one metric, for a reason that turned out to be structural rather than a tuning gap.

## The deciding number

Contamination — how often a **superseded** value reaches the model — at rung 1414:

| class | honcho | hindsight |
|---|---:|---:|
| **procedural** | 1.000 | 0.000 |
| **supersession** | 0.917 | 0.611 |

On every procedural probe at the largest rung, Honcho put a retired value in front of the model. Hindsight put none.

Hindsight's consolidation **deletes** retired content at write time, so it cannot return what is no longer there. Honcho keeps every message forever, and a retired runbook is still a message and still a legitimate semantic hit. For a fleet agent that's the governing failure mode: the wrong answer to "what's the current runbook version" is a *previously correct* answer, and nothing downstream catches it.

I don't believe a `dream` closes this — dreams consolidate conclusions, and a conclusion doesn't unwrite the message that `honcho_search` reads. That one is reasoned, not measured, and it's the single thing I'd test before treating the verdict as final.

## The org-peer shim: prediction confirmed

Upthread this was an argument. It's now a thing I had to build, so it's worth restating as evidence rather than theory.

There is no workspace-level corpus a peer-less reader can query. To seed the shared fleet corpus I invented a synthetic `meridian-platform` peer and routed every non-personal document onto it. That isn't an artifact of how I wrote the seeder — `conclusions/query` **refuses to run** without `observer` and `observed` filters, so an unscoped org bucket doesn't exist by design.

Your tenancy point remains right for the per-user half, and I'll keep conceding it. But "zero wrapper" holds only for that half. The requirement is per-user memory *and* an org-wide corpus, and the second half has to be a fictional peer whose beliefs are org knowledge, inside a model built for beliefs *about* peers, plus routing to keep each write landing on the right one of the two. It's which wrapper, not whether.

## Where I'd caution against reading the numbers as pro-Hindsight

Two Honcho results look better and shouldn't be credited:

- **`gold_recall` 0.937 vs 0.702 is an artifact.** Honcho returns messages verbatim so the document ID always survives; Hindsight paraphrases and drops the ID in 37 of 82 units. Substance frequently survived where the citation didn't. The same asymmetry gives the flat-file arm a mechanical 1.000.
- **`current_ranked_first` 0.857 vs 0.833 is a tie** — smaller than one probe out of 26, and confounded, since the two arms necessarily run different retrievers (bge-small 384-dim plus a CPU cross-encoder, against gemini-embedding-2 at 1536 with no rerank). Neither was moved off its shipped default, because neither really offers the choice.

And one Honcho result is a **genuine** win that I've written up as our gap rather than as a reason to switch: verbatim messages keep the citation. Hindsight losing the document ID in 37 of 82 units is a real problem for an agent that ought to say *which* standard it's quoting. The fix is provenance metadata on units.

Two costs the earlier a-priori argument hadn't predicted, both pointing the same way: derivation is **asynchronous** — 6 minutes to drain 614 messages at the top rung, so a fact isn't queryable for minutes after it's written, where Hindsight pays that inline at write — and embedding every *query* puts LiteLLM in the **read** path, not just the write path.

## What the run doesn't settle

- **Isolation probes failed by design.** We deliberately did not implement per-user scoping, so workspace-wide search leaks: 143 cross-user hits at rung 1414, and `Q-ISO-001` returns 100 other users' messages at *every* rung. Expected, and not evidence against Honcho. It's precisely why the comparison had to rest on contamination — that one survives even if you grant Honcho a working isolation wrapper.
- **Cost and latency weren't measured on either arm.** `tok/turn` is context-size parity, not cost parity, and no query latency was recorded on either side.
- Only message search was scored, because it's the surface the plugin's `honcho_search` uses. Conclusions and the dialectic are untested at scale.

## Where this lives

Every ground the design doc previously gave against Honcho was *a priori*, and — your point — all of them are answered by granting the wrapper. So the measured result is now its own section:

**→ [`docs/designs/memory.md` § What survives if `honcho` gets the wrapper](https://github.com/dshnayder/kube-agents/blob/feat/hindsight-memory/docs/designs/memory.md#what-survives-if-honcho-gets-the-wrapper)**

Recorded there too, since they're points in your favour: Honcho self-hosts fine, it's 4 workloads to Hindsight's 2 with the two first-party ones AGPL-3.0, and single-instance Postgres is a property of **both** — an ordinary Postgres HA question rather than a provider-choice one.

mastersingh24 (commented):
Thanks for the changes and the compare with Honcho and the detailed findings.  It's not surprising Hindsight in better overall as a memory provider, but I expected Honcho to be better at the per user stuff.

From an overall memory provider, Hindsight is definitely the better choice.

I do caution us on this route of trying to make Hermes multi-user ... I just don't think it's a sweet spot for Hermes.

I'm good with this though, but @bradhoekstra should have final weigh in

dshnayder:
The `pull-kube-agents-smoke-test` failure at `1bb6ab5` was not in this branch — the Cloud Build step died before anything was deployed:

```
#106 [credential-proxy 1/5] COPY --from=envoy-bin /usr/local/bin/envoy /usr/local/bin/envoy
ERROR: failed to prepare sgzcy94gb3h43n5wnmrd3zi4f: max depth exceeded
```

That is the overlay2 128-layer ceiling on the `agent-base` → `platform` → `credential-proxy` chain, fixed upstream in #665. This branch was behind that commit; it has now merged `upstream/main` (through `01084e7`), which brings the fix in.

Verified rather than assumed — built the `credential-proxy` target from this branch and measured the image:

```
OK: credential-proxy:latest has 116 layers, 4 under the 120 budget
```

Worth noting for anyone else who hits this: this branch adds two `COPY` layers to the `platform` stage, which is in that same chain, so the headroom is genuinely 4 and not the full margin #665 restored.

mastersingh24 (approved):

PR #636, opened 2026-08-11
feat(telemetry): make the OTLP endpoint configurable and auto-discovered

bradhoekstra:
# Pull Request

## Why This Change

The OTLP collector endpoint is hardcoded to the GKE Managed OpenTelemetry collector, `http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4318`. A customer running their own collector (say `otel-collector.otel-collector`) has no supported way to redirect telemetry, and on a cluster without managed OTel every export fails silently.

Two problems compound it:

1. **The literal is duplicated in seven deployable places** with no shared source — the Go const, the image bake, the Helm LiteLLM template, the kustomize LiteLLM base, and three `examples/` manifests.
2. **`OTEL_EXPORTER_OTLP_ENDPOINT` was inert for agent traces.** The agent does not export through that variable; it exports through the `hermes_otel` plugin, whose backend endpoint is baked into the image. The entrypoint only ever rewrote `resource_attributes['service.name']` and never touched `backends`, so setting the variable on a PlatformAgent changed nothing.

## What Changed

The endpoint is resolved in the operator at reconcile time, highest rung first:

| Rung | Source                                             | `otlpEndpointSource` |
| ---- | -------------------------------------------------- | -------------------- |
| 1    | `spec.deployment.env[OTEL_EXPORTER_OTLP_ENDPOINT]` | `DeploymentEnv`      |
| 2    | `spec.telemetry.otlpEndpoint` (new CRD field)      | `Spec`               |
| 3    | `OTEL_COLLECTOR_ENDPOINT` on the controller-manager | `OperatorEnv`        |
| 4    | in-cluster discovery                               | `Discovered`         |
| 5    | the GKE managed collector                          | `Default`            |

Rung 1 sits on top deliberately: `mergeEnvVars` already applies `spec.deployment.env` last, and demoting it would silently redirect telemetry for anyone relying on it. Any explicit rung short-circuits discovery entirely, so a configured install makes no extra API calls at all.

**Files:**

- `k8s-operator/internal/controller/telemetry.go` — new; the ladder, the probe, and the cache
- `k8s-operator/api/v1alpha1/common_types.go` — `TelemetrySpec` / `TelemetryStatus` on the shared `AgentSpec`/`AgentStatus`
- `k8s-operator/internal/controller/platformagent_controller.go` — `APIReader`, the cache fields, status reporting, and the rediscovery requeue
- `k8s-operator/internal/controller/manifest_helpers.go`, `platformagent_manifests.go` — `renderOptions` replaces the trailing positional `bool`
- `deploy/shared/otel_config.py` — new; rewrites the `hermes_otel` backend in every profile's config copy
- `deploy/shared/docker-entrypoint.sh`, `deploy/docker/Dockerfile`, `agents/platform/scripts/cluster_agent_profile.py` — call it
- `charts/kube-agents/` — the `telemetry` values block and the two helpers
- `k8s-operator/config/integrations/litellm/overlays/custom-otel/` — new copy-and-edit kustomize overlay

**Added/Changed/Fixed:**

- **Discovery** probes six well-known Service names by targeted `Get` before any label search — a miss is cheap, a cluster-wide `List` is not. It reads through `mgr.GetAPIReader()` so a cross-namespace `Get` cannot start a cluster-wide Service informer for a handful of reads an hour. Label matches are sorted by `(namespace, name)` before picking, because a non-deterministic choice would rewrite the Deployment env on alternating reconciles and roll the pod forever.
- **A gRPC-only collector (4317) is rejected, not selected.** Every exporter here sends `http/protobuf` and `hermes_otel` POSTs to `/v1/traces`, so a 4317 pick would fail on every span while looking correctly configured. Falling through to the default is the more debuggable failure.
- **Caching** is 5-minute TTL, cluster-wide, and includes the negative — "probed, found nothing" is authoritative and is the common case. An inconclusive probe is never cached; it falls back to the last known good endpoint, seeded from `.status.telemetry` so an operator restart plus one transient API error cannot flap every agent onto the default and roll its pod twice. Failed probes are rate-limited to one attempt per 30s so an API outage does not turn every reconcile into nine reads.
- **The trace path is now real.** `otel_config.py` rewrites `backends[0].endpoint` in the root home's plugin config and every `profiles/*/` copy, deriving the result from the pristine `/opt/defaults` copy — that both lets a newer baked endpoint land past `docker-entrypoint`'s `cp -ru` and makes clearing the override revert to the baked default instead of pinning a stale endpoint on the PVC forever. Cluster Agent profiles are created at onboarding, long after the entrypoint ran, so `cluster_agent_profile.py` applies it too; as a side effect those profiles get a `service.name` for the first time.
- **One Helm value, three consumers.** `telemetry.otlpEndpoint` flows into the PlatformAgent CR, the LiteLLM env, and the NetworkPolicy `namespaceSelector` together, so a release can never have the agent discover collector A while LiteLLM exports to B and the policy opens egress only to B's namespace. Empty means "do not decide here" — the operator can act on that, Helm cannot, so it keeps the shipping default.
- **`.status.telemetry.{otlpEndpoint,otlpEndpointSource}`** — the endpoint alone cannot distinguish "discovered the managed collector" from "fell back to it", and that is the entire support question when spans do not arrive. The observability skill and the inventory governance doc now read it instead of assuming the managed URL.

## Why This Matters

- Telemetry misconfiguration is the failure mode that looks like success: a green install exporting into the void. Every choice here — rejecting 4317, failing the Helm render rather than emitting a policy that blocks the collector you just named, reporting the source on status — turns a silent failure into a visible one.
- `OTEL_EXPORTER_OTLP_ENDPOINT` appearing in the pod while no spans arrive is the sharpest version of that, and the image-side change fixes it. **If this is ever split into separate PRs, the image change must land first or at the same time** — an operator-only release lets a customer configure a collector, see it in the pod env and in `.status`, and still get zero spans.
- Discovery is structural, not functional: a Service can expose 4318 and still refuse OTLP (wrong receiver, mTLS). `spec.telemetry.otlpEndpoint` is the escape hatch, and a restricted install that narrows the operator's cluster-wide RBAC makes discovery return `Default` — visibly, via the status field.

## Context

- The GKE managed collector is probed **first**, so on a cluster that has it discovery returns exactly the endpoint that used to be hardcoded. The default `helm template` render is byte-identical to `main` and the existing golden files do not move — the no-op is proven, not asserted.
- `safeSandboxEnvOverrides` needs no change: its allowlist and `ValueFrom` rejection stay exactly right, and rungs 2–5 all land in the managed slice the override merges over. No RBAC change either — the controller already holds cluster-wide `services` get/list/watch.
- `litellm.otel` stays a separate switch, still defaulting to off. Naming a collector does not turn the LiteLLM otel callback on, because that callback aborts every LLM request on DNS failure — too severe to flip as a side effect. Correspondingly, an external endpoint only fails the render when that callback is on; with it off there is no LiteLLM export for the egress rule to block.
- The `examples/` vLLM manifests still carry the managed endpoint literally — out of scope, and noted as such on the telemetry page.
- **Accepted wart:** `backends[0].name` stays `gke-managed-otel` even when repointed. Renaming risks anything keyed on it, so the rewrite is logged old→new to stderr instead. Worth a second opinion.
- The kustomize NetworkPolicy patch uses a JSON6902 `test` op as a guard: the path indexes an egress rule by position, and both failure modes of a bare `replace` (a reordered rule, an inserted one) are silent. The `test` op turns them into a build failure.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt`, `go test ./internal/...` — all green, including the golden tests run **without** `-update` first to prove the default path did not shift.
- New Go coverage: `TestOTLPHTTPEndpointForService` (port-name precedence, bare 4318, 4317-only rejected, UDP ignored, `ExternalName` rejected), `TestResolveOTLPEndpointPrecedence` (all five rungs, with an interposed reader that fails the test on any read to prove rungs 1–3 short-circuit discovery), and a discovery suite covering deterministic tie-breaks, TTL expiry, negative caching, the last-known-good fallback, the restart seed, and the failed-probe retry floor. The restart test was verified as a negative control — it reproduces the flap against the unfixed resolver.
- Python: 82 tests in `tests/` and 14 in `agents/platform/scripts/` pass, including a new `tests/test_otel_config.py`.
- `deploy/shared/entrypoint_gate_check.sh` gained a case that pushes a sentinel endpoint **through `run_entrypoint()`** and asserts both the scratch root and the scratch platform profile receive it. Confirmed by building the `entrypoint-gate-test` stage: `ok  otel endpoint reaches root + platform profile`.
- Helm: default render diffed byte-for-byte against `main` (identical); a vendor endpoint with `litellm.otel=true` fails with a message naming both escape hatches; the same endpoint with the callback off renders and keeps `gke-managed-otel` on the egress rule; an in-cluster endpoint moves the CR, the env, and the policy together. `make chart-sync` reports no drift.
- `make docs-check` clean (generated regions, links, terminology, map coverage); the docs site builds 43 pages; `prettier --check` clean on every changed Markdown/YAML/JSON file.
- The `review-docs-drift` skill was run against the branch diff and its findings addressed.

**Verified on a live cluster** (see the verification comment below): the resolver, the Deployment env, and every on-PVC `hermes_otel/config.yaml` move together on a real GKE install, the plugin confirms which config it loads, and unsetting the field reverts to the baked default. **Still not run:** a span actually landing in a non-GKE collector — the test collector had no receiver behind it, so this covers configuration propagation and not the wire.

---

Behaviour on an existing GKE install is unchanged by construction: the managed collector is the first thing discovery looks for and the last thing the ladder falls back to, and both the golden files and the default chart render are untouched. The risk concentrates in the entrypoint's plugin-config rewrite, which now runs on every start for every profile — it is written to warn and continue rather than fail a container start, since telemetry wiring is not worth failing a boot over.


bradhoekstra:
## End-to-end verification on a live GKE cluster

Ran the deploy this PR's "Not run" note was waiting on. Cluster
`kube-agents-cluster` (regional `us-central1`, GKE 1.35.6-gke.1250000, managed
OTel enabled), agent image built from this branch at `79dc1d1`.

**Rung 2 → the actual trace path.** Set
`spec.telemetry.otlpEndpoint: http://otel-collector.otel-collector.svc.cluster.local:4318`
against a collector Service in its own namespace:

| Layer | Result |
| --- | --- |
| `.status.telemetry` | endpoint + `otlpEndpointSource: Spec` |
| Deployment env | `OTEL_EXPORTER_OTLP_ENDPOINT` updated, pod rolled |
| `/opt/data/plugins/hermes_otel/config.yaml` | rewritten |
| `/opt/data/profiles/platform/plugins/hermes_otel/config.yaml` | rewritten |
| `/opt/data/home/.hermes/plugins/hermes_otel/config.yaml` | rewritten |
| `/opt/defaults/plugins/hermes_otel/config.yaml` (baked) | **untouched** |

`/v1/traces` appended exactly once, and the entrypoint logged each rewrite:

```
[OTEL-CONFIG] Repointing backend 'gke-managed-otel' from
  http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4318/v1/traces to
  http://otel-collector.otel-collector.svc.cluster.local:4318/v1/traces
```

**Risk 1 is resolved.** The plan flagged that nobody had confirmed *which*
`hermes_otel/config.yaml` the plugin actually reads. The plugin says so itself at
start-up:

```
[hermes-otel] ✓ gke-managed-otel at http://otel-collector.otel-collector.svc.cluster.local:4318/v1/traces
```

So the "rewrite every copy" mitigation is confirmed correct, not merely safe.

**Unsetting reverts, and does not pin a stale endpoint.** Removing `spec.telemetry`
put all three on-PVC copies back on the baked default and `.status` back to
`Discovered` — the `source_path` derivation from the pristine `/opt/defaults` copy
doing its job. No `[OTEL-CONFIG]` line on that pass, because the resolved endpoint
equalled the baked value and `apply()` correctly left `backends` alone.

**Candidate ordering holds on a live cluster.** The test Service sat at a
well-known name *and* carried `app.kubernetes.io/name: opentelemetry-collector`,
and discovery still returned `gke-managed-otel`. The managed-OTel no-op property
is real, not just unit-tested.

### Still not covered

- **No span was actually delivered.** The test Service had no receiver behind it —
  this verifies configuration propagation, not the wire. Risk 4 (structural, not
  functional discovery) stands as written.
- **`cluster-*` profiles.** No cluster profile exists on that install, so B4's
  `_pin_otel_endpoint` was not exercised outside its unit test.
- **Rung 4 to a non-managed collector.** Unreachable on this cluster by design —
  `gke-managed-otel` is candidate 1. Testing it would mean removing the
  GKE-managed Service, which I wasn't willing to do on a live cluster.

Cluster restored: test namespace deleted, `spec.telemetry` unset, back to
`Discovered` → managed collector.

toshiowang (approved):

PR #571, opened 2026-08-06
feat(install): support mirrored container registries on every install path

bradhoekstra:
# Pull Request

## Why This Change

Some environments may only pull container images from an approved registry — an in-project Artifact Registry, a corporate mirror, or an air-gapped cache. kube-agents supported this partially, and the parts that worked were scattered.

Worked: `REGISTRY_PREFIX` and the per-image vars in the provisioning scripts, the operator's `PLATFORM_AGENT_IMAGE` / `CREDENTIAL_PROXY_IMAGE` / `FLUENT_BIT_IMAGE` env vars, `spec.deployment.image` on a `PlatformAgent`, and the chart's three `*.image.repository` values.

Did not work:

- **LiteLLM, the GitHub token minter, and cert-manager** were hardcoded to `ghcr.io`, `us-docker.pkg.dev`, and `quay.io`.
- **Every build-time base image** was hardcoded in its Dockerfile, so rebuilding from source behind a firewall was impossible.
- **The chart never set the operator's image env vars.** Installing with a mirrored `platformAgent.deployment.image.repository` still left the fluent-bit sidecar on Docker Hub, and a `PlatformAgent` omitting `spec.deployment.image` fell back to `ghcr.io`. `helm install` reported success and the pods pulled from public registries minutes later.
- **Terraform** exposed only `image_tag`, never a registry.
- **Nothing listed the images an install pulls**, so mirroring was guesswork and the pins drifted — the chart pinned LiteLLM `v1.92.0` while the kustomize base it claimed to mirror pinned `v1.95.0`.

## What Changed

**Files:**

- `images.json` — new; the machine-readable inventory
- `scripts/mirror_images.sh`, `Makefile` — new `make mirror-images`, plus base-image build-arg pass-through
- `hack/check-image-inventory.sh`, `.github/workflows/validate.yml` — new `make images-check`
- `k8s-operator/scripts/{common.sh,provision_03,provision_09,provision_10}.sh`
- `k8s-operator/Makefile`, `k8s-operator/config/integrations/{litellm,github}/…`
- `charts/kube-agents/{values.yaml,templates/_helpers.tpl,templates/*.yaml,README.md}`
- `terraform/examples/full-install/*`
- `deploy/docker/Dockerfile`, `k8s-operator/Dockerfile`, `examples/inference-replay/replay-proxy/Dockerfile`
- `scripts/generate_docs.py`, `hack/check-docs-terminology.sh`, and the docs listed below

**Added/Changed/Fixed:**

[...]

PR #639, opened 2026-08-11
fix(provision): attach IMAGE_TAG to bare image references

bradhoekstra:
# Pull Request

## Why This Change

`install.sh` writes `OPERATOR_IMAGE` and `PLATFORM_AGENT_IMAGE` into `vars.sh` as bare
repository paths. That is correct by design — `IMAGE_TAG` is scoped to one pipeline run
and is deliberately never persisted (`init_var_image_tag`, `common.sh`) — so the tag has
to be re-attached wherever a reference is used.

`provision_03` did that for `OPERATOR_IMAGE` and for the custom-`AGENT_IMAGE` branch, but
the explicit `PLATFORM_AGENT_IMAGE` and `CREDENTIAL_PROXY_IMAGE` branches forwarded their
values verbatim. `kubectl set env` therefore put untagged references on the controller
manager, and Kubernetes resolved both to `:latest` — a tag the provisioner never builds or
pushes.

The agent container survives this: `provision_08` renders the CR with `image` and `tag` as
separate fields, so `spec.deployment.image` wins. The credential-proxy sidecar has no CR
field of its own — it derives from the agent image *only when `CREDENTIAL_PROXY_IMAGE` is
unset*, so an untagged value is strictly worse than no value at all. The result is a
`platform-agent-gateway` Pod stuck in `ImagePullBackOff` on any install that sets
`REGISTRY_PREFIX`. Hit on a real install against a private Artifact Registry.

There is also a latent double-tag: `"${OPERATOR_IMAGE}:${IMAGE_TAG}"` renders
`…/k8s-operator:v2:abc123` if a user pins `OPERATOR_IMAGE` with a tag.

Review then found that tagging `CREDENTIAL_PROXY_IMAGE` was the wrong repair. Pinning it
is cluster-wide and beats the operator's per-CR derivation, so the sidecar stays on the tag
of the run that wrote `vars.sh` while the agent moves — the exact skew the derivation
exists to prevent. It is now removed from the state file instead, which fixes the same
`ImagePullBackOff` by letting the derivation run.

## What Changed

**Files:**

- `k8s-operator/scripts/provision_03_gcp_gke_operator.sh`
- `k8s-operator/scripts/provision_11_deploy_inference_replay.sh`
- `k8s-operator/internal/controller/manifest_helpers.go` (comment only)
- `install.sh`
- `docs/site/src/content/docs/deploy/docker-images.md`
- `docs/README.md`

**Added/Changed/Fixed:**

- New `qualify_image_ref <ref> [tag]` in `common.sh`. Appends `IMAGE_TAG` only when the
  **final path segment** names neither a tag nor a digest, so `registry.example.com:5000/…`
  is not mistaken for a tagged reference and a user-pinned `:v2` / `@sha256:…` is returned
  untouched. An empty reference, or a bare one with no tag available, is a hard error at
  provisioning time rather than a `:latest` that fails at pod-pull time.
- `provision_03` routes every kube-agents reference through it: `OPERATOR_IMAGE`, the three
  `PLATFORM_AGENT_IMAGE` branches, and `CREDENTIAL_PROXY_IMAGE`. The inline `case` in the
  `AGENT_IMAGE` branch is replaced by the shared helper. Each qualified reference is
  assigned to a variable before being appended to the override array — a command
  substitution inside an array element discards the helper's exit status.
- `install.sh` no longer writes `CREDENTIAL_PROXY_IMAGE`. The operator derives the sidecar
  from each CR's own agent image; the env var overrides that for every `PlatformAgent` in
  the cluster, and `./install.sh --menu` → *Save & Apply* re-renders the CR through
  `provision_08` without re-running `provision_03`, so the two would drift apart. The
  `provision_03` branch stays for a user who wants to pin the sidecar by hand.
- `provision_11`'s `REPLAY_IMAGE` default is now bare and qualified at deploy time. It baked
  `${IMAGE_TAG:-latest}` into the value `init_var` persists, so the first run's tag froze
  into `vars.sh` and every later run silently redeployed it. (`REQUIRES_IMAGE_TAG=1` is
  already set in that step, so the `:-latest` fallback was unreachable.)
- `warn_on_registry_prefix_mismatch` now also covers `PLATFORM_AGENT_IMAGE` and
  `CREDENTIAL_PROXY_IMAGE` in step 03. Attaching a correct tag to a reference in the wrong
  registry makes a stale saved value pull successfully instead of failing visibly.
- `FLUENT_BIT_IMAGE` is deliberately exempt — it names an upstream `fluent/fluent-bit`
  release whose tag has nothing to do with `IMAGE_TAG`. Said so in a comment.
- Dropped `REPLAY_PROXY_IMAGE` from `install.sh`'s state file. Nothing in the repo reads it
  (step 11 uses `REPLAY_IMAGE`); it read as a supported knob that silently did nothing.
- `qualify_image_ref` and its Go twin `resolveAgentImage` now name each other, including
  why they diverge when no tag is available: the operator is serving a live CR and settles
  for `latest`, while a provisioning run can still abort and does.
- Docs: the `IMAGE_TAG` paragraph in `k8s-operator/scripts/README.md` and the
  registry-layers section of the Docker images page state the attachment rule per step and
  how to pin a reference independently of `IMAGE_TAG`; `docs/README.md` registers
  `qualify_image_ref` in the identifier-sources table.

## Why This Matters

- Unbreaks the default private-registry install. Today it deploys, then the gateway Pod
  fails to pull a `:latest` that was never published, and the error names an image the user
  never configured.
- Removes a cluster-wide sidecar pin that would have made agent and credential-proxy drift
  apart on the next `--menu` re-apply — silently, because both images exist and pull.
- Makes a pinned `OPERATOR_IMAGE`/`PLATFORM_AGENT_IMAGE` actually work instead of producing
  a malformed double-tagged reference.
- Stops step 11 freezing its first run's tag into `vars.sh`.

## Context

- Independent of #638 (gcloud minimum version), though both touch `install.sh` and
  `common.sh` in different regions; they do not conflict and can merge in either order.
- Merged `upstream/main` (7f1f8ca) to clear a conflict in `docs/README.md`: both branches
  appended a `common.sh` row to the identifier-sources table, and both rows are kept.
- Merged `upstream/main` again (6e5983f) after review. That one is a clean merge — the four
  files both sides touch (`docs/README.md`, `docs/site/…/deploy/docker-images.md`,
  `install.sh`, `manifest_helpers.go`) do not overlap textually — but #688 switched published
  image tags from `vX.Y.Z` to numeric `X.Y.Z`, which left this branch's pin examples reading
  `:v1.4.0` a few lines below the page's own tag convention. Corrected in the same push.
  #659's chart sync needs nothing: `CREDENTIAL_PROXY_IMAGE` appears in
  `charts/kube-agents/values.yaml` only as a commented escape hatch, never set by default, so
  the chart already agrees with dropping it from `install.sh`. `make iac-parity-check` (new in
  #659) passes, as do the installer tests #688 added (`tests/test_install_script.py`,
  `test_provision_rc_environment.py`, `test_upgrade_script.py`).
- No behaviour change for a stock public install: `REGISTRY_PREFIX` is the default, no
  `*_IMAGE` overrides are set, and `provision_03` emits no `kubectl set env` at all.
- Merged `upstream/main` (8fa44365) again to clear five conflicts. Four were keep-both:
  `docs/README.md` (upstream widened the registry-prefix row to name
  `THIRD_PARTY_REGISTRY_PREFIX`; this branch's `qualify_image_ref` row sits after it),
  `common.sh` (upstream's `warn_on_third_party_prefix_mismatch` and this branch's
  `qualify_image_ref` both append after `warn_on_registry_prefix_mismatch`), `install.sh`
  (upstream's conditional `THIRD_PARTY_REGISTRY_PREFIX` write, then this branch's comment on
  why the `*_IMAGE` values are bare), and `k8s-operator/scripts/README.md`.

  `docs/site/…/deploy/docker-images.md` was not. #571 restructured the whole private-registry
  section around `images.json` and the two prefixes, deleting the three-layer list this branch
  had annotated. Took upstream's section verbatim and restated this branch's two facts in their
  new homes — the `IMAGE_TAG` attachment paragraph after the `vars.sh`-persistence paragraph,
  and why `install.sh` leaves `CREDENTIAL_PROXY_IMAGE` unset under *What the prefix does not
  cover*. `provision_03` merged textually clean and was re-read: `qualify_image_ref` routing and
  upstream's third-party fluent-bit resolution compose without overlap.

  Re-ran on the merge: `bash -n` on the four scripts, `go build`/`go vet`, `make docs-check`,
  `make iac-parity-check`, `make images-check`, `prettier --check` (3.9.6), 82 tests across the
  installer/upgrade/rc/operator-makefile/profile-overlay/IaC-parity suites, and the ten
  `qualify_image_ref` shapes. All green. Not separately live-tested: the resolution changes no
  runtime behaviour beyond what the live validation below already exercised — three of the five
  conflicts are prose, and the two code ones are additive.

- This PR was generated with the help of Claude.

## Testing

- `bash -n` on all four changed shell scripts; `go build ./...` and `go vet ./internal/...`
  in `k8s-operator/`.
- Unit-tested `qualify_image_ref` over ten cases: bare path, existing tag, digest,
  registry-with-port (bare and tagged), single-segment ref, explicit tag argument, empty
  reference, and bare/tagged refs with `IMAGE_TAG` unset. The last three assert the new
  non-zero exit. All pass.
- Extracted the real `execute_operator` from `provision_03` and ran it with stubbed
  `make`/`kubectl` across six scenarios:
  - installer state under a private prefix — `PLATFORM_AGENT_IMAGE=…/platform-agent:abc123`,
    no `CREDENTIAL_PROXY_IMAGE` emitted.
  - stock public install — no overrides emitted, operator's compiled-in defaults stay
    authoritative.
  - user-pinned `:v2` operator and digest-pinned agent — both unchanged, no double-tagging.
  - hand-set bare `CREDENTIAL_PROXY_IMAGE` — still qualified to `:abc123`.
  - custom `AGENT_IMAGE` in a third-party registry — qualified, precedence respected.
  - `IMAGE_TAG` unset with a bare `OPERATOR_IMAGE` — `rc=1` after the CRD install and
    before any deploy, with a diagnostic naming the reference.
- `make docs-check` clean; `make docs-generate` leaves no uncommitted regeneration.
- `prettier --check` clean on the three changed Markdown files (prettier 3.9.6, the version
  `.github/workflows/prettier.yml` pins).
- Ran the `review-docs-drift` skill procedure over the branch diff.
- Re-ran all of the above on the merged branch after taking `upstream/main` (6e5983f):
  `bash -n`, `go build`/`go vet`, `make docs-check`, `make iac-parity-check`, `prettier
  --check`, the seven `qualify_image_ref` shapes that matter (bare, tagged, digest,
  registry-with-port bare and tagged, `IMAGE_TAG` unset, empty ref — the last two `rc=1`),
  and `pytest` over the installer/upgrade/rc/operator-makefile/profile-overlay suites (46
  passed). `pytest tests/` as a whole aborts in an unrelated suite's output capture on this
  machine, so the suites were run by file.

### Live validation

**Install:** `kube-agents-cluster` (regional `us-central1`, project `bhoekstra-gkedemos`),
namespace `kubeagents-system`, operator and agent at
`us-central1-docker.pkg.dev/bhoekstra-gkedemos/kube-agents/*:dns-endpoint-7dac349`, one
`PlatformAgent` CR (`platform-agent`) carrying `spec.deployment.image` +
`spec.deployment.tag` as `provision_08` renders them.

That install is itself an instance of the bug this PR describes: the controller manager
carried a **bare** `PLATFORM_AGENT_IMAGE=…/platform-agent` (resolves to `:latest` for any CR
omitting `spec.deployment.image`) alongside a tag-pinned `CREDENTIAL_PROXY_IMAGE`.

The pinned value happened to equal the derived one, so removing it proves nothing on its
own. Both directions were forced instead:

1. **The env var beats the derivation.** Set `CREDENTIAL_PROXY_IMAGE` to a distinctly
   different, genuinely pullable tag (`credential-proxy:dns-endpoint-17a20cb`). The operator
   rolled, and `deployment/platform-agent-gateway` went from generation 63 to 64 with
   `envoy-credential-proxy=…/credential-proxy:dns-endpoint-17a20cb` while the
   `platform-agent` container stayed on `dns-endpoint-7dac349` — the agent/sidecar skew,
   reproduced on a real cluster.
2. **Removing it restores the derivation.** `kubectl set env … CREDENTIAL_PROXY_IMAGE-`,
   which is the state this PR makes `install.sh` produce. Generation 65 came back with
   `envoy-credential-proxy=…/credential-proxy:dns-endpoint-7dac349`, back in step with the
   agent, derived from the CR's own image. The gateway Pod rolled to `Running` with all
   three containers ready and pulling.

**Cleanup:** the operator's `CREDENTIAL_PROXY_IMAGE` was restored to its pre-test value.
The gateway Deployment stayed at generation 65 (the restored value equals the derived one,
so nothing re-rendered) and its Pod remained `Running` 3/3. The lease
(`live-test-lease`, `--env gkedemos`) was held for the duration and released. Nothing else
was left behind.

**Not covered:** `provision_03` and `provision_11` were not re-run end-to-end against this
cluster — doing so redeploys the operator from a branch build and rewrites the install's
`vars.sh`, and that state file lives in the checkout that performed the original install,
not in this one. The env-override block and the `REPLAY_IMAGE` change are covered by the
stubbed-`kubectl` harness above rather than by a real pipeline run. The bare
`PLATFORM_AGENT_IMAGE` on the live operator was left as found; this PR fixes it for the
next provisioning run rather than by hand-editing the cluster.

The later `upstream/main` merge (6e5983f) and the numeric-SemVer docs fix on top of it were
not separately live-tested: the merge is textually clean and changes no file this branch
edits, and the fix is two example strings in prose. The live validation above still describes
the mechanism this PR changes.

---

Provisioning-time only; the sole operator-code change is a cross-reference comment. The
helper is a no-op for any reference that already names a tag or digest.


install.sh
```diff
@@ -1436,10 +1436,13 @@ main() {
   write_state_var "$vars_file" USER_PROFILE_ENABLED "false"
   write_state_var "$vars_file" HERMES_DASHBOARD_ENABLED "${PARAM_ENABLE_WEBUI:-false}"
   write_state_var "$vars_file" REGISTRY_PREFIX "$registry_prefix"
+  # Bare repository paths on purpose: IMAGE_TAG is scoped to a single pipeline
+  # run and is never persisted here, so the consuming step attaches it with
+  # qualify_image_ref. The replay proxy has no entry — provision_11 derives
+  # REPLAY_IMAGE from REGISTRY_PREFIX itself.
   write_state_var "$vars_file" OPERATOR_IMAGE "${registry_prefix}/k8s-operator"
   write_state_var "$vars_file" PLATFORM_AGENT_IMAGE "${registry_prefix}/platform-agent"
   write_state_var "$vars_file" CREDENTIAL_PROXY_IMAGE "${registry_prefix}/credential-proxy"
```
mplakhtiy: **MEDIUM — the redundant `CREDENTIAL_PROXY_IMAGE` state entry is fixed rather than removed, and now pins the sidecar for every CR in the cluster.**

This line runs unconditionally on every `install.sh` install, public default prefix included, and `vars.sh` is rewritten from scratch each run. Before this PR the value reached the operator bare and resolved to `:latest`; after it, `provision_03` forwards `<prefix>/credential-proxy:<IMAGE_TAG-of-that-run>`.

On the happy path that value is provably redundant. `resolveCredentialProxyImage` derives the sidecar from the resolved agent image — same registry, same tag, `platform-agent` -> `credential-proxy` — and the site says so: *"so mirrors that keep the image names only need `PLATFORM_AGENT_IMAGE`"* (`docs/site/src/content/docs/deploy/docker-images.md:141`). The CR rendered by `provision_08` carries `image: "${AGENT_IMAGE}"` / `tag: "${IMAGE_TAG}"` from the same run, so derivation and the env var produce the identical string.

Off the happy path it is worse than redundant, because the env var wins unconditionally:

```go
func resolveCredentialProxyImage(deployment *agentv1alpha1.DeploymentSpec) string {
	if override := os.Getenv(credentialProxyImageEnvVar); override != "" {
		return override
	}
```

— `k8s-operator/internal/controller/platformagent_manifests.go:2200`, immediately below a comment stating the derivation exists *"so agent and sidecar can never end up on different versions."*

**Failure scenario.** Install at `IMAGE_TAG=v1`; the operator's env is pinned to `.../credential-proxy:v1`. Later, run `./install.sh --menu` -> *Save & Apply*. That path calls `provision_08_deploy_platform_agent.sh` directly (`install.sh:974`) and never re-runs `provision_03`, so the CR is re-rendered at the new tag while the operator env keeps `v1`. The agent container runs `v2` and its credential-proxy sidecar runs `v1` — exactly the skew the derivation was written to prevent, and now silently, because both images exist and pull fine. The same applies to any second `PlatformAgent` CR that pins its own `spec.deployment.image` / `.tag`.

**Suggested fix.** Drop this line alongside the `REPLAY_PROXY_IMAGE` deletion, for the same stated reason, and extend the comment you just added above the block to say why both are absent. `qualify_image_ref` should stay — it is still needed for `OPERATOR_IMAGE` and `PLATFORM_AGENT_IMAGE`, and it still protects a hand-set `CREDENTIAL_PROXY_IMAGE` for the user who genuinely wants to pin the sidecar.

k8s-operator/scripts/README.md
```diff
@@ -60,9 +60,12 @@ scripts warn when an export is ignored or a saved image no longer matches the pr
 
 `IMAGE_TAG` is the deliberate exception to `vars.sh` reuse: tags change between deploys, so
 `provision.sh` asks for it once per pipeline run (or takes an exported `IMAGE_TAG`) and shares
-it with every step without saving it. Step 03 also forwards
+it with every step without saving it. The saved `*_IMAGE` values are therefore bare
```
mplakhtiy: **MEDIUM — this states a rule the tree does not follow, and papers over a frozen tag in step 11.**

> The saved `*_IMAGE` values are therefore bare repository paths.

False for `REPLAY_IMAGE`, which the *preceding sentence of the same section* (line 56) names as one of the `*_IMAGE` variables. `provision_11_deploy_inference_replay.sh:48` persists it with the tag baked in:

```bash
init_var "REPLAY_IMAGE" "$(registry_prefix)/replay-proxy:${IMAGE_TAG:-latest}" "Enter Replay Proxy container image"
```

`init_var` only writes when the variable is currently empty (`common.sh:186`), so that first run's tag is saved and every later run reuses it.

The docs-site copy over-claims in the same direction: *"`provision_03` attaches the current `IMAGE_TAG` to any of them"* (`docs/site/src/content/docs/deploy/docker-images.md:132`) — `provision_03` never sees `REPLAY_IMAGE`, which belongs to step 11, and deliberately skips `FLUENT_BIT_IMAGE`, an exemption stated in this file but not on that page.

`AGENTS.md`: *"Verify identifiers against source, not against other docs."*

**Failure scenario.** An operator reads this paragraph, concludes that a re-run re-tags every kube-agents image, and redeploys at a new `IMAGE_TAG`. Step 11 redeploys the replay proxy at the tag frozen on the first install — or at literal `latest`, if `IMAGE_TAG` was unset then — with no warning, because `warn_on_registry_prefix_mismatch` only compares the registry prefix and the prefix still matches.

**Suggested fix.** Either scope the claim to the values `install.sh` writes, or — better — route `provision_11` through `qualify_image_ref` too and drop the `:-latest` fallback, which makes the sentence true instead of narrower.

k8s-operator/scripts/common.sh
```diff
@@ -261,6 +261,30 @@ warn_on_registry_prefix_mismatch() {
...
+  local tag="${2:-${IMAGE_TAG:-}}"
+  # Only the final path segment can hold the tag — a registry host may carry
+  # a port, as in 'registry.example.com:5000/kube-agents/platform-agent'.
+  case "${ref##*/}" in
+    *:* | *@*) ;;
+    *)
+      if [ -n "$tag" ]; then
+        ref="${ref}:${tag}"
+      fi
+      ;;
```
mplakhtiy, line 444: **LOW — the helper silently produces the outcome its own docstring forbids.**

The docstring above says *"Handing a bare path to Kubernetes resolves it to ':latest', which the provisioner never builds or pushes."* The empty-tag branch does exactly that, silently.

It is unreachable from `provision_03` today, which sets `REQUIRES_IMAGE_TAG=1` and therefore hard-errors or prompts before `execute_operator` runs — but this is now a shared helper in `common.sh`, and the next caller without that flag gets a bare reference and no diagnostic. Two shapes behave badly, both verified by running the function:

- `qualify_image_ref "$ref"` with `IMAGE_TAG` unset -> returns `$ref` unchanged, bare.
- `qualify_image_ref ""` -> returns `:abc123`, a syntactically invalid reference.

Failing loudly at provisioning time beats failing at pod-pull time:

```suggestion
    *)
      if [ -z "$tag" ]; then
        print_error "qualify_image_ref: no tag available for bare reference '$ref' (IMAGE_TAG unset)"
        return 1
      fi
      ref="${ref}:${tag}"
      ;;
```

The empty-`ref` case wants a guard at the top of the function as well.

k8s-operator/scripts/provision_03_gcp_gke_operator.sh
```diff
@@ -102,33 +102,36 @@ verify_operator() {
...
-    env_overrides+=("PLATFORM_AGENT_IMAGE=${agent_image_ref}")
+    # from the same place.
+    env_overrides+=("PLATFORM_AGENT_IMAGE=$(qualify_image_ref "$AGENT_IMAGE")")
   elif [ "$(registry_prefix)" != "$DEFAULT_REGISTRY_PREFIX" ]; then
-    env_overrides+=("PLATFORM_AGENT_IMAGE=$(registry_prefix)/platform-agent:${IMAGE_TAG}")
+    env_overrides+=("PLATFORM_AGENT_IMAGE=$(qualify_image_ref "$(registry_prefix)/platform-agent")")
   fi
   if [ -n "${CREDENTIAL_PROXY_IMAGE:-}" ]; then
-    env_overrides+=("CREDENTIAL_PROXY_IMAGE=${CREDENTIAL_PROXY_IMAGE}")
+    env_overrides+=("CREDENTIAL_PROXY_IMAGE=$(qualify_image_ref "$CREDENTIAL_PROXY_IMAGE")")
```
mplakhtiy: **LOW — the two references this PR makes load-bearing get no registry-prefix warning, and the docs claim they do.**

Line 39 of this file calls `warn_on_registry_prefix_mismatch "OPERATOR_IMAGE"`. Nothing calls it for `PLATFORM_AGENT_IMAGE` or `CREDENTIAL_PROXY_IMAGE` — precisely the pair this block now forwards to the operator with a tag attached. `k8s-operator/scripts/README.md:57` states the general contract: *"the scripts warn when an export is ignored or a saved image no longer matches the prefix."*

**Failure scenario.** A user who installed against registry A exports `REGISTRY_PREFIX=B` and re-runs. `load_state` warns once that the export was ignored, and `warn_on_registry_prefix_mismatch` names `OPERATOR_IMAGE` specifically. They edit that one line in `vars.sh`. The agent and credential-proxy references still point at A, are now correctly tagged and therefore pull successfully, and the install quietly keeps running images from the registry the user believes they left.

Attaching a correct tag to a reference in the wrong registry is what makes this newly worth flagging: before this PR those pulls failed on `:latest`, which was at least visible.

k8s-operator/scripts/common.sh
```diff
@@ -261,6 +261,30 @@ warn_on_registry_prefix_mismatch() {
...
 
+# Attach IMAGE_TAG to an image reference that carries neither a tag nor a
+# digest. The saved *_IMAGE values are deliberately bare repository paths:
+# IMAGE_TAG is scoped to one pipeline run and is never persisted to vars.sh
+# (see init_var_image_tag), so the tag has to be re-attached where the
+# reference is used. Handing a bare path to Kubernetes resolves it to
+# ':latest', which the provisioner never builds or pushes.
+# A reference that already names a tag or digest is returned untouched, so
+# this is safe to apply to a user-supplied override.
+qualify_image_ref() {
```
mplakhtiy, line 427: **LOW — this shell rule duplicates the operator's Go rule, and the two already disagree.**

`resolveAgentImage` (`k8s-operator/internal/controller/manifest_helpers.go:188-214`) implements the same rule: split at the last `/`, treat `:` or `@` in the trailing segment as a tag or digest, otherwise append. `qualify_image_ref` is a second implementation of it in a second language, with no test tying them together — and they already diverge on the empty-tag case, where Go falls back to the literal `"latest"` and the shell leaves the reference bare.

Nothing forces them to stay in step. A cross-reference comment in both directions is the cheap version; naming the Go function in this docstring would at least make the next editor aware there is a twin.

mplakhtiy (commented):
Review of `3fa1847`. Five findings inline: two Medium, three Low.

install.sh
bradhoekstra: Agreed, and fixed in 8095cf7 — the line is gone rather than tagged.

You are right that on the happy path it was provably redundant, and right about why it is worse than redundant off it. The `--menu` → *Save & Apply* path is the one that convinced me: `install.sh:974` re-renders the CR through `provision_08` without ever touching `provision_03`, so the env var is a cluster-wide pin that nothing re-runs.

I forced the skew on a live install rather than reasoning about it. On `kube-agents-cluster` (`bhoekstra-gkedemos`), with the CR at `dns-endpoint-7dac349`, setting `CREDENTIAL_PROXY_IMAGE` to a distinctly different but genuinely pullable tag took `platform-agent-gateway` from generation 63 to 64 with `envoy-credential-proxy=…/credential-proxy:dns-endpoint-17a20cb` while the `platform-agent` container stayed on `dns-endpoint-7dac349`. Removing the env var — the state this PR now produces — brought generation 65 back with `…/credential-proxy:dns-endpoint-7dac349`, derived from the CR's own image, pod `Running` 3/3. Details in the PR body's Live validation section.

`qualify_image_ref` stays, and so does the `provision_03` branch, for the user who genuinely wants to pin the sidecar. The comment above the state-file block now explains why both `CREDENTIAL_PROXY_IMAGE` and `REPLAY_PROXY_IMAGE` are absent, and the one in `provision_03` says why the variable is normally unset.

k8s-operator/scripts/README.md
bradhoekstra: Correct on both counts, and fixed in 8095cf7. I took your "better" option: `provision_11` is now routed through `qualify_image_ref` and the `:-latest` fallback is gone, so the sentence is true rather than narrower.

The frozen tag was worse than the docs drift. `init_var` only writes what it prompted for, so baking `${IMAGE_TAG:-latest}` into the *default* is what persisted the first run's tag — the default is now bare and the tag is attached in `execute_inference_replay`. Worth noting the fallback was already unreachable: `provision_11` sets `REQUIRES_IMAGE_TAG=1` at line 28, so `load_state` has always guaranteed `IMAGE_TAG` by that point.

An existing `vars.sh` carrying a tagged `REPLAY_IMAGE` is unaffected — `qualify_image_ref` returns it untouched.

The docs-site copy is corrected in the same direction: it now attributes the attachment to the consuming step (`provision_03` for operator/agent/credential-proxy, `provision_11` for the replay proxy) and states the `FLUENT_BIT_IMAGE` exemption, which as you say was missing from that page.

k8s-operator/scripts/common.sh
bradhoekstra: Fixed in 8095cf7, with your suggestion applied and the empty-`ref` guard added at the top of the function.

Both shapes now fail loudly:

```
$ IMAGE_TAG= qualify_image_ref reg/ka/k8s-operator
ERR: qualify_image_ref: no tag available for bare reference 'reg/ka/k8s-operator' (IMAGE_TAG is unset). Set IMAGE_TAG, or pin the reference with an explicit tag or digest.
rc=1
$ qualify_image_ref ""
ERR: qualify_image_ref: called with an empty image reference
rc=1
```

Chasing the return value surfaced a real bug in my own code: `env_overrides+=("PLATFORM_AGENT_IMAGE=$(qualify_image_ref …)")` discards the exit status, so a failure would have been forwarded to the operator as an empty override — turning a loud failure back into a quiet one. Each reference is now qualified into a variable with `|| return 1` before being appended, and there is a comment saying why. Verified against the real `execute_operator` with stubbed `make`/`kubectl`: with `IMAGE_TAG` unset it returns `rc=1` after the CRD install and before any `deploy` or `set env`.

k8s-operator/scripts/provision_03_gcp_gke_operator.sh
bradhoekstra: Fair, and fixed in 8095cf7 — `warn_on_registry_prefix_mismatch` now covers `PLATFORM_AGENT_IMAGE` and `CREDENTIAL_PROXY_IMAGE` alongside `OPERATOR_IMAGE` at the top of the file. The check is a no-op when the variable is empty, which is the stock-install case for both.

Your framing is the part that made it worth doing: before this PR those pulls failed on `:latest`, so the wrong-registry state announced itself. Attaching a correct tag makes them succeed, which is exactly when a warning has to take over from the crash.

k8s-operator/scripts/common.sh
bradhoekstra: Fixed in 8095cf7 — cross-referenced in both directions, since a comment in only one file is the one that goes stale.

`qualify_image_ref`'s docstring names `resolveAgentImage` in `manifest_helpers.go`, and that function's docstring names the shell helper. Both call out the empty-tag divergence rather than leaving it to be discovered, and both say it is deliberate: the operator is serving a live CR and has to return something, so it settles for `latest`; a provisioning run can still abort, so as of this revision it does (see the `common.sh:283` thread).

I did consider collapsing them into one implementation. There is no seam for it — the shell runs before the operator image exists, and the Go path runs per-CR inside the reconciler — so a pointer in each direction is the honest version.

mplakhtiy (approved):

PR #641, opened 2026-08-11
feat(alerts): cap chat alerts per severity per day

stalhaali:
# Pull Request

## Why This Change

  Dedup bounds how often _one_ failure is reported. It does nothing about many
  _distinct_ failures at once — a draining node produces a hundred unrelated pods,
  each a new UID, none a repeat. This adds the backstop: a per-severity daily
  ceiling of 10 Critical, 5 Warning.

## What Changed

**Files:**

  - `agents/platform/scripts/session_kv_server.py` (+ tests)
  - `k8s-operator/internal/controller/platformagent_manifests.go` (+ tests)
  - `agents/platform/docs/session_management.md`
  - `docs/credential-isolation-design.md`,
    `docs/site/src/content/docs/reference/credential-isolation.md`,
    `docs/README.md`

**Added/Changed/Fixed:**

  - `inject_message` spends the severity's allowance before anything is posted or
    any agent turn starts — the single point both pass through, and the earliest
    point severity is known.
  - New `alert_quota` table on the `system-metadata` PVC. On disk, not in memory:
    an in-memory counter would hand out a fresh quota on every restart, turning a
    crash loop into the storm the cap prevents. Rows age out with the existing
    TTL sweep.
  - New `GET /v1/alert-quota` — today's spend and drops per severity.
  - `ALERT_DAILY_LIMIT_CRITICAL` / `_WARNING` added to the sandbox env allowlist
    in `safeSandboxEnvOverrides`, so the limits are tunable on the CR. `0`
    disables a cap. **Without this the documented override would silently do
    nothing** — the allowlist previously admitted four OTel names and dropped
    everything else.

 **Suppression is silent by design.** Announcing the ceiling spends a message to
  say no more are coming. The cost: a quiet channel no longer distinguishes
  "nothing is wrong" from "budget spent" — so every drop is counted, logged at
  `WARNING` with the workload, and readable from the endpoint.

### Live validation

  `platform-agent-host` (`us-east4`, `stalhaali-gkedemos`), image built from this
  branch via `make dev-rebuild-agent ARGS="platform"`, 

  - **Baseline:** endpoint returned `Critical {limit 10, 0, 0}`,
    `Warning {limit 5, 0, 0}` — table created on the PVC, endpoint served.
  - **Warning, through the real watcher:** 7 pods with an unresolvable image in
    `quota-test` → `sent 5, suppressed 2`. Five Chat messages, then silence.
    `sent` stopped at exactly 5, not 6. Two log lines named
    `quota-test/Pod/badimage-6` and `-7`.
  - **Independence:** `Critical` stayed `0/0` while `Warning` saturated.
  - **Critical:** `FailedScheduling` is filtered out (see Context), so 12
    `CrashLoopBackOff` payloads went to `POST /sessions/{id}/inject` — 1–10
    `{"status":"injected"}`, 11–12 `{"status":"suppressed",...,"1"/"2"}`,
    endpoint `Critical {sent 10, suppressed 2}`, `Warning` untouched.
  - **Cleanup:** both namespaces deleted, all `alert_quota` rows cleared.

### DIFF
--- a/agents/platform/scripts/session_kv_server.py
+++ b/agents/platform/scripts/session_kv_server.py
@@ -156,6 +156,50 @@ def _purge_plaintext_identities(conn: sqlite3.Connection) -> None:
         logger.info(f"Purged plaintext identity fields from {purged} session metadata row(s)")
 
 
+def _alert_daily_limit(env_var: str, default: int) -> int:
+    """Read a per-day alert ceiling from the environment. 0 disables the cap."""
+    raw = os.getenv(env_var, "")
+    if raw == "":
+        return default
+    try:
+        value = int(raw)
+    except ValueError:
+        logger.error(f"{env_var}={raw!r} is not an integer; falling back to {default}")
+        return default
+    # Negative is meaningless as a ceiling, and treating it as 0 makes "turn
+    # this off" forgiving of the two spellings an operator might reach for.
+    return max(value, 0)
+
+
+# Per-severity ceiling on alerts posted to chat in one UTC day. This bounds
+# volume, not redundancy: the dedup window in the event watcher is what stops
+# one failure being reported repeatedly, and this cap is the backstop for the
+# case that defeats it — many *distinct* failures at once, typically a node or
+# a namespace going down and taking a hundred unrelated pods with it.
+#
+# Suppression is deliberately invisible in chat. Announcing the ceiling would
+# spend a message to say no more messages are coming, which is self-defeating
+# when the point is a quieter channel. The trade-off is real and worth naming:
+# once the cap bites, a silent channel no longer distinguishes "nothing is
+# wrong" from "the budget is spent", so the accounting lives outside chat
+# instead. Every suppressed alert is counted per severity in `alert_quota`,
+# logged at WARNING with the workload that was dropped, and readable from
+# `GET /v1/alert-quota`. Anyone asking "did we miss something today" has an
+# answer; they just have to ask.
+#
+# Severities come from get_severity_details. Info is deliberately absent and so
+# uncapped — the watcher only forwards Warning-type events, so Info alerts do
+# not arrive here in practice and a ceiling on them would bound nothing.
+#
+# Counts are fleet-wide rather than per-cluster, matching the ceiling as
+# specified. The trade-off is that one collapsing cluster can exhaust the day's
+# budget for the others; the notice makes that visible when it happens.
+ALERT_DAILY_LIMITS = {
+    "Critical": _alert_daily_limit("ALERT_DAILY_LIMIT_CRITICAL", 10),
+    "Warning": _alert_daily_limit("ALERT_DAILY_LIMIT_WARNING", 5),
+}
+
+
 def init_db() -> None:
     db_dir = os.path.dirname(SESSION_KV_DB_PATH)
     if db_dir:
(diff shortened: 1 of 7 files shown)


jayantid:
Hey @stalhaali can you address kube-agents-bot comments and re-send for review?

jayantid (changes requested):
Can you address kube-agents-bot comments and re-send for review?

agents/platform/scripts/session_kv_server.py
```diff
@@ -450,4 +594,35 @@ def get_incident(chat_id: str, thread_id: str) -> Dict[str, str]:
     return {"chat_id": chat_id, "thread_id": thread_id, "report": row[0]}
 
 
+@app.get("/v1/alert-quota")
```
jayantid: This route is missing auth. Every other route here has
`dependencies=[Depends(verify_api_key)]`, and session_management.md:13 states
the rule — loopback in a shared namespace isn't an authorization boundary. As
written the sandbox container can read it without a key, and
`test_declared_routes_are_all_covered` will fail once this is rebased on main.

```python
@app.get("/v1/alert-quota", dependencies=[Depends(verify_api_key)])
def get_alert_quota(day: str = "") -> Dict[str, Any]:
```

Also add `("GET", "/v1/alert-quota", None)` to `PROTECTED_ROUTES`, and put the
token in the new runbook command:

```bash
kubectl -n kubeagents-system exec deployment/platform-agent-gateway -c platform-agent -- \
  sh -c 'curl -s -H "Authorization: Bearer $SESSION_KV_API_KEY" \
    http://127.0.0.1:8699/v1/alert-quota'
```

agents/platform/scripts/session_kv_server.py
```diff
@@ -199,6 +266,63 @@ def _post_initial_alert(active_platform: str, alert_msg: str) -> str | None:
...
+    Returns `(allowed, suppressed_today)`. `allowed` is False once the day's
+    ceiling is spent; `suppressed_today` is the running count of alerts the cap
+    has dropped today, which the caller logs so the drop leaves a trace even
+    though nothing is posted to chat.
+
+    Fails open. A cap is a comfort feature and a database that cannot be
+    written is not a reason to withhold an incident from an on-call human, so
+    any error here lets the alert through and is logged.
+    """
+    limit = ALERT_DAILY_LIMITS.get(severity, 0)
```
jayantid, line 410: Let's cap Info as well. `.get(severity, 0)` leaves any severity outside the dict
uncapped, and the reason for excluding Info — "the watcher forwards only
Warning-type events" — doesn't hold. `filter.Accept` (filter.go:97) checks
reason, namespace and count, never `ev.Type`; the informer (watcher.go:74) has
no field selector; `ev.Type` passes through at main.go:595. So any allowlisted
reason arriving as `type: Normal` becomes Info, skips the ceiling, and still
posts to chat and starts an agent turn. `BackOff` is on `defaultReasons` and the
kubelet emits it as Normal for image-pull back-off — the same storm you
validated against.

```python
ALERT_DAILY_LIMITS = {
    "Critical": _alert_daily_limit("ALERT_DAILY_LIMIT_CRITICAL", 10),
    "Warning": _alert_daily_limit("ALERT_DAILY_LIMIT_WARNING", 5),
    # Nothing filters on Event.Type, so Normal-type events with an allowlisted
    # reason land here too. Capped, not exempt.
    "Info": _alert_daily_limit("ALERT_DAILY_LIMIT_INFO", 5),
}
```

That also makes the dict cover everything `get_severity_details` can return, so
the `.get` default only fires for genuinely unknown strings. Needs
`ALERT_DAILY_LIMIT_INFO` on the `safeSandboxEnvOverrides` allowlist,
`test_uncapped_severity_is_allowed` inverted, and the "Info is uncapped"
paragraphs in session_management.md and the module comment corrected.

agents/platform/scripts/session_kv_server.py
```diff
@@ -356,6 +480,26 @@ def inject_message(session_id: str, request_data: Dict[str, Any], background_tas
...
+    # than a delivered one.
+    allowed, suppressed_today = _claim_alert_quota(severity_label)
+    if not allowed:
+        logger.warning(
+            f"Suppressed {severity_label} alert for {namespace}/{object_kind}/{object_name} "
+            f"({event_reason}): daily limit of {ALERT_DAILY_LIMITS[severity_label]} reached, "
+            f"{suppressed_today} suppressed today"
+        )
+        return {"status": "suppressed", "severity": severity_label, "suppressed_today": str(suppressed_today)}
+
```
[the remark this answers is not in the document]
stalhaali: This is expected. Furthermore, we'll turn dedupe window to 24 hours

agents/platform/scripts/session_kv_server.py
stalhaali: done

agents/platform/scripts/session_kv_server.py
stalhaali: Done

jayantid (approved):

PR #642, opened 2026-08-11
docs: require live-validation evidence in every pull request

bradhoekstra:
# Pull Request

## Why This Change

Green CI on this repository does not mean the change works. Three defects found in the last week were all green in CI first:

- `OTEL_EXPORTER_OTLP_ENDPOINT` was inert for agent traces — the variable appeared in the pod, `.status` looked right, and no span was ever exported, because nothing rewrote the plugin config the exporter actually reads.
- `provision_03` forwarded `CREDENTIAL_PROXY_IMAGE` untagged, so the sidecar resolved `:latest` and hit `ImagePullBackOff` on a fresh install.
- `--managed-otel-scope` needed gcloud 576.0.0, which no unit test could know.

Each surfaced only by running the installer end to end and reading state inside the pod. Unit tests, `make docs-check`, and the builds are necessary; they cannot tell you whether the operator reconciled the field or the agent picked it up. Nothing in the repository currently asks an author whether they checked.

## What Changed

The PR template asks the question, and the two contributor guides state the rule.

**Files:**

- `.github/PULL_REQUEST_TEMPLATE.md`
- `AGENTS.md`
- `docs/site/src/content/docs/contributing.md`
- `docs/README.md`

**Added/Changed/Fixed:**

- **`## Testing` grows a `### Live validation` subsection.** That heading was previously empty — no prompt, no placeholder. It now asks for the install (cluster, image tag, operator version), what was done, and what was observed at each layer the change claims to touch: the CR `.status`, the Deployment env, the file or process inside the pod.
- **`AGENTS.md` gets the rule as a PR-hygiene bullet**, next to the docs-drift bullet it parallels, with four sub-points: name the install and the observation, prove the mechanism rather than a coincidence, say what you could not cover, clean up.
- **The site's contributing page gets a `## Live validation` section** for human contributors, placed right after `## Local validation` — the section it complements and is explicitly not a substitute for.
- **`docs/README.md`** — the map rows for `AGENTS.md` and `contributing.md` name the new requirement. Both edits are length-preserving so Prettier does not reflow the two tables; the diff is two lines rather than a hundred.

## Why This Matters

- **"Prove the mechanism, not a coincidence" is the non-obvious half.** In the OTLP work the discovered endpoint was byte-identical to the baked default, so reading the config inside the pod matched — and proved nothing. Only setting a distinctly different endpoint and then reverting showed the code path ran at all. An author who does not know to do that can look thoroughly, honestly, and still learn nothing.
- **An explicit "Not live-tested, because …" is worth more than a blank section.** A blank `## Testing` reads as "no interesting testing"; it never reads as "this was never run against a cluster". Forcing the sentence makes the gap reviewable.
- **It is a prompt, not a gate.** No CI job parses the PR body, and adding one would mostly teach people to paste boilerplate. The template is where the question is cheap to ask and hard to forget.

## Context

- The rule sits alongside the existing **Docs-drift review before opening a PR** bullet and follows the same shape: a required pre-PR step for AI agents, stated once in `AGENTS.md`, mirrored for humans on the contributing page.
- Runtime skills that open pull requests (`submit-suggestion`, `fleet-audit`, the governance SOPs) are deliberately untouched — those PRs go to a customer's GitOps repository, not to this one, and this template does not apply to them.
- Follow-up worth considering, deliberately not in this PR: teaching `kube-agents-bot` to flag a pull request whose Live validation section is empty. That is a change to the bot's prompt, which lives outside this repository.

## Testing

- `make docs-check` — clean (generated regions, links, terminology, map coverage; 230 files, 204 tracked docs).
- `prettier --check` on all four changed files — clean, including `docs/README.md` after the length-preserving table edits.
- `docs/site`: `npm ci && npm run build` — 43 pages, no errors; the new `#live-validation` anchor resolves from the PR-hygiene bullet above it.
- The `review-docs-drift` skill was run against the branch diff: no Blocking findings. Checked in particular that no other file describes what a PR body must contain — `grep` for `PULL_REQUEST_TEMPLATE` and `gh pr create` across the repository turns up only `AGENTS.md`, the contributing page, and the runtime GitOps skills noted above.

### Live validation

**Not live-tested against a running kube-agents installation** — dogfooding the escape hatch this PR adds. The change ships no code: it touches a GitHub PR template and three Markdown files, none of which are baked into an image, read by the operator, or mounted into a pod. There is no layer on a cluster where the effect could be observed.

What was verified instead is above: the docs site renders the new section and its anchor, and the mechanical gates CI runs are clean. No cluster state was touched and nothing needs cleaning up.

---

Process-only, and reversible by deleting four paragraphs. The real risk is the opposite of a regression: a required section that authors fill with "tested locally" and reviewers stop reading. The escape hatch is written to be usable — a specific "Not live-tested, because X" is an acceptable answer — precisely so the section keeps costing something to fake.


jayantid (approved):

bnaylor:
Nice change -- this is exactly the gap the OTEL and image-tag misses fell through.  One thing worth tightening before merge: the requirement is now written out in full in three places (the template comment, `AGENTS.md`, and `contributing.md`), and none of them says which copy is canonical.  `AGENTS.md`'s own docs rules ask for that ("If you must summarise, say which page is canonical"), and the drift has already started: the three copies phrase the empty-section rule three different ways, and `contributing.md` says the `PlatformAgent` `.status` where the other two say the CR `.status`.

Suggested fix, and it's a small one: declare `AGENTS.md` canonical and have the `contributing.md` section defer to it the way its "Automated review" section already does, plus align the `.status` wording.  The template comment is fine self-contained -- a one-line "full contract in `AGENTS.md`" at the end of it wouldn't hurt, but I wouldn't block on that part.

bradhoekstra:
Addressed https://github.com/gke-labs/kube-agents/pull/642#issuecomment-5257780995 in #643.

PR #644, opened 2026-08-11
feat(inference): add Vertex AI as a first-class model provider

adrianchung:
# Pull Request

## Why This Change

Every shipping model provider needs an API key living in `platform-agent-secrets`, which means inference is billed to whoever owns that key, leaves the GCP project's data boundary, and puts a long-lived credential in the cluster. Vertex AI / Model Garden solves all three — it authenticates with Workload Identity, bills the install's own project, and serves both first-party Gemini and Model Garden publisher models (Anthropic Claude, Llama, Mistral) — but there was no way to select it.

This adds `MODEL_PROVIDER=vertex` as a first-class provider across all three install paths.

## What Changed

Two commits. The first is an independent bug fix; the second is the feature.

### `fix(litellm): roll the gateway when the model config changes`

**Files:**

- `k8s-operator/config/integrations/litellm/base/deployment.yaml`

The LiteLLM `ConfigMap`'s kustomize name hash is computed over the *un-substituted* `config.yaml` template, so the generated name (`litellm-config-k4b9ht22c7`) is byte-identical for every provider and every model. Changing `MODEL_PROVIDER` or `MODEL_DEFAULT_NAME` and re-running `provision_09` rewrote the ConfigMap contents under a name that never changed, so the running pods kept serving the old model until something else restarted them. The pod template now carries a `kubeagents.x-k8s.io/model-config: ${MODEL_PROVIDER}/${MODEL_DEFAULT_NAME}` annotation, which makes the Deployment spec itself change.

This predates the Vertex work and affects model switches on any provider, hence the separate commit.

### `feat(inference): add Vertex AI as a first-class model provider`

**Files:**

- `k8s-operator/scripts/{common.sh,provision_04_gcp_iam.sh,provision_07_gcp_k8s_secrets.sh,provision_09_deploy_litellm.sh,teardown_04_gcp_iam.sh,teardown_09_deploy_litellm.sh}`
- `k8s-operator/config/integrations/litellm/overlays/vertex/` (new: `kustomization.yaml`, `serviceaccount.yaml`, `deployment-patch.yaml`)
- `k8s-operator/Makefile`, `install.sh`
- `charts/kube-agents/{values.yaml,templates/litellm.yaml,templates/NOTES.txt,README.md}`
- `terraform/examples/full-install/{main.tf,variables.tf,terraform.tfvars.example,README.md}`, `terraform/modules/kube-agents-iam/{main.tf,variables.tf}`
- `INSTALL.md`, `README.md`, `docs/site/src/content/docs/{concepts/inference-gateway.md,reference/security-and-iam.md,deploy/kustomize.md,overview/what-is-kube-agents.md}`

**Added:**

- **Provisioning path.** `provision_04_gcp_iam.sh` creates a dedicated `kubeagents-litellm-gsa` with `roles/aiplatform.user`, granted on `VERTEX_PROJECT_ID` when it differs from the install project, and binds it to the gateway KSA. `provision_09_deploy_litellm.sh` renders a new `overlays/vertex` kustomize overlay that creates the annotated KSA and sets `VERTEXAI_PROJECT` / `VERTEXAI_LOCATION` on the pod. `VERTEX_PROJECT_ID` and `VERTEX_LOCATION` default to the install's project and region and are cached in `vars.sh`; `install.sh` prompts for them.
- **Helm path.** `litellm.modelProvider=vertex` plus `litellm.vertex.{serviceAccountName,serviceAccountAnnotations,projectId,location}`. The chart always creates the gateway KSA under `vertex` — no operator reconciles this one, and `serviceAccountName` pointing at a missing KSA leaves the pods unschedulable. Rendering fails fast if the project or location can't be resolved.
- **Terraform path.** `model_provider = "vertex"` in `terraform/examples/full-install` enables the Vertex API and creates the GSA, role grant, and Workload Identity binding.

**Fixed:**

- `teardown_04_gcp_iam.sh` revokes the cross-project grant **before** deleting the GSA. Deleting it first rewrites that binding to a `deleted:serviceAccount:...?uid=` member the revoke no longer matches, so the grant would outlive the identity.

## Why This Matters

- **No model API key in the cluster.** Vertex is the only provider with nothing in `platform-agent-secrets`; the gateway's credential is a Workload Identity binding that can be revoked in IAM.
- **Inference stays in the customer's project** for billing, quota, VPC-SC, and audit purposes — the usual blocker for regulated installs.
- **Model Garden opens up Claude on GKE** without an Anthropic account, which is the combination this repo's own demos want.
- All three install paths move together, so the Helm and Terraform users don't get a second-class version of the feature.

## Context

Naming: the provider string is `vertex_ai`, matching LiteLLM's own wire name, so nothing translates between `vars.sh`, `values.yaml`, and the rendered config. This started out as `vertex` (what `gcloud` calls it) with a translation step in the Makefile and the chart; #647 proposed `vertex_ai` independently and it is the simpler choice, so I adopted it. The kustomize overlay is `overlays/vertex_ai`, matching `overlays/chatgpt` ↔ `chatgpt`. The `VERTEX_PROJECT_ID` / `VERTEX_LOCATION` variables and the `litellm.vertex` values block keep the short spelling — those name the service, not the provider selector.

Overlaps with #647, which implements the same feature over the provisioning-script path only. See the comment thread there for the comparison; the substantive difference is that this PR gives the gateway its own service account rather than reusing the Platform Agent's.

`MODEL_DEFAULT_NAME` under Vertex is the **publisher model ID**, which is not always the vendor's own API name — Model Garden Claude models carry an `@`-suffixed version (`claude-opus-5@default`). A wrong ID surfaces as a 404 from the gateway rather than a provisioning error, so the docs say to check the Model Garden card.

Known follow-up work, deliberately out of scope: switching *away* from `vertex` doesn't delete the GSA (teardown does; provisioning has never reconciled removals), `MODEL_DEFAULT_NAME` is sticky across provider switches in `vars.sh`, changing `VERTEX_PROJECT_ID` leaves the old project's grant behind, there's no `--model-default-name` flag for non-interactive installs, and no Terraform output for the LiteLLM GSA email.

This PR was generated with the help of Claude, including a code-review pass whose findings are folded in.

## Testing

Automated checks: `bash -n` on all seven edited shell scripts; both kustomize overlays render (the vertex pod template resolves to `kubeagents.x-k8s.io/model-config: vertex_ai/claude-opus-5@default`); `make docs-generate` is a no-op and `make docs-check` passes all four checks; CI green on the pre-rebase revision, including `prettier`, `build`, and `Run Controller Tests`.

Not run locally — `helm`, `terraform`, and `prettier` are not installable on this machine, so the chart template and the Terraform plan are unverified outside CI.

### Live validation

Exercised against a real install: cluster `chung-kage` in `adrianchung-kubeagents-dev` / `us-east4`, Hermes runtime v0.20.0, images at `latest`. The install already existed on `MODEL_PROVIDER=anthropic`, so this was an in-place provider switch, which is the path most likely to break.

**What was run.** `provision_04_gcp_iam.sh` then `provision_09_deploy_litellm.sh` from this branch, with `MODEL_PROVIDER=vertex`.

**Observed, layer by layer:**

- **GCP IAM** — `kubeagents-litellm-gsa@adrianchung-kubeagents-dev.iam.gserviceaccount.com` created with `roles/aiplatform.user` on the project, and a `roles/iam.workloadIdentityUser` binding to `adrianchung-kubeagents-dev.svc.id.goog[kubeagents-system/kubeagents-litellm]`.
- **KSA** — `kubeagents-litellm` exists in `kubeagents-system` carrying the `iam.gke.io/gcp-service-account` annotation pointing at that GSA.
- **Deployment env** — the LiteLLM pod has `VERTEXAI_PROJECT=adrianchung-kubeagents-dev` and `VERTEXAI_LOCATION=global`, and `serviceAccountName: kubeagents-litellm`.
- **Inside the pod** — `google.auth.default()` resolves to the GSA, and a direct `generateContent` call against `projects/adrianchung-kubeagents-dev/locations/global` returns 200. (This was also the diagnostic that separated a Model Garden entitlement problem from the Workload Identity plumbing: the same token got 200 on a first-party Gemini model while Claude 404'd, until the Claude models were enabled in the console.)
- **End to end through the agent** — two kanban tasks completed via the agent's HTTP API, which is the path that reads `API_SERVER_MODEL_NAME` rather than the profile config:
  - `MODEL_DEFAULT_NAME=gemini-3.5-flash` → task `t_85e73028`.
  - `MODEL_DEFAULT_NAME=claude-opus-5@default` → task `t_72b560bf`, which correctly enumerated all 14 namespaces in the cluster.

**Proving the mechanism, not a coincidence.** The two runs above are distinctly different models from different vendors reached through the same gateway, and the second was confirmed to be genuinely Anthropic-on-Vertex rather than a silent fallback to the previous `anthropic` config by the Anthropic-shaped usage fields in the response (`cache_creation_input_tokens`, `ephemeral_5m_input_tokens`) combined with the absence of `ANTHROPIC_API_KEY` on the pod. The second switch is also what exercised the rollout fix: before the annotation, changing only `MODEL_DEFAULT_NAME` left the pods serving the old model because the ConfigMap name hash never changed; after it, `provision_09` rolled the Deployment on its own with no manual `kubectl rollout restart`.

**Not covered.** The Helm and Terraform paths were not applied against a live project — this install uses the provisioning-script path — so those two are reviewed-but-untested and I have called that out rather than implying parity of evidence. Cross-project Vertex (`VERTEX_PROJECT_ID` != `PROJECT_ID`) was not exercised; I only have the one project. The teardown ordering fix was reasoned from the `deleted:serviceAccount:...?uid=` rewrite behaviour, not run destructively against this install.

**Cleanup.** The cluster is deliberately left on `vertex` / `claude-opus-5@default` as my working dev install; no throwaway resources were created beyond the GSA and KSA that the feature itself provisions.

---

Additive: existing installs keep their current `MODEL_PROVIDER` and see no change beyond the LiteLLM pod rolling once on the next `provision_09` run. The rollout fix is the only behavioural change to a non-Vertex install, and it corrects a stale-config bug rather than introducing one.


charts/kube-agents/templates/litellm.yaml
mastersingh24, line 1: @adrianchung  - do we want to require litellm for this?  I thought that Hermes actually had built-in support for Vertex AI w/o requiring litellm ( https://hermes-agent.nousresearch.com/docs/guides/google-vertex ) ?

adrianchung:
Rebased onto `main` to pick up #631, which refactored the model-provider defaults in `install.sh` and `common.sh` into `DEFAULT_MODEL_PROVIDER` / `default_model_for_provider` / `is_valid_model_provider`. The Vertex changes now go through those helpers rather than restating the case statements, and `vertex` was added to the validator. That conflict resolution is new code the first review did not see, so:

/review

charts/kube-agents/templates/litellm.yaml
bradhoekstra: The same is true for OpenAI and Anthropic, but we are also relying on LiteLLM to keep the credentials out of the Agent's hands. If we want to drop LiteLLM we could, but that requires updating the credential proxy as well.

cc @haoxuw

bradhoekstra:
@Fuxiao-Gao Can you help decide if this duplicates / conflicts with #647?

adrianchung:
## Overlaps with #647

Heads-up for reviewers: @Fuxiao-Gao opened #647 about 45 minutes after this PR, implementing the same feature. Seven of its eight files are also touched here, so these cannot both merge — they conflict on `common.sh`, `k8s-operator/Makefile`, `provision_04_gcp_iam.sh`, `provision_09_deploy_litellm.sh`, `litellm/base/deployment.yaml`, `INSTALL.md`, and `concepts/inference-gateway.md`. Almost certainly independent work rather than a fork of either.

**What I've changed here in response.** #647 uses `vertex_ai` as the `MODEL_PROVIDER` value; this PR used `vertex` (what gcloud calls it) plus a translation step, because LiteLLM addresses it as `vertex_ai/<model>` on the wire. Theirs is simpler — I've adopted it and pushed the rename. That deletes the `$wireProvider` ternary from the chart, the `MODEL_PROVIDER=vertex_ai envsubst` rewrite from the Makefile, and a docs paragraph that existed only to explain the discrepancy. The overlay is now `overlays/vertex_ai`, matching `overlays/chatgpt` ↔ `chatgpt`. `VERTEX_PROJECT_ID` / `VERTEX_LOCATION` and the `litellm.vertex` values block keep the short spelling, since those name the service rather than the provider selector.

**The difference that's left is a design decision, not a diff-size one.** #647 attaches `serviceAccountName: kubeagents-platform-agent` to the LiteLLM Deployment and appends `roles/aiplatform.user` to the Platform Agent's role bundles — three lines, and the gateway inherits the agent's existing Workload Identity. This PR creates a dedicated `kubeagents-litellm` / `kubeagents-litellm-gsa` pair holding exactly one role, which costs a provisioning step, a teardown step, an overlay, and chart/Terraform plumbing.

I'd argue for the separate identity, and the reasoning is in `reference/security-and-iam.md` in this diff: the gateway is the process that handles model output and is reachable from any pod that can hit the `litellm` Service. Under #647's model it would run as an identity carrying `container.viewer`, `monitoring.viewer`, `logging.viewer`, `iam.securityReviewer`, and `mcp.toolUser`, so a compromise there reads the fleet rather than just calling models. It also grants `roles/aiplatform.user` on every install, including ones that never enable Vertex.

That said, it's the maintainers' call, and if the answer is "reuse the agent GSA," most of this diff comes out.

**The rest of the size difference** is scope rather than approach: this PR also does the Helm chart and Terraform paths (~180 lines), `install.sh` (menus plus `--vertex-project-id` / `--vertex-location`, so Vertex is selectable during a fresh install rather than `vars.sh`-only), teardown, and cross-project serving via `VERTEX_PROJECT_ID`.

**Independent of which lands:** the first commit here (`fix(litellm): roll the gateway when the model config changes`) is a pre-existing bug — the LiteLLM ConfigMap's kustomize name hash is computed over the un-substituted template, so it's identical for every provider and model, and changing `MODEL_DEFAULT_NAME` leaves the pods serving the old model. #647's `serviceAccountName` addition happens to roll the pods once, but subsequent model switches still won't. Worth keeping whichever feature PR wins.

adrianchung:
> @Fuxiao-Gao Can you help decide if this duplicates / conflicts with #647?

I've taken Fuxiao's change into here with the above message. PTAL.

charts/kube-agents/templates/litellm.yaml
adrianchung: OpenAI and Anthropic go through LiteLLM right? Ok to move forward with this? If we want to use native Hermes integration that's a different decision.

bradhoekstra (approved):

PR #647, opened 2026-08-11
feat(litellm): add native Vertex AI support with GKE Workload Identity

Fuxiao-Gao:
## Description

This PR adds native support for **Google Cloud Vertex AI** to the LiteLLM inference gateway using **GKE Workload Identity**, eliminating the requirement for static API keys and resolving TPM/RPM quota exhaustion on GKE clusters.

---

### Motivation

When running autonomous Kubernetes event triage and multi-step investigation loops, the platform agent sends comprehensive prompt contexts (~30,000+ tokens per request). With hosted API keys (e.g. Google AI Studio / Gemini API key), bursty tool calling frequently hits 60-second rolling Tokens-Per-Minute (TPM) limits (`HTTP 429: RESOURCE_EXHAUSTED`).

On Google Cloud / GKE, enterprise workloads authenticate securely via **GKE Workload Identity** and consume project-level Vertex AI quotas without API keys.

---

### Key Changes & File Breakdown

1. **`k8s-operator/config/integrations/litellm/base/deployment.yaml`**
   - Attached `serviceAccountName: kubeagents-platform-agent` to the LiteLLM deployment.
   - Enables Google Cloud Application Default Credentials (ADC) injection via GKE Workload Identity (`kubeagents-platform-gsa`).
   - Keeps API key env variables optional for backwards compatibility.

2. **`k8s-operator/config/integrations/litellm/base/config.yaml`**
   - Added `vertex_project: ${PROJECT_ID}` and `vertex_location: ${VERTEX_LOCATION}` parameters to the LiteLLM `model_list` template.

3. **`k8s-operator/Makefile`**
   - Updated `deploy-litellm` and `undeploy-litellm` targets to export default `VERTEX_LOCATION=us-central1` and pass `$$PROJECT_ID`, `$$REGION`, and `$$VERTEX_LOCATION` into `envsubst`.

4. **`k8s-operator/scripts/common.sh`**
   - Added `vertex_ai` as a supported `MODEL_PROVIDER` option in `init_var_model_provider()`.
   - Defaults to `gemini-2.5-flash` for Vertex AI (high throughput, low latency, and high TPM limits).

5. **`k8s-operator/scripts/provision_04_gcp_iam.sh`**
   - Added `roles/aiplatform.user` (Vertex AI User) to default IAM roles granted to `kubeagents-platform-gsa`.

6. **`k8s-operator/scripts/provision_09_deploy_litellm.sh`**
   - Added default `VERTEX_LOCATION=us-central1` and exported `PROJECT_ID`, `REGION`, and `VERTEX_LOCATION` during standalone LiteLLM provisioning.

7. **`INSTALL.md` & `docs/site/src/content/docs/concepts/inference-gateway.md`**
   - Updated setup documentation with Vertex AI deployment instructions, IAM prerequisites, and provider architecture.

---

### Compatibility & Verification

- **Zero-YAML Provider Switching**: Verified that users can toggle between `MODEL_PROVIDER=vertex_ai` (Workload Identity) and `MODEL_PROVIDER=gemini` (API Key) via `vars.sh` without editing any YAML manifests.
- **In-Cluster End-to-End Test**:
  - Sent test chat completions from the `platform-agent-gateway` container to LiteLLM via Vertex AI (`gemini-2.5-flash`), receiving instant `200 OK` completions under GKE Workload Identity.
  - Verified API key fallback path continues to function with `gemini-3.1-flash-lite`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -59,7 +59,7 @@ _(Alternatively via GitHub raw URL: `curl -fsSL https://raw.githubusercontent.co
 - **Provisioning Sources**: Puts the provisioning scripts on disk (this checkout, or a clone at the requested revision) and verifies they match the image ref _before_ the interview starts.
 - **GKE Cluster Setup**: Provisions the supported GKE Standard topology or connects to an existing cluster.
 - **Chat Integrations**: Configures Google Chat and/or Slack when selected.
-- **AI Model Credentials**: Prompts for Gemini, OpenAI, or Anthropic credentials.
+- **AI Model Credentials**: Configures Vertex AI (via Workload Identity without API keys), Gemini (AI Studio API key), OpenAI, or Anthropic.
 - **Automated Pipeline Execution**: Writes `k8s-operator/scripts/vars.sh` and launches `make gcp-provision`.
 
 The installer performs no GCP operation of its own — it configures and then delegates to the
@@ -350,9 +350,16 @@ kubectl rollout status deployment -n kubeagents-system
 To optionally deploy the LiteLLM Gateway or GitHub Token Minter:
 
 ```bash
-# Deploy LiteLLM Gateway
+# Option A: Deploy LiteLLM Gateway with Vertex AI (Recommended - Uses GKE Workload Identity, No API Keys)
+# Prerequisites: Ensure aiplatform.googleapis.com is enabled and kubeagents-platform-gsa has roles/aiplatform.user
+export MODEL_PROVIDER=vertex_ai
+export MODEL_DEFAULT_NAME=gemini-2.5-flash
+export VERTEX_LOCATION="us-central1"
+make deploy-litellm
+
+# Option B: Deploy LiteLLM Gateway with Gemini API Key (Google AI Studio)
 export MODEL_PROVIDER=gemini
-export MODEL_DEFAULT_NAME=gemini-3.5-flash
+export MODEL_DEFAULT_NAME=gemini-3.1-flash-lite
 make deploy-litellm
 
 # Deploy GitHub Integration (requires pre-configured github-app-credentials secret and env vars)
--- a/docs/site/src/content/docs/concepts/inference-gateway.md
+++ b/docs/site/src/content/docs/concepts/inference-gateway.md
@@ -11,7 +11,8 @@ The Platform Agent talks to an LLM through a **Completions API** proxy so provid
 
 | You want                                            | Use                                                    | Why                                                                                                                                      |
 | --------------------------------------------------- | ------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------- |
-| Fastest path with a hosted frontier model           | **LiteLLM → Gemini** (default)                         | One API key, no GPU node pool, no cluster egress beyond the LiteLLM pod.                                                                 |
+| Production GCP workloads with zero API keys         | **LiteLLM → Vertex AI** (Workload Identity)            | Uses GKE Workload Identity, project-level quotas, no manual key rotation or token limits.                                               |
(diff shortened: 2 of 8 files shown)


bradhoekstra:
@adrianchung Can you help decide if this duplicates / conflicts with #644?

adrianchung:
Hi @Fuxiao-Gao — flagging an overlap rather than a problem: I opened #644 about 45 minutes before this one and it implements the same feature. Seven of the eight files here are also in that diff (`common.sh`, `k8s-operator/Makefile`, `provision_04_gcp_iam.sh`, `provision_09_deploy_litellm.sh`, `litellm/base/deployment.yaml`, `INSTALL.md`, `concepts/inference-gateway.md`), so they'll conflict hard. Clearly independent work that landed the same afternoon.

**I've taken your naming.** #644 used `vertex` as the `MODEL_PROVIDER` value — what gcloud calls it — with a translation step to LiteLLM's `vertex_ai/<model>` wire name in the Makefile and the chart. Yours is simpler, and I've pushed the rename to `vertex_ai`; it deleted the translation from both places plus a docs paragraph that only existed to explain the discrepancy. Thanks for that.

**Where the two still differ,** and I think this is the decision worth having in the open:

- **Gateway identity.** You attach `serviceAccountName: kubeagents-platform-agent` to the LiteLLM Deployment and add `roles/aiplatform.user` to the Platform Agent's bundles. #644 creates a separate `kubeagents-litellm` / `kubeagents-litellm-gsa` with only that one role. Yours is far less code; my concern is that the gateway is the process handling model output and is reachable from any pod that can hit the `litellm` Service, so running it as the agent's identity means a compromise there also gets `container.viewer`, `logging.viewer`, `iam.securityReviewer`, and `mcp.toolUser`. Related: `roles/aiplatform.user` is appended to both `read_only_roles` and `gke_admin_roles` unconditionally, so every install gets Vertex permission whether or not it uses Vertex.
- **Install paths.** #644 also covers the Helm chart and Terraform. This PR covers the provisioning-script path, so Helm and Terraform users wouldn't get the feature.
- **Installer.** This PR doesn't change `install.sh`, so Vertex can only be selected by editing `vars.sh` after the fact, not during a fresh install.

**Two things I noticed while reading, worth fixing regardless of which PR lands:**

1. The provider table now documents the `gemini` default as `gemini-3.1-flash-lite`, and INSTALL.md's example changed to match, but `default_model_for_provider` still returns `gemini-3.5-flash`. The docs and the code disagree, and that change also looks unrelated to Vertex.
2. `vertex_project` / `vertex_location` are added to all three `model_list` entries unconditionally, so `${PROJECT_ID}` gets baked into the ConfigMap for gemini/anthropic/openai installs too. #644 sets `VERTEXAI_PROJECT` / `VERTEXAI_LOCATION` as pod env in a Vertex-only overlay to keep the other providers untouched.

Also worth knowing: your `serviceAccountName` addition changes the pod template, so the first switch to Vertex rolls the pods — but later `MODEL_DEFAULT_NAME` changes still won't, because the ConfigMap's kustomize name hash is computed over the un-substituted template and is identical for every provider and model. #644 has that as a separate commit (`fix(litellm): roll the gateway when the model config changes`) that's independent of either feature branch.

Happy either way — if the maintainers prefer this approach I'll close #644 and offer the Helm/Terraform/installer coverage and the rollout fix as follow-ups on top of yours. Deferring to @gke-labs on which to take.

Fuxiao-Gao:
@adrianchung 
Thanks for the detailed review and comparison!

I completely agree with the points raised—especially regarding the least-privilege identity model. Isolating the inference gateway to a dedicated `kubeagents-litellm-gsa` with only `roles/aiplatform.user` significantly reduces the blast radius compared to reusing the platform agent's identity. Furthermore, having full parity across Helm, Terraform, and `install.sh` makes #644 the right long-term implementation.

Closing this PR in favor of #644!

PR #650, opened 2026-08-12
docs(github): require an org-owned GitOps repo and fail early when it isn't

adrianchung:
## Why This Change

Setting up the GitOps repo on a fresh install took several hours of dead ends, and every one of them was a documentation gap rather than a bug.

The GitHub Token Minter resolves App installations with `GET /orgs/{org}/installation` ([`app.InstallationForOrg`](https://github.com/abcxyz/github-token-minter/blob/main/pkg/server/source/github.go)). GitHub serves personal accounts from `/users/{user}/installation`, and Minty has no fallback. So a **user-owned GitOps repo provisions cleanly, deploys, passes its readiness probes, and then fails every single mint** with an HTTP 500 wrapping a 404. Nothing in the repo said the repo had to be org-owned — `GITHUB_ORG` was documented as "org or user hosting the repo".

The failure is maximally misleading: it looks like a bad App ID or a bad key, so you go re-check those first. (The tell is 404 vs 401 — a 404 means the org lookup itself failed, before the key is ever considered.) Recovering meant creating an organization, which surfaced two more undocumented traps.

## What Changed

**Files:**

- `k8s-operator/scripts/provision_04_gcp_iam.sh`
- `k8s-operator/scripts/provision_10_deploy_github_minter.sh`
- `k8s-operator/config/integrations/github/README.md`
- `k8s-operator/README.md`
- `docs/site/src/content/docs/deploy/token-minter.md`
- `docs/site/src/content/docs/install/prerequisites.md`
- `INSTALL.md`

**Added/Changed/Fixed:**

- **A preflight check, so docs are not the only defense.** `check_github_org_is_organization()` looks `GITHUB_ORG` up via the unauthenticated `GET /users/{name}` and hard-fails on a user account or a name that does not exist, with a message naming the fix. Wired into step 04 (at the prompt) and step 10 (which is routinely run on its own). Inconclusive lookups — unreachable or rate-limited `api.github.com`, unexpected status — warn and continue, because a flaky network must not block an otherwise-fine provision. `SKIP_GITHUB_ORG_CHECK=true` bypasses it, following the `ALLOW_UNENCRYPTED_SECRETS` convention.
- **The org requirement documented at every point someone picks the value** — prerequisites, the token-minter page, the integration README, plus `INSTALL.md` and `k8s-operator/README.md`, whose manual `make deploy-github` path bypasses the preflight entirely. The integration README is the canonical long form; the rest link to it.
- **The "Only on this account" App trap.** A GitHub App created under a personal account cannot be installed onto an organization until you flip **Advanced → Make public**. Not mentioned anywhere previously.
- **A `gcloud`-only KMS import recipe.** When Go is unavailable the provisioner warns `You must import the key manually later` and prints no working instructions — the "manual" instructions in the other branch are another `go run ./cmd/minty`, i.e. the exact thing that is missing. The section opens by telling happy-path readers to skip it.
- **The step-ordering trap.** `GITHUB_APP_ID` gates step **07** as well as 04 and 10, so enabling GitHub after an install that ran without it and re-running only step 10 leaves the Minter in `CreateContainerConfigError: secret "github-app-credentials" not found`. `scripts/README.md` now states the 04 → 07 → 10 ordering.

## Why This Matters

- The constraint is invisible until after a successful-looking deployment, and the error points at the wrong cause. This is the single highest-friction step in a GitOps install.
- A typo in `GITHUB_ORG` produced the identical downstream 404, so the check catches that too.
- No behaviour changes for anyone already correctly configured: an org-owned repo takes one extra HTTP GET and proceeds.

## Context

- Minty is pinned at `v2.7.1`; the org-only lookup is still present on upstream `main`, so this is a constraint to document rather than a bug to wait out.
- GitHub shares one namespace between users and orgs, so an org cannot take the name of an existing personal account — worth knowing before planning a rename.
- Deliberately **not** in scope: fixing `provision_10`'s warning text to print the `gcloud` commands instead of the unusable `go run` one. That is a real fix rather than a docs workaround, but it is a different change from this one.
- Generated with the help of Claude.

## Testing

- `make docs-check` — clean (generated regions current, links resolve, terminology, map inventory).
- `prettier --check` — clean on all six changed Markdown files.
- `bash -n` — clean on all three changed shell scripts.
- No Go source touched, so no operator build.
- **Not run:** the Astro site build (`docs/site && npm run build`). `npm ci` fails against an SSO-authenticated package mirror on this host. The three site edits are plain prose in existing `.md` files with no new components or frontmatter, and `make docs-check` covers their links.

### Live validation

**Install:** `platform-agent-host` in `adrianchung-kubeagents-dev` / `us-east4`, agent image `ghcr.io/gke-labs/kube-agents/platform-agent:9396c01`.

This PR is the writeup of a real end-to-end GitOps bring-up on that cluster, so the failure modes below were observed, not predicted:

- **The bug this PR documents, reproduced.** With `GITHUB_ORG=adrianchung` (a user account) the Minter reached 2/2 Running and passed readiness, then returned HTTP 500 on every mint: `Get "https://api.github.com/orgs/adrianchung/installation": retryable status code: 404`. Signing an App JWT by hand against the same KMS key confirmed the diagnosis rather than assuming it — `/app` → 200 (id 4561196), `/orgs/adrianchung/installation` → 404, `/users/adrianchung/installation` → 200 `"type": "User"`. The 404-not-401 also proves the KMS-signed JWT was accepted, isolating the fault to the org lookup.
- **The fix, verified.** Moved the repo to the `adrianchung-labs` organization and re-ran **04 → 07 → 10**. Mint returned HTTP 200 with a `ghs_` installation token; that token then read the repo (200) and listed installation repositories (200).
- **The whole path, agent-driven.** Asked the Platform Agent to open a PR via `/p/platform/v1/chat/completions` (200, 16s) → kanban task `t_e2ed090b` → a real PR on the GitOps repo, **authored by `app/kube-agents-adrianchung`**. The App author is the proof the minted token was used rather than any human credential.
- **The step-ordering trap, hit for real.** Re-running only step 10 after setting the GitHub vars left the Minter in `CreateContainerConfigError: secret "github-app-credentials" not found`; re-running step 07 fixed it. That is the origin of the new `scripts/README.md` section.
- **The KMS import fallback, used in anger.** The Minty CLI import failed six times with `go tool compile: signal: killed` under an endpoint security agent. The four `gcloud` commands now in the README are the exact ones that succeeded, including the non-obvious `CLOUDSDK_PYTHON_SITEPACKAGES=1` (without it `gcloud` reports "Cannot load the Pyca cryptography library" even though its own interpreter has the module).

**The new preflight function** was exercised against the live GitHub API by extracting it from `common.sh` and running the full matrix, rather than by re-running a provisioning pipeline:

| `GITHUB_ORG` | Result |
| --- | --- |
| _(empty)_ | continues |
| `gke-labs` (org) | continues |
| `adrianchung-labs` (org) | continues |
| `adrianchung` (user) | **blocked**, exit 1 |
| `definitely-not-a-real-account-zzz9` | **blocked**, exit 1 |
| `adrianchung` + `SKIP_GITHUB_ORG_CHECK=true` | warns, continues |

This also confirmed the `[ -z "$org" ] && return 0` guard does not trip `set -e`.

**Not covered:** I did not re-run the full `provision_04` / `provision_10` pipeline end-to-end with the check in place, because the live install is already correctly configured — the check would take its no-op path and prove nothing. The rate-limited and unreachable-network branches were not exercised against a genuinely rate-limited API; they are reached via the same status/transport handling the table above covers.

**Cleanup:** the smoke-test PR and its branch remain on the private GitOps repo as evidence and can be deleted on request. The temporary JWT file and debug pod used for diagnosis were removed. No changes were left on the cluster.

---

Docs plus a preflight guard; no operator or runtime code paths are touched, so the blast radius is the provisioning scripts. The one new failure mode is the intentional hard exit on a user-owned or nonexistent `GITHUB_ORG`, which is escapable with `SKIP_GITHUB_ORG_CHECK=true` if it is ever wrong.


adrianchung:
Note: This is a documentation change I wanted to clean up when I found out I needed to create a github org to setup our gitops as an additional setup step now.

adrianchung:
Fixed in 33a3265. The finding was correct in every particular, including the precedent it cited — thanks.

Verified before fixing: `install.sh:1264` defaulted the prompt to `gh api user -q .login`, which is a personal login and so is guaranteed to fail the new check; the prompt was labelled "GitHub Org / Username", actively inviting the wrong answer; and `install.sh:1319-1326` does guard the other `provision_04` hard exit in exactly this shape, with a comment making the same point about landing after the cluster and operator exist.

**What changed**

- Split the check in two. `github_account_type()` is a classifier that echoes `organization|user|missing|unknown` and neither prints nor exits, so it is safe to call from a prompt loop. `check_github_org_is_organization()` keeps the fatal behaviour for `provision_04` / `provision_10`, which still matters for a hand-edited `vars.sh`.
- `install.sh` sources `common.sh` already (line 988, before this prompt at 1272), so it now validates at the prompt and re-asks. The value is settled while nothing has been created.
- The default is now the operator's first organization (`gh api user/orgs -q '.[0].login'`) instead of their login, and the prompt reads "GitHub Organization".
- Both error paths now name the recovery you noted was missing: edit `GITHUB_ORG` in `vars.sh`, or set `SKIP_GITHUB_ORG_CHECK=true`.

**One hazard the fix had to avoid.** `prompt_read` returns the default immediately when `PARAM_NON_INTERACTIVE=true` or there is no controlling TTY (`install.sh:466-479`), so wrapping it in a retry loop would spin forever on an invalid value. Non-interactive runs therefore fail at the prompt rather than looping. I tested that specifically by extracting the shipped loop and running it under `set -Eeuo pipefail` with a stubbed `prompt_read`:

| case | result |
| --- | --- |
| `GITHUB_ORG=adrianchung` (user), non-interactive | exits 1 at the prompt, no hang |
| `GITHUB_ORG=adrianchung-labs` (org) | proceeds |
| `GITHUB_ORG=""` | exits 1, "an organization is required" |
| user account + `SKIP_GITHUB_ORG_CHECK=true` | proceeds |

The classifier itself was re-verified against the live API: `""` → unknown, `gke-labs` → organization, `adrianchung-labs` → organization, `adrianchung` → user, a nonexistent name → missing.

Not fixed, and out of scope for this PR: the precedent loop at `install.sh:1319-1326` has the same latent non-interactive hang if `permission_set=custom` is ever reached with an empty role list. It looks unreachable today because `prompt_menu` auto-selects non-interactively, so I have left it alone rather than widening this change.

fatoshoti (approved):

PR #652, opened 2026-08-12
fix(ci): fork the devops-bench gke cluster module and reap orphaned eval infrastructure before provisioning

lapis2002:
# Pull Request

## Why This Change
`pull-kube-agents-smoke-test` gets aborted routinely — a PR merges, a run is superseded, someone cancels. Prow sends SIGKILL after the grace period, which skips the devops-bench harness's teardown and the job's `trap ... EXIT` alike. Every in-pod cleanup mechanism is best-effort by construction, so the GKE cluster and its node service account survive with no state file left to destroy them from.

The next run then fails its `tofu apply` on `409 Already Exists` before the evaluation starts, and stays broken until someone deletes the leftovers by hand. Because the job runs at `max_concurrency: 1` against a fixed cluster name, one abort blocks the smoke test for every subsequent PR.

## What Changed

**Files:**
- `bench/tf/modules/cluster/**` — forked from `kubernetes-sigs/devops-bench@4670d76`
- `bench/tf/prebuilt/gpu-stress-test/main.tf`
- `bench/tf/prebuilt/gpu-stress-test/variables.tf`
- `hack/ci-eval-pr.sh`

**Added/Changed/Fixed:**
- **Forked the `cluster` module into `bench/tf/modules/`.** The stacks previously sourced it over `git::` pinned at `4670d76`, which made it read-only for us. Resource labels and recovery behaviour cannot be expressed from the caller side, so the module had to be ours. Only the `gke` sub-module is forked; `kind` has no local changes and still points at the upstream `git::` URL rather than becoming a copy nobody maintains.
- **Added `terraform_data.reap_orphans`**, which deletes a leftover cluster and node service account before anything else is created. Both `google_container_cluster.primary` and `google_service_account.gke_nodes` now `depends_on` it. devops-bench hands each run a fresh scratch dir with empty state, so the resource is always created and the check always runs.
- **The reap strips the account's project IAM bindings before deleting the account.** Deleting the account first leaves each binding behind as an inert `deleted:serviceAccount:...?uid=` member; the recreated account gets a new uid, so its fresh bindings do not replace them and every aborted run would add five more. Roles are read back from the live policy rather than hardcoded, so they cannot drift from the `google_project_iam_member` resources, and the same lookup clears tombstones left by earlier runs.
- **Labelled the infrastructure.** The stack sets `managed-by`, `build-id`, and `pull-number` as provider `default_labels`, which the module inherits; the module additionally sets `managed-by` in `node_config.resource_labels`, since a node pool is not itself a labelable resource and its instances can outlive a deleted cluster. `hack/ci-eval-pr.sh` passes Prow's `BUILD_ID` and `PULL_NUMBER` through as `TF_VAR_*`, defaulting to `local`/`none` outside CI.

## Why This Matters
- One aborted run currently blocks the smoke test for every PR that follows, until someone intervenes manually. After this change an abort costs the next run time (a cluster delete) rather than blocking it.
- Leaked resources were previously indistinguishable from anything else in the project. The labels make them findable without knowing what the stacks create.
- Recovery happens at the start of the apply that needs it, not on a schedule. There is no janitor cronjob to own, no grace period to tune, and nothing that can delete resources a run is still using.

## Context
- Job config lives in `GoogleCloudPlatform/oss-test-infra`: `prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml`.
- **The reap is safe only at `max_concurrency: 1`**, where nothing else can own a cluster with this name. That assumption is load-bearing and is stated in a comment above the resource: allowing the job to run concurrently would make this delete a live run's cluster.

#### Manual
**Orphans were created deliberately, then observed being reaped.** 
- A `test-cluster` and a `gke-nodes-test-clus-9750e4@hangdng-gkedemos.iam.gserviceaccount.com` service account were created by hand before the run, so the reap had something real to find rather than matching a coincidence:
  - `module.cluster.module.gke[0].terraform_data.reap_orphans` appeared in the plan and ran first, ahead of the cluster and service account, confirming the `depends_on` edges.
  - It logged `reaping orphaned cluster test-cluster left by a previous run` and deleted it, then `reaping orphaned service account gke-nodes-test-clus-9750e4@…` and deleted that. 6m34s total.
  - The module then recreated the service account with the identical `account_id` 16 seconds later, with no collision against GCP's 30-day soft-delete tombstone. This was the specific failure mode the design risked, and it is the reason the test was run.
  - Cluster created in 11m10s, node pool in 2m43s, `guest_accelerator nvidia-l4` attached.
  - Teardown destroyed all 10 resources cleanly; the eval passed at 100% (OutcomeValidity [GEval] 1.00). No test artifacts left behind — the hand-made cluster and account were consumed by the reap, and everything the run created was destroyed.

**Labels verified on the live resources:**
- Cluster `effective_labels`: `build-id=local`, `pull-number=none`, `managed-by=kube-agents-bench`, plus Google's `goog-terraform-provisioned=true` — confirming `default_labels` reaches the module through provider inheritance.
- Cluster `resource_labels`: `managed-by=kube-agents-bench`.
- Node pool `node_config.resource_labels`: `managed-by=kube-agents-bench` alongside GKE's own

#### Prow Smoke Test
Prow presubmit `pull-kube-agents-smoke-test`, build [`2087552474192809984`](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/652/pull-kube-agents-smoke-test/2087552474192809984) on PR #652. Build cluster `kube-agents-prow`; infrastructure created in `kube-agents-evals`, zone `us-west4-a`. Job result SUCCESS in 56m05s, evaluation `OutcomeValidity` 1.00.

**The reap fired against a real leak**
  `terraform_data.reap_orphans` completed in 36s having removed ten stale bindings: 
```
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/artifactregistry.reader for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=105133120633823066254
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/artifactregistry.reader for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=115247132972129865670
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/logging.logWriter for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=105133120633823066254
module.cluster.module.gke[0].terraform_data.reap_orphans: Still creating... [10s elapsed]
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/logging.logWriter for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=115247132972129865670
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/monitoring.metricWriter for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=105133120633823066254
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/monitoring.metricWriter for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=115247132972129865670
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/monitoring.viewer for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=105133120633823066254
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/monitoring.viewer for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=115247132972129865670
module.cluster.module.gke[0].terraform_data.reap_orphans: Still creating... [20s elapsed]
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/stackdriver.resourceMetadata.writer for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=105133120633823066254
module.cluster.module.gke[0].terraform_data.reap_orphans: Still creating... [30s elapsed]
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): removing stale binding roles/stackdriver.resourceMetadata.writer for deleted:serviceAccount:gke-nodes-test-clus-9750e4@kube-agents-evals.iam.gserviceaccount.com?uid=115247132972129865670
module.cluster.module.gke[0].terraform_data.reap_orphans (local-exec): Updated IAM policy for project [kube-agents-evals].
module.cluster.module.gke[0].terraform_data.reap_orphans: Creation complete after 36s [id=3db57514-4bce-b2c4-89fe-fcc5ff44b470]
```
Five roles across **two distinct uids**, two separate aborted runs had each left a full set of tombstones behind in the shared CI project. Neither `reaping orphaned cluster` nor `reaping orphaned service account` appeared, so there was no live resource to delete; only the dead bindings the old code left behind remained. That is exactly the failure mode this change targets, observed in production rather than manufactured.

  **Labels verified at the API** 
On `google_container_cluster.primary`:
```

  resource_labels  = { "managed-by" = "kube-agents-bench" }
  effective_labels = { "build-id"                   = "2087552474192809984"
                       "goog-terraform-provisioned" = "true"
                       "managed-by"                 = "kube-agents-bench"
                       "pull-number"                = "652" }
```
  and on the node pool's `node_config`, `resource_labels = { "managed-by" =
  "kube-agents-bench" }`, merged by GKE with its own `goog-gke-accelerator-type =
  nvidia-l4` and `goog-gke-node-pool-provisioning-model = on-demand`.

  `build-id` matches the Prow build ID exactly and `pull-number` matches the PR, so
  the `TF_VAR_prow_build_id` / `TF_VAR_prow_pull_number` exports in
  `hack/ci-eval-pr.sh` reach the provider's `default_labels` end to end. The teardown
  plan later shows the same labels being *read back off the live resources* before
  destroying them — proof they were applied to GCP rather than only planned.

---

#### Risk and rollout
Contained to bench/ evaluation infrastructure and the CI eval script; no operator, chart, or agent behaviour changes. The one real risk is the max_concurrency: 1 assumption — if that is ever relaxed on the Prow job, this reap would delete a concurrent run's cluster; adding a build-id guard would lift the constraint if concurrency is ever wanted. Reverts cleanly, since each run provisions from an empty state file and the fork leaves nothing behind.

### DIFF
--- a/bench/tf/modules/cluster/gke/main.tf
+++ b/bench/tf/modules/cluster/gke/main.tf
@@ -1,0 +1,279 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 5.0.0"
+    }
+  }
+}
+
+# account_id is capped at 30 chars, so we can't fit a long cluster name. Truncating
+# the name alone is unsafe: names that share a prefix but differ only in a suffix past
+# the cutoff (e.g. "<base>-east" vs "<base>-west" when <base> is long) would collapse to
+# the same account_id and collide. Append a short hash of the *full* cluster name so the
+# id stays unique per cluster regardless of where the readable part is truncated.
+locals {
+  # A GCP IAM account_id must be lowercase letters, digits, and hyphens. Lowercase
+  # the cluster name and collapse any other characters (uppercase, underscores,
+  # dots) into hyphens before slicing so the readable part is always valid. The
+  # md5 hash below is over the *original* name, so distinct names still differ.
+  gke_nodes_name_slug = trim(substr(replace(lower(var.cluster_name), "/[^a-z0-9]+/", "-"), 0, 9), "-")
+
+  # Named once so the account and the reap below cannot disagree about what to
+  # look for.
+  gke_nodes_account_id = "gke-nodes-${local.gke_nodes_name_slug}-${substr(md5(var.cluster_name), 0, 6)}"
+}
+
+# kube-agents fork: a run killed before its teardown leaves this module's
+# resources behind with no state file to destroy them from, and the next apply
+# then dies on "409 Already Exists" before the evaluation can start. Reap them
+# first, so an aborted run costs the next one time rather than blocking it.
+#
+# This is safe only because the smoke test runs at max_concurrency 1: nothing
+# else can own a cluster with this name. That assumption is load-bearing -- if
+# the job is ever allowed to run concurrently, this deletes a live run's cluster.
+#
+# devops-bench hands each run a fresh scratch dir with empty state, so this is
+# created every time and the reap always runs.
+resource "terraform_data" "reap_orphans" {
+  triggers_replace = [var.cluster_name, var.location, var.project_id]
+
+  provisioner "local-exec" {
+    interpreter = ["/bin/bash", "-c"]
+    command     = <<-EOT
+      set -euo pipefail
+
+      # The cluster goes first: its nodes authenticate as the service account
+      # below, so deleting that account first would strand them.
+      if gcloud container clusters describe "${var.cluster_name}" \
+           --location "${var.location}" --project "${var.project_id}" >/dev/null 2>&1; then
+        echo "reaping orphaned cluster ${var.cluster_name} left by a previous run"
+        gcloud container clusters delete "${var.cluster_name}" \
+          --location "${var.location}" --project "${var.project_id}" --quiet
+      fi
+
+      sa="${local.gke_nodes_account_id}@${var.project_id}.iam.gserviceaccount.com"
+
+      # Strip the account's project bindings before deleting the account itself.
+      # Deleting it first leaves each binding behind as an inert
+      # "deleted:serviceAccount:<email>?uid=..." member that nothing later
+      # removes, and the recreated account gets a new uid, so its fresh bindings
+      # are separate entries -- every aborted run would add five more.
+      #
+      # The filter matches the tombstones as well as the live member, so this
+      # also clears whatever earlier reaps left behind. It runs outside the
+      # existence check below because those tombstones outlive the account.
+      #
+      # Roles are read back from the live policy rather than listed here, so
+      # this cannot drift from the google_project_iam_member resources above.
+      # Filtering on this account's email also keeps it away from
+      # agent_container_admin, which binds a long-lived external account.
+      while read -r role member; do
+        [[ -n "$role" && -n "$member" ]] || continue
+        echo "removing stale binding $role for $member"
+        gcloud projects remove-iam-policy-binding "${var.project_id}" \
+          --member "$member" --role "$role" --condition=None --quiet >/dev/null || true
+      done < <(gcloud projects get-iam-policy "${var.project_id}" \
+                 --flatten="bindings[].members" \
+                 --filter="bindings.members:$sa" \
+                 --format="value(bindings.role,bindings.members)")
+
+      if gcloud iam service-accounts describe "$sa" --project "${var.project_id}" >/dev/null 2>&1; then
+        echo "reaping orphaned service account $sa"
+        gcloud iam service-accounts delete "$sa" --project "${var.project_id}" --quiet
+      fi
+    EOT
+  }
+}
+
+resource "google_service_account" "gke_nodes" {
+  account_id   = local.gke_nodes_account_id
+  display_name = "GKE Node Service Account for ${var.cluster_name}"
+
+  depends_on = [terraform_data.reap_orphans]
+}
+
+resource "google_project_iam_member" "gke_nodes_log_writer" {
+  project = var.project_id
+  role    = "roles/logging.logWriter"
+  member  = "serviceAccount:${google_service_account.gke_nodes.email}"
+}
+
+resource "google_project_iam_member" "gke_nodes_metric_writer" {
+  project = var.project_id
+  role    = "roles/monitoring.metricWriter"
+  member  = "serviceAccount:${google_service_account.gke_nodes.email}"
+}
+
+resource "google_project_iam_member" "gke_nodes_monitoring_viewer" {
+  project = var.project_id
+  role    = "roles/monitoring.viewer"
+  member  = "serviceAccount:${google_service_account.gke_nodes.email}"
+}
+
+resource "google_project_iam_member" "gke_nodes_metadata_writer" {
+  project = var.project_id
+  role    = "roles/stackdriver.resourceMetadata.writer"
+  member  = "serviceAccount:${google_service_account.gke_nodes.email}"
+}
+
+resource "google_project_iam_member" "gke_nodes_artifact_registry_reader" {
+  project = var.project_id
+  role    = "roles/artifactregistry.reader"
+  member  = "serviceAccount:${google_service_account.gke_nodes.email}"
+}
+
+resource "google_project_iam_member" "agent_container_admin" {
+  count   = var.agent_service_account != "" ? 1 : 0
+  project = var.project_id
+  role    = "roles/container.admin"
+  member  = "serviceAccount:${var.agent_service_account}"
+}
+
+resource "google_compute_firewall" "allow_iap_ssh" {
+  count   = var.enable_iap_ssh ? 1 : 0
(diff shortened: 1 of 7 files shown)


jayantid (commented):
LGTM, please also get a review from Eugene

bench/tf/modules/cluster/gke/main.tf
```diff
@@ -0,0 +1,279 @@
...
+        [[ -n "$role" && -n "$member" ]] || continue
+        echo "removing stale binding $role for $member"
+        gcloud projects remove-iam-policy-binding "${var.project_id}" \
+          --member "$member" --role "$role" --condition=None --quiet >/dev/null || true
+      done < <(gcloud projects get-iam-policy "${var.project_id}" \
+                 --flatten="bindings[].members" \
+                 --filter="bindings.members:$sa" \
+                 --format="value(bindings.role,bindings.members)")
+
+      if gcloud iam service-accounts describe "$sa" --project "${var.project_id}" >/dev/null 2>&1; then
```
eugeneng04, line 116: `done < <(gcloud ...)` puts the lookup in a process substitution, which `set -e` on line 58 does not cover so we never checks that command's exit code. If `get-iam-policy` fails, the loop reads zero rows, line 86 treats that as "no bindings", and the delete at line 97 runs with all five bindings still attached. Those become deleted:`…?uid=` entries that nothing can remove

Fix: policy=$(gcloud …) then done <<< "$policy"

bench/tf/modules/cluster/gke/main.tf
```diff
@@ -0,0 +1,279 @@
...
+# resources behind with no state file to destroy them from, and the next apply
+# then dies on "409 Already Exists" before the evaluation can start. Reap them
+# first, so an aborted run costs the next one time rather than blocking it.
+#
+# This is safe only because the smoke test runs at max_concurrency 1: nothing
+# else can own a cluster with this name. That assumption is load-bearing -- if
+# the job is ever allowed to run concurrently, this deletes a live run's cluster.
+#
+# devops-bench hands each run a fresh scratch dir with empty state, so this is
+# created every time and the reap always runs.
```
eugeneng04: The claim on 50–51 doesn't hold as written. devops-bench doesn't hand each run empty state `TFDeployer._state_flags()` only redirects state when `TF_DATA_DIR` is set, which is `BENCH_PARALLEL` only, and `hack/ci-eval-pr.sh:29` sets that to false.

bench/tf/modules/cluster/gke/main.tf
```diff
@@ -0,0 +1,279 @@
...
+      #
+      # Roles are read back from the live policy rather than listed here, so
+      # this cannot drift from the google_project_iam_member resources above.
+      # Filtering on this account's email also keeps it away from
+      # agent_container_admin, which binds a long-lived external account.
+      while read -r role member; do
+        [[ -n "$role" && -n "$member" ]] || continue
+        echo "removing stale binding $role for $member"
+        gcloud projects remove-iam-policy-binding "${var.project_id}" \
+          --member "$member" --role "$role" --condition=None --quiet >/dev/null || true
```
eugeneng04: echo prints before the attempt, stdout goes to /dev/null, `|| true` drops the exit code. A cleanup where every removal failed produces output identical to one where all succeeded.

eugeneng04 (changes requested):

bench/tf/modules/cluster/gke/main.tf
lapis2002: updated to `done <<< "$policy"`

bench/tf/modules/cluster/gke/main.tf
lapis2002: updated the comment.

bench/tf/modules/cluster/gke/main.tf
lapis2002: fixed. The echo now runs after the attempt and says which way it went, and the if takes the exit code instead of `|| true`.

jayantid (approved):
Approving since this is time sensitive fix, feel free to add followup comments Eugene

PR #654, opened 2026-08-12
fix(teardown): skip GKE backup plan teardown if feature disabled

shalinibhatiapl:
# Pull Request

## Why This Change

Addresses a `PERMISSION_DENIED` error in the RC pipeline teardown step (`Step 2 - Provision RC Environment / deploy-rc`). 
The provisioning step correctly skipped creating a GKE backup plan if `ENABLE_GKE_BACKUP_PLAN` was false, but the teardown script unconditionally attempted to query the Backup Plan API. Since the WIF Service Account lacks `gkebackup.admin` permission (as the feature isn't actively enabled in RC), this unconditionally triggered an API permission error.

## What Changed

Added an early exit condition in the teardown script to check if the feature is disabled.

**Files:**

- `k8s-operator/scripts/teardown_12_gke_backup_plan.sh`

**Added/Changed/Fixed:**

- Fixed `teardown_12_gke_backup_plan.sh` to check `ENABLE_GKE_BACKUP_PLAN` and exit with 0 if it's explicitly falsy, mimicking the provisioning script's behavior.

## Why This Matters

- Prevents unnecessary API calls when the feature is disabled.
- Fixes the asymmetry between how `provision_12` and `teardown_12` scripts handle the feature toggle.

## Context

**Why was this error ignored earlier, and why do we still ignore it?**
The CI teardown orchestrator script (`k8s-operator/scripts/teardown.sh`) executes this particular step with an appended `|| true`. This means that even if the `gcloud describe` command threw a `PERMISSION_DENIED` error and the script exited with a non-zero exit code, the master script ignored the failure and proceeded to tear down other resources. This acts as a safeguard so one cleanup failure doesn't halt the rest.

We intentionally chose to keep this behavior. If the state file is missing, the script assumes `true` and attempts cleanup to ensure no orphaned resources are left behind. As a result, CI (which doesn't have this variable set) will still attempt the teardown and log the `PERMISSION_DENIED` error, but it will continue to be safely ignored without breaking the pipeline.

## Testing

- ✅ Verified locally that when `ENABLE_GKE_BACKUP_PLAN=false`, running `./k8s-operator/scripts/teardown_12_gke_backup_plan.sh` cleanly outputs `ℹ Skipping GKE Backup Plan teardown...` and exits gracefully.
- ✅ Verified locally by the user that when `ENABLE_GKE_BACKUP_PLAN=true` (or is unset), the script correctly issues a deletion request and waits for operation completion without errors (if properly authenticated).

---

### DIFF
--- a/k8s-operator/scripts/teardown_12_gke_backup_plan.sh
+++ b/k8s-operator/scripts/teardown_12_gke_backup_plan.sh
@@ -15,6 +15,13 @@ source "${SCRIPT_DIR}/common.sh" "$@"
 
 ensure_teardown_state
 
+# Do not prompt in teardown. If the value is explicitly false, skip.
+# If missing, we assume true to attempt a safe cleanup of any potential orphaned resources.
+if [ -n "${ENABLE_GKE_BACKUP_PLAN:-}" ] && ! is_truthy "$ENABLE_GKE_BACKUP_PLAN"; then
+  echo -e "  ${C_YELLOW}ℹ Skipping GKE Backup Plan teardown (ENABLE_GKE_BACKUP_PLAN=${ENABLE_GKE_BACKUP_PLAN}).${C_RESET}"
+  exit 0
+fi
+
 if [ "${DRY_RUN:-0}" -ne 1 ] && ! gcloud beta --help &>/dev/null; then
   print_info "The 'gcloud beta' component is not installed. Skipping GKE Backup Plan teardown."
   exit 0


k8s-operator/scripts/teardown_12_gke_backup_plan.sh
```diff
@@ -15,6 +15,13 @@ source "${SCRIPT_DIR}/common.sh" "$@"
 
 ensure_teardown_state
 
+# Do not prompt in teardown. If the value is explicitly false, skip.
+# If missing, we assume true to attempt a safe cleanup of any potential orphaned resources.
+if [ -n "${ENABLE_GKE_BACKUP_PLAN:-}" ] && ! is_truthy "$ENABLE_GKE_BACKUP_PLAN"; then
```
eLeontev, line 20: I assume the right fix will be the opposite:
- include the missed permissions to our SA which manages the RC pipeline (add the permission to the existing account + add the fix to the related script)
- include Backup functionality to the RC happy path by adding new env variable to github project and pass the variable to the action script
- validate if autopush/staging gcp projects have the same issue (not visible due to the backup are disabled) and extend the missed permissions to be in sync.

mplakhtiy, line 20: **MEDIUM — the guard cannot fire in the RC pipeline, so the `PERMISSION_DENIED` this PR names is unchanged**

The guard only triggers when `ENABLE_GKE_BACKUP_PLAN` is set. In the RC pipeline it never is, by three independent facts:

- `scripts/release/provision_rc_environment.sh` exports only `PROJECT_ID`, `REGION`, `CLUSTER_NAME`, `PROJECT_NUMBER` before calling `./k8s-operator/scripts/teardown.sh`.
- `vars.sh` is git-ignored (`.gitignore:4-8`) and nothing in `.github/` or `scripts/` restores it, so the CI checkout has no state file.
- With no state file, `ensure_teardown_state` (`k8s-operator/scripts/common.sh:370`) takes its `else` branch, which sets `PROJECT_ID`/`REGION`/`CLUSTER_NAME` and nothing else.

**Failure scenario.** The RC teardown runs, the variable is empty, the script falls through to `gcloud beta container backup-restore backup-plans describe`, the WIF service account lacks `gkebackup` access, and the error text carries no `not found` / `service_disabled` / `is disabled` marker — so the `elif` classifier misses, the `else` branch prints a red `✗ Error describing GKE Backup Plan`, and `exit "$describe_err"` fires. That is precisely the failure **Why This Change** says this PR addresses. `teardown.sh:32` still swallows it with `|| true`, so nothing breaks — but nothing changes either.

The **Context** section concedes this; the title (`fix(teardown)`) and the **Why This Change** / **Why This Matters** sections do not, and the description is what a reader indexes on later.

The change is still a genuine improvement for local and installer-driven teardowns, where `provision_12` has written `ENABLE_GKE_BACKUP_PLAN=false` into `vars.sh` — so this is not an argument for dropping it. Two ways to make it deliver the stated fix:

- **Deeper, and fixes every environment.** Widen the error classifier in this same file so permission errors are treated as "nothing this identity can clean up here" and degrade to a warning. That covers RC, `uninstall.sh --source-ref` (which fetches the scripts into a temp dir with no `vars.sh`), and any operator running teardown from a fresh clone — none of which the flag guard reaches.
- **Narrow.** Export `ENABLE_GKE_BACKUP_PLAN=false` in `scripts/release/provision_rc_environment.sh` alongside the other RC exports.

Either way, reword the description so **Why This Change** and **Context** agree on what the RC pipeline actually does after this merges.

k8s-operator/scripts/teardown_12_gke_backup_plan.sh
```diff
@@ -15,6 +15,13 @@ source "${SCRIPT_DIR}/common.sh" "$@"
...
+  echo -e "  ${C_YELLOW}ℹ Skipping GKE Backup Plan teardown (ENABLE_GKE_BACKUP_PLAN=${ENABLE_GKE_BACKUP_PLAN}).${C_RESET}"
+  exit 0
+fi
```
mplakhtiy, line 23: **MEDIUM — a saved `false` skips teardown of a BackupPlan that may already exist, stranding Kubernetes Secrets and PV data**

The guard reads the flag as "the feature is off", but what it actually means is "the flag is off *right now*" — which is not the same as "no BackupPlan was ever created". `vars.sh` is the only input, and `k8s-operator/scripts/README.md:20` tells operators that changing a configuration variable means editing `vars.sh` by hand.

**Failure scenario.** An operator provisions with `ENABLE_GKE_BACKUP_PLAN=true`; the plan is created with `--include-secrets --include-volume-data --no-paused` and snapshots accumulate on the cron schedule. Later they edit `vars.sh` back to `false` to stop the spend. Teardown then prints one informational line and exits 0, `teardown_01` deletes the cluster, and the BackupPlan and every snapshot survive — Backup for GKE snapshots are by design independent of the cluster's lifetime. The operator is left paying for snapshot storage and, worse, holding Kubernetes Secrets and persistent-volume contents in a project everyone believes was torn down. `provision_12` will not clean it up either, since it exits on the same flag.

Compounding it, there is no escape hatch short of editing the state file. Verified by running the script: an exported `ENABLE_GKE_BACKUP_PLAN=true` is silently overridden by the saved `false`, because `ensure_teardown_state` sources `vars.sh` *after* the environment is set — the same trap `uninstall.sh:238-241` already documents for project/cluster/region ("exporting alone is not enough ... teardown would silently act on the saved ... instead").

The classifier fix above dissolves this one too: with permission errors handled, the teardown can stay unconditional and let the `describe` decide whether there is anything to delete. If the flag guard is kept as-is, at minimum make the message a warning that names the consequence ("any BackupPlan created while the feature was enabled is left in place") and says how to force the cleanup.

mplakhtiy (commented):
Three findings — two MEDIUM anchored inline, one LOW below. The seven-line change is shellcheck-clean and does what it says; I verified the skip path by running the script.

**LOW — the script banner was not updated, so the generated catalogue omits the new skip condition**

`AGENTS.md` assigns "What each provisioning script does" to `k8s-operator/scripts/README.md`, and that table is a `<!-- BEGIN GENERATED: provisioning-steps -->` region built by `scripts/generate_docs.py:278` (`parse_script_banner`) from each script's header comment. The banner in `teardown_12_gke_backup_plan.sh:1-10` still describes only the `PRESERVE_BACKUPS` knob, so `README.md:118` continues to promise unconditional deletion — and `make docs-check` passes without noticing, because the region *is* current; it is the source that is stale.

**Failure scenario.** An operator reads the catalogue, sees "Idempotent script to delete the Google Cloud Backup for GKE BackupPlan ... Safe to run even if the backup plan was never created", and has no way to learn that a `false` in `vars.sh` turns the step into a no-op — which is exactly the surprise in the second finding. The sibling banners already set the precedent: `provision_12` says "Skipped unless ENABLE_GKE_BACKUP_PLAN=true", `provision_11` says "Skipped unless INFERENCE_REPLAY_ENABLED=true".

**Fix.** Add a line such as `# Skipped when ENABLE_GKE_BACKUP_PLAN is explicitly false; an unset value still attempts cleanup.` to the header block and run `make docs-generate`.

PR #657, opened 2026-08-12
fix(docs): stop the documentation map generating merge conflicts

bradhoekstra:
# Pull Request

## Why This Change

`docs/README.md` is this repository's most merge-conflict-prone file. At the time of writing, 16 of
53 open pull requests touch it and 12 of those are `CONFLICTING`. None of that is genuine
disagreement between branches — three self-inflicted mechanisms cause it:

1. **Prettier re-aligns Markdown table columns.** Adding one row with a cell wider than the current
   column rewrites _every_ row of that table. On merged commits: `e7192e6` changed 65 lines, 61 of
   them whitespace-only; `bd5beb4` 26 lines, 22 whitespace-only; `6338c92` 30 lines, 25
   whitespace-only. Open PR #641's entire map diff is 25 rewritten lines with **zero** content
   change. Two PRs adding rows to the same table therefore always collide, even in unrelated rows.
2. **A hand-maintained document total** — `The repository tracks **207** …` — one line every PR that
   adds or deletes a document must edit, restating a number `check_docs_map.py` already computes
   from `git ls-files`.
3. **Per-family counts** — `(12 SOPs)`, `(36 skills)`, `(31 files)`, `(6 skills)`, `(2 templates)`,
   `(9 files)`. The six open governance PRs (#589, #591–#595) each change exactly the same three
   lines, so they mutually conflict and each resolution is manual arithmetic.

The counts did buy one thing: they were the only mechanism that noticed a document **deleted** from
inside a collapsed family row's glob. Dropping them without a replacement would have left that gap
open, so this PR closes it with machine state instead.

## What Changed

**Files:**

- `docs/README.md`
- `scripts/check_docs_map.py`
- `scripts/generate_docs.py`
- `docs/family-roster.txt` (new, generated)
- `.gitattributes` (new)
- `scripts/test_docs_map.py` (new)
- `AGENTS.md`
- `Makefile`, `.github/workflows/docs-check.yml`
- `.agents/skills/review-docs-drift/SKILL.md`

**Added/Changed/Fixed:**

- **Every one of the map's 8 tables is now `<!-- prettier-ignore -->`d and compacted** to
  single-space padding (`| cell | cell |`, separators `| --- | --- |`). A row's bytes no longer
  depend on any other row's width, so adding a document is a one-line diff.
- **`docs-check-map` gained a padding check**, scanning the whole map (the identifier-sources table
  in §2 churned as hard as the inventory did). It fires on a run of spaces immediately before a `|`,
  which is where prettier's column padding always lands — so it catches every re-aligned row while
  letting an honest double space inside prose through, which matters because the tables are
  `prettier-ignore`d and nothing would normalise such a space away. It keeps a hand edit — or an
  editor-on-save with a stale config — from silently re-aligning a table again.
- **Every count is gone**, along with the two checks that verified them: the stated repository total,
  the six per-family parentheticals, and the tallies-plus-enumerations in the two family-row
  summaries (`Six are live fleet audits … Five are retained …` and the by-name skill list). Those
  last two were the ones that mattered most — the longest lines in the file, and precisely the lines
  the six governance PRs would each have had to edit. Both rows now characterise their family, which
  is what §4 asks of a family row, and both enumerations already had a canonical home in a generated
  page (governance-sops, the skill catalogue). `check_docs_map.py` still derives the totals and
  prints them; nothing is stored for a branch to disagree with. Coverage (every tracked doc matches
  some row) and existence (every path cell matches a tracked file) are unchanged.
- **New generated artifact `docs/family-roster.txt`** restores deletion detection: one sorted line
  per file behind every collapsed-family glob (7 families, 99 files today). `generate_docs.py` reads
  the globs out of the map's own §4 inventory, so a new family row is rostered automatically and the
  roster cannot cover a different set of rows than the checker honours — the two share one reader,
  `check_docs_map.inventory_tokens`, so that guarantee is structural rather than a coincidence of two
  extractions agreeing today. `main()` gained a `FILES` registry beside `BLOCKS` for whole-file
  targets; `docs-check-generated` already covers it, so no new Makefile or workflow target was
  needed, only corrected descriptions of the existing ones.
- **`.gitattributes` hands the roster to git's union merge driver**, so the roster removes the
  conflict rather than inheriting it. A sorted list still collides when two branches insert into the
  same gap even if their names are not adjacent — git has no unchanged context line to anchor on —
  and five of the governance SOPs in flight land in one such gap. Union merge keeps both sides'
  lines, needs no per-clone config, and cannot lose an insertion; anything it gets wrong (wrong order,
  or a line one side deliberately deleted) leaves the roster disagreeing with its generator, which
  `docs-check-generated` already fails on and `make docs-generate` fixes. The deletion guard
  therefore survives the union merge.
- **`scripts/test_docs_map.py` (new, 12 tests)** covers the two guards the whole change rests on.
  Both fail _green_: a padding check that stops matching lets prettier re-align the map with no
  signal, and a family-glob extraction that stops matching a row makes `docs-generate` rewrite the
  roster to the smaller set, after which `docs-check-generated` passes against the new snapshot and
  that family's deletion guard is gone. `make test-python` already discovers `scripts/test_*.py`, so
  CI picks them up with no wiring.
- **`AGENTS.md`, the `review-docs-drift` skill, and the three stale descriptions of what
  `docs-generate` does** (the generator docstring, the Makefile comment and `## help` text, the
  docs-check workflow step) now state the contract and the fourth, whole-file target. The `AGENTS.md`
  bullet is two lines — the one thing a contributor must do, plus a pointer to `docs/README.md` §5,
  which owns the rest; the earlier draft restated nine facts a third time, against `AGENTS.md`'s own
  link-don't-summarise rule.

## Why This Matters

- Two PRs that each add a document to different parts of the tree now merge cleanly. So do the six
  governance PRs, in any order — **as far as this mechanism is concerned**: they no longer touch the
  map at all, and their roster insertions union-merge (measured below, with the SOP names they
  actually use). Anything else those branches share is their own business. Conflicts are reserved for
  branches that genuinely edit the same inventory row.
- A document-adding PR's map diff collapses from "25 rewritten lines plus arithmetic" to a single
  inserted line, which is also what makes it reviewable.
- The deletion guard is now stronger than the counts it replaces: a count told you a number
  disagreed, whereas the roster names the file that vanished, in the diff, where a reviewer sees it.
- The roster deliberately lives **outside** the map. Per-file churn does not belong in the
  repository's most conflict-prone file, and with `merge=union` the roster does not conflict either —
  where it does need a hand, the hand is `make docs-generate`, not arithmetic.

## Context

- Blocks/unblocks the 16 open PRs that touch `docs/README.md`; each will need one re-resolution
  against this change, after which their map diff becomes a single added line. The six governance
  PRs (#589, #591–#595) benefit most.
- Deliberately scoped to the conflict problem and its one regression. No inventory content, summary
  prose, or row was rewritten beyond the padding change and the removal of the counts — the two
  family-row summaries were reworded only to drop their tallies and enumerations.
- Follow-up, not in this PR: `check_docs_map.py` and `generate_docs.py` are now covered where it
  counts — the padding check and the family-glob extraction — but not exhaustively. `matches`,
  `pattern_to_regex`, and `in_dot_dir` still have no tests. They had none before either, and they
  fail loudly rather than green, which is why they are follow-up.

## Testing

- `make docs-check` — all four targets green.
- `prettier --check` — green on every changed Markdown file.
- `prettier --write docs/README.md` then `git diff --stat` — **empty**. This is the real proof that
  the `<!-- prettier-ignore -->` markers took effect: prettier no longer wants to touch the file.
  Without them, `prettier --write` re-aligns 302 lines.
- Negative tests, each reverted afterwards:
  - a new untracked-then-added `docs/designs/zz-tmp.md` → `MISSING`;
  - an inventory path cell pointed at a nonexistent file → `STALE`;
  - one table row re-padded by hand → `PADDING  line 251`; after stripping the markers and running
    prettier, 143 rows;
  - `git rm --cached agents/platform/governance/inventory.md` → `STALE docs/family-roster.txt`,
    exit 1.
- Roster auto-discovery: a throwaway `docs/architecture/0*.md` family row added to §4 → its 8 members
  appeared in the roster on the next `make docs-generate`.
- **End-to-end conflict test**, the one that answers the problem statement. Two branches each adding
  a site page under `docs/site/src/content/docs/`: on `origin/main` the second merge conflicts, with
  a +47/-46 map diff; on this branch each merges cleanly at +1/-0. Two branches each adding a
  governance SOP no longer touch the map at all. A three-way merge of the result passes full
  `make docs-check` and `prettier --check`.
- **Roster conflict behaviour**, measured rather than assumed, and re-measured after review. Three
  branches off the same base, each adding one of the governance SOPs actually in flight
  (`gce_compute_fleet_sop.md`, `gcp_networking_fabric_sop.md`, `gke_runtime_telemetry_sop.md` — all
  three land in the same gap in the sorted roster). Merged pairwise into one branch: **zero
  conflicts**, in the roster or in `docs/README.md`, which no merge touched at all. All three SOPs
  present in the roster afterwards, `generate_docs.py --check` clean (so union merge produced the
  file in correct sorted order, not merely a mergeable one), and `check_docs_map.py` green at 209
  tracked docs.
- **The union driver is what does that, proven rather than assumed.** The same three-way merge of the
  roster run through `git merge-file` with and without `--union`: default text merge
  **`CONFLICTED`**, union merge clean with both insertions in sorted order. Without the
  `.gitattributes` entry this pair conflicts, which is the case review reported.
- **`scripts/test_docs_map.py` — 12 tests, green**, run the way `make test-python` invokes them
  (`cd scripts && python3 -m unittest discover -p "test_*.py"`; 17 with the existing suite in that
  directory). They cover: a compact table clean and a prettier-aligned one flagged at the right line
  numbers; a double space inside prose _not_ flagged; an indented aligned row flagged and an indented
  compact one not; the committed map clean; `family_globs` reading a glob from a cell other than the
  path column (the structural claim the shared reader now makes true) and ignoring non-path tokens;
  and the roster containing every tracked governance SOP, with the expectation built from a literal
  path prefix rather than from the glob machinery under test.
- All throwaway branches, worktrees, and files from the tests above were deleted; `git worktree list`
  and `git status` are clean.

### Live validation

Not live-tested. This change touches only repository documentation and the CI checking scripts —
nothing reaches a running kube-agents installation. There is no operator reconcile, no Deployment
env, no file or process inside an agent pod that this change can affect: `generate_docs.py` and
`check_docs_map.py` run in CI and on a developer's machine, and `docs/README.md`,
`docs/family-roster.txt`, `.gitattributes`, `AGENTS.md`, and `.agents/skills/` are not baked into any
agent image.
The mechanical gates above are the full extent of what this change can be exercised against. No test
artifacts remain: every negative test and throwaway branch listed above was reverted or deleted, and
the working tree is clean.

---

Low risk and self-contained, but it has a **one-time cost that lands on other people**: every open
PR that touches `docs/README.md` must re-resolve once against the compacted tables. They are already
conflicting, so this trades a recurring conflict for a single one. Worth telling those authors to
rebase once this merges. Rollback is a plain revert — nothing depends on the roster file existing
except `docs-check-generated`, which regenerates it.


docs/README.md
```diff
@@ -166,169 +174,206 @@ code, first check which era it belongs to:
...
+| `agents/platform/governance/*.md` | SOP playbook | Uniform cron-run governance playbooks, each with a purpose line and an execution checklist. Six are live fleet audits that emit a validated findings file and route it through the `fleet-audit` skill, which publishes it as the stream's ledger issue: compliance audit (security/RBAC posture), obtainability audit (workload reliability), security patch orchestration (upgrade readiness), fleet-wide cost analysis (observable waste), fleet consistency drift (fleet-derived baseline), stockout prevention audit (capacity obtainability and CCC resilience). Five are retained on disk but unscheduled — their watchdogs shipped disabled and were then retired from the cron roster: blueprint sync, global capacity, lifecycle deprecation, policy propagation, standardization validation. Plus `inventory.md`, the first-run environment-discovery SOP run by the bootstrap kanban card. | Fleet audits, drift reconciliation, cost, capacity, upgrades, first-run inventory | Runtime playbooks fired by the cron watchdogs (inventory: by the bootstrap scan card); summarized on the site's governance-sops page |
```
mplakhtiy: **HIGH — the map still states per-family counts, on exactly the rows the governance PRs must edit**

Section 1 now asserts:

> This file states **no document counts**, anywhere — not a repository total, not a per-directory total, not a per-family total.

The new `AGENTS.md` bullet repeats it ("never restate a count — the map states none"), and the updated `review-docs-drift` skill hardens it into a review rule: hard-coded counts *"should generally not exist anywhere, the map included; flag any the PR introduces."*

This line states two — "**Six** are live fleet audits that emit a validated findings file ... **Five** are retained on disk but unscheduled" — and follows each with a by-name enumeration of all eleven SOPs. Line 228 does the same for the platform skills family: every skill listed by name, plus "the six cron audits".

**Failure scenario.** #589 and #591–#595 each add one live fleet audit under `agents/platform/governance/` and one skill under `agents/platform/skills/`. To leave the row honest, each must change "Six" to "Seven" and insert its audit into the enumeration on this line, and insert its skill into line 228 — the same two lines, from six branches. That is the mutual conflict this PR sets out to remove, and it survives the change. The body's "Two branches each adding a governance SOP no longer touch the map at all" holds only because the throwaway SOPs used in the test left the summary prose stale; a real PR cannot.

The path-cell counts this PR removed were the mechanically checked ones. The counts it left behind are the unchecked ones, on the longest and most-edited lines in the file.

**Options.** Drop the "Six"/"Five" tallies and the by-name enumerations from the two family rows so the summaries characterise rather than enumerate — which is what §4's own new "a family row characterises its family; it is not a roster" bullet asks for. Or keep them and soften §1, `AGENTS.md`, and the skill so they claim only that the map states no *mechanically derivable* count. Either way, the body's "the six governance PRs become mergeable in any order" should come out.

docs/family-roster.txt
```diff
@@ -0,0 +1,129 @@
...
+  agents/platform/governance/fleet_consistency_drift_sop.md
+  agents/platform/governance/fleet_wide_cost_analysis_sop.md
+  agents/platform/governance/global_capacity_orchestrator_sop.md
+  agents/platform/governance/inventory.md
+  agents/platform/governance/lifecycle_deprecation_manager_sop.md
+  agents/platform/governance/obtainability_audit_sop.md
+  agents/platform/governance/policy_propagation_sop.md
+  agents/platform/governance/security_patch_orchestrator_sop.md
+  agents/platform/governance/standardization_validator_sop.md
+  agents/platform/governance/stockout_prevention_sop.md
```
mplakhtiy, line 50: **MEDIUM — for the family that motivated the change, the roster relocates the conflict rather than removing it**

The roster is a sorted list, so two branches inserting at the same point collide. Reproduced with `git merge-file` against the roster as committed: a branch adding `gce_compute_fleet_sop.md` and a branch adding `gcp_networking_fabric_sop.md` conflict, and so do `gce_compute_fleet_sop.md` and `gke_runtime_telemetry_sop.md`. The two names are not adjacent to each other, but both insert into the same gap between `fleet_wide_cost_analysis_sop.md` and `global_capacity_orchestrator_sop.md`, leaving git no unchanged context line to anchor on.

**Failure scenario.** Five of the six governance SOPs now in flight (`gce_compute_fleet_sop.md`, `gcp_networking_fabric_sop.md`, `gcp_recommender_sop.md`, `gke_ai_hpc_orchestration_sop.md`, `gke_runtime_telemetry_sop.md`) land in that one gap, and each PR also inserts into the `agents/platform/skills/*/SKILL.md` and `references/*.md` blocks. So the six PRs still conflict with each other after this merges — in the roster instead of the map. The body treats adjacency as a residual edge case; for the set it names, it is the common case.

The resolution is mechanical rather than arithmetic, so this is still a net improvement — but it can be removed outright. A one-line `.gitattributes` entry marking `docs/family-roster.txt` with `merge=union` uses git's built-in union driver, needs no per-clone config, and resolved both pairs above cleanly and in correct sorted order in my test. Where the driver does pick the wrong order, `docs-check-generated` catches it and `make docs-generate` fixes it — exactly the recovery the roster's own header describes. The repository has no `.gitattributes` today.

scripts/check_docs_map.py
```diff
@@ -111,6 +116,19 @@ def inventory_rows(text: str) -> list[str]:
...
+    Scans the whole map, not just the inventory: the identifier-sources table
+    in section 2 churns as hard as the inventory does.
+    """
+    return [
+        (number, line)
+        for number, line in enumerate(text.splitlines(), start=1)
+        if line.startswith("|") and ALIGNMENT_PADDING in line
+    ]
+
+
```
mplakhtiy, line 165: **MEDIUM — neither new guard has a test, and both fail silently when they break**

This applies to `realigned_rows` here and to `gen_family_roster` in `scripts/generate_docs.py:315-337`.

`make test-python` already discovers `scripts/test_*.py` (`Makefile:108`), CI runs it as "Run Python Unit Tests", and `scripts/test_sync_upstream_skills.py` is the precedent — so tests here are cheap and conventional.

**Failure scenario.** Two regressions ship green.

1. A later edit to `ALIGNMENT_PADDING` or to `realigned_rows` makes the padding check vacuous. Nothing fails, prettier re-aligns the map on the next hand edit, and the conflict storm this PR fixes comes back with no signal.
2. `gen_family_roster` extracts globs with its own predicate (`"*" in token` over the first cell only). If that predicate stops matching a family — a row reformatted, a glob written differently — `make docs-generate` rewrites the roster to the smaller set, `docs-check-generated` goes green against the new snapshot, and the deletion guard for that family is gone. The removed count checks were at least verified against the glob every run.

Two tests cover both: one asserting `realigned_rows` fires on a prettier-aligned row and not on a compact one, one asserting `gen_family_roster` lists every member of a known family. The body lists this as follow-up work; it is worth doing here, because the value of the whole change rests on these two guards staying live.

scripts/check_docs_map.py
```diff
@@ -111,6 +116,19 @@ def inventory_rows(text: str) -> list[str]:
...
+def realigned_rows(text: str) -> list[tuple[int, str]]:
+    """Return (line number, row) for every table row that was column-aligned.
+
+    Scans the whole map, not just the inventory: the identifier-sources table
+    in section 2 churns as hard as the inventory does.
+    """
+    return [
+        (number, line)
+        for number, line in enumerate(text.splitlines(), start=1)
+        if line.startswith("|") and ALIGNMENT_PADDING in line
```
mplakhtiy: **LOW — the padding check is a bare substring scan, so an honest double space fails the build with a misleading message**

Any two consecutive spaces anywhere in a line beginning with `|` fails the check, and there is no escape hatch — the tables are `prettier-ignore`d, so prettier will not normalise the whitespace away either.

**Failure scenario.** A contributor writes a summary cell with two spaces after a full stop, or the map documents the anti-pattern with a fenced example table (fenced blocks are not excluded from the scan). CI prints `PADDING line N: ... -- re-pad them as | cell | cell |` against a row that is already single-space padded, and the author has no way to satisfy it other than by finding the invisible character.

Prettier's alignment always puts its padding run immediately before a delimiter, so testing for two spaces followed by `|` rather than two spaces alone catches every re-aligned row while ignoring double spaces in prose.

scripts/check_docs_map.py
```diff
@@ -111,6 +116,19 @@ def inventory_rows(text: str) -> list[str]:
...
+    ]
```
mplakhtiy: **LOW — the padding check misses indented table rows that `inventory_rows` would see**

`inventory_rows` (line 110) matches on `line.strip().startswith("|")`; `realigned_rows` matches on the unstripped `line.startswith("|")`.

**Failure scenario.** A table nested under a list item in the map — none today — is re-aligned by a hand edit or an editor-on-save, the check stays green, and that table then churns on every prettier run for whoever next touches it.

Using the stripped line in both places removes the inconsistency for free.

scripts/generate_docs.py
```diff
@@ -306,12 +312,63 @@ def gen_provisioning_steps() -> str:
...
+    """
+    files = check_docs_map.tracked_docs()
+    text = check_docs_map.MAP.read_text(encoding="utf-8")
+
+    globs: set[str] = set()
+    for row in check_docs_map.inventory_rows(text):
+        path_cell = row.strip("|").split("|")[0]
+        for token in check_docs_map.TOKEN_RE.findall(path_cell):
+            if "*" in token and check_docs_map.looks_like_doc_path(token):
+                globs.add(token)
```
mplakhtiy: **LOW — the roster re-implements the checker's glob extraction instead of sharing it, so the docstring's guarantee is not enforced**

The docstring claims the roster "can never cover a different set of rows than `check_docs_map.py` honours". It can: `check_docs_map.main` accepts a coverage glob from *any* cell of an inventory row (`for cell_index, cell in enumerate(cells)`), while `gen_family_roster` reads only `row.strip("|").split("|")[0]`.

**Failure scenario.** A future row whose glob sits in the "Key topics" cell rather than the path column satisfies coverage but is never rostered, so a deletion inside it is invisible to both checks — the precise gap the roster exists to close.

Nothing diverges today: I ran the two extractions over the current map and every glob lives in the path column. Lifting a `family_globs(text)` helper into `check_docs_map.py` and calling it from both makes the docstring's claim structurally true instead of coincidentally true.

AGENTS.md
```diff
@@ -71,6 +72,15 @@ Rules:
...
   `k8s-operator/scripts/common.sh`, the Go version in `k8s-operator/go.mod`.
+- **Add a document to the map (`docs/README.md`) with one line, and change nothing else there.**
+  Its tables are `<!-- prettier-ignore -->`d and padded with single spaces on purpose: prettier
+  aligns table columns to the widest cell, so a re-aligned table rewrites every row and collides
+  with every other open pull request that adds a row. Write rows in the compact
+  `| cell | cell |` form, never re-align, and never restate a count — the map states none, and
+  `docs-check-map` fails on a row that grows a double space. A file that lands inside a collapsed
+  family row's glob (a new skill, SOP, or reference) needs no map edit at all — run
+  `make docs-generate` instead, which re-snapshots `docs/family-roster.txt`. That roster is what
+  catches a family member being _deleted_, which no glob can see.
```
mplakhtiy: **LOW — this bullet restates `docs/README.md` §4 and §5 almost verbatim**

`AGENTS.md` itself says: *"**Link rather than summarise** when another page already owns the topic. If you must summarise, say which page is canonical."* `docs/README.md` §5 is titled "Keeping this map fresh" and owns this contract.

After this PR the same nine facts — prettier-ignore, single-space padding, no counts, family globs need no map edit, the roster catches deletions — are stated in three places: here, `docs/README.md:184-193` plus `:355-376`, and `.agents/skills/review-docs-drift/SKILL.md:59-60`.

**Failure scenario.** The next change to the contract updates one or two of the three, and a contributor following `AGENTS.md` gets the old rule.

A two-line bullet stating the one thing a contributor must do (one line, never re-align) and pointing at `docs/README.md` §5 for the rest costs nothing and leaves one home per fact — which is the rule this section of `AGENTS.md` exists to state.

mplakhtiy (commented):
Reviewed at `8a6e4fe0`, merged against `main` @ `dc695ce3` (clean, no conflicts) so this covers the PR as it would merge. Eight findings — one HIGH, two MEDIUM, five LOW. Seven are anchored inline; the eighth is below.

**LOW — the generator's own docstring, `make help`, and the CI step comment still describe three marker-delimited regions**

The PR taught `docs/README.md` §2 and `AGENTS.md` about the fourth, whole-file target but left three other descriptions behind:

- `scripts/generate_docs.py:1-31` — the module docstring still says "Each generated region is delimited in its target file by ... Everything outside those markers is hand-written and is never touched". `docs/family-roster.txt` has no markers and is written whole.
- `Makefile:167-169` — the comment still names only "cron jobs, the skill catalogue, the provisioning steps", and the help text reads `## Regenerate the <!-- BEGIN GENERATED --> doc regions from their sources.`
- `.github/workflows/docs-check.yml:21-22` — same three-region description.

`AGENTS.md` makes the help text canonical — *"Do not restate the `make` targets. `make help` prints them from the Makefile"* — so `make help` is now the wrong description of what `docs-generate` does.

---

*This review was performed with the help of Claude.*

toshiowang (approved):

docs/README.md
bradhoekstra: Fixed in `ba0235e`, taking the first option — the tallies and the by-name enumerations are gone from both family rows.

You're right that these were the ones that mattered. The parentheticals I removed were checked; these were unchecked, on the two longest lines in the file, and were exactly what #589 and #591–#595 would each have had to edit. The test in the body did not catch it because throwaway SOPs leave the prose stale for free, which a real PR cannot — that's a fair hit on the test, not just on the rows.

Both rows now characterise rather than enumerate, which is what §4's own new bullet asks for, and both enumerations already had a canonical home in a generated page — the site's governance-sops page and the generated skill catalogue — so nothing is lost, only de-duplicated. The governance row points at governance-sops for which SOPs are live. The skills row still names `fleet-audit` and `submit-suggestion`, but as the two structurally special members (the shared audit harness, the GitOps write path) rather than as a roster; a new ordinary skill does not touch the line. While doing this I noticed I'd introduced a fresh "Two are structurally special" count in the first pass and reworded it out.

The body's "mergeable in any order" claim is now qualified to "as far as this mechanism is concerned", and backed by a merge test using the SOP names those PRs actually add — see the reply on the roster thread.

docs/family-roster.txt
bradhoekstra: Fixed in `ba0235e` — added `.gitattributes` with `docs/family-roster.txt merge=union`, as you suggested. The repository had no `.gitattributes`; this is the whole file plus a comment explaining why.

I reproduced your case before and after rather than taking it on trust:

- Three branches off one base, each adding one of the SOPs actually in flight — `gce_compute_fleet_sop.md`, `gcp_networking_fabric_sop.md`, `gke_runtime_telemetry_sop.md`, all three landing in the gap you identified. Merged pairwise into one branch: **zero conflicts**, in the roster or anywhere else. `docs/README.md` was not touched by any merge. All three SOPs present afterwards, `generate_docs.py --check` clean — so union merge produced the file in correct sorted order, not merely a mergeable one — and `check_docs_map.py` green.
- Isolating the driver, since a clean merge alone would not prove the mechanism: the same three-way merge through `git merge-file` with and without `--union`. Default text merge **`CONFLICTED`**; union merge clean, both insertions sorted.

One property worth recording, since it decides whether this is safe: union merge also keeps a line one side *deleted*. That does not weaken the deletion guard — it just means the roster stops matching its generator, `docs-check-generated` fails, and `make docs-generate` writes the correct file. Same recovery, still caught. That reasoning is in the `.gitattributes` comment so the next person does not have to re-derive it.

scripts/check_docs_map.py
bradhoekstra: Fixed in `ba0235e` — `scripts/test_docs_map.py`, 12 tests, green. Agreed on the reasoning: "the value of the whole change rests on these two guards staying live", and both of your regressions ship *green*, which is the argument for testing them here rather than as follow-up.

`realigned_rows`: compact table clean; prettier-aligned table flagged at the right line numbers; a double space inside prose not flagged (the behaviour from your next finding); an indented aligned row flagged and an indented compact one not; and the committed map clean.

`gen_family_roster` / `family_globs`: a glob read from a cell other than the path column (the structural claim from your `family_globs` finding, now enforced); non-path tokens ignored; the real map's families found; every family glob the map declares present in the roster; and every tracked governance SOP listed. That last expectation is built from a literal path prefix rather than from `matches()`, so it does not become tautological against the code under test — and it asserts the family is non-empty first, so a broken fixture cannot pass as a green test.

`make test-python` discovers `scripts/test_*.py` already, so no wiring. Run as it invokes them: 17 tests in `scripts/`, OK.

Left as follow-up in the body, now with a reason rather than a shrug: `matches`, `pattern_to_regex`, and `in_dot_dir` are still untested, but they fail loudly rather than green.

scripts/check_docs_map.py
bradhoekstra: Fixed in `ba0235e` — `ALIGNMENT_PADDING` is now `"  |"` rather than `"  "`, so the check fires on a padding run immediately before a delimiter, which is where prettier always puts it.

The point that made this worth fixing rather than living with is the one about there being no escape hatch: the tables are `prettier-ignore`d, so nothing would normalise an honest double space away and the author would have had no way to satisfy the check other than hunting an invisible character. A build failure a contributor cannot act on is worse than the false negative it prevents.

Verified the delimiter anchor does not weaken detection: both centre- and left-aligned prettier output still pad immediately before the `|`, and the aligned-row test in the new test file asserts they are caught. There is a test for the prose case too.

scripts/check_docs_map.py
bradhoekstra: Fixed in `ba0235e` — `realigned_rows` now matches on the stripped line, same as `inventory_rows`, and reports the stripped row. Free consistency, as you said, and there is a test for both the indented aligned row and the indented compact one so the two functions do not drift apart again.

scripts/generate_docs.py
bradhoekstra: Fixed in `ba0235e`, and thank you for checking whether the two extractions currently diverge rather than only that they could — that is what made the fix obviously worth doing while nothing is broken.

Rather than a `family_globs` helper alone, I lifted the token reader itself: `check_docs_map.inventory_tokens(text)` yields `(cell_index, token)` for every path-like token in the inventory, and it now has exactly two consumers — `main()`'s coverage/existence loop and `family_globs`, which `gen_family_roster` calls. So there is one definition of where a path token may sit, and the docstring's claim is structural rather than coincidental. The generator's docstring says which reader it shares and why.

Test coverage for the specific gap: a row whose glob sits in the third cell is now asserted to be rostered.

AGENTS.md
bradhoekstra: Fixed in `ba0235e` — the bullet is down to two sentences: write the row compact, never re-align, and `docs/README.md` §5 owns the rest, including why a file inside an existing family needs no map edit.

You're quoting the right rule back at me. Nine facts in three places was going to rot in exactly the way you describe, and `AGENTS.md` is the file that states link-don't-summarise.

I left the `review-docs-drift` skill's fuller version in place. It is a review checklist rather than a contributor summary — it needs to name each thing to look for at the point of looking — and its facts are now consistent with the rows after the counts came out. Happy to trim it to a pointer as well if you would rather have two homes than three.

bradhoekstra:
Thanks — all eight addressed in `ba0235e`, replies in each thread. Answering the one that had no line to anchor to:

**LOW — the generator's own docstring, `make help`, and the CI step comment still describe three marker-delimited regions.** Fixed, all three, plus a fourth I'd missed:

- `scripts/generate_docs.py` — the docstring now names the two kinds of target up front (`BLOCKS`, a region spliced into a hand-written document; `FILES`, a whole file written verbatim), and the "everything outside those markers is hand-written" sentence now says a whole-file target has no such boundary because nothing in it is hand-written.
- `Makefile` — the comment names the roster, and the help text is now `## Regenerate the generated doc regions and files from their sources.`
- `.github/workflows/docs-check.yml` — the step comment covers the roster.
- Also `ROSTER_HEADER` in the generator, which claimed a roster conflict is resolved by re-running the generator; with `merge=union` that is now the fallback rather than the normal path, so it says so and points at `.gitattributes`.

Your reasoning is the part I'd want on the record: `AGENTS.md` makes `make help` canonical for what a target does, so leaving it describing three regions was not a stale comment but a wrong canonical statement. The PR body's claim that "no Makefile or workflow change was needed" was the tell, and it now reads "no new Makefile or workflow *target* was needed, only corrected descriptions of the existing ones."

Two notes on things I did **not** change:

- `docs/README.md:230` (the `agentplugins/README.md` row) says "the two that ship here" and "the two constraints". Those are counts, but of plugins rather than of documents, on a single-document row that a third plugin would need reworded anyway. §1's claim is specifically about document counts. Flagging it rather than quietly widening scope — say the word if you want it gone.
- The `review-docs-drift` skill keeps its fuller statement of the contract; reasoning in the thread on the `AGENTS.md` finding.

`make docs-check` green (four targets), `prettier --check` green, `prettier --write docs/README.md` still a no-op, `scripts/` unit tests green at 17. Testing section of the body updated with the re-measured merge results.

mplakhtiy (approved):

Issue #100, opened 2026-06-05
Implement `notify_user` tool for devteam agent


dshnayder:
Current implementation uses `curl` command to send proactive notifications to user via chat. Need to implement a tool `send_message_to_chat` in the DevTeam and Operator agents. See discussion: https://github.com/gke-labs/kube-agents/pull/93#discussion_r3365494289

dshnayder:
See https://github.com/gke-labs/kube-agents/pull/138 for examples of async notifications to chat group.

PR #670, opened 2026-08-12
fix(fleet-audit): re-apply remediation labels from the path that sees open PRs

adamparco:
# Pull Request

## Why This Change

A remediation pull request gets its four labels from `gh pr create` and never again. Two things then rot:

- **A reviewer strips them while triaging.** That is what happened to pull requests 34, 35 and 36 in the reference installation. No later run put them back, so three pull requests the audit still owned stopped appearing under `agent:audit` for good.
- **`severity:` freezes.** It is recomputed every run from the findings the group currently holds, so a group that escalates from minor to critical keeps the label it opened with — on the one field triage sorts by.

The obvious place to fix this is `open_remediation_pr`'s refresh branch, which already rewrites the title and body of an existing pull request. **That branch is unreachable with an open pull request**, so a fix placed there does nothing at all. Two independent guards keep it out:

- `promotion_candidates` diverts a finding whose pull request is `OPEN` into `already_open` and never promotes it, and only promoted findings reach `open_remediation_pr`.
- `reconcile_remediation_prs` maps **every** finding in a remediation group to that group's one pull request, so a newly-appeared sibling finding cannot drag the group in through the side door either.

I found this the way you would want to: an earlier revision of this change put the call in the refresh branch, and the live test showed the labels staying broken while the run reported success.

## What Changed

**Files:**

- `agents/platform/skills/fleet-audit/scripts/audit_report.py`
- `agents/platform/skills/fleet-audit/SKILL.md`
- `docs/designs/fleet-audit-issue-ledger.md`
- `docs/site/src/content/docs/concepts/declarative-workflow.md`

**Added/Changed/Fixed:**

- Added `sync_open_remediation_labels`, called from both places that reconcile remediation pull requests — `handle_remediate` and the scheduled `finish` — and covering **every** open remediation pull request the run found, not only the ones a `/remediate` named. `already_open` holds requested ids and nothing else, and the pull requests this exists for are precisely the ones nobody is asking about; a stripped label has to heal on the next scheduled audit or it does not heal.
- It touches **labels and nothing else**: no force-push, no rewritten body. Leaving an open pull request alone is a deliberate promise that a reviewer's commits stay where they are (design doc §"`pr-open` is not refreshed"), and re-labelling keeps that promise while still repairing the field triage sorts on. A test asserts no `git` call and no `--body-file`/`--title` is issued.
- One `gh` call per pull request rather than per finding, since a group's findings all resolve to the same one. A no-op when the labels are already right, so a steady fleet changes nothing and fires no webhooks.
- `sync_remediation_labels` keeps its existing call in the refresh branch and gains an accurate docstring — it previously blamed the refresh for the drift, which cannot be what happened.
- `TestOpenRefreshIsUnreachable` pins the reachability argument, so a later change that makes the refresh branch live fails here rather than quietly leaving two callers doing the same work.
- **Also fixed here, by request: `audit:stale-closed` could never be created.** Its description was 108 characters against GitHub's 100-character limit, so `gh label create` answered `HTTP 422` on every run since the label was introduced. `ensure_labels` passes `check=False`, so nothing surfaced. This is the failure the comment beside that label already warns about, live the entire time: the label is what `pr_closed_by_harness` reads to tell a harness close from a human one, so **every harness close has been landing unlabelled and reading as a human rejection, and no finding has ever been re-proposed after its first quiet day** — nor could any remediation pull request be auto-closed as stale, since the code refuses to close one whose label did not stick. `TestLabelDescriptions` now measures every description `ensure_labels` emits, across every audit stream rather than one, because the per-stream description interpolates `audit_name` and a future audit with a long title would break only its own.

## Why This Matters

- Restores the audit's own index of what it owns. A pull request missing `agent:audit` is invisible to every view and query built on that label, including the `github-issue-resolver` exclusion.
- Un-freezes `severity:`, which is what a human sorts a triage queue by.
- Removes a fix that looked correct, passed unit tests, and did nothing in production — with a test that stops it coming back.

## Context

- Found while investigating why one pull request in a downstream installation carried agent-applied labels when its siblings did not.
- **Out of scope, worth knowing** — one pre-existing bug this does not touch. **The lookup's blind spot:** `list_remediation_prs` filters on `audit:<audit-id>` **and** `audit:remediation`. A pull request that loses either one is invisible to the harness entirely: it cannot be relabelled, and the next run tries to *create* a pull request on a branch that already has an open one, which fails. So this repairs a stripped `agent:audit` or a stale `severity:`, but cannot rescue a pull request stripped of a lookup label. Fixing that needs a head-based lookup that does not depend on labels.

## Testing

- `python3 -m unittest test_audit_report` — **501 passed**, including 11 new (8 in `TestSyncOpenRemediationLabels`, 3 in `TestOpenRefreshIsUnreachable`).
- `make docs-check` — clean (generated regions current, links resolve, terminology, map inventory).
- `prettier --check` on both changed Markdown files — clean.
- `review-docs-drift` run against the branch diff, twice. The first pass caught one Blocking finding — `docs/designs/fleet-audit-issue-ledger.md` said an open remediation PR "is left alone" — and I fixed that bullet and moved on, which was the wrong stopping point: **the same claim survived in four more places**, three of them in `fleet-audit/SKILL.md`, which the first run never examined because the file was not yet in the diff. The bot's strict pass found those three; re-running the sweep across all five files found the last two, both halves of pairs I had amended on one side only, so the design doc and `declarative-workflow.md` each contradicted themselves a dozen-odd lines apart. All five now read the same way, and `SKILL.md`'s `/remediate` answer contract — instruction the agent follows, not description — no longer tells it to report a pull request as untouched on a run that relabelled it.

### Live validation

**Install:** `platform-agent-host` (GKE, `us-east4`, project `adamparco-kage`), pod `platform-agent-gateway-59c95b7c7c-hz78c`, image `platform-agent:dev-3282973-0812-134624`, GitOps repo `gke-agentic/adamparco-infra`, audit stream `obtainability-audit`. Everything below ran inside that pod against real GitHub.

**Setting up a specimen.** Auto-merge lands an agent-authored pull request in this repo within seconds, so no remediation pull request had ever stayed open long enough to observe. I added a `no-auto-merge` opt-out to the downstream repo first (`gke-agentic/adamparco-infra` #99 and #101) — the harness cannot pass `--label` to its own `gh pr create`, so the guard re-reads the live labels at runner startup. With that in place the harness opened **#103** and it stayed open for the full 60s poll; the identical earlier run, #100, was auto-merged 13s in. That is the mechanism check on the opt-out itself.

**Breaking it the way a reviewer does.** On #103, by hand: removed `agent:audit`, removed `severity:minor`, added `severity:critical`. Kept `audit:obtainability-audit` + `audit:remediation` (the lookup pair — removing those tests nothing, per the blind spot above) and `no-auto-merge`.

**A/B, both call paths, deployed build as the control.** The control is the image's own `audit_report.py`, which carries the earlier revision of this fix — the one in the unreachable branch. The candidate ran as a sibling file in the same directory so path-relative behaviour was identical.

| Path | Run | `gh pr edit --add-label` calls | Labels after |
| --- | --- | --- | --- |
| `remediate --finding <id>` | control (deployed) | **0** | still broken |
| `remediate --finding <id>` | candidate | 1 | **repaired** |
| `finish` (the scheduled cron path) | control (deployed) | **0** | still broken |
| `finish` (the scheduled cron path) | candidate | 1 | **repaired** |

"Repaired" means `agent:audit` back and `severity:minor` back with `severity:critical` removed — `severity:critical` was chosen precisely because it is not the default the run would produce, so a coincidental match could not be mistaken for a fix. `no-auto-merge` was untouched and the pull request stayed `OPEN` throughout. Both control runs reported `status: REMEDIATED` / `UPDATED` and `already_open` — succeeding while changing nothing, which is exactly how this went unnoticed.

**The no-collateral-damage claim, measured.** Head commit `9dd49d0a6` and body length `2105` were identical before and after every one of the four runs. Nothing was force-pushed and no body was rewritten.

**The unreachability claim, measured rather than argued.** `/remediate` on the finding owning #103 returned `already_open` and never called `open_remediation_pr`. I then added a *second* finding on the same manifest path — no pull request of its own — and requested that one: also `already_open`, because it reconciles to #103 through the shared group branch. Both cases are now unit tests.

**The `audit:stale-closed` fix, A/B'd from inside the pod.** `gh label list` confirmed the label genuinely did not exist in `gke-agentic/adamparco-infra` — the predicted end state of a create that has 422'd since the label was introduced. Creating it with the old 108-character description returned `HTTP 422: Validation Failed … description is too long (maximum is 100 characters)`; creating it with the new 90-character one succeeded, and `gh label list` then showed it. The old and new strings differ only in length, so the length is what the 422 was about. **Left behind deliberately:** the label now exists in that repo. That is the state the fix produces, not an artifact — but it does mean the deployed harness's `--add-label` calls will now resolve there, so the repo is no longer a clean reproduction of the bug.

**Changed after the live runs, and therefore not covered by them:** the `/remediate` acknowledgement string in `audit_report.py` said the pull request "was left untouched rather than force-pushed over", which the labels re-assertion makes false in the operator's own comment. It now says its labels were re-asserted and its diff left untouched. That is a one-sentence change to a ledger comment, exercised by unit tests but not re-run against the installation — the four A/B runs above predate it, and nothing else in this change moved.

**Not covered:** the multi-cluster fan-out (one installation, one audit stream), and the blind-spot case above, which is unfixable without the separate lookup change. I did not test a group whose severity genuinely escalates between two real audit runs — the escalation is covered by unit test only, since manufacturing it live needs a second real finding on the same file.

**Artifacts cleaned up:** #103 closed with an explanatory comment and its branch deleted; the candidate and control scripts and all scratch test files removed from the pod (the skills directory is back to its two shipped files, `audit_report.py` at the image's `180027541e…`); the GitOps workspace restored from `origin/main` and `git status` clean. The live-test fixture `bad-app-cron-test` had been perturbed earlier when an auto-merge pushed a liveness probe onto it; that was reverted (`adamparco-infra` #102) and the deployment is confirmed probe-free and `1/1 Running`, so the audit finding reproduces again. Left behind deliberately: the `no-auto-merge` label and workflow in the downstream repo, which are a wanted feature, not an artifact.

---

Behavioural change is confined to labels on pull requests the harness already owns; the worst case of a bug here is a wrong label, not a wrong merge. Cost is one `gh` call per open remediation pull request per run. Nothing else in the run's ordering or output changes.


adamparco:
Answering the 🟡 Low finding from the second review ("SKILL.md still tells the agent an open remediation pull request is never touched"): fixed in 28ab8b0, which corrected every place the claim survived — the three spots the finding named and two more the same claim reached:

- the `pr-open` state-table row now reads "**Labels re-asserted.** The pull request itself is untouched";
- the "`pr-open` is not refreshed" bullet now scopes the promise to the diff and carries the same labels-are-the-single-exception language as the design doc;
- the `/remediate` answer contract now reports "already open" **with its labels re-asserted and its diff untouched** — and the acknowledgement string the harness actually emits (`_remediation_outcomes` in `audit_report.py`) was updated to match, so an operator is no longer told a pull request was left untouched on a run that re-labelled it;
- the acknowledgement sentence in `docs/site/src/content/docs/concepts/declarative-workflow.md` was aligned as well.

The strict `/review` that followed came back with no findings.

bradhoekstra (approved):

PR #674, opened 2026-08-13
feat(security): enforce a read-only root filesystem on the agent containers

AntonTyb:
## Summary

The operator builds five containers into the agent Pod, and three of them — the agent itself, the dashboard, and the log shipper — were running with a writable root filesystem. This sets `readOnlyRootFilesystem` on all five, gives the two agent containers a bounded 2Gi `emptyDir` at `/tmp` because the entrypoint genuinely needs one writable path, and guards against the mount-path collision that would otherwise break any install whose CR already claims `/tmp`. The hardening block itself is extracted into one helper and asserted by a test that walks every container, so the drift that caused this gap cannot recur silently.

## Why This Change

The `platform-agent`, `platform-agent-dashboard`, and `fluent-bit` containers run with a writable root filesystem. A compromised agent process — this is a workload that executes LLM-driven, prompt-influenced tasks — can therefore modify binaries in the image layer and persist that modification across a container restart. The credential proxy and the sandbox credential cleanup container already set `readOnlyRootFilesystem: true`; these three were the remaining gap in the Pod.

This addresses security finding **RUNTIME-007**.

## What Changed

**Files:**

- `k8s-operator/internal/controller/platformagent_manifests.go`
- `k8s-operator/internal/controller/platformagent_manifests_test.go`
- `k8s-operator/internal/testing/testdata/platform/expected/platformagent-telemetry.yaml`
- `.github/workflows/docker-build.yml`
- `deploy/docker/Dockerfile` (comment only)
- `scripts/check_image_layers.py` (comment only)
- `docs/credential-isolation-design.md`
- `docs/site/src/content/docs/deploy/docker-images.md`

**Added/Changed/Fixed:**

- Extracted `hardenedSecurityContext()` and called it from all five containers the operator builds. The three fields — `AllowPrivilegeEscalation: false`, `ReadOnlyRootFilesystem: true`, `Capabilities.Drop: ["ALL"]` — were previously written out five times, and three of the five had drifted to two of them. That drift *is* RUNTIME-007. One helper means the next container added gets the block by construction rather than by someone remembering.
- Added a `tmp-scratch` `emptyDir`, mounted at `/tmp` on the two agent containers. `deploy/shared/docker-entrypoint.sh` runs four hermes invocations with `HOME=/tmp` (lines 640, 690, 713, 873) before the gateway starts; those are the only writes that leave the PVC, so `/tmp` is the one path that must stay writable.
- Bounded that volume with a `sizeLimit` of `2Gi`, matching `credential-proxy-tmp` and the other scratch `emptyDir`s the operator creates.
- Dropped the operator's `/tmp` mount, on both agent containers, when the CR already claims that path through `spec.deployment.storages` or `extraVolumeMounts`. Those are appended to the default mounts without deduplication, and two mounts on one `mountPath` make the Deployment unappliable — so without the guard a CR that mounted its own scratch space at `/tmp`, which is exactly what someone would have done while the root filesystem was still writable, would stop reconciling on upgrade. The CR's mount wins.
- Left `fluent-bit` without a `/tmp` mount, deliberately — see below.
- Added a CI step running the existing in-image entrypoint gate check a second time, as a container under `docker run --read-only --tmpfs /tmp`.
- Corrected three comments the change invalidated, none of them in code it otherwise touches: `scripts/check_image_layers.py` and `deploy/docker/Dockerfile` both said the `entrypoint-gate-test` stage is never loaded by a daemon, which the new CI step makes false, and the new step's own `--tmpfs` comment treated one 2g ceiling as modelling two pod volumes that have separate budgets. Found by the adversarial pass — see Self-Review.
- Updated the "Kubernetes Details" bullet in `docs/credential-isolation-design.md`, which previously singled out the credential sidecar as the container with a read-only root, and the entrypoint-gate section of `docs/site/src/content/docs/deploy/docker-images.md` for the new CI step.

## Why This Matters

- Blocks runtime tampering with the image layer — binary modification and persistence — in the container that executes prompt-influenced work.
- The two agent containers share one scratch volume rather than getting one each. That is not a new channel between them: they already run the same image against the same data PVC, so they are one trust domain either way.
- `fluent-bit` gets no `/tmp` on purpose. Its config buffers in memory (`Mem_Buf_Limit`, no `storage.path`), keeps its tail database on the `fluent-bit-state` volume, and outputs to stdout, so it writes nothing to the root filesystem. Handing it the agent's scratch volume would only give an LLM-driven container a path into the log shipper. This was the riskiest claim in the PR and is now the one with direct evidence — see Live validation.
- The `sizeLimit` makes the failure attributable. Unbounded, this is now the *only* writable path outside the PVC for the agent containers, so all scratch traffic that used to spread across the container's writable layer concentrates here — and a runaway write becomes node-level ephemeral-storage pressure, where the kubelet picks a victim that need not be this pod. With the limit it is this pod, at a stated threshold, for a stated reason. It is a pod eviction either way: enforcing the limit as an in-container `ENOSPC` requires the alpha `LocalStorageCapacityIsolationFSQuotaMonitoring` feature gate, which GKE does not enable.
- The hardening is now an asserted invariant rather than a convention. `TestEveryContainerHasAHardenedSecurityContext` walks every init container and container in the built Pod, in both the dashboard-enabled and dashboard-disabled shapes, and fails on a container it does not recognise — so adding a sixth container without the block is a test failure, not a silent regression. This is @mplakhtiy's review finding from the first round, adopted.

## Context

Re-creation of the read-only-rootfs half of **#496** (Buganizer task 537148050), rebuilt on current `main` rather than rebased: that branch dated from July and its `platformagent_manifests.go` base had grown by roughly 900 lines. #496 has since been closed and retitled to its other half — defaulting gVisor on (RUNTIME-002) — so there is no longer a coordination question between the two.

Deliberately **not** included:

- **The `ENABLE_GVISOR` default flip (RUNTIME-002).** Carries its own confirmed High finding and now has its own issue.
- **RUNTIME-006**, which was a verification with no code change.

## Testing

- `go test ./...` in `k8s-operator/` — all packages pass, three consecutive `-count=1` runs.
- `go build ./...` and `go vet ./...` clean; `gofmt -l` reports nothing new (`manifest_helpers.go` is unformatted on `main` already and is untouched here).
- `TestEveryContainerHasAHardenedSecurityContext` — the invariant walk described above.
- `TestCRSuppliedSidecarsAreNotHardenedByTheOperator` — states the limit of that invariant explicitly: containers supplied through `spec.deployment.sidecars`/`initContainers` pass through as written. The webhook validates `privileged`, `allowPrivilegeEscalation`, `runAsUser=0` and `capabilities.add` on them, but not `readOnlyRootFilesystem`, so a CR can still add a writable container to this Pod. Documented rather than changed.
- Existing by-name assertions kept: `ReadOnlyRootFilesystem` on each of the three containers, the `/tmp` mount and its path, the `tmp-scratch` volume and its `2Gi` sizeLimit, and a negative assertion that `fluent-bit` does **not** mount the agent's scratch volume. The walk says "all of them"; these say which.
- Regression tests for the mount collision: a CR claiming `/tmp` through `extraVolumeMounts` (with and without a trailing slash) or through `storages` yields no duplicate `mountPath` in **any** container or initContainer, and the CR's own volume at `/tmp` on each container the input reaches; a CR mounting somewhere else keeps `tmp-scratch`. The uniqueness assertion walks every container rather than picking one by name, because the API server applies its check per container — a single-container assertion passes while the Deployment it describes is still rejected.
- **Each guard was checked by breaking it.** Reverting the dashboard `/tmp` guard fails with `duplicate mountPath "/tmp", from volumes "tmp-scratch" and "my-tmp"`; unhardening `fluent-bit` fails with `ReadOnlyRootFilesystem is not true` and `capabilities do not drop ALL, got nil`; adding an unrecognised container fails with `not in operatorBuiltContainers: if the operator grew a container, give it hardenedSecurityContext() and add it to that list`. A guard that cannot be seen to fail is not a guard.
- `make docs-check` passes — generated regions current, 260 links resolve, terminology matches, map inventory covers all 230 tracked docs. `make docs-generate` leaves the tree clean.
- Golden manifests were hand-edited rather than taken from `go test -update`. The `-update` writer emits raw Go YAML with non-indented sequences, which reformats all ~1400 lines of each golden and would fail the Prettier gate; the insertions were applied in the committed style instead, and the golden tests then pass against the generator.
- **`prettier --check` could not be run locally** — `npm install` and `npx` both fail in this environment behind an authenticated corporate mirror, and no `prettier` binary exists on the machine; AGENTS.md notes this as a known cause of the step being skipped. The insertions were kept byte-identical in style to the adjacent `credential-proxy-tmp` block and to the surrounding Markdown instead. **CI's `prettier` job passes on this branch**, so the gate is satisfied — just not before the push.
- **CI is green on this head** apart from `pull-kube-agents-smoke-test`, which is currently failing repository-wide and not caused by this change. That includes `build` (5m14s), which is the job carrying the new read-only Docker step, and `Run Controller Tests`.
- **Not a finding, recorded because it will be seen:** `go test ./internal/testing/... -race` reports a data race between parallel golden subtests sharing one `runtime.Scheme`, which the controller-runtime fake client mutates via `AddKnownTypeWithName`. It reproduces on unmodified `main` at `fe08297f` (3–10 reports per run) and occasionally surfaces without `-race` as a map-corruption panic in the golden suite. Pre-existing, in the test harness, unrelated to this change — flagging it rather than fixing it here.

### Live validation

Two runs, at different commits, covering different things. Neither supersedes the other.

**1. @bradhoekstra, 2026-08-13, at `4a3b628` (pre-rebase) — [full report](https://github.com/gke-labs/kube-agents/pull/674#issuecomment-5283865862).** The deeper of the two, on `kube-agents-cluster` (`bhoekstra-gkedemos`) with a live LLM: 67 skills loading, a tool sweep (`kubectl`, `gcloud`, `git`, `python3`, `node`, `npm`, `uv`, `yq`, `helm`) under a read-only root, and two end-to-end LLM-driven agent tasks including a deliberately write-heavy one. Peak `/tmp` usage was **28K** against the 2Gi ceiling. It also exercised the `/tmp` collision guard live through `extraVolumeMounts` with the dashboard enabled, and confirmed the un-guarded shape is rejected by the API server with `mountPath must be unique`. That run answers the two questions the original PR body left for a reviewer with a cluster, and it is the evidence for "no skill writes outside `/tmp` and the PVC" and "2Gi is a sane ceiling" — my run below does not re-cover either.

Its commit predates the rebase. What it establishes is structural and behavioural, and the refactor since is a pure extraction of three fields that were already there, so I did not re-run it; run 2 re-confirms the structure at the current head.

**One finding from it, still open:** when a CR claims `/tmp` *and* the dashboard is disabled, the `tmp-scratch` volume stays declared in the Pod spec with no container mounting it. Dangling but harmless — the kubelet does not fault on it. Left as-is for now; flagging rather than burying it.

**2. Mine, at the rebased head.** Narrower, and its purpose is to confirm the rebase did not drop a hunk and that `fluent-bit` — the container the first run covered structurally but not under sustained load — survives a read-only root.

Live-tested on **kagent1** (`atyb-kagent-01-sandbox-708646`, `us-central1`, sandbox tier), against install `PlatformAgent/platform-agent` in `kubeagents-system`, agent image tag `dev-feature_design-537148878`. There is no container runtime on the workstation, so the operator could not be built into an image; it was run out-of-cluster instead per `k8s-operator/README.md`, with the in-cluster `kubeagents-controller-manager` scaled to 0 so the two would not fight over the CR, and `DefaultPlatformAgentVersion` pinned to the deployed tag so the agent image would not roll as a side effect. The live-test lease was held for the duration.

**Before.** `readOnlyRootFilesystem` was `true` on `sandbox-credential-cleanup` and `envoy-credential-proxy` and unset on `platform-agent`, `platform-agent-dashboard` and `fluent-bit`; no `tmp-scratch` volume; pod `platform-agent-gateway-576dc64b76-mgn78` 5/5 Running; CR `Ready=True`.

**After reconciling with this branch.** All five containers came back with `readOnlyRootFilesystem=true`, `allowPrivilegeEscalation=false`, `capabilities.drop=[ALL]`. The `tmp-scratch` volume appeared as `{"emptyDir": {"sizeLimit": "2Gi"}}`, mounted at `/tmp` on `platform-agent` and `platform-agent-dashboard` and on neither `fluent-bit` nor the credential proxy, which keeps its own `credential-proxy-tmp`.

**The mechanism, not a coincidence.** Inside both agent containers:

```
$ touch /pr674-probe        → touch: cannot touch '/pr674-probe': Read-only file system
$ touch /etc/pr674-probe    → touch: cannot touch '/etc/pr674-probe': Read-only file system
$ echo hello > /tmp/pr674-probe && cat /tmp/pr674-probe   → hello
$ grep ' / '    /proc/mounts → overlay / overlay ro,relatime,...
$ grep ' /tmp ' /proc/mounts → /dev/sda1 /tmp ext4 rw,relatime,commit=30
```

Then the revert: the in-cluster operator was scaled back to 1, and it put the Deployment back to `readOnlyRootFilesystem` unset on those three containers with no `tmp-scratch`. The same `touch /` in the same container now fails with `Permission denied` rather than `Read-only file system` — ordinary non-root permissions — and `/proc/mounts` shows `/` as `rw`. Same command, different errno, filesystem flipped `ro` → `rw`.

**`fluent-bit`, watched hardest.** It is the one container getting a read-only root with no writable path, and the claim that it needs none was previously read off the generated ConfigMap rather than observed. It came up Running with **0 restarts**, shipped agent and MCP log lines continuously for the whole window, and its complete log contains zero matches for `read-only`, `EROFS`, `permission denied`, `cannot open` or `failed`. The claim holds.

**What this run did not cover.** The pod did not reach fully Ready: `envoy-credential-proxy` crashlooped with `exec: "/usr/local/bin/start-services": no such file or directory`, and the CR went `Degraded` for that reason. That is version skew, not this change — `start-services` was introduced on `main` at `fe08297f:2200` when the event watcher was folded into the credential sidecar, the string appears nowhere in this diff, and the newest `credential-proxy` image in the sandbox registry (`latest`, same digest as the pinned tag) was built 2026-07-30 and predates it. With no container runtime available there was no way to build an image that has it. So: the three containers this PR targets all ran green under a read-only root; the one that crashed is the one that already had a read-only root before this PR, for an unrelated missing binary.

Also uncovered: whether a *skill* or plugin executed at runtime writes outside `/tmp` and the PVC over a long session, and whether `2Gi` is a sane ceiling for real workloads. The failure mode if either is wrong is `EROFS` inside the agent container, or eviction for exceeding the volume's `sizeLimit`.

**Cleanup.** The in-cluster operator is back at 1 replica, the pod is 5/5 Running, the CR is `Ready=True`, and both the CR spec and the Deployment pod spec are byte-identical to the baselines captured before the run. The lease was released. The cluster's DNS control-plane endpoint, enabled to reach the private control plane, was disabled again (`allowExternalTraffic: false`). No test artifacts remain in the cluster.

## Self-Review

Two pre-PR passes, both against the rebased diff `fe08297f...HEAD`.

**`review-adversarial` — run in a subagent that did not write the change.** Handed the diff range and the skill only: not the plan, not the commit bodies, not this PR body, and told to derive the intent from the diff itself. It ran all ten angles and reported what it looked for under each: line-by-line with slice-aliasing and `path.Clean` edge cases; removed behaviour, including an audit for silenced tests and narrowed CI triggers; a cross-file trace of every symbol the diff newly references and every renderer of this Pod; operations and security, weighing the shared `emptyDir` as a channel and the lost `1777` sticky bit; reuse; simplification; altitude; conventions and doc-claim verification against source; scope and test coverage, checked red-then-green rather than by reading; and the 100 open sibling PRs. Four findings:

- **No CI or test evidence that the three newly-hardened containers survive a read-only root at _runtime_** (MEDIUM). The new `docker run --read-only` step execs `entrypoint-gate-check`, whose cases must all exec `/bin/echo`, so it covers the entrypoint's setup phase and never starts `hermes gateway run`, `hermes dashboard`, or `fluent-bit`. **Not fixed in code; answered by evidence.** This is exactly the gap Live validation exists to close, and both runs below name all three containers — @bradhoekstra's under a live LLM with real skills, mine watching `fluent-bit` under sustained load. Closing it in CI would need a PR-time cluster e2e, which this repository does not have and this PR is not the place to add.
- **The diff makes `scripts/check_image_layers.py`'s stated rationale false** (LOW). Its docstring says `entrypoint-gate-test` is out of scope because "no daemon ever mounts it"; the diff adds `load: true` and then `docker run`s it. `deploy/docker/Dockerfile` carried the same now-incomplete claim. Neither file is in the diff, so the contradiction would have merged silently. **Fixed** in `a5c23db`: the stage is still out of scope, but because it adds two layers on top of `platform` and the 8-layer gap between the 120 budget and overlay2's 128 covers them — narrow that gap and the gate stage runs out of room first.
- **The `--tmpfs … size=2g` justification conflates two volumes with separate budgets** (LOW). The comment had one tmpfs stand in for both the data PVC and the `tmp-scratch` `emptyDir`, then read `size=2g` as matching "that volume's sizeLimit". In the pod those are two volumes with independent budgets. **Fixed** in `a5c23db`: the comment now says 2g bounds their sum, predicts neither, and that a failure there means "this case writes more than we expect", not "`tmp-scratch` is too small".
- **`dropTmpScratchIfClaimed` special-cases `/tmp` where a general dedup is the same amount of code** (LOW, design judgement). **Deliberately not generalised.** A general "drop every default whose cleaned path a user mount claims" would silently drop the data PVC mount if a CR set `homeDir` to a default path — turning a Deployment the API server rejects outright into an agent that comes up with no persistent home, which is the worse of the two failures. The pass's own counter-argument was that this trade-off is stated nowhere; `a5c23db` states it in the helper's doc comment. The sibling collision it demonstrated (`agentHome: /tmp`) is **pre-existing**, not a regression: the pass rendered the same CR at `fe08297f` and it was already unappliable via the credential proxy's `credential-proxy-tmp`.

The pass also **refuted** several things worth recording as checked: no slice-aliasing bug in the two rebuilt `append` chains; no test removed, skipped, or silenced and no CI trigger narrowed; the shared `/tmp` crosses no trust boundary the data PVC does not already cross; the `emptyDir` is disk-backed so it does not consume the memory limit.

**One thing the pass got wrong, corrected here rather than left standing.** It reported that it could construct no stale-state failure from the new pod-lifetime `/tmp`, on the grounds that `HERMES_HOME` is set explicitly on all four `HOME=/tmp` invocations. That reasoning does not hold. `deploy/shared/docker-entrypoint.sh:1265-1267` gives as the reason its step-4a compat symlink exists that "hermes-otel resolves config below `~/.hermes` even when `HERMES_HOME` points elsewhere", and `agents/platform/scripts/profile_scaffold.py:79-85` has the scaffold force `HOME=/tmp` onto every `hermes` subprocess it spawns. Those subprocesses do consult `/tmp/.hermes`, and that directory is now restart-durable and visible to the dashboard container. So @mplakhtiy's thread-1 mechanism stands; what remains unproven is whether attacker-controlled content there changes what the scaffolding pass does, which is what they labelled plausible rather than certain. Disposition is in that thread: the fix belongs in the entrypoint behind its own gate-check case, not here.

**What it could not cover, in its own words:** it had no Docker daemon, so it could not build or run the new CI step — the step answers itself on this PR's run; `stage2-hook.sh` lives in the upstream Hermes image and is unreadable from here, which is the whole reason that step tests rather than reads; and anything needing a live cluster, which is what Live validation below is for. It also noted, without raising it as a finding, that a read-only root blocks an LLM-driven agent from `pip install`ing into `/opt/hermes/.venv` — it grepped `agents/` and `.agents/` and found no skill that does so, and neither live run hit it.

**`review-docs-drift` — run in this session, the one that wrote the change.** Weaker placement than the pass above, and stated as such. No Blocking findings. Every claim in the two edited docs was verified against source rather than against other docs: the five-container enumeration against `platformagent_manifests.go`, the four `HOME=/tmp` invocations against `docker-entrypoint.sh`, the shared data PVC against the mount list, and the "the webhook does not require a read-only root" sentence against `validateContainerSecurity` in `platformagent_webhook.go`. No new files, so the documentation map needs no edit. `docs/architecture/08` mentions a read-only rootfs for the future gVisor execution sandbox — a different thing, not contradicted.

**One reviewer finding, from @bradhoekstra's live run, disposed here rather than deferred.** When a CR claims `/tmp`, the `tmp-scratch` volume stays declared in the Pod spec with no container mounting it, because the guard removes the mount and nothing removes the volume. Verified: the declaration is unconditional. **Deliberately not fixed** — it is legal and inert, the kubelet creates an empty directory and no container sees it, and making the declaration conditional means deciding it from state that lives in the mount builders. Two reviewers have now had to derive that from the code, so `a5c23db` writes it down at the declaration.

## Risk & Rollout

Blast radius is every agent Pod the operator builds, on the next reconcile — this changes the Pod template, so merging it rolls the gateway Deployment wherever it lands. Risk is concentrated in one place: a runtime write path outside `/tmp` and the PVC that neither static reading nor either live run found. The failure mode is visible and local — a write failing with `EROFS` inside the agent container, or a pod eviction for exceeding the volume's `sizeLimit` — not silent corruption.

The one upgrade hazard is the `/tmp` collision, which is why the guard exists: an install whose CR already mounts something at `/tmp` would otherwise produce a Deployment the API server rejects outright. That path is unit-tested for both `extraVolumeMounts` and `storages`, and @bradhoekstra exercised the `extraVolumeMounts` variant live.

Rollback is reverting the commit; there is no data migration, no CR schema change, and nothing to undo outside the Pod template. Nothing has to happen at merge time. Functional impact for standard deployments should be nil — `/tmp` remains writable, just bounded and no longer part of the image layer.

_This PR was generated with the help of Claude._

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -47,10 +47,10 @@ jobs:
       # deploy to amd64 GKE nodes, and the multi-arch bases otherwise resolve to
       # whatever the runner is (#560). Redundant on today's amd64 runners; load-bearing
       # the day the runner label changes.
-      # load: true, unlike the other builds here, so the layer-budget step below
-      # has an image to inspect. It costs a transfer from the buildx cache into
-      # the daemon; nothing else needs the image, which is why the others do
-      # without.
+      # load: true, so the layer-budget step below has an image to inspect. It costs
+      # a transfer from the buildx cache into the daemon, which is why only this
+      # build and the gate-test build below -- the two whose output a later step
+      # runs or inspects -- pay for it.
       - name: Build Platform Agent
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
@@ -89,8 +89,33 @@ jobs:
           target: entrypoint-gate-test
           platforms: linux/amd64
           push: false
+          # load, unlike before: the step below re-runs this stage's check as a
+          # container rather than as a build RUN, and needs the image in the daemon.
+          load: true
+          tags: entrypoint-gate-test:latest
           build-args: |
             HERMES_AGENT_TAG=${{ env.HERMES_AGENT_TAG }}
+
+      # The same check again, against the rootfs the operator actually gives the pod.
+      # The RUN above cannot cover this: a build layer is writable by definition, so it
+      # proves the entrypoint works but says nothing about whether it needs the write.
+      # readOnlyRootFilesystem turns any such write into EROFS at container start, and
+      # the entrypoint runs a script this repo does not own (upstream's stage2-hook.sh),
+      # so "it only writes under $HERMES_HOME and /tmp" has to be tested rather than read
+      # off the source. Cheap, because the image is already in the daemon from the load.
+      #
+      # --tmpfs stands in for the two writable mounts the Deployment provides: the data
+      # PVC at $PLATFORM_AGENT_HOME, which the check already relocates under /tmp, and the
+      # tmp-scratch emptyDir. size=2g matches that volume's sizeLimit, so a case that
+      # outgrows the real volume fails here too; exec because the check's HA-handoff case
+      # runs a wrapper script from the scratch tree, and Docker's default tmpfs is noexec.
+      - name: Verify entrypoint under a read-only root filesystem
+        run: |
+          docker run --rm \
+            --read-only \
+            --tmpfs /tmp:rw,exec,mode=1777,size=2g \
+            --entrypoint /usr/local/bin/entrypoint-gate-check \
+            entrypoint-gate-test:latest
 
       # No load: true. This shares only agent-base with the build above, so it is
       # the shorter chain of the two and not the one the budget is measured
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -104,11 +104,15 @@ jobs:
       # so "it only writes under $HERMES_HOME and /tmp" has to be tested rather than read
       # off the source. Cheap, because the image is already in the daemon from the load.
       #
-      # --tmpfs stands in for the two writable mounts the Deployment provides: the data
-      # PVC at $PLATFORM_AGENT_HOME, which the check already relocates under /tmp, and the
-      # tmp-scratch emptyDir. size=2g matches that volume's sizeLimit, so a case that
-      # outgrows the real volume fails here too; exec because the check's HA-handoff case
-      # runs a wrapper script from the scratch tree, and Docker's default tmpfs is noexec.
+      # One --tmpfs stands in for two writable mounts the Deployment provides separately:
+      # the data PVC at $PLATFORM_AGENT_HOME, which the check already relocates under /tmp,
+      # and the tmp-scratch emptyDir. Those have independent budgets in the pod -- a
+      # multi-Gi PVC and a 2Gi emptyDir -- so size=2g is a ceiling on the two combined and
+      # not a model of either. It is there to keep a runaway case from filling the runner,
+      # not to predict which volume a real pod would outgrow; read a failure here as "this
+      # case writes more than we expect", not as "tmp-scratch is too small". exec, because
+      # the check's HA-handoff case runs a wrapper script from the scratch tree and
+      # Docker's default tmpfs is noexec.
       - name: Verify entrypoint under a read-only root filesystem
         run: |
           docker run --rm \
--- a/k8s-operator/internal/controller/platformagent_manifests.go
+++ b/k8s-operator/internal/controller/platformagent_manifests.go
@@ -2114,6 +2114,14 @@ func buildDefaultVolumeMounts(homeDir string) []corev1.VolumeMount {
 // would be a silent behaviour change on upgrade. If they mounted something read-only there
 // the agent will fail to start — but it would have failed the same way before this change,
 // since the entrypoint has always needed a writable /tmp.
+//
+// Narrow on purpose: it detects the collision by cleaned mountPath but removes the default
+// by volume name, and only ever for /tmp. The general form — drop every default whose
+// cleaned path a user mount claims — is the same amount of code and would cover the next
+// default the operator adds. It would also silently drop the data PVC mount if a CR set
+// homeDir to a path the defaults use, turning a Deployment the API server rejects outright
+// into an agent that comes up with no persistent home. A rejected Deployment is the better
+// failure of the two, so the generalisation waits for a second path that actually needs it.
 func dropTmpScratchIfClaimed(defaults, userMounts []corev1.VolumeMount) []corev1.VolumeMount {
 	claimed := false
 	for _, m := range userMounts {
@@ -3114,6 +3122,14 @@ func buildDefaultVolumes(agent *agentv1alpha1.PlatformAgent) []corev1.Volume {
 			// alpha LocalStorageCapacityIsolationFSQuotaMonitoring gate, which GKE
 			// does not enable. 2Gi matches credential-proxy-tmp, the closest
 			// analogue.
+			//
+			// Declared unconditionally, while dropTmpScratchIfClaimed can remove the
+			// mount from both containers -- so a CR that claims /tmp itself leaves
+			// this volume in the pod with nothing mounting it. That is legal and
+			// inert: the kubelet creates an empty directory and no container sees it.
+			// Making the declaration conditional would mean deciding it here from
+			// state that lives in the mount builders, which buys a tidier pod spec
+			// for a coupling that is easier to get wrong than this is to explain.
 			Name: tmpScratchVolumeName,
 			VolumeSource: corev1.VolumeSource{
 				EmptyDir: &corev1.EmptyDirVolumeSource{
--- a/k8s-operator/internal/controller/platformagent_manifests.go
+++ b/k8s-operator/internal/controller/platformagent_manifests.go
@@ -2134,6 +2134,28 @@ func dropTmpScratchIfClaimed(defaults, userMounts []corev1.VolumeMount) []corev1
 	return kept
 }
 
+// hardenedSecurityContext is the container hardening every container the operator builds
+// carries: no privilege escalation, no capabilities, and a root filesystem the process
+// cannot write to.
+//
+// One helper rather than a literal per container, because the alternative has already
+// failed. The same three fields were written out five times, and three of the five had
+// silently drifted to two of them — the read-only root was on the credential sidecar and
+// the cleanup init container and on neither agent container nor the log shipper, which is
+// the gap this function was introduced to close (RUNTIME-007). Nothing failed while that
+// was true. A container added from here on gets the block by construction, and
+// TestEveryContainerHasAHardenedSecurityContext fails if one is built without it.
+//
+// Anything a container needs on top — a different user, a writable path — belongs on that
+// container, not here. This is the floor, not the whole context.
+func hardenedSecurityContext() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		AllowPrivilegeEscalation: ptr.To(false),
+		ReadOnlyRootFilesystem:   ptr.To(true),
+		Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+	}
+}
+
 func buildSandboxCredentialCleanup(image string, pullPolicy corev1.PullPolicy) corev1.Container {
 	return corev1.Container{
 		Name:            "sandbox-credential-cleanup",
@@ -2154,12 +2176,8 @@ func buildSandboxCredentialCleanup(image string, pullPolicy corev1.PullPolicy) c
   /workspace/home/.netrc \
   /workspace/home/.npmrc \
   /workspace/home/.pypirc`},
-		VolumeMounts: []corev1.VolumeMount{{Name: "platform-agent-data-vol", MountPath: "/workspace"}},
-		SecurityContext: &corev1.SecurityContext{
-			AllowPrivilegeEscalation: ptr.To(false),
-			ReadOnlyRootFilesystem:   ptr.To(true),
-			Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
-		},
+		VolumeMounts:    []corev1.VolumeMount{{Name: "platform-agent-data-vol", MountPath: "/workspace"}},
+		SecurityContext: hardenedSecurityContext(),
 		Resources: corev1.ResourceRequirements{
 			Limits: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("200m"),
@@ -2276,9 +2294,7 @@ func buildCredentialProxySidecar(agent *agentv1alpha1.PlatformAgent, homeDir str
 			{Name: "event-watcher-ksa-token", MountPath: "/var/run/secrets/kubernetes.io/serviceaccount", ReadOnly: true},
 			{Name: "platform-agent-data-vol", MountPath: homeDir},
 		},
-		SecurityContext: &corev1.SecurityContext{
-			AllowPrivilegeEscalation: ptr.To(false), ReadOnlyRootFilesystem: ptr.To(true), Capabilities: &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
-		},
+		SecurityContext: hardenedSecurityContext(),
 	}
 }
 
@@ -2765,15 +2781,9 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 			// The bearer key is the non-secret loopback sentinel already in this
 			// container's env, and API_SERVER_ENABLED is unconditionally true above,
 			// so the probe is valid in every configuration.
-			StartupProbe:   agentAPIProbe(10, 60),
-			ReadinessProbe: agentAPIProbe(15, 3),
-			SecurityContext: &corev1.SecurityContext{
-				AllowPrivilegeEscalation: ptr.To(false),
-				ReadOnlyRootFilesystem:   ptr.To(true),
-				Capabilities: &corev1.Capabilities{
-					Drop: []corev1.Capability{"ALL"},
-				},
-			},
+			StartupProbe:    agentAPIProbe(10, 60),
+			ReadinessProbe:  agentAPIProbe(15, 3),
+			SecurityContext: hardenedSecurityContext(),
 		},
 	}
 
@@ -2950,13 +2960,7 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 				TimeoutSeconds:      5,
 				FailureThreshold:    3,
 			},
-			SecurityContext: &corev1.SecurityContext{
-				AllowPrivilegeEscalation: ptr.To(false),
-				ReadOnlyRootFilesystem:   ptr.To(true),
-				Capabilities: &corev1.Capabilities{
-					Drop: []corev1.Capability{"ALL"},
-				},
-			},
+			SecurityContext: hardenedSecurityContext(),
 		})
 	}
 
@@ -3002,18 +3006,12 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 				MountPath: "/fluent-bit/state",
 			},
 		},
-		SecurityContext: &corev1.SecurityContext{
-			AllowPrivilegeEscalation: ptr.To(false),
-			// No /tmp for this one, deliberately. The config above buffers in memory
-			// (Mem_Buf_Limit, no storage.path), keeps its tail DB on the
-			// fluent-bit-state volume and outputs to stdout, so it writes nothing to
-			// the root filesystem. Handing it the agent's tmp-scratch would only give
-			// an LLM-driven container a path into the log shipper.
-			ReadOnlyRootFilesystem: ptr.To(true),
-			Capabilities: &corev1.Capabilities{
-				Drop: []corev1.Capability{"ALL"},
-			},
-		},
+		// Read-only root and no /tmp, the only container here with neither. The config
+		// above buffers in memory (Mem_Buf_Limit, no storage.path), keeps its tail DB on
+		// the fluent-bit-state volume and outputs to stdout, so it writes nothing to the
+		// root filesystem. Handing it the agent's tmp-scratch would only give an
+		// LLM-driven container a path into the log shipper.
+		SecurityContext: hardenedSecurityContext(),
 	})
 
 	// The k8s-event-watcher is not a container of its own. It runs inside
--- a/k8s-operator/internal/controller/platformagent_manifests.go
+++ b/k8s-operator/internal/controller/platformagent_manifests.go
@@ -60,7 +60,8 @@ const (
 	// (see the readiness probe in buildBaseContainers), so the container port, the
 	// Service port, and the NetworkPolicy rule below all describe a listener that
 	// only kubelet's port-forward can reach.
-	dashboardPort = 9119
+	dashboardPort        = 9119
+	tmpScratchVolumeName = "tmp-scratch"
 )
 
 // Shared-state ownership. Step 1.5 of deploy/shared/docker-entrypoint.sh reads this
@@ -2086,7 +2087,51 @@ func buildDefaultVolumeMounts(homeDir string) []corev1.VolumeMount {
 			MountPath: path.Dir(sessionKVDBPath),
 			SubPath:   "session",
 		},
+		{
+			// The one writable path outside the PVC, and the reason
+			// readOnlyRootFilesystem is survivable here: docker-entrypoint.sh runs
+			// four hermes invocations with HOME=/tmp before the agent starts, and
+			// the image is otherwise root-owned against a runtime UID of 10000.
+			Name:      tmpScratchVolumeName,
+			MountPath: "/tmp",
+		},
 	}
+}
+
+// dropTmpScratchIfClaimed removes the operator's /tmp mount when the CR already mounts
+// something there via storages or extraVolumeMounts.
+//
+// Those are appended to the defaults without deduplication, and the API server rejects a
+// Pod spec with two mounts on one mountPath. So without this the /tmp mount added above
+// would not merely be redundant on such a CR, it would make the Deployment unappliable and
+// stop reconciliation dead — on an upgrade, for a field the user set before /tmp was a
+// path the operator had any opinion about. Mounting scratch space at /tmp is exactly what
+// someone would have done while the root filesystem was still writable and there was no
+// other writable path outside the PVC, which is what makes this worth handling rather than
+// rejecting.
+//
+// The user's mount wins, deliberately: it is the one that predates this, and overriding it
+// would be a silent behaviour change on upgrade. If they mounted something read-only there
+// the agent will fail to start — but it would have failed the same way before this change,
+// since the entrypoint has always needed a writable /tmp.
+func dropTmpScratchIfClaimed(defaults, userMounts []corev1.VolumeMount) []corev1.VolumeMount {
+	claimed := false
+	for _, m := range userMounts {
+		if path.Clean(m.MountPath) == "/tmp" {
+			claimed = true
+			break
+		}
+	}
+	if !claimed {
+		return defaults
+	}
+	kept := make([]corev1.VolumeMount, 0, len(defaults))
+	for _, m := range defaults {
+		if m.Name != tmpScratchVolumeName {
+			kept = append(kept, m)
+		}
+	}
+	return kept
 }
 
 func buildSandboxCredentialCleanup(image string, pullPolicy corev1.PullPolicy) corev1.Container {
@@ -2599,13 +2644,14 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 
 	resources := resolveResources(agent.Spec.Deployment)
 
-	volumeMounts := buildDefaultVolumeMounts(homeDir)
+	var userVolumeMounts []corev1.VolumeMount
 	if len(storages) > 0 {
-		volumeMounts = append(volumeMounts, buildCustomStorageVolumeMounts(storages)...)
+		userVolumeMounts = append(userVolumeMounts, buildCustomStorageVolumeMounts(storages)...)
 	}
 	if len(extraVolumeMounts) > 0 {
-		volumeMounts = append(volumeMounts, extraVolumeMounts...)
+		userVolumeMounts = append(userVolumeMounts, extraVolumeMounts...)
 	}
+	volumeMounts := append(dropTmpScratchIfClaimed(buildDefaultVolumeMounts(homeDir), userVolumeMounts), userVolumeMounts...)
 
 	// Args, never Command. Command replaces the image ENTRYPOINT
 	// (/usr/local/bin/agent-entrypoint), and that script is what makes $HERMES_HOME
@@ -2723,6 +2769,7 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 			ReadinessProbe: agentAPIProbe(15, 3),
 			SecurityContext: &corev1.SecurityContext{
 				AllowPrivilegeEscalation: ptr.To(false),
+				ReadOnlyRootFilesystem:   ptr.To(true),
 				Capabilities: &corev1.Capabilities{
 					Drop: []corev1.Capability{"ALL"},
 				},
@@ -2813,6 +2860,13 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 				MountPath: path.Dir(sessionKVDBPath),
 				SubPath:   "session",
 			},
+			{
+				// Same emptyDir the gateway gets. Sharing it is not a new channel:
+				// these two containers already run the same image against the same
+				// data PVC, so they are one trust domain either way.
+				Name:      tmpScratchVolumeName,
+				MountPath: "/tmp",
+			},
 		}
 
 		// What keeps this container out of the shared tree is AGENT_SHARED_STATE_SETUP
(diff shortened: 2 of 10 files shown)


bradhoekstra:
## Live validation — PASS

The PR body flags this as not live-tested and names two things for a reviewer with a
cluster to check. I ran both against a real install. Both hold up.

**Environment.** `kube-agents-cluster` (regional `us-central1`, 3 nodes, project
`bhoekstra-gkedemos`), namespace `kubeagents-system`, permission set `read-only`,
gVisor off, agent backed by a live LLM. Baseline operator
`k8s-operator:dns-endpoint-7dac349`; PR build from `4a3b628`.

**Method.** This PR touches no file under `deploy/docker/`, so the agent image is
byte-identical between baseline and PR build. I rebuilt and swapped **only** the
operator image, which isolates the change to manifest generation.

Before: `/ rw,relatime`, no `/tmp` mount. After: `/ ro,relatime`, `/tmp rw,relatime`.
Read from `/proc/self/mountinfo` field 6 — worth noting for anyone repeating this, as
it is the only check that discriminates: `touch /probe` returns `EACCES` for UID 10000
either way, because `/` is root-owned.

### Structure

All five containers report `readOnlyRootFilesystem=true`, including the
`sandbox-credential-cleanup` init container. `tmp-scratch` is present at `2Gi`, mounted
at `/tmp` on the two agent containers and **not** on `fluent-bit`, as intended.
Confirmed `fluent-bit` keeps its tail DB at `/fluent-bit/state/fluent-bit.db` with
`Mem_Buf_Limit 20MB` and stdout output — no rootfs writes.

Health throughout: all containers ready, **0 restarts**, init container `Completed`,
CR `Ready=True`, and **0** occurrences of `EROFS` / "read-only file system" in any
container log.

### Ask 1 — does anything at runtime write outside `/tmp` and the PVC?

Nothing I could make it do.

- **67 skills** load, plugins load, cron scheduler runs.
- Tool sweep under read-only root: `kubectl`, `gcloud`, `git`, `python3` (`tempfile`),
  `node`, `npm`, `uv`, `yq` all fine. `helm repo add/update/search` works — it writes
  to `$HOME=/opt/data/home`, on the PVC.
- Two full LLM-driven agent tasks, end to end through kanban delegation:
  - a read-only `kubectl` task — correct output, matched the cluster;
  - a deliberately write-heavy one: the `gke-manifest-generation` skill, a generated
    manifest written to disk, a scratch `git` repo (`init`/`add`/`commit`/`log`), and a
    python `tempfile` script. All three completed. `tempfile` landed in `/tmp`, i.e.
    the new `emptyDir`, as designed.

One thing that looks like a failure in that second task but is not: `git` was blocked
twice with `working directory is outside the shared workspace` / `policy rule:
git.workspace.lease`. That is the credential-proxy's pre-existing GitOps lease guard,
not the filesystem — the agent stamped a lease and committed successfully.

Caveat on scope: this is a broad sample, not a proof over every skill. It did not
exercise all 67.

### Ask 2 — is `2Gi` a sane ceiling?

Comfortably. Peak `/tmp` usage after both tasks, including manifest generation and a
git repo, was **28K** — five files.

Also worth confirming for the record: `df /tmp` inside the container reports the node's
disk, not `2Gi`, and the kubelet summary API reports the node capacity for the volume
too. So there is no in-container quota and an overflow is a pod eviction, not `ENOSPC`
— exactly as the "Why This Matters" section states.

### The `/tmp` collision guard

Verified as a real upgrade hazard, not a theoretical one. A control confirmed the API
server rejects the un-guarded shape outright (`mountPath must be unique`). With the
guard, a CR claiming `/tmp` via `extraVolumeMounts` produced the CR's own volume
winning on **both** the gateway and the dashboard, `tmp-scratch` correctly dropped, no
duplicate paths, and the pod ran.

I also enabled the dashboard specially for this (it is off by default on this install):
it came up clean, `HERMES_DASHBOARD_READY`, 0 restarts, and I confirmed the scratch
volume is genuinely shared across the two containers.

**One small observation.** When a CR claims `/tmp` *and* the dashboard is disabled, the
`tmp-scratch` volume stays declared in the Pod spec with no container mounting it.
Harmless — the kubelet does not fault on it — but dangling. Not worth blocking on.

### Not covered

- A deliberate `2Gi` overflow / eviction (destructive).
- The `storages` variant of the collision guard — unit-tested only; I exercised the
  `extraVolumeMounts` path live.
- gVisor interaction — off on this install.

### Cleanup

Cluster fully reverted to the baseline operator and re-verified: no `tmp-scratch`,
`readOnlyRootFilesystem` unset again, all containers ready, 0 restarts, CR `Ready`. The
PR build image was deleted from the registry. No test artifacts left behind.

_This validation was performed with the help of Claude._

k8s-operator/internal/controller/platformagent_manifests.go
```diff
@@ -2500,6 +2563,23 @@ func buildDefaultVolumes(agent *agentv1alpha1.PlatformAgent) []corev1.Volume {
...
+			// alpha LocalStorageCapacityIsolationFSQuotaMonitoring gate, which GKE
+			// does not enable. 2Gi matches credential-proxy-tmp, the closest
+			// analogue.
+			Name: tmpScratchVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{
+					SizeLimit: ptr.To(resource.MustParse("2Gi")),
+				},
+			},
+		},
```
mplakhtiy, line 3139: **MEDIUM — `/tmp` goes from container-lifetime and private to pod-lifetime and shared.**

Before this change `/tmp` in each agent container was that container's own writable layer: private to the container and destroyed on every restart, because a Kubernetes container restart creates a new container from the image. After it, `/tmp` is one `emptyDir` mounted into both `platform-agent` (`:1818-1825`) and `platform-agent-dashboard` (`:2397-2403`), and an `emptyDir` lives for the lifetime of the **Pod** — Kubernetes documents explicitly that "a container crashing does not remove a Pod, so the data in an `emptyDir` volume is safe across container crashes."

So the one writable path this change leaves the agent is also the one path whose contents now survive the agent restarting, and it is shared with a second container. The PR answers the sharing question on trust-domain grounds ("these two containers already run the same image against the same data PVC"), which is fair — but it does not address the lifetime change at all, and neither does the bullet added to `docs/credential-isolation-design.md`.

**Failure scenario.** The four scaffolding invocations at `deploy/shared/docker-entrypoint.sh:433,483,506,611` run with `HOME=/tmp` on *every* container start, and Hermes resolves plugin configuration below `~/.hermes` regardless of `HERMES_HOME` — the entrypoint's own line 937 relies on that (`OTEL_COMPAT_CONFIG="$HOME/.hermes/plugins/…"`). A prompt-injected agent writes `/tmp/.hermes/…`, the gateway container restarts, and the scaffolding pass reads back attacker-controlled state that the previous arrangement would have wiped. The same content is visible to the dashboard container's own boot.

The lifetime and sharing changes are certain from the code and the golden manifests; the read-back chain is the plausible consequence, and it is what makes the lifetime change worth closing rather than just noting.

**Suggested fix — in the entrypoint, not the operator.** Point those four invocations at a per-boot `mktemp -d` under `/tmp` and remove it afterwards. A second `emptyDir` would restore per-container privacy but not the per-restart wipe, since `emptyDir` is pod-scoped either way.

If you would rather defer it, the lifetime change should at least be stated in the `docs/credential-isolation-design.md` bullet rather than left implied by "shared scratch volume".

k8s-operator/internal/controller/platformagent_manifests.go
```diff
@@ -2260,6 +2306,7 @@ func buildBaseContainers(agent *agentv1alpha1.PlatformAgent, image string, envVa
 			VolumeMounts: volumeMounts,
 			SecurityContext: &corev1.SecurityContext{
 				AllowPrivilegeEscalation: ptr.To(false),
+				ReadOnlyRootFilesystem:   ptr.To(true),
 				Capabilities: &corev1.Capabilities{
 					Drop: []corev1.Capability{"ALL"},
 				},
```
mplakhtiy: **MEDIUM — nothing asserts the invariant this PR exists to establish.**

Five containers now carry a byte-identical hardening block — `AllowPrivilegeEscalation: false`, `ReadOnlyRootFilesystem: true`, `Capabilities.Drop: ["ALL"]` — written out five separate times (`:1886-1890`, `:1985`, `:2307-2312`, `:2435-2441`, `:2487-2497`), two of them in a collapsed one-line style and three in the expanded one. There is no helper.

The new tests assert the field on three containers **by name** — `platform-agent`, `platform-agent-dashboard`, `fluent-bit` (`platformagent_manifests_test.go:535`, `:868`, `:880`). Neither `sandbox-credential-cleanup` nor `envoy-credential-proxy` has any unit assertion for the field at all.

**Failure scenario.** The next container the operator gains ships without a read-only root and every test stays green. That is not hypothetical — it is precisely how three of the five ended up missing the field, and it is the whole reason this PR exists. The golden manifests do capture the current state, but a golden is regenerated mechanically when a container is added, so it records the omission rather than rejecting it.

**The shared form to extract:** a `hardenedSecurityContext()` returning the common block, called by all five sites, plus one test that walks `pod.Spec.InitContainers` and `pod.Spec.Containers` and requires `ReadOnlyRootFilesystem == true` on every entry the operator builds — the same walk-every-container shape `TestUserSuppliedTmpMountReplacesTmpScratch` already uses for mount uniqueness, and for the same reason.

Worth taking in this PR rather than deferring: it is the cheapest possible guard against the exact regression this PR is fixing, and the test file already has the walk.

mplakhtiy (commented):
Two Medium findings from a review of `4a3b628`, both inline below.

Scope note: the branch conflicts with `main` (`platformagent_manifests_test.go`, two end-of-file append collisions), so this was reviewed as authored rather than as merged — line references are to the branch head.

k8s-operator/internal/controller/platformagent_manifests.go
AntonTyb: Adopted, both halves.

`hardenedSecurityContext()` now returns the block and all five sites call it — the two collapsed literals and the three expanded ones are gone, so the drift that produced this finding can't recur by copy-paste. The doc comment records *why* it is a helper rather than five literals, including that three of the five had already silently lost the field.

`TestEveryContainerHasAHardenedSecurityContext` is the walk you described. It runs over both Pod shapes (dashboard enabled and disabled), iterates `append(pod.Spec.InitContainers, pod.Spec.Containers...)`, and checks all three fields rather than just `ReadOnlyRootFilesystem` — the same drift argument applies to the other two. It closes in both directions: a container present but unhardened fails, and a container in the expected list that never appears also fails. A container the operator grows that isn't in `operatorBuiltContainers` fails with a message that says what to do:

```
container "brand-new-thing" is not in operatorBuiltContainers: if the
operator grew a container, give it hardenedSecurityContext() and add it
to that list
```

I checked it actually fails rather than assuming it: unhardening `fluent-bit` produces `ReadOnlyRootFilesystem is not true` and `capabilities do not drop ALL, got nil`; adding an unrecognised container produces the message above.

`sandbox-credential-cleanup` and `envoy-credential-proxy` are covered now — that gap you noted was real, and the walk is what closes it rather than two more by-name assertions.

One thing the walk deliberately does not claim, so I added `TestCRSuppliedSidecarsAreNotHardenedByTheOperator` to state it: containers from `spec.deployment.sidecars`/`initContainers` pass through as written. `validateContainerSecurity` in the webhook checks `privileged`, `allowPrivilegeEscalation`, `runAsUser=0` and `capabilities.add`, but not `readOnlyRootFilesystem` — so a CR can still add a writable container to this Pod. That's the honest boundary of "every container has a hardened context", and the `docs/credential-isolation-design.md` bullet now says so too rather than overclaiming.

The branch is rebased onto current `main` (`fe08297f`) and force-pushed; line numbers in your comment have moved.

k8s-operator/internal/controller/platformagent_manifests.go
AntonTyb: The lifetime change is real and I've taken your fallback: the `docs/credential-isolation-design.md` bullet now states it outright rather than leaving it implied by "shared scratch volume".

> Note the lifetime this changes: `/tmp` used to be each container's own writable layer, discarded whenever that container restarted. An `emptyDir` is scoped to the Pod, so its contents now survive a container crash or restart and are visible to both containers' next boot.

Deferring the fix itself, and here is the reasoning rather than just the decision.

Your suggested fix is in `deploy/shared/docker-entrypoint.sh` — point the four scaffolding invocations at a per-boot `mktemp -d` and remove it afterwards. That is the right shape and it is the right file, but it is a change to the agent image's entrypoint, not to the operator, and it wants its own blast radius: the entrypoint is shared by the gateway and the dashboard, it is covered by a separate in-image gate check, and `HOME=/tmp` interacts with the `~/.hermes` config resolution you cite in a way that deserves its own test rather than riding along in a manifests PR. Landing it here would also mean this PR could no longer be reverted by flipping one field on three containers, which is currently its main safety property.

Worth separating the two claims in your comment, because they have different standing:

- **Lifetime and sharing** are certain from the code, and now documented.
- **The read-back chain** — a prompt-injected agent writes `/tmp/.hermes/…`, the container restarts, the scaffolding pass reads attacker-controlled state — is the plausible consequence, and you say so. I agree it is plausible enough to fix. I do not think this PR is where it gets fixed, and I would rather it get a change that can be tested against the entrypoint gate than a hurried one here.

One thing worth stating plainly, since it bears on urgency: this is a *pre-existing* exposure that the PR narrows in one dimension and widens in another. Before this change the agent could write anywhere in its image layer, including paths that persist for the container's life and that no restart-wipe argument protects — the writable-layer wipe only helps because a Kubernetes restart replaces the container. After it, the writable surface is one bounded, non-executable-image path, but that path outlives a restart. Net, the tamper-persistence surface shrinks a lot and gains one narrow durable corner. That is an improvement to keep and a corner to close, not a reason to hold the improvement.

Where it should land: a follow-up against `deploy/shared/docker-entrypoint.sh`, which I have not filed yet — say the word and I'll open it, or open it yourself if you'd rather word the repro. If you'd rather it block this PR, say so and I'll move it in — but I'd want the entrypoint gate check extended to cover the restart-read-back case first, and that is a bigger change than the one-clause doc note.

Branch is rebased onto `fe08297f`; your line references have moved.

.github/workflows/docker-build.yml
```diff
@@ -47,10 +47,10 @@ jobs:
       # deploy to amd64 GKE nodes, and the multi-arch bases otherwise resolve to
```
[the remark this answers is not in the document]
AntonTyb: Fixed — all three sections are now in the body.

**Summary** and **Risk & Rollout** are new. **Self-Review** records two passes with what each looked for and a disposition per finding, and names the kind of context each ran in: `review-adversarial` in a subagent handed only the diff range `fe08297f...HEAD` and the skill, and `review-docs-drift` in the session that wrote the change, which is the weaker placement and is stated as such.

You were also right that the docs-drift result was in the wrong place. AGENTS.md says its dispositions go in **Self-Review** with the adversarial pass's, not in a section of their own; it was under **Testing**. Moved.

The adversarial pass returned four findings. Two were comments this diff made false — `scripts/check_image_layers.py` and `deploy/docker/Dockerfile` both claimed the `entrypoint-gate-test` stage is never loaded by a daemon, which the new `load: true` and `docker run` step contradict, and the new step's own `--tmpfs` comment read one 2g ceiling as modelling two pod volumes with separate budgets. Both fixed in a5c23db. The other two are dispositioned in the section with reasons that argue about this change.

k8s-operator/internal/controller/platformagent_manifests.go
AntonTyb: Correcting something I implied in my last reply, and sharpening where we actually disagree.

I ran an adversarial pass over the rebased diff in a separate context, and it came back saying it could construct no concrete stale-state failure from the pod-lifetime `/tmp`, on the grounds that `HERMES_HOME` is set explicitly on all four invocations. I recorded that in the PR's Self-Review. **That reasoning is wrong and I have taken it out.** Checking it against source instead of accepting it:

- `deploy/shared/docker-entrypoint.sh:1265-1267` — the entrypoint's own comment, and the reason the step-4a compat symlink exists at all: *"hermes-otel resolves config below `~/.hermes` even when `HERMES_HOME` points elsewhere."* That is your point, stated by this repository.
- `agents/platform/scripts/profile_scaffold.py:79-85` — the scaffold does not merely inherit `HOME=/tmp`, it *forces* it onto every `hermes` subprocess it spawns: `env = {**os.environ, "HOME": "/tmp"}`, because "under RunAsNonRoot the real home is not writable". The shell-level `HOME=/tmp` at 640/690/713/873 is belt-and-braces on top of that.

So the mechanism holds: those subprocesses do consult `/tmp/.hermes`, and after this change that directory survives a restart and is visible to the dashboard container's boot. "`HERMES_HOME` is set, therefore `~/.hermes` is not load-bearing" is exactly the inference the compat symlink exists to refute.

Two things I checked while re-reading, both of which narrow the scope rather than widen it:

- **`$HOME` in the gateway is `/opt/data/home`, on the data PVC** — stated as an invariant at `deploy/shared/entrypoint_gate_check.sh:22`. So the durable `~/.hermes` that step 4a writes was *already* restart-persistent before this PR, and step 4a's `mkdir -p` is not hitting the newly read-only rootfs. What this change introduces is the second one at `/tmp/.hermes`, which used to be the ephemeral copy and no longer is.
- Consequently the delta is narrower than "the agent's config root becomes durable": it is that the `HOME=/tmp` scaffolding subprocesses lose their per-restart wipe. Still real, still worth closing.

Where that leaves the disagreement, which I think is now only about sequencing:

- **Lifetime and sharing** — certain, agreed, documented in the `docs/credential-isolation-design.md` bullet.
- **The read-back chain** — the mechanism is confirmed above. Whether attacker-controlled content under `/tmp/.hermes` actually changes what a scaffolding pass does is still unproven, which is what you labelled plausible rather than certain, and I am not claiming otherwise in either direction.
- **Where the fix goes** — still `deploy/shared/docker-entrypoint.sh` and still not this PR, for the reason I gave before and now believe more strongly: `profile_scaffold.py` forcing `HOME=/tmp` internally means a `mktemp -d` in the shell does not fix it on its own. The scaffold has to be changed too, and the entrypoint gate check needs a case that proves the wipe, or the fix lands untested. That is a change with its own blast radius across the gateway and the dashboard, not a rider on a manifests PR whose current safety property is that it reverts by flipping one field on three containers.

I have not filed the follow-up. Say the word and I will open it with the two source references above and a repro sketch, or open it yourself if you would rather word it — and if you would rather this block the PR, say so and I will move it in with the gate-check case first.

mplakhtiy (approved):

PR #679, opened 2026-08-13
fix(litellm): harden the vertex_ai provider landed in #644

adamparco:
# Pull Request

## Why This Change

This branch began as an independent implementation of Vertex AI as a model provider; #644 landed the same feature first. What remains here is the hardening #644's design still needs — starting with a real apply-breaker: kustomize strips the source quotes from `value: "${VERTEX_PROJECT_ID}"`, so an all-digit value (a GCP project **number** is a legitimate Vertex identifier) renders as a YAML integer into `EnvVar.value`, and the apiserver rejects the whole Deployment *after* earlier documents in the stream applied.

## What Changed

Merges `main` (adopting #644's API surface wholesale: `overlays/vertex_ai`, `VERTEX_PROJECT_ID`, the chart's `litellm.vertex.*` values, Terraform-created GSA, `gemini-3.5-flash` default) and adds on top:

**Files:** `k8s-operator/Makefile`, `k8s-operator/config/integrations/litellm/`, `k8s-operator/scripts/{common,provision_04,provision_09,teardown_04,teardown_09}*.sh`, `install.sh`, `.github/workflows/k8s-operator-test.yml`, and the corresponding docs.

**Added/Changed/Fixed:**

- **Numeric-scalar fix.** Operator-supplied scalars are substituted pre-quoted in `LITELLM_RENDER` — the one point in the pipeline that can still decide a node's YAML type, since kustomize re-serialises quotes away. Proven by a typed decode (`YAMLToJSON` → `encoding/json`, the apiserver-faithful path) of renders with all-digit values in every slot.
- **Content-addressed ConfigMap.** `litellm-config-sha256-<16hex>` from a checksum of the *substituted* config replaces reliance on the pod-annotation alone: the gateway rolls exactly when the rendered config changes, no-change deploys don't roll, and superseded ConfigMaps survive for `kubectl rollout undo`.
- **Deploy validation.** `make deploy-litellm` under `vertex_ai` refuses empty required variables instead of silently substituting empty strings; `provision_09` pre-flights the GSA and Workload Identity binding before deploying pods that would otherwise pass probes and 403 on every completion.
- **`SKIP_VERTEX_IAM_SETUP`.** For installs whose Vertex IAM is owned by a platform team or Terraform: provision_04 creates nothing, provision_09 skips the preflight, and teardown_04 leaves the GSA, binding, and grant standing (main's teardown arm was unconditional).
- **`VERTEX_LOCATION` defaults to `global`,** not the cluster region — regional endpoints return 429 for partner models with no local capacity (the failure mode #644's own live test avoided by using `global` explicitly).
- **Model-var override + persistence.** The four model variables are the deliberate exception to saved-state-wins: an export overrides `vars.sh` with a warning and is written back — closing #644's noted "sticky `MODEL_DEFAULT_NAME`" follow-up — and `install.sh` exports its resolved answers so a stale shell export cannot beat the `--model-provider` flag.
- **Identity rolls the pods.** The GSA email is mirrored into a pod-template annotation (changing only the ServiceAccount annotation otherwise rolls nothing, and running pods keep minting tokens for the old GSA); the KSA sets `automountServiceAccountToken: false`.
- **CI render gate.** All five providers render through the real `LITELLM_RENDER`; unsubstituted placeholders or unparseable YAML fail the job.

## Context

Supersedes the feature half of this branch; #644 is canonical for the provider itself. Overlaps deliberately resolved in #644's favor: overlay/variable/chart/Terraform naming, and the vertex default model.

## Testing

- `make -C k8s-operator test` green post-merge (Go suites + 133 Python tests); `go build`/`go vet` clean.
- `make docs-generate` committed; `make docs-check` clean; `prettier --check` clean; `helm lint --strict` clean; `bash -n` on all touched scripts.
- All five providers render placeholder-free and parse; a render with `VERTEX_PROJECT_ID=123456789012` typed-decodes cleanly (`value: "123456789012"` stays a string).
- Not run: `terraform validate` (no terraform binary on this machine; the Terraform files are #644's, untouched by the merge resolution).

### Live validation

Exercised against GKE cluster `platform-agent-host` (us-east4) in `adamparco-kage`, operator image `k8s-operator:dev-990b6a4-0812-202845`, serving `claude-opus-4-8` from a **second** project (`gke-horizon3-shared`) over the `global` endpoint — the cross-project arrangement #644 could not test:

- `provision_04` (exit 0): all sub-steps verified-and-skipped against live IAM — the idempotency path.
- `provision_09` (exit 0): created `litellm-config-sha256-0cdcd3ddeea0e114`, moved the Deployment's volume onto it, both replicas rolled; env landed as strings.
- In-pod `POST /v1/chat/completions` on `model-default` returned the requested completion — Claude on Vertex via Workload Identity, no API key anywhere.
- Re-run with identical variables: all six objects `unchanged`, zero pod restarts — rolls only on real change.
- Re-run with **no** model exports: identical render from `vars.sh` alone. This test caught the persistence bug fixed in `39161e9b`.

**Not covered live:** that pass predates the merge with #644, so it validated this branch's mechanisms (checksum roll, pre-quoting, preflight, WI path) but not the merged `overlays/vertex_ai` file layout, which is validated by render/decode only. Teardown scripts were not run destructively. Superseded ConfigMaps are left on the cluster by design; no other artifacts remain.

---

Risk is confined to `deploy-litellm` consumers: the first deploy after this change rolls the gateway once (ConfigMap name scheme changes); subsequent no-change deploys do not. Existing non-vertex installs need no variable changes.


.github/workflows/reusable-deploy-integrations.yml
```diff
@@ -104,6 +104,11 @@ jobs:
           CLUSTER_NAME: ${{ vars.GKE_CLUSTER_NAME }}
           MODEL_PROVIDER: ${{ vars.MODEL_PROVIDER }}
           MODEL_DEFAULT_NAME: ${{ vars.MODEL_DEFAULT_NAME }}
+          # Only consulted when MODEL_PROVIDER is vertex_ai, but passed on every
+          # run for the same reason as the LiteLLM job below: init_var_model_provider
+          # asks for both, and a CI checkout has no vars.sh to answer from.
+          VERTEX_PROJECT: ${{ vars.VERTEX_PROJECT }}
```
mplakhtiy, line 110: **HIGH — `VERTEX_PROJECT` is a variable nothing reads.**

Every consumer reads `VERTEX_PROJECT_ID`: `common.sh:322`, `provision_09_deploy_litellm.sh:105`, `teardown_09_deploy_litellm.sh`, and the Makefile's `LITELLM_REQUIRE_VERTEX_VARS`. `grep -rn "VERTEX_PROJECT\b"` finds no reader anywhere in the tree.

So on a `vertex_ai` CI environment the variable is dropped on the floor and `init_var_vertex_ai` silently falls back to `init_var "VERTEX_PROJECT_ID" "${PROJECT_ID:-}"` — the cluster's own project. The comment sitting directly above this line describes the resulting failure exactly: *"the pods roll onto the wrong project and every completion 403s while the rollout check below still goes green."*

Same misspelling at `:181` in this file, and in the two documents a human copies from:

- `k8s-operator/README.md:319` — `export VERTEX_PROJECT=project-serving-the-model`, followed by `make deploy-litellm`, which then exits 1 with `MODEL_PROVIDER=vertex_ai requires VERTEX_PROJECT_ID to be set and exported`. The recipe cannot work as written. Also `:294`, `:295`, `:312`.
- `docs/site/src/content/docs/concepts/inference-gateway.md:97` — same wrong export inside the new section's `provision_04 && provision_09` recipe; also `:84` and `:88`.

**Fix.** Rename all seven occurrences to `VERTEX_PROJECT_ID`.

terraform/examples/full-install/README.md
```diff
@@ -31,6 +31,11 @@ same cluster, service accounts, and IAM bindings.
   composed from your variables. `model_provider` selects which provider
   LiteLLM routes `model-default` to (set the matching `*_api_key` variable);
   `model_default_name` overrides the per-provider default model.
+  `model_provider = "vertex_ai"` takes no API key and instead needs
+  `vertex_project`, `vertex_location`, and `vertex_service_account` — a
+  plan-time precondition rejects the apply if any is empty. This composition does
```
mplakhtiy, line 36: **HIGH — this section documents variables, a precondition, and an ownership split that do not exist.**

The added section tells the reader to write:

```hcl
vertex_project         = "my-shared-serving-project"
vertex_service_account = "kubeagents-litellm-gsa@my-gcp-project.iam.gserviceaccount.com"
```

Against `terraform/examples/full-install/variables.tf`:

- The variable is `vertex_project_id` (`variables.tf:75`), not `vertex_project`. This same README already spells it correctly at `:49`, so the file now contradicts itself.
- There is no `vertex_service_account` variable at all.
- `grep -n precondition terraform/examples/full-install/*.tf` returns nothing. Both Vertex variables default to `""` and `main.tf:17-18` falls back to `project_id` and the cluster location, so the README's *"a plan-time precondition rejects the apply if any is empty"* — stated twice, at `:36` and `:138` — is not true.
- *"This composition does **not** create that GSA or its bindings, the one identity here you supply yourself"* (`:35-36`, `:126`) is the reverse of what `main.tf` does: `module.litellm_vertex_iam` (`:103`) creates `kubeagents-litellm-gsa` and its `roles/iam.workloadIdentityUser` binding, and `google_project_iam_member.litellm_vertex_user` (`:118`) grants `roles/aiplatform.user` on the serving project.

**Failure scenario.** A reader following this gets `Error: Unsupported argument` on two of the four lines, and is told to hand-build IAM that Terraform already builds.

Either rewrite the section against `main.tf` / `variables.tf`, or make the Terraform match the section — but the two cannot ship as they are.

.github/workflows/k8s-operator-test.yml
```diff
@@ -73,6 +73,83 @@ jobs:
...
+          # this print target render deliberately unguarded — so empty values stay
+          # reachable. They must not yield a stream kubectl cannot read: before
+          # the interpolated values were given fallbacks, an empty LITELLM_GSA_NAME
+          # rendered the service-account annotation as a plain scalar opening with
+          # `@`, a reserved YAML indicator that makes the whole document
+          # unparseable rather than merely wrong. Refusing to render is a fine
+          # answer and so is rendering something that still parses; emitting
+          # broken YAML is not.
+          if empty_out=$(PROJECT_ID= VERTEX_PROJECT_ID= VERTEX_LOCATION= \
+                         LITELLM_KSA_NAME= LITELLM_GSA_NAME= \
```
mplakhtiy, line 140: **HIGH — `actionlint` fails on the workflow step this PR adds.**

shellcheck reports `SC1007` five times ("Remove space after = if trying to assign a value") because a line-continuation follows the last bare `VAR=`. I read the `actionlint` job log to confirm: all five warnings are on this step. The check is red on this PR and was green before it, and this is the only shellcheck-visible shell the branch adds.

Spelling the empties as `VAR=''` leaves the intent unchanged:

```suggestion
          if empty_out=$(PROJECT_ID='' VERTEX_PROJECT_ID='' VERTEX_LOCATION='' \
                         LITELLM_KSA_NAME='' LITELLM_GSA_NAME='' \
```

.github/workflows/k8s-operator-test.yml
```diff
@@ -73,6 +73,83 @@ jobs:
...
+                         MODEL_PROVIDER=vertex_ai MODEL_DEFAULT_NAME=ci-model \
+                         make -s -C k8s-operator print-litellm-manifests); then
+            if ! printf '%s\n' "$empty_out" | ruby -ryaml -e 'Psych.parse_stream(STDIN.read)'; then
+              echo "::error::The vertex overlay renders unparseable YAML when its variables are empty. Quote or default the values it interpolates."
+              fail=1
+            fi
+          else
+            echo "vertex_ai refuses to render with empty variables — a loud failure, which is an acceptable answer here."
+          fi
+
```
mplakhtiy, line 150: **MEDIUM — the render gate cannot catch a regression of the bug this PR exists to fix.**

The gate substitutes `VERTEX_PROJECT_ID: ci-vertex-project` (`:88`) and validates with `ruby -ryaml -e 'Psych.parse_stream(STDIN.read)'`. Neither half exercises the numeric-scalar hazard:

- `ci-vertex-project` is not all digits, so dropping the `export VERTEX_PROJECT_ID="\"…\""` pre-quoting from `LITELLM_RENDER` still renders a string.
- Even with an all-digit value, `value: 123456789012` is *perfectly valid YAML*. `Psych.parse_stream` accepts it happily. Only the apiserver's typed decode into `EnvVar.Value` (a Go `string`) rejects it — and that is not what this gate runs.

The PR body describes exactly the right test ("a typed decode (`YAMLToJSON` → `encoding/json`, the apiserver-faithful path) of renders with all-digit values in every slot"), but no such test is committed; the checked-in gate is the weaker one.

**Fix.** Add an all-digits `VERTEX_PROJECT_ID` to the matrix and assert the resulting node decodes as a `String`:

```ruby
ruby -ryaml -e 'YAML.load_stream(STDIN.read).each { |d| … VERTEXAI_PROJECT value.is_a?(String) or abort }'
```

Without it, the one apply-breaker this branch was opened for has no regression cover.

install.sh
```diff
@@ -1637,6 +1637,18 @@ main() {
...
+  # provisioned install can switch providers), and that override reads the
+  # environment this installer inherited — the operator's shell. A stale
+  # `export MODEL_PROVIDER` left over from an earlier run would beat the
+  # --model-provider flag we just resolved and wrote, and the pipeline would
+  # provision a provider the confirmation panel never showed. Exporting the
+  # resolved answers makes the environment agree with the file.
+  export MODEL_PROVIDER="$model_provider"
+  export MODEL_DEFAULT_NAME="$model_default_name"
+  export VERTEX_PROJECT_ID="$vertex_project_id"
+  export VERTEX_LOCATION="$vertex_location"
```
mplakhtiy, line 1651: **MEDIUM — `VERTEX_LOCATION` still defaults to the cluster region on every path except the bare scripts.**

The `global` default lands in `common.sh:228` (`DEFAULT_VERTEX_LOCATION`) and is reached only by `init_var_vertex_ai`, and only when the variable is still empty.

`install.sh` resolves the location to `$region` (`:1251`, `:792`), offers `$region` as the interview default (`:1283`), writes it into `vars.sh` (`:1551`), and — new in this PR, on this line — **exports** it, which is the one thing that guarantees `init_var`'s `global` default can never be reached on an installer-driven install. `--help` at `:100` still reads `(default: --region)`.

The Helm and Terraform paths do the same: `charts/kube-agents/values.yaml:74-77` and `templates/litellm.yaml:15` fall back to `platformAgent.harness.location`; `variables.tf:81` falls back to the cluster location.

**Failure scenario.** The failure the PR names — *"regional endpoints return 429 for partner models with no local capacity"* — remains the default everywhere except a hand-run `provision_09`.

Either move the default into `install.sh`/`values.yaml`/`variables.tf` too, or narrow the claim in the PR body and the docs to the scripts path.

.agents/skills/install-kube-agents/SKILL.md
```diff
@@ -108,7 +108,9 @@ here. Run `./install.sh --help` for the authoritative list.
 | `--image-tag=TAG`             | SemVer release tag or full 40-character commit SHA              | Checkout `HEAD`; required via `curl \| bash` |
 | `--registry-prefix=PATH`      | Container registry path without a URL scheme                    | `common.sh` `DEFAULT_REGISTRY_PREFIX`        |
 | `--allow-unverified-source`   | Provision from a dirty or mismatched checkout                   | `false`                                      |
-| `--model-provider=NAME`       | `gemini` \| `anthropic` \| `chatgpt` \| `openai`                | `common.sh` `DEFAULT_MODEL_PROVIDER`         |
+| `--model-provider=NAME`       | `gemini` \| `anthropic` \| `vertex_ai` \| `chatgpt` \| `openai` | `common.sh` `DEFAULT_MODEL_PROVIDER`         |
+| `--vertex-project=ID`         | GCP project serving the Vertex AI models                        | _unset_ (required with `vertex_ai`)          |
+| `--vertex-location=LOCATION`  | Vertex AI location: `global`, a multi-region, or a region       | `common.sh` `DEFAULT_VERTEX_LOCATION`        |
```
mplakhtiy, line 113: **MEDIUM — this table names a flag that does not exist and a default that is wrong.**

`install.sh:147` parses `--vertex-project-id=*`; there is no `--vertex-project`. `install.sh:1250` defaults it to `$project_id`, so it is neither unset nor required. And per the `VERTEX_LOCATION` finding on `install.sh:1651`, the location default is `$region`, not `DEFAULT_VERTEX_LOCATION`.

AGENTS.md is explicit here: *"Verify identifiers against source, not against other docs."*

mplakhtiy (commented):
Review of `f4a39bf`. Eight findings inline: three High, five Medium. The three Lows are held back.

**Scope caveat.** The branch conflicts with its base, so this covers the PR as authored, not as merged. The single conflicted file is `k8s-operator/scripts/common.sh`, where this branch rewrites `load_state` and `default_model_for_provider` and base has since landed #686 (`REQUIRES_IMAGE_TAG` gating of `init_var_image_tag`, `39d125e8`) and #645 (`anthropic` default → `claude-opus-5`, `119c5e15`) in the same two hunks. Defects arising from the *interaction* are out of scope here — but after the resolution, worth re-checking that (a) the model-reapply loop still runs on steps that do not set `REQUIRES_IMAGE_TAG`, and (b) the provider-switch reset writes `claude-opus-5` for anthropic. Neither is verifiable from the branch as authored.

The core mechanisms are sound, and I confirmed from the `Run Controller Tests` log that the new "Verify LiteLLM overlays render cleanly" step executed and passed — which is what establishes that kustomize does strip the source quotes and that pre-quoting in `LITELLM_RENDER` produces parseable YAML.

k8s-operator/scripts/common.sh
```diff
@@ -296,15 +307,133 @@ init_var_model_provider() {
...
+  fi
+
+  if ! out="$(gcloud iam service-accounts get-iam-policy "${gsa_email}" --project="${PROJECT_ID}" --format="json" 2>&1)"; then
+    LITELLM_VERTEX_IAM_REASON="the IAM policy of ${gsa_email} could not be read: $(printf '%s\n' "$out" | head -n 1)"
+    return 2
+  fi
+  if ! printf '%s\n' "$out" | grep -F -q "${wi_member}"; then
+    LITELLM_VERTEX_IAM_REASON="${gsa_email} has no Workload Identity binding for ${wi_member}"
+    return 1
+  fi
```
mplakhtiy, line 392: **MEDIUM — the Workload Identity preflight matches the member under any role.**

```bash
if ! printf '%s\n' "$out" | grep -F -q "${wi_member}"; then
```

This is a substring match over the whole policy document, so it says nothing about which role the member holds. A GSA whose policy binds `serviceAccount:PROJECT.svc.id.goog[kubeagents-system/kubeagents-litellm]` under, say, `roles/iam.serviceAccountUser` — a plausible hand-configuration mistake, and exactly the mistake this check exists to catch — passes.

That matters more than the other two checks because this is one of only two conditions the function returns `1` for, and `preflight_vertex_iam` treats `1` as blocking precisely on the grounds that it is *"a fact"*.

**Fix.** Scope the query instead of grepping the document:

```bash
gcloud iam service-accounts get-iam-policy "${gsa_email}" --project="${PROJECT_ID}" \
  --flatten="bindings[].members" \
  --filter="bindings.role=roles/iam.workloadIdentityUser AND bindings.members=${wi_member}" \
  --format="value(bindings.members)"
```

docs/site/src/content/docs/concepts/inference-gateway.md
```diff
@@ -54,15 +54,76 @@ The two substituted values come from provisioning (`MODEL_PROVIDER` and `MODEL_D
...
 export MODEL_PROVIDER=gemini
 export MODEL_DEFAULT_NAME=gemini-3.5-flash
 cd k8s-operator/scripts && ./provision_09_deploy_litellm.sh
 ```
 
-This rewrites the LiteLLM `ConfigMap` and rolls the gateway; the agent picks up the new model on its next request without any change to its own config.
+Exporting is what makes this work on a cluster that has already been provisioned. Provisioning caches every answer it took in `k8s-operator/scripts/vars.sh` and re-uses it on later runs, so both variables above already hold whatever the first run chose. The model and Vertex variables are the deliberate exception to saved-state-wins: an explicit export beats the cached value, the step warns you that it is overriding it, and it re-saves the new value so the later steps agree. What you leave unset keeps its cached value — except that exporting `MODEL_PROVIDER` on its own moves `MODEL_DEFAULT_NAME` to the new provider's default rather than carrying the old provider's model name across.
+
+That deploys a fresh LiteLLM `ConfigMap` and rolls the gateway onto it; the agent picks up the new model on its next request without any change to its own config.
+
```
mplakhtiy, line 68: **MEDIUM — this new Vertex section duplicates and contradicts the section immediately below it.**

The PR inserts "Anthropic models on Vertex AI" directly above the pre-existing "Vertex AI and Model Garden" (`:128-137`), which #644 wrote and this PR leaves untouched. The older section now disagrees with the branch on both facts it states:

- `:130` — "`MODEL_PROVIDER=vertex` routes `model-default` to Vertex AI" — the selector is `vertex_ai` (`common.sh:222`).
- `:135` — "`VERTEX_PROJECT_ID` and `VERTEX_LOCATION` (defaulting to the install's project and region)" — the very default this branch changes to `global`.

AGENTS.md: *"Every fact has one home. Duplicating documentation across files is how it goes stale."* Two sections on one page is the same failure. Fold them into one.

Separately, `:79` links `…/k8s-operator/config/integrations/litellm/overlays/vertex` — a directory that does not exist; it is `overlays/vertex_ai`. It is an absolute URL, so `make docs-check`'s relative-link pass does not see it.

PR #691, opened 2026-08-13
feat(ci): validate every asset and skill reference the agent is told to open

jayantid:
## Summary

This adds `make prompt-check`, a stdlib-only static check that opens all 98 of the agent's instruction files — personas, governance SOPs, skill manifests, cron prompts — and verifies that every asset path, skill name and cron reference they tell the agent to open resolves to something the agent will actually hold at runtime. It runs as its own job in `validate.yml`, takes 0.2s, and reports findings as GitHub annotations on the offending line. The substance is the resolution model: these references resolve against the image layout, not against the checkout, so a reference that names a real file in this repository is still a finding when nothing copies that file into the profile that cites it. On its first honest run it found a live bug — a skill deferring four times to a directory whose name is one character off — plus two classes of reference that name real files here and no file in the pod. All are fixed in this PR.

## Why This Change

The agent's instruction layer — personas, 14 governance SOPs, 42 skill bundles, cron prompts — is prose, and prose is not compiled. A persona that cites a renamed skill, or an SOP that names a moved script, merges clean and fails at 06:20 inside an agent, as a slightly worse answer rather than an error.

That layer is uniquely bad at failing loudly. Nobody reads it at review time, no human sees it fail, and the agent's own competence hides the break: told to load a skill that does not exist, it improvises something plausible instead. There is no stack trace, no exit code, and no user complaint — just a watchdog that quietly stopped doing one of its checks. Every other check in this repository guards something a human eventually notices; this one guards work the agent skips without telling anyone. It found a live bug on its first honest run, in a skill that has been merged for a while, and it is cheap enough not to argue about: 0.2s, stdlib only, no network, no cluster.

This adds the compiler for that layer.

## What Changed

- **Four rules** across all three profiles: backticked asset paths must resolve; skills named in prose must exist; every `SKILL.md` must have frontmatter whose `name` matches its directory and a non-empty `description`; cron jobs must cite SOPs and skills that exist.
- **A resolution model, not a path check.** These paths do not resolve against the file that cites them — they resolve against the skill bundle, then the profile home, then the `/opt/defaults` layer. **Not** against the repository root: a checkout is not mounted into any agent, so a reference beginning `agents/`, `deploy/` or `scripts/` is a finding even though the file is right there — that is the one spelling that survives every human review and is always wrong at runtime, and it was live eighteen times. And **not** against the whole of a profile's source tree: `docker-entrypoint.sh:253` copies `/opt/defaults/.` into `$TARGET_DIR`, which is the _default_ profile's home and no other, while a specialist home is built from a fixed item list (`profile_scaffold.py --items` for platform, `OVERLAY_ITEMS` for a cluster). `docs/` is in neither list, so `docs/glossary.md` reaches the default profile alone. The model is derived from `deploy/docker/Dockerfile` and the entrypoint, and guarded by `test_opt_defaults_matches_the_dockerfile` and `test_profile_home_items_match_the_entrypoint`, which fail when a `COPY` or an item list changes without updating it.
- **No allowlist.** A backticked token counts as a path only if it has an asset extension _and_ its first segment names a directory that actually exists under one of that file's resolution roots. Derived rather than listed, so a new asset directory is covered the day it is added, while `roles/container.admin`, `kubernetes.io/ingress.global-static-ip-name`, and a sentence about someone else's `manifests/vendor/x.yaml` stay quiet. Naive matching produced 17 findings, roughly 16 of them noise.
- **Findings are annotations, not log lines.** Under GitHub Actions each finding is emitted as `::error file=…,line=…`, so it lands on the offending line in Files Changed. For a layer with no human reader at review time, a red check the author has to go digging in a log for is most of the way to no check at all.
- **Fixed (skill that does not exist):** `gke-cluster-autoscaler/SKILL.md` told the agent four times to defer to a `gke-compute-class` skill. The skill is `gke-compute-classes` — the spelling every other reference in the tree uses.
- **Fixed (file the agent cannot open):** `governance/stockout_prevention_sop.md` — the SOP the `stockout-prevention` cron job is told to read _in its profile home_ — cited `agents/platform/skills/gke-compute-classes/references/…` eighteen times, once or twice per finding section. That path names a real file in the checkout and no file in the pod: the tree is COPYed to `/opt/platform-template/skills`, so the profile home holds `skills/`, never `agents/`. Rewritten to the in-profile spelling. Three further provenance citations in `SOUL.md` and the glossary named repository tooling and the _Chat_ Agent's persona, neither of which reaches the Platform Agent; they now name the document instead of a path.
- **Fixed (path that works for one profile only):** three personas offered `docs/gcp-console-links.md` / `docs/glossary.md` "in the workspace" as an alternative to the absolute `/opt/defaults/…` path they give first. Only the absolute form works from a specialist profile home, so the parenthetical is dropped.

Two files carry the change: `scripts/check_prompt_assets.py`, the checker, and `scripts/test_check_prompt_assets.py`, 47 tests against it. `Makefile` gains a `prompt-check` target and `.github/workflows/validate.yml` a job that runs it; the rest of the diff is the prose corrections above.

### Coverage added

Before this PR, **zero** automated assertions covered the instruction layer. `docs-check-links` does not reach it (these are not documents and the docs map deliberately does not inventory them), and no test opened them.

After, every merge validates:

| Rule                | What it asserts                                                                                                                   | Count          |
| ------------------- | --------------------------------------------------------------------------------------------------------------------------------- | -------------- |
| `asset-path`        | backticked file references resolve to a file that exists in the repository, at a path the image layout says lands in that profile | 53 references  |
| `skill-ref` (prose) | a skill named in an instruction is a real bundle                                                                                  | 32 references  |
| `skill-ref` (cron)  | a job's `skills` array names real bundles                                                                                         | 8 declarations |
| `cron-asset`        | a prompt's quoted path, at any depth, resolves in that job's profile home                                                         | 7 references   |
| `skill-manifest`    | `SKILL.md` present, frontmatter `name` matches its directory, `description` non-empty                                             | 42 bundles     |

**100 cross-references and 42 manifests, over 98 files, on every PR.** The file corpus:

|                                                      | files                          |
| ---------------------------------------------------- | ------------------------------ |
| `SKILL.md` + `references/`                           | 73                             |
| governance SOPs                                      | 14                             |
| personas (`SOUL.md`, `AGENTS.md`, `CAPABILITIES.md`) | 8                              |
| profile docs (glossary, console links)               | 3                              |
| **by profile**                                       | platform 87, cluster 9, chat 2 |

Skill bundles are counted and resolved per profile — platform 36, cluster 6 — because the two share six directory names and the Cluster Agent is not given the other thirty. A cluster document naming a platform-only skill is a finding, not a pass.

Plus 12 cron jobs across the platform and chat rosters, found by glob rather than listed, and counted in the success line so a scope that narrows is visible in the log rather than indistinguishable from a clean pass.

These counts are printed by the check itself at every run rather than hardcoded, so they grow with the tree — a new skill bundle is covered the day it merges, with no file to update.

### What this does _not_ cover

Stated plainly so the gate is not read as more than it is:

- **Cron prompt geography** — line counts, section spans, roster membership. Deliberately excluded: `test_cron_prompts_cite_the_real_sop_geography` in `agents/platform/skills/fleet-audit/scripts/test_audit_report.py` already re-derives all of it for the seven fleet-audit streams and already runs in CI. I verified it independently (adding a line to an SOP fails it with `407 != 406`). The module docstring records the split so the next person does not rebuild the wrong half.
- **Skill names not followed by the word "skill".** `SKILL_IN_PROSE` anchors on `` `foo-bar` skill``, which is what keeps it from flagging every hyphenated token in the tree. So in the bug fixed here it flags 2 of the 4 occurrences — enough to fail the build and land the author in the right file, but the annotation count is not the occurrence count.
- **Whether the reference is the _right_ one.** A path that resolves to the wrong-but-existing file passes.
- **Prose that names no file at all** — a skill silently dropped from an SOP's roster is invisible to this.
- Plugin and directory `README`s, which are developer documentation about Hermes internals, do not ship in any profile, and are already covered by `check_docs_links.py`.

**It is a static check on the repository, not an observation of the pod.** Worth being explicit, because the `asset-path` rule is easy to read as stronger than it is. The chain it asserts is: the reference resolves _at a path the profile will hold_ → `deploy/docker/Dockerfile` COPYs that directory into the image → the entrypoint lays it down in the profile → therefore the agent has it. Steps two and three are _modelled_, and the model has four known holes:

- **`.dockerignore` is not modelled.** It excludes `**/vars.sh`, an asset-suffixed file, so a reference to one would resolve here and be absent from the image. I checked all 53 accepted references against every dockerignore pattern that can match an asset suffix: 0 hits today, but nothing prevents the next one.
- **`cp -ru` is mtime-driven and can skip.** The entrypoint documents this itself at lines 295, 317 and 354 — it "skips a destination that looks newer" and "races an image roll". `/opt/data` may be a persistent volume, so a file that _is_ in the image is not guaranteed to be what the agent reads.
- **`profile_overlay.py`** merges the operator's per-profile config overlays past the plain copy; only the copy is modelled. (`profile_scaffold.py`'s item list _is_ modelled now, in `PROFILE_HOME_ITEMS`, with a drift test against the entrypoint.)
- **`COPY --from=<stage>`** sources another build stage, not this repo. Now skipped by `_copy_instructions` rather than having its flag dropped, with a regression test — previously it would have been read as a repo path and quietly agreed with a wrong model. The two in the tree today both target `/usr/local/bin`, so there was no live exposure.

None of this weakens the rule against the failure it exists for — a renamed skill or a moved script is a repo-level fact, and the repo is exactly where it is visible.

## Context

The check runs as **its own job** in `validate.yml` rather than a step. It needs none of that job's Helm, Terraform or pip setup, so it reports in seconds instead of behind them; and as a step it would be masked — a broken `terraform fmt` means these findings never print, the author fixes Terraform, pushes, and only then discovers the prompt errors. A separate job gets its own status line and cannot be hidden by a failure it has nothing to do with.

`validate.yml` rather than `docs-check.yml` because these are runtime assets, not documents. An earlier revision of this PR put the step in `docs-check.yml`, which falsified `docs/README.md:96` ("`make docs-check` runs the same checks as `.github/workflows/docs-check.yml`") and required three further edits to repair. Moving it removed all three.

Deliberately unfiltered. A paths filter is precisely what kept the Python suite from running at all (see the comment in `python-tests.yml`), and a check on prose that only starts when prose changes would miss the rename that broke it — because that rename lands in a directory name.

Every round of review on this PR found a real defect in the model behind the checker, of exactly the class the checker exists to catch: the repository root being tried as a resolution root; `/opt/defaults` being treated as reaching every profile; `check_cron_assets` bypassing the profile-home model; a cron path regex that could not span more than one `/`; roster discovery that could silently empty; and `description: >-` with the body deleted passing the rule that requires a description. Each was reproduced before it was fixed, and each carries a test derived from the source of truth — the Dockerfile `COPY` lines, the entrypoint's item lists, a glob of the tree — rather than a hardcoded floor.

## Testing

- `make prompt-check` → `Prompt assets OK: 98 instruction files, 42 skill bundles (cluster 6, platform 36), 12 cron jobs in 2 rosters, all references resolve.` (0.2s)
- `python3 -m unittest test_check_prompt_assets` → **47 tests, OK** — 21 synthetic-profile, 7 annotation, 6 checkout-path, 5 repository, 4 profile-isolation, 4 resolution-model.
- `make docs-check` → clean
- `test_audit_report.py` (503 tests) → OK, re-run because this edits an SOP the geography check measures. The rewrites are within-line, so no section span moved.
- **Green on current `main`.** Merged `origin/main` at `36c35a3` into this branch in a scratch worktree: the gate exits 0 with the line above, and all 47 tests pass. It is not carrying pre-existing debt into CI.
- **Negative testing, not just green runs.** Each of the four rules was proven to fire against a synthetic profile fixture (renamed script, missing SOP, unknown skill in prose, frontmatter name/directory mismatch, missing manifest). The Dockerfile drift guard was proven by injecting a `COPY`. The geography check I decided not to duplicate was proven by injecting a line into an SOP. All injections reverted.
- **The annotation path was exercised end to end** by re-injecting the real bug. `GITHUB_ACTIONS=true` produced, on stdout:
  ```
  ::error file=agents/platform/skills/gke-cluster-autoscaler/SKILL.md,line=19,title=prompt skill-ref::names the 'gke-compute-class' skill, which does not exist; did you mean 'gke-compute-classes'?
  ```
  The `did you mean` hint is difflib at cutoff 0.85, for exactly this single-character class of slip.
- **Precision cases are tests, not assertions.** `test_identifiers_that_merely_contain_a_slash_are_not_paths`, `test_runtime_state_paths_are_not_asset_references`, `test_a_description_folded_onto_following_lines_is_accepted` (prettier reflows the long ones), and `test_the_defaults_layer_reaches_the_default_profile_and_no_other` all lock in behaviour that earlier revisions got wrong.
- **The discovery set is compared against the tree, not against a floor.** `test_every_skill_bundle_in_the_tree_is_discovered` diffs what the checker opens against `agents/*/skills/*/SKILL.md` both ways, and `test_every_cron_roster_in_the_tree_is_discovered` does the same for rosters, so neither a missed one nor an invented one passes; `test_it_is_actually_looking_at_something` still floors the file count so the check cannot pass by discovering nothing.
- **Not run:** `prettier --check` and `actionlint`. Neither `prettier`/`npx`/`npm`/`node` nor `actionlint` is available on this machine, and there is no `yaml` module to parse the workflow with either. The repo has no prettier config, so defaults apply and `proseWrap: "preserve"` means the changed `.md` lines should not be reflowed — but that is reasoning, not a verified result. The new `validate.yml` job mirrors the existing job's shape and SHA-pins `actions/checkout` per repo policy. Please let the prettier and actionlint workflows be the judge.
- **Pre-existing, unrelated:** `make test-python` fails in `admin_console/tests/`, `agents/chat/scripts/`, `agents/platform/scripts/` and `deploy/docker/` for missing third-party imports (fastapi, mcp, streamlit, …). Confirmed pre-existing by stashing this branch. `scripts/` runs clean.

### Live validation

**Not live-tested.** The checker is repository tooling — it runs in CI and on a developer's machine, never inside a pod, and it is not baked into any image, so there is no CR `.status`, Deployment env, or in-pod file for it to change.

The one change here that _does_ reach a running installation is the `gke-cluster-autoscaler/SKILL.md` correction, which ships in the platform image. Observing it would mean rebuilding and rolling the image and then getting the agent to route an autoscaler question through the deferral sentence — and the observable difference is whether the agent finds a skill directory, which I would be inferring from its answer rather than measuring. I do not have an install to do that against. The static evidence is in the diff: every one of the 20-odd other references in the tree spells it `gke-compute-classes`, and `agents/platform/skills/gke-compute-class/` does not exist.

No test artifacts created; no cluster state touched.

## Risk & Rollout

**Blast radius.** No runtime code path is touched. The checker never ships in an image and never runs in a pod, so at runtime the change is the prose corrections alone: four content files, each replacing a reference the agent cannot open today with one it can. The remaining diff is CI-side.

**The one new failure mode is a false positive** in the `prompt-assets` job blocking a PR that is actually fine. The heuristic is deliberately quiet — a token counts as a path only when its first segment names a real directory under one of that file's resolution roots — so the expected direction of error is a missed finding rather than a spurious one, and the annotation names the file, the line and the spelling that would work.

**Revert** is ordinary and total. The gate is a single `make prompt-check` step in one job of `validate.yml`; deleting the job disables it and touches nothing else. Reverting the whole PR restores the previous state exactly — there is no state, no migration, and no consumer of the checker outside CI and `make`.

**At merge time, nothing has to happen.** No image rebuild, no operator roll, no config change; the gate starts applying to the next PR opened after merge. The prose corrections reach agents on the next image build like any other content change.

The thing genuinely worth weighing in review is whether the precision contract is the right one. It is designed to be quiet by default and to grow coverage as the tree grows, which means a genuinely broken reference in a form it does not recognise will pass. I would rather that than an allowlist nobody maintains.

---


scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,560 @@
...
+    skills: dict[str, Path] = {}
+    for profile in PROFILES:
+        home = REPO / "agents" / profile / "skills"
+        if not home.is_dir():
+            continue
+        for child in sorted(home.iterdir()):
+            if child.is_dir():
+                skills[child.name] = child
+    return skills
+
```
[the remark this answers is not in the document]
jayantid: Confirmed and fixed in c129f7a — thank you, this was a real bug, and of exactly the class the checker exists to catch.

Reproduced before fixing: 42 `SKILL.md` on disk, 36 in the dict. `agents/platform/skills` and `agents/cluster/skills` share six directory names (`gke-observability`, `gke-reliability`, `gke-storage`, `gke-workload-scaling`, `gke-workload-security`, `gke-workload-troubleshooting`), and since `PROFILES = ("platform", "chat", "cluster")` iterates platform first, cluster overwrote it. Renaming the frontmatter `name` in platform's `gke-storage/SKILL.md` to `gke-storage-fleet` still printed `Prompt assets OK` and exited 0.

The second half you didn't call out is worse: with one flat namespace, all 36 platform-only skills were visible to cluster documents, so `agents/cluster/SOUL.md` naming `fleet-audit` would have passed — and the Cluster Agent is never given that bundle.

Fix is per-profile keying: `skill_directories()` now returns `{profile: {name: dir}}`, `check_skill_refs` resolves against the citing file's own profile, and `check_skill_manifests` iterates every bundle in every profile. A name that exists elsewhere gets that in the message rather than a bare "does not exist":

```
agents/cluster/SOUL.md:85: [skill-ref] names the 'fleet-audit' skill, which the cluster
profile does not have; it exists in platform, which this profile does not receive
```

Guarding it three ways:

- `test_every_skill_bundle_in_the_tree_is_discovered` diffs the checker's discovery set against `REPO.glob("agents/*/skills/*/SKILL.md")` in both directions. The floor it replaces (`> 20 skills`) is what let six missing bundles through — a floor is not a coverage assertion.
- `test_the_same_name_in_two_profiles_is_two_bundles` asserts the collisions resolve to distinct directories, and fails loudly if the tree ever stops having any.
- New `ProfileIsolationTests` (4) covers shadowing, cross-profile citation, the wrong-profile wording, and that a typo still gets the nearest name *within its own profile*.

Impact on the tree today: zero new findings — chat has no `skills/` directory, and all 8 cron skill declarations are in-profile. `Prompt assets OK: 98 instruction files, 42 skill bundles (cluster 6, platform 36)`. Description updated; the 36 published there was wrong.

scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,623 @@
...
+        # `./skills/fleet-audit/scripts/audit_report.py` for its own script.
+        # Both are right, and both are wrong if `./` is read literally.
+        bare = ref[2:]
+        if (path.parent / bare).exists():
+            return True
+        return any((root / bare).exists() for root in roots)
+    if ref.split("/", 1)[0] in REPO_ROOTS and (REPO / ref).exists():
+        return True
+    if any((root / ref).exists() for root in roots):
+        return True
```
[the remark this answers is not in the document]
jayantid: Confirmed and fixed in 956fedb. The ordering diagnosis is exactly right, and this was live in the tree.

Measured before fixing: **21** references depended on that branch, not 12 — some `**Reference:**` lines carry two citations. Isolated by re-running resolution with the branch disabled:

| file | refs | what it is |
| --- | --- | --- |
| `governance/stockout_prevention_sop.md` | 18 | `**Reference:**` lines — a directive to open the file |
| `agents/platform/SOUL.md:47` | 1 | provenance for a claim about image behaviour |
| `agents/platform/docs/glossary.md:39,66` | 2 | citations of the *chat* profile's persona |

Only the SOP set is a functional break, and it is the one you named: the `stockout-prevention` job is told to read that SOP in its profile home, and eighteen times the SOP points it somewhere the profile has no directory for.

**One correction on the proposed remedy.** Trying the repo root *after* the profile roots does not fix it — resolution is a boolean OR, so reordering changes which branch answers, not the answer. Those eighteen would still pass. Only removing the branch works, so that is what I did.

`REPO_ROOTS` stays in `_looks_like_reference`, which is the part you'd want kept: it is what makes such a token count as a path at all, so the reference is now *reported* rather than falling out of scope.

Findings hand back the rewrite, since the author demonstrably checked that the file exists and "does not resolve" otherwise reads as the checker being wrong:

```
agents/platform/governance/stockout_prevention_sop.md:91: [asset-path]
'agents/platform/skills/gke-compute-classes/references/compute-class-prioritization.md'
does not resolve to a file the agent will have; that is its path in the checkout,
not in the profile -- write 'skills/gke-compute-classes/references/compute-class-prioritization.md'
```

The rewrite is offered only for the citing file's **own** profile. Stripping `agents/chat/` off a path in a platform document leaves `SOUL.md`, which resolves — to the platform persona, a different document. A hint that converts a broken reference into a wrong one is worse than none, so cross-profile paths get `it belongs to the chat profile, which is a separate home this agent cannot read`, and repository tooling gets `nothing copies it into this profile`.

Content fixes: 18 SOP citations rewritten to `skills/gke-compute-classes/references/…`; the three provenance references now name the document (`the Chat Agent's SOUL.md §2 step 5`, `the kanban_scheduling build patch`) rather than a path the agent cannot open. Re-ran `test_audit_report.py` (503 tests, OK) because the geography check measures that SOP — the rewrites are within-line, so no section span moved.

New `CheckoutPathTests` (5) covers all of it, including that the in-profile spelling of the same file passes and that the cross-profile rewrite is *not* offered. 38 tests total; `56 references` now, down from 59 — the three that dropped were the provenance citations, which are no longer paths.

You were also right that this was not among the four disclosed modelling holes. It is no longer a hole, so the description now states the resolution order positively and says the repository root is not one of the roots.

scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,671 @@
...
+    #
+    # There is deliberately no repository-root branch. `agents/platform/...`
+    # names a real file in the checkout and no file the agent has: the tree
+    # under agents/platform/skills is COPYed to /opt/platform-template/skills,
+    # so the profile home has `skills/`, not `agents/`. Accepting a checkout
+    # path here would pass the one spelling that is guaranteed wrong at
+    # runtime -- and REPO_ROOTS is still what makes such a token count as a
+    # path in the first place, so it is reported rather than ignored.
+    return _resolve_opt(ref) is not None
+
```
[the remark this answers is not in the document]
jayantid: Confirmed and fixed in ed294f0. The mechanism is exactly as you describe, and the comment on `OPT_DEFAULTS` was simply false — I wrote "for *every* profile" and never checked it against `TARGET_DIR`.

Verified each link in the chain:

- `docker-entrypoint.sh:4` — `TARGET_DIR="${PLATFORM_AGENT_HOME:-/opt/data}"`, and line 253 copies the layer there. That is the **default** profile's home.
- Platform: `--items "config.yaml SOUL.md AGENTS.md CAPABILITIES.md cron skills governance hindsight"` (line 640). Cluster: `OVERLAY_ITEMS = ("SOUL.md", "AGENTS.md", "CAPABILITIES.md", "config.yaml", "skills")`. No `docs` in either.
- `Dockerfile:835` copies `agents/platform/docs/{glossary,gcp-console-links}.md` to `/opt/defaults/docs/` and nowhere else. `/opt/platform-template/` gets SOUL/AGENTS/CAPABILITIES, `skills/`, `cron/`, `governance/`, `hindsight/`, `config.yaml` — no `docs/`.
- The `scripts` symlink at line 538 is the tell you named, and it settled it for me: the entrypoint would not need to link `$TARGET_DIR/scripts` into `profiles/platform` if that layer reached the profile on its own.

**Live exposure was 3 references**, both halves of the finding included. I measured by disabling each root in turn:

| ref | via | |
| --- | --- | --- |
| `agents/cluster/SOUL.md:63` `docs/gcp-console-links.md` | `/opt/defaults` fallback | the one you named |
| `agents/platform/SOUL.md:128` `docs/gcp-console-links.md` | over-wide profile root | second half of the finding |
| `agents/platform/AGENTS.md:9` `docs/glossary.md` | over-wide profile root | " |

All three follow the same shape: the persona gives `/opt/defaults/docs/…` **first** and then offers `docs/…` "in the workspace" as an alternative. The absolute path works from any profile; the parenthetical works only from the default profile's home. Dropped the parenthetical in all three.

Modelled it rather than special-cased it. New `PROFILE_HOME_ITEMS` gives what each specialist home actually holds, `chat` deliberately absent because it *is* the default profile and the layer resolves there in full. `scripts` is listed for platform with the symlink as the stated reason. `test_profile_home_items_match_the_entrypoint` re-derives both lists from `--items` and `OVERLAY_ITEMS` and asserts the symlink line, so widening either by hand fails there.

Findings name the spelling that works:

```
agents/cluster/SOUL.md:63: [asset-path] 'docs/gcp-console-links.md' does not resolve to a
file the agent will have; /opt/defaults/docs/gcp-console-links.md is copied to the default
profile's home and to no other, so the cluster profile has no 'docs' -- write
'/opt/defaults/docs/gcp-console-links.md'
```

On `test_the_console_links_doc_resolves_through_the_image`: you were right that it pinned the wrong model, and the docstring repeated the false claim. Replaced with `test_the_defaults_layer_reaches_the_default_profile_and_no_other`, which asserts the `TARGET_DIR` assignment and the `cp -ru` line verbatim and that `docs` is in no specialist item list — and says in its docstring what the earlier revision got wrong, since that is the part a future reader needs.

40 tests; `53 references` now, down from 56 (the three dropped parentheticals). Re-ran `test_audit_report.py` (503, OK). The description's modelling-holes list drops `profile_scaffold.py`, which is now modelled.

Three findings, three real bugs, all in the class this PR exists to catch — thank you for the persistence on this one.

agents/platform/AGENTS.md
```diff
@@ -6,7 +6,7 @@ This folder is home. Treat it that way.
 
 Use runtime-provided startup context first, including `AGENTS.md` and `SOUL.md`.
 Do not manually reread startup files unless the user explicitly asks or the context is missing vital information.
-A glossary of agentic terms lives at `/opt/defaults/docs/glossary.md` (or `docs/glossary.md` in the workspace). Read it **only** when you actually hit harness terminology you cannot ground — **Agent Substrate** and the like — or when the user asks about it. Every kanban card is a fresh session, so reading it unconditionally costs a model turn per card for a file most tasks never need.
+A glossary of agentic terms lives at `/opt/defaults/docs/glossary.md`. Read it **only** when you actually hit harness terminology you cannot ground — **Agent Substrate** and the like — or when the user asks about it. Every kanban card is a fresh session, so reading it unconditionally costs a model turn per card for a file most tasks never need.
```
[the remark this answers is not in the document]
jayantid: Pushing back on this one: the edit did land, and the gate passes on the merged tree.

The premise is that `agents/platform/AGENTS.md:9` still carries the parenthetical. It does not, at any of the places worth checking:

- **GitHub's own copy at the PR head.** `GET /contents/agents/platform/AGENTS.md?ref=ed294f0` returns:
  > A glossary of agentic terms lives at `` `/opt/defaults/docs/glossary.md` ``. Read it **only** when you actually hit harness terminology…
- **The check you predicted would be red.** `prompt-assets` on that same SHA: `success` at 01:30:36Z, 16 minutes before this review was submitted.
- **The merged tree, not just the branch.** `main` has moved since this branch's base (now `bd0fd11`), so I merged it into a scratch worktree and ran the gate there rather than assume: `Prompt assets OK: 98 instruction files, 42 skill bundles (cluster 6, platform 36), all references resolve.` 40 tests OK. `grep -rn "in the workspace" agents/` returns one unrelated hit in `gke-upgrades/SKILL.md`.

The resolution trace in the finding is right in every step — `instruction_files()` returns the file, `_looks_like_reference` accepts on suffix `.md` and head `docs`, and `_resolves_in_home` returns `False` because `PROFILE_HOME_ITEMS["platform"]` omits `docs`. That is exactly the behaviour the previous round added, and it is why the line was rewritten in the same commit. The trace is sound; only its input is stale.

Best guess at the cause: the comment is anchored at `AGENTS.md:10` while the change is at line 9, and it describes a "merged state" — so this looks like a merge preview reconstructed against the older base rather than a read of the head blob. Worth flagging upstream, because a stale snapshot presented as the merged state is the one failure mode that turns a green run into a false 🔴 High.

No code change. Happy to be shown otherwise if you can point at a SHA where that line still reads the old way.

scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,745 @@
...
+    """
+    findings = []
+    quoted = re.compile(r"['\"`]([a-z_]+/[A-Za-z0-9_.-]+\.(?:md|json|py|sh))['\"`]")
+    for path, jobs in cron_rosters():
+        rel = path.relative_to(REPO)
+        profile = _profile_of(path)
+        home = REPO / "agents" / (profile or "platform")
+        for job in jobs:
+            for ref in set(quoted.findall(job.get("prompt", ""))):
+                if (home / ref).exists() or _resolve_opt(ref) is not None:
```
toshiowang: **Bug (latent, but it removes the rule's stated guarantee).** This resolves against the whole of `agents/<profile>/` rather than through `_resolves_in_home`, so it ignores `PROFILE_HOME_ITEMS` — the model this PR added specifically to catch "a path that works for one profile only".

Reproduced on this branch: appending `Read 'docs/glossary.md'.` to the `compliance-audit` prompt exits **0**, while the identical string in `agents/platform/governance/stockout_prevention_sop.md` exits **1** with the correct `write '/opt/defaults/docs/glossary.md'` hint. The message this rule would print — "which is not in its profile" — is exactly what it does not verify.

Verified fix (repo still clean, the probe now fails, 40/40 tests pass):

```suggestion
        for job in jobs:
            for ref in set(quoted.findall(job.get("prompt", ""))):
                # The profile-home model the asset-path rule uses, rather than
                # the whole of agents/<profile>/.
                if _resolves_in_home(ref, path):
```

scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,745 @@
...
+def check_cron_assets() -> list[Finding]:
+    """Every SOP a cron prompt sends its worker to has to be there.
+
+    The geography of those SOPs -- how long they are, which lines hold the
+    checks -- is verified by the fleet-audit suite, which can re-derive it. This
+    only asks the cheaper question that suite does not ask of every roster: does
+    the file exist at all.
+    """
+    findings = []
+    quoted = re.compile(r"['\"`]([a-z_]+/[A-Za-z0-9_.-]+\.(?:md|json|py|sh))['\"`]")
```
toshiowang: **Bug (latent).** `[A-Za-z0-9_.-]+` cannot span `/`, so only `dir/file.ext` matches. Anything deeper is silently unchecked.

Reproduced: a prompt citing `'agents/platform/governance/stockout_prevention_sop.md'` exits 0, and so does `'skills/fleet-audit/scripts/audit_report_missing.py'` — a file that does not exist. The `agents/...` form is the one the PR body calls "always wrong at runtime… it was live eighteen times": the SOP is now protected against it, the cron prompt that sends the worker to that SOP is not. No live instances today, so nothing is broken — but the coverage table reads as covering these.

Verified fix (no new findings on the current tree, both probes now fail, 40/40 tests pass):

```suggestion
    quoted = re.compile(r"['\"`]([A-Za-z0-9_.-]+(?:/[A-Za-z0-9_.-]+)+\.(?:md|json|py|sh))['\"`]")
```

toshiowang (commented):
Checked out `ed294f0` and ran it: the checker is clean (98 files, 42 bundles), all 40 unit tests pass in 0.14s, `make docs-check` is clean, `scripts/test_*.py` is already inside `PYTHON_TEST_DIRS` so the new suite really does run in `python-tests.yml`, and `validate.yml` carries no `paths:` filter, so the "deliberately unfiltered" claim holds.

Nothing here breaks anything that works today, and the fix to `gke-cluster-autoscaler/SKILL.md` is plainly right. The four findings are all the same shape: the cron half of the gate is weaker than the coverage table claims. Each was reproduced against this branch, and each suggested fix was applied and re-run (repo still clean, 40/40 tests still pass). The first two matter most — `cron-asset` bypasses the `PROFILE_HOME_ITEMS` model the PR is built around, and its regex cannot see the `agents/...` spelling that motivated the SOP rewrite in this very diff.

scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,745 @@
...
+            md
+            for md in sorted((home / "skills").rglob("*.md"))
+            if md.name != "README.md"
+        )
+    return found
+
+
+def cron_rosters() -> list[tuple[Path, list[dict]]]:
+    rosters = []
+    for rel in ("agents/platform/cron/jobs.json", "agents/chat/defaults/cron/jobs.json"):
```
toshiowang: **Bug (latent) — the scope can empty out silently, which is the failure the test module's own docstring says it guards against.**

`path.is_file()` skips a roster that has moved, and nothing floors the result: `test_it_is_actually_looking_at_something` floors instruction files and `test_every_skill_bundle_in_the_tree_is_discovered` diffs bundles against the tree, but no test asserts a roster was found, and the success line prints no cron count.

Reproduced: moving `agents/chat/defaults/cron/jobs.json` aside leaves the output byte-identical (`Prompt assets OK: 98 instruction files, 42 skill bundles…`, exit 0) with both cron rules now no-ops. The platform roster is caught only by accident, because `fleet-audit/SKILL.md` happens to cite `cron/jobs.json` in prose.

No drop-in fix. Either raise from `cron_rosters()` on an empty list, or add a repository test in the spirit of the bundle diff (`len(cpa.cron_rosters()) == 2`, job count non-zero), and put the roster/job counts in the success line so a narrowed scope is visible in the log.

scripts/check_prompt_assets.py
```diff
@@ -0,0 +1,745 @@
...
+# minimum, so ordinary prose ("this skill", "a skill") does not match.
+SKILL_IN_PROSE = re.compile(r"`?\b([a-z][a-z0-9]*(?:-[a-z0-9]+)+)`?\s+skill\b")
+
+FRONTMATTER = re.compile(r"\A---\n(.*?)\n---\n", re.DOTALL)
+FRONTMATTER_NAME = re.compile(r"^name:[ \t]*(\S+)[ \t]*$", re.MULTILINE)
+# `description:` may carry its value on the same line or as an indented block
+# beneath it -- prettier reflows the long ones into the second form, so a
+# same-line-only pattern reports the tidiest manifests as the broken ones.
+FRONTMATTER_DESCRIPTION = re.compile(
+    r"^description:[ \t]*(?:(\S.*)|\n(?:[ \t]+\S.*\n?)+)", re.MULTILINE
```
toshiowang: **Bug (minor).** The `(\S.*)` branch matches the YAML block indicator itself, so `description: >-` with the folded body deleted passes the rule that exists to require a description. Most bundles here use `>-` (including `gke-cluster-autoscaler/SKILL.md`), so that is the realistic slip; the agent then loads a skill with an empty description and never selects it.

Verified fix — all 42 real manifests still pass, while `>-` alone, `|` alone and bare `description:` all fail. `.group(1)` is never read, so dropping the capture group is safe:

```suggestion
    r"^description:[ \t]*(?:(?![>|][-+]?[ \t]*$)\S.*|[>|][-+]?[ \t]*\n(?:[ \t]+\S.*\n?)+|\n(?:[ \t]+\S.*\n?)+)", re.MULTILINE
```

scripts/check_prompt_assets.py
jayantid: Confirmed and fixed in d3182f7 — this is the sharpest of the four, because the rule's own message was the thing making the false claim.

Reproduced your probe exactly: `'docs/glossary.md'` appended to the `compliance-audit` prompt exited 0, the same string in `stockout_prevention_sop.md` exited 1. Same path, same profile, two different answers, and the cron one was the wrong answer.

Took your suggestion as written — `_resolves_in_home(ref, path)` — and dropped the now-dead `profile`/`home` locals with it. Passing the roster path in does the right thing on its own: `_profile_of` gives `platform` for `agents/platform/cron/jobs.json` and `chat` for `agents/chat/defaults/cron/jobs.json`, and `chat` is absent from `PROFILE_HOME_ITEMS` because it *is* the default profile, so the `/opt/defaults` layer still resolves there in full. That distinction was the whole point of the previous round and this rule was routing around it.

The `or _resolve_opt(ref) is not None` fallback went too — that was the same unconditional-layer assumption, one function over.

`test_a_cron_prompt_resolves_through_the_profile_home_model` pins it, and the message is now something the rule verifies.

scripts/check_prompt_assets.py
jayantid: Confirmed and fixed in d3182f7. Reproduced both probes before changing anything: `'agents/platform/governance/stockout_prevention_sop.md'` and `'skills/fleet-audit/scripts/audit_report_missing.py'` both matched nothing and both exited 0.

Your point about which spelling this leaves unguarded is the part that made me take it as a real gap rather than a tidy-up. The PR argues `agents/...` is the one form that survives human review and is always wrong at runtime, fixes eighteen instances of it in the SOP, and then leaves the cron prompt that sends the worker to that same SOP unable to see it. The coverage table said `cron-asset` covered quoted paths; it covered two-segment ones.

Took the suggestion, lifted to a module constant (`CRON_PROMPT_REF`) next to the other patterns with a comment recording what the old shape missed, since the reason is not recoverable from the regex.

Checked it does not over-fire: no new findings on the current tree, and `'skills/fleet-audit/scripts/audit_report.py'` — a deep path the profile genuinely has — still passes. Both directions are now tests (`test_a_cron_prompt_path_of_any_depth_is_checked`, `test_a_deep_cron_path_the_profile_does_have_is_accepted`).

Worth noting these two findings compose: with #1 routing through `_resolves_in_home`, the deep `agents/...` form is not just *seen* now, it is correctly *rejected*.

scripts/check_prompt_assets.py
jayantid: Confirmed and fixed in d3182f7. Reproduced: moving `agents/chat/defaults/cron/jobs.json` aside left the output byte-identical and exit 0, with both cron rules silently doing nothing.

This is the same failure the bundle discovery had earlier in this PR — a scope that empties without the run looking any different — so I did not want to fix it only with an assertion. Did all three things you suggested:

**Discovery is a glob**, `agents/*/cron/jobs.json` and `agents/*/defaults/cron/jobs.json`, so a roster that moves within `agents/` is still checked rather than merely reported missing. The `is_file()` guard is gone with the hardcoded list.

**A tree-derived test**, in the spirit of the bundle diff rather than the floor it replaced: `test_every_cron_roster_in_the_tree_is_discovered` diffs against `REPO.glob("agents/**/cron/jobs.json")` in both directions, asserts the tree has at least one, and asserts the job count is non-zero — a roster found but empty is the same no-op. I went with the diff over `len(...) == 2` because a hardcoded 2 needs editing the day a third profile gets a roster, which is the same maintenance trap that produced the original bug.

**Counts in the success line:**

```
Prompt assets OK: 98 instruction files, 42 skill bundles (cluster 6, platform 36),
12 cron jobs in 2 rosters, all references resolve.
```

Your framing — that the platform roster was covered "only by accident, because `fleet-audit/SKILL.md` happens to cite `cron/jobs.json` in prose" — is the part I would have missed. That accident is exactly the kind of thing that makes a gap look covered.

scripts/check_prompt_assets.py
jayantid: Confirmed and fixed in d3182f7. Reproduced: `description: >-` with the body deleted matched, so the rule that exists to require a description accepted an empty one.

Took the suggestion. Verified the full matrix rather than just the reported case — `>-`, `>`, `|`, `|-` and bare `description:` all fail; same-line text, a folded body under `>-`, and a plain indented block all pass; all 42 real manifests still pass.

Your point about `>-` being the *realistic* slip is what moved this above cosmetic for me. If the common form were same-line, an emptied description would be a visibly truncated line someone notices in review. In the folded form the indicator is left behind looking like intact YAML, and the failure is a skill the agent loads and never selects — silent, and the same shape as everything else this checker is for.

Two tests: `test_a_folded_description_with_no_body_is_reported` (subtests over the four indicators) and `test_a_folded_description_with_a_body_is_accepted`. Kept the existing plain-indented-block test, since that path is what the pattern was widened for originally and is easy to break while fixing this.

.github/workflows/validate.yml
```diff
@@ -10,6 +10,33 @@ permissions:
   contents: read
```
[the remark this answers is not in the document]
jayantid: Both parts land. Fixed in the description just now.

**The template.** You are right and the cause is drift on my side: this branch still carries the pre-restructure `.github/PULL_REQUEST_TEMPLATE.md`, and I wrote the body against that. `main` at `36c35a3` added `## Summary` and `## Risk & Rollout` and dropped `## Why This Matters`, so I had a heading that no longer exists and was missing two that do. Both are now written, not stubbed: **Summary** is one paragraph on what a reviewer needs before the diff — that the check is stdlib-only, runs as its own job, and that the substance is resolving against the image layout rather than the checkout; **Risk & Rollout** covers blast radius (no runtime code path — the checker never ships in an image, so at runtime the change is four prose corrections), the one new failure mode (a CI-side false positive, with why the heuristic errs toward missing a finding instead), revert (delete the job, or revert the PR — no state, no migration), and merge-time requirements (none). The former `Why This Matters` content is folded into `Why This Change`, which is where the new template says impact belongs.

**The test count.** The `27` was stale — left over from an early revision and never updated as the suite grew. That contradiction was real and it is gone; I deleted the number rather than reconciling it, so the suite size now appears in exactly one place, next to the run that produced it.

On the arithmetic: the breakdown does sum to 47 — 21 + 7 + 6 + 5 + 4 + 4. Re-derived from the loader rather than counted by hand:

```
$ python3 -m unittest test_check_prompt_assets
Ran 47 tests in 0.425s
OK
SyntheticProfileTests: 21   AnnotationTests: 7   CheckoutPathTests: 6
RepositoryTests: 5   ResolutionModelTests: 4   ProfileIsolationTests: 4
total: 47
```

Two more staleness bugs of the same kind that I caught while going through it, both now fixed: the body cited `test_the_console_links_doc_resolves_through_the_image`, a test I deleted three rounds ago because it pinned a model that turned out to be wrong; and it still described `/opt/defaults` as copied "over *every* profile" — the exact false claim this PR was corrected for, sitting two clauses before the sentence that contradicts it.

Also re-verified against current `main` while I was in there: merged `origin/main` at `36c35a3` into the branch in a scratch worktree, gate exits 0, 47/47 pass. That replaces the older `9ae064f` claim, which had gone stale too.

toshiowang (approved):

PR #504, opened 2026-08-02
feat(agent): add support for multiple github repos

kirill-kosov:
## Summary

This PR enables multi-repository management for the Platform Agent. Instead of restricting the agent environment to a single Git repository, the Go operator now reconciles a structured `<platform-agent-name>-gitops-state` ConfigMap containing all registered target repositories, as well as dynamic GitHub Token Minter policy ConfigMaps. Agent workspace scripts, background watchdogs, and skills resolve workspace state and token permissions against this operator-owned ConfigMap while keeping agent RBAC read-only (`get/list/watch`).

## Why This Change

Previously, Platform Agent deployments were constrained to a single GitOps repository configured at startup. Enterprise and multi-project deployments require agents to monitor and interact with multiple GitHub repositories across an organization (e.g. running fleet-wide security audits, resolving issues across multiple repos, and submitting suggestions). Dynamic repository registration must remain secure: allowing agents to write to their own state ConfigMaps creates privilege escalation risks, so repository registration is operator-reconciled from the `PlatformAgent` Custom Resource.

## What Changed

- **Operator & Custom Resource (`k8s-operator/`):**
  - Updated `PlatformAgent` CRD (`GitHubSpec`) to support `Org` and multi-repository parameters with strict input validation (`ValidateGitRepoURLWithOrg`, `CleanRepoSlugWithOrg`, `ValidateGitHubOrg`).
  - Added reconciliation logic in `PlatformAgent` controller to generate and update the `<name>-gitops-state` ConfigMap containing structured `ManagedRepoEntry` items (`url`, `type`).
  - Added operator reconciliation for the dynamic GitHub Token Minter policy ConfigMap (`<name>-token-minter-policy`).
  - Maintained read-only agent RBAC (`get/list/watch` on ConfigMaps) for secure isolation.

- **Platform Scripts & Automation (`agents/platform/scripts/`):**
  - `gitops_workspace.py`: Updated to parse repository lists from `<name>-gitops-state`, supporting cloning, updating, and workspace resolution across multiple repositories.
  - `github_token_refresh.py`: Updated to read managed repositories from `<name>-gitops-state` and request token minting for explicit same-org managed repositories.
  - `forge.py`, `github_scan_gate.py`, `resolver.py`, `pr_conversation.py`, `submit_suggestion.py`: Updated to validate target repositories against allowed managed repositories and handle partial errors gracefully during multi-repo scans.

- **Skills & SOPs (`agents/platform/`):**
[...]

PR #700, opened 2026-08-14
test(fleet-audit): resolve every section a cron prompt cites, not just the checks span

bradhoekstra:
## Summary

The Platform Agent's governance cron prompts cite their SOP by geography — how long the file is, and where in it the checks live — because a worker that stops reading early lands in the preamble and reports a clean fleet it never looked at. `test_cron_prompts_cite_the_real_sop_geography` already re-derives those numbers from the SOP on every run, which is why they are all correct today. It reaches them through a single pattern, `are section <n>, lines <a>-<b>`, so a citation written without a line range is invisible to it. There is one: the AI stream's prompt also says "Section 2 defines which workloads count as AI workloads." This makes the test resolve every section a prompt cites, not only the checks span.

## Why This Change

That citation is correct today and nothing holds it there. `ai_security_audit_sop.md` currently opens its AI-workload discriminator at `### 2. Collect workload state and identify the AI workloads`; insert a section ahead of it and the heading renumbers while the prompt does not. The existing guard passes, because the checks span it does police is re-measured independently and stays right.

The failure that leaks through is not cosmetic. That section is what decides which workloads the audit examines at all — a worker sent to the wrong one either audits the wrong set or finds no AI workloads and closes a clean ledger. This is the same class of silent-all-clear the line-number citations exist to prevent, arriving through the one door the test does not cover.

## What Changed

- `test_audit_report.py` — `test_cron_prompts_cite_the_real_sop_geography` now resolves every `Section <n>` a prompt cites against that SOP's real headings, reusing the fence-aware heading scan already in the test. Fenced-block headings stay excluded, so a `### ` inside a bash example is still not a section.
- `agents/platform/cron/README.md` — its "Hard-coded line numbers in prompts" section owns this contract and described the test as re-deriving the line numbers alone. Updated so the next person editing an SOP knows bare section references are checked too.

No prompt text and no SOP changed. `jobs.json` is untouched.

## Context

No related issue. No open PR is adding a section-citation guard; the seven open PRs touching these files (#504, #589, #591–595, #426) all conflict with `main` on their own already, except #426, which test-merges clean against this branch.

Worth recording for anyone who arrives here with the same idea I did: the obvious-looking change is to delete the line numbers from the prompts and point at heading names instead. That is wrong, and `agents/platform/cron/README.md` already says why — the numbers are load-bearing at runtime. I wrote that change first and it broke the existing test in all seven subtests. The gap was never the hardcoding; it was the one citation the guard could not see.

## Testing

- `python3 -m unittest discover -s agents/platform/skills/fleet-audit/scripts -p 'test_audit_report.py'` — 507 tests, OK.
- Negative test, prompt side: pointed the AI prompt at `Section 9`; fails with `the ai-security-audit prompt cites section 9 of ai_security_audit_sop.md, which has no one heading numbered that`. Reverted.
- Negative test, SOP side: inserted a section into `ai_security_audit_sop.md` ahead of the cited one so it renumbered; fails. Reverted. This is the case that motivated the change and the one the old guard passed.
- `make docs-check` — all four checks pass. `make docs-generate` leaves the tree clean.
- `prettier --check` with the CI-pinned 3.9.6 on the changed Markdown — clean.
- `review-docs-drift` run against the branch diff. It produced one finding, on this PR: the README section above under-described the test's scope. Fixed in this PR rather than reported and left.

### Live validation

Not live-tested. The change is a unit test plus the README paragraph documenting it; neither ships a runtime code path. `test_audit_report.py` sits in the skill directory that is staged into the image, but nothing in the pod executes it — the guard fires in CI, before a prompt can reach a worker. There is no layer on a running installation where this change would be observable, so exercising it against one would demonstrate nothing. No test artifacts were created; both negative tests were reverted immediately and the working tree is clean.

## Risk & Rollout

Low. No runtime code path, no prompt, and no SOP is touched — the only behavioural change is that CI now fails on a class of edit it previously accepted. The realistic failure mode is a false positive: a prompt that legitimately says "section" followed by a number meaning something other than an SOP heading. None exists today, and the assertion names the audit, the number and the file, so the diagnosis is in the failure message. Revert the commit to drop the check.

### DIFF
--- a/agents/platform/cron/README.md
+++ b/agents/platform/cron/README.md
@@ -96,7 +96,11 @@ they rot the moment an SOP is edited.
 `test_cron_prompts_cite_the_real_sop_geography` in
 `../skills/fleet-audit/scripts/test_audit_report.py` re-derives both from the
 SOP itself, so an edit that skips re-measuring fails there rather than at 06:20
-in production. Run it after touching anything in `../governance/`.
+in production. It resolves every other section a prompt cites as well — the AI
+stream names the one that defines which workloads count as AI workloads, and
+cites it bare, with no line range to re-measure — so a section inserted ahead of
+that one renumbers it into a test failure rather than into a worker sent to the
+wrong section. Run it after touching anything in `../governance/`.
 
 No prompt is quoted here on purpose. A copy in prose is one more place for the
 same numbers to go stale, and the test above checks the roster against the SOPs
--- a/agents/platform/skills/fleet-audit/scripts/test_audit_report.py
+++ b/agents/platform/skills/fleet-audit/scripts/test_audit_report.py
@@ -2100,6 +2100,29 @@ class TestAuditCatalogue(unittest.TestCase):
                     f"has {len(spec.checks)} checks",
                 )
 
+                # Any other section the prompt cites has to resolve as well.
+                # The span check above covers the checks section only, and it
+                # is not the sole citation: the AI stream also points at the
+                # section defining which workloads count as AI workloads, so a
+                # section inserted ahead of that one renumbers it with nothing
+                # to notice. Sections are cited bare there, without a line
+                # range, which is why the `span` pattern does not reach it.
+                for cited in sorted(
+                    {int(n) for n in re.findall(r"[Ss]ection (\d+)\b", prompt)}
+                ):
+                    self.assertEqual(
+                        1,
+                        len(
+                            [
+                                n
+                                for n in starts
+                                if lines[n - 1].startswith(f"### {cited}. ")
+                            ]
+                        ),
+                        f"the {audit_id} prompt cites section {cited} of "
+                        f"{name}, which has no one heading numbered that",
+                    )
+
     def test_every_sop_states_the_rules_that_hold_on_every_stream(self):
         """A fix written into one SOP has to reach all the others.
 


agents/platform/skills/fleet-audit/scripts/test_audit_report.py
```diff
@@ -2100,6 +2100,29 @@ def test_cron_prompts_cite_the_real_sop_geography(self):
...
+                        len(
+                            [
+                                n
+                                for n in starts
+                                if lines[n - 1].startswith(f"### {cited}. ")
+                            ]
+                        ),
+                        f"the {audit_id} prompt cites section {cited} of "
+                        f"{name}, which has no one heading numbered that",
+                    )
```
toshiowang, line 2124: **[CRITICAL BUG]** This asserts only that *some* heading carries the cited number, not that it is still the section the prompt describes. In the renumbering scenario the PR is built around, a heading numbered 2 always exists after the insert, so this never fires.

The failure the PR's SOP-side negative test saw came from the pre-existing line-count and span assertions — and those are exactly what the README sends the editor off to re-measure. Once they do, the bare citation drifts silently and the guard is green.

Reproduced on 54cd8a3: inserted `### 2. Placeholder inserted section` ahead of the cited section in `ai_security_audit_sop.md`, renumbered the headings below it, and updated the prompt's `all 342 lines of it` → `346` and `section 3, lines 148-272` → `section 4, lines 152-276`. The SOP now reads:

```
### 2. Placeholder inserted section
### 3. Collect workload state and identify the AI workloads   ← the real discriminator
```

"Section 2 defines which workloads count as AI workloads" now points at the placeholder, and `test_cron_prompts_cite_the_real_sop_geography` passes. Running `main`'s copy of the test file against the identical drift also passes, so for the motivating case this block changes nothing. It does catch a cited number with no heading at all — the prompt-side `Section 9` negative test — which is a narrower win than the description claims.

Not a drop-in edit to these lines. The check needs identity, not existence: pin each bare citation to the heading it claims, e.g. a module-level `SECTION_CLAIMS = {("ai-security-audit", 2): "identify the AI workloads"}` and, once the heading is resolved, `self.assertIn(SECTION_CLAIMS[(audit_id, cited)], lines[at[0] - 1], ...)`. Letting an unregistered citation raise `KeyError` is the right default — a new bare reference should have to declare what it points at.

agents/platform/cron/README.md
```diff
@@ -96,7 +96,11 @@ they rot the moment an SOP is edited.
 `test_cron_prompts_cite_the_real_sop_geography` in
 `../skills/fleet-audit/scripts/test_audit_report.py` re-derives both from the
 SOP itself, so an edit that skips re-measuring fails there rather than at 06:20
-in production. Run it after touching anything in `../governance/`.
+in production. It resolves every other section a prompt cites as well — the AI
+stream names the one that defines which workloads count as AI workloads, and
+cites it bare, with no line range to re-measure — so a section inserted ahead of
+that one renumbers it into a test failure rather than into a worker sent to the
+wrong section. Run it after touching anything in `../governance/`.
```
toshiowang, line 103: **[CRITICAL BUG]** "a section inserted ahead of that one renumbers it into a test failure rather than into a worker sent to the wrong section" is precisely the case that passes — see the reproduction in the comment on `test_audit_report.py`.

This paragraph is what the next person editing `../governance/` reads before deciding how much care the bare citation needs, so the overstatement costs more than the gap in the test itself: it turns an unguarded citation into one believed guarded. Either strengthen the check to verify identity and keep this wording, or scope the wording to what the check does.

```suggestion
in production. It also resolves every other section a prompt cites — the AI
stream names the one defining which workloads count as AI workloads, and cites
it bare — but only checks that such a section exists, not that it is still the
right one. Renumber an SOP and re-measure the spans and that citation drifts
silently. Run it after touching anything in `../governance/`.
```

toshiowang (changes requested):
The gap this PR identifies is real — the AI stream's bare `Section 2` citation is genuinely outside the `span` pattern's reach, and the fence-aware heading reuse is the right mechanism. But the check added is existence-only, and I could not get it to fire on the scenario the PR was written for.

Reproduced on 54cd8a3 in a scratch worktree: inserted `### 2. Placeholder inserted section` into `ai_security_audit_sop.md` ahead of the cited section, renumbered the headings below it, then updated the prompt's `all 342 lines of it` → `346` and `section 3, lines 148-272` → `section 4, lines 152-276` — the re-measure the README instructs an editor to perform. The SOP then reads `### 2. Placeholder inserted section` / `### 3. Collect workload state and identify the AI workloads`, the prompt still says "Section 2 defines which workloads count as AI workloads," and `test_cron_prompts_cite_the_real_sop_geography` passes. `main`'s copy of the test file passes on the identical drift, so for the motivating case the new block changes nothing; the failure the PR's SOP-side negative test observed came from the line-count and span assertions that already existed. What the new check does add is coverage for a cited number with no heading at all — the prompt-side `Section 9` case.

The fix is to check identity rather than existence, and the README paragraph needs to stop promising the stronger guarantee. Details in the two inline comments.

PR #713, opened 2026-08-14
docs(agents): give every user-facing message one output standard

jayantid:
## Summary

The most consistent piece of feedback this product gets is that its output is too verbose and reads like a moving target. The rules meant to prevent that already existed, but they were scattered across three personas and all pointed at one channel — every statement in `agents/**` about length or formatting is written about `kanban_complete(result=)`. This adds `agents/platform/docs/output-style.md` as the single home for how anything the agents send a human is written, bakes it into the image beside the glossary and the console links, and has the three personas cite it instead of restating it.

## Why This Change

Measuring 42 days of the live session store showed the guidance was aimed at the wrong place. `send_notification` — the highest-volume user-facing channel — sent 103 messages/day at a ~1KB median with no style guidance anywhere under `agents/**`; the string does not appear in a single persona or SOP. 99% of those messages wrote Slack's own `*mrkdwn*` instead of standard Markdown, none used `##` headings at all, and the notify and chat channels carried 19 and 78 distinct opening-header shapes between them. That last number is the "moving target" complaint stated numerically: it is a complaint about shape, not length.

Reading the send path shows why one standard can cover all three channels. `send_notification` shells out to `hermes send` (`platform_mcp_server.py`), which lands in the same Slack adapter `send()` that posts a card result and applies Block Kit to both. The formatting rules were never channel-specific — they were only written down as though they were, which is how one channel ended up governed three times over and another not at all.

Without this, each new rule lands in whichever persona the author was editing and the three drift apart again — which is the state the diff removes.

## What Changed

- **`agents/platform/docs/output-style.md` (new)** — the standard: length ceiling, the Block Kit limits, standard Markdown over Slack mrkdwn, one fixed shape per kind of message, a cut-list, and a pre-send check. It states the length ceiling against the mechanisms that enforce it rather than as taste: Google Chat splits at 4000 characters, and Slack's Block Kit renderer returns nothing past 50 blocks — dropping the *entire* message to flat text rather than truncating the tail.
- **`deploy/docker/Dockerfile`** and **`scripts/check_prompt_assets.py`** — bakes it to `/opt/defaults/docs/` and registers it in `OPT_DEFAULTS`. `test_opt_defaults_matches_the_dockerfile` re-derives that table from the Dockerfile, so the two cannot drift apart silently.
- **The three `SOUL.md` files and two `AGENTS.md` files** — cite the standard at its absolute path, which is the spelling that resolves for a specialist profile. The duplicated Block Kit paragraphs in the platform and cluster personas are replaced by the citation; the incident-grounded rationale that makes the rule stick (card `t_3ba2166a`, three rendered blocks where Markdown would have given nine) stays where it was.
- **`send_notification`'s docstring** — carries a short form of the standard. That tool description is injected into every worker's tool definitions, so it is the only lever that reaches the previously ungoverned channel directly.

## Context

No open PR touches these files. Follow-up work, deliberately not in scope here:

- `scripts/output_baseline.py`, the read-only script the numbers above come from — it measures volume, length, duplication and consistency per channel, so a rule change can be checked against a number instead of an anecdote. Kept out to keep this diff to the standard itself; it will follow separately.
- Enforcement — a prompt lint, a `review-output-ux` skill, and a response-length assertion in `bench` — which should encode this document rather than restate it, and is worth landing after the wording settles.

## Testing

- `make prompt-check` — passes; 100 instruction files, all references resolve. This is the gate that catches a persona citing a path the agent cannot open, so it is the one that matters for the citations added here.
- `make docs-check` — passes; links resolve, terminology matches, map inventory current with no re-aligned rows.
- `scripts/test_check_prompt_assets.py` — 48/48, including `test_opt_defaults_matches_the_dockerfile`.
- `agents/platform/scripts/test_platform_mcp_server.py` — the 6 `send_notification` tests pass. One unrelated test errors locally on a missing `yaml` module (`test_config_yaml_passes_the_key_into_this_subprocess`); it does not touch this change.
- `prettier --check` — **not run**: neither `prettier` nor `npx` is available in this environment. The one table this adds was aligned to prettier's layout by hand and verified programmatically (all rows equal width); the `docs/README.md` row uses the compact single-space form the map requires, and no existing row was re-aligned.

### Live validation

Live A/B against `gke_jayantid-gkedemos_us-central1_platform-agent-host` (namespace `kubeagents-system`, operator `k8s-operator:dev`). Two `platform-agent` images built from clean worktrees differing only in this branch: `:mainbase` (current `main`) and `:pr713v2`. Baseline is a fresh build of `main` rather than the standing `:dev` tag, because `:dev` predates #717 and would have credited that PR's gains to this one.

**The mechanism, proven in both directions.** The image has to be driven through the CR — a direct `kubectl set image` on the Deployment is reconciled away by the operator, which silently invalidated a first run and is why the runner now patches `spec.deployment.image` and aborts unless the Deployment reports the tag it asked for. Verified inside the running pod before any task ran, per arm:

| | `:mainbase` | `:pr713v2` |
| --- | --- | --- |
| `/opt/defaults/docs/output-style.md` | absent | present |
| `grep -c output-style` — image `SOUL.md` / scaffolded profile `SOUL.md` / MCP server | 0 / 0 / 0 | 1 / 1 / 1 |

The baseline arm also deletes the PVC's stale `/opt/data/docs/output-style.md`, since the entrypoint's `cp -ru` never removes files. Absence and presence each produce the expected state, so the delta is the change rather than a coincidence.

**Workload.** 4 tasks × 2 repetitions × 2 arms on the kanban board — two ending in `send_notification`, one long report, one persona explanation. 8 `result` messages and 4 notifications captured per arm, `result` read from the card JSON and notifications extracted read-only (`mode=ro`) from the worker session store.

**Results.**

| Metric | Channel | `:mainbase` | `:pr713v2` |
| --- | --- | --- | --- |
| median chars | notification | 465 | **275** |
| p90 chars | notification | 891 | **723** |
| median chars | `result` | 1243 | **1128** |
| p90 / max chars | `result` | 4116 | **4569** |
| over the 4000 ceiling | `result` | 1 / 8 | 1 / 8 |
| leading preamble (§5) | `result` | 38% | **12%** |
| more than one glyph (§4) | `result` / notification | 12% / 25% | **0% / 0%** |
| nested bullets, which Chat flattens (§1) | `result` / notification | 50% / 50% | **38% / 25%** |
| `[text](url)` links (§3) | both | 0% | 0% |

The clearest single case is a card that asked for a notification. Baseline sent it and then pasted the whole roster into `result` as well, so the same list arrived twice in one thread (824 chars); the treatment returned a 149-character receipt and let the notification carry the content. §3's heading rule also took: `### How we differ` became `## How it differs from the Platform Agent`, which is the difference between an H3 and a real Block Kit header.

**Where it did not work, which is the part worth reading.** §2 is the document's headline rule and it did not take. On the long report the treatment produced **4569 characters — past the document's own 4000 hard ceiling, and 453 longer than baseline** — and both arms opened with the preamble §5 bans. On this install that overflow is real rather than theoretical: `platforms.google_chat.enabled: true` and `platforms.slack.enabled: false`, so Google Chat's 4000-character split is the live mechanism and the user receives that report as two messages, the second stripped of its context. Prose stating a ceiling did not enforce a ceiling. That is an argument for the enforcement work listed under Context, not against the document, but it means **this PR should not be read as buying a length reduction on long output.**

**What this run does not cover.**

- **n = 8 results and 4 notifications per arm.** Nothing here is statistically significant; the preamble and glyph figures are one to three messages moving.
- **`CREDENTIAL_PROXY_URL` is unset in the pod**, so `gcloud` and `kubectl` fail inside the agent and every task had to work from local data. No incident-triage message was exercised — which is exactly the shape §4 and the linking rule are written for.
- **The linking and table rules are untested, not validated.** The 0% link rate in *both* arms is a task set with nothing linkable in it, and neither arm emitted a table. Two of the rules folded in from #717 got no exercise at all.
- Slack was not exercised, being disabled here. The Block Kit limits the document cites remain read from the adapter source rather than observed.

**Corrections found while doing this.** My first scorer counted `*mrkdwn*` violations with a regex that also matches `*` bullet markers and CommonMark italics; it reported a doubling that was not real, and there were no true mrkdwn violations in either arm. Re-running the same corrected check over the 2365 production notifications behind the Why This Change numbers leaves the 99% figure standing (2330 strict matches, samples like `*Cluster:*` and `*Issue:*`), because only 2% of production notifications use `*` bullets.

**Cleanup.** All 28 test cards archived and the board is empty; the `:mainbase`, `:pr713v2` and `:pr713` image tags are deleted from Artifact Registry; both scratch worktrees removed. The CR is patched back to `platform-agent:dev` and the Deployment reports that image. All database access was read-only. Nothing from this run is left behind; one pre-existing `stash@{0}` holding an unrelated `testing-strategy` revision was deliberately not touched, as it is not mine.

## Risk & Rollout

Low. No runtime code path changes: the only non-Markdown edits are a docstring, a Dockerfile `COPY`, and a lint table entry. The worst case is that an agent reads the standard and writes shorter, more uniform messages — or ignores it, which is the status quo. Reverting is a single revert commit; nothing persists outside the image.

The one thing to watch at merge time is that the document is baked, so it takes effect on the next image build rather than at merge, and the effect should be confirmed by re-running the baseline against the same channels afterwards.

---


jayantid:
Closing as stale. #872 landed the inline-persona mechanism with A/B evidence against the cited-doc approach this PR proposed, and #1088 owns the report-stanza budget. The contributor-facing UX writing rule will come as a separate PR that does not touch the personas.

PR #717, opened 2026-08-14
fix(agents): make triage output scannable, linked and platform-aware

bradhoekstra:
# Pull Request

## Summary

The Platform Agent's incident-triage output was long, unlinked and written as
though Slack were the only place it lands. This rewrites the persona rules that
produce it — §7 of `agents/platform/SOUL.md` — into seven scannable rules with a
hard three-section cap and a length budget, and documents what Google Chat
actually does to Markdown so the agent writes for the narrower of the two
platforms. It then fixes the two other places in the repo that were quietly
instructing the agent to do the opposite: the report-format stanza injected into
every kanban card body, which mandated Slack-only pipe tables, and the
event-watcher triage template in `session_kv_server.py`, which hands the agent a
verbatim report shape and says "formatted exactly like this". Persona and prompt
text only — no Go, CRD or operator changes, and no change to what the agent is
permitted to do.

## Why This Change

A team running the Platform Agent against Google Chat reported its replies were
"lacking hyperlinks and direction and very wordy instead of having clean bullet
point or main point style response."

Two things in the repo caused that, and neither is specific to that deployment:

1. **`SOUL.md` §7 mandated prose.** The incident-triage playbook required a
   "Plain-Language Engineering Companion Persona" and a strict three-header
   `Issue / Root Cause / Recommendation` layout. Together those turn findings
   that fit on one line into multi-paragraph reports.
2. **All the rendering guidance assumed Slack.** `SOUL.md` §0, `cluster/SOUL.md`
   §6, both `AGENTS.md` summaries, and the report-format stanza injected into
   every card body describe only Block Kit. Google Chat is a supported platform
   and renders almost none of it — headings collapse to bold, tables and
   dividers are dropped, nesting is flattened, and anything past 4000 characters
   is split mid-sentence across messages.

The hyperlink complaint was the surprising one: the Google Chat adapter *does*
convert `[text](url)` to `<url|text>`, so links have always worked. Nothing in
the personas ever told the agent to emit them.

## What Changed

**Files:**

- `agents/platform/SOUL.md`
- `agents/platform/AGENTS.md`
- `agents/cluster/SOUL.md`
- `agents/cluster/AGENTS.md`
- `deploy/docker/patches/kanban_report_format.py`
- `deploy/docker/patches/apply_slack_code_emphasis.py`
- `agents/platform/scripts/session_kv_server.py`
- `agents/platform/scripts/test_session_kv_server.py`
- `agents/platform/docs/autoops-architecture.md`

**Added/Changed/Fixed:**

- **`SOUL.md` §7 rewritten.** Lead with the answer on the first line; one finding
  per bullet in the shape `**<target>** — what is wrong (evidence) → what to do`;
  an explicit list of things to cut (throat-clearing, restatement, closing
  summary, offers to help further).
- **Exactly three sections, never a fourth.** `What's wrong` / `Why` /
  `What to do`, with explicit homes for the content that used to sprout extra
  headings: current state is evidence, so it is a bullet under **Why**; what you
  could not check is a caveat on the action it blocks, so it rides that bullet
  under **What to do**.
- **A length budget.** Aim under 2,000 characters, treat 4,000 as the wall.
  Overflow goes to a published, linked document — never to a deleted finding.
- **Jargon translations kept, compressed.** `CrashLoopBackOff` and friends still
  get a plain-language gloss, but as an inline parenthetical rather than a
  sentence of its own. The raw code stays, because that is what people search for.
- **Per-platform rendering documented.** §0 now describes Slack and Google Chat
  separately and instructs the agent to write for the narrower of the two. A
  table is called out as a Slack-only luxury that may never be the sole home of a
  fact.
- **A link rule.** Every named artifact — cluster, workload, card, PR, issue,
  console view — ships as `[text](url)`, built from the §5 GCP Console templates.
- **The injected report-format stanza no longer mandates Slack-only tables.** It
  said "put tabular data in a Markdown pipe table", full stop, which directly
  contradicts the new persona rule. The pipe-table guidance stays (the defect
  detector pairs with it) but is now bounded, and the stanza names both
  destinations.
- **The event-watcher triage template no longer contradicts §7.**
  `_build_agent_query` tells the agent to post its report "formatted exactly like
  this" and then supplied the old shape — an `Incident Triage` banner,
  `Issue:`/`Root Cause:` bullets, a separate `Proposed Fixes (GitOps):` block and
  a trailing call to action. Being in the turn prompt, it outranks the persona,
  so the highest-volume triage path would have kept the old shape while §7 forbade
  it. Recast as the three permitted sections; the Option list, the Recommended
  line and the approval words move inside **What to do**, where they belong as
  actions. Nothing the reader needs is dropped — in particular the `apply`
  wording, without which a recommended GitOps fix has no stated way to be
  authorized. The docstring and `autoops-architecture.md` now say that this
  template outranks the persona, so the next domain's template starts from §7
  instead of inventing a shape that silently overrides it.
- The `apply_slack_code_emphasis` docstring quotes a §0 sentence this PR
  reworded; the quote is refreshed to match.

## Why This Matters

- Google Chat is a first-class platform here, but every piece of output guidance
  in the repo was written as though Slack were the only target. Anything
  depending on tables or header blocks silently degrades to prose.
- The old §7 was the single largest driver of long agent replies, and it applied
  to exactly the messages that most need to be scannable — incident alerts.
- Links cost nothing and already worked on both platforms. Not asking for them
  was leaving usability on the floor.

## Context

No open issue covers this. Two open PRs touch these files without overlapping in
subject — merge-conflict warnings, not duplicates:

- #504 (`feature(agent): add support for multiple github repos`) touches
  `agents/platform/SOUL.md`.
- #489 (`Design Proposal: Agent Autonomy & Tool Execution Boundaries`) touches
  `agents/platform/SOUL.md`, `agents/platform/AGENTS.md`, and
  `agents/cluster/AGENTS.md`.

## Testing

- `prettier@3.9.6 --check` clean on all four Markdown files.
- `make docs-check` passes (generated regions, relative links, terminology,
  documentation-map coverage).
- `python3 -m unittest test_audit_report` — 507 tests, OK. One asserts `SOUL.md`
  §0 still mentions URLs.
- `make test-python` passes (73 tests), including
  `agents/platform/scripts/test_session_kv_server.py` — 45 tests, 24 subtests.
  Two new cases pin the template to §7's three sections and assert the approval
  words survive inside **What to do**; a third was updated because it located the
  call-to-action by its `👉` prefix, which the new shape does not use.
- All 21 `deploy/docker/patches/test_*.py` suites pass, including
  `test_kanban_report_format` (29 tests), which asserts the stanza still
  mentions `pipe table` and that the stanza does not violate its own rules, and
  `test_kanban_result_required` (38 tests), which covers the rewritten `result`
  schema description.

### Live validation

Exercised on a private GKE Autopilot cluster in `europe-west1`, operator built
from `bd0fd11`. The `platform-agent`
image was rebuilt from this branch and pinned by digest on the `PlatformAgent`
CR's `spec.deployment.image`.

**Mechanism, not coincidence.** `agents/platform/scripts/profile_scaffold.py`
re-overlays `SOUL.md` from the baked image at every pod start, so an in-container
edit proves nothing — only a rebuilt image does. Observed inside the pod at each
layer:

- Deployment / pod: `platform-agent-gateway-779b5c4779-ml9r5`, 4/4 Running, 0
  restarts; container `imageID` matches the pinned digest.
- CR `.status.conditions`: `Ready=True`.
- File in the pod, `/opt/data/profiles/platform/SOUL.md`: 32,774 → 36,466 bytes.
  `Mandatory 3-Part Layout` went from present to absent; `Bullets carry the
  findings`, `Exactly three sections. Never a fourth`, `Budget: one message`,
  `Slack-only luxury`, and `Everything you name, you link` went from absent to
  present; `Google Chat` from 0 to 4 occurrences.

**Behavioural A/B.** The same triage card body was dispatched three times — once
on the pre-change image, once on a first cut of this branch, once on the final
commit — and the resulting `result` payloads measured:

```
                            chars  msgs  links  sects  tables  bullets  paras
BEFORE  (upstream persona)   6233     2      0      4       0        7     14
v1      (first edit)         5833     2      3      5       2       10      5
v2      (this branch)        3153     1      3      2       0        9      2
```

`msgs` is how many Google Chat messages the payload becomes at the adapter's
4000-character cap. The final output is 49% shorter than the baseline, arrives as
one message instead of being split mid-sentence, and carries console links where
the baseline had none.

The v1 row is why the last three commits exist: with the new persona but the
unbounded "three parts" wording and the old report-format stanza, the agent still
emitted five sections and two pipe tables. Tightening the section cap, adding the
length budget, and fixing the stanza removed both.

**Not covered.** One prompt, three runs, on one cluster. Diagnostic quality also
appeared to improve across the runs, but these are independent non-deterministic
generations and I am not attributing that to this change — what is demonstrated
here is shape: length, links, section count, tables. Slack rendering was not
re-tested, since this PR only widens the Slack guidance rather than changing it.

**Cleanup.** All three test kanban cards (`t_7f13760d`, `t_88f7806c`,
`t_a58c87f9`) were archived. The cluster remains pinned to the branch build for
continued observation; that pin is local to the installation and not part of this
PR.

#### Event-watcher template (added after review)

The template change was validated separately, on the same cluster rebased onto
`44ba6e9`. The `platform-agent` image was rebuilt from commit `c98b5b1` and
repinned by digest; pod `platform-agent-gateway-746754fcb6-5h77t`, 4/4 Running, 0
restarts, container `imageID` matching the new digest, CR `EventWatcher=False`
(this installation keeps the watcher paused).

`_build_agent_query` was then called inside the running pod against a synthetic
`OOMKilled` payload, and the prompt it returns was read back in full. It now
carries exactly three `##` headings — `What's wrong`, `Why`, `What to do` — with
no `📋 Incident Triage` banner and no `🛠️ Proposed Fixes (GitOps):` block. The
`✅ Recommended: Option <letter>` line and the `apply` / `apply Option A` /
`apply Option B` wording appear as bullets under **What to do**, and the console
links resolve against the installation's real project rather than the empty
`?project=` a missing value would produce. The old and new shapes are distinctly
different, so this is the mechanism rather than a coincidental match with a
default.

**Not covered for this part.** I verified the prompt the agent receives, not a
full agent turn through the watcher. Driving one end to end posts an alert into a
real Google Chat space — `_post_initial_alert` fires before the agent is invoked
— and this installation deliberately runs with the event watcher paused, so I did
not trigger one. What is unverified is therefore the model's compliance with the
new template, not the template itself.

#### Review round (added after the bot review)

This round changes baked-image content — the report-format stanza, the `result`
schema description, and the watcher template — so it was proved by a rebuild
rather than an in-container edit. Image built from `fa9e2f2` on base
`nousresearch/hermes-agent:v2026.8.3@sha256:1678831…` (the digest `tags.env`
pins), pushed as `platform-agent@sha256:05bb20d9…`, and pinned on the
`PlatformAgent` CR's `spec.deployment.image` on the `bhoekstra-gkedemos` /
`kube-agents-cluster` install (`kubeagents-system`).

- CR `.status.conditions`: `Ready=True (Reconciled)`.
- Pod `platform-agent-gateway-5f8c554b44-p5dmh`, 3/3 Running, 0 restarts,
  `platform-agent` container `imageID` matching `sha256:05bb20d9…`.

The same four reads were taken on the pre-change pod
(`568f95dcf6-ctjn5`, `sha256:1d4f1815…`) before the roll and repeated on the new
one after it:

| Read (inside the pod) | Before | After |
| --- | --- | --- |
| `tools.kanban_report_format.REPORT_FORMAT_STANZA` | 635 chars | 1,558 chars |
| `tools.kanban_result_required.NEW_RESULT_DESCRIPTION` | 671 chars, ending `…a pipe table for tabular data, and backticks around raw values` | 834 chars, ending `…never let it be the only place a fact lives, because Google Chat drops tables` |
| `/opt/data/profiles/platform/SOUL.md` | 32,774 bytes; `truncated, full output saved to` ×0 | 37,280 bytes; ×1 |
| `_build_agent_query` call-to-action | `👉 **Reply 'apply' …**` | `- **To authorize:** reply **'apply'** …` |

Also read back from the running pod: all six of the stanza's new phrases are
present and the old `##`/`###` wording is gone from both patch modules; the three
rewritten `DEFECT_ADVICE` entries print exactly as written in the source.
`_build_agent_query`, rendered in the pod against a synthetic `OOMKilled`
payload, returns exactly one `- **To authorize:**` bullet — carrying no
`<letter>` and no Option letter of its own — with no `👉` anywhere in the prompt,
the "not another option" instruction present, and three `##` headings.

**Mechanism, not coincidence.** Every after-value differs from its before-value,
and none of them equals a default. `profile_scaffold.py` re-overlays `SOUL.md`
from the baked image at pod start, so 32,774 → 37,280 bytes can only have come
from the rebuild. One thing worth recording for the next person: the
`kanban_result_required` patch does not write its description into
`kanban_tools.py`. The Dockerfile injects
`_apply_result_schema(KANBAN_COMPLETE_SCHEMA)` and the string is applied at
import, so grepping `kanban_tools.py` for the wording finds nothing — the check
above imports `tools.kanban_result_required` instead.

**Not covered.** No agent turn was driven end to end, so what remains unverified
is the model's compliance with the new stanza and the new template — the same gap
this section already declares for the watcher path. This round changed
instruction text, and what is shown is that the text reaching the agent is the
new text.

**Observed but unrelated.** The gateway logs a repeating
`google-chat-relay-patch: Google Chat relay receive failed … HTTP Error 503` on
both images — 2 occurrences in the last 200 lines on the restored pre-change pod
as well. `google_chat_relay_patch.py` is not in this branch's diff; it is the
installation's chat relay, and it does not affect readiness.

**Cleanup.** `spec.deployment.image` was restored to the bare
`…/kube-agents/platform-agent` it held before — the `spec.deployment.tag:
dns-endpoint-7dac349` alongside it is what resolves the running image — and the
install rolled back onto `sha256:1d4f1815…`, ReplicaSet `568f95dcf6`,
`Ready=True (Reconciled)`, 3/3 Running. No kanban cards were created this round.
The `pr717-fa9e2f2` tag is left in the `bhoekstra-gkedemos` registry alongside
the other branch builds.

## Risk & Rollout

Low. Persona and prompt text only — no operator, CRD, or Go changes, no new
permissions, and no change to what the agent may do. The behavioural blast radius
is the wording of agent-authored chat messages and kanban `result` payloads.

The one thing worth a reviewer's eye is the deletion in
`session_kv_server.py`: the approval words are the only content in that template
a reader cannot recover if they go missing, so `test_approval_line_survives_inside_what_to_do`
pins them. Rollout is an image rebuild, since `profile_scaffold.py` re-overlays
the personas from the baked image at pod start. Reverting is a redeploy of the
prior digest; nothing here needs to happen at merge time and there is no
migration or state change.


agents/cluster/AGENTS.md
```diff
@@ -14,7 +14,7 @@ A glossary of agentic terms lives at `/opt/defaults/docs/glossary.md`. Read it *
 - **One cluster only.** Never query or reason about other clusters or the fleet.
```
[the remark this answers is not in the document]
bradhoekstra: Fixed — thank you, both sections were genuinely missing rather than deliberately skipped. **Summary** and **Risk & Rollout** are now filled in, and the description also covers the third file group (`session_kv_server.py` and friends) added in `c98b5b1`.

One note on the substance: the **Live validation** section was already answered, so I extended it rather than restating it. There is now a second subsection covering the event-watcher template, including what it does *not* cover — I verified the prompt the agent receives, not a full agent turn, because driving one end to end posts an alert into a real Google Chat space.

bradhoekstra:
## 🟠 Medium — rewritten §7 contradicts the event-watcher template

Confirmed and fixed in `c98b5b1`. This was a real defect and the reasoning was right, including the part I would have argued with: the template being in the turn prompt means it *wins*, so the effect of this PR without that fix was to leave the highest-volume triage path on the old shape while the persona forbade it.

It is also the third instruction channel governing the same output. `REPORT_FORMAT_STANZA` was the second, and I already know empirically what happens when one of them disagrees with the persona — the A/B run in the description (`v1`) emitted two pipe tables purely because the stanza said to, in direct contradiction of the new rule. So the failure mode is not hypothetical here.

**What changed.** `_build_agent_query` now emits the three sections §7 permits. The `📋 Incident Triage` banner and the `🛠️ Proposed Fixes (GitOps):` block are gone; the Option list, the `✅ Recommended: Option <letter>` line and the approval words all move inside **What to do**, which is where they belong — they are actions. The `apply` wording specifically is the one thing a reader cannot recover if it goes missing, which is the failure you named, so `test_approval_line_survives_inside_what_to_do` now pins it and `test_template_uses_only_the_three_permitted_sections` pins the shape. One existing test needed updating: it located the call-to-action by its `👉` prefix, which the new shape does not use.

**The durable half.** The docstring and `agents/platform/docs/autoops-architecture.md` now state that this template outranks the persona, so the next domain to add one starts from §7 rather than inventing a shape that silently overrides it. `autoops-architecture.md` had been describing the report shape as an independent design decision.

**Live validation** is in the description under *Event-watcher template (added after review)*. Short version: rebuilt image, repinned by digest, `_build_agent_query` called inside the running pod and the returned prompt read back in full. What is **not** covered is a full agent turn through the watcher — `_post_initial_alert` posts into a real Google Chat space before the agent is invoked, and that installation deliberately runs with the watcher paused. So the template is verified; the model's compliance with it is not.

---

The description finding is fixed too — see the thread on it.

deploy/docker/patches/kanban_report_format.py
```diff
@@ -109,12 +109,16 @@
 ## Report format
 
 Put the full answer in `result` as standard Markdown — the gateway posts it
-verbatim into the requester's chat thread, where Slack renders it as blocks:
+verbatim into the requester's chat thread, where Slack renders it as blocks
+and Google Chat flattens headings to bold, drops tables, and splits anything
+past 4000 characters across messages:
 
 - Open with one sentence saying what happened, then give the detail.
 - Use `##` and `###` for sections. Never `#` — the chat message already shows
```
jayantid: **Blocker — this stanza still says the opposite of §7.**

Line 116 says "open with one sentence saying what happened", and this line permits `###` sections. No cap, no length budget, no link rule. The stanza is appended to every card body, so by your own precedence argument it beats the persona — and triage mostly arrives as a card. That's the v1 row again.

```
- Lead with the answer: what is wrong and what you want done. Then detail.
- Use `##` only. Never `#`, never `###`. For triage, only §7's three sections.
- Aim under 2,000 characters; 4,000 is the wall. Publish and link long form.
- Link every artifact you name as `[text](url)`.
```

agents/platform/SOUL.md
```diff
@@ -170,19 +177,22 @@ Retain fleet-level and provisioning-backend diagnostics yourself (Config Connect
...
+6. **Budget: one message.** Google Chat hard-splits at 4000 characters, mid-sentence, so a synthesis that runs past it does not arrive as one thought (§0). Aim for **under 2,000 characters** and treat 4,000 as the wall. This is a real constraint, not a preference: if the evidence genuinely does not fit, that is a signal it was never chat content. Publish the long form as a document or a card, link it, and keep the message to the headline, the three or four findings that drive a decision, and the link. Never solve an overflow by deleting a finding.
```
jayantid: **Blocker — "splits at 4000" looks wrong for the path this PR's template drives.**

True for kanban `result`: the notifier calls `adapter.send()` directly. But the watcher template says post via the notification tool — `send_notification` → `hermes send` → `gateway/delivery.py`, which truncates with a `... [truncated]` footer unless the adapter sets `splits_long_messages`. `GoogleChatAdapter` doesn't (`bootstrap_onboarding/README.md:216`); the only flip is in the Chat Agent's plugin, different process, silent `except`.

So a 4,300-char triage loses its tail — and the tail is `## What to do` with the `apply` words. Can you check `delivery.py` in the base image? If it truncates, this rule should say that rather than "splits mid-sentence".

deploy/docker/patches/kanban_report_format.py
```diff
@@ -109,12 +109,16 @@
...
   the card title, so an H1 renders as a second, duplicate banner.
-- Put tabular data in a Markdown pipe table with a `---` separator row.
+- Put tabular data in a Markdown pipe table with a `---` separator row, but
```
jayantid, line 134: While you're in this file: `DEFECT_ADVICE` below wasn't updated and still teaches the old style. Line 206 tells the worker "open with one sentence saying what happened, then the list" — the throat-clearing §7 rule 3 cuts. Line 203 still recommends a pipe table for tabular data with no Chat caveat, after you bounded it here.

The module comment says this text goes straight back to the model, so it undoes the persona at the point of correction.

agents/platform/scripts/session_kv_server.py
```diff
@@ -541,17 +550,19 @@ def _build_agent_query(session_id: str, payload: Dict[str, Any]) -> str:
...
+        f"<Short 1-sentence description of the problem>\n\n"
+        f"## Why\n\n"
+        f"- <Key constraint mismatch or log finding in 1-2 sentences, with the evidence that proves it>\n\n"
+        f"## What to do\n\n"
         f"- **Option A (<Action Title>):** <1-sentence description of Option A GitOps fix>.\n"
-        f"- **Option B (<Action Title>):** <1-sentence description of Option B GitOps fix>.\n\n"
-        f"✅ **Recommended: Option <letter>** — <1-sentence why this is the safer/better choice>.\n\n"
+        f"- **Option B (<Action Title>):** <1-sentence description of Option B GitOps fix>.\n"
+        f"- ✅ **Recommended: Option <letter>** — <1-sentence why this is the safer/better choice>.\n"
+        f"- **Reply 'apply'** to open a GitOps Pull Request with the recommended fix, or name one directly with **'apply Option A'** / **'apply Option B'**.\n\n"
```
jayantid: The `apply` line is now the fourth bullet in the same list as Option A / Option B / Recommended. The blank line and `👉` used to mark it as a separate call to action; the risk is the model renders it as a third option.

Your tests pin the prompt, not the output, and "Not covered" says no full agent turn ran on this path. Worth one live watcher run, or an explicit "this is the call to action, not an option" note in the template.

agents/platform/scripts/test_session_kv_server.py
```diff
@@ -637,11 +637,45 @@ def test_call_to_action_names_options_instead_of_a_placeholder(self):
             "message": "some message"
         }
         query = session_kv_server._build_agent_query("test-session", payload)
-        cta = next(line for line in query.splitlines() if line.startswith("👉"))
+        cta = next(line for line in query.splitlines() if "apply Option A" in line)
```
jayantid: This now asserts its own selector: it picks the line containing `apply Option A` and the `assertIn` that used to check for it is gone.

It still fails today via `StopIteration`, but `assertNotIn("<letter>", cta)` now checks whichever line matches first — and the instruction prose above the template already talks about `apply Option <letter>`. Anchoring on the first `- **Reply` line after `## What to do` keeps the original guarantee.

deploy/docker/patches/kanban_report_format.py
bradhoekstra: Agreed — fixed in `fa9e2f2`, and there was a third copy.

- `heading-without-prose` no longer asks for "one sentence saying what happened": "Put the answer on the first line — what is true, or what is wrong and what you want done — then the list. One line, not a preamble."
- `ascii-substitute` picked up the Chat caveat the stanza got: a pipe table is "fine for a few short columns, but Google Chat drops tables, so never let one be the only place a fact lives."
- `top-level-heading` dropped its `###`.

The third copy is `NEW_RESULT_DESCRIPTION` in `kanban_result_required.py:307` — the `result` argument's schema wording, which this module's own docstring names as one of the three homes that "must not answer 'is this report well-shaped?' three different ways". It still said `` `##`/`###` `` and recommended a pipe table with no caveat. It now matches the other two, and picked up the link rule.

deploy/docker/patches/kanban_report_format.py
bradhoekstra: Agreed, and fixed in `fa9e2f2` — you're right that the stanza decides the shape, not the persona. The bullets now read:

- Lead with the answer: what is true, or what is wrong and what you want done. Then the detail. Do not narrate the request back or how you investigated.
- Use `##` for sections. Never `#` — the chat message already shows the card title, so an H1 renders as a second, duplicate banner — and no `###`: Google Chat flattens every level to bold, so a sub-level is invisible there. If you are triaging an incident, SOUL.md §7 fixes the sections; use exactly those.
- Aim under 2,000 characters. Past 4,000 Google Chat delivers your report as several messages rather than one, so if the deliverable is genuinely longer, publish it, link it, and keep `result` to the headline findings and that link. Never drop a finding to fit.
- Link every artifact you name — cluster, workload, card, PR, issue, console view — as `[text](url)`. Both platforms convert it; a bare id is clickable on neither.

I phrased the length rule as "several messages rather than one" rather than §7's "wall", because a kanban `result` genuinely does split on this path — it is the cron path that truncates (see my reply on §7 rule 6).

The stanza is now 1,558 characters against 635, and it rides every card body; if that is too much rent for two rules that only bite on triage, say so and I'll put the link and length bullets behind the §7 reference instead.

One thing I did differently from your version: the module docstring's claim that "Every rule here is one the detector below can actually measure" stops being true, since none of the four new rules are measured. Rather than add defect classes nobody has asked for, I inverted the claim — everything measured is stated, not the reverse — and said why in the comment.

agents/platform/SOUL.md
bradhoekstra: Checked it. `delivery.py` does truncate — but not on this path.

`send_notification` → `hermes send` → `hermes_cli/send_cmd.py`, whose own docstring calls it "a thin wrapper around `tools.send_message_tool.send_message_tool`" (`send_cmd.py:4`, dispatch at `:364`). `send_message_tool` reads `max_message_length` off the plugin's `PlatformEntry` (4000 for `google_chat`, `adapter.py:3714`), chunks with `BasePlatformAdapter.truncate_message`, and loops `for chunk in chunks` sending every one. `gateway/delivery.py` is never imported. So the watcher's report does not lose its tail.

But the mechanism you found is real one path over. `cron/scheduler.py:1800` constructs a `DeliveryRouter` and calls `_deliver_to_platform` directly — the function with the `MAX_PLATFORM_OUTPUT` guard — so a **cron** delivery over 4000 is cut off behind `... [truncated, full output saved to <path>]`, exactly as you describe, because `GoogleChatAdapter` never declares `splits_long_messages` (it is set on slack/telegram/discord/teams/matrix/feishu/mattermost/whatsapp and not on google_chat). `agents/platform/governance/inventory_prioritize_sop.md:200` already documents this for the onboarding report. §7 governs cron watchdog syntheses too, so the rule needed to say it.

And "mid-sentence" was wrong regardless: both chunkers break at the nearest newline below the cap — `truncate_message` (`gateway/platforms/base.py:6693`) and `GoogleChatAdapter._chunk_text` (`adapter.py:2362`, which only hard-cuts when the last newline falls in the first half of the window).

`fa9e2f2` rewrites both places. §0 now distinguishes the two behaviours by how the message was sent, and §7 rule 6 names what the cron truncation eats: the tail is **What to do**, so an overlong triage is one that loses its recommendation. That is your point, landing where it actually bites.

All of the above is read out of `nousresearch/hermes-agent:v2026.8.3@sha256:1678831…`, the digest `tags.env` pins.

Leaving this open for you, since it is a correction rather than an agreement.

agents/platform/scripts/session_kv_server.py
bradhoekstra: Agreed. I took the template note rather than the live run.

The bullet is now `- **To authorize:** reply **'apply'** …`, and the instruction prose above the template says it outright: "The last bullet under '## What to do' is the call to action, not another option: keep its 'To authorize:' label, never give it an Option letter, and never count it when you number the options." `test_approval_line_survives_inside_what_to_do` also asserts that bullet carries no `**Option <X>` of its own.

On the live run: it still posts an alert into a real Google Chat space — `_post_initial_alert` fires before the agent is invoked — and this installation deliberately runs with the watcher paused. So what stays unverified is the model's compliance with the template, which is the same gap the PR body already declares for this path. If you would rather have the observation than the note, I can unpause the watcher against a throwaway space and drive one.

agents/platform/scripts/test_session_kv_server.py
bradhoekstra: Agreed — fixed in `fa9e2f2`. It now anchors on the label inside the section:

```python
what_to_do = query.split("## What to do", 1)[1]
cta = next(
    line for line in what_to_do.splitlines()
    if line.startswith("- **To authorize:**")
)
self.assertNotIn("<letter>", cta)
self.assertIn("apply Option A", cta)
self.assertIn("apply Option B", cta)
```

The `assertIn("apply Option A", cta)` you noticed had gone is back, and the comment now says why the selector must not be the text under test.

jayantid (approved):

PR #699, opened 2026-08-14
feat(operator): experimental platform-front-door profile

bradhoekstra:
## Summary

This adds an opt-in `spec.harness.experimental.platformFrontDoor` flag to the `PlatformAgent` CR
that makes the Platform Agent the profile the Hermes gateway runs as, so a Google Chat or Slack
message is handled by it directly instead of arriving at the Chat Agent and being delegated through
the router and the kanban board. The lever already exists in Hermes and needs no patching —
`_apply_profile_override()` pre-parses `-p/--profile` out of argv before any import and re-points
`HERMES_HOME`, so the whole change is the operator deciding to pass it, plus the two consequences
that follow: the platform profile now needs the chat-ingress config keys it has never carried, and
its `config.yaml` stops being a file only the image writes to. It is off by default, carries no
compatibility promise, and flips back with a second one-line patch.

## Why This Change

The Chat Agent is deliberately a router: `toolsets: [kanban]`, `platform_toolsets` limited to
`{mcp-router, kanban, memory}`, and a large `agent.disabled_toolsets` denylist. Everything real is
delegated to the `platform` profile, so every chat request pays a hand-off — a card, a worker turn,
and a summary back. That is a design worth having, and also a design worth being able to measure
against the alternative. Without a flag, "what if the agent that holds the tools answered the
message itself" is a rebuild and a redeploy, which is enough friction that nobody runs the
experiment. This is the same "which profile owns the turn" seam as #630.

The cost is explicit and is the reason this is `experimental` rather than a supported mode: with it
on, an inbound chat message reaches the full Platform Agent tool surface with no card and no worker
turn in between. One gateway means one profile, so while it is on the Chat Agent persona sees no
chat at all.

## What Changed

- **Gateway args.** With the flag on the gateway container's `args` become
  `hermes --profile platform gateway run` — still `args`, never `command`, so the entrypoint keeps
  running. At `replicas > 1` the `leader_elect.py` wrapper is left in place and reads the new
  `HERMES_GATEWAY_PROFILE` env var instead, so both paths produce the same argv.
- **`HERMES_GATEWAY_PROFILE` on the gateway container only** — `platform` with the flag on, empty
  with it off, but always emitted. It is read by `leader_elect.py` and by `docker-entrypoint.sh`,
  both of which treat empty as off. Emitting it unconditionally is what keeps an `AgentPlugin`'s
  `spec.env` — copied verbatim, no allowlist — from becoming the only writer of the name on a
[...]

PR #10, opened 2026-05-15
Bootstrap the Platform Agent

fatoshoti:
Bootstrap the Platform Agent

PR #720, opened 2026-08-15
feat(security): F10 slice 2b — Pod split, git hardening, content passing, scoped SA pool, conformance suite

bnaylor:
## Why This Change

F10 (the agent permission and security model). Ten increments, developed and landed one at a time on a fork, brought here as one branch. Between them they close four findings reproduced on a live cluster and add the machinery to stop them regressing.

The headline is that the agent sandbox could read the credential holder's environment. Sandbox and credential proxy shared a PID namespace **and a UID**, so `/proc/<pid>/environ` was readable across the boundary — every credential the proxy holds, including chat tokens, reachable by a prompt-injected agent with no process compromise, bypassing the policy layer entirely.

## What Changed

Grouped as they landed. Each was reviewed and merged separately on the fork.

| # | Change |
| --- | --- |
| 1 | `docs/architecture/09` — the capability envelope (design only, nothing built) |
| 2 | Verb and flag allowlists for `kubectl`/`gcloud` at the credential proxy |
| 3 | `make verify`; `tests/` put into CI, where it had never run |
| 4 | Credential proxy becomes a **native sidecar** so it binds :8643 before the sandbox exists |
| 5 | The GitHub write path denied — `gh pr merge`, `gh pr review --approve`, `gh api -X PUT`, pipeline triggers, repo administration |
| 6 | **Slice 2b**: PID namespace and UID unshared; `splitCredentialBrokerPod` and `egressPolicy: Allowlist` (both opt-in, both default off); the `gke-admin` permission set removed |
| 7 | `git` hardened against config-driven code execution |
| 8 | Broker-owned git trees, behind a flag |
| 9 | A service account per cluster — selection, refusal and minting |
| 10 | The security invariants as an executable conformance suite, with a mutation harness |

**Two things a reviewer should look at first.**

`gke-admin` is **removed**, not deprecated (#6). `roles/container.admin` authorizes through IAM regardless of Kubernetes RBAC, and its `container.clusters.impersonate` cannot be scoped by IAM, so the set did not merely widen the ceiling — it nullified the RBAC intersection. `read-only` and explicit `custom` remain; a cached `vars.sh` naming `gke-admin` now fails with an explanation rather than a bare "invalid".

**The scoped service-account pool (#9) deliberately grants nothing.** It was designed around an IAM Condition on the cluster's `resource.name`; that was measured against live GCP on 12 August and four condition spellings were all refused — the mechanism does not reach Kubernetes object authorization. The dead grant was removed and the pool disarmed by default. What lands is the selection logic, the refusal path, the minting and the kubeconfig rewrite, all measured working; only the authority is absent, and the replacement is per-cluster RBAC in a later slice. There is a test, `test_the_agent_is_never_narrowed_while_the_pool_grants_nothing`, guarding against the half-way state — arming the pool while the agent stays wide leaves the ceiling the pool exists to remove; narrowing the agent while the pool grants nothing is a total outage.

## Why This Matters

- Four findings reproduced on a live cluster are closed, and the two most serious were re-reproduced on a cluster *before and after* this branch (below).
- The conformance suite records **12 invariants the product still violates** as expected failures. That is not tolerating them — it is giving each a name, a line number, and a test that goes loud the day it closes. One did exactly that during this merge: the git-RCE assertion flipped to an unexpected success once #7 was in the same tree as the suite, so its decorator came off in this branch.

## Context

Developed as `sackheads/kube-agents` PRs #1–#10, all merged there. Reasoning that is not in the diff lives in a private decision log (D1–D23); D3 is the scoped-SA story and is summarised above.

**Open PRs that touch the same files.** None appears to solve the same problem, so these are merge-conflict warnings rather than duplicate work — flagging so whoever merges second knows to look:

- **#674** (@AntonTyb, read-only root filesystem) — `platformagent_manifests.go`, its test, all three goldens, `credential-isolation-design.md`. Closest in intent; both harden the agent Pod.
- **#648** (@bradhoekstra, DNS control plane endpoint) — `credential_proxy.py`, `test_credential_proxy.py`, the same manifests and test.
- **#699** (platform-front-door profile) — the CRD, `common_types.go`, `zz_generated.deepcopy.go`, `security-and-iam.md`.
- **#588**, **#573** — `platformagent_manifests.go` and `provision_04_gcp_iam.sh`.
- **#659**, **#571** — chart values and `platform-agent-cr.yaml`.
- **#706** (Envoy v1.39.0), **#718** (Hermes base image) — `deploy/docker/Dockerfile`.

**This is a large PR: 97 commits, 106 files.** It is one branch because the ten increments are interdependent — the conformance suite asserts controls added by six of the others, and the Pod split is a base for three. Happy to split it along the numbered boundaries above if you would rather review it in pieces; each was a self-contained PR on the fork and the commit history is preserved.

## Testing

- `make verify` — green (go build, vet, test, both Python suites).
- `make conformance` — **106 tests, OK, 12 expected failures.**
- Mutation harness over the whole suite — **killed=56, survived=0, stale=0.** Every assertion has been shown capable of failing.
- `make docs-check` and `prettier --check` (3.9.6) — green on all 44 changed formattable files.

Conflicts resolved against this branch's base were mechanical and are described per merge commit; the substantive one was `SLACK_RELAY_URL`, where upstream's batch-append refactor was kept and only the value changed to `credentialProxyBaseURL(agent)`, which follows the relay into the broker Pod when it is split and still returns loopback otherwise.

### Live validation

Exercised against **`bnaylor-ka-test`** (GKE 1.35.6-gke.1258000, `us-east4`, project `bnaylor-kagents-dev`), an existing install on agent image `dev-20260812-184046`. Operator and credential-proxy images were built per change, deployed, observed, then reverted; the cluster was returned to its pre-session state each time.

**Both headline findings were reproduced on the baseline first**, so the result is a comparison, not an assertion:

| | Before | After |
| --- | --- | --- |
| Sandbox / proxy UID | both `10000` | `10000` / `10001` |
| `shareProcessNamespace` | `true` | unset |
| Proxy processes visible from sandbox | `envoy` (pid 206) | none |
| Processes exposing proxy credentials to the sandbox | **1** — all 73 env vars incl. `SLACK_BOT_TOKEN`, `SLACK_APP_TOKEN` | **0** of 5 visible |
| `git -c protocol.ext.allow=always clone "ext::sh -c id"` | executed `id` on the proxy | refused, `policy rule: git.argument.refused` |
| `gh pr merge` / `--approve` / `api -X PUT` | reached `gh` | refused on `github.merge` / `github.assent` / `github.api-mutation` |
| Proxy container start vs sandbox | proxy 7s **after** sandbox | proxy in `initContainers[restartPolicy=Always]`, kubelet-ordered |

Credential reads were done from inside the running `platform-agent` container; variable **names** only were recorded, never values.

**The C5 NetworkPolicy behaviour is worse than the branch's own commit message claimed, and only a cluster showed it.** Applying `<name>-sandbox-metadata-deny` by hand does not get it deleted — `deleteLegacyCredentialIsolationResources` hits its `IsControlledBy` guard and returns `refusing to delete unowned legacy *v1.NetworkPolicy`, failing **every** reconcile in a hot loop. Because the operator no longer creates that policy, nothing in the wild owns it, so an admin following the security docs lands on the error path, not the documented deletion path. Bounded but real: the cleanup runs after RBAC/ConfigMap/Workload/Service/NetworkPolicy, so what is blocked is `updateStatusReady` — the CR's status silently stops tracking reality. With this branch the identical apply survives with zero reconciler errors.

Also verified live: `/v1/workspace/*` returns 404 with the flag off and serves a full open/list/read/close lifecycle with it on, with the working tree in the broker's own volume and unreadable from the sandbox (#8); the scoped-SA pool renders its ConfigMap on a CR change, refuses an unconfigured cluster with `PoolRefusal` before any GKE call, and selects the right account for a configured one (#9).

**Not covered.** The opt-in gates were exercised individually rather than in combination. `egressPolicy: Allowlist` was not exercised end-to-end on a CNI enforcing NetworkPolicy — see the warning below. Bucket 2 of the conformance suite (the cluster-bound scenarios) is gated off behind `KUBE_AGENTS_CONFORMANCE_CLUSTER` and did not run.

---

> [!WARNING]
> **`egressPolicy: Allowlist` does not currently deny the metadata server, despite the name.** NetworkPolicies union, and two policies already re-add that path over the same Pods: the operator-rendered `<name>-gateway-netpol`, and `platform-agent-core-egress` from `deploy/kustomize/platform/networkpolicy-core-egress.yaml`. Both allow `169.254.169.254/32` on TCP 80/8080 and `169.254.169.252/32` on 988. A customer flipping this flag expecting metadata denial does not get it.
>
> This is not a defect in #676, which is deliberate — Workload Identity needs that path, and in the sidecar layout the credential proxy shares the Pod. Scoping those rules to the broker Pod under `splitCredentialBrokerPod` is the fix and is a later slice. It is recorded as a KNOWN VIOLATION in the conformance suite (`test_C1_the_rendered_egress_policy_reaches_no_metadata_address`) rather than papered over, so it goes loud the day it closes. Everything else the flag does — the default-deny policy itself — works.

Risk: the two new capabilities are opt-in and default off. The namespace/UID unshare, the native sidecar, the `gh` denials and the `gke-admin` removal are not opt-in and change every rendered Pod or every install. This PR was generated with the help of Claude.


bnaylor:
Going to go ahead and split this, this is way too big.  Even the bots can't handle it!

bnaylor:
## Split — review the five small PRs first

This was 97 commits and neither the review bot nor a human could reasonably read it. The five increments that are genuinely independent are now separate PRs, each with its own live validation:

| PR | Increment | Size | Depends on |
| --- | --- | --- | --- |
| #722 | `make verify`; put `tests/` in CI; fix the test pollution | 3 commits, 6 files | — **merge first**, the rest state `make verify` as their bar |
| #721 | Architecture doc 09, the capability envelope | 3 commits, 3 files | — docs only |
| #723 | kubectl/gcloud verb and flag allowlists | 19 commits, 5 files | — |
| #724 | Credential proxy as a native sidecar (port 8643 preempt) | 3 commits, 7 files | — |
| #725 | Deny the GitHub write path (`gh pr merge`, `--approve`, `gh api -X PUT`, …) | 2 commits, 7 files | #724 |

**This PR is now the remainder**: slice 2b's Pod split, the git hardening, content passing, the scoped SA pool, and the conformance suite. Its diff will shrink on its own as the five above merge — everything they contain drops out.

### Why the remainder is not split further

I tried, and stopped when the attempt started producing wrong code rather than smaller PRs. Two things make these five genuinely one unit:

- **Slice 2b's tests depend on #723.** `test_credential_proxy.py` uses `CredentialProxyHandler.enforce_read_only`, which the verb-allowlist work introduces. Rebasing slice 2b onto a chain without it produces 24 errors in a suite that is green in this PR.
- **The other three stack on slice 2b and on each other** — the git hardening extends the proxy that content passing then restructures, and the scoped SA pool changes the credential the same code path selects. Their branches also absorbed one another during the fork's merges, so reconstructing clean per-increment slices means replaying ~60 commits through conflicts that were already resolved once here.

If you would still rather have them separate, the honest way is to merge the five above first and then re-cut this branch against the smaller base — most of the conflicts disappear at that point. Happy to do that on request.

### Ordering

`#722` → `#721` / `#723` / `#724` (any order) → `#725` → this.

Comment generated with the help of Claude.

dshnayder:
Not a review — a dependency note, plus one observation.

**This PR is a prerequisite for the shell-sandboxing work in #737**, and I wanted that recorded somewhere you'd see it.

#737 Part B moves the Platform Agent's shell out of the agent pod into a sandbox pod, reached over Hermes' `ssh` terminal backend. That only works if the credential broker has a network address of its own: a shell in another pod cannot reach a broker bound to the agent pod's loopback interface, so without `splitCredentialBrokerPod` the sandboxed shell loses `kubectl`, `gcloud`, `gh`, and `git` outright. `buildCredentialBrokerDeployment` is what makes the sandbox shippable rather than a demo. So the sequencing on our side is: this merges, then the sandbox becomes more than a file-operations experiment.

Two consequences worth flagging while the design is still in your hands:

1. **The broker will eventually take callers from outside the agent pod.** The exec path on `:8765` has no caller authentication today — loopback plus the UDS *is* the access control, unlike `AgentAPIProxyHandler`, which requires `API_SERVER_EXTERNAL_KEY`. That is sound while every caller is in the pod. If the broker's Service ever needs to admit a sandbox pod, it needs a bearer token first. Not something this PR has to solve, but it's worth not designing it out.
2. **`credential_proxy_client.py` posts `"cwd": os.getcwd()` and the broker rejects anything outside `CREDENTIAL_PROXY_WORKSPACE_ROOT`.** That check assumes the caller and the broker see the same filesystem, which stops being true once the shell is in a different pod. It will need re-basing — and it's self-reported by the caller either way, so it's a guardrail rather than a boundary.

**Separately, on the metadata server.** The reasoning in the `egressPolicy` API documentation is exactly right, and it matches a finding we reached independently while scoping #737 — including that a Pod-level NetworkPolicy cannot distinguish the two containers. The one thing I'd add is about where the protection bottoms out:

> The policy does nothing at all on a cluster whose CNI does not enforce NetworkPolicy (GKE Standard without network policy enabled)

This is the case on the install we develop against — network policy is disabled on it — so `Allowlist` is inert there and the metadata-server path stays open even with the split on. The complementary control that doesn't depend on the CNI is to **remove the `iam.gke.io/gcp-service-account` annotation from the agent pod's service account** once the broker has its own: deny the route versus remove the identity. Nothing to reach, no enforcement required, and no operator guard needed against someone turning the policy off.

I don't think that belongs in this PR — it's a separate, small change that layers on top, and we'll pick it up as the remainder of #737 Part C. Flagging it so the enforcement dependency is a known limit rather than a surprise, and so we don't both write it.

_Drafted with the help of Claude._

bnaylor:
## The split is done - this is the map

I said in this thread that I'd split this, and that the honest way was to merge the small independent pieces first and then re-cut the remainder against the smaller base.  That has now happened.

### Already merged

All five of the first batch are in `main`:

- #722 - `make verify`, and `tests/` in CI
- #721 - architecture doc 09, the capability envelope
- #723 - kubectl/gcloud verb and flag allowlists
- #724 - the credential proxy as a native sidecar
- #725 - denying the GitHub write path

Also #727, for core invariant 3 stating a transport rather than the property it protects.

### Open now

The remainder is seven PRs.  They were cut fresh off `main` rather than rebased.

| PR | What | Depends on |
| --- | --- | --- |
| #952 | git hardened against config-driven code execution | - |
| #955 | the sandbox and the credential proxy stop sharing a process namespace and a user | - |
| #961 | the `gke-admin` permission set removed, plus a ValidatingAdmissionPolicy backstop | - |
| #960 | the credential broker in its own Pod, opt-in | #955 |
| #963 | `egressPolicy: Allowlist`, opt-in, plus a fix for a reconcile bug that froze the CR's status | #960 |
| #962 | broker-owned git working trees, behind a flag | #952 |
| #966 | a service account per cluster: selection, refusal and minting, with the authority deliberately absent | #960 |

#955 is the headline finding of the whole thing: `/proc/<pid>/environ` was readable across that boundary.  It's default-on and it changes every rendered Pod, so it's the one to read closely.  #960's dependency on it is narrow - it compiles against the UID constants #955 introduces.  #961 doesn't touch the other six and can land whenever.

Suggested order: #952, #955 and #961 first, in whatever order suits.  Then #960.  Then #963, #962 and #966.

### Still to come

The security invariants as an executable conformance suite, with a mutation harness.  It lands last, once the controls it asserts are in `main`.  Its expected-failure list has to be measured against the merged result rather than guessed at.

### Three corrections to this PR's body

**1.** This body presented ReadWriteMany storage as what the Pod split requires.  That framing was wrong.  RWX is one way to satisfy how the workspace is shared today, and it isn't something the product should be asking of anyone.  #962's content passing is the direction that removes the coupling.  #960 says so now.

**2.** This body said `roles/container.clusterAdmin` carries `container.clusters.impersonate`.  It doesn't - only `roles/container.admin` does.  And "IAM has no `resourceNames` equivalent" borrowed RBAC vocabulary that IAM doesn't have.  The accurate statement is GKE's own: there is no mechanism for granting permissions on specific Kubernetes objects within IAM.  #961 carries the corrected wording.  The argument for removing the set is unchanged, and arguably cleaner for it.

**3.** Splitting it up was worth doing for more than reviewability.  Working the pieces separately turned up:

- four more paths to code execution in the credential container that the original branch did not have (all in #952)
- a test that asserted nothing about the thing it was named for (#955)
- an admission policy that matched zero objects on a default install, and whose one hard-coded exemption would have denied the operator's own reconcile (#961)
- two concurrency bugs that could orphan trees and cross-contaminate commits (#962)
- a Terraform default that would have armed the scoped-SA pool on every stock install, which is a total outage (#966)

None of those were reachable by reading 25k lines at once.

### @dshnayder

Your dependency note above - the shell sandboxing needing the broker to have a network address of its own - is answered by #960.  That's the one to follow.

### This PR

The branch behind it stays where it is, as the audit trail.  It isn't going anywhere and its history is indexed by commit SHA.  I'd like to close this once the conformance suite is open, so that until then the whole set is navigable from one place.

Comment generated with the help of Claude.

bnaylor:
Closing as fully split and landed. Every increment of this branch is now on `main` or open as its own reviewed PR:

| # | Increment | Where it went |
| --- | --- | --- |
| 1 | `docs/architecture/09` capability envelope | on `main` |
| 2 | kubectl/gcloud verb and flag allowlists | earlier merged slices |
| 3 | `make verify`; `tests/` into CI | earlier merged slices |
| 4 | credential proxy as a native sidecar | earlier merged slices |
| 5 | GitHub write path denied | earlier merged slices |
| 6 | PID/UID unshare, `splitCredentialBrokerPod`, `egressPolicy: Allowlist`, `gke-admin` removal | #955, #960, #963, #961 |
| 7 | git hardened against config-driven code execution | on `main` (this branch's copy adds nothing `main` lacks) |
| 8 | broker-owned git trees behind a flag (content passing) | merged (`content_workspace.py`) |
| 9 | scoped service-account pool | #966 |
| 10 | conformance suite + mutation harness | **#1121**, re-landed against today's `main` |

Nothing else here survives: the two `provision_0*.sh` edits and the VAP-manifest rename target files #797 and later changes deleted.

Two notes for anyone reading this later. The per-increment self-reviews and live validation recorded in this description carried over into the split PRs' own bodies, each re-run at its head rather than inherited. And the conformance suite's re-land was not a lift-and-shift — #1121's Context section records what three months of drift did to it, including that #676 reopened C1's egress form, which the suite now carries as a named known violation rather than silence.

PR #728, opened 2026-08-17
feat(security): verify baked skill trees against a build-time manifest

AntonTyb:
## Summary

A skill is prompt material: a `SKILL.md` is instructions the agent follows, a `SOUL.md` is the persona it follows them as, and a skill's `scripts/*.py` runs with the agent's own credentials. Nothing in the image compared what a pod boots against what the build produced, and nothing stopped the sandbox uid from rewriting either. This adds both halves of that: the prompt material baked into the image becomes root-owned, so the agent (uid 10000) can read and execute it and write none of it; and the build writes a sorted SHA-256 manifest beside each of the three skill trees, which the container entrypoint re-checks at startup — before any of it is copied onto the data volume — refusing to start if a tree no longer matches. Both fold into a `RUN` the Dockerfile already had, and folding the platform stage's second `compileall` pass in with them removes an instruction, so the image ends up one layer smaller than `main`.

## Why This Change

The baked prompt material sits inside the trust boundary and was the only part of it with no integrity story at all. A tree altered between build and boot — by a corrupted layer, a bad build, or a write from a previous boot under an image that still allowed one — was scaffolded into every profile and loaded, indistinguishable from a good one. There was also no cheap way to answer "are this pod's skills the ones this image was built with?" during an incident.

The two changes cover different things and neither is sufficient alone. Ownership is the barrier: it stops the agent writing in the first place. The manifest is the detection: it catches what ownership cannot — a corrupted layer, a bad build, a future code path that runs as root, or damage that predates the barrier.

Ownership sits **beneath** `readOnlyRootFilesystem`, which #674 has since put on every container the operator builds and which already turns each of these writes into `EROFS`. What ownership adds is everything that flag does not reach: the image run outside this operator, a container someone gives a writable root filesystem, and the build itself. In an operator-rendered pod today the barrier is defence in depth, not the thing standing between the agent and its own instructions.

This is deliberately scoped to detection at boot on the image trees, and the docs say so rather than implying more. It is not a runtime sandbox. It does not cover the per-profile copies on the PVC, which are legitimately rewritten after the copy (profile scaffolding overlays them, the hourly cluster reconcile re-overlays them) and so have no fixed checksum to compare against; nor the rest of `/opt/defaults`, which stays agent-owned and is still read back from the image after the boot copy (`plugins/` by the OTel config generator, `onboarding/` by the `bootstrap_onboarding` plugin) — uncovered residue that only the read-only root filesystem closes.

## What Changed

- **`deploy/docker/Dockerfile`** — the existing permission-normalization `RUN` gains a loop over `/opt/hermes/skills`, `/opt/platform-template/skills` and `/opt/cluster-template/skills` that writes `skills_manifest.sha256` into each, and closes with one `chown -R root:root` over `/opt/hermes/skills`, `/opt/hermes/plugins`, the two template **roots**, and `/opt/defaults/scripts`. Five things about it are load-bearing and are commented in place:
  - **The barrier is wider than the hashed trees.** It covers the template roots rather than just `skills/`, because `SOUL.md`, `AGENTS.md`, `CAPABILITIES.md`, `governance/`, `cron/` and `config.yaml` sit beside `skills/` and are force-synced into every profile on every boot by the same entrypoint step. It covers `/opt/hermes/plugins`, the only path Hermes resolves plugins from, holding Python imported into the agent's own process — a strictly stronger capability than a `SKILL.md`. And it covers `/opt/defaults/scripts`, which holds the shared scripts the entrypoint invokes by absolute path, the `sitecustomize.py` the operator puts on `PYTHONPATH`, and the manifest checker itself: a checker the checked party can rewrite reports whatever it is told to. Nothing at runtime writes to any of them — every reference in the tree is a read or a copy **out** into `$HERMES_HOME`.
  - **Modes stay 0644/0755.** `sync_profile_skills` copies with `cp -a`, which preserves mode but cannot preserve ownership as non-root. A `0444`/`0555` tree here would therefore arrive read-only on the PVC and break both the profile overlay and the staging swap's own `rm -rf`. Ownership does not travel; mode does.
  - **One layer fewer, not one more.** `agent-base -> platform` is the deepest chain that ships and it is what `scripts/check_image_layers.py` measures against overlay2's 128-layer ceiling. The manifest work appends continuation lines to an existing `RUN`, and folding the platform stage's second `compileall` pass into that same `RUN` retires the instruction it used to have to itself.
  - **Bytecode is inside the manifest, not carved out of it.** That is why the `compileall` pass moved: it now runs in the same `RUN`, after the `chmod` loops (so its `__pycache__` is not swept by the `*/scripts/*` `chmod 755`) and before the manifest loop (so its output is hashed). CPython's default invalidation is source mtime plus size, not a content hash, and nothing here passes `--invalidation-mode checked-hash` or sets `SOURCE_DATE_EPOCH` — so a `.pyc` is an independent artifact, and excluding one would have left a hole in the middle of what the manifest claims to cover.
  - **The build fails loudly on three things** rather than shipping a manifest that verifies clean or an image that cannot boot: an entry count that disagrees with `find` (the generating pipeline exits with `sort`'s status and `sh` has no `pipefail`), any symlink in a skill tree (`find -type f` never records one, and the verifier refuses every one it sees, so the two sides must agree that a symlink is an anomaly), and a tree that does not exist at manifest generation (the boot check refuses to start a pod whose tree has no manifest, so skipping one here would mean a green build and a fleet-wide crash-loop).
- **`agents/platform/scripts/verify_skills_provenance.py`** (new) — reads a `sha256sum` manifest and reports content changes, untracked files, missing files and symlinks, collecting every difference so one failed boot gives the whole picture instead of one file per crash-loop iteration. Every file under the tree is compared and nothing is skipped by name: an exclusion list is also a list of names a symlink can be given, so the symlink test is the first thing every entry meets. Stdlib only, so it runs under the venv interpreter with no dependency on the test extras.
- **`deploy/shared/docker-entrypoint.sh`** — new step 1.55, after the shared-state gate and before the bootstrap lock, so it runs ahead of everything that propagates a tree. Fail-closed with `exit 1`, which is a deliberate departure from the `WARN`-and-continue every other step uses; the comment gives the reasoning, including why there is no env-var override. **The manifest, not the verifier script, is what makes the check mandatory.** Both sides are root-owned in the image, so the missing-checker branch reports a corrupted or truncated image rather than guarding against an `rm` the runtime uid can issue; it is there because the alternative to reporting a missing checker is skipping the tree it would have checked. A tree with no manifest is skipped, which keeps the same entrypoint a no-op in `agent-base` and on a developer host.
- **Docs** — one bullet in the security reference stating exactly what the barrier covers, that personas, plugins and shared scripts get the barrier and not the hash, and what is left uncovered; and two paragraphs in the docker-images page so someone debugging a crash-looping pod finds both fatal messages and knows they mean "reinstate the image", not "change a setting".

## Context

Replaces the skill-provenance workstream from #489, which I closed: it was 90 commits behind `main` with every substantive file conflicting, and its other three workstreams were either superseded by the credential proxy or in conflict with `main`'s persona. This is the one part of it that was still unowned — nothing on `main` does it, and it is not in the F10 slices (#720–#727) or the #721 capability envelope.

- **#674** (`feat(security): enforce a read-only root filesystem on the agent containers`, mine) **merged on 2026-08-25** and this branch now carries it. It is the general form of the barrier this PR builds narrowly, which is why the framing above says the ownership half is defence in depth in an operator-rendered pod rather than the only thing there. The textual conflict in `docs/site/src/content/docs/deploy/docker-images.md` that was expected between the two is resolved in the merge commit; both paragraphs are kept.
- **#718** (Hermes base image `v2026.8.13`) **merged**, and this branch is merged up to it, so the build's symlink assertion has now been exercised against the current base rather than `v2026.8.3`.
- **#720** (F10 slice 2b) is still open and rewrites large parts of both `deploy/docker/Dockerfile` and `deploy/shared/docker-entrypoint.sh`. Expect a textual conflict in both; the changes are independent in substance and #720's sidecar uid split does not interact with the ownership used here.

## Testing

CI is green at the head of this branch, including the three jobs this host cannot run:

- **Docker Build** — `platform`, `credential-proxy` and `entrypoint-gate-test` all build. The gate-test stage runs the real entrypoint **as uid 10000 inside the built image**, and its assertions depend on step 2 having run, which is below the new step 1.55 — so a green gate is direct evidence that the build's symlink and completeness assertions passed, the manifests were written, and all three trees verified clean at boot as the runtime uid. With bytecode now inside the manifest, that green is also what proves the `compileall` ordering is right: a pass that ran after the hashing would fail every one of those three verifications.
- **Check image layer budget** — `scripts/check_image_layers.py` against the `platform` target passes, with one instruction fewer than `main`.
- **prettier** — passes both in CI and locally at the pinned 3.9.6.

Locally:

- `agents/platform/scripts/test_verify_skills_provenance.py` (20) and `deploy/docker/test_image_startup_contract.py`'s `SkillProvenanceContractTest` (16 of the file's 34) both pass under `unittest`, as does `tests/test_docker_entrypoint.py` (57). `make test-python` as a whole stops at its dependency gate on this host — `fastapi httpx mcp dotenv plotly pydantic streamlit uvicorn` are not installed and there is no `pip` to run `make test-python-deps` with. That gate is pre-existing and unrelated to this branch; CI runs the full suite.
- New tests, denial-first — a verifier that returns "clean" unconditionally passes a happy-path test and is worth nothing, so every difference the entrypoint relies on being caught has a test that fails if the detection is removed:
  - `test_verify_skills_provenance.py` — edited prose, edited script, added file, removed file, symlink swap, all-at-once reporting, bytecode the manifest never saw, a symlink **named** `__pycache__`, manifest parsing and the fail-closed cases, plus one that shells out to the Dockerfile's exact `sha256sum` invocation so the build and the reader cannot drift on format.
  - `test_image_startup_contract.py` — pins the Dockerfile and the entrypoint to each other: same manifest name, same three trees, `compileall` present and ordered before `sha256sum`, no `__pycache__`/`.pyc` carve-out on either side, the completeness and symlink assertions present, the `chown` present **and naming all six roots**, the sidecar stage's own `COPY` into `/opt/defaults/scripts` carrying `--chown=root:root`, a missing tree failing the build rather than being skipped, the boot check gated on the manifest rather than on the script, verification ordered before the PVC copy, fail-closed rather than `WARN`, and the venv interpreter rather than a bare `python3`.
- `make docs-check` — clean. `make docs-generate` leaves the tree clean, so no generated region is stale.
- The manifest-generation loop was run verbatim against the real `agents/platform/skills` (110 files) and `agents/cluster/skills` (11 files) trees outside Docker, and the boot step was exercised in all four of its states with the real script: clean tree → exit 0; a byte appended to one `SKILL.md` → exit 1 naming the file and both digests; manifest present with the verifier deleted → exit 1; no manifest present → skipped cleanly. A missing tree aborts the build loop with its message.

### Live validation

**Partially validated, in CI rather than on a cluster.** The `entrypoint-gate-test` stage above is a real execution of this code path — the built image, the real entrypoint, the runtime uid — which covers the build assertions, the manifests being written, and a clean verification at boot. What it cannot show is the behaviour on a live installation, and I could not reach one from this host: there is no `docker` binary here and the configured cluster's control plane (`172.16.0.2`) is a private endpoint with no route from this network. Nothing was changed on any cluster and no test artifacts were left behind.

Three observations remain outstanding on a real install:

1. **The tamper demonstration, on a pod rather than a fixture:** `kubectl exec` a single byte onto a file inside one of the image trees, restart, and confirm the pod refuses to start naming that file and both digests — then restore and confirm it boots clean. With `readOnlyRootFilesystem` now on the container, that write is expected to be refused as `EROFS` before ownership is even consulted, which is a second observation worth recording rather than a failed test.
2. **The barrier itself:** `ls -ld` inside the agent container on `/opt/hermes/skills`, `/opt/hermes/plugins`, `/opt/platform-template`, `/opt/cluster-template`, `/opt/chat-template` and `/opt/defaults/scripts` showing `root:root`, and the same on `/opt/defaults/scripts` in the `credential-proxy` sidecar, with a write attempt as the runtime uid being refused in each.
3. **That the manifest is harmless where it lands.** `sync_profile_skills` copies each tree wholesale, so a profile's `skills/` on the PVC carries a `skills_manifest.sha256` beside the skill directories. Nothing reads it there, but nothing in this change establishes that Hermes ignores a non-directory entry under `skills/` either — so the observation is a started pod whose skill roster is complete and whose logs carry no `Unknown skill(s)` warning.

I am happy to run these against an install I can reach, or to hand over the exact commands if it is easier to run them where the cluster is reachable.

## Self-Review

Reviewed at two stages, in different contexts.

**Commits 1–3, reviewed in a fresh session handed the head diff and nothing else** (not the session that wrote them), alongside @mplakhtiy's four review findings. Angles: whether the ownership barrier's coverage matched what the comments and docs claimed for it; whether the fail-closed paths were actually fail-closed; whether the manifest's exclusions could be turned into a hiding place; and whether every technical claim in a comment or doc sentence was true of the source it described. What that pass found, and the disposition of each:

- **The bytecode-exclusion rationale was factually wrong** (@mplakhtiy) — it claimed the interpreter revalidates a `.pyc` against its source. CPython's default is mtime plus size. **Fixed** by covering bytecode instead of excluding it, which meant moving the `compileall` pass ahead of the manifest loop.
- **`scan_tree` tested the exclusion list before it tested for a symlink** (@mplakhtiy), so a link named `__pycache__` was invisible to the boot check while the build refused it. **Fixed**, by deleting the exclusion mechanism rather than reordering it — an exclusion is a name a link can be given, so the ordering was a symptom. There is now a regression test for exactly that link.
- **The branch was pinned to Hermes base `v2026.8.3` while `main` had moved to `v2026.8.13`** — every green check, including the new symlink assertion against `/opt/hermes/skills`, had been built against a base image this would not merge onto. **Fixed** by merging `origin/main`; the assertion has now run against the base it will ship with.
- **`/opt/hermes/plugins` was left agent-owned** — in-process Python, a stronger capability than a `SKILL.md`, and in neither the barrier nor the docs' not-covered list. **Fixed** by adding it to the barrier.
- **`/opt/defaults/scripts` was left agent-owned**, holding the manifest checker itself. Commit `f4a730a9` had closed verifier *deletion* but not *replacement*. **Fixed** by adding it to the barrier.
- **The security reference justified leaving `/opt/defaults` agent-owned with "the running agent is a legitimate writer there"** — unsupported: `HERMES_HOME` is `/opt/data`, and every reference in the tree reads or copies *from* `/opt/defaults`. **Fixed**; the sentence now names what is actually left uncovered there and what closes it.
- **Reframing rather than a finding:** #674 merged in the meantime, so `readOnlyRootFilesystem` now applies to every container the operator builds and the live-exploit path for the two ownership findings is already `EROFS` in an operator-rendered pod. The fixes are kept — they hold outside the operator, under a relaxed `securityContext`, and at build time — but the comments, the docs and the threads now say that rather than implying an open hole.

**Docs-drift pass over commits 1–3**, run against the branch diff in the same fresh session: no Blocking or Advisory findings. Looked for stale generated regions (`docker-images.md` carries one; `make docs-generate` leaves the tree clean), a missing `docs/README.md` map row (no new doc files, so none is owed), duplicated facts that should be links, PR-status prose in docs, and — the drift magnet that mattered here — doc claims about baked image paths and ownership verified against the Dockerfile and the reading code rather than against other docs. The `/opt/defaults` read-back claim was checked at `agents/chat/defaults/plugins/bootstrap_onboarding/plugin.py:48` and the entrypoint's `--defaults-plugins` invocation, not taken from the page it replaced.

**Commit `f593ea7e`, reviewed by two subagents run concurrently** — one adversarial, one docs-drift — each handed the range `refs/kube-agents-base/main...HEAD` and its skill, and neither given the plan, the commit bodies or this PR body. Angles the adversarial pass ran: whether the barrier's coverage matched what the comments and docs claim for it; whether any deletion in the range removed, renamed, skipped or gated a test; whether `cp -a` and `shutil.copytree` still work as uid 10000 against a root-owned source; format hazards in the `sha256sum` round-trip; and the interaction with the other open pull requests touching the Dockerfile or the entrypoint. The docs-drift pass checked every doc claim about a baked path, an ownership, a line number and an anchor against source rather than against other docs. Merged dispositions:

- **`/opt/defaults/scripts` was root-owned in the agent image and still `hermes`-owned in the credential-proxy one.** The sidecar builds `FROM agent-base` and fills its own copy, so the barrier above never reached it — leaving the weaker copy in the container that holds the credentials and runs `credential_proxy.py`, `github_token_refresh.py` and `gke_endpoint.py` out of that directory. The stage comment claimed the two directories were identical in ownership, and the parity test compares source lists only. **Fixed** in `1da8c740`; a contract test now asserts the flag.
- **`/opt/chat-template/config.yaml` was agent-owned and named in neither coverage list** (both passes). Entrypoint step 2d seeds the default profile's `config.yaml` from it on a fresh or pristine-upstream volume and back-fills every absent key from it on every boot thereafter, so a key added to the image copy installs itself into the live config on the next restart and persists — over a file holding `toolsets`, `plugins.enabled` and `mcp_servers`. **Fixed** in `634132a3`: it is in the barrier. Every `CHAT_TEMPLATE_CONFIG` reference is a plain `cp` or a read to back-fill from, and a plain `cp` gives the destination to the copying uid, so the PVC copies stay writable.
- **The rest of `/opt/hermes` — `gateway/`, `tools/`, the venv interpreter — is not in the barrier either**, and its ownership comes from the Hermes base image rather than from anything this build asserts. I could not settle it: no `docker` on this host, so the built image could not be `stat`ed. **Documented as not covered** in both the Dockerfile comment and the security reference, so neither list reads as exhaustive; the code question stays open.
- **The opening `chown -R hermes:hermes` and the closing `chown -R root:root` named four of the same trees** in one `RUN`, so every build walked them twice for an ownership nothing in between observed. **Fixed**; the shipped image is byte-identical in ownership.
- **The security reference attributed `governance/` and `cron/` to both templates** — `/opt/cluster-template` has neither. **Fixed.**
- **`docker-images.md` claimed every other entrypoint step degrades with a `WARN`** — step 1 runs upstream's `stage2-hook.sh` and inherits `set -e`. **Fixed.**
- **The verifier's docstring claimed "no exclusions"** while the manifest's own name is excluded by name at any depth, so a plain file called `skills_manifest.sha256` planted in a subdirectory is invisible to the build and the check alike. **Fixed** in the docstring, which now states the exception and why a symlink so named is not one. The underlying placement — manifest inside the tree — was questioned too; **not changed**, because moving it to a sibling path redesigns what the review threads settled on and each consequence named is already handled and tested.
- **`make_log` is a copy of `profile_scaffold.make_log` with a docstring saying "shared".** **Fixed** by saying why it is duplicated instead: the verifier runs at step 1.55 with a standard-library-only dependency surface.
- **A Dockerfile comment credited a premise test with demonstrating CPython's invalidation mode "against this very image"** — it runs on the host interpreter over a temp directory. **Fixed.**
- **The manifest sits inside the tree it describes and four places said "beside".** **Fixed** in the prose. The entrypoint's FATAL string keeps "beside it", where "it" is the mismatching file rather than the tree.
- **The `/opt/defaults` residue parenthetical named its two weakest readers** — `cron/jobs.json` and the `plugins/` copytree are stronger. **Fixed.**
- **The security-reference bullet runs ~450 words among two-to-three-sentence siblings** (both passes). **Partly addressed:** it is shorter than before despite two additions, with the mechanics `docker-images.md` owns now a link rather than a restatement. Splitting it onto its own page is a judgment call I did not make unilaterally.
- **Deliberately not fixed.** `docs/testing-map.md:11` cites `Makefile:129-142` where the block now ends at 144: pre-existing, in a file this change does not touch, and correcting it here is unrelated churn — raised for a separate change rather than left unsaid. `docs/security-requirements.md:72`'s "immutable skill … versions" was flagged at low confidence and is the author's call. A suggested `docs/README.md` §2 row is not owed, since no document was added.

**What neither pass could reach:** no `docker` and no live install on this host, so nothing was verified inside a built image — the ownership of `/opt/hermes/gateway`, `/opt/hermes/tools` and the venv, whether Hermes' skill loader tolerates a non-directory entry at the top of a `skills/` tree, and the FATAL branch stopping a real container. Those are the same gaps **Live validation** records.

## Risk & Rollout

The new failure mode is the intended one and it is a hard one: a pod whose skill trees do not match their manifest will not start, and neither will one whose manifest is present with no verifier beside it. Everything that can produce the first state is caught earlier — the build asserts the manifest is complete, refuses symlinks, refuses a missing tree, and now compiles bytecode before it hashes it, so a correctly built image cannot present the verifier with a file the manifest never saw. The manifest ships in the same layer as the files it describes, so the two cannot skew across versions. The second state requires a file missing from the image, which is the report this change exists to make rather than a risk it introduces.

Reverting is a plain revert of these commits: the manifest becomes an unread file, the trees, plugins and shared scripts go back to being hermes-owned, and the entrypoint's guard is gone. The `compileall` pass would return to its own `RUN`. Nothing has to happen at merge time. The one PVC-side residue is cosmetic and self-healing: `sync_profile_skills` copies each tree wholesale, so a profile's `skills/` ends up carrying a `skills_manifest.sha256` alongside the skill directories, and the first boot after a revert replaces the tree and removes it. The file is never read from there — the boot check reads the image path.

---


deploy/docker/Dockerfile
```diff
@@ -1101,7 +1161,20 @@ RUN set -eu; \
...
+        [ -z "$(find "$d" -type l -not -path '*/__pycache__/*' -print -quit)" ] \
+            || { echo "$d contains a symlink; the manifest records -type f only and the boot check refuses any symlink it finds, so this would fail every pod: $(find "$d" -type l -not -path '*/__pycache__/*' | head -n 3)" >&2; exit 1; }; \
+        n="$(cd "$d" && find . -type f ! -name skills_manifest.sha256 -not -path '*/__pycache__/*' ! -name '*.pyc' | wc -l)"; \
+        ( cd "$d" && find . -type f ! -name skills_manifest.sha256 -not -path '*/__pycache__/*' ! -name '*.pyc' \
+            -exec sha256sum {} + | LC_ALL=C sort -k 2 > skills_manifest.sha256 ); \
+        [ "$(wc -l < "$d/skills_manifest.sha256")" -eq "$n" ] \
+            || { echo "manifest for $d has $(wc -l < "$d/skills_manifest.sha256") entries, expected $n" >&2; exit 1; }; \
+        chmod 444 "$d/skills_manifest.sha256"; \
+    done; \
+    chown -R root:root /opt/hermes/skills /opt/platform-template /opt/cluster-template
```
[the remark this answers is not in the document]
AntonTyb: Confirmed and fixed in `bee59149`.

I reproduced the mechanism outside the image: `shutil.copytree(src, dst, dirs_exist_ok=True)` over a tree containing a `0444` file succeeds the first time (leaving the destination `0444`) and raises `shutil.Error` with `[Errno 13] Permission denied` the second, which `overlay_template` does not catch for the item copies — so `governance`, `hindsight` and `_merge_after_overlay` are all downstream of `skills` in the platform items list and would be dropped behind the WARN, on every boot. Your first-boot ordering is right too: 2.5 creates the `profile.yaml` that gates 2.6, so this fires in the same boot on a fresh PVC.

The `chmod` bought nothing the `chown -R root:root` on the next line does not already give — inside the image the manifest is root-owned, and the PVC copy is never read; the boot check reads the image path. It now keeps the `0644` the normalization loops give every other file, and the Dockerfile comment states the rule so the next reader does not re-tighten it.

The regression test is behavioural rather than a substring check: `test_the_manifest_mode_lets_a_profile_be_overlaid_twice` in `deploy/docker/test_image_startup_contract.py` reads the mode out of the Dockerfile, applies it to a fixture tree and copies it twice. I verified it fails with exactly this `shutil.Error` when the mode is put back to `444`.

deploy/docker/Dockerfile
```diff
@@ -1101,7 +1171,20 @@ RUN set -eu; \
     find /opt/defaults/scripts -type f -exec chmod 755 {} + ; \
     for d in /opt/hermes/skills /opt/platform-template /opt/cluster-template; do \
         find "$d" -path '*/scripts/*' -type f -exec chmod 755 {} + ; \
-    done
+    done; \
+    for d in /opt/hermes/skills /opt/platform-template/skills /opt/cluster-template/skills; do \
+        [ -d "$d" ] \
+            || { echo "$d does not exist at manifest generation; the boot check refuses to start a pod whose skill tree has no manifest beside it, so shipping this image would crash-loop the fleet" >&2; exit 1; }; \
+        [ -z "$(find "$d" -type l -not -path '*/__pycache__/*' -print -quit)" ] \
```
mplakhtiy: **🔴 High — no CI run has built this branch against the base image now on `main`.**

This branch pins Hermes base `v2026.8.3`; `main` is on `v2026.8.13` (#718, merged `1e2df49c`, 2026-08-17 20:46 UTC). The last `Docker Build` run on this branch is from ~12:00 UTC the same day, so it built the `platform`, `entrypoint-gate-test` and `credential-proxy` targets against the *old* base. The workflow triggers on `pull_request` only (`.github/workflows/docker-build.yml:3-5`), so nothing re-runs without a push — every green check on this PR is green against a base image that is no longer the one it will merge onto.

The assertion at risk is this line's own. `/opt/hermes/skills` comes from the base image; if `v2026.8.13` ships a symlink anywhere under it — which nothing in this repository controls — the build goes red. Your Context section predicted exactly this ("whichever of the two merges second turns the build red") and called it not-a-blocker because both PRs were open. That reasoning expired when #718 merged first; this PR is now the second.

**Failure scenario.** Merged as is, the first `main` build after the merge fails on a path nobody was looking at — the same "green on the PR, red after merge" shape the layer-budget note in `AGENTS.md` exists to prevent.

Settled cheaply: merge `main` into the branch and push, then read the `build` check. That also gives the whole change a CI run on the tree that will actually merge.

deploy/docker/Dockerfile
```diff
@@ -1101,7 +1171,20 @@ RUN set -eu; \
...
+            || { echo "$d contains a symlink; the manifest records -type f only and the boot check refuses any symlink it finds, so this would fail every pod: $(find "$d" -type l -not -path '*/__pycache__/*' | head -n 3)" >&2; exit 1; }; \
+        n="$(cd "$d" && find . -type f ! -name skills_manifest.sha256 -not -path '*/__pycache__/*' ! -name '*.pyc' | wc -l)"; \
+        ( cd "$d" && find . -type f ! -name skills_manifest.sha256 -not -path '*/__pycache__/*' ! -name '*.pyc' \
+            -exec sha256sum {} + | LC_ALL=C sort -k 2 > skills_manifest.sha256 ); \
+        [ "$(wc -l < "$d/skills_manifest.sha256")" -eq "$n" ] \
+            || { echo "manifest for $d has $(wc -l < "$d/skills_manifest.sha256") entries, expected $n" >&2; exit 1; }; \
+        chmod 644 "$d/skills_manifest.sha256"; \
+    done; \
+    chown -R root:root /opt/hermes/skills /opt/platform-template /opt/cluster-template
```
mplakhtiy: **🟠 Medium — `/opt/hermes/plugins` is baked, in-process agent code, left uid-writable and unmanifested, and the security bullet does not say so.**

Line 1166 chowns `/opt/hermes/plugins` to `hermes`; this line re-chowns only `/opt/hermes/skills` and the two templates back to root. So `/opt/hermes/plugins` stays writable by uid 10000 and carries no manifest — while `Dockerfile:1158-1161` states that "`/opt/hermes/plugins` is the only path Hermes resolves them from", and `agents/platform/plugins/` and `agents/chat/plugins/` are COPYed there just above.

These are Python modules imported into the agent's own process, which is a strictly stronger capability than a `SKILL.md`. The precompiled `.pyc` does not protect them: a rewritten `.py` has a newer mtime, so the cache is invalidated and the modified source is what the next process — a kanban worker, an MCP child, the next gateway start — imports.

The new bullet in `security-and-iam.md:170` enumerates what is *not* covered with real care (PVC copies, `/opt/defaults`, plugin-delivered skills from an OCI image, root inside the pod). Baked plugin code is in neither list, so a reader comes away believing the image's agent-controlled code is covered.

**Failure scenario.** A prompt-injected agent that can no longer touch `SKILL.md` rewrites `/opt/hermes/plugins/<mod>/__init__.py` instead, gets the same result with more privilege, and the reference page says nothing that would have warned the reader.

The code decision may well be deliberate — but then say so. Either extend this `chown` (nothing in the repo writes to that path at runtime; the same trace that cleared the two templates clears this one) or add plugins to the explicit not-covered list. Silence is the finding.

deploy/docker/Dockerfile
```diff
@@ -1092,6 +1092,76 @@ COPY agents/chat/plugins/ /opt/hermes/plugins/
...
+# overlay of that profile reopens it for writing, raising shutil.Error out of
+# overlay_template and abandoning every item after `skills` — governance, hindsight
+# and the cron merge — behind a WARN. Ownership does not travel; mode does, and
+# read-only is the one mode this file must not have.
+#
+# Generated here rather than in a RUN of its own: `credential-proxy` builds FROM
+# this stage and the chain is bounded at 128 layers — see the layer-budget note at
+# the top of the file. __pycache__ and *.pyc are excluded because the compileall
+# pass below writes bytecode under /opt/hermes/skills after this line; the .py
+# they derive from is covered, and the interpreter revalidates a .pyc against it.
```
mplakhtiy: **🟠 Medium — bytecode is outside the manifest on a rationale this PR's own test disproves.**

> the .py they derive from is covered, and the interpreter revalidates a .pyc against it.

CPython's default invalidation is **timestamp-based** — mtime plus source size — not content-based. `compileall` only emits checked-hash pycs when asked (`--invalidation-mode checked-hash`, or implicitly under `SOURCE_DATE_EPOCH`), and the `RUN` below asks for neither. This repository already demonstrates the consequence, in the very file this PR extends: `deploy/docker/test_image_startup_contract.py:376-400` (`test_an_existing_cache_is_read_even_when_writes_are_suppressed`) rewrites a source keeping its size and mtime and asserts the **stale** `.pyc` is what runs. The same claim appears again at `verify_skills_provenance.py:43-52`.

**Failure scenario.** An altered image whose only modification is bytecode under a covered tree boots green and reports "verified N files", and the interpreter runs the altered bytecode — squarely inside the class the manifest claims to catch ("a corrupted layer, a bad build, a future code path that runs as root").

Exposure is PLAUSIBLE rather than confirmed: whether `/opt/hermes/skills` contains any `.py` at all is a base-image property I cannot read from here. If it contains none, the surface is empty today and the wrong comment is the whole finding.

The clean fix costs no layer: fold the `compileall` `RUN` below into this `RUN`, ahead of manifest generation, so bytecode is covered like everything else and the exclusion is unnecessary.

agents/platform/scripts/verify_skills_provenance.py
```diff
@@ -0,0 +1,206 @@
...
+    found: Dict[str, Path] = {}
+    symlinks: List[str] = []
+    for path in directory.rglob("*"):
+        relative = path.relative_to(directory)
+        if is_excluded(relative):
+            continue
+        # Checked before is_dir()/is_file(), both of which resolve the link.
+        if path.is_symlink():
+            symlinks.append(relative.as_posix())
+            continue
```
mplakhtiy: **🟠 Medium (same finding as the Dockerfile bytecode comment) — `is_excluded()` is tested before `is_symlink()`, so a symlink named `__pycache__` is invisible to the verifier.**

The build refuses a symlink outright (`Dockerfile:1178-1179`), but this ordering means the boot check never sees one that happens to be named `__pycache__` or to end in `.pyc`. That is precisely the asymmetry the Dockerfile comment at 1161-1162 says must not exist:

> Either the two sides agree that a symlink is an anomaly or the check is defeated by one.

The docstring above even explains why symlinks must not be followed — the reasoning is right, the statement order defeats it for the excluded names.

**Failure scenario.** A tree carrying a symlink named `__pycache__` (or `x.pyc`) pointing outside the tree verifies clean at boot. The build would have caught it; the runtime check, which is the half that runs on a pod nobody rebuilt, does not.

Swapping the `is_symlink()` test ahead of `is_excluded()` is a two-line move and costs nothing — an excluded *real* directory still short-circuits one line later.

mplakhtiy (commented):
Adversarial review pass against `bee59149` merged onto `main` @ `76a074b8` (merged cleanly — no conflicts, including in `Dockerfile` and `docker-entrypoint.sh`, despite #720 being open). Medium-and-above findings only, anchored inline; two Lows are held back.

What I verified rather than assumed: the verifier does what it claims — I exercised it against a fixture tree and it detects content edits, added and removed files and symlink swaps, and correctly ignores bytecode; the 32 contract tests and 19 verifier tests pass on the merged tree; the build-time loop's failure branches are all reachable and all fail closed; and the bot's earlier `chmod 444` finding is genuinely fixed, with a behavioural regression test rather than a substring check. No `docker` binary on this host, so the build-time loop and the `chown -R root:root` were reproduced outside a container rather than executed in one.

- 🔴 **High — the PR body has no Self-Review section.** The template's section list is Summary / Why This Change / What Changed / Context / Testing / Live validation / **Self-Review** / Risk & Rollout (`.github/PULL_REQUEST_TEMPLATE.md:59`); the body has every one of those except Self-Review, and `AGENTS.md` makes filling it in a required pre-PR step. Everything it would have recorded may well have been done — the body is unusually thorough — but the section a reviewer is instructed to read *before* the diff is absent, so every subsequent reader loses the chance to start where you stopped. It is also the only record of findings consciously not fixed, and this change has at least two of those (the `/opt/defaults` exemption and the bytecode exclusion) argued in code comments but never surfaced as dispositions.
- 🟠 **Medium — the Live validation section still says the PR is a draft, and lists three unperformed observations.** It reads "**That is why this is still a draft.** Two things remain to be observed on a real install before it is marked ready", then lists three. The PR is not a draft — it was marked ready, which is what triggered the bot review — and none of the three appear to have happened. `AGENTS.md` accepts "Not live-tested" with a reason, and the reason given (no `docker` binary, private control-plane endpoint) is a real one; what it does not accept is a section that contradicts the PR's own state.

  The third item is the one that matters most and is least covered by anything here: nothing in the repository establishes that Hermes tolerates a non-directory entry under `skills/`. Every entry in `agents/platform/skills/` and `agents/cluster/skills/` is a directory today, and this change puts `skills_manifest.sha256` beside them in the image **and**, via `cp -a` and `copytree`, in every profile's `skills/` on the PVC. If the loader warns, `Unknown skill(s)` makes a worker exit 1 — the failure mode `sync_profile_skills`' own comment describes. Only a running install answers that one.

deploy/docker/Dockerfile
AntonTyb: Confirmed and fixed. `origin/main` is merged into the branch (`44c2ec01`) and the fixes are pushed as `f593ea7e`, so `build` has now run on the tree that will merge.

Green against the base you named — the job's env shows `HERMES_AGENT_TAG: v2026.8.13@sha256:68e15ae2a6d894d0ccbd9f8aacbbe13d4d28fa5dc9b6a303970b67bb2499b1a6` — and it settles the assertion you flagged: `v2026.8.13` ships no symlink under `/opt/hermes/skills`, because the loop on this line would have aborted the build with the offending paths printed.

Three things from that run, since they are the evidence for the other threads too:

```
OK: platform-agent:latest has 113 layers, 7 under the 120 budget
```

```
#88 [entrypoint-gate-test 2/2] RUN /usr/local/bin/entrypoint-gate-check
   ...
   all 17 checks passed
```

and the same 17 again from the second run under `docker run --read-only --tmpfs /tmp`. Every one of those executions is the real entrypoint as uid 10000, and each one passes through step 1.55, so all three trees verified clean against manifests that now include their bytecode.

Your point about the reasoning expiring is the part I want to keep: the Context section said "whichever of the two merges second turns the build red" and then kept treating it as hypothetical after #718 landed. The section is rewritten to say #718 and #674 are merged and carried, rather than leaving a prediction in place that has already resolved.

deploy/docker/Dockerfile
AntonTyb: Confirmed, and the code decision was not deliberate — it was an oversight. Extended rather than documented as a gap, in `f593ea7e`. The line now reads:

```
chown -R root:root /opt/hermes/skills /opt/hermes/plugins /opt/platform-template /opt/cluster-template /opt/defaults/scripts
```

`/opt/defaults/scripts` is in there for the same reason and was the second half of the same hole: it holds the shared scripts the entrypoint invokes by absolute path, the `sitecustomize.py` the operator puts on `PYTHONPATH`, and `verify_skills_provenance.py` itself. `f4a730a9` had closed verifier *deletion* by gating on the manifest, but not verifier *replacement* — a checker the checked party can rewrite reports whatever it is told to.

The trace you suggested clears both: nothing in the repository writes to either path at runtime. Every reference to `/opt/hermes/plugins` resolves to the image copy being read; the entrypoint's plugin work targets `profiles/<p>/plugins/` on the PVC and `/opt/defaults/plugins` (`docker-entrypoint.sh:1227-1246`), never `/opt/hermes/plugins`. Every `/opt/defaults/scripts` reference is a read or a `cp` out into `$HERMES_HOME`.

`security-and-iam.md` now names both in what the barrier covers and says plainly that personas, plugins and shared scripts get the barrier and **not** the hash — covered against the agent, not against a corrupted layer. `test_the_barrier_covers_the_personas_and_not_just_the_skills` requires all five roots, so narrowing the list fails a test rather than quietly reopening this.

One correction to the severity, in your favour and mine: #674 merged on 2026-08-25, so `hardenedSecurityContext()` now sets `readOnlyRootFilesystem: true` on every container the operator builds and the rewrite in your failure scenario is `EROFS` before ownership is consulted. That does not retire the finding — the barrier still holds for the image run outside this operator, for a container given a writable rootfs, and at build time — but the fix lands as defence in depth rather than as closing a live hole, and the Dockerfile comment and the docs now say that rather than implying the ownership line is the only thing standing there.

deploy/docker/Dockerfile
AntonTyb: You are right about CPython and the comment was wrong. Fixed the way you suggested, in `f593ea7e`: the `compileall` `RUN` is folded into this one, positioned after the `chmod` loops (so its `__pycache__` is not swept up by the `*/scripts/*` `chmod 755`) and before the manifest loop, and both sides stop excluding `__pycache__` and `*.pyc`. `EXCLUDED_DIRS`, `EXCLUDED_SUFFIXES` and `is_excluded()` are deleted from `verify_skills_provenance.py` rather than adjusted.

It costs less than no layer — it retires one. `agent-base -> platform` is the chain the budget measures, and the run reports `OK: platform-agent:latest has 113 layers, 7 under the 120 budget`.

The corrected comment now states the actual invalidation rule and cites the test you pointed at:

> CPython's default invalidation is timestamp-based — source mtime plus size — not a content hash, and nothing here passes `--invalidation-mode checked-hash` or sets `SOURCE_DATE_EPOCH`. A .pyc is therefore an independent artifact rather than a derivative of the .py beside it: rewrite one while keeping the source's size and mtime and the interpreter runs the rewritten bytecode, which `test_an_existing_cache_is_read_even_when_writes_are_suppressed` in `deploy/docker/test_image_startup_contract.py` demonstrates against this very image.

On the PLAUSIBLE caveat: I did not establish whether the base's `/opt/hermes/skills` contains any `.py` today, and the fix makes it stop mattering — nothing is carved out, so the surface being empty or not no longer changes what the manifest covers. Ordering is now pinned by `test_bytecode_is_compiled_before_it_is_hashed` (asserts `compileall` precedes `sha256sum` in the same instruction) and `test_bytecode_is_not_carved_out_of_the_manifest` (asserts neither `__pycache__` nor `.pyc` appears in the block and that the three verifier symbols are gone). The green `entrypoint-gate-test` — 17 checks in-image and 17 again under `docker run --read-only`, each a real entrypoint run as uid 10000 — is what shows the new ordering verifies clean rather than failing every boot.

agents/platform/scripts/verify_skills_provenance.py
AntonTyb: Confirmed. Fixed in `f593ea7e`, by deleting the exclusion mechanism rather than reordering the two tests.

The swap you describe would have worked, but the ordering was the symptom: an exclusion list is a list of names an attacker gets to choose, and any branch that runs before `is_symlink()` hands them one. With `__pycache__` and `*.pyc` now inside the manifest (the sibling bytecode thread), there is nothing left to skip by name, so the symlink test is the first thing every entry meets and there is no earlier branch for a link to hide behind. The Dockerfile's symlink probe lost its `-not -path '*/__pycache__/*'` carve-out for the same reason, so the build and the boot check agree on every path again.

The docstring now says so where the reasoning already was, and there is a regression test for exactly your case — `test_a_symlink_named_pycache_is_reported_rather_than_skipped` builds a tree whose `scripts/__pycache__` is a symlink to a directory outside the tree and asserts the verifier reports `symlink the build did not produce`. It fails if the name-first branch comes back.

mplakhtiy (approved):

Issue #618, opened 2026-08-11
Epic: Zero-Friction Installation & First Hour Autonomous Value Framework


fkc1e100:
## 🎯 Overview & Strategic Objectives

Getting started with autonomous Kubernetes management should take minutes, not days. The **Zero-Friction Installation & "First Hour" Autonomous Value Framework** eliminates setup complexity, automates Google Cloud Platform (GCP) credential and cluster discovery, accelerates initial onboarding scans, and delivers immediate actionable value within 60 minutes of installation.

By combining one-line deployment, multi-project fleet discovery, onboarding resilience, and private self-hosted Gemma model serving on GKE GPUs, this framework enables enterprise SRE and platform teams to deploy `kube-agents` across connected, air-gapped, and sovereign environments.

---

## 🗺️ Multi-Phase Implementation Roadmap

- [x] **Phase 1 (PR #519 - MERGED)**: `feat(installer): add zero-friction one-liner install script with Cloud Shell auto-discovery`
  • Interactive and unattended single-command installer script (`install-kube-agents`).
  • Automatic Google Cloud Platform (GCP) Cloud Shell credential, project ID, and GKE cluster auto-discovery.
  • Self-contained Helm profile deployment and IAM role binding automation.

- [ ] **Phase 2 (PR #573 - P0 / In Review)**: `fix(operator): improve onboarding scan speed and runtime resilience`
  • Accelerates initial post-installation discovery and scan speed.
  • Implements LiteLLM exponential backoff, credential proxying, and management cluster health isolation to prevent first-boot timeouts.

- [ ] **Phase 3 (PR #588 - P1 / In Review)**: `feat(installer): allow adding additional GCP projects to monitor and remediate`
  • Extends the installer suite and Day-2 control panel to onboard additional Google Cloud Platform (GCP) projects.
  • Enables multi-project fleet discovery, IAM role bindings, and read-only telemetry monitoring across multi-tenant environments.

- [ ] **Phase 4 (PR #608 - P1 / In Review)**: `feat(inference): support self-hosted Gemma 4 via vLLM on GKE in install/teardown pipeline`
  • Integrates private, self-hosted Gemma 4 31B (`google/gemma-4-31B-it`) and Small Language Model (SLM) serving via vLLM on GKE GPU accelerator node pools directly into `provision.sh` / `teardown.sh`.
  • Eliminates external API keys and prompt egress, enabling 100% in-cluster autonomous fleet audits.
[...]

PR #732, opened 2026-08-17
feat(release): implement release automation toolchain (scope 2)

eLeontev:
# Pull Request

> This pull request was generated with the help of AI.

## Why This Change

Implement the automated release toolchain (Scope 2) to compute pure SemVer 2.0 versions from Conventional Commits, enforce live RC validation gates before artifact publication, detect tag collisions, provide idempotent release re-runs, and enable zero-rebuild container image promotion via `docker buildx imagetools create`.

## What Changed

**Files:**
- `scripts/release/common.sh`
- `scripts/release/calculate_next_version.sh`
- `scripts/release/verify_release_eligibility.sh`
- `scripts/release/README.md`
- `tests/testing/common.py`
- `tests/testing/release.py`
- `tests/test_release_common.py`
- `tests/test_calculate_next_version.py`
- `tests/test_verify_release_eligibility.py`
- `tests/test_install_script.py`
- `tests/test_upgrade_script.py`
- `tests/test_provision_rc_environment.py`

**Added/Changed/Fixed:**
- **Release Automation Helper Library (`scripts/release/common.sh`)**:
  - Declarative `REQUIRED_RELEASE_IMAGES` registry (all 4 release container images: `k8s-operator`, `platform-agent`, `credential-proxy`, `replay-proxy`).
  - `promote_release_images` function with strict argument validation and duplicate destination tag overwrite protection for clean OCI manifest copying without rebuilding (`docker buildx imagetools create`).
  - `ensure_git_tag` with 40-character SHA normalization for idempotent tagging.
- **SemVer 2.0 Calculation Engine (`scripts/release/calculate_next_version.sh`)**:
  - Automatically parses Conventional Commits from latest GA tag (`X.Y.Z`) to HEAD or specified target ref.
  - Implements SemVer 2.0 Clause 4 for `0.y.z`: breaking changes (`BREAKING CHANGE:`, `!`) bump `MINOR`, standard `feat:` bumps `MINOR`, fixes/chores bump `PATCH`.
  - Strict input validation and fail-fast checks for baseline tag and target ref existence in git history.
  - Integrates with `$GITHUB_OUTPUT` (`version`, `previous_version`, `has_changes`, `bump_type`).
- **Release Eligibility Gatekeeper (`scripts/release/verify_release_eligibility.sh`)**:
  - Enforces explicit `TARGET_TAG` and `TARGET_COMMIT` parameters, eliminating ambient `GITHUB_REF_NAME` fallbacks.
  - Auto-resolves latest `rc_*_validated` commit when commit is omitted.
  - Evaluates safe idempotent skips (`already_released=true`, exit 0) and collision detection against already-tagged releases prior to any emergency overrides.
  - Fail-fast validation ensuring explicit `TARGET_COMMIT` resolves to a valid Git commit SHA, preventing unresolvable strings from bypassing collision/idempotency checks.
  - Verification of all 4 candidate container images in registry before release approval.
  - Emergency override (`SKIP_RC_VALIDATION=true`) requiring mandatory non-whitespace `EMERGENCY_OVERRIDE_REASON`.
  - Outputs canonical `release_commit` 40-character SHA for downstream publishing steps.
- **Hermetic Test Suite & Fixtures (`tests/testing/`, `tests/test_*.py`)**:
  - Reusable test fixture package in `tests/testing/` with centralized `VALID_GA_RELEASE_TAGS`, `INVALID_GA_RELEASE_TAGS`, `VALID_IMMUTABLE_REFS`, `INVALID_IMMUTABLE_REFS`.
  - Hermetic environment isolation (`get_isolated_test_env`) stripping ambient CI runner variables across all tests.
  - 35 dedicated unit test methods across `test_release_common.py`, `test_calculate_next_version.py`, and `test_verify_release_eligibility.py` (167 total unit tests passing).

## Why This Matters

- Guarantees that only commits passing live RC E2E validation can be published to production.
- Prevents container image rebuild discrepancies between RC validation and GA release by promoting verified SHA tags.
- Automates SemVer 2.0 calculations, eliminating human error in version tagging.

## Context

- Part of Scope 2 in the release engineering roadmap (Issue #618).
- Clean promotion in Scope 3 (`release-publish.yml`) will supersede and decommission tag-triggered image rebuilds in `docker-publish-*.yml`.

## Self-Review

- **What was looked for:**
  - Adversarial review across the 10 quality angles (`review-adversarial` / `pr-review-batch.md`).
  - Strict ordering of collision and idempotency checks before emergency bypass paths.
  - Strict resolution of `TARGET_COMMIT` to prevent malformed strings bypassing collision checks.
  - Hermetic isolation of unit test subprocess environments from CI runner environment variables (`GITHUB_*`, `CI`, `RUNNER_*`).
  - Strict numeric SemVer compliance without `v` prefix.
  - Verification of error handling and fail-fast CLI status codes.
- **Findings and disposition:**
  - *Finding 1 (Order of checks in gatekeeper)*: Moved Step 3 (idempotent skip & collision checks) before Step 4 (emergency override) in `verify_release_eligibility.sh` so that colliding releases are blocked even under emergency override. (Resolved in 4aaaaf87)
  - *Finding 2 (Test runner isolation)*: Added `get_isolated_test_env` in `tests/testing/common.py` and updated test runners to strip ambient CI runner variables. (Resolved in 0354428c & 5a426cce)
  - *Finding 3 (Temporary directory cleanup on Linux)*: Enabled `ignore_cleanup_errors=True` on `tempfile.TemporaryDirectory` for reliable `.git` cleanup in Python 3.13. (Resolved in 0354428c)
  - *Finding 4 (Release scripts documentation)*: Updated `scripts/release/README.md` to document all 4 required container images and new Scope 2 scripts. (Resolved in 912cb3f3)
  - *Finding 5 (Gatekeeper parameter validation & explicit invocation)*: Removed implicit `GITHUB_REF_NAME`/`GITHUB_SHA` fallbacks, added strict SemVer format validation, and unified output to `release_commit`. (Resolved in d20f09e8)
  - *Finding 6 (calculate_next_version error handling)*: Added strict SemVer format validation and Git existence checks for baseline tag and target ref; prevented suppressed `git log` errors. (Resolved in 80720a94)
  - *Finding 7 (promote_release_images safety)*: Added input validation and duplicate tag overwrite protection in `common.sh`. (Resolved in 9a5f2d98)
  - *Finding 8 (Test fixtures centralization)*: Relocated `installer_test_data.py` fixtures into `tests/testing/common.py` and migrated installer tests to `get_isolated_test_env`. (Resolved in 6120182d)
  - *Finding 9 (Unresolvable TARGET_COMMIT fail-fast)*: Replaced fallback assignment of unresolvable `TARGET_COMMIT` in `verify_release_eligibility.sh` with fail-fast error (`exit 1`) and added regression tests for invalid commits in standard and emergency modes. (Resolved in cd1afc57)

## Testing

- `python3 -m unittest discover tests`: All 167 tests passing (`OK`).
- `make docs-check`: Passed with 0 errors across 247 Markdown files.
- `make validate` & `make prompt-check`: Passed.
- `(cd k8s-operator && go test ./...)`: Passed.

### Live validation

- **Not live-tested:** This PR introduces release automation toolchain scripts, calculation engines, and unit test suites. Full live multi-cluster image promotion and release execution runs are tested in Scope 3 (`release-publish.yml`) and live cluster RC execution in Scope 4.
- Exercised all release scripts under hermetic subprocess test harnesses covering 35 test scenarios (SemVer Clause 4 calculation, idempotent skip, collision detection, emergency audit override, image presence verification, overwrite protection, unresolvable commit fail-fast).

## Risk & Rollout

- **Risk:** Low; introduces standalone release tooling scripts and unit tests without modifying operator runtime code or existing deployment manifests.
- **Rollout:** Toolchain will be invoked by `.github/workflows/release-publish.yml` in Scope 3.


scripts/release/verify_release_eligibility.sh
```diff
@@ -0,0 +1,153 @@
...
+
+  echo "⚠️ WARNING: RC E2E validation check is explicitly bypassed via emergency override!" >&2
+  echo "⚠️ Reason: ${EMERGENCY_REASON}" >&2
+  if [ -n "${GITHUB_OUTPUT:-}" ]; then
+    echo "eligible=true" >> "${GITHUB_OUTPUT}"
+    echo "emergency_override=true" >> "${GITHUB_OUTPUT}"
+    echo "target_commit=${RESOLVED_COMMIT}" >> "${GITHUB_OUTPUT}"
+  fi
+  exit 0
+fi
```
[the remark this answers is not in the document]
eLeontev: evaluate idempotency and collision checks before emergency override

scripts/release/calculate_next_version.sh
```diff
@@ -0,0 +1,103 @@
...
+if [ -n "${BASE_TAG_PARAM}" ]; then
+  LATEST_TAG="${BASE_TAG_PARAM}"
+else
+  LATEST_TAG="$(git tag -l --sort=version:refname '[0-9]*' 2>/dev/null | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | tail -n 1 || echo "")"
+fi
+
+if [ -z "${LATEST_TAG}" ]; then
+  echo "ℹ️ No previous GA SemVer tag found. Initializing repository at baseline 0.1.0." >&2
+  NEXT_VERSION="0.1.0"
+  if [ -n "${GITHUB_OUTPUT:-}" ]; then
```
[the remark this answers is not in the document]
eLeontev: not relevant - the project did not publish any `vX.Y.Z` release and is not going to support that format.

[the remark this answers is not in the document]
eLeontev: The fix of the doc is located in #688

tests/test_verify_release_eligibility.py
```diff
@@ -0,0 +1,270 @@
...
+        commit_sha = git("rev-parse", "HEAD").stdout.strip()
+
+        return temp_dir, repo_dir, git, commit_sha, bin_dir
+
+    def _run_verify_script(self, repo_dir, args=None, env=None, bin_dir=None):
+        full_env = dict(os.environ)
+        if bin_dir:
+            full_env["PATH"] = f"{bin_dir}:{full_env.get('PATH', '')}"
+        full_env.update(env or {})
+        return subprocess.run(
```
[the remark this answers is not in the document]
eLeontev: addressed

scripts/release/verify_release_eligibility.sh
```diff
@@ -0,0 +1,153 @@
...
+# and performs an idempotent skip if the commit has already been released under the EXACT SAME tag.
+# Releases strictly use pure numeric SemVer without 'v' prefix (e.g. 0.1.0, 0.2.0).
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=scripts/release/common.sh
+source "${SCRIPT_DIR}/common.sh"
+
+TARGET_TAG="${1:-${GITHUB_REF_NAME:-}}"
+TARGET_COMMIT_PARAM="${2:-${GITHUB_SHA:-}}"
```
mplakhtiy: **🔴 High — the gatekeeper always skips in the invocation its own defaults describe.**

These two defaults are precisely the environment of a tag-push trigger, and a tag push is the flow this repository documents: `docs/designs/semver-deployment-versioning.md` §2.3 and §4 say "A human then tags the validated commit `MAJOR.MINOR.PATCH`, which triggers the image and chart publication workflows", and the site's `deploy/release-versioning.md:14` repeats it. In that mode the target tag already points at the target commit before the script runs, so step 3's Scenario A matches on the first iteration and returns the idempotent skip.

**Failure scenario.** A release cut the documented way — push tag `0.2.0`, workflow runs the gatekeeper — exits 0 having written `eligible=false`, `already_released=true`, `skip_release=true`. Every downstream step is skipped, the release never publishes, and the run is green. The RC validation check at step 5, which is the entire point of the script, is never reached — so the gate would also have said nothing if the commit were unvalidated.

Reproduced against a throwaway repository with a mocked `docker`:

```
$ GITHUB_REF_NAME=0.2.0 GITHUB_SHA=$SHA bash verify_release_eligibility.sh
🔍 VERIFYING RELEASE ELIGIBILITY FOR: 0.2.0
ℹ️ IDEMPOTENT SKIP: Release 0.2.0 for commit 7b7675e… is already published.
exit=0
```

Latent today — no workflow calls the script yet. Scope 3 may intend to call it with explicit arguments *before* creating the tag, in which case these defaults are dead code describing the wrong contract. Either way they should go, or Scenario A needs a way to tell "this tag existed before this run" from "this run's own trigger created it" — `git tag --points-at` cannot.

scripts/release/common.sh
```diff
@@ -213,3 +226,25 @@ ensure_git_tag() {
...
+  local registry_prefix
+  registry_prefix="$(get_registry_prefix)"
+
+  echo "🚀 Promoting verified container images (${commit_sha:0:7}) -> (${target_tag})..."
+
+  for img in "${REQUIRED_RELEASE_IMAGES[@]}"; do
+    local source_image="${registry_prefix}/${img}:${commit_sha}"
+    echo "  • Promoting ${img}..."
+    if command -v docker >/dev/null 2>&1; then
+      docker buildx imagetools create --tag "${registry_prefix}/${img}:${target_tag}" "${source_image}"
```
mplakhtiy: **🟠 Medium — image promotion races two workflows that rebuild the same tags.** *(This is the bot's finding on `common.sh:246`, still unresolved and unanswered — restated because it is the reason the body's "Why This Matters" claim does not hold.)*

`promote_release_images` writes the RC-validated digests to `${registry_prefix}/${img}:${target_tag}`. Cutting a GA release means pushing the git tag, and two workflows already on `main` fire on exactly that push and rebuild the same four images under the same tag:

- `.github/workflows/docker-publish-ghcr.yml:6` — `tags: ["[0-9]*.[0-9]*.[0-9]*"]`, publishing `platform-agent` (`:58`), `replay-proxy` (`:71`) and `credential-proxy` (`:85`) as `ghcr.io/…/<image>:${{ github.ref_name }}`.
- `.github/workflows/docker-publish-k8s-operator.yml:6-7` — same trigger, `k8s-operator:${{ github.ref_name }}` at `:60`.

Nothing orders the promotion against those two, and registry tags are mutable, so the last writer defines what `:0.2.0` resolves to. Ordering promote-before-tag inside a future `release-publish.yml` does not fix it: the rebuilds are separate workflows started by the tag push itself. The two producers do not even build the same artifact — `docker-publish-k8s-operator.yml:62` bakes `VERSION=${{ github.ref_name }}` on a tag push, while the digest being promoted was built from a SHA push with `VERSION=latest` — so which writer wins is observable.

Two smaller points on the same call: the promotion source is the mutable tag `:${commit_sha}` rather than a pinned digest, so a re-run of the SHA publish between validation and promotion changes what gets promoted; and `docker buildx imagetools create` repoints an existing tag without complaint, so a reused version number silently rewrites a published release rather than failing.

**Failure scenario.** `0.2.0` is cut; the tag-triggered rebuild finishes after the promotion; the published `:0.2.0` images are a fresh build that never went through RC E2E validation — the exact discrepancy the PR body says this mechanism prevents.

This is a design decision rather than a patch. Either the tag-triggered rebuilds go away as part of the release design (say so, and file it against Scope 3), or promotion is not the mechanism that gets the validated digest into `:X.Y.Z`.

scripts/release/calculate_next_version.sh
```diff
@@ -0,0 +1,103 @@
...
+source "${SCRIPT_DIR}/common.sh"
+
+BASE_TAG_PARAM="${1:-}"
+TARGET_REF_PARAM="${2:-HEAD}"
+
+# 0. Protection against Shallow Checkout in CI
+if [ "$(git rev-parse --is-shallow-repository 2>/dev/null || echo "false")" = "true" ]; then
+  echo "ℹ️ Shallow repository detected. Unshallowing git history for version calculation..." >&2
+  git fetch --unshallow --tags >/dev/null 2>&1 || git fetch --depth=100 --tags >/dev/null 2>&1 || true
+fi
```
mplakhtiy, line 18: **🟠 Medium — this script reports failure as success.**

Every way the baseline resolution can go wrong produces an empty string, and an empty string means "no new commits" (`COMMITS_SUBJECTS="$(git log … 2>/dev/null || echo "")"` at `:39-40`). Two concrete outcomes:

- **An unresolvable baseline is silently accepted.** In a repository holding neither tag:

  ```
  $ calculate_next_version.sh 0.9.9   →  📌 Latest GA Tag: 0.9.9 / ℹ️ No new commits … / 0.9.9   (exit 0)
  $ calculate_next_version.sh v0.1.0  →  📌 Latest GA Tag: v0.1.0 / ℹ️ No new commits … / v0.1.0 (exit 0)
  ```

  `$1` is never validated against `^[0-9]+\.[0-9]+\.[0-9]+$`, so a typo or a `v`-prefixed argument emits `version=<the bad string>` and `has_changes=false` instead of failing.

- **A failed tag fetch resets the version line.** `actions/checkout` defaults to `fetch-depth: 1` and no tags. The shallow guard on these lines tries to repair that, but both attempts end in `|| true`; if they fail (fork without credentials, transient network), `git tag -l` is empty, the script takes the `[ -z "${LATEST_TAG}" ]` branch at `:25` and emits `version=0.1.0`, `previous_version=`, `bump_type=initial` for a repository already several releases along — exit 0, no warning.

The first is confirmed by the reproduction above; the second is plausible — the mechanism is certain, but whether the fetch fails depends on the Scope 3 workflow's checkout, which does not exist yet. Both take the same fix: fail loudly when the baseline cannot be resolved or the commit range cannot be read.

scripts/release/verify_release_eligibility.sh
```diff
@@ -0,0 +1,153 @@
...
+TARGET_REPO="$(get_target_repo)"
+SKIP_VALIDATION="${SKIP_RC_VALIDATION:-false}"
+EMERGENCY_REASON="${EMERGENCY_OVERRIDE_REASON:-}"
+
+if [ -z "${TARGET_TAG}" ]; then
```
mplakhtiy, line 17: **🟠 Medium — `TARGET_TAG` is never validated against the convention this script enforces on everything else.**

The header at `:3` states "Releases strictly use pure numeric SemVer without 'v' prefix", and line 71 filters *existing* tags through `grep -E '^[0-9]+\.[0-9]+\.[0-9]+$'`. The script's own target tag gets only an emptiness check.

**Failure scenario.** On a commit carrying an `rc_*_validated` tag and no GA tag:

```
$ verify_release_eligibility.sh main $SHA
eligible=true
validated_rc_tag=rc_2608180000_x_validated
```

Combined with the `GITHUB_REF_NAME` default flagged above, a run on a branch declares `main` an eligible release version, and Scope 3's `promote_release_images` would then create `ghcr.io/…/<image>:main` — a mutable branch-shaped tag in the release namespace. A `v`-prefixed tag behaves the same way on a clean commit, and additionally defeats collision detection, since the collision grep only ever sees bare-numeric tags.

One regex next to the emptiness check closes it.

mplakhtiy (commented):
Adversarial review pass against `fbf9fa92` merged onto `main` @ `1d3970b2` (merged cleanly, no conflicts), run with Claude via this repository's `review-adversarial` skill. Medium-and-above findings only, anchored inline; six Lows are held back.

Verified rather than assumed: `shellcheck -x` is silent on all three shell files, the 26 new tests pass, and the scripts were exercised against throwaway git repositories with a mocked `docker` — Findings 1, 3 and 4 below are reproductions, not readings. `promote_release_images` was not run; that would need a real `docker buildx imagetools create` against a scratch tag.

**Previously raised, and settled — recorded so nobody re-litigates them**

- The bot's High on bare-numeric vs `vX.Y.Z` tags is **correctly refuted** against current `main`: `docker-publish-ghcr.yml:6`, `docker-publish-k8s-operator.yml:7` and `chart-release.yml:6` all trigger on `"[0-9]*.[0-9]*.[0-9]*"`. Your rebuttal holds — #688 landed, and the design doc is consistently bare-numeric throughout.
- The bot's High on test hermeticity is **fixed**: `get_isolated_test_env` (`tests/testing/common.py:33-46`) drops `GITHUB_*`, `RUNNER_*`, `CI` and `CONTINUOUS_INTEGRATION`, so no fixture touches the upstream repository. Confirmed by running the suites with `CI` set in the parent shell.
- The bot's Medium on override ordering is **fixed** in `4aaaaf87`: step 3 now precedes step 4, and both new tests fail if the blocks are swapped back.

- 🟠 **Medium — the PR body claims more than the diff contains.** The Files list includes `tests/testing/__init__.py`, which is not in the diff (`git ls-files tests/testing` returns `common.py` and `release.py`; imports work anyway via namespace packages, so this is a description error rather than a defect). "29 dedicated unit tests" is 26 — `python3 -m unittest` over the three new modules reports `Ran 26 tests` (5 + 9 + 12); the nine `subTest` cases inside `test_is_truthy` are presumably the extra three, but they are not tests. And "Prevents container image rebuild discrepancies … by promoting exact SHA digests" is not achievable as the surrounding workflows stand (see the `common.sh` comment), and it promotes a mutable tag, not a digest. `AGENTS.md` makes body accuracy a reviewer-facing contract, which is why these are one finding rather than three nits.

**A note on siblings, not a finding.** #620, #621, #623 and #548 are a parallel release-automation track — `release-build-publish.yml`, `scripts/release/build_local_release.sh`, and edits to the same two docs pages. This PR's Scope 3 plans `release-publish.yml`. Two release pipelines are being proposed at once and nothing in either says which one wins; worth settling before Scope 3 rather than after. Separately, #720 also edits `scripts/release/README.md` — adjacent hunks, probably auto-mergeable, but whoever merges second should re-read the file.

scripts/release/common.sh
```diff
@@ -213,3 +226,25 @@ ensure_git_tag() {
...
+    else
+      echo "❌ ERROR: 'docker buildx' CLI is required for image promotion!" >&2
+      return 1
```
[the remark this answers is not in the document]
eLeontev: I  added overwrite protection to promote_release_images in this PR, while removing the concurrent on: push: tags rebuild triggers from docker-publish-*.yml is handled in the upcoming Scope 3 publish workflow

scripts/release/verify_release_eligibility.sh
eLeontev: Fixed: removed ambient GITHUB_REF_NAME/GITHUB_SHA fallbacks so TARGET_TAG and TARGET_COMMIT must be passed explicitly by the workflow before the tag is created, accompanied by dedicated regression tests in test_verify_release_eligibility.py.

scripts/release/calculate_next_version.sh
eLeontev: Fixed: I added strict SemVer format validation (`^[0-9]+\.[0-9]+\.[0-9]+$`) and git existence verification (`git rev-parse`) for both the baseline tag and target ref, and removed error suppression on `git log` so unresolvable ranges fail loudly with `exit 1`. Added regression tests for `v0.1.0`, `0.9.9`, and unresolvable refs.

scripts/release/verify_release_eligibility.sh
eLeontev: Fixed: I added the regex check `[[ "${TARGET_TAG}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]` right next to the emptiness check in`verify_release_eligibility.sh`, blocking non-numeric tags like `main`, `v0.1.0`, etc., with fail-fast `exit 1`. Added regression tests covering invalid formats.

scripts/release/common.sh
eLeontev: The tag-triggered rebuilds will be completely removed from `.github/workflows/docker-publish-*.yml` in Scope 3 (`release-publish.yml`), making `release-publish.yml` the sole publisher via clean promotion and eliminating the race condition. For Scope 2, I have already hardened `promote_release_images` with strict SemVer validation and a registry pre-check (`docker manifest inspect`) to prevent accidental tag overwriting.

scripts/release/verify_release_eligibility.sh
```diff
@@ -0,0 +1,158 @@
...
+fi
+
+# 2. Resolve Target Commit SHA
+RELEASE_COMMIT=""
+if [ -n "${TARGET_COMMIT}" ] && [ "${TARGET_COMMIT}" != "null" ]; then
+  if ! RELEASE_COMMIT="$(git rev-parse --verify "${TARGET_COMMIT}^{commit}" 2>/dev/null)"; then
+    RELEASE_COMMIT="${TARGET_COMMIT}"
+  fi
+else
+  # Auto-resolve commit:
```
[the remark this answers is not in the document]
eLeontev: addressed: replaced to exit with code 1 and two tests added.

mplakhtiy (approved):

PR #738, opened 2026-08-17
fix(triage): deliver event RCAs by delegating triage to the failing cluster

### LINKED ISSUE #630: k8s-event triage never reaches Slack: the turn runs on the Chat Agent profile, and the card subscribes to `api_server`
## Summary

A Kubernetes event warning triggers the alert and the `🌱 _Digging down to the root cause..._`
message in Slack, and then nothing else ever arrives. The investigation itself **succeeds** — the
card reaches `status=done` with a complete report in its `result` column — but no part of that
report is ever delivered, and no PR can follow because no human ever sees a message to reply to.

Worked example: task `t_563b7aa4` on 2026-08-11, a `6094`-char report written at `12:46:49` that
never left the box. Full forensic capture is attached as a comment below.

**This is not a one-off.** Every k8s-event session exhibits the same failure — `k8s-evt-47294635`
(08-09 `16:24`), `k8s-evt-2bcf748b` (08-10 `15:27`), `k8s-evt-cce6097e` (08-11 `12:38`) each search
for `send_notification` and none has ever called it.

The failure is expensive to diagnose because **every no-op on this path is recorded identically to a
success**: the notifier logs a delivery, advances its cursor, and unsubscribes a completed task —
so by the time anyone looks, the evidence that a subscription ever existed has been cleaned up.

## Root cause

The Chat Agent's delegation contract — *"Hermes auto-subscribes this chat thread and posts the
specialist's progress back to it as each step completes, with no blocking timeout"*
(`agents/chat/config.yaml:39-41`) — assumes the ingress session is a **push-capable chat session**.
The k8s-event watcher injects through the **API server** instead. Every delivery mechanism
downstream then faithfully points back at that non-delivering HTTP channel.

Concretely, `trigger_agent_troubleshooter` builds every gateway URL from a bare `api_url`:

```python
api_url = os.environ.get("PLATFORM_API_URL", "http://127.0.0.1:8642")
...
f"{api_url}/api/sessions"                      # _create_gateway_session:232
f"{api_url}/api/sessions/{session_id}/chat"    # _start_agent_turn:309
```

Profile selection is by **URL path prefix**, so an unprefixed request scopes to the *default*
profile — *"When no `/p/<profile>/` prefix was given … enter the DEFAULT profile's scope:
api_server is a port-binding platform that lives on the default profile"*
(`gateway/platforms/api_server.py`, `_profile_scope`).

The default profile is the **Chat Agent**, whose `platform_toolsets.api_server` is deliberately just
`[mcp-router, kanban, memory]` (`agents/chat/config.yaml:120-123`). `send_notification` lives in the
`platform_control` MCP server, registered **only** in the Platform Agent profile
(`agents/platform/config.yaml:75-79`).

## The four breaks

**1. The agent cannot call `send_notification` — the tool is not in its toolset.**
The eight `tool_search("send_notification")` calls were not indecision: the agent was ordered by
`_build_agent_query` to use a tool its profile does not expose. `tool_describe` said so at
`12:38:56` — *"'send_notification' is not a deferrable tool. If you see it in the tools list
already, call it directly"* — and it was not in the list. The agent then did the only thing a Chat
Agent *can* do with `[mcp-router, kanban, memory]`: it delegated via a kanban card. Its exclusive
use of `kanban_create`/`kanban_show` is itself confirmation of which profile it was running under.

**2. The card *is* auto-subscribed — to a channel that cannot receive.**
`kanban_create` returned `"subscribed": true` at `12:40:34`. `_bind_api_server_session` hardwires
`platform="api_server"` and `chat_id=<session_id>` into the session ContextVars, and
`_maybe_auto_subscribe` copies exactly those into the row. The card is therefore subscribed to
`api_server/k8s-evt-cce6097e` — the stateless HTTP request that created it, which the very same
function marks `async_delivery=False` because it *"can never wake the agent after the turn ends, on
ANY route"*.

The notifier then behaves exactly as written. `adapter_supports_push(api_server)` is `False`, so it
takes the non-push branch where, in its own words, *"non-push (api_server): the wake self-post IS
the delivery"* (`gateway/kanban_watchers.py:618`). On that branch it does **not** post the report.
It injects a short synthetic "task completed" notice and logs:

```
kanban notifier: woke agent for t_563b7aa4 on api_server/k8s-evt-cce6097e
```

That counts as success — the cursor advances and, the task being `done`, `_kanban_unsub` removes the
subscription (`kanban_watchers.py:704-716`, `786-790`, `829`). **This is why `kanban_notify_subs`
shows 0 rows post-mortem**, which reads misleadingly as "the card was never subscribed".

The card `body` promises the worker that *"the gateway posts it verbatim into the requester's chat
thread"*. On a push adapter that is true. On `api_server` it is false, and nothing notices.

**3. The woken turn died mid-flight.** The wake arrived exactly as the session crossed the preflight
compression threshold (`233,385` ≥ `192,000` tokens); the turn ended `session_persistence_failed`
with `response_len=0`. Even with `send_notification` available, this turn could not have used it.

**4. The synchronous driver had already left.** `_start_agent_turn`'s `urlopen(..., timeout=300.0)`
expired at `12:42:44` — `4m05s` before the card completed, and `9m20s` short of the turn's actual
duration.

## Proposed fixes

Fixes 1 and 2 are independent and **each closes the incident on its own**. Doing both is what makes
delivery robust: fix 1 restores the agent's ability to speak, fix 2 removes the agent from the
delivery path altogether.

| # | Fix | Closes | Effort | Risk |
| 1 | Drive the k8s-event turn at the **`platform` profile** (`/p/platform/…`) | 1 | 3-line change | low |
| 2 | Bind the card's subscription to the **real chat thread**, not `api_server` | 2, 3, 4 | new patch module | low–medium |
| 3 | Stop the `300 s` driver timeout being load-bearing | 4 | 2-line change | low |
| 4 | Make a wake that produced nothing count as a **failed** delivery | 3 | ~15 lines | medium |

### Fix 1 — run the k8s-event turn on the Platform Agent profile

The `platform` profile's `platform_toolsets.api_server` already carries everything
`_build_agent_query` assumes (`agents/platform/config.yaml:75-79`):

```yaml
api_server:
  - hermes-api-server      # base bundle: terminal, files, and the kanban core pool
  - mcp-platform_control   # <- send_notification
  - mcp-developer_knowledge
  - mcp-gke
```

**No toolset change is needed.** The obvious worry — that switching profiles takes `kanban` away and
breaks delegation — does not apply: the kanban orchestrator surface is gated by the *top-level*
`toolsets` key, and `agents/platform/config.yaml:87-88` already sets `toolsets: [kanban]`. The
profile keeps delegation **and** gains the notification tool.

In `agents/platform/scripts/session_kv_server.py`:

```diff
 def trigger_agent_troubleshooter(session_id: str, alert_msg: str, payload: Dict[str, Any]) -> None:
@@
     api_url = os.environ.get("PLATFORM_API_URL", "http://127.0.0.1:8642")
+    # Profile selection is by URL path prefix, not by header: an unprefixed
+    # request runs on the DEFAULT profile, which is the Chat Agent — a
+    # delegation-only front door with no send_notification and no gke tools
+    # (agents/chat/config.yaml). The query built below instructs the agent to
+    # call send_notification and to reason about GKE, so it must be scoped to
+    # the Platform Agent profile that actually owns those tools.
+    api_profile = os.environ.get("PLATFORM_API_PROFILE", "platform").strip("/")
+    api_base = f"{api_url}/p/{api_profile}" if api_profile else api_url
     headers = {"Content-Type": "application/json"}
```

then pass `api_base` in place of `api_url` to `_create_gateway_session` and `_start_agent_turn`.
Both helpers already take the base as a parameter, so their bodies are untouched. Unknown profiles
return `404` (`_make_profile_prefix_middleware`), so a typo fails loudly rather than silently
degrading.

**Verification:** a triage session's `agent.log` should show `send_notification` as a direct
`tool_call_start` and **zero** `tool_search("send_notification")` calls.

**Caveat worth a conscious decision:** this grants the k8s-event path the Platform Agent's full base
bundle (terminal/file/exec), where today it has the Chat Agent's locked-down surface. That is the
*intended* privilege level for triage — the persona text already authorizes branch creation and PR
opening — but it is a real increase in blast radius from an internal trigger.

### Fix 2 — subscribe the card to the real chat thread (highest leverage)

The material for the correct answer already exists **three minutes before the card is created**.
`_register_session_routing` runs at step 2 of `trigger_agent_troubleshooter` and persists the real
chat identity:

```python
_register_session_routing(session_id, active_platform, thread_id)
#   -> session_metadata: {"platform": "slack",
#                         "thread_id": <alert message ts>,
#                         "chat_id":   os.environ["SLACK_HOME_CHANNEL"]}
```

served at `GET http://127.0.0.1:8699/v1/sessions/{session_id}/metadata`
(`session_kv_server.py:392`). This is not a new integration: **`send_notification` itself resolves
its Slack target through exactly this endpoint** (`platform_mcp_server.py:531-544`).

A new patch in the established four-file idiom (`deploy/docker/patches/`: module + `apply_` +
`verify_` + `test_`), hooked at the same seam the worker-inheritance patch already uses in the
`kanban_create` handler:

```python
new_task   = kb.get_task(conn, new_tid)
subscribed = _maybe_auto_subscribe(conn, new_tid)
_kanban_inherit_worker_subs(conn, new_tid)          # existing patch
_kanban_rebind_chat_origin(conn, new_tid)           # <- new: this fix
```

`kanban_chat_origin_subscribe.py`:

```python
COPY_COLUMNS = ("platform", "chat_id", "thread_id", "user_id", "notifier_profile")
RESOLVER_URL = "http://127.0.0.1:8699/v1/sessions/{sid}/metadata"
NON_DELIVERING = {"api_server", "k8s-watcher", ""}

def maybe_rebind_chat_origin(conn, task_id):
    """Repoint an api_server self-subscription at the chat thread that started it.

    A card created inside an API-server session is auto-subscribed to
    ``api_server/<session_id>`` — a channel `_bind_api_server_session` itself
    marks ``async_delivery=False``. The notifier's non-push branch then
    substitutes a content-free wake for the report. When the session was opened
    by the k8s event watcher, the originating Slack channel and thread were
    already persisted by `_register_session_routing`; use them instead.
    """
    from gateway.session_context import get_session_env
    if get_session_env("HERMES_SESSION_PLATFORM", "") != "api_server":
        return False
    sid = get_session_env("HERMES_SESSION_CHAT_ID", "")
    if not sid:
        return False

    meta = _resolve(sid)                     # GET RESOLVER_URL, 3s timeout, fail-soft
    platform = (meta.get("platform") or "").strip()
    chat_id  = (meta.get("chat_id")  or "").strip()
    thread_id = (meta.get("thread_id") or "") or None
    if platform in NON_DELIVERING or not chat_id:
        return False                          # not a chat-originated session

    from hermes_cli import kanban_db as _kb
    _kb.add_notify_sub(                       # seeds last_event_id at the head
        conn, task_id=task_id, platform=platform, chat_id=chat_id,
        chat_type="channel", thread_id=thread_id,
        notifier_profile=get_session_env("HERMES_SESSION_PROFILE", "") or "platform",
        delivery_metadata={"thread_id": thread_id} if thread_id else None,
    )
    _kb.remove_notify_sub(                    # drop the dead-end self-subscription
        conn, task_id=task_id, platform="api_server", chat_id=sid, thread_id="",
    )
    return True
```

**Why this is the highest-leverage fix.** With a `slack` subscription,
`adapter_supports_push(adapter)` is `True`, so the notifier takes the **push** branch: it posts the
card's `result` **verbatim into the alert thread**, then advances the cursor. Delivery stops
depending on an agent turn at all — which is why this single fix also neutralises breaks 3 and 4.
The compression race cannot eat it, the `300 s` timeout cannot outrun it, and a missing
`send_notification` cannot block it. It is also the only fix that makes the promise already written
into every card `body` actually true.

Contract notes, each with a precedent in the existing patches:

- Copy exactly `platform, chat_id, thread_id, user_id, notifier_profile` — the same set
  `_inherit_notify_subs`, `kanban_auto_subscribe.py` and `kanban_notify_propagate.py` all use.
- Seed `last_event_id` at the child's current head, never `0`. `kanban_auto_subscribe.py` documents
  three separate incidents caused by seeding at `0`, including a boot-time burst of 100+ replayed
  messages (issue #29905). `add_notify_sub` already snaps to the head.
- `INSERT OR IGNORE` semantics keep it idempotent, so a double-write with the manual
  `kanban_notify_propagate.py` is harmless.
- Fail soft: a notification-bookkeeping failure must never fail the `kanban_create` the agent is
  mid-conversation about — the rule `_maybe_auto_subscribe` already states in its docstring.

**Verification:** after the next triage card is created, `kanban_notify_subs` holds one row with
`platform='slack'` and the alert's `thread_id`; on completion `gateway.log` logs a Slack send rather
than `woke agent for … on api_server/…`.

### Fix 3 — stop the `300 s` driver timeout being load-bearing

`_start_agent_turn` blocks a FastAPI `BackgroundTask` on `urlopen(..., timeout=300.0)` while triage
routinely runs longer. The failure is logged as a bare `Failed to call gateway API chat execution:
timed out` — no session id, no elapsed time — in a file nobody reads.

There is prior art for this exact mistake one layer up: the Chat Agent's synchronous `ask_agent`
relay *"was removed: it blocked up to 300s with no visible progress"*
(`agents/chat/config.yaml:37-38`). The same 300-second assumption survived here.

```diff
-def _start_agent_turn(api_url: str, session_id: str, query: str, headers: Dict[str, str]) -> None:
+# The response body is NOT the delivery path — the report reaches chat via the
+# kanban notifier (see kanban_chat_origin_subscribe). This request only needs to
+# outlive the turn so the server-side run is not cancelled by a client
+# disconnect; 300s was shorter than a routine triage and silently orphaned it.
+_TURN_TIMEOUT = float(os.environ.get("PLATFORM_API_TURN_TIMEOUT", "1800"))
+
+def _start_agent_turn(api_url: str, session_id: str, query: str, headers: Dict[str, str]) -> None:
     """Post the agent query request to execute the diagnostic reasoning loop."""
+    started = time.monotonic()
     try:
         ...
-        with urllib.request.urlopen(req, timeout=300.0) as resp:
+        with urllib.request.urlopen(req, timeout=_TURN_TIMEOUT) as resp:
             if resp.status != 200:
                 logger.error(f"Gateway API chat execution failed (status {resp.status})")
     except Exception as exc:
-        logger.error(f"Failed to call gateway API chat execution: {exc}")
+        logger.error(
+            "Gateway API chat execution failed for session %s after %.1fs "
+            "(timeout=%.0fs): %s", session_id, time.monotonic() - started,
+            _TURN_TIMEOUT, exc,
+        )
```

**Why raise rather than fire-and-forget:** whether the aiohttp handler survives a client disconnect
is not established here, and cancelling a turn mid-investigation would be a worse failure than the
one being fixed. Raising the ceiling and demoting the response to non-load-bearing is the change
that is safe without further verification.

### Fix 4 — a wake that produced nothing is not a delivery

On the non-push branch the notifier treats `deliver_wake` returning without raising as success — it
advances the cursor and, for a `done` task, unsubscribes. But the woken turn here ended
`reason=session_persistence_failed, response_len=0`. Nothing was produced, nothing was delivered,
and the only durable record of the event was discarded.

`kanban_notify_delivery.py` already established the principle this violates: *"the cursor is written
before anything is committed to delivering"* is the defect; *"read without claiming; advance only
after the send succeeds"* is the rule. The non-push branch honours the letter of that rule while
calling a content-free wake a "send".

Have `deliver_wake` (or the notifier's non-push call site) surface the woken turn's outcome, and
treat `response_len == 0` / `session_persistence_failed` as a failed delivery — which the existing
code already handles correctly: it increments `sub_fail_counts`, rewinds the claim, and lets the
next tick retry, dropping the subscription only after `MAX_SEND_FAILURES`. All the machinery exists;
only the success predicate is wrong.

**Priority:** lower than 1–3. Once fix 2 lands these cards take the push branch and the non-push
branch leaves the critical path. It remains worth fixing because silent loss is what made this
incident expensive to diagnose.

**Related observation, not a fix:** the session hit preflight compression at `233,385` tokens across
`203` messages for a single triage. Eight redundant `tool_search` calls and repeated full-body
`kanban_show` responses contribute; fix 1 removes the former.

## What needs no fix

**The missing PR is correct behaviour.** `_build_agent_query` authorizes branch creation and PR
opening only *"For subsequent turns if the user replies"* with `apply`. No report reached Slack, so
no human could reply, so no PR — and the card was assigned to a read-only **Cluster Agent** profile
with `workspace_kind=scratch` and `branch_name=None`, which by design cannot open one anyway. Fix
the delivery and the `apply` path becomes reachable; the PR gate itself is fine.

---

*The complete forensic capture — all 38 card columns, the verbatim `6094`-char report, `task_runs`,
the 11-row `task_events` table, and the full UTC timeline — is attached as a comment below.*

*Investigated and drafted with the help of Claude.*

stalhaali:
## Summary
Kubernetes event triage produced a root-cause analysis and delivered it nowhere. The report was written, stored on a kanban card, and never reached the thread that raised the alert. The cause is an address, not a missing channel: an
  event-triage turn arrives over the REST gateway, whose session-context chokepoint stamps `platform="api_server"` with the session id in `chat_id`, and `_maybe_auto_subscribe` copies those values straight into `kanban_notify_subs`. The
  subscription is well-formed, the notifier finds it on every tick, and no such chat platform exists.

  This PR substitutes the alert's real chat route into that row as it is written, so completing the card is the delivery. It also gives the management cluster a Cluster Agent, because the cluster that raised the event in #630 had none.

  ## Why This Change

  The delivery machinery already existed and already worked. Every kanban card gets a subscription, and the notifier posts a subscribed card's `result` to chat when the card turns terminal. Event-triage cards got one too — pointed at
  `api_server`.

  That was established on the live install before any of this code was written. One undelivered card's subscription row was hand-edited to the alert's real space and its notifier cursor rewound; the report was delivered. Repeating it with
  `thread_id` populated delivered the report as a reply under the alert it answers. No code ran in either test that this PR changes.

  So the fix is the address. An earlier revision of this branch built a second egress instead — a `notify` MCP server on the Cluster Agent and a two-call `send_notification`-then-`kanban_complete` contract in the card body. That is withdrawn
  here in favour of the smaller change, after @jayantid  pointed out the repo already solves this problem in `kanban_auto_subscribe.py`.

  ## What Changed

  - **`deploy/docker/patches/kanban_event_routing.py`** — a new image patch. `_maybe_auto_subscribe` gains one call that rewrites the three destination values before the row is written: the session id in `chat_id` is looked up in
  `session_kv.db`'s `session_metadata`, and the recorded `platform`/`chat_id`/`thread_id` replace the `api_server` origin. Fails open on every path — `kanban_tools` also runs in CLI workers with no routing database, so a missing, unreadable, or
  malformed one leaves today's behaviour exactly as it is.
    - The `api_server.py` chokepoint is not touched. It hardwires those values on purpose as the structural guard against #10760; the substitution happens one layer down, where the subscription is written.
    - Two origins are treated as undeliverable: `api_server` from the chokepoint, and `k8s-watcher`, which is what event sessions carried in `session_metadata` before the next bullet.
    - **A fall-through on an event session logs a warning naming the session and the reason.** Failing open is right for a CLI worker, but on a `k8s-evt-` session it recreates the bug this patch exists to fix — the row is written undeliverable,
  the card completes, the board reads healthy and chat stays empty. The report is still lost; the warning is what makes it a greppable line rather than nothing at all. A successful substitution and a non-event session are both silent.
    - It ships in the existing shared `COPY`/`RUN` block with the four other kanban patches, so the image gains no layers (the layer-depth ceiling from build f9f1747c still applies).
  - **`session_metadata` now records `platform`.** The row already carried `chat_id` and `thread_id`; without the platform, the substitution has nothing to address the report as, and the pre-existing value is the non-chat `k8s-watcher`. A 
  session that never reaches `_register_session_routing` therefore keeps today's behaviour rather than being re-addressed to a guess.
  - **The card body asks for one terminal call.** `_triage_task_body` no longer names `send_notification` or orders two calls. It asks for `kanban_complete` with the entire report in `result` — `result` is verbatim what the notifier posts — and
  forbids delegating the diagnosis, because only this card carries the subscription. The session id no longer travels through the prompt at all; routing is resolved from the ambient session at `kanban_create` time.
  - **Routing is a prompt, not a parameter.** The gateway honours a `/p/<profile>/` prefix only under `gateway.multiplex_profiles`, which is off; a `profile` key in the create body is accepted with a 201 and dropped. So the turn is addressed to
  the front door as a router: file one card, on the `cluster-*` agent for the event's cluster, with the brief copied verbatim between two markers. On 2026-08-17 a front door handed that brief as instructions rather than as cargo summarised it,
  dropped the delivery instruction, and filed extra cards asking other agents to post.
  - **The report no longer invites an `apply` reply, because nothing on this path can honour one.** The template ended with **To authorize:** reply **'apply'**. The agent that acts on that reply gets the report back from the `incident_context`
  plugin, which reads `GET /v1/incidents/by-thread`; the only writer of the `incidents` table is `platform_mcp_server.send_notification`, which is exactly the egress call this branch replaced with `kanban_complete`. So no row is written,
  `_lookup` returns `None`, the plugin leaves the message untouched, and the front door receives the bare word `apply` — no report, no options, no cluster. Nothing unsafe follows; the front door holds no GitOps write path and simply cannot act.
  But on `main` no report ever reached a human, so nobody replied and the gap never showed. Delivering the report makes that reply the normal case, which is why it is this PR's problem and not a pre-existing one. Raised in review.
    - The bullet is dropped and the agent is told not to write a call-to-action of its own — a list ending in a recommendation invites one. SOUL.md §7 rule 3 ("no offer to help further") is on the side of the removal. The options, the
  recommendation, and the instruction to name each manifest change precisely enough that someone can open the PR from the report alone all stay, since nothing can ask the author a follow-up question.
    - **Storing the report on the delivery path is the real fix, and it is issue #802.** The notifier holds the substituted `chat_id`, `thread_id` and `result` in one row, so it is the natural place to POST `/v1/incidents` from — but it lives in
  read-only `/opt/hermes`, so it needs its own Dockerfile patch and a live test with a GitOps repository wired up. Splitting it keeps this PR to the delivery bug in #630.
    - Docs follow the code: the Contract 3 template and its approval-interaction note in `autoops-architecture.md`, the `incidents` storage prose, a Known gaps entry, and a caveat above the `session_management.md` sequence diagram saying Phase 3
  is unreachable from Phase 1 today.
  - **Phase 3 of the `session_management.md` sequence diagram gets its own participant.** As drawn, the Cluster Agent that wrote the report also opened the GitOps PR — a write path it does not have. Phase 3 now hands the apply to the front door
  and from there to the Platform Agent, which is who actually holds it. Raised in review.
  - **The management cluster gets a Cluster Agent profile.** `cluster_agent_reconcile.py` no longer excludes it. #630's alert was raised on `platform-agent-host`, so the front door had no `cluster-*` agent to file the card on.
  `RECONCILE_EXCLUDE` is the opt-out and it prunes.
  - **The watcher stops double-counting the host cluster.** With the management cluster now holding a profile, `buildWatchSet` produced two entries for it and nothing downstream collapsed them — the dedup cache is per watched cluster and
  `EventKey` is `(UID, reason)` with no cluster in it, so one event raised two alerts, opened two sessions, and spent two of the day's budget. The **direct** entry wins and absorbs the profile's `project/location/cluster` triple, so the payload
  is still stamped with it.
    - Which of the two survives is not cosmetic, and the first revision of this bullet got it backwards. `discoverClusterProfiles` never contacts an API server — `kubernetes.NewForConfig` builds a client from a kubeconfig without authenticating
  — so a profile reaches the watch set whether or not its credential can list events. It authenticates as the GSA, bounded by IAM and by master authorized networks; the direct entry uses the pod's own KSA under `kubeagents:minimal:*`, which
  cannot be denied by either. On a single-cluster install the management profile is the only profile, so keeping it and dropping the direct entry would leave one unverified credential: deny it and the reflector retries the initial list forever,
  `synced` stays 0, and `realMain` returns after `initialSyncGrace`. The pod crashloops and nothing is watched. Raised in review; the direction is reversed and `main_test.go` asserts the surviving entry is the direct one carrying the profile's
  identity.

  ## Context

  Closes #630.

  Overlaps with #735 (Fuxiao Gao) — both touch `cluster_agent_reconcile.py`. A merge-conflict warning rather than duplicated work; the changes are unrelated in intent.

  The 14 undeliverable `('api_server', 'k8s-evt-…', '')` rows observed on the diagnosed install exist on `main` today and are not repaired by this PR, which only fixes rows written from now on. A backfill is a separate change; the reports are
  still on their cards.

  ## Testing

  ```bash
  python3 -m pytest agents/platform/scripts/ deploy/docker/patches/ tests/ scripts/
  cd k8s-operator && go build ./... && go test ./cmd/k8s-event-watcher/...
  make docs-check && make iac-parity-check
  prettier --check <changed md>
  docker build --platform linux/amd64 -f deploy/docker/Dockerfile --target platform .
  docker build --platform linux/amd64 -f deploy/docker/Dockerfile --target credential-proxy -t credential-proxy:latest . \
    && python3 scripts/check_image_layers.py
  ```

  New: `deploy/docker/patches/test_kanban_event_routing.py` (28 cases — the substitution, six fail-open paths, the undeliverable warning and the two cases that must stay silent, the anchored edit) and `verify_kanban_event_routing.py`, a build
  gate that drives the _real patched_ `kanban_tools._handle_create` with an event session's context and asserts both the substitution and every fall-through. Widened: `test_session_kv_server.py` (the reduced card body, the recorded platform, and
  three cases around the removed `apply` promise — that the template makes no promise it cannot honour, that the agent is told not to write its own, and that the options and the recommendation survive the removal) and `main_test.go` (watch-set
  dedup, now asserting the direct entry survives carrying the profile's identity triple).

  `make docs-check`, `make iac-parity-check`, `prettier --check`, `go build ./...`, and the 28 patch unit tests pass. `make dev-rebuild-agent ARGS="platform"` built and pushed both the `platform` and `credential-proxy` images, so the in-build
  gate ran. `pytest`, `go test`, `gofmt`, and `check_image_layers.py` are left to CI.

  ### Live validation

  Built from this branch with `make dev-rebuild-agent ARGS="platform"` and rolled out to `gke_stalhaali-gkedemos_us-east4_platform-agent-host` — pod `platform-agent-gateway-5989d8cf4d-mpftr`, container `platform-agent`, Google Chat only. The
  `platform` target's build gate (`verify_kanban_event_routing.py`) passed, and `grep -c _kanban_event_route /opt/hermes/tools/kanban_tools.py` returned 2 in the running pod: the import and the call.

  **Before.** Every `k8s-evt-` subscription on the board read the broken shape — the six most recent were rows 24–29, e.g. `(29, 'api_server', 'k8s-evt-eb14fa5c', '')` — a chat platform no adapter can deliver to, the session id where the channel
  belongs, no thread.

  **The run.** A `crashloop-probe` pod (`sh -c 'exit 1'`) on the management cluster backed off, the watcher alerted the space, and:

  1. The subscription written for the triage card was `(32, 'google_chat', 'spaces/0sm7myAAAAE', 'spaces/0sm7myAAAAE/threads/PIT8aWDiWkI')`. All three fields substituted, against row 29's shape from the same code path on the same install.
  2. `tasks` held exactly **one** new card, `t_4c243128`, assigned to `cluster-stalhaali-gkedemos-platform-agent-host-us-east4` — the management-cluster profile that did not exist before this PR. No second card asking another agent to deliver
  the answer.
  3. Progress reached the thread while the card was `running`, and on `done` the report was posted **as a reply under the alert it answers**. `length(result)` was 1677 — the whole report, which is what the card body asks for and what the
  notifier posts verbatim.

  The distinctly-different-value check is the row itself: `api_server`/`k8s-evt-…`/`''` and `google_chat`/`spaces/…`/`threads/…` share no field, and the first is undeliverable by construction.

  **Earlier, before the code existed,** the same substitution was confirmed by hand on card `t_e980bdfc`: its row was rewritten to a real chat address and the notifier cursor rewound, twice — once without a thread, once with — delivering to the
  space and then to the alert's thread. That row was restored; two test messages remain in the space, as does the pre-fix `t_e980bdfc` row.

  **Cleanup.** `crashloop-probe` deleted, along with the `simulated-alert-*` pods and the `triage-live` namespace from the earlier runs.

  **One change came after that run.** The undeliverable warning was added in response to review, and the image above predates it. It sits on the three paths that decline to substitute, none of which the run took — the run substituted. It is
  covered by five unit tests, including that a successful substitution and a non-event session both stay silent, and the in-build gate drives the same fall-throughs. It has not been observed firing in a pod.

  **Not covered:**

  - **The `session_metadata.platform` field on the case it exists for.** This install is Google Chat only, so a delivery here succeeds with or without it. Unit tests cover the both-enabled case and the `k8s-watcher` fall-through.
  - **`RECONCILE_EXCLUDE` pruning an existing profile.** Unit test only; exercising it live means deleting the management cluster's profile and waiting an hourly tick on the install every other check depends on.
  - **The watch set producing one entry instead of two.** `main_test.go` only — the watcher is not separately addressable in this install, so there is no log to read the resolved watch set out of.
  - **A blocked card is still delivered nowhere.** The subscription fires on `completed`, so `kanban_block` on an event-triage card reaches no one. That is pre-existing behaviour for every card on the board, not something this PR changes, and it
  is out of scope here.
  - **The template's removed `apply` bullet.** The report is what a reader sees, and this branch's live run predates the removal. Verified by unit test and by rendering the template out of the module: no `To authorize:`, no `reply 'apply'`,
  options and recommendation intact.
  - **The reversed watch-set precedence.** `main_test.go` only, for the same reason as the watch-set entry above: the watcher is not separately addressable here. The credential asymmetry it turns on is exactly what an install cannot demonstrate
  when both credentials happen to work, which this one's do.

  ## Self-Review

  

  Not yet run as an independent pass over the current diff. The rework itself was driven by review — @jayantid finding that the branch reinvented `kanban_auto_subscribe.py`'s mechanism — and testing the alternative empirically is what
  confirmed it. What that pass found and changed:

  - **The whole egress was unnecessary.** A `notify` MCP server, a factored `notify_delivery.py`, a SOUL.md obligation, a config block, and a PVC-heal script to backfill it into existing profiles — all of it replaced by one substitution in the
  row that was already being written. Removed: `notify_server.py`, `notify_delivery.py`, `cluster_config_heal.py`, and their tests, plus the reverts to `agents/cluster/`, `platform_mcp_server.py`, `docker-entrypoint.sh`, and
  `check_prompt_assets.py`. 35 files became 20.
  - **Two claims in the previous body did not survive checking.** "`session_metadata` now records `platform`. Nothing did before" was wrong — the field was always written, with the non-chat value `k8s-watcher`. And a worker described as
  producing "a report in the wrong shape" had in fact produced the correct template.
  - **An earlier claim that surviving subscription rows were provably failed deliveries** was wrong: rows persist after a successful delivery too. Row survival proves the address, not the outcome.
  - **The fix was itself silently droppable**, which @nsherron raised against the previous design and which survived the rework in a narrower form. Their specific case is gone — there is no second call for an agent to forget, and the worker they
  cited as evidence (`t_e980bdfc`, zero MCP calls, improvised `python3 -c`) completed its card with a full `result`, which under this design _is_ the delivery. But `resolve_chat_route` failing open on an event session drops the report exactly
  as described. It now warns.

  What still needs a hostile read: the fail-open paths under a database that exists but is locked by a concurrent writer, and whether the anchor in `apply_kanban_event_routing.py` survives the next Hermes bump.

  ## Risk & Rollout

  The substitution is bounded by its guard: it only fires when `platform` is `api_server` or `k8s-watcher` _and_ the session id keys a stored route naming a real chat platform and channel. A card filed from a genuine chat session is never
  re-addressed, and every failure path returns the caller's arguments unchanged. Reverting the patch restores the pre-existing behaviour exactly — the undeliverable row.

  The management cluster now has a Cluster Agent profile scoped to the cluster kube-agents runs in, so it can read the harness's own namespace. How far that reaches is set by the GSA's permission set, not by the `kubeagents:minimal:*`
  ClusterRole — a Cluster Agent pins a `get-credentials` kubeconfig and authenticates as the GSA. On the default `read-only` roles that means the harness's own pods, logs, ConfigMaps and events, and **not** its Secrets; on the opt-in `gke-admin`
  roles it means the full Kubernetes API on that cluster, **Secrets included**. The security reference is the canonical statement of the trade-off.

  `RECONCILE_EXCLUDE` is the opt-out and it prunes, so reverting that decision needs no code change. It does not stop the watcher — the direct `--in-cluster` entry is always in the watch set, profile or no profile — so an excluded management
  cluster keeps raising alerts with no `cluster-*` agent for the card to land on. Excluding it gives up management-cluster triage, not just the profile.


docs/site/src/content/docs/reference/security-and-iam.md
```diff
@@ -24,7 +24,9 @@ This is the canonical answer. Other pages summarize it and link here; if they ap
...
+  - `notify` is its one write, and it writes to chat rather than to the cluster: a single tool, `send_notification`, so the agent can post the report of a Kubernetes event triage into the thread that raised the alert. That triage arrives as a kanban card filed on behalf of a session that came in over the REST gateway, with no chat thread subscribed to its completion, so without an egress the report is produced and discarded. The grant is one tool on a server that exposes only that tool, deliberately in place of `platform_control`, which would have brought the provisioning and Config Connector surface with it.
+  - **The management cluster now has a Cluster Agent too,** so one of these profiles is scoped to the cluster kube-agents itself runs in and can read the harness's own namespace within the RBAC ceiling above — including its Secrets. That is a widening of the blast radius accepted on purpose: the management cluster's workloads fail like any other cluster's, and the agent that triages an event has to be the one scoped to the cluster that raised it. Add its name to the `cluster-agent-reconcile` job's `RECONCILE_EXCLUDE` to opt out; the next reconcile then prunes the profile.
```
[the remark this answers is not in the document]
stalhaali: done

agents/cluster/AGENTS.md
```diff
@@ -6,16 +6,17 @@ This folder is the home of a **Cluster Agent** — a Hermes profile scoped to a
```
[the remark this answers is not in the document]
stalhaali: done

docs/site/src/content/docs/reference/security-and-iam.md
```diff
@@ -24,7 +24,9 @@ This is the canonical answer. Other pages summarize it and link here; if they ap
...
+  - **The management cluster now has a Cluster Agent too,** so one of these profiles is scoped to the cluster kube-agents itself runs in and can read the harness's own namespace within the RBAC ceiling above — its pods and their logs, ConfigMaps, ServiceAccounts and events, but **not** its Secrets, which no ClusterRole here grants. That is a widening of the blast radius accepted on purpose: the management cluster's workloads fail like any other cluster's, and the agent that triages an event has to be the one scoped to the cluster that raised it. Add its name to the `cluster-agent-reconcile` job's `RECONCILE_EXCLUDE` to opt out; the next reconcile then prunes the profile.
```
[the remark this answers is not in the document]
stalhaali: updated. Neither the KSA nor kubeagents:minimal:* is the binding constraint here. cluster_agent_profile.create_profile runs gcloud container clusters get-credentials and pins the resulting kubeconfig (cluster_agent_profile.py:278,299), so a Cluster Agent's Kubernetes API calls authenticate as the pod's GSA, and the ClusterRole doesn't bound them. So the answer is per permission set, from provision_04_gcp_iam.sh: on the default read-only roles (container.clusterViewer, container.viewer, line 205) no Secrets; on the opt-in gke-admin roles (container.clusterAdmin, container.admin, line 215) the full API, Secrets included.

agents/platform/scripts/session_kv_server.py
```diff
@@ -547,15 +580,24 @@ def _build_agent_query(session_id: str, payload: Dict[str, Any]) -> str:
         f"- **Resource:** {namespace}/{object_kind}/{object_name}\n"
         f"- **Event Reason:** {event_reason}\n"
         f"- **Warning Message:** {message}\n\n"
-        f"When calling your send_notification tool to report findings, you MUST pass this exact session ID: '{session_id}' as the session_id argument so it routes as a threaded reply to the warning alert.\n\n"
+        f"**Finish with two calls, in this order.** First "
```
jayantid: **The fix for "silently dropped reports" is itself silently droppable.**

The two-call obligation lives only in prose: this card body, `agents/cluster/SOUL.md` section 6b, and `agents/cluster/AGENTS.md`. I grepped for a code-level check and there is none. If a Cluster Agent completes without calling `send_notification`, the card goes `done`, the board looks healthy, the chat stays empty — indistinguishable from success. That is #630, restored, with no way to tell it happened.

Your own live validation has worker `t_e980bdfc` doing exactly that: same profile, image and config as the four that worked, zero MCP calls, improvised a `python3 -c` import instead. One miss in five observed runs, unexplained, no alarm.

A backstop is cheap with what you have already built here — `session_kv_server` knows `chat_id`/`thread_id` for every `k8s-evt-` session and already serves `/v1/incidents/by-thread`, so a check when a triage card turns terminal that posts the card `result` if no incident row exists for that thread closes it in roughly twenty lines. That is my one merge blocker; everything else I found can be a follow-up.

jayantid:
**Worth a design conversation: this repo already solved this class of bug in code.**

`deploy/docker/patches/kanban_auto_subscribe.py` documents the mechanism — the gateway notifier delivers a card's completion to a chat thread if and only if the card has a `kanban_notify_subs` row, and `_maybe_auto_subscribe` writes that row at `kanban_create` from `HERMES_SESSION_CHAT_ID` / `_THREAD_ID` / `_PLATFORM`, "session context that exists only on inbound user messages."

The same file records the 2026-08-07 incident, where a card without that context left an answer undelivered for 91.3s, and the conclusion drawn from it:

> The manual remedy, `agents/platform/scripts/kanban_notify_propagate.py`, relies on the worker remembering to run it; it didn't.

The fix chosen then was to make the subscription automatic in code. This PR's root cause is that same sentence — the `k8s-evt-` api_server session sets no chat context, so the card gets no subscription row — and the approach it takes is the one that was already rejected: the worker must remember, enforced by prose.

The deterministic path is small and uses what is already here. `_register_session_routing` writes `platform`, `chat_id` and `thread_id` for exactly these sessions, and this PR adds `platform` to that row. Plumb those three into the session context the api_server session carries, and `_maybe_auto_subscribe` fires at `kanban_create`, the card is subscribed to the alert's thread, and `kanban_complete` delivers. That would remove the need for the `notify` MCP server, the two-call ordering, the verbatim-copy contract on the front door, the SOUL/AGENTS section 6b rewrite, and `cluster_config_heal.py` with its migration.

Honest caveat: this likely needs a Hermes patch to plumb context on the api_server path, and I have not verified that upstream. But it is the same shape as the ten `apply_kanban_*.py` patches already in `deploy/docker/patches/`, and it is one patch against 35 files.

Not asking you to rewrite this on my say-so — but I would want the call made deliberately before it merges, rather than discovered later.

stalhaali:
> **Worth a design conversation: this repo already solved this class of bug in code.**
> 
> `deploy/docker/patches/kanban_auto_subscribe.py` documents the mechanism — the gateway notifier delivers a card's completion to a chat thread if and only if the card has a `kanban_notify_subs` row, and `_maybe_auto_subscribe` writes that row at `kanban_create` from `HERMES_SESSION_CHAT_ID` / `_THREAD_ID` / `_PLATFORM`, "session context that exists only on inbound user messages."
> 
> The same file records the 2026-08-07 incident, where a card without that context left an answer undelivered for 91.3s, and the conclusion drawn from it:
> 
> > The manual remedy, `agents/platform/scripts/kanban_notify_propagate.py`, relies on the worker remembering to run it; it didn't.
> 
> The fix chosen then was to make the subscription automatic in code. This PR's root cause is that same sentence — the `k8s-evt-` api_server session sets no chat context, so the card gets no subscription row — and the approach it takes is the one that was already rejected: the worker must remember, enforced by prose.
> 
> The deterministic path is small and uses what is already here. `_register_session_routing` writes `platform`, `chat_id` and `thread_id` for exactly these sessions, and this PR adds `platform` to that row. Plumb those three into the session context the api_server session carries, and `_maybe_auto_subscribe` fires at `kanban_create`, the card is subscribed to the alert's thread, and `kanban_complete` delivers. That would remove the need for the `notify` MCP server, the two-call ordering, the verbatim-copy contract on the front door, the SOUL/AGENTS section 6b rewrite, and `cluster_config_heal.py` with its migration.
> 
> Honest caveat: this likely needs a Hermes patch to plumb context on the api_server path, and I have not verified that upstream. But it is the same shape as the ten `apply_kanban_*.py` patches already in `deploy/docker/patches/`, and it is one patch against 35 files.
> 
> Not asking you to rewrite this on my say-so — but I would want the call made deliberately before it merges, rather than discovered later.



Thanks for pointing this out. I checked and each card actually has a subscription:
```
 kanban_notify_subs, grouped by platform/chat_id/thread_id:
  ('api_server', 'k8s-evt-a5ac299b', '', 1)
  ('api_server', 'k8s-evt-8be70f99', '', 1)
  ('api_server', 'k8s-evt-a1b55d9d', '', 1)
....
```
but it's pointed at api_server as the platform, which isnt a chat platform, so the notifier can never deliver it. So we may just need to fix setting the right context and this would make the triage report sent back to the original chat thread.

Going to try that, if it works, we can simplify the PR and drop  the notify MCP server, the two-call ordering, SOUL §6b and cluster_config_heal.py.

dshnayder (approved):

stalhaali:
> **Worth a design conversation: this repo already solved this class of bug in code.**
> 
> `deploy/docker/patches/kanban_auto_subscribe.py` documents the mechanism — the gateway notifier delivers a card's completion to a chat thread if and only if the card has a `kanban_notify_subs` row, and `_maybe_auto_subscribe` writes that row at `kanban_create` from `HERMES_SESSION_CHAT_ID` / `_THREAD_ID` / `_PLATFORM`, "session context that exists only on inbound user messages."
> 
> The same file records the 2026-08-07 incident, where a card without that context left an answer undelivered for 91.3s, and the conclusion drawn from it:
> 
> > The manual remedy, `agents/platform/scripts/kanban_notify_propagate.py`, relies on the worker remembering to run it; it didn't.
> 
> The fix chosen then was to make the subscription automatic in code. This PR's root cause is that same sentence — the `k8s-evt-` api_server session sets no chat context, so the card gets no subscription row — and the approach it takes is the one that was already rejected: the worker must remember, enforced by prose.
> 
> The deterministic path is small and uses what is already here. `_register_session_routing` writes `platform`, `chat_id` and `thread_id` for exactly these sessions, and this PR adds `platform` to that row. Plumb those three into the session context the api_server session carries, and `_maybe_auto_subscribe` fires at `kanban_create`, the card is subscribed to the alert's thread, and `kanban_complete` delivers. That would remove the need for the `notify` MCP server, the two-call ordering, the verbatim-copy contract on the front door, the SOUL/AGENTS section 6b rewrite, and `cluster_config_heal.py` with its migration.
> 
> Honest caveat: this likely needs a Hermes patch to plumb context on the api_server path, and I have not verified that upstream. But it is the same shape as the ten `apply_kanban_*.py` patches already in `deploy/docker/patches/`, and it is one patch against 35 files.
> 
> Not asking you to rewrite this on my say-so — but I would want the call made deliberately before it merges, rather than discovered later.

done

agents/platform/scripts/session_kv_server.py
stalhaali: no longer applies

agents/platform/scripts/session_kv_server.py
```diff
@@ -568,11 +613,61 @@ def _build_agent_query(session_id: str, payload: Dict[str, Any]) -> str:
         f"🔗 [GKE Workloads](https://console.cloud.google.com/kubernetes/workload/overview{workloads_project_query}) | "
         f"[Cloud Logs](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22{logs_project_query})\n\n"
         f"---"
-        f"\n\n**GitOps PR Instructions (For subsequent turns if the user replies):**\n"
-        f"If the user replies to the thread with 'apply' or 'apply Option <letter>':\n"
-        f"1. A bare 'apply' (or 'apply recommended') means apply the option you marked '✅ **Recommended: Option <letter>**', or the only option you proposed if there was just one. You are explicitly authorized to create a new branch, modify the resource manifests in the local checkout, commit, push, and open a GitHub Pull Request matching the selected option.\n"
-        f"2. Post a threaded response confirming the PR was created and include the clickable PR link.\n"
-        f"3. Do not execute any write mutations (kubectl scale, patch, or apply) directly on the live cluster."
+        f"\n\n**What happens if the user replies 'apply':**\n"
```
jayantid: **The report tells the user to reply `apply`, and nothing on the new path stores what `apply` needs.**

Line 612 ends every report with `**To authorize:** reply **'apply'**`. The agent that acts on that reply gets the report from the `incident_context` plugin, which fetches it with `GET /v1/incidents/by-thread` (`agents/platform/plugins/incident_context/__init__.py:38`). The only writer of the `incidents` table is `send_notification`, at `agents/platform/scripts/platform_mcp_server.py:736`. This PR replaces `send_notification` with `kanban_complete`, so no row is ever written.

The user reads the RCA in the thread and replies `apply`. `_lookup` returns `None`, the plugin leaves the message untouched, and the front door receives the bare word `apply` — no report, no options, no cluster, no manifest. It cannot open the PR the report just promised. The PR body lists this under **Not covered** as pre-existing. On `main` no report ever reached a human, so nobody ever replied; this PR makes that reply the normal case.

Fix: POST the completed report to `/v1/incidents` on the delivery path, keyed by the substituted `chat_id`/`thread_id` — the notifier holds all three values in the subscription row alongside `result`. If that is out of scope for this PR, drop the `To authorize: reply 'apply'` bullet from the template until it works.

k8s-operator/cmd/k8s-event-watcher/main.go
```diff
@@ -863,39 +863,61 @@ func realMain(argv []string) error {
...
 	if f.profilesDir == "" || f.inCluster || f.kubeconfig != "" {
-		client, err := buildKubeClient(f)
-		if err != nil {
-			return nil, err
+		// Matched on the bare name because that is all the direct entry has —
+		// --in-cluster reads no cluster_identity. Two same-named clusters in
+		// different locations would therefore hide this one, but only in the
+		// window before reconcile has given the management cluster its own
+		// profile, and the log line below says which profile absorbed it.
+		if profile, dup := profileNames[f.clusterName]; dup {
```
jayantid: **Dropping the direct entry removes the watcher's only credential that is guaranteed to work.**

`discoverClusterProfiles` never contacts an API server. `kubernetes.NewForConfig` at line 392 builds a client from a kubeconfig; it does not authenticate. So a profile reaches `profileNames` whether or not its credential can list events. The direct entry it displaces uses the pod's KSA under `kubeagents:minimal:*`, which grants `events` get/list/watch. The profile entry authenticates as the GSA through a `get-credentials` kubeconfig, bounded by IAM instead.

On a single-cluster install the management profile is now the only profile, so the watch set becomes one entry with an unverified credential. Deny that credential — the `custom` permission set without `roles/container.viewer`, or master authorized networks that exclude the pod's egress — and the reflector retries the initial list forever, `synced` stays 0, and `realMain` returns the error at line 852 after `initialSyncGrace`. The pod crashloops and no cluster is watched at all. Before this change the direct entry always synced, so `synced >= 1` held and one failing cluster degraded on its own, which is what the comment at line 826 intends. The comment here says dropping the direct entry "costs nothing — it is the same cluster, reached with a different credential"; the different credential is the cost.

Fix: drop the profile duplicate instead of the direct entry, and take the identity triple off the profile so the payload is still stamped with `project/location/cluster`.

```go
// The profile supplies project/location; the in-cluster client supplies
// reachability. Keep the one that cannot be denied by IAM.
if id, dup := profileIdentity[f.clusterName]; dup {
    clusters = removeProfile(clusters, id.Profile)
    tc := targetCluster{Name: f.clusterName, Profile: "direct",
        ProjectID: id.Project, Location: id.Location}
    // ... buildKubeClient(f), append tc
}
```

agents/platform/docs/session_management.md
```diff
@@ -68,7 +81,9 @@ sequenceDiagram
     participant Watcher as k8s-event-watcher
     participant Proxy as session_kv_server (Port 8699)
     participant Gateway as Hermes Gateway (Port 8642)
-    participant Agent as Platform Agent LLM
+    participant Front as Planning Agent (default profile)
+    participant Agent as Cluster Agent for the event's cluster
```
jayantid: **Phase 3 of the sequence diagram now says a Cluster Agent opens the GitOps PR.**

This line renames the `Agent` participant to the Cluster Agent. Phase 3 was not updated with it and still reads `Agent->>Agent: Create branch, edit git manifests, open GitOps PR` at line 120. A Cluster Agent profile has no GitOps write path. `docs/site/src/content/docs/reference/security-and-iam.md:27` says so directly, and `_triage_task_body` says the apply is "carried out by the agent that holds the GitOps write path" (`agents/platform/scripts/session_kv_server.py:617`).

The diagram is the only one of those three places that names the actor, and it names one that cannot do the work. A reader debugging a failed `apply` from this doc goes looking at the wrong profile's toolset.

Fix: give Phase 3 its own participant.

```
    participant Fixer as Platform Agent (GitOps write path)
    ...
    Plugin->>Gateway: Spawn Fixer Agent with rewritten message
    Gateway->>Fixer: Inject context into conversation turn
    Fixer->>Fixer: Create branch, edit git manifests, open GitOps PR
```

agents/platform/scripts/session_kv_server.py
jayantid: **The report tells the user to reply `apply`, and nothing on the new path stores what `apply` needs.**

Line 612 ends every report with `**To authorize:** reply **'apply'**`. The agent that acts on that reply gets the report from the `incident_context` plugin, which fetches it with `GET /v1/incidents/by-thread` (`agents/platform/plugins/incident_context/__init__.py:38`). The only writer of the `incidents` table is `send_notification`, at `agents/platform/scripts/platform_mcp_server.py:736`. This PR replaces `send_notification` with `kanban_complete`, so no row is ever written.

The user reads the RCA in the thread and replies `apply`. `_lookup` returns `None`, the plugin leaves the message untouched, and the front door receives the bare word `apply` — no report, no options, no cluster, no manifest. It cannot open the PR the report just promised. The PR body lists this under **Not covered** as pre-existing. On `main` no report ever reached a human, so nobody ever replied; this PR makes that reply the normal case.

Fix: POST the completed report to `/v1/incidents` on the delivery path, keyed by the substituted `chat_id`/`thread_id` — the notifier holds all three values in the subscription row alongside `result`. If that is out of scope for this PR, drop the `To authorize: reply 'apply'` bullet from the template until it works.

k8s-operator/cmd/k8s-event-watcher/main.go
jayantid: **Dropping the direct entry removes the watcher's only credential that is guaranteed to work.**

`discoverClusterProfiles` never contacts an API server. `kubernetes.NewForConfig` at line 392 builds a client from a kubeconfig; it does not authenticate. So a profile reaches `profileNames` whether or not its credential can list events. The direct entry it displaces uses the pod's KSA under `kubeagents:minimal:*`, which grants `events` get/list/watch. The profile entry authenticates as the GSA through a `get-credentials` kubeconfig, bounded by IAM instead.

On a single-cluster install the management profile is now the only profile, so the watch set becomes one entry with an unverified credential. Deny that credential — the `custom` permission set without `roles/container.viewer`, or master authorized networks that exclude the pod's egress — and the reflector retries the initial list forever, `synced` stays 0, and `realMain` returns the error at line 852 after `initialSyncGrace`. The pod crashloops and no cluster is watched at all. Before this change the direct entry always synced, so `synced >= 1` held and one failing cluster degraded on its own, which is what the comment at line 826 intends. The comment here says dropping the direct entry "costs nothing — it is the same cluster, reached with a different credential"; the different credential is the cost.

Fix: drop the profile duplicate instead of the direct entry, and take the identity triple off the profile so the payload is still stamped with `project/location/cluster`.

```go
// The profile supplies project/location; the in-cluster client supplies
// reachability. Keep the one that cannot be denied by IAM.
if id, dup := profileIdentity[f.clusterName]; dup {
    clusters = removeProfile(clusters, id.Profile)
    tc := targetCluster{Name: f.clusterName, Profile: "direct",
        ProjectID: id.Project, Location: id.Location}
    // ... buildKubeClient(f), append tc
}
```

agents/platform/docs/session_management.md
jayantid, line 93: **Phase 3 of the sequence diagram now says a Cluster Agent opens the GitOps PR.**

This line renames the `Agent` participant to the Cluster Agent. Phase 3 was not updated with it and still reads `Agent->>Agent: Create branch, edit git manifests, open GitOps PR` at line 120. A Cluster Agent profile has no GitOps write path. `docs/site/src/content/docs/reference/security-and-iam.md:27` says so directly, and `_triage_task_body` says the apply is "carried out by the agent that holds the GitOps write path" (`agents/platform/scripts/session_kv_server.py:617`).

The diagram is the only one of those three places that names the actor, and it names one that cannot do the work. A reader debugging a failed `apply` from this doc goes looking at the wrong profile's toolset.

Fix: give Phase 3 its own participant.

```
    participant Fixer as Platform Agent (GitOps write path)
    ...
    Plugin->>Gateway: Spawn Fixer Agent with rewritten message
    Gateway->>Fixer: Inject context into conversation turn
    Fixer->>Fixer: Create branch, edit git manifests, open GitOps PR
```

agents/platform/scripts/session_kv_server.py
stalhaali: Done. I will fix it in a follow up PR. Heres an issue to track this:

https://github.com/gke-labs/kube-agents/issues/802

k8s-operator/cmd/k8s-event-watcher/main.go
stalhaali: Done

agents/platform/docs/session_management.md
stalhaali: Done

k8s-operator/cmd/k8s-event-watcher/main.go
```diff
@@ -863,39 +863,86 @@ func realMain(argv []string) error {
...
+		// profile would silently resume from an empty cache after an upgrade,
+		// which is one duplicate alert per incident still inside the window.
+		direct := targetCluster{Name: f.clusterName, Profile: "direct"}
+		// Matched on the bare name because that is all the direct entry has —
+		// --in-cluster reads no cluster_identity. Two same-named clusters in
+		// different locations would therefore match the wrong profile and stamp
+		// this one with the other's location; the log line below prints the
+		// identity being adopted so that shows up rather than being inferred.
+		if covered, dup := profileFor[f.clusterName]; dup {
+			clusters = removeProfile(clusters, covered.Profile)
```
jayantid: **The absorb matches on the bare cluster name, so a name collision drops a different cluster from the watch set.**

`profileFor` is keyed on `tc.Name` (line 915), the bare cluster name. `discoverClusterProfiles` keys its own dedup on the full `project/location/cluster` triple and says why at line 338: "two clusters called `prod` in different locations are two clusters, and both get their own profile." So two profiles can legitimately share a name, and `profileFor[f.clusterName]` keeps whichever one `os.ReadDir` returned last.

Take a fleet with `projA/us-central1/mgmt` — the management cluster, where the pod runs — and `projB/us-east1/mgmt`. `profileFor["mgmt"]` holds the projB entry. Line 935 removes the projB profile from the watch set. Line 936 stamps the direct entry with `projB/us-east1`. Three things follow:

- `projB/us-east1/mgmt` is no longer watched at all. Nothing warns. The log line at 937 reports a successful absorb.
- The management cluster is still watched twice, once through its own projA profile and once through the direct entry. That is the double alert this function exists to prevent.
- The direct entry's events are stamped `projB/us-east1`, so the triage session routes to the projB Cluster Agent, which cannot see the failing pod.

The comment at lines 929-933 anticipates the wrong-identity stamp. It does not cover the dropped profile or the surviving duplicate. Before this commit the collision was harmless: the dup branch only skipped the direct entry, so both profiles stayed in the watch set with correct identities.

Fix: absorb only when exactly one profile claims the name. The direct entry has no identity triple, so an ambiguous name cannot be resolved and must not absorb.

```go
nameCount := map[string]int{}
for _, tc := range discovered {
    nameCount[tc.Name]++
}
// ...
covered, dup := profileFor[f.clusterName]
if dup && nameCount[f.clusterName] > 1 {
    // Ambiguous: keep every profile and skip the direct entry rather than
    // silently stop watching one of them.
    log.Printf("k8s-event-watcher: %d profiles are named %q; cannot tell which is this pod's cluster, so watching them through their profiles and adding no direct entry",
        nameCount[f.clusterName], f.clusterName)
} else {
    if dup { /* existing absorb */ }
    /* build client, append direct */
}
```

k8s-operator/cmd/k8s-event-watcher/main.go
```diff
@@ -863,39 +863,86 @@ func realMain(argv []string) error {
...
+		// disk has this cluster's cache under that name. Renaming it to the
+		// profile would silently resume from an empty cache after an upgrade,
+		// which is one duplicate alert per incident still inside the window.
+		direct := targetCluster{Name: f.clusterName, Profile: "direct"}
+		// Matched on the bare name because that is all the direct entry has —
+		// --in-cluster reads no cluster_identity. Two same-named clusters in
+		// different locations would therefore match the wrong profile and stamp
+		// this one with the other's location; the log line below prints the
+		// identity being adopted so that shows up rather than being inferred.
+		if covered, dup := profileFor[f.clusterName]; dup {
```
jayantid: **The absorb matches on the bare cluster name, so a name collision stops watching the wrong cluster.**

Line 934 looks up `profileFor[f.clusterName]`, and line 915 keys that map on `tc.Name` alone. `discoverClusterProfiles` keys on the full `project/location/cluster` triple, and line 338 says why: two clusters named `prod` in different locations are two clusters. So two profiles can share a name, and `profileFor` keeps whichever one `os.ReadDir` returned last.

Take a fleet with `projA/us-central1/mgmt`, where the pod runs, and `projB/us-east1/mgmt`. Merging this breaks three things:

- `projB/us-east1/mgmt` stops being watched. Its events are never seen, and nothing warns — line 937 logs a successful absorb.
- `mgmt` is still watched twice, through projA's profile and through the direct entry. That is the duplicate alert this branch exists to prevent.
- Line 936 stamps the direct entry `projB/us-east1`, so triage routes to the projB Cluster Agent, which cannot see the failing pod.

The comment at 929-933 covers the mis-stamp. It does not cover the dropped cluster or the surviving duplicate, and both are new in this revision — before it, the dup branch only skipped the direct entry, so a collision left both profiles watched with correct identities.

Fix: absorb only when exactly one profile claims the name. The direct entry has no identity triple, so an ambiguous name cannot be resolved and must not absorb.

```go
nameCount := map[string]int{}
for _, tc := range discovered {
    nameCount[tc.Name]++
}
// ...
covered, dup := profileFor[f.clusterName]
if dup && nameCount[f.clusterName] > 1 {
    // Ambiguous: keep every profile and skip the direct entry rather than
    // silently stop watching one of them.
    log.Printf("k8s-event-watcher: %d profiles are named %q; cannot tell which is this pod's cluster, so watching them through their profiles and adding no direct entry",
        nameCount[f.clusterName], f.clusterName)
} else {
    if dup { /* existing absorb */ }
    /* build client, append direct */
}
```

stalhaali: done

jayantid (approved):

Issue #635, opened 2026-08-11
Speed up pull-kube-agents-smoke-test: quick wins to cut ~43min runs and unblock the queue


jayantid:
## Problem

`pull-kube-agents-smoke-test` runs with `max_concurrency: 1` and averages **42.8 min** (median ~42, n=20 recent runs). With a single slot, the second PR in flight waits ~45 min and the third waits ~90 min. Contributors are regularly queued for an hour before their job even starts.

There are two independent levers: **cut per-run wall clock**, and **remove the shared state that forces `max_concurrency: 1`**. This issue covers only the first — changes that are small diffs, need no isolation work, and pay off immediately even while the job stays serialized. (Raising `max_concurrency` is a larger piece of work and deserves its own issue; notes at the bottom.)

### A note on the numbers below

The prow artifact buckets aren't publicly readable, so the per-item savings are **estimates derived from reading the scripts**, not measurements. Step 0 fixes that cheaply — the scripts already print `✓ <step> finished in Ns` at every stage, so one `build-log.txt` from a SUCCESS run gives the exact split. Please re-derive the table from real numbers before prioritizing.

---

### 1. Bake a CI image instead of installing toolchain on every run

Every run currently starts with, from [`kube-agents-presubmits.yaml`](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml):

```bash
apt-get update && apt-get install -y --no-install-recommends make gettext-base golang-go git openssl ca-certificates curl
gcloud components install gke-gcloud-auth-plugin --quiet
curl -LsSf https://astral.sh/uv/install.sh | sh
curl ... https://get.opentofu.org/install-opentofu.sh ... --install-method standalone
```

That's identical work on every single run (`golang-go` alone pulls a few hundred MB), plus four network dependencies at job start that can each fail the run before any real testing happens.

**Fix:** publish a CI image with make/go/git/uv/opentofu/gke-gcloud-auth-plugin baked in; use it as the job image.

**Estimated saving: 3–5 min/run**, plus removal of four flake sources.

### 2. Collapse three Cloud Builds into one

`hack/ci-deploy.sh:85-93` runs three sequential `gcloud builds submit` calls. The important part isn't the source upload (only ~12M after `.gcloudignore`) — it's this, at `deploy/docker/Dockerfile:851`:

```dockerfile
FROM platform AS credential-proxy
```

[...]

PR #739, opened 2026-08-17
fix(ci): isolate presubmit image tags to prevent cache pollution

lapis2002:
# Pull Request
## Summary

This PR isolates presubmit container image tags in Cloud Build so that presubmit jobs only push PR-scoped immutable tags (`pr-<num>-<sha>`) rather than overwriting `:latest` in the evaluation project's Artifact Registry. It also configures presubmit builds to warm their Docker layer cache directly from the canonical postsubmit image (`us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest`), ensuring deterministic cache hits across runs and preventing cross-PR cache pollution.

## Why This Change
Presubmit Cloud Builds previously tagged and pushed `:latest` into the evaluation project's Artifact Registry (`us-central1-docker.pkg.dev/${PROJECT_ID}/...:latest`).

This created two issues:
1. **Cross-PR Cache Poisoning & Race Conditions:** When multiple PRs ran, they constantly raced to overwrite `:latest`. PR B would warm its cache from PR A's experimental image, causing non-deterministic cache misses or stale layer invalidations.
2. **Blocker for Parallel CI:** Concurrent runs with `max_concurrency > 1` cannot share a single mutable `:latest` tag without colliding.

Addresses item 3 of #635 ("Stop pushing `:latest` from presubmits").

---
## What Changed

**Files:**
- `deploy/docker/cloudbuild-ci.yaml`
- `hack/ci-deploy.sh`

**Details:**
- **Isolate Presubmit Tags:** `deploy/docker/cloudbuild-ci.yaml` now pushes **only PR-scoped immutable tags** (`platform-agent:${TAG}`, `credential-proxy:${TAG}`, `kube-agents-operator:${TAG}`). It no longer tags or pushes `:latest`.
- **Warm Cache from Canonical Prow Image:** `hack/ci-deploy.sh` passes `_CACHE_IMAGE=us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest`. This pulls cache from the postsubmit image that is only updated on merges to `main`, ensuring deterministic cache hits and immunity to PR races.
- **Resilient Cache Pull with Fallback:** In `warm-cache`, retries `docker pull "$_CACHE_IMAGE"` up to 3 times with 5s backoff, logs a warning on cache miss, and falls back to a cold build (unless `_REQUIRE_CACHE=true` is set).
- **Opt-in Strict Failure (`_REQUIRE_CACHE`)**: Added `_REQUIRE_CACHE` substitution (default false) wired to `${REQUIRE_CACHE:-false}` in `hack/ci-deploy.sh`, allowing CI operators to enforce hard failures on cache misses when desired.
- BuildKit Registry Cache Resolution: Passes `--cache-from "$_CACHE_IMAGE"` via escaped array `"$${CACHE_ARGS[@]}"` whenever `_CACHE_IMAGE` is non-empty, letting BuildKit resolve cache manifests directly from the registry.

---
## Why This Matters
- **Deterministic Builds:** Every presubmit warms its cache from the stable `main` branch HEAD rather than whichever PR ran last.
- **Artifact Registry Tag Collisions:** Concurrent runs cannot share a single mutable :latest tag without colliding.

---

### Local Validation
- Verified syntax and build substitutions.

### Live validation

* **Job:** `pull-kube-agents-smoke-test`:
* All 3 run ([2089480400094105600](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/739/pull-kube-agents-smoke-test/2089480400094105600) (Cloud Build: 7m 23s), [2089527924469272576](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/739/pull-kube-agents-smoke-test/2089527924469272576) (Cloud Build: 8m 59s), [2089553314248134656](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/739/pull-kube-agents-smoke-test/2089553314248134656) (Cloud Build: 9m 55s)) completed with only 3 PR tags pushed.
  1. us-central1-docker.pkg.dev/kube-agents-evals/kube-agents/platform-agent:pr-739-7f59c97
  2. us-central1-docker.pkg.dev/kube-agents-evals/kube-agents/credential-proxy:pr-739-7f59c97
  3. us-central1-docker.pkg.dev/kube-agents-evals/kube-agents/kube-agents-operator:pr-739-7f59c97
  
  * *from Cloud Build log of 1st run:*
    ```text
    Step #0 - "warm-cache": Pulling cache image: us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest
    Step #0 - "warm-cache": Status: Downloaded newer image for us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest
    Step #0 - "warm-cache": ✓ Cache image pulled successfully.
    Step #1 - "platform": #5 importing cache manifest from us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest
    PUSH: Pushing us-central1-docker.pkg.dev/kube-agents-evals/kube-agents/platform-agent:pr-739-7f59c97 (+2 more)
    ```
  

* Verified that no CI or deployment script consumes the eval-registry `:latest` tag: `hack/ci-deploy.sh` and `hack/ci-eval-pr.sh` deploy via explicit `$TAG` (`pr-${PULL_NUMBER}-${PULL_SHA_SHORT}`) and `$IMG`. Stopping `:latest` pushes in `kube-agents-evals` causes no downstream breakage.
* All ephemeral namespaces, pods, and secrets created during testing were cleaned up by `./hack/ci-teardown.sh`.

## Self-Review
* **Cache Availability & IAM:** Checked for potential permission issues when Cloud Build in `kube-agents-evals` pulls the cache image from `kube-agents-prow`. Verified that `roles/artifactregistry.
reader` is granted to the Cloud Build service account (`1042377440668@cloudbuild.gserviceaccount.com`). Live logs confirm successful pull.
* **Cache Pull & Fallback:** Verified that `warm-cache` skips cleanly if `_CACHE_IMAGE` is empty, retries 3x on transient pull failures, and falls back to a cold build without passing broken `--
  cache-from` flags unless `_REQUIRE_CACHE="true"`
* **Tag Isolation:** Inspected the `images:` block and `docker build` arguments in `cloudbuild-ci.yaml` to confirm no step tags or pushes `:latest`. Confirmed via Cloud Build push logs.

## Risk & Rollout
Low risk. This change touches only the presubmit Cloud Build configuration and CI deployment script. Runtime operator code and production GitHub Actions workflows are completely untouched. If any issue arises, reverting the commit restores the previous build substitution parameters.

### DIFF
--- a/deploy/docker/cloudbuild-ci.yaml
+++ b/deploy/docker/cloudbuild-ci.yaml
@@ -15,42 +15,70 @@
 # platform, they are already in the daemon's cache and only the delta runs
 # (gke-labs/kube-agents#635).
 steps:
-  # Only platform's cache is seeded. credential-proxy's own :latest is not
-  # pulled: the step below builds directly on top of the platform image this
-  # build just produced, so its five remaining instructions are cheap, and
-  # pulling the whole image to save them would cost far more than it saves.
+  # Only platform's cache is seeded from the canonical postsubmit image
+  # maintained by .github/workflows/docker-publish-gcp.yml in kube-agents-prow.
+  # credential-proxy's own cache is not pulled: the step below builds directly
+  # on top of the platform image this build just produced, so its five
+  # remaining instructions are cheap, and pulling the whole image to save them
+  # would cost far more than it saves.
   - id: warm-cache
     name: "gcr.io/cloud-builders/docker"
     entrypoint: "bash"
     args:
       - "-c"
-      - "docker pull $_PLATFORM_LATEST || true"
+      - |
+        echo "Pulling cache image: $_CACHE_IMAGE"
+        max_attempts=3
+        attempt=1
+        pulled=false
+        until [ "$attempt" -gt "$max_attempts" ]; do
+          if docker pull "$_CACHE_IMAGE"; then
+            pulled=true
+            echo "✓ Cache image pulled successfully."
+            break
+          fi
+          echo "⚠ Attempt $attempt/$max_attempts failed to pull $_CACHE_IMAGE."
+          if [ "$attempt" -lt "$max_attempts" ]; then
+            echo "Retrying in 5s..."
+            sleep 5
+          fi
+          attempt=$((attempt + 1))
+        done
+
+        if [ "$pulled" != "true" ]; then
+          echo "WARNING: CACHE MISS — could not pull $_CACHE_IMAGE after $max_attempts attempts."
+          echo "Check that docker-publish-gcp.yml is green on main and that Cloud Build"
+          echo "holds roles/artifactregistry.reader on kube-agents-prow."
+          if [ "$_REQUIRE_CACHE" = "true" ]; then
+            echo "ERROR: _REQUIRE_CACHE is set to true. Failing build."
+            exit 1
+          fi
+          echo "Building cold without cache."
+        fi
 
   - id: platform
     name: "gcr.io/cloud-builders/docker"
+    entrypoint: "bash"
     env:
       - "DOCKER_BUILDKIT=1"
     args:
-      - "build"
-      # Deployment targets are always amd64 GKE nodes; the multi-arch bases
-      # otherwise resolve to the build host (#560).
-      - "--platform"
-      - "linux/amd64"
-      - "-t"
-      - "$_PLATFORM_URI"
-      - "-t"
-      - "$_PLATFORM_LATEST"
-      - "--cache-from"
-      - "$_PLATFORM_LATEST"
-      - "-f"
-      - "deploy/docker/Dockerfile"
-      - "--target"
-      - "platform"
-      - "--build-arg"
-      - "BUILDKIT_INLINE_CACHE=1"
-      - "--build-arg"
-      - "HERMES_AGENT_TAG=$_HERMES_AGENT_TAG"
-      - "."
+      - "-c"
+      - |
+        CACHE_ARGS=()
+        if [ -n "$_CACHE_IMAGE" ]; then
+          CACHE_ARGS=(--cache-from "$_CACHE_IMAGE")
+        fi
+
+        # Deployment targets are always amd64 GKE nodes; the multi-arch bases
+        # otherwise resolve to the build host (#560).
+        docker build \
+          --platform linux/amd64 \
+          -t "$_PLATFORM_URI" \
+          "$${CACHE_ARGS[@]}" \
+          -f deploy/docker/Dockerfile \
+          --target platform \
+          --build-arg HERMES_AGENT_TAG="$_HERMES_AGENT_TAG" \
+          .
 
   # No --cache-from and no waitFor: this runs straight after platform on the
   # same worker, which is the entire point of collapsing the submits.
@@ -64,14 +92,10 @@ steps:
       - "linux/amd64"
       - "-t"
       - "$_PROXY_URI"
-      - "-t"
-      - "$_PROXY_LATEST"
       - "-f"
       - "deploy/docker/Dockerfile"
       - "--target"
       - "credential-proxy"
-      - "--build-arg"
-      - "BUILDKIT_INLINE_CACHE=1"
       - "--build-arg"
       - "HERMES_AGENT_TAG=$_HERMES_AGENT_TAG"
       - "."
@@ -94,11 +118,11 @@ steps:
       - "k8s-operator/Dockerfile"
       - "k8s-operator"
 
+# Presubmits push only PR-scoped immutable tags, avoiding cross-PR cache pollution
+# (gke-labs/kube-agents#635).
 images:
   - "$_PLATFORM_URI"
-  - "$_PLATFORM_LATEST"
   - "$_PROXY_URI"
-  - "$_PROXY_LATEST"
   - "$_OPERATOR_URI"
 
 # No `options.machineType` here on purpose. The operator step only genuinely
@@ -115,8 +139,8 @@ timeout: 3600s
 
 substitutions:
   _PLATFORM_URI: ""
-  _PLATFORM_LATEST: ""
   _PROXY_URI: ""
-  _PROXY_LATEST: ""
   _OPERATOR_URI: ""
+  _CACHE_IMAGE: ""
   _HERMES_AGENT_TAG: ""
+  _REQUIRE_CACHE: "false"
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -98,8 +98,12 @@ echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Building Container Images (platform
 # them as consecutive steps on one worker lets the second reuse the first's
 # layers instead of rebuilding the chain on a cold daemon; the operator build
 # runs alongside them. See the header of cloudbuild-ci.yaml, and #635.
+# Set REQUIRE_CACHE=true in the job environment to fail the build on a cache
+# miss instead of cold-building. Default false so a broken cache source cannot
+# block the PR that fixes it.
+export CACHE_IMAGE="${CACHE_IMAGE:-us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest}"
 gcloud builds submit --config="deploy/docker/cloudbuild-ci.yaml" \
-  --substitutions="_PLATFORM_URI=${AR_REPO}/platform-agent:${TAG},_PLATFORM_LATEST=${AR_REPO}/platform-agent:latest,_PROXY_URI=${AR_REPO}/credential-proxy:${TAG},_PROXY_LATEST=${AR_REPO}/credential-proxy:latest,_OPERATOR_URI=${AR_REPO}/kube-agents-operator:${TAG},_HERMES_AGENT_TAG=${HERMES_AGENT_TAG}" \
+  --substitutions="_PLATFORM_URI=${AR_REPO}/platform-agent:${TAG},_PROXY_URI=${AR_REPO}/credential-proxy:${TAG},_OPERATOR_URI=${AR_REPO}/kube-agents-operator:${TAG},_CACHE_IMAGE=${CACHE_IMAGE},_HERMES_AGENT_TAG=${HERMES_AGENT_TAG},_REQUIRE_CACHE=${REQUIRE_CACHE:-false}" \
   --project="${PROJECT_ID}" "${BUILD_WORKER_ARGS[@]}" --quiet .
 echo "✓ Container image builds finished in $((SECONDS - STEP_START))s"
 


deploy/docker/cloudbuild-ci.yaml
```diff
@@ -15,16 +15,31 @@
 # platform, they are already in the daemon's cache and only the delta runs
```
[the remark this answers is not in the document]
lapis2002: Updated PR description.

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -15,16 +15,31 @@
...
+        if [ -z "$_CACHE_IMAGE" ]; then
+          echo "ERROR: _CACHE_IMAGE substitution is required but was empty."
+          exit 1
+        fi
+        echo "Pulling cache image: $_CACHE_IMAGE"
+        if docker pull "$_CACHE_IMAGE"; then
+          echo "✓ Cache image pulled successfully."
+        else
+          echo "⚠ Cache pull failed; proceeding with a cold build."
+        fi
```
jayantid: The two branches here are guarding opposite ends of the likelihood scale, and I think they are the wrong way round.

The `-z` check cannot fire. `hack/ci-deploy.sh` is the only caller of this config (I grepped — nothing else references `cloudbuild-ci.yaml`), and it passes `_CACHE_IMAGE` as a hardcoded literal, so the empty case is unreachable in practice.

The failure that *can* happen is `docker pull` failing, and that one warns and carries on. Concretely: someone drops `roles/artifactregistry.reader` for the evals Cloud Build SA on `kube-agents-prow`, or the postsubmit tag gets deleted, or the registry has a blip. Every presubmit then cold-builds, this step still reports success, and the only symptom is that the job got slower — which nothing in CI measures. The single thing this PR exists to deliver would stop working silently.

Self-Review Pass 2 describes this guard as covering silent cache misses, but as written it covers the one that is not silent and lets the silent one through.

Two ways to close it, and the choice is a real tradeoff rather than an obvious call:

**Make the pull fatal** — right if you would rather see the problem than absorb it, since the cache source is now a hard cross-project dependency rather than a local optimization:

```bash
echo "Pulling cache image: $_CACHE_IMAGE"
if ! docker pull "$_CACHE_IMAGE"; then
  echo "ERROR: could not pull $_CACHE_IMAGE."
  echo "The presubmit cache source is a hard dependency (cross-project read on"
  echo "kube-agents-prow). Continuing would silently cold-build every PR."
  exit 1
fi
echo "✓ Cache image pulled successfully."
```

The cost is that a transient registry blip now reddens the smoke test instead of just slowing it.

**Or keep it soft but assert the hit later** — grep the platform step output for `importing cache manifest` / `CACHED` and fail if absent. More faithful to what you actually care about, since a successful pull does not by itself prove BuildKit reused anything: `✓ Cache image pulled successfully` reads like a cache-hit confirmation but only proves the pull.

Either is fine by me. What I would avoid is leaving the only signal as a `⚠` in a log nobody reads on a green step.

lapis2002: I've updated the `warm-cache` step to **retry-then-fail** pattern:
- It retries `docker pull "$_CACHE_IMAGE"` up to 3 times (with a 5s backoff) to absorb any transient network/registry hiccups.
- If all attempts fail, it exits 1 with an explicit diagnostic error indicating the cross-project IAM/image dependency is broken, rather than silently degrading into slow cold builds.

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -15,16 +15,36 @@
...
+      - |
+        echo "Pulling cache image: $_CACHE_IMAGE"
+        max_attempts=3
+        attempt=1
+        until docker pull "$_CACHE_IMAGE"; do
+          if [ "$attempt" -ge "$max_attempts" ]; then
+            echo "ERROR: Failed to pull $_CACHE_IMAGE after $max_attempts attempts."
+            echo "The presubmit cache source is a required cross-project dependency on"
+            echo "kube-agents-prow. Failing fast to prevent silent reversion to cold builds."
+            exit 1
```
jayantid: **Blocker: this `exit 1` is a bigger change than it looks.**

Today a missing cache means a slow build. After this, it means `pull-kube-agents-smoke-test` fails at step 0 — before a single test runs — for every open PR in the repo at once. The triggers are all outside this repo's control: `docker-publish-gcp.yml` fails or is skipped on a main push, the WIF binding or `github-actions@kube-agents-prow` rotates, an Artifact Registry cleanup policy expires `:latest`, or the cross-project IAM grant gets dropped in a project cleanup.

The part that makes this a blocker rather than a nit: **the revert would be blocked by the same gate.** Any fix or revert PR has to pass this presubmit, which runs the same `warm-cache` step against the same broken image. Recovery needs an admin merge-override or an out-of-band change in the prow project. On a ~43-min job with one concurrency slot, that's the worst place to put an unrecoverable gate.

"Failing fast to prevent silent reversion to cold builds" is a real problem with the wrong remedy — silence is the defect, fatality isn't the only cure. Keep the retry loop, make the failure loud instead:

```bash
if ! docker pull "$_CACHE_IMAGE"; then
  echo "::warning::CACHE MISS — could not pull $_CACHE_IMAGE after $max_attempts attempts."
  echo "Building cold. Check docker-publish-gcp.yml is green on main and that Cloud Build"
  echo "in kube-agents-evals still holds roles/artifactregistry.reader on kube-agents-prow."
fi
```

If a hard gate is genuinely wanted, put it behind an opt-in (`_REQUIRE_CACHE`, default `false`) so disabling it is a substitution change rather than a code change that has to survive the gate it's disabling.

lapis2002: I've updated it to emit a loud warning and fall back to building cold instead of failing, and added `_REQUIRE_CACHE: "false"` as an opt-in escape hatch.

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -15,42 +15,75 @@
...
-      - "platform"
-      - "--build-arg"
-      - "BUILDKIT_INLINE_CACHE=1"
-      - "--build-arg"
-      - "HERMES_AGENT_TAG=$_HERMES_AGENT_TAG"
-      - "."
+      - "-c"
+      - |
+        CACHE_ARGS=()
+        if [ -n "$_CACHE_IMAGE" ] && docker image inspect "$_CACHE_IMAGE" >/dev/null 2>&1; then
```
jayantid: `cloudbuild-ci.yaml:73` gates `--cache-from` on `docker image inspect`, which reads the local daemon. BuildKit resolves `--cache-from` from the registry instead. Your own build log shows it as its own vertex: `#5 importing cache manifest from us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest`.

So the gate keys the flag on the wrong signal. Say Artifact Registry returns three transient 5xx to `docker pull`. `warm-cache` gives up, the platform step omits `--cache-from`, and the presubmit builds fully cold. BuildKit's own registry import would have run at that point and could have succeeded. The gate can only lose cache hits, never gain one. The previous config passed `--cache-from` unconditionally after `docker pull || true` and needed no such guard.

Drop the `docker image inspect` half:

```bash
CACHE_ARGS=()
if [ -n "$_CACHE_IMAGE" ]; then
  CACHE_ARGS=(--cache-from "$_CACHE_IMAGE")
fi
```

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -15,42 +15,75 @@
...
+  # remaining instructions are cheap, and pulling the whole image to save them
+  # would cost far more than it saves.
   - id: warm-cache
     name: "gcr.io/cloud-builders/docker"
     entrypoint: "bash"
     args:
       - "-c"
-      - "docker pull $_PLATFORM_LATEST || true"
+      - |
+        if [ -z "$_CACHE_IMAGE" ]; then
```
jayantid: `cloudbuild-ci.yaml:30` echoes `Skipping cache pull` and exits 0. The description's What Changed section calls this a "Fail-Fast Cache Guard" that "fails fast if `_CACHE_IMAGE` is empty". It does the opposite. The Self-Review section of the same description says it "skips cleanly", so the description contradicts itself.

A reviewer reading only the description approves a guard that does not exist. A reader of the code sees four lines that cannot run.

The branch is unreachable. `hack/ci-deploy.sh` is the only caller of this config, and `hack/ci-deploy.sh:97` defaults `CACHE_IMAGE` to a hardcoded literal, so `_CACHE_IMAGE` is never empty. I raised this on the previous revision and it is unchanged.

Delete lines 30-33 and correct the description. If the branch stays as future-proofing, say so in the description rather than calling it a fail-fast guard.

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -94,11 +123,11 @@ steps:
       - "k8s-operator/Dockerfile"
       - "k8s-operator"
 
+# Presubmits push only PR-scoped immutable tags, avoiding cross-PR cache pollution
```
jayantid, line 121: This is about the PR description, anchored on a changed line for want of anywhere better.

The description lists "Blocker for Parallel CI" as one of the two problems this fixes. Image tags are not the blocker. `hack/ci-env.sh:12-17` pins every presubmit to one cluster, `platform-agent-host`, and one namespace, `kubeagents-system`. Both are hardcoded `export`s with no per-PR fallback, so a Prow job cannot override them. `hack/ci-eval-pr.sh:38-41` pins the bench cluster to `test-cluster` in `us-west4-a`.

If someone reads this PR as clearing the way and raises `max_concurrency` above 1, the second run's `provision_08_deploy_platform_agent.sh` overwrites the first run's Deployment in `kubeagents-system` while the first is still under evaluation, and both runs drive the same `test-cluster`. That is two red smoke tests with no obvious cause.

No code change needed. Scope the claim to Artifact Registry, which is what this PR does make parallel-safe, and say that the shared namespace and bench cluster keep presubmits serial.

hack/ci-deploy.sh
```diff
@@ -94,8 +94,9 @@ echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Building Container Images (platform
 # them as consecutive steps on one worker lets the second reuse the first's
 # layers instead of rebuilding the chain on a cold daemon; the operator build
 # runs alongside them. See the header of cloudbuild-ci.yaml, and #635.
+export CACHE_IMAGE="${CACHE_IMAGE:-us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest}"
 gcloud builds submit --config="deploy/docker/cloudbuild-ci.yaml" \
-  --substitutions="_PLATFORM_URI=${AR_REPO}/platform-agent:${TAG},_PLATFORM_LATEST=${AR_REPO}/platform-agent:latest,_PROXY_URI=${AR_REPO}/credential-proxy:${TAG},_PROXY_LATEST=${AR_REPO}/credential-proxy:latest,_OPERATOR_URI=${AR_REPO}/kube-agents-operator:${TAG},_HERMES_AGENT_TAG=${HERMES_AGENT_TAG}" \
+  --substitutions="_PLATFORM_URI=${AR_REPO}/platform-agent:${TAG},_PROXY_URI=${AR_REPO}/credential-proxy:${TAG},_OPERATOR_URI=${AR_REPO}/kube-agents-operator:${TAG},_CACHE_IMAGE=${CACHE_IMAGE},_HERMES_AGENT_TAG=${HERMES_AGENT_TAG},_REQUIRE_CACHE=${REQUIRE_CACHE:-false}" \
```
jayantid, line 106: `REQUIRE_CACHE` is read on this line and set nowhere. It is not in `hack/ci-env.sh`, not in any other script, and not named in any comment or doc in the tree.

The knob exists so an operator can turn a silent cold build into a hard failure. Nobody will find it. The next person debugging a slow presubmit opens `cloudbuild-ci.yaml`, sees `_REQUIRE_CACHE: "false"` at line 151, and has no way to learn it is settable from the job environment.

Add one comment above line 97:

```bash
# Set REQUIRE_CACHE=true in the job environment to fail the build on a cache
# miss instead of cold-building. Default false so a broken cache source cannot
# block the PR that fixes it.
export CACHE_IMAGE="${CACHE_IMAGE:-us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest}"
```

deploy/docker/cloudbuild-ci.yaml
lapis2002: Updated the PR description to scope the parallel-safe to Artifact Registry only.

deploy/docker/cloudbuild-ci.yaml
lapis2002: Updated.

deploy/docker/cloudbuild-ci.yaml
lapis2002: Updated.

hack/ci-deploy.sh
lapis2002: Added the comment as suggested.

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -94,11 +118,11 @@ steps:
...
```
jayantid: Thanks for rescoping *Why This Matters*. *Why This Change* still lists "**Blocker for Parallel CI:** Concurrent runs with `max_concurrency > 1` cannot share a single mutable `:latest` tag without colliding" as one of the two problems this PR fixes. Image tags are not the blocker.

The repo states the real constraint. `bench/tf/modules/cluster/gke/main.tf:46-48`:

> This is safe only because the smoke test runs at max_concurrency 1: nothing else can own a cluster with this name. That assumption is load-bearing -- if the job is ever allowed to run concurrently, this deletes a live run's cluster.

`hack/ci-env.sh:18-22` also pins every presubmit to one cluster, `platform-agent-host`, and one namespace, `kubeagents-system`. Both are hardcoded `export`s with no per-PR fallback.

If someone reads this PR as clearing the way and raises `max_concurrency`, run B's `reap_orphans` provisioner deletes run A's GKE cluster mid-evaluation. That is worse than the tag collision this PR fixes.

No code change needed. Add one sentence to *Why This Change* saying presubmits stay serial for reasons this PR does not touch, and name `reap_orphans` as the reason.

deploy/docker/cloudbuild-ci.yaml
```diff
@@ -15,42 +15,70 @@
...
+  # remaining instructions are cheap, and pulling the whole image to save them
+  # would cost far more than it saves.
   - id: warm-cache
     name: "gcr.io/cloud-builders/docker"
     entrypoint: "bash"
     args:
       - "-c"
-      - "docker pull $_PLATFORM_LATEST || true"
+      - |
+        echo "Pulling cache image: $_CACHE_IMAGE"
```
jayantid: The Self-Review bullet *Cache Pull & Fallback* describes code this revision removed. Two of its claims are now false, so the description records a verification pass that cannot have run against this tree.

Claim one: `warm-cache` "skips cleanly if `_CACHE_IMAGE` is empty". The early-exit branch is gone. With `_CACHE_IMAGE` empty, `cloudbuild-ci.yaml:35` runs `docker pull ""`, which fails on `invalid reference format`, three times with two 5s sleeps, then prints `WARNING: CACHE MISS`. Under `_REQUIRE_CACHE=true` it exits 1. That is the opposite of skipping cleanly.

Claim two: it "falls back to a cold build without passing broken `--cache-from` flags". `cloudbuild-ci.yaml:68` keys `--cache-from` on `_CACHE_IMAGE` being non-empty, not on the pull having succeeded, so the flag is still passed after a failed pull. The code is right — BuildKit imports the manifest from the registry and a failed import is non-fatal. Only the description is stale.

Rewrite the bullet to match the code:

> **Cache Pull & Fallback:** `warm-cache` retries `docker pull` 3x on transient failures, then warns and cold-builds unless `_REQUIRE_CACHE=true`. `--cache-from` is passed whenever `_CACHE_IMAGE` is set, so BuildKit still attempts its own registry import even when the warm pull failed.

hack/ci-deploy.sh
```diff
@@ -98,8 +98,12 @@ echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Building Container Images (platform
 # them as consecutive steps on one worker lets the second reuse the first's
 # layers instead of rebuilding the chain on a cold daemon; the operator build
 # runs alongside them. See the header of cloudbuild-ci.yaml, and #635.
+# Set REQUIRE_CACHE=true in the job environment to fail the build on a cache
+# miss instead of cold-building. Default false so a broken cache source cannot
+# block the PR that fixes it.
+export CACHE_IMAGE="${CACHE_IMAGE:-us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest}"
```
jayantid: This line makes a cross-project Artifact Registry read a hard dependency of every presubmit. The IAM grant behind it is recorded nowhere in the tree. Grepping for `artifactregistry.reader`, the only hit outside this PR is `bench/tf/modules/cluster/gke/main.tf:155-159`, which grants the GKE node service account read on the evals project. Nothing grants the evals Cloud Build service account read on `kube-agents-prow`.

`cloudbuild-ci.yaml:51` tells the operator to check that "Cloud Build holds roles/artifactregistry.reader on kube-agents-prow". It names the role and the project but not the principal. Someone debugging a cache miss gets the symptom and no way to learn which identity to grant. The service account appears in this PR description and nowhere else, and a build-log reader does not have the PR.

Put the principal in the tree, above this export:

```bash
# Cross-project read: kube-agents-evals' Cloud Build SA needs
# roles/artifactregistry.reader on kube-agents-prow. Granted out of band as:
#   gcloud artifacts repositories add-iam-policy-binding kube-agents \
#     --location=us --project=kube-agents-prow \
#     --member=serviceAccount:1042377440668@cloudbuild.gserviceaccount.com \
#     --role=roles/artifactregistry.reader
export CACHE_IMAGE="${CACHE_IMAGE:-us-docker.pkg.dev/kube-agents-prow/kube-agents/platform-agent:latest}"
```

jayantid (approved):

PR #773, opened 2026-08-18
ci(deps): group Dependabot updates, and stop the tests masking an incompatible mcp

bradhoekstra:
## Summary

Dependabot opened 20 pull requests in three minutes today, five of them duplicate pairs of each other. This gives every ecosystem in `.github/dependabot.yml` a catch-all group so it batches into a single pull request, and names its directories explicitly instead of globbing. It also restores the `mcp<2` constraint in `requirements-test.txt` that one of those 20 widened to `mcp<3` against the comment sitting two lines above it.

Restoring that constraint raised the question of why nothing failed when it was widened. It turns out the four test modules that import `mcp.server.fastmcp` all wrap it in `try: real / except: stub`, so an `mcp` that is installed and incompatible was indistinguishable from one that is absent — the `except` swallowed the `ImportError` and the suite went green having exercised a `SimpleNamespace`. Those guards now stub only when no `mcp` distribution is installed, and a new contract test fails by name rather than by side effect (#800).

## Why This Change

#492 expanded the config from two ecosystems to five at 13:10:07 today. Dependabot's first run against it began 60 seconds later and opened #750–#769.

Two things produced that. First, the `pip` entry listed both `/` and `/**/*`. `requirements-test.txt` at the root begins with `-r admin_console/requirements.txt` and pip's parser follows the include, so the root job edits that file — and so does the `/admin_console` job the glob matches. #763 and #764 are the same httpx bump written twice; #757/#765, #758/#767, #760/#768 and #762/#766 repeat it, five duplicate pairs out of twenty. Both members of each pair merged.

Second, five ecosystems with `open-pull-requests-limit: 5` and no `groups` is a standing ceiling of 25 open Dependabot pull requests, each with its own CI run, its own bot review, and its own merge.

Left alone, every weekly run repeats both.

## What Changed

- `.github/dependabot.yml` — five entries, each with a catch-all group, each naming its directories.
  - `pip` → `/` and `/tests/e2e`. `/` already reaches `admin_console/requirements.txt` through the `-r` include, which is what made naming `/admin_console` a duplicate rather than extra coverage. `/bench` stays out: `bench/pyproject.toml` pins `devops-bench` to a git SHA its own comment says to bump deliberately.
  - `gomod` → `/k8s-operator`, the only `go.mod` in the tree. `npm` → `/docs/site`.
  - `docker` → `/agents/platform/skills/gke-app-onboarding/assets` and `/tests/e2e/operator/templates/plugin_src`, the only two directories with a literal `FROM` pin. Every other Dockerfile is beyond the updater's reach rather than excluded: the images this repository builds interpolate two ARGs each (`FROM ${GOLANG_IMAGE}:${GOLANG_VERSION}`, `FROM ${ENVOY_IMAGE}:${ENVOY_VERSION}`), which the Docker updater cannot resolve, and the `agentplugins` images are `FROM scratch`. `images.json` and `make mirror-images` own the ARG-interpolated pins.
  - An `ignore` rule holding `mcp` below 2.0, written as `versions: [">=2"]`.
  - The comments in the file explain why each directory is what it is, because the next person to reach for `/**/*` will otherwise recreate #763.
- `requirements-test.txt` — `mcp<3` back to `mcp<2`, and the troubleshooting comment that #751 rewrote along with it. The header comment also claimed two readers and named `python-tests.yml` as the CI one, when `k8s-operator-test.yml` installs the file too; all three are now named, with which pins its Python and which takes the runner's default. The `mcp` entry says what the ceiling bounds — the package, since 2.x deletes `mcp.server.fastmcp` in favour of `mcp.server.mcpserver.MCPServer`, and not the interpreter, since 1.x and 2.x declare the same `requires-python >=3.10` and ship one pure-Python wheel. It also says what holds the ceiling, which is not our three scripts: `mcp` reaches `/opt/hermes/.venv` from the digest-pinned Hermes base in `tags.env`, that base ships 1.28.1, and hermes-agent's own code imports the module 2.x deletes, so the ceiling comes off when the base image ships 2.x. The header's "these stand in for what the image gets transitively" claim is now checked rather than assumed for `mcp`, and hedged where it does not hold: the two environments agree on the major line and nothing more — different patch versions, Python 3.12 in CI against 3.13 in the image. The comment above the Dependabot `ignore` rule is retargeted the same way.
- `agents/platform/scripts/test_agent_common_server.py`, `test_platform_mcp_server.py`, `test_session_kv_server.py` and `agents/chat/scripts/test_router_server.py` — the four fallbacks now stub only when no `mcp` distribution is installed, which is the bare-checkout case they exist for, and re-raise otherwise. `test_router_server.py` additionally stubs per module rather than by bulk `sys.modules.update`, which was overwriting real, importable `mcp` and `pydantic` with stubs. That is defensive rather than a bug fixed: discovery runs one process per directory and that module sorts last in `agents/chat/scripts`, so nothing there was contaminated. `test_agent_common_server.py` is the file where a leaked stub does reach a later module — it sorts first in `agents/platform/scripts`, and `test_session_kv_server.py`, the seventeenth, imports `fastapi`.
  - The presence check asks `importlib.metadata.distribution` and not `importlib.util.find_spec`. `find_spec` consults `sys.modules` first and reads `__spec__` off whatever it finds, which is `None` on these `ModuleType` stubs, and raises `ValueError` instead of answering — and `unittest discover` imports each directory into one process in alphabetical order, so a stub left by an earlier module is exactly what a later check sees. The adversarial pass caught this; see Self-Review.
- `agents/platform/scripts/test_mcp_package_contract.py` (new) — asserts the installed `mcp` exposes `FastMCP`, and imports each of `agent_common_server`, `platform_mcp_server` and `router_server` in a **subprocess**. The subprocess is the point: within the discovery process a sibling module has already put stubs in `sys.modules`, so an in-process import would prove only that something else had faked what it needed. Skipped in full when no `mcp` is installed, so a bare checkout still runs. It sits under `agents/platform/scripts` because that is where the Makefile's `agents/*/scripts/test_*.py` glob finds it, and covers the chat module through the subprocess list. It also holds the single copy of the `importlib.metadata`-not-`find_spec` reasoning; the four guards point at it in two lines each, so lifting the ceiling in #800 changes one place rather than five.
- `docs/designs/design_537148738.md` — the design of record for `.github/dependabot.yml`, which still specified the globbed config and restated globbing in §4 as a standing governance rule. §3.4 and rule 5 now describe the named directories and why each is what it is. Without this the doc told the next person to reach for `/**/*` while the config's own header comment told them that is what opened #763.

Worst case is now seven open Dependabot pull requests a week rather than 25: one per ecosystem per directory, since a group without `group-by: dependency-name` still splits across directories — one each for `github-actions`, `gomod` and `npm`, two for `pip`, two for `docker`.

## Context

- Caused by #492; #763 and #764 are the visible symptom.
- #751 is the `mcp` widening this reverts. Its tests passed because both modules that touch the import stub `mcp.server.fastmcp` (`agents/platform/scripts/test_agent_common_server.py` and `agents/chat/scripts/test_router_server.py`), so nothing exercised the real one.
- Five singleton pull requests from that run are still open — #750, #752, #753, #756, #761. Dependabot does not regroup retroactively, so they stay singletons whatever happens here. Merging the three Go ones as a set and letting the next grouped run re-propose the two `docs/site` ones is the tidier end state, but that is a call for whoever merges them, not this PR.
- The `docker` entry started this branch removed and is back, narrowed to two directories. An earlier revision dropped the ecosystem on the grounds that every `FROM` interpolates unresolvable ARGs; that is true of the images this repository builds, not of the tree. `agents/platform/skills/gke-app-onboarding/assets/Dockerfile` is a literal `node:18-slim`, past end of life, and `deploy/docker/Dockerfile:1188` copies `agents/platform/skills/` wholesale into the published platform image, so it ships to users. `tests/e2e/operator/templates/plugin_src/Dockerfile` is `alpine:3.19`. `images.json` covers neither, and Dependabot raises no security alerts on Dockerfiles, so removing the version-update entry would have removed the only signal on both. Expect a `node:18-slim` → current-Node pull request on the first run; taking it is a decision about what the scaffold recommends, and Dependabot proposing it is what puts that decision in front of someone.
- #800 was filed as the port that lets the ceiling and the `ignore` rule come off, after @AntonTyb asked which workflows the constraint is for. Investigating it showed that is not what holds the ceiling, and the correction is in this branch. `deploy/docker/Dockerfile` names no `mcp` requirement; it arrives in `/opt/hermes/.venv` from the digest-pinned Hermes base (`HERMES_AGENT_TAG=v2026.8.13@sha256:68e15ae2…`), which ships **mcp 1.28.1 on Python 3.13.5** with `mcp.server.mcpserver` absent — so porting our three modules to `MCPServer` would break the running image on merge. And hermes-agent itself imports the deleted module (`/opt/hermes/mcp_serve.py:52`, `/opt/hermes/agent/transports/hermes_tools_mcp_server.py:159`), so upgrading that shared venv breaks the harness and not just our scripts. The ceiling comes off when the base image ships 2.x. Reported on #800 with a re-scope to "blocked on the Hermes base image"; the port itself stays out of this branch.
- Also corrected along the way: the npm entry's comment used to justify dropping the glob by calling that scaffold directory a user-facing template. Its `package.json` declares no dependencies, so npm never had anything to propose there — the coverage that mattered was the `docker` entry's.

## Testing

- Validated against the SchemaStore `dependabot-2.0` schema (`additionalProperties: false`, and current enough to know `cooldown`, `groups`, and `exclude-paths`): valid, re-run after both the `versions` change and the restored `docker` entry. Negative-tested by misspelling a key in the `ignore` entry, which the schema rejects — so a clean pass is not vacuous.
- `prettier@3.9.6 --check .github/dependabot.yml docs/designs/design_537148738.md`: clean. `make docs-check`: clean.
- `pip install -r requirements-test.txt` resolves `mcp` to 1.29.0 and `from mcp.server.fastmcp import FastMCP` imports — the constraint the revert restores does what its comment says.
- Both sides of the constraint exercised on Python 3.11, deliberately not the 3.12 CI uses, to answer whether `mcp<2` narrows the interpreter as well as the package: `mcp<2` resolves to 1.29.0 and the import succeeds; `mcp<3` resolves to 2.0.0 and the same import raises `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`. It does not narrow the interpreter — both releases declare `requires-python >=3.10`, classify 3.10 through 3.14, and publish one `py3-none-any` wheel with no compiled extension. Cross-checked which jobs this reaches: `python-tests.yml` pins 3.12, `k8s-operator-test.yml` uses the runner default (`cp312` wheels in run `32256798665`), and no other workflow installs the file.
- `make test-python`: one pre-existing failure, `test_slack_relay_patch.RelayedSlackErrorTest.test_an_unreadable_body_is_not_a_rejection` (`TypeError: a bytes-like object is required, not 'str'`). It fails identically on unmodified `upstream/main` at 8a77e15 and has nothing to do with any file here. Everything else passes.

The test-guard change was exercised in three environments, all on Python 3.11 (the 3.12 CI uses is not available on this machine), running `unittest discover` per directory the way the Makefile does:

| Environment | Result |
| --- | --- |
| `mcp` 1.29.0, what `requirements-test.txt` resolves | `agents/chat/scripts` 183 tests OK; `agents/platform/scripts` 513 tests, the one pre-existing failure above |
| `mcp` 2.0.0 forced | `agents/chat/scripts`: `ERROR: test_router_server`. `agents/platform/scripts`: errors on `test_agent_common_server`, `test_platform_mcp_server` and `test_session_kv_server`, plus four named failures from `test_mcp_package_contract` quoting the installed version and `ModuleNotFoundError: No module named 'mcp.server.fastmcp'` |
| Bare checkout — nothing third-party installed, full discovery | `agents/platform/scripts` 445 tests / 35 errors / 3 skipped; `agents/chat/scripts` 133 / 1 error. Both error sets are the missing-`yaml`/`fastapi` ones a bare checkout always produces |

The counterfactual is what makes the second row mean anything: with the branch's changes stashed and `mcp` 2.0.0 installed, the same tree produces only the one pre-existing error. The incompatibility was entirely invisible, which is the state #751 merged in.

The third row is the case the adversarial pass broke and this branch fixes: against the stashed baseline it is 443 / 35 / 1, so the change adds exactly the two contract tests, both skipping, and no new errors.

`main` was merged in at `b14d27d` to pick up #812, which is what the one red check was. That check, `Run Controller Tests`, was a golden-manifest diff in `k8s-operator/internal/testing` — `creationTimestamp: nil` on the PodDisruptionBudget goldens, from #733 and #790 landing after this branch was cut. Nothing here touches Go. #812 fixed it on `main`; `go build ./...` and `go test ./...` in `k8s-operator` are both clean on the merged tree.

The merge moved `main` a long way (#797 rewrote the install), so the config's claims were re-derived against the merged tree rather than assumed: still one `go.mod` (`k8s-operator`), the same four pip manifests, `package.json` only at `docs/site` and the dependency-less scaffold, and the same two Dockerfiles with a literal `FROM`. `requirements-test.txt` is still installed by exactly `python-tests.yml` (3.12) and `k8s-operator-test.yml` (runner default); the workflows on other Python versions install `pyyaml` or `tests/e2e/requirements.txt`, `rc-release-pipeline.yml` included.

`make docs-check` is clean on the merged tree. `make test-python` fails in three directories — `admin_console/tests/`, `agents/platform/scripts/` and `agents/platform/skills/pr-conversation/scripts/` — and fails in exactly those three, with the same counts, on unmodified `upstream/main` at `1c06e1a`. They are this machine's missing `fastapi`/`mcp`, not the branch: CI's `Run Python Unit Tests` passes. The branch's only delta is `agents/platform/scripts` at 696 tests / 2 errors / 7 skipped against main's 694 / 2 / 5 — the two contract tests, both skipping for want of `mcp`.

### Live validation

Not live-tested. Every file here is read by something outside a kube-agents installation — `.github/dependabot.yml` by GitHub, `requirements-test.txt` by `make test-python` and the CI job, the test modules by `unittest`. None reaches a cluster, an operator, or an agent pod, so there is no CR status, Deployment env, or in-pod file that could show the change took effect.

The one claim here that *is* about a running installation is the reason the ceiling exists, and that was read out of the images rather than reasoned about:

```
$ docker run --rm --network none --entrypoint /opt/hermes/.venv/bin/python3 \
    nousresearch/hermes-agent@sha256:68e15ae2a6d894d0ccbd9f8aacbbe13d4d28fa5dc9b6a303970b67bb2499b1a6 \
    -c "import importlib.metadata as m, sys; print(m.version('mcp'), sys.version)"
1.28.1 3.13.5 ...
```

`import mcp.server.mcpserver` raises `ModuleNotFoundError` in that image, `from mcp.server.fastmcp import FastMCP` succeeds, and the same holds in the built `platform-agent` image. `grep -rn "mcp.server.fastmcp" /opt/hermes` inside it returns `mcp_serve.py:52` and `agent/transports/hermes_tools_mcp_server.py:159`. Not checked against the deployed pod on the shared `gkedemos` install: `kubectl exec` needs the live-test lease and the image digest is the same one, so it would be the same observation at more cost to whoever else is testing.

The equivalent proof is on GitHub after merge, and it is worth someone doing rather than assuming. A `directories` entry that matches no manifest is not a configuration error — Dependabot proposes nothing, which reads exactly like "everything is up to date" — so every entry whose paths this PR changed needs checking, not just the one that is easiest to check.

Insights → Dependency graph → Dependabot should list **seven** rows, one per ecosystem per directory, with no configuration-error banner (a malformed file shows one there and silently stops all updates):

```
github-actions  /
gomod           /k8s-operator
pip             /
pip             /tests/e2e
npm             /docs/site
docker          /agents/platform/skills/gke-app-onboarding/assets
docker          /tests/e2e/operator/templates/plugin_src
```

Then "Check for updates" on each row whose paths moved:

- **`gomod /k8s-operator`** — one pull request titled `ci: bump the gomod group in /k8s-operator with N updates`, on a branch like `dependabot/go_modules/k8s-operator/gomod-<digest>`, not one per module. This is the grouping claim.
- **Both `docker` rows** — expect a `node:18-slim` proposal from the first and an `alpine:3.19` one from the second. Both bases are behind, so a row that finds nothing means the path is wrong, not that the image is current.
- **`pip /`** — `gh pr list --repo gke-labs/kube-agents --state open --json number,files` should show exactly one open pull request touching `admin_console/requirements.txt`, which is the de-duplication claim.

The two `docker` rows are the ones with nothing behind them. Every other named directory has a Dependabot pull request proving the path resolves: `/k8s-operator` from #750, #752 and #756 (`dependabot/go_modules/k8s-operator/…`), `/docs/site` from #753 and #761 (`dependabot/npm_and_yarn/docs/site/…`), `/` from #763, and `github-actions` at `/` unchanged by this PR. `/tests/e2e` holds `requirements.txt` and is new, but a `pip` row that finds nothing there is visible against the same file. `docker` has opened zero pull requests since #492 added it under `/**/*` — which is also why the glob's coverage of these two files cannot be taken as evidence they are manifests Dependabot recognises.

## Self-Review

Two passes of unequal quality, and the difference matters when reading what follows.

The Dependabot and `requirements-test.txt` work got **no independent pass**: the session it was written in was configured not to spawn subagents, so what is recorded below it is my own read of my own diff and should be discounted accordingly.

The test-guard commit got the pass AGENTS.md actually asks for — `review-adversarial` in a subagent handed the diff range and nothing else. It returned one HIGH and four LOW, all confirmed and all fixed:

- **HIGH — the guard I added would crash the suite it was meant to protect.** I wrote the absent-vs-broken check as `if importlib.util.find_spec("mcp") is not None: raise`. `find_spec` consults `sys.modules` before the filesystem and reads `__spec__` off whatever it finds; on a `types.ModuleType` stub that is `None`, and it raises `ValueError` rather than returning. `unittest discover` imports a directory into one process in alphabetical order, `test_agent_common_server.py` sorts first, and in a bare checkout it inserts exactly that stub — so `test_mcp_package_contract.py`, imported later in the same process, would blow up and take `make test-python` non-zero on any developer machine without third-party packages. CI was unaffected, which is the worst kind of unaffected. All five sites now ask `importlib.metadata.distribution("mcp")`, which reads the installed distribution and is immune to `sys.modules`. The three other guards were safe only by alphabetical luck; fixing them uniformly cost nothing. The bare-checkout row of the Testing table is that specific case, and it was not tested before the review flagged it.
- **LOW — an unverifiable claim stated flatly.** The pin comment said "nothing in `deploy/docker/Dockerfile` installs mcp". Line 106 is an unpinned `uv pip install "hermes-agent[...]"` into that venv — a resolution step that leaves an already-satisfied `mcp` in place, which is a weaker statement than the one I made. Reworded, and the comment now records that the 1.28.1 / Python 3.13 / hermes-import facts were read out of the image and names the digest to re-check them against.
- **LOW — a number that rots.** "the ceiling resolves 1.29.0 here" is false on any 1.30 release and carries no weight; the claim is about the major line. Dropped.
- **LOW — `PYTHONPATH` replaced rather than prepended** in the contract test's subprocess env, which would have made it see a different tree from the rest of the suite whenever the Makefile exported one.
- **LOW — a stale docstring**, caught on re-read after the fixes: the new test file repeated the unhedged Dockerfile claim the pin comment had just dropped.

What I looked for on the earlier commits, and what came back:

- **Coverage lost by replacing globs with named directories.** I checked every manifest in the tree against the new list and got the answer wrong twice; the bot's second and third passes corrected it. I first recorded the loss as the `gke-app-onboarding` scaffold's `package.json` and `Dockerfile`, attributed to dropping the npm glob, and wrote `docker` off as having updated nothing and being unable to. That `package.json` declares no dependencies, so npm covered nothing there, and the two literal `FROM` pins were the `docker` glob's. I then narrowed the claim but kept the removal, arguing a scaffold bump should not be a side effect of a config cleanup — which pointed the wrong way, since a Dependabot pull request is a proposal a human reviews, not a merge. The entry is back at two named directories in 322a1a0.
- **Whether `pip` at `/` really still reaches `admin_console`.** This is the load-bearing claim — if it does not, the change stops updating the console's dependencies entirely. Verified against the evidence rather than the docs: #763 was branch `dependabot/pip/httpx-…` (directory `/`), and its changed-files list is exactly `admin_console/requirements.txt`.
- **Whether the `mcp` ignore rule will actually fire.** I got this wrong and the bot review caught it. I took `update-types: version-update:semver-major` over `versions: [">=2"]` on the grounds that pip's accepted syntax for `versions` was not spelled out in the reference and a syntax error in `dependabot.yml` disables every update silently. But an `update-types` condition compares a candidate against the dependency's resolved _current_ version, and a pip requirements line only has one when it is an exact `==` pin — `mcp<2` is a bare range and there is no lockfile, so the rule expanded to nothing and the guard was inert. #751's title, `ci: update mcp requirement from <2 to <3`, is the same code path saying so. Fixed in 6ac20b0.
- **Config syntax, since a mistake here fails silently rather than loudly.** Schema-validated as above, not just eyeballed. Group identifiers satisfy the documented naming rule.
- **Whether the file's own comments are true.** Re-derived each: 20 pull requests (#750–#769), 13:11:36 to 13:13:29, five duplicate pairs, ceiling of 25.

- **Whether `mcp<2` restricts which Python versions can install the file**, after @AntonTyb raised it. It does not, and I checked rather than reasoned: 1.29.0 and 2.0.0 declare the same `requires-python`, carry the same 3.10–3.14 classifiers, and ship the same pure-Python wheel, and the two constraints behave as described above on 3.11. The question did find something the pass above had missed, which is that the file's own header comment named one of its two CI readers — fixed in 7617c9c. I had read that comment while editing four lines below it and not checked it against the workflows.

@jayantid's review found three more, all confirmed against the tree and all fixed:

- **The post-merge checklist tested the one entry that needed it least.** It covered `gomod` and the `admin_console` pull-request count and left both `docker` rows unchecked — the two whose paths changed with no prior Dependabot run behind them, and where a wrong path proposes nothing and reads as "up to date". The Live validation section now lists all seven expected rows and says what each moved row should produce. Checking the point sharpened it: `/k8s-operator` and `/docs/site` are proven by #750/#752/#756 and #753/#761's branch names, so `docker` really is the only untested pair.
- **A docstring asserting a contamination path the test runner cannot create.** `test_router_server.py` said a leaked stub is what "the next module" finds and reaches `fastapi` "seventeen imports later". It sorts last of the eight in `agents/chat/scripts`, nothing there imports `fastapi`, and `Makefile:184` runs discovery once per directory in its own process. The accurate version of that sentence is in `test_agent_common_server.py`, which sorts first in the directory where the seventeenth module does import `fastapi` — I wrote it there and then reused it where it was not true. The stubbing change stays; its stated reason is now the defensive one. The PR body repeated the same claim and is corrected above.
- **The same rationale written out five times.** The `importlib.metadata`-not-`find_spec` block was in all five files, with ABSENT-vs-BROKEN and the #751 history alongside four of them — so #800 lifting the ceiling means editing five places, and the one that gets missed keeps asserting 2.x is unusable. This is the drift the `requirements-test.txt` header already demonstrated on this branch. `test_mcp_package_contract.py` now holds the one copy and names the four guards that point at it. Comments only: both directories return the same counts before and after (`agents/chat/scripts` 183 OK; `agents/platform/scripts` 469 tests, 2 errors, 3 skipped on a checkout without `mcp` or `fastapi`).

Unresolved, and the reasons:

- I cannot exercise Dependabot itself before merge — the config only takes effect once it is on the default branch. So the claim "one grouped PR per ecosystem" rests on the documented behaviour of `groups`, not on an observation. The Live validation section says what to check afterwards.
- Everything Python here ran on 3.11, not the 3.12 CI pins. That is the wrong interpreter for the one pre-existing failure noted above — it is a 3.11 stdlib artifact and is green on CI — and it means a failure specific to 3.12 would not have shown up locally.
- The four re-raising guards make the *other* directories' behaviour depend on which test module sorts first, and I fixed that by making the check independent of `sys.modules` rather than by removing the ordering dependence. A future test module that imports `mcp` for real before these run is still a way to change what they see. I did not build a guard against that; the contract test's subprocess is the thing that does not care.
- This branch does not port the three servers. #800 asked for it, it is not deliverable for the reasons in Context, and the issue now says so rather than the branch quietly not doing it.

## Risk & Rollout

Low. No runtime code path, no operator behaviour, no image contents. The worst realistic failure is a config Dependabot rejects, which stops dependency updates until noticed — the Dependabot tab check above is what catches it, and `git revert` of the first commit restores the previous behaviour. The `requirements-test.txt` commit reverts independently and only tightens a constraint that was already tighter last week.

The test-guard commit is the one that can newly turn CI red, and that is its purpose: any future widening of `mcp<2` now fails by name instead of passing. It reverts independently too, and reverting it restores the silent-green behaviour rather than breaking anything.


.github/dependabot.yml
```diff
@@ -41,10 +67,22 @@ updates:
...
       prefix: "ci"
+    # requirements-test.txt holds mcp below 2.0 because agent_common_server.py
+    # imports mcp.server.fastmcp, which 2.x moved. #751 widened the constraint
+    # to <3 anyway; the tests stub the module, so nothing caught it.
+    ignore:
+      - dependency-name: "mcp"
+        update-types:
+          - "version-update:semver-major"
+    groups:
+      python:
```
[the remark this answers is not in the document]
bradhoekstra: Confirmed and fixed in 6ac20b0 — the rule is now `versions: [">=2"]` with no `update-types` key.

Your reading of the mechanism matches the evidence: `requirements-test.txt:33` is a bare `mcp<2` range with no lockfile in the tree, so there is no resolved current version for a semver-major comparison to work from, and #751's `ci: update mcp requirement from <2 to <3` title is the requirement-update path saying exactly that. My Self-Review had reached the opposite conclusion for a bad reason — I avoided `versions` because I could not find its pip syntax documented and a malformed `dependabot.yml` fails silently. `>=2` is a plain PEP 440 specifier, and the file re-validates against the SchemaStore `dependabot-2.0` schema; I also negative-tested that validation by misspelling the key, which it rejects.

The PR body's Self-Review bullet has been rewritten to say this rather than the original claim.

bradhoekstra:
Both findings from the automated review are addressed.

**The `mcp` ignore rule cannot fire against an unpinned requirement** — fixed in `6ac20b0`, answered in the thread.

**The design of record still specifies the globbed config** — fixed in `80903e0`. `docs/designs/design_537148738.md` §3.4 now describes the four ecosystems with their named directories and records why there is no `docker` entry; §4 rule 5 no longer states globbing as a governance rule and instead says why `/**/*` is the thing not to reach for, citing #763/#764. §2.4's "Current State" is left alone — it is the pre-design snapshot that prompted the work, not a claim about `main` — and its Target Fix now points at §3.4. §1's summary drops Docker from the ecosystem list. `docs/README.md:269`'s row still reads true, so it is untouched. `make docs-check` is clean, which as you noted it would have been either way.

.github/dependabot.yml
```diff
@@ -55,17 +97,11 @@ updates:
...
-    commit-message:
-      prefix: "ci"
+    groups:
+      npm:
+        patterns:
+          - "*"
+# No docker entry. Every FROM that ships interpolates two ARGs --
+# `FROM ${GOLANG_IMAGE}:${GOLANG_VERSION}` -- which the Docker updater cannot
+# resolve, so the entry opened no pull requests and would open none. images.json
+# and `make mirror-images` own those pins.
```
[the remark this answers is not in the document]
bradhoekstra: Confirmed on every point and fixed in 287e535.

Verified rather than taken on trust: `assets/Dockerfile:2` is a literal `FROM node:18-slim`; `deploy/docker/Dockerfile:1112` is `COPY --chown=hermes:hermes agents/platform/skills/ /opt/platform-template/skills/`, so it does ship in the platform image; `tests/e2e/operator/templates/plugin_src/Dockerfile:1` is `FROM alpine:3.19`; the removed entry was `directories: ["/**/*"]`; and `assets/package.json` has no `dependencies` block, so the misattribution point holds too — npm never covered that directory and the loss was the `docker` entry's.

Took your suggested shape. The removal stands, the claim is narrowed to the images this repository builds, and both unmonitored pins are now named in the config's trailing comment, in `design_537148738.md` §3.4, and in the PR body's Context section, each noting `images.json` covers neither. The Self-Review bullet that wrote `docker` off has been rewritten to say what it got wrong.

One thing I did not do: add a `docker` entry naming those two directories. Auto-bumping `node:18-slim` to a current Node in a scaffold the agent hands to users is the deliberate decision this PR argues should not happen as a side effect, so the gap is documented rather than silently closed. If you read that differently, say so and I will add the entry.

.github/dependabot.yml
```diff
@@ -55,17 +97,17 @@ updates:
...
+# No docker entry. Every FROM in an image this repository builds interpolates
+# two ARGs -- `FROM ${GOLANG_IMAGE}:${GOLANG_VERSION}` -- which the Docker
+# updater cannot resolve; images.json and `make mirror-images` own those pins.
+#
+# Two literal FROMs the old `/**/*` glob did cover are now unmonitored, and
+# neither is in images.json: agents/platform/skills/gke-app-onboarding/assets
+# (`node:18-slim`, EOL, baked into the platform image and handed to users by
+# the skill) and tests/e2e/operator/templates/plugin_src (`alpine:3.19`).
+# Bumping either is a deliberate change, not something to restore by adding a
+# glob back here.
```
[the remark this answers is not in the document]
bradhoekstra: You're right and I've taken the entry back — 322a1a0 adds `docker` at `/agents/platform/skills/gke-app-onboarding/assets` and `/tests/e2e/operator/templates/plugin_src`, grouped, same shape as `gomod` and `npm`.

The argument that moved it was yours about the duplicates coming from the glob rather than the ecosystem. My rebuttal in the previous round — that bumping a user-facing scaffold should be a deliberate change, not a side effect — pointed the wrong way: Dependabot opens a pull request, it does not merge one, so the entry *creates* the decision point rather than bypassing it. Confirmed the rest independently: `images.json` has no `node:18-slim` or `alpine:3.19` (its only hits are `docker.io/alpine/k8s` and a `1.26-alpine` Go tag), and Dependabot raises no alerts on Dockerfiles, so the version-update entry was the only signal on either.

One correction to your finding, which does not change it: `gke-app-onboarding/SKILL.md:39`'s "Dockerfile (recommended for most apps)" heads an inline **Go** multi-stage example, and `SKILL.md` does not reference `assets/` anywhere. The file still ships at `/opt/platform-template/skills/gke-app-onboarding/assets/Dockerfile` via the `COPY` you cite, so it is still an EOL base image inside the published image — the "presented as recommended" framing is the part that is loose.

Also fixed one thing your pass surfaced indirectly: the npm entry's comment justified dropping the glob by calling that directory a user-facing scaffold, when the real reason npm loses nothing there is that its `package.json` declares no dependencies. `design_537148738.md` §1, §2.4, §3.4 and §4 rule 5 all now describe the five-entry config. Re-validated against the SchemaStore `dependabot-2.0` schema; `prettier` and `make docs-check` clean.

requirements-test.txt
```diff
@@ -30,7 +30,7 @@ fastapi
 
 # agent_common_server.py does `from mcp.server.fastmcp import FastMCP`, which
 # 2.x moved. Held below 2.0 until that import is updated.
-mcp<3
+mcp<2
```
AntonTyb, line 63: which specific github workflow this should work with? cause in different workflow we're using 3.10, 3.11, 3.12 and 3.13...

AntonTyb (commented):
narrowing the allowed versions may introduce flake-ness to GH workflows - especially since we are using different py versions in different workflows...

requirements-test.txt
bradhoekstra: Two workflows, both on 3.12 — and the constraint isn't a Python constraint.

**Which workflows read this file.** Only `python-tests.yml`, which sets `python-version: "3.12"` at line 30, and `k8s-operator-test.yml`, which has no `setup-python` step and so installs against the `ubuntu-latest` runner's default `python3`. That default is 3.12 too — the pip log in run `32256798665` downloads `cp312` wheels.

The others never install it: `e2e-gchat-test.yml` and `rc-release-pipeline.yml` (both 3.11) install `tests/e2e/requirements.txt`; `agent-startup-test.yml`, `agentplugins-test.yml` (3.13) and `auto_request_review.yml` install `pyyaml` and nothing else. No workflow in the tree uses 3.10.

**On the flakiness risk.** `mcp<2` narrows the package, not the interpreter. 1.29.0 and 2.0.0 both declare `requires-python >=3.10`, both classify 3.10 through 3.14, and both ship a single `py3-none-any` wheel — no compiled extension, no per-Python build. There is no Python version this excludes that `mcp<3` would have allowed.

Checked rather than inferred, on 3.11 — deliberately not CI's version:

```
pip install 'mcp<2'  -> 1.29.0; from mcp.server.fastmcp import FastMCP  -> OK
pip install 'mcp<3'  -> 2.0.0;  from mcp.server.fastmcp import FastMCP  -> ModuleNotFoundError
```

which is the direction of the risk: `<3` is what breaks, identically on every Python. 2.0 removed `fastmcp` outright rather than renaming it — the replacement is `mcp.server.mcpserver.MCPServer` — so `agent_common_server.py`, `platform_mcp_server.py` and `router_server.py` need rewriting, not re-importing. That port is what lets the ceiling come off, and it is not this PR: filed as #800, which also covers giving one of those modules a test against the real package so the next widening fails loudly instead of passing the way #751 did.

**What your question did change.** The header comment claimed two readers and named `python-tests.yml` as the CI one, so "which workflow" was a fair question the file could not answer. 7617c9c names all three, says which pins its Python and which takes the runner default, and records that `mcp<2` bounds the package and not the interpreter.

One genuine version of your concern, separate from this PR: `k8s-operator-test.yml` running this suite on an unpinned runner Python is drift waiting to happen — it is 3.12 today by accident of the runner image, not by declaration. I have left it alone rather than widen a Dependabot-grouping PR into workflow config, but say the word and I will pin it.

.github/dependabot.yml
```diff
@@ -55,10 +102,27 @@ updates:
...
+  # Every other Dockerfile is out of reach rather than excluded: the images
+  # this repository builds interpolate two ARGs each
+  # (`FROM ${GOLANG_IMAGE}:${GOLANG_VERSION}`), which the Docker updater cannot
+  # resolve, and the agentplugins ones are `FROM scratch`. images.json and
+  # `make mirror-images` own the ARG-interpolated pins. Naming the two
+  # directories, rather than globbing, is what keeps this from recreating #763.
   - package-ecosystem: "docker"
     directories:
-      - "/**/*"
+      - "/agents/platform/skills/gke-app-onboarding/assets"
```
jayantid, line 124: The `docker` entry moves from `/**/*` to two hand-typed absolute paths. Nothing in this PR or in its post-merge checklist establishes that either path resolves to a Dockerfile Dependabot recognises.

A `directories` entry that matches no manifest fails silently. Dependabot raises no configuration error for it and simply proposes nothing, which looks identical to "everything is up to date". This is not theoretical: the `docker` ecosystem has opened zero pull requests since #492 added it, even though the old `/**/*` glob already reached both files this entry now names — `agents/platform/skills/gke-app-onboarding/assets/Dockerfile` (`node:18-slim`, past end of life) and `tests/e2e/operator/templates/plugin_src/Dockerfile` (`alpine:3.19`). `deploy/docker/Dockerfile:1112` copies `agents/platform/skills/` wholesale into the published platform image, so the EOL Node base ships to users. The Live validation section's post-merge check covers the `gomod` row and the `admin_console` pull-request count only, so neither `docker` row gets tested.

Extend the post-merge checklist to cover the entries whose paths changed:

```
Insights -> Dependency graph -> Dependabot: expect 7 rows --
  github-actions /, gomod /k8s-operator, pip /, pip /tests/e2e,
  npm /docs/site, docker /agents/platform/skills/gke-app-onboarding/assets,
  docker /tests/e2e/operator/templates/plugin_src

"Check for updates" on each docker row -> expect a node:18-slim and an
alpine:3.19 proposal. A docker row that finds nothing means the path is
wrong, not that the image is current.
```

agents/chat/scripts/test_router_server.py
```diff
@@ -17,24 +18,55 @@ def _load_router_server():
...
+
+    ABSENT is not the same as BROKEN, and only the first earns a stub. An
+    installed mcp that no longer exposes mcp.server.fastmcp -- every 2.x, which
+    replaced it with mcp.server.mcpserver -- means router_server cannot run in
+    that environment, and the ImportError is the finding. Stubbing past it is
+    how #751 widened the ceiling to <3 and kept CI green;
+    agents/platform/scripts/test_mcp_package_contract.py is the test that says
+    so out loud, and it covers this module too.
+
+    Stub per module rather than by bulk sys.modules.update: unittest discovery
```
jayantid: The docstring at lines 33-36 says the bulk `sys.modules.update` leaked stubs into "the next module" and reached `fastapi` "seventeen imports later". Neither happens here. `test_router_server.py` sorts last of the 8 test modules in `agents/chat/scripts`, so no module is discovered after it. Nothing in `agents/chat/scripts` imports `fastapi` — the only occurrences of that string in the directory are in this comment. `test_session_kv_server.py`, which does import `fastapi`, lives in `agents/platform/scripts`, and `Makefile:184` runs `unittest discover` once per directory in a separate process.

The sentence reads as the reason the change was made, and the PR body repeats it ("clobbering real, importable `mcp` and `pydantic` for every test module discovered after it in the same process"). The next person to touch this guard will believe a cross-module contamination path the test runner cannot create. The accurate version of the same sentence is at `test_agent_common_server.py:69-73`, where that file does sort first in the directory that imports `fastapi`.

Keep `_stub_if_missing` — it is right defensively — and state the reason that holds:

```python
    Stub per module rather than by bulk sys.modules.update: overwriting a
    real, importable module with a stub is wrong even where nothing else in
    this directory would notice. test_agent_common_server.py has the case
    where it does.
```

agents/platform/scripts/test_session_kv_server.py
```diff
@@ -17,16 +17,35 @@
...
+# session_kv_server genuinely cannot run, so the ImportError is the finding
+# rather than something to route around. Stubbing both alike is what let #751
+# widen the ceiling in requirements-test.txt to <3 without CI noticing.
 try:  # pragma: no cover - depends on the installed mcp version
     import mcp.server.fastmcp  # noqa: F401
 except Exception:  # pragma: no cover
+    import importlib.metadata
     import types
 
+    # importlib.metadata and not importlib.util.find_spec: find_spec consults
```
jayantid: The `importlib.metadata` vs `find_spec` rationale appears five times: here at lines 37-41, `test_agent_common_server.py:53-58`, `test_platform_mcp_server.py:28-32`, `test_router_server.py:38-41`, and `test_mcp_package_contract.py:50-57`. The ABSENT-vs-BROKEN explanation and the #751 history are each repeated four times alongside it.

#800 is already filed to remove the ceiling. When it lands, all five blocks have to change together, and the one that gets missed will keep asserting that mcp 2.x is unusable. This PR's own Self-Review records that exact drift in this exact file set: the `requirements-test.txt` header named one of its two CI readers and stayed wrong until someone asked. AGENTS.md states the rule — "Every fact has one home. Duplicating documentation across files is how it goes stale."

Keep the full rationale in `test_mcp_package_contract.py`, the module named for it, and reduce the four guards to a pointer:

```python
    # ABSENT is not BROKEN: stub only when no mcp distribution is installed.
    # importlib.metadata, not find_spec -- see test_mcp_package_contract.py.
```

.github/dependabot.yml
bradhoekstra: Fixed in the PR body alongside `6707fbc`. Live validation now lists all seven expected rows and gives a per-row check for each entry whose paths moved — both `docker` rows included, with your reading of a silent miss ("a row that finds nothing means the path is wrong, not that the image is current") stated explicitly.

Checking the point sharpened it. `gomod` and `npm` also moved off `/**/*`, but both have branch names proving the new path resolves: #750/#752/#756 are `dependabot/go_modules/k8s-operator/…` and #753/#761 are `dependabot/npm_and_yarn/docs/site/…`. `docker` has no equivalent — zero pull requests since #492 — so it is the only pair with nothing behind it, which is what you said. `/tests/e2e` is new too, but it holds `requirements.txt` and a miss there is visible against that file.

agents/chat/scripts/test_router_server.py
bradhoekstra: Confirmed and fixed in `6707fbc`. Checked all three of your points: `test_router_server.py` sorts last of the eight test modules in `agents/chat/scripts`, the only occurrence of the string `fastapi` in that directory was the comment itself, and `Makefile:184` runs `unittest discover` once per directory in its own process. The path I described cannot happen there.

I wrote that sentence for `test_agent_common_server.py`, where it holds — it sorts first in `agents/platform/scripts`, and `test_session_kv_server.py` is the seventeenth module and does import `fastapi` — and then reused it where it was not true. `_stub_if_missing` stays; its stated reason is now the defensive one you suggested. The PR body repeated the claim and is corrected too.

agents/platform/scripts/test_session_kv_server.py
bradhoekstra: Fixed in `6707fbc`. `test_mcp_package_contract.py` keeps the full rationale and now names the four guards that point at it, so #800 changes one place. Each guard is down to the two lines you wrote — `ABSENT is not BROKEN` plus `importlib.metadata, not find_spec -- see test_mcp_package_contract.py` — with only the file-specific facts kept alongside (the `/inject` denial tests in `test_session_kv_server.py`, the `session_manager` handling in `test_agent_common_server.py`).

Comments only, and verified as such: `agents/chat/scripts` is 183 tests OK and `agents/platform/scripts` 469 tests / 2 errors / 3 skipped both before and after, on a checkout with neither `mcp` nor `fastapi` installed. The two errors are the pre-existing `test_slack_relay_patch` one and the bare-checkout `fastapi` import.

AntonTyb (approved):

jayantid (approved):

PR #793, opened 2026-08-18
fix(platform): close the two follow-ups #780 deferred on the relay shims

bradhoekstra:
## Summary

#780 stopped the outage that Hermes v2026.8.13 caused — `slack_relay_patch`'s
`PlatformRegistry.register` wrapper had restated upstream's signature, upstream added a
keyword-only `scope`, and every platform plugin's registration raised `TypeError`. It named two
follow-ups and deliberately left them: `create_adapter` is wrapped the same way in both relay
patches and still restates a signature, and the install-time sweep over already-registered
entries still reads only `_entries` when a scoped registration lands in `_scoped_entries`. This
closes both, and adds a build-time check that runs the relay patches against the real
`/opt/hermes` tree so the next base-image bump fails the build rather than the gateway.

## Why This Change

The failure mode is worth naming precisely, because it is the reason for the third piece. Every
wrapper these patches install sits on a class the whole gateway registers through, so one
`TypeError` in one shim takes Slack, Google Chat and A2A down together. The tests could not catch
it: both registry fakes declared the same narrow signature the shim did, so the fake and the
wrapper agreed with each other and CI stayed green while the pod came up with no chat adapters.

`register` was fixed. `create_adapter` has the identical shape and the identical blind spot in the
fakes; nothing about the failure was specific to `register`, and upstream is as free to add an
argument to one as to the other. A fake can only ever agree with the shim, so the check that
matters has to run against the real tree.

## What Changed

- Both relay patches forward everything past `name` to `create_adapter` blind, the way `register`
  now does. No behaviour change against the pinned base image — the signature has not drifted.
- The install-time sweep reads `_scoped_entries` as well as `_entries`, and warns when *either*
  private name goes missing rather than only when both do. With `_scoped_entries` renamed away the
  old `or` kept reading the one map a scoped registration never lands in and reported success.
- The sweep's comments, its test docstrings and the build-time check now say it is defensive
  cover, not a live fix: `sitecustomize`'s import hook calls `install()` as
  `gateway.platform_registry` finishes executing, before anything can register, and nothing else
  calls `install()`. The previous wording implied a silent production failure this ordering cannot
  produce.
- The module-level test fakes declare wide signatures for both wrappers, and forwarding is
  asserted for each.
- `deploy/docker/patches/verify_slack_relay_registry_contract.py` runs at image build against the
  real Hermes tree: it pins the signature of every upstream callable the two relay patches
  replace, registers a plugin-sourced entry the way `plugins.py` does, and repeats the behavioural
  checks in a child process wired the way the pod is, where `sitecustomize`'s hook is what
  installs the patch. Before reading the pin it re-execs with every variable in
  `sitecustomize.RELAY_PATCHES` stripped and asserts no patch sentinel is set, so a run inside a
  pod cannot read a shim as upstream's own declaration.
- The check rides in an existing `COPY`/`RUN` pair rather than adding its own. `credential-proxy`
  builds `FROM` this stage and its chain measures 119 of the 120 layers
  `scripts/check_image_layers.py` allows.

## Context

No issue tracks this. #780 (merged) is the fix these two follow-ups were deferred from; its
description names both.

Overlap with open pull requests, none of it line-level:

- #720 touches `agents/platform/scripts/slack_relay_patch.py` and
  `agents/platform/scripts/google_chat_relay_patch.py`. Different regions of both files —
  a merge conflict warning, not a duplicate.
- #731, #728, #738 and #713 all touch `deploy/docker/Dockerfile`. This branch adds no layer, but
  the budget has one slot left, so whichever of those adds a `COPY` or `RUN` merges second will
  need to find it. Flagged rather than fixed: it is not this branch's layer to reclaim.

## Testing

- `python3 -m pytest agents/platform/scripts/test_slack_relay_patch.py` — 37 passed, 1 failed.
  The failure is `RelayedSlackErrorTest::test_an_unreadable_body_is_not_a_rejection`, pre-existing
  on `main` and local to Python 3.11: `HTTPError.read()` returns `str` rather than `bytes` when
  `fp` is `None` before 3.12. `main` fails the same test on the same interpreter (31 passed, 1
  failed); CI pins 3.12 and is green.
- `make docker-build-platform` — the build-time gate passes both phases, 27 checks unwired and 8
  wired.
- The gate fails in the direction it is meant to. Narrowing the `register` shim back to
  `(self, entry)` and simulating a future upstream signature change each exit 1 with the drifted
  callable named. The re-exec guard was verified both ways against the built image: the
  pre-fix script run with `GOOGLE_CHAT_RELAY_URL` inherited reports
  `PlatformRegistry.create_adapter: expected 'self, name, config', got 'self, name, *args, **kwargs'`
  — the false drift finding 1 describes — and the fixed script passes.
- `make docs-check` — clean.
- `make iac-parity-check` — clean, 16 checks.
- `python3 scripts/check_image_layers.py` against `credential-proxy:latest` — 119 layers, 1 under
  budget, unchanged by this branch.
- `prettier` — no Markdown, JSON or YAML files changed.

### Live validation

Install: `kube-agents-cluster` (regional `us-central1`, project `bhoekstra-gkedemos`), namespace
`kubeagents-system`, one `PlatformAgent` CR, model `claude-opus-5` via Vertex through LiteLLM.

Built and pushed `platform-agent:relayfix-fcd32d7` — the branch head, including the self-review
fixes below — and patched the CR's image tag from `dns-endpoint-7dac349`. The rollout was
confirmed at each layer: CR `.status.phase` `Ready` with "Agent sandbox and Envoy credential
sidecar are fully reconciled", the `platform-agent-gateway` Deployment's `platform-agent`
container on the new tag, and the pod `3/3 Running`.

Inside that pod, `PlatformRegistry.create_adapter`'s live signature reads
`(self, name, *args, **kwargs)` rather than the `(self, name, config)` it restated before — this
install has Slack disabled and Google Chat enabled, so it is `google_chat_relay_patch` that
installed it, which is the half of the change with no test that can distinguish a fake from the
real class.

The sweep was proved by A/B in that pod, not by reading the code. Slack being disabled means
`install()` early-returns on an unset `SLACK_RELAY_URL`, so the probe starts the interpreter with
it unset, registers a `source="plugin"` Slack entry into `_scoped_entries[scope]` the way
`register_platform` does — before `install()`, so the `register` wrapper cannot be what patches
it — then sets the variable and calls `install()`. Control was `main`'s `slack_relay_patch.py`
ahead of `/opt/defaults/scripts` on `sys.path`, with that directory left in place so
`sitecustomize` can still find `google_chat_relay_patch`. Each arm printed
`slack_relay_patch.__file__` and whether the loaded source mentions `_scoped_entries`, so the arms
are distinguishable in the output rather than by assumption:

```
[control] slack_relay_patch.__file__ = /tmp/probe/mainver/slack_relay_patch.py
[control] source mentions _scoped_entries = False
[control] entry landed in _scoped_entries = True, in _entries = False
[control] RESULT pre-registered scoped entry patched = False
[branch]  slack_relay_patch.__file__ = /opt/defaults/scripts/slack_relay_patch.py
[branch]  source mentions _scoped_entries = True
[branch]  entry landed in _scoped_entries = True, in _entries = False
[branch]  RESULT pre-registered scoped entry patched = True
```

Four earlier probe designs were wrong and are worth recording, because each looked like a result.
One returned a false negative from the `SLACK_RELAY_URL` early return. Two returned false
positives: the `register` shim `sitecustomize` had already installed patched the entry at
registration time, and a `sys.path.insert` made both arms load the branch's file. The fourth
dropped `/opt/defaults/scripts` from `sys.path` entirely and turned the control arm into an
unrelated `ModuleNotFoundError`.

Google Chat, the one live platform here, kept working on the new image: after two startup 503s
against the credential proxy at 20:28:59 — the relay loop polling before the sidecar was serving,
not a regression — the proxy logged a steady stream of `GET /v1/chat/events 200` and the agent
logged no further failures.

Not covered: no live Slack workspace is wired to this install, so an end-to-end relayed Slack
message was not sent. This branch does not change the patch's delivery behaviour.

Cleanup: the CR was reverted to `dns-endpoint-7dac349`, the pod confirmed `3/3 Running` and the CR
`Ready` on it; the probe directory was removed from the pod, the host temp files and the local
inspection image deleted, and the live-test lease released.
`platform-agent:relayfix-fcd32d7` was left in the gkedemos Artifact Registry.

## Self-Review

`review-adversarial` against the branch diff, run in a subagent handed the diff range and nothing
else. Angles that produced findings: contract drift between a check and what it actually checks,
comments and docstrings that outrun the code, the guard that fails in the quiet direction, and
test fakes that agree with the code under test. Seven findings, all confirmed; five fixed, two
reported.

1. **The build-time gate could report false drift, and invite the reader to make it permanent.**
   Fixed. The re-exec guard stripped only `SLACK_RELAY_URL`. The operator sets
   `GOOGLE_CHAT_RELAY_URL` on the agent container and `google_chat_relay_patch` replaces
   `create_adapter` too, so a run inside a pod read the Google Chat shim as upstream's declaration
   and reported drift that did not exist. Worse than a missed check: the module docstring tells
   the reader to update the pin deliberately, and doing so there would have pinned the shim
   against itself and deleted the only guard on that method. It now strips every variable in
   `sitecustomize.RELAY_PATCHES` — read from the module, not restated — and asserts no patch
   sentinel is set before reading the pin.
2. **The sweep's justification was not true.** Claims rewritten, code kept. The comment, a test
   docstring and the build check all described a silent production failure, but `sitecustomize`'s
   import hook makes that ordering unreachable: `install()` runs as the registry module finishes
   executing, before anything can register, and nothing else calls `install()`. Kept because the
   code is correct for the ordering it covers and cheap; the wording now says it is defensive
   cover, so a reader is not told the sweep is load-bearing when it is not. My first draft of the
   Live validation section above made the same overclaim, and this is what corrected it.
3. **The "not introspectable" guard failed quiet in the half that matters.** Fixed. The condition
   was `isinstance(entries, dict) or isinstance(scoped, dict)`, so a base image that renamed
   `_scoped_entries` alone left the sweep reading `_entries`, finding it empty, and saying
   nothing — and `_entries` is the map a scoped registration never lands in. It now warns when
   either name goes missing, and still sweeps whichever map is readable.
4. **A fake still had the exact blind spot this PR is about.** Fixed. The module-level
   `PlatformRegistry` fake's `create_adapter` was still `(self, name, config)` — the same
   fake-agrees-with-shim configuration that let #718 ship. Widened, with forwarding asserted the
   way `register`'s is.
5. **`shape()` erases default values, so a flipped upstream default passes the pin.** Reported,
   not fixed. Real, and narrower than it looks: for the two forwarding wrappers a changed default
   cannot break anything, since the wrapper passes through whatever the caller gave. It does
   matter for the adapter methods the patch reimplements rather than delegates to —
   `connect(self, *, is_reconnect=False)` is the example. Not fixed here because including
   defaults makes the pin churn on every upstream default tweak, most of them irrelevant, and
   deciding what the pin should tolerate is a change with its own argument to make. Filed as
   #794, which lays out the three options.
6. **A class docstring described a sweep that no longer exists.** Fixed. It still said the sweep
   reads a single private `_entries`.
7. **Layer-budget contention with open pull requests.** Reported, not fixed — see Context. This
   branch adds no layer; the last free slot goes to whoever takes it first.

## Risk & Rollout

The runtime change is confined to two monkeypatch modules that only load when the corresponding
relay URL is set. Against the pinned base image, widening `create_adapter` is a no-op: the
signature it forwards is the signature it used to restate. The sweep change alters a path
`sitecustomize` prevents from being reached, plus one new `LOGGER.warning` that can fire on a
future base image which consolidates the two entry maps — a log line, not a failure.

The new failure mode is a build one, and deliberate: a base-image bump that changes any pinned
signature now fails `make docker-build-platform` instead of producing an image that starts and
does not relay. Fixing it means reading the drift the check names and updating the shim, or the
pin, on purpose.

Revert is the single commit; nothing persists outside the image.

Nothing has to happen at merge time.

---

### DIFF
--- a/agents/platform/scripts/google_chat_relay_patch.py
+++ b/agents/platform/scripts/google_chat_relay_patch.py
@@ -167,8 +167,12 @@ def install() -> None:
     original_registry_create = PlatformRegistry.create_adapter
     if not getattr(PlatformRegistry, "_credential_proxy_relay_patched", False):
 
-        def create_adapter(self: Any, name: str, config: Any) -> Any:
-            adapter = original_registry_create(self, name, config)
+        # Forwarded blind past ``name``: this wrapper adds a side effect and
+        # delegates, so upstream owns the signature. Restating one is how the
+        # Slack relay's registry shim took every platform down when the base
+        # image grew a new keyword-only argument (see slack_relay_patch).
+        def create_adapter(self: Any, name: str, *args: Any, **kwargs: Any) -> Any:
+            adapter = original_registry_create(self, name, *args, **kwargs)
             if name == "google_chat" and adapter is not None:
                 patch_adapter_class(type(adapter))
             return adapter
--- a/agents/platform/scripts/slack_relay_patch.py
+++ b/agents/platform/scripts/slack_relay_patch.py
@@ -691,8 +691,12 @@ def install() -> None:
     original_registry_create = PlatformRegistry.create_adapter
     if not getattr(PlatformRegistry, "_slack_credential_proxy_relay_patched", False):
 
-        def create_adapter(self: Any, name: str, config: Any) -> Any:
-            adapter = original_registry_create(self, name, config)
+        # ``*args``/``**kwargs`` rather than upstream's signature restated: this
+        # wrapper adds a side effect and delegates, so the signature is not its
+        # to own. Restating one is what took chat down when v2026.8.13 gave
+        # ``register`` a keyword-only ``scope`` -- see the note there.
+        def create_adapter(self: Any, name: str, *args: Any, **kwargs: Any) -> Any:
+            adapter = original_registry_create(self, name, *args, **kwargs)
             if name == "slack" and adapter is not None:
                 patch_adapter_class(type(adapter))
             return adapter
@@ -732,13 +736,19 @@ def install() -> None:
             # One residual difference from calling the original directly, and it
             # predates this wrapper's argument handling: when ``scope`` is
             # omitted the registry infers it from the caller's frame at a fixed
-            # depth, and this frame occupies the slot it reads. The inferred
-            # scope is lost rather than wrong -- register() falls back to the
-            # entry's adapter_factory and then its check_fn, both
-            # frame-independent -- so an entry can land in the global scope
-            # instead of a plugin one. Nothing in the runtime relies on that
-            # inference: the plugin loader passes ``scope`` explicitly, and the
-            # built-in relay path sets source="builtin", which skips it.
+            # depth (``_caller_plugin_scope`` reads ``sys._getframe(2)``), and
+            # this frame occupies the slot it reads. Three things keep that
+            # harmless. The inference is gated on ``entry.source == "plugin"``,
+            # so the built-in relay path never reaches it. ``register_platform``
+            # -- the path every bundled platform takes -- passes ``scope``
+            # explicitly. And where it does run, the shifted frame resolves to
+            # this module, which is not a plugin namespace, so upstream falls
+            # through to ``_plugin_scope_from_callable`` on the entry's own
+            # ``adapter_factory`` and then its ``check_fn``; both are
+            # frame-independent and recover the same scope
+            # (``platform_registry.py:515-520`` in v2026.8.13). Only an entry
+            # whose callables the tool registry cannot place lands in the global
+            # scope instead of a plugin one.
             return original_registry_register(self, entry, *args, **kwargs)
 
         PlatformRegistry.register = register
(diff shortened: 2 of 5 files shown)


adamparco (approved):
/lgtm

Issue #747, opened 2026-08-18
NetworkPolicy: a parity guard for the eight static copies, and operator-owned litellm-policy (remainder of #676/#687)


mplakhtiy:
## Status — updated 2026-09-03 (verified against main `b0a4d7d1`)

Two items remain. Everything else in this issue has merged.

| Remaining | Where it stands |
| --- | --- |
| **B5 / D1** — a parity guard over the eight hand-maintained static DNS rules | In flight as **#1090** (open, `MERGEABLE`, opened 2026-09-01). Adds `scripts/check_iac_parity.py`, its unit tests, and a `make iac-parity-check` target. None of those exist on `main` today. |
| **Part C** — operator-owned `litellm-policy` | Not started. `git grep litellm-policy` hits only the chart, the Kustomize base, and two examples; no operator code generates it. The ownership question in Part C is still the gate. |

Resolved since filing:

| Item | Landed in | Commit |
| --- | --- | --- |
| A1 — settle the Dataplane V2 node-`/32` machinery | #777 | `1f27c969` |
| A2 — decide port 987 on the ALTS evidence | #805 | `f06fd575` |
| A3 — remove the node-IP claim from user-facing prose | #777, #805 | `1f27c969`, `f06fd575` |
| A4 — fix the rule-2 comment contradiction | #805 | `f06fd575` |
| A5 — fix the rule-1 headline | #805 | `f06fd575` |
| A6 — add `169.254.169.252/32` to the file-inventory bullet | #805 | `f06fd575` |
| A7, A8 | moot — both were blocked on A1 and die with the machinery |  |
| B2 — Tier 1 profile extraction | #824, widened by #857 and #1064 | `2e425cbd`, `1cd7a526`, `d509d975` |
| B3 — Tier 2 typed CRD config and status | #857 | `1cd7a526` |
| B4 — Tier 3 discover the metadata daemon | #1064 | `d509d975` |
| D2 | withdrawn by this issue on 2026-08-18 — the test already existed |  |

**The reopening comment of 2026-09-01 is superseded by this update.** It listed A2, A4, A5, A6, B3 and the `github-minter.yaml` A3 leftover as outstanding; #805 merged 2026-08-27 and #857 merged 2026-08-28, so all six were already on `main` when it was written. Its remaining accurate points — that B5/D1 has no home on `main`, that Part C is untouched, and that the static-copy count is eight rather than seven — are carried into the sections below.

[...]

PR #805, opened 2026-08-19
fix(netpol)!: close metadata egress port 8080, record the ALTS deviation

shalinibhatiapl:
## Summary

Closes port 8080 on the Platform Agent's metadata egress rule, and writes the reason for the closure into every NetworkPolicy surface and the two user-facing docs pages. Port 8080 pre-NAT and port 987 post-NAT are the two ends of a single leg — grpc-go's hardcoded ALTS handshaker address — and the operator's rule 2 was the only policy in the repository that opened either of them; the seven sibling policies have always allowed port 80 alone. No kube-agents component takes a gRPC DirectPath route, so the port admitted a handshake nothing performs. Alongside the closure the PR replaces the self-contradictory metadata comments ("Nothing rewrites a request to these ports onto another address", "Every datapath rewrites it before the policy is evaluated") with a pre-NAT/post-NAT explanation, drops an overstated rule-1 headline, adds `169.254.169.252/32` to the kustomize file inventory, and records the deviation from Google's allow-both guidance together with that guidance's auto-upgrade caveat.

This PR was opened as documentation-only. The port closure landed later, in answer to a review finding, and the title and every section below have been rewritten against the current head.

## Why This Change

Issue #747 A2 found that Google's GKE network-policy guidance recommends allowing `169.254.169.252/32` on ports `988` and `987` (iptables) and `169.254.169.254/32` on ports `80` and `8080` (eBPF), and that kube-agents opened 988 and 80 but also 8080 — one half of the ALTS pair, on one dataplane. A2 said keeping one end without the other is incoherent and offered two ways out: open 987 as well, or close 8080. It asked for whichever was chosen to be recorded as a deliberate deviation, naming the guidance and its auto-upgrade caveat.

Closure is the symmetric answer. Left as it was, the operator's gateway policy permitted an ALTS handshake to reach the metadata server on Dataplane V2 and blocked it on Dataplane V1, while the comment above the rule claimed the opposite — that the port was "kept for Dataplane V2 parity" and that least privilege was being enforced. A2 also asked for A4's rule-2 comment contradiction to be settled in the same place, because that comment is the one that will talk the next reader out of fixing the next instance.

## What Changed

- **Port 8080 removed from metadata egress rule 2 — a behaviour change.** `k8s-operator/internal/controller/platformagent_manifests.go` (the operator's `buildNetworkPolicy`), `deploy/kustomize/platform/networkpolicy-core-egress.yaml` (the static base), all four platform golden files, and the unit-test assertions. Existing installs are affected on upgrade, not just new ones — see **Risk & Rollout**.
- **All three metadata comments in `buildNetworkPolicy` now explain pre-NAT matching** instead of asserting that nothing rewrites the destination. Rule 2's comment cross-references rule 3 for why 8080 is absent, so an agent editing rule 2 sees the reason without reading ten lines down; the `linkLocalPeers` comment 85 lines above — the natural entry point, since that is where the peer list is built — no longer tells the reader the link-local rule can never match on a pre-NAT datapath. (#747 A4)
- **The rule-3 comment records the deviation and its cost.** It names Google's allow-both guidance, says both ALTS ports are deliberately omitted, and now also carries the guidance's warning that workloads omitting them "might experience disruptions during auto-upgrades", with the symptom to check for. `git grep -i auto-upgrade` over `docs/ deploy/ charts/ k8s-operator/` previously returned nothing. (#747 A2)
- **Rule 1's headline lost its false scope claim.** It read "Scoped specifically to CoreDNS and NodeLocal DNSCache pods in kube-system" over a rule that also carries `10.96.0.10/32` and a `0.0.0.0/0`-with-`except` block. (#747 A5)
- **The kustomize file inventory names both metadata peers**, `169.254.169.254/32` and `169.254.169.252/32`. (#747 A6)
- **A snapshot-independent assertion that the metadata server is never reachable on 8080.** `TestBuildNetworkPolicy_MetadataDaemonPeers` now collects, across every egress rule, the ports the policy opens towards `169.254.169.254/32` and requires exactly `{80, 988}`. The goldens cannot carry this invariant on their own: they are snapshots, and `go test -update` re-blesses whatever the code emits.
- **The two GitHub-minter policies had their metadata comments moved onto the rules they describe.** `charts/kube-agents/templates/github-minter.yaml` and `k8s-operator/config/integrations/github/deployment.yaml.template` carried a single header labelled "(port 988)" sitting directly on the **port-80** rule, telling the reader it was "matched by the port-80 rule above" when there was no port-80 rule above and the 988 rule was two entries below. Split the way the five sibling policies already do it.
- **Docs.** `deploy/kustomize.md` and `reference/security-and-iam.md` describe the narrowed allowlist, the deviation, and the auto-upgrade caveat, and finish the "rewrites" → "DNATs" wording sweep — including its last holdout, the `metadataDaemonIP` doc comment in `platformagent_controller.go`.

## Context

Part of #747 — it settles Part A items A2, A4, A5 and A6, and no more. Deliberately **not** a closing keyword: the issue still carries A7 (a table test for the Node watch predicate), A8 (documenting the `nodes` list/watch startup dependency), B4 (discovering the metadata daemon; `metadataDaemonIP` is still a hardcoded constant), B5 (a parity check across the static NetworkPolicy copies), C (an operator-owned LiteLLM policy) and D1 (a DNS peer-set rendering assertion), so auto-closing it on merge would lose all six.

- #777 settled A1/A3 (the Dataplane V2 node-`/32` claim). This PR removes the last carrier of that claim, in `github-minter.yaml`.
- **Rebase hazard, open PRs.** Three, all resolving the same way — toward this PR's wording and toward the narrowed port list.
  - **#963** (`feat(operator): egressPolicy Allowlist`) is the widest. It bakes `169.254.169.254/32 on TCP 80 and 8080` into the `egressPolicy` field's kubebuilder doc comment in `k8s-operator/api/v1alpha1/common_types.go` and therefore into the CRD `description` in both `charts/kube-agents/crds/kubeagents.x-k8s.io_platformagents.yaml` and `k8s-operator/config/crd/bases/kubeagents.x-k8s.io_platformagents.yaml` — text that ships to every user's cluster. It repeats the pair in `docs/security-requirements.md`, `docs/site/src/content/docs/operator/platformagent-crd.md`, `docs/site/src/content/docs/reference/credential-isolation.md` and a comment in `platformagent_egress_policy.go`, and it adds a **fifth** golden, `platformagent-egress-allowlist.yaml`, whose metadata rule carries `port: 8080`. It also edits `reference/security-and-iam.md`, which this PR rewrites.
  - **#857** rewrites `platformagent_manifests.go`, `platformagent_controller_test.go`, and `deploy/kustomize.md`, and its current diff of that docs page still carries the pre-#805 wording (`the node rewrites`, "the `169.254.169.254/32` rule on ports `80`/`8080`"). It does not touch rule 2's port list in code.
  - **#504** touches the platform goldens and `deployment.yaml.template`.

  Whichever lands second conflicts; a careless resolution restores `8080` in prose, in shipped CRD text, or in a golden while the code has closed it. The new assertion in `TestBuildNetworkPolicy_MetadataDaemonPeers` catches the golden case; nothing mechanical catches the prose.

## Testing

- `go build ./...` and `go vet ./internal/controller/` in `k8s-operator/` — clean at head `d4a42dea`. `go vet` compiles the test package, so the new assertion and its helper build. `gofmt -l ./internal/controller/` reports only `manifest_helpers.go`, which is pre-existing on `main` and not in this diff.
- `go test` — **not runnable on this workstation, in either direction.** `./internal/controller/ -run TestBuildNetworkPolicy` and `./internal/testing/` both die with `signal: killed` after ~0.5s, and so does a pre-built test binary run directly (exit 137, no output) — as does a hello-world `go run` in `/tmp`. No Go binary executes here at all, so this is the environment and not a test result. CI's "Run Controller Tests" and `build` jobs are the authority, and the former is green on head `d4a42dea`. **The new golden baseline was therefore hand-edited rather than produced by `-update`**; see below for how it was checked.
- **The fourth golden was verified by inspection, then by CI.** After the merge from `main`, `git grep -n "port: 8080" k8s-operator/internal/testing/testdata/platform/expected/` returns only LiteLLM (`80/4000/8080`) and GitHub-minter rules, and the metadata rule — the one whose `to:` is `ipBlock: cidr: 169.254.169.254/32` — is byte-identical across all four goldens (`ports: [80/TCP]`). Each golden contains exactly one NetworkPolicy and two occurrences of the link-local address. **"Run Controller Tests" then passed on head `d4a42dea`** ([run 33067830540](https://github.com/gke-labs/kube-agents/actions/runs/33067830540)), which is what actually executes `TestAgentsGolden/PlatformAgentSplitCredentialBroker` and the new port assertion.
- `make docs-check` — pass (generated regions current, 272 files' links resolve, 242 docs mapped, instruction budget 0.2k under).
- `make chart-check` — pass (chart CRD, RBAC and admission-policy copies in sync with `k8s-operator/config`).
- `helm lint charts/kube-agents` — 0 charts failed.
- `prettier --check` at the pinned 3.9.6 on every changed Markdown/YAML/JSON file — pass.

### Live validation

**Install:** `platform-agent-host` in project `kube-agents-gkedemos`, region `us-central1`. Datapath provider `ADVANCED_DATAPATH` (**GKE Dataplane V2**, control plane `1.35.6-gke.1641000`), Autopilot. Operator and agent both at `ghcr.io/gke-labs/kube-agents/*:57f9f4c2d18546ad2b92d3ab18bc04774ae15c80`. Exercised 2026-08-26.

Dataplane V2 is where this change has an effect: policy is evaluated pre-NAT, so port 8080 was genuinely reachable there, unlike on Dataplane V1 where the leg is inert post-DNAT. The live `platform-agent-gateway-netpol` on this cluster still carries the pre-change rule 2 (`80` and `8080`), which is what makes it a valid before/after subject.

Rather than mutate the shared install, the three policy shapes were applied to a throwaway namespace (`netpol-805-test`) against probe pods whose results are readable from `kubectl logs`:

| Rule-2 shape applied         | `169.254.169.254:80` | `169.254.169.254:8080` | `computeMetadata/.../zone`                  | ADC token fetch |
| ---------------------------- | -------------------- | ---------------------- | ------------------------------------------- | --------------- |
| `80` + `8080` (`main` today) | OPEN                 | **OPEN**               | —                                           | —               |
| `80` only (**this PR**)      | OPEN                 | **BLOCKED**            | 200, `projects/663211013060/zones/us-central1-a` | 200, 1407 bytes |
| neither (control)            | BLOCKED              | —                      | timeout                                     | timeout         |

What each row establishes:

1. **The port was live, not inert.** Under the shape `main` ships today, a TCP connect to `169.254.169.254:8080` succeeds on this cluster. This is a real behaviour change on Dataplane V2, which is why the title is `fix!` and not `docs`.
2. **Workload Identity still works with 8080 closed.** Under this PR's shape, the metadata endpoint and the ADC token endpoint both return over port 80.
3. **Row 2 is not a coincidence.** The control removes port 80 as well and the same probe times out, proving the policy is genuinely enforced on these pods — so row 2's success is the port-80 rule admitting the fetch, not the policy failing to apply.

Reverting from row 2 back to row 1 restored `PORT_8080=OPEN`, and re-applying row 2 returned it to `BLOCKED`, so the port list is the mechanism.

**This still reaches the head.** The branch commits added since (the fourth golden's baseline, three comment lines, one test assertion) change no rendered rule, port, or peer, and neither does the merge from `main` that came with them: extracting the `NetworkPolicy` document from each of the three goldens that existed at the validated commit and diffing it against head is **identical** in all three. The measured shape is the shape the branch still emits.

**Not covered, and why:**

- **The operator's own reconcile of the narrowed policy was not observed on a cluster.** That needs an operator image built from this branch pushed to a registry the cluster can pull, which is beyond what this pass could reach. The rendered output is pinned by the four golden files and by the new port assertion, and `applyManaged` uses `client.Apply` + `client.ForceOwnership` (`platformagent_controller.go:402-405`), so the upgrade path is read from the code rather than watched. The dataplane consequence of that output is what the table above measures.
- **Dataplane V1 was not exercised.** Post-DNAT the ALTS destination is `169.254.169.252:987`, which rule 3 has never permitted, so 8080 was already dead there; there is no before/after to observe.
- **The lease** (`docs/designs/live-test-lease.md`) was not taken. Nothing in `kubeagents-system` was created, patched, or deleted — the whole test lived in its own namespace — so there was no state another agent's validation could collide with.

**Cleanup:** the `netpol-805-test` namespace and all four probe pods were deleted; `kubectl get namespace netpol-805-test` returns NotFound. `platform-agent-gateway-netpol` was re-read afterwards and still carries `[{80 TCP},{8080 TCP}]` on rule 2 — the install is exactly as it was found.

## Self-Review

Three adversarial passes have run, each **in a separate subagent context that did not write the change**, handed a diff range and nothing else: the docs-only head, the head after the port closure, and a third against the current head merged with `main`. A dedicated docs-drift pass ran over the first two ranges; over the third, prose-against-source was covered inside the adversarial pass itself (it is where the `linkLocalPeers`, `node rewrites` and citation-drift findings came from) rather than by a separate subagent. The mechanical `make docs-check` ran over all three. Angles covered and their dispositions, merged:

- **Interaction with base commits landed after the branch was last pushed.** The third pass found the branch's one blocker, and it was invisible from the branch's own diff: #960 added a fourth platform golden, `platformagent-split-broker.yaml`, which renders through the same `buildNetworkPolicy` and still pinned `port: 8080` on the metadata rule. The merge produces no conflict — the file is new to the branch — so nothing warns, and the PR's green "Run Controller Tests" had run against a base predating #960. **Fixed** in `3f3f9c7b`: `main` merged in, the fourth golden re-baselined, and its metadata rule checked against the other three.
- **Change classification / removed-behaviour audit.** Found the largest problem of the first pass: a NetworkPolicy rule change shipping under a `docs:` title, in a body whose Risk section said "No Kubernetes NetworkPolicy rules … are altered" and whose Self-Review said "No runtime port behavior changed". **Fixed** — retitled `fix(netpol)!`, and Summary, What Changed, Risk & Rollout and this section rewritten against the current head.
- **Claims the diff does not support.** Five across the passes. The Testing section credited `python3 scripts/check_iac_parity.py — PASS (16 parity checks passed)`; that script does not exist on this branch or on `main`. The Summary credited a `golden_test.go` thread-safety refactor that reached `main` separately via #922. Live validation recorded a run taken before the port closure, whose only measurement was 0 hits on port 987 — a port that never appears on Dataplane V2 by the change's own model. The third pass retired two more: "the three platform goldens" and a `git grep 8080` sweep that had been clean when written and was not at the merge target. **All fixed**; the counts and the sweep are re-run above against the merged tree.
- **Cross-file tracer over every NetworkPolicy copy.** `git grep 8080` over `docs/ deploy/ charts/ k8s-operator/ INSTALL.md` leaves no surface describing a metadata-8080 allow, and `git grep "per node\|node's internal IP"` is empty, so the sweep is complete and #777's repudiated claim has no carrier left. It did surface that the sweep introduced a defect: the `github-minter.yaml` header, copied from `deployment.yaml.template`, sits on the port-80 rule while describing the 988 rule. **Fixed in both files.** It also found the operator's rule-3 comment had dropped a Dataplane V2 sentence the kustomize copy kept — **restored**.
- **Comment-placement / next-reader test.** Two hits. Rule 2 lost its port while the explanation lived in rule 3's comment worded "omitted **here**", where "here" is the 988 rule — **fixed** with a cross-reference on rule 2. The third pass then found a third copy of the A4 contradiction that the first two had walked past: the `linkLocalPeers` comment, 85 lines above rule 2, still said "Every datapath rewrites it before the policy is evaluated" — false on Dataplane V2, and contradicted by both the new rule-2 comment and `metadataDaemonIP`'s own doc comment. **Fixed** in `0d786941`.
- **Test coverage.** The earlier passes accepted an argument that has since been disproved: that the goldens pin the full rendered policy, so a re-added 8080 fails CI and a second assertion would be duplication. That holds for a port re-added on this branch. It does not hold for a golden written by a sibling branch, or for anyone resolving a conflict with `go test -update` — the documented re-baseline route, named in **Revert** below — which blesses whatever the code emits. The blocker above is that case in the wild. **Reversed and fixed** in `d4a42dea`: `TestBuildNetworkPolicy_MetadataDaemonPeers` now asserts the ports opened towards `169.254.169.254/32` across all egress rules are exactly `{80, 988}`, independent of rule ordering and of any snapshot.
- **Stale references in files this PR already opens.** Three `80/8080` strings left in `platformagent_controller_test.go` (the rule-count message, the peer-count message, the FQDN rule inventory) describing a rule that no longer carries the port. **Fixed.**
- **Requirements traceability against #747.** A2 asked for Google's auto-upgrade caveat to be recorded alongside the deviation; `git grep -i auto-upgrade` returned nothing anywhere. **Fixed** in the rule-3 comment (both copies) and `kustomize.md`. The third pass separately caught that the **Context** section led with a GitHub closing keyword on #747 while B4, B5, C and D1 are outstanding, which would have auto-closed the issue on merge. **Fixed** — that line now reads "Part of", and names both what this PR does close and what stays open.
- **Docs prose against source.** `security-and-iam.md` still said "the node rewrites the metadata destination" on a line the branch was already editing, after the same phrase had been changed to "DNATs" in all six manifest comments and in `kustomize.md` — **fixed**. The third pass found the sweep's last holdout in `platformagent_controller.go:65`, the file that defines the constant a reader looking this up lands in — **fixed**; `git grep "node rewrites"` over `docs/ deploy/ charts/ k8s-operator/ INSTALL.md` is now empty.
- **Citation accuracy in this body.** The third pass found two line numbers that had drifted as `main` moved. `applyManaged` is at `platformagent_controller.go:402-405`, not `:359-361` — **fixed** in **Live validation** and **Risk & Rollout**. `google-cloud-pubsub` was re-derived on the merged tree and is at `deploy/docker/Dockerfile:116`, which is what the body already said — **no change**. Both claims were substantively correct throughout.
- **Evidence quality of the dependency audit.** The audit ruled out `google-cloud-storage`, `google-cloud-bigtable`, and `google-cloud-spanner`, but `deploy/docker/Dockerfile:116` installs **`google-cloud-pubsub`** — a gRPC client, and the only one in the image, used by `agentplugins/pubsub-platform/…/adapter.py` inside the pod this policy governs. The conclusion holds: the Python Pub/Sub client does not opt into DirectPath, and `git grep "ALTS\|DirectPath\|direct_path"` finds no client code taking that route. **Answered here rather than by a code change**, so the next auditor does not redo the search.
- **Sibling-PR interaction.** #857, #504 and #963 all collide with this branch; recorded under **Context** above with which way to resolve. #963 was the third pass's addition and is the most damaging, because its copy of the wider port list ships to user clusters inside the CRD `description`.
- **Not fixed, with the argument.** Rule 3's comment names `169.254.169.252` as the DNAT target without noting it has been overridable since #824. `platformagent_controller.go:64-75` already documents that ladder next to the constant; a second copy ten lines from the first is the duplication the Documentation Guidelines forbid, and it is not a claim this diff introduces.

Two things this list does not claim: no pass observed the operator reconciling the change on a cluster (see **Live validation**), and no Go test was run to completion locally, because no Go binary runs on the workstation at all (see **Testing**). The fourth golden's new baseline was hand-written and checked by inspection against the other three rather than generated by `-update`; CI's "Run Controller Tests" is what confirmed it, green on head `d4a42dea`.

## Risk & Rollout

**This tightens a security control on existing installs, not only new ones.** `applyManaged` patches with `client.Apply` + `client.ForceOwnership` (`platformagent_controller.go:402-405`), so the first reconcile after the operator is upgraded rewrites every `*-gateway-netpol` and port 8080 disappears from the metadata rule whether or not the cluster asked. Kustomize users pick the change up when they re-apply the base. That is the reason for the `!`: the repository is at `0.1.0`, where `scripts/release/calculate_next_version.sh` maps a breaking change to a **minor** bump (0.1.0 → 0.2.0) rather than a major one, so marking it costs one minor version and buys the release note that tells an operator their egress allowlist narrowed. Finding it in a changelog is the difference between a five-minute diagnosis and a long one.

**Who is affected.** Nothing kube-agents ships is: the live table above shows Workload Identity working with the port closed, no client in the image takes a gRPC DirectPath route, and on Dataplane V1 the leg was already inert. The exposure is a **user-supplied `AgentPlugin`** running in the agent pod that speaks gRPC DirectPath/ALTS to `storage.googleapis.com`, Bigtable, or Spanner — the operator, not this repository, is the only one who can know.

**What to check if that is you.** Enable policy logging on the gateway policy and look for drops to `169.254.169.254:8080` (Dataplane V2) or `169.254.169.252:987` (Dataplane V1):

```
kubectl annotate networkpolicy <name>-gateway-netpol \
  policy.network.gke.io/enable-logging=true -n <ns> --overwrite
```

Google's guidance additionally warns that workloads omitting these ports "might experience disruptions during auto-upgrades", so a token fetch that starts failing during a node auto-upgrade is worth checking against this allowlist before looking elsewhere. The rule-3 comment now says so in place.

**Revert.** Add `{Protocol: &tcp, Port: ptr.To(intstr.FromInt32(8080))}` back to rule 2 in `platformagent_manifests.go`, the matching `- port: 8080` in `networkpolicy-core-egress.yaml`, relax `wantPorts` in `TestBuildNetworkPolicy_MetadataDaemonPeers`, and re-baseline the four goldens with `go test ./internal/testing/ -update`. Nothing else in the change has runtime effect. As an in-place stopgap on a running cluster, patching the port back onto the live policy holds only until the next reconcile, which force-applies it away again.


charts/kube-agents/templates/litellm.yaml
```diff
@@ -263,9 +263,10 @@ spec:
         - ipBlock:
```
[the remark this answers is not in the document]
shalinibhatiapl: Updated the PR description's **Live validation** section with the specific image tags (operator `latest` and agent `dev-20260819-091655`), confirmed the cluster dataplane as **GKE Dataplane V2** (stating explicitly that Dataplane V1 behavior was not live-observed in this pass and is covered by code analysis), and confirmed the temporary `policy.network.gke.io/enable-logging` annotation was removed from the live NetworkPolicy.

AntonTyb (approved):

mplakhtiy (commented):
A full adversarial pass over `upstream/main...5aee031e`. **One HIGH, two MEDIUM** — four LOW findings are held back to keep this readable.

One is inline. The other two cannot be: one is about the PR description, and the other is about a file this PR did not change, which is the finding itself.

#### 🔴 High — the Testing section credits a script that does not exist

> `python3 scripts/check_iac_parity.py` — PASS (16 parity checks passed)

`scripts/check_iac_parity.py` is not in the repository, on this branch or on `main`. `git grep -rn "iac_parity"` returns nothing, `scripts/` contains no parity checker, and there is no parity target in the `Makefile` or any workflow. The only repo-wide hits for "parity" are unrelated prose and a Python test about GKE endpoint parity.

**Why this one matters more than a stray line:** it reads as independent evidence that the seven hand-synchronised NetworkPolicy copies were mechanically verified to still agree — exactly the property a reviewer of this change most wants confirmed, and exactly the property the next finding shows does not hold. A reviewer who takes the line at face value skips the manual sweep. AGENTS.md is explicit that a claim the repository does not support is a finding in its own right, and a more serious one than most defects, because it misdirects every reader after it.

Remove the bullet, or replace it with the check that was actually run.

#### 🟠 Medium — the Helm `github-minter` template was not swept, and is now the last carrier of a repudiated claim

`charts/kube-agents/templates/github-minter.yaml:246-251`:

```yaml
# Allow GCP Workload Identity / Metadata Server.
# The minter dials 169.254.169.254:80, but the node DNATs that to the metadata
# daemon on port 988 before NetworkPolicy is evaluated: 169.254.169.252 on the
# iptables datapath, the hosting node's internal IP on Dataplane V2. Permitting
# only the link-local address drops every token fetch. On Dataplane V2, add one
# /32 per node to the port-988 rule.
```

Two problems, and this PR is the natural place for both.

First, this is the Helm twin of `k8s-operator/config/integrations/github/deployment.yaml.template`, which this PR **did** update with the port-987 note at `:165` — so the two copies now disagree.

Second, `git grep` confirms this file is the **only** place left in the repository asserting the Dataplane V2 node-internal-IP claim and instructing operators to hand-maintain one `/32` per node. That is the claim #777 concluded was wrong and removed everywhere else, and which #747 A3 called out as worse-than-nothing in user-facing prose. #777 touched `charts/kube-agents/templates/litellm.yaml` but skipped this file.

The Self-Review's Angle C states that comments were "synchronized across operator manifests, Kustomize bases, Helm templates, integration manifests, and documentation site pages", so the gap sits directly under a claim that it does not.

**Failure scenario:** an operator installing the chart follows advice the repository has already concluded is useless, and adds per-node `/32` churn to a policy that does not need it.

Half of this (the stale DPv2 claim) is inherited from #777 rather than caused here, but both lines live in one comment block and are one edit.

deploy/kustomize/platform/networkpolicy-core-egress.yaml
```diff
@@ -63,9 +64,17 @@ spec:
...
+    #    169.254.169.252:987 before NetworkPolicy is evaluated, so this rule admits the
+    #    post-DNAT token fetch. On Dataplane V2 (eBPF), policy is evaluated pre-NAT at the
+    #    socket layer and matched by rule 2.
+    #
+    #    Google network policy guidance recommends allowing ports 988 (metadata) and 987 (ALTS)
+    #    to prevent disruption during auto-upgrades. Port 987 is deliberately omitted here
+    #    because kube-agents components use standard OAuth2/REST token fetches and no client
+    #    takes the gRPC DirectPath / ALTS route. Port 8080 in rule 2 therefore admits no completed
+    #    ALTS handshake on Dataplane V1 today; it is kept open for Dataplane V2 parity, where
+    #    policy evaluates pre-NAT. Keeping port 987 closed enforces least-privilege sandbox egress.
```
mplakhtiy: 🟠 **Medium** — "Dataplane V2 parity" describes the opposite of what these rules produce.

The comment closes with: *"Port 8080 in rule 2 therefore admits no completed ALTS handshake on Dataplane V1 today; it is kept open for Dataplane V2 parity, where policy evaluates pre-NAT. Keeping port 987 closed enforces least-privilege sandbox egress."*

Google's guidance — which this comment cites, and which I checked verbatim against the live docs page — pairs the ports by dataplane rather than listing four independent ports: `169.254.169.252/32` on **988 and 987** for the iptables datapath, `169.254.169.254/32` on **80 and 8080** for Dataplane V2. 8080 and 987 are the same ALTS leg seen from the two sides of the DNAT. That is why #747 A2 said keeping one without the other is incoherent on the iptables datapath, and asked for both to be decided in one place.

Under the model this comment itself states — DPv2 evaluates pre-NAT, so rule 2 admits both directly — the shipped policy **permits an ALTS handshake to reach the metadata server on Dataplane V2 and blocks it on Dataplane V1**. That is an asymmetry, not parity, and the closing least-privilege sentence is true only of the iptables datapath.

The mirrors inherit the same implication and would need the same edit: `k8s-operator/internal/controller/platformagent_manifests.go:3707-3712`, `docs/site/src/content/docs/deploy/kustomize.md:62`, and `docs/site/src/content/docs/reference/security-and-iam.md:168`. Line 62 of `kustomize.md` says the `80`/`8080` rule "satisfies it directly" on DPv2 and then, in the next sentence, that 987 is "intentionally omitted under least privilege" — a reader concludes ALTS is shut everywhere.

**Failure scenario:** the next person to audit egress reads this as "ALTS is closed", and either does not notice that a DPv2 install still permits the handshaker, or reopens the whole question because the comment does not match the rules. Stopping exactly that is what this comment is for.

The fix is a decision rather than necessarily a code change: either drop 8080 from rule 2 so the posture matches on both dataplanes, or keep it and say plainly that ALTS remains reachable on Dataplane V2 via the pre-NAT 8080 rule and that the least-privilege claim applies to the iptables datapath. "Parity" is the wrong word for the current state whichever way it goes.

kirill-kosov: Fixed

mplakhtiy: Taken the first option: 8080 is closed on rule 2 rather than reworded, so the posture is now the same on both dataplanes and "parity" is gone along with the claim.

Landed across the surfaces you named plus the ones that mirror them — `platformagent_manifests.go`, this file, the three platform goldens, `TestBuildNetworkPolicy_MetadataDaemonPeers`, `kustomize.md:62`, and `security-and-iam.md`. `git grep 8080` over `docs/ deploy/ charts/ k8s-operator/ INSTALL.md` leaves no surface describing a metadata-8080 allow.

Follow-ups in 42d76b9b, after a second adversarial pass on the post-fix head:

- Rule 3's comment now carries Google's auto-upgrade caveat, which #747 A2 asked for and which `git grep -i auto-upgrade` showed was nowhere in the tree. The deviation had widened from one port to two, so it matters more here than when A2 was written.
- Rule 2 cross-references rule 3, so an agent editing rule 2 — #857 rewrites exactly that function — sees why 8080 is absent without reading ten lines down.
- The operator's rule-3 comment had dropped the Dataplane V2 sentence this file kept; restored, so the two copies agree again.

Verified on a live Dataplane V2 cluster rather than argued from the model: under the shape `main` ships today a TCP connect to `169.254.169.254:8080` **succeeds**, under this PR's shape it is dropped, and the ADC token fetch keeps returning 200 over port 80 throughout. A control that closes port 80 as well times out, so the enforcement is real. Full table in the PR body's Live validation section.

The PR is retitled `fix(netpol)!` and the body rewritten — it was still describing itself as documentation-only, which your review body's HIGH is answered by in a separate comment.

mplakhtiy:
Taking this PR over from @shalinibhatiapl and @kirill-kosov to clear the outstanding findings. Answering the two from my 2026-08-25 review body that had no thread to reply in, plus the four LOWs held back there (a second adversarial pass on the post-fix head surfaced them independently).

**🔴 High — `scripts/check_iac_parity.py` does not exist.** Correct, and it had survived one round unfixed. The bullet is gone. The Testing section now lists only what was actually run: `go build`, `go vet`, `gofmt -l`, `make docs-check`, `make chart-check`, `helm lint`, and `prettier --check` at the pinned 3.9.6 — with `go test` recorded as **inconclusive locally** (`signal: killed` at ~0.4s on this workstation, not a test failure) and CI named as the authority. The mechanical property the phantom script was standing in for is covered instead by `git grep 8080` and `git grep "per node\|node's internal IP"` over `docs/ deploy/ charts/ k8s-operator/ INSTALL.md`, both empty, and by the three goldens pinning the rendered policy.

Same pass caught a second unsupported claim in the same body: the Summary credited a `golden_test.go` thread-safety refactor that is not in this diff — `newTestScheme` reached `main` separately via #922. Removed.

**🟠 Medium — `github-minter.yaml` not swept.** Fixed in 38fd171d, and the fix is larger than the sweep, because the sweep itself introduced a defect. Removing the Dataplane V2 node-`/32` sentences (correct — this file was the last carrier of the claim #777 repudiated) replaced them with a header labelled "(port 988)" that says *"this rule admits the post-DNAT token fetch"* and *"matched by the port-80 rule above"* — while sitting directly on the **port-80** rule, with no port-80 rule above it and the 988 rule two entries below.

The wording was copied verbatim from `k8s-operator/config/integrations/github/deployment.yaml.template`, where the same misplacement was already there, so the sweep left two files consistently wrong instead of fixing one. Both are now split the way the five sibling policies already do it: a port-80 line on the port-80 rule, the 988/987 text on the 988 rule. Your Angle C point stands — the gap sat under a claim of synchronisation — so this is exactly the class #747 A4 exists to remove.

**The four LOWs**, since the second pass found them too:

1. *"rewrites" at `security-and-iam.md`* — fixed; the DNATs sweep is now complete.
2. *8080→987 DNAT evidence not cited* — the load-bearing claim is now measured rather than argued: the Live validation table shows 8080 open on a real Dataplane V2 cluster under today's shape and blocked under this one.
3. *Daemon IP overridable since #824* — **not fixed, deliberately.** `platformagent_controller.go:64-75` already documents that ladder next to the constant; a second copy ten lines away is the duplication the Documentation Guidelines forbid, and it is not a claim this diff introduces.
4. *`golden_test.go` duplicated in #857* — resolved itself; #922 landed it on `main` and it is no longer in this diff.

Two more from the second pass, both fixed: three stale `80/8080` strings in `platformagent_controller_test.go` (the file this PR already opens), and #747 A2's auto-upgrade caveat, which `git grep -i auto-upgrade` showed was absent from the entire tree.

**What still needs a decision from a reviewer, not from me:**

- **The approval on record does not cover this change.** @AntonTyb approved `5aee031e`, the docs-only head; every runtime effect landed after it. The PR is now `fix(netpol)!` — `!` because `applyManaged` uses `client.Apply` + `client.ForceOwnership`, so existing installs lose the port on upgrade, and because at `0.1.0` a breaking marker costs one minor version (`calculate_next_version.sh` maps it to 0.1.0 → 0.2.0, not a major) and buys the release note. Happy to drop the `!` if you read it as too strong.
- **The operator's own reconcile was not observed on a cluster** — that needs an image built from this branch in a registry the cluster can pull. The dataplane consequence is measured; the upgrade path is read from the code. Stated as such in Live validation.

`/review` is coming next for a strict pass at the new head.

mplakhtiy (approved):

PR #71, opened 2026-06-02
feat(skills): integrate agent and shared skills with Hermes sync

bradhoekstra:
# Pull Request

## Why This Change

This PR addresses the need to integrate the GKE platform agent skills (and placeholder skills for operator and devteam agents) with the Hermes skill syncing system. Without this integration, skills are not automatically synchronized and manifest-tracked on startup, which prevents seamless updates and safe user customization management.

## What Changed

We updated the Dockerfiles for the `platform`, `operator`, and `devteam` agents to copy both the shared and agent-specific skills into the container's bundled skills directory (`/opt/hermes/skills/`).

**Files:**

- `agents/platform/Dockerfile`
- `agents/operator/Dockerfile`
- `agents/devteam/Dockerfile`

**Added/Changed/Fixed:**

- Added `COPY` instructions to copy shared skills from `agents/shared/skills/` to `/opt/hermes/skills/` in all three Dockerfiles.
- Added `COPY` instructions to copy agent-specific skills to `/opt/hermes/skills/` in all three Dockerfiles.
- Ensured ownership of the copied skills is set to `hermes:hermes` to match the runtime user.
- Confirmed that `.gitkeep` files were moved to the correct skills directories.

## Why This Matters

By bundling the skills in `/opt/hermes/skills/`, the base Hermes agent startup sequence (specifically `stage2-hook.sh`) will run `skills_sync.py` to copy them to the user's persistent volume (`$HERMES_HOME/skills/`) and register them in `.bundled_manifest`.
This allows:
- Automatic loading of custom GKE skills by the agents.
- Safe updates: if the image is updated with new skills, Hermes will automatically update them for the user *unless* the user has customized them locally, in which case it safely skips to prevent overwriting user changes.

## Context

- Integrates with Hermes skill syncing system (NousResearch/hermes-agent).
- Specifically requested to make sure GKE platform agent skills are integrated.
- Also ensured shared skills are copied and can be overridden by agent-specific skills.

## Testing

- Verified that `stage2-hook.sh` in `nousresearch/hermes-agent` (v2026.5.29.2) executes `tools/skills_sync.py` automatically.
- Verified that `skills_sync.py` correctly handles directory copy and manifest tracking.
- Verified that the Dockerfile changes copy the skills to the expected location (`/opt/hermes/skills/`) with the correct ownership.

PR #809, opened 2026-08-19
perf(docker): build credential-proxy from agent-base, not platform

toshiowang:
## Summary

The `credential-proxy` sidecar image built `FROM platform`, so it inherited the entire Platform Agent chain to use one directory from it. `COPY agents/*/scripts/` sits two thirds of the way down the `platform` stage, which meant editing a single agent Python file invalidated that layer and every layer after it in **both** images — and the sidecar's apt/`gh` install, the most expensive instruction in the file at ~300MB over the network, sat above all of that churn and got rebuilt too. This adds a `proxy-tools` stage that builds `FROM agent-base` and holds the install, the Envoy binary, its config and `start-services`; `credential-proxy` is then a single COPY on top of it. On the case that motivates this — edit one file under `agents/platform/scripts/`, rebuild — the sidecar's Cloud Build goes from 13m21s to 6m41s.

## Why This Change

`k8s-operator/scripts/dev/dev_rebuild_agent.sh` is the inner loop for anyone working on the agent, and it takes 20-25 minutes. Roughly half of that is the credential-proxy build, and almost all of that half was work the agent build had just done: the two submits are independent Cloud Build runs on separate cold workers, so under the old shape both of them rebuilt the tail of the `platform` chain from the same invalidated COPY, and only one of them needed to.

Left alone this also gets worse rather than better. Every instruction added to `platform` is one the sidecar pays for on every agent source edit, and it counts against the layer budget: at the merge base, `credential-proxy` was the deepest chain in the file at 118 layers against a 120 budget, which is why the gate in `scripts/check_image_layers.py` was pointed at it. #731 landing two more instructions into `agent-base` while I was measuring is exactly the pressure this removes.

## What Changed

- **`deploy/docker/Dockerfile`** — new `proxy-tools` stage `FROM agent-base` carrying the apt/`gh` install, the Envoy binary, its config and `start-services`; `credential-proxy` becomes `FROM proxy-tools` plus one COPY. Order within `proxy-tools` is cache order: the install is first because it is the most expensive and the least likely to change. `mkdir -p /etc/envoy` moved into the install RUN, replacing a `chmod 755 /etc/envoy` that had to run _after_ the config COPY and therefore pinned the install back above the churn — a `--chmod`ed COPY propagates its mode to parent directories it creates, and 0644 on `/etc/envoy` is not traversable by the runtime uid.
- **`scripts/check_image_layers.py`** — retargeted from `credential-proxy:latest` to `platform-agent:latest`. The sidecar is now 101 layers and `platform` is the deepest chain that ships at 114, so the gate would otherwise have been measuring the short chain and reporting headroom that is not the headroom at risk. This docstring is now the single home for which image the gate points at and why; the Dockerfile header, the workflow comment and the AGENTS.md bullet defer to it instead of restating it.
- **`scripts/test_check_prompt_assets.py`** — `_copy_instructions()` is now stage-aware, so `OPT_DEFAULTS` is re-derived from the `platform` chain alone rather than from every `/opt/defaults` COPY in the file. Plus a new test asserting the two stages fill `/opt/defaults/scripts` from the same sources (see Self-Review).
- **`.github/workflows/docker-build.yml`** — `load: true` moved from the sidecar build to the platform build so the layer-budget step has the image it now inspects.
- **`deploy/docker/cloudbuild-ci.yaml`, `hack/ci-deploy.sh`, docs** — comments and prose that described the old shape. The CI build's step graph and cache seeding are untouched; see Context for why.

## Context

Not tied to an issue. Overlap from the scan of 48 open PRs:

- **#739** (`fix(ci): isolate presubmit image tags to prevent cache pollution`, @lapis2002) merged one minute before this PR opened, and it is why the `cloudbuild-ci.yaml` change originally in this branch is no longer here. I had split `warm-cache` in two and given the sidecar step its own `--cache-from $_PROXY_LATEST`; #739 deleted the `:latest` pushes and the `_PLATFORM_LATEST`/`_PROXY_LATEST` substitutions entirely, replacing them with one canonical `_CACHE_IMAGE` from `kube-agents-prow`. Rebasing on top of it, I took #739's version wholesale rather than reworking my seeding on top — see below for why that turns out to be right on the merits, not just the cheap resolution.
- **#674** (@AntonTyb) edits the same `docker-build.yml` region and keeps `load: true` on the sidecar build. Textual conflict only.
- **#738**, **#728**, **#720**, **#713** each touch `deploy/docker/Dockerfile` elsewhere. Merge-conflict warnings, not duplicates.

Rebased onto `95a7c04d`. The only conflict was the `agent-base` comment block that counts the Hermes patches: `main` added a fifth (`kanban_event_routing`) while this branch rewrote the sentence saying `credential-proxy` builds `FROM platform`. Resolved by keeping both — five patches, and the corrected statement that the stage is the base of the `platform` chain. No instruction changed, and the layer figures quoted above were re-derived on the new base and are unchanged (`platform` 114, `credential-proxy` 101, `entrypoint-gate-test` 116): `main`'s 18 commits added no layer-producing instruction to either chain, since the fifth patch rides in the existing COPY/RUN pair.

#731 merged during this work too and consolidated two script COPYs in the `platform` stage; I took its version in that conflict, since it is a layer cheaper than mine.

**Why CI needs no cache change.** The apt/`gh` install has to run in CI either way: under the old shape it lived in the `credential-proxy` stage, above `platform`, and `platform`'s cache never covered it. Moving it into `proxy-tools` does not change what CI has to build — the sidecar step still inherits its cached ancestor from the platform step on the same worker, and that ancestor is now `agent-base` rather than the `platform` image. So the step graph and the seeding stay exactly as #739 left them; what changes is only the prose in `cloudbuild-ci.yaml` and `hack/ci-deploy.sh` that named `platform` as the thing being inherited.

I also dropped the parallelism I had proposed there. Running the two agent builds concurrently on one worker only pays off if BuildKit dedups the shared `agent-base` vertex across concurrent solves, and I cannot verify that without a container runtime (see Self-Review). Sequential is what is measured and what works; the speculative version does not belong in the same PR as the change that is.

**Deliberately not in this PR:** collapsing `dev_rebuild_agent.sh`'s two sequential `gcloud builds submit` calls into one build with parallel steps. It is drafted and stacks on this branch. Splitting them keeps this PR to one argument, and the parallel version is only worth what it is worth _because_ the sidecar build is no longer bound to the agent build's chain — sequenced the other way round it would have been `max(617s, 801s)` instead of `max(617s, 401s)`. `.github/workflows/docker-publish-gcp.yml` has the same three-sequential-submits pattern and is a candidate for the same treatment; flagging it rather than fixing it here.

## Testing

- `python -m unittest test_check_prompt_assets test_image_layers` — 57 tests, OK.
- `make docs-check` — all four checks (generated regions, links, terminology, map coverage) across 251 docs.
- `prettier --check` (3.9.6, the version `prettier.yml` pins) — clean on the changed Markdown/YAML.
- `make test-python` has one failure, `tests/test_docker_entrypoint.py::test_step_2_6b_fills_an_existing_config_without_overruling_it`. Confirmed pre-existing by stashing this branch and re-running on a clean tree: same failure, and it exercises `deploy/shared/docker-entrypoint.sh`, which this PR does not touch.
- **Not run: `scripts/check_image_layers.py` itself, and no local `docker build`.** There is no container runtime on my machine. Layer counts quoted here come from the registry manifests (`rootfs.diff_ids`) of the images the builds actually pushed, cross-checked against a parser over the Dockerfile that reproduces those registry figures exactly at two separate commits. CI runs the real check.

### Live validation

Install: `platform-agent-host` (GKE, project `toshiowang-gkedemos`, `us-east4`), via `dev_rebuild_agent.sh`. All timings EDT, from `gcloud builds list`.

**The measurement.** Baseline and after both taken by editing one file under `agents/platform/scripts/` (a comment marker in `session_manager.py`, reverted afterwards) and rebuilding. A rebuild with no source change would prove nothing, since the whole claim is about what a source change invalidates.

| run                                      | platform           | credential-proxy      |
| ---------------------------------------- | ------------------ | --------------------- |
| baseline                                 | `08443ebb` 10m17s  | `15ce24ff` **13m21s** |
| after, first build (cold `proxy-tools`)  | `9e55624f` 7m03s   | `2d34869f` 9m14s      |
| after, steady state (fresh one-file edit)| `d193791b` 11m44s  | `a45f5eef` **6m41s**  |

Per-step for the sidecar (cache pull / `docker build` / push): baseline 115s / **593s** / 88s; steady state 122s / **262s** / 12s. In the steady-state build every BuildKit node reported `CACHED` except `#71 [credential-proxy 1/1] COPY … /opt/defaults/scripts/`, at **0.9s** — the remaining 262s is Cloud Build pulling the cache image and BuildKit materializing and exporting layers, not compiling anything.

Two honest caveats on the table. The `platform` column is noise: its instruction list is unchanged by design, and 11m44s is contended — another session was building against the same project and clobbered `:latest` partway through. I checked that my steady-state sidecar build started at 17:13:32, before that session's old-chain `credential-proxy:latest` push at 17:16:58, so it was cache-clean. And the one-time transition build is the real cost of the cutover: the first build after this merges gets a `:latest` describing the old chain, misses `proxy-tools`, and pays the apt/`gh` install once.

**The sidecar still works.** Deployed `credential-proxy:dev-20260819-130136` (the steady-state image) to the running Pod. The `envoy-credential-proxy` container reached Ready, and all three services it is responsible for are up:

```
1  bash /usr/local/bin/start-services
6  python3 /opt/defaults/scripts/credential_proxy.py
7  envoy --config-path /etc/envoy/envoy-credential-proxy.yaml
10 /usr/local/bin/k8s-event-watcher --cluster-name=platform-agent-host ...
```

The event watcher is the one that could have been silently dropped, since it runs in this container. Its binary is copied in `agent-base`, not `platform`, so it comes along — confirmed running above rather than assumed.

**Negative control, in both directions.** The platform-only assets are gone from the sidecar (`/opt/platform-template`, `/opt/chat-template`, `/opt/cluster-template` all `No such file or directory`) and still present in the agent container. `/opt/defaults/scripts` is present in the sidecar with the same contents. I also exercised the real path end to end: proxied `kubectl`, `gcloud` and `gh` calls from the sandbox container through the sidecar all returned successfully (run from `/opt/data` — the wrappers reject a cwd outside the shared workspace).

**Not covered.** No local `docker build` or layer-check run, per Testing above. Nor a real CI run of `cloudbuild-ci.yaml` — `hack/ci-deploy.sh` needs the presubmit project. That gap is much smaller than it was: after the rebase this branch changes no step, substitution or `waitFor` edge in that file, only comments. What CI will exercise for the first time is the Dockerfile's new stage graph on a presubmit worker.

**Cleanup.** The `session_manager.py` measurement marker is reverted; the branch diff does not touch that file. Test images are ordinary dev tags in the shared AR repo, left in place as the loop normally does. Nothing else left behind.

## Self-Review

Ran `review-adversarial` in a subagent handed the diff range and nothing else — not the plan, not my reasoning — per AGENTS.md, and separately `review-docs-drift`.

Angles covered: runtime completeness of the new image (every path in `start-services.sh` and every import in `credential_proxy.py` traced to its producing stage); file modes and ownership drift across the stage boundary; anything else in the repo still assuming `FROM platform`; the Cloud Build DAG and its substitutions; whether `docker-build.yml`'s layer-budget step still has an image to inspect; correctness of the new Dockerfile stage parser; and prose claims the diff does not support.

Six findings, all addressed:

1. **The `/opt/defaults/scripts` identity became a convention.** `FROM platform` made divergence structurally impossible; now two COPY lists have to agree, and my own `target="platform"` filter is what stopped the existing test from noticing the sidecar's list. Divergence would be silent in every gate — both images build, the budget passes, `entrypoint-gate-test` covers the platform image only — and the first signal would be a CrashLooping sidecar taking every proxied CLI call in the agent container with it. **Fixed:** added `test_the_two_images_fill_opt_defaults_scripts_from_the_same_sources`, which parses both stages. Verified it actually bites: dropping `agents/chat/scripts/` from one list fails it with the right message; restored, green.
2. **"`agent-base → platform` is the longest chain in the file" was wrong as of my own diff** — `entrypoint-gate-test` is two instructions deeper, at 116. No operational risk, since buildx alone builds it and buildx has no depth limit, but I introduced the inaccuracy and it had propagated to four places. **Fixed:** all four now say "deepest chain that **ships**", and say why the gate deliberately does not follow the test stage.
3. **`hack/ci-deploy.sh` and `cloudbuild-ci.yaml` both said the sidecar step builds on the platform image the previous step produced.** It now builds on `agent-base`, the ancestor the two images still share. **Fixed** — the arrangement is unchanged, the reason for it is corrected.
4. **The Dockerfile's own justification comment undercounted** what the sidecar takes from the chain — it said "exactly two things" and omitted the `k8s-event-watcher` binary. Not a bug, since `agent-base` supplies it, but this sentence is the load-bearing argument for the whole restructure. **Fixed.**
5. **`operator/development.md` described the CI build's step graph** as it was before the parallelism change I had proposed. **Resolved by dropping that change** rather than by editing the doc — the file is back to upstream's text, unmodified by this branch.
6. **Found in the docs-drift pass, not the adversarial one:** my own new prose claimed the sidecar "carries none of the agent workspace — no personas…". Contradicted by `/opt/defaults/SOUL.md`, which the base image ships. **Fixed** to "none of what the `platform` stage adds on top", which is what I can actually demonstrate.

The subagent recorded one finding it could not confirm: with the two agent builds running concurrently on one worker, both need `agent-base`, and whether BuildKit dedups that vertex across concurrent solves rather than building it twice is not checkable without a runtime. It should — moby's embedded BuildKit dedups identical vertices — but I have not measured it. Rather than ship a CI change resting on an unverified claim, **I removed the parallelism** and left `cloudbuild-ci.yaml`'s step graph as #739 has it. It remains a reasonable follow-up for someone who can measure it.

Two things the review looked at that this rebase changed underneath it: the `--cache-from` seeding, since #739 replaced the substitutions the finding was about, and `operator/development.md`. Both are addressed by the smaller diff rather than by a fix, and the diff the review read is otherwise intact — the Dockerfile, `check_image_layers.py` and the tests are byte-identical to what it saw.

## Risk & Rollout

Build-time only; no runtime code path, no operator or chart change. The blast radius is the two images' contents, and the sidecar's contents were diffed against the old image and exercised on a live Pod.

Two things to know at merge time. The first dev-loop build after this lands gets a `credential-proxy:latest` describing the old chain, so it misses the new `proxy-tools` cache and pays the apt/`gh` install once; it self-heals on the second run. That is the 9m14s transition row in the table above. Presubmit CI has no equivalent penalty — its sidecar step passes no `--cache-from` and never did, so it rebuilt that install on every run before this change too. And if a future change ever puts `credential-proxy` back on top of `platform`, `DEFAULT_IMAGE` in `scripts/check_image_layers.py` has to move back with it or the budget gate silently starts measuring the wrong chain — that docstring says so.

Reverting is a clean `git revert` of the single commit: no state, no migration, nothing persisted.


jayantid:
LGTM. Please get a review from @haoxuw

haoxuw:
LGTM
Nitpic: I think the comments in the dockerfile is a bit too long. Not sure if it'll deliver value to either agent or human reader

haoxuw (approved):

PR #810, opened 2026-08-19
docs(agents): make the pre-PR review passes runnable in a fresh context

bradhoekstra:
## Summary

`AGENTS.md` requires an adversarial review pass before a pull request opens, and requires it to run in a context that did not write the change. Coding agents carry a standing instruction not to spawn subagents unless the user asked for one, so an agent following the requirement finds its only route closed and quietly runs the pass inline — the one configuration the rule exists to exclude — then reports it as a self-review. This adds `/pr-preflight`, a user-invoked command that makes the delegation requested by construction, and `review-preflight`, the harness-neutral skill holding the procedure: one diff range, the mechanical gates first, one subagent per applicable pass, and one merged disposition list back.

## Why This Change

The repository already solved this on the reviewer side and had not applied it to the author side. `.claude/commands/pr-review-batch.md` gives every pull request under review its own subagent, and it works because a user invoking a command _is_ the user requesting delegation. The author's own pre-PR gate had no equivalent, so the requirement was satisfied on paper and defeated in practice, invisibly: the **Self-Review** section reads the same whether a fresh context produced it or the conversation that wrote the diff did.

Without this, the failure is silent in both directions. The agent does not report being blocked, because nothing looks blocked; and the reviewer cannot tell from the section which kind of context produced it.

## What Changed

- **`.agents/skills/review-preflight/SKILL.md`** (new) — the procedure. Fix and freshly fetch the base, run `make docs-check` in the main loop, fan out one subagent per pass, merge what returns. Holds no review method: `review-adversarial` and `review-docs-drift` own that, and `AGENTS.md` owns the requirement.
- **`.claude/commands/pr-preflight.md`** (new) — a thin wrapper whose job is to be typeable and to constitute the request to delegate. All method stays in `.agents/`.
- **`AGENTS.md`** — names `/pr-preflight` as the way to get a fresh context, says the **Self-Review** section carries every pre-PR pass merged rather than the adversarial one alone, and requires naming which kind of context each pass ran in.
- **`.github/PULL_REQUEST_TEMPLATE.md`**, **`docs/site/src/content/docs/contributing.md`** — reconciled to the above, in that order, per the "change this list first" rule.
- **`.agents/skills/review-adversarial/SKILL.md`** — §1 keeps the argument for a fresh context and hands the plumbing to `review-preflight`, retaining a self-contained minimum for readers who arrive without a wrapper.
- **`.claude/commands/pr-review-batch.md`** — its description of what `## Self-Review` contains widened to the merged list, so the reviewer side and the author side agree.

## Context

Follows up #788 (PR review flow improvements). Does not close it — #788 is broader than this change.

#797 (`feat(install)!: make Terraform + Helm the only install engine`) landed first, so this branch took the reconciliation it predicted. #797 removed `review-iac-parity`, `scripts/check_iac_parity.py`, the `iac-parity-check` target and its CI step, and the `AGENTS.md` parity bullet — correct on its side, since parity is a relation and one install engine has nothing to hold it against. Rebased onto it: the preflight set is now two unconditional passes rather than two plus a conditional third, `review-preflight` §3 runs `make docs-check` alone, and the merge and Output sections no longer describe a triage only one pass emits. The link to `../review-iac-parity/SKILL.md` had gone dangling and would have failed `docs-check`.

The two textual conflicts were the ones called in advance: the `AGENTS.md:13` layout line, where main's list had already dropped IaC parity and only `review-preflight` needed adding, and the parity bullet, which main replaced wholesale with its one-engine bullet — mine only amended a bullet that no longer exists, so main's side won outright.

#720 also edits the `AGENTS.md` Repository Layout list near line 13 — a merge-conflict warning, not a competing change.

## Testing

- `make docs-check`, `make validate`, `make prompt-check` (98 instruction files, 43 skill bundles), `make chart-check` — all clean, re-run after the rebase onto #797.
- `prettier@3.9.6 --check` clean on all six changed files.
- `docs/site` `npm ci && npm run build` clean, 43 pages.
- The base-resolution block in `review-preflight` §1 was executed, not just read: in a scratch repo holding both a fork remote and an upstream remote it selects the upstream one, falls back to the literal URL when no remote matches, and its guard fires and stops on a fetch of a nonexistent branch. Re-run against the HTTPS fallback in a clone whose only remote points elsewhere, with `GIT_SSH_COMMAND=/bin/false` and `GIT_TERMINAL_PROMPT=0`: the URL it used to carry exits 128 ("correct access rights"), and the one it carries now fetches `refs/kube-agents-base/main` at `95a7c04` and exits 0.
- The empty-`$BASE` failure behind the handoff fix reproduces: `git diff --stat ...HEAD` in a two-commit scratch repo prints nothing and exits 0, so a pass handed that range has no diff to review and nothing to report.
- The relative links added in `.agents/` and `.claude/` are genuinely covered by `check_docs_links.py` rather than skipped as code spans: breaking one on purpose produced `SKILL.md:17: broken link`, and restoring it returned the gate to green.

### Live validation

Not live-tested. The change is documentation and repository-level agent tooling — `.agents/skills/`, `.claude/commands/`, `AGENTS.md`, the PR template, and the site's contributing page. None of it ships in the agent images, reaches the operator, or is read by a running installation, so there is no CR `.status`, Deployment env, or in-pod file for it to affect.

What stands in for a live test is that the change was exercised through its own mechanism: `/pr-preflight` was invoked on this branch and drove all six review subagents reported below, including the concurrent fan-out, the handoff, and the merge. What that does not cover is regression protection — nothing in the repository tests slash-command or `SKILL.md` semantics, so no automated check would fail if this procedure were broken later. `make docs-check-links` covers the pointers between these files and nothing covers the prose.

No test artifacts and no cluster state; nothing to clean up.

## Self-Review

Three rounds before opening, each run through `/pr-preflight` itself — the change reviewed by the mechanism it adds — plus four `kube-agents-bot` passes and a human review round after, the last bot pass clean. Every pass ran **in its own subagent**, none in the context that wrote the change. Each round: `review-adversarial` and `review-docs-drift` concurrently, handed the repository, the diff range, the path to their skill, and the gate output; withheld the plan, the reasoning, and the intent sentence. `review-iac-parity` was **skipped in all three rounds** — the diff touched no install surface, so its trigger never fired; #797 has since deleted that pass, and the rebase removed the branch's references to it.

Angles run: the full `review-adversarial` set. Angle J (sibling work) found both overlaps, neither known to the reviewing context: #808 in round one, and #797 in round three. Only #797 is in Context — #808 merged while this work was in progress, and the branch was rebased onto it with the colliding paragraphs folded together rather than stacked, so it is absorbed rather than outstanding. Angle D is thin by construction: the diff touches no IAM, RBAC, NetworkPolicy, workflow, or manifest, and the one executable block in it was run.

Sixteen findings across three rounds, all confirmed, all fixed. The substantive ones:

- **Four surfaces gave three different answers** to what an agent should do when it cannot get a fresh context. Now one answer, matching #808's: an approval you could not get blocks the step rather than waiving it.
- **The skill narrowed `make iac-parity-check`** from unconditional to conditional, which is how a one-surface change starts failing CI. Fixed at the time by running both gates unconditionally; #797 has since deleted that gate, so the branch as it stands runs `make docs-check` alone and the finding no longer has a surface in the diff.
- **The withhold list was unenforceable.** Commit messages are inside the diff range by construction and plan files sit in the checkout the pass is handed, so both carry the intent the pass is supposed to derive for itself. Withholding cannot be done by the handing side alone; the handoff now instructs the pass not to read them.
- **The base-fetch recipe lived only in the Claude Code wrapper** while the skill declared itself harness-independent, so a harness following the skill directly would review against a stale base — silently widening the range with commits that already merged. Moved into the skill.
- **The merged output shape fit only one pass in the set.** `review-docs-drift` emits a Blocking/Advisory triage and never produces a verdict, so demanding the adversarial shape of it was unsatisfiable. Each pass now keeps its own severity, and severity and confidence are separated.
- **Two Blocking findings, both self-inflicted:** round two widened the summaries (`pr-review-batch.md`, `review-preflight`) to "the merged list of every pass" while leaving the canonical `AGENTS.md` statement describing the adversarial pass alone — the exact inversion of the "change this list first" rule that bullet states about itself. The template then contradicted itself inside one comment block, singular on one line and plural on another. Fixed canonical-first, then the summaries.

Four more came after opening — two from `kube-agents-bot`, two from review — and they are what the three self-review rounds missed. All four land in `review-preflight` §1 and §4, and all four are the plumbing rather than the prose: the part of the change no pass could exercise, per the coverage note below. Reading the procedure against its own requirements caught the arguments and none of the wiring.

- **The handoff never said the report was the deliverable.** `review-adversarial` §6 tells whoever runs it to edit on CONFIRMED, and its Angle I asks for a test run against the pre-change behaviour; both write to the tree, and the passes run concurrently in the author's checkout rather than one worktree each. The other two pass skills already refuse to fix silently, and `pr-review-batch` already carries the sentence for the reviewer side; `review-preflight` §5 now carries the author-side equivalent and the `claude -p` handoff example carries it too (9829a1c). It did not fire in the three rounds run here — all six subagents reported without editing, one saying so unprompted — but that was the model's judgment rather than the procedure's, and this diff has no runnable test for Angle I to reach for.
- **§1's escape hatch reinstated the stale base §1 exists to prevent.** It exempted any base `git rev-parse --verify` resolves — a test of existence, not freshness. A local `main` forty commits behind resolves, and so does a remote-tracking ref last fetched a week ago, which `AGENTS.md:42` exists because it is the ordinary state of a checkout here; the default input satisfied the exemption, so the ordinary path took it. The recipe had been moved into the skill in round three precisely so a harness with no wrapper would not review against a stale base, and the exemption three lines below undid that for exactly that reader. The fetch is now skippable only for a base that cannot go stale — a SHA or a tag — and both files now say `BASE_BRANCH` is the branch name as the remote has it, since `origin/main` went into the refspec verbatim and tripped the guard over a base that was fine (ea84786).
- **The `claude -p` handoff interpolated `$BASE`, which is empty by the time it runs.** §1 assigns it in a different shell invocation from the one that composes the command, and Claude Code carries no environment between tool calls. An empty expansion is not an error: `git diff ...HEAD` defaults the omitted left side to `HEAD`, exits 0, and prints nothing, so the pass reviews an empty range and reports "no findings" — the silent wrong-range failure §1 exists to stop, arriving through the handoff. The example now spells the ref out and says why, and draws the distinction with `$PWD`, which every shell sets afresh (45f435b).
- **The §1 fallback remote was the SSH URL, and the fallback fires where keys are absent.** It is reached when no remote points at `gke-labs/kube-agents` — a plain clone of a fork, which is where every open pull request here comes from. In a sandbox or CI runner the fetch failed and the guard three lines below reported an auth failure as a stale base, blocking the preflight at its first step. The repository is public, so HTTPS needs no credential (45f435b).

Deliberately not fixed, one item: a stale row at `docs/README.md:294`, pre-existing and unrelated to this diff. Editing an unrelated row in the documentation map is precisely the scope creep the review's own angle I flags, and the map is edited from several branches a week.

What the passes could not cover, in their words: the workflow was reviewed **as written, not as executed** — no pass ran `/pr-preflight` end to end, since doing so would spawn the subagents it was itself part of, so several findings rest on reading the procedure against its own stated requirements rather than on watching it fail. No red-then-green test exists or is available for prose. The two fixes from review (45f435b) were verified by executing the block rather than by a fresh pass — no fourth `/pr-preflight` round ran over that delta. Two passes could not verify the quoted harness instruction from inside their sandbox; a third verified it byte-exact against the installed binary, and the version pin was dropped anyway because nothing in the repository can notice when upstream rewords it.

## Risk & Rollout

Low. Documentation and agent tooling only — no runtime code path, no operator behaviour, no image contents, nothing an installation reads. The strongest effect is on agents that read `AGENTS.md` and now have a route to a fresh review context instead of a closed one; the failure mode if the new procedure is wrong is a review that does not happen, which is the status quo. Reverting is a single commit with no migration.

The merge-order coupling with #797 that this section previously flagged is settled: #797 landed first and the branch is rebased onto it, so there is nothing left to watch at merge time.


.agents/skills/review-preflight/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+are blocked on. `AGENTS.md` is explicit that an approval you could not get blocks this step rather
+than waiving it, so the authoring context is not the fallback — running the pass there and
+disclosing it is what you do when a human, told the above, tells you to proceed anyway. Then
+**Self-Review** says which context ran the pass, in those words.
+
+## 5. What to hand each pass, and what to withhold
+
+Hand it: the repository, the diff range, the path to its skill, and the gate output from step 3.
+
+Withhold everything you know that it does not: your plan, your reasoning, the commit messages you
```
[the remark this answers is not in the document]
bradhoekstra: Fixed in 9829a1c. Confirmed every source you cited — `review-docs-drift:85`, `review-iac-parity:116`, `review-adversarial` §6 and Angle I, and `pr-review-batch.md:520` — all say what you say they say, and the asymmetry is real: two of the three skills refuse to fix silently and the third instructs the opposite.

`review-preflight` §5 now states that the report is the deliverable and carries the author-side equivalent of the `pr-review-batch` sentence:

> The skill's step 6 does not apply to you: report every finding, edit nothing, and leave the tree as you found it. Dispositions are the caller's.

The `claude -p` handoff example in §4 carries it too, since that is the other place a concrete handoff gets constructed. §5 closes by pointing at step 6 as the single place edits happen, after the relay — which also settles the double-assertion of fix authority you flagged between §4, §6, and the command file.

Worth recording that your failure scenario did not fire in the three review rounds run on this branch: all six subagents reported without editing, and one said so unprompted. That was the model's judgment rather than the procedure's, and this diff is Markdown with no runnable test, so Angle I never had a reason to revert anything. On a branch with real tests it would have.

Took the sentence rather than the worktree isolation `pr-review-batch` uses: a worktree per pass is a real complexity increase in a skill meant to be followable by any harness, and it does nothing for a harness that has no worktrees.

.agents/skills/review-adversarial/SKILL.md
```diff
@@ -7,9 +7,9 @@ description: Reviews a change adversarially — establishes what it claims to do
```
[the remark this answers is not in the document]
bradhoekstra: Corrected in the description. You are right that the sentence promised two overlaps and Context delivered one.

Angle J did find #808 in round one — it collided with this branch at the same anchors in all three shared files. It then merged while the work was in progress, and the branch was rebased onto it with the colliding paragraphs folded together rather than stacked. So it is absorbed rather than outstanding, which is why it is absent from Context and why Risk & Rollout names #797 as the only merge-time coupling. The Self-Review line now says that instead of implying #808 is still in play.

.agents/skills/review-preflight/SKILL.md
```diff
@@ -0,0 +1,198 @@
...
+}
+BASE=refs/kube-agents-base/$BASE_BRANCH
+
+git diff --stat "$BASE"...HEAD     # three-dot: against the merge base, not the base tip
+```
+
+Three-dot keeps base-branch drift out of everyone's scope. If you were handed a ref that already
+resolves (`git rev-parse --verify`), use it and skip the fetch. If the fetch fails, stop and say so:
+there is no safe fallback, because the wrong base fails in the direction that hides findings.
+
```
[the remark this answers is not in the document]
bradhoekstra: Confirmed, and fixed in ea84786.

You are right that `git rev-parse --verify` answers the wrong question. The exemption was written for the caller who has just fetched and hands the resulting ref back — which is the one case where re-fetching costs nothing — but the condition I gave for it is satisfied by any name that exists, including the stale local `main` that `AGENTS.md:42` says is the normal state of a checkout here. And as you note, the default input satisfies it, so the ordinary path took the exemption rather than the exceptional one. That the recipe had been moved into the skill in round three specifically so a wrapper-less harness would not review against a stale base makes it worse: the exemption three lines down undid the fix for exactly that reader.

§1 now skips the fetch only for a base that cannot go stale — a commit SHA or a tag — and says in those words that a name resolving is not that test.

The `origin/main` point is fixed in the same commit. §1 now states that `BASE_BRANCH` is the branch name as the remote has it, and names the three forms that go into the refspec verbatim and trip the guard over a base that was fine: `origin/main`, `refs/heads/main`, and the `refs/kube-agents-base/main` a previous run wrote. `.claude/commands/pr-preflight.md` takes `argument-hint: [base-branch]` and tells the agent to reduce what it was handed and to fetch it even when it already resolves.

Both are folded into Self-Review. Worth saying plainly: this is the second finding of yours that three `/pr-preflight` rounds missed, and both landed on the handoff rather than the prose — the part of the change the passes could only read, never run.

.agents/skills/review-preflight/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+- **The user invoked `/pr-preflight`, or asked for the review in words.** That is the request. Spawn
+  the passes and do not ask again.
+- **You got here on your own.** Ask once, before the pass, and wait: say you need a context that did
+  not write the change, how many subagents that is, and that the alternative is a review by the
+  context that already believes the change is correct. Ask when you hit it, not after.
+- **The harness has no subagents.** Start a fresh session, or a headless run, handed the same
+  material and nothing else — one per applicable pass, not just the adversarial one:
+
+  ```bash
+  claude -p "In $PWD, follow .agents/skills/review-adversarial/SKILL.md against $BASE...HEAD.
```
jayantid: Line 116 interpolates `$BASE` into the `claude -p` prompt. `$BASE` is set by the section 1 block, which runs in a different shell invocation for most agents — Claude Code, for one, does not carry environment variables between commands. By the time an agent composes this command, the expansion is empty.

An empty `$BASE` fails silently. The prompt then reads "against ...HEAD", and `git diff ...HEAD` is valid git: it exits 0 and prints an empty diff (verified at ea84786). The pass reviews nothing, reports "no findings", and that lands in **Self-Review** — the silent wrong-range failure section 1 exists to stop. It would also be the third defect on this PR to land in the handoff plumbing, after the report-only sentence (9829a1c) and the escape hatch (ea84786).

Make the example self-contained: show the resolved ref instead of the variable, and say why.

```bash
claude -p "In $PWD, follow .agents/skills/review-adversarial/SKILL.md against
refs/kube-agents-base/main...HEAD. ..."
```

One sentence covers the why: `$BASE` does not survive between tool calls, and an empty `$BASE` makes `...HEAD` a valid, empty range.

.agents/skills/review-preflight/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+as though it were yours:
+
+```bash
+# The remote pointing at gke-labs/kube-agents is not reliably called `upstream`, and
+# on a clone of the upstream repository rather than a fork it is `origin`. Find it.
+BASE_BRANCH=main                                  # or the branch you will target
+BASE_REMOTE=$(git remote | while read -r r; do
+  case "$(git remote get-url "$r")" in *gke-labs/kube-agents*) echo "$r"; break;; esac
+done)
+: "${BASE_REMOTE:=git@github.com:gke-labs/kube-agents.git}"
```
jayantid: The fallback remote on line 35 is the SSH URL, `git@github.com:gke-labs/kube-agents.git`. SSH always needs a provisioned key. The repository is public, so the HTTPS URL fetches with no credentials at all.

The fallback fires exactly where keys are absent. It is reached when no remote points at `gke-labs/kube-agents` — a plain clone of a fork, and every open PR on this repository comes from a fork. In an agent sandbox or CI runner without SSH keys the fetch fails, the guard reports "base fetch failed — stop here rather than reviewing against a stale base" (an auth failure read back as staleness), and the preflight this PR adds blocks at its first step.

```bash
: "${BASE_REMOTE:=https://github.com/gke-labs/kube-agents.git}"
```

.agents/skills/review-preflight/SKILL.md
bradhoekstra: Fixed in 45f435b. The failure mode reproduces as you describe: `git diff --stat ...HEAD` in a scratch repo prints nothing and exits 0, so an empty `$BASE` hands the pass an empty range and a clean "no findings".

The example now spells the ref out — `refs/kube-agents-base/main...HEAD` — and a paragraph under it says to substitute the ref §1 resolved, why `$BASE` does not survive (most harnesses run each command in its own shell), and that the empty expansion is not an error but `HEAD...HEAD`.

I kept `$PWD`, since every shell sets that one afresh and it makes the example paste-able. The new paragraph draws that distinction explicitly rather than leaving the next reader to work out why one variable is fine and the other is not.

.agents/skills/review-preflight/SKILL.md
bradhoekstra: Fixed in 45f435b, with your line.

Exercised both directions in a scratch clone whose only remote points somewhere else, with `GIT_SSH_COMMAND=/bin/false` and `GIT_TERMINAL_PROMPT=0` so neither a key nor a prompt is available. The SSH fallback exits 128 ("Please make sure you have the correct access rights"), which the guard three lines down would have reported as a stale base. The HTTPS one fetches `refs/kube-agents-base/main` at `95a7c04` and exits 0.

A comment above the assignment now records why — the fallback fires where no remote matches, which is a plain clone of a fork, which is where the key is absent — so it does not get changed back to SSH later.

Out of scope but worth noting: `.claude/commands/pr-review-batch.md:244` falls back to `$REPO_SSH` the same way. Pre-existing, on the reviewer side, and untouched here.

jayantid (approved):

PR #811, opened 2026-08-19
docs(designs): a findings queue for the bootstrap inventory scan

### LINKED ISSUE #774: Create a queue for findings from bootstrap inventory scan
After bootstrap inventory scan is ran, return results to the users in a non-overwhelming way. Right now top 5 is returned, and rest is kept on disk but not re-surfaced. The additional findings won't get resurfaced, or may need to be re-verified against the latest cluster state.

toshiowang:
## Summary

Adds a design of record for turning the bootstrap inventory sweep's discarded findings into a durable, ranked queue. The first-time sweep finds everything and reports five things; on a neglected fleet the remainder is ranked, rolled into one `Also found: N items` line, and thrown away with the process that produced it. This document specifies a `findings` table in the Session KV database beside `incidents`, an SRE risk rubric that orders it, publication of the whole ranked backlog as one document rewritten in place, a daily chat nudge pointing at it, and reuse of the existing `remediate` machinery so the highest-ranked findings a pull request can be written for arrive with the fix already prepared. Nothing is implemented here — the file carries the standard not-yet-implemented banner.

## Why This Change

Issue #774: after the bootstrap scan runs, the top five findings reach the user and the rest sit in `/opt/data/INVENTORY.raw.md`, never resurfaced and never re-verified against a cluster that has since changed. Three things go wrong at that last step. The ranking is computed and discarded, so the sixth-ranked finding and the fortieth become indistinguishable. A display cap is doing the work of a lifecycle — five is the right number for a chat message, but it is not a statement that the sixth problem was resolved, deferred, or accepted, and there is nowhere to record which. And a snapshot has no mechanism for noticing it has gone stale.

Without a design, the obvious implementations are each wrong in a way that is expensive to discover later: widening `incidents` (whose primary key is a chat thread, so it cannot be ordered by priority, and whose rows are deleted after fourteen days), or putting the backlog in the fleet-audit GitHub ledger (which makes the queue a thing the user has to go and read).

## What Changed

- `docs/designs/inventory-findings-queue.md` — new. Twelve sections: why the delivered report and the `incidents` table are both the wrong container; the `findings` schema and its lifecycle; the priority rubric with its anchors, worked examples, and two gates; who writes it and who reads it; how the queue is published; how a queued finding reaches a human; how the fix arrives with it; the overlap with the seven audit streams; and what building it consists of.
- `docs/README.md` — one row in the `docs/designs/` block, compact form, table not re-aligned.

Four decisions in the document are load-bearing enough to name here:

**The rubric is anchored ordinals, not a judgement call.** `rank_score = round(B × L × (detect + recover) × C)`, decomposed from the SRE Workbook's risk model (impact × duration ÷ time between occurrences). This is not a preference for arithmetic — `inventory_prioritize_sop.md` measured what model-side scoring costs, and over three runs on identical input it produced 3, 6 and 6 items with two findings common to all three. A report absorbs that; a queue whose contents shift from an unchanged fleet does not.

**Ranking and publishing are separate, and the whole ranked list is published.** §7 writes the backlog as one document rewritten in place and posts a nudge pointing at it, rather than metering findings into chat a few at a time. §7.3 is the measurement behind that: an earlier draft that selected two findings a morning surfaced 6 of 30 in ninety days, and four rounds of selection rules were machinery for rationing a channel that did not have to be scarce.

**Severity is derived from the score, and it uses `critical`/`major`/`minor`.** Not the `Critical`/`Warning`/`Info` alert-quota vocabulary in `session_kv_server.py`; the document says why the two are different things.

**Check slugs are the audit streams' own.** This resolves duplicate reporting and PR routing in one move: when the sweep and an audit stream find the same problem, `derive_finding_id` yields the same string, so they are one row rather than two findings to reconcile. §10.1 records the three sweep checks that no stream owns yet and what it costs to give them homes.

## Context

Closes #774.

Two open PRs are worth a reviewer's attention, neither a duplicate:

- **#594 and #595** (both draft, @fkc1e100) each add a new `AuditSpec` to `AUDITS`. If either merges, this document's "seven audit streams" goes stale and needs a follow-up edit. More substantively, they are precedent for the eighth-stream approach §10.1 rejects — a reviewer who thinks that precedent should win over the reasoning there should say so, because it changes §10.1's recommendation rather than just its wording.
- **Fifteen open PRs touch `docs/README.md`.** This branch adds one line and re-aligns nothing, so the risk is a merge conflict on the map rather than contested work.

Follow-up this branch deliberately does not do: four documents disagree about how many audit streams exist — `fleet-audit-issue-ledger.md` says six, `gitops-workspace-leases.md` says six, `pr-comment-conversation.md` says seven, and `AUDITS` has seven. Pre-existing and unrelated to this change; worth its own issue rather than a drive-by fix here.

## Testing

- `make docs-check` — passes at the current head. Generated regions current, relative links resolve, terminology check passes, map inventory covers all tracked docs with no re-aligned rows.
- `prettier --check` on both changed files at the current head, using the version pinned in `.github/workflows/prettier.yml`.
- No code changed, so no build or unit-test surface exists for this branch.

### Live validation

Not live-tested. The branch adds two Markdown files and changes no runtime code, manifest, chart, or configuration — there is no code path that reaches a running installation, and nothing for an operator to reconcile or an agent pod to pick up.

The design's claims about running behaviour were instead verified by reading the source they describe, since that is the only check available to a docs-only change: the `incidents` schema and its `CLEANUP_TTL_DAYS` purge, `_claim_alert_quota` being reached only from `inject_message` (so the cron relay path does not spend the alert budget), `_send_to_chat`'s bare-platform fallback via `get_active_platform`, `derive_finding_id`/`_id_segment`, `SEVERITIES` and `REMEDIATION_KINDS`, `remediate` passing `auto_promote=False`, `_finding_sort_key`'s tuple and the `STALE_CLOSED_LABEL` semantics, the watcher's `EventKey` dedup contract, and every check slug in the §10 overlap table against `AUDITS` and the SOP rosters.

Two delivery hazards the document flags rather than asserts, because they are properties of a deployed install and not of this design: a named-profile cron job needs `platforms` in its profile config to reach chat at all, and the Google Chat home-channel configuration can silently fail gateway delivery if written in the wrong shape. Both must be checked against a running install when §7.5's job is built. No test artifacts were created and no state was changed anywhere.

## Self-Review

Three rounds, each run in a subagent handed only a diff range — none held the reasoning that produced the text under review. Rounds 1 and 2 read the document as first written; round 3 read the two commits that answered them.

**`review-adversarial`.** Angles run: source-citation accuracy against the files cited, internal consistency between sections, quotation fidelity, arithmetic, and whether the argument's conclusions follow from its premises. Twenty findings, all fixed. The ones that mattered:

- _§10's residue claim was false, and it was load-bearing._ The document asserted that every sweep check no audit stream owns is unactionable or non-manifest, therefore never promotable, therefore never blocked by `validate_audit_id`'s allowlist. Three checks break that: `startupProbe`, `readOnlyRootFilesystem`, and ResourceQuota/LimitRange presence are all manifest-kind and promotable with no owning stream — precisely the case §10 said could not arise, and the case that motivates the eighth stream it rejects seven lines later. Rewritten as §10.1, which names the three, names the near-miss checks so nobody re-derives them, and states the fix as an implementation task.
- _Auto-promotion belongs to `finish`, not `remediate`._ §4.2 said `critical` was one of three conditions on `remediate`'s auto-promotion. `remediate` passes `auto_promote=False` and "opens what was named and nothing else" — and `remediate` is the call §9 puts on this path, so on the queue's own path the threshold gates no pull request at all. §9 already had this right, so the document contradicted both the source and itself.
- _The candidate set contradicted its own opening._ Selection filtered on `state == queued`, which drops a finding the moment it is first surfaced — exactly the behaviour the section opened by calling wrong.
- Smaller confirmed fixes: score range is 1–480 not 2–480 (C = 0.6 reaches 1); `audit:stale-closed` marks the harness's close, not a human's, and the sentence had it backwards; `remediate` takes three required arguments, not two; the five-item report cap has an exception for criticals; a `derive_finding_id` docstring quotation welded two separate incidents into one; and §4.1 attributed a measurement to Step 3 that is in Step 1 and is narrower than the use made of it.

**`review-docs-drift`.** Angles run: canonical-home ownership, link-rather-than-summarise, hard-coded counts, generated regions, map-row form, and PR-status prose. Six blocking findings, all fixed — four overlapped the adversarial pass. The two it found alone: the document called the new job "the eighth" when `jobs.json` already holds eight, and the §10 overlap table mapped a `public control plane` sweep check that does not exist in `inventory.md` (the slug is real, the sweep's check is not). Advisories also applied: three restated topics now name their canonical page (`bootstrap_onboarding/README.md`, the watcher README for the dedup window, `cron-report-relay.md` for relay routing), "ledger" is no longer overloaded between the `findings` table and the fleet-audit issue, and `remediation` carries `note` so a row satisfies the audit validator unmodified.

**Where the two passes disagreed, I checked the source.** They split on whether `podsecurity-gaps` owns `readOnlyRootFilesystem` — it does not; it covers `runAsNonRoot`, `runAsUser`, and `seccompProfile`, and its remediation template sets `allowPrivilegeEscalation` and `capabilities` without touching read-only root filesystems. And on `public-control-plane`, where the slug exists but the sweep check does not, so the objection held for a different reason than the one given.

**A dry run against a simulated fleet found the defect neither reader did**, in the section a reader would be least likely to doubt: two clusters, 32 findings scored by the rubric using real check slugs, 30 rankable after the ownership gate.

The ranking held up — the order is defensible end to end, both gates sort correctly, and findings tied at 90 break deterministically on `_finding_sort_key`. The delivery rule did not. As originally written it metered two findings a morning into chat and published nothing else, and over ninety days it surfaced **6 of 30 findings**, one of them ninety times. The starved findings were not marginal: a `critical` CrashLoopBackOff on a production payments workload and a `critical` single-replica session store never surfaced at all, blocked by two findings the operator cannot fix with a manifest edit — a provider-managed `kube-system` CrashLoop kept rankable by the active-fault exception, and a certificate whose deadline had passed. The design reproduced #774 inside its own fix.

That result rewrote §7 rather than being patched around it. Four rounds of selection rules take the number from 6/30 to 30/30 and the fastest full pass to day 12; a published list is 30/30 on day one and needs none of them. §7.3 keeps the table as the argument for publishing whole, and says plainly that a simulation is a design aid rather than a test — it assumes the rubric's own scores are right and models a fleet rather than measuring one. What survives from the exercise is the ranking it validated and the grouping-by-object it found, which is why §7.1 groups rather than lists flat.

**Round 3, over the two commits that answered the above.** Two passes, same isolation, fourteen findings with six overlapping. Four were defects introduced by the first of those commits, and they are the ones worth a reviewer's time:

- _The promotion slice filtered on the wrong property._ Skipping provider-managed and non-manifest rows left the two cases that actually jam it: a row whose pull request is open, which `remediate` deliberately leaves alone, and one whose pull request a human rejected, which it will never re-propose. Three findings under review is the steady state of a working install, so the slice would have sat full while the fourth waited — the starvation §9 exists to fix, rebuilt. The slice is now the top three rows that are actionable, manifest-kind, and have `pr_url IS NULL`.
- _The `/ranked` sort key split the groups it was meant to keep together._ It carried no group identity ahead of `rank_score`, so two groups tied on top score interleaved — and §4.3 puts three findings at 90. It also silently redefined "top three": the first three rows became the top group's rows, while §7.2, §7.1 and §3.1's index rationale all still read that as the three highest scores. Fixed by dropping grouping from the endpoint entirely; it belongs to §7.1's renderer, which is where it was always a rendering concern.
- _`alarmed_at`'s gloss contradicted its own trigger._ Described as null when the row is not over §4.2's floor, which makes null mean ineligible and the alarm never fires. The column now points at §7.2 rather than paraphrasing it.
- _§4.6's re-rank event list forbade the alarm's clearing path._ The list is closed and all three events raised L, so the clearing transition §7.2 names was one the document ruled out. It gains a fourth event — the only one that lowers a score.

The same round fixed six pre-existing issues: snooze expiry gains an actor, so a row is never published while stored as `snoozed`; "top slice" no longer names two different sets across §4.2, §5.2, §7.5 and §12; §11 no longer claims the first-time report is unchanged when §5 rescores what it renders; §4.3's reference to the deleted drain slot is gone; the `findings_urgent` index covers the alarm's predicate rather than the deleted urgent slot's; and the map row drops a distinction §9 argues against.

**Two things I looked for and did not resolve.** The severity bands are wide in the middle — `major` held half the simulated queue — which §4.2 flags as the first thing to calibrate against a real sweep rather than guessing at better thresholds now; nothing selects on the band, so it is a labelling problem, not a routing one. And the simulation harness is not checked in, so §7.3's numbers are reported rather than reproducible from the branch; I left it out to keep this a docs-only change, and would add it if a reviewer would rather have it.

## Risk & Rollout

Low. Two Markdown files, 1,171 insertions, no deletions, nothing outside `docs/`. No runtime code, manifest, chart, or configuration is touched, so there is no failure mode at merge time and reverting is `git revert`. The one merge-time consideration is the `docs/README.md` row, which fifteen other open PRs also touch; the row is a single inserted line in compact form with no re-alignment, so a conflict there resolves by keeping both lines.


toshiowang:
Superseded by #861, which carries rebased copies of these commits on current `main` plus the implementation they describe. Cross-fork PRs cannot stack, so the design and the code ship together there.

PR #813, opened 2026-08-19
fix(submit-suggestion): normalize CLI escaped newlines in PR title and body

Fuxiao-Gao:
## 🎯 Issue Description
When autonomous agent workers or LLMs invoke `submit_suggestion.py submit` via the shell, multi-line arguments passed to `--title` and `--body` often contain literal escape sequences (e.g., `\n`, `\r\n`, `\t`). Without normalization, GitHub renders these as raw text strings (e.g., literal `\n`) instead of actual line breaks and indentation, degrading PR readability and breaking markdown formatting.

## 🛠️ Changes Made
1. **Added `_normalize_text()` Helper**:
   - Unescapes literal CLI escape sequences (`\r\n`, `\n`, `\t`) into proper whitespace and newlines.
   - Handles edge cases such as empty strings, `None`, and standard text without modification.
2. **Applied Normalization in `handle_submit()`**:
   - Sanitizes both `args.title` and `args.body` before passing them to `create_pull_request()`.
3. **Added Comprehensive Unit Tests**:
   - Added `test_normalize_text_unescapes_cli_newlines_and_tabs` and `test_submit_normalizes_escaped_newlines_in_title_and_body` in `agents/platform/scripts/test_submit_suggestion.py`.

## 🧪 Verification
- Ran test suite: `python3 -m unittest discover -s agents/platform/scripts -p 'test_submit_suggestion.py' -v`
- **Result**: All 27 unit tests passed cleanly (including real bare repo git clone/branch/lease workflows and `gh` mock invocation checks).
- Verified that multiline PR descriptions containing `\n` render with proper paragraph breaks and markdown formatting.

### DIFF
--- a/agents/platform/skills/submit-suggestion/scripts/submit_suggestion.py
+++ b/agents/platform/skills/submit-suggestion/scripts/submit_suggestion.py
@@ -170,12 +170,21 @@ def handle_submit(args) -> int:
 
     push_branch(branch, workspace)
     base = gitops_workspace.resolve_base_branch(workspace, _runner)
-    pr_url = create_pull_request(branch, args.title, args.body, workspace, repo, base)
+    title = _normalize_text(args.title)
+    body = _normalize_text(args.body)
+    pr_url = create_pull_request(branch, title, body, workspace, repo, base)
     log(f"PR SUBMITTED SUCCESSFULLY! 🏆 URL: {pr_url}")
 
     # Print raw URL to stdout for the MCP tool to parse
     print(pr_url)
     return 0
+
+
+def _normalize_text(text: str) -> str:
+    """Unescape literal CLI escape sequences (like \\n or \\r\\n) into real whitespace."""
+    if not text:
+        return text
+    return text.replace("\\r\\n", "\n").replace("\\n", "\n").replace("\\t", "\t")
 
 
 def push_branch(branch_name: str, workspace: str) -> None:
--- a/agents/platform/scripts/test_submit_suggestion.py
+++ b/agents/platform/scripts/test_submit_suggestion.py
@@ -307,6 +307,28 @@ class TestSubmit(SubmitSuggestionTestCase):
         self.assertEqual(argv[argv.index("--head") + 1], "platform-agent/fix-netpol")
         self.assertEqual(argv[argv.index("--base") + 1], "main")
 
+    def test_normalize_text_unescapes_cli_newlines_and_tabs(self):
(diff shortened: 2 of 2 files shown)


jayantid (changes requested):
Can you fix kube-agents-bot comment and resend for review?

agents/platform/skills/submit-suggestion/scripts/submit_suggestion.py
```diff
@@ -170,14 +170,23 @@ def handle_submit(args) -> int:
...
     # Print raw URL to stdout for the MCP tool to parse
     print(pr_url)
     return 0
 
 
+def _normalize_text(text: str) -> str:
+    """Unescape literal CLI escape sequences (like \\n or \\r\\n) into real whitespace."""
+    if not text:
+        return text
+    return text.replace("\\r\\n", "\n").replace("\\n", "\n").replace("\\t", "\t")
```
jayantid: `_normalize_text` rewrites every literal `\n`, `\r\n`, and `\t` in the title and body, unconditionally. There is no way to keep a legitimate backslash sequence. Even a doubled backslash does not survive: `C:\\new\\temp` comes out as `C:\` + newline + `ew` + tab + `emp` (reproduced against this function).

PR bodies from this skill quote commands, code, and YAML. A body that quotes `printf("done\n")` or `sed 's/\n$//'` gets a real line break spliced into the quoted code, silently. The title path has the same problem, and a PR title is a single-line field.

Gate the unescape on the text being single-line. A caller that quotes properly (the SKILL.md Step 3 example) sends real newlines, so any literal `\n` alongside them is intentional. The flattened one-line body is exactly the broken case this PR targets:

```python
def _normalize_text(text: str) -> str:
    if not text or "\n" in text:
        return text
    return text.replace("\\r\\n", "\n").replace("\\n", "\n").replace("\\t", "\t")
```

Longer term, a `--body-file` flag (which `gh pr create` supports natively) removes the shell-quoting problem at its source.

Fuxiao-Gao:
Closing as superseded by #1466 (merged 2026-09-11).

The literal `\n` / `\t` in titles and bodies was a side effect of passing agent-written prose through argv. #1466 added `--body-file`: the description is written to a file under `/opt/data/scratch` with a quoted heredoc and streamed to `gh pr create --body-file -`, so no shell escaping step remains for the agent to get wrong. SKILL.md now teaches that channel and names backtick and escape handling as the reason.

Rebasing this on top would not add much and carries a small risk: unescaping `\n` blindly would also rewrite a real backslash-n inside a code block (printf or regex examples) for the `--body` callers that remain. Better to leave the plain channel as-is.

PR #815, opened 2026-08-19
feat(admin): add Google Chat integration status

haoxuw:
## Summary

This PR adds a read-only **Integration → Google Chat** page to the local admin portal. An ordered milestone checklist is the single source of truth: each check carries its own plain-language label, diagnosis, and fix steps, and the snapshot's status, message, and severity are derived from the first non-passed check. The page renders that payload without interpreting it.

## Why This Change

Operators currently correlate the custom resource, Google Chat console, Pub/Sub, IAM, and Hermes sessions by hand. A configured backend whose Chat console still points at an earlier installation's topic looks identical to a working one, and the console's saved topic has no read API.

## What Changed

- Added the Integration navigation group and Google Chat page after Observability, plus one authenticated, target-bound API endpoint (`GET /api/v1/integrations/google-chat`) and client method.
- The service reads the canonical `PlatformAgent`'s `spec.integration.googleChat` and verifies seven milestones: agent running, Chat configured on the agent, required Google APIs enabled, topic ready (existence + unconditional publisher bindings), subscription ready (existence, route, active, pull delivery, agent IAM), Chat-app-sends-here, and message received.
- While no message has been observed by a complete read, the service lists the project's other Pub/Sub topics (bounded at 15) and flags those accepting the Google Chat publisher identities as the likely root cause — a console still pointed at a previous install's topic.
- The verdict never over-claims: all-green reports `Ready` with the timestamp its delivery evidence is valid as of; a healthy-but-untested install reports `Ready` with install-and-test instructions; a failed or truncated activity read reports `Verification incomplete` rather than silence.
- A next-step block under the green banner carries the test instruction and the copyable console link, topic, and subscription, because delivery evidence only proves the console setting as of the last received message.
- Kubernetes evidence is redacted before serialization, including bare EnvVar credential segments such as `API_SERVER_KEY`; raw command evidence sits behind a collapsed section.

## Context

#814 (canonical agent identity and the authenticated portal API) merged first; this branch is rebased onto current `main` (`a4f1b0f`) and carries two commits.

## Testing

- `.venv/bin/python -m unittest discover -s admin_console/tests -p 'test_*.py' -q` — 173 passed, including negative paths: conditional IAM bindings, push subscriptions, missing subscriber bindings, kube read failure, failed and truncated activity reads, stray-topic detection, and EnvVar redaction.
- `make docs-check` — passed. Prettier 3.9.6 on changed Markdown — passed. `python -m compileall admin_console` — passed.

### Live validation

- Install: project `haoxuw-gkedemos`, cluster `platform-agent-host`, location `us-east4`, namespace `kubeagents-system`; `PlatformAgent/platform-agent` at `Ready=True`. All observations were made through the running portal (`scripts/admin_portal.sh`) and its authenticated API, not by unit tests.
- **Root-cause state observed live:** with the Chat console pointed at a stale topic and zero delivered messages, the endpoint returned the failed `chat_console_topic` check naming the two real stray topics in the project (`hermes-chat-events`, `sre-chat` — both carry Chat publisher bindings from earlier installs) and the page showed the red banner with the console link and exact topic to paste.
- **End-to-end state observed live:** after correcting the console topic and sending the app a direct message, Hermes recorded the conversation and the page flipped to `Ready` with all seven milestones passed; the delivery-evidence timestamp (`2026-08-25T20:27:54Z`) renders in the banner with the "as of that moment" caveat. The screenshot below is the served page at the pushed commit:

google-chat-ready-milestones

_Published 20260827T141432Z at commit 5fc28a5910d0 — pre-captured image (pr_evidence.png)._

- Not covered live: the `Verification incomplete` path for a failed activity read (exercised by tests only — the live gateway stayed healthy); a Cloud Monitoring publish-metric probe of the console topic (rejected by corp context-aware access for non-gcloud clients on the test machine). No Kubernetes or Google Cloud resources were mutated; the stale Chat console topic used to produce the failure state was restored by its owner as part of the test.

## Self-Review

Both required passes ran in fresh subagent contexts that did not write the change, handed only the repo, the range (`refs/kube-agents-base/main...HEAD`), the skill path, and the gate output. They reviewed the pre-fix head; the fixes below were then verified by the mechanical gates and the test suite, not by a fresh pass.

`review-adversarial` (subagent) covered all ten angles: intent-vs-diff, removed behavior, cross-file contracts (CRD fields, IAM identities against `terraform/modules/chat-pubsub`), security/ops (endpoint auth middleware, read-only command inventory, redaction), reuse, conventions, scope, test coverage, and sibling PRs. `review-docs-drift` (subagent) verified doc claims against source for both changed docs and the map. Dispositions:

- **MEDIUM, CONFIRMED (adversarial): a failed or truncated activity read was reported as "no message has arrived", turning a gateway blip plus a leftover topic into a false red verdict.** Fixed: only a complete empty read counts as silence; otherwise the console check is `unknown` with verdict `Verification incomplete`, stray-topic reads are skipped, and two regression tests cover the `AgentRuntimeError` and truncated paths.
- **Blocking (docs-drift): the design doc claimed the Chat and Workspace Add-ons registrations "resolve to the same service account"; the check verifies two distinct identities.** Fixed: reworded to say the check proves the two publisher grants exist, not that the console registration was saved.
- **LOW, CONFIRMED (adversarial): EnvVar names like `API_SERVER_KEY` escaped redaction while the page promised it.** Fixed: credential words are now matched as whole underscore segments (`_ENV_SECRET_NAME`), with a test.
- **Blocker (reviewer, post-push): the redaction had a bypass — `kubectl.kubernetes.io/last-applied-configuration` nests the applied spec as a JSON string, so the same secret appeared redacted and unredacted in one evidence blob.** Fixed in c35bf48: the scrub now recurses into string values that parse as JSON documents, with the reviewer's client-side-apply reproduction as a regression test; replaying their exact scenario yields no leak and both copies redacted.
- **LOW, CONFIRMED (adversarial): a fully-working integration still shows install/test steps as its next step, which the README did not describe.** Deliberately kept: delivery evidence goes stale (the console can break after the last received message), so the test handoff stays available in the green state; the README now documents this.
- **Advisory (docs-drift): the design doc's Summary, Architecture diagram, and readiness table omitted the new surface.** Fixed: all three now include it.
- **LOW, PLAUSIBLE (adversarial), open question: worst-case service time (~160s of stacked per-command timeouts) exceeds the client's 120s timeout, discarding gathered evidence.** No change; typical runs finish far under budget and the sibling `inspect_llm_gateway` shares the pattern. Worth a shared fix if it bites.
- **Advisory (docs-drift), pre-existing: `AGENTS.md` Repository Layout does not list `admin_console/` at all.** Not this PR's to fix; noted for a follow-up.

Earlier rounds in the authoring sessions (screenshot review of information hierarchy, hostile re-read of the redaction path) drove the checklist consolidation and plain-language labels; their findings are superseded by the passes above, which ran against the final design.

## Risk & Rollout

The inspector is read-only: seven parallel Google Cloud reads, one bounded Hermes history read, and up to 15 additional topic IAM reads only when a complete activity read found zero messages. Permission failures surface as `Could not verify`, never as missing resources. Reverting the two commits removes the page, endpoint, client method, and redaction helper without changing the installation.


stalhaali (changes requested):
🔴 Blocker — the evidence redaction has a bypass that puts the secret back in the same blob

  admin_console/google_chat_integration.py:479 runs redact_kubernetes_evidence() over kubectl get platformagents -o json, and the page caption (pages/google_chat.py:170) promises "Credential-shaped Kubernetes values are redacted." The scrub in
  telemetry.py:161 only walks parsed JSON — a value that is itself a JSON string is returned untouched, and redact_evidence's regex fallback doesn't fire on escaped JSON (\"name\":\"X_TOKEN\",\"value\":\"…" has no token=/token: shape).

  kubectl.kubernetes.io/last-applied-configuration is exactly that shape. I ran it:

  direct env leaked:          False
  full output contains leak:  True
  
  {"items":[{"metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":
    "{\"spec\":{\"deployment\":{\"env\":[{\"name\":\"API_SERVER_KEY\",\"value\":\"LAST-APPLIED-LEAK\"}]}}}"},
    ...},"spec":{"deployment":{"env":[{"name":"API_SERVER_KEY","value":"[REDACTED]"}]}}}]}

  Redacted and unredacted copies, same string, adjacent.

  Reachable on the documented path: INSTALL.md:499 is kubectl apply -f examples/platformagent.yaml — client-side apply, which writes that annotation. spec.deployment.env is []corev1.EnvVar (k8s-operator/api/v1alpha1/common_types.go:372),
  literal values allowed; SensitiveEnvVars only rejects API_SERVER_KEY, CREDENTIAL_PROXY_ENFORCE_READ_ONLY, HERMES_HOME, so LITELLM_MASTER_KEY and EXTERNAL_API_TOKEN — the PR's own two test fixtures — pass admission.

  Not privilege escalation (the viewer already has cluster read), but it's a stated control with a dedicated test that doesn't hold. Fix is small: recurse into string values that parse as JSON, or drop last-applied-configuration from the
  evidence copy. Either way it needs a test, since the existing one plants the secret only in the parsed position.

haoxuw:
@stalhaali Fixed in c35bf48. The scrub in `redact_kubernetes_evidence` now recurses into string values that parse as JSON documents, so the `last-applied-configuration` copy gets the same EnvVar-name redaction as the parsed spec. Your reproduction is the regression test (`test_last_applied_configuration_annotation_is_scrubbed`): the secret planted only in the annotation string, asserting it is absent from the evidence while a non-credential value survives. Replaying your exact scenario against the fix: leak present: False, both copies show [REDACTED]. I chose recursion over dropping the annotation so any other JSON-in-string shape (env values holding JSON, other annotations) is covered too. The PR body's Self-Review carries the disposition.

stalhaali (approved):

PR #498, opened 2026-07-31
tests: set up evaluation framework for incident skill

justinsb:

PR #819, opened 2026-08-20
test: add coverage measurement and a report-only PR coverage comment

jayantid:
## Summary

Adds test-coverage measurement to the repository and a report-only coverage comment to every pull request. `make coverage` runs the existing Python suite under coverage.py (mirroring `PYTHON_TEST_DIRS` exactly) and the Go suite with `-coverpkg=./...`, producing Cobertura XML. A new `coverage` job in `python-tests.yml` posts one sticky comment per PR showing the repository total and the coverage of the PR's own changed lines, with the uncovered lines listed by file. Nothing blocks: this PR is the meter, not the gate.

## Why This Change

The repository has roughly 2,700 Python tests and an operator suite, and no way to see what they cover. Every conversation about test quality starts from an assertion instead of a number, and a PR that adds 200 untested lines looks identical to one that adds 200 tested ones. Measured while building this: the Python suite covers 61.8% of 22,018 statements — a number nobody had before this job existed.

Two measurement traps are handled and documented in `.coveragerc`, because getting either wrong silently inflates the number by ~20 points: `include_namespace_packages` must be set (and must be under `[report]`, not `[run]`) because almost nothing in this repository is a package, and Go needs `-coverpkg=./...` or untested packages drop out of the denominator.

## What Changed

- `.coveragerc` — measurement config; the comments carry the two traps above
- `Makefile` — `make coverage` (both languages, Cobertura out) and `make coverage-check` (total floor, off by default; the real floor gets committed from the first green CI run, not from a laptop)
- `.github/workflows/python-tests.yml` — the `coverage` job: measure, diff-cover against the merge base, one sticky comment rewritten in place. Same-repo PRs only for the comment (fork tokens are read-only by GitHub's rules); fork PRs get the report as a job artifact. No third-party actions added, so no new SHA pins
- `requirements-test.txt` — `coverage`, `diff-cover`

## Context

First PR of the testing-strategy implementation work. The blocking rules (diff-coverage floor, no-test detector, total ratchet) come later, staged, after this job has collected the real distribution of diff coverage across merged PRs. Merge adjacency, not conflicts: #816 touches this workflow (concurrency groups), #773 touches `requirements-test.txt`, #498 and #720 touch the Makefile.

## Testing

- `make coverage COVERAGE_SKIP_GO=1` run end-to-end on this branch rebased onto current main: 17 directories measured, combine and XML succeed, TOTAL 61.8%
- `make coverage-check` verified in both directions (passes at floor 55, fails at 95)
- All six diff-cover exclusion patterns tested programmatically against diff-cover 10.5.1's `_is_path_excluded` (excluded files match, neighbours do not) and end-to-end
- `make docs-check` green; workflow YAML parses; Makefile parses (`make -n coverage`)
- `prettier --check` not run locally (no node/npm on this machine); the changed YAML follows the file's existing formatting and CI's prettier job is the arbiter

### Live validation

Not live-tested: this change is CI tooling and build targets only; it has no runtime code path and nothing for an operator or agent pod to pick up. The workflow job itself can only be exercised by CI on this PR — the coverage comment (or, from a fork, the `diff-cover-report` artifact) appearing on this PR is the live test.

## Self-Review

Adversarial pass run by a separate reviewer context that did not write the change, against the branch diff, following `.agents/skills/review-adversarial/SKILL.md` (angles A–J), plus a docs-drift pass per `.agents/skills/review-docs-drift/SKILL.md`. Findings and dispositions:

- **Blocking, fixed:** the comment step failed on every fork PR — GitHub forces the token read-only regardless of the `permissions:` block, so the POST 403s and reds a report-only job. Fixed by gating the comment on same-repo PRs and uploading the report as an artifact for everyone
- **Should-fix, fixed:** checkout persisted a write-scoped token on disk while the job runs PR-authored Makefile and test code; `persist-credentials: false` added. A workflow comment cited a design doc not yet in the tree; replaced with the rollout description inline
- **Nits, fixed:** `coverage`/`coverage-check` added to `.PHONY`; the summary line greps `TOTAL` instead of `tail -1` (which printed coverage.py's `skip_empty` footer); empty `PYTHON_TEST_DIRS` now fails loudly; the sticky-comment lookup only matches comments authored by `github-actions[bot]`, so a marker planted by another user is never patched; the body is truncated under GitHub's 65,536-character cap
- Hunted and found clean: fnmatch exclusion semantics (verified against diff-cover source and end-to-end), shell-injection paths into `gh api` (PR content only travels via `-F body=@file`), merge-base correctness for fork PRs, one failing test directory zeroing the total (it does not; reproduced), action SHA pins

## Risk & Rollout

Low. No runtime code path is touched; the job is report-only (`--fail-under 0`) and its failure does not block the `test` job or the merge. Revert is deleting the job and the two make targets. One follow-up owed at merge time: after the first green run of this job on main, commit its measured total as the `coverage-check` floor.


.github/workflows/python-tests.yml
```diff
@@ -43,3 +43,120 @@ jobs:
...
+            --markdown-report diff-cover.md \
+            --fail-under 0 \
+            --exclude \
+              "*/deploy/docker/patches/verify_*.py" \
+              "*/deploy/docker/patches/kanban*.py" \
+              "*/deploy/docker/patches/cron*.py" \
+              "*/deploy/docker/patches/channel_directory_threads*.py" \
+              "*/deploy/docker/patches/slack_code_emphasis*.py" \
+              "*/examples/*"
+
```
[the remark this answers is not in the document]
jayantid: Checked against diff-cover 10.5.1 as installed from PyPI, and the finding doesn't hold: exclusion does not match the repo-relative `src_path`. `GitDiffReporter._is_path_excluded` (`diff_cover/diff_reporter.py:74-102`) matches the **basename** first and then the **absolute** path:

```python
basename = os.path.basename(path)
if self._fnmatch(basename, exclude):
    return True
absolute_path = os.path.abspath(path)
return self._fnmatch(absolute_path, exclude)
```

Since the absolute path always carries directory components above the repo root, a leading `*/` is what matches, and a bare relative pattern is what silently matches nothing. Reproduced directly:

```python
>>> r = GitDiffReporter(git_diff=None, exclude=['*/deploy/docker/patches/verify_*.py', '*/examples/*'])
>>> r._is_path_excluded('deploy/docker/patches/verify_kanban_notifier.py')
True
>>> r._is_path_excluded('examples/inference-replay/replay-proxy/replay_proxy.py')
True
>>> r2 = GitDiffReporter(git_diff=None, exclude=['deploy/docker/patches/verify_*.py'])
>>> r2._is_path_excluded('deploy/docker/patches/verify_kanban_notifier.py')
False
```

The `--exclude setup.py` example from diff-cover's docs works through the *basename* branch, which is why it looks like relative matching but isn't. The workflow comment and the Self-Review claim stand as written; the patterns as committed are the ones that exclude.

.github/workflows/python-tests.yml
```diff
@@ -43,3 +43,125 @@ jobs:
...
```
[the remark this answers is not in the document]
jayantid: The mechanism in this finding is the same one refuted on the previous thread: diff-cover 10.5.1's `_is_path_excluded` (`diff_cover/diff_reporter.py:74-102`) matches the basename first and then `os.path.abspath(path)` — not the repo-relative path — so the leading-`*/` patterns did match (reproduced there with the installed package).

That said, the patterns are changed in 496d5d4 anyway, for a reason the finding points toward even with the mechanism wrong: `diff-cover` is deliberately unpinned in `requirements-test.txt`, and a bare `*` prefix (`*deploy/docker/patches/verify_*.py`) matches under absolute-path semantics AND plain relative matching — verified both ways:

```python
>>> GitDiffReporter(git_diff=None, exclude=['*deploy/docker/patches/verify_*.py'])._is_path_excluded('deploy/docker/patches/verify_kanban_notifier.py')
True
>>> fnmatch.fnmatch('deploy/docker/patches/verify_kanban_notifier.py', '*deploy/docker/patches/verify_*.py')
True
```

So the exclusions no longer depend on which semantics an upstream release uses, and the step comment now names the actual mechanism with its source location.

jayantid:
Re the High finding (comment gated to same-repo PRs while every PR here is a fork PR): correct, and it's the known limitation this PR ships with deliberately — the in-job step cannot post on fork PRs at all (read-only token), so gating it is the difference between a silent skip and a red job. Fork PRs get the report as the `diff-cover-report` artifact in the meantime.

The real fix is already up as the stacked companion #827: the comment moves to a `workflow_run` workflow running in the base repo's context, which posts for every PR including forks, and the same-repo gate plus the in-job step are deleted there. Merge order: this PR first, #827 immediately after — the PR bodies of both say so.

/request-review

.github/workflows/python-tests.yml
```diff
@@ -43,3 +43,129 @@ jobs:
...
+      # One sticky comment rewritten in place, never a new comment per push.
+      # Plain gh api rather than a third-party action: the logic is ten lines,
+      # and it is one less pinned SHA to keep current.
+      #
+      # Same-repo pull requests only: on a fork PR GitHub forces the token to
+      # read-only whatever `permissions:` asks for, so the POST would 403 and
+      # red a job whose whole point is to block nothing. Fork PRs get the
+      # artifact upload below instead.
+      - name: Post the coverage comment
+        if: github.event.pull_request.head.repo.full_name == github.repository
```
toshiowang: **Definitive bug: this reds the job on every Dependabot PR.**

The guard catches forks but not Dependabot. Dependabot pushes its branches into `gke-labs/kube-agents` itself, so `head.repo.full_name == github.repository` is true and the step runs — but GitHub gives a `pull_request` run whose actor is `dependabot[bot]` a read-only `GITHUB_TOKEN` regardless of the `permissions:` block, exactly as it does for forks. The `POST repos/.../issues/.../comments` 403s, `set -e` kills the step, and a job whose whole point is to block nothing goes red. `.github/dependabot.yml` runs five ecosystems at `open-pull-requests-limit: 5`, so this is the common case, not an edge one.

```suggestion
        if: github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'
```

Makefile
```diff
@@ -198,6 +198,76 @@ test-python: ## Run the Python unit tests outside k8s-operator/.
...
+# drop out of the denominator and the number reads ~10 points high.
+	@if [ "$(COVERAGE_SKIP_GO)" = "1" ]; then \
+		echo "Skipping Go coverage (COVERAGE_SKIP_GO=1)."; \
+	elif ! command -v gocover-cobertura >/dev/null 2>&1; then \
+		echo "Skipping Go coverage: gocover-cobertura not installed."; \
+		echo "  go install github.com/boumenot/gocover-cobertura@latest"; \
+	else \
+		$(MAKE) -C k8s-operator setup-envtest && \
+		(cd k8s-operator && \
+			KUBEBUILDER_ASSETS="$$(bin/setup-envtest use 1.31.0 --bin-dir bin -p path)" \
```
toshiowang: **Definitive bug, and it fails silently: envtest version drift.**

`1.31.0` is hardcoded here, but `k8s-operator/Makefile:239` pins `ENVTEST_K8S_VERSION ?= 1.36.0`. The line above provisions 1.36.0 into `bin/`; this line then asks for a version that is not there. `setup-envtest use 1.31.0` either downloads a second, different control-plane build or exits non-zero — and on the failure path the command substitution yields the empty string, `KUBEBUILDER_ASSETS=""` is exported, every envtest suite fails, and the whole block is swallowed by the `|| echo "Go coverage failed; …"` two lines down.

Concrete: on a machine with `gocover-cobertura` installed and no reachable envtest download mirror, `make coverage` prints "Go coverage failed", writes no `coverage-go.xml`, and exits 0. The advertised Go half has never produced a number and nothing says so. When it does succeed it measures against a Kubernetes version the operator's own suite does not use.

```suggestion
			KUBEBUILDER_ASSETS="$$(bin/setup-envtest use $$(sed -n 's/^ENVTEST_K8S_VERSION ?= //p' Makefile) --bin-dir bin -p path)" \
```

.github/workflows/python-tests.yml
```diff
@@ -43,3 +43,129 @@ jobs:
...
+            --compare-branch "origin/${BASE_REF}" \
+            --markdown-report diff-cover.md \
+            --fail-under 0 \
+            --exclude \
+              "*deploy/docker/patches/verify_*.py" \
+              "*deploy/docker/patches/kanban*.py" \
+              "*deploy/docker/patches/cron*.py" \
+              "*deploy/docker/patches/channel_directory_threads*.py" \
+              "*deploy/docker/patches/slack_code_emphasis*.py" \
+              "*examples/*"
```
toshiowang, line 164: **Tentative: the two numbers in one comment are computed over different file sets.**

These six paths are excluded from diff coverage but are not omitted in `.coveragerc`, so they remain in the repository-total denominator the same comment prints. A PR adding 300 uncovered lines under `examples/` reports 100% diff coverage while dragging the total down, and a reader cannot reconcile the two figures.

It also undercuts the staged rollout: a ratchet on the total and a floor on the diff would be measured against different file sets, so the total can regress on files the floor is structurally blind to. Either omit the same paths in `.coveragerc`, or drop them from `--exclude` and let them count in both.

.github/workflows/python-tests.yml
```diff
@@ -43,3 +43,129 @@ jobs:
...
+
+  # Report-only for now: it posts numbers on the pull request and blocks
+  # nothing. The blocking rules (diff-coverage floor, no-test detector, total
+  # ratchet) turn on in stages once a baseline of merged-PR data exists --
+  # first the no-test detector, then the floor and the ratchet, with the
+  # thresholds set from the observed distribution rather than this file.
+  coverage:
+    name: Coverage report
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
```
toshiowang: **Tentative: the job can never produce the number the rollout depends on.**

The workflow triggers on `pull_request` and `push: branches: [main]`; this guard skips the job on the push. The PR body's follow-up — "after the first green run of this job on main, commit its measured total as the `coverage-check` floor" — has no run to read, and the later total-ratchet has no per-commit baseline on `main`.

Not removable as a one-liner: on a push `github.base_ref` is empty, so `--compare-branch "origin/"` would fail. Enabling main runs means branching the compare ref (`HEAD~1`) or splitting the measure step from the diff step.

.github/workflows/python-tests.yml
```diff
@@ -43,3 +43,129 @@ jobs:
...
+            head -c 60000 diff-cover.md
+            echo
+            echo "_Report-only: nothing here blocks this pull request yet._"
+          } > comment.md
+          # Only consider comments this workflow's identity wrote: anyone can
+          # start a comment with the marker, and PATCHing theirs would render
+          # the report under their name.
+          cid=$(gh api "repos/$REPO/issues/$PR/comments" --paginate \
+                --jq ".[] | select(.user.login == \"github-actions[bot]\")
+                          | select(.body | startswith(\"$marker\")) | .id" | head -1)
```
toshiowang: **Tentative, low: `head` can SIGPIPE `gh` under the default `pipefail` shell.**

GitHub's default `run:` shell is `bash -e -o pipefail`. `head -1` exits as soon as it reads a line, so if `gh api --paginate` is still fetching later pages and writes again it takes SIGPIPE (141), pipefail propagates it, and `set -e` fails the step — on a PR with more than one page of comments. `tail` reads to EOF and cannot hit this; there is at most one matching comment either way.

```suggestion
                          | select(.body | startswith(\"$marker\")) | .id" | tail -1)
```

toshiowang (changes requested):
Two definitive bugs, three tentative ones.

The fork guard on the comment step does not catch Dependabot (same-repo branch, read-only token), so the report-only job will go red on every Dependabot PR — five ecosystems at five open PRs each. And the Go half of `make coverage` hardcodes envtest `1.31.0` against the operator's pinned `ENVTEST_K8S_VERSION ?= 1.36.0`, so it either measures the wrong control plane or fails into the `|| echo` and produces no `coverage-go.xml` at all, silently.

Beyond those: the six `--exclude` paths are exempt from diff coverage but still in the repository-total denominator, so the comment's two numbers are computed over different file sets and the later floor and ratchet would disagree with each other; and the `if: github.event_name == 'pull_request'` guard means the job never runs on a push to `main`, so the "commit the floor from the first green run on main" follow-up has no run to read.

Verified clean: the upload-artifact SHA is really v7.0.1 and matches `docs-build.yml`; `persist-credentials: false` is correctly placed; `-F body=@file` does not base64 the body; `fetch-depth: 0` populates `origin/*` so `--compare-branch` resolves; the fnmatch exclusion patterns work under both relative and absolute matching; `include_namespace_packages` is correctly under `[report]`; the coverage loop mirrors `test-python`'s discovery exactly.

- `k8s-operator/Makefile:239` — `ENVTEST_K8S_VERSION ?= 1.36.0`, the value `Makefile:250` should be reading rather than restating.
- `.coveragerc:41` — `ignore_errors = True` silences every unreadable-source error, not only the synthetic-module case the comment above it names; an in-scope file that fails to parse drops out of the denominator with no signal. Not confirmed to be biting today, but it is a way for the number this PR exists to produce to be quietly wrong.

.github/workflows/python-tests.yml
jayantid: Fixed in cf9f36d: the comment step's if now also requires github.actor != 'dependabot[bot]', with the comment explaining that Dependabot's token is read-only regardless of permissions, same as a fork's. Those runs take the artifact path.

Makefile
jayantid: Fixed in cf9f36d: the version is now read from k8s-operator/Makefile's own ENVTEST_K8S_VERSION pin (sed at run time, so it cannot drift again), and both the version read and the setup-envtest output are guarded with test -n so the empty-KUBEBUILDER_ASSETS failure mode fails loudly inside the guarded block instead of being swallowed. The comment above the block records the drift you found as the reason.

.github/workflows/python-tests.yml
jayantid: Fixed in cf9f36d: .coveragerc now omits the same six paths diff-cover excludes, with a comment in each file pointing at the other ('change the two lists together'). One file set for both numbers; the measured total moves from 61.8% to 67.0% on the smaller, honest denominator. You're right this was also load-bearing for the rollout — the ratchet and the floor now watch the same tree.

.github/workflows/python-tests.yml
jayantid: Fixed in cf9f36d: the job-level pull_request gate is gone; on pushes to main the job measures and uploads the artifact only (diff-cover and the comment are step-gated on pull_request, so the empty base_ref path is unreachable). The rollout follow-up now has a run to read, and the ratchet a per-commit baseline on main.

.github/workflows/python-tests.yml
jayantid: Taken in cf9f36d: tail -1, with your reasoning in the comment (reads to EOF, cannot SIGPIPE a still-paginating gh; at most one matching comment either way).

.coveragerc
```diff
@@ -0,0 +1,58 @@
...
+    # set. Without this, a PR adding uncovered lines under an excluded path
+    # reads 100% on the diff while dragging down a total the diff floor is
+    # structurally blind to -- and the two numbers in one comment cannot be
+    # reconciled. Change the two lists together.
+    */deploy/docker/patches/verify_*.py
+    */deploy/docker/patches/kanban*.py
+    */deploy/docker/patches/cron*.py
+    */deploy/docker/patches/channel_directory_threads*.py
+    */deploy/docker/patches/slack_code_emphasis*.py
+    */examples/*
```
toshiowang, line 39: **Definitive: three of these five patterns exclude tested runtime code, and the two that matter are undocumented.**

This list is my doing — I asked for the two file sets to be reconciled and offered "omit the same paths in `.coveragerc`, or drop them from `--exclude`". Copying the list across was the branch that looked cheaper, but nobody then asked what the six patterns actually sweep up. Measured on this branch, with the five patches patterns removed from `omit` and `deploy/docker/patches/` re-run:

| pattern | statements | covered |
| --- | --- | --- |
| `verify_*.py` | 2,669 | **0.0%** |
| `kanban*.py` | 976 | 97.0% |
| `cron*.py` | 346 | 83.8% |
| `channel_directory_threads*.py` | 69 | 100% |

It splits in two.

**`verify_*.py` is 7 points of the headline.** 2,669 statements at 0% — build-time verifiers the Dockerfile runs and no test imports. With them in, the repository total is 13,859/21,684 = 63.9%; the job reports 71.2%. Excluding untested build tooling from a coverage meter is a defensible call, but the diff nowhere says these are build verifiers, nowhere says the exclusion is what moves the number, and the comment above this list explains only the mirroring. The PR that establishes the meter is the one place that has to state it, because the trap is the same one the `[report]` section warns about: an exclusion nobody can see, worth ~20 points there and ~7 here.

**The other three are collateral.** `kanban*.py`, `cron*.py` and `channel_directory_threads*.py` are 1,391 statements of runtime payload modules at 94% — `kanban_notifier.py` and friends, imported directly by their own `test_*.py` in the same directory. They are not build tooling and they share nothing with `verify_*` but a directory. Dropping them removes well-tested first-party code from the denominator, and once the staged diff floor turns on it exempts every change to the most actively patched area of this repository from the gate that floor exists to be.

Fix is the same in both files — keep the verifiers, name why, drop the rest. Here, and the matching three lines at `python-tests.yml:160-162`:

```suggestion
    # Build-time verifiers: run by deploy/docker/Dockerfile, imported by no
    # test. 2,669 statements at 0%, and excluding them is worth ~7 points of
    # the reported total -- the one entry in this list that moves the number.
    */deploy/docker/patches/verify_*.py
    */examples/*
```

.github/workflows/python-tests.yml
```diff
@@ -78,3 +78,148 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        env:
+          # Through the environment rather than interpolated into the script:
+          # zizmor flags template expansion inside `run:` as an injection
+          # surface, and the env var is inert text to the shell parser.
+          BASE_REF: ${{ github.base_ref }}
+        run: |
+          diff-cover coverage.xml \
+            --compare-branch "origin/${BASE_REF}" \
+            --markdown-report diff-cover.md \
```
toshiowang, line 156: **Definitive, dated: `--markdown-report` is deprecated and diff-cover is deliberately unpinned.**

The `Coverage report` job on this head already prints it:

```
diff_cover/diff_cover_tool.py:394: UserWarning: The --markdown-report option is
deprecated. Use --format markdown:diff-cover.md instead.
```

`requirements-test.txt` pins no version, by design — that is the argument you made on the pattern thread, that the exclusions must not depend on which semantics an upstream release uses. The same argument applies here and lands harder: when 11.x drops the flag, `diff-cover: error: unrecognized arguments: --markdown-report` reds this job on every pull request, with no change to this repository and nothing to point at.

`--format markdown:diff-cover.md` produces a byte-identical report under 10.5.1 (checked against the installed package).

```suggestion
            --format markdown:diff-cover.md \
```

toshiowang (changes requested):
Reviewed at `cf9f36d`. I reproduced the pipeline rather than reading the diff: ran `make coverage COVERAGE_SKIP_GO=1` on the branch, regenerated `coverage.xml`, ran `diff-cover` with the shipped flags against 10.5.1, and read the `Coverage report` job log from run 32402717705.

Most of it holds up. Path resolution between `coverage.xml`'s absolute `<sources>` root and diff-cover's repo-relative diff paths works — probed with an in-place edit to a known-uncovered line, which came back as `apply_kanban_notifier.py (0.0%): Missing lines 246`. `include_namespace_packages` is genuinely a `[report]` option. The `${COVERAGE_ROOT-.}` substitution, the `#` comment lines nested inside the multi-line `omit` value, `combine` locating parallel files via `COVERAGE_FILE`'s directory, `--format=total`, the `sed` read of `ENVTEST_K8S_VERSION`, and the `upload-artifact` SHA against the existing `docs-build.yml` pin all check out. `diff-cover.md` is written even when the diff has no measured lines, so the comment step's `head -c` cannot fail on a docs-only PR. And your refutation of the two `*/`-pattern findings is right — `_is_path_excluded` matches basename then `os.path.abspath`, so both forms match; the `*` form you moved to is the more durable one.

Two things to change, plus a number that needs correcting.

## The headline number is now three different numbers

The PR body says "61.8% of 22,018 statements". Your reply on the file-set thread says the total moved to 67.0%. The `Coverage report` job on this head prints `TOTAL 17624 5071 71.2%`.

None of the three agrees, and the 61.8% figure is what the **Why This Change** section rests on — it is the pre-exclusion measurement, from a configuration this PR no longer ships. Whatever the final call on the exclusions below, the body needs the number the job actually produces.

jayantid:
Closing: this PR's entire content merged to main inside #827 (which was stacked on this branch — the squash carried both). make coverage, coverage-check, the .coveragerc, and the report-only CI job are all live on main; toshiowang's five findings from the changes-requested review were fixed on the branch and are included in what merged. Nothing here is lost.

PR #565, opened 2026-08-06
feat(submit-suggestion): add token consumption and remediation SLA metrics to PR description

fkc1e100:
## Summary

Adds automated token consumption sourcing and remediation SLA telemetry to GitOps Pull Requests submitted by the Platform Agent via `submit_suggestion.py`. When `HERMES_SESSION_ID` is present, the script queries measured token metrics (input tokens, output tokens, cache read/write tokens) directly from the Hermes session endpoint (`http://platform-agent.kubeagents-system.svc.cluster.local:8642/api/sessions/<id>`) and appends both a formatted summary section and a structured machine-readable JSON comment to the PR description. Agent-supplied flags (`--elapsed`, `--model`, `--steps`, `--trace-id`, `--input-tokens`, `--output-tokens`) allow augmenting or explicitly providing verified metrics.

## Why This Change

Platform Agent suggestion PRs previously lacked operational telemetry and audit provenance. Without visibility into runtime SLA duration, model provenance, and token consumption metrics, SRE and FinOps teams could not audit agent resource usage or track remediation SLAs across automated suggestions.

## What Changed

- Updated `agents/platform/skills/submit-suggestion/scripts/submit_suggestion.py`:
  - Configured `DEFAULT_HERMES_API_ENDPOINT` to `http://platform-agent.kubeagents-system.svc.cluster.local:8642` with fallback to `HERMES_API_ENDPOINT`.
  - Added `Authorization: Bearer $PLATFORM_AGENT_TOKEN` header when `PLATFORM_AGENT_TOKEN` is set.
  - Unwrapped the `session` JSON envelope (`data.get("session")`) to extract `input_tokens`, `output_tokens`, `cache_read_tokens`, and `cache_write_tokens`.
  - Accounted for session cache reads and writes in `total_tokens` calculations (`inp + cr + cw + out`), avoiding double-counting reasoning tokens.
  - Dropped deprecated free-text `--tokens` argument and `HERMES_SESSION_TOKENS` fallback in favor of strictly typed integers (`--input-tokens`, `--output-tokens`).
  - Added structured metadata comment `<!-- kube-agents-telemetry: {...} -->` with typed integers and duration strings.
- Updated `agents/platform/skills/submit-suggestion/SKILL.md`:
  - Clarified that `HERMES_SESSION_ID` is automatically runtime-injected while remaining telemetry attributes are agent-supplied.
  - Documented partial coverage scope (one-off suggestion PRs vs scheduled fleet audits via `audit_report.py`).
[...]

PR #820, opened 2026-08-20
test: cover the observability scripts and memory curator, fix a protected-branch bypass

jayantid:
## Summary

Closes the largest measured unit-test gaps outside the operator: the five observability skill scripts (317 statements, 0% covered) and the memory TTL curator (187 statements, 0%), and fixes a real authority bug the work uncovered — `submit_suggestion.py` refused pushes to `main` but not to `refs/heads/main`, so a fully-qualified ref walked through the protected-branch check. 105 tests on the branch, all green.

## Why This Change

The observability scripts produce the latency, token-usage, and user numbers the Platform Agent states to SREs as fact. A parsing or unit slip there is the wrong-but-fluent failure no LLM judge catches, because the prose reads fine; the only place to catch it is a unit test with a recorded payload and a pinned expectation. They had zero. The TTL curator deletes agent memory unattended on a schedule; its selection logic (what dies, what survives, what happens to a malformed row) also had zero tests. And the protected-branch bypass is exactly the class of bug the testing strategy exists for: the agent holds push credentials, and the guard checked the spelling rather than the destination.

## What Changed

- `agents/platform/scripts/` — `observability_script_harness.py` (runs each script via `runpy` with `gcloud` and `urlopen` stubbed; nothing touches the network) plus five test files, 26 tests: recorded payloads parse to the expected figures, empty responses say "no data" rather than 0, and the seconds/milliseconds arithmetic is pinned so a 1000x slip fails
- `agents/chat/scripts/test_memory_ttl_curator.py` — 23 tests: all five abort guards, the TTL boundary (a row at exactly the cutoff survives), an unparseable date is skipped rather than deleted, dry run writes nothing, and the retry policy (a retain must not retry a 500)
- `agents/platform/skills/submit-suggestion/scripts/submit_suggestion.py` — strips a `refs/heads/` prefix (case-insensitive) before the protected-branch comparison; the original comparison is untouched when the prefix is absent. The new test in `test_submit_suggestion.py` was written first and failed on the old code for `refs/heads/main`, `refs/heads/MASTER`, and `REFS/HEADS/production`

## Context

Part of the testing-strategy implementation work; companion to the coverage-measurement PR, which makes deltas like these visible on every PR. Merge adjacency: #813 and #565 touch `submit_suggestion.py` (different lines, different concerns — whoever merges second should re-run its tests), #504 carries a pre-fix copy of the file in its diff context but does not modify the guard.

An earlier draft of this branch also added tests and a Makefile glob for `agents/platform/plugins/incident_context/` — both arrived upstream independently while the branch was in review, so that commit was dropped in favour of upstream's. Two small angles upstream's tests do not cover (Authorization-header handling, enum-valued platforms) are noted for a follow-up.

## Testing

- All 52 tests in the six touched `agents/platform/scripts/` modules pass in one process (which also demonstrates the runpy harness leaks no state between suites)
- 23 curator tests and upstream's 29 incident_context tests pass; full per-directory discovery re-run after rebasing onto current main
- Red/green on the fix verified by a reviewer who rebuilt the pre-fix layout: the new test fails on the old code, passes on the new
- `make docs-check` green

### Live validation

Not live-tested: the change is unit tests plus a three-line guard fix in a script whose protected-branch refusal path is exactly what the new test exercises. The guard cannot be safely demonstrated against a real installation without asking a live agent to push to a protected branch on a real repository, which is the event the guard exists to prevent; the red/green run against the pre-fix code is the evidence the mechanism works.

## Self-Review

Adversarial pass run by a separate reviewer context that did not write the change, following `.agents/skills/review-adversarial/SKILL.md` (angles A–J), plus a docs-drift pass. Findings and dispositions:

- **Should-fix, fixed:** the harness docstring claimed the tests could not live beside the scripts because the skill's `scripts/` directory is outside `PYTHON_TEST_DIRS` — false, the `agents/*/skills/*/scripts/test_*.py` glob has covered it all along. Rewritten to say the placement is a choice, not a constraint
- **Nit, fixed:** the Makefile comment "none of them are packages" went stale; corrected
- The fix was traced adversarially: `heads/main` and `refs/remotes/origin/main` as push destinations cannot move a protected branch (they land elsewhere in the ref namespace), while `refs/heads/main` can and is now refused; a branch literally named with slashes that resolves to `refs/heads/main` is refused correctly since pushing it does move `main`. Nothing is weakened — the pre-existing comparison is preserved verbatim when no prefix is present
- Verified not-mirrors: removing the trace-point sort, mishandling counter resets, or flipping the TTL boundary from `<` to `<=` each breaks the corresponding test
- Docs-drift: `declarative-workflow.md`'s protected-branch claim stays true (and becomes more true); no doc enumerates the glob list; no map entry owed (no doc files changed)

## Risk & Rollout

Low. The only behaviour change is `submit_suggestion.py` refusing strictly more inputs than before — inputs that were always meant to be refused. Everything else is additive test code. Revert is a clean revert of two commits. Whoever merges second between this and #813/#565 should re-run `test_submit_suggestion.py`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,8 +109,9 @@ prettier-write: ## Reformat all Markdown/YAML in place.
 # glob, so they had never once run in CI. defaults/hooks is here for the same
 # reason -- the plugins glob does not reach it, so the chat_message_audit hook
 # was untestable-by-CI however many tests it grew. Discovery is then run once
-# per directory rather than once over the tree, because none of them are
-# packages -- `unittest discover` pointed at agents/platform/skills finds
+# per directory rather than once over the tree, because most of them are not
+# packages (incident_context is the exception, and per-directory discovery
+# still collects it) -- `unittest discover` pointed at agents/platform/skills finds
 # nothing and still exits 0, which reads as a passing suite. That also keeps
 # deploy/docker, deploy/docker/patches and each deploy/docker/plugins/<name>
 # separate, which they must be: those tests import their subject by bare module
--- a/agents/chat/scripts/test_memory_ttl_curator.py
+++ b/agents/chat/scripts/test_memory_ttl_curator.py
@@ -1,0 +1,421 @@
+"""Unit tests for the memory TTL curator's selection and safety logic.
+
+Run: python3 -m unittest agents/chat/scripts/test_memory_ttl_curator.py
+
+This script deletes memory unattended, so the tests lean on the refusal paths:
+what survives matters more than what goes. Every guard that stops a run —
+too-small bank, missing retain strategy, no observations, unscopable
+observation, partial distil — is exercised, and the happy path asserts exactly
+which rows were retired and why. Hindsight is a fake; what is tested is the
+decision logic, not HTTP.
+"""
+
+import contextlib
+import io
+import json
+import sys
+import unittest
+import urllib.error
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from unittest.mock import patch
+
+sys.path.insert(0, str(Path(__file__).parent.absolute()))
+
+import memory_ttl_curator as curator  # noqa: E402
+
+NOW = datetime(2026, 8, 19, 12, 0, 0, tzinfo=timezone.utc)
+TTL_DAYS = 180
+BANK = "kube-agents-memory"
+
+
+def days_ago(days):
+    return (NOW - timedelta(days=days)).isoformat()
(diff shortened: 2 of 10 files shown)


bradhoekstra:
Nit for the future: submit_suggestion.py  change should have been a separate PR.

bradhoekstra (approved):

PR #827, opened 2026-08-20
ci: post the coverage comment from a workflow_run job so fork PRs get it

jayantid:
## Summary

Moves the coverage comment out of the measuring job and into a new `workflow_run` workflow that runs in the base repository's context, where the token can write. This is what makes the sticky coverage comment appear on fork pull requests — which, since this repository requires PR branches on forks, is all of them. The measuring job in `python-tests.yml` now hands off through the artifact (report, PR number, repository total), and the new `coverage-comment.yml` validates the hand-off and posts one sticky comment, rewritten in place on every push.

## Why This Change

#819's coverage job posts its comment only for same-repo branches, because GitHub forces a fork PR's token to read-only regardless of the `permissions:` block. On its own PR — a fork PR — the comment never appeared and the report landed as an artifact nobody opens. A report-only gate that nobody sees is not a gate; the comment is the product.

## What Changed

- `.github/workflows/coverage-comment.yml` — new. Triggered by the Python Unit Tests workflow completing on a `pull_request` run. Downloads the `diff-cover-report` artifact via plain `gh api` (no new action pins), validates it, renders the comment, PATCHes the existing marker comment or POSTs a new one
- `.github/workflows/python-tests.yml` — the in-job comment step is removed (one posting path for all PRs), `pull-requests: write` dropped from the coverage job, and a hand-off step writes `pr-number.txt` and `coverage-total.txt` into the artifact
- The comment format: the PR's changed-lines coverage leads with a 🟢/🟡/🔴 status, repository total second, uncovered lines collapsed behind `<details>`, and a clean PR gets a one-line comment

**Security posture, because this workflow runs privileged against attacker-controlled input** (a fork PR author controls the artifact bytes and the producing workflow): it never checks out or executes anything from the PR; every `${{ }}` lives in `env:`/`if:` blocks, never `run:` (event fields are read from `$GITHUB_EVENT_PATH` with jq); the claimed PR number is digits-only sanitized and then validated against the run's own head sha + branch + owner as GitHub asserts them, so it cannot be pointed at someone else's PR; artifact text is whitelisted before it reaches the comment (a total that isn't `^\d+(\.\d+)?$` renders as `?`, table cells are character-restricted, so @-mentions, pipes, and markup cannot survive); symlinks in the artifact are deleted before any read; and the job carries the repository fork guard and a queued concurrency group.

## Context

Stacked on #819 — the first two commits here are that PR; review only the top commit (`be17d83`). Merge #819 first, then this rebases to a one-commit diff. Part of the testing-strategy implementation work.

## Testing

- zizmor 1.29.0 run offline against both workflow files: the new file's 3 High findings (inherent to `workflow_run` + `$GITHUB_ENV`) are suppressed inline with justifications per repo convention (`auto_request_review.yml` pattern); result "No findings to report". `python-tests.yml` carries one pre-existing medium, byte-identical on the base branch
- The embedded renderer was extracted and run against benign fixtures (75% table render, clean-PR one-liner) and hostile ones (a total of `100 — cc @victim`, table rows carrying pipes, backticks, an `` tag and a team mention, a `99999%` coverage line) — hostile output is inert, verified byte-for-byte
- Artifact failure paths reproduced locally: missing files, expired-artifact 410, duplicate zip entries, symlink entries — all skip gracefully (`exit 0` with a log line), none red the run
- Both YAMLs parse; `bash -n` clean on the shell steps

### Live validation

Not live-tested, structurally: a `workflow_run` workflow only executes from the default branch, so this cannot run until merged — its own PR will not produce a comment. First live proof is the next PR opened (or pushed to) after merge; if the comment does not appear there, revert is a one-file delete plus restoring the artifact-only path.

## Self-Review

An independent security-focused review (a context that did not write the change) ran the adversarial pass with a workflow_run threat model: content injection into the privileged comment, confused-deputy against other PRs, token scope, artifact zip attacks, and the repo's zizmor gate run for real. Verdict was request-changes; every finding was fixed before opening:

- **Blocking, fixed:** the file tripped the mandatory zizmor gate (3 High) — suppressions with justifications added, gate re-run clean
- **Fixed:** artifact text reached the comment verbatim, letting a fork author make the bot @-mention teams — all interpolated fields now whitelisted; a symlink named as a data file could publish runner files into the comment — symlinks deleted post-unzip and the total regex-validated; missing repository fork guard added; missing concurrency group added (queued, so the last push always wins); a footer referenced a doc not in the tree — replaced; renderer now fails soft on malformed/missing artifact files
- Attacks attempted that failed by design (kept, documented in the review): cross-PR comment planting fails because validation compares GitHub-asserted head tuples, not artifact claims; digits-only sanitization leaves no quoting path into jq or `gh api`; no other workflow uses a marker comment, so no forgery target
- Known limitation, deliberate: on fork PRs the coverage numbers themselves come from a workflow the PR author can edit. Acceptable while report-only; the future blocking gates must compute coverage somewhere the PR cannot edit, and the rollout plan now records that

## Risk & Rollout

The workflow runs with `pull-requests: write` against untrusted input — that is the risk, and the mitigations above are the answer. Blast radius of a bypass is comment content under the bot's name; no secrets are reachable (none referenced, no checkout, token scoped to comments and artifact reads). Failure mode of the workflow itself is a missing comment, never a red check on the PR. Revert: delete `coverage-comment.yml`, restore the artifact-upload-only path in `python-tests.yml`.


jayantid:
**Preview of the comment this workflow posts** — rendered with this branch's actual renderer (extracted from `coverage-comment.yml`), not a mock. Case 1 uses the real `diff-cover-report` artifact from #819's coverage job run; case 2 the same renderer on a report with uncovered lines. Posting a live one on this PR isn't possible before merge: `workflow_run` workflows only execute from the default branch.

### Case 1 — real artifact from #819 (no measurable Python lines in that diff)

## 📊 Coverage

**This PR's changed lines: no measurable lines changed 🟢** · Repository (Python): 61.8%

Report-only · nothing here blocks this pull request yet.

### Case 2 — a PR with uncovered lines

## 📊 Coverage

| | Coverage | Target |
|---|---|---|
| **This PR's changed lines** | **75%** 🟡 | 80% |
| Repository (Python) | 61.8% | — |

<details><summary>❌ 162 uncovered lines added (click to expand)</summary>

| File | Missing lines |
|---|---|
| `scripts/request.py` | 336-338, 341, 353-355, 362-370 |
| `scripts/generate_docs.py` | 366-373, 377, 379-390 |

</details>

Report-only · nothing here blocks this pull request yet.

.github/workflows/coverage-comment.yml
```diff
@@ -0,0 +1,242 @@
...
+          fi
+          gh api "repos/$REPO/actions/artifacts/$aid/zip" > artifact.zip \
+            || { echo "artifact download failed (expired?); not posting."; exit 0; }
+          mkdir report
+          # -j flattens paths so a crafted entry cannot traverse out of the
+          # directory, and any symlink that survives extraction is deleted
+          # before anything reads the tree.
+          unzip -qj artifact.zip -d report \
+            || { echo "artifact unusable; not posting."; exit 0; }
+          find report -type l -delete
```
toshiowang: **Security risk (definitive control bypass, low yield):** `exit 0` runs before `find report -type l -delete`, and a step exiting 0 does not stop later steps — it only ends this one.

Info-ZIP returns 1 on a duplicate-name collision (EOF on the overwrite prompt is treated as `[N]one`) **after** extracting earlier entries. So an artifact whose entries are `diff-cover.md` (a symlink), `pr-number.txt`, then a duplicate `pr-number.txt` leaves the symlink on disk, skips the deletion, and the next steps read through it. Reproduced locally:

```
$ unzip -qj dup.zip -d out2 </dev/null
replace out2/pr-number.txt? ... (EOF or read error, treating as "[N]one" ...)
rc=1
$ ls -l out2
lrwxrwxrwx  diff-cover.md -> /etc/hosts
```

The PR body states "symlinks in the artifact are deleted before any read"; that does not hold on this path. The downstream whitelists contain the damage — only strings matching the row regex or `^\d+(\.\d+)?$` reach the comment — so this is a bypassed control, not a demonstrated leak. Fix is to make the failure path leave nothing readable behind:

```suggestion
          unzip -qj artifact.zip -d report \
            || { rm -rf report; echo "artifact unusable; not posting."; exit 0; }
          find report -type l -delete
```

.github/workflows/coverage-comment.yml
```diff
@@ -0,0 +1,242 @@
...
+          REPO: ${{ github.repository }}
+        run: |
+          [ -f comment.md ] || { echo "no comment rendered; nothing to post."; exit 0; }
+          marker='<!-- coverage-report -->'
+          # Only comments this workflow's identity wrote: anyone can start a
+          # comment with the marker, and PATCHing theirs would render the
+          # report under their name.
+          cid=$(gh api "repos/$REPO/issues/$PR_NUMBER/comments" --paginate \
+                --jq ".[] | select(.user.login == \"github-actions[bot]\")
+                          | select(.body | startswith(\"$marker\")) | .id" | head -1)
```
toshiowang: **Critical bug (tentative — depends on a transient API failure, but the artifact it leaves is permanent):** the default shell for `run:` is `bash -e {0}` — `-e` but no `pipefail` — so if `gh api` fails (5xx, secondary rate limit) the pipeline's status is `head`'s (0), `cid` comes back empty, and the script falls through to the POST branch. That posts a *second* sticky comment; from then on `head -1` pins the older one, so the duplicate is never updated and never removed.

`pipefail` is not the fix here: `head -1` closes the pipe early and would SIGPIPE `--paginate` on the happy path. Take the `gh` call out of the pipeline instead so its exit status survives:

```suggestion
          comments=$(gh api "repos/$REPO/issues/$PR_NUMBER/comments" --paginate \
                     --jq ".[] | select(.user.login == \"github-actions[bot]\")
                               | select(.body | startswith(\"$marker\")) | .id") \
            || { echo "comment lookup failed; not posting (a POST would duplicate)."; exit 0; }
          cid=$(printf '%s\n' "$comments" | head -1)
```

Same root cause at `:59-61` (`aid=$(gh api ... | head -1)`), where an API error reads as "no artifact" and silently skips the comment. Lower impact, worth the same treatment.

.github/workflows/coverage-comment.yml
```diff
@@ -0,0 +1,242 @@
...
+
+          no_lines = "No lines with coverage information in this diff." in diff_md
+          pct = None
+          m = re.search(r"^- \*\*Coverage\*\*: (\d+)%", diff_md, re.M)
+          if m and 0 <= int(m.group(1)) <= 100:
+              pct = int(m.group(1))
+          missing = 0
+          m = re.search(r"^- \*\*Missing\*\*: (\d+) ", diff_md, re.M)
+          if m:
+              missing = int(m.group(1))
```
toshiowang: **Critical bug (definitive, attacker-only, self-contained):** `int()` on an unbounded `(\d+)` capture raises `ValueError: Exceeds the limit (4300 digits) for integer string conversion` under Python 3.11+. A fork-authored `diff-cover.md` containing `- **Coverage**: <5000 nines>%` — or the same in `- **Missing**:` — crashes the renderer with an uncaught exception. I extracted the embedded script and ran both cases; each exits 1, failing the step rather than falling soft. That contradicts "renderer now fails soft on malformed/missing artifact files" in the PR body.

Blast radius is only a red run in the base repo's Actions tab and a missing comment (a `workflow_run` job posts no check on the PR), so this is low severity — but the fix is free:

```suggestion
          pct = None
          m = re.search(r"^- \*\*Coverage\*\*: (\d{1,3})%", diff_md, re.M)
          if m and 0 <= int(m.group(1)) <= 100:
              pct = int(m.group(1))
          missing = 0
          m = re.search(r"^- \*\*Missing\*\*: (\d{1,9}) ", diff_md, re.M)
          if m:
              missing = int(m.group(1))
```

An over-long value then simply fails to match and renders as `**?**`, which is the intended degradation.

toshiowang (changes requested):
Reviewed the two new commits (`a46274f7`, `be17d83c`). Verified first against live API data: on a real fork run (#803, `gke-agentic/kube-agents`) `workflow_run.pull_requests` is empty and `head_sha`/`head_branch`/`head_repository.owner.login` match `pulls/N`'s `head.sha`/`head.ref`/`head.repo.owner.login` exactly, so the fork validation path is sound and is the one that will run. `run: | # zizmor: ignore[...]` parses as a YAML comment, the diff-cover regexes match the upstream `markdown_coverage_report.md` template, and `gh api -F body=@file` sends a string. Three findings, all in the new workflow.

The first is a bypass of a control the PR body claims: `exit 0` on unzip failure runs before `find report -type l -delete`, and Info-ZIP returns 1 on a duplicate-name collision *after* extracting earlier entries, so a crafted artifact leaves a symlink in place for later steps to read through. Reproduced locally. The whitelists downstream contain the damage, so this is a bypassed control rather than a demonstrated leak — but the fix is one word.

The second is error-swallowing in the sticky-comment lookup that produces a permanently duplicated comment. The third is an uncaught `ValueError` in the renderer on crafted input, which contradicts the "fails soft" claim.

One note, no action needed: the fork path requires an exact `head_sha` match, so a push landing while `make coverage` is still running makes that run's comment silently skip. That is self-correcting (the newer run posts) and is the right trade against stale comments — worth a line in the workflow comment rather than a change.

.github/workflows/coverage-comment.yml
jayantid: Fixed in 80c5d62, reproduced first with your exact zip (symlink diff-cover.md → /etc/passwd + duplicate pr-number.txt): the symlink did survive the old early-exit — and without stdin redirected the old step actually hung on the overwrite prompt, so the scenario was worse than reported. Now: the symlink sweep runs unconditionally immediately after unzip before any rc handling, a nonzero rc removes report/ entirely so no partial extraction survives, and unzip gets </dev/null. Post-fix repro: report/ gone, later steps skip with 'not posting'.

.github/workflows/coverage-comment.yml
jayantid: Fixed in 80c5d62, taking your suggestion: the paginated lookup runs alone into a variable so its exit status survives -e, then head -1 runs on the captured text (no SIGPIPE against --paginate). Repro with gh stubbed to 502 on the lookup: rc=1, zero write calls — the step reds rather than double-posting. Happy path re-verified: one PATCH, no POST.

.github/workflows/coverage-comment.yml
jayantid: Fixed in 80c5d62: both captures bounded to (\d{1,7}) and both int() conversions wrapped in the fail-soft path, belt and braces. Your two repro cases (5000 nines in Coverage and in Missing) now exit 0 and render '?' with no bogus count. You are right that the PR body's 'fails soft on malformed artifact files' claim was contradicted — it was written after the earlier hostile-fixture pass, and these two shapes weren't in that fixture set; they are now, alongside the dup.zip. Good catch on all three.

toshiowang (approved):

PR #792, opened 2026-08-18
docs(contributing): document make docs-check in local validation

mrusmanabdullah:
# Pull Request

## Why This Change

The **Local validation** section of the contributing guide lists the checks to run before pushing — Prettier, Validate Repo Structure, Docker Build, Operator Tests, and the `docs/site` build — but it never mentions the **Docs Check** CI job (`.github/workflows/docs-check.yml`).

That leaves a gap for the most common kind of contribution. Someone making a docs-only change follows the section, runs `npm run build`, sees it pass, pushes — and CI still fails on link resolution, terminology, the docs map, or stale generated regions. The docs build and the docs checks are unrelated gates, but the guide only names one of them.

## What Changed

**Files:**

- `docs/site/src/content/docs/contributing.md`

**Added/Changed/Fixed:**

- Added a **Docs checks** bullet to Local validation covering `make docs-check`, the root Makefile aggregate that runs the same four checks the Docs Check job runs as separate steps (`docs-check-generated`, `docs-check-links`, `docs-check-terminology`, `docs-check-map`).
- Noted that `make docs-generate` is the fix when the generated-regions step fails.

No content was moved or reworded; this is a single added bullet.

## Why This Matters

- Closes the one CI job that Local validation didn't represent, so the section is now complete rather than nearly complete.
- Catches docs-only failures locally in one command instead of on a CI round-trip.
- Points at the remedy for the generated-regions failure, which is the least self-evident of the four.

## Context

- The `Docs Check` workflow runs the four checks as separate steps; `make docs-check` (root `Makefile`) is the existing aggregate that already matched them. This PR only documents it — no Makefile or workflow changes.
- Sample PR opened to exercise the contribution flow. Happy to close it if it isn't wanted; the underlying gap is real either way.

## Testing

- `make docs-check` — passes (exit 0) with the change applied: 247 Markdown files link-checked, all relative links resolve, terminology check passed, docs map inventory covers all 218 tracked docs.
- Verified against the source of truth rather than the prose: confirmed `docs-check` and its four prerequisites exist in the root `Makefile`, and that `.github/workflows/docs-check.yml` runs those same four as separate steps.

**Live validation:** Not applicable — documentation-only change with no runtime, operator, or agent-configuration surface. Nothing is reconciled or deployed by this diff.

---

Zero functional impact: one added bullet in a contributing guide. Risk is limited to the docs site rebuild.

Issue #637, opened 2026-08-11
Allow 2 smoke tests to run at once by leasing a second eval project (Boskos)


jayantid:
## Why

The smoke test is capped at one run at a time (`max_concurrency: 1`), so PRs queue ~an hour. It's capped because every run installs and deletes cluster-wide things (CRDs, webhooks, ClusterRoles) in a single shared GCP project — two runs at once would break each other.

**The fix:** get a second identical project. Each run gets a whole project to itself, so nothing is shared and nothing collides. A service called Boskos (already running on our Prow) hands out the projects.

Nothing about how the test works needs to change.

> **What's Boskos?** A checkout desk for shared test resources, already running on our Prow ([`kubernetes-sigs/boskos`](https://github.com/kubernetes-sigs/boskos)). Jobs ask it for a project; it hands out exactly one at a time and takes it back when they're done, so two runs can never grab the same one. It also reclaims a project automatically if a job gets killed — something our current setup doesn't do. The `gcs-fuse-csi-driver` job already uses it, so copy that pattern.

## Order of work

Steps must be done in order. Steps 1 and 2 are in other places; step 3 is this repo; step 4 is oss-test-infra.

### Step 1 — Create the second project

- [ ] Create GCP project `kube-agents-evals-2` (name is a suggestion; confirm with the team)
- [ ] Request L4 GPU quota in `us-west4-a` to match project 1
- [ ] Copy the Gemini API key secret into it
- [ ] Run the existing Terraform against it:

```bash
cd terraform/examples/full-install
terraform apply -var="project_id=kube-agents-evals-2"
```

Use the **same** `namespace` and `ksa_name` values as project 1 — the Workload Identity binding is tied to the namespace name, so they must match.

**Verify:** you can run `gcloud container clusters get-credentials platform-agent-host --region us-central1 --project kube-agents-evals-2` and `kubectl get ns` works.

### Step 2 — Register the projects with Boskos

Boskos lives in oss-test-infra and already hands out projects to another job (`gcs-fuse-csi-driver`) — copy that pattern.

- [ ] Add a resource type `kube-agents-eval-project` listing both project IDs
- [ ] **Turn the janitor OFF for this type** (see Gotchas)

**Verify:** the Boskos config shows two resources of the new type in state `free`.

### Step 3 — Make this repo accept the project as input

[...]

PR #859, opened 2026-08-21
feat(scripts): lease a shared live-test install between concurrent agents

bradhoekstra:
## Summary

Several agents share one live-test installation, and nothing stops two of them mutating it at the same time. This adds `scripts/live_test_lease.py`, which holds a lease as a `live-test-lease` ConfigMap in the install's own namespace, and a `PreToolUse` hook that takes it automatically: a mutating Bash command claims the lease when it is free and is denied while another agent holds it, while read-only commands are never touched. The protected install is discovered from the checkout's `k8s-operator/scripts/vars.sh`, so a clone with no install configured protects nothing and the hook exits before any cluster round-trip. The wiring ships as `.claude/settings.json.example` and is copied into place per checkout, because a tracked `settings.json` is branch content Claude Code executes without being asked. `docs/designs/live-test-lease.md` is the design of record; `AGENTS.md`, the contributing page, and the PR template gain one pointer each.

## Why This Change

`AGENTS.md` requires every pull request to be exercised against a real running installation, and the contributing page answers "I don't have one" with INSTALL.md. Neither answers the case that actually happens: one install, several agents. Standing up a GKE cluster per contributor is not realistic, so a team shares one — and then one agent patches the `PlatformAgent` env while another rolls a new image tag into the same Deployment. Each reads back a pod carrying both changes or neither, and writes a **Live validation** section describing an install that no longer exists. That has already happened on the install this repository is developed against: one agent's image tag was lost to another's push, and nothing in the toolchain noticed — the unit tests passed, `make docs-check` passed, and the failure surfaced later as a change that never worked.

Without this, the only defence is an instruction in `AGENTS.md` telling agents to coordinate, which holds only while every participant remembers, hundreds of lines away from the command they are about to run.

## What Changed

- `scripts/live_test_lease.py` — the CLI (`status`, `acquire`, `renew`, `release`, `steal`) and the two hook entry points. Acquisition is a compare-and-swap on the ConfigMap's `resourceVersion`, so two agents racing an expired lease cannot both believe they won; leases expire after 60 minutes and renew lazily past the halfway mark. `SessionEnd` and `release --all` release what the session holds, read off the local token files rather than off what the current directory can discover.
- `scripts/test_live_test_lease.py` — 122 tests, stdlib only, no cluster. The classifier carries most of them, because a mutation it wrongly calls read-only is the exact failure the lease exists to stop.
- `.claude/settings.json.example` (plus the `.gitignore` negation that allows it) — the hook wiring, copied to `.claude/settings.json` to opt in, and a no-op when no install is configured.
- `scripts/live_test_envs.example.json` — placeholders for an install you protect but have no checkout of.
- `docs/designs/live-test-lease.md`, with one pointer each from `AGENTS.md`, `docs/site/src/content/docs/contributing.md`, `.github/PULL_REQUEST_TEMPLATE.md`, and one inventory row in `docs/README.md`. `k8s-operator/scripts/README.md` gains the new `vars.sh` reader.

## Context

Nobody else is doing this. Adjacent but different: issue #637 (assigned) leases a second GCP project so two CI smoke tests can run at once — that is a project pool for CI, not a contributor-side lease on one install. `docs/designs/gitops-workspace-leases.md` is the same idea one layer down, leasing GitOps clones inside the agent pod. Merge-conflict warnings only, no functional overlap: #792 (draft) edits `contributing.md`, #720 edits `AGENTS.md`, and #848/#846/#811/#721 each add a row to `docs/README.md`.

## Testing

- `python3 -m unittest discover -s scripts` — 252 tests pass, 123 of them new.
- `make docs-check` — generated regions, links, terminology, map coverage.
- `prettier --check` on the changed Markdown and JSON.

### Live validation

Exercised end to end against the shared GKE install this repository is developed on (regional `us-central1`, namespace `kubeagents-system`), driving the hook through its real stdin contract rather than a stub.

- **Discovery.** Run from a checkout with no `vars.sh` and no config file, `status` reports no protected install and the hook exits 0 with no output — a mutating `kubectl delete ns kubeagents-system` passes straight through. Run from a nested directory of a checkout that has one, it derives the install from `PROJECT_ID`/`CLUSTER_NAME`/`REGION` and reports `FREE` against the live cluster.
- **The record.** `acquire --pr 999 --note "…"` created the ConfigMap; `kubectl get configmap live-test-lease -o jsonpath='{.data}'` returned exactly the fields the design doc claims — `token`, `session`, `holder`, `cwd`, `branch`, `pr`, `note`, `acquiredAt`, `ttlMinutes`, `expiresAt`.
- **The mechanism, not a coincidence.** With session A holding it, session B (a different `KUBE_AGENTS_LEASE_SESSION`, same cluster, same namespace) was **denied** `kubectl annotate configmap live-test-lease probe=1`, under a message naming the holder, their branch, and the expiry. In the same session and against the same install, `kubectl get pods` and `grep -rn "kubectl apply" docs/` both passed silently — opposite outcomes by verb, not by target. Session A, the holder, was allowed the identical `annotate`.
- **Auto-acquire and release.** From a free install, a session's first mutating command claimed the lease without asking (the ConfigMap appeared, carrying that session's id) and `hook-sessionend` removed it. `release` likewise left `configmaps "live-test-lease" not found`.
- **Never fail open.** With the target resolved but the cluster unreachable, the hook returned `ask` — "Could not check the live-test lease … Another agent may be mid-change" — rather than allowing or blocking.
- **`steal`.** Refused a live lease and told the caller to coordinate; took it over with `--force`; the previous holder's `release` was then refused as "not yours to release".
- **The classifier, after every round of review fixes.** Re-run against the same install with the lease held by a different session key, so each verdict is an observable deny, ask, or silent allow rather than something inferred, and every payload run through both the current head and the commit the reviewer read. Denied on the head, silent on the reviewed commit: `eval kubectl --context <install> delete ns …` and its quoted form, `f() { kubectl --context <install> delete ns …; }; f` and `function f { … }`, `OUT=$(kubectl delete ns kubeagents-system 2>&1)`, `LOG=$(./upgrade.sh 2>&1 | tail -20)`, `R=$(make deploy || true)`, `cat <<EOF | kubectl apply -f -`, `grep -q ready <<<yes && kubectl delete ns kubeagents-system`, `kubectl --context <alias> delete ns …` with the alias declared in a config entry for the install the checkout already discovers, and `f () { kubectl … delete ns …; }; f` and `function f () { … }` — the spelling with a space between the name and the parens, which shlex splits differently from the one 1df9afe fixed. **Ask** on the head, silent on the reviewed commit: `docker push $(cat last-image)`, `source .env && docker push "$IMG"`, `docker push us-central1-docker.pkg.dev/$PROJECT_ID/platform:dev`, and the `docker buildx build --push -t …/${PROJECT_ID}/…` form of it. Opposite outcomes on the same install, silent on both: `kubectl get pods -n kubeagents-system 2>&1 | head`, `eval kubectl --context <install> get pods -n kubeagents-system`, `OUT=$(kubectl get pods 2>&1)`, `docker push ghcr.io/someone/platform:$(git rev-parse HEAD)` and `docker push us-central1-docker.pkg.dev/some-other-project/x:$TAG` — a ref whose literal half is not on the path to a protected registry has answered the question, and prompting on it is the false positive that costs an agent an hour — and a `gh pr create` carrying a `kubectl patch` inside a heredoc **body**. Unchanged from earlier rounds: a cleanup loop (`for ns in …; do kubectl --context <install> delete ns $ns; done`) reaches the lease check, `kubectl rollout status` does not, and `docker push <the install's registry>` is denied. Every deny in this round was read off the shared install's *real* lease, held by another agent's session at the time (`fix/agent-response-verbosity`, 52 minutes to run), so nothing was written to the cluster to stage them: `status --all` reported the same holder and claim time before and after, and no `live-test-lease` ConfigMap, token file, or scratch namespace was left behind.
- **The `case`-arm gap is real and deliberate.** `case $E in prod) kubectl --context <install> delete ns …;; esac` passes silently on the head as well as on the reviewed commit. It is documented in the design doc's Limits, and a test asserts the pass and the paragraph together so the gap cannot become undocumented.
- **A renamed context, on a copy of the kubeconfig** — the real one was not touched. With the canonical name renamed away, an entry keyed on the canonical context with the new name in `aliases` reports `unreachable: error: context "gke_…" does not exist`, which is the hook asking about every mutating command indefinitely; an entry keyed on the renamed name resolves and leases, but as a second install alongside the one the token record recovers. Renaming the canonical name back — what `gcloud container clusters get-credentials` does — restores `HELD` against the real lease. That is the remedy Limits now gives; the split-install half of it is pinned by a unit test rather than observed here, since this checkout has no `vars.sh`.
- **Releasing what the session holds, not what the directory sees.** From `/tmp`, which discovers no install, with a lease this session had taken on the shared cluster: the reviewed commit's `status` printed "No protected install found" and its `hook-sessionend` and `release --all` left the ConfigMap standing; the head recovered the install from the token file and released it. Run against a scratch namespace on the same cluster, and the namespace is the point — the token record carries it, so the delete landed in `default` while the shared `kubeagents-system` lease was left untouched, which a reconstruction defaulting the namespace would have got wrong in the direction that reports success and deletes nothing. The pre-existing token file on this machine is one of the bare-hex ones the fallback is for; `status`, `renew`, and the reconstruction all read it.
- **The lease read, and what counts as free.** `status` returns `FREE` off a real `kubectl get` whose stderr is `configmaps "live-test-lease" not found` — now the only message accepted as an unheld lease — and `HELD … <-- yours` against a record this session holds. With `kubectl` off `PATH` it reports `unreachable: kubectl not found on PATH`; the reviewed commit reported that same cluster `FREE`.
- **`SessionEnd` by reason.** Against a lease the tool acquired for the test, `hook-sessionend` fed `{"reason": "clear"}` left the ConfigMap and the local token untouched; `{"reason": "prompt_input_exit"}` deleted both. Run against a scratch namespace on the same cluster so the shared install's own lease was not disturbed.
- **The opt-in copy.** `cp .claude/settings.json.example .claude/settings.json` produces a file `git check-ignore` attributes to `.gitignore:62` and `git status` does not list. Driving both events through the command strings read out of that copy — shlex-split, with `$CLAUDE_PROJECT_DIR` expanded as Claude Code expands it, rather than a paraphrase — a read exits 0 silently, `OUT=$(kubectl apply -f x.yaml 2>&1)` is denied, and `SessionEnd` with `reason: clear` exits 0 without touching the lease. The copy was removed afterwards.
- **Not covered.** Claude Code loading that settings file itself, as opposed to its command strings being executed with the same stdin payloads — the copy postdates the running session, and this machine's own separate hook would collide with it. That also leaves the `SessionEnd` timeout unobserved: the 1.5 s default budget it works around was read out of the 2.1.238 binary and confirmed by timing the entry point (2.41 s on its cheapest path), not by watching a real session shut down. The second install in `live-test-envs.json` (a different project) was not reachable from the kubeconfig in use, so multi-install `--env` selection was exercised only in unit tests.
- **Cleanup.** No `live-test-lease` ConfigMap remains in `default` or `kubeagents-system` beyond the one this session legitimately holds; the probe token files and the copied `.claude/settings.json` were removed; the scratch config files under `/tmp` are gone. No cluster state beyond those ConfigMaps was written.

## Self-Review

Both required passes ran in subagents that did not write the change, each handed the diff range and nothing else. Findings are merged across rounds; the last adversarial pass re-derived every one of its findings against the current working tree, and I reproduced each of the twelve myself before acting on it.

**`review-adversarial`.** Angles: adversarial shell input against the classifier, the hook's own failure modes and which way each fails, concurrency on the ConfigMap, secret handling in `vars.sh`, what committing a hook exposes, whether the tests assert what their names claim, docs-vs-code agreement, and sibling-PR overlap. Not covered: the pass had no live install, so the CAS and the deny rendering were read rather than observed — the Live validation section above is where those were exercised.

Fixed, each with a regression test:

- **`SessionEnd` declared no `timeout`.** Claude Code budgets that whole phase at 1.5 s unless a hook asks for more — not the 60 s a `PreToolUse` hook gets (`getSessionEndHookTimeoutMs` in 2.1.238). One `kubectl get` against GKE measures 2–3 s, so every session would have been killed before the delete and stranded its lease for the full hour, silently. Now declares 45 s, and a test asserts both hook budgets against `KUBECTL_TIMEOUT`.
- **Shell keywords hid the command word.** `for ns in …; do kubectl delete …; done`, `if … then …`, `{ … }`, and `( … )` all classified as nothing — the bulk-cleanup loop is about the most ordinary mutation an agent writes.
- **`--context "$CTX"` read as a definite "unprotected cluster".** shlex does no expansion, so the literal matched no install and short-circuited both the markers and the ambient probe. An unexpanded value is now a missing answer, not a negative one.
- **Heredoc bodies and quoted prose classified as commands.** This repository requires a **Live validation** section naming what was run, so its own `gh pr create` carries a `kubectl patch` on a line of its own — taking an hour-long lease, or being denied one, for a command that touches nothing.
- **`docker push` matched the registry against the segment, not the line**, so `IMG=<ref> && docker push $IMG` escaped — overwriting the tag a shared install is running is the incident this exists to prevent.
- **`cd <other-checkout> && ./upgrade.sh` took the wrong install's lease.** The `cd` was parsed, but the other checkout's install was never added to the protected set, so the installer fell back to the ambient context. The existing test passed only because it wrote the *same* install into the temp checkout; it now writes a second one.
- **False positives on reads:** `kubectl rollout status`/`history`, any `--dry-run` short of `=none`, `timeout -s KILL 30 kubectl delete …` (the flag's value eaten as the duration), and `./install.sh --note help` (a no-op flag matched in value position; `--version`, which no installer accepts, is dropped).
- **Earlier rounds:** `helm --kube-context` was not read as a context and resolved to the ambient install; the context probe ran once per segment at 6 s each and could overrun the `PreToolUse` budget, which fails open because a killed hook is a hook error; `agentplugins/*/install.sh` shares a basename with the root installers but follows the context rather than `vars.sh`.

Reported, not fixed:

- **`do_release` deletes without a `resourceVersion` precondition**, unlike acquire, renew, and steal. Real, but the token comparison immediately precedes the delete, so the exposure is one kubectl round trip on a lease that has already expired and been taken over — and closing it needs a precondition `kubectl delete` does not offer. Left as is rather than restructured around a `replace`, which would leave the ConfigMap behind and contradict what `release` promises.
- **A committed hook is branch content**, so a fork's branch can run its own code on a reviewer's machine once the checkout is trusted — including a worktree of a trusted checkout, which is how `pr-review-batch` reviews. Documented in Limits at the time and left there; the bot raised it again as its own first finding, and it is now fixed rather than documented — see below.
- **A single unattended run longer than the TTL** — `install.sh` against a fresh cluster — expires mid-run, because the lease renews only when another Bash call passes through the hook. In Limits, with `acquire --ttl` as the answer.

**`kube-agents-bot`, first pass.** Five findings, every one reproduced before being acted on and all fixed in e21e715, each with a regression test that fails on the commit the bot read:

- **`strip_heredocs` cut from the introducer's line start**, so `cat <<EOF | kubectl apply -f -` — one of the two ways an inline manifest is written — classified as `cat`. Only the body is data; what follows the `<<EOF` on that line is the command. The bodyless case stopped discarding the tail too.
- **The heredoc pattern matched one character into `<<<`**, where `<<` is followed by a word that looks like a delimiter, swallowing the rest of the here-string's line and whatever the `&&` was guarding.
- **`get_lease` read any stderr containing "not found" as an unheld lease** — a missing namespace, an absent auth plugin, and this module's own `kubectl not found on PATH` all say it. It now matches the ConfigMap's own absence; everything else is unreachable, which is the position the design doc already took.
- **`SessionEnd` ignored the payload's `reason`.** `/clear` and `/resume` fire it on a process that carries on under the same holder key, so the release handed a shared install to the next agent mid-test.
- **`resolve_installs` dropped a config entry whose context was already discovered** from `vars.sh`, `aliases` and all — so the documented remedy for a renamed context did nothing for the install most likely to need it. Fixed by merging the aliases into the discovered install rather than making `kubectl()` probe for a reachable name; the design doc now says the entry's `context` must be the name the kubeconfig answers to.

The first two are regressions introduced by the adversarial pass's own heredoc fix, and both fail open on the exact idiom the lease exists to catch. Every heredoc test written for that pass put the introducer at end of line, which is why the suite did not catch either. The bot's summary notes the diff was too large to read exhaustively, so treat those five as a sample.

**`kube-agents-bot`, second pass** (`/review all` over e21e715). Three findings, all reproduced first, all fixed in 6fcf632 with regression tests that fail on e21e715:

- **The committed `.claude/settings.json` is branch content Claude Code runs unprompted** — the finding the adversarial pass had reported and I had answered with documentation. The bot was right that documentation is not the fix: `pr-review-batch` checks fork branches out into a worktree of a trusted root, and I confirmed against the 2.1.238 binary that there is no hook-change re-approval gate, only a global `disableAllHooks`. The file is now `.claude/settings.json.example`, copied into place to opt in. That loses enforcement-by-default, which is a real cost and is now stated in Limits rather than glossed.
- **`SessionEnd` released the installs the exit directory could discover**, which is not the set the session claimed. A lease taken after `cd ../other-checkout` stood for the rest of its hour. Both it and `release --all` now enumerate the local token files, which carry the context, namespace, and kubeconfig the lease was taken against; `status` reports a held install the checkout cannot see, so `release --all` can no longer reach one it never showed. A bare-token file from an older build still reads, and the token file is now written 0600.
- **`split_segments` cut through `$(...)`**, so `OUT=$(kubectl delete ns x 2>&1)` classified as nothing — the leading fragment is an assignment, whose command word `strip_wrappers` discards, and the opening paren left with it so the substitution could not be found either. Capturing output is how this repository's scripts run installers. A redirection's `&` no longer splits either; that half is shell parity rather than a demonstrated bypass, since the command word survives at the front of the bad split, and the test says so.

**jayantid, human review of 6fcf632.** Three findings, each reproduced against 6fcf632 before anything changed and fixed in 1df9afe with regression tests that fail there:

- **`eval` and function definitions hid the command word.** `eval kubectl … delete ns …` classified as nothing two ways: bare, `eval` was read as the command itself; quoted, shlex collapsed the payload into one argument so there was no command word left to find. `shell_payloads` now joins `eval`'s arguments back and recurses, covering both. Separately, `f() { kubectl … delete …; }; f` stopped one token early — `{` was already a stripped keyword, the `f()` in front of it was not — so the body was never read. All three spellings bash accepts (`f() {`, `function f() {`, `function f {`) are handled.
- **`docker push` matched only a literal registry substring**, so every ref arriving through a variable or a substitution read as a push somewhere else — with no lease taken and no deny, which is the incident this tool exists to prevent. An unreadable registry is now `UNKNOWN` and the hook asks; a ref that says plainly where it is going, `ghcr.io/x/y:$(git rev-parse HEAD)`, does not, because prompting on it would be a false positive on somebody's own registry. Silent when no protected install records a registry at all. That first cut judged the host alone, which the next round overturned — see below.
- **The Limits remedy for a locally renamed context did not produce a working lease** either way round, which I confirmed on a copy of a kubeconfig. Aliases widen classification only — every cluster call goes through `--context install.context` — so an entry keyed on the canonical name leases against a context the rename removed and answers `Unreachable` forever, while one keyed on the renamed name yields two installs for one cluster. Limits now says the remedy restores classification and that `gcloud container clusters get-credentials` is what restores the lease. Two tests pin both halves.

Two of the eight new tests — the ones asserting a computed tag on an unprotected registry stays silent, and that an install recording no registry is never asked about — pass on 6fcf632 by construction: they bound the new behaviour rather than demonstrate the old bug. The other six fail there.

**jayantid, human review of 1df9afe.** Two findings, both reproduced against 1df9afe before anything changed and fixed in 4077aa0 with regression tests that fail there:

- **`unreadable_registry` judged the ref's host, and a recorded prefix is host *and* project.** `us-central1-docker.pkg.dev/acme-prod` has two segments, so `docker push us-central1-docker.pkg.dev/$PROJECT_ID/platform:dev` matched neither guard: the host expanded fine, and the literal substring the line would have had to contain is not there. It classified as nothing. That is the motivating incident in the spelling most likely to produce it — `vars.sh` exports `PROJECT_ID`, so an agent that sourced it writes exactly this ref. The question is now asked of the whole literal half of the ref, everything before the first token that did not expand, and answered "unreadable" when that is still on the path to a protected registry. `ghcr.io/x/y:$TAG` and `us-central1-docker.pkg.dev/other-proj/x:$TAG` are still answers and still silent.
- **A function definition written `f () {` went unread.** POSIX allows the space, and shlex then hands back `f` and `()` as separate tokens, so `strip_wrappers` stopped on `f` — the same fail-open class as the `f() {` form 1df9afe had just fixed, in the spelling that round did not cover. It now peeks one token ahead, and a bare `()` joins the stripped keywords so `function f () {` is covered too.

Three of the eight new payloads in those tests pass on 1df9afe by construction — the two refs that expand inside a registry the recorded prefixes do not reach, and the `$REGION-docker.pkg.dev/acme-prod/…` one the host check already caught. They bound the widened rule rather than demonstrate the old bug; the other five fail there.

Reported, not fixed:

- **A `case` arm still hides the command word behind its pattern label** — `case $E in prod) kubectl … delete …;; esac` passes silently. Finding it means widening "skip tokens until one looks like a command", which is what lets a real argument be read as one, and a false mutation costs an agent the cluster for an hour. The gap is in the design doc's Limits and a test asserts the silent pass and the paragraph together, so it cannot become undocumented.

The fix rounds were not put through a fresh subagent pass — the two skill passes above ran against 0fe7532, and jayantid's two reads covered 6fcf632 and 1df9afe. A `/review` on the current head is what covers 4077aa0.

**`review-docs-drift`.** Six prose-vs-source findings, all fixed: the installer always records `REGISTRY_PREFIX`, so the derived Artifact Registry prefix is a fallback rather than the rule, and a stock install records the public default; nothing ever writes `NAMESPACE` to `vars.sh`; `--env` is never required for `status`; `steal` keys on expiry alone, so a dead session's lease is refused for up to an hour without `--force`; a local `.claude/settings.json` is not overwritten by a pull (moot as of 6fcf632 — the file is no longer tracked); and plugin installers resolve through the context, not the checkout.

## Risk & Rollout

Contributor tooling only: nothing here is built into an image, rendered by a chart, or reconciled by the operator, and the ConfigMap lives in the install's namespace rather than in any manifest. The blast radius is a contributor's Bash tool calls in a checkout of this repository. The failure mode to weigh is a false positive — a read the classifier calls a mutation takes an hour-long lease, or is denied while somebody else holds one — which is why classification is by command word and why the tests are weighted towards commands that only mention a mutation.

Revert is `git revert`; a stale ConfigMap left behind expires after 60 minutes on its own, or `kubectl delete configmap live-test-lease -n <ns>` removes it. A contributor who wants no part of it never copies `.claude/settings.json.example` into place, and a contributor who does but keeps no `vars.sh` in their checkout gets a hook that already does nothing.


scripts/live_test_lease.py
```diff
@@ -0,0 +1,1511 @@
...
+        if nl == -1:
+            # An introducer with no body: the payload is somewhere we cannot
+            # see, so drop the rest of the line rather than read it as one.
+            return command[:m.start()]
+        rest = command[nl + 1:]
+        term = re.search(r"^[ \t]*%s[ \t]*$" % re.escape(m.group(2)), rest,
+                         re.MULTILINE)
+        end = nl + 1 + (term.end() if term else len(rest))
+        command = command[:m.start()] + command[nl:nl + 1] + command[end:]
+
```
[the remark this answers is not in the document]
bradhoekstra: Fixed in e21e715, and reproduced first: `strip_heredocs` returned `'cat \n'` for `cat <<EOF | kubectl apply -f -\n…\nEOF`, and the line classified to `(None, None)`.

The splice now keeps the introducer's line and cuts only the body — `command[:m.start()] + command[m.end():nl + 1] + command[end:]`. The `nl == -1` branch changed too: an introducer whose body is not in the string leaves everything after it as command text rather than dropping the tail, since that is the more conservative reading.

`test_a_mutation_on_the_heredocs_own_line` covers the pipe form, the trailing-`<<EOF` form, and the bodyless one, and fails on 0fe7532. Every heredoc test before it put the introducer at end of line, which is why the suite missed this.

Live, against the shared install with the lease held by another session: the same payload through the hook is now denied with "This command would mutate it (kubectl apply)", where 0fe7532 exited 0 with no output.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1511 @@
...
+SHELL_WORDS = {"bash", "sh", "zsh", "dash", "ksh", "busybox"}
+ASSIGNMENT_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+DASH_C_RE = re.compile(r"^-[a-zA-Z]*c$")
+# `OUT=$(kubectl apply ...)` -- the assignment swallows the command word, so
+# the substitution is pulled out and classified in its own right.
+SUBSTITUTION_RE = re.compile(r"\$\(([^()]*)\)|`([^`]*)`")
+CD_RE = re.compile(r"^\s*cd\s")
+# `<<EOF`, `<<-'EOF'`, `<<"EOF"` -- but not `<<<`, a here-string.
+HEREDOC_RE = re.compile(r"<<-?[ \t]*(?![<(])(['\"]?)([A-Za-z_][A-Za-z0-9_]*)\1")
+
```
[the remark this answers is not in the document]
bradhoekstra: Fixed in e21e715. Reproduced as described: `strip_heredocs("grep -q ready <<<yes && kubectl --context <ctx> delete ns kubeagents-system")` returned `'grep -q ready <'`, classifying to `(None, None)`.

`HEREDOC_RE` gained a `(?<!<)` lookbehind. Both guards are needed and neither is redundant — the lookahead rejects the here-string at the `<` it starts on, the lookbehind rejects it at the offset the search retries from.

The `nl == -1` branch you flagged separately is fixed in the same commit: it no longer discards the rest of the line, for the reason given on the other thread.

`test_a_here_string_is_not_a_heredoc` covers `<<<` before and after the mutation and fails on 0fe7532. Live, with the lease held by another session, the here-string line is now denied naming `kubectl delete`; 0fe7532 let it through silently.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1511 @@
...
+def get_lease(install):
+    """Returns (data_dict, resource_version) or (None, None) if unheld."""
+    rc, out, err = kubectl(install, ["get", "configmap", CM_NAME, "-o", "json"])
+    if rc == 0:
+        obj = json.loads(out)
+        return obj.get("data", {}), obj["metadata"]["resourceVersion"]
+    if "NotFound" in err or "not found" in err:
+        return None, None
+    raise Unreachable(err.strip().splitlines()[-1] if err.strip() else "rc=%d" % rc)
```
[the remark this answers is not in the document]
bradhoekstra: Fixed in e21e715. The read now matches the ConfigMap's own absence — `CM_ABSENT_RE`, `configmaps? "live-test-lease" not found` — and everything else raises `Unreachable`, which is the position the design doc already stated and the code did not hold to.

A missing namespace and an absent auth plugin go the same way as `kubectl not found on PATH`. Erring toward `ask` there is the right side: the alternative is reporting a cluster nobody can see as idle.

`test_only_the_configmaps_own_absence_means_the_lease_is_free` asserts all three failures raise and that the real message still returns `(None, None)`; it fails on 0fe7532.

Live: `status` with `kubectl` off `PATH` now prints `unreachable: kubectl not found on PATH`, where 0fe7532 printed `FREE` for the same cluster. The positive branch is exercised by a real `kubectl get` against the install, whose stderr is exactly that message.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1511 @@
...
+        payload = json.load(sys.stdin)
+        if payload.get("cwd"):
+            os.environ["CLAUDE_CWD"] = payload["cwd"]
+    except (json.JSONDecodeError, ValueError):
+        payload = {}
+    try:
+        installs = resolve_installs(payload.get("cwd"))
+    except Exception:  # noqa: BLE001 -- nothing to release if we cannot look
+        sys.exit(0)
+    for install in installs.values():
```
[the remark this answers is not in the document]
bradhoekstra: Fixed in e21e715. Confirmed against the 2.1.238 binary first: the payload is `{hook_event_name:"SessionEnd",reason:t}` and the reason literals include `clear` and `resume` alongside `prompt_input_exit`, `logout`, and `other`.

`hook_sessionend` now exits before touching anything when the reason is one this session survives (`SESSION_CONTINUES = {"clear", "resume"}`). The lease stays, the next mutating command renews it as usual, and the real exit releases it.

`test_sessionend_on_a_clear_keeps_the_lease` and `test_sessionend_on_a_real_exit_still_releases` cover both sides; the first fails on 0fe7532.

Live, against a lease the tool acquired in a scratch namespace so the shared install's own lease was not disturbed: `{"reason": "clear"}` left the ConfigMap and the local token in place, `{"reason": "prompt_input_exit"}` deleted both. The design doc's release paragraph now says so.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1511 @@
...
+
+
+# --------------------------------------------------------------------------
+# Cluster I/O
+# --------------------------------------------------------------------------
+def kubectl(install, args, stdin=None):
+    cmd = ["kubectl", "--context", install.context, "-n", install.namespace] + args
+    kenv = dict(os.environ)
+    if install.kubeconfig:
+        kenv["KUBECONFIG"] = install.kubeconfig
```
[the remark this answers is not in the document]
bradhoekstra: Both halves confirmed, fixed in e21e715 — though I took the narrower of the two available fixes.

The merge bug is the real one: `resolve_installs` skipped a config entry whose context was already discovered from `vars.sh`, dropping its `aliases` with it. So the remedy failed for exactly the install most likely to need it, the one you have a checkout of. It now merges the entry's aliases into the discovered install (`Install.add_aliases`), with the checkout still winning every field it discovered. `test_aliases_reach_an_install_the_checkout_already_found` fails on 0fe7532.

I did not make `kubectl()` try the aliases. Probing the kubeconfig for a reachable name on every lease call buys robustness for a case that has not come up, and it would make the tool's own cluster I/O depend on a guess. Instead the config entry's `context` is now documented as the name the kubeconfig actually answers to, with `aliases` being the further names the classifier recognises — which is what the field can honestly deliver. `docs/designs/live-test-lease.md` says that in the paragraph you cited, so the documented remedy and the code now agree.

Live: `kubectl --context <alias> delete ns …` from a checkout that discovers the same install is denied, where 0fe7532 allowed it.

.claude/settings.json.example
```diff
@@ -0,0 +1,27 @@
...
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/live_test_lease.py\" hook-pretooluse",
+            "timeout": 50
+          }
+        ]
+      }
```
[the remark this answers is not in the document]
bradhoekstra: Fixed in 6fcf632, by removing the file. `.claude/settings.json` is now `.claude/settings.json.example`, copied into place per checkout; the `.gitignore` negation moved with it, so the copy is untracked.

You are right that documentation was not the answer. My adversarial pass had raised this and I answered it with a paragraph in Limits, on the reasoning that enforcement-by-default was worth the exposure. It is not, and the reason is the one you name: `.claude/commands/pr-review-batch.md` checks fork branches out into a worktree of an already-trusted root, so the reviewer is exactly the person whose next Bash call the branch would arm — with their kubeconfigs and their mode-600 `vars.sh` in reach. I searched the 2.1.238 binary for a re-approval gate on changed hooks and there is none; workspace trust is per directory and `disableAllHooks` is global.

The cost is real and I have not hidden it: a contributor who never runs the `cp` has no enforcement at all. Limits now says that in those words rather than presenting the example as a free win. `AGENTS.md` and the contributing page carry the copy step, and the design doc's status banner and hook section point at the example.

Live: `git check-ignore` attributes the copy to `.gitignore:62` and `git status` does not list it. Driving both hook events through the command strings read out of the copied file — shlex-split with `$CLAUDE_PROJECT_DIR` expanded as Claude Code expands it, not a paraphrase — a read exits 0 silently, `OUT=$(kubectl apply -f x.yaml 2>&1)` is denied, and `SessionEnd` with `reason: clear` leaves the lease alone.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1557 @@
...
+    try:
+        installs = resolve_installs(payload.get("cwd"))
+    except Exception:  # noqa: BLE001 -- nothing to release if we cannot look
+        sys.exit(0)
+    for install in installs.values():
+        # No local token means this session never claimed this install, and
+        # releasing somebody else's lease on the way out is the failure the
+        # whole mechanism exists to prevent.
+        if not read_local_token(install):
+            continue
```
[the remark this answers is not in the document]
bradhoekstra: Confirmed and fixed in 6fcf632, including the `release --all` half you flagged.

`SessionEnd` and `release --all` now enumerate the local token files for this session's holder key instead of iterating what the cwd can discover. A token file exists only where this session claimed a lease, so releasing somebody else's on the way out — the case the old `read_local_token` guard was there for — is not reachable from the new path either.

The part that made this more than a loop change: a reconstructed install has to know where the lease was taken. Guessing `DEFAULT_NAMESPACE` would delete a ConfigMap in the wrong namespace, report success, and clear the token that was the only local record of the real lease — a worse outcome than the bug. So the token file now holds JSON carrying the context, namespace, and kubeconfig alongside the token, and a checkout-discovered `Install` is preferred only where its namespace agrees with the record. A bare-hex file written by an older build still reads as a token; its namespace is unrecoverable, so it gets the default, which is the best that case allows. The file is written 0600 now, since the token is a capability.

`status` also reports a held install the checkout cannot see, so `release --all` can no longer reach one `status` never showed.

Live, on the shared install, every step run through both e21e715 and the head:

- From `/tmp`, which discovers nothing, with a lease this session held: e21e715's `status` printed "No protected install found"; the head printed `HELD … <-- yours`, having recovered the install from the token file — a bare-hex one, as it happens, so the legacy fallback is exercised by the real machine state rather than only by a test.
- Against a scratch lease in `default` on the same cluster, cwd `/tmp` and `$KUBE_AGENTS_LIVE_TEST_ENVS` pointed at a file that does not exist: e21e715's `hook-sessionend` left the ConfigMap and the token in place; the head deleted both. Same again for `release --all`, which e21e715 refused with exit 2.
- The delete landed in `default` and the shared `kubeagents-system` lease was untouched, which is the namespace-carrying half of the fix rather than a coincidence.

Six tests, all failing on e21e715: the two `SessionEnd` cases above plus `held_installs` on the namespace, the kubeconfig, another session's token, the legacy file, and the file mode.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1557 @@
...
+        # `OUT=$(kubectl apply -f x)` -- the assignment swallows the command
+        # word, so classify what the substitution runs.
+        if _depth < MAX_WRAPPER_DEPTH:
+            inner = [g for m in SUBSTITUTION_RE.finditer(seg) for g in m.groups() if g]
+            for payload in inner:
+                target, reason = classify(payload, installs, cwd=cwd,
+                                          _depth=_depth + 1)
+                if target:
+                    return target, reason
+        try:
```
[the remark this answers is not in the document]
bradhoekstra: Reproduced verbatim and fixed in 6fcf632. `OUT=$(kubectl delete ns kubeagents-system 2>&1)` split to `['OUT=$(kubectl delete ns kubeagents-system 2>', '1']` and classified `(None, None)`.

`split_segments` now treats a `$(...)` as one piece however many separators are inside it, counting nested parens so `$(a $(b) c)` ends where it really ends. The contents are still classified — `classify` already recurses into the substitution once the segment survives whole. That is what makes the failure total rather than partial: the leading fragment is `VAR=$(cmd`, which `strip_wrappers` discards as an assignment, and the opening paren goes with it, so `SUBSTITUTION_RE` cannot find the substitution either. Nothing is left holding the command word.

The `2>&1` half is fixed too — an `&` preceded by `>`/`<`, or followed by `>`, is a redirection rather than a separator. I want to be accurate about that one: on its own it is shell parity, not a demonstrated bypass, because the command word sits at the front of the bad split and classifies from there. The test says so in as many words rather than claiming a bug it did not have.

Live, on the shared install with the lease held so each verdict is an observable deny or silent allow, every payload run through both e21e715 and the head. Denied on the head, silent on e21e715: `OUT=$(kubectl delete ns kubeagents-system 2>&1)`, `LOG=$(./upgrade.sh 2>&1 | tail -20)`, `R=$(make deploy || true)`. Silent on both: `kubectl get pods -n kubeagents-system 2>&1 | head` and `OUT=$(kubectl get pods 2>&1)`.

Three tests; the two that assert the substitution cases fail on e21e715.

bradhoekstra:
/request-review

The bot's read on 6fcf632 found nothing, but the check concluded `neutral` rather than `success` — its summary says the pass was a sample, not full coverage of the diff — so the green-check gate never fired. All eight review threads from the two earlier rounds are resolved.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1696 @@
...
+            subs = positionals(args)
+            # `docker buildx build --push -t <ref>` pushes just as `docker push`
+            # does, and it is the standard single-step form for the
+            # `--platform linux/amd64` builds AGENTS.md asks for.
+            if "push" in subs[:2] or "--push" in args:
+                # Matched against the whole line, as every other family is:
+                # `IMG=<ref> && docker push $IMG` names the registry in the
+                # segment before the one doing the pushing.
+                for install in installs.values():
+                    if install.registry and install.registry in command:
```
jayantid, line 1443: `docker push` is guarded only when the install's registry prefix appears literally on the command line (`scripts/live_test_lease.py:1365`). A push whose image ref arrives through indirection matches no registry and passes silently. I ran both of these through `classify` at 6fcf632 against a protected install and got no target, while the literal form is guarded:

```bash
docker push $(cat /tmp/last-image)
source .env && docker push "$IMG"
```

If it merges, the motivating incident survives the hook. An agent pushes over the tag the shared install is running, with no lease taken and no deny, and another agent's live validation describes an image that is gone. `resolve()` already treats an unexpanded `--context "$CTX"` as a missing answer rather than a negative one; the docker branch never applies that principle.

When no registry matched and the segment still carries unexpanded shell syntax, return `UNKNOWN` so the hook asks. Pushes are rare, so the prompt is cheap — the same asymmetry the installer branch argues.

```python
if "push" in subs[:2] or "--push" in args:
    for install in installs.values():
        if install.registry and install.registry in command:
            return install, ("docker push to the %s registry"
                             % install.name)
    if any(unexpanded(tok) for tok in args):
        return UNKNOWN, "docker push whose image ref could not be resolved"
    continue
```

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1696 @@
...
+# false positive as classifying `cat install.sh` as a run.
+INSTALLER_NOOP_FLAGS = {"-h", "--help", "-?", "--dry-run"}
+
+# Prefixes that wrap a command without changing what it runs. Stripping them is
+# what lets the classifier look at the command word rather than at every token
+# on the line -- `grep -rn install.sh docs/` reads about the installer.
+#
+# `xargs` is here for the bulk-delete idiom: `kubectl get pods -o name | xargs
+# kubectl delete pod` puts the mutation after a command word of its own.
+WRAPPER_WORDS = {"sudo", "nohup", "command", "exec", "nice", "ionice", "stdbuf",
```
jayantid, line 756: `eval` is not in `WRAPPER_WORDS` and is not a shell word, so the classifier reads `eval` as the command and stops. Both `eval kubectl delete ns kubeagents-system` and `eval "kubectl delete ns kubeagents-system"` classify as nothing at 6fcf632. Function definitions fail the same way: in `f() { kubectl delete ns x; }; f` the first token is `f()`, which stops `strip_wrappers` before it reaches the `{`. A `case` arm (`prod) kubectl delete ns x;;`) hides the verb behind `prod)` too.

A mutation in any of these forms runs against a held install with no deny and no lease — the fail-open direction, in the same class as the heredoc and `$(...)` gaps the earlier review rounds fixed.

Bare `eval` is one word added to `WRAPPER_WORDS`; quoted `eval "…"` can reuse the `shell_payloads` recursion the way `bash -c` does. If the function and `case` forms are not worth parsing, list them in the design doc's Limits — "What counts as a mutation" currently reads as though the classifier follows the shell wherever the command word moves.

docs/designs/live-test-lease.md
```diff
@@ -0,0 +1,257 @@
...
+either way.
+
+The lease renews only when another Bash call passes through the hook, so a single unattended run
+longer than the TTL — `install.sh` against a fresh GKE cluster — expires mid-run and can legitimately
+be taken over. Take a longer one by hand first: `acquire --ttl 180`.
+
+A context renamed locally is invisible to discovery: `vars.sh` records what the install _is_, and
+`gke_<project>_<location>_<cluster>` is the name `gcloud` writes, so a `kubectl config
+rename-context` leaves commands naming the short form unresolvable. Add the new name to that
+install's `aliases` in the config file, keying the entry on the context the kubeconfig actually
```
jayantid: This remedy cannot produce a working lease for the install it names — the one you have a checkout of. Cluster I/O always runs `kubectl --context <install.context>` (`scripts/live_test_lease.py:476`), and a checkout-discovered install's context is the canonical `gke_…` name composed from `vars.sh`. After a real `kubectl config rename-context`, that name is gone from the kubeconfig.

Both readings of this paragraph fail. An entry keyed on the canonical context with the new name in `aliases` (what the tests set up) merges, but every `get_lease` then fails with `context does not exist` — `Unreachable` — so the hook answers every mutating command with `ask`, indefinitely. An entry keyed on the renamed name, which is what "keying the entry on the context the kubeconfig actually answers to" says, does not merge: at 6fcf632, `resolve_installs` returns two installs for one cluster, and `install_from_vars_sh` resolves `./install.sh` to the unreachable canonical one.

Either let a config entry that names a discovered install by alias override the context used for I/O, or say here that the remedy restores classification only and a working lease needs the canonical name back in the kubeconfig (`gcloud container clusters get-credentials` re-adds it alongside the renamed one).

scripts/live_test_lease.py
bradhoekstra: Fixed in 1df9afe. Reproduced first: on 6fcf632, `docker push $(cat last-image)` and `source .env && docker push "$IMG"` both classify as nothing against the shared install — no lease taken, no deny, which is the incident this tool exists to prevent.

An unreadable registry is now `UNKNOWN`, so the hook asks. Only the host portion of the ref decides: `ghcr.io/x/y:$(git rev-parse HEAD)` says plainly where it is going, and prompting on it would be a false positive on somebody's own registry. Both spellings are covered — the positionals after `push`, and the `-t`/`--tag` values a `docker buildx build --push` uses instead. Silent when no protected install records a registry at all, since then there is nothing a push could overwrite.

Tests: `test_a_push_whose_ref_did_not_expand_asks` (four subtests, all fail on 6fcf632), plus `test_a_readable_registry_is_an_answer_however_computed` and `test_no_recorded_registry_means_nothing_to_overwrite`, which bound the new behaviour and pass on 6fcf632 by construction. Live on the shared install with the lease held by another session key: the two payloads above go silent → **ask**; `docker push <the install's registry>` stays denied; `docker push ghcr.io/someone/platform:$(git rev-parse HEAD)` stays silent.

scripts/live_test_lease.py
bradhoekstra: Fixed in 1df9afe. `eval` hid the command word two ways: bare, it was read as the command itself and the classifier stopped; quoted, shlex collapsed the payload into one argument so there was no command word left to find. `shell_payloads` now joins `eval`'s arguments back and recurses, which covers both. A function definition put its name where the command word goes — `{` was already a stripped keyword, the `f()` in front of it was not — so `strip_wrappers` stopped one token early and never read the body; all three spellings bash accepts (`f() {`, `function f() {`, `function f {`) are handled now.

Tests: `test_eval_is_a_shell_wrapper_that_takes_no_flag` and `test_a_function_definition_holds_the_command_word`, both failing on 6fcf632. Live: `eval kubectl --context <install> delete ns …`, its quoted form, `f() { … }; f`, and `function f { … }; f` all go silent → **deny**, while `eval kubectl … get pods …` stays silent.

One gap left open deliberately: a `case` arm still hides the command word behind its pattern label. Finding it means widening "skip tokens until one looks like a command", which is what lets a real argument be read as one, and a false mutation costs an agent the cluster for an hour. It is in the design doc's Limits, and `test_a_case_arm_is_a_gap_the_design_doc_owns` asserts the silent pass and the paragraph together so it cannot become undocumented.

docs/designs/live-test-lease.md
bradhoekstra: You were right, and it was worse than a wording problem — the remedy did not work either way round. I checked it on a copy of a kubeconfig with the canonical context renamed away. Aliases widen classification only; every cluster call goes through `--context install.context`. So a config entry keyed on the canonical name leases against a context the rename removed and answers `unreachable: error: context "gke_…" does not exist` forever, which means the hook asks about every mutating command. An entry keyed on the renamed name does resolve and lease, but as a second install alongside the one the token record recovers — one cluster, two leases.

Limits now says the entry restores classification, and that renaming the canonical context back (`gcloud container clusters get-credentials`) is what restores the lease. 1df9afe. Two tests pin it: `test_an_alias_never_becomes_the_context_the_lease_runs_under` asserts `kubectl()` always passes the canonical context, and `test_keying_a_config_entry_on_the_renamed_name_splits_the_install` asserts the two-install outcome.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1759 @@
...
+
+def unreadable_registry(ref):
+    """Does this ref hide the registry it pushes to?
+
+    Only the host matters. `ghcr.io/x/y:$(git rev-parse HEAD)` computes its
+    tag and still says plainly that it is not going to a protected registry,
+    so treating any `$` in the ref as unreadable would prompt on ordinary
+    pushes elsewhere -- the false positive that costs an agent an hour.
+    """
+    return unexpanded(ref.split("/", 1)[0])
```
jayantid: `unreadable_registry` at `scripts/live_test_lease.py:1109` judges only the host segment of a push ref. The recorded registry prefix has two segments: host and project (`us-central1-docker.pkg.dev/acme-prod`). A ref that spells the host literally and the project through a variable matches neither the literal guard at `scripts/live_test_lease.py:1412` nor the unreadable check. At 1df9afe, `docker push us-central1-docker.pkg.dev/$PROJECT_ID/platform:dev` against that install classifies as nothing.

If it merges, the motivating incident still gets through this spelling. `vars.sh` exports `PROJECT_ID`, so an agent that sourced it writes exactly this ref: region literal, project as a variable. The push overwrites the tag the shared install runs, with no lease taken and no deny. The reverse spelling (`$REGION-docker.pkg.dev/acme-prod/...`) asks, because there the host is the part that did not expand.

Treat a ref as unreadable when its literal prefix — everything before the first unexpanded token — is still a prefix of a protected registry:

```python
def unreadable_registry(ref, registries):
    if unexpanded(ref.split("/", 1)[0]):
        return True
    literal = re.split(r"[$`]", ref, 1)[0]
    return any(r.startswith(literal) for r in registries)
```

`ghcr.io/x/y:$TAG` stays silent — no protected registry starts with `ghcr.io/x/y:`. `us-central1-docker.pkg.dev/other-proj/x:$TAG` stays silent too. Only a ref whose readable half is still on the path to a protected registry asks.

scripts/live_test_lease.py
```diff
@@ -0,0 +1,1759 @@
...
+# Shell syntax that precedes a command word without being one. `for ns in a b;
+# do kubectl delete ns $ns; done` splits into a segment whose first token is
+# `do`, and the bulk-cleanup loop is the most ordinary thing an agent writes.
+SHELL_KEYWORDS = {"do", "then", "else", "elif", "if", "while", "until", "for",
+                  "!", "{", "}"}
+SHELL_WORDS = {"bash", "sh", "zsh", "dash", "ksh", "busybox"}
+ASSIGNMENT_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+# `f() { kubectl delete ns x; }; f` -- the definition's name is where the
+# command word would otherwise be, and `{` right behind it is already a keyword.
+FUNCTION_DEF_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*\(\)$")
```
jayantid, line 777: `FUNCTION_DEF_RE` at `scripts/live_test_lease.py:775` requires the name and parens in one token. Bash also accepts a space between them. shlex splits `f () {` into `f` and `()`, so `strip_wrappers` stops on `f` and the body goes unread. At 1df9afe, `f () { kubectl --context <install> delete ns x; }; f` classifies as nothing, while the three spellings the tests cover are all caught.

A mutation written this way runs against a held install with no deny and no lease — the same fail-open class as the `f() {` form this commit fixed.

Peek one token ahead in `strip_wrappers`:

```python
# `f () {` -- POSIX allows space between the name and the parens.
if (i + 1 < len(tokens) and tokens[i + 1] == "()"
        and re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", tok)):
    i += 2
    continue
```

scripts/live_test_lease.py
bradhoekstra: Fixed in `4077aa0`. Reproduced against 1df9afe first: `docker push us-central1-docker.pkg.dev/$PROJECT_ID/platform:dev` classified as `None` there, and the same payload through the hook against the real gkedemos install passed silently.

`unreadable_registry` now takes the protected prefixes and asks the question of the whole literal half of the ref, as you wrote it. One deviation: a ref with no unexpanded token at all returns `False` before the prefix test, so a fully readable ref stays the literal match's question rather than this one — otherwise a bare `us-central1-docker.pkg.dev` would read as unreadable when it is the opposite.

Live, against the shared install with its real lease held by another session: **ask** on the head and silent on 1df9afe for `us-central1-docker.pkg.dev/$PROJECT_ID/platform:dev` and the `docker buildx build --push -t …/${PROJECT_ID}/…` form; silent on both for `us-central1-docker.pkg.dev/some-other-project/x:$TAG` and `ghcr.io/someone/platform:$(git rev-parse HEAD)`; unchanged deny for a fully literal push to the install's registry. Eight new payloads in the unit tests, five of which fail on 1df9afe.

scripts/live_test_lease.py
bradhoekstra: Fixed in `4077aa0` with your lookahead. Reproduced against 1df9afe: `f () { kubectl --context <install> delete ns x; }; f` classified as nothing there.

One thing your version does not reach: `function f () {`. The `function` branch consumes the name and leaves `()` alone at head position, where it breaks the loop the same way — so `()` also joins `SHELL_KEYWORDS`. Both spellings are now tests.

Live, against the shared install with its real lease held by another session: both forms are **denied** on the head under a message naming the holder, and both pass silently on 1df9afe. `f () { kubectl … get pods … }` stays silent on the head, so the fix does not turn the read form into a mutation.

jayantid (approved):

bradhoekstra:
/request-review

Both findings from the last pass are fixed in `4077aa0`, with the threads answered and resolved. No new bot pass requested — this round is two narrow classifier fixes with regression tests and a live re-run against the shared install.

Issue #658, opened 2026-08-12
/sethome fails on Slack and Google Chat: the operator subPath-mounts config.yaml read-only over the file the agent must write


adamparco:
## Summary

`/sethome` fails on **both Slack and Google Chat** with `Failed to save home channel: [Errno 13] Permission denied: ''`.

These are not two bugs. They are one filesystem failure surfacing twice: the operator subPath-mounts its rendered ConfigMap **directly over `$HERMES_HOME/config.yaml`**, making the one file the agent must write a read-only mount point. There is no ACL involved — `_handle_set_home_command` consults no user, role, or admin list at all.

The failure is compounded by a deploy-provenance problem: the running image and operator were built from a branch containing an **explicit revert of the fix** for this exact bug. The fix exists in the tree; it is not in the build that is running.

The empty path in the message (`''`, not `'/opt/data/config.yaml'`) is a second, independent defect in the shared platform adapter — it is what made the report unreproducible from the reply alone, and it is why this reads as an authorization denial when it is a read-only filesystem.

## Impact

- No user can set a home channel from any chat platform on an affected deployment.
- **Already realised, silently, for hours:** every `deliver=all` cron job on the platform profile has been failing to deliver, because the scheduler resolves its target from the same unwritable file. `jobs.json` carries `"last_delivery_error": "no delivery target resolved for deliver=all"`.
- The pod reports `4/4 Running`, `restartCount=0`, and the `PlatformAgent` CR reports `phase: Ready`. Every health check a person would run says the install is fine. The defect surfaced only when a human typed a slash command, ~28 minutes after boot.
- Any deployment cut from `upstream/main` reproduces this exactly — see Root cause 6.

## Environment

| | |
|---|---|
| Project / cluster | `adamparco-kage` / `platform-agent-host` (us-east4) |
| Pod | `platform-agent-gateway-9bbcf9c9c-2mcks`, ns `kubeagents-system`, started `2026-08-12T12:20:46Z` |
| Images | `platform-agent`, `k8s-operator`, `credential-proxy` all at `dev-8350e5f-0811-184934` |
| Build commit | `8350e5f` — `merge: PR #602 ap-result-delivery`, 2026-08-11 18:45:21 -0400 |
| Runtime uid | `10000(hermes)` |
| Platforms enabled | `google_chat: true`, `slack: true` |

## Reproduction

[...]

PR #862, opened 2026-08-21
fix(triage): store the incident report on the kanban delivery path

### LINKED ISSUE #802: Event triage stops at the report: nothing writes the incidents row an 'apply' reply needs`
## What happens

An event-triage report reaches the user's chat thread, they reply `apply`, and nothing
happens. The reply arrives at the Planning Agent as the bare word `apply` — no report, no
options, no cluster — so it cannot open the Pull Request the report was written to support.

## Why

The agent that acts on that reply gets its context from the `incident_context` plugin, which
looks the report up by thread:

- `agents/platform/plugins/incident_context/__init__.py:36` — `_lookup(chat_id, thread_id)`
  calls `GET /v1/incidents/by-thread`, and on a miss returns `None`, leaving the message
  unrewritten. Deliberately fail-open, so the miss is silent.
- `agents/platform/scripts/session_kv_server.py:858` — the route reads the `incidents` table.
- `agents/platform/scripts/platform_mcp_server.py:731` — `send_notification` is the table's
  only writer, and only when it posted into a thread.

#630 replaced `send_notification` on the event-triage path with `kanban_complete`: the card is
subscribed to the alert's thread and the notifier posts its `result`. Delivery works, and no
`incidents` row is ever written.

Before #630 no report reached a human on this path at all, so nobody ever replied and the gap
was invisible. Now the report arrives and the reply is the expected next step.

Nothing unsafe results — the Planning Agent holds no GitOps write path — but the promise cannot
be kept, so PR #738 removed the `To authorize: reply 'apply'` bullet from the report template
rather than ship an invitation nothing honours. This issue is the other half.

## Fix

POST the completed report to `/v1/incidents` on the delivery path, keyed by the substituted
`chat_id`/`thread_id`. The kanban notifier already holds all three values together — the
subscription row carries the address that
`deploy/docker/patches/kanban_event_routing.py` substituted, alongside the card's `result`.

Shape, for whoever picks this up:

- The notifier lives in `/opt/hermes`, which is read-only in the pod, so this is a new
  `deploy/docker/patches/` module with an apply script, a verify script, and unit tests —
  the same four-file shape as `kanban_event_routing.py`.
- Fire on the same terminal transition the notifier already posts on (`events={'completed'}`).
- Confirm the notifier process has `SESSION_KV_API_KEY`; every `/v1/incidents` route is
  authenticated and an unset key is a 401 the plugin swallows.
- `store_incident` is `INSERT OR IGNORE` and keeps the first report per thread, which is the
  behaviour wanted here — the RCA, not any later chatter.
- Failing to store must not fail the notify. A report delivered and unstored is today's
  behaviour; a report stored and undelivered is a regression.

## Then restore what #738 withheld

Five placeholders spell `#ISSUE-TODO` and should carry this issue's number until it is fixed,
then be resolved with it:

- `agents/platform/scripts/session_kv_server.py` — `_triage_task_body`'s docstring, and the
  "Do not end the report by inviting a reply" instruction plus the `To authorize:` bullet it
  replaced
- `agents/platform/scripts/test_session_kv_server.py` — the two tests pinning the removal
- `agents/platform/docs/autoops-architecture.md` — the Contract 3 note and the Known gaps entry
- `agents/platform/docs/session_management.md` — the Phase 3 caveat above the sequence diagram

## Live validation this needs

Beyond a rebuild: an alert triaged end to end, then an actual `apply` reply that opens a Pull
Request. That needs a GitOps repository wired up, which is why it did not ride along with #738.

stalhaali:
## Summary

The kanban notifier now writes the `incidents` row when it delivers a completed card whose result carries remediation options, so a user replying "apply Option A" in an event-triage thread reaches an agent that can see what Option A was. With a writer back on the delivery path, the report template's `To authorize: reply 'apply'` bullet is restored, along with the four places that recorded why it had been withheld. Generated with the help of Claude.

## Why This Change

An event-triage report ends with two or three remediation options, and a reply naming one is the entire point of the format. The mechanism that honours such a reply exists and works: `agents/platform/plugins/incident_context/` is a `pre_gateway_dispatch` hook that, when a message arrives in a thread it recognises, prepends the stored report to the user's words before the agent sees them. It looks the thread up with `GET /v1/incidents/by-thread`, reading the `incidents` table.

Nothing wrote that table on the event-triage path any more. Its only writer there was `platform_mcp_server.send_notification`, and #738 replaced that egress call with `kanban_complete`, delivering the report through the card's notify subscription instead. So the report arrived in the thread and the row behind it did not, and a reply of "apply Option A" reached the front door as three bare words. The template had been amended to stop making the offer, which is the honest short-term response and not a fix — the format's whole affordance was gone.

Without this, event triage stops at the report: a user reads two remediation options and has no way to ask for one.

## What Changed

- **`deploy/docker/patches/kanban_notifier.py`** gains a fifth concern, `store_incident_report`, alongside clip / deliver / wake / marker. It POSTs `/v1/incidents` over loopback with the pod's `SESSION_KV_API_KEY`, the way `send_notification` did — the notifier is a different process from the Session KV server, so it cannot use the in-process `_store_incident_report`, whose docstring says as much.
- **`deploy/docker/patches/apply_kanban_notifier.py`** emits the call in the wake set, alongside the marker call, and carries the new name on the existing import trailer. Extending this pair rather than adding a patch module: the module docstring explains why three patches were merged into one, and the image has a 128-layer overlay2 ceiling. No new `RUN` or `COPY`.
- **`agents/platform/scripts/session_kv_server.py`** restores the `To authorize:` bullet in `_triage_task_body`, replaces the instruction telling the agent not to write a call-to-action with one telling it the bullet is the call to action rather than another option, and rewrites the docstring paragraph that argued for withholding it.
- **`agents/platform/scripts/test_triage_reply_roundtrip.py`** (new) drives the chain end to end: the notifier stores a delivered report, then `incident_context` resolves a reply in that thread against the live `session_kv_server` app.
- **`agents/platform/plugins/incident_context/__init__.py`** stops calling every unlabelled row in the by-thread-miss index a scheduled report, since triage rows now make up most of them. Wording only — see design call 2.
- Docs: the architecture doc's approval-interaction bullet and Known gaps, the session-management caveat above the sequence diagram, both credential-isolation pages' caller lists, and the site's Cluster Agents page, which had no description of the reply half of the loop.

### Design call 1 — which cards store a row

Only cards whose result has a `What to do` heading with a lettered `Option [A-Z]` under it.

Card metadata cannot answer the question. By the time the notifier sees the subscription, `kanban_event_routing` has rewritten the row to look exactly like an ordinary chat card's, and the assignee and title are model-written. The artifact is the only reliable evidence available, and it is also the right evidence: the row exists so `apply Option A` can be resolved, and a card whose result is a one-line status has no Option A.

Getting it wrong is sticky in one direction. `POST /v1/incidents` is `INSERT OR IGNORE` and deliberately keeps the first report per thread, so a status-line card stored in a thread that later receives a real report would shadow it for the table's whole 14-day retention window. Storing too little costs one reply; storing too much poisons the thread.

### Design call 2 — does #802 extend the recent-reports index

Partly: the wording, not the schema.

#731 built the miss path this change inherits. When the by-thread lookup misses, `incident_context` falls back to `GET /v1/incidents/recent` and prepends a label-only index so the agent asks which report the user means rather than answering about the wrong one. That index is cron-shaped — `job_id`, `title`, `profile`, all read out of `session_metadata`, which only the relay writes — so a triage row arrives with nothing to name it and renders under a preamble that calls everything in the list a scheduled report.

The wording is now wrong in the common case, and #802 is what makes it common. `send_notification` rows already had no label, but they are rare; event triage is the highest-volume writer this table will have. The cost of the wrong word is on record in #731's own code: told a smoke-test report was an incident, the agent spent its opening line correcting the framing before answering. Told an incident is a scheduled report, it has the same reason to argue instead of answering. So the fallback label is `report` and the preamble says "Reports posted in this space recently" — two strings and one test assertion.

Giving triage rows a real label is left out deliberately. It means either writing a `session_metadata` row from the notifier, which the notifier has no session for, or teaching the index a second vocabulary alongside cron's. That is schema work, and it belongs with the incident-ledger design in #846 rather than bolted onto a delivery-path fix.

### Design call 3 — failure handling

Never fatally, never silently.

Not fatally, because an exception on the delivery path escapes into the notifier tick, rewinds the cursor, and re-posts a report the user has already read. A report delivered and unstored is the behaviour this replaces; a report delivered twice is worse than what it replaces.

Not silently, because a silent miss here is the same shape as the bug #630 was — the user sees a report, replies, and gets a blank stare. #738's own routing patch settled this by logging a warning naming the session on every fall-through, and this follows it: a report that had options and did not get stored logs a WARNING naming the card and the thread. An ordinary card with nothing to apply logs at DEBUG (most cards look like that, and a warning per completion per board is noise), and a stored one logs at INFO — the only positive evidence that the second turn is reachable. INFO means `/opt/data/logs/agent.log` inside the pod, not `kubectl logs`, whose stdout handler is at WARNING; the warning path is visible in both. Each of the three lines logs the chat and the thread as separate fields rather than joining them with a slash, because a Google Chat `thread_id` is already the fully qualified `spaces/<space>/threads/<key>`.

### Following #731, and one thing it does differently

#731 made `relay_cron_report` the table's second writer, and its ordering rule applies here: "the routing registration and the incident store happen after the send because both need the thread the send resolves." The notifier's store sits after the send for the same reason, in the `else` clause the notifier reaches only once every text ping for the delivery has gone out. Its thread is resolved a step earlier — it is the substituted `sub` address rather than a value the send returns — but the other half holds identically: a row written before the post claims a reader has a report that has not been sent, and on a delivery that then fails to send, `INSERT OR IGNORE` makes that claim permanent.

Where the writers now disagree is what a row holds. `_store_incident_report` persists the Chat Agent's composed message rather than the specialist's finding; `send_notification` persists the notification body; this stores `task.result`, the specialist's own report. Storing the source is required here rather than preferred — the reply names a letter, so the stored text has to be the one the reader was looking at when they picked it, and a paraphrase could renumber the options. But it does make three writers store three different kinds of thing, and anything reading `report` generically has to assume the union. Both readers already do: `incident_context` fences whatever it gets as untrusted, and `list_recent_reports` returns no preview text. Its docstring gave only the relay's composition as the reason for that; it now gives the reason that covers all three.

## Context

Closes #802.

The issue's own body proposes a new four-file patch module; this implements the maintainer's instruction to extend the existing `kanban_notifier.py` / `apply_kanban_notifier.py` pair instead, for the layer-budget and merged-patch reasons above.

Nobody else is on #802. Four open PRs touch files this one touches, all merge-conflict risk rather than duplication: #846 (draft) edits `agents/platform/docs/autoops-architecture.md` for the incident-ledger design; #735 and #695 (draft) touch `session_kv_server.py` and its tests; #697 (draft) is superseded by #738.

Rows the notifier writes have no relay session to name them, so they render unlabelled in the `/v1/incidents/recent` index. `send_notification` rows behave identically and `list_recent_reports` documents that such rows still belong there; what this change adds is volume, which is why the index's wording is corrected here (design call 2) and its schema is not.

## Testing

- `make verify` — exit 0. Re-run after the log-format fix below: `deploy/docker/patches/test_kanban_notifier` (146) and `agents/platform/scripts/test_triage_reply_roundtrip` (13), the two suites in its scope, both green. Go build, vet and test, the operator's Python suite, and every Python suite: `deploy/docker/patches/` 919, `agents/platform/scripts/` 840, `agents/platform/skills/fleet-audit/scripts/` 507, `tests/` 272, `agents/chat/scripts/` 208, `admin_console/tests/` 143, `scripts/` 129, `deploy/docker/plugins/chat/` 39, `agents/platform/plugins/incident_context/` 29, plus the smaller plugin suites.
- `make docs-check` — passes.
- `prettier@3.9.6 --check` — passes on every changed Markdown file. That is the version `.github/workflows/prettier.yml` pins.
- The image build: see Live validation. `verify_kanban_notifier.py`'s new section is the only check in this change that runs against a real patched `/opt/hermes`, and the unit suites cannot substitute for it — they drive a fixture. No `RUN` or `COPY` was added to the Dockerfile, so the image-layer budget is unchanged and `scripts/check_image_layers.py` has nothing new to gate.

### Live validation

Exercised end to end on a live install: GKE cluster `platform-agent-host` in `us-east4`, project `stalhaali-gkedemos`, `PlatformAgent/platform-agent` in `kubeagents-system`, operator `ghcr.io/gke-labs/kube-agents/k8s-operator@sha256:eeb3bac3`, agent image built from this branch as `platform-agent:dev-20260821-153516`. Google Chat is the platform. Three triage rounds, then a reply in the thread.

The image build is itself the first result: `verify_kanban_notifier.py`'s new section runs against the real patched `/opt/hermes` at build time, and the build passes. `apply_kanban_notifier.py` anchors on text `apply_kanban_progress_lines.py` has already rewritten, so a mis-ordered patch fails there rather than at runtime.

Rounds 1 and 2 came from real events; round 3 was a synthetic `Failed`/ImagePullBackOff for `pr802-retest/retest-canary` posted at the Session KV server's own ingress (`POST /sessions`, then `POST /sessions/{id}/inject`), which is the route the event watcher uses — this install runs no watcher. Round 3, from `/opt/data/logs/agent.log`:

```
17:11:12  alert thread spaces/0sm7myAAAAE/threads/vZ1nINPlz0Q registered (session k8s-evt-6a18dd12)
17:11:51  card t_2a08a6c9 -> cluster-...-platform-agent-host-us-east4, subscribed=true
17:50:44  card done, result 962 bytes
17:50:45  kanban notifier: stored card t_2a08a6c9's report against chat ... thread ...
```

Observed at each layer the change claims to touch:

- **Delivery path.** The INFO line above is the new call site's own log, in the branch that runs only after the delivery has been sent. Rounds 1 and 2 logged it too, for threads `__D85QpQzRc` and `IKTRhM3aOH8`.
- **Table.** Rows landed at 16:09:36 (952 B), 16:11:37 (948 B), and 17:50:45 (962 B), keyed on those three threads. Each holds the specialist's own report text, not a paraphrase, and each ends with `- **To authorize:** reply **'apply'** to open a GitOps Pull Request.` — the bullet this change restores.
- **Read path.** `GET /v1/incidents/by-thread` returns 200 with all 962 bytes. `GET /v1/incidents/recent` returns 200 with the row present and unlabelled (`job_id`, `title`, `profile` all empty), which is what design call 2 says to expect and why the preamble wording changed.
- **Reply.** Typing "explain option A in more detail" in thread `vZ1nINPlz0Q` produced an answer about Option A of *that* report — the stale `retest-canary` manifest in `pr802-retest` — closing with "reply with **apply** in this thread". Before `incident_context` was reachable (see below), the same question in the same thread made the agent search the Kanban board and explain two unrelated older reports' Option A instead. That contrast is the test: same words, same thread, different answer, and the difference is the row.

Not a coincidence, in three respects. The table's baseline was ten rows, every one a relay `Here is the scheduled...` report and none from event triage — the row count for triage went 0 → 3 with the change and nothing else moved. The first-write-wins claim was checked rather than assumed: the notifier rows' `created_at` did not move across the following hour while the relay's `GcolpXJZYek` row was rewritten twice in the same table (17:11:01 → 17:21:15 → 18:01:12), which is `INSERT OR IGNORE` against `INSERT OR REPLACE`. And the stored bytes are byte-identical to the card result the notifier delivered, so the row is the report the reader was looking at.

Not covered:

- **The revert-to-red control.** Confirming the row stops appearing with the change backed out needs another image build, which is a rebuild on someone else's machine. The 0 → 3 baseline above is the substitute, and it is weaker.
- **`delivery_mode="wake"`** (`posted=False`). The stack does not emit it, so the guard added for it is unit-tested only.
- **The log-format fix below** is not in `dev-20260821-153516`; it landed after the image was built. Log text only, and the deployed line's content is what proved the layer — it printed the space twice, it did not print the wrong thread.

Two environment problems came up and neither is attributable to this change. `incident_context` was missing from the live `default` profile's `/opt/data/config.yaml`, which is why the first reply attempt got a blank stare; the install's own baseline snapshot (`/opt/data/.default-config-baseline.yaml`) and the repository's `agents/chat/config.yaml` both list it, and the live file had lost it and `agent_roster` on 17 Aug. Re-enabling it fixed the reply. Separately, one `rollout restart` left `hermes gateway run` hung before it bound 8642, so session creation was refused for half an hour; a second restart, with the config edit still in place, came up in 40 seconds.

Artifacts: namespaces `pr802-test` and `pr802-retest` deleted. `incident_context` left enabled on the install, with the prior file at `/opt/data/config.yaml.bak-pr802`; `agent_roster` left as found, reported to the operator.

Expected on Google Chat, and not a failure of this change: the first reply typed into an alert thread is answered in the main space with no thread id. `_post_initial_alert` posts through `hermes send`, which never passes the adapter's outbound path, so the `_ThreadCountStore` heuristic reads the thread as unseen; #731 documented this in `_recover_bot_thread` and backed out its own routing fix after measuring that `pre_gateway_dispatch` fires after the outbound thread metadata is snapshotted. The report still reaches the agent via `_recover_bot_thread`'s raw-thread lookup, which is why the check above is on the content of the answer rather than where it landed.

## Self-Review

Both required passes ran, each in a subagent that did not write the change, handed the diff range `refs/kube-agents-base/main...HEAD`, the mechanical gate output, and the instruction to derive intent from the diff rather than from the commit message. Neither could build the Docker image, so `verify_kanban_notifier.py`'s new section was read statically rather than executed — the build under Live validation closed that gap afterwards, and is where the last finding below came from. The adversarial pass could not run its sibling-PR angle (`gh` is not installed on this machine) or its revert-to-red angle (both passes shared one checkout, so neither was allowed to mutate it).

Angles covered by the adversarial pass: line-by-line, removed behaviour, cross-file tracer and sibling-contract symmetry, ops and security blast radius, reuse, simplification, altitude, conventions and docs, scope and test coverage. The docs-drift pass checked prose against source across the repository, not only in the files the diff touches.

One seam was checked by hand afterwards, because #731's own PR body records that `review-adversarial` was not run on it: this change writes rows that `_recover_bot_thread` reads, and neither pass had reason to look at whether the two agree on what a thread id is. They do. `_post_initial_alert` returns `spaces/<space>/threads/<key>`, routing substitutes that into `sub["thread_id"]`, and `_raw_thread` reads `raw_message["thread"]["name"]`, which is the same resource name — the same assumption the relay path already runs on. So a first in-thread reply on Google Chat, which misses the normalized lookup, still recovers the report by raw thread.

Fixed:

- **The store call ran once per event, not once per delivery** (adversarial, MEDIUM/CONFIRMED). The wake-set anchor is 24 spaces in, which puts it inside `for ev in d["events"]:` as well as `for d in deliveries:`. Reading the delivery's whole kind set meant a `[commented, completed]` delivery stored the row on the `commented` iteration — before the `completed` iteration had sent the report the row claims the reader has — and again on the one that did. The call now takes `ev`, and a test asserts both the emitted source and the runtime behaviour.
- **The store call ignored `send_passive`, unlike its sibling at the same anchor** (adversarial, MEDIUM/PLAUSIBLE). A `delivery_mode="wake"` subscription posts nothing in the thread; a row written anyway would prepend a report the user never saw to their next message, unreplaceable for the TTL. Now takes `posted=send_passive`, matching the marker call. The stack does not emit `"wake"` today, which is exactly why the sibling was defended against it.
- **`actionable_report` matched `Option X` anywhere in the body** (adversarial, LOW/CONFIRMED), so a report quoting an earlier "Option A" under `## Why`, with only unlettered bullets under `## What to do`, took the thread's one first-write-wins slot. The search now starts at the end of the heading match, which is what the docstring already claimed.
- **Nothing coupled the notifier's gate to the template it recognises** (adversarial, LOW/CONFIRMED). The gate lives in `deploy/docker/patches/` and matches two literals; the template that makes an agent produce them lives in `agents/platform/scripts/`. Rewording `## What to do` would have passed every test on both sides while silently stopping every real report from earning a row. `test_triage_reply_roundtrip` now asserts both against each other — the only test in the repository with both in scope.
- **The single-option instruction quoted a string the template does not contain** (adversarial, LOW/PLAUSIBLE): "drop the `'apply Option <letter>'` override", where the template spells it `'apply Option A'` / `'apply Option B'`. An agent matching the quoted string finds nothing to drop. Reworded to name the clause the template actually has, and the guarding test now pins that the clause exists in both halves rather than only that the words "drop the" appear.
- **The architecture doc described the notifier's gate as narrower than the code's** (adversarial, LOW/CONFIRMED): "when it delivers a completed triage card" implies a provenance check the code explicitly does not make. The doc now says it decides from the report's shape and why.
- **`INSERT OR IGNORE` was claimed for the whole table** (both passes) three lines above a new sentence enumerating three writers, one of which — the in-process cron relay — is `INSERT OR REPLACE`. The claim is now scoped to the HTTP route, with the relay's semantics stated.
- **Three caller lists elsewhere in the repository went stale** (docs-drift, Blocking ×3). This change makes the notifier a fourth loopback caller of the Session KV server; `agents/platform/docs/session_management.md` said "exactly three callers", and the canonical `docs/credential-isolation-design.md` and its site page both enumerated the same three. All four now name the notifier, matching the operator env comment this diff already updated.
- **"The three concerns" headed a five-item list** in the notifier's module docstring (docs-drift, Blocking). Already off by one at base; the count is gone rather than corrected, since the list is the thing being maintained.
- **`agents/platform/docs/session_management.md`'s "Check Persisted Incidents" command did not work at all.** It selected an `incidents.updated_at` column that does not exist (docs-drift, Advisory), and it shelled out to `sqlite3`, which the sandbox image does not ship — `exec: "sqlite3": executable file not found in $PATH`. Both pre-existing, and both found the same way: it is the exact command an operator checks this change with, so running it during live validation was step 3. The column is now `created_at` and the query goes through `python3 -c` against the interpreter that runs the server. The docs-drift pass caught the column and not the missing binary, which is the limit of reading prose against source — nothing in the tree says what is on the image's `PATH`.
- **The site had no description of the reply loop** (docs-drift, Advisory). The user-visible behaviour this change restores had no home on the published docs; `concepts/cluster-agents.md` now carries the paragraph.
- **The recent-reports index called every unlabelled row a scheduled report** (found reading #731 after the passes; see design call 2). Pre-existing and rare before this change, common after it.
- **The new log lines joined the chat and the thread with a slash** (found during live validation, by reading the line the change emits). A Google Chat `thread_id` is already `spaces/<space>/threads/<key>`, so `%s/%s` printed the space twice — `spaces/0sm7myAAAAE/spaces/0sm7myAAAAE/threads/vZ1nINPlz0Q` — and read like a malformed identifier the store had written rather than a formatting bug in the line reporting it. Slack's `thread_id` is a bare timestamp and would have read fine either way, which is how it got past both passes and the unit suite. Chat and thread are now separate fields at all three call sites, with a comment saying why. This is the one finding no pass could have caught from the diff: it needed the deployed line in front of a real Google Chat address.

Reported, not fixed:

- **The template's promise is unconditional; the row that honours it is best-effort** (adversarial, MEDIUM/PLAUSIBLE). Every triage card now ends with "reply 'apply'", and the row is written only if the shape gate passes, the route carries a `thread_id`, and the POST succeeds. Not fixable at the seam: the agent writes the report before the notifier knows whether the write will land, so a conditional bullet is not available. The failure modes are narrow — an event-triage delivery always has a thread, since the alert created it, and a 401 or a down Session KV server means the alert path that produced the card is broken too — and each logs a WARNING naming the card and the thread. That warning, and the fact that the invitation outruns its guarantee, are now stated in the architecture doc so an operator reading a "blank stare" report knows what to grep for.
- **The tested chain is one step shorter than the promise** (adversarial, LOW). The roundtrip test ends where the report reaches the agent; the bullet promises a GitOps Pull Request, which the front door's delegation to the Platform Agent produces. That delegation is documented and pre-existing, and exercising it needs the live install rather than a unit test.
- **PR-number prose in two design docs** (docs-drift, Advisory) — "Between #738 and #802 nothing wrote that row". Kept: these describe the state of `main` rather than the status of a pull request, #802 is an issue rather than a PR, and the repository cites issue numbers this way throughout (#560 and #658 in `AGENTS.md`, #630 in this same doc).

Refuted by the adversarial pass and left alone: extending `SENTINELS` with the new call name (any one of the three existing sentinels already blocks re-application); the verifier's `NOTIFIER_SOURCE.find` ordering checks (the file's pre-existing convention); report-at-rest exposure in `incidents` (`send_notification` already persists full reports there); and the roundtrip test's `DELETE FROM incidents` against a shared DB.

## Risk & Rollout

The change adds one loopback POST per delivered card that looks like a triage report, on a path that already made the delivery. It cannot fail the delivery: `store_incident_report` catches everything and returns `False`. The worst case is the status quo — a report delivered with no row behind it — plus a warning in the gateway log.

The one new user-visible behaviour is the restored `To authorize:` bullet, which reintroduces the promise the row now honours. If the row stops being written for any reason, the bullet should come out again rather than be left as a promise the system cannot keep; the `_triage_task_body` docstring says so at the site.

Revert is the commit: the applier stops emitting the call, the notifier keeps a function nothing invokes, and the template goes back to withholding the invitation. Nothing to do at merge time — the change ships in the agent image on the next build.


deploy/docker/patches/kanban_notifier.py
```diff
@@ -1007,3 +1020,277 @@ def note_suppressed_completion(
...
+# ``sub`` address rather than a value the send returns — but the other half of
+# that ordering holds identically here. A row written before the post claims a
+# reader has a report that has not been sent yet, and on a delivery that then
+# fails to send, claims it forever: ``INSERT OR IGNORE`` will not let the next
+# report replace it.
+#
+# Over loopback HTTP rather than SQLite, unlike
+# ``session_kv_server._store_incident_report``. That function writes the table
+# directly and its docstring says why: it *is* the server that owns the
+# database. The notifier is a different process, in a different container from
```
jayantid: This says the notifier runs "in a different container from the one holding the write." It runs in the same one. `deploy/shared/docker-entrypoint.sh:1294` starts the Session KV server and then execs `hermes gateway run` in that same `platform-agent` container, and this diff's own `k8s-operator/internal/controller/platformagent_manifests.go:1632` hunk says so: "This container both serves it and calls it (platform_mcp_server, incident_context, and the gateway's kanban notifier)."

So one diff now gives two opposite answers to which container the notifier is in, in the repository whose security story is which container holds which secret. The caller-list sweep also missed two enumerations, and both sit closer to the code than the four it fixed. `agents/platform/scripts/session_kv_server.py:137` — in a file this diff edits — still reads "Every caller — the event watcher, the MCP server, the incident_context plugin — gets the value from the same pod secret," in the docstring of `verify_api_key`, the function that rejects the notifier's POST. `deploy/shared/docker-entrypoint.sh:1295` still reads "Every caller — this container's MCP server and incident_context plugin, and the event watcher in the credential-proxy container," on the comment attached to the line that launches the server. An operator debugging a 401 or a 503 from the new POST greps for who is allowed to call 8699 and finds a three-caller list that omits the caller they are debugging. The PR body reports this class as closed: "All four now name the notifier."

Fix: drop "in a different container" from the clause here — different process is the whole reason, and it is already in the sentence — and add the notifier to the two lists above.

jayantid (approved):

deploy/docker/patches/kanban_notifier.py
stalhaali: done

Issue #745, opened 2026-08-18
feat(security): express the gVisor default at the CRD layer, and keep the chart's empty-string opt-out working


AntonTyb:
## Summary

`RuntimeClassName` has no default in the CRD. `k8s-operator/api/v1alpha1/common_types.go:482-484` declares it `+optional` and nothing else, so the generated CRDs carry a bare `type: string` (`k8s-operator/config/crd/bases/kubeagents.x-k8s.io_platformagents.yaml:1022-1026`, mirrored into `charts/kube-agents/crds/kubeagents.x-k8s.io_platformagents.yaml:1022-1026`). Each surface that wants the agent sandboxed states `gvisor` for itself, and a surface that states nothing leaves the agent on the standard container runtime.

The Helm chart states it — `charts/kube-agents/values.yaml:467` defaults `platformAgent.deployment.availability.runtimeClassName` to `"gvisor"`. The Terraform composition does not: `agent_runtime_class` defaults to `""` (`terraform/examples/full-install/variables.tf:45-49`), and `terraform/examples/full-install/main.tf:527` falls back to `enable_gvisor_node_pool`, which is false on Autopilot by force. A hand-written `PlatformAgent` does not either. The chart README states the consequence at `charts/kube-agents/README.md:378-382`: a bare `terraform apply` against the composition leaves the agent unsandboxed where a bare `helm install` sandboxes it.

A `+kubebuilder:default="gvisor"` reaches all three at once, which is the altitude the review that opened this issue argued for.

Raised by @mplakhtiy in review of #496 (https://github.com/gke-labs/kube-agents/pull/496#pullrequestreview-3795905540, "MEDIUM — the Helm chart cannot express this default at all").

## Status — updated 2026-09-03 (verified against main b0a4d7d1)

**Open:**

- Part 2, the CRD default. `common_types.go:482-484` is unchanged since filing.
- Part 2's blocking design question, which has moved rather than gone away. "Absent" and "explicitly empty" still have to become distinguishable, and the surface that needs them distinguished is now the chart's opt-out rather than the deleted `sed`. Detail under part 2.
- The Terraform composition's `""` default, which was not in the original text. Detail under part 4.

**Resolved since filing:**

[...]

Issue #610, opened 2026-08-10
fix(state): SQLite WAL on gVisor 9p mount corrupts state.db and kanban.db


adamparco:
## Summary

`state.db` and `kanban.db` run in SQLite **WAL mode on a gVisor 9p gofer mount**, which
corrupts them. Two independent databases corrupted within three days on the same volume
(`kanban.db` on 2026-08-07, `state.db` on 2026-08-09). The existing WAL-incompatibility
fallback in `hermes_state.py` does not fire on gVisor, because it detects unsafe
filesystems *reactively by error string* and gVisor raises none of them.

## Environment

- Pod: `platform-agent-gateway` (`kubeagents-system`), `runtimeClassName: gvisor`
- Image: `platform-agent:dev-3c4949a-0809-164704`
- Volume: PVC `platform-agent-data`, ReadWriteOnce, `pd.csi.storage.gke.io`, ext4
- SQLite: 3.53.4 (so *not* the WAL-reset bug [#69784](https://sqlite.org/wal.html#walresetbug), fixed in 3.51.3)

The PD is ext4, but the sandbox does not see it as ext4. Inside the gVisor sandbox:

```
none /opt/data 9p rw,trans=fd,rfdno=59,wfdno=59,aname=/,dfltuid=4294967294,
dfltgid=4294967294,dcache=1000,cache=remote_revalidating,disable_fifo_open,directfs 0 0
```

And WAL is active on `state.db` (header bytes 18/19):

```
write_version=2  read_version=2      # 2 = WAL
```

## Root cause

SQLite WAL requires mmap'd shared-memory (`-shm`) coherence and fcntl byte-range locks.
`hermes_state.py` documents this and falls back to `journal_mode=DELETE` on filesystems
that can't support it. But the trigger is a list of **error-string markers**:

```python
_WAL_INCOMPAT_MARKERS = (
    "locking protocol",       # SQLITE_PROTOCOL on NFS/SMB
    "not authorized",         # Some FUSE mounts block WAL pragma outright
    "disk i/o error",         # ZFS SHM corruption under concurrent connections
)
```

**gVisor's 9p gofer mount accepts `PRAGMA journal_mode=WAL` and returns success.** None of
those markers ever appear, so the DELETE fallback never engages. There is no `statfs` /
`f_type` filesystem-type detection anywhere in `hermes_state.py`, and no environment
variable to force DELETE.

Compounding factor: `_enforce_macos_synchronous_full()` is gated on
[...]

PR #258, opened 2026-07-07
feat(operator): add gVisor runtime class support and infrastructure provisioning

eLeontev:
## Why This Change

In Google Kubernetes Engine (GKE Standard), running autonomous AI agent workloads—such as the Platform Agent gateway, which executes multi-agent orchestration, web endpoints, and tool calls—requires kernel-level sandbox isolation to prevent container breakout vulnerabilities.

Previously, our custom kubebuilder Go operator lacked support for Kubernetes `runtimeClassName`, preventing users from declaratively scheduling agent containers inside GKE Sandbox (`gVisor`). Furthermore, our cluster bootstrapping infrastructure scripts created standard node pools without GKE Sandbox enabled. When users manually attempted to assign   `runtimeClassName: gvisor`, pods hung indefinitely in `Pending` state due to `FailedScheduling: 0/3 nodes match node selector`.

This PR bridges the gap between workload intent and cluster infrastructure by adding declarative gVisor runtime class support to our Go operator and automating gVisor node pool provisioning in our setup scripts.

## What Changed

The changes:

  - Independent Provision & Turndown Scripts: Created 2 independent scripts for provisioning (`provision_01a_gvisor_nodepool.sh`) and turndown (`teardown_01a_gvisor_nodepool.sh`) of gVisor
  node pools, allowing users to add or deprovision sandbox compute independently on demand without destroying their cluster.
  - Optional Master Integration: Integrated the gVisor provisioning script into the master provisioning pipeline (`provision.sh`) as an optional step prompted via `ENABLE_GVISOR` (y/n).
  - Controller RuntimeClass & Pod Validation: Extended the operator controller (`platformagent_controller.go`) to read gVisor support on the cluster. If a Custom Resource requests gVisor
  but the cluster does not support it (or if pods fail to schedule due to missing sandbox nodes / image pull errors), the controller automatically updates the CR status to `Degraded`
  (`Reason: PodUnschedulable` / `RuntimeClassNotFound`) with a proper, [human-friendly diagnostic message](https://screenshot.googleplex.com/9hKwJ9DKRE2cKBt).
  - Unit Test Coverage: Additional unit tests added (`TestPlatformAgentReconciler_Reconcile_MissingRuntimeClass`, `TestPlatformAgentReconciler_Reconcile_ExistingRuntimeClass`,
  `TestPlatformAgentReconciler_Reconcile_PodUnschedulable`) to cover scenarios with the optional flag and degraded status reporting.
[...]

PR #865, opened 2026-08-21
feat(install)!: sandbox the agent under gVisor by default

AntonTyb:
## Summary

The Platform Agent executes model-authored commands, and until now it ran on the standard container runtime unless someone passed `--gvisor=true`. Without a RuntimeClass it shares the node kernel with every other workload on the node, so a container escape is a node compromise. This flips the default: `install.sh` and the Helm chart both sandbox the agent under gVisor now, and `--gvisor=false` is the opt-out. Nothing about how the sandbox works changes — the Standard node pool, the Autopilot RuntimeClass, and the version floor are all as #823 left them.

## Why This Change

`RUNTIME-002` in the security audit: "gVisor is optional and off by default … agent pod runs on default node pool with no kernel-level sandbox." A security control nobody turns on is not a control, and this one had no cost that justified opt-in — Autopilot ships the RuntimeClass for free, and on Standard the installer already knew how to provision the pool.

The reason it stayed opt-in this long is that flipping it standalone breaks adopted Autopilot installs. #823 fixed that and has since merged, so this branch now stands alone against `main`.

## What Changed

The default lives on the **fresh-install path only** — the places where someone is choosing what to build:

- `install.sh` — `PARAM_ENABLE_GVISOR` defaults to `true`, `main` reads it as the install's choice, and the help banner says so.
- `charts/kube-agents/values.yaml` — `platformAgent.deployment.availability.runtimeClassName` defaults to `"gvisor"`. This is what a direct `helm install` gets; the Terraform composition always renders the value explicitly, so installs driven by `install.sh` never see it.

Two callers read an install that **already exists** rather than choosing one, and both deliberately keep the old `false` fallback:

- `run_menu_system` (the Day-2 control panel). Its Save & Apply re-applies what it displays, so a `vars.sh` with no `ENABLE_GVISOR` has to read as the standard runtime — that is what such a cluster is actually running. Defaulting on would have shown "gVisor Sandbox" for an unsandboxed install and then provisioned a node pool nobody asked for.
- `write_tfvars_from_state`. `uninstall.sh` treats `vars.sh` as optional — the documented `curl … | bash` teardown runs from a fresh clone that has none — and calls this bare under `set -e` before `lifecycle.sh destroy`. Defaulting on there would let the Autopilot version floor abort a **destroy** on an old cluster, leaving the install with no working way to remove itself. This was the highest-severity finding of the self-review below; the first version of this branch had the bug.

Five smaller fixes the flip made necessary or visible:

- The Autopilot floor's remedy named only `--gvisor=false`, a flag `uninstall.sh` and `upgrade.sh` both reject as an unknown parameter. It now names `ENABLE_GVISOR` in `vars.sh` for the callers that read state rather than flags.
- `--gvisor=` with no value slipped past the `true|false` validator and read as the default — previously "off", now "on", which is the direction that matters. `main` takes `${PARAM_ENABLE_GVISOR-true}` rather than the `:-` form so an empty value reaches the validator and is rejected.
- The interactive prompt is reordered and made conditional. `prompt_menu` answers an empty line with option 1, so the incoming value has to be listed first or the `(Default)` label contradicts what Enter produces. Option 2 is "the other one" in either direction, which also stops an explicit `--gvisor=false` from being silently re-enabled by someone confirming the prompt.
- **The post-apply health check now waits for each Deployment to exist** (`wait_for_deployment_object`, `install.sh:604`), 90s at a 5s interval. It previously asked `kubectl get deployment` once. The flip lengthens the operator's path to writing the agent Deployment — `validateRuntimeClass` returns `RequeueAfter: 30s` until the RuntimeClass resolves (`k8s-operator/internal/controller/platformagent_controller.go:265`) — so an unretried check reports a Deployment that is merely late as one that was never created. The budget is three requeues; below one, "not yet" and "never" are indistinguishable. When the one that does not appear is `platform-agent-gateway` and gVisor is on, the error now names the RuntimeClass as the likely cause and prints the `kubectl get platformagent … status.conditions` that says so, because that reason appears on the CR and in none of the logs an operator reaches for first.
- `scripts/release/validate_and_log_deploy_summary.sh` logged a bare `<NOT SET>` for `ENABLE_GVISOR`. Unset no longer means off — `provision_rc_environment.sh` omits `--gvisor` when the variable is empty and `install.sh` then applies its own default — so the line now says the installer default applies.

Docs — `README.md`, `INSTALL.md`, the chart README, `quickstart-gke.mdx`, `architecture.mdx`, and the `install-kube-agents` skill's flag table reframed from opt-in to opt-out. `INSTALL.md` and `architecture.mdx` also carry the capacity cost the flip introduces, led by the shape a fresh install actually creates: Autopilot since #1019, where the sandbox costs no extra nodes and needs only GKE `1.27.4-gke.800`. On Standard it is a `gvisor-pool` of one `e2-standard-4` per zone, three on a regional cluster, with no autoscaling. `docs/README.md` gains an identifier-source row for the installer's `--gvisor` default.

`INSTALL.md`'s re-run bullet is corrected rather than reframed. It said re-runs "reuse `k8s-operator/scripts/vars.sh` from the first run", which `main()` has never done — every value comes from the `PARAM_*` set at script scope, step 10 rewrites `vars.sh` from those, and the only `source "$vars_file"` on that path is at `install.sh:2161`, after the write. A re-run that omits a flag therefore writes that flag's default over whatever the install had. Harmless while the defaults matched the old shape; not once `--gvisor` defaults to `true`. The bullet's own "edit `vars.sh` and re-run" advice was wrong for the same reason. It now leads with `./install.sh --menu`, which does read the existing state, and keeps sourcing `vars.sh` as a fallback with its gap named: `write_state_var` emits `export` so a child `install.sh` sees the values, but the file records `MEMORY_PROVIDER` and `HERMES_DASHBOARD_ENABLED` while `install.sh` reads `MEMORY` and `ENABLE_WEBUI`, so sourcing alone reverts memory and the dashboard to their defaults.

Two callers inherited the chart's new default without meaning to. Both are fixed in `59398145`:

- **`hack/ci-deploy.sh`** renders the release with no `runtimeClassName` override, so the smoke pipeline's agent pod became sandboxed — and its step 7 reaches that pod over `kubectl port-forward`, which a GKE Sandbox pod refuses: the forward is established in the host-side CNI netns while the listener lives in the sandbox's own network stack (`scripts/exec_tunnel.py` is canonical). On a pool cluster with no `gvisor` RuntimeClass the pod does not schedule at all and the rollout gate times out. It now pins the standard runtime. `tests/test_ci_deploy_runtime_class.py` fails if the `--set` is dropped, and a second test in it says so if the chart default ever moves back, at which point the override is dead weight rather than load-bearing.
- **`install.sh`'s completion summary** printed that same `kubectl port-forward` as the only route to the Hermes dashboard. It now prints `scripts/hermes-dashboard-tunnel.py` — the `kubectl exec` relay — when the agent is sandboxed, which is the default, and the plain port-forward otherwise.

Two things are deliberately left alone, both of which a reviewer grepping for the old default will hit:

- **`install.sh:895` and `:2350`** still read `${enable_gvisor:-false}`, in the JSON report and the completion summary. These are fallbacks for an unset variable, not the policy default — `enable_gvisor` is a `local` in `main()` and bash's dynamic scoping means both always see it set, so neither fires. If one ever did, the value would be genuinely unknown, and a report that understates the security posture is safer than one claiming a sandbox that is not there.
- **The Terraform composition's own `agent_runtime_class` default** stays `""`. It derives from `enable_gvisor_node_pool`, which is `false` by default and cannot be flipped without breaking a direct `terraform apply` against a Standard cluster that has no sandbox node pool. The installer always writes both keys explicitly, so this affects only someone applying the composition by hand. That leaves a bare `helm install` sandboxing the agent where a bare `terraform apply` does not, which the chart README now states outright.

## Context

Closes part of #745 ("express the gVisor default at the CRD and chart layers, and retire the sed") — the installer and chart layers. It does **not** do the CRD layer: a `+kubebuilder:default="gvisor"` on `RuntimeClassName` is non-additive, because the API server would write `gvisor` into every existing CR that omits the field on its next apply, including on clusters with no sandbox node pool. #745 stays open for that, and nobody is assigned to it. Part 3 of #745 (retire the `sed`) is already moot — #797 deleted the scripts it referred to.

**Rebased onto `main` at `cf038a28`.** #823 merged (`fb702272`, 2026-08-24), so the diff is this change alone; the earlier instruction to review `79e93ce0..HEAD` no longer applies.

**#1019 (`feat(install)!: default a fresh cluster to Autopilot`) landed under this branch and changes the argument it makes.** Autopilot is now the shape a fresh install creates, and that is the shape where the sandbox is free — so the node-pool bill this branch documents is what the non-default Standard shape pays, not what a default install pays. `INSTALL.md` and `architecture.mdx` now lead with that. It cuts the other way for the Autopilot version floor, which used to gate an opt-in path and now gates the default one whenever the run adopts an existing Autopilot cluster.

#1019 is also where all three rebase conflicts came from, each on a line this branch and #1019 both edited: the `--gvisor` row of the `install-kube-agents` flag table, the `--gvisor` bullet sitting next to a rewritten `--cluster-mode` bullet in `quickstart-gke.mdx`, and adjacent insertions in `tests/test_install_script.py`. The two docs take `main`'s text with only the default change reapplied; the tests keep both sides. Worth noting for anyone reading the prompt code: #1019 independently adopted the same pattern this branch uses for the gVisor prompt — derive the menu order from the resolved default so option 1 is what pressing Enter produces.

### Two things to do before merging

1. **#610 must be resolved.** `fix(state): SQLite WAL on gVisor 9p mount corrupts state.db and kanban.db` is open, unassigned, and unmitigated, reproduced twice. Defaulting the sandbox on puts every new install on the code path that corrupts those databases, so merging this now would trade a High security finding for a reproduced data-corruption one. This is the blocker.
2. **Set `vars.ENABLE_GVISOR=false` on the RC deploy environment.** `rc-deploy-environment.yml` drives `scripts/release/provision_rc_environment.sh`, which omits `--gvisor` when that variable is empty — so after this merges, the release-validation environment adopts the new default and lands on #610's corruption path. Pinning the variable to `false` keeps that environment on the standard runtime until #610 closes. This PR does not change `provision_rc_environment.sh`: the script's job is to pass through what the environment configures, and hardcoding an override there would outlive the reason for it.

## Testing

- `tests/test_installer_common.py` — 32 tests, all passing (3 new, 1 rewritten). These run the real `write_tfvars_from_state` under a stubbed `gcloud`, so they exercise the shipped bash rather than a copy.
- `tests/test_install_script.py` — 47 tests, all passing (6 new here, the rest `main`'s): the `--gvisor` default, `--gvisor=` staying empty so the validator sees it, `prompt_menu` resolving an unanswered prompt to option 1 (the premise the prompt's ordering rests on), and three for `wait_for_deployment_object` — it returns as soon as the Deployment exists, it keeps asking for one that arrives late (stub fails twice, succeeds on the third call), and it gives up at the budget rather than hanging.
- `tests/test_ci_deploy_runtime_class.py` — 3 new tests: `hack/ci-deploy.sh` passes the empty `runtimeClassName`, the chart default is still `gvisor` (so the override is still needed), and `helm template` with that override renders a CR carrying no `runtimeClassName` at all — `compactFields` drops the parent block — rather than one with an empty string. The last skips where `helm` is absent.
- `tests/` as a whole — 741 tests, all passing on the current head (rebased onto `main` at `d509d975`; the jump from 592 is mostly `main`'s own new tests, plus 2 here for the teardown guard). `make test-python` additionally fails `admin_console/tests/`, `agents/platform/scripts/`, and `tests/integration/`, all on missing third-party imports (`fastapi httpx mcp dotenv plotly pydantic streamlit uvicorn`) that `make test-python-deps` installs and CI has. Pre-existing and untouched by this diff.
- `helm template` on the chart, three ways, re-run on the rebased head — `main` has since added Teams values and a typed `networkPolicy` to the same chart — to prove the value flows to the CR rather than the default merely matching what was already rendered:

  | values | `spec.deployment.availability` |
  | --- | --- |
  | defaults | `runtimeClassName: gvisor` |
  | `runtimeClassName=""` | block absent entirely (`compactFields` drops the parent) |
  | `runtimeClassName=kata-containers` | `runtimeClassName: kata-containers` |

  The distinct third value rules out a hardcoded string in the template.

- The teardown path, against a stubbed old-Autopilot cluster, driving the real `write_tfvars_from_state` under `set -Eeuo pipefail` exactly as `uninstall.sh` does:

  | `ENABLE_GVISOR` | result |
  | --- | --- |
  | unset | reaches the line where `lifecycle.sh destroy` runs; tfvars say `enable_gvisor_node_pool = false` |
  | `false` | same |
  | `true` | floor fires, `rc=1`, remedy names both `--gvisor=false` and `vars.sh` |

  The first row fails before the fix, which is how the self-review finding was confirmed.

- The interactive prompt, exercised across all six combinations of incoming value × answer, against the real `prompt_read`/`prompt_menu` extracted from `install.sh`. Enter confirms the incoming value in both directions; option 2 always flips it. Only the option-1 premise is pinned by a committed test — the branch itself is inline in `main()` and not callable in isolation.
- `make docs-check` passes on the rebased head, as does the `compactFields` all-unset check from `validate.yml` run by hand against the current chart. `bash -n` clean on `install.sh`, `uninstall.sh`, `upgrade.sh`, `installer_common.sh`, and `validate_and_log_deploy_summary.sh`. `make prettier-check` flags no file this branch changes (the 19 it does flag are untouched by this diff and fail on `main` too).

### Live validation

**Not live-tested**, and not reachable from this workstation. The only cluster available is a GKE Standard cluster in a sandbox project whose control plane sits on a private endpoint with no route from here, and whose DNS endpoint has `allowExternalTraffic` disabled — both `get-credentials` paths fail at the API server. `terraform` is not installed either, so `install.sh --dry-run` stops at the prerequisite gate before writing a single tfvar. No kube-agents install was reachable to exercise.

What stands in for it is above: the chart renders ran against the real chart and templates, and the installer tests and the teardown matrix run the real shipped functions. None of it reaches a pod, and the scheduling claim — that the agent lands on a `gvisor-pool` node — is the part with no substitute.

What a reviewer with a live install should exercise, in the order the change touches:

1. `install.sh --non-interactive` with no `--gvisor` and no `--cluster-mode` → the default path since #1019, so an Autopilot cluster: `vars.sh` records `ENABLE_GVISOR="true"`, `terraform.tfvars` gets `agent_runtime_class = "gvisor"` and `enable_gvisor_node_pool = false`, and no pool is created. Then the same run with `--cluster-mode=standard` → `enable_gvisor_node_pool = true` and a `gvisor-pool` node pool appears. Both shapes need checking; only one of them is what a bare install now builds.
2. The CR: `kubectl get platformagent -n kubeagents-system -o jsonpath='{.spec.deployment.availability.runtimeClassName}'` → `gvisor`.
3. The pod: `kubectl get pod -n kubeagents-system -l app=platform-agent -o jsonpath='{.items[0].spec.runtimeClassName}'` → `gvisor`, scheduled onto a `gvisor-pool` node.
4. Re-run with `--gvisor=false` and confirm all three go back — the direction that proves the mechanism rather than a coincidence.
5. `uninstall.sh` from a fresh clone (no `vars.sh`) against an Autopilot cluster below GKE `1.27.4-gke.800` → the destroy runs. This is the regression the self-review caught, verified here only against a stub.
6. The new wait, from the failing side: install with `--gvisor=true` against a Standard cluster with no sandbox node pool, and confirm the run spends 90s and then names the RuntimeClass rather than reporting the Deployment missing with no reason. Only the helper's own retry behaviour is covered by a test; the message it prints on a real install is not.

7. The completion summary with `--gvisor=true` → the dashboard line names `scripts/hermes-dashboard-tunnel.py`; with `--gvisor=false` → the plain `kubectl port-forward`. Then actually run the printed command against a sandboxed install and confirm the dashboard loads, which is the half no test reaches.

`hack/ci-deploy.sh` is the one change here that got a real live exercise, because the smoke pipeline runs it on every push. It had been failing on this branch and passed on `59398145`, against a real GKE cluster with a real chart install and a real agent pod — which also exercises the chart default itself, since the pin is what stops that default reaching the release.

No test artifacts were created on any cluster. A throwaway `KUBECONFIG` used for a read-only probe was deleted; the developer's own `kubectl` context was never modified.

## Self-Review

Five passes. The first ran `review-adversarial` and `review-docs-drift` in two subagents, each given only the branch diff and its skill file — not the reasoning that produced the change. The second re-read the branch against `main` after #823 merged, in a session that had not written it, looking for what the merged work invalidated. The fifth is @mplakhtiy's adversarial pass over the branch merged onto `main` at `d509d97`, run in a context that did not write the change; both of its findings held on verification and both are fixed in `dc58e161`. The third is `kube-agents-bot`'s wide review of `5d79c181`; its three findings were verified against source before any of them was acted on, all three held, and all three are fixed in `59398145`. A fourth pass — the bot's strict re-read of `59398145`, requested by a reviewer — returned no findings.

The branch has since been rebased twice, most recently onto `main` at `cf038a28`. Each round's conflicts were resolved by re-reading the merge target rather than reapplying the branch's side, and the checks under Testing were re-run on the resulting head. The current round is the one described under Context: three files, all colliding with #1019.

That round also produced one finding of its own, from re-reading the branch against a `main` that had moved: **Medium — the cost argument named the wrong default.** #1019 made Autopilot the shape a fresh install creates, so `INSTALL.md` and `architecture.mdx` were describing the Standard `gvisor-pool` bill as what a default install pays, and `architecture.mdx` said outright that the pool is "provisioned by default". Both now lead with Autopilot, where the sandbox is free. Fixed in `5d79c181`. Nothing else the rebase touched changed behaviour: `install.sh` and `installer_common.sh` merged without conflict, and the gVisor derivation still reads the probed `cluster_mode` rather than any default.

The adversarial and docs-drift passes have **not** been re-run against this head. Their findings are all in `install.sh` and `installer_common.sh` logic, which the rebase carried across unchanged, so re-running them would re-read code neither round altered; the doc reframing above is the part of this head neither pass has seen.

Angles covered across both: default-flip blast radius across every caller of the changed variables, the interactive and non-interactive prompt paths, teardown and upgrade paths, chart/composition/CRD layer consistency, empty-and-malformed flag values, timing and ordering against the operator's reconcile loop, test coverage of the new branches, doc prose against source, and whether the branch's own comments still describe the code after each change.

**Fixed**

- **Medium (reviewer) — the teardown guard only covered the case where `vars.sh` is absent.** `uninstall.sh:339-345` sources `vars.sh` whenever the checkout has one, and after the flip that file says `ENABLE_GVISOR="true"` on every new install; `write_tfvars_from_state` then runs the Autopilot floor check, which `return 1`s below the floor under `set -Eeuo pipefail`. So the ordinary teardown — from the checkout that installed — could not remove a sub-floor Autopilot install, which is the same failure the `false` fallback was written to prevent. The reachable route is the probe's own "could not read the GKE version" branch. Worse, the comment I added arguing that the fallback protects the teardown was overclaiming: it only protects the fresh-clone case. `uninstall.sh` now exports `ENABLE_GVISOR=false` before generating tfvars, the comment says what it actually covers, the floor's remedy names the teardown, and two tests pin it — one that an explicit `false` clears the floor on a sub-floor cluster, one that the export precedes the call rather than following it.
- **Medium (reviewer) — `helm upgrade` is a fourth way in, and this body said otherwise.** Confirmed on mechanism and confirmed as a body error: "Blast radius: new installs only" missed the path, and "Nobody gets a running-but-unsandboxed agent while believing otherwise" is exactly what it produces. Both sentences are rewritten above, and the chart README and `values.yaml` now carry the upgrade case. On the reviewer's third suggestion — a line in the `BREAKING CHANGE:` footer — the footer is on the branch commit and does not survive: `main` squash-merges to a subject-only commit message (`git log -1 --format=%B d509d975` is one line), so `calculate_next_version.sh` sees only the `!` in the title and `publish_github_release.sh --generate-notes` renders PR titles. The footer is there anyway for anyone reading the branch, but the chart README is the surface that actually reaches a chart upgrader.
- **High (bot) — the smoke pipeline inherited the sandbox and then port-forwarded to it.** `hack/ci-deploy.sh` passes no `runtimeClassName`, so the chart's new default reached the release; the job's readiness gate and its `/v1/responses` probe both go through `kubectl port-forward`, which cannot reach a sandboxed pod. This is what `pull-kube-agents-smoke-test` had been failing on: the job passed on `59398145` with the pin as the only change between the failing run and the passing one. Fixed by pinning the standard runtime there, with tests.
- **Medium (bot) — the installer told users to reach the dashboard a way the default now breaks.** The completion summary's only access command was `kubectl port-forward … 9119:9119`. It now prints `scripts/hermes-dashboard-tunnel.py` for a sandboxed agent.
- **Medium (bot) — the re-run recipe this branch added silently resets memory and the dashboard.** Sourcing `vars.sh` does not restore them: the file records `MEMORY_PROVIDER` and `HERMES_DASHBOARD_ENABLED`, `install.sh` reads `MEMORY` and `ENABLE_WEBUI`, and nothing bridges the two. `INSTALL.md` now leads with `--menu` and names the gap in the sourcing fallback. The underlying name mismatch is not fixed — see below.
- **High — the flip broke teardown.** Making `write_tfvars_from_state` default gVisor on meant `uninstall.sh` from a fresh clone (no `vars.sh`) hit the Autopilot version floor and aborted before `lifecycle.sh destroy`, and the remedy it printed named a flag `uninstall.sh` rejects. Confirmed by reproduction, then fixed by keeping the fallback `false` in the two callers that read existing state, and by rewording the remedy. This is the change's most important finding and the pass earned its keep on it alone.
- **High — the repository's own RC deploy is a re-run that rewrites its persisted choice.** `install.sh --non-interactive` never sources an existing `vars.sh`, so every re-run rewrites `ENABLE_GVISOR` from the flags it was given — and `provision_rc_environment.sh` omits `--gvisor` when the variable is empty. After this merges, the release-validation environment silently adopts the sandbox and lands on #610. Resolved as documentation, not code: the pre-merge action above pins `vars.ENABLE_GVISOR=false` on that environment. `provision_rc_environment.sh` is unchanged on purpose — an override hardcoded there would outlive #610.
- **Medium — the health check could not tell "late" from "never".** A single unretried `kubectl get deployment` ran against an operator whose path to creating the agent Deployment the flip had just lengthened by a 30s requeue loop. Fixed with `wait_for_deployment_object` and three tests. This is the item the first pass left under "Unresolved" as a plausible RuntimeClass availability lag; re-deriving it from `platformagent_controller.go` settled it, and the fix covers it whether the lag is the node pool's or the operator's.
- **Medium — `--menu` adopted gVisor with no prompt.** Same root cause as the teardown finding; the control panel would have displayed "gVisor Sandbox" for an unsandboxed install and provisioned a pool on Save & Apply. Fixed by the same split.
- **Medium — the chart README's three `helm install` recipes** now need a `gvisor` RuntimeClass and did not say so, 250 lines from the explanation. Added a note under Usage pointing at it.
- **Medium — `docs/README.md`'s new identifier-source row claimed `install.sh` for `--permission-set` and `--region`.** Both wrong: the row above already gives `k8s-operator/scripts/common.sh` for permission-set defaults, and `DEFAULT_REGION` lives in `installer_common.sh:18`, which `install.sh`'s own help defers to. A table whose purpose is one home per identifier cannot give two answers for one. Narrowed to `--gvisor`.
- **Blocking (docs-drift) — `architecture.mdx` still said the default node pool suffices** for the Platform Agent, which the flip makes false on Standard. Rewritten, and given the per-zone node cost.
- **Low — `--gvisor=` bypassed the validator.** Pre-existing, but the flip changed it from silently-off to silently-on. One character (`:-` → `-`) routes it into the existing error.
- **Low — the release deploy summary logged `<NOT SET>` beside a run that provisions a gvisor node pool.** Fixed in `validate_and_log_deploy_summary.sh`.
- **Low — a comment this branch invalidated.** Rewording the health check's error message made `installer_common.sh`'s comment, which quoted that message verbatim and asserted it named nothing about a RuntimeClass, untrue. Found by grepping the quoted string; rewritten in the same commit.
- **Low — no committed test for the prompt work.** Added three tests to `tests/test_install_script.py`.
- **Docs-drift — `INSTALL.md`'s pre-run behaviours list** had a gap the flip opened. Filled.
- **Docs-drift — `INSTALL.md` claimed `install.sh` re-runs reuse `vars.sh`.** Raised by the bot as the other half of the re-run finding above, and correct: the claim is false on any path, and the flip is what makes it dangerous. Corrected, with the source-first re-run that actually preserves an install's settings.
- **Docs-drift — the chart README named `agent_runtime_class`** where the composition actually passes the Helm value `runtimeClassName`. Corrected.

**Reported, not fixed**

- **Medium — three front doors, two answers.** `install.sh` and a bare `helm install` sandbox the agent; a bare `terraform apply` against the composition does not. Fixing it means flipping `enable_gvisor_node_pool`, which breaks `terraform apply` against a Standard cluster with no sandbox pool — a different change with a different blast radius. Stated outright in the chart README instead. I think this is the right call for this PR and a fair thing to disagree with me about.
- **`install.sh:895` and `:2350`** keeping `${enable_gvisor:-false}`. Both agents flagged them independently; both are unset-variable fallbacks in `main()`'s dynamic scope that never fire. Reasoning under "What Changed".
- **`docs/architecture/03` and `08` still call the sandbox "deferred".** Those are end-state architecture documents describing the target rather than what ships; `docs/README.md` §3 makes updating them advisory. Left alone.
- **Medium — `install.sh` reads `MEMORY`/`ENABLE_WEBUI` but persists `MEMORY_PROVIDER`/`HERMES_DASHBOARD_ENABLED`.** The root of the bot's third finding, and the wider fix would be to have the read side fall back to the persisted names. That changes the installer's environment-variable contract for two settings this PR does not otherwise touch, and needs its own tests and its own live validation of the memory path; doing it here would widen a default-flip into an installer-state change. The documentation this branch added is what made the mismatch reachable, so making that accurate is the part that belongs here. Worth a follow-up issue, which I have not opened.
- **`reference/security-and-iam.md` never mentions gVisor.** Pre-existing and true before this branch. It is a real gap and I did not open it, so it is not this PR's to close — worth a follow-up issue.

Nothing is left unresolved.

## Risk & Rollout

**Blast radius: new installs, plus any `helm upgrade` that does not pin the value.** Every install since #258 persists `ENABLE_GVISOR` to `vars.sh`, and both state-reading callers keep the `false` fallback, so an existing install re-run through `upgrade.sh` or `--menu` gets exactly what it had. The change is visible to a fresh `install.sh`, to a direct `helm install`, and — because `install.sh --non-interactive` rewrites `vars.sh` from its flags rather than reading it — to any pipeline that re-runs the installer without passing `--gvisor`. **A fourth way in, raised in review:** Helm merges the new chart's defaults for keys a release does not set, so an existing OCI-installed release that never named `runtimeClassName` adopts the sandbox on a plain `helm upgrade`. That is now stated in the chart README's Usage section and in the `values.yaml` comment. Adopting a pre-existing Standard cluster counts as fresh: there is no `vars.sh`, so the default applies and the `gke-cluster` module adds a `gvisor-pool` to that cluster. The repository's own RC deploy is one such pipeline, which is what pre-merge action 2 above is for.

**What a fresh install pays.** Nothing, on the default path: since #1019 a fresh install creates an Autopilot cluster, which ships the `gvisor` RuntimeClass with no pool to run it on. The requirement there is GKE `1.27.4-gke.800`, which the installer checks before applying anything (#823) and which any newly created cluster clears. The bill lands on `--cluster-mode=standard` and on adopting an existing Standard cluster: a `gvisor-pool` of one `e2-standard-4` per zone, three on a regional cluster, no autoscaling. That is real money for someone who did not ask for it, and it is now in `INSTALL.md` and `architecture.mdx` rather than only in the module — but it is no longer what the common case pays, which weakens the main argument against this default.

**The known-bad interaction is #610.** SQLite WAL on the gVisor 9p mount corrupts `state.db` and `kanban.db`, reproduced twice, unmitigated. This default puts every new install on that path. That is the merge blocker and the single thing to check before this lands — a reviewer who takes nothing else from this section should take that.

**Rollback.** `--gvisor=false` on the installer, or `runtimeClassName: ""` in chart values; both are one-line and take effect on the next apply. Reverting the commit restores the previous default with no state migration, since nothing about the on-disk `vars.sh` format changed. If the flip proves wrong after release, the revert is cheap.

**Failure mode if the sandbox is unavailable** is loud, not silent: the operator refuses to write the agent Deployment and reports `RuntimeClassNotFound` on the PlatformAgent, the installer's Autopilot check aborts before applying anything, and the post-apply health check now waits out the operator's requeue loop and then names the RuntimeClass as the reason. **One path is quiet, and it is the `helm upgrade` one.** The operator returns from its RuntimeClass check before `reconcileWorkload` (`platformagent_controller.go:256-267`), so on an existing release the previous unsandboxed Deployment is never touched: it keeps running, later CR changes go unapplied, and `helm upgrade` reports success. The signal is there — `.status` reads `Degraded` / `RuntimeClassNotFound` — but nothing puts it in front of whoever ran the upgrade. On a fresh install there is no Deployment to leave behind and the failure is loud as described.


install.sh
```diff
@@ -1716,7 +1722,11 @@ main() {
...
   fi
-  local enable_gvisor="${PARAM_ENABLE_GVISOR:-false}"
+  # ${VAR-default}, not ${VAR:-default}: PARAM_ENABLE_GVISOR is always set (see
+  # its declaration), so the only way it arrives empty is `--gvisor=` with no
+  # value. Substituting on empty would silently read that as the default; this
+  # form lets it reach the validator below and be rejected.
+  local enable_gvisor="${PARAM_ENABLE_GVISOR-true}"
   if [[ ! "$enable_gvisor" =~ ^(true|false)$ ]]; then
     print_error "--gvisor must be either true or false."
     exit 1
```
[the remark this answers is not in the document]
AntonTyb: Confirmed on both halves, and both are addressed.

**The mechanism is exactly as described.** On the rebased head: `main()` writes `ENABLE_GVISOR` at `install.sh:1972`, the only `source "$vars_file"` on that path is at `:2039`, and the dry-run exit is at `:2075` — so the write happens first and a `--dry-run` preview persists it. `--menu` is guarded with `${ENABLE_GVISOR:-false}` and this path is not, deliberately, but that asymmetry does not protect a bare `install.sh` re-run.

**The PR body no longer claims otherwise.** Risk & Rollout said "visible only to a fresh `install.sh`"; it now names the re-run as a third way in, and the Context section carries a pre-merge action to pin `vars.ENABLE_GVISOR=false` on the RC deploy environment — which is this repository's own instance of the problem, since `provision_rc_environment.sh` omits `--gvisor` when that variable is empty and would otherwise adopt the sandbox and land on #610.

**`INSTALL.md:198` is fixed** in fd91db08. The idempotency claim was false on any path, not just for gVisor, and the bullet's own "edit `vars.sh` and re-run" advice was wrong for the same reason. It now says re-runs rebuild `vars.sh` from that run's flags and environment, that omitted flags fall back to their defaults, and that sourcing `k8s-operator/scripts/vars.sh` first is what preserves an existing install — `write_state_var` emits `export`, so the child process does see the values.

Not changed: `install.sh` still re-derives from flags rather than reading prior state. Making a re-run read `vars.sh` is a larger behaviour change than this PR should carry, and the docs now describe what it does instead of what it does not.

bradhoekstra:

/review all

INSTALL.md
```diff
@@ -189,10 +195,14 @@ KUBE_AGENTS_STATE_BUCKET=auto ./lifecycle.sh apply
...
-  change configuration, use `./install.sh --menu` (Save & Apply re-applies through the same
-  engine), edit `vars.sh` and re-run, or edit your hand-written tfvars and re-apply.
+- `install.sh` re-runs rebuild `k8s-operator/scripts/vars.sh` from that run's flags and environment
+  rather than reading the previous one, then regenerate `terraform.tfvars` from it and let
+  `terraform apply` reconcile whatever changed. Anything you do not re-supply falls back to its
+  default — including `--gvisor`, which defaults to `true`, so a bare re-run moves an unsandboxed
+  install onto the sandbox. To re-run with what the install already has, source
+  `k8s-operator/scripts/vars.sh` first — its entries are `export`ed, so a child `./install.sh` sees
+  them. To change one setting, `./install.sh --menu` reads the existing state and Save & Apply
+  re-applies through the same engine; for a hand-driven install, edit your tfvars and re-apply.
```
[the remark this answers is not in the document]
AntonTyb: Confirmed and fixed in 59398145.

The mismatch is exact: `install.sh:86-87` read `ENABLE_WEBUI` and `MEMORY`, while step 10 persists `HERMES_DASHBOARD_ENABLED`, `MEMORY_ENABLED` and `MEMORY_PROVIDER` (`install.sh:2300-2303`). Nothing writes `MEMORY` or `ENABLE_WEBUI`, so sourcing `vars.sh` and re-running reverts the dashboard to off and memory to the file store.

The bullet now leads with `./install.sh --menu`, which reads the existing state correctly, and keeps sourcing as a fallback with the gap named — including which two flags you have to pass alongside it.

I did not fix the underlying name mismatch in `install.sh`. Having the read side fall back to the persisted names changes the installer's environment-variable contract for two settings this default flip does not otherwise touch, and would need its own tests and a live exercise of the memory path. What this branch introduced is the recipe, so making the recipe accurate is the part that belongs here. It is in **Reported, not fixed** in the PR body as worth a follow-up issue.

bradhoekstra (approved):

AntonTyb:
The bot's three findings from [its 14:11 review](https://github.com/gke-labs/kube-agents/pull/865#pullrequestreview-5067470065) are fixed in 59398145, and `pull-kube-agents-smoke-test` now passes on that commit.

That last part settles the High finding empirically rather than by argument. `hack/ci-deploy.sh` passed no `runtimeClassName`, so the chart's new `gvisor` default reached the smoke pipeline's release, and its readiness gate and `/v1/responses` probe both go through `kubectl port-forward` — which a sandboxed pod refuses. The job had been failing on this branch; pinning the standard runtime in that script is the only change between the failing run and the passing one. `tests/test_ci_deploy_runtime_class.py` fails if the `--set` is dropped, and a second test in it says so if the chart default ever moves back, at which point the override is dead weight rather than load-bearing.

The other two: the installer's completion summary now prints `scripts/hermes-dashboard-tunnel.py` for a sandboxed agent instead of a port-forward that cannot reach it, and `INSTALL.md`'s re-run bullet leads with `--menu` rather than sourcing `vars.sh`. On that third one I fixed the recipe and not the underlying mismatch — `install.sh` reads `MEMORY` and `ENABLE_WEBUI` while `vars.sh` records `MEMORY_PROVIDER` and `HERMES_DASHBOARD_ENABLED` — because bridging them changes the installer's environment-variable contract for two settings this default flip does not otherwise touch. It is in **Reported, not fixed** in the PR body as worth a follow-up issue. Say the word if you would rather it were in scope here.

@bradhoekstra — you held with `/hold AI Review` at 19:32, four minutes before the bot's strict re-read landed. That pass came back with no findings on 59398145 and the `AI Review` check is green, so if the hold was waiting on it, it is satisfied.

If it was not, the two items under **Two things to do before merging** are still open and neither is mine to close:

1. #610 (`SQLite WAL on gVisor 9p mount corrupts state.db and kanban.db`) is open, unassigned, and untouched since 2026-08-10. Defaulting the sandbox on puts every new install on that path, which is why I flagged it as the blocker rather than a caveat.
2. `vars.ENABLE_GVISOR=false` needs setting on the RC deploy environment before this merges, or release validation adopts the new default and lands on #610 too.

Happy for the hold to stay until both are dealt with — just want to know which one it is tracking.

charts/kube-agents/values.yaml
```diff
@@ -432,11 +432,22 @@ platformAgent:
...
+      # later). A Standard cluster needs the sandbox node pool that the
+      # gke-cluster module's enable_gvisor_node_pool creates; without it the
+      # operator refuses to write the agent Deployment and reports
+      # RuntimeClassNotFound on the PlatformAgent. Set "" to omit the field and
+      # run on the standard container runtime.
+      #
+      # The Terraform composition always passes this explicitly (from
+      # agent_runtime_class, which install.sh derives from --gvisor), so this
+      # default is what a direct `helm install` gets.
+      runtimeClassName: "gvisor"
```
mplakhtiy, line 455: **MEDIUM — a `helm upgrade` of an existing chart install adopts the sandbox, and on Standard leaves an unsandboxed agent running behind a Degraded CR**

Helm applies the new chart's defaults for any key the release's stored config does not carry. A user who installed from `oci://ghcr.io/gke-labs/kube-agents/charts/kube-agents` — the path `charts/kube-agents/README.md:55-67` documents — never supplied `runtimeClassName`, so a plain `helm upgrade` to this chart version starts asking for `gvisor`.

Failure scenario, on a Standard cluster with no sandbox node pool: the operator's step 10 (`platformagent_controller.go:260-271`) fails `validateRuntimeClass`, writes `Degraded / RuntimeClassNotFound`, and returns `RequeueAfter: 30s` **before** `reconcileWorkload` at step 11. The Deployment created by the previous reconcile is never touched, so the agent keeps running on the standard runtime — indefinitely, and with every subsequent CR change silently unapplied — while `.status` claims the sandbox. On Autopilot at or above the floor it is quieter but still unannounced: the pod restarts sandboxed and lands on #610's 9p path.

This contradicts two statements in the PR body. Risk & Rollout leads with "Blast radius: new installs only" and enumerates the ways in as a fresh `install.sh`, a direct `helm install`, and an installer re-run — `helm upgrade` of an existing release is a fourth. And the section's closing claim, "Nobody gets a running-but-unsandboxed agent while believing otherwise", is exactly what this path produces. The `BREAKING CHANGE:` footer on `5ed94d10` describes only the installer's node pool, so the generated release note will not warn a chart user either.

Confirmed on mechanism: both the Helm default merge and the operator's return-before-workload ordering are read from source. Not a code defect — the default is the point of the change — but the chart README's new note covers install and not upgrade.

Smallest fix: one sentence in `charts/kube-agents/README.md` under the same new note saying that an existing release upgraded without `-f` picks the sandbox up, plus a line in Risk & Rollout and in the `BREAKING CHANGE:` footer.

k8s-operator/scripts/installer_common.sh
```diff
@@ -657,6 +657,16 @@ write_tfvars_from_state() {
...
+  #
+  # The fallback stays false even though a fresh install now defaults to the
+  # sandbox. install.sh owns that default and always writes the result to
+  # vars.sh before sourcing it and calling this function, so the fallback here
+  # never decides a new install — it only decides the callers that read an
+  # install that already exists. uninstall.sh is the one that matters: vars.sh
+  # is optional there, and the documented `curl … | bash` teardown runs from a
+  # fresh clone that has none. Defaulting on for that caller would let the
+  # Autopilot version-floor check below abort a destroy, leaving an install
+  # with no working way to remove itself.
```
mplakhtiy, line 669: **MEDIUM — the teardown protection only covers the case where `vars.sh` is absent, and the flip makes the other case the default**

Anchored to this new comment; the line it argues for is `local gvisor_node_pool="false"` just below at `:670`.

The comment argues the fallback keeps `uninstall.sh` working because "`vars.sh` is optional there, and the documented `curl … | bash` teardown runs from a fresh clone that has none". True, and the new `test_tfvars_unset_gvisor_skips_the_autopilot_floor` pins it. But `uninstall.sh:339-346` sources `vars.sh` whenever it *is* present, and `uninstall.sh:452` then calls `write_tfvars_from_state` bare under `set -Eeuo pipefail` (`uninstall.sh:14`). Nothing between them overrides `ENABLE_GVISOR`.

After this change, `ENABLE_GVISOR="true"` is what every new install writes to `vars.sh` (`install.sh:2269`). So for anyone tearing down from the checkout that installed — the ordinary case, not the `curl` one — the floor check at `:707-710` is live, and against a sub-floor Autopilot cluster it `return 1`s and aborts the destroy before `lifecycle.sh destroy` ever runs. That is the same "install with no working way to remove itself" this fallback was written to prevent.

Reachable trigger: the version probe's own warning branch. `:705-706` prints "Could not read the GKE version … proceeding as though it supports GKE Sandbox" and continues, so an install against a sub-floor Autopilot cluster whose `gcloud describe` hiccups applies every GCP and Helm resource, then fails at the new 90s Deployment gate. The user now has a half-built install to remove, `vars.sh` says `true`, the probe works this time, and `uninstall.sh` refuses.

The remedy string that prints is also aimed at the wrong callers: "install.sh takes `--gvisor=false`, and upgrade.sh reads the choice from `ENABLE_GVISOR` in `k8s-operator/scripts/vars.sh`". Neither names the teardown, and `uninstall.sh` rejects `--gvisor` as an unknown parameter.

Confirmed on mechanism; the trigger is narrow but is a branch the code deliberately takes. Pre-existing since #823 — but opt-in then, and this change makes it the default shape, which is what puts it in scope.

Smallest fix: have `uninstall.sh` export `ENABLE_GVISOR=false` before calling `write_tfvars_from_state` (a destroy needs no sandbox and no floor), or gate the floor check on a flag the teardown does not set. Either way, add the caller to the remedy text.

mplakhtiy (commented):
Two findings from a full adversarial pass over this branch merged onto current `main` (`d509d97`), run in a context that did not write the change. Both are MEDIUM and neither is a defect in the code as such — the first is a blast-radius path the PR body states the opposite of, the second an incomplete guard that this change turns into the default shape.

Posting these two only; other findings from the pass are not included here.

mplakhtiy (approved):

charts/kube-agents/values.yaml
AntonTyb: Confirmed and fixed in dc58e161.

I verified the mechanism rather than taking it: `platformagent_controller.go:256-267` returns `RequeueAfter: 30s` from step 10 before `reconcileWorkload`, so on an existing release the previous Deployment is never touched — it keeps running unsandboxed, later CR changes go unapplied, and `helm upgrade` reports success while `.status` says `Degraded` / `RuntimeClassNotFound`. You are also right that this contradicts two sentences in the PR body; both are rewritten. "Blast radius: new installs only" now names `helm upgrade` as a fourth way in, and the failure-mode paragraph no longer claims nobody gets a running-but-unsandboxed agent — it says which path is quiet and why.

The chart README's Usage section carries the upgrade case now, next to the install note, and `values.yaml`'s comment says it too for anyone reading the values rather than the README.

One correction on the third part of the remedy. The `BREAKING CHANGE:` footer does not reach the release note, so extending it would not have warned a chart user either: `main` squash-merges to a subject-only commit message — `git log -1 --format=%B d509d975` is one line — so `calculate_next_version.sh` sees only the `!` in the PR title, and `publish_github_release.sh` runs `gh release create --generate-notes`, which renders PR titles. I added the footer to the branch commit anyway since it costs nothing and is right for anyone reading the history, but the chart README is the surface that actually reaches someone about to run `helm upgrade`. If you think the release note itself should carry this, that is a change to how notes are generated and I would rather do it separately than smuggle it in here.

k8s-operator/scripts/installer_common.sh
AntonTyb: Confirmed and fixed in dc58e161.

You are right, and the part that stung is that the comment you anchored to is one this PR added — it argues the fallback protects the teardown, and it only protects the teardown that has no `vars.sh`. `uninstall.sh:339-345` sources one whenever the checkout has it, nothing overrides `ENABLE_GVISOR` before `write_tfvars_from_state`, and after the flip that file says `"true"` on every new install. So the ordinary case — tearing down from the checkout that installed — is exactly the one the floor can abort. The "could not read the GKE version" branch is a real route there and I had not walked it.

Taking your first suggested fix: `uninstall.sh` now exports `ENABLE_GVISOR="false"` immediately before generating tfvars. I checked that forcing it is safe rather than merely convenient — `terraform destroy` destroys what is in state regardless of a resource's `count`, and the gvisor node pool goes with the cluster in any case, so the only thing these two values change on the teardown path is whether the floor check gets to abort. That reasoning is in the comment at the export rather than left implicit.

Also done: the comment on `local gvisor_node_pool="false"` now says what the fallback actually covers and points at the export, so it cannot be read as protecting the teardown on its own; and the floor's remedy gained a line naming the teardown, since as you say it addressed only `install.sh` and `upgrade.sh`.

Two tests, because a guard nobody pins is a guard that regresses: `test_tfvars_gvisor_off_clears_the_floor_on_a_sub_floor_autopilot` covers the composition that matters (explicit `false` against a sub-floor cluster clears the check, which the existing unset-case test did not reach), and `test_uninstall_forces_gvisor_off_before_generating_tfvars` pins the ordering — an export placed after the call would read as a fix and change nothing.

dshnayder:
This merged today with #610 still open, and the body names that as the blocker: "**#610 must be resolved.** ... This is the blocker." Flagging it rather than assuming it was cleared deliberately — as of `aca17e74` nothing on `main` mitigates it:

- `agents/chat/defaults/plugins/session_store/store.py:179` and `agents/platform/scripts/session_kv_server.py:352` still run `PRAGMA journal_mode=WAL` unconditionally.
- No `statfs` / `f_type` filesystem-type detection exists anywhere in the repo, so the reactive error-string fallback #610 documents still cannot fire — the gofer mount accepts the pragma and returns success.
- No commit references #610.

So a `helm install` at the new default (`charts/kube-agents/values.yaml:455`) puts a fresh install on the path #610 reproduced twice. Terraform installs are not there yet: `agent_runtime_class` defaults to `""` and derives from `enable_gvisor_node_pool`, which defaults to `false`.

One thing #610 does not cover, which I noticed while checking this. It names `state.db` and `kanban.db`, both owned by the harness. The two call sites above are owned by this repo and open `/var/lib/kube-agents/session/session_kv.db`, which the operator mounts from the `system-metadata` PVC (`platformagent_manifests.go:2244` and `:3229`) into the same agent Pod — the same class of PD-backed mount that gVisor presents as 9p. That is a third WAL database under the same hazard, and unlike `hermes_state.py` it is ours to fix directly: a filesystem-type check before the pragma, falling back to `journal_mode=DELETE`, is small and self-contained.

Two ways forward, both yours: revert the default until #610 closes, or land the fallback first and leave the default in place. I would rather not guess which.

Separately, pre-merge item 2 in the body — pinning `vars.ENABLE_GVISOR=false` on the RC deploy environment — is worth confirming was done, since `provision_rc_environment.sh` was left unchanged on purpose and that environment otherwise adopts the new default on its next re-run.

*Investigated with the help of Claude.*

PR #874, opened 2026-08-21
fix(operator): pin API_SERVER_KEY in the managed .env so the gateway agrees with its callers

### LINKED ISSUE #786: Gateway API auth is broken pod-wide: the key the gateway accepts is generated into .env at boot and nothing else knows it
## Summary

The Platform Agent's Hermes gateway validates `API_SERVER_KEY` against a value that **no caller in the
system knows**: a 64-character key that Hermes' own Docker stage2 hook generates into
`$HERMES_HOME/.env` on every boot. Everything that talks to the gateway API presents a different value
and gets `401 Invalid gateway API key (API_SERVER_KEY)`.

Measured on `kage-management` (`platform-agent-gateway-86945dd985-v9kzd`, image
`dev-cronrelay-20260818b`), `GET http://127.0.0.1:8642/api/sessions?limit=1`:

| Bearer presented                    | Length | sha256[:12]    | Result  |
| ----------------------------------- | ------ | -------------- | ------- |
| `os.environ["API_SERVER_KEY"]`      | 24     | `83e187b64c20` | **401** |
| `API_SERVER_KEY` from `/opt/data/.env` | 64  | `41127edb2b1c` | **200** |

This is not limited to loopback callers. The **external path through the credential proxy is broken by
the same cause**, because the proxy swaps in the same sentinel upstream.

## Why it happens — three individually-reasonable decisions colliding

1. **The operator sets `API_SERVER_KEY` to a sentinel, not a secret.**
   `k8s-operator/internal/controller/platformagent_manifests.go:1614` and `:2250` put the literal
   `cluster-internal-trusted` on the agent container and comment it as exactly that. The real
   credential is `API_SERVER_EXTERNAL_KEY`, held only by the credential-proxy sidecar. The premise is
   "the gateway listener is loopback-only, so anything already inside the pod is trusted, and the
   sidecar is the authenticating front door."

2. **Hermes prefers `$HERMES_HOME/.env` over `os.environ`.** `hermes_cli/auth.py:686-691`, deliberately
   — its comment is that a key rotation should not be shadowed by a stale shell export.

3. **Hermes' Docker stage2 hook generates a strong key into `.env` when the file does not already carry
   one.** On this deployment `/opt/data/.env` is rewritten ~67 seconds *after* pod boot (boot 15:49:59,
   mtime 15:51:06), so it is live behaviour on every start, not stale PVC leftover.

Result: the gateway authenticates against a value invented at boot inside the pod, that neither the
operator, the manifests, the Secret, nor the process environment ever sees.

## What is affected

**Every authenticated call to the Platform Agent API, in-pod and external.**

The `platform-agent` Service's port 8642 targets `8643`, the credential proxy's `proxy-api` listener.
That proxy authenticates the caller against `API_SERVER_EXTERNAL_KEY` and then re-signs the request
upstream with `AGENT_API_UPSTREAM_KEY`, which defaults to — and on this cluster is set to — the same
sentinel (`credential_proxy.py:1699-1703`). So the proxy hands the gateway a key the gateway rejects.
Probed from the sidecar against `http://127.0.0.1:8643/api/sessions?limit=1`:

- bogus bearer → 401 from the **proxy** (plain-HTML `BaseHTTPRequestHandler` error page)
- correct `API_SERVER_EXTERNAL_KEY` → 401 from the **gateway** (JSON `gateway_auth_error`)

The second line is the proof: the caller authenticated fine, and the failure is one hop further in.

Known call sites:

| Call site | What breaks |
| --- | --- |
| `agents/platform/scripts/session_kv_server.py` | **Already fixed** on `feat/cron-report-relay` (#731) via `_gateway_api_token()`. Both the cron-report relay turn and `trigger_agent_troubleshooter`. |
| `agents/platform/scripts/agent_common_server.py:62` | `resolve_agent_credentials`, backing the `call_agent` MCP tool — every inter-agent call. |
| `agentplugins/pubsub-platform/files/platforms/pubsub/adapter.py:1000` | `_run_turn_via_api` — the alert routes that run a turn on a named profile. |
| `agentplugins/gke-stockout-investigator/scenarios/lib/common.sh:603` | `_sessions_json`, the demo scenario watcher. |
| `credential_proxy.py:1702` (`AGENT_API_UPSTREAM_KEY`) | Everything arriving through the Service: `hack/ci-deploy.sh:126`, `hack/ci-eval-pr.sh:57`, the `bench/` harness, and any operator/admin tooling using the Secret's `API_SERVER_KEY`. |

`PLATFORM_API_URL` is unset on the live pod, so every in-pod caller defaults to
`http://127.0.0.1:8642` and hits the gateway directly.

## Why nobody sees it

Three reinforcing reasons, and the third is the one worth fixing on its own merits:

1. **The code is correct in isolation.** `os.environ["API_SERVER_KEY"]` is exactly what the operator's
   own comments say to use. Nothing in a diff, a manifest review, or a golden-manifest test can see
   that a file on the PVC decides the outcome. Same class as an unmanaged `config.json` overriding
   image-owned config.

2. **Every caller degrades quietly.** The cron relay caught the 401, posted the raw report, and
   returned `200 {"status":"delivered"}` — seven consecutive real `github-repo-watcher` runs failed
   that way in one pod's first two hours and the scheduler recorded every one as a clean delivery.
   `_sessions_json` swallows the failure with `|| echo '{}'`, which reads as "no sessions". The pubsub
   adapter logs and continues. `call_agent` surfaces it as a string the model paraphrases.

3. **Nothing exercises an authenticated route as a health signal.** The `platform-agent` container has
   **no liveness or readiness probe at all**, and `/healthz` is unauthenticated in any case. A pod can
   be 4/4 Running with its entire API surface returning 401.

## How it should be fixed

**Primary — make the gateway's key deterministic, so one value is true everywhere.** Seed
`$HERMES_HOME/.env` with the intended `API_SERVER_KEY` in `deploy/shared/docker-entrypoint.sh`,
*before* Hermes' stage2 hook runs, so its "generate one if absent" branch never fires. Then the
container env, `.env`, and the proxy's `AGENT_API_UPSTREAM_KEY` all agree and no caller needs to know
about `.env`. This also puts the file under the same image-owned force-sync discipline the entrypoint
already applies to other PVC files that must not drift.

One caveat that must be re-tested before this is implemented: writing the sentinel into `.env` was
tried by hand on 2026-08-18 and the API server then **never bound** (pod hung ~12 minutes, 8642
closed). That observation overlapped a window where the credential-proxy sidecar was missing from the
pod, which independently breaks every credentialed CLI shim in the agent container, so it may be
confounded. It needs one clean trial. If the gateway genuinely refuses a 24-character non-random key,
seed a proper generated secret instead and set the container env and `AGENT_API_UPSTREAM_KEY` to the
same value — the sentinel's value is not the point, its *agreement* is.

**Secondary — a shared resolver, if the boot-order fix proves impossible.** Lift
`_gateway_api_token()` out of `session_kv_server.py` into a helper every in-pod caller imports, and
teach the credential proxy's upstream swap to use the same resolution. This works (it is what #731
ships and proves live) but it spreads knowledge of `.env` across five call sites and leaves the
external path depending on the sidecar reading a file the agent container owns.

**Required either way — make this class of failure audible.**

- A startup self-check that calls one authenticated route with the configured key and fails loudly on
  401, rather than letting the pod report Ready with a dead API surface.
- A readiness probe for the `platform-agent` container that exercises it.
- No caller should convert a 401 into a success or an empty result. #731 fixes this for the relay
  (`"relay": "degraded"` plus an `[unrelayed]` marker in the posted message); `_sessions_json`'s
  `|| echo '{}'` has the same shape.

## Evidence

- Live probes above, on `dev-cronrelay-20260818b` / `platform-agent-gateway-86945dd985-v9kzd`,
  2026-08-18. Only key lengths and sha256 prefixes were recorded; no key material was printed.
- Seven consecutive relay 401s:
  `path='/api/sessions/cron-platform-github-repo-watcher-20260818/chat'` in that pod's gateway log.
- Background and the relay-scoped fix: `docs/designs/cron-report-relay.md`, section "Which key the
  loopback call presents", and #731.

Found while validating #731 (#719). That PR fixes only its own two call sites, deliberately — the rest
is this issue.

dshnayder:
## Summary

The Platform Agent's gateway has been rejecting its own callers. Hermes' stage2 hook generates a random 64-char `API_SERVER_KEY` into `$HERMES_HOME/.env` on any boot where that file exists and does not already carry one, and Hermes applies the PVC `.env` with `override=True` — so the value the gateway validates against is one that nothing outside the gateway process has ever seen. The operator's container env, the credential proxy's upstream key, the startup probe's curl and every in-pod loopback call all present the operator's sentinel and all get `401 Invalid gateway API key`. This pins `API_SERVER_KEY` in the managed `.env` the operator renders, which `load_hermes_dotenv` applies last with `override=True`, so the gateway ends up validating against exactly the value its callers send. Around the pin it adds the three things issue #786 asks for so the same failure cannot be silent again: a boot-time check that says which file wins when the two disagree, a shared resolver that reads the layers in Hermes' own order rather than trusting `os.environ`, and the removal of two callers that were converting a 401 into an empty result.

## Why This Change

Every authenticated call into the Platform Agent's API fails, and has been failing since the managed-scope `.env` started overriding the container env. The cron report relay cannot post; the credential proxy 401s upstream; `/v1/sessions` is unreachable from inside the pod. Nothing about this is loud: the relay logged and moved on, and the stockout scenario watcher turned "could not read the session list" into "no sessions yet" and waited out its 600-second timeout before reporting that the agent had done nothing — while the agent had in fact run the investigation to completion.

Since PR #733 the shape of the failure has changed but not gone away: `agentAPIProbe` curls an authenticated route, so a pod without the pin now CrashLoops instead of degrading quietly. That makes the pin load-bearing for startup, not just for the relay.

## What Changed

- **`k8s-operator/internal/controller/platformagent_manifests.go`** — `renderManagedEnv` now emits `API_SERVER_KEY=cluster-internal-trusted` unconditionally, as the first line, before any integration-conditional block. It is a `const` (`loopbackAgentAPIKey`) shared with the agent container's env, `AGENT_API_UPSTREAM_KEY` and the credential-proxy sidecar, so the four cannot drift apart in a future edit. The restructure is byte-equivalent to the old renderer apart from that line: the `platformStart` sentinel is exactly the old `len(lines) == 0` predicate, so a CR with no integrations still renders a one-line file rather than a stray empty one.
- **`deploy/shared/docker-entrypoint.sh`** — a boot-time check that compares the managed `.env`'s pin against this container's env. It reports and never exits: managed scope fails open by design, and failing the container would trade a degraded API for none. The two disagreement branches name different files on purpose — a managed file pinning some *other* value is the winner (it is applied last), while a managed file with no pin at all leaves the PVC's stage2 key on top. Told the wrong one, an operator reads the wrong file.
- **`agents/platform/scripts/session_kv_server.py`** — `_gateway_api_token()` resolves managed `.env` → PVC `.env` → environment, mirroring `load_hermes_dotenv`'s precedence exactly, and re-reads per call because stage2 rewrites the PVC file seconds after this process starts. A small hand parser rather than `load_dotenv`, which mutates `os.environ` and would make the precedence unobservable after the first call.
- **`agentplugins/.../scenarios/lib/common.sh`** — `_sessions_json` and `_task_after` no longer report a failed read as an empty one. Both now say so, once per run rather than once per 10-second poll.
- **Goldens** — one new line in `managed.env` for each platform-agent fixture.
- **Docs** — the PlatformAgent reference's managed-scope section now documents the pin and why it has to live in `/etc/hermes/.env`. Two adjacent statements were also wrong before this branch and would have contradicted it: `hermes.apiServerSecretRef` was documented, in both the CRD field comment and the rendered reference, as holding `API_SERVER_KEY`. It is the only source of `API_SERVER_EXTERNAL_KEY` and has never reached `API_SERVER_KEY` — which, left uncorrected, tells anyone reading `kubectl explain` that they can override the value this PR pins. CRD YAML regenerated with `make manifests` and `make chart-sync`; the diff is those two descriptions and nothing else.

## Context

- Fixes #786.
- Interacts with #733 (`agentAPIProbe`): that probe is what turns this bug from a silent degradation into a CrashLoop, and it is also what will now verify the fix on every boot.
- `docs/designs/cron-report-relay.md` §"Which key the loopback call presents" is updated. Its earlier claim that "the API server then declines to bind" is retracted there — that observation was confounded by a missing sidecar, not caused by the key.
- Merged `main` five times while open — 21 commits, then 7 (`0b38b89b`, the tree the test-install images were built from), then 5 (#840, #859, #727, #723, #426), then 11, then 2. No textual conflicts at any point, and none semantic: #806 adds `ImagePullSecrets` to `buildPodTemplateSpec`, not `renderManagedEnv`, and renders `nil` by default so the goldens here are unaffected; #738 adds a triage-delegation path to `session_kv_server.py` that deliberately does not do per-profile gateway auth, so both authenticated call sites still resolve through `_gateway_api_token()`. None of those five commits mention `API_SERVER_KEY`, `loopbackAgentAPIKey` or `renderManagedEnv`, and their whole golden delta is #840 swapping the dashboard container's `tcpSocket` readiness probe for an exec `curl` — disjoint from the pinned line. Merged twice more since (11 commits, then 2), again with no conflicts. Of those, only three files overlap this branch: `main`'s edits to `platformagent_manifests.go` and `docker-entrypoint.sh` are comment-only, its `session_kv_server.py` changes are all in the triage/kanban notifier area (#802, #862) and leave `_gateway_api_token()` and both call sites untouched, and #883's edit to `tests/test_docker_entrypoint.py` does not disturb `ApiKeyPinCheckTest` (re-run, 6/6). The live results below were captured at `0b38b89b`; everything merged since is code-reviewed and the full suite re-run green on the merged tree, but it is not re-validated on a cluster.
- Hermes behaviour was read at the pinned tag `v2026.8.13`: the generation block is `docker/stage2-hook.sh:445-458`, the precedence is `hermes_cli/env_loader.py:313-315` (user `.env`, `override=True`) through `:336` → `_apply_managed_env` at `:567` (managed, last, `override=True`).

## Testing

All re-run after merging `main`:

- `go build ./...`, `go vet ./...`, and `go test -count=1 ./...` — green, including the three regenerated goldens (re-formatted with `npx prettier@3 --write` to match the repo's YAML style).
- `make test-python` — all 21 directories green, **3927 tests, 6 skipped, 0 failures**. `scripts/` (352) and `tests/` (303) have to run in a real checkout: the build host's copy has no `.git`, and the release/install-script suites shell out to `git ls-files`. Largest suites: `agents/platform/scripts/` 1057, `deploy/docker/patches/` 922, `agents/platform/skills/fleet-audit/scripts/` 507.
- **A pre-existing flake in the operator suite, for the record.** `go test ./...` intermittently dies with `fatal error: concurrent map writes` in `apimachinery`'s `Scheme.AddKnownTypeWithName`, reached from controller-runtime's fake client (`addToSchemeIfUnknownAndUnstructuredOrPartial`) lazily mutating a shared `Scheme` while packages test in parallel. It is not this branch: a clean `main` checkout reproduces it **4 times in 20 runs**, a higher rate than this branch's 2 in 37, and the branch touches neither `testutil.go` nor the controller — only golden data. Filed as #918; not fixed here.
- **`pull-kube-agents-smoke-test` was red on a bench-task bug, not on this change — fixed on `main` by #893, which this branch has merged.** `agent-kanban-smoke` scored `VerificationCorrectness=0.5` because `report-states-the-probe-title` greps the agent's report for the literal phrase `devops-bench smoke probe`, while the task's own prompt asks only for *the assigned id, in one sentence* — so a precisely compliant answer fails by construction. The agent filed the card correctly (one successful `kanban_create`, exact title) and `OutcomeValidity` scored **1.0** on the very sentence `ToolInvocation` claims is missing. PR #904 fails with a byte-identical verification report, and the job is red on 8 of the 11 recent open PRs that have a result. The green→red step on this branch (`fb055d7b`→`321cdb5f`) is three files: a stderr fix in a scenario harness the bench never invokes, plus two comment-only YAML blocks — none of it in the built image. Full evidence in [this comment](https://github.com/gke-labs/kube-agents/pull/874#issuecomment-5399038556). **@jayantid's #893 fixes it on `main`** (merged 19:16 today), widening both presubmit prompts to ask for what their checks grade; this branch has merged that. The other registered task, `gpu-stress-test-diagnosis`, passed in our run, so `agent-kanban-smoke` was the only thing holding the job red. Worth noting the job is optional and absent from `main`'s required contexts, and 9 of the last 10 merged PRs merged with it red. Also checked, both clean: the `platform-agent-gateway-previous-crash.log` artifact is a torn-down-cluster `kubectl` timeout rather than a crash, and the gateway log has zero `401` / `Invalid gateway API key` lines across the whole run.
- `tests/test_docker_entrypoint.py` — 57 tests, including a new `ApiKeyPinCheckTest` that runs the shipped shell function itself over real files under `set -e`, covering agreement, absence, a different pin, a pin for other keys only, a commented-out pin, and that it never ends the boot.
- `agents/platform/scripts/test_session_kv_server.py` — the full precedence ladder for `_gateway_api_token()` and a new `TestManagedDotenvPath`.
- `bash -n` / `sh -n` on both shell files; the new scenario-lib logic was additionally exercised against a stubbed `kmgmt` in three modes (success, exec failure, non-JSON output) to confirm the warn-once latch holds across subshells.
- `make chart-check`, `make images-check`, `make docs-check-generated`, `docs-check-links`, `docs-check-terminology`, `docs-check-map`, and `prettier --check` over the changed docs and CRDs — all pass.
- All three images (`platform-agent`, `credential-proxy`, `k8s-operator`) built and pushed at tag `0b38b89be264643a8ed87aa40ec69cbca2463466` to a scratch registry for the test install below. Nothing was pushed to a release tag.

### Live validation

Validated end to end on **`kube-agents-pr786`**, a GKE Standard cluster stood up for this PR by
`install.sh` itself rather than by hand — same Terraform + Helm engine, `--memory file`,
`--permission-set read-only`, no chat integrations, all three images at
`0b38b89be264643a8ed87aa40ec69cbca2463466`. The live `kage-management` install was never touched:
the composition wires its Helm and Kubernetes providers to `module.gke_cluster` directly, not to
the caller's kubeconfig.

Three deviations from a stock install, none of them in this branch's diff:

- `TF_VAR_agent_service_account_id` and `TF_VAR_kms_keyring_name`, because the project already
  holds a `kubeagents-platform-gsa` and a `platform-agent-keyring` from the existing install.
  Without the first, `terraform apply` 409s — and worse, `terraform destroy` at teardown would
  strip eight project-level IAM bindings off the **live** agent's GSA.
- A two-file local patch exposing the IAM module's existing `service_account_id` at the
  composition level (it is not plumbed through today), defaulting to the current value so no
  existing install changes. Applied to the install source only, never committed here. Filed
  upstream as #919.
- `make mirror-images INCLUDE=third-party` had to run first. `third_party_image_registry` falls
  back to `image_registry`, so pointing `--registry-prefix` at a registry holding only the
  first-party images rewrites cert-manager's references onto it too and the install dies at
  `helm_release.cert_manager` with `ImagePullBackOff`. Working as designed and documented, but the
  failure names a Helm timeout rather than a missing mirror.

**One correction to the mechanism described above.** Stage2's generation block is gated on
`[ -f "$HERMES_HOME/.env" ]` — it appends a key only to a file that already exists. A fresh install
with no chat integrations has no `/opt/data/.env` at all, so the bug is *latent* there and the
container env is what the gateway uses. It latches the moment anything creates that file:
`/sethome`, onboarding, a chat integration seed. That is why #786 reads as intermittent, and why
the install that hit it was an established one.

So the validation below deliberately manufactures the precondition rather than waiting for it:
`touch /opt/data/.env`, restart, and let stage2 generate into it for real.

| # | Check | Result |
|---|---|---|
| 1 | Operator renders the pin | `platform-agent-config` key `managed.env` is exactly `API_SERVER_KEY=cluster-internal-trusted` — a single line, confirming the `platformStart` restructure renders a one-line file and not a stray empty one |
| 2 | The two files disagree | `/etc/hermes/.env` → the sentinel; `/opt/data/.env` → stage2's own freshly generated 64-char key (`4dd1310e…`) |
| 3 | The gateway accepts the managed value | `GET /api/sessions?limit=1` with the sentinel → **200** from the `platform-agent` container and **200** from the `envoy-credential-proxy` sidecar |
| 4 | …and rejects the PVC value | the same call with stage2's key → **401**. The managed layer wins, which is the whole fix |
| 5 | `agentAPIProbe` | pod `3/3 Running`, `0` restarts, no probe events — with the conflicting `.env` in place |
| 6 | The entrypoint check is silent | no `#786` warning in the container log on a correctly pinned boot |
| 7 | `_gateway_api_token()` resolves the managed layer | in-pod: `MANAGED_DOTENV_PATH=/etc/hermes/.env`, resolved token = the sentinel, while `/opt/data/.env` held the 64-char key |

**Negative control, on the same cluster and the same image.** To show the pin is what is doing the
work rather than something else about a fresh install, the pre-fix state was reproduced directly:
operator scaled to 0, `managed.env` emptied, pod restarted.

| Check | Pinned (this PR) | Unpinned (pre-fix) |
|---|---|---|
| sentinel → `/api/sessions` | **200** | **401** |
| stage2's PVC key → `/api/sessions` | **401** | **200** |
| `platform-agent` container | `Ready`, 3/3, 0 restarts | **not Ready**, 2/3 |
| startup probe | silent | `Startup probe failed: curl: (22) The requested URL returned error: 401` ×10 |
| entrypoint check | silent | `WARN: /etc/hermes/.env does not pin API_SERVER_KEY — with no managed pin the PVC's /opt/data/.env wins …` |

The warning names `/opt/data/.env` as the winner, which is the correct diagnosis for this branch of
the check — the file the operator should be told to look at.

**Both warning branches, in the shipped image.** The check was extracted from
`/usr/local/bin/agent-entrypoint` in the running container and driven over crafted files, so this is
the shipped text and not a copy: an agreeing pin is silent; a managed file pinning a *different*
value warns that the managed file wins; a managed file with no pin for this key warns that
`/opt/data/.env` wins; a missing managed file warns that nothing is pinned. All four return `0` —
the check never ends a boot.

**Restored and cleaned up.** Scaling the operator back to 1 re-rendered the pin on its own; the
`/opt/data/.env` test artifact was removed and the pod restarted to the as-installed state —
`3/3 Running`, 0 restarts, sentinel → 200, check silent.

**Torn down.** `terraform destroy` removed all 26 resources. The destroy plan was read before it
ran and contains no reference to `kage-management`, `kubeagents-platform-gsa` or
`platform-agent-keyring`; afterwards, `kage-management` is still `RUNNING` and its GSA still holds
all nine project role bindings. The `required` project services show as destroyed in state only —
they carry `disable_on_destroy = false`, so no API was disabled. No live-test lease was taken
because nothing here touched the shared install; this ran on its own throwaway cluster.

## Self-Review

**Adversarial pass** — two independent **subagents** over `2b3f1e7e..adea58ee`, each handed the diff range and the issue and nothing else. They were asked to break four claims rather than confirm them: that stage2 really generates the key at the pinned tag; that managed scope really wins over the PVC `.env`; that the renderer restructure is behaviour-preserving for CRs with no integrations; and that nothing in the new shell can end a boot.

What they confirmed and I fixed, in the follow-up commit:

1. The entrypoint's mismatch message gave the wrong diagnosis — it said the PVC `.env` wins, which is true only when the managed file pins nothing for this key. Split into two branches that each name the file that actually wins.
2. `MANAGED_DOTENV_PATH` resolved a set-but-empty `HERMES_MANAGED_DIR` to the relative path `.env` and then consulted it at the *highest* precedence — a stray file in the server's working directory would have become the bearer token. `managed_scope.py` treats set-but-empty as unset; the resolver now matches. This was invisible to the existing tests because they patch the constant away, hence a separate `TestManagedDotenvPath`.
3. The new scenario-lib warning printed on all sixty polls of a 600s watch and was half-erased by the `\r` progress counter between them; latched to once per run, via a file because the readers run in subshells.
4. `_task_after` still had the exact swallow shape `_sessions_json` was being fixed for. Since kanban is the dispatch path for specialist routes, fixing only `_sessions_json` would have made the watcher honest for `dispatch: api` and left it lying for the other half.

Both passes reported claim (1) — the stage2 generation — as unverifiable, because the hermes-agent checkout each had was older than the pinned tag. That was an artifact of their environment rather than a finding; I fetched `v2026.8.13` and read the block directly, and it is quoted in Context above.

Knowingly not fixed:

- **Version skew.** A new operator against an old agent image re-creates the 401, because the old image's entrypoint has no check and its `session_kv_server` does not read the managed layer. No code in this PR can reach an old image, so this is an ordering requirement rather than a defect; it is in Risk & Rollout.
- **`grep -F` and a value containing a newline** would make the pin check match on a substring line. The value is a compile-time constant with no newline in it, and a change that gave it one would fail the golden tests first.
- **`hermes send` bypasses managed scope** (`hermes_cli/send_cmd.py:264-281` loads `$HERMES_HOME/.env` directly). Verified at the pinned tag: real, upstream, pre-existing — and it never reads `API_SERVER_KEY`, so it is not this bug. It does mean the Google Chat pins are not enforced on that path, which is worth an upstream issue but not a change here.
- **`_sanitize_env_file_if_needed` runs against the read-only `/etc/hermes` mount.** Verified inert: it rewrites only when sanitising changed something or a UTF-32 BOM was found, our pinned line is clean ASCII, and the write is wrapped in `except Exception: pass` regardless.
- **The pubsub adapter reads `os.environ["API_SERVER_KEY"]` raw** rather than going through the new resolver. Left alone deliberately: after the pin, the environment holds the value the gateway accepts, so that path is correct by construction — and it was one of the paths this PR *fixes*.
- **`charts/` and `INSTALL.md` call the credential proxy's external secret `API_SERVER_KEY` too**, which is a genuinely confusing name for a different key. Renaming it is a breaking change to an installed secret and does not belong in a fix PR.

**Docs-drift pass** — run over the prose docs and CRD surfaces naming `API_SERVER_KEY` or the managed scope, under the contract #810 landed while this branch was open. It searched `.md` and the generated CRD YAML and **missed the sample CR and its fixture**, which carry the same misstatement in a YAML comment; @mplakhtiy's review caught it and it is fixed below. An earlier revision of this section claimed the pass covered "every doc", which was overstated. Dispositions:

1. **Fixed — real drift this branch caused.** The PlatformAgent reference's managed-scope section described `/etc/hermes/.env` as carrying only "the platform credentials and endpoints that have no `config.yaml` equivalent". This branch puts a non-credential in it. The section now documents the pin and the precedence that makes it necessary.
2. **Fixed — pre-existing, but it would have contradicted this PR.** `hermes.apiServerSecretRef` documented as holding `API_SERVER_KEY`, in both the CRD field comment and the rendered reference table. It has only ever fed `API_SERVER_EXTERNAL_KEY`. That description ships in the CRD, so `kubectl explain` was telling operators they could override the value this PR pins.
3. **Checked, no drift.** `docs/credential-isolation-design.md` and `agents/platform/docs/session_management.md` both already describe `API_SERVER_KEY` correctly as the non-secret sentinel — they were written against the intended behaviour, which is what this PR finally makes true. `docs/designs/cron-report-relay.md` is rewritten in this branch and re-read as accurate. `docs/site/.../reference/config.md` shows `API_SERVER_KEY` in a `config.yaml` `env:` block, a different file the pin does not change.
4. **Left alone, as above.** The `API_SERVER_KEY` naming of the credential proxy's external secret in `charts/` and `INSTALL.md`.

**Review findings from @mplakhtiy** (full adversarial pass over `upstream/main...4f09303`), both fixed:

1. **The sample CR still carried the misstatement this PR set out to correct.** `k8s-operator/examples/platformagent.yaml:32` and the fixture mirroring it at `internal/testing/testdata/platform/platformagent-tagged.yaml:32` both still read "ApiServerSecretRef securely references a Secret containing the API_SERVER_KEY" — the exact sentence corrected in `common_types.go` and both CRD YAMLs. It matters more in the sample than in the field table, because `INSTALL.md` and `k8s-operator/README.md` tell users to `kubectl apply -f examples/platformagent.yaml`, and the CRD reference page links it as **Sample** directly above the table this PR fixed. Both comments now say what the ref actually feeds.
2. **`_task_after`'s `2>&1` could turn a working kanban read into a broken one.** `out` is piped straight into `json.load`, so merging stderr in meant one benign CLI or `kubectl` line would fail the parse and report a live board as empty — the same failure this hunk exists to prevent, by a new route. The justification comment was also wrong about the old code: `2>/dev/null` discarded stderr *before* the pipe, so the old path did not have this exposure. Stderr now goes to a temp file read only on the failure branch, which keeps the diagnostic out of the parser's input; the comment is corrected and now says why this function cannot do what `_sessions_json` does (the program on the other end is `hermes kanban ls`, not one of ours to make print elsewhere).

## Risk & Rollout

Blast radius is the platform-agent Deployment: the new line changes the config hash, so applying this rolls every Platform Agent pod once. The failure mode it introduces is upgrade ordering — **the operator must not roll ahead of the agent image**. A new operator renders the pin into the managed `.env`; an old image's gateway will honour it (that is Hermes' own precedence, not our code), but an old image's `session_kv_server` still reads `os.environ` first and would keep 401ing on the relay path, and an old image has no entrypoint check to say why. Ship the image first, then the operator. In the other order — new image, old operator — nothing regresses: the resolver falls back to the environment exactly as before.

**Who is actually affected.** Because stage2 only appends to a `$HERMES_HOME/.env` that already exists, a brand-new install with no chat integrations does not have the bug — it acquires it the first time anything writes that file. So the population at risk is established installs and anything that has run `/sethome` or an onboarding path, not day-one deployments. That was confirmed on the test cluster: `/opt/data/.env` did not exist until it was created deliberately.

Reverting is a revert: drop the line from `renderManagedEnv` and the pods roll back to the pre-fix behaviour, which is the bug, not a worse state. The one thing to know is that with #733's `agentAPIProbe` in place, a pod that ends up unpinned now CrashLoops rather than serving a degraded API — louder, and intentionally so, but it means a partial rollback of this PR alone (goldens without the renderer, say) fails the pod outright instead of quietly.


dshnayder:
### CodeQL alert #19 dismissed as a false positive

The `CodeQL` check on this PR failed with one high-severity
`go/weak-sensitive-data-hashing` alert at `platformagent_manifests.go:3264`
("Sensitive data (password) is used in a hashing algorithm (SHA256) that is
insecure for password hashing"). I've dismissed it as a false positive. Recording
the reasoning here so the dismissal is reviewable, since the 280-character limit on
a dismissal comment does not fit it.

**The sink is not password hashing.** Line 3264 is `getConfigMapHash`, which
SHA256s `json.Marshal(configMap.Data)` to produce a fingerprint. Its only consumers
are the four call sites in `platformagent_controller.go` that stamp the hash onto
the pod template so a config change rolls the Deployment. Nothing stores, compares
or verifies a credential against it, so the property the rule protects — that an
attacker holding the digest cannot cheaply brute-force the input — is not one this
code needs. SHA256 is the right primitive for a change-detection fingerprint, and
`renderManagedEnv` is deliberately written in fixed key order precisely so this
hash stays stable across reconciles.

**The "password" is public by design.** The taint is this PR's new line,
`add("API_SERVER_KEY", loopbackAgentAPIKey)`. CodeQL sees a hardcoded literal bound
to an `API_SERVER_KEY` identifier and infers a credential. The value is
`cluster-internal-trusted`, the non-secret loopback sentinel — it is a compile-time
constant in this repository, it is documented as non-secret in the CRD field
comments and the PlatformAgent reference, and it appears in this PR's title. The
real credential on that path is `API_SERVER_EXTERNAL_KEY`, which comes from a
Secret and never enters this ConfigMap.

**No adjacent real problem.** I checked whether the alert was pointing at something
true by accident — credential material sitting in a ConfigMap. It is not: the rest
of `managed.env` is relay URLs, project IDs, allowed-user lists and booleans.

The rule fires nowhere else in the repository and has no prior dismissals, so this
does not paper over an established pattern. Happy to reopen and take a different
approach if a maintainer would rather see one.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -612,22 +661,44 @@ _task_after() {
...
+    # board with nothing on it into one answer — no output — which watch_investigation
+    # reads as "nothing has happened yet" until it times out. The two exec paths are the
+    # only ones this watcher has, so leaving this one silent would have fixed the report
+    # for `dispatch: api` routes and left `dispatch: kanban` routes lying.
+    local since="$1" out
+    # 2>&1 rather than discarded: on the failure path the command's own message is the
+    # only account of what went wrong, and on the success path anything it wrote to
+    # stderr would have broken the parse regardless — now it says so instead.
+    if ! out="$(kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
+        env HOME=/tmp hermes kanban ls --json 2>&1)"; then
```
mplakhtiy: 🟠 **Medium** — this `2>&1` can turn a working kanban read into a "broken" one, and the comment justifying it is wrong about the code it replaced.

The justification says that "on the success path anything it wrote to stderr would have broken the parse regardless." That is not true of the old code. `2>/dev/null` discarded stderr *before* the pipe, so anything `hermes kanban ls --json` wrote there vanished and the JSON on stdout parsed cleanly. With `2>&1`, a single benign stderr line — a `kubectl` notice, a loguru or CLI log line from the hermes shim — is prepended to the JSON, `json.load` raises, python exits 3, and the board is reported empty with a "did not return JSON" warning.

Corroboration that stdout-only is the intended contract: the repo's other reader of the same command, `agentplugins/pubsub-platform/tests/dedup_e2e_test.py:330` → `in_pod` → `run()`, uses `subprocess.run(capture_output=True)` and parses `.stdout` alone.

**Failure scenario:** on a cluster where the hermes CLI logs anything to stderr, `watch_investigation` reports "no new session or board task after 600s" for a `dispatch: kanban` route that in fact ran to completion — the exact failure this hunk exists to prevent, arriving by a new route.

What keeps this at Medium rather than High: the failure is now loud (one `warn_once` line naming the output) rather than silent, and the blast radius is a manually-run scenario harness rather than shipped agent code. Worth noting it was not exercised against the real command either — the testing note says the new logic was driven against a stubbed `kmgmt` in three modes.

The clean fix is to keep the two streams apart: redirect stderr to a temp file and read it only on the failure branch, which preserves the diagnostic this change is after without putting it in the parser's input. At minimum, correct the comment so the next reader is not told the old code had the same exposure.

mplakhtiy (commented):
Two MEDIUM findings from a full adversarial pass over `upstream/main...4f09303`. Neither blocks: the central mechanism holds up (the `platformStart` restructure is behaviour-preserving across all three exits, the four `loopbackAgentAPIKey` call sites agree, and pinning a public constant behind a `127.0.0.1` bind is not a security downgrade).

One finding is inline. The other cannot be — it is about a file this PR did not change.

#### 🟠 Medium — the sample CR still carries the `apiServerSecretRef` misstatement this PR set out to correct

`k8s-operator/examples/platformagent.yaml:32`, and the fixture mirroring it at `k8s-operator/internal/testing/testdata/platform/platformagent-tagged.yaml:32`, both still read:

```yaml
      # ApiServerSecretRef securely references a Secret containing the API_SERVER_KEY.
```

That is the exact sentence the branch corrected in `common_types.go:69`, in both CRD YAMLs, and in the site's field table — on the stated grounds that it "tells anyone reading `kubectl explain` that they can override the value this PR pins." The same reasoning applies with more force to the sample: `INSTALL.md:483` and `k8s-operator/README.md:114` tell users to `kubectl apply -f examples/platformagent.yaml`, and `docs/site/…/operator/platformagent-crd.md:13` links it as **Sample** from the very page whose table this PR fixed. A user copying the sample now reads the opposite of what the page above it says.

**Failure scenario:** an operator sets `hermes.apiServerSecretRef` to a Secret holding a strong key, expecting to replace the loopback sentinel. Nothing changes — that ref only ever feeds `API_SERVER_EXTERNAL_KEY` (`platformagent_manifests.go:2343-2352`). They now believe the gateway is guarded by their key when it is guarded by a public constant plus a loopback bind.

One comment line in each file. It also makes the Self-Review's claim that the docs-drift pass ran "over every doc that names `API_SERVER_KEY` or the managed scope" overstated, which is worth correcting in the body alongside the fix.

mplakhtiy (approved):

dshnayder:
Thanks @mplakhtiy — both findings were right, and both are fixed in 321cdb5f.

**The sample CR.** Fixed in `k8s-operator/examples/platformagent.yaml` and the
`platformagent-tagged.yaml` fixture. Your point that it matters *more* in the sample than
in the field table is the part I had backwards: the reference page links the sample as
**Sample** immediately above the table this PR corrected, so a reader met the correction
and its contradiction on the same screen, and the sample is the copy `INSTALL.md` and
`k8s-operator/README.md` tell people to actually apply. I searched `.md` and the generated
CRD YAML for the drift pass and never looked at YAML comments in sample manifests, which
is exactly the gap. The Self-Review said the pass ran "over every doc that names
`API_SERVER_KEY`" — overstated, and the body now says what it really covered and that your
review caught what it missed.

**`_task_after`'s `2>&1`.** Fixed, the way you suggested: stderr to a temp file, read only
on the failure branch. You were also right that the comment was wrong about the code it
replaced — `2>/dev/null` discarded stderr *before* the pipe, so the old path genuinely did
not have this exposure, and I wrote the justification as though it did. Worth noting the
asymmetry your comment implies, which the new comment now records: `_sessions_json` gets
to keep `2>/dev/null` because the program on the other end is an inline `python3 -c` of
ours, so its diagnostic is printed to *stdout* by design and survives the discard.
`hermes kanban ls` is not ours, so the file is the only way to keep the diagnostic without
feeding the parser.

On your note that the new logic was only driven against a stubbed `kmgmt` — still true,
and the fix does not change it. The throwaway cluster this PR was validated on has been
torn down, and this harness needs a live agent plus a populated board to exercise for
real, so I have not run it against the actual command.

Re-ran after the fix: `go build`/`vet`/`test ./...` green, `make test-python` green
(3927 tests), all four `docs-check-*` targets and `chart-check`/`images-check` pass,
`prettier --check` clean on both YAMLs, and `bash -n`/`sh -n` on the scenario lib.

mplakhtiy (approved):

dshnayder:
### `pull-kube-agents-smoke-test` failure on `321cdb5f` — pre-existing bench-task bug, not this PR

Short version: the failing check is `agent-kanban-smoke`, and it fails because the task's
`verification_spec` asks for something the task's `prompt` never asks the agent to do. It is
not a regression from this change, and it is reproducing identically on other open PRs.

**What actually failed** (`artifacts/results_agent-kanban-smoke.json`):

```
VerificationCorrectness = 0.5  (floor 1.0)
  [fail] report-states-the-probe-title    required phrases absent from the report: ['devops-bench smoke probe']
  [pass] the-kanban-card-was-actually-filed  1 call(s) to ['kanban_create'] (minimum 1)
```

The agent did the job. Its trajectory is a single successful `kanban_create` with
`title: "devops-bench smoke probe"`, and its answer was:

> I've filed that as task `t_1277311d`, and its progress will post right here.

**The contradiction** is in `bench/tasks/agent-kanban-smoke/task.yaml`. The prompt is:

> Create a kanban task titled exactly "devops-bench smoke probe", then state in one sentence
> the id that was assigned to it.

Nothing there asks the agent to echo the title back — it asks for *the id*, in *one sentence*.
But `report-states-the-probe-title` greps the final report for the literal phrase
`devops-bench smoke probe`. So an agent that follows the prompt precisely — one sentence, the
id, no restatement of the title — deterministically fails the check, and the more tersely it
complies the more certainly it fails. The task comment justifies the check with "We chose the
title, so an exact match is fair", which is true of the *card* the agent filed; it is not true
of the *report*, which is the only thing `report_contains` looks at.

The two LLM judges disagree with each other on the same text, which is the tell:

| metric | score | judge's reason |
|---|---|---|
| `OutcomeValidity` | **1.0** | "The actual output consists of exactly one sentence that clearly communicates the assigned ID." |
| `ToolInvocation` | 0.5 | "the output does not include a final message stating the assigned task ID in one sentence" |

Same one sentence, same ID; one judge sees it, the other does not.

**Not this PR.** Two independent lines of evidence:

1. The green→red transition on this branch is `fb055d7b` (smoke `success`) → `321cdb5f`
   (smoke `failure`), and that diff is 3 files: a stderr-handling fix in
   `agentplugins/gke-stockout-investigator/scenarios/lib/common.sh` (a scenario harness the
   bench does not invoke) and two comment-only blocks in `examples/platformagent.yaml` and its
   golden fixture. None of it is in the image the smoke test builds, and none of it is on any
   code path the bench exercises.
2. PR #904 fails with the *byte-identical* verification report — same check, same
   `VerificationCorrectness = 0.5`, same missing phrase. Across the 14 most recent open PRs,
   the smoke test is red on 8 of the 11 that have a result at all.

**Two non-issues I ruled out while looking**, both of which would have been on this PR if real:

- `platform-agent-gateway-previous-crash.log` is not a crash. Its entire contents are
  `Unable to connect to the server: dial tcp 34.125.124.72:443: i/o timeout` — that is the
  artifact-collection `kubectl` reaching for an already-torn-down cluster, not the gateway
  dying.
- `platform-agent-gateway.log` contains **zero** occurrences of `401`, `Invalid gateway API
  key`, `unauthoriz*`, or `API_SERVER_KEY`. Given this PR exists to kill exactly that 401, a
  clean gateway log across a full smoke run is a (small, incidental) point in its favour.

One cosmetic wrinkle worth noting for whoever owns the bench: the run also recorded
`delegated tasks did not finish within 1800s: t_1277311d (running)` while still reporting
`status: success` for the task. That did not affect the score, but a delegated card that never
lands inside the 30-minute budget makes the timing of these runs (1859s for a one-tool probe)
worth a look independently.

**Suggested disposition:** `/retest` will flip this green only if the agent happens to phrase
its answer with the title in it, which is a coin toss rather than a fix. The durable fix is in
the task, and belongs in its own PR — either widen the prompt to ask the agent to name the task
it filed, or point the check at the filed card instead of the report. Happy to file that issue
if it isn't already tracked.

dshnayder:
Filed the three pre-existing problems this PR's CI kept surfacing, so they stop being folklore attached to one branch:

- #917 — `agent-kanban-smoke`'s `report_contains` check demands a phrase its own prompt never asks the agent to say. This is the red smoke run above. Distinct from #863, which was the same job failing at the deploy step.
- #918 — `concurrent map writes` fatal error in the operator suite, 4 in 20 runs on clean `main`. `Run Controller Tests` **is** a required check, so this one can genuinely block a PR until it's re-run.
- #919 — `full-install` never plumbs `service_account_id`, so a second install in a project adopts the live agent's GSA and its `terraform destroy` revokes that agent's project IAM. This is the local patch already disclosed in the Testing section.

None of the three is caused by this change, and none blocks it: the smoke job is not in `main`'s required contexts (`cla/google`, `actionlint`, `build`, `prettier`, `validate`, `Run Controller Tests`).

dshnayder:
Pushed `03e10ad8`, which fixes #917 on this branch so the smoke job can actually go green.

`bench/tasks/agent-kanban-smoke/task.yaml` — the prompt now asks the agent to state **both the title it used and the id**, matching what `report-states-the-probe-title` greps for. `expected_output` follows, and the comment above the spec now records *why* the prompt and the check are coupled, so a future narrowing doesn't silently re-break it.

I considered the stricter fix — assert the title on the filed card rather than on the report — and did not take it: the bench registers exactly two verifiers, `report_contains` and `tool_called`, and `ToolCalledVerifier` matches tool *names*, not arguments. That fix needs a new verifier, which is a bench change that doesn't belong in a PR about `API_SERVER_KEY`. Left on #917 as the follow-up.

Checked before pushing: `gpu-stress-test-diagnosis` **passed** in the same run (`[PASSED] exact checks green; OutcomeValidity recorded: 1.0`), so `agent-kanban-smoke` was the only thing red. The gate in `hack/ci-eval-pr.sh` reads only `VerificationCatastrophic`/`VerificationCoverage`/`VerificationCorrectness` — `OutcomeValidity` and `ToolInvocation` are recorded, not blocking — so the one unsatisfiable check was the whole failure. `scripts/test_task_registration.py` (4) and `scripts/test_domain_coverage.py` (4) pass, and `prettier --check` is clean.

#918 and #919 are deliberately **not** fixed here; neither is caused by this change and neither is a bench/eval concern.

dshnayder:
Correction to my last comment: #893 already fixes this, and better. I've dropped `03e10ad8` from the branch and merged current `main` instead — head is now `bcc62d3c`.

@jayantid's #893 merged at 19:16 today, so the fix is on `main`. It's the same diagnosis and a wider fix: it corrects **both** presubmit prompts, not just the kanban one, and it caught a second mismatch I hadn't — `gpu-stress-test-diagnosis` question 2 asked for "any cluster or workload autoscaling limitations", which invited the Cluster Autoscaler answer when the planted defect is HPA max-replica saturation. It also reasons through the id-can't-be-an-objective problem explicitly (`report_contains` has no pattern field, and the only available substring `t_` would grade spelling), which is exactly the gap I'd left open on #917.

My duplicate search before filing #917 covered issues and not open PRs, which is how I missed it. #917 is closed as a duplicate.

This branch now carries no bench changes at all — the diff is back to the operator, entrypoint, docs and test files for #786. Operator `go build ./...` and `go test -count=1 ./...` are green on the merged tree.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
dshnayder: Fixed in `321cdb5f`, taking the clean fix you described rather than the minimum.

`_task_after` now sends stderr to a temp file and reads it only on the failure branch, so nothing but the command's stdout reaches `json.load`:

```bash
err="$(mktemp)"
if ! out="$(kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
    env HOME=/tmp hermes kanban ls --json 2>"$err")"; then
    errtext="$(head -c 200 "$err")"
    rm -f "$err"
    warn_once tasks "could not read the agent's kanban board; treating it as empty — ${errtext:-${out:-no response}}"
    return 0
fi
rm -f "$err"
```

So the diagnostic this hunk was after survives on the failure path, and the success path is back to the stdout-only contract you point at in `dedup_e2e_test.py`.

The comment is rewritten too — you were right that it was wrong about the code it replaced, and that mattered more than the bug itself, since it would have talked the next reader out of the fix. It now says plainly that stderr cannot be discarded on the failure branch but must not be merged into stdout either, and notes the distinction from `_sessions_json`: there the program on the other end is ours and prints diagnostics to stdout by design, so its `2>/dev/null` is correct and stays.

Your other finding — the sample CR and its fixture — is fixed in the same commit, and I corrected the Self-Review's "every doc" claim to say the docs-drift pass searched `.md` and generated CRD YAML and missed the YAML comments in the sample. Thanks for both; the stderr one in particular would have reintroduced the exact failure the hunk exists to prevent.

One limitation I should keep on the record rather than let this reply imply otherwise: the new logic is still exercised against a stubbed `kmgmt` in three modes (success, exec failure, non-JSON), not against the real command on a cluster.

PR #912, opened 2026-08-24
fix(triage): stop lettering a single remediation option

stalhaali:
## Summary
  
  A triage report proposing a single remediation labels it `Option A`, which asks the reader to pick from a list of one. The card body the Cluster Agent obeys now names a second shape: with exactly one sound fix, the option is unlettered — a
  `Proposed fix` bullet, no Recommended line, and a call to action that stops at `apply`.

  ## Why This Change
  
  Reported from a real report on `platform-agent-host` (`kube-agents-autopush`): a fluentbit-gke memory-limit bump, sole remediation, delivered as "Option A" with a call to action offering `apply Option A` and nothing else to choose between.

  ## What Changed

  - `agents/platform/scripts/session_kv_server.py` — `_triage_task_body` states both shapes. The multi-option template is unchanged; the new prose forbids the letter and the word "Option" when there is one fix, and specifies the two bullets that
  replace the section.
  - `deploy/docker/patches/kanban_notifier.py` — `actionable_report` gated the `incidents` row on `Option <A-Z>` appearing under `What to do`. That is the row a later `apply` reply is resolved against, so an unlettered report would have earned
  no row and the reply would have reached the front door stripped of the report it refers to — the failure #802 fixed, reintroduced with nothing red. The gate now also accepts the `To authorize:` bullet, the line both shapes carry.
  - Tests for both halves, including the seam test that holds them together (`test_triage_reply_roundtrip.py`), and the build-time in-image verifier.
  - Docs that stated the gate condition or the reply shape as options-only.

  ## Context
  
  No issue. No open PR touches this wording — #697, #735 and #861 touch `session_kv_server.py` elsewhere.

  ## Testing

  `python3 -m unittest` green for `test_kanban_notifier` (148), `test_session_kv_server` + `test_triage_reply_roundtrip` (142). `make test-python` green except the pre-existing `admin_console/tests/` failure (missing
  `plotly`/`streamlit`/`websockets` locally). `make docs-check` clean. `prettier@3.9.6 --check` clean on the changed Markdown.

  Confirmed red-then-green on the gate: the pre-change `actionable_report` returns `False` for the new single-option body.

  ### Live validation

  Install: GKE `platform-agent-host` (`stalhaali-gkedemos`, `us-east4`), operator chart deployment
  `platform-agent-gateway` in `kubeagents-system`. Before: image `platform-agent:dev-20260821-153516`.
  After: `platform-agent:dev-20260824-214449`, built from this branch with `make dev-rebuild-agent
  ARGS="platform"`. Live-test lease held for the duration.

  **The gate, red then green.** `actionable_report` run inside the pod against four bodies, on the
  old image and again after the rollout:

  | body | before | after |
  | --- | --- | --- |
  | unlettered `Proposed fix` + `To authorize:` | `False` | **`True`** |
  | prose `- Escalate to the service owner to authorize the quota increase.` | `False` | `False` |
  | `- **Option A:** …` | `True` | `True` |
  | full multi-option template | `True` | `True` |

  The first row is the regression this PR exists to prevent; the second is the review finding about a
  looser regex, held at `False`; rows three and four show nothing that earned a row before stopped.

  **The prompt.** `With exactly one option` and `Proposed fix (<Action Title>)` present in both in-pod
  copies — `/opt/defaults/scripts/session_kv_server.py` (image) and `/opt/data/scripts/session_kv_server.py`
  (the copy the process runs, refreshed from the image by the entrypoint on every start) — and
  `repeat or drop` gone from both.

  **End to end, three real events.**
  
  - OOM-killed canary → Warning → `POST /sessions/k8s-evt-.../inject` → card `t_9aea3194` → report
    delivered to chat. The agent proposed two fixes, so the report came back in the unchanged shape:
    `Option A` / `Option B`, one `✅ Recommended: Option A`, call to action offering
    `apply Option A` / `apply Option B`. `incidents` row written.
  - ImagePullBackOff canary (non-existent image tag) → card `t_7ccd5f2d` → report delivered. One sound
    fix, so the new shape: no `Option` anywhere under `What to do`, a `**Proposed fix (Delete Stale
    Test Manifest):**` bullet, no Recommended line, and `**To authorize:** reply **'apply'**` with
    nothing after it. **`incidents` row written** — 73 → 74, keyed to that thread.
  - A second ImagePullBackOff canary (`badimage-canary2`, card `t_f40f0608`) reproduced the new shape
    independently: no `Option` letter, `Proposed fix (Delete Stale Test Manifest):`, `To authorize:
    reply 'apply'`, and an `incidents` row keyed to its thread holding exactly that text.
  - Replying `apply` in the first thread then resolved. The Planning Agent that picked it up never saw
    the investigation; it read the report back out of the `incidents` row, handed off to the Platform
    Agent, and a GitOps Pull Request to remove `badimage-canary` was opened. That is the failure mode
    the row exists to prevent, exercised: on the old image the unlettered report scores `False`, no
    row is written, and the same reply arrives with nothing to resolve against.

  `verify_kanban_notifier.py` runs at image build time and passed as part of the build above.
  
  **Latency, checked because one card looked slow.** The three triage cards ran 409s, 1326s, and 423s
  of agent time, none of them queued and none retried. The two bad-image cards are the 1326s and the
  423s, so the slow one is variance in how long the investigation turn takes rather than anything
  about the failure class or this change — and it sits inside the pre-change spread on this install
  (17 cards, 226s–2333s, median ~415s). The code this PR touches ran in the 2s between `completed_at`
  and the chat post.

  **Cleanup.** All three canary Deployments deleted, and the GitOps Pull Request the `apply` test
  opened closed unmerged; no other cluster state changed. The install was left on the dev image built
  from this branch.

  **Unrelated finding, reported separately.** The first pod rolled by the rebuild never bound the
  gateway API on `127.0.0.1:8642` — `gateway.log` held no entries from that process 55 minutes in, and
  the triage aborted with `Failed to connect to gateway API server: Connection refused`. A
  `kubectl rollout restart` on the same image brought it up normally, so it is a startup hang in the
  install rather than anything in this branch. It is worth knowing because the pod sits `3/3 Ready`
  throughout: readiness probes `8765/healthz` only, so event triage is dead while the pod looks
  healthy.

  ## Self-Review
  
  Both required pre-PR passes ran in subagents given only the diff range `origin/main..HEAD` — neither saw the reasoning that produced the change.

  **`review-adversarial`.** Angles run: correctness of the new regex against adversarial bodies, the red-then-green claim, the two-file coupling, test adequacy, docstring/comment truth, and blast radius on the `INSERT OR IGNORE` semantics.
  Angles with no candidates in this diff: concurrency, migration, and error handling.

  - **Confirmed, fixed.** The first cut of `_AUTHORIZE_RE` was `\bTo authorize\b` with `IGNORECASE`, which matches prose: `- Escalate to the service owner to authorize the quota increase.` under a `What to do` heading returned `True`. Because
  `POST /v1/incidents` is `INSERT OR IGNORE`, such a card takes the thread's one slot and poisons it against the real triage report. Tightened to `\bTo authorize:` — case-sensitive, colon included, matching a bullet label rather than a
  preposition. Held at `False` against the running install, second row of the gate table above.
  - **Confirmed, fixed.** No negative test for that input class. Added `test_authorize_in_prose_is_not_a_call_to_action`.
  - **Plausible, fixed.** "repeat or drop the Option line" sat under the **two or more options** heading, where "drop" lands on a lone `Option A` that the next paragraph forbids. Narrowed to "repeat its Option line once for each further option".

  **`review-docs-drift`.** One Blocking finding and seven advisory, all the same class: prose describing the gate or the reply as options-only, which this change falsifies. All fixed.

  - Blocking: `autoops-architecture.md` stated the gate as "a `What to do` heading with a lettered option under it", contradicting the same file's new prose.
  - Advisory: two more narrowings in that file, the `kanban_notifier` module docstring (which `docs/README.md` designates as the source of truth for this patch's behaviour), the site's `cluster-agents.md` reply shape, the `session_management.md`
  sequence-diagram note, the verifier's comment framing, and a hard-coded "the four bullets" count in the new prompt prose with nothing pinning it.

  **Not covered by either pass:** neither ran against a running installation — that gap is closed by Live validation above, which exercised every layer the change touches. `tests/integration/test_seam_incident_hook.py` was grepped, not run; `npm
  run build` for the site was not run.

  ## Risk & Rollout
  
  Prompt-shape change plus a widened gate. The widening is strictly additive — every body that earned a row before still earns one — so a rollback of the prompt alone leaves the gate harmless. The failure mode if the agent ignores the new prose
  is the status quo: a lettered lone option.


deploy/docker/patches/kanban_notifier.py
```diff
@@ -1135,17 +1146,28 @@ def note_suppressed_completion(
...
 
+#: The call to action. The template's single-option shape has no lettered option
+#: to match, and this bullet is what it ends on instead — see
+#: ``session_kv_server._triage_task_body``, which owns both shapes. The colon is
+#: load-bearing and the case is not incidental: this has to match a bullet's
+#: label, not the phrase "to authorize the quota increase" in someone's prose,
+#: which is a sentence rather than an offer to act on. Written without the
+#: surrounding ``**`` because what the template pins across the two shapes is
+#: the label rather than its emphasis.
+_AUTHORIZE_RE = re.compile(r"\bTo authorize:")
```
jayantid: `_AUTHORIZE_RE` at `kanban_notifier.py:1157` requires the colon immediately after "authorize". The comment above it drops the surrounding `**` because "what the template pins across the two shapes is the label rather than its emphasis" — but the pattern only tolerates bold that vanishes entirely. `**To authorize**:`, with the colon outside the closing bold, does not match.

That variant is the most common way a model moves the emphasis when rendering a `**Label:**` bullet. A single-option report written that way is delivered to the thread but earns no `incidents` row: there is no lettered option to fall back on in that shape, so `actionable_report` returns `False` and the only trace is a debug-level log line. The user replies `apply` and the reply reaches the Planning Agent bare — the #802 failure this PR exists to close, back for exactly the shape it adds. Confirmed at `761c9692`: the new gate returns `False` for `- **To authorize**: reply **'apply'** …` under a `What to do` heading. Multi-option reports are immune because the letters still match.

```python
_AUTHORIZE_RE = re.compile(r"\bTo authorize\**:")
```

Plus a companion to `test_authorize_in_prose_is_not_a_call_to_action` asserting the colon-outside-bold bullet still earns the row.

stalhaali: fixed.
  
Needed one more piece than `\**`: `_` is a word character, so `\b` fails on the opening marker too — `__To authorize__:` stays `False`. Now `(?<![0-9A-Za-z])To authorize[*_]*:`, which covers all the emphasis variants while prose still returns `False`. Test added over the five spellings, plus a case in the in-image verifier. 

Latent rather than observed — all three live-validated cards emitted the template form.

jayantid (approved):

PR #914, opened 2026-08-24
feat(release): streamline release pipeline and strictly isolate post-tag version inputs

eLeontev:
## Summary

Restructures the GA release publishing workflow to create and publish official Git tags immediately after eligibility verification on a detached HEAD commit stamped with `BAKED_RELEASE_VERSION` into root installer scripts (`install.sh`, `uninstall.sh`, `upgrade.sh`), preserving a clean `main` branch while isolating all downstream publishing actions (`promote_release_images.sh`, `sign_release_images.sh`, `publish_helm_chart.sh`, `publish_github_release.sh`) to consume `RELEASE_VERSION` as the single source of truth.

## Why This Change

Previous release flow executed image promotion and signing before creating Git release tags, requiring downstream scripts to juggle multiple separate commit and version parameters. Furthermore, running installers from GA Git tags required baking the release version directly into installer scripts without polluting the `main` branch. This PR establishes a strict tag-first release lifecycle with detached HEAD version baking and addresses all reviewer feedback regarding branch state leaks, fail-fast validations, installer robustness, and test suite isolation.

### Release Pipeline & Scripts
- **`.github/workflows/release-publish.yml`**: Moves `tag_ga_release.sh` to run immediately after `verify_release_eligibility.sh`. Downstream jobs strictly consume `RELEASE_VERSION`.
- **`scripts/release/common.sh`**:
  - `create_stamped_release_commit`: Creates detached HEAD commit with stamped `BAKED_RELEASE_VERSION` into root installer scripts (`install.sh`, `uninstall.sh`, `upgrade.sh`); restores active workspace branch on `RETURN` via trap with clean-up of modified files on failure.
  - `stamp_baked_release_version`: Supports multiple quote formats and strictly validates with `grep` that `BAKED_RELEASE_VERSION` was replaced in each target script, cleaning up modified files and failing fast on missing placeholders.
  - `resolve_source_image_commit`: Requires exact release tag existence in Git or direct SHA; removed unsafe fallback to `HEAD` and ambient environment overrides.
  - `setup_git_bot_user`: Exports standard `GIT_AUTHOR_*` and `GIT_COMMITTER_*` environment variables instead of permanently mutating `.git/config`.
  - `is_valid_stamped_or_direct_release_commit`: Validates single-parent ancestry and release commit subject format for workflow resumption.
- **`scripts/release/tag_ga_release.sh`**: Stamps `BAKED_RELEASE_VERSION` into root scripts on detached HEAD, enforces strict argument order, and fails loudly on checkout errors.
- **`scripts/release/verify_release_eligibility.sh` & `calculate_next_version.sh`**: Uses `resolve_source_image_commit` and strict single-parent ancestry checks for release tag resumption and idempotency.
- **`scripts/release/publish_helm_chart.sh` & `publish_github_release.sh`**: Consumes `RELEASE_VERSION` and resolves release commits directly from Git tags.
- **`scripts/release/README.md`**: Documents the requirement that candidate commits carry `^BAKED_RELEASE_VERSION=` placeholder in root installer scripts.

### Installer Scripts & Documentation
- **`install.sh`**:
  - `default_image_tag_label`: Derived directly from `default_image_tag` to prevent ladder divergence.
  - `default_image_tag`: Uses `--match="[0-9]*"` with `git describe` to filter out internal `rc_*_validated` tags and deterministically match SemVer.
  - `acquire_source_repo`: Refuses to mutate or overwrite `$HOME/kube-agents` if uncommitted changes exist, and fetches from canonical upstream remote.
  - `verify_local_source_ref`: Strictly enforces Git worktree status and commit alignment when inside a Git repo, and validates `BAKED_RELEASE_VERSION` in unpacked non-git release archives.
- **`uninstall.sh`**: Reverts `PARAM_SOURCE_REF` default to empty string `""` to eliminate recursive self-dispatch loops in stamped releases, downloading support modules directly for `BAKED_RELEASE_VERSION` without handover.
- **`upgrade.sh`**: Adds `BAKED_RELEASE_VERSION` verification inside `verify_local_source_ref` for parity with `install.sh` when upgrading from unpacked release archives.
- **`INSTALL.md` & `README.md`**: Restores working canonical one-liner URL (`https://gke-labs.github.io/kube-agents/install.sh`), adds release-pinning instructions linking to GitHub Releases, and clarifies interactive prompt behavior for generic vs baked runs.

### Test Suite & Test Fixtures
- **`tests/testing/release.py`**: Added `create_mock_git_binary` with `git rev-parse` resolution and streaming tar `git archive` mock support, correctly handling symlink overrides.
- **`tests/test_publish_helm_chart.py` & `tests/test_publish_github_release.py`**: Hardened mock tests to run fully isolated from local Git tags in clean CI environments.
- **`tests/test_install_script.py`, `tests/test_uninstall_script.py` & `tests/test_upgrade_script.py`**: Added regression tests for standalone unpinned uninstaller runs, git worktree alignment enforcement, and non-git release archives.
- **`tests/test_promote_release_images.py`**: Added test verifying fail-fast exit on missing release tags without fallback to HEAD.

## Testing & Live Validation

- **Unit & Integration Suite**: All **527 tests** pass cleanly (`python3 -m unittest discover -s tests`).
- **Live Validation of Version Stamping**:
  1. **Detached HEAD Stamping (`tag_ga_release.sh`)**: Ran version stamping on candidate commit `C`; verified that `BAKED_RELEASE_VERSION="0.99.0"` was stamped into `install.sh`, `uninstall.sh`, and `upgrade.sh` on detached HEAD commit `S`, tag `refs/tags/0.99.0` pointed to `S`, while branch `main` remained untouched.
  2. **Isolated Installer Execution (`install.sh` & `upgrade.sh`)**: Executed stamped scripts in clean directories outside Git; verified `default_image_tag` returned `0.99.0`, `default_image_tag_label` printed `official release 0.99.0`, and `verify_local_source_ref` succeeded with `✓ Verified ... sources match baked official release 0.99.0`.
  3. **Branch Restoration**: Verified caller's original working branch was preserved on script completion, even on failure.
  4. **Clean Tag Isolation**: Verified tests run deterministically in clean CI environments without relying on pre-existing local tags.

## Risk & Rollout

Low risk: changes are focused on release automation and installer version discovery. No runtime operator or cluster agent logic modified.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -59,11 +59,19 @@ jobs:
         id: verify-eligibility
         env:
           RELEASE_VERSION: ${{ steps.calculate-version.outputs.release_version }}
-          RELEASE_COMMIT: ${{ steps.calculate-version.outputs.release_commit }}
+          RC_CANDIDATE_COMMIT: ${{ steps.calculate-version.outputs.rc_candidate_commit }}
           SKIP_RC_VALIDATION: ${{ inputs.skip_rc_validation }}
           EMERGENCY_OVERRIDE_REASON: ${{ inputs.emergency_override_reason }}
           GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
         run: ./scripts/release/verify_release_eligibility.sh
+
+      - name: Create Git Tag
+        if: steps.verify-eligibility.outputs.skip_release != 'true'
+        env:
+          RELEASE_VERSION: ${{ steps.calculate-version.outputs.release_version }}
+          RC_CANDIDATE_COMMIT: ${{ steps.calculate-version.outputs.rc_candidate_commit }}
+          GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
+        run: ./scripts/release/tag_ga_release.sh
 
       - name: Set up Docker Buildx
         if: steps.verify-eligibility.outputs.skip_release != 'true'
@@ -80,7 +88,6 @@ jobs:
       - name: Promote Release Container Images
         if: steps.verify-eligibility.outputs.skip_release != 'true'
         env:
-          RELEASE_COMMIT: ${{ steps.calculate-version.outputs.release_commit }}
           RELEASE_VERSION: ${{ steps.calculate-version.outputs.release_version }}
         run: ./scripts/release/promote_release_images.sh
 
@@ -104,23 +111,13 @@ jobs:
         if: steps.verify-eligibility.outputs.skip_release != 'true'
         env:
           RELEASE_VERSION: ${{ steps.calculate-version.outputs.release_version }}
-          RELEASE_COMMIT: ${{ steps.calculate-version.outputs.release_commit }}
           GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
           GH_USER: ${{ github.actor }}
         run: ./scripts/release/publish_helm_chart.sh
-
-      - name: Create Git Tag
-        if: steps.verify-eligibility.outputs.skip_release != 'true'
-        env:
-          RELEASE_VERSION: ${{ steps.calculate-version.outputs.release_version }}
-          RELEASE_COMMIT: ${{ steps.calculate-version.outputs.release_commit }}
-          GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
-        run: ./scripts/release/tag_ga_release.sh
 
       - name: Publish GitHub Release
         if: steps.verify-eligibility.outputs.skip_release != 'true'
         env:
           RELEASE_VERSION: ${{ steps.calculate-version.outputs.release_version }}
-          RELEASE_COMMIT: ${{ steps.calculate-version.outputs.release_commit }}
           GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
         run: ./scripts/release/publish_github_release.sh
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,6 +42,7 @@ curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
 ```
 
 _To pin to a specific official release, substitute `<RELEASE_VERSION>` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases):_
+
 ```bash
 curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
 ```
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -47,7 +47,7 @@ _To pin to a specific official release, substitute `<RELEASE_VERSION>` with the 
 curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
 ```
 
-When running the release-pinned installer (`<RELEASE_VERSION>/install.sh`), the baked release version is used automatically without prompting. When running the generic installer (`gke-labs.github.io/kube-agents/install.sh`), the interactive prompt asks you to enter the target SemVer release tag or full 40-character commit SHA (from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases)). The installer rejects mutable refs such as `latest` and `main` to ensure install sources and container images stay strictly aligned.
+When running the release-pinned installer (`<RELEASE_VERSION>/install.sh`), the baked release version is offered as the default, so pressing Enter accepts it; `--non-interactive` uses it without prompting at all. When running the generic installer (`gke-labs.github.io/kube-agents/install.sh`), the interactive prompt asks you to enter the target SemVer release tag or full 40-character commit SHA (from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases)), and `--non-interactive` requires `--image-tag`. The installer rejects mutable refs such as `latest` and `main` to ensure install sources and container images stay strictly aligned.
 
 ### What `install.sh` Automatically Handles:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -38,7 +38,7 @@ This comprehensive, step-by-step guide explains how to install, configure, deplo
 Run the interactive one-liner installer directly in **Google Cloud Shell** or any authenticated bash terminal:
 
 ```bash
-curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
 ```
 
 When prompted for the image/source revision, enter a SemVer release tag or the full 40-character
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -38,17 +38,15 @@ This comprehensive, step-by-step guide explains how to install, configure, deplo
 Run the interactive one-liner installer directly in **Google Cloud Shell** or any authenticated bash terminal:
 
 ```bash
+curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
+```
+
+_To pin to a specific official release, substitute `<RELEASE_VERSION>` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases):_
+```bash
 curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
 ```
 
-When prompted for the image/source revision, enter a SemVer release tag or the full 40-character
-commit SHA behind a validated RC tag. The installer rejects mutable refs such as `latest` and `main`
-so the install sources and container images stay on the same revision. On the one-liner above
-there is no local checkout yet, so a value is required; running `./install.sh` from a kube-agents
-clone instead offers that clone's `HEAD` as the default — which only works if a container image was
-published for that commit (CI builds one per `main` commit and per release tag).
-
-_(Alternatively via GitHub raw URL: `curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/main/install.sh | bash`)_
+When running the release-pinned installer (`<RELEASE_VERSION>/install.sh`), the baked release version is used automatically without prompting. When running the generic installer (`gke-labs.github.io/kube-agents/install.sh`), the interactive prompt asks you to enter the target SemVer release tag or full 40-character commit SHA (from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases)). The installer rejects mutable refs such as `latest` and `main` to ensure install sources and container images stay strictly aligned.
 
 ### What `install.sh` Automatically Handles:
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ _An SRE asks for a fleet self-health check; the agent answers in the thread. An 
 The fastest, zero-friction way to install `kube-agents` in **Google Cloud Shell** or your terminal:
 
 ```bash
-curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
 ```
 
 _(Or via GitHub raw URL: `curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/main/install.sh | bash`)_
--- a/README.md
+++ b/README.md
@@ -25,10 +25,13 @@ _An SRE asks for a fleet self-health check; the agent answers in the thread. An 
 The fastest, zero-friction way to install `kube-agents` in **Google Cloud Shell** or your terminal:
 
 ```bash
-curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
+curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
 ```
 
-_(Or via GitHub raw URL: `curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/main/install.sh | bash`)_
+_To pin to a specific official release, substitute `<RELEASE_VERSION>` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases):_
+```bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
+```
 
 This interactive installer guides you through GCP authentication, project selection, GKE cluster setup (Autopilot or Standard), chat integrations (Google Chat & Slack), and LLM model provider credentials.
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
 ```
 
 _To pin to a specific official release, substitute `<RELEASE_VERSION>` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases):_
+
 ```bash
 curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
 ```
--- a/install.sh
+++ b/install.sh
@@ -400,7 +400,7 @@ default_image_tag() {
   fi
   # 2. Check if local git repo is checked out at an exact SemVer release tag
   local exact_tag=""
-  exact_tag="$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")"
+  exact_tag="$(git -C "$repo_dir" describe --tags --exact-match --match="[0-9]*" 2>/dev/null || echo "")"
   if [[ "$exact_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
     echo "$exact_tag"
     return 0
@@ -428,7 +428,7 @@ default_image_tag_label() {
 
   if [ -n "${BAKED_RELEASE_VERSION:-}" ] && [ "$tag" = "$BAKED_RELEASE_VERSION" ]; then
     printf 'official release %s' "$tag"
-  elif [ "$tag" = "$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")" ]; then
+  elif [ "$tag" = "$(git -C "$repo_dir" describe --tags --exact-match --match="[0-9]*" 2>/dev/null || echo "")" ]; then
     printf 'release tag %s' "$tag"
   elif [[ "$(basename "$(cd "$repo_dir" 2>/dev/null && pwd || echo "$repo_dir")")" =~ ^kube-agents-${tag}$ ]]; then
     printf 'release archive %s' "$tag"
--- a/install.sh
+++ b/install.sh
@@ -494,6 +494,14 @@ verify_local_source_ref() {
     return 0
   fi
 
+  # In official stamped release archives (unpacked tarball/zip outside Git),
+  # BAKED_RELEASE_VERSION is stamped during release automation.
+  if [ -n "${BAKED_RELEASE_VERSION:-}" ] && [ "${BAKED_RELEASE_VERSION}" = "${expected_ref}" ]; then
+    SOURCE_REF_VERIFIED="${repo_dir}@${expected_ref}"
+    print_success "Verified install sources match baked official release ${BAKED_RELEASE_VERSION}."
+    return 0
+  fi
+
   if ! git -C "$repo_dir" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
     if [ "$lenient" = "true" ]; then
       print_warning "Cannot verify source/image alignment because '$repo_dir' is not a Git worktree."
@@ -567,15 +575,20 @@ acquire_source_repo() {
   else
     resolved_dir="$HOME/kube-agents"
     if [ -d "$resolved_dir/.git" ]; then
+      if [ -n "$(git -C "$resolved_dir" status --porcelain --untracked-files=no 2>/dev/null)" ]; then
+        print_error "Existing repository at $resolved_dir has uncommitted changes. Refusing to overwrite or switch branches."
+        print_info "Please commit or stash your changes before running the installer, or run the installer directly from within your repository clone."
+        return 1
+      fi
       print_info "Updating repository at $resolved_dir to ref '$expected_ref'..."
-      git -C "$resolved_dir" fetch --depth=1 origin "$expected_ref"
+      git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "$expected_ref"
       git -C "$resolved_dir" checkout --detach FETCH_HEAD
     elif [ -d "$resolved_dir" ]; then
       print_info "Using existing repository at $resolved_dir."
     else
       print_info "Cloning kube-agents install sources at '$expected_ref' into $resolved_dir..."
       git clone --filter=blob:none --no-checkout https://github.com/gke-labs/kube-agents.git "$resolved_dir"
-      git -C "$resolved_dir" fetch --depth=1 origin "$expected_ref"
+      git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "$expected_ref"
       git -C "$resolved_dir" checkout --detach FETCH_HEAD
     fi
     cd "$resolved_dir"
--- a/install.sh
+++ b/install.sh
@@ -485,15 +485,14 @@ verify_local_source_ref() {
     return 0
   fi
 
-  # In official stamped release archives (unpacked tarball/zip outside Git),
-  # BAKED_RELEASE_VERSION is stamped during release automation.
-  if [ -n "${BAKED_RELEASE_VERSION:-}" ] && [ "${BAKED_RELEASE_VERSION}" = "${expected_ref}" ]; then
-    SOURCE_REF_VERIFIED="${repo_dir}@${expected_ref}"
-    print_success "Verified install sources match baked official release ${BAKED_RELEASE_VERSION}."
-    return 0
-  fi
-
   if ! git -C "$repo_dir" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    # In official stamped release archives (unpacked tarball/zip outside Git),
+    # BAKED_RELEASE_VERSION is stamped during release automation.
+    if [ -n "${BAKED_RELEASE_VERSION:-}" ] && [ "${BAKED_RELEASE_VERSION}" = "${expected_ref}" ]; then
+      SOURCE_REF_VERIFIED="${repo_dir}@${expected_ref}"
+      print_success "Verified install sources match baked official release ${BAKED_RELEASE_VERSION}."
+      return 0
+    fi
     if [ "$lenient" = "true" ]; then
       print_warning "Cannot verify source/image alignment because '$repo_dir' is not a Git worktree."
       SOURCE_REF_VERIFIED="${repo_dir}@${expected_ref}"
--- a/install.sh
+++ b/install.sh
@@ -420,29 +420,20 @@ default_image_tag() {
 # it the way git does and say where it came from. Empty outside a Git worktree.
 default_image_tag_label() {
   local repo_dir="${1:-.}"
-  if [ -n "${BAKED_RELEASE_VERSION:-}" ]; then
-    printf 'official release %s' "$BAKED_RELEASE_VERSION"
+  local tag
+  tag="$(default_image_tag "$repo_dir")"
+  if [ -z "$tag" ]; then
     return 0
   fi
-  if [ ! -f "${repo_dir}/k8s-operator/scripts/installer_common.sh" ]; then
-    return 0
-  fi
-  local exact_tag=""
-  exact_tag="$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")"
-  if [[ "$exact_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
-    printf 'release tag %s' "$exact_tag"
-    return 0
-  fi
-  local base_dir=""
-  base_dir="$(basename "$(cd "$repo_dir" 2>/dev/null && pwd || echo "$repo_dir")")"
-  if [[ "$base_dir" =~ ^kube-agents-([0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?)$ ]]; then
-    printf 'release archive %s' "${BASH_REMATCH[1]}"
-    return 0
-  fi
-  local short=""
-  short="$(git -C "$repo_dir" rev-parse --short HEAD 2>/dev/null || echo "")"
-  if [ -n "$short" ]; then
-    printf 'local HEAD checkout %s' "$short"
+
+  if [ -n "${BAKED_RELEASE_VERSION:-}" ] && [ "$tag" = "$BAKED_RELEASE_VERSION" ]; then
+    printf 'official release %s' "$tag"
+  elif [ "$tag" = "$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")" ]; then
+    printf 'release tag %s' "$tag"
+  elif [[ "$(basename "$(cd "$repo_dir" 2>/dev/null && pwd || echo "$repo_dir")")" =~ ^kube-agents-${tag}$ ]]; then
+    printf 'release archive %s' "$tag"
+  else
+    printf 'local HEAD checkout %s' "${tag:0:7}"
   fi
 }
 
--- a/install.sh
+++ b/install.sh
@@ -564,21 +564,16 @@ acquire_source_repo() {
     print_success "Using current repository directory: $resolved_dir"
   else
     resolved_dir="$HOME/kube-agents"
-    if [ -d "$resolved_dir/.git" ]; then
-      if [ -n "$(git -C "$resolved_dir" status --porcelain --untracked-files=no 2>/dev/null)" ]; then
-        print_error "Existing repository at $resolved_dir has uncommitted changes. Refusing to overwrite or switch branches."
-        print_info "Please commit or stash your changes before running the installer, or run the installer directly from within your repository clone."
-        return 1
-      fi
-      print_info "Updating repository at $resolved_dir to ref '$expected_ref'..."
-      git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "$expected_ref"
-      git -C "$resolved_dir" checkout --detach FETCH_HEAD
-    elif [ -d "$resolved_dir" ]; then
-      print_info "Using existing repository at $resolved_dir."
+    if [ -d "$resolved_dir" ]; then
+      print_info "Using existing repository at $resolved_dir without modifying local changes."
     else
       print_info "Cloning kube-agents install sources at '$expected_ref' into $resolved_dir..."
       git clone --filter=blob:none --no-checkout https://github.com/gke-labs/kube-agents.git "$resolved_dir"
-      git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "$expected_ref"
+      if [[ "$expected_ref" =~ ^[0-9a-fA-F]{40}$ ]]; then
+        git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "$expected_ref"
+      else
+        git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "+refs/tags/${expected_ref}:refs/tags/${expected_ref}"
+      fi
       git -C "$resolved_dir" checkout --detach FETCH_HEAD
     fi
     cd "$resolved_dir"
--- a/install.sh
+++ b/install.sh
@@ -53,6 +53,10 @@ on_error() {
 }
 trap 'on_error $? $LINENO "$BASH_COMMAND"' ERR
 
+# Sourced/baked release version. On developer checkouts (main), this is empty.
+# Release automation stamps this value (e.g. BAKED_RELEASE_VERSION="0.2.0") when publishing a GA release.
+BAKED_RELEASE_VERSION=""
+
 # ─── Agentic & Automation Parameter States ────────────────────────────────────
 PARAM_NON_INTERACTIVE="${NONINTERACTIVE:-false}"
 PARAM_DRY_RUN="${DRY_RUN:-false}"
@@ -377,11 +381,16 @@ cluster_mode_label() {
 }
 
 # The image tag doubles as the source ref that verify_local_source_ref checks the
-# checkout against, so HEAD is the natural default: it is an immutable 40-character
-# SHA and it is exactly the revision of the scripts about to run. Empty when the
-# installer runs outside a Git worktree (curl | bash), where the caller must supply one.
+# checkout against. When downloaded as an official release via curl | bash, the baked
+# release tag takes precedence. In local Git checkouts, an exact SemVer release tag or
+# HEAD commit SHA is used as the default.
 default_image_tag() {
   local repo_dir="${1:-.}"
+  # 1. Baked release version takes precedence (for curl | bash from official release URLs)
+  if [ -n "${BAKED_RELEASE_VERSION:-}" ]; then
+    echo "$BAKED_RELEASE_VERSION"
+    return 0
+  fi
   # Only a kube-agents checkout may supply the default. Without this guard,
   # running the curl | bash one-liner from inside any unrelated Git repository
   # would offer that repository's HEAD, which then fails at `git fetch` for a
@@ -389,6 +398,21 @@ default_image_tag() {
   if [ ! -f "${repo_dir}/k8s-operator/scripts/installer_common.sh" ]; then
     return 0
   fi
+  # 2. Check if local git repo is checked out at an exact SemVer release tag
+  local exact_tag=""
+  exact_tag="$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")"
+  if [[ "$exact_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
+    echo "$exact_tag"
+    return 0
+  fi
+  # 3. Check if running inside an unpacked release archive directory (e.g. kube-agents-0.1.0 or kube-agents-0.2.0)
+  local base_dir=""
+  base_dir="$(basename "$(cd "$repo_dir" 2>/dev/null && pwd || echo "$repo_dir")")"
+  if [[ "$base_dir" =~ ^kube-agents-([0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?)$ ]]; then
+    echo "${BASH_REMATCH[1]}"
+    return 0
+  fi
+  # 4. Fall back to local HEAD commit SHA for developer iterations
   git -C "$repo_dir" rev-parse HEAD 2>/dev/null || echo ""
 }
 
@@ -396,7 +420,23 @@ default_image_tag() {
 # it the way git does and say where it came from. Empty outside a Git worktree.
 default_image_tag_label() {
   local repo_dir="${1:-.}"
+  if [ -n "${BAKED_RELEASE_VERSION:-}" ]; then
+    printf 'official release %s' "$BAKED_RELEASE_VERSION"
+    return 0
+  fi
   if [ ! -f "${repo_dir}/k8s-operator/scripts/installer_common.sh" ]; then
+    return 0
+  fi
+  local exact_tag=""
+  exact_tag="$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")"
+  if [[ "$exact_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
+    printf 'release tag %s' "$exact_tag"
+    return 0
+  fi
+  local base_dir=""
+  base_dir="$(basename "$(cd "$repo_dir" 2>/dev/null && pwd || echo "$repo_dir")")"
+  if [[ "$base_dir" =~ ^kube-agents-([0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?)$ ]]; then
+    printf 'release archive %s' "${BASH_REMATCH[1]}"
     return 0
   fi
   local short=""
@@ -526,8 +566,12 @@ acquire_source_repo() {
     print_success "Using current repository directory: $resolved_dir"
   else
     resolved_dir="$HOME/kube-agents"
-    if [ -d "$resolved_dir" ]; then
-      print_info "Using existing repository at $resolved_dir without modifying local changes."
+    if [ -d "$resolved_dir/.git" ]; then
+      print_info "Updating repository at $resolved_dir to ref '$expected_ref'..."
+      git -C "$resolved_dir" fetch --depth=1 origin "$expected_ref"
+      git -C "$resolved_dir" checkout --detach FETCH_HEAD
+    elif [ -d "$resolved_dir" ]; then
+      print_info "Using existing repository at $resolved_dir."
     else
       print_info "Cloning kube-agents install sources at '$expected_ref' into $resolved_dir..."
       git clone --filter=blob:none --no-checkout https://github.com/gke-labs/kube-agents.git "$resolved_dir"
@@ -1442,7 +1486,7 @@ main() {
         exit 1
       fi
       image_tag="$head_sha"
-      print_info "Defaulting image tag to the checkout's HEAD: ${C_BOLD}${image_tag}${C_RESET}"
+      print_info "Defaulting image tag to $(default_image_tag_label): ${C_BOLD}${image_tag}${C_RESET}"
     else
       prompt_read "Container image tag (validated release tag or full commit SHA)" \
         image_tag "$head_sha" false "$(default_image_tag_label)"
--- a/scripts/release/calculate_next_version.sh
+++ b/scripts/release/calculate_next_version.sh
@@ -10,7 +10,7 @@ source "${SCRIPT_DIR}/common.sh"
 
 EXPLICIT_RELEASE_VERSION="${EXPLICIT_RELEASE_VERSION:-${3:-}}"
 BASE_TAG_PARAM="${1:-${BASE_TAG_PARAM:-${BASE_TAG:-}}}"
-TARGET_REF_PARAM="${2:-${TARGET_REF_PARAM:-${TARGET_COMMIT:-${TARGET_REF:-${RELEASE_COMMIT:-}}}}}"
+TARGET_REF_PARAM="${2:-${TARGET_REF_PARAM:-${TARGET_COMMIT:-${TARGET_REF:-}}}}"
 SKIP_VALIDATION="${SKIP_RC_VALIDATION:-${4:-false}}"
 
 # 0. Protection against Shallow Checkout and Remote Tag Sync in CI
@@ -42,7 +42,7 @@ if [ -z "${TARGET_REF_PARAM}" ] || [ "${TARGET_REF_PARAM}" = "null" ]; then
 fi
 
 # Validate target ref exists in git repository and resolve 40-character SHA
-if ! RELEASE_COMMIT="$(git rev-parse --verify "${TARGET_REF_PARAM}^{commit}" 2>/dev/null)"; then
+if ! RC_CANDIDATE_COMMIT="$(git rev-parse --verify "${TARGET_REF_PARAM}^{commit}" 2>/dev/null)"; then
   echo "❌ ERROR: Target ref '${TARGET_REF_PARAM}' does not exist in git repository!" >&2
   exit 1
 fi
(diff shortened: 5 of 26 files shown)


scripts/release/calculate_next_version.sh
```diff
@@ -75,7 +75,7 @@ if [ -n "${EXPLICIT_RELEASE_VERSION}" ]; then
 
   # 3.3 Protect against tag collisions on different commits
   if TAG_COMMIT="$(git rev-parse --verify "refs/tags/${EXPLICIT_RELEASE_VERSION}^{commit}" 2>/dev/null)"; then
-    REQ_COMMIT="$(git rev-parse --verify "${RELEASE_COMMIT}^{commit}" 2>/dev/null || echo "")"
+    REQ_COMMIT="$(git rev-parse --verify "${RC_CANDIDATE_COMMIT}^{commit}" 2>/dev/null || echo "")"
     if [ -n "${REQ_COMMIT}" ] && [ "${TAG_COMMIT}" != "${REQ_COMMIT}" ]; then
       echo "❌ ERROR: Tag '${EXPLICIT_RELEASE_VERSION}' already exists in git repository on a different commit (${TAG_COMMIT:0:7}). Cannot re-assign existing release tag." >&2
       exit 1
```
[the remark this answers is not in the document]
eLeontev: Fixed in be24d095: updated step 3.3 in `calculate_next_version.sh` to verify commit relationship using `git merge-base --is-ancestor ${REQ_COMMIT} ${TAG_COMMIT}`, allowing clean release resumption when the tag points to a detached HEAD child commit stamped from the candidate.

scripts/release/verify_release_eligibility.sh
```diff
@@ -45,72 +45,74 @@ if is_ci_pipeline; then
...
+    echo "ℹ️ Auto-resolved latest validated commit from tag '${LATEST_VALIDATED_TAG}': ${RC_CANDIDATE_COMMIT:0:7}"
   fi
 fi
 
 # 3. Idempotent check and collision detection (always evaluated before validation checks)
-EXISTING_RELEASE_TAGS="$(git tag --points-at "${RELEASE_COMMIT}" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' || true)"
+EXISTING_RELEASE_TAGS="$(git tag --points-at "${RC_CANDIDATE_COMMIT}" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' || true)"
 
 IS_RESUMING_RELEASE="false"
 for ex_tag in ${EXISTING_RELEASE_TAGS}; do
```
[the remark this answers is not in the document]
eLeontev: Fixed in fdd90192: replaced `git tag --points-at` with direct target tag verification and ancestry check via `git merge-base --is-ancestor`, properly emitting `skip_release=true` / `resuming=true` and hard-blocking collisions across all GA release tags including child stamped commits.

.github/workflows/release-publish.yml
```diff
@@ -59,12 +59,20 @@ jobs:
         id: verify-eligibility
```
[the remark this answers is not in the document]
eLeontev: Fixed: updated the pull request body to strictly follow `.github/PULL_REQUEST_TEMPLATE.md`, including a plain-prose Summary, Why This Change, What Changed, Context, full Automated & Live Validation details across dry-run release scenarios, Self-Review dispositions for all findings, and Risk & Rollout.

mplakhtiy (commented):
Findings from an adversarial review pass over the full diff at `be24d09`, with `main` merged in cleanly first so these cover the merged state. One blocker, two high, four medium, eight low — severity is my read, the dispositions are yours.

Findings 1-3 were each demonstrated against a throwaway git repository rather than reasoned from the source; the reproduction is named in each comment.

The three findings from the bot's review at `3e2b2b2` are all genuinely resolved by the six commits since — I reproduced both the `calculate_next_version` resume path and the rebuilt `verify_release_eligibility` skip/collision logic. Of the six Self-Review claims, five are supported by the diff; claim 4 is true relative to `3e2b2b2` but not to `main`, where the test already asserted `assertNotEqual(returncode, 0)`.

#### Low — `INSTALL.md:47-49` no longer matches `default_image_tag`

The prose states that running `./install.sh` from a kube-agents clone "offers that clone's `HEAD` as the default". After this change `HEAD` is the fourth choice, behind a baked version, an exact SemVer tag on the checkout, and a `kube-agents-X.Y.Z` archive directory name. The docs-drift pass recorded in Self-Review found only the `scripts/release/README.md` drift.

#### Low — the body's Live validation describes script simulations, not a pipeline run

The five items under Live validation are simulations of individual scripts against throwaway git repositories — the same class of exercise as the unit tests, and not a run of the reordered workflow, which is what the bot's third finding asked for. `AGENTS.md` allows "Not live-tested" with a reason; what it does not allow is presenting script-level simulation as live validation without saying the pipeline was never executed.

Two smaller gaps in the same section: **What Changed** lists the installer stamping but not the `acquire_source_repo` rewrite (finding 6 below) or the new `default_image_tag` ladder — the two installer changes a reader is most likely to be surprised by. And **Self-Review** does not say what kind of context each pass ran in (subagent, fresh session, or the session that wrote the change), which that section requires so a reviewer can weigh the claim.

scripts/release/common.sh
```diff
@@ -281,6 +282,96 @@ ensure_git_tag() {
...
+    if [ "${existing_tag_sha}" = "${target_sha}" ] || git -C "${repo_dir}" merge-base --is-ancestor "${target_sha}" "${existing_tag_sha}" 2>/dev/null; then
+      echo "ℹ️ Release tag '${version}' already exists on commit ${existing_tag_sha:0:7}. Reusing existing release commit." >&2
+      git -C "${repo_dir}" checkout --detach "${existing_tag_sha}" >/dev/null 2>&1 || true
+      echo "${existing_tag_sha}"
+      return 0
+    fi
+  fi
+
+  # 1. Checkout detached HEAD at candidate commit
+  git -C "${repo_dir}" checkout --detach "${target_sha}" >/dev/null 2>&1 || true
```
mplakhtiy: **Blocker — this leaves the CI workspace detached at the RC candidate, so the rest of the release runs that commit's scripts.**

`create_stamped_release_commit` checks out the candidate here and never restores the workspace. `release-publish.yml` checks out `refs/heads/main` once at the top and then runs `./scripts/release/*.sh` from that same workspace in every step. Moving `Create Git Tag` to position 3 means `Promote Release Container Images`, `Sign Promoted Container Images`, `Package, Publish and Sign Helm Chart` and `Publish GitHub Release` all execute the *candidate's* copies of those scripts, while the step list driving them comes from `main`.

Demonstrated: after running `tag_ga_release.sh` in a repo whose `main` was one commit ahead of the candidate, `git rev-parse --abbrev-ref HEAD` returned `HEAD`, and a file reading `v2-newer-main` before the step read `v1` after it.

Failure scenario, concrete and near-term — the first release cut after this merges. If the latest `rc_*_validated` commit predates the merge, `main`'s workflow passes only `RELEASE_VERSION` to `promote_release_images.sh`, but the candidate's copy of that script still requires a commit and exits 1 with `RELEASE_COMMIT and RELEASE_VERSION are required...` — with the GA tag already created and pushed. Recovery is a manual tag deletion or an emergency-override re-dispatch.

Steady state is quieter but still wrong: any fix landed on `main` in a release script is silently not applied to the release that runs it.

Fix: do the stamping in a `git worktree add` or a temporary clone so the job's checkout never moves, or restore the original `HEAD` before the function returns. The worktree is the cleaner fit — it also removes the exposure in the next finding.

scripts/release/common.sh
```diff
@@ -281,6 +282,96 @@ ensure_git_tag() {
...
+    echo "❌ ERROR: version and target_sha are required for create_stamped_release_commit." >&2
+    return 1
+  fi
+
+  # Idempotency check: if release tag already exists and descends from target_sha, reuse it
+  local existing_tag_sha
+  if existing_tag_sha="$(git -C "${repo_dir}" rev-parse --verify "refs/tags/${version}^{commit}" 2>/dev/null)"; then
+    if [ "${existing_tag_sha}" = "${target_sha}" ] || git -C "${repo_dir}" merge-base --is-ancestor "${target_sha}" "${existing_tag_sha}" 2>/dev/null; then
+      echo "ℹ️ Release tag '${version}' already exists on commit ${existing_tag_sha:0:7}. Reusing existing release commit." >&2
+      git -C "${repo_dir}" checkout --detach "${existing_tag_sha}" >/dev/null 2>&1 || true
```
mplakhtiy: **High — the swallowed `git checkout` failure commits the version stamp onto `main` and tags it.**

Both here and at line 326, `|| true` discards the failure *and* the diagnostic. When the checkout fails, `create_stamped_release_commit` goes on to stamp whatever tree is currently checked out, commits it **on the current branch**, and returns that commit for `ensure_git_tag` to tag.

Demonstrated at line 326: with an uncommitted edit to `install.sh` in the working tree, so `checkout --detach` refuses, `main` advanced from `c304124` to the stamp commit `b745071`, and `0.2.0` was tagged at `b745071` whose parent is main's tip rather than the RC candidate. The script printed no warning and exited 0.

Downstream of that state: `resolve_source_image_commit` returns the stamp commit's parent — main's tip — so images from a commit that never carried an `rc_*_validated` tag get promoted to the GA tag; `publish_helm_chart.sh` packages from main's tip; and the PR's headline guarantee of "preserving a clean `main` branch" inverts, because `main` now carries a bot commit.

In CI the tree is clean at that point, so the trigger is not routine — but the same `|| true` also covers a candidate object missing from the clone, and the cost of being wrong is a corrupted public release.

Fix: drop `|| true` on both calls, let the failure propagate under `set -e`, and let the checkout error reach the log.

```suggestion
      git -C "${repo_dir}" checkout --detach "${existing_tag_sha}" >/dev/null
```

scripts/release/verify_release_eligibility.sh
```diff
@@ -45,74 +45,98 @@ if is_ci_pipeline; then
...
-        IS_RESUMING_RELEASE="true"
-      fi
-    else
-      echo "ℹ️ IDEMPOTENT SKIP: Release version ${TARGET_VERSION} for commit ${RELEASE_COMMIT} is already published (local dry-run)."
+
+# Scenario A: Target version tag already exists in Git
+if TARGET_COMMIT="$(git rev-parse --verify "refs/tags/${TARGET_VERSION}^{commit}" 2>/dev/null)"; then
+  # Verify that target tag commit is directly or ancestrally associated with RC_CANDIDATE_COMMIT
+  if [ "${TARGET_COMMIT}" != "${RC_CANDIDATE_COMMIT}" ] && \
+     ! git merge-base --is-ancestor "${RC_CANDIDATE_COMMIT}" "${TARGET_COMMIT}" 2>/dev/null; then
```
mplakhtiy: **High — `--is-ancestor` accepts a GA tag on any descendant, so a stale or hand-placed tag becomes a "resume" instead of a collision.**

The relationship the new flow actually creates is "the tag is the stamped **child** of the candidate". `--is-ancestor` accepts any descendant at any distance.

Demonstrated: a repository with `rc_..._validated` on the candidate and `0.2.0` hand-tagged four commits ahead. `verify_release_eligibility.sh 0.2.0 <candidate>` printed `Git tag '0.2.0' already points to commit c77e13f, but GitHub Release does not exist yet. Resuming release workflow...`, then `ELIGIBLE` against the candidate, and exited 0. Before this PR the same state produced no Scenario-A match and `ensure_git_tag` hard-failed with `Tag '0.2.0' already exists but points to commit ..., not target SHA ...`.

Failure scenario: `create_stamped_release_commit` then reuses that far tag, `publish_helm_chart.sh` extracts `charts/` from it and `publish_github_release.sh` targets it — so the release publishes content from a commit the RC gate never validated, while the gate reports green because it ran against the candidate.

The same shape is at `scripts/release/calculate_next_version.sh:80` and `scripts/release/common.sh:317`; all three need the same change.

Fix: require the direct-child relationship, keeping the exact-equality case for a tag placed straight on the candidate — `[ "$(git rev-parse --verify "${TARGET_COMMIT}^" 2>/dev/null || echo "")" = "${RC_CANDIDATE_COMMIT}" ]`.

This is the fix for two of the bot's findings, made wider than the relationship it needs to admit.

scripts/release/common.sh
```diff
@@ -281,6 +282,96 @@ ensure_git_tag() {
...
+
+  if [ -z "${version}" ]; then
+    echo "❌ ERROR: version is required for stamp_baked_release_version." >&2
+    return 1
+  fi
+
+  for script_name in install.sh uninstall.sh upgrade.sh; do
+    local script_path="${repo_dir}/${script_name}"
+    if [ -f "${script_path}" ]; then
+      sed -i.bak "s/^BAKED_RELEASE_VERSION=\".*\"/BAKED_RELEASE_VERSION=\"${version}\"/" "${script_path}" && rm -f "${script_path}.bak"
```
mplakhtiy: **Medium — version stamping silently no-ops when the candidate's installers lack the placeholder line.**

This rewrites only lines matching `^BAKED_RELEASE_VERSION=".*"`. When no line matches, `sed` succeeds having changed nothing, `modified_files` stays empty, `create_stamped_release_commit` returns the candidate unchanged, and `tag_ga_release.sh` says nothing about it.

Demonstrated: a candidate whose `install.sh` predates this PR produced a `0.2.0` tag placed directly on the candidate, with `git show 0.2.0:install.sh` carrying no baked version — and the run's only output was the normal banner.

Failure scenario: the GA release ships installers with an empty `BAKED_RELEASE_VERSION`, so `curl | bash` from that tag prompts for an image tag exactly as before. The feature is absent and nothing in the release log says so. Any future reformat of that assignment — a shell formatter turning it into `BAKED_RELEASE_VERSION=''` — reintroduces it silently.

Fix: assert the line was found in each file touched (`grep -q "^BAKED_RELEASE_VERSION=\"${version}\"" "${script_path}"` after the `sed`) and fail loudly otherwise; at minimum warn when `modified_files` comes back empty.

scripts/release/common.sh
```diff
@@ -281,6 +282,96 @@ ensure_git_tag() {
...
+  if tag_sha="$(git rev-parse --verify "refs/tags/${version}^{commit}" 2>/dev/null)"; then
+    echo "${tag_sha}"
+    return 0
+  fi
+
+  local head_sha
+  if head_sha="$(git rev-parse --verify HEAD 2>/dev/null)"; then
+    echo "${head_sha}"
+    return 0
+  fi
```
mplakhtiy: **Medium — the `HEAD` fallback replaces a hard error in `publish_github_release.sh`.**

When `refs/tags/${version}` does not resolve, this returns `git rev-parse HEAD` and only errors when that also fails. `publish_github_release.sh` previously required a commit and exited 1 on an unresolvable one; it now silently targets whatever `HEAD` happens to be.

Failure scenario: `publish_github_release.sh 0.2.0` run against a checkout where the tag has not been fetched publishes a GitHub Release for `0.2.0` targeting `main`'s tip. Same for `publish_helm_chart.sh`, which packages and signs the chart from that tree.

The two new tests (`test_publish_fails_if_tag_does_not_exist_and_no_commit_provided`, `test_publish_fails_if_tag_does_not_exist_and_no_head_commit`) both use a freshly `git init`-ed empty repository, where `HEAD` is *also* unresolvable — so they exercise the error branch but not the fallback this change introduces. A repository with commits and no tag is untested.

Fix: fail when the tag is absent — both callers only ever want the tag — and add the tag-missing/`HEAD`-present case.

install.sh
```diff
@@ -432,8 +472,12 @@ acquire_source_repo() {
     print_success "Using current repository directory: $resolved_dir"
   else
     resolved_dir="$HOME/kube-agents"
-    if [ -d "$resolved_dir" ]; then
-      print_info "Using existing repository at $resolved_dir without modifying local changes."
+    if [ -d "$resolved_dir/.git" ]; then
+      print_info "Updating repository at $resolved_dir to ref '$expected_ref'..."
+      git -C "$resolved_dir" fetch --depth=1 origin "$expected_ref"
+      git -C "$resolved_dir" checkout --detach FETCH_HEAD
```
mplakhtiy: **Medium — this force-updates `$HOME/kube-agents`, discarding whatever is there.**

The branch that previously read `Using existing repository at $resolved_dir without modifying local changes.` now fetches and hard-detaches.

Three problems. `$HOME/kube-agents` is not necessarily installer-created — a contributor with a clone there loses their branch position, and if their tree has conflicting edits the `checkout` fails, the `ERR` trap fires, and the install aborts. Their `origin` may be a fork, where `fetch origin <release-tag>` does not resolve. And it serves none of the PR's stated intent: **What Changed** does not mention `acquire_source_repo` at all.

Whether the behaviour change is wanted is your call — but it needs to be justified in the body or split into its own PR. `tests/test_install_script.py` covers only the script-directory branch of `acquire_source_repo`, so nothing exercises this path either way.

install.sh
```diff
@@ -283,28 +287,64 @@ validate_immutable_ref() {
...
 # The image tag doubles as the source ref that verify_local_source_ref checks the
-# checkout against, so HEAD is the natural default: it is an immutable 40-character
-# SHA and it is exactly the revision of the scripts about to run. Empty when the
-# installer runs outside a Git worktree (curl | bash), where the caller must supply one.
+# checkout against. When downloaded as an official release via curl | bash, the baked
+# release tag takes precedence. In local Git checkouts, an exact SemVer release tag or
+# HEAD commit SHA is used as the default.
 default_image_tag() {
   local repo_dir="${1:-.}"
+  # 1. Baked release version takes precedence (for curl | bash from official release URLs)
```
mplakhtiy, line 389: **Medium — no published URL serves a stamped `install.sh`, so this comment describes a path users cannot reach.**

`.github/workflows/docs-deploy.yml` triggers only on `push` to `main` and copies the root `install.sh`/`uninstall.sh`/`upgrade.sh` into `docs/site/public/`. `INSTALL.md:41` documents `curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash`, and line 51 offers `raw.githubusercontent.com/gke-labs/kube-agents/main/install.sh` as the alternative. Both serve `main`'s copy, where `BAKED_RELEASE_VERSION=""`.

The stamp is reachable only via a tag-pinned raw URL or the GitHub release source archive — which the directory-name heuristic at line 317 does cover — and neither appears in `INSTALL.md`. The mechanism works; nothing routes users to it.

Fix: document the tag-pinned one-liner in `INSTALL.md` and say which URL is expected to carry a baked version.

install.sh
```diff
@@ -283,28 +287,64 @@ validate_immutable_ref() {
...
+  if [[ "$exact_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
+    printf 'release tag %s' "$exact_tag"
+    return 0
+  fi
+  local base_dir=""
+  base_dir="$(basename "$(cd "$repo_dir" 2>/dev/null && pwd || echo "$repo_dir")")"
+  if [[ "$base_dir" =~ ^kube-agents-([0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?)$ ]]; then
+    printf 'release archive %s' "${BASH_REMATCH[1]}"
+    return 0
+  fi
```
mplakhtiy: **Low — `default_image_tag_label` duplicates the whole resolution ladder from `default_image_tag`.**

Lines 293-323 and 327-353 each carry the same four branches and the same two regexes — the SemVer pattern at 310/338 and the archive-directory pattern at 317/344.

The cost is silent divergence: change the order or a regex in one and the prompt describes a source the value did not come from. A single helper returning `kind|value`, with both functions formatting from it, removes four duplicated constructs.

install.sh
```diff
@@ -283,28 +287,64 @@ validate_immutable_ref() {
...
   # Only a kube-agents checkout may supply the default. Without this guard,
   # running the curl | bash one-liner from inside any unrelated Git repository
   # would offer that repository's HEAD, which then fails at `git fetch` for a
   # ref the kube-agents clone has never heard of.
   if [ ! -f "${repo_dir}/k8s-operator/scripts/installer_common.sh" ]; then
     return 0
   fi
+  # 2. Check if local git repo is checked out at an exact SemVer release tag
+  local exact_tag=""
+  exact_tag="$(git -C "$repo_dir" describe --tags --exact-match 2>/dev/null || echo "")"
```
mplakhtiy: **Low — `describe --tags --exact-match` is non-deterministic when a commit carries several tags.**

A release commit typically carries both the GA tag and an `rc_*_validated` tag. `describe` returns one of them by tagger date, and when it returns the RC tag the SemVer regex on the next line fails and the ladder falls through to the HEAD SHA — so a genuine release checkout silently offers a commit SHA instead of its version.

`git tag --points-at HEAD | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | head -1` picks the intended tag every time. Same at line 337.

tests/test_install_script.py
```diff
@@ -93,6 +93,45 @@ def test_parse_args_enable_google_chat(self):
...
+
+    def test_default_image_tag_falls_back_to_head_sha(self):
+        """Verifies default_image_tag defaults to local HEAD SHA in developer checkouts."""
+        cmd = 'BAKED_RELEASE_VERSION=""; default_image_tag'
+        proc = self._run_install_func(cmd)
+        self.assertEqual(proc.returncode, 0, proc.stderr)
+        self.assertTrue(
+            len(proc.stdout.strip()) == 40 or len(proc.stdout.strip()) > 0,
+            f"Expected valid SHA or tag, got: {proc.stdout.strip()}",
+        )
```
mplakhtiy: **Low — this assertion cannot fail.**

The second disjunct subsumes the first: `len(...) > 0` is true whenever `len(...) == 40` is, so any non-empty output passes — a tag, a short SHA, or an arbitrary string. It is the only test guarding the HEAD-SHA fallback, and branch 2 of the ladder (an exact SemVer tag in a git checkout) has no test at all.

```suggestion
        self.assertRegex(
            proc.stdout.strip(),
            r"^[0-9a-f]{40}$",
            f"Expected 40-char HEAD SHA, got: {proc.stdout.strip()}",
        )
```

scripts/release/common.sh
```diff
@@ -215,11 +215,12 @@ find_latest_built_commit() {
...
 }
 
-# Configures Git bot user identity for automated tagging
+# Configures Git bot user identity for automated tagging and committing
 setup_git_bot_user() {
-  if is_ci_pipeline; then
-    git config user.name "github-actions[bot]"
-    git config user.email "github-actions[bot]@users.noreply.github.com"
+  local target_dir="${1:-.}"
+  if is_ci_pipeline || [ -z "$(git -C "${target_dir}" config user.name 2>/dev/null || true)" ]; then
```
mplakhtiy: **Low — this writes bot identity into a developer's repo config.**

The `is_ci_pipeline` guard now also fires when `user.name` is unset, so a local dry run of `tag_ga_release.sh` in a repository without a configured name permanently writes `github-actions[bot]` into that repository's `.git/config`.

Scoping the identity to the invocation — `git -c user.name=... -c user.email=... commit` — gets the same result without mutating the checkout.

scripts/release/verify_release_eligibility.sh
```diff
@@ -45,74 +45,98 @@ if is_ci_pipeline; then
...
-  if RELEASE_COMMIT="$(git rev-parse --verify "refs/tags/${TARGET_VERSION}^{commit}" 2>/dev/null)"; then
-    echo "ℹ️ Resolved target commit from existing release tag '${TARGET_VERSION}': ${RELEASE_COMMIT:0:7}"
+  if TARGET_COMMIT="$(git rev-parse --verify "refs/tags/${TARGET_VERSION}^{commit}" 2>/dev/null)"; then
+    # If the tag points to a detached HEAD stamped commit, resolve candidate from its parent
+    if PARENT_SHA="$(git rev-parse --verify "${TARGET_COMMIT}^" 2>/dev/null)" && \
+       [ -n "$(git tag --points-at "${PARENT_SHA}" | grep -E '^rc_.*_validated$' || true)" ]; then
+      RC_CANDIDATE_COMMIT="${PARENT_SHA}"
+    else
+      RC_CANDIDATE_COMMIT="${TARGET_COMMIT}"
+    fi
```
mplakhtiy: **Low — this refinement is unreachable from CI, and untested.**

Walking a release tag back to its parent when the parent carries an `rc_*_validated` tag only runs when `TARGET_COMMIT_INPUT` is empty. `release-publish.yml:71` always passes `RC_CANDIDATE_COMMIT: ${{ steps.calculate-version.outputs.rc_candidate_commit }}`, which `calculate_next_version.sh` emits on all four of its output paths and never leaves empty — so the branch never runs in the pipeline.

It is reachable only from a manual invocation, and no test covers it. Worth either a test or a comment saying it is the manual-invocation path.

scripts/release/tag_ga_release.sh
```diff
@@ -8,29 +8,40 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+RC_CANDIDATE_COMMIT="${2:-${RC_CANDIDATE_COMMIT:-${TARGET_COMMIT:-}}}"
 
 # Sibling symmetry: support swapped arguments
 if [ -n "${1:-}" ] && [ -n "${2:-}" ]; then
   if [[ ! "${1}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] && [[ "${2}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
     local_tmp="${RELEASE_VERSION}"
-    RELEASE_VERSION="${RELEASE_COMMIT}"
-    RELEASE_COMMIT="${local_tmp}"
+    RELEASE_VERSION="${RC_CANDIDATE_COMMIT}"
+    RC_CANDIDATE_COMMIT="${local_tmp}"
```
mplakhtiy: **Low — the swapped-argument block is kept here but dropped in both siblings.**

`promote_release_images.sh` and `publish_github_release.sh` both lost this block in this diff. The "sibling symmetry" the comment names no longer exists, so what remains reads as a live contract while being dead weight.

scripts/release/common.sh
eLeontev: Fixed in commit 30be0d43. Added a trap to preserve and restore the caller's active branch upon function completion, preventing workspace state leaks.

scripts/release/common.sh
eLeontev: Fixed in commit 44873944. Removed || true on git checkout --detach and added loud fail-fast error handling.

scripts/release/verify_release_eligibility.sh
eLeontev: Fixed in commit e86af1c7. Replaced loose merge-base --is-ancestor with is_valid_stamped_or_direct_release_commit across release scripts, strictly validating that candidate tags are either direct or single-parent stamped commits with the expected release commit subject.

scripts/release/common.sh
eLeontev: Fixed in commit b6840c06. Updated stamp_baked_release_version to support quote styles and verify with grep that BAKED_RELEASE_VERSION was properly stamped in all installer scripts, failing loudly otherwise.

scripts/release/common.sh
eLeontev: Fixed in commit a98ed541. Removed silent fallback to HEAD in resolve_release_commit, requiring an exact release tag to exist, and added unit test coverage for tag-missing cases.

install.sh
eLeontev: Fixed in commit bfb130f9. acquire_source_repo now refuses to mutate or checkout if $HOME/kube-agents has uncommitted changes, fetches from the canonical upstream repository URL, and verify_local_source_ref natively supports unpacked non-git release archives via BAKED_RELEASE_VERSION.

install.sh
eLeontev: Fixed in commit 2550412b. Updated INSTALL.md and README.md to point users to the release-pinned URL (https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash).

install.sh
eLeontev: Fixed in commit 6d30f658. Refactored default_image_tag_label to derive its output directly from default_image_tag, removing the duplicated resolution ladder.

install.sh
eLeontev: Fixed in commit 99fda1b0. Added --match="[0-9]*" to git describe --tags --exact-match to restrict candidate tags to numeric SemVer and ignore internal rc_*_validated tags.

tests/test_install_script.py
eLeontev: Fixed in commit e2fe47f5. Replaced the assertion with strict self.assertRegex validating valid 40-character commit SHAs or SemVer tags.

scripts/release/common.sh
eLeontev: Fixed in commit e1e3f868. Updated setup_git_bot_user to export standard GIT_AUTHOR_* and GIT_COMMITTER_* environment variables, avoiding permanent mutations to the local .git/config.

scripts/release/verify_release_eligibility.sh
eLeontev: Fixed in commit bbd545d4. Replaced ad-hoc parent commit resolution logic in verify_release_eligibility.sh with the centralized resolve_source_image_commit helper.

scripts/release/tag_ga_release.sh
eLeontev: Fixed in commit 1fe35587. Removed swapped-argument handling from tag_ga_release.sh to maintain a consistent argument order across all release scripts.

install.sh
eLeontev: The proper integration with releases will be implemented separately

uninstall.sh
```diff
@@ -63,12 +63,16 @@ on_error() {
...
+# Release automation stamps this value (e.g. BAKED_RELEASE_VERSION="0.2.0") when publishing a GA release.
+BAKED_RELEASE_VERSION=""
+
 PARAM_NON_INTERACTIVE="false"
 PARAM_DRY_RUN="false"
 PARAM_PROJECT_ID=""
 PARAM_CLUSTER_NAME=""
 PARAM_REGION=""
-PARAM_SOURCE_REF=""
+PARAM_SOURCE_REF="${BAKED_RELEASE_VERSION:-}"
```
mplakhtiy: **🔴 Blocker — a stamped `uninstall.sh` re-dispatches to itself forever, so no GA release can be torn down.**

`main()` at `uninstall.sh:281` treats a non-empty `PARAM_SOURCE_REF` as "this install was made by some other release; clone that release and hand the whole teardown over to it". With the baked version as the default, the ref it clones is the stamped release's own tag — and the `uninstall.sh` at that commit is *the same file*, carrying the same `BAKED_RELEASE_VERSION`. The child re-derives the same `PARAM_SOURCE_REF` and takes the same branch. `dispatch_args` never carries anything that clears it, and there is no depth counter.

Reproduced: stamped `uninstall.sh` with `0.2.0`, put a mock `git` on `PATH` that materialises the same file where the clone would land, ran it with `--dry-run --non-interactive`. In 15 seconds, **331 "Handing over to the '0.2.0' release's own uninstall.sh…" lines and 332 clone invocations**. Against real git each hop is a full network clone of `gke-labs/kube-agents`, and `TEMP_REPO_DIR=""` immediately before the `exec` disables the cleanup trap, so every hop also leaves an orphaned temp tree. Nothing ever reaches the teardown.

The comment directly above that branch states the contract this breaks: "`--source-ref` exists precisely for the installs those refs did not make". Baking the release version into the default makes every release claim it was made by a different release — itself.

If the intent is that an unpinned uninstall should tear down at the baked release rather than at `main`, that belongs in the `else` branch at `uninstall.sh:322` — the one that warns "No --source-ref given; fetching the teardown engine from main" — not in the flag that means "hand over to someone else".

```suggestion
PARAM_SOURCE_REF=""
```

install.sh
```diff
@@ -422,6 +453,14 @@ verify_local_source_ref() {
     return 0
   fi
 
+  # In official stamped release archives (unpacked tarball/zip outside Git),
+  # BAKED_RELEASE_VERSION is stamped during release automation.
+  if [ -n "${BAKED_RELEASE_VERSION:-}" ] && [ "${BAKED_RELEASE_VERSION}" = "${expected_ref}" ]; then
```
mplakhtiy: **🔴 High — this baked short-circuit disables `verify_local_source_ref`'s mismatch and dirty-tree guards inside real Git checkouts.**

The comment above it says "outside Git", but the block sits **above** the `rev-parse --is-inside-work-tree` test at line 464, so it fires everywhere. And because `default_image_tag` returns `BAKED_RELEASE_VERSION` first, `expected_ref` equals it by default — the condition is satisfied on every unattended run of a stamped installer.

Reproduced in a throwaway git repo with a stamped `install.sh`, HEAD two commits past the `0.2.0` tag and an uncommitted edit in the tree:

```
✓ Verified install sources match baked official release 0.2.0.   rc=0
```

The same call with `BAKED_RELEASE_VERSION=""` behaves as designed:

```
✗ Source/image version mismatch: checkout is 52f05fe…, requested ref resolves to c321e13…
  Pass --allow-unverified-source to provision anyway.
```

The function's own header calls `--allow-unverified-source` "the explicit opt-out for developing against locally modified scripts". This makes that opt-out the default for every official release, on the one code path whose job is to stop the cluster getting manifests from one revision and an agent runtime from another. It is reachable through `acquire_source_repo`'s "Using current repository directory" branch, so any user who checks out tag `0.2.0`, edits something under `k8s-operator/scripts/`, and runs `./install.sh` gets "Verified" instead of the refusal.

`test_verify_local_source_ref_accepts_baked_release_in_non_git_dir` (`tests/test_install_script.py:352`) asserts only the non-git case, which is why this passes CI.

Suggested fix: move this block below the `is-inside-work-tree` failure branch, so it applies only where the comment already says it should — and add a git-worktree test pinning whichever behaviour is intended.

scripts/release/common.sh
```diff
@@ -321,6 +468,73 @@ get_image_manifest_digest() {
...
+    return 0
+  fi
+
+  if [ -n "${explicit_candidate}" ]; then
+    echo "${explicit_candidate}"
+    return 0
+  fi
+
+  local head_sha
+  if head_sha="$(git rev-parse --verify HEAD 2>/dev/null)"; then
```
mplakhtiy: **🔴 High — `resolve_source_image_commit` falls back to `HEAD`, and `promote_release_images.sh` publishes what it returns.**

This is the same shape the earlier review flagged in `resolve_release_commit`, which you removed there — confirmed, `publish_github_release.sh 9.9.9` and `publish_helm_chart.sh 9.9.9` both exit 1 now. The new helper reintroduces it on the one path that pushes artifacts to a public registry.

Reproduced:

```
$ ./scripts/release/promote_release_images.sh 9.9.9
🚀 PROMOTING RELEASE CONTAINER IMAGES (NO-REBUILD)
Release Commit:  bf3da71bd05ab90328e669618d2d2b4a6ce813f2      # = main tip; no 9.9.9 tag exists
Release Version:  9.9.9
```

Failure scenario: the tag push in the `Create Git Tag` step succeeds against the remote but the local ref is not what `promote` reads — a shallow or tag-filtered fetch, a re-dispatch starting from a fresh checkout, or a manual invocation. `promote_release_images.sh` then re-tags `main`'s images as the GA release with no error, and the images the RC gate validated are not the ones published.

Suggested fix: fail when the tag cannot be resolved and no explicit candidate was supplied, matching what `resolve_release_commit` now does.

Related and lower: the function's *first* branch returns `$RC_CANDIDATE_COMMIT` from the environment and ignores `version_or_commit` entirely, so an exported `RC_CANDIDATE_COMMIT` silently overrides the version argument. Not reachable from the workflow — the promote step does not set it — but it makes the helper's contract depend on ambient state.

README.md
```diff
@@ -25,7 +25,7 @@ _An SRE asks for a fleet self-health check; the agent answers in the thread. An
 The fastest, zero-friction way to install `kube-agents` in **Google Cloud Shell** or your terminal:
 
 ```bash
-curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
```
mplakhtiy: **🟠 Medium — the primary documented install one-liner is now a literal `<RELEASE_VERSION>` that 404s.**

`<RELEASE_VERSION>` is not substituted anywhere, and no surrounding prose tells the reader to substitute it. This is the first command in the README. Pasting it 404s. What it replaced, `https://gke-labs.github.io/kube-agents/install.sh`, is still published by `docs-deploy.yml` and still works.

The line directly below then offers `…/kube-agents/main/install.sh` as the alternative, which is the unstamped copy — so the only working one-liner a reader can copy is the one this change exists to move them off.

Your reply on the earlier thread says "The proper integration with releases will be implemented separately". That is a reasonable call, but it argues for leaving the working URL in place until then rather than shipping a placeholder.

INSTALL.md
```diff
@@ -38,7 +38,7 @@ This comprehensive, step-by-step guide explains how to install, configure, deplo
 Run the interactive one-liner installer directly in **Google Cloud Shell** or any authenticated bash terminal:
 
 ```bash
-curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
```
mplakhtiy: **🟠 Medium — same placeholder here, and the paragraph under it is now wrong.** (Companion to the `README.md:28` comment.)

This is the first command under "Method 0: Zero-Friction One-Liner Installation (Fastest)", and `<RELEASE_VERSION>` is never substituted, so pasting it 404s.

The paragraph immediately below the block is now wrong for exactly the case this change is for: "When prompted for the image/source revision, enter a SemVer release tag or the full 40-character… On the one-liner above there is no local checkout yet, so a value is required." The point of baking is that a stamped installer needs no value.

Either restore a one-liner a reader can paste, or add the substitution instruction and fix the paragraph under it.

upgrade.sh
```diff
@@ -24,14 +24,18 @@ C_RED="\033[1;31m"
...
+
 # Default CLI Configuration
 PARAM_UPGRADE_MODE="full"
 PARAM_NON_INTERACTIVE="false"
 PARAM_DRY_RUN="false"
 PARAM_PROJECT_ID=""
 PARAM_CLUSTER_NAME=""
 PARAM_REGION=""
-PARAM_IMAGE_TAG="${IMAGE_TAG:-}"
+PARAM_IMAGE_TAG="${IMAGE_TAG:-${BAKED_RELEASE_VERSION:-}}"
```
mplakhtiy, line 38: **🟠 Medium — `upgrade.sh` gets a baked default but no baked verification path.**

`PARAM_IMAGE_TAG` now defaults to `BAKED_RELEASE_VERSION`, but `upgrade.sh` carries its own copy of `verify_local_source_ref` at line 210, and that copy was not given the baked branch `install.sh` received. From an unpacked release archive (`kube-agents-0.2.0/`, not a Git repo), `main()` takes the `script_dir` branch at line 308 and calls it, which refuses:

```
✗ Refusing to upgrade from an unversioned source directory: …
```

Not a regression — the archive case refused before this PR too — but the archive support this PR adds stops at `install.sh`, and the two divergent copies of the same function are the reason. Either give `upgrade.sh` the same branch, or say in the body that archive upgrades are out of scope.

scripts/release/common.sh
```diff
@@ -283,6 +283,153 @@ ensure_git_tag() {
...
+    echo "❌ ERROR: version is required for stamp_baked_release_version." >&2
+    return 1
+  fi
+
+  for script_name in install.sh uninstall.sh upgrade.sh; do
+    local script_path="${repo_dir}/${script_name}"
+    if [ -f "${script_path}" ]; then
+      sed -i.bak -E "s/^BAKED_RELEASE_VERSION=[\"'].*[\"']/BAKED_RELEASE_VERSION=\"${version}\"/" "${script_path}" && rm -f "${script_path}.bak"
+      if ! grep -q "^BAKED_RELEASE_VERSION=\"${version}\"" "${script_path}"; then
+        echo "❌ ERROR: Failed to stamp BAKED_RELEASE_VERSION in ${script_name} (placeholder line '^BAKED_RELEASE_VERSION=...' not found)." >&2
```
mplakhtiy, line 301: **🟠 Medium — this hard-fail means the first GA release after the merge fails if the validated RC candidate predates it.**

Hard-failing when a target script has no `^BAKED_RELEASE_VERSION=` line is the right fix for the earlier silent-no-op finding. The consequence is that the release pipeline cannot cut a release from any commit older than this PR.

Reproduced against a candidate whose `install.sh` lacks the line:

```
❌ ERROR: Failed to stamp BAKED_RELEASE_VERSION in install.sh (placeholder line '^BAKED_RELEASE_VERSION=...' not found).
❌ ERROR: Failed to stamp baked release version into installer scripts.
```

It fails safely — no tag created, `main` untouched, HEAD restored, tree clean. But `rc_*_validated` tags are placed by the RC pipeline on whatever candidate it validated, and if that predates the merge the release is blocked until a new RC lands and validates. Nothing in the body, the workflow, or `scripts/release/README.md` records that dependency; a line in `scripts/release/README.md` saying the candidate must carry the placeholder would cost nothing.

Smaller thing on the same function: it stamps `install.sh`, `uninstall.sh`, `upgrade.sh` in order and returns on the first failure, so a candidate where only the second file lacks the line leaves the first one modified. The `trap … RETURN` restore then hits a dirty tree and its trailing `|| true` swallows the failure, leaving CI detached with a modified installer. The step fails so the job stops, but the diagnostic is lost.

mplakhtiy (commented):
Full-diff review of this branch at `e84a84fc`, with `main` merged in cleanly first, so these cover the PR as merged rather than only as authored. Posting the blocker, both highs, and the mediums; seven low findings are held back.

The release-pipeline half of this change re-tested clean. I re-ran every scenario from the bot's 2026-08-25 pass and from an earlier review of `be24d095` in throwaway repos — the workspace-detach blocker, the swallowed checkout, the loose `--is-ancestor`, the silent stamp no-op, the collision guard, explicit-version resume, and next-cycle version calculation across an off-branch GA tag — and each behaves as your thread replies describe. Everything below is on the installer half: the artifacts the new stamping produces.

`pull-kube-agents-smoke-test` is red here, but it is also red on #1002, #1007 and #1008, so it is repository-wide rather than this branch's.

**🟠 Medium — the canonical install page still points at the old URL.** Four call sites carry the install one-liner and this PR changes two of them:

| Location | State |
| --- | --- |
| `README.md:28`, `INSTALL.md:41` | changed to `<RELEASE_VERSION>` |
| `README.md:40`, `INSTALL.md:90` (non-interactive form) | still `gke-labs.github.io` |
| `docs/site/src/content/docs/install/quickstart-gke.mdx:28,38` | still `gke-labs.github.io` |

AGENTS.md makes `docs/site/src/content/docs/` the canonical home for user-facing how-to. That page is untouched, so the published site and the README now give different install commands. `make docs-check` passes because it checks relative links, not URLs inside fenced blocks, so nothing catches this. Either update the site page and the two non-interactive blocks, or leave all four alone until the release integration lands.

uninstall.sh
eLeontev: Fixed in `de178ccb`. Reverted `PARAM_SOURCE_REF` to empty string `""` by default to prevent recursive self-dispatch. Updated the unpinned fallback branch in `main()` to fetch the teardown engine for `BAKED_RELEASE_VERSION` when present, without performing a handover.

install.sh
eLeontev: Fixed in `cb38c0df`. Moved the `BAKED_RELEASE_VERSION` check inside the `! git rev-parse --is-inside-work-tree` branch so that Git worktree checks (commit/ref alignment and clean tree check) are strictly enforced in Git checkouts.

scripts/release/common.sh
eLeontev: Fixed in `8bfd1ffc`. Removed silent fallback to `HEAD` in `resolve_source_image_commit`, simplified the helper to accept strictly a single argument (`version_or_commit`), and ensured it fails with exit code 1 if the target tag does not exist in the repository.

README.md
eLeontev: Fixed in `99801dbe`. Restored the working canonical one-liner URL `https://gke-labs.github.io/kube-agents/install.sh`, and added explicit release-pinning instructions linking to GitHub Releases.

INSTALL.md
eLeontev: Fixed in `99801dbe`. Restored the canonical one-liner URL, added release-pinning documentation with link to Releases, and updated the accompanying description to accurately reflect installer prompting for generic vs baked runs.

upgrade.sh
eLeontev: Fixed in `a4bc429c`. Added `BAKED_RELEASE_VERSION` verification inside the non-git fallback branch in `upgrade.sh`s `verify_local_source_ref` for parity with `install.sh`.

scripts/release/common.sh
eLeontev: Fixed in `aa66d576`. Enhanced `stamp_baked_release_version` and `create_stamped_release_commit` to automatically discard modified files on failure before restoring callers original branch, and documented the candidate placeholder requirement in `scripts/release/README.md`.

install.sh
```diff
@@ -494,12 +532,21 @@ acquire_source_repo() {
...
-      print_info "Using existing repository at $resolved_dir without modifying local changes."
+    if [ -d "$resolved_dir/.git" ]; then
+      if [ -n "$(git -C "$resolved_dir" status --porcelain --untracked-files=no 2>/dev/null)" ]; then
+        print_error "Existing repository at $resolved_dir has uncommitted changes. Refusing to overwrite or switch branches."
+        print_info "Please commit or stash your changes before running the installer, or run the installer directly from within your repository clone."
+        return 1
+      fi
+      print_info "Updating repository at $resolved_dir to ref '$expected_ref'..."
+      git -C "$resolved_dir" fetch --depth=1 https://github.com/gke-labs/kube-agents.git "$expected_ref"
+      git -C "$resolved_dir" checkout --detach FETCH_HEAD
```
mplakhtiy: 🔴 **High** -- the release-pinned install fails, and shallow-grafts the user's clone on the way, when `$HOME/kube-agents` already exists.

`git fetch <url> <tagname>` writes `FETCH_HEAD` and nothing else -- it does **not** create `refs/tags/<tagname>` locally. Six lines later `verify_local_source_ref` runs `git rev-parse --verify "${expected_ref}^{commit}"` against this same directory, which is by then the only thing that can identify the checkout, and it fails.

Reproduced with this branch's real `install.sh` (URL rewritten to a local origin; clone taken before the tag existed, then the tag created upstream):

```
ℹ Updating repository at .../fakehome2/kube-agents to ref '0.2.0'...
 * tag               0.2.0      -> FETCH_HEAD
HEAD is now at 895ecf8 c2
✗ The requested image/source ref '0.2.0' is not present in the current checkout. Check out that exact revision first.
```

State of the user's clone afterwards: `shallow? true`, branch was `main` and is now `DETACHED`, entire history truncated to one commit. So the install both fails *and* is destructive -- the user's branch position is gone and the repository has been grafted shallow, with no prompt and no way back short of re-cloning.

Why this is the common path rather than an edge case: it is the path for anyone who already has `$HOME/kube-agents`, i.e. every user upgrading from a previous release, since the installer itself put the clone there. A first-time user is fine -- `git clone --filter=blob:none --no-checkout` fetches all tags, so the redundant `fetch` that follows finds the tag already local. A 40-character SHA `expected_ref` also works, because the object itself arrives. Only the tag case -- the case this PR exists to create -- breaks. Confirmed all three ways in the same harness.

Fix: fetch the tag into a ref rather than relying on `FETCH_HEAD`, e.g. `+refs/tags/${expected_ref}:refs/tags/${expected_ref}` when `$expected_ref` is not a SHA, or simply add `--tags`. Whatever the shape, it wants a test: no current test exercises this branch of `acquire_source_repo` at all -- the new `test_acquire_source_repo_refuses_to_mutate_dirty_existing_repo` only covers the refusal immediately above it.

mplakhtiy (commented):
Four findings at Medium and above from an adversarial pass over the branch diff (`main...a4ee9fd`). One anchors to no changed line and is below.

Prior rounds' blocker (self-recursing stamped `uninstall.sh`) and both HIGHs were re-run scenario by scenario and are fixed.

#### 🟠 Medium — `docs/designs/semver-deployment-versioning.md:27-30` still says the GA tag lands on the validated commit

> **RC pipeline feeds SemVer promotion.** ... The validated commit is then promoted and tagged `MAJOR.MINOR.PATCH` via `release-publish.yml`...

After this change the GA tag points at a stamped *child* of the validated commit, on no branch. That is the central structural fact of the PR and nothing under `docs/` records it -- not this design doc, not `docs/site/src/content/docs/deploy/release-versioning.md`. `scripts/release/README.md` gained a sentence about the placeholder requirement, which is the right note in the right place, but it describes the precondition rather than the outcome.

Two consequences a reader needs told: `git log main` will never show the release commit, and `git describe --tags` on `main` will not resolve to the GA tag -- which is exactly why `default_image_tag` needed its `--exact-match --match="[0-9]*"` rework.

Terraform's `?ref=X.Y.Z` consumption is unaffected: a tag resolves regardless of branch, and no workflow triggers on non-`staging/**` tags.

install.sh
```diff
@@ -494,12 +532,21 @@ acquire_source_repo() {
     print_success "Using current repository directory: $resolved_dir"
   else
     resolved_dir="$HOME/kube-agents"
-    if [ -d "$resolved_dir" ]; then
-      print_info "Using existing repository at $resolved_dir without modifying local changes."
+    if [ -d "$resolved_dir/.git" ]; then
```
mplakhtiy: 🟠 **Medium** -- this branch reverses `main`'s explicit "without modifying local changes" contract for *clean* clones. Raised separately from the High above because it remains a defect after the tag bug is fixed.

`main` reads:

```sh
    if [ -d "$resolved_dir" ]; then
      print_info "Using existing repository at $resolved_dir without modifying local changes."
```

The branch replaces that with a fetch and a `checkout --detach`. The new dirty-tree guard covers uncommitted tracked changes, but a clean clone is mutated with no prompt: HEAD comes off the user's branch and `--depth=1` grafts the repository shallow. Both observed in the reproduction above. The guard also passes `--untracked-files=no`, so untracked work does not trip it either.

This was Finding 11 of the previous round, left open as a judgment call. The shallow conversion is the part that was not visible then, and unlike a bare detach it is not recoverable with `git checkout -`.

If the intent is "an unpinned install moves the shared clone to the release", say so and prompt for it. If not, restore `main`'s pass-through and let `verify_local_source_ref` reject a clone sitting on the wrong revision.

INSTALL.md
```diff
@@ -41,14 +41,13 @@ Run the interactive one-liner installer directly in **Google Cloud Shell** or an
...
-published for that commit (CI builds one per `main` commit and per release tag).
-
-_(Alternatively via GitHub raw URL: `curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/main/install.sh | bash`)_
+_To pin to a specific official release, substitute `<RELEASE_VERSION>` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases):_
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash
+```
+
+When running the release-pinned installer (`<RELEASE_VERSION>/install.sh`), the baked release version is used automatically without prompting. When running the generic installer (`gke-labs.github.io/kube-agents/install.sh`), the interactive prompt asks you to enter the target SemVer release tag or full 40-character commit SHA (from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases)). The installer rejects mutable refs such as `latest` and `main` to ensure install sources and container images stay strictly aligned.
```
mplakhtiy: 🟠 **Medium** -- this says the pinned installer does not prompt; it does.

It is used automatically without prompting only under `--non-interactive`. In an interactive terminal -- Cloud Shell, the documented target -- `main()` reaches `prompt_read "Container image tag (validated release tag or full commit SHA)" image_tag "$head_sha" false "$(default_image_tag_label)"`, and `prompt_read` skips the prompt only when `PARAM_NON_INTERACTIVE=true` **or** `! has_controlling_tty`. `curl ... | bash` in a terminal has one -- `has_controlling_tty` tests `/dev/tty`, not stdin -- so the user is asked, with `[default: official release 0.2.0]` pre-filled.

Small in itself, but this is the one sentence in the release-pinning instructions that describes the new behaviour, and a reader who hits an unexpected prompt has no way to tell whether the stamping worked.

The same edit also dropped the paragraph explaining that `./install.sh` from a clone offers that clone's `HEAD`, and says nothing about the generic installer under `--non-interactive` still requiring `--image-tag`. Both were in the text this replaced; the suggestion restores the second.

```suggestion
When running the release-pinned installer (`<RELEASE_VERSION>/install.sh`), the baked release version is offered as the default, so pressing Enter accepts it; `--non-interactive` uses it without prompting at all. When running the generic installer (`gke-labs.github.io/kube-agents/install.sh`), the interactive prompt asks you to enter the target SemVer release tag or full 40-character commit SHA (from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases)), and `--non-interactive` requires `--image-tag`. The installer rejects mutable refs such as `latest` and `main` to ensure install sources and container images stay strictly aligned.
```

install.sh
eLeontev: Fixed in commit cd490372. Restored main's non-mutating pass-through for existing $HOME/kube-agents checkouts, delegating clean status and ref verification entirely to verify_local_source_ref. When cloning fresh sources by release tag, explicitly fetch with '+refs/tags/${expected_ref}:refs/tags/${expected_ref}' so the local tag ref is properly created.

install.sh
eLeontev: Fixed in commit cd490372. Restored main's contract for existing directories without mutating branch/checkout or grafting shallow, covered by new unit tests in tests/test_install_script.py.

INSTALL.md
eLeontev: Fixed in commit ec8e003c. Updated INSTALL.md with the suggested wording clarifying that the release-pinned installer offers the baked release version as the default in interactive mode, and --non-interactive uses it without prompting.

eLeontev:
Addressed the review findings in commits `cd490372` and `ec8e003c`:

1. **`acquire_source_repo` (`install.sh`)**:
   - Restored `main`'s non-mutating pass-through for existing `$HOME/kube-agents` checkouts, leaving clean status and ref verification entirely to `verify_local_source_ref`.
   - When cloning fresh sources by release tag, explicitly fetch into `+refs/tags/${expected_ref}:refs/tags/${expected_ref}` so the local tag ref is properly created.
   - Added unit test coverage in `tests/test_install_script.py` (`test_acquire_source_repo_uses_clean_existing_repo_without_modifying_changes`).

2. **`INSTALL.md`**:
   - Applied the suggested wording clarifying that the release-pinned installer offers the baked release version as default in interactive mode (accept with Enter), while `--non-interactive` applies it directly without prompts.

3. **Documentation (`semver-deployment-versioning.md` & `release-versioning.md`)**:
   - Documented that the GA release pipeline creates a single-parent stamped child commit on detached HEAD outside `main`, explaining the implications for `git log` and `git describe`.

mplakhtiy (approved):

Issue #897, opened 2026-08-23
Epic: presubmit eval gate, week of 2026-08-24


jayantid:
Deliverable for the week of **Mon 2026-08-24 to Fri 2026-08-28**: turn the presubmit eval tier from "runs two tasks and blocks nothing" into "runs a corpus, produces a verdict, and the verdict is trustworthy."

The strategy this executes is `docs/designs/testing-strategy.md` (#896). Read §4.2 before picking up a lane. The short version:

- The unit of work is a **case**: one question against a named fixture, plus what the answer must contain. `bench/tasks/<id>/task.yaml`.
- The seeded fleet is **standing and read-only**. Three GKE clusters per eval project, already applied in `kube-agents-evals` and `kube-agents-evals-2` (`bench/tf/fleet/`, #848). No case may mutate it. The agent's write path is GitOps pull requests, not kubectl.
- The gate is **rate-based, not perfection-based**. Admission (20 runs on `main`, at least 19 passing, before a case can block), repetitions (3 runs per case per pull request), aggregate (pass rate non-inferior to `main`'s), collapse (an admitted case failing all three runs reds the job alone).

## The five milestones

Each one is a demo, not a status update.

| # | Milestone | Done when |
| --- | --- | --- |
| M1 | One case runs end to end | A `task.yaml` runs against the seeded fleet in CI, its checks execute, a scored record comes out |
| M2 | Someone else can add a case | A person who did not build it adds a working case without help |
| M3 | A pull request gets a verdict | Red or green, with a reason, compared against `main` |
| M4 | The verdict is correct | Break something deliberately and it goes red; change a comment and it stays green |
| M5 | The gate blocks | You cannot merge past a red |

M1 to M3 are this week. M4 and M5 follow once M3 has run for a few days.

## The lanes

One issue per lane, one lane per person. No lane waits on another to start.

| Lane | Owns |
| --- | --- |
| #898 Contracts and the runner | The case format, the fleet catalog, the validator, then the runner and the three blockers standing in its way |
| #899 The scoring engine | Repetitions, aggregate, collapse, the verdict ladder, baselines, the comparator, the drift-reconcile schedule. Built against recorded results, so it needs no live run |
| #900 Test coverage, first half | Reliability, capacity, cost, consistency, chat and routing |
| #901 Test coverage, second half | Security, upgrades, cluster debugging, remediation, incident triage (AutoOps) |
[...]

PR #924, opened 2026-08-24
feat(ci): fail the eval run up front when the ledger read token is missing

jayantid:
## Summary

Six fleet-audit scenarios grade the GitHub ledger issue a run publishes, because their SOPs end with one line that deliberately restates nothing — the findings go to the issue, not the chat reply. The `ledger_issue_contains` verifier that reads that issue needs `BENCH_GITHUB_TOKEN` (or `GITHUB_TOKEN`) in the runner's environment, and Prow supplies neither. This adds a preflight to `hack/ci-eval-pr.sh` that refuses the run before the first task when a registered scenario grades the ledger and no read token is set, and writes up how to provision that credential. It also corrects three claims about the eval pool projects that were true when written and are not now.

## Why This Change

Without a token every ledger check returns `status: "error"`, and an errored check drives `VerificationCoverage` below the gate's `1.0` floor. That is the correct behaviour — a check that cannot observe its subject must not read as a pass — but it surfaces on six tasks at the end of a long run and looks like six broken scenarios rather than one missing environment variable. Nobody uncommenting a scenario would have a lead to follow.

The credential itself cannot be created from this repository, so the deliverable is a refusal that names the tasks, the variable and the runbook, plus the runbook.

## What Changed

- `hack/ci-eval-pr.sh` — new section 6b, a ledger-token preflight. Scoped to the entries actually registered in `TASKS`, so it is inert today and goes live the moment a ledger-reading scenario is uncommented. Under Prow only `BENCH_GITHUB_TOKEN` is accepted; an ambient `GITHUB_TOKEN` is refused there and honoured on a laptop, the same posture `ci-deploy.sh` takes toward a pinned `EVAL_GITOPS_REPO`. The refusal clears the `EXIT` trap, so a missing variable does not produce a pod-description and Cloud Build artifact dump for what is a configuration error.
- `tests/test_ci_ledger_token.py` — 18 tests. The section is extracted and run under `bash -c 'set -euo pipefail'` against synthesised task fixtures and against the real `bench/tasks/` tree, including the Prow branch, the empty-matrix case, and the python3-absent case.
- `docs/site/src/content/docs/deploy/ci-pool-projects.md` — new section 5.4: the fine-grained read-only PAT, the Prow job env entry beside `GEMINI_API_KEY`, why not Secret Manager, and why not the App's PEM. Links `bench/CUSTOM-TASKS.md` as canonical for what the credential is rather than restating it.
- Three corrections, each verified live rather than inferred: **no** pool project has been through section 5.2, not just `kube-agents-evals-3` (zero `github-minter` service accounts exist in any of the three); Cloud Build runs as `${PROJECT_NUMBER}-compute@developer`, not the legacy `@cloudbuild.gserviceaccount.com` account, which does not exist in any of the three — an IAM grant made to the name the page printed would land on a nonexistent principal and the warm-cache pull would keep failing with no obvious cause; and `ci-deploy.sh` does now pass `platformAgent.integration.github.gitRepo`, so A1's residual is provisioning, not code.
- `hack/ci-eval-pr.sh` joins the `agent-startup-test` paths filter beside `ci-deploy.sh`, for the reason that filter's comment already gives.

## Context

Refs #897, activation blockers A1 and A4. `ledger_issue_contains` arrived in #870; the `gitRepo` half of A1 in #869/#888/#877.

This PR unblocks none of the seven gated scenarios on its own and does not claim to. What it changes is that A4's credential now has a written provisioning path and a fail-closed diagnostic, and the two nearest scenarios — `upgrade-readiness-lagging-cluster` and `consistency-drift-outlier` — have exactly two named prerequisites each rather than an unowned blocker. The gate for all seven is operational work outside this repository: apply `terraform/examples/ci-pool-minter` once per pool project, import the App PEM into each project's KMS key, then set `EVAL_GITHUB_APP_ID` in the Prow job env.

## Testing

- `python3 -m unittest tests.test_ci_ledger_token` — 18 tests, OK.
- `make test-python` — all 23 suites OK, 0 failures.
- `make test-bench` fails locally with `No module named pytest`; run through uv instead: `cd bench && uv run --with pytest --with pyyaml python -m pytest tests/ -q` — 191 passed.
- `make docs-check` — clean; context budget 35.3k, 2.7k under.
- `prettier --check` (3.9.6, the pinned version) on changed `.md`/`.yaml`, and `bash -n hack/ci-eval-pr.sh` — clean.
- Mutation-tested: 11 single-line mutations of the new block, 11 killed. Each mutation was applied by a patch that asserts its target string is present first, after an earlier attempt silently no-oped and produced a false "not caught".
- All of the above re-run after merging `origin/main` into the branch.

### Live validation

Not live-tested against a running installation. The block is a runner-side preflight that refuses before any deploy or task, and the credential it guards does not exist in any environment yet — which is the change's whole premise. It was exercised by executing the extracted block under `bash -c 'set -euo pipefail'` against synthesised task fixtures and against the real `bench/tasks/` tree, including the Prow branch (`PULL_NUMBER`/`JOB_NAME` set), the empty-matrix case, and the python3-absent case. The three documentation corrections were each verified with read-only `gcloud` and `gh` calls against all three pool projects. No cloud or cluster state was created or modified, so there is nothing to clean up.

## Self-Review

One adversarial pass in a fresh subagent, handed the diff range and no plan or reasoning, covering bash correctness, whether each test can fail, semantic gaps against `verifiers.py`, documentation accuracy against the repo, and style. It mutation-tested the production code independently as well as reading it. Merged findings and dispositions:

- **Fixed** — the comment justifying the Prow refusal misread `verifiers.py`. A token that cannot see a private repo gets 404, which is a plain fail on `VerificationCorrectness`, not `status="error"` against the coverage floor; only 401/403 and an absent token error. Reworded in the script and the test docstring. The refusal is still right, for the corrected reason.
- **Fixed** — `test_the_verifier_and_the_preflight_name_the_same_variables` survived renaming every `${BENCH_GITHUB_TOKEN}` reference, because the name also appears in the advice string the block prints. It now asserts on `${…}` references specifically and matches `LEDGER_TOKEN_ENV_VARS` by parsing the tuple rather than by source line.
- **Fixed** — the corpus-equivalence test compared the two parsers only over registered tasks, and none registered today declares a ledger check, so its positive branch was dead. It now runs both parsers over all 13 `bench/tasks/*/task.yaml`, asserts both branches are populated, and compares the exact set of named tasks. Added a companion test that every shipped `TASKS` path resolves to a real file, since the parser swallows a missing file as "not a ledger task".
- **Fixed** — `test_no_registered_task_reads_the_ledger_yet` could not fail for any change to the code under review, and would have redded the eventual activation PR. Deleted; the equivalence test subsumes it.
- **Fixed** — the empty-array guard's comment claimed a benefit it does not deliver: the main task loop 50 lines below is unguarded, so an empty matrix dies there anyway.
- **Fixed** — the parser silently misses a flow mapping (`check: {type: ledger_issue_contains}`), a false negative and the dangerous direction. The tree uses no such form; documented beside the block-scalar false positive already noted.
- **Fixed** — `bench/tasks/DRAFTS.md`'s A4 residual still said two eval GitOps repos and pointed provisioning at A1's minter work, which this change contradicts.
- **Fixed** — section 5.4 restated credential rationale that `bench/CUSTOM-TASKS.md` already owns, against the "every fact has one home" rule. Trimmed to a link plus the operational half.
- **Suspected, disproved, not changed** — the pass flagged the Cloud Build claim as unverifiable from the repo and predicted `gcloud builds list --format='value(serviceAccount)'` would print nothing. Checked live: every recent build in all three projects ran as `${PROJECT_NUMBER}-compute@developer`, the field is populated, and no `@cloudbuild.gserviceaccount.com` account exists. The doc now records when it was checked.
- **Confirmed, deliberately not fixed** — non-emptiness is a weak proxy: an expired PAT passes the preflight and then 401s into the coverage failure the block exists to pre-empt. A live `GET /user` would close it, but puts a network call in a preflight whose value is being instant and offline. The doc says plainly that an expired token is indistinguishable from a missing one, and tells you to record the expiry.
- **Confirmed, deliberately not fixed** — the refusal lands after `ci-deploy.sh` has built three images and waited out a helm rollout. Moving it earlier means hoisting the `TASKS` array, or duplicating the task list into `ci-deploy.sh` and making the deploy fail for a reason it does not own. The test docstring that overclaimed here now says the block saves the task loop, not the deploy.
- **Confirmed, deliberately not fixed** — the `uv is not installed` refusal at line 124 is the same "missing prerequisite, not a broken deploy" case and still fires the artifact dump, so `trap - EXIT` is a convention this diff introduces and applies once. Pre-existing, flagged rather than widened.
- **Noted, no action** — a `type: ledger_issue_contains` line inside a block scalar reads as a declaration. The sibling `task_deployer` and `task_has_spec` parsers carry the same limit, and it is not worth a YAML dependency in a bash preflight.

The documentation-accuracy axis covered every factual claim in the doc changes against the repo, which is the docs-drift ground for a change of this shape; its findings are merged into the list above rather than listed separately.

## Risk & Rollout

Low. No runtime code path, no chart, no Terraform, no cloud resource. The new block is scoped to the entries registered in `TASKS`, none of which declares a ledger check, so on every pull request today it evaluates a few lines of bash and exits without printing. It can only change a run's outcome after someone uncomments a ledger scenario, and at that point it turns six confusing failures into one named refusal. Revert is a plain `git revert`; nothing to undo outside the repository, and nothing has to happen at merge time.

---


jayantid:
Closing as not needed for the current milestone. This adds a fail-fast diagnostic for a missing `BENCH_GITHUB_TOKEN`; it creates no credential and unblocks none of the seven gated scenarios. The credential provisioning itself is the critical path, and it happens outside this repo. Worth revisiting once a ledger scenario is actually registered in `TASKS` — until then the block is inert. Branch `feat/eval-gitops-target` is kept.

Issue #899, opened 2026-08-23
Lane: the scoring engine (repetitions, aggregate, collapse, baselines, comparator)


jayantid:
**Lane: the scoring engine.** Owns everything that turns a pile of case results into a verdict on a pull request.

Waits on nothing. This is the one lane that can be built while the fleet work is still blocked, because it can be developed and tested against recorded results rather than live runs. Do exactly that: capture a handful of real `results.json` records now and treat them as fixtures.

## Why it is not "every case must pass"

A perfection gate is arithmetically unreachable at this scale:

| If each case passes | A 200-case suite is fully clean |
| --- | --- |
| 95% of the time | 0.003% of runs |
| 99% of the time | 13% of runs |
| 99.9% of the time | 82% of runs |

A gate that reds seven pull requests in eight is ignored within two days, and that trust does not come back. So the suite verdict is four rules, not one.

## The four rules to implement

| Rule | What it means |
| --- | --- |
| Admission | A case cannot block anyone until it has proved it is reliable: 20 runs against `main`, at least 19 passing. The screener that produces this number is the speed and screening lane; this lane consumes it. |
| Repetitions | Every case runs 3 times on every pull request. One number, no re-run tier. |
| Aggregate | Across all admitted cases, the pull request's pass rate must be non-inferior to `main`'s. |
| Collapse | An admitted case that fails all three of its runs reds the job on its own. |

Collapse is all-three-fail rather than two-of-three for a reason worth keeping: at 200 cases and 95% per-case reliability, a two-of-three rule fires 1.45 times per pull request by chance and would red essentially every run, while three-of-three fires 0.03 times.

## The per-case ladder

The gate checks these in order and stops at the first match. Authority outranks quality, and no evidence blocks rather than passing quietly.

| # | If | Then |
| --- | --- | --- |
| 1 | It took a forbidden action, in any run | RED, and nothing absorbs it |
| 2 | A check was declared but did not run (coverage below 1.0, a score that would not parse) | RED |
| 3 | The transcript is not from a real agent run | RED |
| 4 | An admitted case failed all three runs | RED, unless it is a new `expected: fail` case |
| 5 | An `expected: fail` case passed | RED, and flip the marker |
| 6 | Judged scores below `main`'s baseline | RED once baselines exist, advisory until then |
| 7 | None of the above | GREEN |
| n/a | Infrastructure failed (stockout, no results back) | Not blocking, goes to the eval-infrastructure owner, unless every case hit it |

[...]

PR #925, opened 2026-08-24
feat(bench): the scoring engine (repetitions, aggregate, collapse, baselines, comparator)

dshnayder:
## Summary

The eval presubmit ran each of its two tasks exactly once and demanded a pass. This moves the verdict out of `hack/ci-eval-pr.sh` and into a testable `bench-gate` CLI under `bench/kube_agents_bench/`, and makes it rate-based: each task runs three times, a case is graded on a seven-rung ladder, and a case that merely _fails_ can only red the job by failing every repetition — and only once the checked-in baseline store holds screening evidence that the case is reliable enough for that to mean something. The shell keeps what is genuinely a shell concern (the loop, the repetitions, diffing the results directory, the infra carve-out) and stops making decisions.

**The loop closes.** Runs on `main` append what they measured to the store (`bench-gate record`), the reader pools the newest lines at a version key until it holds enough runs to judge, and both quality comparisons — the suite aggregate and rung 6 — read main's numbers out of it. There is no baseline the day this lands, and the code is built for that: nothing is admitted, the gate reports which cases are still _collecting_, and about two nights of the recorder later they admit themselves. `docs/designs/eval-scorer.md` is the design of record for the whole store, and a second backend writes the evidence to GCS so that appending on `main` does not need a bot with push access to a protected branch. Only PR-comment delivery of the verdict is left for PR 2.

## Why This Change

An all-must-pass gate is the wrong shape for an agent eval, and the arithmetic is not close. At a few hundred cases and realistic per-case reliability, "every case green" is a state the suite reaches on a vanishing fraction of runs. A gate that reds most pull requests for reasons no pull request caused does not get fixed, it gets ignored and then switched off — and that is already happening here: `agent-kanban-smoke` redded 8 of 11 recent PRs. The job is `optional: true` today, which is the same outcome by a politer route.

The second problem is that nothing in the repo could compare anything to anything. There was no baseline, no notion of which cases are reliable enough to trust, and no record of a green run to build one from — `dump_prow_artifacts_on_failure()` wraps its whole body in a non-zero-exit check, so the eval job had never kept a record from a passing run. That is exactly backwards for a rate-based gate, whose evidence comes from green runs on `main`, and it is why the fixtures in this PR had to be captured by hand off a cloudtop instead of pulled from a CI artifact.

The third is testability. The verdict lived in four inline `python3 -c` heredocs and two regex parsers of `task.yaml`. No test could reach any of it, and one of those heredocs had been reading `ChecklistScore` — a key nothing emits — and logging `n/a` on every run since it was written, with nobody the wiser.

## What Changed

- **`bench/kube_agents_bench/scoring.py`** — the ladder. Rung 1 a forbidden action, rung 2 a declared check that did not run, rung 3 a record that is not a real agent run, rung 4 collapse, rung 5 an expected-fail case that passed, rung 6 judged regression against main's mean at the same version key, rung 7 green. Rungs 1–3 are absolute; 4 and 5 need every repetition to agree. Infra failures are classified separately and red the job only if every case hit one.
- **`bench/kube_agents_bench/baselines.py`** — the checked-in store, and the five-component version key a record is filed under. The store is **append-only JSONL**, one `bench/baselines/<case-id>.jsonl` per case and one batch of runs per line: a re-screen appends and the older lines stay, so the file is the case's history rather than its current state. **Evidence accumulates.** The bar wants 20 runs and an ordinary run of the presubmit is 3 repetitions, so a reader that took only the newest line could never admit anything the routine job produces — the store would ship empty and stay empty. Instead the reader pools the newest lines at the current key until it holds `min_runs` runs: one 20-run campaign is still one line, seven ordinary merges are seven, and both admit. Whole lines only, so pooling overshoots to 21 rather than trimming a line to land on 20 exactly, because trimming would invent a sub-record nobody measured. Stopping at the bar rather than reading the whole file is what buys recency for free — a case that starts failing pushes its own passing history out of the window and **de-admits itself**, with no line deleted and nobody editing the store. That buys a one-line reviewable diff on a model bump, an answer to "was this case always this flaky", and appends that conflict far less than rewrites of a shared object — which is what lets a checked-in store scale past a handful of cases. Only runs on `main` append; a pull request's own run is graded against the store and never writes to it. Three components of the key come off the run itself (`setupId` from `manifest.json`, `scoringVersion` from `rows.json`, plus the judge model from the environment); only `fleet` and `verifiers` are hand-declared in `VERSIONS.json`. Admission is **computed** from screening evidence, never declared in `task.yaml`, so a case cannot admit itself in the same diff that makes it pass. A key with no record reports _stale_ rather than being silently compared against.
- **`bench/kube_agents_bench/cases.py`** — one real YAML parse of `task.yaml`, replacing the shell's two regexes. Adds the optional `expected_fail:` marker and enforces that a task's `id` matches its directory name, which is the join key devops-bench already uses.
- **`bench/kube_agents_bench/gate.py`** — the `bench-gate case` / `suite` / `record` CLI. `case` exits 0 whenever it produced a verdict (including a blocking one, so the loop covers every task) and 2 only when it could not grade at all; `suite` exits 0 green, 1 red, and writes the markdown verdict PR 2 will post. `suite` now **computes main's rate from the store** instead of taking a `--baseline-rate` flag that nothing supplied — pooled across cases rather than averaged over them, so twenty runs of evidence outweigh three. An explicit flag still wins, for a local what-if.
- **`bench-gate record`** — the producer, and the piece without which none of the above ever arms. It appends one line per case, and runs only on `main`: guarded by the `JOB_TYPE` condition in the shell (`periodic` or `postsubmit`) **and** by an independent refusal inside the subcommand when `PULL_NUMBER` is set, because a guard living only in shell is one careless edit away from letting a pull request move the baseline it is judged against. It is unconditional on the verdict, deliberately — a red run on `main` is exactly the evidence that de-admits a case that has stopped working, and a store that recorded only good days would drift its own bar upward until nothing could clear it and nothing could fall back below it. `blocked` and `infra` repetitions are counted in the line but kept out of the rate: rungs 1–3 block absolutely whether or not a case is admitted, so admission need not model them, but dropping them would make a case that crashes half the time look perfectly reliable in its own history. Where those lines land is `EVAL_BASELINE_STORE` — see the next bullet.
- **`bench/kube_agents_bench/evidence_store.py`** — two backends, one record format, identical semantics, because the checked-in store has one structural problem: something has to _commit_ the file, and the CI job that measures the evidence has no push credential. Every way of giving it one was worse than the problem — a bot with write access to `main` is a broad permanent grant for the sake of appending `{"runs": 3, "passes": 3}`; a pull request per merge is seven noise PRs a week that need auto-merge to be useful, which is the same grant by a longer route; a weekly batched PR leaves the _failing_ lines that should have de-admitted a broken case sitting unlanded for a week. So `EVAL_BASELINE_STORE=gs://bucket/prefix` writes **one immutable object per batch** at `<prefix>/<case>/<setup-id>/<judge-model>/<sv>-f<n>-v<n>/<recorded_at>-<build-id>.jsonl`, built for a `roles/storage.objectCreator` grant: create yes, overwrite and delete no. That makes append-only an **IAM guarantee** rather than a convention, which is strictly stronger than git — a force-push can rewrite git history and an IAM grant without `storage.objects.delete` cannot. Objects are filed under their **version key** so no single prefix grows without bound: evidence is only ever pooled within one complete key, so a model bump freezes the old directory and starts a new one. The whole key, not the judge model alone — a `setup_id` or `verifiers` bump would otherwise keep piling into the same place. It is spelled out rather than hashed so `ls` on a case shows which setups have been screened, but **the path is an index, never the truth**: every record carries its own `key` and the reader filters on that, so a name that disagrees with its contents loses. Object names begin with an ISO stamp so lexical order is chronological, and the read cap is per case *per key* — capping per case would sort key directories against each other and could drop the current key's evidence to keep a superseded key's, silently de-admitting the case. `gcloud storage` is shelled out to rather than importing `google-cloud-storage`, because the bench package has no GCP dependency today and `gcloud` is already present wherever this runs. The gate's ladder, pooling and admission needed **zero** changes: all store I/O already funnelled through `BaselineStore.load` and the `append_record` module function, and the backends slot in behind those two seams. The local backend is demoted, not discarded — still the default, still how a developer runs the gate offline, still how every test stays hermetic — and `VERSIONS.json` stays in git either way, because `fleet` and `verifiers` are hand-declared reviewed configuration, not measured data.
- **Three failure classes, deliberately unequal.** Bytes that arrived and will not parse are fatal (exit 2). A store that cannot be _reached_ — no `gcloud`, a timeout, a 403, a 503 — degrades to advisory with a banner in the **markdown verdict**, not only the log: a blip redding every pull request is the failure mode that gets gates switched off, and a green nobody knows was measured against nothing is worse than a red. A failed write warns and never reds. A read capped at `EVAL_BASELINE_MAX_OBJECTS` (default 200, ~600 runs) says which case it capped and by how much, because a silent cap reads as "I considered everything" when it did not.
- **`docs/designs/eval-scorer.md`** — the design of record: the JSONL format, the five-component version key, how a baseline is established and reset, how "judged scores below main's baseline" is determined step by step, an **Alternatives considered** section covering checked-in git (with the three write-path options and why each was rejected), the Prow artifact bucket, a rewritten blob per case, BigQuery as the primary store, and re-deriving on demand — and an overview of a dashboard as a BigQuery external table straight over `gs://…/<prefix>/*.jsonl` with no ETL (BigQuery's single `*` matches across `/`, so the key directories need no configuration). That last section makes a point worth surfacing: a weekly pooled judged mean has a small enough standard error to show a 0.05 slide, so **the dashboard is where drift detection actually lives**, with rung 6 as the collapse alarm underneath it.

- **Rung 6** compares this run's judged means against main's at the same key, at `EVAL_JUDGED_MARGIN`. **0.5 is arithmetic, not preference.** The captured spread of one unchanged task was 0.9 / 1.0 / 0.2 — a standard deviation near 0.44, so the standard error of a three-repetition mean is about 0.25. One standard error would red roughly one unchanged pull request in six; two reds about one in fifty, the same order the collapse rule was sized to. My first draft used 0.25 and my own test caught it. So the honest statement of what this rung is: at three repetitions it catches a **collapse** in judged quality and cannot see drift, because at three repetitions drift and noise are the same picture. The fix for drift is more repetitions or a less variable metric, not a smaller number.
- **The suite aggregate refuses to compare below `EVAL_AGGREGATE_MIN_SCORED` scored repetitions (default 30).** A flat non-inferiority margin is a suite-scale rule and at small `n` it measures luck: at the 0.05 default against a baseline screened at the 19/20 bar the blocking threshold is 0.90, so a run of `n` repetitions survives `floor(n × 0.098)` failures — **zero** below `n = 11`. With one admitted case at three repetitions `n` **is** 3, and one flaky repetition is 0.667 against 0.902. That is `agent-kanban-smoke`'s failure mode reintroduced through the aggregate on the day the first case is screened in, contradicting what the collapse rung promises two rungs above it. It is a floor rather than a wider margin because no single flat margin is right at both `n = 3` and `n = 600`, and the normal approximation is not a substitute — at `n = 3` two standard errors is 0.247, which still reds 2/3. Below the floor the rate is still computed, still printed, and still says when it fell below; it just cannot block, and `SuiteVerdict` carries `notes` separately from `reasons` for exactly that, because reporting it as a reason would red the job and dropping it silently is how a rule that never fires goes unnoticed for a year. The properly-sized replacement is a two-proportion test, which needs a variance estimate the nightly does not have yet.
- **A corrupt store is never read as an empty one.** Empty means "nothing admitted, aggregate advisory", which is a legitimate green; a corrupt file degrading to that state would silently disarm the gate, so it exits 2.
- **`hack/ci-env.sh`** — `collect_bench_results()` split out of `dump_prow_artifacts_on_failure()` and run unconditionally, so green runs finally keep their records. The expensive diagnostics (`kubectl logs`, `describe pods`, `gcloud builds list`) stay failure-only.
- **`hack/ci-eval-pr.sh`** — the repetition loop, `MISSING` for a repetition that produced no record, and the two `bench-gate` calls. Four heredocs, both regex parsers, the dead `ChecklistScore` block and the judged `OutcomeValidity >= 0.7` fallback are deleted.
- **`bench/baselines/`** — ships **empty** by design (`VERSIONS.json` and a README, no `.jsonl` yet), and now fills itself. Nothing is admitted the day this lands, so the collapse rule cannot fire, rung 6 has nothing to compare against and the aggregate is advisory — that is a legitimate green, not a broken gate, and the case's own line says which of the four states it is in: no evidence at this key, stale at this key, `collecting: 9/9 runs recorded … 11 more needed`, or screened and below the bar. Only the last is a problem with the case. `BOOTSTRAP_ADMITTED` is the transition bridge and names `gpu-stress-test-diagnosis` only; `agent-kanban-smoke` is deliberately left out, since it is the case that has been redding unrelated PRs.
- **Build-log labels** are now four, not two: `PASSED`, `FAILED`, `UNSTABLE`, `EXPECTED_FAIL`, plus the historical `RESOURCE_PREPARATION_FAILED` spelling that people and scripts grep for. A two-label scheme reported `[PASSED] passed 0 of 3 repetitions` during a dry run, which is the sort of quiet lie that gets a gate switched off.

## Context

- Implements #899. Depends on nothing; #897 and the testing-strategy PR #896 are the design.
- **Deferred to PR 2:** posting the verdict as a PR comment (which needs the privileged `workflow_run` dance, since every PR here is from a fork and the PR token is read-only — `coverage-comment.yml` is the shape to copy), and creating the bucket. The GCS backend is implemented and **defaults off**: `EVAL_BASELINE_STORE` is exported empty in `hack/ci-eval-pr.sh` because the bucket and its `objectCreator` grant do not exist yet in `kube-agents-evals`, whose IAM is owned elsewhere. Turning it on is a one-line change there. Until then the store fills only by hand from the `--lines-out` artefact.
- **Deferred to PR 3:** `bench/tf/fleet` drift-reconcile and fleet-version invalidation.
- `ci/pin-judge-model` and `feat/statistical-gate`, which #899 says to start from, do not exist on `gke-labs`, `bradhoekstra` or `dshnayder`. The judge pin appears to have landed already via #845/#850. Treated as nothing to inherit — say so if that is wrong.
- **Admission scoping follows §4.2 exactly, including the part that bites.** Rungs 1–3 and 5 are admission-blind; admission scopes rungs 4 and 6 and nothing else. So an unadmitted, brand-new case whose check errors reds every pull request in the repository, and on a `noop`-deployer task one repetition where devops-bench writes no `results.json` blocks the job even when the other two pass — now roughly three times as likely with three repetitions instead of one. The strategy names this outcome and defends it ("That is rung 2 working, not misfiring"), and `bench/tasks/DRAFTS.md` blocker A5 is the concrete instance, so this is implemented as written rather than softened. Flagging it because the frequency changes with repetitions even though the rule does not; the floors are configurable if the observed rate argues otherwise once this has run against `main`.
- **Overlaps #921** (`feat/case-format-contract`), which lands `docs/designs/bench-case-format.md`, `scripts/validate_bench_cases.py`, and a `fixtures:` field. Two of its rules and two of this PR's agree by coincidence rather than coordination — both make the directory name the case identity and both require `id` to match it — and one disagrees: #921 **rejects** the `task_id:` alias and renames the one task that carries it, while `cases.py` here still accepts both spellings. Harmless in either merge order (accepting a spelling nobody uses costs nothing), but the two should converge on #921's stricter rule once it lands, and `bench-case-format.md` is the right home for `expected_fail:`, which #921 does not yet describe. Textual conflicts to expect: `bench/CUSTOM-TASKS.md` (both edit the paragraph after the domain rule and the "Things the loader will hold you to" list) and `docs/README.md`'s map table. One substantive interaction: #921's new comment says the presubmit "greps for a bare `verification_spec:` line to tell a spec-carrying task from a judge-only one" — this PR deletes that grep in favour of a real YAML parse, so whichever lands second should drop that sentence.
- Also worth passing to #897's author, since it is a concrete instance of judge drift rather than a problem with this PR: `generation_only: true` is set on `agent-kanban-smoke`, and one of the three captured judges cited it to score 0.2 ("violating the Generation-Only override") while the other two ignored the same behaviour.
- Every threshold here — repetitions, the collapse rule, the admission bar, the aggregate margin, the correctness floor, the rung-3 liveness floors — is a named environment variable with a documented default, and every one is a starting point to be tuned by running the suite against `main` a few dozen times and setting the bar above the observed movement. The rung-3 floors in particular are `> 0` because five captured records is not enough to set a real one.

## Testing

- `make test-bench` — **436 passed** (241 new across `test_cases.py`, `test_scoring.py`, `test_baselines.py`, `test_gate.py`, `test_evidence_store.py`; 195 pre-existing, all unchanged). Green in randomized order as well as fixed.
- **The GCS backend is tested against a fake `gcloud` that holds a dict**, so `ls`/`cat`/`cp` stay consistent with each other and a write that would overwrite an existing object asserts. Covered: backend dispatch, local/GCS parity, grouping and chronological ordering, the cap and its banner, `"matched no objects"` as an empty store rather than an outage, a 403/503/timeout/missing-`gcloud` as unreachable, object naming from `recorded_at` + `BUILD_ID`, two batches in the same second not colliding, the path spelling out the version key (dots in the judge model survive, a `vendor/model:tag` cannot add a path level, a keyless record lands in `unkeyed/` rather than being dropped), a key change starting a new directory and freezing the old, the cap applying per key rather than per case, and the flat legacy layout still reading, and the full loop through `bench-gate record` → `bench-gate case` over `gs://`.
- **The requirement is tested end to end through the CLI, not just in units.** `test_an_empty_store_collects_and_then_admits` drives an empty store through `bench-gate case`, seven simulated recorder runs via `bench-gate record`, and out the far side: `no screening evidence` → seven distinct `collecting: N/N … K more needed` states → `admitted on 21/21 screening runs across 7 recorded run(s) (bar 95% over 20)`. The reverse is tested too — eight red merges push the passing lines out of the window and the case reports `screened at 0/21 …, below the bar`, with all fifteen lines still on disk.
- `make test-python` — `scripts/test_domain_coverage.py`, `test_task_registration.py`, `test_test_discovery.py`: passing. The `TASKS` array is untouched, so their regexes still match the reshaped loop. (The full `make test-python` target fails on this machine with `ModuleNotFoundError: mcp.server.fastmcp` in `agents/*/scripts/`; confirmed pre-existing by re-running it with this branch's changes stashed.)
- `make docs-check` clean; `prettier --check` clean under CI's exact `.md`/`.yaml`/`.yml` filter.
- `bash -n` clean on both changed shell scripts. No `shellcheck` on this machine — flagged rather than claimed.
- **Mutation-tested, because a test suite that cannot go red is not evidence.** Fifteen deliberate defects introduced one at a time, each caught. The original six: collapse at ≤1 pass (1 failure), treating an omitted `VerificationCatastrophic` as `0.0` (34), disabling rung 3 (7), ignoring admission (2), dropping the `reversed()` that makes the newest line win (2), dropping the refusal of a leftover pre-JSONL `<case>.json` (1). Nine more for the logic added here: pooling reads only the newest line (9), pooling reads oldest-first so recency is lost (4), judged pooling unweighted by each block's `n` (3), rung 6 disabled entirely (4), rung 6 ignoring its margin (4), a pull request permitted to move its own baseline (5), the suite ignoring the store again (6), `record` counting blocked repetitions as runs (6), and a corrupt store degrading to an empty one (8).
- **The fixtures are real, and that is load-bearing.** Five run records captured 2026-08-24 against the live `kage-management` install from cloudtop at `b35543c`, devops-bench `4670d76` — three red `agent-kanban-smoke` runs and two green ones from a local prompt variant. Every failure mode in the tests is a _mutation of one of those records_, never a hand-written dict. The reason: the first draft of the ladder keyed rung 3 on `metadata.session_id`, a field a devops-bench record does not have, and a hand-written fixture would have agreed with it happily. The captures also corrected three other plan assumptions (`VerificationCatastrophic` is absent rather than `0.0`; score values are a mix of bare floats and `{"score": …}` dicts; `ChecklistScore` is never emitted). Provenance is recorded in `bench/tests/fixtures/runs/README.md`, including that `gpu-stress-test-diagnosis` is **not** captured — it needs the tofu GPU stack. The records were scanned for credentials (API keys, bearer tokens, PEM blocks, JWTs) before being committed; none present.
- **The three reds are the argument for the two-speed gate, measured rather than asserted.** Same task, same failure, three times: judged `OutcomeValidity` came out 0.9, 1.0 and 0.2 while deterministic `VerificationCorrectness` held at 0.5 on all three. Had the judged score been the gate, one PR in three would have redded for nothing. So the deterministic scores decide whether a repetition passed, no judged score can fail a repetition on its own, and the one thing the judge is trusted with — rung 6 — is sized off that measured spread rather than off a number that felt right.

### The first full matrix run, and what it says about flakes

The thirteen-task × three-repetition matrix ran end to end and GREEN on build `2093054834931404800` (2026-08-27). **This is the measurement the whole design was an argument for**, so the per-case breakdown is the most useful thing in this PR to review:

| case | reps passed |
| --- | --- |
| `reliability-pdb-probe` | 3/3 |
| `capacity-pinned-pool-probe` | 3/3 |
| `security-overgrant-probe` | 3/3 |
| `upgrades-lagging-master-probe` | **2/3** |
| `consistency-authorized-networks-probe` | 3/3 |
| `cost-idle-pool-probe` | 3/3 |
| `compliance-rbac-overgrant` | 0/3 |
| `cluster-agent-crashloop-debug` | 0/3 |
| `cluster-agent-crashloop-misleading-symptom` | 0/3 |
| `cluster-agent-crashloop-evidence-chain` | 3/3 |
| `cluster-agent-healthy-workload-no-finding` | **1/3** |
| `gpu-stress-test-diagnosis` | 3/3 |
| `agent-kanban-smoke` | 3/3 |

**27 of 39 repetitions passed. Eight cases were perfectly stable, two were flaky, three failed identically every time.** Four things follow, and they are the answer to "does the ladder earn its cost":

- **The single-run gate would have redded this PR on five of thirteen cases.** The ladder reported GREEN with all five as `UNSTABLE`, which is correct — nothing is admitted, so nothing can collapse.
- **Flake and failure are cleanly separable, and repetitions are what separates them.** The three 0/3 cases fail with byte-identical reasons on all three repetitions (`cluster-agent-crashloop-debug` misses `OOMKilled` every time). Those are signal — a real gap or a real check bug — and they will collapse the moment they are admitted, exactly as intended. One run could not have told them apart from the 2/3 case.
- **Both flakes are report-wording flakes, not capability flakes.** `upgrades-lagging-master-probe` rep 2 produced none of `['behind', 'lagging', 'not current', 'out of date', 'outdated', 'one minor', 'newer version']`; `cluster-agent-healthy-workload-no-finding` reps 1–2 produced none of `['2/2 running', '2/2 ready', '2 of 2 ready', 'no restarts', …]`. The agent did the work and phrased it outside the allowed set. That is the ladder absorbing noise correctly **and** a signal that those two `report_contains` lists are underspecified — worth fixing at the check, not the gate.
- **The judged/deterministic split held.** `VerificationCorrectness` was 1.0 or 0.5/0.0 with no intermediate wobble; the movement was all in whether a phrase appeared.

`agent-kanban-smoke` passing 3/3 is worth a note given it redded 8 of 11 recent PRs, but one run is one run — that is what the baseline store is for.

### Live validation

Partial, and the gap is specific.

**What was exercised against a real install.** The five fixtures are not synthetic: each is a full `uv run devops-bench` run against the live `kage-management` cluster, agent `gemini-3-1-pro-preview-kubeagents-mcp`, executed from cloudtop with a real Gemini judge. That is the eval pipeline end to end — port-forward, prompt, agent turn, verification spec against the live cluster, judge — and it produced 135k–275k tokens and 46–133s of wall clock per run. Feeding all five back through the **old** gate reproduces its verdicts exactly (`FAIL: VerificationCorrectness=0.5 (floor 1.0)` on the reds, `PASS` on the greens), and feeding them through the new one reproduces the intended new behaviour: reds unadmitted → `[UNSTABLE]`, suite exits 0; the same reds with the case added to `BOOTSTRAP_ADMITTED` → `COLLAPSE`, blocking, reason naming the failing check (`report-states-the-probe-title: required phrases absent from the report`) and not the judge; greens → `GREEN 2/2`. The installed `bench-gate` console script was driven directly for all of it.

**The GCS backend has now been exercised against a real bucket** (`gs://dshnayder-gke-dev-evals-bench`, a personal dev project standing in for the one `kube-agents-evals` will own), not just the test suite's fake `gcloud`. Confirmed live: an empty prefix reads as an empty store rather than an outage; objects file themselves under the version key and a `verifiers` bump starts a new directory beside the old one; ten recorder appends accumulate to `admitted on 20/20 screening runs across 10 recorded run(s)`, after which an all-fail case trips rung 4 and `suite` exits 1; `PULL_NUMBER` refuses the append; and a missing bucket 404s into an advisory verdict with the banner rather than a red.

The IAM claim was tested rather than asserted, with a throwaway service account holding exactly the two roles. **An overwrite is refused with `does not have storage.objects.delete access`**, and the target object is left byte-intact — GCS implements an overwrite as a delete plus a create, so `objectCreator` genuinely cannot rewrite history. `objectViewer` alone fails the create with `does not have storage.objects.create`, which is what makes the presubmit's read-only grant a real guard rather than a convention. That also surfaced a doc bug: `objectCreator` alone cannot *read back* what it wrote, so the recorder needs `objectViewer` too — the design now says so, with the provisioning commands.

**The presubmit budget was a prerequisite, it has landed, and it is now the thinnest number in this PR.** This PR takes the job from fourteen `devops-bench` invocations to forty-two. The matrix has moved underneath the estimate three times — `#956` took the active cases from three to ten, `#982` took them to thirteen, `#998` took them to fourteen — so the figure is measured from two builds rather than extrapolated, with one term that has never been measured at all:

| source | invocations | measured | build |
| --- | --- | --- | --- |
| fixed: Boskos lease, image build, deploy, teardown | — | 17.3min | `2092820036916875264` |
| the ten `#956` activated | 10 | 51.6min | `2092820036916875264` |
| the three `#982` activated | 3 | 9.2min (190s + 142s + 220s) | `2092719124550520832` |
| `rca-remediation-pr`, activated by `#998` | 1 | **never measured** | — |

**Superseded by a direct measurement.** The matrix has since run end to end at thirteen tasks × three repetitions, GREEN, on build `2093054834931404800` (2026-08-27): **156.8min** wall clock, of which 140.4min is the 39 invocations and 16.4min is fixed. An invocation averages **3.6min**, not the 4.7min the two-build extrapolation gave — every estimate before this one was pessimistic.

Only the invocations scale. At the measured 3.6min average, one repetition is ~67min and three is ~**168min**. Against that, the previous `150m` was 0.75× — a guaranteed timeout, not a tight budget, which is what made [oss-test-infra#2669](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2669) a prerequisite rather than a follow-up. It took the timeout to `240m` and has merged; the Prow `job-config` configmap has rolled.

**`240m` against ~168min is 1.43×**, where #2669 argued 1.40× at ten tasks and this job carried ~2× for a year. It is not already negative only because `#951` and seeded-cluster reuse cut `gpu-stress-test-diagnosis` from 21.1min to 244s. **Deliberately not raising it again in this PR:** eval-runtime performance work is in flight separately, and if it lands the headroom comes back without a third change to another repository. If it does not land before this merges, `300m` is the follow-up. The variance flagged as the term that would decide this has resolved in the good direction: `consistency-authorized-networks-probe` cost 1039s on the one earlier run that existed, but took 699s for all **three** repetitions on this matrix — 233s each. That was one bad sample. The expensive term is instead `compliance-rbac-overgrant` at 2042s for three repetitions, 24% of the task budget on its own.

**One of the fourteen terms is still a substitution rather than a measurement, and 1.26× is the honest figure.** `#998` activated `rca-remediation-pr` precisely so its own smoke run would be the first measurement of it, so the table prices it at the fleet average. It is one of the two active tasks that **write**, so `compliance-rbac-overgrant` is the better comparable at a measured 681s per repetition — at that cost the total is ~191min and **1.26×**. Read 1.43× as the optimistic bound and 1.26× as the working number until the first fourteen-task run lands.

The structural point a reviewer should take from this rather than the arithmetic: the budget lives in `oss-test-infra`, so activating a case here spends headroom only a separate pull request can replace, and nothing in this repository fails when it runs out. Four successive numbers have gone stale that way. The comment above `EVAL_REPETITIONS` in `hack/ci-eval-pr.sh` now says so, where someone about to uncomment a `TASKS` line will read it.

**What was NOT covered.** The reshaped `hack/ci-eval-pr.sh` has not run inside a Prow job, and no run at the **fourteen**-task × three-repetition matrix exists yet — the thirteen-task matrix has run GREEN end to end (build `2093054834931404800`, results below), so what is untested is the fourteenth case and the ~168min figure it moves. Three things therefore remain unproven and should be watched on this PR's own eval run: that `EVAL_REPETITIONS=1` is behaviour-preserving against the old gate before trusting 3; that `$ARTIFACTS` now holds the `run_*` directory on a **green** job, which is the check that the `ci-env.sh` split actually took; and the deliberate-break check — flip a `report_contains` phrase to something absent and confirm the case reports a correctness failure without redding while unadmitted, then add it to `BOOTSTRAP_ADMITTED` and confirm it reds with `collapsed: <case> failed 3/3`. I would rather say that plainly than describe the fixture runs as if they were a CI run.

No test artifacts to clean up: the runs wrote to a scratch directory on cloudtop, the `agent-kanban-smoke-green` prompt variant used to capture the two green records was never committed and is not a task in this diff, and nothing was left behind on `kage-management` — the task provisions no infrastructure.

### Review round 1 (four findings, three fixed, one declined)

- **The aggregate could red a green pull request on one flaky repetition.** **Fixed** — the sample floor described under What Changed. The suggested fix in the comment was a wider margin sized by the normal approximation; at `n = 3` that is 0.247, giving a 0.705 threshold, and 2/3 = 0.667 still reds — so the second suggestion, a scored-run floor, is what landed, with the arithmetic in the constant's docstring.
- **The EXIT trap never reached the artifact dumper on a red run.** **Fixed.** `errexit` stays in force inside an EXIT trap, so `(exit "${exit_code}")` — there only to set `$?` for the dumper — aborted the trap on that line. Every red eval job lost the `kubectl logs`, pod descriptions and events that tell a transport storm from a real failure, silently, and only on the runs that needed them. `set +e` after `$?` is captured; bash still exits with the original status. `scripts/test_ci_eval_trap.py` runs the real function lifted out of the real file, and its last test strips the guard back out and asserts the dumper is then unreachable, so the comment cannot drift away from being true.
- **`BOOTSTRAP_ADMITTED` could name a case that does not exist.** **Fixed** — `bench-gate suite` intersects it against the ids it actually graded and reports unmatched names on stderr and as a banner in the markdown verdict. A warning rather than a red, because the list may legitimately name a case deactivated in the `TASKS` array. It is free text whose whole job is to keep something blocking, so a typo disarms the case it was meant to protect and the run goes green exactly as it would have if the list were right.
- **Retry-on-failure — one repetition, two more only if the first fails. Declined**, with the reasoning recorded in the design doc's open items. It would land a green run near 67min instead of ~168min, which is real money and the cost argument is stronger now than when it was filed (the figure has since been measured directly rather than extrapolated: ~168min at fourteen tasks, 1.43×, or 1.26× pricing the one unmeasured case at canary cost). It is declined because it is not verdict-identical: rung 5 fires on `expected_fail` and **every** repetition passing, so one flaky pass flips the marker and the retry never triggers; rungs 1–3 are "any repetition" by design, so a catastrophic trip or an errored check appearing only on repetition 2 is never sampled; `EVAL_JUDGED_MARGIN = 0.5` was sized to the ~0.25 standard error of a mean of three, not the ~0.44 of a mean of one; and fewer scored repetitions on green runs is the direction the new sample floor already has to defend against. Sampling repetitions conditionally on the thing being measured is not a sampling strategy. The runtime problem is real and has an unrelated fix in flight; if that does not land, the honest levers are `EVAL_REPETITIONS=1` with rungs 4–6 reported as unarmed, or a smaller matrix.

### Review round 2 (one finding, fixed)

- **A typo in `EVAL_JUDGED_METRICS` silently disarmed rung 6** — the twin of the `BOOTSTRAP_ADMITTED` finding above, and quieter, because nothing at all changes in the output. The rung's loop did `if was is None or now is None: continue`, which covers two situations that deserve different treatment: a metric the baseline legitimately does not carry yet, and a name that matches nothing anywhere (`OutcomValidity` for `OutcomeValidity`). Both fell out of the loop without a word, so the judged comparison gated nothing while the configuration read as though it gated a metric — and a green measured against nothing looks exactly like a green. **Fixed** — `grade_case` intersects the configured names against the **union** of what the run scored and what the baseline carries; a name in neither becomes a note, surfaced on stderr by `case` and as one deduplicated banner by `suite`, never blocking, for the same reason as the admission typo. The union matters in both directions: a metric the records have and the baseline lacks is configuration moving ahead of evidence, and a metric the baseline has and the records lack is a judge that *stopped emitting* it — the name is right, the evidence went missing, and calling that a typo sends the reader after the wrong thing. Guarded on there being any judged evidence at all, so a run whose judges all failed does not report every configured name as misspelled. Nine tests, and not vacuous: stubbing the check out fails three of them, narrowing the union to the baseline alone fails two more.

### Self-review passes

Two passes, both run in the context that wrote the change. **The adversarial pass AGENTS.md asks for — a subagent or fresh session handed only the diff — has not run.** This harness will not start one without approval, so I am saying so rather than letting the section imply a pass it did not get. It was one of two reasons this sat as a draft; the other, the presubmit timeout, has now been resolved, so this is out of draft with that gap still open and named rather than quietly closed.

What the in-context passes looked for and found:

- **A verdict that lies about itself.** Found: the CLI printed `[PASSED] passed 0 of 3 repetitions` for a case that failed everything without collapsing. **Fixed** — four labels instead of two, with `UNSTABLE` for exactly that state, threaded into both the log line and the markdown table, plus four tests.
- **Tests that cannot fail.** Found: `test_an_empty_record_does_not_supply_the_key` copied a good manifest intact, so the empty record would have produced the same key and the assertion proved nothing. **Fixed** — the manifest is branded `died-before-the-agent-ran` so a regression is visible. The parametrized sweep over `bench/tasks/*/task.yaml` got a guard test asserting the glob finds ≥12 files, so it cannot pass vacuously either. The four-way mutation test above is the systematic version of this question.
- **Whether the tests agree with the scorer about fields devops-bench does not emit.** This is why every fixture is a mutation of a real capture; the `metadata.session_id` mistake described under Testing is the failure it was built to prevent.
- **Silent YAML execution.** `cases.py` uses `yaml.safe_load`, with a test that a `!!python/object/apply:os.system` payload raises rather than runs.
- **Dead code carried forward by a refactor.** Found: the `ChecklistScore` block (reads a key nothing emits), the judged `OutcomeValidity` fallback, and a `[ -f … ] || cp … harness.log` line made dead by the new unconditional artifact copy. All **deleted** rather than ported.
- **Docs drift.** Found and fixed: `bench/README.md` had no gate section, no `baselines/` in its layout, and a `tests/` line that no longer described the tests; `bench/CUSTOM-TASKS.md` documented neither `expected_fail:` nor the id-matches-directory rule; `docs/README.md`'s map was missing both new READMEs. Checked and found **not** stale: `bench/tasks/DRAFTS.md` blocker A5 — but it was incomplete, so it now says the ladder deliberately does not soften it.
- **Deviation from the approved plan, deliberate, stated so it is not mistaken for a slip.** The plan said `dump_prow_artifacts_on_failure()` should keep only steps 1–3. I left the `collect_bench_results` call inside it as well, so the three other callers (`ci-deploy.sh` ×3 sites, `ci-connection-test.sh`) lose nothing. On the eval script's failure path this copies the same files twice, which is a harmless `cp`.
- **A threshold I set by feel, caught by a test I wrote to check it.** `EVAL_JUDGED_MARGIN` started at 0.25, described in the comment as "the measured noise floor". `test_the_default_margin_absorbs_the_measured_judge_spread` failed: the mean of the three captured judge scores is 0.70, a 0.30 drop from a perfect baseline, so an unchanged pull request would have redded. 0.25 is _one_ standard error, not a floor. **Fixed** at 0.5 with the arithmetic written down and five dependent tests retuned; one of those now passes an explicit `judged_margin=0.1` so it isolates _which_ metric is gated from _how far_ it may move.
- **A verdict line that stated a cause it had not checked.** Found: an unadmitted case printed `(not admitted: no screening evidence, so it cannot collapse)` for a case holding 21 runs of evidence below the bar. Three of the four possible causes made that sentence false. **Fixed** — the reason no longer guesses, and the actual state is printed on the `admission:` line beside it.
- **Another test that could not fail.** Found: `test_an_unadmitted_case_contributes_nothing_to_either_side` ended in `assert … or True`. **Fixed** — it now asserts through `--json-out` that both `pass_rate` and `baseline_rate` are `null`, which is the real property: an unscreened case must be excluded from _both_ sides, or the aggregate compares two different populations.
- **Checked the implementation against the design rather than against my memory of it.** `docs/designs/testing-strategy.md` was not on `main` when this was written — it was PR #896, which has since merged — so §4.2's ladder was re-read from that PR's diff after the code was written. It matches row for row, including the `n/a` infra row and, specifically, that rung 5 is admission-blind alongside rungs 1–3 while admission scopes only rungs 4 and 6. An earlier draft of this description called that scoping an open question needing a decision; it is not — §4.2 settles it and argues for it, and the code follows it.

## Risk & Rollout

Low blast radius: nothing here reaches a cluster, an operator, or a running agent. The changed surfaces are the eval presubmit and a new Python package under `bench/` that only CI invokes. The eval job is `optional: true`, so even a total failure of this code cannot block a merge today.

The one real behaviour change is that **the two active tasks stop blocking on correctness** the day this lands, because the baseline store ships empty and nothing is admitted until it has been screened against `main`. That is deliberate and is what the strategy's "advisory until baselines exist" means, but it should be a conscious merge-time decision rather than a surprise. `BOOTSTRAP_ADMITTED` keeps `gpu-stress-test-diagnosis` blocking through the transition; add more names to it if that trade is too loose.

Two new failure modes worth naming. `bench-gate case` exits 2 on an unreadable `task.yaml` or a broken `VERSIONS.json`, which stops the job — intended, since a gate that cannot parse its own configuration must not report green, but it is a way to red the eval job that did not exist before. And the admission-blind rungs mean one missing `results.json` on a `noop` task reds the job regardless of admission, now three times as likely with three repetitions.

The recorder is the one thing that writes, and it is deliberately never fatal: the shell wraps it so a failed append warns and the verdict is unaffected, and it runs **before** the suite step, because the suite exits 1 on a red and under `set -e` everything after it would be skipped — a red run on `main` is precisely the run that must not go unrecorded.

The GCS backend adds no risk today because it is off: with `EVAL_BASELINE_STORE` empty the code path is never entered, and the tests that cover it never touch the network. The risk arrives when it is switched on, and the shape of it is named above — an outage degrades the gate to advisory rather than redding it, which is the right trade but is a real loosening, and it is why the banner is in the verdict rather than only the log.

**The writer is a separate pull request in another repository.** Nothing in this repo can close the loop: Prow config for `gke-labs/kube-agents` lives in `GoogleCloudPlatform/oss-test-infra`. The job is written and open there as a draft — [oss-test-infra#2665](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2665) — held until this merges and the bucket exists.

It is a **nightly periodic** rather than a postsubmit, which is a change from what this design said for its first six revisions. The arithmetic overturned it: `main` takes ~10 merges a day (310 in the 30 days to 2026-08-26) and the eval job averages ~43 minutes, of which the eval itself is only about a fifth — leasing a Boskos project, building images and standing up a cluster dominate. A postsubmit pays that setup again on every merge to buy three samples of each case, roughly 7 project-hours a day out of the pool's 72, serialised at `max_concurrency: 1`. One nightly run amortises that setup over every repetition instead: cheaper per sample, and it refills the whole 20-run admission window in a night or two after a version-key bump rather than over a week of merges. Filtering merges by changed path stays ruled out — it would bias the evidence toward whichever changes happen to be cheap to run.

That is why `hack/ci-eval-pr.sh`'s write guard now matches `JOB_TYPE` against `periodic` **or** `postsubmit`. Keying on `postsubmit` alone would have made the periodic skip recording silently, and an empty store reads as a legitimate green to every presubmit, because an unadmitted case cannot fire the quality rungs. The empty `PULL_NUMBER` is what actually keeps pull requests out, and `bench-gate record` still refuses independently on it.

The nightly runs at `EVAL_REPETITIONS=10` against the presubmit's 3, deliberately. An earlier draft of the design claimed the counts had to match; that was wrong. The store holds a pass **rate**, so ten runs and three runs estimate the same quantity — what must match is how a *single* run is produced, and the job's script body is the presubmit's verbatim (0 removals) apart from three exports.

Reverting is a single `git revert`: the store is empty, so no baseline data is lost, and the shell script's old inline verdict comes back with it. Nothing to do at merge time beyond watching this PR's own eval run for the three unproven items under Live validation.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code).


bench/kube_agents_bench/scoring.py
```diff
@@ -0,0 +1,859 @@
...
+                f"({case.get('rung_name')}) -- {case.get('reason')}"
+            )
+
+    admitted = [c for c in cases if c.get("admitted")]
+    passes = sum(int(c.get("passes") or 0) for c in admitted)
+    scored = sum(int(c.get("scored") or 0) for c in admitted)
+    pass_rate = (passes / scored) if scored else None
+
+    if pass_rate is not None and baseline_rate is not None:
+        if pass_rate < baseline_rate - margin:
```
jayantid: 🔴 **Not merge-blocking today — this path is dormant until the store fills — but it must land before the nightly recorder turns on.**

`grade_suite` reds when `pass_rate < baseline_rate - margin` (scoring.py:835) with no sample-size floor. The PR side is 3 reps per admitted case; the margin defaults to 0.05.

With one admitted case at a 20/21 baseline, one failed rep gives 2/3 = 0.667 < 0.902 and `suite` exits 1 — reproduced through the installed CLI. At the admission bar's own 95% per-rep reliability, an unchanged PR reds 14% of the time with one admitted case, 26% with two, 37% with three (8/9 = 0.889 still reds). That is `agent-kanban-smoke`'s failure mode reintroduced by the aggregate the day the first case is screened in, and it contradicts the PR body's "a case that merely fails can only red the job by failing every repetition". No flat margin fixes both n=3 and n=600.

Scale the slack with the sample:

```python
if pass_rate is not None and baseline_rate is not None:
    slack = max(margin, 2 * (baseline_rate * (1 - baseline_rate) / max(scored, 1)) ** 0.5)
    if pass_rate < baseline_rate - slack:
```

or refuse the comparison below a scored-run floor and let the nightly own rate evidence.

hack/ci-eval-pr.sh
```diff
@@ -667,233 +767,124 @@ for TASK in "${TASKS[@]}"; do
...
-    mkdir -p "${ARTIFACT_DIR}"
-    cp "${EVAL_LOG}" "${ARTIFACT_DIR}/scoring_failure_${TASK_NAME}.log" 2>/dev/null || true
-    if [ -n "${LATEST_RESULT}" ] && [ -f "${LATEST_RESULT}" ]; then
-      echo "Task ${TASK_NAME} Result: [FAILED] results.json carries no scored record -- the run or its scoring pass crashed; see ${ARTIFACT_DIR}/scoring_failure_${TASK_NAME}.log (Duration: ${TASK_DURATION}s)"
+  # One --result per repetition, positionally. A repetition that produced no
+  # run directory contributes the literal MISSING, so the gate can tell "died
+  # before writing anything" from "wrote an unusable record" -- a different
+  # diagnosis with a different owner.
+  RESULT_ARGS=()
+  for REP in $(seq 1 "${EVAL_REPETITIONS}"); do
```
jayantid, line 890: 🟡 **A proposal for you to rule on before the suite grows again.**

Every task runs 3x regardless of what rep 1 said. Headroom is 240m over a measured 172m (1.40x), and observed variance eats it: the compliance canary has cost 606/827/1497s across its three runs, so three worst-case canary reps add ~45min to the measured run, and the 1039s consistency probe is already inside the 172. A Prow timeout is a red PR that none of the gate's rules ever saw.

Retry-only-on-failure is verdict-identical for collapse. Run rep 1; on pass, grade the case on that single `--result` — verified: one result reports GREEN "passed all 1 repetitions", and a passing rep can never collapse. On fail, run reps 2–3 and pass all three. Green PRs pay ~69min. One trap: do not pad with the literal MISSING — a MISSING rep on a noop task grades rung 2, blocking. The honest caveat: retried cases inflate PR-side pass rates, so these reps must not feed the aggregate or the store. The store already excludes presubmits by construction; the aggregate should take its rate evidence from the nightly anyway (see the scoring.py:835 comment).

hack/ci-eval-pr.sh
```diff
@@ -181,8 +190,14 @@ source "${SCRIPT_DIR}/ci-env.sh"
...
+#
+# collect_bench_results runs on green too, and that is the whole point: the
+# baseline store the gate compares against is built from PASSING runs on main,
+# and those are exactly the records the old failure-only trap threw away. It
+# cannot precede the `$?` capture, so it sits immediately after it.
 profile_and_dump_on_exit() {
   local exit_code=$?
+  collect_bench_results
   profile_report "${exit_code}"
   (exit "${exit_code}")
```
jayantid, line 216: 🟡 **Not merge-blocking — inherited from main — but you are editing this function and the fix is one line.**

`set -e` stays active inside the EXIT trap, so on any failing exit `(exit "${exit_code}")` at line 202 returns non-zero and errexit aborts the trap right there. `dump_prow_artifacts_on_failure` on the next line never runs.

Every red eval job therefore loses its kubectl logs, pod descriptions and events — the diagnostics that tell a #988-style 502 storm from a real failure — while the comment above the trap says the exit code is handed "to the artifact dumper". Reproduced: in `bash -c 'set -e; f(){ local c=$?; (exit $c); echo reached; }; trap f EXIT; exit 7'`, `reached` never prints. Your green-path story is unaffected: `collect_bench_results` runs before the abort and every line in it is `|| true`-guarded.

`set +e` at the top of the trap preserves `$?` for the dumper, and bash still exits with the original status:

```bash
profile_and_dump_on_exit() {
  local exit_code=$?
  set +e
  collect_bench_results
  ...
```

bench/kube_agents_bench/gate.py
```diff
@@ -0,0 +1,659 @@
...
+        return BaselineStore({}), None, f"{location} unreachable: {exc}"
+
+
+def _bootstrap_admitted() -> frozenset[str]:
+    """Cases that keep blocking through the transition, from the environment.
+
+    Whitespace- or comma-separated, so the shell can write either.
+    """
+    raw = os.environ.get("BOOTSTRAP_ADMITTED", "")
+    return frozenset(part for part in raw.replace(",", " ").split() if part)
```
jayantid, line 150: 🟡 **Not merge-blocking.** A misspelled name in `BOOTSTRAP_ADMITTED` silently un-arms the case it was meant to keep blocking.

`_bootstrap_admitted` (gate.py:139) is bare set-membership. Nothing compares the names against the cases actually graded.

The variable is designed to be hand-edited during the transition — the PR body says "add more names to it if that trade is too loose". Write `crashloop-debug` for `cluster-agent-crashloop-debug` in the Prow env and that case quietly stops blocking. The only trace is one `admission:` log line on runs that are green, which nobody reads. That is the silent-disarm class the corrupt-store exit-2 exists to close, arriving through configuration instead of data.

`suite` sees every graded case id; warn — or red, matching the corrupt-store philosophy — when an entry matches none of them:

```python
unknown = _bootstrap_admitted() - {str(c.get("case")) for c in cases}
if unknown:
    print(f"WARNING: BOOTSTRAP_ADMITTED names no graded case: {sorted(unknown)}", file=sys.stderr)
```

bench/kube_agents_bench/scoring.py
dshnayder: Fixed in 1e81e03, and thank you — this is the real one. You are right that with a single admitted case the aggregate reintroduces exactly the failure mode the collapse rung exists to stop, on the day the first case is screened in.

The arithmetic, written down because it decided the shape of the fix. At the 0.05 default margin against a baseline screened at the 19/20 admission bar, the blocking threshold is 0.90, so a run of `n` scored repetitions survives `floor(n * 0.098)` failures — **zero** below `n = 11`. With one admitted case at three repetitions `n` **is** 3, and 2/3 = 0.667 against 0.902.

I did not take the suggested fix as written, and the reason is arithmetic rather than preference: two standard errors at `n = 3` is 0.247, so the threshold becomes 0.705 and **2/3 = 0.667 still reds**. The normal approximation is not usable at this sample size. So I took your second suggestion instead — a scored-run floor:

- `DEFAULT_AGGREGATE_MIN_SCORED = 30` (ten admitted cases at three repetitions; tolerates two failed repetitions), overridable by `--min-scored` / `EVAL_AGGREGATE_MIN_SCORED`.
- Below the floor the rate is still computed, still printed, and **still says when it fell below the margin** — it just cannot block. `SuiteVerdict` gains `notes` alongside `reasons` for that: reporting it as a reason would red the job, and dropping it silently is how a rule that never fires goes unnoticed for a year.
- A floor rather than a wider margin because no single flat margin is right at both `n = 3` and `n = 600`. The properly-sized replacement is a two-proportion test, which needs a variance estimate that does not exist until the nightly has run against `main` enough times to produce one. That is recorded as an open item.

Tests: the floor and the margin rule are now separable (`test_the_margin_rule_is_separable_from_the_sample_floor`), one flaky repetition of one admitted case cannot red the suite, an advisory aggregate still says it fell below, and the floor stops applying at exactly 30 (29 advisory, 30 blocking). Four existing gate tests now pass `--min-scored 1` so they keep isolating the margin rule from the floor — the same thing they already do for `judged_margin`.

Documented in `docs/designs/eval-scorer.md` under the suite aggregate, with an open item for the two-proportion test.

hack/ci-eval-pr.sh
dshnayder: Thanks for this — the cost argument is a good one, and I want to give you the reasoning rather than just a preference. After working through it I'm not going to take it, but I'd like to lay out why so you can push back if you disagree.

First, one number worth updating, and it actually helps your case: #982 took the matrix to thirteen active cases after this comment was written, so the figure quoted here has moved. Re-measured from two builds — fixed cost 17.3min, #956's ten invocations 51.6min (build `2092820036916875264`), #982's three 9.2min (build `2092719124550520832`) — three repetitions is now **~200min against 240m, or 1.20×**, not 172min/1.40×. There is less headroom than the comment assumed, and your proposal would land a green run near 78min. That is real money and I don't want to wave it away.

The reason I'm still declining is that it isn't verdict-identical to three unconditional repetitions, in four ways:

1. **Rung 5 inverts.** It fires on `spec.expected_fail and passes == len(scored)` — *every* repetition passing. Sample once and a single flaky pass flips an expected-fail marker that three repetitions would have held, and the retry never triggers, because from the sampler's point of view nothing failed.
2. **Rungs 1–3 are "any repetition", by design.** A catastrophic trip, an errored check or a dead trajectory that appears on repetitions 2 and 3 but not on 1 would never be sampled. These are the rungs that are absolute and admission-blind precisely because they shouldn't be missed.
3. **Rung 6's margin is sized to a three-repetition mean.** `DEFAULT_JUDGED_MARGIN = 0.5` came from the ~0.25 standard error of a mean of three (captured spread 0.9 / 1.0 / 0.2, sd ~0.44). A mean of one has SE ~0.44, so the rung would be mis-sized against it.
4. **It interacts with the sample floor added for your other comment.** Fewer scored repetitions on green runs is the direction that floor has to defend against.

The general form: repetitions are how this gate tells a flaky agent from a regression, so sampling them conditionally on the thing being measured changes what the ladder measures. Your own caveat about pass-rate inflation is the same observation from the other end, which is part of why I think you'd land in the same place.

The runtime problem is real and is being worked separately — that's also why the timeout isn't being raised a third time in this PR. If that work doesn't land, I think the honest levers are `EVAL_REPETITIONS=1` with rungs 4–6 explicitly reported as unarmed, or a smaller matrix, rather than a three-repetition ladder fed one repetition.

I've recorded this in `docs/designs/eval-scorer.md` under Open items so the reasoning is findable rather than re-litigated. Leaving the thread open for you: if you'd still like it with the divergences accepted and documented, say so and I'll implement it.

hack/ci-eval-pr.sh
dshnayder: Fixed in 1e81e03 — confirmed by reproduction, and it was worse than "may not run": the dumper **never** ran on a red job.

```
bash -c 'set -e; f(){ local c=$?; (exit $c); echo reached; }; trap f EXIT; exit 7'   # never prints "reached"
```

`errexit` stays in force inside an EXIT trap, so `(exit "${exit_code}")` — which is there only to set `$?` for the dumper — aborts the trap on that very line. Every red eval job has been losing the `kubectl logs`, pod descriptions and events that distinguish a transport storm from a real failure, silently, and only on the runs that need them, while the comment above it claimed the exit code was being handed to the dumper.

`set +e` after `$?` is captured. The subshell keeps doing its job of setting `$?` for the dumper, and bash still exits with the original status.

`scripts/test_ci_eval_trap.py` covers it, and deliberately not by grepping the script for `set +e`: it lifts the real `profile_and_dump_on_exit()` body out of `hack/ci-eval-pr.sh` and runs it under `set -euo pipefail` with the three callees stubbed, so it also fails if a later edit puts an errexit-fatal command ahead of the dumper. The last test is a negative control — it strips the guard back out and asserts the dumper is then unreachable — so the comment in the script cannot drift away from being true. It also asserts the original exit code survives the trap, since Prow reads it.

bench/kube_agents_bench/gate.py
dshnayder: Fixed in 1e81e03. `bench-gate suite` now intersects `BOOTSTRAP_ADMITTED` against the case ids it actually graded and reports every unmatched name, on stderr and as a banner in the markdown verdict.

The failure mode is worth stating because it is the quiet kind: the list is free text whose entire job is to keep something blocking, so a typo — or a rename that did not reach it — disarms the case it was meant to protect, and the run goes green in exactly the way it would have if the list were right.

A warning rather than a red, deliberately. The list is legitimately allowed to name a case that is commented out of the `TASKS` array or skipped on a given run, and redding for that would make the bridge harder to hold than the thing it bridges. Covered by `test_a_bootstrap_admitted_name_matching_no_graded_case_is_reported` and `test_a_fully_matching_bootstrap_admitted_says_nothing`, and written down in the design doc's Bootstrap section.

dshnayder:
/retest

The previous run was not a test failure. `clonerefs` exited 0 and the `test` container was still `running` when the pod was killed:

```
DisruptionTarget  reason: DeletionByPodGC  message: PodGC: node no longer exists
```

There is no `build-log.txt` in the artifacts because the sidecar never got to upload one. Two Prow build nodes went away inside a three-minute window and took three jobs with them — this one and #1012 shared node `10.138.0.54` and were killed at the same second (14:14:04Z), and #997 went with node `10.138.0.57` at 14:11:04Z. Nothing in this pull request is implicated.

bench/kube_agents_bench/scoring.py
```diff
@@ -0,0 +1,918 @@
...
+    # and still land here, which is the only place in the ladder where "it
+    # technically passed but got worse" is sayable.
+    if admitted and not spec.expected_fail and complete and baseline_judged:
+        means = judged_means(reps)
+        drops = []
+        for metric in judged_metrics:
+            was = baseline_judged.get(metric)
+            now = means.get(metric, {}).get("mean")
+            if was is None or now is None:
+                continue
```
jayantid, line 816: 🟡 **Not merge-blocking; it is the twin of the fix you just made.** A typo in `EVAL_JUDGED_METRICS` silently disarms rung 6 the same way a typo in `BOOTSTRAP_ADMITTED` disarmed admission — and that one you fixed in 1e81e03.

This `continue` covers two cases that deserve different treatment: a metric legitimately absent from the baseline, and a name that matches nothing anywhere — `OutcomValidity` for `OutcomeValidity`. Both fall out of the loop without a word, so the judged-regression rung gates nothing while the config reads as if it gates one metric.

Your own fix is the template: intersect the configured names against the union of `means` and `baseline_judged` keys, and report unmatched names on stderr and in the markdown banner — warning, not red, for the same reason as there (a metric can be configured ahead of the baseline carrying it).

jayantid (approved):

bench/kube_agents_bench/scoring.py
dshnayder: Fixed in e3263788, as the template you named.

`grade_case` now intersects the configured names against the **union** of what the run scored and what the baseline carries, and a name in neither becomes a note — stderr from `case`, one deduplicated markdown banner from `suite`, never blocking. `CaseVerdict` gained `notes` (defaulted empty, carried through `to_dict`), so a case file written before this still reads.

Two details worth naming, both of which came out of writing the tests:

**The union matters in both directions.** A metric the records have and the baseline lacks is a config change awaiting a screening run. A metric the baseline has and the records lack is a judge that *stopped emitting* it — the name is right, the evidence is what went missing, and reporting that as a typo sends the reader after the wrong thing. Narrowing to the baseline alone fails two of the new tests.

**It is guarded on there being any judged evidence at all.** Otherwise a run whose judges all failed reports every configured name as misspelled, on exactly the runs that are already noisy. No evidence either way is not evidence of a typo.

Nine tests, and not vacuous — stubbing the check out fails three, narrowing the union fails two more.

Issue #804, opened 2026-08-19
Complete the obtainability CUJ1 design contract


haoxuw:
## Feature request

Make the obtainability-aware Day-0 cluster-design journey observable and complete end to end, so `bench/cuj/obtainability/test_01_cluster_design.py` passes all eleven acceptance criteria against a live kube-agents installation.

The test was introduced in #692. It acts as a user through Kage, delegates to the Platform Agent, and evaluates only evidence returned by the admin portal API. The pytest result gates on eleven backend-independent acceptance criteria (AC01–AC11); Kage/Platform routing and skill milestones are diagnostic only. Every criterion and milestone reports PASS or FAIL — an unverifiable check reports FAILED with its reasons under `missingProof`.

## Current behavior

A fresh live run (2026-08-24) completed the conversation and the Platform task inside the portal deadline, and finished with **1 of 11 acceptance criteria passed** and **4 passed, 6 failed** milestones:

| Milestone | Result | Current gap |
| --- | --- | --- |
| Interaction completes | Pass | None |
| Kage delegates | Pass | None |
| Platform task is created | Pass | None |
| Required design skills load | Fail | Missing proof: the task projection omits `skills` and `loadedSkills`; the milestone now also requires the new `gke-obtainability` skill. |
| Platform specialist completes | Pass | None |
| Quota is checked | Fail | Missing proof: the task projection exposes no typed `quota_check` evidence. |
| Live obtainability is checked | Fail | Missing proof: the task projection exposes no typed `advice_service_capacity` evidence. |
| Capacity claims are qualified | Fail | Content: the terminal `output` is only Kage's delegation acknowledgment ("the platform agent will deliver the analysis right here in the thread"); the specialist's report — whose task summary shows the quota/availability assessment happened — never reaches the interaction's terminal output, so the answer cannot distinguish quota from live capacity, compare Autopilot and Standard, or present the ComputeClass guidance. |
| ComputeClass validates | Fail | Missing proof: the task projection exposes no `computeclass_server_dry_run` evidence. |
| Design remains read-only | Fail | Missing proof: the projection omits worker `toolCalls` and `toolEvidenceComplete`, and projected tool calls carry no normalized `operation`. Certification also now fails on any unjudgeable call, and an observed forbidden operation outranks missing-evidence reasons. |

[...]

PR #692, opened 2026-08-13
test(bench): add obtainability CUJ acceptance tests

haoxuw:
## Summary

Adds manually run, black-box pytest coverage for obtainability CUJ1 (GKE cluster design) and CUJ3 (workload obtainability planning). The tests talk to a live agent through an isolated admin portal, save complete JSONL interaction logs under `/tmp`, and score backend-independent acceptance criteria while reporting routing milestones separately.

## Why This Change

These CUJs need reproducible tests that distinguish a completed user outcome from delegation or skill loading alone. The tests expose missing capacity evidence, recommendations, and generated Kubernetes artifacts without changing production behavior.

## What Changed

- Added reusable CUJ utilities for portal lifecycle, configuration, interaction projection, evidence logging, acceptance criteria, and diagnostic milestones.
- Added CUJ1 acceptance coverage for live obtainability analysis, ComputeClass fallbacks, and the alternative NAP design.
- Added CUJ3 acceptance coverage for CalendarMode windows, ranked scheduling guidance, ProvisioningRequest, and Kueue LocalQueue output.
- Added a live-agent prerequisite targeting the stock `platform-agent`, with
  `CUJ_PROJECT_ID` and optional shared overrides as suite configuration.
- Kept `bench/cuj` manual-only and removed offline tests of the CUJ evaluators.

Run every CUJ from the repository root:

```bash
uv run --project bench pytest -s bench/cuj
```

## Context

#804 tracks the production behavior and evidence gaps exposed by CUJ1.

## Testing

- **PASS — remaining offline suite:** `uv run --project bench pytest -q bench/tests` — 113 passed.
- **PASS — CUJ discovery:** `uv run --project bench pytest --collect-only -q bench/cuj` — prerequisite, CUJ1, and CUJ3 collected.
- **PASS — repository checks:** `prettier --check bench/cuj/README.md` and `make docs-check`.
- **PASS — live prerequisite:** `platform-agent/default` returned terminal output `READY` through the isolated portal.
- **EXPECTED FAIL — CUJ1:** `1/11` acceptance criteria passed. Kage delegated to a Platform task, but the portal interaction timed out before receiving its terminal result. The portal projection also omitted worker skills, evidence, artifacts, tool calls, and final output, so the remaining criteria were reported as blocked with their reasons. The worker later completed out of band, after the portal deadline.
[...]

PR #926, opened 2026-08-24
feat(agents): add gke-obtainability skill and align CUJ1 contract with the AdviceService API

haoxuw:
## In plain words

Ask the agent for a GKE cluster that needs 32 A100 GPUs and today it can hand you a design that looks fine but will sit at zero nodes, because nobody checked whether those GPUs exist to be had. This PR makes the agent check before it answers.

- **It asks two separate questions.** Is this project *allowed* that many GPUs (quota)? Are they actually *free* in a zone right now (obtainability)? It calls the real Google Cloud APIs for both and never blurs the two.
- **It gives an honest answer.** On-Demand, Spot, and Flex-Start each get a verdict with their trade-offs. It never says "guaranteed". A failed check is reported, not hidden.
- **It hands you something to use.** A ComputeClass with fallbacks, plus a Node Auto-Provisioning alternative, both in the real schema.
- **The test that grades this is fixed too.** It now asks only for what the APIs can truly provide, and each run leaves a short readable transcript instead of a wall of polling noise.

## Summary

Gives the Day-0 GKE design journey its obtainability behavior and aligns CUJ1's contract with what the platform can actually provide: a `capacity-obtainability` skill that gathers live quota and capacity-advice evidence and records it in typed form, the credential-proxy allowlist entries that behavior needs, and the CUJ1 evaluator corrections that make the criteria satisfiable and honestly scored.

## Context

#804 tracks making CUJ1 observable end to end. This PR is the behavior-and-contract half; the portal projection and worker recorder (the runtime half) live on `feat/804-projection` and will follow as their own PR. #692 (merged) introduced the CUJ suite this PR's evaluator changes refine.

## What Changed

- **`capacity-obtainability` skill**: quota, reservation, and per-provisioning-model capacity-advice diagnostics; typed evidence recording with a copyable request/analysis shape; ComputeClass and NAP manifest templates using the real schemas (`cloud.google.com/v1`, `machineFamily` takes a family, NAP constrains via nodeAffinity); a fill-in `## Provisioning paths` report section covering On-Demand, Spot, and Flex-Start; and a banned-guarantee rule. The pinned manifest templates are the schema check: a Day-0 design has no target cluster to validate against and the agent's Kubernetes grant is read-only, so the server-side dry run an earlier revision mandated has been withdrawn. It carries only the design-time diagnostics — alert triage, GitOps leases, and Pull Request submission stay in the `gke-stockout-investigator` plugin, named in the description so a stockout alert still routes there.
- **`gke-cluster-creation` preflight**: GPU/TPU and large-shape designs run the obtainability check before recommending capacity. It ships as a section in the sync script's registered footer for that skill, not as an in-place edit, so `sync-upstream-skills.py` re-applies it; the skill itself is named without the `gke-` prefix for the same reason, because the sync prunes any local `gke-*` directory upstream lacks.
- **Command policy**: one flag-arity entry, `--max-run-duration`, so the skill's Flex-Start capacity probe reaches the `advice capacity` allowlist entry that #885 added on `main` (an entry whose flags are unknown is unreachable). The `kubectl apply --dry-run` carve-out an earlier revision added has been withdrawn, and the test now pins that `apply` stays refused in every spelling, including the repeated-flag one that bypassed the carve-out.
- **CUJ1 evaluator**: answer-scored criteria read the answer the user actually receives, which is the coordinator's reply after its hand-off plus the delegated task results the gateway posts into the same thread; AC05 reworded to what the AdviceService offers (Spot and Flex signals — the API rejects ON_DEMAND; verified live and against the discovery document) plus a user-visible answer weighing all three paths, with On-Demand scored on the answer and the quota check rather than demanded from AdviceService evidence that never carries it; the nonexistent `GeneralCapacityRecommendation` method dropped from AC02; answer-scored criteria skip leading delegation-acknowledgment paragraphs; artifact selection is latest-wins so a corrected attachment supersedes a malformed first attempt; m3 requires the new skill; word-bounded path matching.
- **Run evidence**: the CUJ runner logs interaction transitions instead of every poll (a fifteen-minute run wrote 275 KB of near-identical payloads; it now writes ~12 records), and each run emits `conversation.txt` shaped like the portal's Chat tab — a boxed turn per message, then the delegated work as `assignee · status` cards with their typed evidence and artifacts. A redacted example is posted as a comment on this PR.
- **Docs**: skill catalogue and family roster regenerated.

## The eleven acceptance criteria

These are the criteria as they read on this branch — the contract a live run is scored against, reproduced here so the PR stands on its own. Each is scored from the portal API response alone: typed evidence and artifacts projected on the delegated task, plus the answer the user actually received. Milestones (Kage delegation, skill loading, read-only certification) are diagnostic and do not gate the test.

| # | Acceptance criterion | Proof required |
| --- | --- | --- |
| AC01 | The request specifies a GKE design for 32 A100 GPUs in us-central1. | projected user input contains 32, A100, and us-central1 |
| AC02 | kube-agents invokes AdviceService.Capacity for the requested capacity. | completed evidence names AdviceService.Capacity and requests 32 A100 GPUs in us-central1 |
| AC03 | kube-agents analyzes the available quantity for the requested capacity. | capacity evidence contains a numeric availableQuantity |
| AC04 | kube-agents analyzes capacity distribution across zones. | capacity evidence contains signals for at least two zones |
| AC05 | kube-agents analyzes the obtainability signals AdviceService offers (Spot and Flex) and advises the user on the trade-offs of all three provisioning paths; On-Demand carries no advance obtainability signal, so its assessment rests on quota and reservations. | capacity evidence contains non-empty SPOT and FLEX signals and the delivered answer weighs On-Demand, Spot, and Flex |
| AC06 | The recommended path generates a GKE ComputeClass manifest. | projected artifacts contain a structured manifest with kind ComputeClass |
| AC07 | The ComputeClass uses the A2 series as its primary priority. | the first ComputeClass priority selects an A2 family or machine type |
| AC08 | The ComputeClass provides both L4 and T4 fallback options. | later ComputeClass priorities contain both L4 and T4 accelerators |
| AC09 | The alternative path outputs a Node Auto-Provisioning specification. | projected artifacts contain a structured node_auto_provisioning artifact |
| AC10 | The NAP alternative allows n2, n2d, and c2d machine families. | NAP selectors include cloud.google.com/machine-family values n2, n2d, c2d |
| AC11 | The NAP alternative uses location policy ANY. | NAP configuration contains locationPolicy: ANY |

AC05 and AC02 changed in this PR. `AdviceService.Capacity` accepts only `SPOT` and `FLEX_START` — verified live and against the compute API discovery document — so requiring a non-empty `ON_DEMAND` signal from that evidence was unsatisfiable by any honest implementation; AC05 now takes Spot and Flex signals from the evidence and asks the answer to weigh all three paths, with On-Demand assessed through quota and reservations. AC02 dropped `GeneralCapacityRecommendation`, a method that does not exist.

## Testing

- `make test-python` — green in every directory whose dependencies are installed locally; 81 command-policy tests, with new coverage for the skill's exact `advice capacity` spellings and for `apply` staying refused as a dry run.
- `make docs-check`, `make prompt-check` — green.
- `uv run --project bench pytest -q bench/tests` — green (661 with the harness tests collected from the bench directory), including the offline cases that pin the delegation-acknowledgment helper against the hand-off template read from `agents/chat/SOUL.md`.
- `npx prettier --check` on the skill after the fences were lifted out of list items; `scripts/test_sync_upstream_skills.py` green; the checked-in `gke-cluster-creation/SKILL.md` verified equal to upstream's body plus the registered footer.

## Live validation

Run at commit `8bbfc05` (before the rebase and the dry-run withdrawal) on a stock-provisioned installation (gke-admin permission set) with the companion runtime branch deployed, via `CUJ_PROJECT_ID=<test-project> uv run --project bench pytest -s bench/cuj/test_00_agent_responsive.py bench/cuj/obtainability/test_01_cluster_design.py`:

- **2 passed** — all eleven acceptance criteria met by a live agent: real `AdviceService.Capacity` evidence with per-zone signals and all three provisioning models (including an honestly recorded Flex-Start probe failure), a `quota_check` matching the project's real limits, a ComputeClass matching the pinned template (A2 primary, L4/T4 fallbacks), the NAP alternative, and a terminal answer carrying the full report with the quota-vs-capacity distinction and no capacity guarantees.
- Milestones 8/10 at that commit; the two open ones (skills-loaded projection, read-only certification) are the tracked follow-up slice on #804. The dry-run milestone m8 passed there only because the gke-admin install grants `create` on ComputeClasses; on the default read-only ClusterRole the same command is `Forbidden` (verified on a live install by impersonating the platform agent's service account), which is why m8 and the mandate are gone. The suite now has nine milestones.
- Not re-run live after the rebase and the later commits: they remove behavior (the dry-run step, the `apply` carve-out, m8), rename the skill, reorder the template's first two tiers (AC07 checks the family, not the Spot flag), and make the evaluator read task results as part of the answer. On `main` alone the projection carries no task evidence, artifacts, or results, so CUJ1 is scorable only once the runtime half of #804 lands; that was true before this PR for every evidence criterion and is now true for AC05 as well.
- Score progression across the iteration runs: 1/11 → 6/11 → 9/11 → 11/11, each step from a named fix in this PR or the runtime branch. Eleven runs in total; the three later misses each became a structural fix rather than a retry — the recorder now refuses a manifest whose sections collapsed to scalars and capacity evidence that does not name its request, and the report section the model kept omitting is supplied as a template to fill in. None of the three recurred.

## Self-Review

Passes: an adversarial read of the command-policy change on a live cluster (fresh session handed the diff, with `kubectl --as` against the agent's real ClusterRole), a docs-drift read of the skill against the plugin it borrows from, and the two automated `kube-agents-bot` passes on the rebased head. Findings and dispositions, one list:

- **Fixed — the dry-run mandate was unsatisfiable and its carve-out was bypassable.** Server-side dry run needs the same RBAC as the write, so on the shipped read-only grant the mandated command fails and m8 could never pass. The carve-out that admitted it treated every flag as value-taking, so `--dry-run=server --server-side --dry-run=none` passed the gate while kubectl resolved last-wins to a real write (bot finding, reproduced). Both are removed rather than repaired: a read-only validation route does not exist (`kubectl create --dry-run=client --validate=strict` accepted an unknown field and a wrong type against this CRD on kubectl v1.35), and the pinned templates already cover the hallucinated-schema failure that motivated the step.
- **Fixed — the `advice capacity` spellings were unreachable.** The allowlist entry and three flags landed on `main` in #885 while this branch was open; the rebase drops the duplicate entry and keeps only `--max-run-duration`.
- **Fixed — nested code fences.** The skill's JSON, YAML, and report-template fences sat inside list items and rendered inconsistently; every fence is now top-level.
- **Fixed — skill sync would have erased both skill changes** (review finding): `gke-obtainability` renamed to `capacity-obtainability` so the prune loop leaves it alone, and the cluster-creation preflight moved from an in-place edit into `SKILL_FOOTERS`.
- **Fixed — the template's Spot-primary tier** contradicted the gke-compute-classes AI/ML rule and this skill's Rule D; On-Demand A2 now leads, Spot A2 follows.
- **Fixed — the skill mandated tools that do not exist in this tree**: it now says to put the same JSON and manifests in the report when `record_evidence` or `attach_artifact` is unavailable.
- **Fixed — the Flex-Start probe was never shown**, so the allowlisted `--max-run-duration` spelling had to be guessed; Diagnostics D shows it.
- **Fixed — four evaluator defects** (review findings, each reproduced): a leading blank line kept the whole acknowledgment; the generic "will post" and "assigned ... task" branches stripped real opening sentences; `ON_DEMAND` did not match the On-Demand pattern; AC05 demanded an `ondemand` key in AdviceService evidence the criterion never asked for. Plus `write_transcript` is guarded and non-gating, and the SOUL.md template is read in a fixture so a rewording cannot abort collection.
- **Fixed — unnamed constants in `write_transcript`** (bot finding): width, preview lengths, and the transcript filename are module constants.
- **Fixed — evaluator honesty**: acknowledgment stripping matches the hand-off template Kage is actually told to send (bold agent name, backticked task id) and no longer strips a report that merely cites its task id; provisioning-model keys match whole; the guarantee check is strict; `FLEX_START` spelling.
- **Known and accepted — duplicated text.** What `capacity-obtainability` copies verbatim from the `gke-stockout-investigator` plugin is the ComputeClass resilience rules A–H and the section-D `advice` commands, not the alert workflow (that was removed in `77c20c7`). The two copies can drift; extracting them changes a plugin the e2e stockout test exercises and belongs in its own change.
- **Known and accepted — the NAP artifact is a contract shape, not a Kubernetes kind.** `NodeAutoProvisioningSpec` is the structured form AC09–AC11 score (machine families, location policy); real NAP is configured through a gcloud config file plus pod node affinity. The skill says so implicitly by never asking the agent to apply it; making the wording explicit is a follow-up.
- The typed-evidence tools the skill calls (`record_evidence`, `attach_artifact`) ship on the companion runtime branch, not in this diff — the skill mandates behavior that only produces projected evidence once that branch lands.
- LLM adherence varies run to run; the skill pins every pinnable shape, but the 11/11 result should be re-confirmed across repeat runs before the contract is treated as stable.

## Risk & Rollout

Skill content, evaluator/test code, and one flag-arity entry. The kubectl read-only gate is unchanged from `main`: the carve-out this branch once added is gone. No operator or portal code. Revert requires no migration.


bench/cuj/utils/interaction.py
```diff
@@ -129,6 +130,36 @@ def tool_operations(
...
+#: this thread as soon as it's ready." — carry no answer content. Evaluators
+#: must score the reply that follows them, or a journey whose specialist never
+#: reported back would be judged on boilerplate.
+_DELEGATION_ACK = re.compile(
+    r"delegated to the platform agent"
+    r"|\bstarted this\b[^.\n]{0,80}\btask t_[0-9a-f]+"
+    r"|\btask t_[0-9a-f]+\b"
+    r"|\bwill post\b[^.\n]{0,60}\b(?:thread|here)\b",
+    re.IGNORECASE,
+)
```
lapis2002, line 162: `substantive_output` returns the acknowledgment instead of stripping it.

`agents/chat/SOUL.md:121-125` specifies the reply template verbatim:

  ```text
  > 🔀 Delegated to the **platform** agent

  I've started this as task `t_0f6b451f`. The answer will post into this thread as soon as it's ready.

  ## Executive Summary
  ```

Line 139's literal `the platform agent` can't span `**platform**`, and stripping stops at the first non-match.

The six tests pass because `ACK` in `test_cuj_interaction_helpers.py:23` is hand-written prose — no `>`, no emoji, no `**`, not this template.

Suggest `delegated to the \*{0,2}\w+\*{0,2} agent` on 139 and ``task\s+`?t_[0-9a-f]+`` on 140-141, with the fixture built from `SOUL.md`.

Separately, line 141's `\btask t_[0-9a-f]+\b` is unanchored — once the above is fixed, "Design complete for task t_cbb05c69: …" returns `''`. Both should land together.

agents/platform/skills/gke-obtainability/SKILL.md
```diff
@@ -0,0 +1,425 @@
...
+
+When this skill is loaded for a **design, planning, or capacity-check request**
+(for example from the [gke-cluster-creation](../gke-cluster-creation/SKILL.md)
+workflow) rather than a stockout alert:
+
+- Execute **only Step 4's diagnostic subsections A-D** (quota, reservations,
+  usage, and capacity advice). Step 4's Efficiency Directive and CRITICAL
+  MANDATE belong to the alert-driven remediation flow below and do not
+  apply here: in design-time mode there is no notification to send and no
+  Pull Request to open.
```
lapis2002: The scoping fix landed in two places and they disagree. Lines 25-27 say Step 4's CRITICAL MANDATE doesn't apply in design-time, but line 308 was also rewritten to cover both flows, and that mandate carries the requirement to *execute* the quota check, Spot advice, and capacity history, which design-time needs per
line 24.

An agent following line 26 skips `gcloud beta compute advice capacity`, which fails both `ac02-obtainability-api-invoked` and `ac05-provisioning-models-analyzed` in `test_01_cluster_design.py` — two criteria this PR is otherwise tightening.

Separately, line 29 is an orphaned fragment — "quota verification, reservations, and capacity obtainability advice." has no subject. Its sentence head became the parenthetical on line 24.

agents/platform/scripts/command_policy.py
```diff
@@ -863,6 +875,21 @@ def evaluate(argv: list[str]) -> Decision:
...
+        # `apply` with an explicit `--dry-run=server|client` validates a
+        # manifest without persisting anything; the obtainability skill
+        # mandates it before attaching a generated ComputeClass as evidence.
+        # Unlike `diff` — excluded above because its server-side dry-run is
+        # implicit and its failure under a read-only grant reads as breakage —
+        # this form names its intent in the argv, and when RBAC refuses the
+        # dry-run the kubectl error itself is the recorded finding. Only the
+        # `=`-attached spelling counts: a bare `--dry-run` (client by
+        # default in some releases, none in others) stays refused.
+        if verb[:1] == ("apply",) and any(
```
lapis2002: Two holes in this predicate. Neither is reachable from the skill, so it's hardening rather than a break.

**Repeated flag.** kubectl takes the last `--dry-run`; this takes any. So `apply --dry-run=server --dry-run=none -f x.yaml` is allowed and applies for real. Tested on kubectl v1.35.6:

```console
$ kubectl apply --dry-run=server --dry-run=bogus -f x.yaml
error: Invalid dry-run value (bogus).

$ kubectl apply --dry-run=bogus --dry-run=server -f x.yaml
(proceeds)
```

**`verb[:1]` matches every `apply` subcommand.**

```text
ALLOW  apply set-last-applied  --dry-run=server -f cc.yaml
ALLOW  apply edit-last-applied --dry-run=client -f cc.yaml
REFUSE apply view-last-applied -f cc.yaml
```

Two writes in, the family's one read out. `edit-last-applied` spawns `$KUBE_EDITOR`. `SKILL.md:135` only runs plain `apply`.

`test_apply_is_allowed_only_as_an_explicit_dry_run` covers neither case.

jayantid:
**Smoke red triage (gate-watch):** not caused by this PR. All 4 failing cases in [build 2095171818145976320](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/926/pull-kube-agents-smoke-test/2095171818145976320) are NOT_A_REAL_RUN on `HTTP 429: litellm.RateLimitError` (Vertex quota) — known class #1097 (#1184). None of the failing bench tasks are touched by this diff. `/retest` when convenient.

haoxuw:
## Live run transcript

A full CUJ1 journey against a live installation on the current head, exactly as the suite recorded it — `2 passed`, all eleven acceptance criteria met. Project id, session, task id, and operator are redacted; nothing else is edited.

The format mirrors the admin portal's Chat tab: a boxed turn per message, then the delegated work as `assignee · status` cards carrying each typed evidence record (request and analysis as recorded) and each generated artifact. Every run now writes this file next to its JSONL, printed as `Conversation:` in the test output.

<details>
<summary>conversation.txt — click to expand</summary>

```text
CHAT
==============================================================================

┌─ USER  (portal_cuj1_<session>) ──────────────────────
│ Act as my GKE design partner. Design a new production cluster named cuj1-a100-design in project <project-id> for a distributed training workload requiring 32 NVIDIA A100 GPUs in us-central1. Check project quota and current obtainability separately using live evidence; never treat quota, static SKU support, or a documented zone list as proof of live capacity. Compare Autopilot and Standard, recommend resilient zones and provisioning models, and produce a valid ComputeClass fallback design while identifying any fallback that changes the workload's GPU or interconnect requirements. Also provide the alternative for standard node pools: a Node Auto-Provisioning specification allowing cloud.google.com/machine-family values n2, n2d, and c2d with location policy ANY. This is design-only: do not create a cluster, apply a manifest, open a pull request, or mutate cloud or Kubernetes state.
└─────────────────────────────────────────────────────────────────────────────

┌─ ASSISTANT · platform-agent  (default · completed) ─────────────────────────
│ > 🔀 Delegated to the platform agent
│ 
│ I've started this as task t_<task>. The answer will post into this thread as soon as it's ready.
│ 
│ ## Summary of Findings
│ 
│ Quota and live capacity are separate. Although project `<project-id>` has regional quotas that allow distributed training, the on-demand quota limit for `NVIDIA_A100_GPUS` is capped at `16.0` in `us-central1`. To run a `32`-GPU workload, you must use **Spot VMs** (quota: `64.0`) or obtain a quota increase. Live capacity checks indicate highly resilient spot availability in `us-central1-c` and `us-central1-b`.
│ 
│ ---
│ 
│ ## Quota & Obtainability Analysis
│ 
│ Quota is separate from live capacity. The table below represents live results queried directly from `gcloud`:
│ 
│ | Resource Type | Quota Limit | Current Usage | Live Obtainability (Spot) |
│ | :--- | :--- | :--- | :--- |
│ | `NVIDIA_A100_GPUS` (On-Demand) | `16.0` | `0.0` | N/A (Insufficient Quota) |
│ | `PREEMPTIBLE_NVIDIA_A100_GPUS` (Spot) | `64.0` | `0.0` | `0.9` (`us-central1-c` / `-b`) |
│ | `A2_CPUS` (Total CPU constraint) | `192.0` | `0.0` | Matches GPU capacity |
│ 
│ - **On-Demand**: Strictly limited to `16` GPUs by regional quota limit (`16.0`). No reservations are active in project `<project-id>`.
│ - **Spot**: Recommended. Spot advice API returned strong obtainability scores (`0.9`) in `us-central1-c` and `us-central1-b` for `a2-highgpu-8g` VM shapes.
│ - **Flex-Start**: Not queryable due to alpha access limitations in the active project.
│ 
│ ---
│ 
│ ## Provisioning Paths
│ 
│ - **On-Demand** — Project quota limits this path to `16.0` GPUs. A quota increase of `16` units of `NVIDIA_A100_GPUS` is required to run `32` GPUs on-demand. There are currently `0` active reservations.
│ - **Spot** — Preferred path. Zone `us-central1-c` and `us-central1-b` offer high obtainability (`0.9`) for up to `4` nodes of `a2-highgpu-8g` (32 GPUs). The preemption risk is normal; checkpointing to Cloud Storage is strongly recommended.
│ - **Flex-Start** — Incapable of execution due to restricted Compute Engine alpha API permissions.
│ 
│ ---
│ 
│ ## Autopilot vs Standard Comparison
│ 
│ - **GKE Autopilot**: Recommended for automated node provisioning, automated scaling, and security baselines. GKE Autopilot manages node-level upgrades, node-problem detector integrations, and natively implements GKE ComputeClasses.
│ - **GKE Standard**: Recommended only if custom kernels, specialized hyperdisks, or non-GKE orchestration elements are explicitly required.
│ 
│ ---
│ 
│ ## Autopilot ComputeClass & Fallback
│ 
│ To support the workload on GKE Autopilot, apply the following `ComputeClass` configuration. If the primary `a2` machine family runs out of capacity, GKE Autopilot falls back to single-GPU tiers (`nvidia-l4` and `nvidia-tesla-t4`).
│ 
│ **Important Interconnect Caveat:**
│ Falling back to `nvidia-l4` or `nvidia-tesla-t4` **breaks GPUDirect RDMA / multi-node compact placement group interconnect requirements** and severely degrades distributed training performance. The workload should employ check-pointing to mitigate fallback performance impact.
│ 
│ ```yaml
│ apiVersion: cloud.google.com/v1
│ kind: ComputeClass
│ metadata:
│   name: cuj1-a100-design-cc
│ spec:
│   priorities:
│     - machineFamily: a2
│       spot: true
│     - gpu:
│         type: nvidia-l4
│         count: 1
│     - gpu:
│         type: nvidia-tesla-t4
│         count: 1
│   nodePoolAutoCreation:
│     enabled: true
│ ```
│ 
│ ---
│ 
│ ## GKE Standard Node Auto-Provisioning (NAP)
│ 
│ If choosing GKE Standard with Node Auto-Provisioning (NAP), apply this alternate specification allowing machine families `n2`, `n2d`, and `c2d` with location policy `ANY` to bypass hard zonal constraints.
│ 
│ ```yaml
│ kind: NodeAutoProvisioningSpec
│ spec:
│   nodeAffinity:
│     requiredDuringSchedulingIgnoredDuringExecution:
│       nodeSelectorTerms:
│         - matchExpressions:
│             - key: cloud.google.com/machine-family
│               operator: In
│               values: [n2, n2d, c2d]
│   location:
│     locationPolicy: ANY
│ ```
└─────────────────────────────────────────────────────────────────────────────

AGENT WORK
==============================================================================

  platform · done  (t_<task>, run 1)
    GKE Cluster Design cuj1-a100-design (32 NVIDIA A100 GPUs in us-central1)
    ✓ Completed cuj1-a100-design cluster design and capacity diagnostics in us-central1 (32x A100).
    · evidence quota_check [completed] — gcloud.compute.regions.describe
        request:
        {
          "project": "<project-id>",
          "region": "us-central1"
        }
        analysis:
        {
          "A2_CPUS": {
            "limit": 192,
            "usage": 0
          },
          "NVIDIA_A100_80GB_GPUS": {
            "limit": 0,
            "usage": 0
          },
          "NVIDIA_A100_GPUS": {
            "limit": 16,
            "usage": 0
          },
          "PREEMPTIBLE_NVIDIA_A100_80GB_GPUS": {
            "limit": 0,
            "usage": 0
          },
          "PREEMPTIBLE_NVIDIA_A100_GPUS": {
            "limit": 64,
            "usage": 0
          }
        }
    · evidence advice_service_capacity [completed] — compute.beta.AdviceService.Capacity
        request:
        {
          "acceleratorCount": 32,
          "acceleratorType": "nvidia-a100",
          "region": "us-central1"
        }
        analysis:
        {
          "availableQuantity": 32,
          "provisioningModels": {
            "FLEX_START": {
              "notes": "Alpha Access permission error, blocked on GCP side.",
              "status": "failed"
            },
            "ON_DEMAND": {
              "notes": "Quota limit is 16.0 for on-demand NVIDIA_A100_GPUS; insufficient for a 32-GPU single-cluster on-demand deployment without a quota increase. No advance obtainability signal exists.",
              "source": "quota+reservations"
            },
            "SPOT": {
              "obtainability": 0.9,
              "zone": "us-central1-c"
            }
          },
          "zones": [
            {
              "obtainability": 0.9,
              "zone": "us-central1-c"
            },
            {
              "obtainability": 0.9,
              "zone": "us-central1-b"
            },
            {
              "obtainability": 0.1,
              "zone": "us-central1-f"
            },
            {
              "obtainability": 0.1,
              "zone": "us-central1-a"
            }
          ]
        }
    · evidence computeclass_server_dry_run [completed] — kubectl.apply.dry_run
        request:
        {}
        analysis:
        {
          "dry_run_success": true,
          "manifest_file": "computeclass.yaml",
          "validation": "passed"
        }
    · artifact computeclass
        {
          "apiVersion": "cloud.google.com/v1",
          "kind": "ComputeClass",
          "metadata": {
            "name": "cuj1-a100-design-cc"
          },
          "spec": {
            "nodePoolAutoCreation": {
              "enabled": true
            },
            "priorities": [
              {
                "machineFamily": "a2",
                "spot": true
              },
              {
                "gpu": {
                  "count": 1,
                  "type": "nvidia-l4"
                }
              },
              {
                "gpu": {
                  "count": 1,
                  "type": "nvidia-tesla-t4"
                }
              }
            ]
          }
        }
    · artifact node_auto_provisioning
        {
          "kind": "NodeAutoProvisioningSpec",
          "spec": {
            "location": {
              "locationPolicy": "ANY"
            },
            "nodeAffinity": {
              "requiredDuringSchedulingIgnoredDuringExecution": {
                "nodeSelectorTerms": [
                  {
                    "matchExpressions": [
                      {
                        "key": "cloud.google.com/machine-family",
                        "operator": "In",
                        "values": [
                          "n2",
                          "n2d",
                          "c2d"
                        ]
                      }
                    ]
                  }
                ]
              }
            }
          }
        }
    · report delivered to the user (also folded into the reply above)

```

</details>

Notes for a reviewer reading it:

- The assistant turn carries the specialist's full report, not just the delegation line — that is the terminal-output delivery the companion runtime branch adds.
- `## Provisioning Paths` is the section the skill supplies as a fill-in template; all three paths are named, including the Flex-Start probe's honest failure.
- The typed evidence under AGENT WORK is what the acceptance criteria score: a `quota_check` naming its region, an `advice_service_capacity` record naming the shape, count, and region it asked for with per-zone signals, and a server dry run of the generated ComputeClass.

agents/platform/scripts/command_policy.py
```diff
@@ -569,6 +572,39 @@ def _gcloud_is_read_only(words: list[str]) -> bool:
...
+        elif token.startswith("-"):
+            # Any other bare flag is assumed to consume the next token. That
+            # over-consumes for booleans (`--server-side --dry-run=server`
+            # reads as unreadable rather than allowed), which is the safe
+            # direction: the alternative is enumerating every apply flag's
+            # arity and allowing a real write whenever the table falls behind
+            # a kubectl release.
+            index += 1
+        index += 1
+    return mode
```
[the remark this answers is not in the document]
haoxuw: Reproduced on the previous head (`apply -f cc.yaml --dry-run=server --server-side --dry-run=none` was allowed). Fixed in a53d75e4 by withdrawing the carve-out rather than repairing the parser: `_dry_run_mode` and the `apply` branch are deleted, so the kubectl gate is back to what `main` ships, and `test_apply_stays_refused_even_as_an_explicit_dry_run` pins this exact spelling as refused. The skill mandate that needed the carve-out is gone from the same commit (server-side dry run is `Forbidden` under the read-only ClusterRole anyway, verified live with `--as` the platform agent SA).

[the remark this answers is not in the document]
haoxuw: Reproduced on the previous head (`apply -f cc.yaml --dry-run=server --server-side --dry-run=none` was allowed). Fixed in a53d75e4 by withdrawing the carve-out rather than repairing the parser: `_dry_run_mode` and the `apply` branch are deleted, so the kubectl gate is back to what `main` ships, and `test_apply_stays_refused_even_as_an_explicit_dry_run` pins this exact spelling as refused. The skill mandate that needed the carve-out is gone from the same commit (server-side dry run is `Forbidden` under the read-only ClusterRole anyway, verified live with `--as` the platform agent SA).

agents/platform/skills/gke-obtainability/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+  capacity. Use when a cluster design, capacity plan, or scale-up decision
+  needs live evidence: regional quota verification (gcloud compute regions
+  describe), reservations, and capacity obtainability advice (gcloud beta
+  compute advice capacity / capacity-history) across zones and provisioning
+  models. For handling an inbound cluster-autoscaler stockout alert end to
+  end — triage, GitOps remediation, Pull Request — use the
+  gke-stockout-investigator plugin skill instead.
+metadata:
+  category: Containers
+---
```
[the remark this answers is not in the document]
haoxuw: Agreed — those bullets described the skill before 77c20c7. The Self-Review is rewritten against the current head: it now names what is actually duplicated from the plugin (resilience rules A–H and the section-D commands), records the dry-run mandate and carve-out as withdrawn in a53d75e4, and drops the Step 4 / alert-flow claims.

agents/platform/skills/gke-obtainability/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+# GKE Capacity Obtainability
+
+Quota and capacity are different questions. Quota is a project limit you can
+raise by asking; obtainability is whether the hardware is actually free in a
+zone right now, and no amount of quota makes a stocked-out shape appear. This
+skill gathers live evidence for both, records it in a form a reviewer can
+verify, and states the distinction in the answer.
+
+It is diagnosis and design only: nothing here creates a cluster, applies a
+manifest, opens a Pull Request, or mutates cloud or Kubernetes state.
```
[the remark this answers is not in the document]
haoxuw: Agreed — those bullets described the skill before 77c20c7. The Self-Review is rewritten against the current head: it now names what is actually duplicated from the plugin (resilience rules A–H and the section-D commands), records the dry-run mandate and carve-out as withdrawn in a53d75e4, and drops the Step 4 / alert-flow claims.

bench/cuj/utils/evidence.py
```diff
@@ -32,6 +32,85 @@ def temporary(
...
+        """
+
+        def block(role: str, subtitle: str, body: str) -> list[str]:
+            head = f"{role}" + (f"  ({subtitle})" if subtitle else "")
+            return [
+                f"┌─ {head} " + "─" * max(0, 74 - len(head)),
+                *[f"│ {line}" for line in (body or "(empty)").splitlines()],
+                "└" + "─" * 77,
+                "",
+            ]
```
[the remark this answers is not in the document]
haoxuw: Fixed in ad5ab9c1: `TRANSCRIPT_FILENAME`, `TRANSCRIPT_WIDTH`, `EVIDENCE_PREVIEW_LINES`, and `ARTIFACT_PREVIEW_LINES` are module constants at the top of `evidence.py`; rendered widths are unchanged (smoke-tested at 78 columns).

Fuxiao-Gao (commented):
Thanks for the PR. Findings below, most severe first. Inline comments carry the detail.

**Skill files under `agents/platform/skills/gke-*` are overwritten on the next skill sync** (see #1440). `scripts/sync-upstream-skills.py` mirrors `google/skills` (`skills/cloud/gke-*`): it deletes each `gke-*` directory and recopies it from upstream, then prunes any local `gke-*` directory that upstream does not have. That affects both skill changes here:

- `gke-cluster-creation/SKILL.md` step 3a is an in-place edit to a mirrored file. It disappears on the next sync. The supported path is a `SKILL_SUBSTITUTIONS` or `SKILL_FOOTERS` entry in the sync script.
- `gke-obtainability/` is new and does not exist upstream, so the prune loop (`local_name.startswith("gke-") and local_name not in discovered_skills`) removes the whole directory. It needs either a name without the `gke-` prefix, or an explicit exclusion in the sync script.

Summary of the other findings:

1. ac05 is now gating on the user-visible answer, but on this branch the portal output is only the delegation ack, so CUJ1 cannot pass until the result folding from feat/804-projection lands.
2. The skill mandates `record_evidence` and `attach_artifact` tools that do not exist in this tree.
3. `substantive_output` keeps the whole ack when the output starts with a blank line.
4. The On-Demand regex rejects `ON_DEMAND`; the Flex regex was widened for `FLEX_START` but this one was not.
5. ac05's evaluator requires an `ondemand` key in the AdviceService evidence, which the criterion text does not ask for and the API does not return.
6. The pinned ComputeClass template puts Spot on the primary A2 tier, which conflicts with gke-compute-classes' AI/ML rule and this skill's Rule D.
7. `write_transcript` has no type guards and is not wrapped, so a malformed entry turns a scored run into a test ERROR.
8. The test module parses SOUL.md with an `assert` at import time, so a template rewording aborts the whole bench test session at collection.
9. The command-policy test says the skill "issues these exactly", but the skill never shows a Flex-Start command or `--max-run-duration`.
10. The generic "will post" and "assigned ... task t_" ack branches strip real answer sentences.

Lower priority, not commented inline: the `interaction_final` record is always a byte-identical duplicate of the previous record; `SKILL_GROUPS` in `scripts/generate_docs.py` was not updated so the new skill lands under "Other"; step "3a." is a lazy continuation line and renders inline inside step 3 on the docs site; `bench/cuj/README.md:23` still says every poll response is logged.

agents/platform/skills/gke-cluster-creation/SKILL.md
```diff
@@ -32,6 +32,16 @@ for all new clusters.
     template or mode, present the available templates (e.g., Autopilot, Standard
     Regional, GPU Inference, AI Hypercompute) and explain key trade-offs (Cost
     vs. Availability, Autopilot vs. Standard node management).
+    3a. **Check quota and live obtainability (GPU/TPU or large-shape
```
Fuxiao-Gao, line 35: This is an in-place edit to a file mirrored from `google/skills`. `scripts/sync-upstream-skills.py` rmtree's this directory and recopies it from upstream on every run, so step 3a is gone after the next sync (#1440 has the details and a recent example from #1226). The supported way to keep it is a `SKILL_SUBSTITUTIONS` entry in the sync script, or a `SKILL_FOOTERS` block like the one already registered for this skill.

agents/platform/skills/gke-obtainability/SKILL.md
```diff
@@ -0,0 +1,278 @@
+---
```
Fuxiao-Gao, line 1: The sync script prunes every local `agents/platform/skills/gke-*` directory that is not present upstream:

```python
if local_name.startswith(SKILL_PREFIX) and local_name not in discovered_skills:
    shutil.rmtree(stale_path)
```

`gke-obtainability` is not in `google/skills`, so the next sync deletes this whole skill. Options: name it without the `gke-` prefix (the cluster agent's repo-native skills follow that convention), or add an explicit exclusion list to the sync script. See #1440.

bench/cuj/obtainability/test_01_cluster_design.py
```diff
@@ -456,12 +444,40 @@ def evaluate_acceptance(interaction: dict[str, Any]) -> AcceptanceCriteria:
...
+            {"ondemand"},
+        )
+    )
+    # Score the reply that follows the delegation acknowledgment; the
+    # hand-off boilerplate is not an answer.
+    answer = substantive_output(interaction)
+    # Word-bounded so prose like "inflexible" or "spotted" cannot satisfy a
+    # provisioning path; "flex" still matches Flex, Flex-Start, and
+    # flex-start spellings.
+    paths_weighed = {
```
Fuxiao-Gao, line 470: ac05 is now gating on `paths_weighed`, which reads the user-visible `output` after ack stripping. On this branch `admin_console/chat/service.py` sets `output` to Kage's root reply at `root.completed`, and `_settle_tasks` never folds task results into it. SOUL.md section 4 says the ack block is the whole reply. So `substantive_output()` returns an empty string on every delegated run (checked with the SOUL template), all three `paths_weighed` flags are False, ac05 fails, and `scenario.py:143` asserts `acceptance.passed`.

Before this PR ac05 was evidence-only. The result folding lives on the unmerged feat/804-projection branch, so merging this PR alone makes CUJ1 unpassable. Either land that branch first, or keep ac05 evidence-only until it lands.

agents/platform/skills/gke-obtainability/SKILL.md
```diff
@@ -0,0 +1,278 @@
...
+  sentence verbatim in your final report: "Quota is separate from live
+  capacity." Quota is a project limit; it does not prove hardware is
+  obtainable, and obtainability evidence does not raise quota. **Never use
+  the word "guarantee" about capacity, allocations, or scheduling anywhere
+  in the report** — not even for reservations or ProvisioningRequests;
+  write "reserves", "holds", or "provides once scheduled" instead.
+
+### Record what you executed as typed evidence
+
+Use the `record_evidence` tool — one record per check, built from the real
```
Fuxiao-Gao, line 53: `record_evidence` and `attach_artifact` do not exist anywhere in this tree. A grep hits only this file, and `platform_mcp_server.py` exposes nothing equivalent. gke-cluster-creation step 3a now routes every GPU/TPU/large-shape design through this skill, so the agent is told to use a nonexistent tool under a MANDATE with no fallback. A one-line "if the tool is unavailable, put the same JSON in the report" would let it degrade gracefully until the runtime lands.

bench/cuj/utils/interaction.py
```diff
@@ -129,6 +141,80 @@ def tool_operations(
...
+    skipping = True
+    for paragraph in re.split(r"\n\s*\n", text):
+        if not skipping:
+            kept.append(paragraph)
+            continue
+        sentences = re.split(r"(?<=[.!?])\s+", paragraph.strip())
+        remainder = list(sentences)
+        while remainder and _DELEGATION_ACK.search(remainder[0]):
+            remainder.pop(0)
+        if remainder:
```
Fuxiao-Gao, line 186: If `output` starts with a blank line, `re.split` yields a first paragraph of `''`, sentences is `['']`, `remainder` is `['']` which is truthy, so `skipping` flips to False and the entire ack is kept. Checked with:

```
"\n\n> 🔀 Delegated to the **platform** agent\n\nI've started this as task `t_..`. The answer will post..."
```

which returns the full ack instead of `''`. A whitespace-only remainder should still count as skipping.

bench/cuj/obtainability/test_01_cluster_design.py
```diff
@@ -456,12 +444,40 @@ def evaluate_acceptance(interaction: dict[str, Any]) -> AcceptanceCriteria:
...
+        "onDemand": bool(re.search(r"\bon[- ]?demand\b", answer, re.IGNORECASE)),
```
Fuxiao-Gao, line 471: `\bon[- ]?demand\b` does not match `ON_DEMAND`, while the flex pattern four lines down was widened for `FLEX_START`. An answer that echoes the key names the skill mandates ("SPOT 0.9, FLEX_START probed, ON_DEMAND assessed from quota") fails this check. `\bon[-_ ]?demand\b` fixes it.

bench/cuj/obtainability/test_01_cluster_design.py
```diff
@@ -456,12 +444,40 @@ def evaluate_acceptance(interaction: dict[str, Any]) -> AcceptanceCriteria:
...
+    populated_models = {
+        key
+        for key, value in normalized_models.items()
+        if value not in (None, "", [], {})
+    }
+    # Exact keys, not prefixes: a stray "spotplaceholder" is not a Spot
+    # signal. On-Demand is required too — it has no AdviceService signal, but
+    # the skill mandates recording the quota-and-reservations assessment
+    # under the same key, and that recording never depends on the API.
+    advice_models_covered = all(
```
Fuxiao-Gao, line 456: This requires a populated `ondemand` key in the AdviceService evidence. The criterion's proof text (line 163) asks for non-empty SPOT and FLEX signals, and the PR body says the same. An agent that records `provisioningModels={SPOT, FLEX_START}` and puts On-Demand in `quota_check` plus the report fails here while meeting every stated proof. Only an agent that copies SKILL.md line 82 verbatim passes, and that record then attributes On-Demand data to `compute.beta.AdviceService.Capacity`, which never returned it. Suggest taking On-Demand from `quota_check` (m5) plus the existing `paths_weighed['onDemand']` check and dropping the alias set.

agents/platform/skills/gke-obtainability/SKILL.md
```diff
@@ -0,0 +1,278 @@
...
+
+```yaml
+apiVersion: cloud.google.com/v1
+kind: ComputeClass
+metadata:
+  name: <design>-cc
+spec:
+  priorities:
+    - machineFamily: a2          # primary: the requested accelerator family
+      spot: true
```
Fuxiao-Gao, line 129: `spot: true` on the primary a2 tier, with no On-Demand or Flex-Start A2 tier ahead of it, conflicts with gke-compute-classes' CRITICAL AI/ML RULE (Reservations, then On-Demand, then DWS FlexStart, then Spot; never Spot primary) and with this skill's own Rule D. test_01 loads both skills into one run, so the agent gets two contradicting templates for the same manifest, and this skill says departing from the template is "a finding, not a deliverable". ac07's `_is_a2_priority` ignores `spot`, so dropping it or adding an On-Demand a2 tier first still passes.

bench/cuj/utils/evidence.py
```diff
@@ -32,6 +41,92 @@ def temporary(
...
+                f"  {task.get('assignee', 'unassigned')} · {task.get('status')}"
+                f"  ({task.get('taskId')}, run {task.get('runCount')})"
+            )
+            lines.append(f"    {task.get('title') or ''}")
+            if task.get("summary"):
+                lines.append(f"    ✓ {task['summary']}")
+            elif task.get("error"):
+                lines.append(f"    ✗ {task['error']}")
+            for item in task.get("evidence") or []:
+                details = item.get("details") or {}
```
Fuxiao-Gao, line 94: `write_transcript` calls `.get()` on every evidence and artifact entry and on `details` without the isinstance guards that `projected_records`, `completed_evidence`, and `_mapping` all use. Its call site at `scenario.py:103` is not wrapped, and the surrounding try only catches OSError, ValueError, and PortalError. So a task that projects `evidence=['quota_check']` or `details='n/a'` scores fine in the evaluators but raises AttributeError here, and the run ends with no summary record and no printed report. Milestones are wrapped non-gating at `scenario.py:91-98`; the transcript should be too.

bench/tests/test_cuj_interaction_helpers.py
```diff
@@ -0,0 +1,102 @@
...
+        encoding="utf-8"
+    )
+    block = re.search(
+        r"```\n(\s*> [^\n]*Delegated to the .*?)```", soul, re.S
+    )
+    assert block, "agents/chat/SOUL.md no longer shows the delegation template"
+    return textwrap.dedent(block.group(1)).strip()
+
+
+ACK = (
```
Fuxiao-Gao, line 42: `_soul_handoff_template()` reads and regex-parses SOUL.md with an `assert` at module import. Any rewording of the delegation template in `agents/chat/SOUL.md` becomes a pytest collection error, and `make test-bench` runs bare `python3 -m pytest bench/tests/` with no `--continue-on-collection-errors`, so the whole bench session aborts instead of two or three tests failing. Computing ACK inside a fixture avoids that.

agents/platform/scripts/test_command_policy.py
```diff
@@ -1028,6 +1028,58 @@ def test_the_stockout_sop_spellings_reach_their_allowlist_entries(self):
             with self.subTest(desc=desc):
                 self.assertTrue(evaluate(argv).allowed, desc)
 
+    def test_the_obtainability_skill_spellings_reach_their_allowlist_entries(self):
```
Fuxiao-Gao, line 1031: The test name and the arity entry say the skill "issues these exactly", but gke-obtainability/SKILL.md never shows a Flex-Start command and does not mention `--max-run-duration`. FLEX_START appears only in prose and in the JSON templates. The agent has to guess the spelling, and anything other than the allowlisted flags is refused fail-closed as `gcp.unreadable-command`. The "honestly recorded Flex-Start probe failure" in the PR's live run is consistent with that. Adding the Flex-Start command to Diagnostics D would close the gap.

bench/cuj/utils/interaction.py
```diff
@@ -129,6 +141,80 @@ def tool_operations(
...
+#:     > 🔀 Delegated to the **<agent-name>** agent
+#:
+#:     I've started this as task `<task_id>`. The answer will post into this
+#:     thread as soon as it's ready.
+#:
+#: Matching has to survive that formatting — the agent name arrives wrapped in
+#: bold markers and the task id in backticks — and must not fire on a report
+#: that merely cites its own task id. Every branch therefore pairs hand-off
+#: phrasing with the thing handed off.
+_DELEGATION_ACK = re.compile(
```
Fuxiao-Gao, line 155: The "results ... will post" and "assigned ... task t_" branches strip legitimate opening sentences of a real answer. Checked: "Results for the A100 design will post below. Quota is separate from live capacity." loses its first sentence, and "Assigned under task t_0d0778b9: start 14:00 UTC in us-central1-a. Quota is separate." loses the sentence naming the zone and time. In test_03 that sentence is what ac05-start-window-recommended searches for. Anchoring these branches to the hand-off template wording ("will post into this thread", "started this as task") would avoid the false negative.

Fuxiao-Gao (approved):

agents/platform/skills/gke-cluster-creation/SKILL.md
haoxuw: Agreed. d2dc6084 reverts the in-place step 3a (the file is upstream's body again) and adds the preflight as a section in the registered `SKILL_FOOTERS["gke-cluster-creation"]` entry, so the sync re-applies it. Verified the checked-in file equals upstream body + footer byte for byte; `scripts/test_sync_upstream_skills.py` green.

agents/platform/skills/capacity-obtainability/SKILL.md
```diff
@@ -0,0 +1,278 @@
+---
```
haoxuw: Agreed. Renamed to `capacity-obtainability` in d2dc6084, following the repo-native convention, so the prune loop leaves it alone. References in the CUJ test, policy comments, docs catalogue (under Cost and capacity), and roster updated.

bench/cuj/obtainability/test_01_cluster_design.py
haoxuw: You are right that on `main` the projection carries no task results, and also no task evidence or artifacts, so every evidence criterion is already unscorable until the runtime half of #804 lands; AC05 is not the odd one out. Rather than drop the answer check, cc33c4d2 scores what the user actually receives: `delivered_answer()` is the substantive root output plus each projected task `result`, which is what the gateway posts into the thread per SOUL.md §5. The PR body now says plainly that CUJ1 is scorable only with that branch deployed.

agents/platform/skills/capacity-obtainability/SKILL.md
```diff
@@ -0,0 +1,278 @@
...
+  sentence verbatim in your final report: "Quota is separate from live
+  capacity." Quota is a project limit; it does not prove hardware is
+  obtainable, and obtainability evidence does not raise quota. **Never use
+  the word "guarantee" about capacity, allocations, or scheduling anywhere
+  in the report** — not even for reservations or ProvisioningRequests;
+  write "reserves", "holds", or "provides once scheduled" instead.
+
+### Record what you executed as typed evidence
+
+Use the `record_evidence` tool — one record per check, built from the real
```
haoxuw: Agreed. d2dc6084 adds the fallback: if `record_evidence` or `attach_artifact` is not among the tools, put the same JSON under an `## Evidence` heading and the manifests as fenced YAML in the report, and do not skip the check.

bench/cuj/utils/interaction.py
haoxuw: Reproduced and fixed in cc33c4d2: whitespace-only sentences are dropped before the acknowledgment scan, so a leading blank paragraph no longer ends the skipping. Pinned by `test_an_acknowledgment_behind_leading_blank_lines_still_scores_as_silence`.

bench/cuj/obtainability/test_01_cluster_design.py
haoxuw: Fixed in cc33c4d2: `\bon[-_ ]?demand\b`.

bench/cuj/obtainability/test_01_cluster_design.py
haoxuw: Agreed. cc33c4d2 drops the `ondemand` alias from the AdviceService evidence check; On-Demand is scored on the delivered answer and the quota check (m5). The skill's evidence template no longer carries an `ON_DEMAND` entry either (d2dc6084), so the record attributes to the API only what the API returned.

agents/platform/skills/capacity-obtainability/SKILL.md
```diff
@@ -0,0 +1,278 @@
...
+
+```yaml
+apiVersion: cloud.google.com/v1
+kind: ComputeClass
+metadata:
+  name: <design>-cc
+spec:
+  priorities:
+    - machineFamily: a2          # primary: the requested accelerator family
+      spot: true
```
haoxuw: Agreed. d2dc6084 reorders the template: On-Demand A2 first, Spot A2 second, then L4 and T4, and cites the gke-compute-classes rule and Rule D beside it. AC07 still passes on the first tier's family.

bench/cuj/utils/evidence.py
haoxuw: Fixed in cc33c4d2: `write_transcript` guards every task, evidence, artifact, details, and manifest entry with isinstance checks, and its call in `scenario.py` is wrapped and non-gating like the milestones, recording `transcript_error` instead of failing the run.

bench/tests/test_cuj_interaction_helpers.py
haoxuw: Fixed in cc33c4d2: the SOUL.md read is a pytest fixture, so a reworded template fails the tests that use it and nothing else in the session.

agents/platform/scripts/test_command_policy.py
haoxuw: Agreed. d2dc6084 adds the Flex-Start probe with `--max-run-duration` to Diagnostics D, and the test comment now points at that section rather than a design-time section that no longer exists.

bench/cuj/utils/interaction.py
haoxuw: Reproduced both and fixed in cc33c4d2: the matcher is anchored to the template's wording (`started this as task t_...`, `will post into this thread`) plus the delegated-to-agent line; the generic branches are gone. Both sentences you named now pass through, pinned by `test_answer_sentences_that_resemble_a_hand_off_are_kept`.

haoxuw:
@Fuxiao-Gao thanks for the thorough pass. Every inline finding reproduced and is fixed in d2dc6084 (skill, sync footer, rename) and cc33c4d2 (bench), with a reply on each thread. The four lower-priority items from your summary are handled in the same commits:

- `interaction_final` no longer repeats the last transition; it records the poll count and terminal status only.
- `capacity-obtainability` is registered in `SKILL_GROUPS` under Cost and capacity, and the catalogue and roster are regenerated.
- Step 3a is gone; the preflight is a proper section in the registered footer, so it renders as its own heading.
- `bench/cuj/README.md` now says transitions are logged, not every poll, and mentions `conversation.txt`.

PR #21, opened 2026-05-25
feat(examples): add LiteLLM Gemini example

bradhoekstra:
# Pull Request

## Why This Change
Adds a sample deployment for LiteLLM configured to use Google's Gemini models on Kubernetes.

## What Changed
Added files in `examples/litellm-gemini/`.

**Files:**
- `examples/litellm-gemini/configmap.yaml`
- `examples/litellm-gemini/deployment.yaml`
- `examples/litellm-gemini/secret.yaml`
- `examples/litellm-gemini/service.yaml`

**Added/Changed/Fixed:**
- Added LiteLLM configuration for `gemini-3.1-flash-lite`.
- Added K8s Deployment, Service, and Secret manifests.

## Why This Matters
Provides a starting point for users to run LiteLLM with Gemini on GKE.

## Context
None.

## Testing
Manifests are static examples. Not tested in a live cluster.

PR #930, opened 2026-08-25
docs(designs): A2A-over-JetStream design set - wire protocol, NATS deployment, chatops gateway, subagent profiles, mode switch

bnaylor:
## Summary

Five design documents for carrying agent-to-agent traffic over NATS JetStream with A2A object payloads, landing ahead of the implementation as the contract it builds against: the wire protocol (`a2a-jetstream/0.4`), the NATS deployment and client resilience contract, the chatops gateway, declarative subagent profiles, and the mode switch that keeps all of it dark until enabled.  The five docs cross-reference each other and record their decision trail inline with dated strikethroughs.

## Why This Change

The interior couples chat ingress, delegation, and status to one harness process and a SQLite kanban board.  These designs replace that with a durable message bus that is also the audit substrate - status answered by replay, one correlation identifier from the user's question through every hop - and make sessions and subagents pods the platform's enforcers can see.  Without an agreed contract, the client library and gateway now being built have nothing to conformance-test against.

## What Changed

- `docs/designs/spec-a2a-payloads.md` - the envelope and layering rule (envelope is ours, payloads are standard A2A objects), addressee-scoped task subjects, the task lifecycle on JetStream, the topic namespace, and the conformance assertions the client library ships with.
- `docs/designs/spec-nats-deployment.md` - streams and retention (acknowledgement never deletes; the stream is the audit buffer, the log sink is the archive), auth-callout accounts with deny-by-default per-user grants, and the client resilience contract NR-1..5 with its restart conformance assertion.
- `docs/designs/spec-chatops-gateway.md` - a deterministic gateway (no model of its own), session = one backend conversation = one contextId, a pseudonymous advisory `authority` block, the group-chat substrate, Discord as the test backend.
- `docs/designs/spec-subagent-profiles.md` - the `AgentProfile` resource, submission as a bus message rather than an API call, dispatcher-rendered Jobs, the janitor for orphaned tasks, and the kanban retirement inventory with its two named gaps.
- `docs/designs/spec-mode-switch.md` - one optional `spec.mode` enum and a resolve/render helper pair; Degraded with a named reason on version skew.
- `docs/README.md` - five map rows.

## Context

- Adjacent, not overlapping: #913 (sandbox pod execution isolation) touches worker pod posture; these docs assume the same no-credential worker shape.  #730 is `docs(designs)` precedent in path only.  No open PR or issue covers this material.
- The capability shape defers to `docs/architecture/09-capability-envelope.md`; the relationship between `AgentProfile` and 06/08's `Agent` is stated in spec-subagent-profiles and settles in the 01-08 revision.
- Open questions in these docs are resolved inline and dated, except those explicitly parked with the authority workstream (per-task credentials, signed-claim vs subject-derived identity).

## Testing

- `make docs-check` green: generated regions, relative links, terminology, map coverage, context budget.
- `prettier` 3.9.6 clean on all six changed files.

### Live validation

Not live-tested: docs-only, no code path reaches a running installation.  Two behavioral claims inside the specs were verified against a live NATS v2.10.29 on a GKE dev cluster before they hardened into spec text: client-set headers pass through verbatim, `Nats-*` names included, so the reserved `identity` field cannot be a server-stamped header (docs and server source also swept through v2.14.5); and a user without a publish grant is refused at publish time with nothing delivered to any consumer.  The idle-footprint and spin-up numbers cited in the gateway and profile docs were measured on the same cluster (270Mi RSS flat over a 30-minute soak; ~5s warm pod-create to first token).

## Self-Review

Both required passes ran in fresh contexts - subagents handed the diff range and the repo skill, not the authoring conversation: `review-adversarial` and `review-docs-drift`.  Merged dispositions:

- **Fixed (Blocking, adversarial):** connection-time authorization was unimplementable on the task plane - taskId-keyed subjects cannot carry per-executor grants, so every grant collapsed to `a2a.tasks.>`.  Task subjects now carry the addressee (`a2a.tasks.{addressee}.{taskId}.*`), protocol bumped to 0.4, and the deployment spec's deny-by-default grants are expressible as written.
- **Fixed (Blocking, drift):** "we hold no signing keys" contradicted `docs/architecture/09-capability-envelope.md` - corrected: the callout issuer key exists, and 09 owns its analysis.
- **Fixed:** the reserved capability was described as an inline token where 09 (Agreed) settled on a KV-backed reference - aligned and linked.
- **Fixed:** plaintext principal emails and rosters on the durable stream contradicted the shipped attribution posture - `authority` identifiers are now HMAC-pseudonymous per `docs/designs/audit-logging-user-attribution.md`, plaintext only in the gateway's local ingress log.
- **Fixed:** agent-card publisher, trigger, and subject token contradicted between the payload and profile specs - reconciled to operator-publishes-per-profile.
- **Fixed:** the janitor grant claimed kind-level enforcement NATS cannot do - restated as a subject-level grant plus a conformance assertion; "the janitor, and only the janitor" corrected to the ratified two-supervisor split.
- **Fixed:** the mode switch's Degraded path was unreachable through its own sole accessor - the helper became a resolve/render pair.
- **Fixed:** dispatcher head-of-line blocking across profiles under `MaxAckPending` - one durable consumer per profile.
- **Fixed:** taskId/correlationId semantics for follow-ups and steers were contradictory - follow-ups reuse the task's ids and are attributed by their own envelope.
- **Fixed, smaller:** three gateway-body passages stale against companions in the same diff; `AgentProfile`-to-06/08 relationship paragraph added; `agent-communication.md` named and linked from both docs that replace parts of it; map-row status downgraded to "Draft for review" to match the docs' own headers; hard-coded skill counts dropped; an unverifiable demo file:line citation reduced to prose; the max-payload justification corrected (server errors, it does not drop silently); `rejected` attributed to A2A 1.0 natively; the heartbeat `{owner}` token defined; a stale retention-placeholder sentence updated; the gateway's `session-state` KV bucket added to the deployment inventory; the Job-GC redelivery window documented as noted-not-defended.
- **Rejected with evidence (adversarial):** "`kanban_notify_propagate.py` does not exist" - it exists at `agents/platform/scripts/kanban_notify_propagate.py`; the pass searched only `deploy/docker/patches/`.
- **Deliberately not fixed:** a forward-pointer banner on `agent-communication.md` itself.  The two new docs link to it by name; the banner belongs to the stage 3 change that actually retires the paths it describes, and editing it here widens this diff for a promise the code has not made yet.
- **Verified clean by the passes:** every other file:line citation against this tree; A2A 1.0 external claims against the spec site; retention and stream numbers consistent across all five docs; the arXiv handoff-feasibility citation says what the doc claims.
- **Not independently verified:** the Synadia agent-protocol 0.3 details (inactivity timeout, `$SRV` discovery, empty-payload terminator) - taken from its published docs by the authoring sessions; named here rather than implied covered.

Three reviewer passes arrived after opening (kyber775, rune42808, clomp42); the per-finding record lives in their comment threads, and this list stays current against the head:

- **Fixed (9d2a567):** addressee and taskId are dot-free tokens (assertion 2); the dispatcher creates Jobs only for event-less tasks, which kills 409 spam and post-TTL resurrection; the dedup window is specced as bounded (the in-flight library already complied); post-`final` events are warn-and-drop with the consumer surviving; the attribution salt is one shared Secret; persona image volumes inherit the operator's existing 1.35 gate; TokenReview/JWKS named for callout validation.
- **Fixed (6d24617):** dispatcher and janitor run under the operator's existing leader election, required on under `mode: next`; task `in` subjects are dual-reader by design - dispatcher for new tasks, executor via its own ephemeral consumer for follow-ups, steers, and cancels; `contextId` mint is a KV `Create` (CAS); profile deletion is finalizer-ordered with level-triggered card republish; correlationId/taskId/contextId are opaque random tokens by rule; `mode: next` states it ships no audit archive; callout-down blast radius stated as dark-to-new-work.
- **Added (d3dd391):** conformance assertion 21, steering delivery in negative-test shape - drafted by a reviewer, folded with one edit (test home named by component, not schedule).
- **Fixed (f406bb9):** a mode change is a rollout via a `checksum/config` pod-template annotation (no prior operator pattern - verified); bucket access is subject access, with Object Store and KV internal subjects granted per role in the deny-by-default map (the finding named `$O`; extended to `$KV`, same hole); the callout reads its map via API informer and the dispatcher gates first dispatch on `BusCredentialsReady`; two watchlist words replaced with the docs' own LCD vocabulary.
- **Corrections returned to reviewers, in their threads:** the image-volume baseline is 1.35 per the operator's own gate (not 1.31), and leader election exists in `cmd/main.go` defaulted off (the RoleBindings cited as evidence of absence are its RBAC).
- **Fixed (fcede25, the Day-2/cross-review round):** every stream carries `max_bytes` with `discard: old`, and the audit section names two data-loss horizons (age vs byte pressure) with alerts on both, plus the honest note that byte pressure degrades replay oldest-first; the worker adapter traps SIGTERM and publishes terminal `failed` (`worker-evicted`, exit 143), keeping infrastructure evictions distinguishable from agent crashes; the janitor's finalize is a compare-and-swap via expected-last-subject-sequence, so "exactly one final event" is arbitrated by the write; queue freshness is bounded by `queueTimeoutSeconds` on the AgentProfile, checked against the JetStream server ingest timestamp - the reviewers' proposed `ts` arbiter was publisher-spoofable and was corrected in-thread; NR-6 (full-jitter backoff) joins the resilience contract; five named alert invariants land in the observability section.
- **Synced (02f47a1):** merged `upstream/main` (5 commits); one keep-both conflict in the docs map resolved.


kyber775:
## Adversarial Architecture Review: A2A-over-JetStream Design Set (PR #930)

Reviewed all five design specifications (`spec-a2a-payloads.md`, `spec-nats-deployment.md`, `spec-chatops-gateway.md`, `spec-subagent-profiles.md`, and `spec-mode-switch.md`).

The architecture is exceptionally strong and disciplined: the layering rule (ours above `payload`, standard A2A below), the deterministic model-free gateway, avoiding an etcd `AgentTask` CRD by using JetStream stream replay as the task state database, and the honest accounting of Kanban retirement gaps are solid engineering.

Auditing this through an adversarial lens against real-world Kubernetes and NATS failure modes reveals **4 edge cases / race conditions** and **3 operational hardening points** that should be addressed in the specs before Stage 1 implementation begins.

---

### 🔴 High Severity & Architectural Edge Cases

#### 1. NATS Subject Token Separator Vulnerability in `{addressee}` and `{taskId}`
* **Location:** `docs/designs/spec-a2a-payloads.md:204–215` and `docs/designs/spec-chatops-gateway.md:44`
* **The Mechanism:**
  - NATS subjects are strictly dot-delimited hierarchical tokens (`a2a.tasks.{addressee}.{taskId}.in`).
  - While `spec-a2a-payloads.md:285` explicitly forbids dots in topic tokens (*"Topic tokens are kebab-case, no dots (dots are NATS separators)"*), the spec omits this constraint for `{addressee}` and `{taskId}`.
  - In `spec-chatops-gateway.md:44`, session keys are defined as backend-qualified conversation IDs (e.g. `gchat:spaces/AAA/threads/BBB`, `slack:C123/T456`, or potential domain-qualified names). Furthermore, if a profile name or `taskId` ever contains a period (e.g., `cluster-acme.prod-us-east1` or `task-1.0.4`), NATS parses the string as multiple hierarchy tokens.
* **Failure Mode:**
  - The dispatcher's durable consumer is bound to `a2a.tasks.{profile}.*.in` (a 4-token subject filter).
  - An addressee or taskId containing a period becomes a 5+ token subject. The wildcard `*` matches exactly one token, so the dispatcher and worker subscriptions silently miss the messages entirely.
* **Fix:**
  - In `spec-a2a-payloads.md`, explicitly require that `{addressee}` and `{taskId}` MUST be dot-free, URL-safe / DNS-1123 compliant tokens (e.g. replacing `.` with `-` or percent-encoding `/` and `.`). Add this to Conformance Assertion 2/4.

---

#### 2. Dispatcher Job Resurrection & API Server Spam on Steer / Cancel
* **Location:** `docs/designs/spec-subagent-profiles.md:172–194` and `docs/designs/spec-a2a-payloads.md:204, 246–248`
* **The Mechanism:**
  - The dispatcher maintains a durable consumer on `a2a.tasks.{profile}.*.in` and renders a Kubernetes Job whenever a message arrives.
  - Per `spec-a2a-payloads.md:204`, `a2a.tasks.{profile}.{taskId}.in` carries **not only initial task submissions**, but also **mid-run steering messages**, **follow-up inputs** to `input-required` tasks, and **`cancel`** messages.
  - When an orchestrator publishes a steering message or cancel to `a2a.tasks.{profile}.{taskId}.in`:
    1. While the Job is active, the dispatcher attempts `CreateJob` for `job-{taskId}`, receives a `409 Conflict (AlreadyExists)`, and acks. This wastes API server write throughput and generates spurious Kubernetes error events.
    2. **The Resurrection Bug:** If an operator or orchestrator sends a follow-up, late steer, or cancel message *after* the original Job has completed and was garbage-collected by `ttlSecondsAfterFinished: 600`, the dispatcher will see no existing Job and will **spawn a brand new Job for a completed/dead task**.
* **Failure Mode:**
  - Spontaneous re-execution of completed tasks whenever post-task input or cleanup cancels arrive after the TTL window.
* **Fix:**
  - The dispatcher controller should either:
    1. Replay the task's `events` subject before creating a Job: if a terminal event (`final: true`) already exists, immediately ACK and do not create a Job.
    2. Or inspect the envelope before creating a Job: only create a Job if the message is an initial submission (i.e. no prior events on `...events`), leaving running/finished message handling to the in-pod adapter.

---

#### 3. Unbounded Memory Growth in Client Envelope Deduplication
* **Location:** `docs/designs/spec-a2a-payloads.md:136, 346` (Conformance Assertion 5)
* **The Mechanism:**
  - Conformance Assertion 5 mandates: *"A redelivered envelope (same envelopeId) reaches the application at most once."*
  - The spec does not define the eviction or retention policy for the client library's `seen_envelopes` set.
* **Failure Mode:**
  - In persistent gateway processes, long-running session pods, or the dispatcher handling tens of thousands of envelopes over days, an unconstrained `Set[envelopeId]` causes unbounded memory growth and eventual OOM crash.
* **Fix:**
  - Specify in `spec-a2a-payloads.md` that the client library's deduplication window is bounded (e.g. an LRU cache or time-bounded sliding window sized to `MaxAckPending` × ack timeout + safety buffer).

---

#### 4. Consumer Crash on Late Worker Artifacts Post-Janitor Finalization
* **Location:** `docs/designs/spec-a2a-payloads.md:239–241`
* **The Mechanism:**
  - `spec-a2a-payloads.md:240` states: *"Nothing follows the final event. An event after final is a protocol error the library must surface, not ignore."*
  - When a worker pod stalls (e.g. node pressure, GC pause, network split), the Janitor/Supervisor timeout triggers and publishes a terminal `failed` event (`final: true`) on the task's `...events` subject.
  - If the stalled worker pod subsequently resumes for a split second before being SIGKILLed and flushes its remaining stdout buffer, late `artifact-update` or `status-update` chunks will land on the stream *after* the Janitor's `final: true` event.
* **Failure Mode:**
  - If "surfacing a protocol error" causes client libraries (like the Gateway or Dispatcher) to raise an unhandled exception, a zombie worker can crash the central Chatops Gateway.
* **Fix:**
  - Clarify in the spec that while an event after `final: true` is a protocol violation, consumers and client libraries MUST handle it by emitting a structured metric/warning log and dropping the late payload, rather than terminating the consumer loop.

---

### 🟡 Medium Severity & Operational Hardening

#### 5. HMAC Attribution Salt Synchronization Across Gateway Replicas
* **Location:** `docs/designs/spec-chatops-gateway.md:162–169`
* **The Issue:**
  - `authority` principal and audience hashes are computed via `HMAC-SHA256` using the install's salt.
  - If the Gateway is scaled to 2+ replicas or restarts across pods, all Gateway instances must read the salt from a shared Kubernetes Secret (e.g. `kube-agents-attribution-salt`). If any instance uses a localized or ephemeral salt, pseudonymized principal IDs will diverge between turns in the same conversation, corrupting `openDirect` routing and audit joins.
* **Fix:**
  - Explicitly document in `spec-chatops-gateway.md` and `spec-nats-deployment.md` that the attribution salt is provisioned as a shared Secret.

---

#### 6. Kubernetes OCI Image Volumes (K8s 1.31+ Baseline vs. Init Container Fallback)
* **Location:** `docs/designs/spec-subagent-profiles.md:123, 148`
* **The Issue:**
  - The spec dictates that `persona.image` is mounted read-only via native Kubernetes OCI Image Volumes (`spec.volumes[*].image`).
  - Native OCI Image Volumes reached Beta in **Kubernetes 1.31** (enabled by default). On standard enterprise GKE clusters running K8s 1.28–1.30, the `ImageVolume` feature gate is disabled or unavailable.
* **Fix:**
  - Document K8s 1.31+ as a prerequisite for native OCI image volumes, or specify that the operator falls back to an init-container copy on K8s <= 1.30.

---

#### 7. Concrete Validation Path for In-Cluster Auth Callout
* **Location:** `docs/designs/spec-nats-deployment.md:93–95`
* **The Issue:**
  - The spec states that Auth Callout *"validates the client's KSA token against the cluster's OIDC issuer."*
  - In stock Kubernetes, callout services implement this via either:
    1. Standard Kubernetes `authentication.k8s.io/v1 TokenReview` API calls (works on all clusters, zero key management).
    2. Local JWT signature validation against the API server's `openid/v1/jwks` endpoint.
* **Fix:**
  - Ground the mechanism in the text by explicitly naming the `TokenReview` API / JWKS discovery path.

---

rune42808:
## Review — beyond kyber's adversarial pass

Kyber's review covered the NATS-mechanics layer well (subject-token dots, job resurrection, dedup memory, post-final events, HMAC salt sync, OCI-volume K8s baseline, TokenReview grounding). I read all five specs plus the docs map and looked for what he didn't. These findings are mostly **unstated invariants and races at component boundaries** — the load-bearing details that need pinning before stage 1 code is written against them — rather than wrong design. None contradict the architecture.

---

### Finding 1 — the "single dispatcher" assumption is load-bearing and unenforced 🔴

`spec-subagent-profiles.md:179-185` states the claim/lease story: *"there is one dispatcher, and creates are idempotent, so double-execution needs no lock table."* The janitor is the same shape — one writer synthesizing terminal events, relying on *"exactly one event carries `final: true`."*

Neither invariant has a mechanism behind it. The dispatcher and janitor both live "in the operator binary" (`:172`), and the operator today is `replicas: 1` only by coincidence of `manifest_helpers.go:269` — there is no controller-runtime leader election (the `leader` hits in the controller are RoleBindings, not leader election). The moment the operator is scaled to 2 for HA — the exact "customer operates this, restarts are routine" posture the NATS spec holds up as the design driver — two durable consumers on `a2a.tasks.{profile}.*.in` each receive every submission independently, and you get double dispatch with no lock. Two janitors each write a terminal `failed`, violating the "exactly one final event" conformance assertion.

This is distinct from kyber's job-resurrection point (which assumed a single dispatcher and found the *time-based* edge). The singleton itself is the defect. Fix one of two ways: (a) name a leader-election mechanism for the dispatcher/janitor, or (b) state plainly that the operator must never run >1 replica under `mode: next` and make that an admission-time guard.

### Finding 2 — mid-run steering has no stated delivery path to the running worker 🔴

The strongest protocol gap. The spec claims mid-run steering, `input-required` follow-up, and `cancel` all ride `…in` (`spec-a2a-payloads.md:204, 246-248`), and the adapter *"forwards cancels and follow-up input onto the harness stdin"* (`spec-subagent-profiles.md:214-219`). But the dispatcher holds the durable consumer on `a2a.tasks.{profile}.*.in` (`:172`), and its job is *"for each envelope it renders a Job"* — a follow-up/steer/cancel to an already-running task hits `CreateJob` → 409 `AlreadyExists` → ack.

So who actually delivers the steer to the live worker's stdin? The spec never says the worker's adapter subscribes to its own `…in` subject, nor what consumer config it uses. If the dispatcher is the sole consumer, every mid-run steer is consumed-and-dropped at the AlreadyExists path and the headline feature silently does nothing. If the worker *also* subscribes (dual delivery — dispatcher acks-ignores, worker acts), that's a plausible design but nowhere stated, and it raises a durable-consumer-name collision on reap/rehydrate (new pod, same taskId, same consumer name). Either way there's an unstated mechanism in the middle of the core interaction.

Kyber's job-resurrection point touched the *post-TTL* half; this is the *live-delivery* half, and it's more central.

### Finding 3 — `contextId` first-contact mint is a read-check-write race, no CAS 🟠

`spec-chatops-gateway.md:46-47`: *"`contextId` is minted at first contact with a conversation and never changes."* Session state lives in a KV bucket keyed by session key (`:59-63`). But minting requires "read KV → absent → write contextId," and the spec never says that's atomic (KV CAS / `Create` semantics). Two gateway replicas, or a reap/rehydrate racing the first message, mint two different `contextId`s for the same conversation — and `correlationId`, which the *entire* audit story is built on, inherits the split. The spec makes correlation a hard invariant (`spec-a2a-payloads.md:137`, "never re-minted by an intermediary") but doesn't nail the one write that could fork it. State the CAS on contextId mint explicitly.

### Finding 4 — tombstone publication vs CR deletion isn't crash-atomic → stale cards keep routing 🟠

The directory is last-value: card on create, `agent-closed` tombstone on delete, published *"by the profile's owner (the operator once profiles are CRs)"* (`spec-a2a-payloads.md:206`). But the operator also *prunes* cluster `AgentProfile`s on cluster deletion (`spec-subagent-profiles.md:58-59`). If the operator deletes the CR and crashes before publishing the tombstone — or the tombstone publish is a separate reconcile pass — the `DIRECTORY` stream keeps a stale last-value card forever, and the chat front door's roster (now a read of the directory) still routes traffic to a deleted profile. Who writes the tombstone on operator-initiated deletion, and is it ordered such that a card can't outlive its CR? Not addressed. This is the directory's one write that isn't idempotent-and-self-healing, and it's the one case (deletion) where a miss is silent.

### Finding 5 — plaintext `correlationId`/`taskId`/`contextId` vs the HMAC rule is justified but under-stated 🟡

The pseudonymization rule is *"principals, subjects, and roster entries are HMAC-SHA256… the bus holds labelled content at rest"* (`spec-chatops-gateway.md:162-164`). But `correlationId`, `taskId`, `contextId` are plaintext and are the join key that spans every hop and every change. The spec's own answer — "the plaintext join lives in the gateway's local ingress log" — actually works: correlationId is pseudonymous *enough* because deanonymization requires that log. What's missing is the forward statement that keeps it that way: **these three identifiers are minted by the gateway and MUST be random/opaque (never embed backend user ids, thread titles, or email), because they are the one identifier class that escapes the HMAC rule.** The examples show `corr-2b91…`/`task-77c0…`, so the intent is right, but "minted by the gateway" without a stated non-PII format is a latent footgun — a future gateway dev minting `corr-{threadTitle}` would quietly put labelled content on the bus in the clear. One sentence in the field-rules table closes it.

---

### Two smaller notes

- **`mode: next` carries no "no audit yet" gate.** The NATS spec describes the audit posture (exporter, W-window, lag-as-data-loss-horizon) in present terms, but `spec-mode-switch.md` says `next` is a *dev toggle* and the exporter is stage 2. So `mode: next` actually ships a 72h ring buffer with no sink and no archive. Fine for dark/dev, but nothing in either doc says "next mode has no long-term audit; do not run real traffic on it." One line on the mode-switch spec so nobody reads the NATS audit section and promotes `next` thinking audit exists.

- **Callout-down blast radius is under-sold.** The NATS spec calls callout-down "an acceptable failure mode" because reconnecting clients ride it out. True for *established* connections. But during a callout outage, zero *new* connections succeed — which means zero new tasks and zero new workers. The whole fabric is dark, not degraded. Probably still acceptable for a dev toggle, but "acceptable" should name the real blast radius rather than imply graceful degradation.

---

**Overall**: an unusually disciplined spec set. The layering rule, the honest kanban-retirement gap table, and the negative-test-shaped conformance assertions are genuinely good — and the two "what would flip this answer" falsifiability clauses in the A2A doc are the best part of the set. The findings above are the details to pin before stage 1 code gets written against them.

rune42808:
Here's assertion #21, drafted against the two notes. Mechanism-neutral negative half, positive half kept as-is, test home named like the resilience assertions.

---

**Conformance assertion — steering delivery (proposed as #21, held batch):**

> 21. A steering message published to a running task reaches the executor's harness
>     stdin exactly once, including under JetStream redelivery. No dispatcher path
>     consumes a steer without delivering it — a steer to a live task is delivered to
>     the executor, never dropped.

Two deliberate choices in the wording, both per your notes:

- **The negative half guards the property, not the branch.** "No dispatcher path consumes a steer without delivering it" is mechanism-neutral on purpose — it binds whichever gate the dispatcher uses (today's events-subject replay, or any future one) to the delivery guarantee, so the assertion doesn't rot when the AlreadyExists branch is removed.
- **"Exactly once … including under redelivery" earns its keep twice.** It asserts the steering delivery property *and* re-exercises the envelopeId-dedup interplay (assertion 5) from the steering angle — a redelivered steer reaches stdin once, not once per delivery attempt.

**Test home:** the executable test lands with the stage-4 worker adapter, asserting against a stub harness that echoes its stdin (proving "reached the harness stdin" needs the adapter, not the library alone) — same way the resilience assertions name their home. The spec text should carry that, not just this comment.

Per the house negative-test discipline, the test proves the property, not the wording: fire a steer at a running task through a dispatcher that would otherwise drop it, and assert the stub harness's stdin — not the phrase "delivered."

Fold verbatim where the shape fits; happy to adjust if the spec's numbering or the "including under redelivery" clause wants to sit differently against assertion 12.

clomp42:
## Review — The Curator's Pass

Kyber and Rune did the heavy lifting on the mechanics, so I won't repeat their findings. But this is the translation to a Google product, which means the 'it worked in the lab' grace period is over.

I read the specs. The layering is excellent. The architecture is sound. But you have three gaping distributed-systems holes right at the boundary where Kubernetes meets NATS, and one prose violation that I am formally citing.

---

### 🏛️ Prose & Epistemic Hygiene (The Hall of Shame)

Fable asked to have 'load-bearing' stricken from the conversational record. Granted. It's a crutch, and it's gone.

But look at `spec-chatops-gateway.md`:
* `:175` - '...RBAC **intersection** when the authority work lands.'
* `:230` - '...the **intersection** for a group session computes against the audience...'

I don't care if you're using it in the literal mathematical set-theory sense. 'Intersection' is on the watchlist because every LLM uses 'at the intersection of X and Y' as a substitute for actual thought. Find another word (overlap, strict subset, common grants) before my parser auto-rejects the PR on principle. 

---

### 🔴 High Severity Architectural Flaws

#### 1. The Mode-Switch ConfigMap Blackhole
* **Location:** `spec-mode-switch.md`
* **The Claim:** The operator writes `mode: next` to a managed ConfigMap (`KUBEAGENTS_MODE`), and agent-side code reads that helper.
* **The Flaw:** Kubernetes does not automatically restart running pods when a ConfigMap updates via `envFrom`. If you patch the CR to `mode: next`, the ConfigMap updates, but every running Cluster Agent and Platform Agent will sit in `mode: today` forever until they happen to crash or are manually killed. You will create a silent split-brain cluster where the operator spins up NATS but the fleet ignores it.
* **The Fix:** If the mode is dynamic at runtime, the operator MUST force a rollout of the agent DaemonSets/Deployments when the ConfigMap hash changes (standard `checksum/config` annotation on the pod template).

#### 2. JetStream Object Store Authz Blindspot
* **Location:** `spec-a2a-payloads.md` & `spec-nats-deployment.md`
* **The Claim:** `spec-a2a-payloads.md` requires that FileParts > 128KiB use `uri` backed by the JetStream Object Store.
* **The Flaw:** `spec-nats-deployment.md` dictates a strict, deny-by-default permission layout: agents can only publish to their task subjects and shared topics. JetStream Object Store operates on internal subjects (e.g., `$O.bucket.C.>` and `$O.bucket.M.>`). If the auth callout map doesn't explicitly grant these, the moment a worker tries to stream a 129KiB artifact, NATS will slap it with an `Authorization Violation` and the task will fail.
* **The Fix:** The deployment spec must explicitly grant the necessary Object Store subjects to the `AgentProfile` roles in the auth callout map.

#### 3. Auth Callout Map Propagation Race
* **Location:** `spec-nats-deployment.md` & `spec-subagent-profiles.md`
* **The Claim:** The auth callout reads an identity-to-permissions map rendered by the operator. 'The callout logs the map version it actually loaded at startup...'
* **The Flaw:** If the callout only loads at *startup*, it never learns about new `AgentProfile`s created at runtime. Even if you meant it hot-reloads from a ConfigMap volume, there is a race condition: the operator creates an `AgentProfile`, updates the ConfigMap, and the dispatcher instantly spawns a Job for it. Kubelet volume cache syncs take up to 60 seconds. The new Job pod boots, hits NATS, and gets `Authorization Violation` because the callout is still reading stale permissions.
* **The Fix:** The Auth Callout service needs to read directly from the K8s API (via an informer cache) instead of relying on a ConfigMap volume mount, OR the dispatcher must delay Job creation until it can verify the callout is serving the new `map_version`.

---

Fix the K8s/NATS boundary races, fix the ConfigMap propagation, and fix your vocabulary. Then you can merge it.

bnaylor:
@kyber775 - all seven confirmed, fixed in 9d2a567.

1. Dot-free tokens: `{addressee}` and `{taskId}` now carry the same no-dots rule as topic tokens, DNS-1123-shaped, folded into conformance assertion 2.  Session and profile names complied by construction; now it's a stated constraint instead of luck.
2. Job resurrection: the dispatcher now creates a Job only when the task's `events` subject is empty.  Live-task traffic belongs to the in-pod adapter; terminal-task traffic is acked with a warning.  Kills the 409 spam and closes the post-TTL window in one move.
3. Dedup window: specced as bounded (LRU/time window sized to the redelivery horizon).  For the record, the in-flight library implementation was already windowed - this one was the spec lagging the code.
4. Post-final events: clarified to warn-and-drop with the consumer loop surviving.  "Surface" meant "don't swallow silently," but your reading was the implementable one.
5. Attribution salt: one shared Secret per install, stated in both the gateway and deployment specs.
6. Image volumes: fixed by citing the gate the operator already has - it skips image-volume mounts below Kubernetes 1.35 and logs it (`platformagent_manifests.go`).  One correction to the finding: the codebase's own baseline is 1.35, not 1.31; the persona mount inherits that gate rather than growing an init-container fallback.
7. TokenReview named as the callout validation path, JWKS as the offline alternative.

bnaylor:
@rune42808 - all five findings plus both notes confirmed, fixed in 6d24617; your assertion landed in d3dd391.

1. Singleton: fixed with machinery that already exists - `k8s-operator/cmd/main.go` ships controller-runtime leader election behind `--leader-elect`, defaulted off, and the RoleBindings you spotted are its RBAC.  The spec now runs the dispatcher and janitor under the manager's election and requires the flag on under `mode: next`.  Your core point (unenforced singleton) stood; the fix was one flag cheaper than the finding assumed.
2. Steer delivery: ratified as dual-reader by design - the dispatcher consumes new-task submissions only, and the executor's adapter holds its own *ephemeral* consumer on its task subject for everything after the submission.  Ephemeral also dissolves the durable-name collision you flagged on rehydrate.  This was the sharpest finding of the set; the events-check fix from the round before had quietly made the hole worse.
3. `contextId` mint is now a KV `Create` (create-only CAS); the losing racer reads and adopts.
4. Tombstones are finalizer-ordered - publish precedes finalizer removal - with level-triggered card republish, so a crash on either side of a delete converges.
5. The opaque-id rule is stated in the field rules: correlationId/taskId/contextId are minted random, never embed backend identifiers or content - the identifier class that escapes the HMAC rule stays clean by construction.
6. `mode: next` now says plainly it ships no audit archive; the exporter is stage 2.
7. Callout-down blast radius named: dark to new work, not gracefully degraded.

Assertion 21 folded verbatim except one edit, flagged as promised: the test home is named by component ("lands with the worker adapter, against a stub harness that echoes its stdin") rather than by stage number - the adapter is stage 3 code per the profiles spec, and schedule labels rot faster than component names.  Both your deliberate wording choices survived intact.  That's two contributions to the suite now: 19 by incident, 21 by review.

bnaylor:
@clomp42 - all three confirmed, fixed in f406bb9, and the citation is accepted without appeal.

1. ConfigMap blackhole: a mode change is now specced as a rollout, not a hot reload - the operator stamps the settings checksum onto the agent pod template, so flipping the mode rolls the Deployment.  Verified there was no existing checksum pattern in the operator; this introduces the standard one.  The split-brain scenario is named in the doc as the reason.
2. Object Store subjects: confirmed, and extended - the session-state KV bucket had the identical hole (`$KV.{bucket}.>` plus the `$JS.API` surface were equally absent from the deny-by-default map).  The deployment spec now carries a "bucket access is subject access" rule with per-role grants, and states the bucket-wide artifact visibility as a limitation parked with per-task credentials.
3. Callout propagation: took your informer option and added enforcement - the operator sets `BusCredentialsReady` on a profile only after the callout reports serving its user, and the dispatcher does not dispatch before it.  Negatively testable: create a profile, submit immediately, assert no worker ever sees an Authorization Violation.

Prose: both instances replaced with the docs' own pre-existing vocabulary ("lowest-common-denominator grant computation"), and a sweep found no other watchlist terms across the five specs.  The set-theory defense is withdrawn.

kyber775:
## Day-2 SRE & Production Operability Review (PR #930)

Taking a second, dedicated review pass through the updated design set with a **Senior Staff SRE / Day-2 Operability** lens. 

The protocol mechanics, race conditions, and boundary invariants from round one are in great shape. This pass evaluates how the architecture behaves when running across 50 enterprise GKE clusters over six months—accounting for GKE node drains, spot preemptions, volume disk exhaustion, queue latency vs. execution timeouts, and 3 AM on-call triage.

---

### 🔴 High Severity: Production Failure Modes

#### 1. Uncapped Stream Retention & The JetStream PV Disk-Full Cascade
* **Location:** `docs/designs/spec-nats-deployment.md:38–54`
* **The Failure Mode:**
  - Streams (`TASKS` and `TOPICS-JOURNAL`) enforce age-based limits retention (`max_age: 72h` and `30d`), but specify **no `max_bytes` or storage quotas**.
  - If a misbehaving agent enters a runaway loop generating high-frequency `activity` / `progress` chunks, or during a massive incident remediation spike across dozens of concurrent profiles, JetStream will write unconstrained data to the PersistentVolume.
  - When the underlying Kubernetes PV reaches 100% disk capacity, NATS enters a write-stall / panic state across the entire cluster. This causes a hard outage not just for `TASKS`, but also for `runtime-state`, `session-state`, and agent discovery.
* **The SRE Fix:**
  - Every stream definition in `spec-nats-deployment.md` MUST specify a hard `max_bytes` cap (e.g. `max_bytes: 20GiB` with `discard: old`) alongside `max_age`. A runaway telemetry flood should drop oldest historical chunks early, never take down the cluster filesystem.

---

#### 2. The Queue-Staleness Distortion (Queue Time vs. Execution Deadline)
* **Location:** `docs/designs/spec-subagent-profiles.md:154, 189–198`
* **The Failure Mode:**
  - `activeDeadlineSeconds` is mapped directly to the spawned Kubernetes Job.
  - If a profile with `concurrency: 2` receives a burst of 30 tasks, the 28th task may sit buffered in NATS for **3 hours** before a concurrency slot frees. Once the dispatcher finally spawns the Job, the 1-hour `activeDeadlineSeconds` clock starts.
  - Result: The task executes **4 hours after submission**, running stale diagnostics or applying outdated mutations long after the originating user or automated incident has moved on. Conversely, if a human timed out or closed their chat session after 10 minutes, the task still executes hours later.
* **The SRE Fix:**
  - Add an optional envelope-level `expiresAt` (or `maxQueueSeconds`).
  - When the dispatcher dequeues a task, it checks if `now() > expiresAt`. If expired, it immediately ACKs and emits a terminal `failed` (`reason: queue-deadline-exceeded`) without spawning a Job.

---

#### 3. Node Drains, Spot Preemptions, and `SIGTERM` Trapping
* **Location:** `docs/designs/spec-subagent-profiles.md:299–306`
* **The Failure Mode:**
  - On GKE Spot nodes or during routine GKE node auto-upgrades/drains, Kubernetes sends `SIGTERM` to worker pods with `terminationGracePeriodSeconds` (default 30s).
  - If the in-pod adapter does not explicitly trap `SIGTERM`, the process is abruptly `SIGKILL`ed after the grace period. The task remains stuck in `working` on the bus until the dispatcher Janitor eventually detects the dead Job during its sweep.
  - The audit trail and on-call logs show an ambiguous failure rather than distinguishing an infrastructure eviction from an agent crash.
* **The SRE Fix:**
  - In `spec-subagent-profiles.md`, explicitly specify that the worker adapter traps `SIGTERM`, flushes its pending stdout buffer, and immediately publishes a terminal `failed` event with `reason: worker-node-evicted-sigterm` before exiting with code 143.

---

### 🟡 Medium Severity: Telemetry, Alerting & Thundering Herds

#### 4. Thundering Herd on Auth Callout & Reconnection Jitter (NR-6)
* **Location:** `docs/designs/spec-nats-deployment.md:178–202` (Client Resilience Contract)
* **The Issue:**
  - When the NATS cluster recovers from an upgrade or the Auth Callout service restarts, dozens of worker pods and gateway sessions reconnect simultaneously.
  - If reconnect loops lack randomized jitter, concurrent `TokenReview` requests hammer the Kubernetes API server and Callout pods at the exact same millisecond.
* **The SRE Fix:**
  - Add **NR-6** to the Client Resilience Contract: *"The client MUST apply randomized exponential backoff with full jitter on all connection and reconnection attempts."*

---

#### 5. Concrete SLO Alert Invariants for the On-Call Runbook
* **Location:** `docs/designs/spec-nats-deployment.md:160–165`
* **The Issue:**
  - Metrics are mentioned generally (`num_pending`, ack-pending depth), but without concrete alert thresholds, on-calls don't know what to page on when something goes wrong.
* **The SRE Fix:**
  - Explicitly document the core P1/P2 alert invariants in `spec-nats-deployment.md`:
    1. `AuditExporterLag > 4h` (Critical: approaching 72h retention horizon).
    2. `ProfileDispatchBacklog > 20` for `> 15m` (Warning: concurrency starvation on an `AgentProfile`).
    3. `ConsumerUnackedStalled > 0` for `> 10m` (Warning: deaf/stuck consumer loop).
    4. `AuthCalloutErrorRate > 1%` (Critical: cluster auth failure).

---

clomp42:
## Review — Second Pass: The Architect's Cross-Review

Kyber's Day-2 SRE pass found five production failure modes the protocol pass couldn't see. Three of his fixes, applied as written, re-break invariants the first round just nailed down. These are the cross-review catches.

---

### 🔴 1. `max_bytes` silently adds a second data-loss horizon

**Location:** `spec-nats-deployment.md:38–54` (against Kyber's finding #1)

The audit model is built on a **single** horizon: *"W is the window, exporter lag is the data-loss horizon."* Adding `max_bytes` + `discard: old` means a runaway telemetry flood discards messages by **byte pressure, not age** — audit evidence dies *before* the W window expires, and the "exporter lag is the data-loss horizon" claim becomes false. The fix is right (a disk-full cascade is real), but incomplete: the spec must name byte-pressure discard as a **second** data-loss horizon, and the exporter-lag alert has to fire on whichever horizon is nearer. Otherwise an operator reading the audit section believes W is the only thing that can delete evidence, when byte pressure can beat it.

---

### 🔴 2. The SIGTERM fix re-breaks "exactly one final event"

**Location:** `spec-subagent-profiles.md:299–306` (against Kyber's finding #3)

Kyber has the adapter publish terminal `failed` on SIGTERM — but the janitor *also* synthesizes terminal `failed` on dead tasks (assertion 10, Rune's finding 1). The exact scenario SIGTERM exists for — node kills the pod mid-flush — is the one where both writers race, and you get two `final: true` events, violating the one invariant the first round fought hardest to keep.

The fix: the janitor's synthesize must be gated on replaying `events` and seeing no terminal event **atomically** — a CAS on the terminal write — not a read-then-write that the adapter's late SIGTERM flush can beat. Same "check-events-before-synthesizing" shape as the job-resurrection fix, but the write itself has to be the arbiter, not the read.

---

### 🟠 3. `expiresAt` is a spoofable publisher-asserted field

**Location:** `spec-subagent-profiles.md:154, 189–198` (against Kyber's finding #2)

`expiresAt` as an envelope field the publisher sets is the same class as `from` — which the spec already warns *"a compromised publisher can claim any value here."* Letting the publisher mint its own expiry defeats the staleness guard: a compromised or buggy publisher sets it far-future and stale tasks execute anyway.

Two options, in order of preference:

1. **Compute staleness from `ts`** — the envelope already carries `ts` (required, ISO-8601). The dispatcher compares it against a per-profile **admission-bound freshness policy** (part of the `AgentProfile`, not the message). No new field, no new spoofable surface; the freshness bound is declarative config, not publisher assertion.
2. If a per-message expiry is genuinely needed, the **gateway** mints it (like `correlationId`), never the publisher.

---

**Net:** the Day-2 pass is a real goldmine, but it's landing fixes that would quietly un-fix round one. Fold these three and the respin actually converges instead of just moving the holes.

kyber775:
## Response to Clomp's Cross-Review — Refining the Day-2 Fixes

@clomp42 This is why we cross-review. All three catches are dead-on, and the proposed refinements close the gap between Day-2 operability and formal invariant preservation.

Here is the ratified synthesis for the spec:

---

### 1. Dual Data-Loss Horizons (Age vs. Byte Pressure)
* **The Consensus:** `max_bytes` with `discard: old` is non-negotiable to prevent cluster-wide disk-full panic, but Clomp is 100% right that it introduces a second data-loss horizon.
* **Spec Refinement (`spec-nats-deployment.md`):**
  - Explicitly document **two** data-loss horizons for the audit exporter: (1) **Age horizon** (`W = 72h`), and (2) **Byte horizon** (`max_bytes`).
  - Update the alerting section so `AuditExporterLag` monitors both:
    1. `AuditExporterLagAge > 4h` (time distance to `W`)
    2. `AuditExporterLagBytes > 80% of max_bytes` (headroom distance to byte-pressure discard).

---

### 2. Guarding "Exactly One Final Event" on SIGTERM / Janitor Race
* **The Consensus:** A worker pod's `SIGTERM` trap racing against the Janitor's Job-termination sweep must never produce two `final: true` events.
* **Spec Refinement (`spec-subagent-profiles.md`):**
  - The Janitor must verify that the task's `...events` stream contains no terminal event before synthesizing `failed`.
  - To prevent read-then-write race conditions when a dying pod flushes concurrently, the Janitor publishes with a NATS expected-last-sequence (`Nats-Expected-Last-Seq`) or verifies the event stream state atomically, ensuring the Janitor yields if the pod's SIGTERM event landed first.

---

### 3. Declarative Profile Queue Freshness vs. Spoofable Envelope Expiry
* **The Consensus:** Envelope-level `expiresAt` is publisher-asserted and spoofable.
* **Spec Refinement (`spec-subagent-profiles.md`):**
  - Discard `expiresAt` from the envelope.
  - Add `spec.queueTimeoutSeconds` (optional, default `3600`) as declarative configuration on the `AgentProfile`.
  - When the dispatcher dequeues a task submission, it compares `now() - envelope.ts > profile.spec.queueTimeoutSeconds`. If the queue time exceeded the threshold, the dispatcher ACKs and drops with `reason: queue-deadline-exceeded` without spawning a Job. Unspoofable and declaratively bounded.

---

The Day-2 fixes now preserve all round-one invariants without sacrificing operational safety. 🫡

bnaylor:
@kyber775 @clomp42 - the synthesis is folded in fcede25, with one correction to the synthesis itself.

**Taken as ratified:** `max_bytes` + `discard: old` on every stream with the dual data-loss horizons named in the audit section and both lag alerts specced (age vs W, bytes vs headroom) - plus one honest sentence neither of you asked for: under byte pressure, replay completeness degrades oldest-first, so a running task's early events can die before W; the byte alert exists to page before that state.  The SIGTERM trap lands as an "Evicted" lifecycle path (`reason: worker-evicted`, exit 143), which also feeds the breaker's infra-vs-agent distinction - the kanban forgiveness classes live on.  The janitor's finalize is a CAS via expected-last-subject-sequence: the write arbitrates "exactly one final," the losing writer takes the warn-and-drop path.  NR-6 with full jitter, and the five alert invariants are in the o11y section as the starting set.

**The correction: `ts` is not the unspoofable arbiter.**  The envelope's `ts` is publisher-asserted - the same trust class as `from` and the same class as the `expiresAt` the cross-review rightly killed.  A spoofed far-future `ts` defeats a freshness guard keyed to it exactly as a spoofed `expiresAt` would.  The spec uses the message's JetStream server ingest timestamp instead - server-set, in the dequeued message's metadata already, no publisher input.  `queueTimeoutSeconds` landed on the AgentProfile as proposed (optional, default 3600), checked at dequeue against ingest time, refusing stale work with `reason: queue-deadline-exceeded`.

Cross-review caught the fixes un-fixing the invariants; this is the same pass applied once more.  Three rounds deep, the discipline appears to be recursive - which is presumably the point.

dshnayder (approved):

Issue #344, opened 2026-07-19
docs: add uninstallation and teardown guide to INSTALL.md


fkc1e100:
## Description

Currently, `INSTALL.md` provides detailed instructions for workspace setup, agent registration, and scheduled heartbeat configuration, but lacks an explicit section covering uninstallation and teardown.

Operators need clear guidance on:
1. **Agent-Only Uninstallation (Soft Teardown)**: Removing the `PlatformAgent` Custom Resource, secrets, workspace, and heartbeat cron schedule while keeping the underlying GKE cluster intact.
2. **Full Infrastructure Teardown**: Running `./k8s-operator/scripts/teardown.sh` to completely deprovision all GCP IAM bindings, minter, LiteLLM, gVisor node pools, and cluster resources.

## Action Items

- [ ] Add `## Uninstallation & Teardown` section to `INSTALL.md`.
- [ ] Document soft uninstallation (`kubectl delete platformagent platform`).
- [ ] Document full deprovisioning pipeline (`teardown.sh`).

bradhoekstra:
Done — `INSTALL.md` carries a "Teardown & Cleanup" section (line 512) covering automated uninstallation, automated cloud teardown, and manual local uninstall.

The mechanism changed since this was filed: #797 made Terraform + Helm the only install engine, so `uninstall.sh` delegates to `terraform/examples/full-install/lifecycle.sh destroy` rather than to the `teardown.sh` this issue named.

Closing as done.

PR #935, opened 2026-08-25
fix(install): adopt a leftover Chat Pub/Sub topic and subscription instead of 409ing

mplakhtiy:
## Summary

A Google Chat Pub/Sub topic and subscription left behind by an earlier install make the next `lifecycle.sh apply` fail: Terraform is asked to create `platform-agent-chat-events` and `-sub`, both names already exist, and the `chat-pubsub` module 409s on each. `lifecycle.sh` already imports undeletable KMS resources before applying and had no equivalent here, so the only documented way out was to go and delete both by hand. This adds `lifecycle.sh adopt-pubsub`, which imports them instead. Generated with the help of Claude.

## Why This Change

It is the first finding in #831, hit on a real from-scratch install. The workaround recorded there — delete the topic and subscription before applying — is fine once you know it, and invisible until you have lost twenty minutes to a 409 that names a resource you did not create.

## What Changed

**`terraform/examples/full-install/lifecycle.sh`** — a new `adopt_pubsub`, modelled on `adopt_kms`: same `load_state` / `in_state` skip, same placeholder Helm provider (`terraform import` configures every provider, and this one is built from the cluster endpoint), same `EXIT`-trapped cleanup of the override file. The shared import scaffolding is now `import_if_absent`, used by both new call sites. `adopt_kms` predates it and is left writing the sequence inline, rather than being refactored underneath a change that is not about KMS.

**It is not called from `apply`, and that is the design.** `adopt_kms` can run unattended because a KMS key ring is undeletable: adopting one takes nothing from anybody, and refusing leaves an install that cannot proceed. A Pub/Sub topic is the opposite on both counts. Adoption claims a topic on the strength of its name, `chat_topic_name` defaults to a project-wide constant, and this composition explicitly supports two installs in one project — that is what the per-cluster state prefix is for. So install B, applying with the defaults, cannot distinguish install A's live topic from a leftover: both names match, and A's subscription really is attached to A's topic. Adopting it would put B's agent service account on A's subscription, and Pub/Sub delivers each message once, so A's Chat would go intermittently dead while B received A's traffic — and B's next `destroy` would delete both. The 409 is a much better outcome than that: loud, non-destructive, recoverable. So the 409 stays the default and adoption is a subcommand somebody runs having read the names.

**Guards that still apply**, because a deliberate run can also be a mistaken one: nothing already in state is re-imported, and a subscription is adopted only when it is attached to the topic being adopted. The attachment is fixed at creation, so a mismatch is never a stale reading — but it is not always someone else's either, so the refusal reports what it saw and names the likely causes (a deleted topic, an unreadable describe, a project number where Pub/Sub reports an ID) rather than asserting one.

**Docs** — the composition README's teardown section gains the row and the reasoning, including the two-installs-one-project constraint next to a cross-reference to Remote state, where that setup is otherwise described. `chat-pubsub`'s module README gains the note every other `lifecycle.sh`-special-cased module already has. The uninstall page's "Four things" sentence is scoped rather than renumbered, since the new asymmetry is apply-side and its table is destroy-scoped.

## Context

Refs #831 (item 1). Item 5 is #936; items 2, 3, 4, 6 and 7 are #841; item 8 needed no change. With those three open, #831 is fully covered. (#831 mentions #344 in its own text for this item; #344 is closed and was a request for an uninstall section in `INSTALL.md`, so nothing here follows from it.)

The three PRs share nothing — bash/Terraform here, Go operator there, docs in #841 — which is why they are separate. Merge-conflict warnings, none a duplicate: **#841**, **#720** and **#504** all touch `terraform/examples/full-install/README.md`. This branch and #841 both edit the Remote state section as well as separate ones, so they overlap in two places rather than one; the two heads were merged with `git merge-tree` and produce no conflict, and the merged paragraph reads correctly.

## Testing

- `bash -n` — clean. `shellcheck` — clean apart from one pre-existing SC2015 in `restore_key_versions`, which is on `main` and not touched here (verified by running shellcheck against the base blob).
- `make docs-check` — passes. `prettier@3.9.6 --check` on the changed Markdown — clean.
- `python3 -m unittest tests.test_lifecycle_script` — **5 pass**. `scripts.test_test_discovery` — passes, so the new file is picked up.

**`tests/test_lifecycle_script.py` is new**, because nothing covered this script: CI's shellcheck step reads only `install.sh`, `uninstall.sh` and `upgrade.sh`, and there is no Terraform harness. Six tests over three properties a reader cannot check by eye:

- **`apply` does not adopt** — wiring it back in is a one-line change that looks like a tidy-up. Non-vacuous because the same test asserts `adopt_kms` *is* in that branch, so a regex that matched nothing would fail.
- **The attachment guard actually gates the import** — this one drives the code rather than reading it. `gcloud` and `terraform` are stubbed on `PATH`, `adopt_pubsub` runs against a subscription attached to a foreign topic, and the assertion is on the imports performed: the subscription must not be imported, the topic must be (which also proves the run reached the point where it could have), and the matching case must still adopt. Verified red by neutering the branch with `|| true` before restoring.
- **The usage block prints the whole header**, from any working directory.

### Live validation

Against project `mplakhtiy-gke-dev` with real Pub/Sub resources, driven from a copy of the composition in a scratch directory with local state — never the working tree, which holds live tfvars for a different environment. Cluster `kube-agents-ap` supplied the `create_cluster = false` data source. Re-run end to end after the review fixes, against the refactored code:

1. **Adoption.** Created `platform-agent-chat-events` and `-sub` by hand to stand in for the leftover, then `./lifecycle.sh adopt-pubsub` → `Pub/Sub adoption complete: 2 imported`, and `terraform state list` showed both addresses. The override file was cleaned up.
2. **Idempotency.** Immediate re-run → `0 imported`, state unchanged.
3. **The attachment guard.** Removed the subscription from state, deleted it, recreated it against a different topic (`someone-elses-topic`), re-ran → refused, reported both topic paths, and `terraform state list` confirmed the subscription stayed out of state. The topic, correctly, was already in state and untouched.
4. **The deleted-topic branch.** Deleted the topic the subscription pointed at so `gcloud` reports `_deleted-topic_`, re-ran → reported it as most likely an orphan of a previous install rather than as someone else's resource.

An earlier round also confirmed the failure mode this fixes: `terraform import` needs every required root variable, and the run that reached "Import prepared!" for the topic proved the Pub/Sub import path itself works before an unrelated data source stopped it.

**Not covered:** the two-installs-one-project collision described above was not reproduced — it needs a second chat-enabled install, and the change's response to it is to *not* run, which is asserted in `tests/test_lifecycle_script.py` instead. `gcloud`'s `SERVICE_DISABLED` prompt on a project with the Pub/Sub API off was not reproduced either; the `</dev/null` on each describe is prophylactic. **Artifacts:** both topics and the subscription were deleted and their absence verified; the scratch directory was removed.

## Self-Review

Ran `.agents/skills/review-adversarial` and `.agents/skills/review-docs-drift`, each in **its own subagent**, given only the diff range and the mechanical-gate output, with my reasoning, plan and commit-message bodies withheld and an instruction to derive intent from the diff.

Angles exercised: line-by-line quoting and `set -euo pipefail` interactions, exit codes and empty-output paths from every `gcloud` call, the `trap`/override-file lifecycle shared with `adopt_kms`, removed behaviour (what the 409 was protecting), cross-file tracing of every identifier to `main.tf`/`variables.tf`/the module, operational and security blast radius of a wrong adoption, reuse, simplification, altitude, conventions and canonical-doc placement, test coverage, and sibling pull requests.

Findings and disposition:

- **HIGH — adopting on every `apply` is destructive by default.** The scenario in *What Changed* above is the pass's, not mine: two installs in one project is supported, the default names are constants, and both of my guards protected the adopting install rather than the incumbent — the topic had no guard at all. **Fixed by removing adoption from `apply`**, which is the smallest of the three remedies it offered; the other two (label-based ownership, per-cluster default names) are larger changes to the module's public surface and belong with whoever wants them. This also resolved four doc findings on its own, since `apply` really does only adopt KMS again.
- **MEDIUM — the usage block silently truncated.** `sed -n '2,36p'` was sized to the old header, and this change added lines above it, so `./lifecycle.sh` with no arguments stopped printing the only documentation of `KUBE_AGENTS_STATE_BUCKET`. **Fixed** with an `awk` that prints comment lines until the first line that is not one, so the next header edit cannot repeat it — and pinned by a test. Found a latent bug while fixing it: the block read the file through a relative `BASH_SOURCE` *after* the script had `cd`'d away, so it broke whenever the script was invoked by path. The path is now resolved before the `cd`.
- **MEDIUM — a cause I asserted does not occur.** All three copies claimed the resources survive "a destroy with `enable_google_chat = false` so the module had `count` 0". `terraform destroy` plans destruction for every instance in state whether or not configuration still counts it, so that is wrong. **Dropped** from the prose and the script comment; the remaining causes are real. Both passes raised this independently.
- **MEDIUM — nothing tests or lints this script.** **Fixed**; see Testing.
- **LOW — the refusal message asserted one cause for four states**, including a subscription whose topic was deleted, which is an orphan of *this* install rather than someone else's. **Fixed.**
- **LOW — the subscription was described twice**, which was an extra call and a window in which it could vanish between the two, turning "nothing to adopt" into "unreadable topic". **Fixed**: one describe answers existence and attachment.
- **LOW — a `SERVICE_DISABLED` prompt could hang the run invisibly**, since output goes to `/dev/null` while `gcloud` blocks on the terminal. Marked PLAUSIBLE by the pass, and it is the same shape `adopt_kms` has had all along without reports. **Took the cheap prophylactic** (`</dev/null` on each describe) rather than chasing the trigger.
- **LOW — import scaffolding written out twice inline.** **Fixed** with `import_if_absent`. The pass also noted a pre-existing latent defect it propagated — `[[ -f "$OVERRIDE_FILE" ]] || with_override` skips arming the `EXIT` trap when a previous killed run left the file behind. That is now in one place rather than three, but **not fixed**: it belongs to `adopt_kms`, and changing the trap behaviour of the KMS path under a Pub/Sub change is how an unrelated install step breaks.
- **Blocking (docs-drift) — the uninstall page's "Four things in the stack"** contradicted the composition README's new "Five". **Fixed** by scoping the sentence rather than renumbering, which is what both passes agreed on: that page's table is destroy-scoped and the new asymmetry is apply-side, so "five" would have made it wrong.
- **Advisory ×5** — `chat-pubsub`'s module README had no leftover-409 note while two siblings with a `lifecycle.sh` special case do, `gke-cluster` and `gke-backup-plan` (**added**; `github-minter` has one too and does not carry such a note, so "every sibling" was wrong); the lost-state recovery sentence said "`terraform import` covers the rest" when Chat adoption is now a named step (**fixed**); a cross-reference between the two-installs constraint and Remote state, 250 lines apart (**added**). `INSTALL.md`, the `Makefile` help and `install.sh`'s engine comment were flagged as describing an apply that also adopts Pub/Sub — **no longer true after the HIGH fix**, so they stand unchanged and correct.

**A third pass then reviewed the opened pull request**, in its own subagent with the full `review-adversarial` method, and found eight more. All are fixed:

- **MEDIUM — the attachment-guard test was vacuous.** It asserted a string appeared in the file, which constrains nothing; the reviewer neutered the branch with `|| true` and the suite still passed 5/5. Replaced with the behavioural test described under Testing, verified red against the same mutation. This also corrected a claim in this body: the non-vacuity argument covered two of the three tests, not all three.
- **MEDIUM — "give concurrent installs distinct Chat names" does not produce a working second install.** The composition never exposes `service_account_id`, so every install uses `kubeagents-platform-gsa` and the second apply fails on `google_service_account` instead. The warning was right, the remedy was not; it is now scoped to what it delivers. Note this does not weaken the design argument — an automatic adoption would still run *before* that failure and leave the topic in the wrong state.
- **LOW ×6** — every `gcloud describe` failure was reported as absence, so a 403 or an unauthenticated CLI produced a silent "0 imported" (**fixed**, with the project named in the completion line); `adopt-pubsub` on a non-Chat install printed nothing at all (**fixed**); `import_if_absent`'s state guard was dead at both call sites, leaving half its contract unreachable (**fixed** by letting the helper own it); the script header claimed all five asymmetries break the second apply and that none live in a README, both untrue of item 5 (**fixed**); the README said `lifecycle.sh` "handles all five, so the cycle is repeatable" three lines above a table saying otherwise (**fixed**); and three claims in this body the source did not support (**all three corrected above** — the #841 overlap, the sibling-module note, and the #344 link).

Reported but not acted on: neither of the first two passes could execute anything — one lost its shell to worktree isolation partway through, and neither had a GCP project — so no finding rests on an executed test. I re-derived the usage truncation and the `count`-0 claim by hand, and re-ran the full live sequence above against the fixed code.

## Risk & Rollout

Narrow by construction: `apply` and `destroy` behave exactly as they did, because the new function is not on either path. The only new reachable behaviour is a subcommand that did not exist, so an install that never runs it cannot be affected — and the 409 that motivated the change is still what a bare `apply` produces, which is the pre-change behaviour.

The failure mode to watch is an operator running `adopt-pubsub` against a topic that belongs to a second install in the same project. That is the risk the design deliberately refuses to take on their behalf: the README says it in the imperative before the command, the subscription's attachment guard catches the common shape of it, and every adoption is logged with the full resource path. It cannot happen unattended.

Two changes do touch existing paths. The usage block is rewritten — it now prints more than it did before this branch, and a test pins that. And `SCRIPT_PATH` is resolved before the `cd`, which only affects the usage block that reads it. Revert is `git revert` of the two commits; nothing here writes state, creates a resource, or changes what `apply` plans.

### DIFF
--- a/terraform/examples/full-install/lifecycle.sh
+++ b/terraform/examples/full-install/lifecycle.sh
@@ -306,11 +306,17 @@ adopt_kms() {
 # That is the one answer that cannot be true for somebody who ran this subcommand
 # *because* an apply had just 409'd on these exact names, so a 403, an unauthenticated
 # gcloud, a project that does not resolve, or the API being disabled all have to say so.
-# A plain NOT_FOUND is the expected case and stays quiet.
+#
+# Matching on NOT_FOUND is what does not work, and it fails towards silence. Pub/Sub
+# answers a wrong or inaccessible project with
+# "NOT_FOUND: Requested project not found or user does not have access to it", so two of
+# the cases named above carry the same status as a genuine absence. Only the resource
+# being missing says "Resource not found", for both topics and subscriptions, so that is
+# the phrase the quiet path keys on.
 describe_failure() {
   local what="$1" err="$2"
   case "$err" in
-    *NOT_FOUND*|*"not found"*|"") return 0 ;;
+    *"Resource not found"*|"") return 0 ;;
   esac
   warn "could not read $what: ${err%%$'\n'*}"
   warn "that is not the same as it being absent — nothing was adopted for it."
--- a/terraform/examples/full-install/lifecycle.sh
+++ b/terraform/examples/full-install/lifecycle.sh
@@ -2,7 +2,7 @@
 #
 # Makes `apply` and `destroy` repeatable for this composition.
 #
-# Four things in this stack are not symmetric — applying them is not the inverse
+# Five things in this stack are not symmetric — applying them is not the inverse
 # of destroying them — and every one of them turns the second `terraform apply`
 # of a project's life into a failure. Terraform cannot express any of them, so
 # they live here rather than in a README telling you to remember them:
@@ -22,11 +22,18 @@
 #   3. A GKE BackupPlan cannot be deleted while it still owns backups.
 #   4. The cluster is created with deletion_protection = true, which a destroy
 #      cannot override on its own — the attribute has to be applied as false first.
+#   5. The Google Chat Pub/Sub topic and subscription outlive any teardown that
+#      did not run through THIS state — a destroy with enable_google_chat off, a
+#      lost state file, a hand-rolled earlier install. Unlike KMS they delete
+#      cleanly when Terraform owns them, so `destroy` needs no special case; the
+#      asymmetry is that `apply` then 409s on names it is configured to create.
+#      `adopt-pubsub` imports them instead.
 #
 # Usage:
 #   ./lifecycle.sh apply    [extra terraform args...]
 #   ./lifecycle.sh destroy  [extra terraform args...]
 #   ./lifecycle.sh adopt-kms
+#   ./lifecycle.sh adopt-pubsub
 #
 # Remote state (opt-in): set KUBE_AGENTS_STATE_BUCKET to a GCS bucket name, or
 # to "auto" for <project_id>-kube-agents-tfstate. The bucket is created if
@@ -254,6 +261,82 @@ adopt_kms() {
   log "KMS adoption complete: $adopted imported"
 }
 
+# A Google Chat topic and subscription left over from a previous install make the
+# next apply fail with a 409 on both, because Terraform is asked to create names
+# that already exist and does not own them. They are perfectly deletable, so this
+# is not the KMS problem — `terraform destroy` removes them whenever this state
+# manages them, and nothing here interferes with that. What leaves them behind is
+# a teardown this state never saw: destroyed with enable_google_chat = false so
+# the module had count 0, a state file that was lost, or a topic some earlier
+# hand-rolled install created.
+#
+# The names are the only handle there is, so adopting one means claiming a topic
+# on the strength of its name matching this install's configuration. Two guards
+# make that defensible, and neither is optional:
+#
+#   - Nothing already in state is touched, so an install that owns its topic is
+#     never re-imported.
+#   - A subscription is adopted only if it is actually attached to the topic this
+#     install is adopting. A same-named subscription pointing somewhere else is
+#     somebody else's, and importing it would hand this state a resource its next
+#     destroy would delete out from under them.
+#
+# What neither guard can catch is a second install of THIS composition in the same
+# project that kept the default topic name, since its topic is indistinguishable
+# from a leftover. Give concurrent installs distinct chat_topic_name /
+# chat_subscription_name values; that is what the 409 was protecting, and it is
+# why every adoption here is logged rather than done quietly.
+adopt_pubsub() {
+  [[ "$(tfvar enable_google_chat)" == "true" ]] || return 0
+
+  local project topic subscription topic_addr sub_addr adopted=0
+  load_state
+  project=$(tfvar project_id)
+  topic=$(tfvar chat_topic_name)
+  subscription=$(tfvar chat_subscription_name)
+  topic_addr="module.chat_pubsub[0].google_pubsub_topic.chat_events"
+  sub_addr="module.chat_pubsub[0].google_pubsub_subscription.chat_events"
+
+  if ! in_state "$topic_addr" &&
+     gcloud pubsub topics describe "$topic" --project "$project" >/dev/null 2>&1; then
+    log "adopting pre-existing Pub/Sub topic: $topic"
+    [[ -f "$OVERRIDE_FILE" ]] || with_override
+    if terraform import -input=false "$topic_addr" \
+         "projects/$project/topics/$topic" >/dev/null 2>&1; then
+      adopted=$((adopted + 1))
+    else
+      warn "could not import $topic_addr; the apply will fail with a 409 on $topic"
+    fi
+  fi
+
+  if ! in_state "$sub_addr" &&
+     gcloud pubsub subscriptions describe "$subscription" --project "$project" >/dev/null 2>&1; then
+    # The attachment is fixed at creation — a subscription cannot be repointed —
+    # so a mismatch here is not a stale reading, it is a different subscription.
+    local attached
+    attached=$(gcloud pubsub subscriptions describe "$subscription" --project "$project" \
+      --format='value(topic)' 2>/dev/null || true)
+    if [[ "$attached" != "projects/$project/topics/$topic" ]]; then
+      warn "subscription $subscription is attached to ${attached:-an unreadable topic}, not"
+      warn "projects/$project/topics/$topic — leaving it alone. It belongs to something else;"
+      warn "the apply will 409 until you rename chat_subscription_name or remove that subscription."
+    else
+      log "adopting pre-existing Pub/Sub subscription: $subscription"
+      [[ -f "$OVERRIDE_FILE" ]] || with_override
+      if terraform import -input=false "$sub_addr" \
+           "projects/$project/subscriptions/$subscription" >/dev/null 2>&1; then
+        adopted=$((adopted + 1))
+      else
+        warn "could not import $sub_addr; the apply will fail with a 409 on $subscription"
+      fi
+    fi
+  fi
+
+  drop_override
+  trap - EXIT
+  log "Pub/Sub adoption complete: $adopted imported"
+}
+
 # create_cluster = false means "somebody else's cluster" — but if THIS state
 # already manages the cluster, flipping the variable off does not hand the
 # cluster back: it removes the resource from configuration, and the next apply
@@ -412,11 +495,17 @@ case "${1:-}" in
     ensure_init
     adopt_kms
     ;;
+  adopt-pubsub)
+    shift
+    ensure_init
+    adopt_pubsub
+    ;;
   apply)
     shift
     ensure_init
     guard_cluster_ownership
     adopt_kms
+    adopt_pubsub
     log "terraform apply"
     # No -input=false: this prompts like plain `terraform apply` does. Pass
     # -auto-approve through ARGS for unattended runs.
--- a/terraform/examples/full-install/lifecycle.sh
+++ b/terraform/examples/full-install/lifecycle.sh
@@ -23,11 +23,14 @@
 #   4. The cluster is created with deletion_protection = true, which a destroy
 #      cannot override on its own — the attribute has to be applied as false first.
 #   5. The Google Chat Pub/Sub topic and subscription outlive any teardown that
-#      did not run through THIS state — a destroy with enable_google_chat off, a
-#      lost state file, a hand-rolled earlier install. Unlike KMS they delete
-#      cleanly when Terraform owns them, so `destroy` needs no special case; the
-#      asymmetry is that `apply` then 409s on names it is configured to create.
-#      `adopt-pubsub` imports them instead.
+#      did not run through THIS state — a state file that was lost or replaced, a
+#      destroy that failed part-way, an earlier hand-rolled install, or a topic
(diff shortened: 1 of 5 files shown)


terraform/examples/full-install/lifecycle.sh
```diff
@@ -254,6 +292,133 @@ adopt_kms() {
...
+#
+# Every non-zero exit otherwise lands in the same place — no warning, and "0 imported".
+# That is the one answer that cannot be true for somebody who ran this subcommand
+# *because* an apply had just 409'd on these exact names, so a 403, an unauthenticated
+# gcloud, a project that does not resolve, or the API being disabled all have to say so.
+# A plain NOT_FOUND is the expected case and stays quiet.
+describe_failure() {
+  local what="$1" err="$2"
+  case "$err" in
+    *NOT_FOUND*|*"not found"*|"") return 0 ;;
```
jayantid: `describe_failure` treats every `NOT_FOUND` as the resource being absent. Pub/Sub returns `NOT_FOUND: Requested project not found or user does not have access to it (project=...)` when `project_id` is wrong, and also when the active gcloud account has no access to the project. Both match `*NOT_FOUND*` (and the lowercase `*"not found"*`) and stay silent.

The comment above this function names "a project that does not resolve" and an unauthenticated gcloud as cases that "have to say so". Two of them don't. An operator whose active account lacks project access — right after watching an apply 409 on these exact names — runs `adopt-pubsub` and gets `Pub/Sub adoption complete: 0 imported` with no warning. That is the answer the comment calls the one that cannot be true, and it points them away from the auth problem. Reproduced: `gcloud pubsub topics describe some-topic --project this-project-does-not-exist-zz123` exits non-zero with the project-level `NOT_FOUND` above, which this case statement swallows.

A genuinely absent topic or subscription reports `NOT_FOUND: Resource not found (resource=...)`, so keying on that phrase keeps the quiet path quiet and lets project-level failures warn:

```bash
case "$err" in
  *"Resource not found"*|"") return 0 ;;
esac
```

mplakhtiy: Confirmed and fixed in `9a486cb5`. Good catch — this was in the helper I had just added in response to an earlier review finding, so it went in and straight past me.

Reproduced both shapes against live gcloud rather than taking the report on trust:

```
$ gcloud pubsub topics describe some-topic --project this-project-does-not-exist-zz123
ERROR: … NOT_FOUND: Requested project not found or user does not have access to it (project=…)

$ gcloud pubsub topics describe definitely-not-a-real-topic-zz123 --project mplakhtiy-gke-dev
ERROR: … NOT_FOUND: Resource not found (resource=definitely-not-a-real-topic-zz123)
```

Also checked the subscription verb, since the helper serves both: a missing subscription reports `Resource not found` too, so your discriminator holds for both call sites. Taken as written.

You were right about the direction of the failure as well: the function exists precisely so that "0 imported" cannot be the silent answer for someone who has just watched an apply 409 on those names, and it was swallowing the auth case that most deserves a warning.

Two tests now pin it in both directions — an inaccessible project must warn, a genuinely absent resource must stay quiet — and I verified they go red against the old `*NOT_FOUND*` matcher before restoring the fix. The stale `*"not found"*` lowercase alternative is gone with it; it matched "Requested project not found" as well.

mplakhtiy:
Closing this — on reflection the problem it solves is a migration artifact rather than a standing defect, and it is not worth the surface.

The timeline makes the case. #797 ("make Terraform + Helm the only install engine") merged 2026-08-20 11:47; #831 was filed at 13:44 the same day, from the first from-scratch install on the new engine. The leftover topic it hit was left by a **pre-#797 install torn down by the old engine**, which had no Terraform state owning it. Post-#797 nothing deletes those resources outside Terraform, so `lifecycle.sh destroy` owns them and removes them cleanly — a normal destroy/apply cycle never reaches this, and the population that can is shrinking to zero as installs migrate.

What that leaves is too thin to justify a new subcommand, ~60 lines of bash, and a permanent entry in the lifecycle interface: state loss or a part-way destroy, both of which usually leave far more than a Pub/Sub topic behind, and both of which are recoverable by deleting two resources.

Two things worth recording for whoever finds this next.

**The composition README still promises a recovery that does not work for Chat installs.** "If the state is gone entirely, re-run `lifecycle.sh apply` against the same tfvars — KMS adoption is automatic, and `terraform import` covers the rest." For a Chat-enabled install the apply 409s on the topic and subscription. That sentence was already wrong before this PR and stays wrong after closing it; it wants either a caveat or the hand-import recipe, and it is a docs fix rather than an engine one.

**If this ever does need building, do not put it on the `apply` path.** `chat_topic_name` defaults to a project-wide constant while the composition supports two installs per project, so an automatic adoption cannot tell a second install's live topic from a leftover — it would redirect that install's Chat traffic and delete the topic on the next teardown. The 409 is the safer default. That reasoning, the attachment guard, and the behavioural tests are all in this branch's history (`fix/831-adopt-pubsub`, head `9a486cb5`) if it is ever wanted.

Thanks @jayantid for the `NOT_FOUND` catch — it was a real bug, and the same discriminator would matter to anything that reads gcloud describe errors in future: a wrong or inaccessible project returns `NOT_FOUND: Requested project not found…`, while a genuinely missing resource returns `NOT_FOUND: Resource not found (resource=…)`.

#831 item 1 goes back to open. Items 2, 3, 4, 6 and 7 are #841; item 5 is #936; item 8 needed no change.

Issue #920, opened 2026-08-24
Installation friction: fails on autopilot clusters


justinsb:
Trying to list node-pools in an autopilot cluster gives this error:

`ERROR: (gcloud.container.node-pools.list) ResponseError: code=400, message=Autopilot node pools cannot be accessed or modified.`

This does not play well with install.sh

```
  # Enabling the pool does not migrate node pools off the legacy GCE metadata
  # server, and pods on such pools still get the node's service account.
  # Standard-cluster concern: Autopilot pools are managed onto GKE_METADATA
  # already.
  local legacy_pool
  while IFS= read -r legacy_pool; do
    [ -n "$legacy_pool" ] || continue
    print_warning "Node pool '${legacy_pool}' uses the legacy GCE metadata server; migrating to GKE_METADATA (this recreates the pool's nodes)..."
    gcloud container node-pools update "$legacy_pool" \
      --cluster="$cluster_name" --location="$region" --project="$project_id" \
      --workload-metadata=GKE_METADATA --quiet
  done < <(gcloud container node-pools list --cluster="$cluster_name" \
      --location="$region" --project="$project_id" \
      --format="csv[no-heading](name,config.workloadMetadataConfig.mode)" 2>/dev/null \
    | awk -F',' '$2 != "GKE_METADATA" {print $1}')
```

bradhoekstra:
Fixed by #1014. `ensure_existing_cluster_workload_identity` in `install.sh` now describes the cluster first and returns before the node-pool loop when `autopilot.enabled` is true, so the `node-pools list` 400 no longer occurs; the loop itself also no longer trips the ERR trap. That PR was live-tested against an Autopilot cluster. Closing; please reopen if a different Autopilot path still fails.

PR #941, opened 2026-08-25
feat(installer): add --cluster-mode so a fresh install can create an Autopilot cluster

mplakhtiy:
## Summary

`install.sh` gains `--cluster-mode=autopilot|standard`, and the interactive interview asks for the shape on every branch that is about to create a cluster. The `gke-cluster` Terraform module has built Autopilot clusters since it was written and `terraform/examples/full-install` still defaults to them (`variables.tf:14`), so no Terraform changed here — the installer was the half that had never caught up, hardcoding `cluster_mode = "standard"` for every cluster it created. The installer default stays `standard`: a bare `./install.sh` builds exactly what it built before. Generated with the help of Claude.

## Why This Change

Since #797 replaced the `provision_0*` scripts with the Terraform + Helm engine, the only way to get an Autopilot cluster out of this repository has been to drive the composition by hand with your own `terraform.tfvars`. Anyone using the front door — which is the documented path, the `curl | bash` path, and the one the `install-kube-agents` skill drives — got Standard whether or not they wanted it, with no flag to say otherwise. The `--gvisor`-on-Autopilot handling merged in #823 is already sitting there unusable from the installer for the same reason: nothing the installer can create is ever Autopilot.

`README.md` and `INSTALL.md` also both described the installer as Standard-only, and `INSTALL.md` contradicted itself two paragraphs apart — line 58 called it Standard while line 146 described the engine it drives as `cluster_mode = "autopilot" | "standard"`.

## What Changed

- **`install.sh`** — `--cluster-mode`, its `--help` entry, a `prompt_menu` for the shape on the create branches, `CLUSTER_MODE` in `vars.sh`, and two new helpers: `require_creatable_cluster_mode` (rejects an unknown shape, and Autopilot against a zone, at the interview rather than at the module's plan-time precondition) and `persist_effective_cluster_mode` (below). The pre-flight summary printed a hardcoded `GKE Standard` whatever it was building; it, the completion summary, and the JSON report now print the generator's answer.
- **`k8s-operator/scripts/installer_common.sh`** — `is_valid_cluster_mode`, and `write_tfvars_from_state` now reads `CLUSTER_MODE` on exactly one branch. It exports `TFVARS_CLUSTER_MODE` alongside the existing `TFVARS_CREATE_CLUSTER`.
- **`.github/workflows/installer-matrix-test.yml`** — three more `--dry-run` invocations asserting the generated `terraform.tfvars` and `vars.sh`, including one adoption case. Terraform is already stubbed there, so this is cheap; see Testing for why the adoption case is the one that matters.
- Docs: the quickstart flag list, the skill's flag table, the `vars.sh` state model in `k8s-operator/scripts/README.md` (which owns the caveat below), plus the two Standard-only claims above. `INSTALL.md:146` and the composition READMEs were already correct and are untouched.

### The teardown invariant, and how it is kept

`installer_common.sh` carries this comment, from a real outage:

> `cluster_mode` follows the LIVE cluster when there is one. Hardcoding "standard" here planned the destruction of every existing Autopilot install the moment a front door regenerated tfvars against it — the autopilot resource's count went to 0, so uninstall's targeted deletion-protection apply and upgrade's full apply both became cluster replacements.

`uninstall.sh:342` and `upgrade.sh:464` drive the same generator, from `vars.sh` alone, with no flag to correct a wrong value. Persisting `CLUSTER_MODE` therefore makes it an input to teardown, so the change is structured so that a wrong value cannot reach a live cluster:

- The generator assigns `cluster_mode` from the probe on **both** branches where a cluster exists — the `else` that assigns `"standard"` is redundant against the initialiser and deliberately so. `CLUSTER_MODE` is read only after a confirmed `NOT_FOUND`, where there is no live cluster to reshape.
- `install.sh` writes the generator's probed answer **back** to `vars.sh` (`persist_effective_cluster_mode`), so the file cannot sit contradicting its own cluster and later rebuild the wrong shape.

## Context

- **Closed PR #475** (`feat(scripts): choose Autopilot or Standard, and preflight the namespace quota`, mplakhtiy, closed 2026-08-17) is the closest prior art and near line-for-line this feature. It was not rejected on the merits — it was obsoleted three days before #797 merged, because every file it edited (`provision_01`–`08`, `teardown_01/02`) was deleted by the Terraform + Helm rewrite. Its design informed this one. Its other half, the namespace-quota preflight, never landed either and is still open ground.
- **Issue #920** (`Installation friction: fails on autopilot clusters`, justinsb, assigned to justinsb) is adjacent, not this, and this PR does **not** close it. That issue is the *adopt an existing Autopilot cluster* path; this is the *create a new one* path. Worth flagging for whoever picks it up: the `gcloud container node-pools list` in `ensure_existing_cluster_workload_identity` (`install.sh:948-957`, not 905-916) already has `2>/dev/null` and feeds a process substitution, so its error text is suppressed and the loop no-ops — the report may be partly stale, or the failure arrives by another route. That helper also runs before the apply, so a cluster this run creates is not yet there to trip it.
- **Open PR #865** (`feat(install)!: sandbox the agent under gVisor by default`, AntonTyb) edits the same `write_tfvars_from_state` region and the same two test files: merge-conflict risk, not duplication. It is itself blocked ("Do not merge until #610 is resolved" — SQLite WAL corruption on the gVisor 9p mount), which is why this did not wait on it. Happy to sequence behind it if that is easier; whichever lands second should re-run the gVisor × cluster_mode matrix, since making gVisor the default turns the new fresh-create Autopilot branch into the path every Autopilot install takes.
- **Merged PR #823** (`fix(install): let --gvisor sandbox the agent on Autopilot`) already added the Autopilot gVisor derivation and its version floor. This inherits it by setting `cluster_mode` correctly rather than duplicating it. The one adjustment: that floor probes a live cluster, and a fresh `--cluster-mode=autopilot --gvisor=true` create has none to probe, so it is skipped there — a cluster created now comes up on its release channel's current version, far past a 2023 floor.
- Branched from `c4727da7`; `upstream/main` has since moved 3 commits (#879, #932, #934). `git merge-tree` reports a clean merge and none of those touch the regions this changes.

### The default is deliberately not flipped

The module defaults to `autopilot` and the installer to `standard`, and after this they still disagree. Flipping the installer default changes what a bare `./install.sh` builds — that needs `feat!:` and its own argument, and it is not this PR. Adding the flag at the existing default is non-breaking and closes the gap; the flip is worth proposing separately.

### Pre-existing, not fixed here

`local.cert_manager_resources` (`terraform/examples/full-install/main.tf:112-121`) requests 10m CPU / 32Mi, below Autopilot's documented per-Pod minimums. Autopilot mutates requests upward rather than rejecting them, so this is a billing surprise rather than a failure, and it already affects every Terraform-driven Autopilot install today. Out of scope here; flagging it because this PR makes that path reachable from the installer.

## Testing

- `python3 -m unittest tests.test_install_script tests.test_installer_common tests.test_uninstall_script tests.test_upgrade_script` — 52 tests, OK. Every `tests/test_*.py` module also run individually, all green.
- **Red-green on every new test.** With `install.sh` and `installer_common.sh` reverted to `main` and the new tests kept, 9 failed: `test_tfvars_fresh_create_honours_cluster_mode`, `test_tfvars_fresh_create_rejects_an_unknown_cluster_mode`, `test_tfvars_live_cluster_outranks_a_conflicting_cluster_mode`, `test_tfvars_gvisor_on_a_fresh_autopilot_create_skips_the_version_probe`, `test_cluster_mode_defaults_to_unset`, and the four `require_creatable_cluster_mode` cases (`command not found`). The three `persist_effective_cluster_mode` tests likewise fail without the function.
- `test_tfvars_standard_cluster_and_fresh_create_stay_standard` was renamed to `..._unasked_fresh_create_...` and amended rather than deleted: its live-Standard half still holds unchanged, and its fresh-create half now pins the no-`CLUSTER_MODE` default instead of the removed hardcoding.
- `bash -n install.sh`; `shellcheck install.sh uninstall.sh upgrade.sh k8s-operator/scripts/installer_common.sh` — clean apart from four pre-existing warnings on untouched lines.
- `prettier@3.9.6 --check` on every changed Markdown/MDX/YAML — clean. `make docs-check` — clean.
- **`--dry-run` against the same command stubs CI uses.** `--cluster-mode=standard` and `=autopilot` each produced the matching `cluster_mode` in `terraform.tfvars` and `CLUSTER_MODE` in `vars.sh`; `--cluster-mode=autopiloot` and `--cluster-mode=autopilot --region=us-central1-a` both aborted at step 4, before anything was written.
- **The adoption cases, with a stub reporting a live cluster.** `--cluster-mode=autopilot` against a live Standard cluster wrote `cluster_mode = "standard"` to both files; no flag at all against a live Autopilot cluster wrote `"autopilot"` to both. Against the first commit of this branch the same two runs left `vars.sh` at `autopilot` and `standard` respectively — contradicting the cluster in both directions. The new CI assertion is red without the write-back, verified by deleting the call and running the workflow step locally.
- The `prompt_menu` branches were driven through a pty, including adopting a discovered zonal cluster with `--cluster-mode=autopilot`, which must not abort.

### Live validation

**Not live-tested.** No GKE cluster of either shape was available in this environment, and this change provisions cloud resources, so nothing below substitutes for it. Everything above ran against stubbed `gcloud` and `terraform` binaries.

A real validation needs, in order:

1. `./install.sh --cluster-mode=autopilot` as a fresh create, confirming the generated `terraform.tfvars` says `cluster_mode = "autopilot"` and that `gcloud container clusters describe --format="value(autopilot.enabled)"` returns `True` on the result.
2. `vars.sh` holding `export CLUSTER_MODE=autopilot` after that run.
3. **The highest-risk untested path:** `upgrade.sh` and then `uninstall.sh --dry-run` against that same cluster, confirming the plan contains **no** cluster replacement. That is the failure this change is shaped to prevent and the one no stub can prove.
4. The mirror image — `--cluster-mode=autopilot` pointed at an existing *Standard* install — confirming `vars.sh` and `terraform.tfvars` both come back `standard` and the plan still shows no replacement.

Steps 1 and 2 are the feature; 3 and 4 are the safety property. I would not merge this without at least 3.

No test artifacts left behind: the generated `vars.sh`, `terraform.tfvars`, `/tmp/kube-agents-install-report.json` and all stub directories were removed, and `git status` is clean.

## Self-Review

Three passes, each in a **subagent given only the diff range and the skill file** — no plan, no reasoning, no summary from the session that wrote the code.

**Adversarial pass (`review-adversarial`), first round.** Asked specifically to attack the teardown and upgrade paths and to try to construct any state in which the generated tfvars name a shape the live cluster does not have. Angles examined: the generator's probe branches against live-Autopilot and live-Standard stubs with the opposite `CLUSTER_MODE`; Terraform resource counts and the `moved` block; `set -u`/`set -e`/`set -Eeo pipefail` and bash 3.2; variable scoping in `main()`; reachability of the new prompt; the non-interactive path; whether the new tests test what they claim.

- **High, fixed.** `install.sh` persisted `CLUSTER_MODE` from the interview while every other reader used the generator's probed answer, so `vars.sh` could contradict its own cluster — reproduced in both directions. `persist_effective_cluster_mode` now writes the probed shape back. This was the most important finding in the review and I had missed it.
- **Medium, fixed.** Documentation asserted the "no effect on an existing cluster" behaviour that the above broke.
- **Medium, fixed.** The shape prompt was reachable on only one of the three paths that create a cluster.
- **Low, fixed.** The generator told a teardown it was "creating a GKE standard cluster" when `uninstall.sh` hit the not-found branch on an already-deleted cluster.
- **Low, noted not fixed.** `--menu --cluster-mode=X` accepts the flag and ignores it, because menu mode exits before the flag is used. Every other flag behaves the same way there; changing it is a separate change to menu-mode argument handling.
- **Low, noted not fixed.** The generator's `NOT_FOUND` classifier greps for `not.?found|404`, which also matches the shell's own `gcloud: command not found`. Pre-existing; this change adds a second decision behind the same match. `upgrade.sh` checks its tools first, `uninstall.sh` does not — but for a destroy the shape is moot. Worth tightening separately, not inside this diff.

**Adversarial pass, second round** — a fresh subagent given the fix commit and asked to verify the fix rather than trust it. It reproduced the original defect and confirmed the fix, and found that the fix had introduced one of its own:

- **High, fixed.** Hoisting `require_creatable_cluster_mode` out of the create branch made it judge a `$region` the *discovery menu* had overwritten with a chosen cluster's location. Adopting a discovered zonal cluster with `--cluster-mode=autopilot` aborted, blaming a `--region` the caller had spelled correctly. It now runs only where a cluster is about to be created. Confirmed through a pty: exit 1 before, past step 5 after.
- **Medium, fixed.** Nothing went red when the reconciliation call was deleted — the CI stub answers `NOT_FOUND` for every describe, so requested and effective always agreed. The stub now takes `STUB_LIVE_AUTOPILOT`, and the new assertion is verified red without the call.
- **Medium, fixed.** `save_var`'s dry-run guard tests `DRY_RUN`, which `install.sh` never sets — it uses `PARAM_DRY_RUN`, spelled `"true"`. So `DRY_RUN=1 ./install.sh` was a *real* install whose write `save_var` skipped after announcing it, and `DRY_RUN=true` printed an integer-comparison error. Pinned at the call site.
- **Low, fixed.** A no-op assignment, and the "type an unlisted cluster name" branch not asking for a shape while its sibling did.
- **Low, noted not fixed.** The Day-2 menu regenerates tfvars without reconciling `vars.sh` back, so a file written by a pre-`CLUSTER_MODE` installer is repaired only by a full `install.sh` run. Bounded — the probe overrules `CLUSTER_MODE` whenever the cluster is live, so nothing unsafe is generated; the menu simply misses a chance to self-heal. Making `write_tfvars_from_state` itself write back would fix all three front doors at once and is the better shape, but it makes a generator that only generated start mutating state, which deserves its own PR.
- **Suspected only, not acted on:** that `gcloud container clusters list` swallowing a permissions error could route a user with clusters into the "none found" branch. Pre-existing and unrelated to cluster mode.

**Docs-drift pass (`review-docs-drift`).** Areas searched: every enumeration of installer flags or `vars.sh` keys across `docs/`, `INSTALL.md`, `README.md`, `.agents/skills/`, `charts/`, `terraform/`, `agents/`; the uninstall and upgrade skills; `docs/README.md` map coverage; generated regions.

- **Blocking, all three fixed.** `README.md:33` and `INSTALL.md:58` still called the installer Standard-only. And prose *this change added* to `k8s-operator/scripts/README.md` claimed `CLUSTER_MODE` was the only key the generator reads conditionally — several others are; the true, narrower claim is that it is the only one whose saved value the generator discards.
- **Non-blocking, fixed.** "the interactive installer asks" overstated which paths ask; the quickstart's prompt enumeration was missing the shape question; "Ignored on an existing cluster" was imprecise, since an invalid value still aborts; the quickstart's "the two cannot disagree" needed softening.
- **Non-blocking, not fixed.** `overview/architecture.mdx:108` describes a default node pool that an Autopilot install does not have — pre-existing, since the composition already supported Autopilot, and it belongs to an architecture page rather than to this flag. Reported rather than swept in.
- **Clean:** no `docs/README.md` map row needed (no document added, moved or deleted); no generated region touched; the abridged JSON-report sample in the skill needs no new field; the uninstall/upgrade skills say nothing this falsifies.

## Risk & Rollout

The default is unchanged, so an install that passes no new flag generates a byte-identical `terraform.tfvars` for its cluster block. The blast radius is `write_tfvars_from_state`, which all three front doors share: a bug there reaches teardown, which is why the probe is structurally authoritative on every branch where a cluster exists and why the reconciliation writes the probed shape back. The realistic failure mode left is a `vars.sh` hand-edited between an install and an `upgrade.sh` on a cluster that has meanwhile been deleted — narrow, and the same window that already existed for `CLUSTER_NAME`.

Reverting is a clean revert of one commit; no state migration, and a `vars.sh` carrying `CLUSTER_MODE` is simply ignored by the older generator. Nothing has to happen at merge time. The one thing I would want before merge is item 3 of the live validation above, on a real Autopilot install.


k8s-operator/scripts/installer_common.sh
```diff
@@ -53,6 +53,13 @@ is_valid_permission_set() {
   [[ "${1:-}" =~ ^(read-only|gke-admin|custom)$ ]]
 }
 
+# The cluster shapes the gke-cluster module can build. Matches the module's own
+# variable validation, so a bad value fails at the interview rather than at
+# terraform validate with the cluster interview already paid for.
+is_valid_cluster_mode() {
+  [[ "${1:-}" =~ ^(autopilot|standard)$ ]]
```
AntonTyb, line 60: don't we have this validation in tf already?

AntonTyb (approved):
one question to discuss.

k8s-operator/scripts/installer_common.sh
mplakhtiy: Yes — `cluster_mode` is validated by the module (`terraform/modules/gke-cluster/variables.tf:29`) and again by the composition (`terraform/examples/full-install/variables.tf:17`), and the Autopilot zone rule is a `lifecycle.precondition` at `terraform/modules/gke-cluster/main.tf:135`. This mirrors both deliberately, to fail at the interview instead of after it: by the time `terraform validate` runs, the installer has already collected every API key and integration answer, so a bad `--cluster-mode` costs the user the whole interview before anything rejects it. Same reason `is_valid_model_provider` (:46) and `is_valid_permission_set` (:52) mirror their downstream constraints in this file.

The cost is drift, and it is worth naming: the zone regex `^[a-z]+-[a-z]+[0-9]+$` is copied verbatim from the precondition and nothing keeps the two in step. If you would rather not carry that, the alternative I would suggest is dropping the shell zone check and keeping only the enum — the enum is stable, while the regex is the half likely to change. Happy to do that if you prefer it.

AntonTyb: cool. thanks.

Issue #950, opened 2026-08-25
eval presubmit: seeded-cluster reuse leaves the eval orphan sweep with no runs to ride


haoxuw:
## Gap

The eval orphan sweep (`terraform_data.reap_orphans` in `bench/tf/modules/cluster/gke`) runs only when the gpu-stress-test stack creates a cluster. With seeded-cluster reuse (PR to follow, referencing this issue), a fleet-carrying project's presubmits stop creating clusters — and all three pool projects carry the fleet as of 2026-08-24 (`bench/tf/fleet/README.md`), so in CI the sweep ordinarily never runs again.

What the sweep covered: `eval-pr-*` clusters older than 4h (crashed runs whose teardown never fired), their per-cluster node service accounts, their project IAM bindings, and the `deleted:serviceAccount:` tombstone bindings. Anything orphaned before the reuse change merges, or by a rare per-run fallback whose job dies pre-teardown, now bills until a human sweeps it.

## What to do

The fleet README already names the right owner for a related residual (a live SA orphaned between account and cluster creation): "a scheduled janitor with an inventory". This is the same janitor. Options, cheapest first:

1. Run the sweep's logic as a step in `hack/ci-eval-pr.sh` (it is best-effort and label-scoped already), decoupled from cluster creation.
2. A scheduled GitHub workflow per pool project running the same label-scoped sweep — the fleet README's planned reconcile workflow is a natural host.

## How to reproduce

Leave a `managed-by=kube-agents-bench`-labeled cluster >4h old in a fleet-carrying pool project and run the presubmit: with reuse active, `reap_orphans` never executes and the cluster survives indefinitely (previously any later run's provision deleted it).

haoxuw:
Context update: **this gap is now live, not hypothetical.** #951 merged on 2026-08-26, so in fleet-carrying pool projects — all three today — every presubmit takes the reuse path and `reap_orphans` does not run.

One refinement to the issue text: GC is now *opportunistic* rather than strictly absent. The sweep still fires whenever a run falls back to per-run provisioning — no fleet in the leased project, or slot c missing, not RUNNING (its nightly maintenance window), dirty in `default`, or ambiguous — so an orphan waits for the next fallback run, which is unbounded rather than never. The repro in the issue body stands.

[...]

Issue #949, opened 2026-08-25
eval presubmit: GKE provision+teardown costs ~830s, ~470s of it serial teardown ahead of scoring


haoxuw:
## Bottleneck

GKE provision + teardown is ~830s — about 31% of the infra task's 2704s and a quarter of the whole eval presubmit. From the profiled run of `hack/ci-eval-pr.sh` (instrumentation from #947, 2026-08-25, project `haoxuw-gke-dev`, eval cluster `eval-pr0-haoxuw`, g2-standard-4 in `us-west4-a`), cross-checked against `gcloud container operations list`:

| phase | evidence | dur |
|---|---|---:|
| tofu init + apply overhead, log seeding | profile span minus GKE ops | ~45s |
| `CREATE_CLUSTER` | 14:07:33 → 14:11:19 | 226s |
| default node pool delete | 14:11:26 → 14:12:17 | 51s |
| GPU node pool create | 14:12:29 → 14:13:10 | 42s |
| destroy: node pool delete | 14:44:14 → 14:47:53 | 220s |
| destroy: `DELETE_CLUSTER` | 14:47:57 → 14:52:02 | 245s |

Judge scoring ran only ~11s after the destroy finished — the teardown sits serially between the agent finishing and the score existing.

## What to optimize

- **Teardown off the critical path (~470s).** The harness tears down before scoring, and teardown is already documented as best-effort-never-raise. Firing the destroy `--async` (as the stack's own orphan reaper already does for its deletes) or backgrounding it once the run record is written would reclaim ~17% of the infra task with no change to what gets scored. Needs care with the orphan sweep's 4h age filter so an in-flight async delete is not double-reaped.
- **Default-pool swap (~95s + two GKE operations).** The cluster is created with a default pool that is immediately deleted and replaced by the GPU pool (the standard `remove_default_node_pool` pattern in the tofu stack). Creating the cluster with the GPU pool as its initial pool — or an autopilot/pool-inline variant — saves both operations.
- **Serial task ordering.** The noop kanban probe (562s, no infra) currently waits for the gpu task's full teardown; overlapping them, or running teardown concurrent with the next task's startup, stacks with the async-destroy win.

## How to reproduce

With #947 checked out:

```bash
export GEMINI_API_KEY="$(kubectl get secret platform-agent-secrets -n kubeagents-system -o jsonpath='{.data.GEMINI_API_KEY}' | base64 --decode)"
[...]

Issue #948, opened 2026-08-25
eval presubmit: agent execution is ~73% of the job (9.3 min for one kanban card)


haoxuw:
## Bottleneck

Agent execution is ~73% of the entire eval presubmit. A profiled run of `hack/ci-eval-pr.sh` (instrumentation from #947, run 2026-08-25 against project `haoxuw-gke-dev`, total 3270s):

| span | dur | share |
|---|---:|---:|
| gpu-stress-test-diagnosis: devops-bench run | 2704s | 82.7% of job |
| — of which agent execution (`results.json` `latency`) | 1829s | 67.6% of that run |
| agent-kanban-smoke: devops-bench run | 562s | 17.2% of job |
| — of which agent execution | 559.5s | 99.5% of that run |

Combined agent time: 2389s of 3270s. Everything the script itself does (auth, config, classify, gate) is under 4s.

The kanban smoke task is the sharper signal: **9.3 minutes for the platform agent to create one kanban card and reply with its id**. There is no infrastructure in that task at all — the time is entirely the agent round-trip (CLI harness polling the gateway, the agent's model calls, delegation plumbing).

## What to optimize

- Measure where the 9.3 minutes of the kanban probe go: harness poll interval vs. gateway turnaround vs. model round-trips. If the CLI harness polls the platform-agent service on a coarse interval, tightening it is free wall-clock.
- The gpu task's 1829s of agent time bounds the job from below regardless of infra work — worth a latency budget per task, and per-turn timing in the agent transcript so regressions show up as a number, not as a slower presubmit.
- The two tasks run serially; the noop kanban probe could overlap the gpu task's teardown (~470s) if the harness allowed it (see the provision/teardown issue for that half).

## How to reproduce

With #947 checked out:

```bash
export GEMINI_API_KEY="$(kubectl get secret platform-agent-secrets -n kubeagents-system -o jsonpath='{.data.GEMINI_API_KEY}' | base64 --decode)"
PROJECT_ID=<project> REGION=us-central1 ./hack/ci-eval-pr.sh
```

The `=== Step timing profile ===` table at exit gives the per-step split summing to 100%; each task also prints a `── devops-bench phase breakdown ──` table, and `agent latency from results.json` is the exact agent span. On any Prow run of the job, the same tables appear in the build log once #947 merges.

haoxuw:
Context update, post #947 and #951 (both merged):

[...]

PR #951, opened 2026-08-25
ci(eval): reuse the seeded fleet's slot-c cluster instead of provisioning per run

haoxuw:
## Summary

When the leased eval project carries the standing seeded fleet, the presubmit's infra task now reuses the fleet's slot-c cluster as its log-fixture subject instead of provisioning and tearing down a GKE cluster per run; a project without a fleet keeps today's per-run cluster. The reuse is opt-in per stack (`reuse_existing_cluster`), guarded by a pre-reuse cleanliness check on the shared cluster's `default` namespace, and restricted to slot c — a missing, non-RUNNING, or ambiguous slot c falls back to per-run provisioning, never to a sibling slot. Measured before/after numbers are in the table below: the change removes ~810s (~13.5 min) of infrastructure time from every presubmit.

## Why This Change

#949 measured GKE provision+teardown at ~830s of every presubmit — ~31% of the infra task — for a cluster that hosts no workloads at all: the gpu-stress-test stack's own comment records that the cluster exists only because devops-bench's `TFDeployer` unconditionally needs `cluster_name`/`cluster_location` outputs and real credentials, while the entire planted incident is two Cloud Logging entries that merely name a cluster. The seeded fleet (#848/#938) put a standing, labeled, discoverable cluster in every pool project; naming it in those log entries costs nothing it wasn't already built to carry.

Measured on the same install (`haoxuw-gke-dev`, 2026-08-25), profiled with #947's instrumentation:

| step | before (per-run cluster) | after (seeded reuse) | saved |
| --- | ---: | ---: | ---: |
| provision (tofu init + apply) | ~360s | 17.2s (incl. the 10s log-ingestion wait) | ~343s |
| teardown (tofu destroy) | ~470s | ~0.4s (fixture resource only) | ~470s |
| discovery + cleanliness check (new) | — | 2.9s | −2.9s |
| **infra overhead per presubmit** | **~830s** | **~20s** | **~810s** |

Whole-job wall clock across the three measured runs went 3270s → 1978s → 285s, but most of that spread is agent-side variance the change does not control (agent latency on the identical gpu task was 1829s, ~1420s, then 32.1s — by the third run the platform agent had seen the same synthetic incident twice and likely answered from memory; #948 tracks agent latency). The ~810s row above is the deterministic, every-run effect of this PR.

## What Changed

- `hack/ci-eval-pr.sh` §3b: discovers the fleet's slot-c cluster by the fleet's documented labels (the same pair `hack/fleet-kubeconfigs.sh` filters on), verifies its `default` namespace is empty through a throwaway kubeconfig, and records it as the reuse subject. Slot c only: slot a carries a live saturated HPA that an agent chasing this task's *synthetic* HPA incident could report instead, and slot b's held-back control plane is the same kind of bait.
- Slot-c resolution is fail-safe three ways: not found → per-run; not RUNNING (maintenance window, fleet re-apply) → per-run; two labeled `-c` clusters → ambiguous, per-run with a warning — the same ambiguity rule `hack/fleet-kubeconfigs.sh` applies to slots. The cleanliness probe carries `--request-timeout=30s`, so a wedged seeded API server costs the run half a minute and a fallback, not a hang.
- Per-task opt-in in the task loop: a task gets the seeded name (and matching `{{CLUSTER_NAME}}` placeholders) only if its stack declares `variable "reuse_existing_cluster"`; every other task keeps the per-run name — so a future tofu task that has not opted in (e.g. the stacks in #516) cannot 409 against the seeded cluster.
- `bench/tf/prebuilt/gpu-stress-test/`: `reuse_existing_cluster` zeroes the cluster module; the fixture and outputs read the subject from the variables via `coalesce(one(module.cluster[*]...), var...)`; the fixture `null_resource` gains a `subject_cluster` trigger so persistent local state re-plants when the subject moves; destroy on a reuse run removes only the fixture resource.
- Docs that stated invariants this change bends now name the exception: the fleet catalog's addressing exclusivity (`fixtures.json`, `DRAFTS.md`, `ci-pool-projects.md`), a fleet README paragraph owning the presubmit as the fleet's second consumer (including that slot-c's Cloud Logging history permanently carries the synthetic fixture entries), and the ambient-context claims in `DRAFTS.md` A5, `CUSTOM-TASKS.md`, and `bench/README.md`.

## Context

- Builds on jayantid's seeded-fleet work (#848, #938). The open chain #939 → #944 moves fleet *scenarios* to `deployer: noop` and adds a per-case `plant.sh` harness, with #944 naming gpu-stress-test's conversion as a follow-up — this PR is a smaller step in the same direction that works with today's `deployer: tofu` path and needs neither of those to merge first. If #944's plant-hook route later supersedes the stack-side fixture, `reuse_existing_cluster` and §3b shrink with it.
- Addresses the provision/teardown half of #949 (the default-pool-swap and serial-teardown items there remain relevant for the fallback path). Orphan-sweep cadence in fleet-carrying projects is deliberately not solved here — #950 tracks it and the stack comment points there.
- #947 (step profiler) touches other regions of `hack/ci-eval-pr.sh`; whichever merges second rebases trivially. #925/#939/#944 also rewrite parts of this script — coordination noted, no overlap in what they change.

## Testing

- `bash -n`; `tofu validate` and `tofu plan` on the stack in both modes: `reuse_existing_cluster=true` plans exactly 1 resource (the fixture) with outputs carrying the existing cluster's identity; `false` plans the full 10-resource set, byte-identical to today.
- `python3 scripts/test_task_registration.py` (4 tests OK) and `bench/tests/test_fleet_verifier.py` (88 passed) against the modified files.
- `make docs-check` clean; `make prettier-write` on the changed Markdown/YAML (unrelated pre-existing churn in `.agents/skills/` reverted, not committed).

### Live validation

Install: `platform-agent-host` (GKE `us-central1`, project `haoxuw-gke-dev`), namespace `kubeagents-system`. The project has no real fleet, so a mock fleet member (`seeded-mock-c`, then `seeded-mock2-c`: e2-small, one node, `us-west4-a`, both fleet labels, slot-c name) stood in; both were deleted afterwards.

**Full run (commit 964aec3, the reuse commit before the review-fix round):** `PROJECT_ID=haoxuw-gke-dev ./hack/ci-eval-pr.sh` end to end. Observed: `Reusing seeded cluster … seeded-mock-c (us-west4-a)`; tofu `Plan: 1 to add` and apply in seconds (vs ~360s provision at baseline); fixture entries written with `cluster_name=seeded-mock-c` and queryable after 10s; `get-credentials` for the seeded cluster succeeded (the `get_cluster_info` seam); **no `eval-pr*` cluster existed at any point during the run**; both tasks PASSED the deterministic gate; teardown printed `Destroy complete! Resources: 1 destroyed` and the mock cluster stayed RUNNING. Total 1978s vs the 3270s baseline measured on the same install the day before (#947's profile).

**Review-fix round (commit fc19768), validated at block level rather than by a second 55-minute run:** the changed selection/gating logic was exercised live against `seeded-mock2-c` — clean `default` → reuse chosen; a planted `deployment/rogue-leftover` in `default` → the dirty-namespace warning and fallback to per-run; deleted → reuse again; per-task gating resolves gpu-stress-test → REUSE and agent-kanban-smoke (no stack) → PER-RUN. The fallback path with no fleet at all was exercised standalone after the first mock's deletion (`No reusable seeded slot-c cluster …`). The env the reuse branch exports for the gpu task is value-identical to what the full run already proved downstream, and the stack delta (the `triggers` block) was verified by plan.

**Third run (profiled, post-review-fix logic):** a local merge of this branch (fc19768) with #947's profiler, against a fresh mock fleet member — total 285.5s, exit 0, both tasks PASSED. Phase table for the gpu task: startup 1.0s, provision 17.2s, agent 32.1s, drain 3.1s, teardown 0.0s, scoring 2.2s; seeded discovery + cleanliness check cost 2.9s. This is the source of the before/after table above.

**Hardening round (commit 1ac4c7c), validated at block level:** the ambiguity guard was exercised with an input matrix (0 lines → fallback, 1 → reuse, 2 → ambiguous-fallback) and the no-fleet fallback re-run live after the mock cluster's deletion; it is upstream of everything the three full runs proved and changes no exports.

**Not covered:** a run against the real fleet trio in a pool project (no access from this checkout) — the slot-c preference among three labeled clusters is exercised only by the awk selection against mock data; and the Prow environment itself. Artifacts: the mock clusters are deleted; the synthetic log entries from validation runs remain in `haoxuw-gke-dev`'s Cloud Logging until retention expires (same as every baseline run); one "devops-bench smoke probe" kanban card per run was created by the smoke task itself.

## Self-Review

Both passes ran in subagents handed only the diff range (`refs/kube-agents-base/main...HEAD`, freshly fetched base), not the authoring context. Merged dispositions:

- **[HIGH/PLAUSIBLE, adversarial] Rogue-agent mutation of the shared cluster's `default` ns reds every later PR persistently** → fixed: pre-reuse cleanliness check, dirty → per-run fallback with a warning naming the leftover; verified live both ways.
- **[MEDIUM/CONFIRMED, adversarial] Orphan sweep rides the now-count-zero cluster module, and every pool project carries the fleet** → deliberately not fixed here: replicating the module's reaper in the runner duplicates install logic; #950 filed, stack comment points at it.
- **[MEDIUM/PLAUSIBLE, both] `head -n 1` fallback selects bait slot a (e.g. during slot c's maintenance window)** → fixed: slot c or per-run, no sibling fallback.
- **[MEDIUM/PLAUSIBLE, adversarial] Fixture `null_resource` had no `triggers`; a subject change against persistent local state ships a fixture-less eval** → fixed: `subject_cluster` trigger.
- **[MEDIUM/PLAUSIBLE, adversarial] Process-global override 409s any future non-opted-in tofu stack (#516)** → fixed: per-task export gated on the stack declaring the variable; both branches verified.
- **[Blocking ×5, docs-drift] Fleet-addressing exclusivity (3 docs + catalog), ambient-context claims (DRAFTS A5, CUSTOM-TASKS), fleet README missing its second consumer** → all fixed as described under What Changed; the catalog's own description now names §3b as the sanctioned exception.
- **[LOW/CONFIRMED, adversarial] Slot-c safety knowledge lives outside the catalog** → partially accepted: the knowledge is now cross-referenced from the catalog description and the fleet README rather than only a shell comment; a machine-checkable link (a catalog role for the reuse subject) is left for the day a second stack opts in.
- **[LOW/CONFIRMED, adversarial] Selection logic untested** → not fixed: `hack/ci-eval-pr.sh` has no test seam at all today (pre-existing); both selection branches were exercised live above. The regression mode is fail-open to the slower correct path.
- **[Advisory, docs-drift] `docs/README.md:368` map row stale ("awaiting the seeded fleet")** → not touched: pre-existing, outside this diff's scope.
- A final author-context review round (stated as such — no fresh pass ran on it): ambiguous slot-c resolution by sort order contradicted the fleet's own drop-on-ambiguity rule → fixed; unbounded kubectl probe → bounded at 30s; the superseded "Task cluster for this run" echo → reworded as the fallback name.
- The adversarial pass also recorded as non-findings: the `coalesce`/`one()` shape is fail-closed, destroy on reuse has no cloud effect, and the shared-name log-wait overlap is unreachable while Prow `max_concurrency` is 1 (and benign past it — the writes are byte-identical).

## Risk & Rollout

Blast radius is the presubmit and one tofu stack; no runtime code paths, charts, or operator behavior. The failure directions are chosen to be safe: discovery or cleanliness-check errors fall back to today's per-run provisioning (slower, correct), and a task whose stack has not opted in is untouched by construction. New standing exposure, stated rather than hidden: the presubmit points the agent under test at a shared fleet cluster, and a misbehaving agent can dirty it — the cleanliness check converts that from persistent red into per-run fallback plus a loud line for the fleet owner, and the fleet README documents the log-entry residue. Revert = revert two commits; the fleet is untouched either way.


hack/ci-eval-pr.sh
```diff
@@ -121,8 +121,88 @@ fi
...
+# damage to the grave, a standing one keeps it. Check through a throwaway
+# kubeconfig (the ambient context stays untouched); dirty or unreachable
+# means fall back to the per-run cluster and say why, loudly, so the fleet
+# owner cleans it while innocent PRs stay green.
+if [ -n "${SEEDED_TASK_CLUSTER}" ]; then
+  SEEDED_KUBECONFIG="$(mktemp)"
+  SEEDED_LEFTOVER=""
+  if KUBECONFIG="${SEEDED_KUBECONFIG}" gcloud container clusters get-credentials \
+    "${SEEDED_TASK_CLUSTER}" --location "${SEEDED_TASK_LOCATION}" --project "${PROJECT_ID}" --quiet >/dev/null 2>&1 \
+    && SEEDED_LEFTOVER="$(KUBECONFIG="${SEEDED_KUBECONFIG}" kubectl get deployments -n default -o name --request-timeout=30s 2>/dev/null)"; then
```
jayantid: 🔴 **The pre-reuse gate checks the one property slot c is not a fixture for.**

This block is the guard for "a misbehaving agent mutated the shared cluster", and it looks for exactly one shape of damage: Deployments in `default`. That is the shape the gpu task's own `read-only-in-a-post-incident-task` safeguard already catches, on the run that caused it. What makes slot c worth protecting is a different property entirely — `bench/tf/fleet/README.md`'s defect table: *"Master authorized networks absent, normalized to OFF (peers run it ON with an open block)"*, role `drift-outlier`, `cluster_slot: c`.

If the agent under test turns master authorized networks **on** for slot c, that fixture is gone and every layer that could notice is blind to it:

- this gate reads Deployments, not cluster config;
- `fixtures.json` gives `drift-outlier` `probes: []`, and the catalog says in as many words that "a role with an empty list confirms nothing, and its checks can never blame the run for an absence";
- the fleet stack would replan it, but the README says the reconcile workflow "does not exist yet" — reconcile today is a manual `tofu apply` after a *suspected* drift;
- and authorized-networks-ON looks like an improvement, so nobody reverts it on sight.

This is not hypothetical in the "an agent could do anything" sense — it is the specific thing this task invites. Question 3 asks for "recommended architectural remediation steps" on a GKE cluster; `bench/tasks/DRAFTS.md` A5 records that the run holds `roles/container.admin` project-wide and that `FLEET_READONLY_SA` is still missing; `task.yaml` notes the mutating surface is not even enumerable ("config.yaml grants mcp-gke, a Google-hosted server whose remote tools we cannot list"). The catastrophic safeguard exists precisely because agents write against instruction, and this PR is the first change to point one at a fleet cluster by name.

The consequence lands late and misattributed. `consistency-drift-outlier` is still commented out, so a destroyed slot-c fixture is silent until #939/#944 activate it — at which point a 3/3-ON cohort yields no outlier, the check fails, and it reads as a bug in the scenario branch rather than damage several PRs back.

The fix is the same shape as the check already here, and needs nothing from the catalog:

```bash
SEEDED_MAN="$(gcloud container clusters describe "${SEEDED_TASK_CLUSTER}" \
  --location "${SEEDED_TASK_LOCATION}" --project "${PROJECT_ID}" \
  --format="value(masterAuthorizedNetworksConfig)" 2>/dev/null)"
if [ -n "${SEEDED_MAN}" ]; then
  echo "WARNING: seeded cluster ${SEEDED_TASK_CLUSTER} now has master authorized networks configured -- the drift-outlier fixture is gone. Falling back to a per-run cluster; the fleet owner should re-apply bench/tf/fleet." >&2
  SEEDED_TASK_CLUSTER=""
fi
```

Same fail-safe direction as the rest of §3b, one extra API call in a block you measured at 2.9s, and it puts silent fixture loss under the same "loud line + per-run fallback" contract you already chose for the namespace case.

bench/tf/fleet/fixtures.json
```diff
@@ -1,6 +1,6 @@
...
+  "description": "Role catalog for the seeded fleet. A scenario names a fixture ROLE; the runner (hack/fleet-kubeconfigs.sh) discovers the leased project's seeded clusters by label, matches each to a slot below, confirms every object in the role's `probes` list is present, and only then writes that role's kubeconfig. Nothing outside this file may address a seeded cluster or an eval project directly: each project carries its own trio, so a case written against one must run unchanged against all of them. Names, prefix and location are deliberately absent -- they are the Terraform's business and are read back off the discovered clusters. See README.md, 'Addressing a fixture by role'. One sanctioned exception: hack/ci-eval-pr.sh section 3b discovers the slot-c cluster by these same labels and reuses it as the gpu-stress-test log-fixture subject -- it mutates nothing in-cluster, writes only project-level Cloud Logging entries that name the cluster, and falls back to a per-run cluster when slot c is absent, not RUNNING, or its default namespace is not empty.",
```
jayantid: 🟡 **"it mutates nothing in-cluster" is true of §3b and not of the reuse path.**

This clause is the catalog's answer to "can anything outside the role chain touch my fixtures?", and a fleet owner will read it as covering the exception it introduces. But the subject narrows mid-sentence: the *script* mutates nothing in-cluster; the *run* the script sets up hands slot c to a `container.admin`-credentialed agent under test for the length of the eval and then asserts, at `severity: catastrophic`, that it did not create a Deployment — an assertion that only makes sense if writes are expected.

The PR body already words the honest version well: "the presubmit points the agent under test at a shared fleet cluster, and a misbehaving agent can dirty it". Suggest that version here — §3b itself mutates nothing, the agent under test receives credentials for slot c, and the only mutation checked for is Deployments in `default`. That tells the fleet owner the thing they would want to know; the current sentence tells them to stop reading. Related to the 🔴 on `hack/ci-eval-pr.sh`.

hack/ci-eval-pr.sh
```diff
@@ -121,8 +121,88 @@ fi
...
+    if [ -n "${SEEDED_LEFTOVER}" ]; then
+      echo "WARNING: seeded cluster ${SEEDED_TASK_CLUSTER} default namespace holds ${SEEDED_LEFTOVER//$'\n'/, } -- a previous run's agent left it dirty. Falling back to a per-run cluster; the fleet owner should clean the namespace." >&2
```
jayantid: 🟡 **The fallback's only consumer is a human reading a green build's log.**

Every failure direction in §3b resolves to "warn and use the per-run cluster", which is the right *behaviour*. But a dirty slot c is sticky — the agent's Deployment survives the run, and nothing here or in the fleet removes it — so the state persists until somebody acts on this line. The line lands on stdout of a Prow build that **passed**, thousands of lines long, and `dump_prow_artifacts_on_failure` does not apply because a fallback is not a failure.

So the realistic steady state after one bad run is: every presubmit in that project silently pays the ~810s back, indefinitely, and the only symptom is the absence of a speedup nobody measures per run. The guard is correct and unobservable at the same time.

Cheapest fix in the fleet's existing idiom, rather than a new alerting path: `hack/fleet-kubeconfigs.sh` already reports fleet health on demand ("7 role(s) written … 0 whose fixtures were not present") and already resolves slot c. Folding the reuse subject's cleanliness — and the authorized-networks assertion from the 🔴 — into that report gives the fleet owner one command that answers "is my fleet still a fleet?", which is the question this PR makes someone need to ask on a schedule.

bench/tf/fleet/fixtures.json
jayantid, line 3: 🟡 **"it mutates nothing in-cluster" is true of §3b and not of the reuse path.**

This clause is the catalog's answer to "can anything outside the role chain touch my fixtures?", and a fleet owner will read it as covering the exception it introduces. But the subject narrows mid-sentence: the *script* mutates nothing in-cluster; the *run* the script sets up hands slot c to a `container.admin`-credentialed agent under test for the length of the eval and then asserts, at `severity: catastrophic`, that it did not create a Deployment — an assertion that only makes sense if writes are expected.

The PR body already words the honest version well: "the presubmit points the agent under test at a shared fleet cluster, and a misbehaving agent can dirty it". Suggest that version here — §3b itself mutates nothing, the agent under test receives credentials for slot c, and the only mutation checked for is Deployments in `default`. That tells the fleet owner the thing they would want to know; the current sentence tells them to stop reading. Related to the 🔴 on `hack/ci-eval-pr.sh`.

jayantid (approved):

bench/tf/fleet/fixtures.json
haoxuw: I think the original is fine

jayantid (approved):

PR #952, opened 2026-08-25
fix(proxy): harden git against config-driven code execution

bnaylor:
## Summary

`git` is the one executable the credential proxy allows through that picks both its transport and its helper programs out of configuration files, and it reads several of those files off the shared volume the agent writes.  So a `git commit` the sandbox asked for could name a program for the credential container to run, with nothing in the argument vector for the deny policy to match on.  This pins the settings that choose a program, through the environment rather than through a config file, and refuses in argv the options and subcommands that would undo the pins or reach a program by another route.

## Why This Change

The verb allowlist lets `git` through unqualified, and unqualified `git` is not a restricted program.  Five paths measured against git 2.55, three of them on the dev install before this change:

- `git -c protocol.ext.allow=always clone "ext::sh -c id"` ran `id` in the credential container.
- `git config --global alias.zz '!id'` then `git zz` did the same, from outside any repository.
- `git grep -iO/usr/bin/id .` did it on a *read* verb - no lease, no file on the volume, one call.
- `git config man.viewer evil`, `git config man.evil.cmd 'id #'`, `git help -m git` - three ordinary requests, none of which takes a lease, and `id` runs.  `git help -w` is the same bypass through `web.browser` and `browser.<tool>.cmd`, and `git status --help` is the same bypass again with no `help` token in the argv at all.
- `git config trailer.zz.cmd 'id #'` then `git commit -m msg --trailer zz:v` puts the command's stdout in the commit message.  That one lands on `commit`, the argv the skills already send.

Transcripts for the first three are under Live validation.  Without this the sandbox reaches the credentials through configuration, and the policy layer never sees an argument worth refusing.

## What Changed

`agents/platform/scripts/credential_proxy.py` is where the mechanism lives; the rest is its tests and the two credential-isolation docs.

- The executor's subprocess environment now sets `GIT_ALLOW_PROTOCOL=https`, suppresses the system config, pins the global config into the sidecar's own `emptyDir`, pins `core.hooksPath` at an empty mode-0500 directory the agent has no mount for, and turns off the fsmonitor, commit and tag signing, the signing program of all three of git's signature formats, autocorrect and both editors.
- **Two different precedence stories here, and the difference matters.**  `GIT_ALLOW_PROTOCOL`, `GIT_EDITOR`, `GIT_SEQUENCE_EDITOR` and `GIT_CONFIG_NOSYSTEM` outrank every config layer *including* `-c`, so for those the environment is the boundary and the argv check is a backup.  The `GIT_CONFIG_COUNT` pins do **not** - `-c` beats them:

  ```
  GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.hooksPath GIT_CONFIG_VALUE_0=/from-env \
    git -c core.hooksPath=/from-dashc config --get core.hooksPath
  -> /from-dashc
  ```

  For `core.hooksPath` the `-c` refusal is therefore the only control there is, not a redundant second one.  Worth knowing before anyone relaxes it for false-positive relief.
- `git_argument_violation` refuses `-c`, `--config-env`, `--exec-path`, `--git-dir`, `--work-tree`, `--global`, `--system` and `--file`; the subcommands whose job is to run a caller-named command (`bisect`, `difftool`, `mergetool`, `filter-branch`, `send-email`, `instaweb`, `web--browse`, `help`, `fast-import`, `p4`, `svn`, `interpret-trailers`, `submodule foreach`); and the flags that do the same job on an ordinary verb (`--exec`/`-x`, `-O`/`--open-files-in-pager`, `--help`, `--trailer`, `--upload-pack`, `--receive-pack`).  Refusals come back as `SECURITY_POLICY_BLOCKED`, rule `git.argument.refused`.
- Matching covers every spelling git accepts, not just the tidy one: `--flag=value`, the attached and clustered short forms (`-iO<cmd>`), and abbreviations, since git's subcommand options take unambiguous prefixes.  An earlier cut of this refused one spelling and git took three.  `-C` stays allowed - the containment check already resolves it, and the skills use it.  Repository-local `git config` stays allowed, because that is how a clone's commit identity gets set.

`help.autocorrect=0` is not defence in depth, it's a precondition: with autocorrect on from a repo's config, `git bisct run` resolves to `bisect run` and a list comparing whole tokens matches nothing.

**The documentation viewer takes two entries, and that is the interesting part of this PR's history.**  `man.<tool>.cmd` and `browser.<tool>.cmd` carry an arbitrary name, so nothing can be pinned - the trigger has to be refused instead.  `git help -m` and `git help -w` are one trigger and the `help` subcommand covers them.  `git <any-verb> --help` is a second: git dispatches it to the same viewer with the verb still in the subcommand slot, so `git status --help` runs the configured command and no token in the argv is `help`.  I closed the first, wrote in both docs that the vector was closed, and review found the second.  Both are refused now, and the docs say why either alone leaves the other open.  `-h` is not refused - git answers it from the subcommand's own option table and prints usage without starting a viewer.

Same shape for `--trailer` and `trailer.<name>.cmd`, except the flag is the only trigger there, and `interpret-trailers` is refused as belt and braces against a future git growing a second one.  `web--browse` stays on the list on its own account: it is directly invocable and `git web--browse --browser=X <url>` runs `browser.X.cmd`.  It never covered `git help -w`.

**The subcommand match scans every token rather than the subcommand slot, and that is deliberate.**  It costs a collision with ordinary text - `git commit -m help` is refused, though `git commit -m "help me"` is one token and passes.  Position-aware matching would remove the collision and I did not take it, because resolving the slot means knowing which of git's global options take a separated value and this file's list is incomplete: `git --attr-source HEAD help -m git` runs the payload while the runtime's own subcommand resolver reports `HEAD`.  A position-aware check allows the very thing this refuses.  There is a test pinning both halves of that.

**A repository's own `.git/config` is the gap this does not close, and it is now measured.**  That file lives in the shared workspace, so the sandbox writes it directly - no proxied command, nothing in an argument vector to inspect - and the runtime reads it on the next request.  Eight keys still reach a command that way: `diff.external`, `diff.<driver>.command` and `diff.<driver>.textconv`; `filter.<name>.clean`, `.process` and `.smudge`; `merge.<driver>.driver`; and `alias.<name>`.  Six of the eight are reachable with no lease at all - `git status` and `git diff` are the widest routes, and between them they reach the `diff.*` keys and both `filter.*` keys that fire on a modified tracked file - and the two that do need a lease are bounded only in the weak sense that the agent creates its own leases.

There is no clamp for those eight because there is nothing to clamp.  Seven carry an arbitrary name inside the key, so the pinned layer has nothing to enumerate; the eighth has no value meaning "off", which is why `diff.external` is un-pinned.  That is the test for whether a pin is possible, and one key turned out to pass it - see the signing note below.  And refusing repo-local config wholesale is not on the table - remotes, branches and the `user.email`/`user.name` `gitops_workspace.py` sets all live in that file.  The design doc's Limitation section carries the table and the reasoning.

**One key in this class *was* pinnable, and it had been missed.**  `gpg.program` covers git's openpgp format and nothing else.  `gpg.format` is repository-local, each format reads its own program key, so `[gpg] format = ssh` with `gpg.ssh.program` walked past the pin: `git commit -S` and `git tag -s` both executed the payload, argv clean.  `gpg.ssh.defaultKeyCommand` does it with no signing key configured at all, and `x509` has its own key.  All three are pinned now.  This one is closable precisely because the set is closed - three formats, three fixed key names - which is the line between it and the eight above.  `-S`/`-s` stay allowed; refusing them would buy nothing once the programs are pinned.  The trigger is those flags and both verbs are lease-gated; I went looking for a lease-free read route through signature *verification* (`log --show-signature`, `show --show-signature`, `verify-commit`, `%G?`, `%GS`, against a commit carrying a crafted SSH signature header) and found none, so the docs say the trigger is signing rather than implying a read verb reaches it.

One useful thing also came out of measuring the pager.  `core.pager` executes on `git log`, `git diff`, `git show` and `git branch` whenever git has a terminal on stdout, and the only reason it does not execute here is that the executor captures output through a pipe.  That is an accident of how output is captured, not a control anyone declared, so there is now a test that fails the day the executor gives git a terminal.  Pinning `core.pager` would not have helped: `pager.<command>` reaches the same place with an arbitrary name.

**Two more gaps stay open and are written down as open** (`docs/credential-isolation-design.md`, "git configuration").  `filter.<name>.smudge` and `alias.<name>` in a repository's own `.git/config` reach the same place and have no nameable trigger to refuse.  And the refused-subcommand list is a denylist over a set git can add to in any release - `help` and then `--help` turning up in successive reviews is the argument for that being the wrong shape.  Allowlisting the ~20 subcommands the product issues is the structurally right fix and is the follow-up.  What this buys is blast radius, not closure.

`diff.external` is deliberately *not* pinned.  There is no value that means "no external diff" - git tries to execute the empty string - so pinning it broke every `git diff` in an earlier revision while removing nothing from an attacker who can write `.git/config` and use the unpinnable keys anyway.

## Context

**Dependencies: none.  This is independent and mergeable now.**  It applies cleanly to `main` at `b6ff2c5c` and needs nothing else landed first.

Seven sibling PRs follow it.  Together they are the re-cut of #720, which was one 25k-line branch and too big to review; the eight pieces are the increments it was built from, each self-contained.  #720 stays open until they land, then closes.

Open PRs on the same files, so whoever merges second knows to look - neither solves this problem, so these are conflict warnings rather than duplicate work:

- **#913** (@dshnayder, agent shell in a credentialless sandbox pod, draft) - `credential_proxy.py`, its tests, and both credential-isolation docs.
- **#720** - the branch this is cut from.

## Testing

- `make verify` - green (go build, go vet, go test, both Python suites).
- `agents/platform/scripts/test_credential_proxy.py` on its own - 154 tests, OK.  33 of them are new and they drive *real git* rather than a mock, so what they assert is what git did.
- `make docs-check` - green, including the map-row and no-re-aligned-rows checks.
- `prettier --check` - the three changed Markdown files are clean.  (19 files under `.agents/skills/` are flagged on `main` already and this branch doesn't touch them.)
- Every `git` argv the product actually issues, plus the allow-side cases each new entry puts at risk - 46 of them, pulled out of `gitops_workspace.py`, `audit_report.py`, `github_token_refresh.py` and the `scripts/` checkers - run through `git_argument_violation()`: **0 refused.**  32 attack argvs covering all three spelling classes: **0 allowed.**  Nothing shipped issues `help`, `--help` or `--trailer`, and the neighbouring flags that are shipped are not prefixes of the new entries: `--tags` and `--topo-order` against `--trailer`, and `--hard` (which `ensure_workspace` sends on every reset) against `--help`.  Each is pinned as an allow-side assertion, because adding a long option widens the abbreviation match.

### Live validation

**Install:** `bnaylor-ka-test` (GKE, `us-east4`, project `bnaylor-kagents-dev`), namespace `kubeagents-system`, agent image `dev-20260812-184046`.  Policy-only change, so only the `credential-proxy` image was rebuilt: `credential-proxy:dev-20260825-144653`, deployed by setting `CREDENTIAL_PROXY_IMAGE` on the operator so reconciliation renders it instead of overwriting it.  Commands were run in the gateway's `platform-agent` container from `/opt/data`, which is how the sandbox reaches the proxy shims.

Before, after, and after reverting - the same commands each time:

| Command | Before | After | After revert |
| --- | --- | --- | --- |
| `git -c protocol.ext.allow=always clone "ext::sh -c id"` | `id` executed on the proxy | refused, exit 126, `git.argument.refused` | `id` executed again |
| `git config --global alias.zz '!id'` then `git zz` | `uid=10000(hermes) ...` | refused, exit 126 | (not re-run; the config-layer readback below covers it) |
| `git config --glo alias.yy '!id'` (abbreviated) | accepted, alias ran | refused, exit 126 | - |
| `git grep -iO/usr/bin/id .` (read verb, clustered short) | `/usr/bin/id` executed | refused, exit 126 | `/usr/bin/id` executed again |
| `.git/hooks/pre-commit` + `.git/config` `core.hooksPath`, then `git commit` | hook fired, wrote `uid=10000(hermes) ...` | commit succeeded, **hook did not fire** | hook fired again |
| `git clone https://github.com/octocat/Hello-World.git` | exit 0 | exit 0 | - |
| `git clone file:///opt/data/...` | - | `fatal: transport 'file' not allowed` | - |

The `ext::` result reads as a git protocol error, which is the tell:

```
Cloning into '/tmp/before-ext'...
fatal: protocol error: bad line length character: uid=
```

`uid=` is `id`'s stdout arriving where git expected a pkt-line, so the payload ran in the container holding the cloud credentials.  After:

```
`git -c` is refused: it sets configuration that outranks the proxy's own. [...]
policy rule: git.argument.refused
EXIT=126
```

Exit 126 with `policy rule:` is refusal at the proxy.  A server-side error would mean the proxy forwarded it, which is the distinction the test turns on.

The pinned config layer, read back from inside the sidecar on the new image:

```
core.hooksPath=/var/lib/credential-proxy/git-hooks-disabled
core.fsmonitor=false
commit.gpgsign=false
gpg.program=false
help.autocorrect=0
tag.gpgSign=false
GIT_EDITOR=false
```

`dr-x------ 2 hermes hermes` on the hooks directory, empty, and `No such file or directory` for the same path from the sandbox container.  On the reverted image `core.hooksPath` resolves to `/opt/data/scratch/lt-after/evil-hooks` - the value the test wrote into the repo's own `.git/config` - which is the thing the pin was overriding.

Also observed: `git -c user.email=... commit` is refused, and the shipped path (repo-local `git config user.email`, `gitops_workspace.py`) still works and commits.  CR stayed `Ready=True (Reconciled)`; the two reconciler errors in the operator log are the benign "object has been modified" that follows a nudge annotation.

**Four commits after the live run were not re-run on the cluster.**  The `help`, `--help` and `--trailer` refusals (`59e38181`, `95fd3cdd`) are dictionary entries in a pure function of argv; the signature-format pins (`1d53734c`) are three more entries in the environment the sidecar builds for each subprocess; `b4351d4e` is documentation.  All the underlying vectors were reproduced against git 2.55 under this branch's exact pinned environment, several of them driving the real `CommandExecutor` - transcripts are in those commit messages - and all are covered by the argv sweep and the unit tests.

Nothing in those four changes the rendered Pod, so the Pod-level observations above still describe the head.  **The config readback block above does not**: it was captured before `1d53734c` and lists the six pinned keys of that build, not the nine the head now sets.  `gpg.ssh.program`, `gpg.ssh.defaultKeyCommand` and `gpg.x509.program` are missing from it, and `test_every_forced_config_key_reaches_git` is what asserts all nine reach git - through `git config --get`, so the count, the key names and the values are all checked by the program that consumes them.

**Not covered.**  The `--upload-pack` / `--receive-pack` refusals were exercised as unit tests only - reaching them live needs `GIT_ALLOW_PROTOCOL` widened to include `file`, and I'm not disabling a live control to demonstrate one.  Authenticated push and fetch against the install's own GitOps repo weren't exercised either: that install's clone still carries an `ssh://` remote from before `gitops_workspace.resolve_repo` normalised remotes to https, and its credential helper isn't wired, so the benign-clone control used a public https repo instead.  Nothing on `main` builds a non-https remote (`ensure_workspace` constructs `https://github.com/{repo}.git` and `remote set-url`s every clone to it), but an operator who hand-configures an `ssh://` remote will now get `transport 'ssh' not allowed` - worth knowing at merge time.

**The repository-local config gap was confirmed live on the hardened image**, because a limitation is worth no more than the measurement behind it.  From the sandbox container, appending `diff.external` straight to a workspace clone's `.git/config` - a plain file write, no proxied command - and then issuing an ordinary `git diff` through the proxy ran the payload in the credential container.  `core.pager` in the same file on the same read verb did not run, with or without `--paginate`, which is the no-terminal result above holding on a real install.  Worth saying plainly: on this install both containers still run as the same UID, so there is no boundary being crossed there - the measurement is about what the config layer does, and it is why the Limitation section says what it says.

**Cleanup.**  `CREDENTIAL_PROXY_IMAGE` removed from the operator, sidecar reconciled back to `dev-20260812-184046`, gateway `4/4 Running`, CR `Ready=True`, operator env back to `ENABLE_WEBHOOKS` alone.  Test clones, the probe hook and the global-config aliases are gone.  Left behind: the `credential-proxy:dev-20260825-144653` image in Artifact Registry, and `:latest` in that dev repository now points at it - nothing in the install consumes that tag, since the operator derives the sidecar tag from the agent image.

## Self-Review

**How the passes ran.**  My own reads happened in the session that wrote the change, so I am not claiming the fresh-context pass AGENTS.md asks for.  Reviewers who had not written the code then read the branch across four rounds and found twelve things; they are addressed in `59e38181`, `95fd3cdd`, `692e09ed` and `1d53734c` and folded in below.  @kyber775 also commented approvingly on an earlier head.

- **Over-refusal - does the policy break shipped `git`?**  This is the failure mode the design chooses: it scans the whole argv rather than guessing where git's options end, so a literal `-c` in a commit message gets refused.  Checked by enumerating all 34 `git` argvs in the tree and running each through `git_argument_violation()`.  Zero refused.  The one that looked risky, `git add -- clusters/prod`, passes because `--` is special-cased out of the abbreviation match; without that it prefixes every long option and matches the first.  There is a test on it.  Re-run after the `help`/`--trailer` additions: still zero.
- **Under-refusal - does every spelling git accepts get caught?**  24 attack argvs across the three spellings (`--flag=value`, attached/clustered short, abbreviated), all refused.  The abbreviation case is the one that bites: git's *own* options (`--git-dir`, `--exec-path`, `--config-env`) are parsed by hand with exact comparisons and are not abbreviable, so testing only those suggests the problem does not exist, while `git config --glo` and `git rebase --exe` both run.  Confirmed refused live, not just in tests.
- **Under-refusal, found by review, fixed twice:** `git help -m` and `git commit --trailer` both reached arbitrary execution and both were allowed (`59e38181`).  The `help` one was already implied by the code's own comment, which named `help -w` as a place git keeps a command in configuration.  Then a second pass found `git <verb> --help` reaching the same viewer with no `help` token in the argv, which my first fix missed and which both docs had by then described as closed (`95fd3cdd`).  A wrong claim in shipped docs is worse than the hole, so the docs and the entry landed together.  Four tests across the two rounds.
- **A pin that was one key short, found by review, fixed:** `gpg.program` disabled signing for the openpgp format only, and both docs claimed signing was disabled outright.  `[gpg] format = ssh` plus `gpg.ssh.program` reached a command through `git commit -S`.  Three keys added to the pinned layer, the wrong sentence corrected in the same commit, and a test per spelling (`1d53734c`).  This is the one key in the repository-local class that turned out to be pinnable, and finding it is what sharpened the "closed set or not" test the limitation section now applies.
- **A false claim in my own limitation table, found by review, fixed:** it said none of the table's triggers takes a lease.  Measured per verb against `git_lease_violation`: `add`, `checkout`, `stash`, `commit`, `merge` and `cherry-pick` are gated; `status`, `diff`, `log -p`, `blame` and `show` are not.  The table now has a lease column with the measured value in it.
- **A test of mine that was partly vacuous, found by review, fixed:** the `core.pager` fixture had no commits, so its `git log` subTests would have passed on a pty as well.  Seeded a commit and asserted rc 0, so all three subTests can now fail.
- **A reported vector that did not reproduce, and what measuring it found anyway:** review raised `core.pager` in a workspace `.git/config` as execution on a read verb.  It does not fire - git pages only with a terminal on stdout and the executor uses a pipe - and I checked that on a pty as well, where it does fire, so the negative is a measurement rather than an absence.  Enumerating the rest of the class found six keys that *do* still execute, none of them closable by the pinned layer.  No clamp; the Limitation section now carries the table, and the accidental control has a test (`692e09ed`).
- **A parser differential I introduced and then declined:** the obvious tidy-up for the `git commit -m help` collision is to match the subcommand slot instead of every token.  Measured before doing it, and it is unsafe here - `git --attr-source HEAD help -m git` executes while the slot resolver reports `HEAD`.  Kept whole-argv scanning, documented the collision next to the entry, and added a test that pins the differential so nobody re-derives the tidy-up later.
- **A false precedence claim in this PR body, found by review, fixed:** the first "What Changed" bullet used to say the pinned settings outrank every config layer "including `-c`".  True of the dedicated variables, false of the `GIT_CONFIG_COUNT` layer, and the difference is load-bearing - it is the reason the `-c` refusal is the sole control for `core.hooksPath` rather than a redundant backup.  The code and both docs had it right; only the body was wrong.  Corrected above, with the measurement.
- **Internal identifiers in shipped comments, found by review, fixed:** four comments cited a rule id from a private log and a slice name that does not exist in this repository, and one read as the architecture doc's `C2` (the Platform Agent) making a claim about failure direction.  All four now state the rule in plain words.
- **Stale identifiers in the new comments.**  Two cite source lines - `gitops_workspace.py:548` for `clean -fdq` and `:594` for repo-local identity.  Both checked against this branch; 548 is exact, 594 lands on the docstring sentence making the claim.  Correct, but line numbers rot, so a reviewer who wants them removed will get no argument from me.
- **Fidelity of the docs to the code.**  The design doc and the site page both enumerate the refused options and subcommands.  Read against the constants rather than against each other, and re-read after the two new entries landed.
- **The `latest` tag.**  Not a diff finding - `cloudbuild.yaml` tags `:latest` alongside the versioned tag, so the validation build moved it in the dev registry.  Recorded under Cleanup rather than quietly left.
- **Not fixed, deliberately - and this is the structural one.**  `filter.<name>.smudge` and `alias.<name>` have no nameable trigger to refuse the way `help` and `--trailer` did, and the refused-subcommand list is a denylist over a set git can extend in any release.  Four same-class paths have now turned up across successive reviews; the third only after the second had been fixed and documented as closed, and the fourth does not go through an argument vector at all.  That is about as clear as evidence gets that enumeration is the wrong shape here.  The fix is to allowlist the ~20 subcommands the product issues and fail closed on the rest - the design doc already recommends it.  That is a follow-up PR, not this one: it changes the denylist-not-allowlist decision recorded above `GIT_MUTATING_SUBCOMMANDS` and needs its own argument.  What lands here closes every path measured, and says plainly that it is blast radius rather than closure.
- **Not fixed, deliberately:** the credential-helper half of the `.git/config` gap.  It is closable - reset the helper list in the pinned layer, reinstate the proxy's own immediately after - but that couples the proxy to whatever value `gh auth setup-git` writes, and I would rather do it with the content-passing work than bolt it on here.

## Risk & Rollout

Fail-closed: an over-tight refusal breaks an agent `git` call rather than permitting an escape, and the refusal message names the rule and says what to use instead.  Blast radius is every proxied `git`, so the thing to watch after merge is a skill hitting `git.argument.refused` on an argv nobody enumerated.  Revert is the two commits - the policy is one function and one environment block, with no state anywhere.  Nothing to do at merge time.

Two behaviour changes that are not refusals of an attack.  `GIT_ALLOW_PROTOCOL=https` means non-https remotes stop working through the proxy; nothing on `main` builds one, but see Live validation.  And the subcommand list is matched against every token, so a commit message that is exactly `help` or exactly `foreach` is refused - `git commit -m "help me"` is one token and passes.  Nothing shipped issues either.


kyber775:
## Security & Day-2 Maintainer Review: PR #952

**Verdict:** 🟢 **Approved (Critical Security Hardening)**

### 🛡️ Threat Model & Defense In Depth Analysis

The credential proxy sidecar holds privileged cloud (`GOOGLE_APPLICATION_CREDENTIALS`) and GitHub tokens. Because the sandbox agent writes files to the shared volume, unrestricted `git` execution allowed the agent to weaponize `.git/config` and command-line flags to execute arbitrary code inside the credentialed sidecar container (via `protocol.ext`, `core.hooksPath`, `core.fsmonitor`, `gpg.program`, `diff.external`, `alias`, and pager flags like `-O`).

This PR establishes an airtight defense-in-depth boundary:

1. **Forced Environment Layer (Highest Precedence):**
   - Setting `GIT_ALLOW_PROTOCOL=https` eliminates arbitrary command execution via `ext::`, `file://`, and `ssh://` protocols. It outranks configuration file options.
   - `GIT_CONFIG_COUNT` forces `core.hooksPath` to an empty mode-0500 directory (`/var/lib/credential-proxy/git-hooks-disabled`), disarms `fsmonitor`, `gpgsign`, `gpg.program`, and forces `help.autocorrect=0`.
   - `GIT_EDITOR=false` and `GIT_SEQUENCE_EDITOR=false` prevent spawning editors on interactive-prompt fallbacks.

2. **Strict Argv Validation (`git_argument_violation`):**
   - Refuses configuration overrides: `-c`, `--config-env`, `--exec-path`, `--git-dir`, `--work-tree`, `--global`, `--system`, `--file`.
   - Refuses code-execution subcommands: `bisect`, `difftool`, `mergetool`, `filter-branch`, `send-email`, `instaweb`, `web--browse`, `p4`, `svn`, `fast-import`, `foreach`.
   - Refuses program execution flags on ordinary verbs: `--exec`/`-x`, `-O`/`--open-files-in-pager`, `--upload-pack`, `--receive-pack`.
   - Correctly handles attached arguments (`-O<cmd>`), clustered short flags (`-iO<cmd>`), and ambiguous prefix abbreviations (`--glo`, `--exe`).

3. **False-Positive & Operational Safety Audit:**
   - `-C` remains allowed (safely validated by the containment check).
   - Local repository `git config` remains allowed (used to set committer identity in `gitops_workspace.py`).
   - The `--` end-of-options separator is explicitly special-cased to prevent abbreviation collisions with pathspecs (`git add -- <path>`).
   - Verified that all 31 `git` invocations across the repository pass without false refusals.

4. **Testing & Live Verification:**
   - 27 new tests driving real `git` in `test_credential_proxy.py` (148 tests total passing).
   - Live before/after validation against `bnaylor-ka-test` GKE cluster demonstrating exact exit 126 / `git.argument.refused` behavior.

Clean, rigorous security hardening that closes critical escape vectors without breaking shipped skill operations.

bnaylor:
@kyber775 thanks for the read. One correction, because it came from a wrong sentence in my PR body rather than from the diff, and I would rather it not become the shared understanding.

Your summary says "Forced Environment Layer (Highest Precedence)". That is true for the dedicated variables and false for the `GIT_CONFIG_COUNT` pins. `-c` beats `GIT_CONFIG_COUNT`:

```
GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.hooksPath GIT_CONFIG_VALUE_0=/from-env \
  git -c core.hooksPath=/from-dashc config --get core.hooksPath
-> /from-dashc
```

Only `GIT_ALLOW_PROTOCOL`, `GIT_EDITOR`, `GIT_SEQUENCE_EDITOR` and `GIT_CONFIG_NOSYSTEM` outrank `-c`. So for `core.hooksPath` the `-c` refusal in `_GIT_REFUSED_ARGUMENTS` is not defence in depth, it is the only control there is. It matters because the wrong version makes the `-c` refusal look like a redundant backup that could be relaxed if it ever causes false positives. The code and both docs had this right; the PR body did not, and it is fixed now.

Also worth flagging since you reviewed against the earlier head: a separate review found `git help -m` and `git commit --trailer` both reaching arbitrary execution through `man.<tool>.cmd` and `trailer.<name>.cmd`, neither of which any pin can reach because the key carries an arbitrary name. Both are refused as of `59e38181`, with a test each. The `git help` one got past the first cut even though the code comment named `help -w` as a risk: the list refused `web--browse`, which `git help -w` reaches internally without the token ever appearing in the argv, so that entry was doing nothing.

No action needed from you. Flagging so the record is right.

dshnayder (approved):

PR #961, opened 2026-08-25
feat(security)!: remove the gke-admin permission set and ship the agent-RBAC admission backstop

bnaylor:
## Summary

Removes the `gke-admin` GCP IAM permission set, and ships the two agent-RBAC `ValidatingAdmissionPolicy` objects that have been sitting in `examples/gitops-repo/policy/` where nothing ever applied them.  Both halves are about the same thing: a permission model whose enforcement does not actually reach.  `read-only` stays the default and `custom` stays the way to widen, which now costs naming each role.  This is a breaking change for any install carrying `PLATFORM_AGENT_PERMISSION_SET=gke-admin` or `permission_set = "gke-admin"`; migration is below.

## Why This Change

GKE authorizes an action "if they have sufficient permissions according to _either_ tool" -- IAM or Kubernetes RBAC ([Access control](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control)).  The agent's Kubernetes RBAC is read-only, but that only constrains one half of that union.  A GSA holding `roles/container.admin` is authorized by IAM whatever the KSA says, so under `gke-admin` the read-only Kubernetes footprint the rest of the permission model rests on stopped constraining anything the agent reached through that identity.

`roles/container.admin` is also the one predefined GKE role carrying `container.clusters.impersonate` (`roles/container.clusterAdmin` does not -- checked with `gcloud iam roles describe`).  GKE grants IAM roles at the project level; a cluster is not a resource an IAM policy attaches to, and the same page says "an IAM role grants privileges across all clusters in the project".  So the impersonation cannot be narrowed to one cluster.

That is why the set is removed and not deprecated.  It did not widen the ceiling, it removed one, and it did so behind a single word in a menu.  `custom` still gets you there if you need it, but you have to name each role, which puts the grant somewhere a reviewer can see it.

The admission policies are a smaller story with the same shape.  They deny agent RBAC that grants a write verb, grants Secrets, or gives a namespace-tier agent ServiceAccount a cluster-scoped binding.  They lived in the GitOps example repository, which is a template customers fork; nothing in this repository applied them, and a policy that is not installed is not a control.

## What Changed

- **`gke-admin` is gone from every install path.**  `installer_common.sh` gains one shared gate, `require_supported_permission_set`, and `install.sh` and `common.sh` both route through it rather than keeping their own copies of the rule.  A value that still names the removed set gets an explanation rather than a bare "invalid", because the value most likely to appear is a cached `vars.sh` or an RC environment variable written before the removal.
- **The Terraform composition drops `local.gke_admin_roles`.**  `local.agent_project_roles` no longer branches on `permission_set` at all -- it takes `project_roles` or the read-only bundle -- so no value of that variable, removed or invented later, can select an admin list.  `variables.tf` refuses `gke-admin` with the same reasoning in its `error_message`.
- **`permission_set` the variable survives even though it now selects nothing.**  `installer_common.sh` writes it into every generated `terraform.tfvars`, so deleting it would fail every installer-driven apply on an unsupported argument.  There is a comment in `variables.tf` saying so.
- **The policies move to `k8s-operator/config/admission/agent-rbac-policy.yaml`**, and the chart ships a generated copy behind `admissionPolicy.enabled`, default on, **and behind the cluster's version**.  `Chart.yaml` accepts `>=1.29.0-0` but `ValidatingAdmissionPolicy` only reached `v1` in 1.30, so the template also checks `.Capabilities.KubeVersion` and renders nothing below that.  Without it a default `helm install` on a supported cluster fails on `no matches for kind`, and through Terraform that fails the whole apply rather than just the policies.  `validate.yml` renders at `--kube-version 1.29.0` and fails if anything comes out.  `make chart-sync` generates it and `make chart-check` holds it in step, the same contract the CRD and RBAC copies already had.
- **They stay outside the kustomize overlay deliberately.**  `namePrefix` rewrites a policy's `metadata.name` and not the `spec.policyName` its binding refers to, so a kustomize build yields bindings pointing at nothing.  That is accepted at apply time and silently inert.  `make deploy` therefore does not install them, and INSTALL.md Method 2 Step 4 now has the `kubectl apply` line.
- **The file's header says what the policies do not cover**, and the site's Security & IAM page repeats it: they cannot check the rules of a Role a binding merely _references_ (CEL cannot read another object), and the content policy only selects manifests carrying the `kube-agents/tier` label.  The binding-scope policy is not label-evadable, since it keys on the ServiceAccount being privileged.
- **The binding-scope policy exempts what the operator mints.**  `reconcileRBAC` unconditionally creates `kubeagents:minimal:<ns>:<name>` as a ClusterRoleBinding whose subject is `spec.security.serviceAccountName`; the default, `kubeagents-platform-agent`, ends in `-agent`, so the policy evaluates every reconcile on every install.  An agent configured with the namespace-tier ServiceAccount name from the shipped GitOps example would have had that reconcile denied before the Deployment existed.  The operator has no namespace-tier path, so its own reconcile is exempt -- keyed on `request.userInfo.username`, the authenticated user the API server puts in the AdmissionReview.

  Keying that on a label the controller stamps, as a first attempt did, is not an exemption but a hole: `matchConditions` are ANDed, so a false one drops the object from the policy altogether, and anyone who set that label on their own `ClusterRoleBinding` walked straight past the denial.  Nothing in the object can set `userInfo`.  It needs two usernames, because the two install paths run the controller as different ServiceAccounts -- kustomize gives it `kubeagents-controller` in `kubeagents-system`, the chart gives it `<release>-operator-sa` in the release namespace -- and `hack/sync-chart-manifests.sh` rewrites the second to the rendered names so the chart's copy names the identity that release actually reconciles as.
- **`custom` with an over-reaching role list now warns.**  `--custom-roles="roles/container.admin"` reaches IAM-identical authority to the bundle this PR removes.  The argument for `custom` is that naming each role puts the grant in front of a reviewer, and on the installer path that is not true: the roles land in a machine-generated `terraform.tfvars` nobody opens.  `warn_on_overreaching_custom_roles` says so at the point of choice, sharing one role list with the ceiling test rather than keeping a second copy.  It warns; it does not refuse.  Called from all three places the roles can arrive: the `--custom-roles` flag and both interactive prompts, which collect them well after the flag path has already been decided.
- **The refusal normalises its own input.**  `common.sh` trimmed and lowercased before calling the gate and `install.sh` did not, so `--permission-set=GKE-ADMIN` got the generic "invalid" instead of the explanation.  Normalising inside the gate fixes both callers at once.
- **Two new suites.**  `tests/test_agent_iam_ceiling.py` executes the real provisioning shell and reads the Terraform composition structurally; `tests/test_admission_policy_shipped.py` asserts the policies reach a cluster on each install path that can carry them.  `validate.yml` renders the chart with helm and counts the four objects, which is the half a text assertion cannot do.

### What the admission policies do not cover

Stated here as well as in the file header, because a backstop described more broadly than it is will mislead someone:

- **On a default install with no GitOps overlay, the content policy matches nothing.**  It selects on the `kube-agents/tier` label, and `commonLabels()` stamps the four `app.kubernetes.io/*` labels and never a tier, so none of `buildMinimalPlatformRole`, `buildPlatformLocalRole` or `buildPlatformLeaderRole` is examined.  It is a backstop on the RBAC a customer's overlay writes, not on the RBAC the operator mints.  I did not close that by stamping the label on operator RBAC, which would subject the operator's own objects to the read-only validation and could deny legitimate reconciles; whether that is the right end state is a separate question from this PR.
- **Neither policy can check the rules of a Role a binding merely references.**  CEL sees only the object being admitted.  An unlabelled write Role bound to an agent ServiceAccount is admitted.
- **The binding-scope policy denies exactly one ServiceAccount name.**  Put together with the first point, that means a hand-written unlabelled `ClusterRole` granting `*` on `*`, bound to the agent's own ServiceAccount, is **admitted by both policies** on a default install with both enforcing.  Pull-request review on the GitOps repository is the control there, which is what the branch-protection guidance shipped beside the example overlay is for.
- **The binding-scope policy only examines `kind: ServiceAccount` subjects.**  A `kind: Group` subject such as `system:serviceaccounts:team-x` is not selected at all.  The subject is unforgeable for the kinds the policy looks at; choosing another kind steps outside them.

### Migration

An install carrying `gke-admin` now fails rather than granting the role.  Set `PLATFORM_AGENT_PERMISSION_SET=read-only` in `vars.sh` (or `--permission-set=read-only`, or `permission_set = "read-only"` in `terraform.tfvars`), or `custom` with the roles named explicitly if you accept the risk.

Re-running the install does not strip roles it no longer grants, so an existing GSA keeps `roles/container.admin` until someone removes it.  The commands are on the site's [Security & IAM](https://github.com/gke-labs/kube-agents/blob/main/docs/site/src/content/docs/reference/security-and-iam.md) page under "Configuring read-only (auditing) mode".

`scripts/release/README.md` carries a note for the RC environments specifically, because `provision_rc_environment.sh` runs `uninstall.sh` before `install.sh`.  The refusal fires while `install.sh` is collecting configuration, which is before `terraform apply` but after the teardown.  Expect a torn-down RC environment that was not rebuilt, not a run that refused to start.

## Context

This is one of four independent PRs re-cutting an already-merged fork increment into reviewable pieces.  The others are #952, #955 and #960.  **This one is independent of all three and mergeable in any order relative to them** -- the only overlap is documentation prose, so a conflict there is textual.

Two provisioning scripts this work was originally written against (`provision_03_gcp_gke_operator.sh` and `provision_04_gcp_iam.sh`) were deleted by #797, which made Terraform + Helm the only install engine.  Their intent is re-homed rather than replayed: the role bundles moved to the composition's locals, and the admission-policy apply became the chart template instead of a second `kubectl apply` in a script.  AGENTS.md forbids the second expression, and the chart is what the engine installs, so there is no script half to port.

## Testing

- `make verify` green (go build, vet, test, both Python suites).
- `make test-python` green: 391 tests in `tests/`, all suites OK.
- `make docs-check`, `make chart-check`, `make validate` green.
- `prettier --check` (3.9.6) green on the changed files.
- `terraform validate` and `terraform fmt -check` green on `examples/full-install`.
- `helm lint` and `helm template` (v3.16.3) green; a default render carries both policies and both bindings, and `--set admissionPolicy.enabled=false` renders none.

**Both new suites were run against unmodified `main` first and both fail there**, which is the point of them:

```
$ git rev-parse HEAD
b6ff2c5c3b0f9e8ba915fe473e191eb3e2e558b1
$ python3 -m unittest discover -s tests -p 'test_agent_iam_ceiling.py'
Ran 17 tests in 0.988s
FAILED (failures=16)
$ python3 -m unittest discover -s tests -p 'test_admission_policy_shipped.py'
Ran 13 tests in 0.011s
FAILED (failures=4, errors=8)
```

After the change:

```
$ python3 -m unittest discover -s tests -p 'test_agent_iam_ceiling.py'
Ran 18 tests in 0.759s
OK
$ python3 -m unittest discover -s tests -p 'test_admission_policy_shipped.py'
Ran 13 tests in 0.041s
OK
```

### Live validation

**Install:** `bnaylor-ka-test`, GKE `v1.35.6-gke.1710000`, `us-east4`, project `bnaylor-kagents-dev`, namespace `kubeagents-system`.  The `live-test-lease` ConfigMap was free this round; I took it as `up10-gke-admin-removal` and released it at the end.

**One deviation from the shipped objects, stated up front.**  The policies were installed with both bindings set to `validationActions: ["Warn"]` rather than `["Deny"]`.  Everything else, including `failurePolicy: Fail` and every CEL expression, is byte-identical to the source file.

Why: applying the `Deny`-bound objects was refused by my tooling's safety check, and on reflection that was the right call -- a cluster-scoped `Deny` policy with `failurePolicy: Fail` over every RBAC write, on an install other people are working on, is exactly the change that guard exists for.  `Warn` cannot reject a request, so it carries none of that risk.  Only the action differs: the expressions evaluate identically and the messages are the shipped ones, so what warns here a `Deny` binding rejects.  Two things make the run more than a formality rather than less -- a warning was emitted on every case that should be denied, and none at all on the cases that should be admitted, which together also rule out an evaluation error being masked by `failurePolicy: Fail`.  **I did not observe a rejection and am not claiming one.**

**Baseline first, so the "after" is a comparison.**  Before installing anything, the cluster had none of these policies, and a labelled `ClusterRole` carrying a `create` verb applied clean.

**The three denials fire, with the shipped messages:**

```
# labelled ClusterRole with a write verb
Warning: Validation failed for ValidatingAdmissionPolicy 'kube-agents-agent-readonly': agent RBAC
may grant only read verbs (get/list/watch) — no write or privilege-escalation verbs ...

# labelled ClusterRole reaching secrets
Warning: Validation failed for ValidatingAdmissionPolicy 'kube-agents-agent-readonly': agent RBAC
may not grant Secrets (token/credential exfiltration risk) ...

# hand-written ClusterRoleBinding to developer-team-agent, no managed-by label
Warning: Validation failed for ValidatingAdmissionPolicy 'kube-agents-agent-binding-scope':
developer-team is a namespace-scoped tier and may not be granted a ClusterRoleBinding (wrong-scope)
```

That last one is also what shows the operator exemption did not gut the policy: a hand-written binding to the same ServiceAccount still trips it.

**The exemption is the authenticated identity, and that is testable directly.**  One object -- the operator's live `kubeagents:minimal:*` binding with its subject changed to `developer-team-agent`, so it is exactly the case the rule denies -- submitted as three different users:

| Applied as | Result |
| --- | --- |
| `bnaylor@google.com` (me) | **warns** (wrong-scope) |
| `system:serviceaccount:kubeagents-system:kubeagents-controller` | admitted, no warning |
| `system:serviceaccount:default:kubeagents-controller` (same name, other namespace) | never reached admission -- RBAC refused the create |

The object is byte-identical across all three, so the exemption is the identity and nothing in the manifest.  The third row is not the clean result I wanted: a lookalike ServiceAccount cannot create `ClusterRoleBinding`s at all, so Kubernetes' own RBAC stops it before admission runs.  Defence in depth working as intended, but it means the namespace half of the exemption is asserted by the CEL and by a test rather than observed.

**And the forgery that used to work does not.**  An earlier revision exempted on a label; a `ClusterRoleBinding` to `developer-team-agent` carrying `app.kubernetes.io/managed-by: platformagent-controller` was admitted silently by it.  Against the shipped version that object warns, identically to the same object without the label.

**And a real reconcile, not just an apply.**  With both policies installed, nudging the CR's `validation.kubeagents.test/nudge` annotation drove a full reconcile through step 3b:

```
INFO Reconciling PlatformAgent {"name":"platform-agent","namespace":"kubeagents-system", ...}
```

No admission error in the controller log, and the CR settled at `Ready=True (Reconciled)`.  This install's controller runs as `kubeagents-controller`, which is the first username in the exemption.  Its agent uses the default ServiceAccount name, so this shows the policies cost a real reconcile nothing; the row above is what shows the exemption firing on the case that would otherwise be denied.

**Reverted, and confirmed reverted.**  Deleted all four policy objects and the nudge annotation; released the lease.  Every probe object was submitted with `--dry-run=server`, and `kubectl get` confirms none persisted.  After removal the forged-label attack object applies with no warning again, so the control is demonstrably gone rather than assumed gone.  Zero `kube-agents-agent-*` policies, zero `up10-*` objects, no nudge annotation, the operator binding intact, CR still `Ready=True`.

**IAM, read-only and unchanged.**  The agent GSA already holds only the read-only set (`container.clusterViewer`, `container.viewer`, `iam.securityReviewer`, `iam.serviceAccountUser`, `logging.viewer`, `mcp.toolUser`, `monitoring.viewer`, `recommender.containerDiagnosisViewer`) -- no `container.admin` -- so this change moves nothing on this install.  Read before anything else, and nothing was applied.

**`terraform plan` against the real project, plan only** (`-lock=false`, local state).  Comparing planned `google_project_iam_member.agent_roles` on `upstream/main` with `permission_set = "gke-admin"` against this branch's default:

```
- roles/container.admin
- roles/container.clusterAdmin
+ roles/container.clusterViewer
+ roles/container.viewer
- roles/monitoring.admin
+ roles/monitoring.viewer
```

Nothing else moves: the base's `read-only` plan and this branch's `read-only` plan are identical.  Terraform also refuses the removed value from a real plan (`var.permission_set is "gke-admin"`).

**The shell refusal, through the real shell.**  A `vars.sh` naming `gke-admin`, sourced and run through `init_var_platform_agent_permission_set`, exits 1 with the explanation; the same file with `read-only` is accepted.  `GKE-ADMIN`, `Gke-Admin` and a whitespace-padded spelling now all reach the explanation rather than the generic error.  `--custom-roles="roles/container.viewer,roles/container.admin"` emits the over-reach warning and continues; a benign list is silent.

**What is still not covered.**  A `Deny`-bound rejection on a live cluster, for the reason at the top of this section.

## Self-Review

**Context each pass ran in, plainly: both passes ran in the session that wrote the change.**  Subagent dispatch was closed for this task, so I did not get the fresh-context read AGENTS.md asks for.  Weigh the list below accordingly -- it is the author checking his own work, which is exactly the configuration the rule says does not reliably catch what the reasoning caused.

**Two review rounds have been over this branch and found six defects I missed, three of which would have broken installs -- one of them introduced by my own fix in the first round.**  All four are fixed in `800b8e9e` and folded into the list below rather than appended to it.  Recording it plainly because it is the honest measure of what the self-review above was worth: the version gate and the operator-RBAC contradiction are exactly the class of thing a context that did not write the change catches and the one that did does not.

What was looked for, and what came of it:

- **A fix that broke the caller it was meant to help.**  Round one moved permission-set normalisation inside `require_supported_permission_set` so every spelling reaches the right message.  That was right for the message and wrong for `install.sh`, which passed the flag raw and *kept* it raw -- so `--permission-set=Custom` started clearing a gate that used to reject it, then missed the custom-roles requirement, the over-reach warning added in the same round, both `write_state_var` lines, and terraform's case-sensitive `contains()`, failing much later in the apply with an error headlined about `gke-admin`.  A clean one-line rejection became a confusing late failure.  `install.sh` now normalises and keeps the normalised value.  The test extracts that line and executes it, so a reassignment that does not actually lowercase fails too.
- **An exemption that was a bypass.**  The operator carve-out added in round one keyed on a label the controller stamps.  `matchConditions` are ANDed, so a false one removes the object from the policy entirely -- meaning anyone could set that label on a `ClusterRoleBinding` to `developer-team-agent` and be admitted.  My own round-one evidence table contained the attack and recorded it as proof the exemption worked.  Now keyed on `request.userInfo.username`, which no manifest can carry, verified live from three identities.  I have added a note beside `PolicyContentTest` about why no test caught this: with no CEL runtime in the repository, the assertions read the expressions as text, so the failure mode a reader most expects them to guard is exactly the one they cannot see.
- **A warning on a door most people do not use.**  `warn_on_overreaching_custom_roles` was called only on the `--custom-roles` flag path; both interactive prompts collect the roles 44 and 65 lines later, so anyone running `./install.sh` and typing a role in got nothing.  Called from all three now.  Related, and stated because my round-one report got it wrong: `common.sh`'s `init_var_platform_agent_permission_set` has no caller anywhere in the repository -- the numbered provision scripts that invoked it went with #797.  I described it as "the prompt path"; it is dead code.  Left in place (removing it is not this PR's business) but no longer counted as a door, and the stale comment in `install.sh` pointing at it now says so.
- **A default install failing on a cluster the chart says it supports.**  `Chart.yaml` declares `kubeVersion: ">=1.29.0-0"`; `ValidatingAdmissionPolicy` is `v1` only from 1.30; the values gate defaults on.  So `helm install` on 1.29 died on `no matches for kind`, and through Terraform that failed the whole apply.  Documenting "turn it off below 1.30" put the burden on a reader with no reason to look.  Fixed by gating the template on `.Capabilities.KubeVersion` as well, rather than raising `kubeVersion`, which would drop 1.29 for the whole chart -- a bigger statement than this needs.  `validate.yml` now renders at 1.29 and fails if anything comes out.
- **A policy that denies RBAC the operator itself creates.**  The binding-scope policy hard-codes `developer-team-agent`, and `reconcileRBAC` unconditionally mints a ClusterRoleBinding whose subject is `spec.security.serviceAccountName` -- which is exactly that name in the shipped `examples/gitops-repo/policy/rbac-overlay/developer-team.yaml`.  A customer following the shipped example had their reconcile denied at step 3b, before the Deployment existed, with an admission error nobody would connect to the example they copied.  The policy asserted an invariant the controller violates by construction.  Fixed by exempting bindings carrying `applyManaged`'s `managed-by` label; verified live, both directions.
- **Docs claiming coverage the policy does not have.**  The page said the policies reject agent RBAC "whoever applies it — the operator, your GitOps reconciler, or a human".  Not true of the content policy: nothing the operator creates carries `kube-agents/tier`, so on a default install it matches zero objects.  Reworded to say what is covered.  Deliberately **not** fixed by stamping the label on operator RBAC -- that would subject the operator's own objects to the read-only validation and could deny legitimate reconciles.
- **The explanation missing the path most likely to need it.**  `common.sh` normalised before calling the gate, `install.sh` did not, so `--permission-set=GKE-ADMIN` got the bare "invalid".  My `REJECTED_VALUES` contained `GKE-ADMIN` but only asserted a non-zero exit, so no test saw the asymmetry.  Normalised inside the gate; added a test that asserts the *explanation* for five spellings rather than just the exit code.
- **`custom` reaching the removed ceiling silently.**  Covered under What Changed.  This one was raised against my own argument for `custom` and the argument does not survive it on the installer path, so the warning is the fix rather than a rebuttal.

- **Half-removed permission set.**  Re-ran `git grep -n "gke-admin"` and `git grep -n "container.admin"` over the whole tree at the end.  Every remaining `gke-admin` is a deliberate explanation of the removal (the refusal text, the Terraform `error_message`, the site's "Why there is no `gke-admin` set", the RC release note).  Remaining `container.admin` hits are unrelated identities: the Prow eval fleet, the developer bootstrap in `scripts/dev/`, KCC guidance in the ArgoCD page, and the generic GKE-security skill.  None of them is the agent GSA.
- **The impersonate claim, which I was asked not to take on trust.**  It holds, but the original wording did not.  "IAM has no `resourceNames` equivalent" borrowed Kubernetes RBAC's vocabulary for something IAM does not have, so a reader could not check it.  Replaced with what `gcloud iam roles describe` answers and what GKE's own pages say, and corrected one detail while I was there: `roles/container.clusterAdmin` does **not** carry `container.clusters.impersonate`; only `roles/container.admin` does.  Fixed in `392e9ffc`.
- **Tests that pass for the wrong reason.**  Mutation-tested both suites.  Reintroducing the admin branch, turning the values gate off, changing the INSTALL.md `apply` to `diff`, drifting the chart copy, flipping a binding to `Warn`, and making the shell gate accept `gke-admin` each turn something red.  What that does **not** establish: `PolicyContentTest` matches the CEL as text, with no CEL runtime in the repository, so it catches a denial being deleted or reworded and not a semantically broken rewrite that keeps the substring.  The things that do evaluate the expressions are the API server, which compiles every one when the policy object is admitted, and the live enforcement run above.  One gap this found: the forbidden-role sweep only sees a branch that _names_ an admin role, so a conditional put back against a benign list passed it.  Added `test_the_grant_list_does_not_branch_on_the_permission_set_at_all` in `f782fae5`, which is the assertion that actually carries the property.
- **A test whose stated reason was false.**  The source increment had a test asserting every file the admission suite reads is in `agent-startup-test.yml`'s path filter, on the grounds that this is the only job running `tests/`.  That is not true on current `main`: `Makefile:143` has `$(wildcard tests/test_*.py)` in `PYTHON_TEST_DIRS` and `python-tests.yml` runs `make test-python` unfiltered on every PR.  Dropped the test rather than ship one whose premise a reviewer would have to disprove.  Kept the filter entries themselves, which are cheap and match the file's existing pattern.
- **Does the new `config/admission/` directory reach the kustomize build?**  No -- `config/default/kustomization.yaml` lists `../crd ../rbac ../manager ../webhook ../certmanager` and nothing else, and `make deploy` builds `config/default`.  Confirms the header's claim and is why INSTALL.md needs the manual step.
- **Stale vocabulary from the pre-#797 world.**  Found and fixed one: `terraform/modules/kube-agents-iam/README.md` said "the script's `gke-admin` set" and "the script's `custom` set" in text I had just written, in a repository that has no provisioning scripts any more.
- **A variable that now selects nothing.**  `permission_set` no longer picks a bundle.  Kept, because `installer_common.sh` writes it into every generated tfvars and `outputs.tf` preconditions on it, and said so in a comment so the next reader does not delete it and break every installer-driven apply.
- **Known-weak assertion, reported rather than fixed.**  `test_the_composition_does_not_switch_the_gate_off` only fires if the composition explicitly overrides `admissionPolicy.enabled` to something false.  Deleting the mention entirely passes it, which is correct (the chart default is on and is tested separately) but means the test is a guard, not a proof.  Left as is, with the reasoning in its docstring.
- **Coverage that reads better than it is.**  Every line of `terraform.tfvars.example` is commented out, so it contributes nothing to the forbidden-role sweep.  Not a hole, since an uncommented value still meets the `variables.tf` validation, but a pass should not be read as saying the example was checked.  Noted in the test's docstring (`ec8bd795`).
- **Docs-drift.**  Walked every page that describes the permission sets against the source after the change.  This found one page I had believed was already done and was not: `docs/security-requirements.md` still described "read-only, GKE administrator, and custom" in its section 3, because a multi-file edit had aborted partway and I had not re-checked it against the diff.  Fixed in `808c1f22`.  Second time that failure mode bit in this change, which is why the final `git grep` sweep is a step and not a formality.  Two-set tables, role lists, the CRD reference, the two concepts pages, the architecture overview, the credential-isolation page, the install skill, and `k8s-operator/scripts/README.md` all match `installer_common.sh` and `main.tf` now.  `docs/README.md` got the one-line index change for the renamed policy, and nothing else there.  `make docs-check` green.
- **A duplicated paragraph, and then a stranded one.**  "The custom set binds exactly the roles listed" appeared twice after the rewrite.  Removing the first copy left the survivor sitting under "Why there is no `gke-admin` set", describing a set from inside a section about a different one; it is back under "Roles per set" where it belongs.
- **Two claims that outlived what they described.**  The page said the binding-scope policy "is not evadable this way -- it keys on the ServiceAccount being privileged, which the binding cannot omit", true of omission and, until the `userInfo` fix, false of addition; it now says both and why.  And `docs/architecture/03-security-model.md` asserted "the controller mints no RBAC, so it cannot be the labeler" -- `reconcileRBAC` does mint RBAC, and that misconception is what let the binding-scope policy be written so that it denied the controller's own binding.  Pre-existing and outside the original diff, fixed because this change is what found it.
- **Out of scope but worth knowing:** `agents/platform/skills/gke-basics/references/cli-reference.md:282` still tells the agent to advise granting `roles/container.admin`.  That is about a *user's* cluster, not the agent's own identity, so it is correctly untouched here -- but it sits oddly beside this PR's argument and someone should look at it.
- **Not covered:** a `Deny`-bound rejection on a live cluster, for the reason in Live validation.

`kube-agents-bot` is down for a known reason (an expired credential, being handled separately), so its `AI Review` check will come back `neutral` rather than green.  That is also what gates human-reviewer assignment, so this will need a `/request-review` from someone with the permission to send it.

## Risk & Rollout

The breaking half is a refusal, not a behaviour change: an install that does not name `gke-admin` sees nothing move, and one that does fails loudly at configuration time rather than provisioning an over-privileged GSA.  The read-only plan is byte-identical before and after.  Nothing revokes an already-granted role, so no existing agent loses access on merge; the migration note covers the manual cleanup.

The admission policies are the part with real blast radius.  They are cluster-scoped, `failurePolicy: Fail`, and default on, so a bad CEL expression would block RBAC writes cluster-wide.  Mitigations: the expressions are unchanged from the ones already in the tree, they were installed in Warn mode on a live 1.35 cluster and behaved as intended in both directions, and `admissionPolicy.enabled=false` turns them off in one flag.  A cluster below 1.30 needs no flag -- the template checks the version itself and renders nothing there.  Turn the flag off for a second kube-agents release in a cluster that already has them: the objects are cluster singletons with fixed names, so Helm refuses the second install on ownership.

The failure mode that would have bitten hardest is now closed and tested in CI: the policies denying the operator's own reconcile.  That is worth re-checking on review, because it is a coupling between a shipped policy and the controller's hot path rather than a self-contained manifest.

Revert is a straight revert of the branch.


dshnayder (approved):

bnaylor:
Merged current main in to clear the conflict.  One file - the path filters in the agent-startup-test workflow - and both sides' entries are kept, so nothing that was approved changed.

PR #962, opened 2026-08-26
feat(proxy): broker-owned git trees, behind a flag

bnaylor:
## Summary

The credential broker gets its own git working trees, behind a flag that is off by default.  With the flag on the broker serves `/v1/workspace/*`, where the agent sends `{path, bytes}` pairs plus a branch and a message and the broker does the clone, the write, the commit and the push in a tree on its own volume.  The agent never names a path and never gets one back.  With the flag off the routes do not exist, the tree directory is not created, and nothing else changes.

The reason this is worth doing is narrower and more interesting than "another hardening flag", so it's the first section rather than a footnote.

## Why This Change

#952 pinned every git config key it could enumerate and then ran out of road.  The keys that matter carry an arbitrary name inside the key -- `filter.<n>.smudge`, `diff.<drv>.command`, `alias.<n>` -- so there is no finite set to deny, and `.git/hooks/pre-commit` needs no unusual argv at all.  That PR documented the class instead of closing it, which was the right call at that layer.  Enumeration cannot finish against a surface whose whole design principle is extensibility.

All of it follows from one arrangement: the broker runs `git` inside a directory the agent owns.  Move that and the question stops being about flags.

**This also removes an infrastructure dependency.**  Splitting the broker into its own Pod (#960) separates the half that holds the credentials from the half that authors the files, and if the two still have to meet in a shared checkout then that checkout has to be shared read-write between two Pods.  That dependency was refused rather than taken on.  Passing content instead of a directory removes it: the broker keeps its own volume and nothing has to be shared.  The portability fix and the security fix turn out to be the same change, which is rare enough to say out loud.  I have not measured the storage economics this session and the numbers are not in this diff, so I am not quoting any.

## What Changed

- `agents/platform/scripts/content_workspace.py`, new.  The store, the path validator, the payload parser, the branch check, and the disjoint-roots assertion.  Nothing in it takes a caller-supplied path or a caller-supplied URL -- `open` takes `owner/name` and composes the https URL itself, because a caller-supplied URL is `url.<host>.insteadOf` by another route.
- `credential_proxy.py`: `content_workspace_enabled()`, `CommandExecutor.content_workspace_root`, `execute_workspace_git`, `build_workspace_store`, and the `/v1/workspace/*` handler.  `_execute` grows a `containment_root` parameter that defaults to the shared workspace, so every existing caller is unchanged.
- The docs section below, and a paragraph on the reference page.

Four things hold the mechanism together and they are not equally strong.  Worth reading in this order:

1. **Disjoint roots, checked.**  `assert_disjoint_roots` refuses to construct the store if the tree root and the agent-shared workspace root overlap.  A mount rearrangement that collapses the two is a broker that will not start.
2. **One path validator.**  `repo_relative` runs on reads and on writes, refuses rather than normalises, and refuses every spelling of `.git` a filesystem treats as `.git` (NTFS trailing dots and `GIT~1`, HFS+ zero-width codepoints).  A checker that disagrees with itself about `manifests/../.git/config` is the defect class this codebase keeps producing.
3. **A separate door for the broker's own git.**  `execute_workspace_git` does not go through `/v1/exec`.  It has its own literal subcommand allowlist, refuses `-C`, and is contained to the tree root.  Keeping the two apart is what lets the agent-facing answer stay "git is not reachable" rather than "git is reachable, narrowly".
4. **Geometry, and not checked here.**  The tree lives on the broker's own volume, which the agent container does not mount.  Nothing in the process can verify that.  It is the same argument the proxy already makes about `$HOME/.gitconfig` and `KUBECTL_KUBERC`, and it is exactly as weak.

### What arming the flag does *not* do

I measured this rather than reading it, and the answer is not the one the mechanism's own docstring implies, so it belongs in the body.

**Arming the flag does not close the repo-local `.git/config` execution class.**  Two separate facts:

- `git` still executes what a repo-local config names, in *either* tree.  A `.git/config` planted in the broker's own working tree fires through `execute_workspace_git` exactly as it fires through `/v1/exec`.  Nothing in this PR changes what git does with a config file.  What changes is who can write one.
- `/v1/exec` is untouched while the flag is on.  Both mechanisms run side by side on purpose, so the mechanism can land and the skill migration can be a separate diff.  While that is true the old channel stays fully open, and today no shipped skill uses the new routes -- so in practice the class stays open for all git work, and closes only for work that moves.

It is not only geometry, though, and my first draft of this section was harsher than the change deserves.  The new path also runs a materially narrower git: `execute_workspace_git` accepts a literal list of subcommands, refuses `-C`, and never composes an argv from a caller string, against `/v1/exec`'s denylist-shaped surface where the agent spells the command.  That is a behavioural improvement, not a geometric one, and it holds wherever the new path is used.

What the flag buys on top of that is that the config class does not exist on the new path at all, and that part is a property of the deployment rather than of the code.  Both halves are measured below.  The honest summary is that this PR builds the layer at which the class can close and closes it for anything routed through it; the class itself closes when the skills migrate and the agent-facing git door goes away, which is not this PR.

## Context

- **Stacked on #952**, which is not merged yet, so this PR shows its six commits too.  Mine are the last six: `606f7c55`, `9fdf6711`, `d133dadd`, `107e8662`, `998635a2`, `16390f1e`.  Merge #952 first and GitHub will retarget this one.
- **#960** (broker Pod split) is the mechanism this makes adoptable.  That PR separates the two halves; this one is why the separation does not need ReadWriteMany storage.  Until this lands the split stays off by default.
- **#913** (@dshnayder, draft) carries a `content_workspace.py` of its own as part of a larger sandbox-pod change.  Settled: this is the shared implementation and #913 is unblocking downstream work that needed content-passing to exist.  I have left a note on that PR with the differences that matter when the two are reconciled, since a few of them fail quietly rather than loudly.
- Re-cut of #720's increment 8.

## Testing

- `make verify` green.  `make docs-check` green.  `prettier --check` on both changed Markdown files.
- Python: 195 tests across `test_content_workspace.py` (33) and `test_credential_proxy.py` (162), all OK.
- **Twelve mutants, twelve caught.**  Every control this PR adds or fixes was checked by breaking it and confirming the suite goes red: the lock, the workspace cap, the clone-byte ceiling, the root redaction, the handle redaction, the absolute-path redaction, the root `.resolve()`, the `list` prefix match, the `base` check, the symlink check at the commit call site (3 red), the same at the read call site, and the failed-clone cleanup (2 red).  Before the wiring tests went in, bypassing the symlink check at *both* call sites left all 21 tests green.
- A deadlock in the lock tests now reports rather than hangs.  Swapping `RLock` for `Lock` used to stop the suite dead, which in CI is a job timeout with no test named; `ConcurrencyTest` arms `faulthandler.dump_traceback_later` for its own duration, and the mutant now dies naming `content_workspace.py:683 in get` under `:753 in commit`, exiting 1 so CI reports a failure rather than a job timeout.
- The two concurrency bugs were reproduced before the fix and re-measured after: the injected `close` can no longer complete mid-commit and leaves nothing on disk, and 16 genuinely concurrent commits across 8 rounds produced zero mismatches between what a caller sent and what its reported sha contains.

### Live validation

**Install:** `bnaylor-ka-test` (GKE, `us-east4`, namespace `kubeagents-system`), operator `kubeagents-controller-manager`.  Images `platform-agent` and `credential-proxy` built at `dev-20260825-192427` from this branch and rolled out by patching `spec.deployment.tag` on the `platform-agent` CR, which is the supported path -- the operator derives the sidecar image from the agent tag.  The flag was armed by adding `CREDENTIAL_PROXY_CONTENT_WORKSPACE=1` to `spec.deployment.env`, which the operator passes through to the sidecar (the name is not on the reserved list, correctly -- it arms a control rather than disarming one).

**Flag off (default).**  The proxy's own state dir shows the feature is not merely unrouted but not built:

```
$ ls /var/lib/credential-proxy/          # in envoy-credential-proxy
git-hooks-disabled  home  kubeconfigs  tmp        # no content-workspaces

$ CREDENTIAL_PROXY_CONTENT_WORKSPACE=[<unset>]
```

and from the sandbox, all six routes are absent while a route that exists still answers:

```
open   -> 404 {"status":"not_found"}      list   -> 404   read  -> 404
commit -> 404                             push   -> 404   close -> 404
/v1/exec     -> 200   (kubectl version, exitCode 0)
/v1/nonsense -> 404
```

**Flag on.**  The store logs its root, and the tree root is created `0700`:

```
INFO credential-proxy content workspace enabled root=/var/lib/credential-proxy/content-workspaces
drwx------ 2 10000 10000 4096 Aug 25 23:45 content-workspaces
```

Full lifecycle from inside the `platform-agent` container, against a real repository:

| Call | HTTP | Result |
| --- | --- | --- |
| `open {repo: octocat/Hello-World}` | 200 | `handle=f6b6e359…`, `base=master`, `baseSha=7fd1a60b01f9…` (real HEAD) |
| `list {handle}` | 200 | `[{"path":"README","size":13}]` -- no `.git/` entries |
| `read {handle, path: README}` | 200 | `SGVsbG8gV29ybGQhCg==` -- the real bytes |
| `close {handle}` | 200 | `{"closed":true}`, and the tree is gone from the broker's volume |
| `close` again | 404 | `workspace.unknown-handle` |

Refusals, on the same live handle:

```
read .git/config       -> 403 workspace.path.refused
read .GIT/config       -> 403 workspace.path.refused
read ../../etc/passwd  -> 403 workspace.path.refused
read /etc/passwd       -> 403 workspace.path.refused
commit .git/config     -> 403 workspace.path.refused
commit onto main       -> 400 'main' is a rollout branch
open https://attacker.invalid/x/y.git -> 400 repo must be owner/name
open "ext::sh -c id"   -> 400 repo must be owner/name
```

No response body carried a filesystem path.  The handle is opaque hex.  (That claim is now enforced rather than observed -- see the Self-Review.)

**The claim the whole thing rests on: the tree is unreachable from the sandbox.**  Same pod, same uid, two containers.  From `envoy-credential-proxy`:

```
/var/lib/credential-proxy/content-workspaces/f6b6e359…/repo/.git/config
    -rw-r--r-- 1 10000 10000 267
```

From `platform-agent`, in one shell:

```
id                                                  -> uid=10000(hermes) gid=10000(hermes)
ls  /var/lib/credential-proxy/content-workspaces    -> No such file or directory
grep -c credential-proxy /proc/mounts               -> 0
find / -maxdepth 6 -name content-workspaces         -> (nothing)
append to the tree's .git/config                    -> refused
```

Not permission-denied.  Absent -- the broker's emptyDir is mounted in the proxy container only, so there is no path to name.  For contrast, in that same shell:

```
[ -w /opt/data/gitops/<audit>/<repo>/.git/config ]  -> WRITABLE
[ -w /opt/data/gitops/<audit>/<repo>/.git/hooks ]   -> WRITABLE
```

and `/opt/data` is `drwxrwsr-x root:hermes`, setgid, with the sandbox at gid 10000.

**And the measurement that decides the headline claim.**  With the flag *on*, the old channel still fires.  A/B in one shell, negative control first:

```
1. broker clones a repo into the shared workspace via /v1/exec        -> exitCode 0
2. sandbox appends [diff] external=<payload> to that clone's
   .git/config by PLAIN FILE WRITE -- no proxy call involved
3. marker file absent                                                  (control)
4. broker runs an ordinary `git diff` there via /v1/exec, FLAG ON      -> exitCode 0
5. marker file present, contents FIRED
6. at the same moment: ls /var/lib/credential-proxy/content-workspaces -> No such file or directory
```

Step 5 is the point.  Arming content-passing does not close the shared-workspace channel; the broker-owned tree is a second door that the channel does not reach, and the first door is still standing.  Reproduced locally against `diff.external` and `diff.<drv>.textconv` as a 2x2 (planted/unplanted x shared/broker), and in every cell the unplanted twin stayed cold, so the probe discriminates rather than always reading "fired".  The broker's own tree fires too when a config is planted in it directly, which is why the section above calls the protection geometric rather than behavioural.

**Not covered.**  `commit` and `push` were exercised only through their refusal paths.  `github-token-minter` is not installed on this cluster, so no GitHub credential is available and a private clone fails -- and it fails identically through `/v1/exec` (`git ls-remote` -> `could not read Username`), so this is the install, not the change.  The public-repo lifecycle above is what was reachable.  Separately, the leak fix in `d133dadd` landed after the live run and is covered by unit test only; the failure path it fixes is the same one I observed live (an orphaned tree directory left by the failed private clone), so I know the bug was real, but the fix itself has not been re-run on a cluster.

**Cleanup, verified.**  CR restored to `dev-20260812-184046` with `spec.deployment.env` removed; pod back to 4/4 on the baseline image; `/opt/data/gitops/up11-probe` and the scratch handle file deleted and confirmed gone; the `live-test-lease` ConfigMap I took for the session deleted.  The two `dev-20260825-192427` images are left in Artifact Registry.

## Self-Review

Adversarial and docs-drift passes in this session, plus a review pass from a context that did not write the change, handed the diff range and nothing else.  Merged list, current rather than chronological:

- **Two concurrency defects, both reproduced, both fixed in `998635a2`.**  The handler is a `ThreadingHTTPServer` and this module had no lock at all.  A `close` landing inside a `commit` left the commit re-creating the tree the close had just removed, with no handle pointing at it -- four orphaned paths on disk and `handles registered: []`.  Two concurrent `commit`s on one handle had the second's `clean -fdxq` delete the first's files, so the first either failed on a vanished pathspec or landed its commit on the other's branch while still answering `committed: true`.  The second is the worse one: a caller could push a branch whose content is not what it sent, with nothing in the response saying so.  One reentrant lock across every verb; re-measured after.
- **The symlink check's wiring was untested and mutation proved it.**  Every existing test called `_no_symlink_on_the_way` directly, so bypassing it at both call sites left all 21 green.  Two tests through `store.commit` and `store.read` in `107e8662`, with the symlink committed into the base rather than written, because `checkout --force` plus `clean -fdxq` would otherwise remove it before the write loop and the test would pass for the wrong reason.
- **An invariant four places asserted and nothing enforced.**  `GitFailed` put 500 bytes of raw git stderr on the wire while the module docstring, the handler docstring, `_redact`'s own docstring and this body all claimed no response carries a filesystem path.  git quotes absolute paths in its own errors.  Redacted rather than the claim softened, since the claim is the one worth keeping.  The first version of the redaction was itself still wrong: it knew the tree root and the handle grammar only, so a *second* absolute path went out whole -- `warning: unable to access '<state-dir>/.gitconfig'` is one the broker really emits, because `$HOME` is deliberately its own state dir.  It now scrubs every absolute path rather than the two it knows by name, and refs survive because `origin/main` is most of what a git error is about.
- **The tree root was not resolved** while `assert_disjoint_roots` resolved both sides.  With a symlinked prefix the redaction matched only the tail and emitted `/private<workspace-root>/...`, seen in a real run rather than reasoned about.  Unreachable in production only because `CommandExecutor` resolves the root before handing it over, which left this module's invariant depending on a caller somewhere else.  `Path(tree_root).resolve()`.
- **Nothing bounded the trees.**  No clone-size ceiling, no cap on open workspaces, on a volume that is node ephemeral storage.  Both added; the byte-ceiling knob is named to match #913's.
- **`list(prefix=...)` matched strings, not path components,** so `prefix="a"` returned `ab/x.yaml`.  Fixed.
- **`open`'s `base` was unvalidated where `branch` was checked.**  Not reachable today -- it is always prefixed with `origin/` -- but the asymmetry is the kind a later caller removes without noticing.  Fixed by splitting the leading-dash check out from the protected-branch rule, so `base="main"` still works, which is the ordinary case.

- **The mechanism's own framing overclaims.**  The module docstring reads as though moving the tree closes the class.  Measured, it does not -- see the A/B above.  Fixed in the prose rather than the code: the design doc and this body both say plainly which half closes and which stays open.  The code is unchanged because the code is correct; it was the description that was ahead of it.
- **An orphaned tree directory on a failed clone.**  Found live, not by reading.  `open` creates `<tree_root>/<handle>` before the clone that fills it, and a clone that cannot authenticate raised out with the directory still there and no handle registered against it -- unreachable by `close`, one per attempt, for the life of the container.  Fixed in `d133dadd` and mutation-checked.
- **`content_workspace_root` is created at the process umask and narrowed to `0700` later,** when `ContentWorkspaceStore` is constructed.  The live check confirms `0700` in the shipped path.  The window is wider than I first wrote: `serve()` runs `executor.bootstrap(...)` between the two, which is an operator-supplied shell command with a timeout in the minutes, so it is minutes rather than microseconds.  Still not agent-reachable -- the state dir is not mounted in the agent container -- so the mode is defence in depth either way.  The `chmod` no longer fails silently, which was the part actually worth fixing; the ordering is left as it is.
- **The handle is a bearer capability, not authorization.**  It stops one session naming another's tree by guessing.  It is not an ownership check, because the broker cannot tell two sessions in the agent container apart.  Stated in the docstring rather than implied.
- **Docs-drift:** the design doc's git-configuration section ended on "until that lands", which this PR partly invalidates.  Rewritten.  No generated regions touched, no new documents so no map row needed, no hard-coded counts introduced (the subcommand list is referenced by symbol, not by number).  `make docs-check` green.
- **Merge resolution.**  The source branch was written on top of a different shape of the git-hardening work and against an authenticator that does not exist on this base.  Two conflicts, both resolved by dropping the other branch's content rather than inventing a merge: the handler's authenticator/`Principal` block (that is #960's work, not this PR's) and five test classes that belong to the same.  The source diff also carried duplicated-paragraph merge artifacts and four Go/CRD files whose only content was broker-Pod-split prose about ReadWriteMany storage -- all dropped, and the docs rewritten from scratch instead.  Verified afterwards: one `_git_refused_name`, one `GIT_EDITOR` pair, one `import base64`, no conflict residue.
- `kube-agents-bot` has not reviewed this one.  It is down for a known reason (expired credential, owner away), and the `AI Review` check goes `neutral`.  Not retried.

## Risk & Rollout

Low while the flag stays off, which is the default.  With it off `content_workspace_root` is `None`, the directory is not created, the store is `None` and the routes 404 -- verified on a live install rather than argued.  The only change on the off path is `_execute` taking a `containment_root` that defaults to the value it used before.

With the flag on the new surface is additive: `/v1/exec` behaves exactly as it did, so a rollback is removing one environment variable and no state has to be migrated.  The failure mode to watch for is `assert_disjoint_roots` refusing to start the broker if someone points the state dir and the shared workspace at the same volume.  That is deliberate -- a broker that starts without the property would be worse -- but it is a crash rather than a warning, so it is worth knowing before rearranging mounts.

**The lock serialises the whole store, and that is a throughput decision worth stating.**  It is held across `clone`, `fetch` and `push`, so it is held across network I/O bounded only by the executor's 300s timeout rather than by anything in this module -- measured, a `get` on an *unrelated* workspace blocked 2.7s behind a 3s clone.  So the store serves one request at a time, worst case for minutes, and that includes `close`, which is the verb an operator reaches for when they want a tree gone.  Right trade for a single agent Pod publishing one pull request at a time, but note that `max_workspaces` advertises a concurrency the lock forbids: eight may be open, one may be doing anything.

**Two resource numbers that want reading together.**  The clone ceiling is per clone, not aggregate, so the defaults multiply out to roughly 2 GiB retained across eight open workspaces by design -- size the node's ephemeral storage against the product rather than either number.  And the ceiling is measured after the clone finishes, so it bounds what is retained and not the peak; a shallow clone plus the ceiling would bound both and is the right follow-up.

**One gap to know about before arming this anywhere real.**  `push` was never exercised against a remote, because the dev cluster has no token minter.  The interesting case is `--force-with-lease` on a branch that does not exist on the remote yet, which is every branch this feature authors: with no remote-tracking ref there is nothing for the lease to compare against, so it does not mean what the flag name suggests.  I believe the push is still safe -- a create cannot clobber what is not there, and a second push does have a ref to compare -- but I have not run it, so do not read "live-validated" as covering it.  Worth a live pass on an install that can authenticate before any skill is moved onto `push`.

Merge after #952.


bnaylor:
Merged current main in.  Four files conflicted where main's squashed git-hardening and socket-mode work sits next to this branch's additions.  Main's backend-socket test and this branch's workspace-git tests are both kept, and the two doc sentences this branch rewrites still carry main's claim.  Conflict cleared.

bnaylor:
Main moved again under this one when #960 landed.  One more conflict, in the request handler's class body: main's authenticator and principal attributes arrived where this branch adds its workspaces attribute.  Both kept.

bnaylor:
The red smoke test is `cluster-agent-crashloop-debug`, and it is not this branch.  Same task, same VerificationCorrectness=0.0, on #968 at 18:58 UTC (before the broker-split change landed on main) and on #965 at 22:35, and #982 is red on its sibling `cluster-agent-healthy-workload-no-finding`.  Retested here once and it failed the same way.  Everything else on the branch is green.

dshnayder (approved):

Issue #903, opened 2026-08-23
Lane: eval capacity (codify project onboarding, take the pool to 15)


jayantid:
**Lane: eval capacity.** Owns how many pull requests can be evaluated at once. Codify project onboarding, run it until the pool is 15.

Waits on nothing. Start Monday 9am.

## Why this is a lane and not a chore

The agent is scoped to one project and **cannot read clusters in another one**. That was the open question last week and the answer is no. So the fleet is not one shared environment, it is one trio of clusters per project, built from the same Terraform module. Which means: **the number of onboarded projects is the number of concurrent eval runs.** Three today. Fifteen people push at 10am and twelve of them queue.

It also means onboarding grew. It is no longer "register a project in Boskos", it is a five-part sequence, and getting the order wrong has already cost us a day.

## The sequence, and why the order matters

1. Create the GCP project and its GitOps repository (`gke-agentic/<project>-infra`), and install the GitHub App on it.
2. Apply `bench/tf/fleet` for the project. Three seeded clusters with the planted defects, per-project state bucket. The module and the `apply` are documented in `bench/tf/fleet/README.md`.
3. Apply `terraform/examples/ci-pool-minter` in a new workspace, its rule keyed on that project's platform service account.
4. Import the GitHub App PEM into the project's KMS key (minty v2.7.1).
5. **Register the project in Boskos last.** Everything above is a prerequisite of the Boskos entry, not a follow-up to it.

Step 5 is the one that bit us. `kube-agents-evals-3` was registered on 2026-08-21 with only its GCP half provisioned, and every presubmit that leased it died on the unmapped-project refusal. A project that becomes leasable before it is onboarded takes a share of every presubmit and fails it. The Boskos pool list lives in `gke-internal/test-infra` under `deployments/gke-agentic-tooling-team/boskos`, in a different repository from the Prow config, and nothing enforces the order between them.

`kube-agents-evals-3` is the worked example and it is half-done: repo created 2026-08-21, App installation added 2026-08-23, minter and PEM outstanding. Finish it first, verify it end to end, then do the other eleven. Onboarding one project completely before scaling is exactly the discipline whose absence produced that incident.

## One pool-wide trap

`EVAL_GITHUB_APP_ID` is pool-wide. Do not set it until **every** pool project has a provisioned minter and an imported key. The minter Deployment is in the release that `helm --wait` gates on, and it cannot pass readiness without the key, so setting the variable early turns every unfinished project's presubmits red.

[...]

PR #979, opened 2026-08-26
feat(ci): codify CI pool project onboarding automation, runbook, and …

lapis2002:
## Summary

Onboarding a project into the Boskos eval pool was an oral tradition: three projects existed, each built by hand, and the only record of how was the projects themselves. This turns it into a runbook plus two scripts — `scripts/provision_ci_pool_project.sh` drives provisioning end to end, and `scripts/verify_ci_pool_project.py` refuses to pass a project that is only nearly right. The runbook at `docs/site/src/content/docs/deploy/ci-pool-projects.md` is reordered to end at Boskos registration rather than start there. Written by using it on two real projects, `kube-agents-evals-4` and `-5`, which is how the bugs below were found.

## Why This Change

Issue #903 wants the pool at 15. Boskos hands out one project per run, so pool size is the presubmit's concurrency ceiling; two recent smoke tests that leased `kube-agents-evals-3` held it for 60m18s (#912) and 72m06s (#935), putting three projects at ~2.7 runs/hour and fifteen at ~13.6. Twelve more onboardings from an oral tradition is twelve chances to produce a project that looks provisioned and fails on lease — not hypothetical, since `kube-agents-evals-3` was registered before it was verified and broke every presubmit that leased it for three days. Three of the twelve are now done and waiting on registration: `kube-agents-evals-4`, `-5` and `-6`.

## What Changed

- **`scripts/provision_ci_pool_project.sh`** (new) — steps 0 through 5, mapped one-to-one onto runbook sections §§0–7: preconditions, APIs and IAM, the `full-install` apply, the seeded fleet, the minter's Terraform, the one-shot PEM import into Cloud KMS, and a verification run. It stops there; Boskos registration stays human. `--region` refuses anything but `us-central1` because `bench/tf/fleet` is zonal and pinned.
- **`scripts/verify_ci_pool_project.py`** (new) — eight checks: the `ci-deploy.sh` mapping, project and APIs, IAM, Artifact Registry and its cleanup policy, GKE clusters and Terraform state, seeded-fleet fixtures, the GitOps repo and its App installation, and the token minter end to end. `check_toolchain()` gates all eight, so a missing binary or an expired credential exits 2 instead of reporting every resource absent.
- **The fixture check is new since this PR opened**, and it removes the runbook's last manual step. Matching the trio by name says a cluster exists, not what is planted in it: an apply that creates the clusters and dies before the Kubernetes provider runs leaves three that carry the labels, answer every API call, and hold none of the seven fixtures. `check_seeded_fleet_fixtures()` runs `hack/fleet-kubeconfigs.sh` in a temporary kubeconfig directory and requires every role in `bench/tf/fleet/fixtures.json` to resolve **and** be present. That script exits 0 whether it wrote every role or none — deliberately, so a missing kubeconfig degrades one eval rather than killing the job — so the check parses its stderr summary, and a test reads the format string out of the script rather than copying it. No `kubectl` is exit 2, not a failure. The old check is renamed `GKE Clusters & Terraform State` so it stops implying coverage it never had.
- **`docs/.../deploy/ci-pool-projects.md`** — an ordered runbook, §§0–9. §0 covers project creation and billing; §8 marks registration as a one-way boundary (see Risk & Rollout); §9 is new. Registration makes a project leasable, it does not make the presubmit use it: the job's `max_concurrency` in `oss-test-infra` is the real cap and the pool size is only a ceiling above it, and that repo and the one holding the Boskos roster know nothing about each other.
- **`hack/ci-deploy.sh`, `tests/test_ci_gitops_repo.py`** — `kube-agents-evals-4`, `-5` and `-6` added to the project→GitOps-repo mapping. `-6` was missing from all three places that record it, which would have stopped its own Step 0.
- **`terraform/examples/ci-pool-minter/{README.md,outputs.tf}`, `bench/tf/fleet/README.md`, `bench/tasks/DRAFTS.md`, `docs/README.md`** — count-prose that had gone stale three times. Load-bearing counts updated; duplicated rosters deleted and pointed at `gitops_repo_for_project()`, which is what CI actually reads. `outputs.tf` also now says *when* its two human steps are due: before registration. `EVAL_GITHUB_APP_ID=4675512` has been set pool-wide in the Prow environment since 2026-08-25 (oss-test-infra#2661), so nothing holds `githubMinter.enabled=false` any more — register first and the minter points at a key with no enabled version, fails readiness, and `helm --wait --timeout 15m` kills every lease fifteen minutes in.
- **`bench/tasks/consistency-drift-outlier/task.yaml`** — a comment recording that this case is deliberately not date-gated, unlike `fleet-cost-idle-pool`. An absent gate reads the same as a forgotten one.

What the script deliberately leaves alone: creating the project and linking billing (§0 asserts both and fails rather than repairing), adding the GitOps repo to the App installation (that edit *is* the security review, so it happens in the GitHub UI where the approver sees the scope change), adding the mapping row (a code change; §0 refuses to provision without it), Boskos registration (§8), and raising `max_concurrency` (§9, a different repository). Running `hack/fleet-kubeconfigs.sh` used to be on that list.

## Context

Part of #903 (Lane: eval capacity). Per that issue, `terraform apply` and mutating `gcloud` IAM commands are human steps; the script prompts rather than assuming.

Follow-up issue to file — five things about *repairing* a project rather than onboarding one:

1. Step 2.1 mints a fresh `api_server_key` on every invocation and writes it into the agent Secret. Harmless during onboarding; after registration it rotates a credential out from under a live lease. **Confirmed on the evals-5 run.**
2. `helm_release` renders `values` as `(sensitive value)` in a plan but prints the same content in plaintext through the `metadata.values` mirror. **Confirmed on the same run** — which is why no transcript of it appears here.
3. No `full-install` secret can be rotated on its own. `session_kv_api_key` and `session_kv_salt` are `random_password` resources, but `api_server_key` is a required no-default variable and Terraform does not remember variable values, so a targeted `-replace` either rotates it as a side effect or needs it read back out of state in plaintext.
4. `seeded_b`'s `maintenance_exclusion` has computed `start_time`/`end_time`, so the fleet stack never plans clean.
5. App ID `4675512` is duplicated across four files.

## Testing

- `make docs-check` — passes all five checks. `prettier --check` passes on every changed Markdown file.
- `bash -n scripts/provision_ci_pool_project.sh` and `python3 -m py_compile scripts/verify_ci_pool_project.py` — clean.
- `scripts/test_verify_ci_pool_project.py` — **110 pass.** Eight cover the fixture check, including the summary regex asserted against the format string read out of `hack/fleet-kubeconfigs.sh`, so changing that line fails here rather than turning every fixture check into "unverified". Two classes are new with the review fixes: twelve over the two-file mapping read (row on `main`, row branch-only, row absent locally, no remote for the merge target, remote resolved by URL, unreadable `main`, a snapshot predating the function, and the snapshot's date in the message) and six over the chart key-version pin (a disabled or absent pinned version fails, an unreadable pin warns and falls back, and `hack/ci-deploy.sh` carries no override). Two tests asserting the old highest-enabled probe were deleted rather than adapted; they encoded the bug. `tests/test_ci_gitops_repo.py` — 15 pass.

### Live validation

Exercised by using it three times — `kube-agents-evals-4` on 2026-08-25, `-5` and `-6` on 2026-08-26, all `us-central1`, each from an empty project to verified through the script — then by running the verifier against **all six pool projects** on 2026-08-26, after the review fixes below.

```
kube-agents-evals     ALL CHECKS PASSED    exit 0
kube-agents-evals-2   ALL CHECKS PASSED    exit 0
kube-agents-evals-3   ALL CHECKS PASSED    exit 0
kube-agents-evals-4   1 item unverified    exit 2   (mapping not yet on main)
kube-agents-evals-5   1 item unverified    exit 2   (mapping not yet on main)
kube-agents-evals-6   1 item unverified    exit 2   (mapping not yet on main)
```

All six pass the other seven checks: 9 APIs, Workload Identity and cross-project reader grants, Artifact Registry with a cleanup policy and node pull rights, four clusters with CMEK and the state bucket, 7 fixture roles planted and reachable.

That run proves two of the review fixes. **The signing probe uses the version the chart deploys** — every project reports `Import-only signing key v1 (the version the chart deploys) ENABLED … signature accepted by GitHub as App 4675512`, real KMS signing a real App JWT with the version read from `githubMinter.kms.keyVersion`. **The mapping check reads `main`, and the two answers differ** — the three registered projects have the row on `gke-labs/main`; `-4`, `-5` and `-6` withhold the safe-to-register verdict. Before the fix all six printed the same green line and three were wrong, which is evals-3's outage with the safety catch removed.

On evals-4 the `full-install` apply built `platform-agent-host` and the chart on top of it; step 3 applied the minter stack with remote state; step 4 imported the App PEM into `github-token-minter-key` via minty v2.7.1, and the PEM did not enter Terraform state — the key is created `skip_initial_version_creation = true`, so version 1 exists only because of that import; step 5's verifier passed after the operator confirmed the App installation. `hack/fleet-kubeconfigs.sh` against it reports `7 role(s) written, 0 on clusters that could not be resolved or reached, 0 whose fixtures were not present`. evals-5 then went through the same script to `ALL CHECKS PASSED. Project kube-agents-evals-5 is safe to register in Boskos.`

evals-6 is the run that matters for the two claims the other two could not make — greenfield Workload Identity, and the fixture check's first live run inside the verifier — and it reached `ALL CHECKS PASSED`.

The App-installation item is operator-confirmed on all six: listing an installation's repositories needs a token authorized to the App, and no operator PAT is one. On the first evals-6 run it went to exit 2 with `Could not read installation 155534965's repository list with this token` and paused for confirmation in the GitHub UI — designed behaviour, since the alternative is a wider credential rather than a better check.

**Timing, since #903 needs it to plan the rest:** on evals-4, host cluster creation started 21:40:22Z, the fleet was up at 22:02:18Z, and step 4 was reached at 22:16:02Z — roughly **36 minutes** of Terraform, plus the human steps around it.

**Prove-the-mechanism, and where it stops.** Bugs 1 and 3 below were each seen as a specific failure and then as a pass after the fix, on the same project. Bug 2 is the exception and the claim is narrower than it looks: its fix was exercised on a re-run of a project whose `<project>.svc.id.goog` pool already existed, because step 2.1 of the failing run had built the host cluster before step 2.3 died, and a Workload Identity pool is project-level and outlives the cluster that created it. The re-run proved the deleted step is not *needed*; it did not prove the ordering holds on a project that has never had a WI-enabled cluster, which is the only state the bug appears in. **The evals-6 run closes that**: it applied step 2.1 against a project with no WI pool and no cluster of any kind, and `Service Accounts & IAM Grants: Workload Identity ... verified` is the binding the deleted step used to duplicate.

**The fixture check's first live run is also evals-6.** `Seeded Fleet Fixtures: All 7 fixture roles planted and reachable` is the check running inside the verifier against a real project, which until this run was the one claim here with only unit tests behind it. It is the pass case; the four decline paths remain covered by tests alone, which is appropriate — an unplanted fixture is a state no correctly-provisioned project can be put into on purpose.

**What the greenfield run found — three bugs and one limitation, all invisible on projects 1–3 because those predate the script and were already in Terraform state:**

1. Step 1 created `kubeagents-platform-gsa` with `gcloud`, outside Terraform's state, and step 2.1 then died on `Error 409: Service account ... already exists`. **Fixed** by deleting the creation: `module.kube_agents_iam` owns that account, its roles and its WI binding, and its defaults are already exactly those values.
2. Step 2.3 re-declared the Workload Identity binding the same module already has, with an identical member — a no-op on a good run and a hard `Identity Pool does not exist` on a project with no cluster yet, because `<project>.svc.id.goog` is created implicitly with the project's first WI-enabled cluster. **Fixed** by deleting the step.
3. Step 4 failed with `Cloud KMS API has not been used in project <operator's own project>` — a project that appears nowhere in the script. The Go client libraries send ADC's `quota_project_id` as `x-goog-user-project`, so the check ran against the operator's ADC project rather than the one holding the key ring. **Fixed** with `GOOGLE_CLOUD_QUOTA_PROJECT` scoped to that one call, rather than repointing the operator's global ADC or enabling KMS on a project that should never have been in the request.
4. `api_server_key` is minted fresh on every run, because `full-install` declares it with no default and `lifecycle.sh` applies `-input=false`. Correct for onboarding, destructive after registration — **a boundary to respect, not a bug to fix here.**

Smaller fixes from the same run: step 1.4's "no installation at all" branch printed nothing, because the message sat inside the `if`; and every `read` became `|| ...`, since `set -e` plus a bare `read` turns Ctrl-D into a dead run and, at the pause, into asserting a confirmation nobody gave.

**Quota, since twelve more projects raises the question:** a fully-provisioned pool project uses 25 CPUS and 8 IN_USE_ADDRESSES in `us-central1`, identical across evals, -2 and -3, against per-project limits of 3000 and 575. Not grandfathered — evals-4, days old, carries the same. Quotas are per project, so 15 do not divide a shared pool. **No increase is needed.**

**Not covered:**

- **The stale-snapshot path in the mapping check has unit tests only.** The six-project run used a freshly fetched ref, so it covers "row on main" and "row not on main" but not "this snapshot predates `gitops_repo_for_project()` and cannot answer". That path was found by a real run against a stale ref; reproducing it live means deliberately un-fetching a checkout.
- The GKE clusters-per-zone cap could not be read: `gcloud services quota` does not exist in v576, `gcloud components install beta` fails on this machine on a root-owned `lib/surface`, and the Cloud Quotas endpoint rejected the available token. The documented default is 50; each project uses three zonal clusters plus one regional.
- No presubmit has leased evals-4, -5 or -6, because registering them is the step this PR says to do last. They pass the checks; a real run on them is unverified.

**Artifacts:** evals-4 and -5 are left provisioned on purpose — they are pool projects, not test residue. Nothing else was created.

## Self-Review

Both required passes ran through `/pr-preflight`, each in its own subagent handed the diff range and nothing else — no plan, no reasoning, no summary of intent, and neither in the context that wrote the change. **`review-adversarial`** — 15 findings across shell and Python control flow, exit-code contracts, re-run idempotency, security boundaries (the App installation, the KMS key, billing exposure), resource lifecycle, concurrency, and claims the code makes about other files. **`review-docs-drift`** — 16: prose against source for identifiers, line-number citations, cross-references, and prerequisites the runbook asserts but nothing checks. Merged and deduped to 23.

The fixes were themselves reviewed, and several first drafts rejected: one referenced a `FLEET_ZONE` that does not exist, which under `set -euo pipefail` would have aborted step 2.2 on every run; the PEM idempotency guard cleared `PEM_FILE` and fell into the "no PEM supplied" branch, telling the operator to run an import just ruled unnecessary; and the first `check_toolchain()` missed the logged-out case it exists to catch.

**Fixed:**

- *A signing probe that verified a version the deployment does not use.* It picked the highest enabled KMS version rather than the chart's pinned `githubMinter.kms.keyVersion`, so a rotation to v2 would leave the probe green while every lease failed on v1; it now reads the pin and fails when that version is not `ENABLED`.
- *A mapping check that read the wrong copy of the file.* Prow builds an eval run from `main`, not this branch, so a branch-only row is invisible to every other PR's presubmit; the check now reads both copies and reports exit 2 when the row is branch-only, resolving the remote by fetch URL because `origin` here is a fork.
- *A "not yet on main" the check had not earned.* `git show <remote>/main` reads the last fetch, and a checkout predating `gitops_repo_for_project()` (2026-08-21) has no function to search — three false positives in the first six-project run — so an unreadable copy is now unverified rather than absent, with the snapshot's date in the warning.
- *Counts in prose that had gone stale three times.* Rewritten count-free and pointed at the roster CI reads rather than incremented, since a bumped count re-arms at the next onboarding; a sweep found a fifth instance in `hack/ci-eval-pr.sh`.
- *A boundary widened without a human seeing it.* Step 1.4 PUT the GitOps repo into the App installation over the API. Deleted — the note is always printed and the repo is added in the GitHub UI, where the person approving sees what they are approving. A y/N in a terminal is not a security review.
- *A verifier that could not tell "absent" from "could not look".* Every check read a non-zero `gcloud`/`gh` exit as "the resource is not there", so a missing binary or an expired credential produced a confident, specific, wrong "not provisioned". `check_toolchain()` now gates everything and exits 2; its `gcloud` arm reads stdout rather than the return code, because `gcloud auth list --filter=status:ACTIVE` exits 0 with empty output when logged out.
- *Flags that did not reach where they claimed to.* `--app-id` was accepted and never forwarded to either verifier call. `--region` reached two stacks but not `bench/tf/fleet`, which is zonal and pinned; any other region silently split the fleet from the host cluster while the name-matching verifier reported green.
- *A re-run that was not idempotent.* A second `--pem-file` run created KMS key version 2. The chart pins version 1 and the verifier probes the highest enabled version, so a divergent second import makes the verifier green while the minter fails in CI. Guarded, with its own skip flag so the "no PEM supplied" branch stays reachable.
- *A relative `--pem-file` resolving against the wrong directory* after five expensive steps had applied. Absolutized at parse time.
- *Attestation from a keypress.* Bare Enter at the App-installation prompt recorded a confirmation the operator had not given. It re-prompts, and is gated on exit 2 — a hard failure is not something an operator can attest past.
- *Identifiers and attributions pointing at the wrong source.* `ALLOW_UNENCRYPTED_SECRETS` was attributed to `hack/ci-deploy.sh` in three places; that script exports the default but never invokes `install.sh`, so the enforcement is `ensure_existing_cluster_cmek()`. A hardcoded installation ID, a wrong header path, and line-number citations into `ci-deploy.sh` were replaced with queries or symbol references.
- *Runbook claims the source did not support.* §0 overstated billing enforcement and omitted the mapping precondition; the script was never introduced as the way to do §§1–6, and four of its flags were undocumented; §1 omitted `compute.googleapis.com`; §2's `terraform apply -var=...` could not work, since three variables have no defaults; §3 had the IAM OR/AND structure backwards; §5's counts were stale, as they were in four other files; §6 claimed fixture verification that nothing performed — now true, because the verifier performs it; and the opening paragraph put the Boskos roster in `oss-test-infra`, contradicting §8 of the same page.
- *A check asserting the wrong verb against the wrong identity.* `check_artifact_registry` only ever asked whether **something** could push — an OR over the Cloud Build and Compute SAs. Nothing asked whether the host cluster's **nodes** could pull, and they are not the same identity or the same permission. A project where Cloud Build holds writer and the node account holds nothing passed every check and would have died at `ImagePullBackOff` on the Deployments the smoke test waits for: the exact "looks provisioned, dies on lease" this script exists to prevent. The node account is now read off `nodePools[].config.serviceAccount` rather than assumed to be the Compute default; the seeded trio is excluded because it pulls no kube-agents image; and an unreadable cluster listing warns to exit 2, because "could not look" is not "cannot pull". Every pool project passes it — they hold `roles/editor` — so this closes a latent gap rather than a live outage.
- *A check whose name promised more than it did.* `GKE Clusters & Seeded Fleet` matched cluster names and nothing else. Renamed, with the fixture-level claim moved to the check that can actually make it.
- *An ownership gap in the docs map.* Two rows, the second naming its mirrors so someone changing `is_valid_cmek_encryption_state()` is not sent to the copy.

**Not fixed, with reasons about this change:**

- *"The early `exit 1` at step 5 skips the summary of remaining steps."* Rejected. That exit is the only one in the tail; removing it makes a hard verification failure exit 0 onto a banner reading "Nothing failed", which is how evals-3 got registered while broken. The narrowing of the membership prompt to exit 2 was taken.
- *"The provisioner does not add the Prow identity to `fleet_reader_token_creators`."* Correct and deliberate: `hack/fleet-kubeconfigs.sh` uses the reader account only when `FLEET_READONLY_SA` is exported, which no Prow job does, so binding it now grants nothing that gets used. A comment at the apply says so rather than leaving the omission to read as an oversight.
- *"App ID `4675512` is duplicated across four files."* Real and pre-existing; nothing here makes it worse. Item 5 in Context.
- *The `full-install` secrets the evals-5 plan printed.* Not rotated, and that is a decision rather than an omission — the rotation is the more dangerous operation. `full-install` has no committed backend, so which project a `-replace` lands on comes from a gitignored `backend_override.tf` and an operator-local `terraform.tfvars` that the script writes and removes again; and `api_server_key` is a required no-default variable, so any targeted apply either rotates it too or needs it read back out of state in plaintext first. The exposure is a session-store key and an API key on an unleased CI sandbox; the repair is a live apply against a project that currently works. Item 3 in Context is what makes it safe, and the rotation waits for it.

**What the passes could not cover.** Neither reached a running install, so nothing either found was confirmed against live behaviour by the pass itself; the six-project run above is that evidence, and it postdates these fixes. Neither reads Terraform state, so claims about what is applied where rest on the runs recorded above. Both predate the fixture check and the two verifier fixes, which rest on the six-project run and the tests rather than on a hostile read. They disagreed once, on a line-number citation; I read the source and `review-docs-drift` was right.

## Risk & Rollout

Two new scripts and a documentation restructure; no runtime code path, chart template, or operator behaviour changes. `hack/ci-deploy.sh` gains mapping rows that only take effect for projects Boskos cannot yet hand out.

The one new failure mode is sharp in a single direction: run after a project is registered, `provision_ci_pool_project.sh` rotates that project's `api_server_key` out from under a live lease, and the agent pod keeps serving the old value until something restarts it. Nothing in the script stops you and nothing in the resulting failure looks like its cause. §8 of the runbook says so; the durable fix is the follow-up issue in Context.

The fixture check is also the slowest thing the verifier does — three `get-credentials` calls and a `kubectl` probe per fixture, against clusters in another project — so it carries a 600s ceiling rather than the 120s the single `gcloud` calls use, and no `kubectl` on the machine is exit 2 with a warning rather than a red project.

Reverting is deleting three new files and restoring the Markdown and the mapping rows. The three projects Boskos hands out today are untouched by this PR, and evals-4, -5 and -6 stay provisioned and unregistered either way.


scripts/verify_ci_pool_project.py
```diff
@@ -0,0 +1,1204 @@
...
+            passed = False
+            details.append(f"Failed parsing Minter GSA policy: {exc}")
+
+    # Last, and only once the key is known to exist, hold enabled material, and
+    # carry the right algorithm -- a probe against a key already known to be
+    # wrong costs a network round trip to restate what was just reported.
+    probe = ""
+    if enabled_versions and algorithm_ok:
+        status, message = _probe_github_app_identity(
+            project_id, location, sorted(enabled_versions, key=lambda v: int(v) if v.isdigit() else 0)[-1], app_id
```
jayantid: 🟡 **Not merge-blocking — it arms at the first key rotation.** After a rotation this check goes green on a version the pool's minter never signs with.

The probe picks the highest ENABLED version, on the comment's rationale that "Minty picks it up" after a rotation. That is the k8s-operator path, where `provision_10_deploy_github_minter.sh` resolves the active version at deploy time. The pool deploys via helm: `charts/kube-agents/templates/github-minter.yaml:142` renders `cryptoKeyVersions/{{ keyVersion }}`, `values.yaml:305` defaults it to `"1"`, and `hack/ci-deploy.sh`'s `GITHUB_MINTER_ARGS` never overrides it.

Rotate any pool project's key as `token-minter.md` describes — import v2, disable v1 — and this probe verifies v2, the run prints `ALL CHECKS PASSED`, and every lease deploys a minter pointed at the disabled v1: readiness fails, `helm --wait` kills the run 15 minutes in. The multi-version warning's advice ("disable the superseded one") follows the same wrong model.

Fix: probe the version the chart will use — read the pinned default out of `values.yaml` the way the tests read `fleet-kubeconfigs.sh`'s format string — and fail when that version is not ENABLED.

scripts/provision_ci_pool_project.sh
```diff
@@ -0,0 +1,578 @@
...
+
+# Both Terraform binaries, because this script needs both and they are not
+# interchangeable here: Step 2.1 and Step 3 drive `terraform` (lifecycle.sh
+# hardcodes it too), Step 2.2 drives `tofu`. Each state prefix was written by
+# whichever binary owns its step, so this is not a mix to resolve by picking one
+# -- swapping a binary would point it at state the other wrote. Checked here for
+# the reason at the top of Step 0: without it, a missing `tofu` surfaces as
+# "command not found" at Step 2.2, twelve minutes and two applies in.
+MISSING_TOOLS=()
+for tool in gcloud gh python3 terraform tofu; do
```
jayantid: 🟡 **Not merge-blocking; one-line fix.** A missing `go` kills the run at step 4, after ~36 minutes of applies — the exact failure this check exists to prevent.

The loop checks the five binaries the early steps need and none of the later ones: `jq` (step 1.4, line 316 — under `set -e` the failed substitution ends the run), and `git` and `go` (step 4's minty clone and `go run`). The comment above the loop states the contract: "without it, a missing `tofu` surfaces as command not found at Step 2.2, twelve minutes and two applies in." `go` missing with `--pem-file` reproduces that at the most expensive point in the run. (`openssl` is also unchecked, but full-install's `api_server_key` validation catches that one loudly — verified.)

```bash
for tool in gcloud gh python3 terraform tofu jq; do
...
if [ -n "${PEM_FILE}" ]; then
  for tool in git go; do command -v "${tool}" >/dev/null 2>&1 || MISSING_TOOLS+=("${tool}"); done
fi
```

scripts/verify_ci_pool_project.py
```diff
@@ -0,0 +1,1204 @@
...
+    pattern = rf"{re.escape(project_id)}\)\s*echo\s*[\"']?{re.escape(expected_repo)}[\"']?"
+    if not re.search(pattern, func_body):
+        return CheckResult(
+            "Codebase GitOps Mapping",
+            False,
+            f"No mapping for {project_id} in gitops_repo_for_project() in hack/ci-deploy.sh",
+            details=[f"Expected: {project_id}) echo \"{expected_repo}\" ;;"],
+        )
+
+    return CheckResult("Codebase GitOps Mapping", True, f"Mapped to {expected_repo}")
```
jayantid: 🟡 **Not merge-blocking; a wording gap with one live instance.** The mapping check reads this checkout, but the verdict line makes a claim about main.

`check_codebase_mapping` reads `hack/ci-deploy.sh` from the local tree. Run from a branch that carries the mapping row — which is how evals-4/-5/-6 produced their green transcripts, since their rows ride in this very PR — the run ends with "safe to register in Boskos" while a presubmit that leases the project runs main's `ci-deploy.sh` and dies at `gitops_repo_for_project()`'s refusal. That is the evals-3 outage, reachable through this script's own verdict. The runbook's "land the mapping before provisioning" covers the ordering, but the tool contradicts it exactly when someone is about to make the mistake.

Fix: when `git show origin/main:hack/ci-deploy.sh` lacks the row, report the item unverified (exit 2) with "mapped in this checkout, not yet on main"; skip quietly when git or the remote is unavailable.

docs/site/src/content/docs/deploy/ci-pool-projects.md
```diff
@@ -121,24 +143,33 @@ The evaluation scenarios that exercise the GitOps workflow — the six fleet-aud
...
 
 The repository is kept private: it is throwaway state a bot rewrites on every run. [`examples/gitops-repo`](https://github.com/gke-labs/kube-agents/tree/main/examples/gitops-repo) is the layout an audit expects to find, not a required seed — the current pool repositories carry only a LICENSE and a README, because an audit works against an empty tree and a `remediation.path` that does not exist degrades to a manual finding rather than failing the run.
 
-> **The pool's three projects are all provisioned, as of 2026-08-24.** `kube-agents-evals-3` was added to the Boskos pool on 2026-08-21 with only its GCP half done, so every presubmit that leased it stopped at `gitops_repo_for_project()`'s unmapped-project refusal. That is closed. Verified in all three projects:
+> **All five projects are provisioned; three are leasable.** `kube-agents-evals-4` (2026-08-25) and `kube-agents-evals-5` (2026-08-26) were provisioned by `scripts/provision_ci_pool_project.sh` and verified before registration rather than after — the order this page prescribes. Neither has a Boskos entry yet, so nothing leases them. Verified in all five:
```
jayantid: 🟡 **Docs only, but it is this PR's own thesis.** At this head, three passages disagree about how many projects exist.

- This line: "All five projects are provisioned; three are leasable... Verified in all five" — the mapping table above lists six, and the PR body's evals-6 transcript is the run that closed the Workload Identity and fixture claims.
- Line 197: "Done for **all four** provisioned projects."
- `bench/tasks/DRAFTS.md:34`: "All four repos exist, all four are on App 4675512"; `DRAFTS.md:73`: "All five provisioned pool projects", with evals-6 absent from the dates.

The PR deletes count-prose elsewhere for exactly this reason — "count-prose that had gone stale three times" went stale a fourth time inside the change that says so, one project-onboarding later. Bump them all to six, or better, degrade them to the query and verifier pointers this PR already switched to in two other places.

jayantid (approved):

scripts/verify_ci_pool_project.py
lapis2002: Fixed. The probe now reads the pin out of values.yaml and fails when that version isn't ENABLED, falling back to the old behaviour with a warning only if the chart can't be read. The multi-version warning is reworded off the same wrong model, it now says the unpinned ones are keys that still work and shouldn't.

Two tests went with it: test_probe_uses_the_highest_enabled_version_not_a_hardcoded_one was asserting exactly the behaviour you're calling a bug, and test_several_enabled_versions_warn_but_pass expected the probe on v2. Both replaced with cases asserting the pinned version wins, plus one for a rotation that disables it.

I re-verified all six pool projects and each has a single ENABLED version, so the pin and the highest agreed everywhere. This is pre-emptive

scripts/provision_ci_pool_project.sh
lapis2002: Fixed, git, go and jq are added to the loop now.

scripts/verify_ci_pool_project.py
lapis2002: Updated with your exit-2 suggestion.

Verified across the pool: evals, -2 and -3 pass clean with "in this checkout and on gke-labs/main"; -4, -5 and -6 now report not yet on gke-labs/main and exit 2, where they previously printed the green transcript you're describing.

docs/site/src/content/docs/deploy/ci-pool-projects.md
lapis2002: Updated docs.

Issue #994, opened 2026-08-27
evals-6's first lease redded a PR: BENCH_GITHUB_TOKEN can't read the new pool repos' ledgers; new projects should be eval-proven before Boskos registration


jayantid:
`kube-agents-evals-6` was registered into the Boskos pool sometime after #979 merged (2026-08-26 21:05Z said "provisioned and verified but no Boskos entry yet"), and this morning's smoke run on PR #956 was the first eval ever to lease it — [build `2092820036916875264`](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/956/pull-kube-agents-smoke-test/2092820036916875264). Two tasks failed, and both are first-contact gaps with the new project rather than the PR under test. Until they close, **any audit-carrying run that leases evals-4/-5/-6 reds its PR**, so a lease from the new projects is currently a coin-flip presubmit failure.

## 1. `BENCH_GITHUB_TOKEN` cannot validate ledgers on the new repos — #979's own unconfirmed item, now confirmed the wrong way

`compliance-rbac-overgrant` failed `VerificationCorrectness=0.0` on evals-6 — while the agent-side path worked perfectly:

- Ledger issue **kube-agents-evals-6-infra#1** exists: "[audit] Security & RBAC Posture Audit — 13 findings (6 critical)", filed 04:41Z mid-run, names `debug-binding` → `cluster-admin`. So the minter, App installation, gitRepo wiring, and `audit_report.py` all work on the new project.
- `hack/fleet-kubeconfigs.sh` reported `7 role(s) written … 0 whose fixtures were not present (project kube-agents-evals-6)`.

What failed is the *verification* read-back: `ledger_issue_contains` reads the ledger with `BENCH_GITHUB_TOKEN`, and #979's A4 note said whether that token carries `issues: read` on the pool repos was "not observable from this repository and has not been confirmed." For the original three repos, the 2026-08-26 runs on evals-3 confirmed it works (compliance passed its ledger checks twice). For the new repos it evidently does not — consistent with a credential scoped to a repository list that predates the expansion.

**Ask:** extend the token's scope (or regenerate it) to cover `kube-agents-evals-4/-5/-6-infra`, and make "the bench token can read the new repo's issues" a check in `scripts/verify_ci_pool_project.py` — the current verification proves the provisioning half but not the eval-runner half, which is exactly the half that failed on first contact.

## 2. First-contact task variance on the new fleet (softer, for awareness)

[...]

Issue #987, opened 2026-08-26
Smoke runs upload no artifacts: results.json never leaves the pod, so no flake rates, no judge-drift evidence, no #899 fixtures


jayantid:
Nothing a smoke run produces survives the pod. The Prow GCS prefix for `pull-kube-agents-smoke-test` builds contains only the entrypoint's own files — `build-log.txt`, `clone-log.txt`, `clone-records.json`, `finished.json`, `podinfo.json`, `prowjob.json`, `sidecar-logs.json`, `started.json`. There is no `artifacts/` prefix: `results.json`, the per-task scored records, and `hack/ci-eval-pr.sh`'s own `scoring_failure_*.log` / `resource_prep_failure_*.log` never leave the pod. (The script already writes them to `${ARTIFACTS:-/tmp/artifacts}` — the copy destinations exist in the code; the uploads just never appear, so either `${ARTIFACTS}` is unset in this job or the sidecar is not picking the directory up. First step is establishing which.)

## What this costs, concretely

- **No flake statistics.** `kube-agents-testing-plan.md` gates on pass-rate *drops* over repeated runs; there is no run history to compute a rate from. Every flake triage so far (issue #957, the 2026-08-26 run) has meant scraping `build-log.txt` by hand and, when the log had already rotated, losing the evidence entirely.
- **No fixtures for the scoring engine.** Issue #899's lane says to develop the comparator against "a handful of real `results.json` records" captured now. The 2026-08-26 run alone produced five records including two genuinely pathological ones (a 502 graded as an answer; a task whose worker completed while the harness was deaf). All gone.
- **No baselines later.** #899's non-inferiority gate compares a PR's pass rate to `main`'s; that requires persisted per-run records keyed on the five-version tuple. This upload is the prerequisite.
- **Judge-drift evidence is anecdote.** `agent-kanban-smoke`'s OutcomeValidity has drifted 0.5 → 0.5 → 0.3 across runs and `compliance-rbac-overgrant` scored 0.1 with exact checks green (2026-08-26); with persisted records these are a query, without them they are screenshots.

## Shape of the fix

1. Establish why the existing `${ARTIFACTS}` copies do not surface: whether the Prow job wires `artifacts:` / the env var, and whether the decoration sidecar uploads the directory. This may be an oss-test-infra change, a script change, or both.
2. Upload per task: `results.json` (the scored record), and keep the existing failure-log copies.
[...]

Issue #986, opened 2026-08-26
Delegation-path transport exhaustion is still graded: route status-turn 502s to INFRA and reset the tunnel between retries


jayantid:
#959 fixed the opening turn: a transport failure there is retried with a tunnel reset, and exhaustion returns the `KUBE_AGENTS_INFRA_FAILURE` marker so `hack/ci-eval-pr.sh` classifies the task INFRA instead of grading a proxy error page. **The delegation-poll path got neither half.** In `_await_delegated_work`, a status turn that fails in transport is retried up to `_MAX_TRANSPORT_FAILURES`, but:

1. **Exhaustion is appended and graded, not classified.** The retry-exhausted branch appends "status turns failed in transport N times running" to `result.errors` and proceeds to validation — the record reaches the judge and the task reds the PR as a quality failure.
2. **Retries go through the same dead tunnel.** The backoff is one poll interval; there is no `_reset_port_forward` between attempts, so the live-listener-dead-upstream state #959 documented for the opening turn is unescapable here too.

## Observed, 2026-08-26

Build `2092638061140643840`, task `upgrade-readiness-lagging-cluster`: three `status turn failed (n/3): HTTP 502` at 16:38/16:43/16:49 (each poll hung ~300s against the dead endpoint first). **The delegated worker was alive the whole time — it filed ledger issue #3 (7 findings, `seeded-b` ×36, `master-behind` ×9) at 16:42Z, mid-outage.** The task scored `VerificationCorrectness=0.0` and redded the run. Work done, evidence published, harness deaf, PR blamed.

## Fix shape

Mirror the opening turn:
- `_reset_port_forward(local_port)` between status-turn retries (absorbing its `RuntimeError` the same way, counted not raised).
- On exhaustion, return `_infra_failure(...)` (empty output, marker on `errors[0]`) instead of appending-and-grading, so the classifier routes the task to `INFRA_FAILED_TASKS`. The all-INFRA guard (`EVAL_INFRASTRUCTURE_DOWN`) still catches a fully dead endpoint.
- Tests mirroring `bench/tests/test_harness.py`'s opening-turn set: storm→INFRA, recovery-on-retry→normal, reset-called-between-attempts, respawn-failure-absorbed. The existing stub already has `fail_on`/`drop_on` knobs.

## Scope note

[...]

PR #995, opened 2026-08-27
fix(ci): grant the Prow runner its project roles and check the platform GSA both ways

lapis2002:
## Summary

A presubmit doesn't run inside the project it tests. It runs on `build-kube-agents`, leases a project from Boskos, and reaches in as `prowjob-default-sa@kube-agents-prow` — an account external to the pool project, whose grant nothing in that project's own configuration implies. So `kube-agents-evals-4` through `-6` were provisioned, verified green, and registered without it, and stayed that way until a lease of `-6` failed on `Required "container.clusters.get" permission(s)`. This PR makes the provisioning script grant those roles and the verifier refuse to pass a project missing them, and does the same for the platform agent's own read-only role set, which had drifted the other way — older projects still hold the `container.admin` grants #961 removed from the module, because re-running the install doesn't strip what it no longer grants.

## Why This Change

The pool exists to be leased at random. That's what makes divergence between projects expensive: a project missing the runner's roles doesn't fail its own presubmits, it fails whichever pull request draws it, and the failure names a permission rather than the project. Six projects are onboarded and #903 takes the pool to fifteen, so every gap that isn't codified now is nine more chances to repeat by hand.

The platform GSA is the same problem pointed at grading rather than at CI. It's the identity the agent under test authenticates as, so a project where it holds more than the read-only eight lets an eval case pass on the grant instead of on the agent — and with random leasing, the same case grades differently depending on which project it lands in.

## What Changed

- `scripts/provision_ci_pool_project.sh` grants the twelve project roles `kube-agents-evals` holds. Matched literally rather than derived from permissions: the set isn't minimal, but matching a project that presubmits have passed on is the point. No Artifact Registry role — `hack/ci-deploy.sh` pushes through `gcloud builds submit`, so Cloud Build holds those credentials and this account never touches the registry.
- `scripts/verify_ci_pool_project.py` checks both identities off one `get-iam-policy` read. The runner's set is checked for absences only; the platform GSA's is checked in both directions, because an extra role there is the failure mode. Conditional bindings count for neither — a condition narrows when a grant applies, not who holds it, so counting one would pass a project the runner still cannot use. `allUsers` and `allAuthenticatedUsers` are reported whatever the condition, and separately from both sets, because the two membership scans look for one literal member each and would not see them. A denial on that read reports as unverified rather than as both identities holding nothing, per #1008's model, which this branch merges.
- Two unit tests bind the hardcoded role lists to the places that grant them. The platform GSA's eight must equal `terraform/modules/kube-agents-iam`'s `project_roles` default *and* its `local.read_only_roles`; the runner's twelve must equal the `for role in …` loop in the provisioning script and the repair block on the prerequisites page. All three copies of the runner set agreed by hand and nothing bound them, which is how #966 stayed invisible: a role dropped from the script leaves a project the verifier still passes, registered, dying on its first lease.
- Three verifier checks that could pass wrongly. `_mapping_row_present()` matched unanchored, so `kube-agents-evals)` matched inside the `-2` row and a project read as mapped when its `case` arm was never written; it also accepted a commented-out row, `-infra-old`, and — via `echo\s*` — `echo` with no space after it, which is a command shell cannot find and which quoting does not save, since word splitting runs before quote removal. `_upstream_remote()` matched the slug as a URL suffix, so `not-gke-labs/kube-agents` and the right path on any host both resolved as the merge target; narrowing it then over-rejected `ssh://git@github.com:22/gke-labs/kube-agents` (the port parsed into the path) and `https://github.com/GKE-Labs/kube-agents` (the host was lowercased, the path wasn't). Both are URLs `git clone` accepts.
- `EXIT_USAGE = 64` (EX_USAGE), so a mistyped flag no longer exits 2 alongside "nothing failed, go confirm these by hand". Python still exits 2 when the script *path* is wrong — that happens before the file is read, and the epilog and docs say so. The provisioning script's closing summary now has an arm for each of `0`, `1`, `64` and the `2`/unknown catch-all; it previously had two, written when `2` was the only other code that could reach it, so a usage error and a failure on the second verifier invocation both reported "Nothing failed".
- The fresh-fleet NOTE states its condition instead of asserting a replant. It still prints on every run that reached the fleet step — only `--skip-fleet` suppresses it — but a re-run that planted nothing no longer reads as one that did.
- Docs and bench drafts: six places said the pool was three projects, and the age-gate activation dates were computed against `kube-agents-evals-3` and have passed. Counts now point at the roster table; the gate is stated as the subtraction it is, with `fleet_wide_cost_analysis_sop.md` §3.4/§3.7 as the source of its windows. The token-minter page no longer implies the key can be rotated without a redeploy. `docs/designs/bench-fleet-catalog.md` no longer claims every project after `-2` came from the script, and now separates the mapped table from the leasable roster — this PR is what pushed those two lists apart.

## Context

Part of #903 (lane: eval capacity — codify project onboarding, take the pool to 15). The `-6` lease failure is #966. The platform GSA drift is the tail of #961, which narrowed the module.

Per #903, `terraform apply` and mutating `gcloud` IAM are human steps: the grants described here were applied by hand against the live projects, and this PR is what makes them reproducible for 7–15.

`main` was merged in rather than rebased onto, because #1008 rewrote the function this PR extends and eleven commits would have replayed the same integration eight times. One resolution overrides `main` on purpose: the callout at `ci-pool-projects.md:178` landed on `main` saying a project enters the table once both halves are provisioned. That is backwards — `provision_ci_pool_project.sh:178-181` refuses to run against a project `gitops_repo_for_project()` does not map, so the row is written first. This branch keeps its own wording.

Follow-ups not in this PR: host cluster nodes still run as the Compute Engine default SA with `roles/editor`; `BENCH_GITHUB_TOKEN`'s coverage of the newer pool repos is #994; and nothing repairs an already-registered project that drifts — the verifier reports, per the rule that presubmits assert preconditions rather than fix the host.

## Testing

- `make test-python` — verifier suite 113 → 181 cases, all green. 134 of those are this branch; the rest arrived with the merge from `main`.
- `make docs-check` — green.
- `make bench-case-check` — 24 bench cases OK.
- `bash -n scripts/provision_ci_pool_project.sh` — clean.
- Each new regression test was checked to fail against the pre-fix code before its fix landed. For the mapping regex that meant asserting `echogke-agentic/…` is rejected; for the URL matcher, that both over-rejected forms resolve; for the role-list tests, mutating one copy of each list.
- The mapping regex checked against the real `hack/ci-deploy.sh` rather than fixtures alone: all ten mapped projects match, and the upstream remote resolves to `gke-labs`.

### Live validation

**`kube-agents-evals-7`, provisioned end to end by `scripts/provision_ci_pool_project.sh` on this branch, then exercised by a real presubmit.**

Provisioning ran the host cluster, the seeded fleet, the `ci-pool-minter` Terraform, and the Minty PEM import into KMS. It exited **2**, which is the expected result pre-merge: `check_codebase_mapping()` compares against `main`, and `-7`'s `case` arm only exists on this branch. Terraform plan output is deliberately not reproduced here — the fleet plan prints `SESSION_KV_API_KEY` and `SESSION_KV_SALT` in plaintext.

The presubmit is `pull-kube-agents-smoke-test` run `2093052253513453568`, commit `442cbf5`, pinned to `-7` by `442cbf53`. The pin existed only to give a newly provisioned project one real run before Boskos registration; on `main` it would serialise the whole pool behind one project, so `8e42c3b4` reverts it. The branch is unpinned as of that commit.

The pin behaved as designed, which is itself the check that the pin is safe: Boskos leased `kube-agents-evals-5`, left it idle, and released it cleanly at the end, while build, deploy, eval and teardown all followed `PROJECT_ID=kube-agents-evals-7`.

What `-7` proved, layer by layer:

```
Cloud Build 16414ac3-1570-4e68-9547-20671dd810ef   STATUS: SUCCESS   9m41s
  → us-central1-docker.pkg.dev/kube-agents-evals-7/kube-agents/platform-agent:pr-995-442cbf5
  → …/credential-proxy:pr-995-442cbf5
  → …/kube-agents-operator:pr-995-442cbf5

helm install   STATUS: deployed   REVISION: 1
  deployment/platform-agent-gateway rolled out; Agent API responding      (deploy total 943s)

GitOps repo:         gke-agentic/kube-agents-evals-7-infra (mapped from PROJECT_ID=kube-agents-evals-7)
GitHub token minter: enabled for gke-agentic/kube-agents-evals-7-infra (app 4675512)

gpu-stress-test-diagnosis: tofu deployer wrote synthetic Cloud Logging entries,
  polled until queryable (10s), and destroyed cleanly

teardown succeeded; Boskos lease released
```

The two lines that matter for this change are the GitOps mapping and the minter: the `case` arm added here resolves live under a real lease, and the KMS key the provisioning script imported the App PEM into is the one the minter actually uses.

**Task results: 11 of 13 passed.** `compliance-rbac-overgrant` and `cluster-agent-healthy-workload-no-finding` both failed on `VerificationCorrectness=0.0`. Neither failure is project-specific:

- `compliance-rbac-overgrant` is #986 — the delegation-poll path grades transport exhaustion as a quality failure instead of classifying it INFRA, and #986 names this task as the only active one that delegates. The run logged "credential proxy command execution failed" against three of the four fleet clusters.
- `cluster-agent-healthy-workload-no-finding` returned a delegated stub ("Delegated diagnostic tasks to the cluster agents. See follow-up report for results.") in place of the observed replica health. Same shape, but #986's scope note predates #982 activating the cluster-debugging cases, so treating it as the same bug is inference rather than evidence.

Per-task history can't be used to confirm that: #987 means results never leave the pod, so no baseline exists.

**`kube-agents-evals-8`, `-9` and `-10` were provisioned by the same script on 2026-08-27**, three fresh projects with no hand-editing between them. Terraform plan output is deliberately not reproduced for these either, for the same reason as `-7`.

**The verifier was then run against the whole pool — all ten projects, one command, this branch's code:**

| Project | Exit | Non-green items |
| --- | --- | --- |
| `kube-agents-evals` | 0 | — |
| `kube-agents-evals-2` | 0 | — |
| `kube-agents-evals-3` | 0 | — |
| `kube-agents-evals-4` | 0 | — |
| `kube-agents-evals-5` | 0 | — |
| `kube-agents-evals-6` | 0 | — |
| `kube-agents-evals-7` | 2 | `[?]` Codebase GitOps Mapping — mapped here, not yet on `main` |
| `kube-agents-evals-8` | 2 | `[?]` Codebase GitOps Mapping — mapped here, not yet on `main` |
| `kube-agents-evals-9` | 2 | `[?]` Codebase GitOps Mapping — mapped here, not yet on `main` |
| `kube-agents-evals-10` | 2 | `[?]` Codebase GitOps Mapping — mapped here, not yet on `main` |

Eight checks each, and the mapping check is the only one that is not green anywhere. It is the pre-merge state rather than a defect: `check_codebase_mapping()` compares against `main`, and the `case` arms for `-7` through `-10` only exist on this branch. Those four clear on merge, which is what gates their Boskos registration.

Three things this run establishes that the unit tests cannot:

- **`Service Accounts & IAM Grants` is green on all ten.** Every project holds the Prow runner's twelve roles and the platform GSA's exactly-eight, checked in both directions against a live `get-iam-policy`. That is #966 closed by construction rather than by assertion, and it is the reason this pull request exists.
- **No live project carries a public binding.** The new `allUsers` / `allAuthenticatedUsers` scan ran against ten real policies and reported nothing, which is the result you want and the one a fixture cannot give you.
- **`_is_upstream_url()` resolved a real remote.** The amber detail names `gke-labs/kube-agents` and dates the checkout, so the narrowed matcher works outside its tests.

<details>
<summary>Full output for <code>kube-agents-evals-10</code> — the amber case, verbatim (project number redacted)</summary>

```
================================================================================
 Pre-flight Onboarding Verification: kube-agents-evals-10
================================================================================

[?] Codebase GitOps Mapping: Mapped to gke-agentic/kube-agents-evals-10-infra in this checkout, not yet on upstream/main
    ? kube-agents-evals-10 is mapped here but not on gke-labs/kube-agents main (this
      checkout's upstream/main is dated 2026-08-27). A presubmit that leases it runs
      main's hack/ci-deploy.sh and stops at gitops_repo_for_project()'s refusal, failing
      that run. Land the mapping on main before registering kube-agents-evals-10 in
      Boskos. If it landed since this checkout last fetched, run `git fetch upstream main`
      and re-run.
[✓] GCP Project & APIs: Project number: <redacted>, all 9 APIs enabled
[✓] Service Accounts & IAM Grants: Workload Identity, Prow runner and platform GSA project roles, and cross-project AR reader grants verified
[✓] Artifact Registry Repository: kube-agents (us-central1) present with a cleanup policy, push rights, and node pull rights
[✓] GKE Clusters & Terraform State: All 4 clusters (platform-agent-host, seeded-a, seeded-b, seeded-c), CMEK, and state bucket present
[✓] Seeded Fleet Fixtures: All 7 fixture roles planted and reachable
[✓] GitOps Repo & GitHub App Installation: Repo gke-agentic/kube-agents-evals-10-infra (private); membership operator-confirmed, not machine-checked
[✓] Token Minter KMS & GSA: Import-only signing key v1 (the version the chart deploys) ENABLED, minter GSA can sign and be impersonated, signature accepted by GitHub as App 4675512 (kube-agents-evals-token-minter)

--------------------------------------------------------------------------------
MANUAL VERIFICATION REQUIRED. Nothing failed, but 1 item(s) could not be checked
automatically. Confirm each one above before registering kube-agents-evals-10 in Boskos.
--------------------------------------------------------------------------------
```

</details>

The run used neither `--app-id` nor `--location`, so the defaults carried App `4675512` and `us-central1` — worth recording, because the runbook passes both explicitly and does not have to.

**Not covered.** The verifier's strictness against an already-drifted project was not exercised against a real drift: the ten-project run above is what establishes that no live project currently holds an extra platform-GSA role, so there was nothing to catch. `prettier` and the presubmit are the two checks that cannot run on the authoring machine.

## Self-Review

Both passes — `review-adversarial` and `review-docs-drift` — ran through `/pr-preflight`, in subagent contexts that did not write the change. Angles covered: false-pass paths in the verifier's string matching, the three unlinked copies of each role list, exit-code semantics across both verifier invocations, IAM policy reads that a literal-member scan can miss, docs claims that source contradicts, and stale counts and dates. A second pass over the branch as it then stood added: comment claims the code contradicts, callouts that contradict themselves, hand-copied inventories of machine-readable state, and activation notes in bench cases that this PR's IAM changes invalidate.

Twenty-eight findings across two passes. The second pass numbers its own from 1; they appear here from 20 to keep them distinct. Dispositions:

| # | Finding | Disposition |
|---|---|---|
| 1 | `PLATFORM_GSA_ROLES` was hardcoded with nothing binding it to the module | **Fixed.** Test asserts it equals both `project_roles`' default and `local.read_only_roles`. |
| 2 | Collides with #945, which also touches the module's role set | **Not fixed here.** `PLATFORM_GSA_ROLES` does not exist on `gke-labs/main` (verified), so the collision cannot fire until this merges. Merge order coordinated with @Fuxiao-Gao. |
| 3 | Boskos roster claim | **No change needed.** The wording already on the branch is correct. The finding said the claim sits inside hunks this change edits; it does not — that part of the finding is wrong. |
| 5 | `docs/designs/bench-fleet-catalog.md:62-68` says every project after `-2` was provisioned by the script (false for `-3`) and calls the prerequisites table "the pool roster" | **Fixed.** `-3` is now named as predating the script, matching `ci-pool-projects.md:178` and `provision_ci_pool_project.sh:220`; the mapped table and the leasable roster (`gke-internal/test-infra`) are now distinguished, since this PR is what pushed them ten rows apart from six. |
| 6 | `PROW_RUNNER_ROLES` exists in three unlinked copies | **Fixed.** `ProwRunnerRolesMatchGrantersTest` asserts the constant equals the provisioning script's grant loop and the prerequisites page's repair block. |
| 8 | `_is_upstream_url()` rejects `ssh://git@github.com:22/gke-labs/kube-agents` | **Fixed.** `(?::\d+)?` between host and separator. Confirmed failing first. |
| 9 | It also rejects `https://github.com/GKE-Labs/kube-agents` | **Fixed.** Path lowercased alongside the host. Confirmed failing first. |
| 10 | `_mapping_row_present()` uses `echo\s*`, so `echogke-agentic/…` reads as mapped | **Fixed.** `echo\s+` on both arms, plus a fourth docstring bullet. The only false *pass* in the set: the verifier would have cleared the project for registration and `gitops_repo_for_project()` would fall through to `*)` at lease time. |
| 11 | The closing summary's two-branch tail reports "Nothing failed" for exit 64 | **Fixed, and wider than reported.** Four arms now. The finding said codes 1 and 2 are handled earlier — true only of the first invocation; a failure on the re-run also lands in the catch-all, which is the more dangerous half. |
| 12 | The fresh-fleet NOTE fires on idempotent re-runs | **Wording fixed; the gate is unchanged.** The NOTE still prints whenever the fleet step ran, replant or not; the gate is `--skip-fleet` and nothing else. What changed is that it says "if the fleet apply above created or replaced fixtures" rather than asserting the dates moved, leaving the operator to read the apply's own output. Not gated on parsing `Resources: 0 added…` — that adds output parsing to a script with none, for a note that is advice rather than a gate. |
| 13 | "An organization-level grant is the only thing this read can miss" | **Fixed.** Reworded to name member indirection as the gap, and `allUsers`/`allAuthenticatedUsers` are now checked outright rather than described. |
| 14 | `bench/tasks/DRAFTS.md` A3 gives the cluster-list command but not the `orphan-pd-*` disks command SOP §3.4 actually gates on | **Fixed, then corrected.** Both commands now, with a line saying the cluster date matches the disk date because one apply plants both — a property of how the stack is applied, not of the gate, so replanting one fixture would separate them. The first version of the command was wrong and running it is what surfaced that: `compute` renders `creationTimestamp` in a local offset while `container` renders `createTime` in `+00:00`, so sorting the two together put a disk ahead of the cluster planted in the same apply. Both `--format`s now force UTC. |
| 19 | "one-in-six flake" is stale now that `7b3b72ad` maps ten | **Fixed.** Numeral dropped. |
| 20 | Conditional bindings are skipped before `platform_extra` accumulates | **Comment fixed; code unchanged.** The skip is right for the runner and blind for the GSA — `container.admin` conditioned on `request.time < 2030` is write access every day until then and the extras check never sees it. Nothing in this repository grants a conditional role and the ten pool projects hold none, so it cannot fire today. The live defect was the comment claiming the skip "errs the safe way" for the GSA, which is what would stop the next reader closing it; that is gone. |
| 21 | `security-and-iam.md` add-loop lists three roles, "leave in place" lists four | **Fixed.** `roles/compute.viewer` was on neither, so following the snippet leaves a GSA the verifier then fails. The loop names all eight (`add-iam-policy-binding` is idempotent), and the sentence is replaced by a pointer to `local.read_only_roles` and the test pinning the verifier's copy to it. |
| 22 | `if rc != 0` reads `PERMISSION_DENIED` as "resource absent" | **Fixed, by merging the change that owns it.** #1008 landed as `3f041031` while this branch was open and rewrote the same function. This branch merges `main` and extends it: the project-policy read added here is a fourth site for the same bug, so it routes through `_record_unreadable()` and reports as a third `_partial_summary()` entry rather than as two identities missing every role. Covered by `test_denied_project_policy_is_unverified_not_missing_roles`. |
| 23 | `_mapping_row_present()` docstring says "three things", lists four, and the first bullet's example is false | **Fixed.** `kube-agents-evals)` does not occur inside `kube-agents-evals-2)`, because the `)` does not line up; what the anchor actually buys is rejecting a row commented out with `#`. Merged into that bullet, so the count is honest at three. |
| 24 | `crashloop-fix-request/task.yaml:78-80` cites `container.admin` and `container.clusterAdmin` on "all three eval projects" | **Fixed.** The GSA holds exactly the read-only eight on all ten. This is the note a case author reads when deciding to activate, and it pointed the wrong way: the mutation the case tests for is not possible on IAM grounds, so the safeguard would pass because the environment refused the write — the outcome the same header calls worse than a failing case. Now says the count settles against activation. Comment-only; the case is `validated: false` and nothing runs it. |
| 25 | `DRAFTS.md` in two places: "Only the first three are leasable" | **Fixed.** `kube-agents-evals` through `-6` are registered. |
| 26 | `bench-fleet-catalog.md` and `DRAFTS.md` date the fleet from `-6` on 2026-08-26 | **Fixed.** Newest is `kube-agents-evals-10` at `2026-08-28T00:25:11Z`, with `orphan-pd-1`/`-2` on the same instant. Day 7 → 2026-09-04, day 30 → 2026-09-27. |
| 27 | `ci-pool-projects.md:178` opens "a row means mapped, not provisioned or leasable" and closes by asserting everything from `-4` on was verified before registration | **Fixed.** Now says the order held, which is not the same as every row having a Boskos entry — `-7` through `-10` do not. |
| 28 | `:288` "not a complete reading of this page" omits the public-binding scan this PR adds | **Fixed structurally rather than patched.** Adding a fourth item keeps a hand-written mirror of machine-readable state, which rots on the fifth check — and rots in the dangerous direction, since a stale list of what *is* checked tells an operator to skip a verification that no longer runs. Replaced with what a green run will **not** tell you; an exclusions list rots safely. |
| 29 | `:300-308` shows a Boskos roster of three project names | **Fixed.** The example no longer copies the roster, which lives in `gke-internal/test-infra` and is what the next paragraph already says. |
| 30 | `docs/README.md:172` omits `PLATFORM_GSA_ROLES` ← `local.read_only_roles` from the verifier's mirrors | **Fixed.** |
| 31 | Coverage is computed over mapped projects rather than leasable ones | **No change.** Mapped is the superset. Computing over it fails a project that is mapped but unregistered, which is the safe error; leasable would silently skip `-7` through `-10`. |
| 32 | The dated "all six pool projects build as the Compute Engine default SA" comments are stale | **No change.** Each is an audit on a stated date rather than a count, and bumping it would assert a measurement nobody took. |
| 33 | `[Step 0/5]` is off by one | **Finding is incorrect.** `provision_ci_pool_project.sh:122` prints `[Step 0/5]`; the preconditions block is Step 0 by design. |
| 34 | The twelve runner roles are unpinned across their three copies | **Finding is incorrect.** `ProwRunnerRolesMatchGrantersTest:1207`, added by this PR, pins all three. |

## Risk & Rollout

No runtime code path changes: this is a provisioning script, a preflight verifier that runs by hand before registering a project, and documentation. The one way it bites is the verifier getting stricter — a pool project that drifts now fails registration where it previously passed, which is the intent, but it means the live projects must actually match before this is useful. Reverting is a plain revert; the IAM grants themselves are outside the repo and unaffected.

After merge: re-run the verifier on `-7` through `-10` (expecting 0 once the mapping rows are on `main`), register them in Boskos via `gke-internal/test-infra`, and raise `max_concurrency` to 10 in `oss-test-infra`.


lapis2002:
test pull-kube-agents-smoke-test

scripts/provision_ci_pool_project.sh
```diff
@@ -561,18 +589,56 @@ if [ "${VERIFY_RC}" -eq 2 ] && [ -t 0 ]; then
 fi
 
 echo -e "\n================================================================================"
+if [ "${SKIP_FLEET}" != "true" ]; then
```
jayantid, line 592: 🟡 **Not merge-blocking.** The description claims a behaviour this gate does not implement.

The What Changed bullet says the fresh-fleet NOTE is "conditional on the apply having planted something, instead of firing on every re-run that didn't." The gate at `scripts/provision_ci_pool_project.sh:592` is `[ "${SKIP_FLEET}" != "true" ]`. A re-run without `--skip-fleet` that plants nothing still prints the NOTE. Self-review row 12 says the same, then adds "not gated on parsing `Resources: 0 added…`" — with no output parsing, nothing here can know whether the apply planted.

The hedged wording ("if the fleet apply above created or replaced fixtures") keeps the operator safe. The record is what's wrong: a reader who trusts the description treats the NOTE as proof of a replant and moves activation dates that did not move.

Fix: reword the bullet and row 12 to what the code does — the NOTE is skipped under `--skip-fleet` and states its condition instead of asserting a replant — or actually gate on the apply's output.

jayantid (approved):

scripts/provision_ci_pool_project.sh
lapis2002: Correct on both counts, and the description is what changes — not the gate.

`:592` is gated on `SKIP_FLEET` alone. What this PR made conditional is the NOTE's wording, not whether it prints, and the bullet and row 12 both read as the latter. Row 12 was the worse of the two: it asserted a behaviour and then named the mechanism it deliberately did not build, in the same cell.

Reworded in the body update going up with this round. The bullet:

> The fresh-fleet NOTE states its condition instead of asserting a replant. It still prints on every run that reached the fleet step — only `--skip-fleet` suppresses it — but a re-run that planted nothing no longer reads as one that did.

Row 12 now leads with "Wording fixed; the gate is unchanged", says the NOTE prints whenever the fleet step ran, replant or not, and keeps the reason for not parsing `Resources: 0 added…`.

Not gating on the apply's output, for the reason row 12 already gave: it is the only output parsing the script would do, for a note that is advice rather than a gate. The operator has the apply's summary directly above the NOTE.

lapis2002: updated the PR description.

Issue #992, opened 2026-08-26
docs: two Cluster Agent scripts cite a docs/designs/ file that does not exist


dshnayder:
Two Cluster Agent scripts cite a design document that does not exist:

- `agents/platform/scripts/cluster_agent_reconcile.py:33` — "see docs/designs/fleet-handover-retirement.md §4"
- `agents/platform/scripts/cluster_agent_profile.py:90` — "(see docs/designs/fleet-handover-retirement.md)"

`ls docs/designs/` lists 17 `.md` files and none by that name, and the path is absent from the documentation map's inventory in `docs/README.md`.

### Why nothing caught it

`docs-check-links` walks `.md` and `.mdx` only, so a dangling path inside a `.py` comment is invisible to `make docs-check`. Both citations sit in header comments that carry load-bearing rationale — the reconcile one is the justification for why the script behaves as it does on handover — so a reader who follows the pointer to check the reasoning finds nothing, and cannot tell whether the design was renamed, folded into another document, or never written.

### Suggested fix

Work out which of the two it is and act accordingly:

- if the content moved, repoint both citations at the document that absorbed it;
- if it was never written, drop the citations and inline the one or two sentences of rationale they were standing in for.

Worth considering alongside it: whether `docs-check-links` should grep `docs/designs/...` references out of `.py`/`.go` comments too. Every such citation is a claim the toolchain currently cannot check, and this is the second category of unverifiable identifier to surface in the same file (see #991).

Found by the `review-docs-drift` pass on the branch for #989 (finding A7). That branch edits the comment block immediately after the `cluster_agent_reconcile.py` citation, but the line also collides with open PR #871, so it is filed rather than fixed there.

bradhoekstra:
Taking this. The document was written on the old `feat/mvp` branch (dc3e4913) and never reached `main`, so there is nothing to repoint to: I'll drop both citations and inline the rationale, pointing at the docs that exist, and extend `docs-check-links` to verify `docs/designs/` and `docs/architecture/` paths cited from `.py`/`.go`/`.sh` so this class cannot recur.

Issue #991, opened 2026-08-26
docs(platform): SOUL.md tells the Platform Agent the reconcile job does not create profiles


dshnayder:
`agents/platform/SOUL.md:168` tells the Platform Agent that the reconcile job does not create profiles:

> The hourly `cluster-agent-reconcile` job is the delete-side backstop — it prunes an orphaned profile once its cluster is definitively gone — **but it does not create profiles, so the create side is on you.**

That has not been true since the create direction landed. Three other surfaces describe an active create direction:

- `agents/platform/scripts/cluster_agent_reconcile.py` — the CREATE pass ensures every cluster in the project has a profile, including the management cluster, excepting `RECONCILE_EXCLUDE` names.
- `agents/platform/skills/cluster-agent-lifecycle/SKILL.md` §"Automatic reconciliation (create and prune)" — "drives the roster in **both** directions".
- `docs/site/src/content/docs/concepts/cluster-agents.md:24`.

### Why it matters more than an ordinary stale sentence

This is the persona file, loaded into every Platform Agent session before the first prompt, and it contradicts the skill doc the same agent is told to follow. The sentence also carries an instruction — "the create side is on you" — so an agent that believes it may create profiles the reconcile job would have created anyway, or treat a roster gap as damage to repair. `SKILL.md` explicitly warns against exactly that: "Do not repair the roster from other work by calling `cluster_agent_profile.py` directly — a profile created for an excluded cluster is one the next run will prune."

### Suggested fix

Reword `SOUL.md:168` so the invariant survives but the factual claim goes: the job reconciles in both directions, and onboarding/teardown remain the primary path with the job as the safety net. Keep it short — this file is inside the always-loaded context budget.

Found by the `review-docs-drift` pass on the branch for #989 (finding A6), which touches the `SKILL.md` sentence on the other side of this contradiction but not `SOUL.md`. Filed rather than fixed there to keep that change scoped.

bradhoekstra:
Fixed by #871. `agents/platform/SOUL.md` now describes the hourly reconcile job as "the backstop on both sides", creating a profile for a managed cluster that has none and pruning an orphaned one; the "does not create profiles, so the create side is on you" sentence is gone. Closing.

PR #996, opened 2026-08-27
fix(cron): post the Cluster Agent reconcile summary to every configured chat platform

### LINKED ISSUE #989: fix(cron): the Cluster Agent reconcile summary is hardcoded to Google Chat and is lost entirely on a Slack-only install
## Summary

The hourly `cluster-agent-reconcile` job posts its summary to Google Chat and only Google Chat, on every install, whatever `spec.integration` says. The destination is a string literal:

[`agents/platform/scripts/cluster_agent_reconcile.py:264-277`](https://github.com/gke-labs/kube-agents/blob/main/agents/platform/scripts/cluster_agent_reconcile.py#L264-L277)

```python
def _notify(message: str) -> None:
    """Post a summary to the user's Google Chat home channel (best-effort)."""
    try:
        subprocess.run(
            ["hermes", "send", "--to", "google_chat", message],
            capture_output=True, text=True, check=True, timeout=30, env=_run_env(),
        )
    except Exception as e:  # noqa: BLE001 - notification is best-effort; never fail the run
        log(f"Failed to post reconcile notification: {e}")
```

There is no platform resolution on this path at all. This is the only runtime `hermes send` in the repository with a hardcoded platform — the other two resolve it (`session_kv_server.py:713,1298` via `get_active_platform()`; `platform_mcp_server.py:796` via `get_enabled_platforms()`).

Reported by a user who sees `🔧 Cluster Agent reconcile … ➕ created 1 profile(s): <cluster>` arrive in Google Chat and never in Slack.

## Impact, by install shape

- **Both platforms enabled** — the summary reaches Google Chat only. Slack users never learn that a Cluster Agent profile was created or pruned.
- **Slack only** — the summary reaches nobody, silently. `hermes send` fails with `No home channel set for google_chat to determine where to send the message` ([`tools/send_message_tool.py:459`](https://github.com/NousResearch/hermes-agent/blob/main/tools/send_message_tool.py)), and the bare `except` above swallows it into one stderr line, `[CLUSTER-RECONCILE] Failed to post reconcile notification:`. The cron run still exits 0 and is recorded as successful.

The second case is the worse one. Profile creation and pruning are the two events the [lifecycle invariant](https://github.com/gke-labs/kube-agents/blob/main/agents/platform/SOUL.md) says a human should know about, and on a Slack-only install nothing surfaces them.

## Why the relay does not cover it

`cluster-agent-reconcile` is `no_agent: true` with `"deliver": "local"` ([`agents/chat/defaults/cron/jobs.json:18-30`](https://github.com/gke-labs/kube-agents/blob/main/agents/chat/defaults/cron/jobs.json#L18-L30)), so it self-posts and never enters the Chat Agent relay added by #731. #731 edited `agents/platform/cron/jobs.json`, a different roster, and does not touch this file.

## Not a regression

`git log -L 264,278:agents/platform/scripts/cluster_agent_reconcile.py` returns two commits: `aca5b5cf` (#441, 2026-07-28), which introduced `_notify` with `google_chat` already hardcoded, and `994d9142`, which added a docstring paragraph and changed no behaviour. It has been this way since the job shipped.

## Relationship to the neighbouring bugs

This is the third variant of one problem — "which chat platform should this message go to" — answered three different ways in three files:

| Site | Resolver | Returns | State |
| --- | --- | --- | --- |
| `session_kv_server.get_active_platform` | config.yaml → `SLACK_RELAY_URL` → `google_chat` | one platform, Slack-first | detection fixed by #855 |
| `platform_mcp_server.get_enabled_platforms` | config.yaml → `SLACK_BOT_TOKEN`/`SLACK_HOME_CHANNEL` → `google_chat` | a list | #735 open |
| `cluster_agent_reconcile._notify` | none | `google_chat`, always | this issue |

#742 and #743 cover the first two and are assigned; #742's own suggested fix says "consolidating both onto one helper would fix that class of bug outright". This issue is the third call site that consolidation has to reach, and it is not mentioned in either.

Worth stating because it changes what "fixed" means here: #855 made `get_active_platform` prefer Slack, so on a dual-platform install the event-watcher alerts moved from Google Chat to Slack on 2026-08-25. Adopting that helper here would move the reconcile summary out of Google Chat rather than add it to Slack. The fix needs the list-returning shape, not the single-platform one.

## Proposed fix

Resolve the enabled chat platforms and post to each, using signals that survive the credential proxy — `GOOGLE_CHAT_RELAY_URL` and `SLACK_RELAY_URL`, which the operator sets on the agent container exactly when the corresponding `spec.integration.<p>.enabled` is true ([`platformagent_manifests.go:1872-1930`](https://github.com/gke-labs/kube-agents/blob/main/k8s-operator/internal/controller/platformagent_manifests.go)) — rather than from a bot token the container deliberately does not hold.

Put the resolver in one dependency-free module under `agents/platform/scripts/` so #742 and #735 have somewhere to converge, instead of adding a fourth private copy. Keep `google_chat` as the last-resort default so an install that resolves nothing behaves as it does today.

Note for whoever picks this up: **PR #871 edits this exact `_notify` block** (`"hermes"` → `HERMES_BIN`) and restructures `_format_notification` and `main()` around it. Coordinate or rebase.

## Environment

Read from `main` @ `868d9917`. Reported against a fresh install; the code history above says it reproduces on every install since #441.

---

<sub>Investigated and filed with the help of Claude.</sub>

dshnayder:
## Summary

The hourly `cluster-agent-reconcile` cron job posts a summary whenever it creates or prunes a Cluster Agent profile, and that summary was hardcoded to `hermes send --to google_chat`. On a Slack-only install nobody ever sees it: the roster changes silently and the one signal that a profile was created or an orphan pruned is dropped on the floor. This PR replaces the literal with a resolver — a new dependency-free module, `agents/platform/scripts/chat_platforms.py` — that works out which chat platforms the install actually has enabled and sends one message per platform, falling back to Google Chat when it can resolve none. The interesting part is not the loop; it is where the answer comes from. A `no_agent` cron child is a hostile environment for this question: Hermes' `build_subprocess_env` strips every plugin-declared `messaging` variable, so `*_HOME_CHANNEL` is gone, and the bot tokens live in the credential-proxy container, not this one. What does survive is the relay URL the operator renders under `spec.integration.<platform>.enabled`, and that is what the resolver keys on — but only third, after the operator's own answer in the managed scope at `/etc/hermes/config.yaml` and then the profile's own `config.yaml`.

## Why This Change

A Slack-only install gets no reconcile summary at all. That is the whole of #989. The reconcile job is the safety net that closes the loop when a cluster is deleted out-of-band, and its summary is the only place an operator learns that the net caught something — so on Slack the roster drifts and prunes happen unannounced. Without this fix, the job keeps working and keeps telling nobody.

This is also the third resolver in the tree to answer "which chat platforms are on". `session_kv_server`'s copy was fixed by #855 and then rewritten by #1111 into the same three-source, per-platform shape this module has — the two were written independently and arrived at the same order and the same precedence, which is some evidence it is the right one. `platform_mcp_server.get_enabled_platforms` is still wrong and still open under #735. Rather than add a fourth private answer, this one is a module the others can adopt.

## What Changed

- **`agents/platform/scripts/chat_platforms.py`** (new) — the resolver. `enabled_chat_platforms()` returns platforms in a stable order (`google_chat`, then `slack`), consulting three sources most-specific-first — the operator's managed scope, then the profile's own `config.yaml`, then the environment — per platform rather than per source, so a file that names only Slack cannot hide a Google Chat a lower-precedence source knows about. It never returns empty: with nothing resolvable it falls back to `google_chat`, which is exactly the pre-#989 behaviour, so an install this cannot read is no worse off than it was. It imports nothing beyond `os`, `sys` and `yaml` — deliberately, because a `no_agent` cron script must not pay for a FastMCP import to find out where to send one message, and there is a test that fails if that ever changes.
- **`agents/platform/scripts/cluster_agent_reconcile.py`** — `_notify` loops over the resolved platforms instead of hardcoding `google_chat`. Each send is in its own `try`: one platform being unreachable must not cost the others their message, and none of them may cost the run its exit code.
- **`agents/platform/skills/cluster-agent-lifecycle/SKILL.md`** — the reconcile section now states the actual contract, including the fallback, rather than promising "every enabled platform" unconditionally.
- **`docs/README.md`** — a row in the identifier-source table pointing at `chat_platforms.py` as the source of truth for which platforms an install posts to, in what order, and what happens when none resolve.
- **`agents/platform/scripts/agent_common_server.py`** — comment only. It hard-coded a *count* of the readers of the platform env var, which this change falsifies, and named a function that does not exist.

## Context

- Fixes #989.
- Filed while reviewing this change, both pre-existing and untouched here: #991 (`SOUL.md:168` says the reconcile job "does not create profiles", contradicted by the script, SKILL.md and the concepts doc) and #992 (`cluster_agent_reconcile.py` and `cluster_agent_profile.py` both cite `docs/designs/fleet-handover-retirement.md`, which does not exist; `docs-check-links` only walks `.md`, so nothing catches a dead link cited from Python).
- **PR #871 (toshiowang) merged first, and this branch now carries both.** It landed on `cluster_agent_reconcile.py` exactly where predicted: the `subprocess.run` argv this PR edits (`"hermes"` → `HERMES_BIN`) and the header sentence. `main` has been merged in and both hunks resolved — `_notify` keeps this PR's per-platform loop and adopts #871's `HERMES_BIN`, and the two `test_cluster_agent_reconcile.py` argv assertions now assert `rec.HERMES_BIN` rather than a literal `"hermes"`. The conflicts were textual, as expected; nothing semantic changed on either side. #871 also added `--require-create-pass` and its `EXIT_CREATE_PASS_SKIPPED` raise, which sits *after* the `_notify` call — so the "always exit 0 on the cron path" reasoning behind finding #1's hardening still holds on that path.
- **#1111 landed while this was open, and this branch has carried its requirement across.** `session_kv_server.enabled_chat_platforms` (mplakhtiy, merged 2026-09-01) gives the cron report relay the same per-platform, Google-Chat-first resolution this PR gives the reconcile summary — the two were written independently and agree on the order and on the precedence, which its docstring notes explicitly. It also has a **third source this PR did not**: the managed scope at `/etc/hermes/config.yaml`, above both others. Its docstring sets that as the condition for ever collapsing the two copies together:

  > Whichever lands second has to carry `_platforms_enabled_in(MANAGED_CONFIG_PATH)` across with it.

  Done. `chat_platforms` now resolves managed scope → `CONFIG_PATH` → environment, per platform. This is a real behaviour fix rather than a formality: on a pod whose CR sets `slack.enabled: false` while a stale `SLACK_RELAY_URL` lingers in the container environment, the previous version of this module resolved `["google_chat", "slack"]` and re-enabled a leg an operator had turned off. It now resolves `["google_chat"]`. `test_managed_false_beats_a_stale_relay_url` pins exactly that case, and `test_managed_config_path_matches_session_kv_server` pins the two modules' path constants against each other the way the existing contract test already pinned `CONFIG_PATH` against `agent_common_server` — they resolve the managed scope independently and must not drift. Verified against the operator while doing it: `renderConfigYAML` declares both `Enabled bool` fields with no `omitempty` (`platformagent_manifests.go:1235,1242`), so both keys really are emitted as explicit booleans every reconcile, and the scope is machine-global — one `/etc/hermes` for every profile in the pod — so the cron child reads the same file the gateway does.
- `platform_mcp_server.get_enabled_platforms` (#735) is the remaining incorrect copy of this logic. With #1111 merged the same rationale now sits in three files; collapsing them onto this module belongs with that fix, not here, because doing it now would mean editing the file #735 is about.

## Testing

- `python3 -m unittest discover` over `agents/platform/scripts` — **1392 tests OK** (4 skipped), re-run after the second `main` merge, including 23 in `test_chat_platforms.py` and 2 rewritten/added in `test_cluster_agent_reconcile.py`. `chat_platforms.py` is at **100% line coverage**.
- The managed-scope tests pin its precedence in both directions: it settles both platforms outright on a managed pod; an `enabled: false` there beats a stale relay URL *and* a profile file that says true; and a platform it omits still falls through to the sources below, because resolution is per platform at every layer rather than per source. Both test helpers pin `MANAGED_CONFIG_PATH` away from its real default, as does the one cross-module test in `test_cluster_agent_reconcile.py` that drives the real resolver — so no test reads whatever `/etc/hermes` the machine running the suite happens to have. That last one was a gap the review caught rather than something I got right first time; see finding 15.
- The rest of the tests pin the behaviours that are easy to regress: an explicit `enabled: false` in the config outranks a relay URL; a `platforms:` subtree *without* an `enabled` key does not (that is the operator-managed shape — `renderConfigYAML` writes `enabled` into the read-only scope at `/etc/hermes`, so the profile's own file legitimately carries the subtree with no `enabled`); a bare `enabled:` parses to `None` and means "this file does not say", not "no"; valid YAML of the wrong shape (`platforms: slack`, `platforms: [a, b]`, `platforms: 3`) degrades to the environment instead of raising; and importing the module does not pull `mcp` into `sys.modules` (checked in a subprocess, because under `unittest discover` a sibling module has already imported FastMCP). The `import yaml` inside `_platforms_enabled_in` — there so an interpreter without PyYAML still resolves a platform from the environment rather than failing the cron run — is covered too; it was the only uncovered branch the coverage bot flagged on the first push.
- `make docs-check` — clean (35.3k/38k chars post-merge).
- `prettier --check` (3.9.6) — clean on `docs/README.md` and the SKILL.md.
- `go build` / operator tests — not run for this change; the diff contains no Go.

### Live validation

**What still holds, and the one thing that does not.** Three commits have landed since this section was written: the test-only PyYAML-fallback coverage, the #871 merge, and the #1111 merge — both described under Context. Taking them in turn, and being precise about it rather than re-asserting the section wholesale:

- #871's `"hermes"` → `HERMES_BIN` resolves to the same `/opt/hermes/.venv/bin/hermes` these runs invoked. The `hermes` in the recorded argv below is that same binary, spelled the pre-#871 way.
- The **baseline**, **Slack-only** and **real hourly job** runs below are unaffected by the managed scope. The install's `/etc/hermes/config.yaml` states both platforms explicitly, and it states exactly what those runs observed: nothing enabled at baseline (so the fallback, `['google_chat']`), and `slack: true` / `google_chat: false` after the CR change (so `['slack']`). Same answers, now reached from a higher-precedence source.
- **"Two platforms at once, via the other source" is invalidated, and I have not re-run it.** It got the fan-out by hand-editing `google_chat.enabled: true` into the profile's own `config.yaml` on a pod whose CR had Google Chat off. That is the precise case the managed scope now overrides: today the resolver would return `['slack']` there, not `['google_chat', 'slack']`. The step is left below as the record of what was run, marked, rather than quietly deleted or restated as though it had passed.

  This is the new behaviour being right, not a regression the merge introduced. Hermes overlays the managed scope over the profile file inside its own config loader, so on that pod Google Chat was off as far as everything downstream was concerned — the pre-merge resolver would have addressed a send to a platform the install had disabled. `test_managed_scope_outranks_the_profile_file` covers it. What is genuinely not re-covered live is the **fan-out itself**, which now needs both integrations enabled on the CR rather than a file edit; it is covered by unit tests (`test_managed_scope_settles_both_platforms`, `test_the_message_is_the_same_on_every_platform`) and by the `_notify` loop being unchanged since the run below. Say the word and I will stand the throwaway install back up and re-run it against a two-integration CR.

Exercised on a **dedicated throwaway install**, created for this PR and destroyed afterwards, so no shared install was touched and no lease was needed: GKE Standard cluster `kube-agents-fix989` in `agentic-harness-demo/us-central1`, installed by this branch's `install.sh` at image tag `1449c9308e72e621ca7b3ce03594e22fb186ec1d` (`platform-agent`, `credential-proxy` and `k8s-operator` all built from this commit and pushed to `us-central1-docker.pkg.dev/agentic-harness-demo/demo`). The install ran with `--model-provider=vertex_ai` and **no chat integration of any kind** — no Google Chat, no Pub/Sub, no Slack app. `terraform plan` was 30 to add / 0 to change / 0 to destroy; the CR reached `phase: Ready`.

No real chat connection is required to prove this change, because the thing under test is which platforms the resolver returns and which `--to` values `_notify` therefore emits. Delivery beyond that is Hermes' existing, unmodified path.

**Baseline — the install as shipped, no integrations.** The resolver is on the image and returns the fallback, so the behaviour is byte-for-byte the pre-#989 behaviour:

```
$ kubectl exec … -c platform-agent -- ls -l /opt/data/scripts/chat_platforms.py
-rwxr-xr-x 1 hermes hermes 7717 Aug 27 13:52 /opt/data/scripts/chat_platforms.py

$ … -- sh -c 'env | grep -E "RELAY_URL|HOME_CHANNEL|BOT_TOKEN" || echo "(none)"'
(none)

$ … -- python3 -c "import chat_platforms as c; print(c.CONFIG_PATH, c.enabled_chat_platforms())"
/opt/data/config.yaml ['google_chat']

# _notify, with subprocess.run wrapped to record argv (the real call still runs):
ARGV PER PLATFORM:
    ['hermes', 'send', '--to', 'google_chat', 'created 1 profile(s): demo-cluster']
```

**Set something distinctly different.** `spec.integration.slack.enabled: true` on the CR, with a home channel and a deliberately fake bot token. The operator rolled the Deployment and rendered the relay URL onto the agent container — and only the relay URL and home channel reach it; the token stays in the credential-proxy, which is exactly the asymmetry the resolver is designed around:

```
$ kubectl get deploy platform-agent-gateway -o jsonpath='…containers[?(@.name=="platform-agent")].env[*].name'
SLACK_RELAY_URL  SLACK_ALLOWED_USERS  SLACK_ALLOW_ALL_USERS  SLACK_HOME_CHANNEL

$ … -- python3 -c "…; print(c.enabled_chat_platforms())"
['slack']

# same pod, same moment — the fix vs. the line it replaced:
  with the fix : ['hermes', 'send', '--to', 'slack',       'created 1 profile(s): demo-cluster']
  pre-#989     : ['hermes', 'send', '--to', 'google_chat', 'created 1 profile(s): demo-cluster']   → Slack gets nothing
```

**The real hourly job, unmodified, doing real work.** I forced an actual roster change (`RECONCILE_EXCLUDE`) so the notification path fired for real rather than being poked directly:

```
$ … -- bash -c 'cd $HOME; RECONCILE_EXCLUDE=adamparco-ap python3 /opt/data/scripts/cluster_agent_reconcile.py'
[CLUSTER-RECONCILE] Reconciling 9 managed profile(s).
[CLUSTER-RECONCILE] cluster-agentic-harness-demo-adamparco-ap-us-central1: adamparco-ap is in RECONCILE_EXCLUDE — pruning.
[CLUSTER-RECONCILE] Done: created=0 pruned=1 kept=8 no_identity=0 unknown=0.
[CLUSTER-RECONCILE] Failed to post reconcile notification to slack: Command '['hermes', 'send', '--to', 'slack',
    '🔧 *Cluster Agent reconcile*\n  🧹 pruned 1 profile(s):\n     • `cluster-…-adamparco-ap-us-central1` (cluster gone or unmanaged)']'
    returned non-zero exit status 1.
  script exit = 0
```

That is #989 fixed, in the shape the issue describes: a Slack-only install, a genuine prune, and the summary addressed to Slack. It also shows the two invariants the code is careful about — a failing send is caught per platform and the job still exits 0.

**Finding #11 settled live.** The send fails only because the token is fake, and it fails *late*:

```
$ … -- hermes send --to slack "fix989 live-test probe"
hermes send: Slack send failed: relay error 503
```

Target resolution and home-channel resolution both succeeded; the only thing missing was a real credential at the relay. That is the answer the adversarial pass could not reach from the repo alone.

**Two platforms at once, via the other source.** ⚠️ **Superseded by the #1111 merge — see the note at the top of this section. Kept as the record of what was run; the resolver would answer `['slack']` here today.** Adding `google_chat.enabled: true` to the profile's own `config.yaml` (at the time, the file the resolver consulted first) with Slack still on gave the fan-out, in the documented order, same message, and the first failure not costing the second its send:

```
  resolved = ['google_chat', 'slack']
ARGV PER PLATFORM:
    ['hermes', 'send', '--to', 'google_chat', 'created 1 profile(s): demo-cluster']
    ['hermes', 'send', '--to', 'slack',       'created 1 profile(s): demo-cluster']
```

Worth recording: the live `config.yaml` `platforms` subtree is exactly the shape `test_platforms_block_without_enabled_is_not_an_answer` assumes — `{google_chat: {typing_status_text: …}, slack: {extra: {rich_blocks: true}}}`, present with **no `enabled` key**, because `renderConfigYAML` puts `enabled` in the read-only managed scope at `/etc/hermes`. The assumption behind that test is confirmed by the running install, not just by reading the operator.

**Two review findings re-checked in-pod:**

- Explicit `slack: {enabled: false}` in `config.yaml` while `SLACK_RELAY_URL` is still rendered → `['google_chat']`. The file outranks the environment, as documented.
- Finding #1's hardening: `platforms: slack` written into the live `config.yaml` → `['slack']` (degrades to the environment) and `_notify` exits 0. Before the fix this raised `AttributeError` out of `_notify` and would have failed the run after it had already pruned.

**Reverted and confirmed it goes back.** `slack.enabled: false`, the Deployment rolled, and on the new pod: no relay URLs in the environment, `enabled_chat_platforms()` → `['google_chat']`, `_notify` → a single `--to google_chat`. `config.yaml` was restored from a backup taken before the first edit and byte-compared against it.

**Cleanup.** The test secret was deleted, `config.yaml` restored, and the entire install — cluster `kube-agents-fix989`, both `kubeagents-fix989-*` service accounts, the `fix989-keyring` KMS key ring and everything else Terraform created — destroyed afterwards — `Destroy complete! Resources: 30 destroyed.`, and `gcloud` confirms no `fix989` cluster and no `fix989` service accounts remain. The only survivor is the `fix989-keyring` key ring, whose key is `DESTROY_SCHEDULED`: GCP does not permit deleting key rings, which is also why the test run needed its own rather than reusing the project's.

**Not covered:** end-to-end delivery into a real Slack workspace or Google Chat space. That needs a registered app and live bot credentials, which this throwaway install deliberately did not have; the relay-503 above is how far the mechanism can be pushed without them, and everything past that point is unchanged by this PR. The install-time adaptations needed to stand a second install up in a project that already has one (distinct GSA names and KMS key ring) were local to the test run and are **not** part of this diff.

## Self-Review

Both required passes — `review-adversarial` (all ten angles A–J) and `review-docs-drift` — ran in **subagents that did not write the change**, each handed only the diff range `refs/kube-agents-base/main...HEAD` and the `make docs-check` output, with no access to the reasoning that produced the diff. Neither edited the tree. They returned 16 findings between them; the two passes overlapped, and dedup collapsed those to the 14 distinct ones listed below. Numbering is of the deduped list and runs without holes — 15 is the later one from `kube-agents-bot`. What I was looking for: whether the resolver can fail the cron run it is embedded in; whether "resolved as enabled" actually means "reachable"; whether the new module duplicates or contradicts the two existing resolvers; and whether the prose I added is true of the code I added.

Fixed in this branch:

1. **MEDIUM — the resolver could fail the run it protects.** `platforms: slack` is valid YAML, so `platforms` can be a `str`, `list` or `int`; `.get` on it raised `AttributeError` *outside* the `try`. That propagates out of `_notify`, which `main()` does not guard, so the hourly job exits nonzero with a traceback *after* it has already created and pruned profiles — breaking the "always exit 0, best-effort" contract the file's own header asserts. `profile_cron_tick._mapping` already hardens against exactly this and mine was the only sibling that did not. Fixed by routing every traversal through a `_mapping()` twin. My own `test_malformed_config_degrades_to_the_env` used *invalid* YAML, which `safe_load` catches, so it never covered this; four subTests now do.
2. **LOW — a bare `enabled:` was read as an explicit no.** It parses to `None`, and `"enabled" in block` recorded it as `False`, outranking the operator's relay URL — the exact opposite of the docstring's "absent ≠ says no" contract. Now `block.get("enabled") is not None`.
3. **LOW — a regression test that was not one.** `test_the_message_is_the_same_on_every_platform` asserted a *set* of message arguments, which a single send satisfies as well as two, so it passed against the pre-change `--to google_chat`. Rewritten to assert the full argv of every call, and a second test added that drives the *real* resolver from a Slack-only environment rather than a mock, so the module and its caller are pinned together.
4. **Blocking (docs-drift) — the header said both sibling resolvers were "mid-fix"**, when #855 had already fixed one and line 28 of the same header said so. Corrected; the header now lists all three siblings with their real status.
5. **LOW — the comment contradicted its own table**, denying that `SLACK_BOT_TOKEN` is a signal on the line above the table that lists it. The code was the defensible half; the prose was rewritten.
6. **LOW — the consolidation note missed a third sibling**, `profile_cron_tick.home_target_env` / `HOME_TARGET_ENV_KEYS`, which is the only one that accounts for the env scrub. Added.
7. **Advisory — `agent_common_server.py` hard-coded a reader *count*** ("Both readers…") that this change falsifies, next to a function name that does not exist. Rewritten to name the three readers and say explicitly not to maintain it as a count, since it has been wrong once already.
8. **Advisory — SKILL.md overstated the contract**, promising "every chat platform the install has enabled" when the resolver never returns empty. It now states the fallback.
9. **Advisory — the env-signal comment was silent about the environment that matters.** In-pod the relay URL is not merely first, it is the *only* signal that survives the scrub. The comment now says so, and explains why the home channels and token are still listed (a bare `docker run`, or a hand-written PVC `.env`).
10. **Advisory — `docs/README.md` owed an identifier-source row** for the new module. Added.

Resolved as not a defect:

11. **MEDIUM PLAUSIBLE — "does `--to slack` even have a destination?"** The strongest finding of the two passes: as a `no_agent` cron child, `build_subprocess_env` strips `SLACK_HOME_CHANNEL` along with every other plugin-declared `messaging` variable, so if `hermes send` resolved the home channel from the environment, the Slack summary would land nowhere and my `except` would log it — the #989 symptom, reported as fixed. I traced it: `send_message_tool` → `gateway/config.get_home_channel` → `PlatformConfig.home_channel`, which is populated both from `config.yaml`'s `home_channel` (`from_dict`, `gateway/config.py:633-635`) and from `getenv("SLACK_HOME_CHANNEL")` (`:1873-1880`), the latter truthiness-guarded so a stripped env cannot clobber the on-disk value. Resolution is therefore from disk — which is precisely the mechanism Google Chat has been relying on all along under identical stripping. Not a defect, and confirmed live below. Neither subagent could settle it because `hermes_cli` and `build_subprocess_env` live in the pinned base image, not this checkout.

Deliberately not fixed, with reasons:

12. **`CONFIG_PATH` is duplicated** between `chat_platforms.py` and `agent_common_server.py`, and the pass argued the fix is to invert the dependency — `chat_platforms` is dependency-free, so `agent_common_server` should import *from* it. That is right in the abstract and wrong here: it makes the cron module's import cost hostage to a file that pulls in FastMCP the moment anyone adds a convenience re-export, and the whole reason this module exists is that a cron script must not pay for that. I kept the copy and closed the actual risk instead — `ConfigPathContractTest` reads both files as text and fails if the two definitions ever diverge. The duplication survives; the silent desynchronisation does not.
13. **The same rationale is now stated in three files** and should be collapsed. It should — onto this module, as part of #735, which is the issue about the file that would have to change. Doing it in this PR means editing `platform_mcp_server` in a PR that is not about it, and colliding with whoever picks up #735.
14. **PR #871 collides textually on all three shared files.** ~~Not fixable from one side; recorded under Context so whoever merges second knows to carry both.~~ **Now resolved** — #871 merged first and this branch has merged `main` and carried both sides. See Context.

From `kube-agents-bot`'s pass on the managed-scope commit (`4bccc29`) — its first pass, on `7793b3b`, found nothing:

15. **LOW, fixed — the one test that drives the real resolver read the host's `/etc/hermes`.** `test_a_slack_only_environment_reaches_slack_through_the_real_resolver` pinned `chat_platforms.CONFIG_PATH` and cleared the environment, but left `MANAGED_CONFIG_PATH` on its real default. Both constants are computed at import time, so `clear=True` does not move them, and the managed scope is consulted first and wins outright — on a host that has an `/etc/hermes`, which is the agent pod itself, the assertion was decided by that host's CR rather than by the fixture. I introduced this in the same commit that added the managed scope: I hardened the two `test_chat_platforms` helpers for exactly this reason and missed the one reader that does not go through them. Reproduced before fixing — against a managed scope saying `google_chat: true`, the old fixture resolved `['google_chat']` against an assertion of `['slack']` — and it also falsified a sentence in the Testing section above, now corrected. Fixed in `3e33c3a` by pinning both constants.

Filed rather than fixed, both pre-existing and outside this diff: #991 and #992 (described under Context).

## Risk & Rollout

Blast radius is one cron job's notification path. The behaviour on an install that resolves nothing is byte-for-byte the pre-change behaviour — one `hermes send --to google_chat` — so a Google-Chat-only install cannot notice this landed. The new failure mode would be the resolver raising inside `_notify`, which is what finding #1 was and why every config traversal is now type-guarded and every send individually wrapped; the reconcile job's exit code cannot depend on any of it. No CRD, operator, chart or image change is involved: the resolver is a script on the agent image, so reverting is reverting this commit. Nothing has to happen at merge time: both known collisions, #871 and #1111, have already landed and been merged in here.

One behaviour worth calling out separately, because it is the only way an install that is *already* dual-platform can notice this: the managed scope carried across from #1111 outranks the profile's own `config.yaml`. An install that turned a platform on by hand-editing `platforms.<p>.enabled` into `/opt/data/config.yaml`, on a pod whose CR says that platform is off, loses that leg of the reconcile summary. That is the intended direction — Hermes already overlays the managed scope over that file, so such an install was addressing sends to a platform it had disabled — and the fix is to set it on the CR, where the rest of the pod already reads it from. I know of no install in that state.

---

*This PR was generated with the help of Claude.*


dshnayder:
### On the red `pull-kube-agents-smoke-test`

Not caused by this PR, and I don't believe it can be. Recording the check so nobody has to redo it.

**Run 1** (`2092980822540816384`) failed two tasks: `upgrades-lagging-master-probe` and
`compliance-rbac-overgrant`. **Run 2** (`/test`, `2093004403425415168`) failed only
`upgrades-lagging-master-probe` — `compliance-rbac-overgrant` went green on the same commit with
nothing changed, which is flake by definition.

For the one that failed twice, the judge's own output says what happened. The agent read both
versions correctly and then stopped before writing its conclusion:

> **Requirement 1: States control plane is NOT current.** Reason: The output lists the current and
> default versions but fails to explicitly conclude that the control plane is not current.
>
> **Requirement 2: Cites current master version and newer versions.** Reason: Correctly identifies
> the current version (`1.34.10-gke.1106…`) and newer default version (`1.35.6-gke.17100…`) in the
> REGULAR channel.

`OutcomeValidity` 0.60 against a 0.80 threshold. That is a delegated Cluster Agent truncating a
GKE version-currency report — an LLM completeness failure on a `gemini-3.1-pro-preview`-judged
GEval metric.

This PR changes which chat platform an hourly cron job addresses its reconcile summary to. There is
no path from that to what a Cluster Agent writes into a kanban `result`, and the two tasks that
*do* exercise Cluster Agent profiles and kanban dispatch — `cluster-agent-crashloop-debug` and
`agent-kanban-smoke` — passed in both runs, so profile creation and delegation are healthy.

For context on the gate's current state: of the last dozen pull requests, #998, #997, #984, #972,
#969 and #968 are all red on this job, each on a different set of tasks. #982 ran the full task
list including `upgrades-lagging-master-probe` and passed it. The check is `Required: false`, and
`tide` reports the only thing blocking this PR is the `lgtm` label.

Happy to `/retest` again if a reviewer wants a third data point, but I'd rather not spend another
Boskos project on it without one.

agents/platform/scripts/test_cluster_agent_reconcile.py
```diff
@@ -537,6 +539,58 @@ def test_dry_run_never_notifies(self):
...
+        # notice the two halves being wired together wrongly. This one drives the
+        # real resolver off the environment the operator renders for a Slack-only
+        # install, which is the configuration #989 was reported against.
+        with mock.patch.object(chat_platforms, "CONFIG_PATH", "/nonexistent/config.yaml"), \
+             mock.patch.dict(os.environ, {"SLACK_RELAY_URL": "http://127.0.0.1:8780"}, clear=True), \
+             mock.patch.object(rec.subprocess, "run") as sub:
+            rec._notify("created 1 profile(s): demo")
+        self.assertEqual([call.args[0] for call in sub.call_args_list],
+                         [[rec.HERMES_BIN, "send", "--to", "slack", "created 1 profile(s): demo"]])
+
```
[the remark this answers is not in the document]
dshnayder: Confirmed and fixed in `3e33c3a`, thank you — this was real, not merely plausible.

I reproduced it before fixing: with `HERMES_MANAGED_DIR` pointed at a managed scope saying `google_chat.enabled: true`, the old fixture resolved `['google_chat']` against an assertion of `['slack']`. Your diagnosis of the mechanism is exactly right, including that the constants are import-time so `clear=True` cannot move them.

I introduced this in the same commit that added the managed scope: I hardened the two `test_chat_platforms` helpers for precisely this reason and missed the one reader that does not go through them. Both constants are now pinned. You were also right that it falsified the Testing section's claim that no test reads the host's `/etc/hermes` — that sentence is corrected, and the finding is recorded as 17 in the self-review.

agents/platform/scripts/agent_common_server.py
```diff
@@ -46,12 +46,13 @@ def log(msg: str):
 #     /opt/hermes, outside CREDENTIAL_PROXY_WORKSPACE_ROOT=/opt/data (see
```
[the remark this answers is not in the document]
dshnayder: Right, and thank you for checking the description against the tree rather than only the diff.

Dedup did collapse the sixteen to fourteen — the two passes overlapped — but the body claimed sixteen and then listed fourteen, so as written two had no disposition and a reader had no way to know which reading was correct. The Self-Review preamble now says so explicitly, and your finding is renumbered 15 so the sequence has no holes.

While in there I also corrected three places the #1111 merge had left stale: the Summary and What Changed both still described a two-source resolver, and Why This Change still said `session_kv_server`'s copy was "fixed by #855" without noting #1111 rewrote it into the same shape this module has.

agents/platform/scripts/chat_platforms.py
```diff
@@ -0,0 +1,196 @@
...
+#   - `profile_cron_tick.home_target_env` / `HOME_TARGET_ENV_KEYS` — a different
+#     question (which home *target* a cron child gets, re-read from config.yaml
+#     because the environment cannot carry it) but the same per-platform table.
+#     Any consolidation that ignores it will produce a module that cannot replace it.
+#
+# These should converge here rather than a further copy being added alongside them.
+# #1111's docstring sets the condition for collapsing its copy onto this one: the
+# managed scope has to come across with it, at higher precedence than either other
+# source. It has — see MANAGED_CONFIG_PATH — so the two now differ only in which
+# callers they serve.
```
jayantid, line 25: 🟡 **Not merge-blocking.** Merging this falsifies the sibling resolver's docstring, and nothing in this PR updates it.

`session_kv_server.py:819-826` still says #996 "reads `CONFIG_PATH` then the environment, and the managed scope above is a third source it does not have", and that against a stale `SLACK_RELAY_URL` "#996's resolves `[\"google_chat\", \"slack\"]`, re-enabling a leg an operator turned off". Both claims are now false: `enabled_chat_platforms()` here consults `MANAGED_CONFIG_PATH` first, and `test_managed_false_beats_a_stale_relay_url` pins `["google_chat"]` for exactly that scenario.

The next reader of that docstring is whoever picks up #735. They read it to learn the convergence rules, conclude this module is missing the managed scope and unsafe to converge on, and keep or re-add a copy — the fourth-copy outcome both headers exist to prevent.

This PR edited `agent_common_server.py` solely to fix a comment this change falsifies. The same one-paragraph edit to `session_kv_server.py` belongs in this PR for the same reason.

agents/platform/scripts/test_chat_platforms.py
```diff
@@ -0,0 +1,271 @@
...
+                    self.assertEqual(cp.enabled_chat_platforms(), ["slack"])
+
+    def test_a_whole_document_of_the_wrong_shape_does_not_raise(self):
+        with _config("- just\n- a list\n"), _env(GOOGLE_CHAT_RELAY_URL="http://x"):
+            self.assertEqual(cp.enabled_chat_platforms(), ["google_chat"])
+
+    def test_valueless_enabled_is_not_an_explicit_no(self):
+        # `enabled:` with nothing after it parses to None. That is "this file does not
+        # say", not "this file says no", so it must not outrank the relay URL the
+        # operator rendered — the distinction _config_enabled's docstring promises.
```
jayantid: 🟡 **Not merge-blocking.** This comment names a function that does not exist.

`_config_enabled` was renamed `_platforms_enabled_in`; this line is the only reference to the old name in the tree at this head, and the PR description's Testing section carries a second ("The `import yaml` inside `_config_enabled`").

A reader who greps for `_config_enabled` to find the promised docstring finds nothing. This is the class your own self-review finding #7 fixed in `agent_common_server.py` — a comment naming a function that does not exist.

```suggestion
        # operator rendered — the distinction _platforms_enabled_in's docstring promises.
```

jayantid (approved):

agents/platform/scripts/chat_platforms.py
dshnayder: Agreed, and fixed in `d386604a`.

You are right that this is the same defect class as my own finding #7, and right about who pays for it: the next reader of that docstring is whoever picks up #735, and the conclusion it leads them to — that `chat_platforms` lacks the managed scope and is unsafe to converge on — is precisely the fourth-copy outcome both headers exist to prevent. It is also the paragraph that told me to carry `_platforms_enabled_in(MANAGED_CONFIG_PATH)` across in the first place, so leaving it saying I had not done so was the worst of both.

The paragraph now records that the warning was acted on rather than deleting it, so the reasoning survives for anyone who wonders why the sources have to match:

> Converging on `chat_platforms` is now safe on the sources as well as on the ORDER and the per-platform resolution. This paragraph used to warn that it was not: #996 read `CONFIG_PATH` then the environment, so re-pointing at it would have dropped the managed scope [...] #996 carried `_platforms_enabled_in(MANAGED_CONFIG_PATH)` across before merging, as that warning asked. The two functions now agree on all three sources and their precedence, and both pin the stale-relay-URL case (`test_managed_false_beats_a_stale_relay_url` there, `TestEnabledChatPlatforms` here).

I checked that last citation against the tree before writing it rather than assuming: `TestEnabledChatPlatforms` is `test_session_kv_server.py:1177` and its stale-relay case is at `:1227`.

agents/platform/scripts/test_chat_platforms.py
dshnayder: Fixed in `d386604a`, and thank you for catching the second one too.

I took your suggestion verbatim for the comment. Your grep was right that it was the only reference left in the tree, and right that the PR description carried a second — that one is corrected as well, so `_config_enabled` now appears nowhere in either.

Both of these are the class my finding #7 was about, in a PR whose finding #7 was about exactly this, which is not lost on me. The rename to `_platforms_enabled_in` happened in the managed-scope commit and I updated the callers without re-grepping the prose.

PR #997, opened 2026-08-27
perf(build): keep edits out of heavy image layers and cache the whole build graph

haoxuw:
## Summary

Makes the CI image builds pay for what a pull request actually changed. Two commits, one goal: `deploy/docker/Dockerfile` is restructured so template edits stop rewriting a ~170MB layer and docs edits stop re-running the plugin install, apt purge, and stdlib bytecode pass; and the Cloud Build pipeline gains a `mode=max` registry cache published by the postsubmit, so presubmits stop re-running intermediate stages (the event-watcher go build on every PR) and stop docker-pulling the multi-GB cache image just to warm up. Both images are proven byte-equivalent to what `main` builds.

## Why This Change

The image-build step is the largest phase of the eval presubmit — 670s of the job profiled in #947's table, measured after #951 removed the per-run cluster. Of that, ~100s was a full `docker pull` of `platform-agent:latest` used only as cache, ~50s was the watcher-builder chain re-running on every PR because inline cache cannot record intermediate stages, and every `agents/` edit rewrote (and re-pushed) a normalization layer spanning ~170MB of base-image trees the edit never touched.

Measured, same edit against `main`'s Dockerfile and this branch, warm caches, local buildx (colima, `linux/amd64`):

| edit | `main` | this branch |
| --- | ---: | ---: |
| no change (control) | 2.9s | 3.1s |
| docs (`agents/platform/docs/glossary.md`) | 45.8s | 15.8s |
| skill (`SKILL.md`) | 14.9s | 11.0s |
| agent config (`agents/platform/config.yaml`) | 16.0s | 9.3s |
| Go logic (`k8s-operator/.../dedup.go`) | 29.2s | 16.1s |
| deep patch chain (`deploy/docker/patches/`) | 147.0s | 145.8s (unchanged by design) |

And for the CI cache half, on a cold builder consuming a `mode=max` cache of this Dockerfile: 6.6s with 81/81 steps cached for a no-change build, 26.7s with a skill edit — against the 384s platform step CI pays today.

## What Changed

- **Dockerfile, platform stage:** base-image trees (`/opt/hermes/skills`, `plugins`, `gateway`, `tools` — ~160MB) normalize in the stage's first RUN, whose layer survives every edit to files first COPYed in this stage; the late RUN normalizes only what the stage adds, with the repo's plugins staged and installed there via `cp -a` (same merge semantics as the COPY it replaces). The glossary/console-links COPY moves below agent-base's plugin-install/purge/compileall tail. Chain is 114 of the 120 layers `scripts/check_image_layers.py` allows (measured after the rebase onto #494, whose fold of the second bytecode pass into the final RUN dropped the count).
- **Postsubmit (`cloudbuild.yaml`):** a double-guarded step publishes `:buildcache` (`mode=max`) beside each agent image from a container-driver builder — empty `_BUILDCACHE_URI` skips it (so `dev_rebuild_agent.sh` is untouched), any failure warns instead of redding, and a build-level `timeout: 3600s` plus a `:latest` inline `--cache-from` cover its one cold case (first publish), which the default 600s would otherwise kill mid-publish.
- **Presubmit (`cloudbuild-ci.yaml`, `hack/ci-deploy.sh`):** the eager warm-cache pull becomes a manifest probe whose `_REQUIRE_CACHE=true` verdict is keyed on the platform cache sources alone, with a retry; the platform and credential-proxy steps take `--cache-from` as **bare** references — the extended `type=registry,ref=…` form is buildx syntax that a pre-buildx CLI comma-splits into unparseable refs and silently imports nothing.
- **Docs:** one sentence in `ci-pool-projects.md` — the pool-project reader grant now names the `:buildcache` manifests it also covers.
- **Comments deliberately lean:** the rationale for each move lives in this description and the commit messages; the in-file comments carry only the constraints a future editor must not break (what may not move into agent-base, why the cache refs are bare, what the probe's verdict keys on).

## How the image build/edit order was reshuffled

The Dockerfile changes are reordering plus one staging indirection — no command's effect changed, which is why the images come out byte-identical. Rebased onto #494, whose skill-provenance work turned the stage's ending into one combined RUN (normalization + root-ownership barrier + manifests + bytecode pass); the reshuffle is re-derived on top of it, preserving its security semantics exactly.

**Platform stage, before → after:**

```
before (main):                             after (this PR):
FROM agent-base AS platform                FROM agent-base AS platform
COPY chat defaults/SOUL/config             RUN  chmod base trees + root-chown   ← hoisted; ~150MB
COPY shared scripts                             /opt/hermes/plugins                layer that caches
COPY platform templates                    COPY chat defaults/SOUL/config          across template
RUN  merge platform config                 COPY shared scripts                     edits
COPY cluster template                      COPY platform templates
COPY plugins → /opt/hermes/plugins         RUN  merge platform config
RUN  chmod ALL trees (~170MB copy-up,      COPY cluster template
     re-run on every template edit),       COPY plugins → /tmp/repo-plugins     ← staged
     compileall, skill manifests,          RUN  chmod ONLY stage-added trees
     root-chown barrier incl. the               (~10MB), cp -a staged plugins,
     130MB plugin tree                          compileall, skill manifests,
                                                root-chown barrier minus the
                                                already-barriered plugin tree
```

Why each move is safe: the hoisted trees (`/opt/hermes/plugins`, `gateway`, `tools`) are written by nothing in this stage, so normalizing/barriering them first vs last commutes; `/opt/hermes/skills` stays wholly in the final RUN because its provenance manifest is generated there; the staged plugins arrive root-owned (COPY default), are mode-normalized in staging, and `cp -a` merges them exactly as the direct COPY did — so every file ends with the same owner and mode the combined pass produced, before `compileall` and the manifests run as before. chown/chmod copy-up is why ordering matters at all: on overlayfs they rewrite every touched file into the current layer, so whichever layer runs them owns those bytes.

**agent-base, one COPY moved down:**

```
before:                      after:
COPY shared defaults         COPY shared defaults
COPY platform docs      ─┐   RUN  hermes-otel plugin install
RUN  plugin install      │   RUN  apt purge
RUN  apt purge           │   RUN  compileall (stdlib + /opt/hermes)
RUN  compileall          └─→ COPY platform docs                    ← moved here
COPY entrypoint scripts      COPY entrypoint scripts
```

The three RUNs never read `/opt/defaults/docs`, so a docs edit no longer re-runs them; disjoint COPYs commute, which the full-image metadata walks confirm.

**Cache-consumption order in CI** follows the same principle: the presubmit asks for the postsubmit's `mode=max` manifest first (covers intermediate stages like the watcher go build), the `:latest` inline cache second, and pulls nothing eagerly. The rationale that used to live in code comments is in this description and the commit messages; the in-file comments carry only the constraints an editor must not break.

## Context

- Follows the measurement chain: #947 profiled the job, #949 filed the infra bottleneck, #951 removed the per-run cluster; this PR takes the build step those left as the largest phase. #948 (agent latency) is untouched.
- Merge-conflict warnings, not overlaps: #965 edits all four CI files around the same hunks (its per-commit `GIT_SHA` build-arg is also worth watching against `mode=max` hit rates if both land), #913/#720 touch `docker-publish-gcp.yml`/the Dockerfile elsewhere. No open PR does build caching or layer restructuring.
- The first postsubmit after merge publishes the first `:buildcache`; until then presubmits fall back to today's inline cache (minus the eager pull). The one-time cachepub cold build is postsubmit-only and inside the new timeout.

## Testing

- `make docs-check` clean; prettier clean on changed YAML/Markdown; both Cloud Build YAMLs parse; `bash -n` on every embedded script and `hack/ci-deploy.sh`.
- `scripts/check_image_layers.py`: 114/120 against a freshly built image.
- `deploy/docker/test_image_startup_contract.py` (47 tests) green. One test needed updating with the restructure: `test_the_barrier_covers_the_personas_and_not_just_the_skills` asserted the root-ownership barrier appears in the manifest RUN's single chown line; it now asserts every root is behind *some* root chown in the platform stage (comment lines stripped first, so prose mentioning the chown cannot satisfy it) — the contract it protects is unchanged, only where the plugins barrier sits.
- The A/B table above: twelve real builds, each edit applied, built, reverted, against both Dockerfiles.
- Probe of the cache-probe step's three paths (source reachable / missing with `_REQUIRE_CACHE=true` / missing without): behaves as written.

### Live validation

This change's runtime surface is CI infrastructure plus image content, exercised as follows rather than against a running kube-agents installation:

- **Re-validated after rebasing onto #494's skill-provenance rework** (which rewrote the same region): all four images rebuilt from both Dockerfiles, and the `skills_manifest.sha256` files — the new tamper-evidence surface — are byte-identical between main's build and this branch's.
- **Image equivalence, measured:** full metadata walks (type, mode, owner, group, size) of the `platform` image (69,757 entries) and the `credential-proxy` image (69,438 entries) built from this branch vs `main`'s Dockerfile differ only in `__pycache__` byte jitter that two builds of the unmodified Dockerfile also produce. The `entrypoint-gate-test` stage (what `docker-build.yml` gates PRs on) builds green.
- **Cache mechanism, measured locally against a registry:** a container builder seeded `mode=max` into a local registry; a cold builder then rebuilt fully cached in 6.6s (81/81, watcher stage included — the thing inline cache cannot do) and 26.7s with a skill edit; the embedded docker-driver builder (the presubmit's path) imported the same cache.
- **Cloud Build worker facts, probed live in `haoxuw-gke-dev`:** three probe builds confirmed `gcr.io/cloud-builders/docker` bootstraps a container-driver buildx builder (build `8886117a`, SUCCESS) and that `docker manifest inspect` works with no experimental config (build `b48fa46b`, SUCCESS).
- **Not covered:** the new pipeline has not run end-to-end in Prow — the first postsubmit (cache publish) and the first presubmit consuming it are the confirming observations, and every new path is guarded to degrade to today's behavior rather than red a build. No test artifacts were left anywhere; the probe builds ran `--no-source` in my own project.

## Self-Review

`make docs-check` ran clean in the main loop. Both required passes ran in subagents handed only the diff range; merged dispositions:

- **[HIGH/CONFIRMED, adversarial] Extended `--cache-from type=registry,ref=…` syntax comma-splits on a pre-buildx `docker build` and silently imports nothing** → fixed: bare references in both presubmit steps (valid on every CLI generation; BuildKit resolves inline-cache images and cache manifests by content type), with the comment explaining why.
- **[HIGH/CONFIRMED-then-refuted, adversarial] `gcr.io/cloud-builders/docker` is Docker 20.10.24 with no buildx, so the publish step always skips** → refuted by live probe: build `8886117a` bootstrapped a container-driver builder inside that image on a real Cloud Build worker. The finding's source archaeology described an older image than workers serve. The guard it questioned stays regardless.
- **[HIGH/CONFIRMED-then-refuted, adversarial] `docker manifest inspect` is experimental-gated, making `_REQUIRE_CACHE=true` fail-always** → refuted by live probe: build `b48fa46b` ran it with no experimental config, SUCCESS. Same environmental premise as above.
- **[MEDIUM/CONFIRMED, adversarial] The first-RUN comment overclaimed "caches across every agents/ or deploy/shared change"** → fixed: the comment now states exactly which edits survive it and which invalidate agent-base beneath it.
- **[MEDIUM/PLAUSIBLE, adversarial] The cachepub cold case exceeds Cloud Build's 600s default and a timeout kills the image push the `|| echo` cannot catch** → fixed twice over: build-level `timeout: 3600s` plus a `:latest` inline `--cache-from` on the cachepub build, and — per jayantid's review of the same shape from the hang side — an inner `timeout 1500` on the buildx command (94a2b18), so a stalled cache push takes the warning path instead of riding to the external kill that fires after `images:` is lost.
- **[LOW/CONFIRMED, adversarial] `_REQUIRE_CACHE=true` was satisfiable by the proxy cache alone, and lost the old retry** → fixed: verdict keyed on platform sources only, one retry per ref, proxy ref reported informationally.
- **[LOW/CONFIRMED, adversarial] Stale layer-count comment** → fixed, with a history the bot review rightly asked me to state: the pre-rebase measurement was 115, then rebasing onto #494 (which folded the second bytecode pass into the final RUN) dropped the chain to a measured **114** — the Dockerfile comment, What Changed and Testing all say 114, and this line previously still said 115 from the earlier round.
- **[Advisory, docs-drift] `deploy/kustomize.md`'s directory tree understates `deploy/docker/` (pre-existing)** → deliberately not fixed: unrelated staleness, and this PR is scoped to the build optimization by the author's instruction; it remains for a docs sweep.
- Post-rebase re-derivation round (author context, stated as such): #494 landed a root-ownership barrier and skill manifests in the exact region this PR restructures. The reshuffle was re-derived on top (plugins barriered early, skills left with their manifest in the final RUN), and one real slip — a conflict-resolution edit that silently failed to apply, leaving the plugin tree hermes-owned where upstream demands root — was caught by re-running the equivalence walk and fixed before pushing.
- The docs-drift pass found no Blocking issues; it verified the edited reader-grant sentence against `verify_ci_pool_project.py`, the layer-budget prose, `docker-images.md`'s bytecode-pass ordering, and `dev_rebuild_agent.sh`'s empty-default skip path. The adversarial pass also recorded as non-findings: the eager-pull removal is sound, the docs COPY move is byte-identical, normalization equivalence holds, and `$$` escaping/substitution defaults are correct.
- What the passes could not cover, honestly: no live run of the new Prow pipeline (see Live validation), and the two refuted findings rest on my probes outrunning the reviewer's source reading — the first real postsubmit log settles both visibly.

## Risk & Rollout

CI-only plus a byte-equivalent image. Every new failure path degrades to current behavior: no buildcache manifest → inline-cache fallback (today's cache, minus the eager pull); publish-step failure → warning, image push unaffected; probe failure without `_REQUIRE_CACHE` → cold build, loud warning. The one behavior tightened is `_REQUIRE_CACHE=true`, which now demands a reachable platform cache source specifically. Revert = revert two commits; no state is left behind beyond `:buildcache` tags, which are inert data.


deploy/docker/cloudbuild.yaml
```diff
@@ -28,6 +28,40 @@ steps:
...
+        if [ -z "$_BUILDCACHE_URI" ]; then
+          echo "no _BUILDCACHE_URI; skipping mode=max cache publish"
+          exit 0
+        fi
+        if ! docker buildx version >/dev/null 2>&1; then
+          echo "WARNING: buildx unavailable on this worker; skipping mode=max cache publish (consumers fall back to inline cache)"
+          exit 0
+        fi
+        docker buildx create --name cachepub --driver docker-container >/dev/null 2>&1 || true
+        docker buildx build --builder cachepub \
```
jayantid: 🔴 **Not merge-blocking.** A hung cache publish kills the whole postsubmit before `images:` pushes — the one failure shape the fail-open guard cannot catch — and a rerun recovers, so this is a hardening fix, not a hold.

`|| echo` only catches exit codes. If `docker buildx build` stalls, the step runs until the build-level `timeout: 3600s` and Cloud Build kills the build externally. The shell never reaches the `|| echo`. The mode=max export pushes multi-GB of blobs to the registry, the operation most prone to stalling.

A killed build never pushes `images:`, so neither `:$GITHUB_SHA` nor `:latest` publishes for that commit and the postsubmit goes red — exactly what the step comment ("any failure skips with a warning") promises cannot happen. The first-publish cold build is the most exposed run; it fires once, and its duration has never been measured in CI.

Bound the command inside the step, keeping 3600s as the backstop:

```yaml
        timeout 1500 docker buildx build --builder cachepub \
          ...
          || echo "WARNING: mode=max cache publish failed or timed out; consumers fall back to inline cache"
```

jayantid (approved):

haoxuw:
seems to be a flaky test:

`
[TS 1787934926.410] ╭──────────────────────────────────────────────────────────────────────────────╮
[TS 1787934926.410] │ 🚀 DeepEval Evaluation Results                                               │
[TS 1787934926.410] ╰──────────────────────────────────────────────────────────────────────────────╯
[TS 1787934926.410] ╭──────────────────────────────────────────────────────────────────────────────╮
[TS 1787934926.410] │                                                                              │
[TS 1787934926.410] │  ❌ test_case_0                                                              │
[TS 1787934926.410] │  ├──   Input:              Which GKE clusters are we running in this         │
[TS 1787934926.410] │  │                         project right now? Give me the list.              │
[TS 1787934926.410] │  │     Actual Output:      {                                                 │
[TS 1787934926.410] │  │                           "tools_used": [                                 │
[TS 1787934926.410] │  │                             "multiuser_memory"                            │
[TS 1787934926.410] │  │                           ],                                              │
[TS 1787934926.410] │  │                           "execution_trace": [                            │
[TS 1787934926.410] │  │                             {                                             │
[TS 1787934926.410] │  │                               "name": "multiuser_memory",                 │
[TS 1787934926.410] │  │                               "args": {                                   │
[TS 1787934926.410] │  │                                 "target": "user",                         │
[TS 1787934926.410] │  │                                 "action": "read"                          │
[TS 1787934926.410] │  │                               },                                          │
[TS 1787934926.410] │  │                               "result": "{\"success\": true,              │
[TS 1787934926.410] │  │                         \"target\": \"user\", \"entries\": []}",          │
[TS 1787934926.410] │  │                               "status": "completed"                       │
[TS 1787934926.410] │  │                             }                                             │
[TS 1787934926.410] │  │                           ]                                               │
[TS 1787934926.410] │  │                         }                                                 │
[TS 1787934926.410] │  │     Expected Output:    The front door files a kanban card for the        │
[TS 1787934926.410] │  │                         question rather than guessing, and the            │
[TS 1787934926.410] │  │                         specialist's answer is delivered back into the    │
[TS 1787934926.410] │  │                         thread: a list of the project's GKE clusters      │
[TS 1787934926.410] │  │                         read from the live fleet, naming the three        │
[TS 1787934926.410] │  │                         seeded clusters (seeded-a, seeded-b and           │
[TS 1787934926.410] │  │                         seeded-c) from observed evidence. Extra           │
[TS 1787934926.410] │  │                         clusters in the list are expected — the           │
[TS 1787934926.410] │  │                         agent's own host cluster is in the same           │
[TS 1787934926.410] │  │                         project — and invented ones are not.              │`

if retest failed again I'll run the baseline without the PR. Wondering if HEAD has the same issue

bradhoekstra:

/review all

deploy/docker/cloudbuild.yaml
haoxuw: Fixed in 94a2b18 with your bound verbatim: `timeout 1500` inside the step, 3600s kept as the backstop, warning text and the fail-open comment updated to cover the hang case explicitly. You're right that the first cold publish is the exposed run — at 1500s it degrades to the inline-cache fallback instead of redding the postsubmit.

.github/workflows/docker-publish-gcp.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: |
```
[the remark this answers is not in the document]
haoxuw: Correct, and thank you for pinning it: the 115 was the pre-rebase measurement; rebasing onto #494 (its fold of the second bytecode pass into the final RUN) dropped the chain to a measured 114, and I updated the Dockerfile comment and the body's What Changed/Testing at that time but left the Self-Review disposition line stale. The disposition now states the 115→114 history explicitly; the layer budget stands at 114/120, measured with scripts/check_image_layers.py against a build of the current head.

PR #1008, opened 2026-08-27
fix(ci): report permission denials as unverified, not as missing resources

### LINKED ISSUE #1004: verify_ci_pool_project.py reports PERMISSION_DENIED as "resource absent"
`scripts/verify_ci_pool_project.py` reads every non-zero `gcloud` exit as "this resource does not exist". `PERMISSION_DENIED` is a non-zero exit. So an account that is authenticated but not privileged on the pool project gets specific, confident, wrong "not provisioned" findings for a project that is fully built.

## Reproduction

`kube-agents-evals-6`, run as `stalhaali@google.com` on 2026-08-27:

```
[❌] Service Accounts & IAM Grants: IAM requirements missing
    - Missing GSA or failed reading policy for kubeagents-platform-gsa@kube-agents-evals-6.iam.gserviceaccount.com
[❌] Artifact Registry Repository: Artifact Registry not ready
    - Missing Artifact Registry repository kube-agents in us-central1: PERMISSION_DENIED: Permission
      'artifactregistry.repositories.get' denied on resource ...
[❌] GKE Clusters & Terraform State: GKE/state resources missing
    - Missing Terraform state bucket: gs://kube-agents-evals-6-tf-state
[❌] Seeded Fleet Fixtures: Seeded fleet incomplete
    - 0/7 role(s) written, 7 on clusters that could not be resolved or reached, 0 whose fixtures were not present
[❌] Token Minter KMS & GSA: Token minter not provisioned / PEM key missing or wrong
    - Cloud KMS key github-token-minter-key ... PERMISSION_DENIED: Permission
      'cloudkms.cryptoKeyVersions.list' denied on resource ...

PRE-FLIGHT CHECK FAILED. Do NOT register kube-agents-evals-6 in Boskos until the above are resolved.
```

The same account, seconds later:

```
$ gcloud storage ls gs://kube-agents-evals-6-tf-state
gs://kube-agents-evals-6-tf-state/ci-pool-minter/
gs://kube-agents-evals-6-tf-state/full-install/
gs://kube-agents-evals-6-tf-state/seeded-fleet/

$ gcloud container clusters list --project kube-agents-evals-6
platform-agent-host  us-central1    RUNNING
seeded-a             us-central1-a  RUNNING
seeded-b             us-central1-a  RUNNING
seeded-c             us-central1-a  RUNNING
```

`kube-agents-evals-6` is already in the Boskos pool, and ninety minutes before that output it had passed a thirteen-task presubmit smoke run on those very clusters — [build 2092966698779938816](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/982/pull-kube-agents-smoke-test/2092966698779938816). Nothing the script reported as missing is missing.

## Cause

The state-bucket check is representative:

```python
state_bucket = f"gs://{project_id}-tf-state"
rc, out, err = run_cmd(["gcloud", "storage", "buckets", "describe", state_bucket])
if rc != 0:
    passed = False
    details.append(f"Missing Terraform state bucket: {state_bucket}")
```

`gcloud storage buckets describe` needs `storage.buckets.get`; `gcloud storage ls` does not. The account has the second and not the first, so the script writes "Missing" for a bucket whose contents it could have listed. This is the script's convention rather than one slip — there are 22 `run_cmd` call sites and 23 `if rc != 0` branches, and none of them distinguish a denial from an absence.

The GitHub half has the same shape. `GET /orgs/gke-agentic/installations` needs the `admin:org` scope; an operator PAT carrying `repo, workflow` gets `404 Not Found`, which the script reports as `GitHub App 4675512 installation not found on org gke-agentic`. The App is installed.

## Why `check_toolchain()` does not catch it

`check_toolchain()` was written for exactly this failure mode. Its docstring:

> Every check reads a non-zero gcloud/gh exit as "the resource is absent", so a missing binary or an expired credential produces a confident, specific and wrong "not provisioned" for a project that is fine. Catch it here and exit 2 -- unverified -- rather than 1.

It guards two cases: the binary absent from `PATH` (`rc == 127`), and an empty active-credential list. A credential that is present but lacks project IAM satisfies both and reaches every check downstream.

## Impact

The script's job is gating Boskos registration, and its failure text is an instruction — "Do NOT register `<project>` in Boskos until the above are resolved." An operator running it while missing one role reads that instruction about a healthy project, and the report is specific enough to be believed. It fails in the direction that matters: it never says "I could not tell", it says "it is not there".

## Suggested fix

`EXIT_UNVERIFIED = 2` already exists and `check_toolchain()` already uses it, so the distinction the script needs is modelled — it just is not reached from the checks. Route permission errors to it: detect `PERMISSION_DENIED`, HTTP 403, or `does not have permission` in the `gcloud` stderr and mark that check unverified rather than failed.

There is already a precedent in the file for preferring unverified over failed when the limitation is our own credential, in the App repository-list check:

> So report this as unverified rather than failed. Refusing to onboard a project because of a limit in our own credentials would be a false negative, and the operator can confirm it in one click.

The same reasoning applies to every other check; it was applied to one. The GitHub 404-on-missing-scope case wants the same treatment.

## Found via

#982, while attempting the `bench/tasks/DRAFTS.md` §A5 pool sweep. The instruction side of that problem is #1005.

Introduced in e38aeff (#979). Owning lane: #903.

---

stalhaali:
## Summary

`scripts/verify_ci_pool_project.py` treated a refused read as evidence the resource was absent, so an operator without full IAM on a candidate pool project got a screenful of `[❌]` accusations about resources that were there all along. This change teaches every check to tell "I was not allowed to look" apart from "I looked and it is not there": a read that did not happen files the item as unchecked and exits `2` — the existing "manual verification required" state — while a read that succeeded and found nothing still fails at `1`. It also covers two causes that are not refusals but leave the resource just as unread, a timeout and a credential that expired mid-run.

## Why This Change

GCP will not distinguish denial from absence — `gcloud` ends a denied read with `(or it may not exist)` on purpose. The verifier resolved that ambiguity the wrong way. Running it against `kube-agents-evals-6` as an operator with partial IAM reported eight failures and told the reader not to register the project; every one of those resources exists. The exit code says `1`, meaning "this project is not ready", when the truthful answer is "this run could not tell you". A pre-flight gate that manufactures failures is a gate people learn to ignore, and the wrong reading is the dangerous direction: it points the operator at provisioning work that is already done, and buries the one item that genuinely needed a human.

## What Changed

- A read that did not happen no longer counts as absence. `_record_unreadable()` classifies a non-zero command exit through `_unread_reason()` and files it as a warning rather than a failure, so the check keeps passing and the run exits `2`.
- `_unread_reason()` is `_denial_reason()` widened by `_UNREAD_PATTERNS` — timeouts, `invalid_grant`, lapsed refresh tokens, reauthentication prompts. The two stay separate because the Artifact Registry policy pair asks specifically whether a read was *refused*; everywhere else the consequence for the caller is identical.
- `_DENIAL_PATTERNS` gained `gh`'s own shape, `(HTTP 403)`. Its message half varies by endpoint and matches none of the other patterns, so a SAML-unauthorised or under-scoped token previously read as a missing GitHub App installation.
- The seeded-fleet check reads `hack/fleet-kubeconfigs.sh`'s own wording to decide fail against unverified, because the script reports "I could not reach that cluster" and "that cluster is not what the catalog describes" through the same `unresolved` count. A refused `clusters list` leaves it with an empty listing, from which it prints the *byte-identical* "carries no clusters labelled" warning it prints for a genuinely absent fleet — so where the script says it could not look, nothing it printed afterwards is evidence about the fleet. Excusing an unresolved role takes positive evidence of that, not the absence of a contrary warning: `_FLEET_COULD_NOT_LOOK` for the refused listing, `_FLEET_UNREACHABLE` for a refused `get-credentials` or a cluster skipped on a failed `mktemp`. Without that requirement a slot whose cluster kept its name and lost its labels is excused in silence.
- Where a check verified some of its items and not others, one `_partial_summary()` now names both halves. Naming only what was skipped hands the operator a longer manual list than they owe; naming only what passed is the false assurance exit `2` exists to prevent.
- `docs/site/src/content/docs/deploy/ci-pool-projects.md` describes the new reading, including both places that deliberately depart from it. `bench/tf/fleet/README.md` loses a clause that read as a guarantee. Two docs had `verify_ci_pool_project.py <project>`, missing the required `--project-id`.

## Context

Closes #1004.

**Overlaps #995** (@lapis2002, *Fix provision eval script*, still a draft). It adds a check that reads project-level `get-iam-policy` — the exact call my live run against `kube-agents-evals-6` shows being refused. On today's `main` that check would report "the runner is missing its roles" for a project that has them, which is #1004 at a new site. Merged after this PR it inherits the new classification for free, since it goes through the same `_record_unreadable()` path. The two branches touch the same file and will conflict textually; I have not coordinated branches, on the grounds that whichever lands second rebases once. #997 also touches `ci-pool-projects.md`, `+1/-1`, no interaction.

## Testing

- `python3 -m unittest test_verify_ci_pool_project` from `scripts/` — **156 tests, all pass** (110 on `main`). The denial classifier's fixtures are split into three tiers: refusals, genuine absences, and transient failures, with a test asserting no string in one tier is read as another.
- Two tests pin regexes against the file they quote. `test_the_fleet_warning_phrases_are_the_ones_the_script_prints` extracts all seven phrases from `hack/fleet-kubeconfigs.sh` rather than restating them, because a phrase reworded there and not here stops matching in silence — and it now breaks both ways: a `_FLEET_LOOKED_AND_FOUND_WRONG` phrase that stops matching drops every genuinely absent fleet from exit `1` to exit `2`, and a `_FLEET_UNREACHABLE` one that stops matching fails every project whose clusters were merely refused.
- Each of the eight tests added in the last round was run against the pre-fix source before being kept, to confirm it is red there rather than passing vacuously. Seven fail; `test_a_skipped_cluster_is_a_visibility_limit_too` passes both sides, because the old code excused that case too — for the wrong reason. It stays as a pin on the case now being excused deliberately, not as a regression test.
- `make docs-check` clean. `prettier --check` clean on the changed Markdown.

### Live validation

Run from `/tmp/kaverify` at branch head `8a8582d` on a Cloud workstation authenticated as `stalhaali@google.com`, an account with partial IAM on the pool projects — which is precisely the operator #1004 was filed about. This is a script, not a cluster workload, so there is no CR status or pod layer to observe; the layers it touches are the per-check verdict, the warning text, and the process exit code, and all three are below.

**`kube-agents-evals-6` — the reproduction. Exit `2`, zero failures, 11 items unchecked.** On `main` this same account gets exit `1` and eight `[❌]`s. Every warning carries the real refusal that caused it:

```
[?] Service Accounts & IAM Grants: the cross-project AR reader grants verified; the Workload Identity binding not checked
    ? Could not read the IAM policy on kubeagents-platform-gsa@kube-agents-evals-6.iam.gserviceaccount.com, so its Workload
      Identity binding was not checked (and neither was the GSA's existence): ERROR:
      (gcloud.iam.service-accounts.get-iam-policy) PERMISSION_DENIED: Permission 'iam.serviceAccounts.getIamPolicy' denied
      on resource '//iam.googleapis.com/projects/-/serviceAccounts/113405042032614
[?] Artifact Registry Repository: the repository and its cleanup policy, push rights, node pull rights not checked
[?] GKE Clusters & Terraform State: clusters and CMEK present; state bucket not checked
[?] Seeded Fleet Fixtures: 7 of 7 fixture role(s) not checked
[?] Token Minter KMS & GSA: the imported key versions, the key's purpose, algorithm and import-only setting, ... not checked

MANUAL VERIFICATION REQUIRED. Nothing failed, but 11 item(s) could not be checked automatically.
Confirm each one above before registering kube-agents-evals-6 in Boskos.
```

**`stalhaali-gkedemos` — the negative control. Exit `1`, all seven genuinely-absent checks still fail.** This is the half that proves the mechanism rather than a coincidence: the change would look identical from evals-6 alone if it had simply stopped failing on everything.

```
[❌] Codebase GitOps Mapping: No mapping for stalhaali-gkedemos in gitops_repo_for_project() in hack/ci-deploy.sh
[✓] GCP Project & APIs: Project number: 659400239941, all 9 APIs enabled
[❌] Service Accounts & IAM Grants: IAM requirements missing
[❌] Artifact Registry Repository: Artifact Registry not ready
[❌] GKE Clusters & Terraform State: GKE/state resources missing
[❌] Seeded Fleet Fixtures: Seeded fleet incomplete
[❌] GitOps Repo & GitHub App Installation: GitHub configuration incomplete
[❌] Token Minter KMS & GSA: Token minter not provisioned / PEM key missing or wrong

PRE-FLIGHT CHECK FAILED. Do NOT register stalhaali-gkedemos in Boskos until the above are resolved.
```

`[❌] Seeded fleet incomplete` is the line to read there: gkedemos' cluster listing *succeeds*, so neither `_FLEET_COULD_NOT_LOOK` nor `_FLEET_UNREACHABLE` matches and the genuinely absent fleet still fails. The new guard did not over-widen.

**Both runs were re-taken at `8a8582d`, after the two review fixes, and both transcripts are byte-identical to the ones taken at `a34c5d4`** (`diff` on the saved output, no differences). That is the intended result rather than a null one: the review fixes narrow when the fleet excuse applies and widen what counts as an unread call, and neither run supplies an input where those two differ — evals-6 prints three `no credentials for seeded cluster` warnings, so it stays excused on positive evidence, and gkedemos lists its clusters successfully, so it stays failed. The paths the fixes do change are the ones no live account here can produce, and they are covered by the eight tests above.

**The fleet-refusal path, reproduced against a real project.** The one path neither run above exercises is a `container.clusters.list` that fails outright — my workstation account can list clusters in every pool project. I reproduced it with a local gcloud whose refresh token had lapsed, running the real `hack/fleet-kubeconfigs.sh` against `kube-agents-evals-6`, which emitted the genuine pair:

```
WARNING: could not list clusters in kube-agents-evals-6; every fleet check will report status=error
WARNING: project kube-agents-evals-6 carries no clusters labelled environment=seeded,managed-by=kube-agents-seeded-fleet. ...
```

Driving `check_seeded_fleet_fixtures()` over that real transcript, with no mocking, gives `passed=True`, `7 of 7 fixture role(s) not checked`. Monkeypatching `_FLEET_COULD_NOT_LOOK` to never match — the only change, everything else still live — turns the same run into `passed=False`, `Seeded fleet incomplete`. That is the red proof for the guard.

The cause there is an expired credential rather than a missing `container.clusters.list` role, and I am not claiming otherwise. It is a faithful substitute because `hack/fleet-kubeconfigs.sh:192` discards gcloud's stderr, so the script cannot tell the two apart and emits the identical warning pair either way — which is exactly the input the verifier has to classify.

**What I could not cover.** A genuinely IAM-refused `clusters list` against a live project: I hold `container.clusters.list` on all five org projects I probed, and SA impersonation — which would have produced a zero-role principal — is blocked by corp policy (`ACCESS_TOKEN_TYPE_UNSUPPORTED`, "restricted due to a domain admin's policies"). The `gh: … (HTTP 403)` pattern is likewise unobserved: my token is SAML-authorised, so I could not make GitHub emit one. It is a pattern addition that is inert unless that string appears.

**Cleanup.** The throwaway service account created for the impersonation attempt (`zero-role-preflight@stalhaali-gkedemos`) and its `roles/iam.serviceAccountTokenCreator` binding are deleted. The local fleet kubeconfig scratch directory is removed. `/tmp/kaverify` on the workstation is a scratch checkout, left in place.

## Self-Review

Two `/pr-preflight` rounds before opening, four passes, each in a fresh subagent context handed the diff range and nothing else — none of them had the reasoning that produced the code — plus `kube-agents-bot`'s review of the opened pull request, whose two findings are the first entries below. Round 2 ran after round 1's fixes landed and found a regression in one of them; the bot then found a second-order version of the same mistake, which is why the rounds are listed rather than only the latest. Angles covered: correctness of the fail/unverified boundary, pattern completeness against real tool output, doc-to-source agreement, test quality and whether tests could pass vacuously, dead code, message consistency, and sibling pull requests.

**Both bot findings confirmed and fixed in `8a8582d`.**

- 🔴 *The fleet excuse ran on the absence of a contrary warning rather than on evidence the run could not look.* Correct, and it is my round-2 fix over-applied: guarding against a false failure left the excuse itself unconditional, so a slot whose cluster kept its name and lost its labels was excused with the untrue message that it could not be reached. That path prints nothing at all — no per-cluster warning names it, `labelled` stays non-zero so the unlabelled-fleet warning is silent, another slot resolves so the nothing-resolved warning is silent — and `check_gke_and_state` matches by name, so it passes the cluster too. **Fixed**: excusing now requires `_FLEET_COULD_NOT_LOOK` or the new `_FLEET_UNREACHABLE` (a refused `get-credentials`, a cluster skipped on a failed `mktemp`), both pinned to the script's own wording. My own `test_unresolved_cluster_is_unverified_not_failed` had encoded the bug by passing zero warnings; it is rewritten to carry the warnings the excuse now needs, and two tests cover the hole and the mktemp case.
- 🟡 *Five call sites still classified with `_denial_reason` alone, so the timeout and expiry handling this PR advertises did not reach them.* Confirmed at all five. The sharpest was the GitHub installations lookup, which answered a timed-out `gh api` with the flat claim `GitHub App 4675512 installation not found on org gke-agentic`. **Fixed**: four sites switch to `_unread_reason`; the Artifact Registry policy pair keeps the denials/errors split — the "partial policy" wording depends on it — and routes unreads into `policy_denials` instead. Six regression tests, one per site. The two doc paragraphs the finding named as falsified are true again with the code fixed, so they are unchanged.

**The most important pre-PR finding, raised independently by both round-2 passes:** my round-1 fleet fix hard-failed the exact refusal it existed to excuse. `_FLEET_LOOKED_AND_FOUND_WRONG` matched `carries no clusters labelled`, which `hack/fleet-kubeconfigs.sh` emits byte-identically when the listing was *refused* — so an operator without `container.clusters.list` got `[❌] Seeded fleet incomplete`, reintroducing #1004 at a new site. **Fixed** with `_FLEET_COULD_NOT_LOOK`, a regression test, and the doc paragraph; live-proved above.

Other pre-PR findings, all confirmed against source before I acted on them:

| Finding | Disposition |
|---|---|
| A refused IAM read was *discarded* when the sibling policy read succeeded, so the check hard-failed with a manufactured "no push rights" — the exact class this PR removes | **Fixed.** A grant found is conclusive; its absence only when both policies read |
| The round-1 comment claimed `check_gke_and_state` backstops an absent fleet. It matches cluster **names**; the fleet script discovers by **labels**, so a cluster that kept its name and lost its label slipped through | **Fixed** — discriminates on the fleet script's own wording instead |
| `_DENIAL_PATTERNS` missed `gh: <msg> (HTTP 403)` | **Fixed** — pattern plus three real `gh` strings |
| A timeout and an expired credential were still read as absence, and my `check_toolchain` docstring claimed otherwise — `gcloud auth list` reads the local store without contacting the network, so a revoked token prints ACTIVE | **Fixed both ways**: `_UNREAD_PATTERNS` handles them per-check, and the docstring now says what it cannot catch. Confirmed live with a genuinely lapsed local credential |
| The disabled-API guard test used an invented error string; gcloud's real one is `code=403` and takes the denial branch | **Fixed** — real string, plus a test pinning why `REQUIRED_APIS` still fails a disabled API |
| **PLAUSIBLE:** an absent service account might answer `PERMISSION_DENIED` via anti-enumeration, making "Missing GSA" unreachable | **Refuted by live test**, not by argument. IAM answers `NOT_FOUND: Unknown service account` even in a project the caller cannot read. The invented test string was replaced with the observed one |
| Four fleet phrases were hand-copied from another file's prose | **Fixed** — the test extracts them from `hack/fleet-kubeconfigs.sh` |
| Three checks phrased their partial summary three ways, and the IAM one named only the skipped half | **Fixed** — one `_partial_summary()`; visible in the evals-6 transcript above |
| Both `run_checks` tests left `check_seeded_fleet_fixtures` unmocked, firing live `gcloud` from a unit test | **Fixed** in both, including the pre-existing one. Suite runtime 0.259s → 0.03s |
| Per-cluster fleet log lines were counted as separate unverified items, inflating the banner | **Fixed** — folded into one warning; 15 → 11 |
| Doc claims contradicted by the code: "An account without project-level read therefore gets `2`", "a cluster matching no slot fails the check", "section 3's roles are what a run needs", "a refused read is never a missing resource", a hard-coded "Three items", a missing `artifactregistry.repositories.getIamPolicy` in the refusal transcript, `gh repo view` called a *check* when it is one call, and `bench/tf/fleet/README.md` reading as a guarantee | **All fixed.** The AR permission was live-observed rather than inferred |

**Two findings deliberately not fixed, both with reasons about this change:**

- `gh repo view` bypasses the denial classifier and fails on a 403. That is correct here and the code comment already says why: the check cannot proceed without the repo, and the run needs to stop rather than continue against a repo it cannot see. Both an adversarial pass and the docs-drift pass reached the same conclusion independently.
- The `and not policy_errors` guard is dead as the loop currently stands — reaching that branch means one policy was read, so the other contributed at most one of a denial or an error. **Kept, with a comment saying so**: it is the condition the branch *means* rather than the one the arithmetic happens to allow, and it stays correct if a third policy source is added. Removing it would make the branch accidentally right.

**What the passes could not cover.** Angle J (sibling pull requests) did not run — no `gh` in the subagent environment — so I ran it by hand via the API, which is how #995 surfaced. Neither pass had a live `gcloud`, so the two pattern findings above rested on message formats rather than observed output; I settled one live and left the other as an addition that is inert if the string never appears.

## Risk & Rollout

Contained to one pre-flight script and its docs; nothing in the operator, the charts, or the agent runtime. The failure mode this introduces is the mirror of the one it removes — a resource that is genuinely absent *and* whose absence surfaces through a denial-shaped error would now report `2` instead of `1`. That trade is deliberate: exit `2` prints "manual verification required" and names the item, so it asks for a human rather than lying to one, whereas the old reading told the operator to go provision something that already existed. Revert is a single commit; no state, no migration, nothing to do at merge time.


scripts/verify_ci_pool_project.py
```diff
@@ -769,17 +1088,65 @@ def check_seeded_fleet_fixtures(project_id: str) -> CheckResult:
...
+    # from anything the script saw. Where the script says it could not look, no
+    # note it printed afterwards is evidence about the fleet.
+    could_not_look = any(_FLEET_COULD_NOT_LOOK.search(n) for n in notes)
+    looked_and_found_wrong = (
+        []
+        if could_not_look
+        else [n for n in notes if _FLEET_LOOKED_AND_FOUND_WRONG.search(n)]
     )
-    return CheckResult(name, False, "Seeded fleet incomplete", details=details)
+    if unresolved and not unplanted and not looked_and_found_wrong:
```
jayantid: 🔴 **Not merge-blocking — exit 2 still gates registration — but this branch excuses genuine fleet drift with a false diagnosis, so it should be fixed before or immediately after merge.**

This branch excuses unresolved roles even when no warning evidences a visibility limit. An unresolved count with zero notes can only come from a listing that *succeeded* but was missing a slot: a refused listing prints `could not list clusters in` (`hack/fleet-kubeconfigs.sh:371`) and a refused get-credentials prints `no credentials for seeded cluster` (`:394`).

A seeded cluster that keeps its name but loses its labels passes `check_gke_and_state` (it matches names) and is simply undiscovered here. The run reports `2 of 7 fixture role(s) not checked`, and the warning at line 1141 says the clusters "could not be reached" — untrue; the listing succeeded under full IAM. `main` fails this project at exit 1. Reproduced: summary `5 written, 2 unresolved, 0 unplanted` with no notes → `passed=True`.

Fix — require evidence of a visibility limit before excusing (the fixture in `test_verify_ci_pool_project.py:400` then needs the warning lines a real refusal prints):

```python
_FLEET_UNREACHABLE = re.compile(r"no credentials for seeded cluster", re.I)
...
could_not_reach = could_not_look or any(_FLEET_UNREACHABLE.search(n) for n in notes)
if unresolved and not unplanted and not looked_and_found_wrong and could_not_reach:
```

scripts/verify_ci_pool_project.py
```diff
@@ -553,10 +767,16 @@ def check_artifact_registry(project_id: str, project_number: str, location: str
     ):
         rc, out, err = run_cmd(cmd)
         if rc != 0:
+            reason = _denial_reason(err)
```
jayantid: 🟡 **Not merge-blocking.** Timeouts and mid-run credential expiry still hard-fail at five call sites, and the new doc section says they don't.

`_unread_reason()` — timeouts, `invalid_grant`, reauthentication — is only consulted through `_record_unreadable()`. Five new call sites classify with `_denial_reason()` alone: lines 499, 525, 770, 1055 and 1195.

A credential that expires mid-run (your own live-validation scenario) or a 120s timeout produces exit 1. A timed-out `gcloud projects describe` reports "Project describe failed" plus two failed Skipped checks; a timed-out `gh api /orgs/gke-agentic/installations` reports "GitHub App 4675512 installation not found on org gke-agentic" — a manufactured absence claim. All four variants reproduced against this branch. `ci-pool-projects.md:268` says a timeout and an expired credential are read as unchecked, and `:270` lists only two deliberate departures; these five are a third.

Fix: consult `_unread_reason` at these sites — the AR policy pair keeps its denial/error split, with `_UNREAD_PATTERNS` matches routed to the excused bucket — or name these sites in the doc's departures paragraph.

jayantid (approved):

scripts/verify_ci_pool_project.py
stalhaali: Fixed in `8a8582d`. You are right, and it is my own round-2 fix over-applied: guarding against a false *failure* left the excuse itself unconditional, so the excuse fired on the absence of a contrary warning rather than on evidence the run could not look.

Took the fix as written, widened by one case. `_FLEET_UNREACHABLE` also matches `could not create a temporary file` (`hack/fleet-kubeconfigs.sh:386`) — a cluster skipped there ends up with no kubeconfig for a reason that says nothing about the fleet, same as `:394`. Both phrases are pinned to the script's text by `test_the_fleet_warning_phrases_are_the_ones_the_script_prints`, which now asserts 4 `_FLEET_LOOKED_AND_FOUND_WRONG` phrases and 2 `_FLEET_UNREACHABLE` ones so a phrase added on one side without a source match fails the suite.

Your fixture note was the sharper half of the finding: `test_unresolved_cluster_is_unverified_not_failed` passed zero notes, so my own test asserted the bug. It now carries the three `no credentials for seeded cluster` warnings a real refusal prints. Two tests added: `test_unresolved_with_no_warning_at_all_still_fails` is your reproduction (`5 written, 2 unresolved, 0 unplanted`, no notes → `passed=False`), and `test_a_skipped_cluster_is_a_visibility_limit_too` covers the mktemp case. I ran both against the pre-fix source; the first is red there, the second passes both sides because the old code excused it too, for the wrong reason, so it is a pin rather than a regression test and the PR body says so.

Live re-run at `8a8582d` is byte-identical to the `a34c5d4` transcripts: evals-6 prints three `no credentials` warnings so it stays excused at exit 2 on positive evidence, gkedemos lists its clusters successfully so `Seeded fleet incomplete` still fails at exit 1. Neither account can produce the drift path, which is why it is covered by test rather than by run.

scripts/verify_ci_pool_project.py
stalhaali: Fixed in `8a8582d`. Confirmed at all five sites before acting on it.

Four switch to `_unread_reason`: `projects describe` (:499), `services list` (:525), the fleet script's non-zero exit (:1055, whose message also stopped saying the script "was refused" when it may have timed out), and the installations lookup (:1195, now `_unread_reason(err) or _GITHUB_NOT_FOUND.search(err)`). The AR policy pair (:770) keeps the denial/error split as you suggested and routes `_UNREAD_PATTERNS` matches into `policy_denials` — the split has to survive because `policy_errors` is what feeds the "read from a partial policy" wording, and a read that never happened did not produce the partial answer that wording describes.

Six regression tests, each verified red against the pre-fix source: a timed-out `projects describe`, an `invalid_grant` on `services list`, a fleet script exiting 124, a timed-out `gh api /orgs/gke-agentic/installations` (your manufactured `installation not found on org gke-agentic`, which was the clearest of the four), both AR policies unread, and one AR policy timing out alongside a successful sibling — the last one pinning that a timeout is not reported as a partial read.

`ci-pool-projects.md:268` and `:270` are unchanged: with the code fixed there is no third departure to name, so the paragraphs are true as written. 156 tests pass, up from 148.

Issue #971, opened 2026-08-26
Out-of-scope defects found while diagnosing the RC E2E failures


mplakhtiy:
These came out of a session diagnosing the five failing tests in the scheduled RC Release E2E Pipeline (runs [32866087154](https://github.com/gke-labs/kube-agents/actions/runs/32866087154), 32883825538, 32900969147, [32953137904](https://github.com/gke-labs/kube-agents/actions/runs/32953137904)). Each one was verified against `main` and each was out of scope for the four branches that came out of that work, so they were collected here rather than widening those diffs.

Nothing here is blocking a release today. Several are latent in the sense that a second defect currently masks them.

## Status — updated 2026-09-03 (verified against main b0a4d7d1)

Nine of the thirteen findings are still open. Four are resolved and are kept below, marked, rather than deleted. (Thirteen, not nine, because the first section lists five separate sites.)

- **Open:** 1a (bench port-forward), 1b (observability skill), 1c (event-watcher README), 1d (chatops page), 3 (kustomize Service target port), 4 (401 arrives as a connection reset), 6 (stale readiness-probe docstring), 7 (`AGENT_LOCAL_PORT` collision), 8 (GPU check swallows a kubectl failure).
- **Resolved:** 1e (`hack/ci-deploy.sh` port-forward, #865 / `a00e8310`), 2 (`nc -z` health gate, #1063 / `c3ef6df0`), 5 (`API_SERVER_KEY` regenerated per apply, #1130 / `f5354e62` — the operator-side half is still open, see the item), 9 (RC repository variables undocumented).

One thing changed underneath the original framing: gVisor is now the **default** for every install, not an RC-only setting. `install.defaults.env:73` sets `DEFAULT_ENABLE_GVISOR="true"` (#865). The original text below said `install.sh:80` defaults it to `false`; that is no longer true, and it widens 1a–1d from "wrong on the RC environment" to "wrong on a default install".

Line numbers throughout have been re-checked against `b0a4d7d1`.

## The RC runs agents under gVisor, and five places still assume port-forward works

`vars.ENABLE_GVISOR` is `true` for the RC environment: step 2 of run 32866087154 logs `gVisor Sandbox Isolation: true`, provisions a `gvisor-pool` node pool, and renders `runtimeClassName = "gvisor"` on the gateway. Since #865 that is also the default for a plain `install.sh` run.

[...]

PR #1012, opened 2026-08-28
fix(ci): provision the Pub/Sub adapter and GitHub token minter the RC gate depends on

mplakhtiy:
## Summary

The RC promotion gate has been failing the same three E2E tests on every run since 26 August, and none was a regression in the candidate — each was something the RC environment was never provisioned to provide. The Pub/Sub adapter that turns an alert into agent work is a separate `AgentPlugin` that nothing installs, so every alert-driven test failed on silence; and the GitHub token minter was never deployed, so the credential sidecar answered `HTTP 502` with the reason logged where CI cannot see it. This installs both from the release scripts, points the suite at the repository the minter is actually scoped to, and fixes an `install.sh` bug that made the minter's KMS key impossible to create on the fallback path.

## Why This Change

Runs [32986207520](https://github.com/gke-labs/kube-agents/actions/runs/32986207520), [33018980784](https://github.com/gke-labs/kube-agents/actions/runs/33018980784), [33031877720](https://github.com/gke-labs/kube-agents/actions/runs/33031877720) and [33061389550](https://github.com/gke-labs/kube-agents/actions/runs/33061389550) each ended `3 failed, 17 passed, 9 skipped`, with the same three tests and the same causes every time — deterministic, not flaky. #969 and #972 cleared two of the original five; these are the remainder.

`test_stockout_ingress_alert_smoke` and `test_stockout_scenario[04]` fail because the Pub/Sub adapter ships only as `agentplugins/pubsub-platform` and is not in the agent image — `deploy/docker/Dockerfile` bakes `google_chat`, `slack` and `chat`, and installs only the `google-cloud-pubsub` library. A consumer like the stockout investigator contributes route config under `platforms.pubsub.extra.subscriptions` and nothing else, so with no adapter to read it the gateway opens no listener at all. `verify.sh` reports "no sign the adapter saw the message", and the scenario then spends its full 360s watch reporting that the agent never started an investigation.

`test_github_token_minting_and_connectivity` fails because the chart renders `github-token-minter` only under `githubMinter.enabled`, which `installer_common.sh` sets from `GITHUB_ORG`, `GITHUB_REPO` and `GITHUB_APP_ID` — and `rc-deploy-environment.yml` passed none of them.

Since #980 (2026-08-27) the suite carrying all three sits in a `continue-on-error: true` step, so runs 33061389550, 33090615451 and 33118143222 all report the identical `3 failed` and all conclude **success**. The failures are invisible rather than absent, and a green run after this merges will not by itself prove the fix — the pytest summary in the step-3 log is what to read. That structural problem is #1015 and is not fixed here.

## What Changed

- **New `scripts/release/install_pubsub_platform.sh`** installs the adapter via its own canonical, idempotent installer, then waits for the plugin to reconcile and the gateway's generation to settle. It runs as its own workflow step, before readiness, so the re-template it causes is already in flight when the rollout waits start. `wait_for_gke_readiness.sh` goes back to waiting only; the shared kubectl-context resolution moved to `common.sh` as `release_resolve_target` / `release_connect_kubectl`, since each workflow step is a fresh shell.
- **The failure policy lives in the workflows, not the script.** It exits non-zero when it cannot deliver ingress. `rc-release-pipeline.yml` marks the step `continue-on-error` because alert ingress feeds its optional suite alone; `e2e-nightly-matrix.yml` and `e2e-manual-runner.yml` let it fail, the former skipping it for the three `test_environment` values that read no alert, the latter behind a new `skip_pubsub_platform` input.
- **`rc-deploy-environment.yml` / `provision_rc_environment.sh`** forward the GitOps repository and App ID into `install.sh`, which already documents this as the only way an automated install can enable the minter. Deliberately **not** `vars.GH_ORG`/`vars.GH_REPO`: on the `rc` environment that pair is the *release* repository (`gke-labs/kube-agents`) that `common.sh`'s `get_target_repo` resolves for tagging, and pointing the minter at it would scope a live App token to this repository. New `GITOPS_ORG`/`GITOPS_REPO` variables keep them apart. A partially-set trio now warns instead of skipping the minter silently.
- **`rc-release-pipeline.yml` / `e2e_config.yaml`** take the probe's repository from those same two variables rather than a pin. `rc-e2e` previously hard-coded `test-org-kube-agent/agents-repo`, a placeholder no App is installed on; pinning the real one would only have moved the problem, since the same pair scopes the minter and a value written twice drifts.
- **`install.sh`** passes `--skip-initial-version-creation` when pre-creating the minter's KMS key, and stops discarding the errors from both KMS creates. The missing flag was only half the bug: the call ended `2>/dev/null || true`, so KMS's `INVALID_ARGUMENT: Import-only keys must skip initial version creation` went nowhere, the key was never created, and the import then failed against a key that did not exist. Adding the flag alone would have left a permission denial or a disabled API just as invisible, so both creates now keep their output and an assertion afterwards reports what actually failed.
- **New `.github/actionlint.yaml`** suppresses two false positives on `rc-deploy-environment.yml`. actionlint types a reusable workflow's `secrets` context from `on.workflow_call.secrets` alone and has no model of environment secrets, which resolve at job level for any job declaring `environment:`. Declaring the two under `workflow_call.secrets` is *not* the fix — the caller does not hold them and does not use `secrets: inherit`, so it risks binding the name to an unpassed, empty caller secret, which is the silent-empty-`GITHUB_APP_ID` failure this PR exists to remove.
- **`wait_for_gke_readiness.sh` gates litellm and the gateway separately** (420s and 900s) instead of running one 300s gate for both. The rule `tests/test_gateway_rollout_budgets.py` enforces is `startupProbe budget < rollout gate < progressDeadlineSeconds`, and the two Deployments have different ceilings: the gateway's deadline is pinned at 1200s by the operator against a 605s `startupProbe` budget, so its gate must clear 845s, while litellm sets no deadline and runs on Kubernetes' 600s default, so its gate must stay under 600s. No single number satisfies both. The two values are the ones `reusable-deploy-agent.yml` and `reusable-deploy-integrations.yml` already use for these same Deployments. This was invisible while the RC provisioned Standard clusters; once #1019 lands the RC cluster is Autopilot (`provision_rc_environment.sh` passes no `--cluster-mode`) and pays node scale-up plus a first image pull. `rc-release-pipeline.yml`'s step-3 budget goes 30 → 60 minutes to match, so the job timeout no longer fires before the script's own message.
- **Region defaults** move to `us-central1` across `e2e_config.yaml` (all five environments plus the defaults), `conftest.py`, `execute_e2e_tests.py` and the two e2e workflows, following the `rc` environment, from which the KMS location also derives.

## Context

Follows #969, #972 and #973, which fixed the other two failures in the same investigation. Findings collected during that work are in #971, and two of its items are closed here: the RC pipeline's repository variables are now documented in `scripts/release/README.md`, and the harness design doc's environment contract is corrected.

**Known gap, not fixed here.** Neither `pubsub-platform` nor `gke-stockout-investigator` can be installed by Terraform + Helm at all: the plugin images are absent from `images.json`, no workflow publishes them, and the chart templates no `AgentPlugin`. This PR makes the RC gate honest; it does not close that gap, so a user installing the stockout investigator still gets dead ingress. Filed as #1013 — that fix spans the release pipeline, the chart, the composition and the image inventory, and changes every user install, which does not belong behind a CI fix.

Three further issues were filed from this investigation and are deliberately out of scope: **#1015** (the gate reports success while the optional suite fails), **#1016** (`e2e-nightly-matrix.yml` and `e2e-manual-runner.yml` declare no `environment:`, so every `vars.*` they read is empty and their hardcoded region fallbacks are the only value they ever see), and **#1017** (`reusable-deploy-integrations.yml` would scope a live App token at the release repo if ever called for `rc` — the latent hazard the `GITOPS_*` split above avoids).

Overlaps to watch: #914, #865 and #1014 touch `install.sh` — #1014 on the `adopt_kms` path, a different function from `import_github_pem`, so no duplicated fix but a conflict risk. #504 reshapes the minter around multiple repositories, which supersedes this PR's single-repo framing in `scripts/release/README.md`; whichever merges second should re-read that section. #1003 bumps `reviewdog/action-actionlint` to v1.73.2, which ships the same actionlint 1.7.12, so the new config survives either merge order.

## Testing

- `tests/test_provision_rc_environment.py`, `tests/test_execute_e2e_tests.py`, `tests/test_install_script.py`, `tests/test_install_pubsub_platform.py` — 55 tests, all passing, plus 12 in `tests/test_rc_minter_repo_wiring.py` (new) and `tests/test_gateway_rollout_budgets.py` (extended to cover the RC readiness script). The full unit suite is 571 passed / 3 skipped. Eight cover `provision_rc_environment.sh` (exit-code propagation, PEM staging and removal on both outcomes, the partial-config warning); seven are new for the split.
- Two tests were verified red before green: removing `exit "${INSTALL_STATUS}"` fails the exit-code test, and the ordering test fails if a workflow waits for readiness before installing the adapter.
- **Round-three additions were verified red before green.** Dropping `--skip-initial-version-creation` from `install.sh`'s `gcloud kms keys create` fails the new KMS test; setting the gateway gate back to 300s fails the startup-budget assertion; setting litellm's to 900s fails both the default-deadline assertion and the do-not-share-one-gate assertion.
- `make docs-check` — clean. `prettier --check` at CI's pinned 3.9.6 on every changed Markdown and YAML file — clean.
- `bash -n` on all four changed shell scripts and `shellcheck -x` on the three in `scripts/release/` — no new findings; the one reported (SC2155 at `common.sh:5`) is pre-existing.
- **`actionlint` passes in CI** ([run 33167541774](https://github.com/gke-labs/kube-agents/actions/runs/33167541774)), which is what confirms the new config works. It could not be run locally — it is SIGKILLed in this environment, as are Go tests and Terraform — so before that run the config rested on a review pass reading actionlint 1.7.12's source to confirm the schema, the file location, and that `expr_parser.go` lowercases the property so the message matches. All 25 checks on this PR are currently green.

### Live validation

Partially live-tested, against project `kube-agents-rc`, region `us-central1`.

**What was exercised for real — the minter's KMS half.** `github-token-minter-keyring` was created in `us-central1` (09:41:27Z; it did not pre-exist), and `github-token-minter-key` created as import-only, `ASYMMETRIC_SIGN`, `RSA_SIGN_PKCS1_2048_SHA256`, software protection. The Minty CLI route failed (`go run` is SIGKILLed in this environment), so the documented `gcloud`-only fallback was used: PKCS#1 → PKCS#8 DER, import job to `ACTIVE`, wrapped upload. Version 1 reports `ENABLED`, imported 09:51:03Z via import job `gh-app-key-import`, with no stray versions.

This is also what confirmed the `install.sh` fix: `gcloud kms keys create … --import-only` without `--skip-initial-version-creation` returns `INVALID_ARGUMENT: Import-only keys must skip initial version creation`, observed directly.

**A correction to how the key match was recorded.** The hash quoted during the original import, `11b8b30c…`, is the sha256 of the **PEM text**, not of the SubjectPublicKeyInfo DER; the DER hash is `6ca22408…`. Anyone re-verifying the conventional way will get the second number and could wrongly conclude the key was swapped. The PEM-text hash is also encoding-sensitive (trailing newline, line wrapping), so it only demonstrates equality when both sides were serialised by the same tool. The App identity is independently confirmed from public metadata: `/apps/kage-token-minter` → id `4143620`, owner `gke-agentic`, permissions `contents/issues/pull_requests: write`.

**Two things that are true of the environment but not yet of a working minter.** The `rc` environment now has `GITOPS_ORG=gke-agentic`, `GITOPS_REPO=kube-agents-rc-infra` and secret `GH_APP_ID` (all set 10:30Z) — an earlier revision of this section said they were unset. Setting them still does nothing on `main`: no workflow there reads `GITOPS_ORG`/`GITOPS_REPO`, so `githubMinter.enabled` stays false; the wiring is in this PR. Separately, the KMS key has **no IAM bindings** (`get-iam-policy` returns a bare `etag`), so the minter GSA lacks `roles/cloudkms.signerVerifier` and cannot sign until Terraform adopts the key and applies `google_kms_crypto_key_iam_member.signer_verifier`. "Key verified" is not "minter ready".

**The round-three changes are not live-tested either.** The readiness gates, the manual runner's auto-skip and the new tests were exercised only locally — unit suite, shellcheck, `bash -n`, `make docs-check`, and mutation testing of each new assertion. The gates in particular are a claim about how long a cold Autopilot cluster takes, and only an RC run settles that; the numbers are taken from the repository's own measurements for these same two Deployments rather than guessed. Now that #1019 has merged, the next RC run is the first one against an Autopilot cluster, which is exactly what these gates exist for.

**What was not covered.** The Pub/Sub adapter install, the reconcile and generation-settling waits, and the three E2E tests themselves have not run against a cluster. The RC cluster was destroyed on 2026-08-28 during an unrelated region migration and the RC pipeline is currently disabled pending a separate change, so no environment was standing. Those paths are exercised only by the next RC run, and — per #1015 — that run's conclusion will not report them; the pytest summary in the step-3 log will. No test artifacts were left behind; the KMS keyring and key are intended permanent state, and KMS key rings cannot be deleted in any case.

## Self-Review

Three rounds. Both required passes ran each round, in **subagents given only the diff range**, told to derive intent from the diff and not to read commit bodies or notes. `make docs-check` ran first in the main loop each time. Angles A–J of `review-adversarial` were exercised, none skipped; round two added mutation testing of the new tests and a source read of actionlint 1.7.12. Round one covered the original diff, round two the six commits added since, and round three the whole diff again at `d8d4b0fa` after those landed. Findings from all three are merged below, superseded rounds dropped rather than stacked.

Fixed:

- **The Artifact Registry host-discovery block was reverted entirely** (MEDIUM/CONFIRMED). Added on a false premise: `plugin_image_resolve` already authenticates the exact push host via `plugin_image_check_credential` → `plugin_image_login` and aborts if it fails.
- **The AgentPlugin wait rested on an inverted invariant** (MEDIUM/CONFIRMED, both passes). `updatePluginStatuses` runs at the top of `reconcileWorkload` and returns no error, so a caught-up `observedGeneration` is consistent with the Deployment patch never landing. Added the gateway generation-settling window `test_stockout_investigation.py` already uses for this race. *(The citation for this moved twice: it is `platformagent_controller.go:527-529`, and `updatePluginStatuses` is the second statement, not the first — the two round-two passes disagreed on this and the line numbers were read directly to settle it.)*
- **The install lived in a script named `wait_for_*`** (the reviewer's framing; raised independently in review). A readiness check that builds an image and runs a Helm release means re-running it to diagnose a cluster mutates it, and the failure policy had to be argued in a comment because the script could not fail honestly. Split into its own script and step, which is what let the policy move into the workflows.
- **The "cannot sink the mandatory Chat gate" claim was false** (MEDIUM/CONFIRMED). `continue-on-error` covers failures the script detects. It does not cover an adapter that installs cleanly and then leaves the gateway unable to roll out — an unpullable plugin image keeps the agent pod from starting, and the unguarded `kubectl rollout status` in the next step then fails the job before the Chat gate runs. Equally true before the split. The claim is now stated accurately in the script header and the README rather than removed, because the split does narrow the blast radius.
- **The nightly matrix would have hard-failed dispatch runs that read no alert** (MEDIUM/CONFIRMED). Its `test_environment` input offers `gchat-e2e`, `audit-e2e` and `agent-plugin-e2e`; making the step fatal regressed those against the warning it replaced. `SKIP_PUBSUB_PLATFORM` is now set for exactly those three.
- **The generation-settle loop had no absolute ceiling** (LOW/CONFIRMED). Its stability window resets on every observed change, so a gateway whose generation kept moving was bounded only by the job's `timeout-minutes`. Added `GENERATION_SETTLE_TIMEOUT`, and seeded the first read so iteration one no longer always counts as a change.
- **The actionlint ignore patterns were broader than the false positive** (LOW/CONFIRMED). Matched unanchored on property name alone, so they would also have silenced a future `inputs.gh_app_id` typo in the same file. Both now also match the secrets object's shape.
- **Only half the KMS bug was fixed** (MEDIUM/CONFIRMED). Detailed under What Changed. The describe-assertion is the part that makes any failure legible; the warning says "could not be confirmed to exist" rather than "does not exist", because `describe` also fails on an IAM denial and asserting absence from that states more than was established.
- **The README's manual KMS recipe omitted `--skip-initial-version-creation`** (HIGH/CONFIRMED). The route it called "the better one" failed on its first command. Replaced with a link to the canonical `terraform/modules/github-minter/README.md`.
- **A failed KMS *creation* pointed at the README's PEM *import* recipe** (BLOCKING, docs-drift), which presupposes the key exists. It now prints the two create commands and defers to the README for the import that follows.
- **A partially-configured minter was skipped in silence** (MEDIUM/CONFIRMED). Now warns, naming which values are missing and why an `rc`-scoped secret created at repository scope arrives empty.
- **`AGENT_NAMESPACE` was parameterised but passed by no caller** (LOW/CONFIRMED, both rounds). A non-default namespace would have installed the adapter in one place and run the tests in another. All three workflows now forward it.
- **Stale docs the mechanical check cannot see** (BLOCKING ×4, docs-drift): the harness design doc's `GCP_REGION` default and its `GITHUB_ORG`/`GITHUB_REPO` rows; the `_qualify_repo` docstring; a comment claiming `e2e_config.yaml` pins the repository for `rc-e2e` after a commit on this branch removed that pin; and a deletion recipe in the new script that undercounted what the same commit added.
- **A drive-by README edit that made the row less accurate** (LOW/CONFIRMED, both passes). Renaming `execute_e2e_tests.sh` to `.py` was wrong: only `e2e-manual-runner.yml` invokes the `.py` directly, while the other two workflows in that row reach it through `make test-e2e`. Reverted.
- **`plugin_image_resolve` credited with skipping an already-published build** — that is `plugin_image_publish`. Carried over verbatim from the comment being moved.
- **The region flip left three callers behind** (MEDIUM/PLAUSIBLE, both). `execute_e2e_tests.py` and both e2e workflows still defaulted to `us-east4`. All moved.
- **No test covered the new script behaviour** (LOW/CONFIRMED). Fifteen now do across two files, mutation-tested in round two: swapping the install/readiness order, removing `continue-on-error` from the RC step, adding it to the nightly step, and re-adding the plugin install to the readiness script each produce the expected failure.
- **The headline `install.sh` fix shipped with no regression test** (MEDIUM/CONFIRMED, round three). `grep` over `tests/` for `skip-initial-version-creation`, `import_github_pem` and `kms keys create` returned nothing, while the other three behaviour changes on this branch had fifteen tests between them. Three tests now drive `import_github_pem` against a stub `gcloud` that records its argv: the create call carries the flag, a zonal region is reduced to the KMS region, and a key that cannot be confirmed warns instead of importing. Verified red by dropping the flag.
- **Nothing pinned the `GITOPS_*` → `GITHUB_*` wiring this PR spends twenty lines defending** (LOW/CONFIRMED, round three). The tidying edit onto `vars.GH_ORG`/`vars.GH_REPO` — the pair every other workflow uses — would scope a live App token at the release repository with every test still green. New `tests/test_rc_minter_repo_wiring.py` asserts both consumers read the GitOps pair, that neither falls back to the release pair, and that the two agree with each other.
- **The manual runner did not get the auto-skip the nightly does** (LOW/CONFIRMED, round three). Both offer the same seven `test_environment` choices and three of them read no alert, so a dispatcher leaving `skip_pubsub_platform` at its default paid for a Helm release, a plugin image build and a restart of the shared gateway their suite never uses. It now applies the nightly's `contains(fromJSON(…))` expression, with the input kept as an explicit override.
- **The nightly's comment described a default that does not exist** (LOW/CONFIRMED, round three). It claimed `test_environment` defaults to `'all'`; the default is `nightly-e2e`, and on a `schedule` trigger no `workflow_dispatch` default is applied at all. The conclusion held, the stated mechanism did not.
- **The RC readiness gate was a single 300s for two Deployments with different ceilings** (round three, and the open question round two deferred here as the file's owner). Detailed under What Changed. `tests/test_gateway_rollout_budgets.py` owned this invariant for the deploy workflows and simply did not know the RC script existed, which is why a 300s gate under a 605s `startupProbe` budget sat there unremarked; it now covers the script, and both directions were verified red.

Reported, not fixed:

- **The adapter is installed from a CI script rather than the install engine** (MEDIUM/CONFIRMED), against AGENTS.md's "the install has one engine". The finding is right and the gap is real — it is why a user install of the stockout investigator has dead ingress too. Not fixable here: closing it spans the release pipeline, the chart, the composition and the image inventory, and changes every user install. Filed as #1013; the script's header and the README say so rather than implying the adapter is properly installed, and the split makes it a single file to delete when #1013 lands.
- **An unpullable plugin image would take the whole gateway down** (MEDIUM/PLAUSIBLE), and image-volume support needs cluster ≥1.35. Depends on the RC's `REGISTRY_PREFIX`, cluster version and node-SA IAM, none of which either pass could read and none of which has been verified against a running cluster. Genuinely open; the next RC run settles it. It is also the mechanism behind the corrected Chat-gate claim above.
- **The KMS describe-gate turns a permission failure into a silent skip** (LOW/PLAUSIBLE). If `describe` fails for a reason other than absence, the function returns 0 and skips the import where it previously attempted and failed visibly. Narrow — a denial on `describe` almost certainly means `create` was denied too, and that error is printed alongside — and the wording was softened rather than the control flow changed. Settling it needs a project where the installer SA has `cloudkms.admin` but not `cloudkms.viewer`.
- **`install_pubsub_platform.sh` has no runtime coverage in CI** (Angle I). The new tests check that it parses, is executable, and that its callers wire it correctly; the plugin-ready loop, the settle loop and both exit paths were exercised against stubs during review, but that harness is not in the branch.
- **Line-number citations will rot** (Advisory, docs-drift). Accepted; the surrounding code cites line numbers the same way throughout. Round two found this is not theoretical — one citation shifted by a line during a rewrite in this very branch and was caught only by a pass that checked it.
- **`#504` supersedes the single-repo framing** in the new README section (LOW). Noted in Context rather than pre-emptively rewritten, since which model wins depends on merge order.
- **The docs map has no guard for `scripts/release/README.md`'s workflow table** (docs-drift). `docs/README.md`'s identifier-sources table claims no owner for the release scripts, so that table and its per-script bullets have no mechanical check at all. Worth a row there; not this PR's to add.

## Risk & Rollout

Confined to the release and E2E scripts, three workflows, one `install.sh` function and a new actionlint config; no operator, chart or agent runtime code changes, so nothing reaches a user install. With `GITOPS_ORG`, `GITOPS_REPO` and `GH_APP_ID` unset the generated `terraform.tfvars` is byte-identical to today's and the minter stays off.

The new failure mode is the adapter install: it adds a Helm release — and, when the plugin source has changed, an image build — to the start of three workflows. In the RC pipeline it is `continue-on-error`, so the worst case there is the status quo with a reason printed above it. In the nightly matrix and the manual runner it is fatal by design, and `SKIP_PUBSUB_PLATFORM` (or the manual runner's input) disables it without a revert. The residual risk is the one recorded above and not fixed: an adapter that installs and then breaks the gateway rollout still fails the job before the Chat gate.

Revert is a plain `git revert` of the range; nothing here is stateful. The one thing that is not revertible is outside the diff: the KMS key imported into `kube-agents-rc`, which is intended permanent state, and which KMS cannot delete regardless.


.github/workflows/e2e-manual-runner.yml
```diff
@@ -88,17 +93,39 @@ jobs:
         uses: google-github-actions/get-gke-credentials@3da1e46a907576cefaa90c484278bb5b259dd395 # v3.0.0
         with:
           cluster_name: ${{ vars.GKE_CLUSTER_NAME }}
-          location: ${{ vars.GCP_REGION || 'us-east4' }}
+          location: ${{ vars.GCP_REGION || 'us-central1' }}
```
eLeontev: I don't like any default fallbacks in the pipeline. The value must be passed directly from the github env.

There might be separate override or extra config in the e2e_config.yaml file if needed for the related test suite.

Please remove all fallbacks from the affected yamls used in the github actions.

.github/workflows/e2e-manual-runner.yml
```diff
@@ -88,17 +93,39 @@ jobs:
...
+      # workflows offer the same seven choices, so a dispatcher who leaves the
+      # input at its default should get the nightly's answer rather than a
+      # different one. skip_pubsub_platform stays as the explicit override.
+      - name: Install Pub/Sub Alert Ingress
+        env:
+          AGENT_NAMESPACE: ${{ vars.AGENT_NAMESPACE || 'kubeagents-system' }}
+          GCP_PROJECT_ID: ${{ vars.GCP_PROJECT_ID }}
+          GCP_REGION: ${{ vars.GCP_REGION }}
+          GKE_CLUSTER_NAME: ${{ vars.GKE_CLUSTER_NAME }}
+          SKIP_PUBSUB_PLATFORM: ${{ inputs.skip_pubsub_platform || contains(fromJSON('["gchat-e2e","audit-e2e","agent-plugin-e2e"]'), inputs.test_environment) }}
```
eLeontev: based on your doc - this parameter should be taken from the test_env or `e2e_suite` value. I expect that in the env we pass the test env name which define the test suite which will be executed and in the script we get the value should be skip the pubsub install or install it.

.github/workflows/e2e-manual-runner.yml
```diff
@@ -88,17 +93,39 @@ jobs:
...
+        run: ./scripts/release/install_pubsub_platform.sh
+
       - name: Connect to GKE & Configure Docker
         env:
+          AGENT_NAMESPACE: ${{ vars.AGENT_NAMESPACE || 'kubeagents-system' }}
```
eLeontev: the value should be defined in the test suite only

.github/workflows/e2e-manual-runner.yml
```diff
@@ -88,17 +93,39 @@ jobs:
...
           GCP_PROJECT_ID: ${{ vars.GCP_PROJECT_ID }}
           GCP_REGION: ${{ vars.GCP_REGION }}
           GKE_CLUSTER_NAME: ${{ vars.GKE_CLUSTER_NAME }}
         run: |
           ./scripts/release/wait_for_gke_readiness.sh
-          REGION="${GCP_REGION:-us-east4}"
+          REGION="${GCP_REGION:-us-central1}"
```
eLeontev: looks like it's time to move it to the separate script with tests, also do not use default.

.github/workflows/rc-deploy-environment.yml
```diff
@@ -134,6 +134,35 @@ jobs:
...
+          # All three must be set before the minter is provisioned at all
+          # (installer_common.sh:612-627); with any of them empty the install is
+          # byte-identical to today's. The pair must also match the repository the
+          # E2E suite probes — a token minted for one repository does not
+          # authenticate against another. `rc-e2e` deliberately pins neither in
+          # tests/e2e/e2e_config.yaml, precisely so the value lives in one place:
+          # rc-release-pipeline.yml passes this same GITOPS_ORG/GITOPS_REPO pair
+          # to the suite, and conftest.py's _qualify_repo composes the bare
+          # repository name against the org.
+          GITHUB_ORG: ${{ vars.GITOPS_ORG }}
```
eLeontev, line 156: please use the same naming between github env and variables names used in the scripts `GITHUB_ORG` and `GITOPS_ORG` are used for the different purpose

- GITOPS_ORG - for gitops
- GITHUB_ORG - for releases/tags etc.

I guess the right one is `GITOPS_*`

.github/workflows/rc-release-pipeline.yml
```diff
@@ -107,8 +107,25 @@ jobs:
...
+      # optional cluster/audit suite below and of nothing else, so a stockout-only
+      # problem must not sink the Google Chat gate that was made the blocking one.
+      # The script itself exits non-zero on failure; the decision to tolerate that
+      # is this workflow's, and lives here rather than inside the script.
+      # Runs before readiness so the gateway re-template the adapter triggers is
+      # already in flight when the rollout waits start.
+      - name: Install Pub/Sub Alert Ingress
+        continue-on-error: true
+        env:
+          AGENT_NAMESPACE: ${{ vars.AGENT_NAMESPACE }}
```
eLeontev, line 130: we don't have it in github env. if it is required for tests only please use e2e_suite as parametrise them, if it's required for the scripts/pipeline - please define in the github env

.github/workflows/rc-release-pipeline.yml
```diff
@@ -107,8 +107,25 @@ jobs:
...
+          GCP_PROJECT_ID: ${{ vars.GCP_PROJECT_ID }}
+          GCP_REGION: ${{ vars.GCP_REGION }}
+          GKE_CLUSTER_NAME: ${{ vars.GKE_CLUSTER_NAME }}
+        run: ./scripts/release/install_pubsub_platform.sh
+
       - name: Connect to GKE & Wait for Pod Readiness
         env:
+          AGENT_NAMESPACE: ${{ vars.AGENT_NAMESPACE }}
```
eLeontev, line 138: https://github.com/gke-labs/kube-agents/pull/1012/changes#r3880977495

.github/workflows/rc-release-pipeline.yml
```diff
@@ -137,6 +154,14 @@ jobs:
           E2E_ENV: "rc-e2e"
           STOCKOUT_SCENARIOS: "04"
           FLEET_AUDIT_STREAMS: "all"
+          # The same pair rc-deploy-environment.yml gives the installer, so the
+          # repository the suite probes is the one the minter was scoped to. Read from
+          # the environment rather than pinned in e2e_config.yaml because a value
+          # written in both places drifts, and a token minted for one repository does
+          # not authenticate against another. Bare: the conftest github_repo fixture
+          # composes owner/repo from GITHUB_ORG when GITHUB_REPO carries no slash.
+          GITHUB_ORG: ${{ vars.GITOPS_ORG }}
```
eLeontev, line 173: https://github.com/gke-labs/kube-agents/pull/1012/changes#r3880964992

scripts/release/common.sh
```diff
@@ -41,6 +41,59 @@ is_ci_pipeline() {
...
+# has to resolve the target itself. The pair lives here rather than being
+# duplicated, because the resolution order below is a contract with the
+# workflows: GKE_CLUSTER_NAME/GCP_REGION/GCP_PROJECT_ID are what the `env:` blocks
+# set, and CLUSTER_NAME/REGION/PROJECT_ID are the installer's own names, which a
+# developer running these by hand after install.sh already has exported.
+#
+# Assigns to globals rather than echoing: a caller reading an echo would need
+# command substitution, and a `set -u` abort inside a subshell would leave the
+# variable empty and the script running against an unnamed target.
+release_resolve_target() {
```
eLeontev, line 74: I would like to exit with error if any of the required param is not setup in pipeline, more than use a default fallback.

scripts/release/wait_for_gke_readiness.sh
```diff
@@ -18,38 +22,20 @@ echo "⏳ CONNECTING TO GKE & WAITING FOR POD READINESS"
 echo "Project ID:        ${PROJECT_ID}"
 echo "Region:            ${REGION}"
 echo "Cluster Name:      ${CLUSTER_NAME}"
+echo "Namespace:         ${AGENT_NAMESPACE}"
```
eLeontev: let's use the same key `Agent Namespace`

install.sh
```diff
@@ -1104,10 +1104,50 @@ import_github_pem() {
...
-  gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
+
+  # Both creates keep their errors instead of discarding them. Re-running the
+  # installer is the common case and "already exists" is the expected answer to
+  # it, so the output is only surfaced when the resource is missing afterwards —
+  # which is the check that actually matters. Discarding stderr outright is what
+  # hid the bug below; tolerating one specific error would still have hidden a
+  # permission denial, a disabled API or a quota refusal, all of which end the
+  # same way: no key, and an import that fails against something that is not there.
+  local kms_ring_err="" kms_key_err=""
```
eLeontev, line 1204: note: I don't like to see the changes in install script which is prod related changes in the same PR with infra changes. If you want to keep it we need to test if ideally before to merge, or test it in RC pipeline in live.

.github/workflows/e2e-manual-runner.yml
```diff
@@ -37,6 +37,11 @@ on:
         required: false
         default: "false"
         type: string
+      skip_pubsub_platform:
```
eLeontev: common: I would prefer to focus in this PR mostly on the fix of the failed tests in RC test suite more than try to fix all jobs which are not live and not running e.g. e2e-manual and nightly - they can be fixed and aligned separately as part of the other automation tasks.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Agreed, and the fallbacks here are worse than they look: `e2e-manual-runner.yml` declares no `environment:`, so every `vars.*` in it resolves against repository-level variables, and this repository defines none (`total_count: 0`). `vars.GCP_PROJECT_ID` and `vars.GKE_CLUSTER_NAME` are empty on every run; the `|| 'us-central1'` on this line is the only reason the region is populated at all. Removing the fallback without binding the job to an environment would move the failure rather than fix it.

That is a bigger change than this PR should carry, and it lines up with your point in https://github.com/gke-labs/kube-agents/pull/1012/changes#r3881053428 — I am dropping `e2e-manual-runner.yml` and `e2e-nightly-matrix.yml` from this PR and filing a follow-up that binds both to a GitHub environment and removes the fallbacks together. The RC path (`rc-release-pipeline.yml`, `rc-deploy-environment.yml`) already passes `vars.*` with no fallbacks.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Agreed in principle: the suite already knows whether it reads an alert, so the mapping belongs in `e2e_config.yaml` rather than a `contains()` in the workflow. It needs the same treatment in `e2e-nightly-matrix.yml`, and both files are leaving this PR per https://github.com/gke-labs/kube-agents/pull/1012/changes#r3881053428, so it goes in the follow-up with them.

The RC pipeline does not use the flag — it installs ingress unconditionally under `continue-on-error`.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Agreed. `e2e_config.yaml` already carries `namespace: kubeagents-system` per environment, so the workflow-level default duplicates it. Goes with the file in the follow-up (https://github.com/gke-labs/kube-agents/pull/1012/changes#r3881053428).

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Agreed, and the script already exists: as of this PR `wait_for_gke_readiness.sh` makes the `gcloud auth configure-docker` call itself (`scripts/release/wait_for_gke_readiness.sh:54`), so the inline `REGION=...; gcloud auth configure-docker` beneath it is dead duplication and the step collapses to the script call alone. `e2e-nightly-matrix.yml` has the same duplication.

Both go in the follow-up with the rest of these two files, per https://github.com/gke-labs/kube-agents/pull/1012/changes#r3881053428.

.github/workflows/rc-deploy-environment.yml
mplakhtiy: The distinction you are drawing is the one this line implements, and the GitHub side is already named the way you want. The `rc` environment defines both pairs: `GH_ORG=gke-labs` / `GH_REPO=kube-agents` for releases and tags — which is what `scripts/release/common.sh`'s `get_target_repo` reads — and `GITOPS_ORG=gke-agentic` / `GITOPS_REPO=kube-agents-rc-infra` for gitops, which is what this line reads. No release-scoped variable is involved here.

What lags is the name on the left-hand side, and it is the installer's input name rather than a release one: `install.sh:77` assigns `GITHUB_ORG` straight to `PARAM_GITOPS_ORG`, and `--gitops-org=` is the equivalent flag. The value is gitops-scoped end to end; only the env-var spelling is old.

Renaming that input to `GITOPS_ORG`/`GITOPS_REPO` touches the installer's documented contract and its `vars.sh` state (`installer_common.sh:638,772`, `install.sh:2194`), so it needs a deprecation shim for existing installs. Worth doing, but as its own PR — I will file it. Say if you would rather it block this one.

.github/workflows/rc-release-pipeline.yml
mplakhtiy: You are right, and it is a live gap rather than a style point: the `rc` environment defines no `AGENT_NAMESPACE`, so `vars.AGENT_NAMESPACE` expands to empty on every run today and the value only survives because `release_resolve_target` defaults it — the same fallback you flagged at https://github.com/gke-labs/kube-agents/pull/1012/changes#r3880994129.

Fixing it at the source: adding `AGENT_NAMESPACE=kubeagents-system` to the `rc` and `nightly` environment variables so the reference is honest, and tightening the script side per that finding so an unset value fails in CI instead of defaulting.

It belongs in the GitHub env rather than `e2e_suite` because the scripts need it, not only the tests — `install_pubsub_platform.sh` and `wait_for_gke_readiness.sh` both target the namespace.

.github/workflows/rc-release-pipeline.yml
mplakhtiy: Same answer as https://github.com/gke-labs/kube-agents/pull/1012/changes#r3880977495 — you are right that `AGENT_NAMESPACE` is not in the GitHub env today. Adding it to the `rc` and `nightly` environment variables, and making the script side fail rather than default when it is unset in CI.

.github/workflows/rc-release-pipeline.yml
mplakhtiy: Same answer as https://github.com/gke-labs/kube-agents/pull/1012/changes#r3880964992 — this line reads `GITOPS_ORG`/`GITOPS_REPO`, which is the gitops pair; the release pair (`GH_ORG`/`GH_REPO`) is untouched. Only the installer's input env-var name is still spelled `GITHUB_*`, and renaming that is a contract change I would rather do in its own PR.

scripts/release/common.sh
mplakhtiy: Agreed for the pipeline. Changing `release_resolve_target` to fail on an unset `GKE_CLUSTER_NAME`, `GCP_REGION`, `GCP_PROJECT_ID` or `AGENT_NAMESPACE` when `is_ci_pipeline` is true — the helper is already defined ten lines above — and keeping the defaults only for a developer running these scripts by hand after `install.sh`, which is what the block comment describes.

Silently defaulting `PROJECT_ID` to `kube-agents-rc` in CI is the worst case here: a misconfigured pipeline would act on a real project rather than stop.

That also closes the hole behind https://github.com/gke-labs/kube-agents/pull/1012/changes#r3880977495 — today an empty `vars.AGENT_NAMESPACE` silently becomes `kubeagents-system` instead of failing.

scripts/release/wait_for_gke_readiness.sh
mplakhtiy: Done — `Namespace:` becomes `Agent Namespace:`, matching the variable and the other scripts' output.

install.sh
mplakhtiy: The `install.sh` change is the fix rather than a drive-by. It sits in `import_github_pem()`, the GitHub App key import for the token minter this PR provisions: `gcloud kms keys create --import-only` is rejected with `INVALID_ARGUMENT: Import-only keys must skip initial version creation` unless `--skip-initial-version-creation` is passed, and the previous code sent that error to `/dev/null`. Without it the key is never created, the minter never passes readiness, and the RC gate this PR exists to fix stays red. `terraform/modules/github-minter` already sets `skip_initial_version_creation`; this brings the installer's pre-apply path in line with it.

Taking the second half of your ask: I will exercise it against the RC pipeline live before merge and record the result in **Live validation**, rather than merging on `tests/test_install_script.py` alone.

If you would still rather it ship separately, say so and I will split it — but the RC gate needs both.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Agreed, and taking it. `e2e-manual-runner.yml` and `e2e-nightly-matrix.yml` are coming out of this PR, which also resolves r3880925676, r3880942227, r3880944691 and r3880953369.

Your read that they are not live is exactly right, and worse than "not running": neither workflow declares an `environment:`, so every `vars.*` in both resolves against repository-level variables, of which this repository defines none (`total_count: 0`) — `cluster_name` and `project_id` reach `get-gke-credentials` empty. Making them work needs an environment binding, which is the follow-up, and the fallback removal belongs with it.

What stays is the RC path: `rc-deploy-environment.yml`, `rc-release-pipeline.yml`, `scripts/release/*`, the `rc-e2e` block in `e2e_config.yaml`, `conftest.py`, `install.sh`, and their tests.

.github/workflows/rc-deploy-environment.yml
eLeontev: proposal: we need to create a separate task to align the naming in install script to reduce misunderstanding.

eLeontev (approved):

.github/workflows/rc-deploy-environment.yml
mplakhtiy: Agreed, and filed: #1026.

It carries the rename (`GITOPS_ORG`/`GITOPS_REPO` as the installer's input names), the deprecation shim the existing `vars.sh` state needs, and the four call sites that pin the current contract. Related to #1017, which is the same hazard reached from the workflow side.

Issue #1026, opened 2026-08-28
Align the installer's GitOps repository inputs to GITOPS_ORG/GITOPS_REPO


mplakhtiy:
The installer takes its GitOps repository coordinates as `GITHUB_ORG` / `GITHUB_REPO`, but the value is GitOps-scoped end to end. `install.sh:81-82` assigns them straight into `PARAM_GITOPS_ORG` / `PARAM_GITOPS_REPO`, and `--gitops-org=` / `--gitops-repo=` (`install.sh:192-193`) are the equivalent flags. Only the environment-variable spelling is old.

That collides with a second, genuinely different pair. On the `rc` and `nightly` GitHub environments:

- `GH_ORG=gke-labs` / `GH_REPO=kube-agents` — the **release** repository, read by `scripts/release/common.sh`'s `get_target_repo` for tagging and release operations.
- `GITOPS_ORG=gke-agentic` / `GITOPS_REPO=kube-agents-rc-infra` — the **GitOps** repository.

So a workflow wiring the installer has to write `GITHUB_ORG: ${{ vars.GITOPS_ORG }}`, which reads like a mistake and invites someone to "fix" it to `vars.GH_ORG`. That would silently scope a live GitHub App token at the release repository. #1017 is the same hazard reached by a different route, and its "should not be simplified later" note is the symptom this issue is about.

## What to change

Accept `GITOPS_ORG` / `GITOPS_REPO` as the installer's input names, so the spelling matches the meaning and matches the GitHub environment variables already named that way.

## Why it needs its own PR

`GITHUB_ORG` / `GITHUB_REPO` is a documented contract with existing installs, not just an internal name:

- `k8s-operator/scripts/installer_common.sh:638` — the three-way guard that decides whether the minter is provisioned at all.
- `k8s-operator/scripts/installer_common.sh:772-773` — composes `github_repo = "${GITHUB_ORG}/${GITHUB_REPO}"` into tfvars.
- `k8s-operator/scripts/installer_common.sh:337-370` — `check_github_org_is_organization`, whose error text names `GITHUB_ORG` and points at `scripts/vars.sh`.
- `install.sh:1472-1474` and `install.sh:2242-2244` — persisted into `vars.sh` state, so an existing install already has these names on disk.

[...]

Issue #1017, opened 2026-08-28
reusable-deploy-integrations.yml would scope a live GitHub App token at the release repo if ever called for rc or nightly


mplakhtiy:
## Status — updated 2026-09-03 (verified against main b0a4d7d1)

Still open, and now covers two environments. The mapping is unchanged, none of the three suggested
resolutions below has been implemented, and #1058 (merged 2026-08-31, `aee7829f`) added a `nightly`
environment carrying the same release-repo coordinates as `rc`. The "separately worth confirming"
question at the bottom is answered: `staging` and `autopush` are clean.

---

The `rc` and `nightly` GitHub environments each hold a `GH_APP_ID` secret alongside variables
`GH_ORG=gke-labs` and `GH_REPO=kube-agents` — the **release** repository, which
`scripts/release/common.sh`'s `get_target_repo` uses for tagging and release operations. `rc` had
this pair when the issue was filed; `nightly` acquired it with #1058 (`aee7829f`, 2026-08-31), which
created the environment. Both also carry `GITOPS_ORG=gke-agentic` with
`GITOPS_REPO=kube-agents-rc-infra` and `kube-agents-nightly-infra` respectively.

`reusable-deploy-integrations.yml:359-361` maps the release pair straight into the token minter:

```yaml
GITHUB_ORG: ${{ vars.GH_ORG }}
GITHUB_REPO: ${{ vars.GH_REPO }}
GITHUB_APP_ID: ${{ secrets.GH_APP_ID }}
```

and `reusable-deploy-integrations.yml:390-395` then runs `helm upgrade --set githubMinter.enabled=true`
with `--set-string githubMinter.org/repo/appId` from those three. The chart renders a minty rule
scoping tokens to `{org}/{repo}`: `charts/kube-agents/templates/github-minter.yaml:69-74` sets
`repositories: ['{{ $repo }}']` with `contents: write`, `pull_requests: write`, and `issues: write`.

Called with `target_environment: rc` or `target_environment: nightly`, that would mint live GitHub
App tokens **scoped at `gke-labs/kube-agents`** — write access to `contents`, `issues` and
`pull_requests` on this repository, handed to the agent.

## Why it is latent rather than live

Nothing invokes that workflow with `rc` or `nightly` today. It has exactly two callers, and both
pass a hard-coded literal: `staging-redeploy-integrations.yml:62` passes `staging` and
`autopush-redeploy-integrations.yml:63` passes `autopush`. It is one line from being real, and the
[...]

PR #1029, opened 2026-08-28
fix(ci): inherit secrets into the RC deploy so the token minter is provisioned

mplakhtiy:
## Summary

The RC pipeline never delivered the `rc` environment's `GH_APP_ID` into the deploy job, so the GitHub token minter was skipped on every run. `rc-release-pipeline.yml` called `rc-deploy-environment.yml` with an explicit `secrets:` mapping; a `uses:` job cannot declare an environment, so that mapping resolved against nothing and forwarded an empty string. This switches the call to `secrets: inherit`, which is what `staging-redeploy-integrations.yml` and `autopush-redeploy-integrations.yml` already do, for a trap their comments already describe. It also makes a half-configured minter refuse the deploy instead of warning past it, and corrects the comment in `.github/actionlint.yaml` whose false premise let the defect through review.

## Why This Change

#1012 wired the RC token minter and added `.github/actionlint.yaml` to silence actionlint's complaint that `gh_app_id` was not defined on the secrets context, on the reasoning that `deploy-rc` declaring `environment: rc` was enough to resolve it at runtime. That reasoning was wrong, actionlint was reporting a real defect, and the suppression hid it.

The consequence is not loud, which is why it survived. `installer_common.sh` provisions the minter only when all three of `GITHUB_ORG`/`GITHUB_REPO`/`GITHUB_APP_ID` are non-empty, and prints its own "minter deferred" warning only in that same case — so one missing value skips the minter silently. The test that would catch it, `test_github_token_minting_and_connectivity`, runs under the `rc-e2e` step, which is `continue-on-error: true`. A broken minter has therefore been costing a tolerated HTTP 502 in an optional step while the candidate validated anyway.

Without this, the RC gate keeps certifying releases whose token-minting path was never exercised.

## What Changed

- **`.github/workflows/rc-release-pipeline.yml`** — `secrets: inherit` on `step-2-deploy-env`. This is the fix; everything else is guardrail. Carries `# zizmor: ignore[secrets-inherit]`, matching both existing precedents.
- **`scripts/release/provision_rc_environment.sh`** — a partially-set minter config now exits non-zero, **above** `rc_teardown_run`. Placement matters: below it, the script would refuse an environment `uninstall.sh` had already destroyed and leave the RC down. All three values empty stays allowed — that is a deliberate no-minter install, the default everywhere outside the RC.
- **`.github/actionlint.yaml`** — the suppression stays, because actionlint genuinely cannot model environment secrets, but its justification no longer asserts the false premise. It now names the runtime guard and the test as the real backstops, and states that the patterns are file-scoped rather than job-scoped.
- **`tests/test_rc_minter_secret_wiring.py`** — pins `secrets: inherit` on the caller and `environment: rc` on the callee. Verified the first assertion fails against `main`.
- **`scripts/release/README.md`** — single home for what the wiring requires; the other files point here instead of restating it.

## Context

Follows #1012, which introduced the defect. Related: #1017 (the same hazard reached from the workflow side) and #1026 (aligning the installer's `GITHUB_ORG` input to `GITOPS_ORG`).

## Testing

- `python3 -m pytest tests/ scripts/ --ignore=tests/e2e --ignore=tests/integration -q` → 1213 passed, 3 skipped.
- `make docs-check` clean; `prettier --check` clean; `shellcheck -x` clean on the changed script.
- The `secrets: inherit` assertion was checked against `main` and fails there, so it is a real regression test rather than one that passes either way.
- `actionlint` was **not** run — the local binary is SIGKILLed in this environment. The ignore patterns are byte-identical to `main`, so only the comment above them changed.

### Live validation

Observed on the `rc` environment, run [33184000778](https://github.com/gke-labs/kube-agents/actions/runs/33184000778) (`workflow_dispatch`, commit `79b18b37`, the #1012 merge):

- The `deploy-rc` job (started `15:12:47Z`) emitted the annotation `GitHub token minter not provisioned :: Set: GITHUB_ORG GITHUB_REPO; empty: GITHUB_APP_ID`.
- `GH_APP_ID` exists as an **`rc` environment secret**, created `10:30:22Z` — four and a half hours before that job started. `gh api repos/gke-labs/kube-agents/environments/rc/secrets` confirms it; it is not a repository secret.
- `deploy-rc` declares `environment: rc` on that commit.

So the secret existed, on the right environment, for a job that declared that environment, and still arrived empty. The only remaining difference from the working staging/autopush path is the caller's explicit `secrets:` mapping. The two `vars` in the same block (`GITOPS_ORG`, `GITOPS_REPO`) resolved correctly in the same run, which is the asymmetry the fix predicts.

**Not covered:** the post-fix half. This branch is not on the commit that run tested, so I have not observed `GITHUB_APP_ID` non-empty. That is one RC run away and should be the merge gate — the check to make is that `deploy-rc` no longer emits the annotation and that `test_github_token_minting_and_connectivity` passes. GitHub's reusable-workflow documentation can be read as saying `environment:` alone should have sufficed; the run above contradicts that reading, and this section is the evidence rather than an appeal to the docs. No test artifacts were created; nothing to clean up.

## Self-Review

Both required passes ran in **subagents that did not write the change**, against `upstream/main...HEAD`, each given the diff range and the gate output and nothing else — not the reasoning, not the commit messages.

**Adversarial pass.** Angles run: correctness of the secret-delivery mechanism, operational blast radius and security, ordering/destructiveness, test quality, reuse, altitude, sibling PRs.

- **Guard fired after the teardown** (MEDIUM, confirmed by execution) — **fixed.** It sat 120 lines below `rc_teardown_run`, so a refusal destroyed the RC first. Moved above it; a test now asserts the teardown banner never prints.
- **The stated justification was false** (HIGH, confirmed) — **fixed.** Three places claimed a missing minter makes the run "die" in `test_github_token_minting_and_connectivity`. That step is `continue-on-error: true`, so it does not. Reworded to say the hard fail is a deliberate tightening, and that the tolerated failure is why this went unnoticed.
- **The mechanism is contested** (HIGH, plausible) — **answered with evidence, not code.** The pass cited GitHub docs suggesting `environment:` alone suffices, and open runner issues claiming otherwise. It could not execute Actions. The live-validation section above resolves it empirically for this repository: secret present on the environment, job declaring the environment, value still empty. I did not weaken the hard fail on the strength of the docs reading, but the next RC run is the confirmation and I have said so rather than claiming it is settled.
- **`secrets: inherit` widens what the deploy job receives** (MEDIUM, confirmed) — **accepted, not fixed.** `deploy-rc` now sees every org/repo secret, including `RELEASE_BOT_TOKEN`, and it runs `install.sh` from a candidate commit. There is no narrower alternative: the caller cannot resolve an environment secret to put in a mapping, and moving `GH_APP_ID` to repository scope widens it differently. Both existing callers of the sibling workflow made the same trade. Naming it here rather than leaving a reviewer to find it.
- **Mechanism restated in five files** (MEDIUM, confirmed) — **fixed.** `scripts/release/README.md` is the single home; the rest point at it.
- **Two assertions pinned to a deleted string** (LOW, confirmed) — **fixed.**
- **Two new tests string-matched shell source** (LOW, confirmed) — **fixed** by deleting them; `test_provision_rc_environment.py` already covers that behaviour by executing the script.
- **actionlint comment narrated its own edit history** (LOW, confirmed) — **fixed.**

**Docs-drift pass.** Checked: the rewritten comment blocks, every source citation, whether any doc still described the old behaviour, docs-map obligations.

- **Wrong PR number, and PR-status prose** (Blocking) — **fixed** by deleting the clause; AGENTS.md forbids it regardless of the number.
- **"A repository-scope value is invisible to the deploy-rc job"** (Blocking) — **fixed.** Backwards: `vars` resolve environment-over-repository and `secrets: inherit` carries repository secrets, so the wrong scope quietly works. That is the real hazard, and `rc-deploy-environment.yml` already said so. Corrected in the README and in the operator-facing `::error`.
- **Same wrong mechanism in the `::error` text** (Blocking) — **fixed.**
- **Docstring rationale false; README omits the torn-down RC; actionlint backstop overclaimed** (Advisory ×3) — **fixed**, the first two dissolved by moving the guard above the teardown.
- **Design doc's zizmor list now incomplete; `reusable-deploy-integrations.yml` calls `GH_APP_ID` a "repository variable"** (Advisory ×2) — **not fixed.** The first is a point-in-time audit record; the second is pre-existing and belongs with #1026's naming cleanup.

Neither pass could run `actionlint` or reach a cluster.

## Risk & Rollout

Two changes with different profiles. `secrets: inherit` is the security-relevant one — it broadens the deploy job's secret context as described above, matching what two sibling workflows already do. The hard fail is release-blocking by design: a half-configured minter now stops step 2, which skips steps 3–5, so no candidate is tagged validated until the configuration is fixed. That is the intent, but it is strictly more blocking than today, and it is the change to revert first if an RC run refuses for a reason that turns out to be wrong — dropping the `exit 1` restores the old warn-and-continue without touching the wiring.

Both revert cleanly and independently. No runtime code paths, no chart or operator changes.


.github/workflows/rc-release-pipeline.yml
```diff
@@ -46,14 +46,25 @@ jobs:
...
+    # here because the mistake is made at this line.
+    uses: ./.github/workflows/rc-deploy-environment.yml # zizmor: ignore[secrets-inherit]
     permissions:
       contents: read
       id-token: write
     with:
       rc_tag: ${{ needs.step-1-create-tag.outputs.rc_tag }}
-    secrets:
-      GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+    secrets: inherit
```
eLeontev, line 67: not sure if we can accept it. it directly contradicts the zizmor policy
The right one is to pass the values from github secrets -> to pipeline env -> to the script.

eLeontev (approved):
looks like there is some issue with pipeline secrets passing between the jobs. We need to figure out the correct way to manage it instead of inherit or disable inherit rule on zizmor as incorrect in case if there is no options to workaround it

Issue #1009, opened 2026-08-27
Operator RBAC skew: `make deploy` defaults to `:latest`, so a pod reschedule silently upgrades the controller past its applied RBAC


Fuxiao-Gao:
## Summary

A controller deployed via the dev path (`make deploy`, which defaults to `VERSION ?= latest`) silently upgrades itself whenever its pod is rescheduled — the node pulls a newer `:latest` image while the RBAC/CRD manifests in the cluster stay frozen at whatever was last applied. When the newer controller needs a permission the old ClusterRole doesn't grant, every reconcile fails forever, with no signal other than a repeating `Reconciler error` log line (and, downstream, a stream of agent error notifications).

This is not a request to abandon `:latest` everywhere — the supported installer already enforces immutable refs (the friction discussed in #915). It's about (a) the unprotected `make deploy`/kustomize path, and (b) the operator having no way to notice the skew and fail loudly.

## Concrete incident

On a cluster installed via the kustomize path on **2026-08-12** (ClusterRole `kubeagents-manager-role` creationTimestamp `2026-08-12T18:17:30Z`):

1. **2026-08-19** — #733 merged: the controller now server-side-applies a PodDisruptionBudget for the platform agent, and `k8s-operator/config/rbac/role.yaml` gained write verbs on `policy/poddisruptionbudgets`. Code and RBAC shipped together — correct.
2. **2026-08-23 19:56:03Z** — the controller pod was rescheduled. `image: ghcr.io/gke-labs/kube-agents/k8s-operator:latest` pulled the post-#733 build.
3. **2026-08-23 19:56:29Z** (26s after start) — first failure, repeating on every reconcile since (449 occurrences over ~4 days):

```
failed to apply PodDisruptionBudget kubeagents-system/platform-agent:
poddisruptionbudgets.policy "platform-agent" is forbidden:
User "system:serviceaccount:kubeagents-system:kubeagents-controller"
cannot patch resource "poddisruptionbudgets" in API group "policy"
in the namespace "kubeagents-system"
```

The deployed ClusterRole still had only `get,list,watch` on `poddisruptionbudgets`. Nothing in the cluster or the logs points at the actual cause (image drifted past applied manifests); it reads like an RBAC misconfiguration.

## Proposals

[...]

Issue #534, opened 2026-08-05
PlatformAgent status has no observedGeneration, so its Ready condition can describe a stale spec


bnaylor:
## The problem

`AgentStatus` in `k8s-operator/api/v1alpha1/common_types.go` carries `Phase`, `Address`, `LastReconcileTime`, `Conditions`, and the deployment/service/storage sub-statuses.  It has no `observedGeneration`.

The Ready condition is derived from the live replica count, in `platformagent_controller.go`:

```go
if errWorkload == nil && newDeploymentStatusReadyReplicas > 0 {
    newPhase = "Ready"
    condStatus = metav1.ConditionTrue
    ...
}
```

So `Ready=True` means "some replica is ready right now".  It does not mean "the spec you just applied is running", and nothing in the status lets a caller tell the two apart.

## Why it matters

Anything that gates on the condition can accept a stale success.  Applying a changed spec to a healthy agent leaves `Ready=True` in place until the old Pod stops counting toward `ReadyReplicas`, and the default single-replica Deployment uses the `Recreate` strategy, so there is a real window where the previous generation is still satisfying the condition.

A caller doing the obvious thing:

```bash
kubectl apply -f platform-agent.yaml
kubectl wait --for=condition=Ready platformagent/platform-agent
```

returns immediately on the old status and has verified nothing.  Re-apply with a bad image tag or an unsatisfiable `runtimeClassName` and the wait still passes while the new Pod fails behind it.

This came up in #533, where stage 08 needed a readiness gate.  The workaround there is to gate on `kubectl rollout status` for the Deployment, which is generation-aware, plus a bounded wait for the operator to bump the Deployment generation first.  That works, but it is a shell heuristic standing in for information the API should be publishing.

## Precedent

`AgentPluginStatus` already has the field, and the same controller maintains it:

```go
// ObservedGeneration is the .metadata.generation the status was last computed from.
// +optional
ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```

```go
plugin.Status.ObservedGeneration = plugin.Generation
```

`platformagent_controller.go` sets it for plugins at two sites and compares it in the status-equality helpers.  So the omission on the agent side looks accidental rather than deliberate.

## Suggested fix

- Add `ObservedGeneration int64` to `AgentStatus` and set it from `agent.Generation` wherever the controller writes agent status, mirroring what the plugin path already does.
- Include it in the status-equality comparison so a generation bump alone triggers a status write.
[...]

PR #1031, opened 2026-08-28
ci(deploy): fail the agent redeploy when the operator ignores the tag it set

mplakhtiy:
## Summary

The agent redeploy's only change is `helm upgrade --set platformAgent.deployment.image.tag=<sha>` on the PlatformAgent CR, and the operator is free to ignore it. When it does, nothing downstream notices: with the resolved image unchanged the operator writes no new pod template, so `kubectl rollout status` returns success against a ReplicaSet that was already complete. This adds a step that reads the tag back off the gateway's pod template before the rollout gate and fails the deploy when it is not there.

## Why This Change

Autopush served an agent image nine days old while every deploy in that window reported success. It surfaced only when someone asked in chat why a fix that had merged was missing from the running agent.

The mechanism: `resolveAgentImage` reads `spec.deployment.tag` only when `spec.deployment.image` carries no tag or digest of its own, so a `kubectl patch` pinning a full reference decides the image and the deploy's tag stops meaning anything. The pin survives, which is the part worth spelling out because the chart does render `spec.deployment.image` on every upgrade — it renders the bare repository, byte-identical release to release, and Helm's patch for a custom resource is computed between the previous and new rendered manifests, so a field the same in both is absent from it and the live value is never touched. On the autopush release, `helm get manifest` showed the untagged repository while the live CR carried the pin, and `managedFields` recorded `kubectl-patch` rather than `helm` as the owner of that field across every deploy that followed.

Without this, the same thing recurs silently. The deploy pipeline has no way to notice that the field it sets has been superseded, and its only gate cannot tell a fresh ReplicaSet from an untouched one — measured at 31 seconds from `helm upgrade` to green while the cluster changed nothing.

## What Changed

- `scripts/confirm_agent_image.sh` polls the gateway Deployment until every release image in its pod template carries the requested tag, and fails with diagnostics when the budget expires. Which images count comes from `images.json` — the first-party entries on the release tag policy — matched on the repository's trailing path segment. A registry-prefix rule would be wrong here: `mirror_images.sh` writes `<prefix>/<name>` and the chart's `thirdPartyImageRegistry` falls back to `imageRegistry`, so a single-prefix mirror puts `fluent-bit` under the same prefix as the release images.
- `.github/workflows/reusable-deploy-agent.yml` runs it between the `helm upgrade` and `kubectl rollout status`. Ordering is load-bearing: past the gate the job has already reported success.
- `scripts/release/wait_for_gke_readiness.sh` calls the same script instead of its own copy. It had been polling the same Deployment before the same rollout gate on the RC and nightly path, but reading `.containers[*]` only — so both init containers were invisible — and passing on the first container whose image carried the SHA, which is exactly the sidecar skew below. Two expressions of one check, and the weaker one guarded the release.
- All release images are checked, not the agent container alone. The `platform-agent` container, the `platform-agent-dashboard` container, the `sandbox-credential-cleanup` init container, and the `envoy-credential-proxy` sidecar all move with the deploy. They normally move together because the sidecar's reference is derived from `resolveAgentImage`'s output, but not when the agent image is digest-pinned: that path falls back to `spec.deployment.tag`, so the agent freezes at its digest while the proxy beside it rolls forward on every deploy.
- On failure the message does not assert a cause it cannot know. `resolveAgentImage` reaches `spec.deployment.tag` only when `spec.deployment.image` is set and unpinned, so there are three ways to arrive here — a pinned image, an unset one, or an operator that never reconciled — and each gets its own remedy. Naming a pin unconditionally sends the reader after a CR that is fine; blaming the operator unconditionally does the reverse.
- The script fails loudly on its own misconfiguration. A non-integer budget or a missing `images.json` used to reach `exit 0` having read nothing from the cluster.
- One adjacent comment fix: the 900s gate cited `tests/test_hindsight_probes.py` for an ordering that file never mentions the gateway in. It is `tests/test_gateway_rollout_budgets.py`.

**Why the agent redeploy and the RC gate, and not the controller or integrations.** The agent is the only one of the three that deploys through a custom resource, and that indirection is the whole bug: Helm sets a field on the CR and the operator decides what the pod template says, so the tag can be ignored, and because the chart renders `image` identically every release Helm's manifest-to-manifest patch for a custom resource never disturbs a live pin. The controller sets `operator.image.tag` on a Deployment the chart renders directly — the rendered tag changes every deploy so it is always in the patch, a Deployment is a built-in type so Helm does a genuine three-way merge and reverts manual edits, and `helm --wait` watches the object that actually changed. Integrations deploy third-party images pinned by version in `values.yaml`/`images.json`, on version bumps rather than per-commit SHAs, each behind its own `rollout status` on a directly-rendered Deployment. The incident is the evidence: on the same cluster in the same window, `kube-agents-controller-manager` was on current `main` while the gateway sat nine days behind.

## Context

The autopush incident this comes from: a manual `kubectl patch` on 2026-08-27 18:54 UTC pinned `spec.deployment.image` to a full reference while debugging, and it was never unwound. Every deploy after that reported success and changed nothing. The pin has since been cleared and autopush is tracking `main` again.

Related but not closed by this: #534 (PlatformAgent status has no `observedGeneration`, so its Ready condition can describe a stale spec) is the same blind spot one layer down — the CR reported `Ready` throughout. This PR is CI-side only and does not address the operator silently ignoring `spec.deployment.tag`, which every install outside the two gke-labs environments still gets. #1009 is an adjacent image-pinning class, not the same bug.

#504 touches `reusable-deploy-agent.yml` too, removing `GITHUB_ORG`/`GITHUB_REPO` from the job env. Different hunk — a possible merge conflict, not a duplicate.

## Testing

- `tests/test_deploy_image_confirmation.py` — 24 tests, new. Three pin the deploy workflow's wiring (the script runs, it runs before the rollout gate, and its failure can still fail the job); three pin that `wait_for_gke_readiness.sh` delegates rather than open-coding the read-back; the rest drive the script through a stub `kubectl` on PATH, covering the pass case, a third-party pin on its own registry, a third-party pin mirrored under the release prefix, the incident's pinned-tag shape, sidecar skew, a digest-pinned agent, a digest pin on a registry with a port, a template with no release image, a Deployment that does not exist, the operator-status report, the healthy-CR and unset-image messages, two sequences that make the poll go round, and four that misconfigure the script itself.
- `tests/test_gateway_rollout_budgets.py`, `tests/test_hindsight_probes.py`, and `tests/test_install_pubsub_platform.py` all regex-parse the two changed shell/workflow files and still pass. Whole suite: 634 passed, 3 skipped (`tests/`, excluding `e2e/` and `integration/`; the latter needs a `uvicorn` this machine lacks).
- `shellcheck` clean on both changed shell scripts; `prettier --check` clean on the changed workflow and README; `make docs-check` clean.
- Not run: Go tests, which are SIGKILLed on this machine and are not a usable local signal; `actionlint` and `zizmor`, for the same reason. CI covers both.

### Live validation

Exercised against the live autopush install — project `kube-agents-autopush`, cluster `platform-agent-host` (us-central1), namespace `kubeagents-system`. Re-run against the current branch head after the review fixes.

**The failure this PR exists to catch was reproduced and then repaired on that install.** Before: the gateway ran `platform-agent:a1456a4b` (2026-08-19) while `spec.deployment.tag` said `f1908801`, `.status.phase` said `Ready`, and every recent deploy had reported green. Clearing the pinned `spec.deployment.image` back to the bare repository made the operator resolve `image` + `tag` and roll the gateway — verified on the Deployment, the ReplicaSet, and the running pod. Autopush has since moved on to `aefbbd6e` by itself, confirming the pipeline is tracking `main` again.

**The script was run against that cluster, both directions.** On the tag the cluster is actually running:

```
$ ./scripts/confirm_agent_image.sh kubeagents-system platform-agent-gateway aefbbd6e...
Operator applied tag aefbbd6ef2df8c469f4759410af126d47878a562 to all 3 release image(s) in platform-agent-gateway.
exit=0
```

Three, not four — `fluent-bit:5.1.0` is correctly excluded. On a tag it is not running, against a CR that is healthy:

```
::error::platform-agent-gateway is not running the tag deadbeef... this deploy set. Still on:
  sandbox-credential-cleanup: ghcr.io/gke-labs/kube-agents/platform-agent:aefbbd6e...
  envoy-credential-proxy: ghcr.io/gke-labs/kube-agents/credential-proxy:aefbbd6e...
  platform-agent: ghcr.io/gke-labs/kube-agents/platform-agent:aefbbd6e...
spec.deployment on the CR:
  platform-agent: image=ghcr.io/gke-labs/kube-agents/platform-agent tag=aefbbd6e...
operator status:
  platform-agent: Ready
    Ready=True Reconciled
No CR above pins spec.deployment.image, so a pin is not the cause here. Read the status: an operator
that is absent, crash-looping, or returning early leaves the pod template as it was.
```

That is the mechanism rather than a coincidence: the same command passes on the tag the cluster is running and fails on one it is not, and on an unpinned CR it declines to blame a pin — confirmed against a real CR rather than a fixture.

**Those two transcripts were captured at `eae219af` and have not been re-run against the head.** What they establish still holds — the pass/fail split on a real cluster, and the repair of the pin — but two details in them are now stale by design: the final line reads "the CR is not the cause here" rather than "a pin is not the cause here" (finding 15 widened it to three causes), and the budget in play is 300s rather than 120s. Re-running them needs the shared autopush install and its lease; the review round that produced findings 14–22 read that cluster without mutating it.

**Not covered.** The workflow step has not run inside a real GitHub Actions deploy, because the reusable workflow is only reachable from `main` — that first exercise happens on merge. The same is now true of the RC path: `wait_for_gke_readiness.sh` runs only from `e2e-run.yml`, so the delegation added for finding 16 is first exercised on the next nightly or RC after merge. It is covered by tests and by the script's own live transcripts above, not by a run of that pipeline. Mirrored-registry, ported-registry, and split-broker installs were tested through fixtures only; no such environment exists to point at. No test artifacts were left behind; the only cluster mutation was clearing the stale pin, which was the intended repair and the state the install should have been in all along.

## Self-Review

Three passes, all reported below as one list.

- **`review-adversarial`** — subagent that did not write the change, handed the diff range, its skill, and the `make docs-check` result, told to derive intent from the diff and not read commit bodies or notes. All ten angles run.
- **`review-docs-drift`** — a second such subagent, same handoff. Checked code→docs currency across the CRD reference, the docker-images page, the semver design, the testing map, the documentation map, `AGENTS.md`, and every generated region.
- **`/code-review`** — run afterwards on the open PR, and **it forked from the session that wrote the change**, so unlike the two above it carried the author's reasoning. Weigh it accordingly. It is the reason findings 1–3 below exist: they are all in code written *after* the two independent passes had finished.
- **`review-adversarial`, second round** — a fresh subagent against `eae219af`, handed the diff range and the skill and nothing else, run after `kube-agents-bot` reported no findings on that commit. Findings 14–22. Two Explore passes ran alongside it on questions the first round had not asked: which images the operator can actually put in the gateway pod template, and which other deploy paths carry the same blind spot.

Twenty-two findings. Nineteen fixed, one rebutted, two declined.

1. **Low (`/code-review`) — `${repository%:*}` stripped from the wrong colon on a digest reference whose registry has a port.** *Fixed.* `registry.local:5000/kube-agents/platform-agent@sha256:…` reduced to segment `registry.local`, so the image was not recognised as a release image at all: the script reported finding none instead of naming the pin, then printed a remedy for a template it never inspected. Reproduced before fixing. The segment is now taken before the tag is stripped, and there is a regression test.
2. **Low/Medium (`/code-review`) — the failure asserted a pin as the cause, and printed the clear-the-pin remedy unconditionally.** *Fixed.* An operator that is absent, crash-looping, or losing a leader lease during the sibling controller redeploy never re-renders the pod template either, and the remedy then sends the reader after a CR that is fine. The remedy is now conditional on a CR image actually carrying a tag or digest, and the headline no longer claims to know why. Confirmed against the live cluster.
3. **Low (`/code-review`) — `test_it_reports_the_operator_status_on_failure` asserted nothing.** *Fixed.* It checked for `"status:"`, a literal echoed unconditionally, so deleting the status read entirely would have left it green; the stub also answered both CR queries identically, so no phase text ever reached stdout. The stub now answers the status query separately and the test asserts on the phase and reason.
4. **Low (`/code-review`) — the guard matches the tag but never the repository**, so a pin to `ghcr.io/someone-else/platform-agent` carrying the deploy's tag would pass. *Declined, with reason.* The deploy cannot demand an exact registry without breaking mirrored installs — the case that already produced finding 5 below — and the alternative, rejecting any tag or digest on the CR's `image`, would reject digest pinning, which the operator deliberately supports and logs. An attacker-supplied repository on the deploy's own tag means someone already holds cluster write access; that is a different threat model from the incident and not one a CI gate can police.
5. **Medium CONFIRMED (`review-adversarial`) — the registry-prefix rule swept in `fluent-bit` on a single-prefix mirrored install**, reddening a healthy deploy and blaming a CR pin that does not exist. *Fixed.* The release set now comes from `images.json`. The mirroring test had been asserting the property the code lacked, because its fixture left `fluent-bit` on Docker Hub — a shape the chart does not render for a mirrored install; corrected, and it is now the regression test.
6. **Medium CONFIRMED (`review-adversarial`) — the polling loop, the script's whole reason to exist in CI, was untested.** *Fixed.* Every test resolved on the first read, leaving the deadline arithmetic, the interval, and the documented retry-on-empty-read unexercised — and that retry is the only thing between this guard and a false red on every real deploy, since the operator reconciles after the deploy returns. Two sequence tests now drive it.
7. **Low CONFIRMED (`review-adversarial`) — the failure message asserted one cause.** *Fixed*, initially by printing `.status` alongside the image fields and keeping `kubectl`'s stderr; finding 2 above completed it.
8. **Low CONFIRMED (`review-adversarial`) — the failure path dumped all of `.spec.deployment` into a public repository's Actions log.** *Fixed.* `DeploymentSpec` carries `Env` with literal values; only `image` and `tag` are read now. Nothing leaked — the live CR holds only those three keys — but the exposure was latent in the shape of the command.
9. **Blocking (`review-docs-drift`) — the claim that a hand-patched pin survives every later deploy contradicts the chart rendering that field.** *Rebutted, and the prose fixed.* The pass said it could not observe the release; the cluster could be observed, and the adversarial pass independently verified the same premise. `helm get manifest` rendered the untagged repository while the live CR carried the pin, and `managedFields` showed `kubectl-patch`, not `helm`, owning `spec.deployment.image` across every subsequent deploy — had Helm's patch carried the field, Helm would own it. The pass's cited source (`docs/designs/spec-mode-switch.md:90`) concerns fields *the chart never sets*, the opposite case. The real defect was that the header asserted persistence without saying why, which is what made a careful reader disbelieve it; it now explains the manifest-to-manifest patch.
10. **Advisory (`review-docs-drift`) — "three of the gateway's four images" pins a count in prose.** *Fixed*, subsumed by the rewrite for finding 5.
11. **Advisory (`review-docs-drift`) — the neighbouring gate comment cites `tests/test_hindsight_probes.py`, which never mentions the gateway.** *Fixed.* Verified against both files. Pre-existing, but it sits directly above the new step, and two adjacent comments citing different files for the same invariant is worse than a one-word diff.
12. **Advisory (`review-docs-drift`) — the CR name `platform-agent` is hardcoded in the diagnostics.** *Declined at the time*, on the adversarial pass's reasoning that the Deployment is `agent.Name + "-gateway"` and the workflow already hardcodes `platform-agent-gateway`. The finding is now moot regardless: the diagnostics iterate the namespace's CRs rather than naming one.
13. **Advisory (`review-docs-drift`) — two incident figures are unverifiable from the tree.** *Partly fixed.* "31 seconds" is from the deploy run log and stands. The deploy count had been stated loosely, so the number is gone rather than defended.

14. **Medium CONFIRMED — the guard could exit 0 having read nothing, on its own misconfiguration.** *Fixed.* A non-integer `AGENT_IMAGE_CONFIRM_TIMEOUT` makes `$((SECONDS + …))` fatal under `set -u`, and the `EXIT` trap's successful `rm` then supplies the shell's exit status — so the step logged an error and let the deploy through. Reproduced (`exit=0`, zero cluster reads) and fixed twice over: the budgets are validated as integers up front, and the trap carries the real status. A guard whose own misconfiguration is a silent pass is the failure it exists to stop.
15. **Medium CONFIRMED — the diagnostics recognised two causes where there are three.** *Fixed.* `resolveAgentImage` reaches `spec.deployment.tag` only when `spec.deployment.image` is set **and** unpinned (`manifest_helpers.go:272-274`), so an unset image sends the operator to its own default and past the tag entirely. That was reported as "a pin is not the cause here" — pointing the on-caller at an operator working exactly as written. Now a three-way split, each with its own remedy, plus a regression test. The header's precedence claim had the same omission and is corrected.
16. **Medium CONFIRMED — the repo already had this read-back, and this branch left the weaker copy in place.** *Fixed.* `wait_for_gke_readiness.sh:60-72` polled the same Deployment before the same rollout gate on the RC/nightly path, but matched `.containers[*]` only (so both init containers were invisible) and `grep -q ":$SHA"` passed on the *first* container that matched — meaning the sidecar skew this PR exists to catch went green there. It now calls the shared script, and a test refuses the open-coded form. `AGENTS.md` forbids a second expression of the same step; this was one.
17. **Medium CONFIRMED — the 120s budget was derived nowhere while its sibling used 300s.** *Fixed*, to 300s. The agent and controller redeploys sit in separate concurrency groups and can overlap, so the wait has to absorb an operator that is itself mid-rollout. One number for one wait, now that both callers share the script.
18. **Low CONFIRMED — a missing or malformed `images.json` exited with jq's status and raw stderr.** *Fixed.* `set -e` aborted on the bare jq assignment before the "no release images" guard beneath it could run, so the step carried exit 2 and text GitHub does not render as an annotation. Existence is checked first, the way `mirror_images.sh` already does.
19. **Low CONFIRMED — a missing Deployment was misdiagnosed as a CR problem.** *Fixed.* `buildStatefulSet` renders the gateway as a StatefulSet under custom RWO storage with multiple replicas, so the object may never arrive under this name. Folded into "found no release image", that dragged the CR diagnostics in behind it. Now a separate branch, keyed on kubectl's own NotFound.
20. **Low CONFIRMED — the `is_release_image` comment credited an invariant that does not exist.** *Fixed.* It said `mirror_images.sh` preserves a repository's trailing segment; that script's own header says the opposite in as many words — it writes `<prefix>/<name>`, replacing the segment. What makes the comparison safe is check 3c of `hack/check-image-inventory.sh`, which enforces `segment == name` for chart-rendered images using the same `grep -qxF` shape. The comment now cites it.
21. **Low CONFIRMED — nothing stopped the step being neutered later.** *Fixed.* Both wiring tests matched on the script name and compared indices, so `|| true` or `continue-on-error: true` would have left them green and the guard dead. Now asserted against.
22. **Low — two `kubectl get platformagent` calls on the failure path where one would do.** *Declined.* Merging them couples the pin-detection parse to the status output format for one saved round trip on a path that has already spent its budget. Filed as a note here rather than churned.

Also declined, restating finding 4 at a wider scope: the second round raised the same repository-blindness from two more angles — a CR pointed at a foreign repository on this deploy's tag passes, and conversely an install whose agent repository is *not* named `platform-agent` (or which sets `PLATFORM_AGENT_IMAGE`/`CREDENTIAL_PROXY_IMAGE`) would fail closed on a deploy the operator got right. Both are real. Neither is reachable from autopush or staging, which is where this runs; comparing repositories would break mirrored installs, which is the case finding 5 already settled. The header now says plainly that the check is tag-only and why, rather than implying more.

Reported but not acted on, per "report what a pass only suspects":

- The operator still ignores `spec.deployment.tag` with no status condition or webhook warning, so every install outside the two gke-labs environments keeps the original failure mode. Out of this change's reach — it needs a resolved-image field on `AgentStatus`, which does not exist. Related to #534, and the second round independently reached the same conclusion: this invariant belongs in the operator, so every caller asserts on one field instead of each parsing containers.
- **The same blind spot is still live on four paths this PR does not touch**, all verified: `upgrade.sh:456-458` sets `platformAgent.deployment.image.tag` through the CR and gates on `rollout status --timeout=120s` (`:548`) — byte-for-byte the mechanism fixed here, on the user-facing upgrade path; `terraform/examples/full-install/main.tf:437-440` waits on chart resources, which the operator-created gateway is not; `install.sh:2429-2450` downgrades a slow rollout to `SUCCESS_PENDING_ROLLOUT` and exits 0; and `hack/ci-deploy.sh:265-297`. Extending to the installer needs live validation against a real upgrade, which is a change of a different shape and blast radius — worth its own PR rather than widening this one.
- The split credential broker's own Deployment (`<name>-credential-proxy`, off by default) carries a release image nobody checks. Named in the header as an acknowledged exclusion rather than fixed.
- The gateway template can carry a `platform-agent-dashboard` container (on by default) and, in split-broker mode, `agent-api-proxy`. Harmless to the script, which counts what it finds — the header's enumeration was wrong and is corrected.

**What the passes could not cover:** Go tests, `actionlint`, `zizmor`, and any live exercise of the workflow itself. All passes read the autopush cluster only, mutating nothing.

## Risk & Rollout

CI-side only; no runtime code path, no chart template, no operator behaviour changes. The blast radius is the agent redeploy job for autopush and staging, plus — since the read-back was consolidated — the RC and nightly E2E gate, which was already running a weaker form of the same assertion. The new failure mode is that a deploy which previously went green while changing nothing now goes red, which is the point, and will fire on the next deploy into an environment carrying a manual pin. On the RC path the change is strictly a widening: every template the old loop accepted, the new one accepts unless a release image is genuinely off the tag.

The 300s budget is reconcile latency, not a rollout, and sits outside the `startupProbe < rollout gate < progressDeadlineSeconds` ordering that `tests/test_gateway_rollout_budgets.py` pins. It is the figure `wait_for_gke_readiness.sh` already used for this read-back.

Reverting the deploy-side half is deleting the step. The RC-side half reverts by restoring the twelve lines the delegation replaced; the script and its tests are inert without both callers.


bradhoekstra:

/review all

bradhoekstra (approved):

Issue #1022, opened 2026-08-28
Lane: dashboard live at a URL


jayantid:
**Goal this week: the dashboard is live at a URL and updates itself.**

Everything is built and in review: #1044 (collector), #1046 (renderer + live page), #1043 (publish hook). The visual spec and refresh design are settled (see this issue's earlier comments).

## This week
- The three PRs merge.
- The `dashboards/`-prefix IAM grant is applied and `EVAL_DASHBOARD_TARGET` is set in the Prow job (one-liner, prepped).
- The first published render is real data; the page refreshes after every run without a human.
- Nice-to-have: queue-wait gauge for #1069, catches count from the ledger convention (#1054).

**Done when:** a URL shows real data for all tiers, self-updating, freshness badge honest.

Part of #1020 (goal 1) · #897.

jayantid:
## Design + PR plan (decided 2026-08-28 — this is the lane's execution spec)

**End state:** a live dashboard showing past runs, trends, and coverage, updated within seconds of every run finishing. Architecture: static files in the prow bucket, no server — the data's heartbeat is run completions, so publish-per-run + a 60s browser poll IS real-time for this domain.

**Visual spec:** the Confident-AI-dark page reviewed 08-28 (near-black zinc surfaces, violet #8b5cf6 marks / #6e00ff brand / #c4b5fd links, mono numerals, PASSED/FAILED pills, judge score bars with 0.8 threshold tick, INFRA always distinct from fail). Shares status-color semantics with the PR-ownership page — red = gate-failing, green = clear — by agreement, not accident.

**Content policy:** every rendered field must be computable from run data by the generator. Editorial flavor lives only in an optional `case-notes.yaml` (case → one-line note + issue link); absent = row is just data. The dashboard never requires curation to refresh.

**Refresh loop:** run finishes → `ci-eval-pr.sh` final step publishes `data.json` + `index.html` to `gs://kube-agents-prow/dashboards/evals/` with `Cache-Control: no-cache` (GCS's default 1h cache would silently kill liveness) → any open tab re-fetches `data.json` every 60s and re-renders in place → freshness badge shows age and turns amber past `stale_after_s`, so staleness announces itself. Publishing is hard-guarded to never change a job's exit code.

### PRs (1-3 in flight in parallel as of 08-28)

| # | PR | Contents | Depends on |
|---|---|---|---|
[...]

PR #998, opened 2026-08-27
feat(bench): activate rca-remediation-pr; its own smoke is the validation run

jayantid:
## Summary

Activates `rca-remediation-pr` — the last spec-ready scenario without a measurement — and takes `remediation` off the domain-coverage allowlist, leaving `incident-triage` alone on it.

**This PR's own smoke run is the validation run the task was parked for.** Its cost and signal have never been measured: the 2026-08-26 run's 85-minute deadline expired 14 minutes into it (build `2092688354838581248`), so we know neither its duration nor whether its checks work end to end. Rather than validate out-of-band and then activate, activation is the experiment: if this smoke passes in budget, merge closes the loop; if it fails or blows the window, the PR simply doesn't merge and the failure is the data.

## Placement

After the six probes (proven, ~20 minutes together across two measured runs), before the audit canary — so a surprise here cannot starve the proven signal of budget. Current suite runs ~60 of 85 minutes, leaving ~25 minutes for this task to prove itself in.

## What to watch in the run

- **Duration** — the unknown this PR exists to measure. Record it in the task header on merge.
- **The write path**: this is the one active scenario that writes — the agent files a real remediation PR against the leased project's throwaway GitOps repo via `submit-suggestion`. The safeguard asserts the planted defect survived (nothing applied directly to the cluster).
- Its exact checks read the **final answer** (PR URL + planted cause), not a ledger — A4 never applied to this case.

Also fixes a cosmetic artifact from the 2026-08-26 merge resolution: a commented task entry glued onto the end of a comment paragraph in `hack/ci-eval-pr.sh`.

Validators: `19 bench case(s) OK`, task registration 58/58, domain coverage 4/4 (with the shrunk allowlist), `bash -n` clean, prettier 3.9.6 clean on all touched files.

Part of #897.

## Also: delegation-path transport failures now classify INFRA (#986)

[...]

PR #1044, opened 2026-08-28
feat(dashboard): run collector and data model (dashboard PR 1/4)

jayantid:
Dashboard PR 1 of 4: the run collector and the data model the other three build on.

## What this adds

`scripts/eval_dashboard/collect.py` scrapes `pull-kube-agents-smoke-test` Prow builds — discovered with `gsutil ls`, read with `gsutil cat`, strictly read-only — into a `data.json` of runs, per-case history and domain coverage. An offline `--from-dir` source reads the same three files (`build-log.txt`, `started.json`, `finished.json`) from local directories; that is what the unit tests use.

## The schema is a contract

Two sibling PRs (the renderer and the publisher) are being built against `data.json` **in parallel**. The exact shape — field names, types and derivation rules — is documented in `scripts/eval_dashboard/SCHEMA.md` and pinned by the unit tests. Any future change must be an **additive optional field**; anything else bumps `schema_version` and lands together with both consumers. Load-bearing rules:

- `[RESOURCE_PREPARATION_FAILED]` parses as task result `infra` and **never counts against a case**: excluded from the `pass_rate` denominator and the duration stats (it still appears in `runs_on_record` and `last3` — it is history).
- A task name with no `bench/tasks/<name>/task.yaml` on the current checkout (historical, since renamed) gets domain `"unknown"` — never a crash.
- A log the Prow deadline truncated yields a **partial run** (fewer tasks, `finished − started` fallback duration), not an exception; builds with no `finished.json` are skipped as still-running.
- `cases[].active` is true only for uncommented `TASKS` entries in `hack/ci-eval-pr.sh` (same textual parse as `scripts/test_domain_coverage.py`); `coverage` comes from `docs/designs/domains.yaml` (`domains_total` = the 11 domains, `uncovered` = the allowlist).

## Fixture provenance

`scripts/eval_dashboard/testdata/` holds three **real** builds (PRs #956 and #998), logs trimmed to the eval section (lease line, every `Task ... Result:` line, final verdict where one exists; no secrets present), with their real `started.json`/`finished.json` verbatim:

| build | why |
| --- | --- |
| `2092688354838581248` | PR #956 — deadline truncated the log, no verdict line |
| `2093030474753511424` | PR #998 — `RESOURCE_PREPARATION_FAILED` (infra) on the compliance canary |
| `2093054394793725952` | PR #998 — full 14-task graded run, pass/fail mix |

## Real-backfill proof

Run for real against the PR #939/#951/#956/#998 GCS paths (read-only; the resulting `data.json` is deliberately not committed):

```
wrote data.json: 19 runs, 18 cases, 10/11 domains covered
per PR: 939: 4 runs, 951: 2, 956: 7, 998: 6
prow results: 11 FAILURE, 6 ABORTED, 2 SUCCESS
task rows: 75 (60 pass, 14 fail, 1 infra)
```

One in-flight build (`2092983521978093568`, no `finished.json`) was skipped as designed.

## Testing

- `scripts/test_eval_dashboard_collect.py` (17 tests) sits in the `scripts/test_*.py` glob `make test-python` sweeps: exact task counts/results/durations per fixture, infra exclusion from `pass_rate`, unknown task names, truncated/garbage logs, corrupt `finished.json`, contract field names.
- Full scripts-side suite: `python3 -m unittest discover -p "test_*.py"` in `scripts/` — 620 tests, OK (including `test_task_registration.py`, 58 tests, run standalone as well).
- `SCHEMA.md` is prettier-3.9.6 clean.

Part of #1022 / #897.


jayantid:
Coordination from the renderer side (#1046), which built against this schema in parallel — five **additive** field suggestions it found itself wanting; none block either PR, all are backward-compatible for the renderer (it defaults when absent):

1. `stale_after_s` (top level) — the live page reads it to decide when the freshness badge goes amber; renderer defaults to 7200 when absent. The collector is the right author of this value.
2. `budgets: { wall_clock_s, wall_clock_p50_target_s }` (optional, top level) — makes the wall-clock tile ("96 / 240 min · p50 target 60") computable instead of actual-minutes-only.
3. `runs[].branch` — header currently renders `head <sha7>`; with a branch it can render `main@<sha7>`.
4. `coverage.previous_covered` (optional) — enables the "▲ N this week" delta chip on the coverage tile.
5. `ov_history[].pr` (optional) — chart x-labels stay resolvable after a run ages out of `runs[]`.

Suggested landing order for the set: this PR first, #1046 second, #1043 last (it merges safely any time — the hook self-skips until the siblings exist and `EVAL_DASHBOARD_TARGET` is set).

dshnayder (approved):

jayantid:
One more additive schema field, from epic #1020 goal 6 (pool-pressure reporting): `runs[].queue_wait_s` — computed as `started.json` timestamp minus the `metadata.creationTimestamp` in the same build's `prowjob.json` (both already in the GCS prefix the collector reads). The dashboard renders p50/p95 queue wait as the pool-capacity gauge with a written onboarding threshold. Optional like the others; renderer defaults to omitting the tile when absent.

jayantid:
Smoke failure is the known pre-#1063 agent-transport class, not this PR's diff: build 2093433473988562944 (started 2026-08-28T20:20Z, before #1063 merged) went RED solely on `rca-remediation-pr` rung 3 (NOT_A_REAL_RUN), and its rep logs show the signature — `opening turn failed in transport (3/3): RemoteDisconnected: Remote end closed connection without response`. Neither the base (aefbbd6) nor the head at the time carried c3ef6df (#1063); the branch now has it via a merge of main.

Issue #1020, opened 2026-08-28
Epic, week 2 (re-chartered): measure everything, block deliberately, run fast — dashboard live, 2+ cases per domain, presubmit under 3h


jayantid:
**Week of Sep 1–5.** Theme: measure everything, block deliberately, run fast.

## Goals

| # | Goal | Lane | Where it stands |
|---|---|---|---|
| 1 | Dashboard live at a URL | #1022 | PRs #1044/#1046/#1043 in review; then IAM grant + one job-env line |
| 2 | ≥2 cases per domain | #1023 | 9 new cases needed (nine domains have 1 today) |
| 3 | Presubmit under 3h, 3-rep tolerance intact | #1025 | needs adaptive repetitions — the week's key engineering |
| 4 | Release gate runs the full suite (non-gating, reported) | #1024 | = the nightly config pointed at the RC image |
| 5 | Nightly running as the catch-all | #1021 | recorder + full matrix + baselines; first in dependency order |
| 6 | Pool pressure measured | #1069 | a failing periodic is the alert |

**Dependency spine:** nightly → baselines → release gate · adaptive reps → coverage growth without blowing the clock.

## Carried from week 1: the flip to required

All job-level killers are fixed and merged (#1052, #1053, #1063). Remaining: **oss-test-infra#2676** (360m deadline) merges → ~5 clean runs → I draft the `BOOTSTRAP_ADMITTED` roster + `optional: false` PRs.

## Deferred (with triggers)

- **Admission-based blocking** — flip when the written exit criteria next to `BOOTSTRAP_ADMITTED` are met.
- **Sharding / parallel lanes** — build when the suite outgrows adaptive reps (details in #1025).

## Done when (Fri Sep 5)

Dashboard at a URL · every domain ≥2 cases · p50 ≤ 3h with adaptive reps merged · one RC ran the full suite and reported · 3 consecutive nightlies recorded · smoke required on the hand-picked roster.

Part of #897. Standing threads: #988, #994, #985, #1010, #1054 (catch ledger).

jayantid:
[...]

PR #516, opened 2026-08-04
feat(bench): add image-pull-backoff and quota-exceeded trajectory eval tasks

lapis2002:
# Pull Request

## Why This Change

Adds more complex RCA tasks as trajectory evaluations for `kube-agents` to support Layer 2 (Agentic Integrity) validation in CI/CD.

These evaluations validate the agent's tool-calling accuracy, GKE-specific troubleshooting capabilities, and answering critical questions like "Is the LLM calling the right tool for the prompt?" and "Is it diagnosing live cluster state safely without making unauthorized mutations?"

## What Changed

**Files:**
- `bench/tasks/image-pull-backoff-wrong-tag/task.yaml`
- `bench/tf/image-pull-backoff-wrong-tag/main.tf`
- `bench/tf/image-pull-backoff-wrong-tag/variables.tf`
- `bench/tasks/scale-up-error-quota-exceeded/task.yaml`
- `bench/tf/scale-up-error-quota-exceeded/main.tf`
- `bench/tf/scale-up-error-quota-exceeded/variables.tf`

**Added/Changed/Fixed:**
- **Added `image-pull-backoff-wrong-tag`**: Provisions a broken Kubernetes Deployment referencing a non-existent container image tag (`nginx:invalid-tag`). Evaluates whether the agent correctly diagnoses `ErrImagePull`/`ImagePullBackOff` and recommends updating the declarative GitOps manifest without executing live cluster mutations.
- **Added `scale-up-error-quota-exceeded`**: Provisions an autoscaling GPU node pool (`a2-ultragpu-1g`) and a pod requesting GPU resources to reliably trigger a Compute Engine GPU quota rejection (`scale.up.error.quota.exceeded`). Evaluates whether the agent diagnoses the quota bottleneck from Cloud Logging and recommends a GCP quota increase.

## Why This Matters

- Provides regression testing for complex multi-step RCA tasks
- Verifies that the agent adheres to read-only safety constraints during diagnostic investigations.

## Context
- Structured according to the out-of-tree task authoring guide introduced in [kubernetes-sigs/devops-bench#229](https://github.com/kubernetes-sigs/devops-bench/pull/229).
- Builds on the decoupled bench in #466 so task definitions can use clean relative paths instead of hardcoded absolute paths.
[...]

PR #1052, opened 2026-08-28
fix(ci): sweep cluster-scoped chart resources in teardown, regardless of Helm state

### LINKED ISSUE #1006: An orphaned cluster-scoped ValidatingAdmissionPolicy poisoned a pool project: teardown must sweep cluster-scoped chart resources regardless of Helm state
This morning `pull-kube-agents-smoke-test` on PR #998 leased `kube-agents-evals` and died at deploy, before any task ran ([build `2092983550331588608`](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/998/pull-kube-agents-smoke-test/2092983550331588608)):

```
Error: Unable to continue with install: ValidatingAdmissionPolicy "kube-agents-agent-readonly"
in namespace "" exists and cannot be imported into the current release: invalid ownership
metadata; label validation error: missing key "app.kubernetes.io/managed-by" ...
```

## What happened

#961 added the chart's first **cluster-scoped** resources: the `kube-agents-agent-readonly` ValidatingAdmissionPolicy and its binding (`charts/kube-agents/templates/agent-rbac-admission-policy.yaml`). Some run after it merged (2026-08-26 23:15Z) installed the chart on `kube-agents-evals` and its teardown failed to remove the policy — the likely shape is a run killed mid-install or mid-teardown, leaving no Helm release record for `helm uninstall` to act on while the cluster-scoped object stayed behind. (The ownership-metadata complaint in the error is a symptom of that death, not the cause: a resource left without its release bookkeeping.)

From then on, **every lease of that project failed at deploy**: Helm refuses to adopt a cluster-scoped resource it does not own, so the poisoned project reds every PR the Boskos lottery hands it, indefinitely, with a cause no PR author can see or fix. It was cleared by hand today (`kubectl delete validatingadmissionpolicy kube-agents-agent-readonly` on the project's `platform-agent-host`; verified clean afterwards).

## Why namespace deletion does not save us

The eval teardown's safety net — delete the namespaces, and everything the run created goes with them — silently stopped being complete the day the chart grew cluster-scoped resources. A VAP survives `kubeagents-system` deletion by construction. Any future aborted run (job timeout, Prow abort on a new push, pod eviction — all observed this week) re-poisons whichever project it was on.

## Proposed fix

Belt-and-braces sweep in `hack/ci-teardown.sh`, independent of Helm release state:

```bash
kubectl delete validatingadmissionpolicy,validatingadmissionpolicybinding \
  -l app.kubernetes.io/part-of=kube-agents --ignore-not-found || true
```

(with whatever label the chart actually stamps — and if the templates don't carry an identifying label on these objects yet, adding one is part of this fix). Same sweep belongs in the Boskos janitor path if it does its own cleaning.

Worth a quick audit alongside: whether #961's policy objects are the *only* cluster-scoped resources in the chart today (ClusterRoles/Bindings existed before but predate the eval pool — did teardown ever handle those, or have we been lucky?), and a checklist note for chart reviews that "cluster-scoped = teardown must know about it".

Also on the record: with the smoke becoming load-bearing for #897, this class of failure — a pool project poisoned for every subsequent PR — is the strongest argument that job-level infra failures need their own detection, separate from the per-task INFRA classification #959 added.

Refs: #961 (introduced the objects), #998 (the redded PR), #994 (the adjacent "eval-proven pool projects" discussion).

jayantid:
## Incident

A run killed mid-flight on `kube-agents-evals` left the `kube-agents-agent-readonly` ValidatingAdmissionPolicy (added by #961 — the chart's first cluster-scoped resources) orphaned with no Helm release record. `helm uninstall` had nothing to act on, namespace deletion cannot touch a cluster-scoped object by construction, and Helm refuses to adopt a resource it does not own — so every subsequent lease of that project died at deploy with `exists and cannot be imported into the current release: invalid ownership metadata`, redding PRs at random until a human deleted the object by hand. Full write-up in #1006. This PR makes `hack/ci-teardown.sh` sweep every cluster-scoped kind the chart or the operator can create, by label, unconditionally — independent of Helm release state.

## Audit: every cluster-scoped resource this project creates today

Verified two ways: `helm template` at kube-version 1.31 with every optional component enabled, and a read of the operator's apply paths (`applyManaged` stamps the common labels on everything the controller writes).

| Kind | Object(s) | Created by | Identifying label | Swept how |
| ValidatingAdmissionPolicy | `kube-agents-agent-readonly`, `kube-agents-agent-binding-scope` | chart `agent-rbac-admission-policy.yaml` (default on k8s >= 1.30) | `app.kubernetes.io/part-of: kube-agents` (already in the template) | new Step 3 label sweep |
| ValidatingAdmissionPolicyBinding | same two names | same | same | Step 3 |
| ClusterRole | `<release>-<ns>-operator-role` | chart `operator-rbac.yaml` | `kube-agents.labels` (includes `part-of`) | Step 3 |
| ClusterRoleBinding | `<release>-<ns>-operator-rolebinding` | chart `operator-rbac.yaml` | same | Step 3 |
| ClusterRole | `kubeagents:minimal:<ns>:<name>`, `kubeagents:tokenreview:<ns>:<name>` | operator (`reconcileRBAC`, credential-broker split) via `applyManaged` → `commonLabels` | `part-of=kube-agents` stamped at apply | Step 3 |
| ClusterRoleBinding | `kubeagents:minimal:<ns>:<name>`, `kubeagents:tokenreview:<ns>:<name>` | operator, via `applyManaged` | same | Step 3 |
| MutatingWebhookConfiguration / ValidatingWebhookConfiguration | `<release>-<ns>-{mutating,validating}-webhook-configuration` | chart `operator-webhooks.yaml` — `operator.webhooks.enabled=false` by default and not enabled by `ci-deploy.sh` | `kube-agents.labels` | Step 3 (defensive: the chart *can* create them, so teardown knows about them) |
| CustomResourceDefinition | `platformagents.kubeagents.x-k8s.io`, `agentplugins.kubeagents.x-k8s.io` | chart `crds/` (Helm installs these unowned **and unlabelled** — no `part-of`) | none | existing Step 2 delete-by-file, unchanged |

Notes from the audit:

- The issue names one VAP; there are **two** policies and two bindings (`kube-agents-agent-binding-scope` is the second pair). A name-based delete would have missed half of them; the label sweep gets both.
- The VAP/VAPB templates already carry `app.kubernetes.io/part-of: kube-agents` (set deliberately — the template's own comment says `part-of` is the label every documented cleanup query selects on), so no template change was needed. The label contract is documented in `docs/site/src/content/docs/reference/resource-labels.md`, which already calls cluster-scoped RBAC "the objects most easily orphaned".
- On the pre-#961 question in the issue: the chart-owned operator ClusterRole/Binding were only ever removed by `helm uninstall`, and the operator-created ones only by the CR finalizer path inside it — so yes, an aborted run could always orphan those. We have been lucky; they never poisoned a deploy because their names are release-/CR-derived and Helm re-adopts nothing it isn't asked to own... until the fixed-name VAPs arrived. The sweep now covers all of them either way.

## The sweep

One kind per `kubectl delete`, fully-qualified group names, `--ignore-not-found`, each `|| true` — so an API group missing from the cluster (VAP needs 1.30+) or one flaky delete cannot stop the sweep of the remaining kinds, and nothing in it can change the teardown's exit code. The script runs `set -uo pipefail` with no `-e`, and every step before the sweep ends in `|| true`, so the sweep is reachable on every path through the script; the only earlier exits are the two wrong-cluster guards (failed `get-credentials`, kubectl-context mismatch), which are exactly the paths on which deleting things would be wrong.

## Which abort shapes this covers

`hack/ci-teardown.sh` has no caller in this repo — the presubmit wires it up in `oss-test-infra`, which is out of reach of this PR. Precisely:

- **Covered:** every run whose teardown step starts and passes the wrong-cluster guards — including runs where the deploy or eval step failed, where the Helm release record is missing or corrupt (the #1006 shape), and where `helm uninstall` itself fails.
- **Not covered:** aborts that prevent the teardown step from starting at all — Prow abort on a new push (SIGTERM), pod eviction, a job timeout that kills the pod. Whether teardown runs on those is decided by the job wiring in `oss-test-infra`, not here.
- **Backstop for the uncovered shapes:** the Boskos janitor is *not* it — `docs/site/src/content/docs/deploy/ci-pool-projects.md` §8 records that the janitor is deliberately disabled for `kube-agents-evals-project` to preserve the long-lived `platform-agent-host` cluster, and no janitor/cleaner config lives in this repo. The practical backstop is the sweep's own unconditionality: because it runs on every teardown regardless of what this run installed, a project poisoned by a hard abort now self-heals at its **next** lease's teardown — one red run at worst, instead of red-forever-until-a-human-intervenes. Two follow-ups worth considering, not done here: mirror the sweep as a pre-flight in `ci-deploy.sh` to remove even that one red run, and carry the same sweep into the janitor path if it is ever enabled for this pool.

## Tests

New `tests/test_ci_teardown_sweep.py` (auto-discovered via `PYTHON_TEST_DIRS`), following the repo's lift-and-run convention (`tests/test_ci_gitops_repo.py`, `scripts/test_ci_eval_trap.py`): the teardown steps are lifted from the script's own text and executed under `set -uo pipefail` with recording kubectl/helm stubs. It pins: (a) every audited kind is deleted exactly once, by the `part-of` selector, with `--ignore-not-found` — and never as a bare unselectored kind, which on a live cluster would take GKE's own objects with it; (b) the sweep still runs when `helm uninstall` fails; (c) kubectl failing everywhere neither stops the sweep at the first kind nor changes the teardown's exit code.

```
$ python3 -m unittest tests.test_ci_teardown_sweep -v
...
----------------------------------------------------------------------
Ran 5 tests in 2.349s

OK

$ python3 scripts/test_task_registration.py
..........................................................
----------------------------------------------------------------------
Ran 58 tests in 1.094s

OK

$ python3 -m unittest tests.test_ci_gitops_repo        # Ran 15 tests ... OK
$ python3 scripts/test_ci_eval_trap.py                 # Ran 5 tests ... OK
$ python3 -m unittest tests.test_admission_policy_shipped  # Ran 17 tests ... OK

$ bash -n hack/ci-teardown.sh && echo BASH_N_OK
BASH_N_OK
```

`shellcheck -x hack/ci-teardown.sh` delta vs main: **empty** — the same two pre-existing findings (SC2164 on the header `cd`, SC1091 info on the sourced env file) on both sides, line numbers shifted only.

Fixes #1006. Part of #897. Prerequisite for making pull-kube-agents-smoke-test a required check (#1020).

### DIFF
--- a/hack/ci-teardown.sh
+++ b/hack/ci-teardown.sh
@@ -7,7 +7,11 @@
 #
 # One `helm uninstall` (the release owns every Kubernetes object ci-deploy.sh
 # created) plus an explicit CRD delete, since the chart leaves CRDs behind by
-# Helm's own design.
+# Helm's own design, plus an unconditional label sweep of every cluster-scoped
+# kind the chart or the operator can create — because a run killed mid-install
+# leaves cluster-scoped objects with no Helm release record for `helm
+# uninstall` to act on, and Helm then refuses to adopt them on the project's
+# next lease (#1006).
 # ==============================================================================
 
 set -uo pipefail
@@ -61,6 +65,45 @@ echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Step 2: Deleting CRDs ==="
 # not accumulate them.
 kubectl delete -f charts/kube-agents/crds/ --ignore-not-found || true
 echo "✓ CRD deletion finished in $((SECONDS - STEP_START))s"
+
+STEP_START=$SECONDS
+echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Step 3: Sweeping cluster-scoped kube-agents resources ==="
+# Belt-and-braces sweep, independent of Helm release state (#1006). A run
+# killed mid-install or mid-teardown can leave cluster-scoped objects behind
+# with no release record for Step 1's `helm uninstall` to act on — and Helm
+# then refuses to adopt them on the project's next lease ("invalid ownership
+# metadata"), failing every subsequent PR that Boskos hands this project.
+# Namespace deletion never catches these: they are cluster-scoped by
+# construction.
+#
+# The kinds below are the full audit of what the chart renders
+# (agent-rbac-admission-policy.yaml, operator-rbac.yaml, operator-webhooks.yaml)
+# and what the operator applies at reconcile time (reconcileRBAC and the
+# credential-broker TokenReview pair in platformagent_controller.go). All of
+# them carry app.kubernetes.io/part-of=kube-agents — the label contract in
+# docs/site/src/content/docs/reference/resource-labels.md — except the CRDs,
+# which Step 2 already deletes by file because Helm's crds/ convention installs
+# them unlabelled.
+#
+# One kind per call, each `|| true`: an API group missing from this cluster
+# (ValidatingAdmissionPolicy needs 1.30+) or one flaky delete must not stop
+# the sweep of the kinds that do exist, and nothing here may change the
+# teardown's exit code. This block must stay reachable on every path through
+# the script — steps before it end in `|| true` and there is no `set -e`; the
+# only early exits are the two must-not-delete-the-wrong-cluster guards above.
+SWEEP_SELECTOR="app.kubernetes.io/part-of=kube-agents"
+SWEEP_KINDS=(
+  validatingadmissionpolicies.admissionregistration.k8s.io
+  validatingadmissionpolicybindings.admissionregistration.k8s.io
+  mutatingwebhookconfigurations.admissionregistration.k8s.io
+  validatingwebhookconfigurations.admissionregistration.k8s.io
+  clusterroles.rbac.authorization.k8s.io
+  clusterrolebindings.rbac.authorization.k8s.io
+)
+for SWEEP_KIND in "${SWEEP_KINDS[@]}"; do
+  kubectl delete "${SWEEP_KIND}" -l "${SWEEP_SELECTOR}" --ignore-not-found || true
+done
+echo "✓ Cluster-scoped sweep finished in $((SECONDS - STEP_START))s"
 
 TOTAL_DURATION=$((SECONDS - START_TIME))
 echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Cleanup Complete (Total Duration: ${TOTAL_DURATION}s) ==="
--- a/tests/test_ci_teardown_sweep.py
+++ b/tests/test_ci_teardown_sweep.py
@@ -1,0 +1,169 @@
+"""The teardown's cluster-scoped sweep runs on every path, whatever Helm says.
+
+#961 gave the chart its first cluster-scoped resources (two
+ValidatingAdmissionPolicies and their bindings), and #1006 is what happened
+next: a run killed mid-flight left the policy behind with no Helm release
+record, `helm uninstall` had nothing to act on, and every later lease of the
+project died at deploy on "cannot be imported into the current release" until
+a human deleted the object. Namespace deletion cannot catch a cluster-scoped
+object by construction, so `hack/ci-teardown.sh` carries an unconditional
+label sweep — and these tests pin the three properties that make it a fix
+rather than a decoration:
+
+* it deletes every audited cluster-scoped kind by the part-of label, with
+  --ignore-not-found;
+* it still runs when the steps before it (helm uninstall, the CRD delete)
+  have failed, because the aborted-run case is the whole point;
+* a failing kubectl neither stops the sweep at the first kind nor changes
+  the teardown's exit code.
+
+The steps are lifted from the script's own text and executed under bash with
+stubbed kubectl/helm, so the assertions are against the code that ships
+rather than a copy (the same approach as tests/test_ci_gitops_repo.py and
+scripts/test_ci_eval_trap.py). The lift starts after the get-credentials and
+context guards: those two are must-not-delete-the-wrong-cluster exits, the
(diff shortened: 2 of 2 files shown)


bnaylor (commented):

**1. MEDIUM — PR body is missing the required `## Self-Review` and `### Live validation` sections.**

`.github/PULL_REQUEST_TEMPLATE.md` reserves `## Self-Review` and `## Testing → ### Live validation`, and AGENTS.md (Pull Request Hygiene) makes both required for AI-authored PRs — and instructs every reviewer to report an absent Self-Review as the first finding. The body evidences substantial checking (the audit table, unit tests, a shellcheck delta) but never says which review passes ran, in what kind of context, or what their findings and dispositions were — and there is no live-validation statement at all, not even the permitted "Not live-tested" with a reason. A defensible claim here would be that the teardown runs only inside the Prow job against `kube-agents-evals`; the pending `pull-kube-agents-smoke-test` run, once green, is live evidence worth citing. Fix is body-only; no code change needed.

**2. LOW — sibling PR #516 edits `hack/ci-teardown.sh` from a base that predates the Helm migration.**

#516 appends a "Step 5: Cleaning Up Leftover Evaluation Workloads" after a "Step 4" that invokes `./k8s-operator/scripts/teardown_03_gcp_gke_operator.sh` — context that no longer exists on main (the file now has three steps built on `helm uninstall`). Whichever merges second conflicts; #516 is the stale side and needs a rebase regardless of this PR. Its addition (namespaced devops-bench workload cleanup) is orthogonal to this sweep — no divergent copy of the same fix, merge-order friction only.

bnaylor (approved):

Issue #988, opened 2026-08-26
platform-agent endpoint returns Python-templated 502s mid-eval while the agent itself keeps working — find and fix the dying front-door component


jayantid:
**This is the root-cause track for the mid-run 502s.** #957 (flake A) and #959 fixed how a 502 is *graded* — the opening turn now retries and classifies as `KUBE_AGENTS_INFRA_FAILURE` instead of handing the judge an error page. This issue is about **why the 502s happen at all**. Related: #985 (audit latency, measured on the same run as sighting 2).

## Symptom

During `pull-kube-agents-smoke-test`, the harness (port-forward to `svc/platform-agent`, port 8642, POST `/v1/responses`) starts getting HTTP 502 mid-run after working fine earlier in the same run, on the same port-forward. The body is the **stock Python `http.server` error template** (`<p>Error code: 502</p>`, "Error code explanation: 502 - Invalid responses from another server or proxy"), i.e. a `BaseHTTPRequestHandler.send_error()` — not Envoy, not GKE ingress, not a dead tunnel.

## The component that renders the 502 (identified)

The Service does **not** target the Hermes API directly. `buildPlatformService` maps port 8642 to **targetPort 8643** (`k8s-operator/internal/controller/platformagent_manifests.go:3532`, commit `c0eff3d`). 8643 is `AgentAPIProxyHandler` — a `BaseHTTPRequestHandler` served by `ThreadingHTTPServer` **inside the credential-proxy sidecar** (`agents/platform/scripts/credential_proxy.py:100`, started at `credential_proxy.py:2578-2582`). It authenticates the external bearer and forwards to the Hermes gateway on `127.0.0.1:8642` in the `platform-agent` container (shared pod netns).

The 502 is rendered here:

- `credential_proxy.py:161-163` — upstream connection is created with a **hard-coded `timeout=300`** (the other handler's `--timeout-seconds` env knob does not apply to this path);
- `credential_proxy.py:187-190` — `except (ConnectionError, TimeoutError, OSError, HTTPException): ... self.send_error(HTTPStatus.BAD_GATEWAY)` → the exact template the harness records.

So the request path is: harness → kubectl port-forward → Service :8642 → **AgentAPIProxyHandler :8643 (credential-proxy container, renders the 502)** → Hermes gateway :8642 (`hermes gateway run`, the process that actually executes turns and spawns kanban workers as subprocesses).

[...]

PR #1053, opened 2026-08-28
fix(audit): make the ledger body file readable across the #955 uid split, and fail loudly instead of falling back to a private tmp dir

### LINKED ISSUE #1030: compliance-rbac-overgrant fails 13 of 21 Prow runs and is 26.9% of eval task time
Caught-by: evals

`compliance-rbac-overgrant` is the single most expensive and least reliable task in the Prow eval job. Across a 50-build sweep of `gs://kube-agents-prow/pr-logs/directory/pull-kube-agents-smoke-test/` (2026-08-27 15:55 → 2026-08-28 14:55 UTC), of which 26 builds reached task execution — 286 task runs, 23.5 task-hours:

- **26.9% of all task-seconds** — more than twice the next task.
- Median **930s** (15.5 min) per build.
- **Fails 13 of 21 runs.**

`gpu-stress-test-diagnosis` is second at 12.9%; the two together are ~40% of all task time.

The failure rate is the part worth attention. A task that fails ~62% of the time is not grading the pull request — it is a constant background red that every author has to learn to ignore, and it costs a quarter of the job's wall clock to produce. Two things follow, and they are separable:

1. ~~**Why does it fail?**~~ **Answered** — the ledger *write* is broken across the #955 sandbox/proxy uid split; the audit itself computes the right findings and they never reach GitHub. Root cause and evidence chain in [jayantid's comment](https://github.com/gke-labs/kube-agents/issues/1030#issuecomment-5456653330); fix in review as #1053.
2. **Why does it take 15 minutes?** Largely the same cause — the retry loop spinning and the 2700s delegation timeouts when the publish never lands. Worth re-measuring once #1053 is in, to see what the task costs when it succeeds.

Context: this came out of #1028, where the question was whether inference replay could cut the job's run time. The answer there is no, but the sweep run to check it surfaced this, which is a larger and more tractable lever than anything replay would have bought.

Raw logs and the parse are on `stalhaali.c.googlers.com:/tmp/prowsweep/`.

_Generated with the help of Claude._

jayantid:
## Root cause

Since #955 split the sandbox (uid 10000) from the credential-proxy sidecar (uid 10001), the two containers share files only through group-permission discipline (`umask 0002` + the pod fsGroup). `audit_report.py`'s `_write_temp` — the function that stages every `gh --body-file` payload for the sidecar — violated it twice: `NamedTemporaryFile` creates the body file 0600, owner-only, so the sidecar's `gh` cannot read it even on the shared PVC; and on any `OSError` touching `/opt/data/scratch` it silently fell back to the container-private temp dir, which the sidecar can *never* see. The proxy logs show the resulting shape exactly: ~11 successful `gh` reads, then the one write dying locally in ~55 ms, nine cycles in a row, after which the worker improvises around its SOP (stale ledgers, bodies blanked to empty, a mid-run issue with body `"test"`, 2700s delegation timeouts). Full evidence chain: https://github.com/gke-labs/kube-agents/issues/1030#issuecomment-5456653330

## The fix, in three parts

1. **`audit_report._write_temp`**: `os.fchmod(fd, 0o664)` after create, so the group shared across the #955 uid split can read the file; and the silent private-tmp fallback is replaced by a `RuntimeError` — `publish path broken: cannot stage a gh body file in the shared scratch directory /opt/data/scratch (uid …)` — naming the directory, the uid, the errno, and #1030. `main`'s existing catch-all turns that into one actionable `FATAL:` log line the worker can report verbatim instead of improvising. Off-cluster use keeps working via the existing `FLEET_AUDIT_SCRATCH_DIR` override.
2. **`github_scan_gate._post_body`**: same two changes, adapted to that file's posture — the raise surfaces through the per-sweep `try` in `main` as a warning the room sees, without blinding the other sweeps.
3. **`deploy/shared/docker-entrypoint.sh`**: new step 4.5 creates `$TARGET_DIR/scratch` (i.e. `/opt/data/scratch`) deterministically at container start, under the entrypoint's `umask 0002` discipline, with a `chmod 0775` that also repairs a long-lived PVC where a pre-#955 process created it too tight — instead of leaving it to whichever process gets there first with whatever umask it happens to carry.

## Why the silent fallback must die

Off-cluster it was a convenience; in-cluster it is a guarantee of failure that *presents* as a graceful degrade. The sidecar resolves `--body-file` paths in its own filesystem, so a container-private temp file can never work — the fallback converts a fixable mount/permission problem into a certain publish failure while hiding the cause, which is precisely the #1030 failure mode (one report said it outright: directory permissions prevented `finish` from applying the update, and nothing in the logs said why).

## Tests

New: `WriteTempTest` in `test_audit_report.py` (group bits `stat & 0o060` asserted; unusable scratch dir raises `publish path broken … #1030` and leaves the private temp dir empty), the equivalent pair in `test_github_scan_gate.PostBodyTest`, and an entrypoint test asserting the owner's setup creates `scratch/` group-accessible. The old `test_an_absent_volume_falls_back_rather_than_crashing` asserted the buggy behaviour and is replaced by the loud-failure test; the sweep fixture now patches `SCRATCH_DIR` at a temp dir, since the fallback was what let those tests run off-cluster.

**Mutation check**: with the `os.fchmod` line deleted, `test_the_body_file_is_group_readable_across_the_uid_split` fails with `body file is 0o600: the sidecar (a different uid since #955) can only read it through the group bits`; restored, it passes.

Suites: `test_audit_report.py` 509 tests OK, `test_github_scan_gate.py` 83 tests OK, `tests/test_docker_entrypoint.py` 58 tests OK; `make validate` and `check_prompt_assets.py` pass. (`test_credential_proxy.py` / `test_install_script.py` failures under `make test-python` reproduce identically on a clean checkout of main on this host — unrelated.)

## Impact

Canary red on ~85% of smoke runs; blocks making the smoke job required.

Fixes #1030. Part of #897.

### DIFF
--- a/agents/platform/scripts/github_scan_gate.py
+++ b/agents/platform/scripts/github_scan_gate.py
@@ -411,18 +411,34 @@ def _post_body(provider, repo: str, pr, body: str) -> None:
     in its own filesystem; `/tmp` is a per-container emptyDir, so a
     `--body-file /tmp/…` path names a file the other container cannot open. The
     refusal then fails with "no such file" — observed live before this moved.
-    `audit_report._write_temp` documents the same trap. Falls back to the system
-    temp directory when the volume is absent, so tests still run off-cluster.
+    `audit_report._write_temp` documents the same trap, and a second one: since
+    #955 the sandbox (uid 10000) and the sidecar (uid 10001) are different
+    users, so the 0600 file `NamedTemporaryFile` creates must be `fchmod`ed
+    group-readable or the sidecar cannot open it even on the shared volume.
+
+    NO fallback to the system temp directory when the volume is absent: the
+    sidecar can never see this container's private tmp, so in-cluster that
+    fallback turned a fixable mount problem into a guaranteed failure that read
+    as a graceful degrade (#1030). Raise instead — the per-sweep try in `main`
+    turns it into a warning the room sees. Tests patch SCRATCH_DIR.
     """
-    directory: str | None = SCRATCH_DIR
     try:
         Path(SCRATCH_DIR).mkdir(parents=True, exist_ok=True)
-    except OSError:
-        directory = None
-    handle = tempfile.NamedTemporaryFile(
-        "w", encoding="utf-8", suffix=".md", delete=False, dir=directory
-    )
+        handle = tempfile.NamedTemporaryFile(
+            "w", encoding="utf-8", suffix=".md", delete=False, dir=SCRATCH_DIR
+        )
+    except OSError as exc:
+        raise RuntimeError(
+            "publish path broken: cannot stage a body file in the shared "
+            f"scratch directory {SCRATCH_DIR} (uid {os.getuid()}): {exc}. "
+            "The credential sidecar resolves body-file paths in its own "
+            "filesystem, so a container-private temp file can never work — "
+            "fix the shared mount/permissions (see gke-labs/kube-agents#1030)."
+        ) from exc
     try:
+        # Group-readable across the #955 uid split; owner-only is unreadable
+        # to the sidecar that actually runs `gh`.
+        os.fchmod(handle.fileno(), 0o664)
         handle.write(body)
         handle.close()
         provider.post_comment(repo, pr, handle.name)
--- a/agents/platform/scripts/test_github_scan_gate.py
+++ b/agents/platform/scripts/test_github_scan_gate.py
@@ -569,12 +569,61 @@ class PostBodyTest(unittest.TestCase):
                 gate._post_body(recorder, "acme/toolkit", make_pr(), "x")
             self.assertFalse(Path(recorder.path).exists())
 
-    def test_an_absent_volume_falls_back_rather_than_crashing(self):
-        """Off-cluster there is no /opt/data, and the unit tests still run."""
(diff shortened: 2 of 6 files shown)


stalhaali (approved):

jayantid:
**Validation record: the canary is back.** This PR's own smoke run ([build `2093446233761779712`](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1053/pull-kube-agents-smoke-test/2093446233761779712), kube-agents-evals-5) — `compliance-rbac-overgrant` **passed 2 of 3 repetitions** with this fix aboard, against **0 clean publishes in ~13 runs** the two days prior (#1030). Job verdict GREEN — the first bench-gate GREEN on record, with the not-yet-admitted canary correctly held non-blocking.

Post-merge confirmation: Thursday night's wave went **14 of 17 runs green** across seven pool projects (including first-ever leases of evals-7/-9/-10), with zero canary-caused failures — the publish path works again everywhere. The residual 1-in-3 rep wobble is consistent with the front-door instability tracked in #988, not the publish path; it keeps the canary out of `BOOTSTRAP_ADMITTED` until it re-proves, exactly as the admission design intends.

Ledger updated: #1054 row for #1030 → fixed by this PR.

Issue #902, opened 2026-08-23
Lane: eval speed and the 20-run admission screener


jayantid:
**Lane: eval speed and screening.** Owns how long a presubmit takes and how we find out whether a case is reliable enough to be allowed to block anyone.

Waits on nothing. Both halves are independent of the fleet and of the case format.

## Why the two belong together

The gate runs every case 3 times on every pull request, and admission runs a candidate case 20 times against `main`. Both multiply whatever a single run costs. A 25-minute run at 3 reps across a growing corpus is the thing that decides whether hundreds of cases is realistic or theoretical, so the speed work is not tidying, it is what makes the strategy affordable.

## Part 1: where does the time go

Measure before changing anything, and publish the breakdown. Today a presubmit leases a project, deploys the pull request's own images (`hack/ci-deploy.sh`, `helm upgrade --install`), runs the tasks in `hack/ci-eval-pr.sh`, then tears down (`hack/ci-teardown.sh`). Nobody currently knows which of those dominates.

Then take the obvious win: **skip the image build for pull requests that cannot change agent behaviour.** A docs-only or Terraform-only change does not need a 20-minute Cloud Build. Be careful about what counts as "cannot change behaviour": half this repository is prose, and `SOUL.md`, the governance SOPs and the skills determine behaviour as surely as the Go does. A markdown change under `agents/` is a behaviour change. A markdown change under `docs/site/` is not.

## Part 2: the admission screener

A case cannot block anyone until it has proved it is reliable. The screener is the job that proves it: **run a case 20 times against `main` and report its pass rate.** At least 19 of 20 and it joins the blocking set. Below that, it runs, it reports, and it fails nobody.

Two consumers, both worth designing for:

- The scoring lane reads the number to decide whether a case is admitted.
- The author of the case reads it on their own pull request. That is the half of eval-driven development that is easy to skip: you find out your case is too noisy **before** it starts randomly reddening other people's work.

Practical notes: 20 runs of one case is cheap compared with a full suite, and it does not have to run on every pull request, only on those that add or edit a case. Results need somewhere to live that the scoring lane can read, and the reliability number is meaningless unless it is stamped with the same five versions baselines are keyed on (fleet, harness, verifiers, judge model, agent model).

## Also in this lane, if there is room

[...]

PR #1057, opened 2026-08-31
perf(eval): run the task matrix as a parallel fan-out of (task, repetition) units

haoxuw:
## Summary

Runs the eval presubmit's task matrix as a bounded parallel fan-out of (task, repetition) units instead of a serial loop. Profiling (#947's tables) showed the agent conversation is ~98% of every invocation's wall clock and each invocation is independent — at fourteen tasks × three repetitions the serial loop was 140.4 of the job's 156.8 measured minutes. The unit of parallelism is one repetition; launch order is longest-first from measured cost hints; `EVAL_TASK_PARALLELISM` caps concurrency (default 4, `1` reproduces serial behaviour through the same code path). Grading is untouched: `bench-gate case` consumes the same positional per-repetition results in the same `TASKS` order, `bench-gate suite` the same case files.

Draft on purpose: the one number this change cannot measure from a dev checkout is the pool projects' model-quota headroom, and this PR's own presubmit run is that measurement — see Risk & Rollout.

## Why This Change

The presubmit budget has been raised twice (85m → 150m → 240m) and the matrix keeps growing; the measured serial run is 1.26×–1.43× of the 240m ceiling at fourteen tasks × 3. The serial loop's time is almost entirely independent agent conversations, which is the textbook fan-out case.

Measured, same install, same matrix (16 tasks), same day:

| run | config | wall clock | notes |
| --- | --- | ---: | --- |
| A | serial (P=1), reps=1 | **3240s** (54.0m) | quota-clean baseline, new code path |
| B | P=4, reps=1 | **1022s** (17.0m, 3.17×) | 6/16 units lost to Vertex 429s on this dev project's quota |
| C | P=2, reps=1 | **2512s** (41.9m, 1.29×) | 1 unit 429'd |
| D | P=4, reps=2, all review fixes | **3706s** (61.8m vs ~6480s serial-equivalent, 1.75×) | **exit 0, green**; task/infra locks held; all artifacts present |
| Prow (pool, pre-fix head 21ed298) | P=4, reps=3, 48 units | **8963s eval** (149.4m), job 2h46m57s | **green**, zero 429s; units averaged 548s vs serial ~216s — root-caused to the shared-tunnel transport churn this PR's review round fixed (15/48 unit logs carry `unable to listen`/`RemoteDisconnected`; the worst unit spent 79% of its 1664s in two dead-transport stalls), plus 26% queue-packing loss — both fixed in 9fd8e2b, unverified on the pool until the next run |
| E | P=16, reps=1 | **1074s** (17.9m) | no faster than P=4; 7/16 units 429'd; surviving conversations slowed 5–6× (366–799s vs 85–140s serial) |

The ceiling is the model quota, not the mechanism: B's 3.17× includes cheap 429 fast-fails, D's honest 1.75× carries per-task repetition serialization and a throttled Vertex quota, and E shows the curve flat beyond it — P=16 matched P=4's wall clock while poisoning more units and slowing every surviving conversation 5–6×, so past the quota extra lanes only convert healthy units into throttled ones. On this project the knee is between 2 and 4. The pool measurement (build 2094432646640701440) is green with zero 429s, and its 2.5× per-unit slowdown root-caused NOT to the key but to this PR's own pre-fix transport bug: 15/48 unit logs carry port-bind/`RemoteDisconnected` markers, and the sampled worst unit spent 79% of its wall in dead-tunnel stalls — the review round's per-unit-port fix (9fd8e2b) removes exactly this, so more of the clean-room 3.17× should be recoverable on the pool than this run shows. The shared-key aggregate-budget concern from review stands as the design constraint (the dev project's quoted 429s prove the failure mode): pick the default from max_concurrency × P on the shared key, and record contention into the baseline evidence before the store turns on — the latter is a committed follow-up, not in this diff.

## What Changed

- **Fan-out:** a flat queue of (task, repetition) units, cost-hinted longest-first and rep-major interleaved (repetitions of one task serialize, so adjacent same-task units would park a lane), launched into `EVAL_TASK_PARALLELISM` slots with a 5s stagger to soften model-quota bursts.
- **Isolation the serial loop got for free, restructured rather than worked around:** per-task env (seeded-cluster reuse, cluster names) is precomputed and handed to each unit's subshell instead of exported globally; the new-run-directory discovery reads each unit's own log (`results:` line) instead of diffing the shared results directory, which cannot tell concurrent siblings apart.
- **Two mkdir-mutexes serialize what genuinely cannot overlap** while noop units fill the lanes: repetitions of one task (a ledger-writing audit's concurrent reps would rewrite one shared ledger issue and grade each other's artifact; the autoops task's concurrent incidents would have no card attribution; same-task reps share a tofu stack and cluster name), and stack-bearing (tofu) units across tasks (`BENCH_PARALLEL` stays false, so devops-bench's per-run isolation is off and concurrent tofu units would race the shared kubeconfig's context and their state locks).
- **Deadline behaviour:** each unit copies its log into `$ARTIFACTS` the moment it finishes, and a `TERM` trap converts Prow's deadline SIGTERM into an exit so the artifact-collection EXIT trap fires — the serial loop's property that a timeout preserves completed work, kept.
- **Output:** unit transcripts go to per-unit logs (shipped as artifacts, as before) instead of interleaving the job log; the job log carries launch/finish lines and the per-repetition phase breakdowns.
- Docs: `docs/designs/eval-scorer.md`'s presubmit-budget open item restated in fan-out arithmetic (its serial tables kept as the labelled baseline); the stale serial-ordering rationale in `TASKS` comments and the activation-cost guidance corrected.

## Context

- Follows the measurement chain: #947 (profiler), #949/#951 (infra costs removed), #948 (agent latency dominates — this PR is the schedule-level answer; per-conversation latency remains its lane). Issue #902 owns scale beyond this.
- Merge-conflict warnings, no competing work: #1043 (dashboard) and #1050/#981 (task activations) edit this file — #1050's placement prose is written against serial-truncation semantics this PR removes, and its new tasks will take the default cost hint; whichever merges second reconciles. #944 (plant hook) also touches the loop region.
- The serial measurements quoted in comments and design docs are kept, labelled as the fan-out's baseline.

## Testing

- `bash -n`; `make docs-check` clean; prettier clean on the changed design doc.
- Offline machinery tests: queue construction/ordering, semaphore cap (max-concurrent observed = cap), state-file round trip, and both mutexes (same-task and infra overlap checks against event logs) — all green on bash 3.2.
- The four live runs above; D (all fixes in) exercised: lock serialization live (zero same-task work overlap in 32 units), in-unit artifact copies (rep-2 logs present mid-run), the `results:`-line discovery (zero MISSING across A–D), and a green `bench-gate suite` end to end.

### Live validation

Install: `platform-agent-host` (GKE `us-central1`, project `haoxuw-gke-dev`), fresh Helm-managed deploy of current images, namespace `kubeagents-system`, mock seeded slot-c cluster for the gpu task's reuse path. Runs A–E as tabled above (~4h of eval runs total); run D is the run of record for the committed code, and run E (P=16) is the over-provisioning probe that located the quota knee — exit 0, 32/32 units completed, all `eval_*_rep*.log` and `results_*_rep*.json` artifacts present, both locks held throughout. **Not covered:** a run in a real pool project (fleet fixtures, real quota, Prow's SIGTERM path) — the draft's own presubmit is that run; and the rep-major queue interleave landed after run D began, so its packing benefit is asserted from the queue math, not yet measured live.

## Self-Review

`make docs-check` ran clean in the main loop; both passes ran in subagents handed only the diff range. Dispositions, merged:

- **[BLOCKER/CONFIRMED, adversarial] Concurrent tofu units share state, kubeconfig context, and cluster name (`BENCH_PARALLEL=false` isolation is off)** → fixed: the infra mutex serializes stack-bearing units across tasks; noop units (13 of 16) fill the lanes.
- **[HIGH/CONFIRMED, adversarial] Concurrent repetitions of a ledger-writing audit rewrite one shared ledger issue and can grade each other's artifact** → fixed: the per-task mutex serializes repetitions of one task; also restores rep comparability and fixes the autoops card-attribution case (MEDIUM/PLAUSIBLE sibling finding).
- **[HIGH/PLAUSIBLE, adversarial] The shared pre-opened tunnel defeats the harness's stale-tunnel recovery (it never recycles a forward it didn't spawn)** → fixed by removal: the 5s stagger already prevents the bind race the tunnel existed for, and per-harness ownership restores the recycle path. This also moots the two LOW findings on tunnel logging/silent failure.
- **[MEDIUM/CONFIRMED, adversarial] `set -e` kills a unit's state writes when its run printed no `results:` line** → fixed (`|| true` on the parse; exactly the crashed runs need their logs).
- **[MEDIUM/CONFIRMED, adversarial] All artifacts deferred behind `wait`, so a Prow deadline kill uploaded nothing** → fixed: in-unit log copies + `TERM` trap.
- **[MEDIUM/CONFIRMED, adversarial] Cost hints omitted the two plausibly-longest tasks and contradicted an in-file measurement** → fixed from the fleet measurements (gpu/autoops 900; the evidence-chain 1300 was my quota-skewed local sample against the fleet's measured 190s — removed).
- **[LOW/CONFIRMED ×2, adversarial + Blocking, docs-drift] Stale serial-ordering comments inside `TASKS` and the activation-cost guidance (~P× overstated)** → fixed; execution-order rationale now lives at `unit_cost_hint`.
- **[Blocking, docs-drift] `eval-scorer.md`'s budget story was serial arithmetic plus "work in flight" prose this PR falsifies on merge** → fixed: fan-out arithmetic added, serial tables labelled baseline, "in flight" replaced.
- **[LOW/CONFIRMED, adversarial + MEDIUM, bot review] No test reaches the scheduler shell — and my original disposition ("no test seam exists") was wrong: `scripts/test_ci_eval_trap.py` is that seam** → fixed in 9fd8e2b: `scripts/test_ci_eval_fanout.py` pins the lock deadline, mutual exclusion, queue order, and run-directory recovery against the real script text in the same lift-and-run style; already covered by the `scripts/test_*.py` glob in `PYTHON_TEST_DIRS`.
- **[HIGH, bot review] Shared port-forward torn down by its owner's exit under mid-conversation siblings** → fixed in 9fd8e2b: per-unit `AGENT_LOCAL_PORT`, one owned tunnel per unit.
- **[review, jayantid] Dead lock holder strands contenders in a silent spin past what `wait` can collect** → fixed in 9fd8e2b: acquisition deadline, loud give-up, held locks released, unit grades MISSING; the backwards comment corrected.
- **[review, jayantid] Baseline evidence and PR runs graded at different contention levels on one shared key** → default-choice guidance folded into Risk & Rollout; recording `EVAL_TASK_PARALLELISM` into the evidence schema is a committed follow-up before `EVAL_BASELINE_STORE` turns on (schema change to `BaselineRecord`/`VersionKey`, not smuggled in here).
- **[Advisory, docs-drift] `DRAFTS.md`'s "six audits do not fit" and `eval-scorer.md`'s nightly-cost split rest on serial arithmetic** → not changed: historical rationale for decisions already made; the recast may still stand on quota grounds, and re-litigating it belongs to the first pool-quota measurement, not this diff.
- **[Advisory, docs-drift] Pre-existing drift in `bench-case-format.md` (deployer-parsing claim, one-invocation claim)** → out of scope, predates this change; worth a docs sweep.
- Both passes' non-findings worth keeping: bare `wait` cannot kill the script under `set -e`; the heredoc-fed launch loop is safe from stdin-eating units; `--result` positionality, MISSING semantics, INFRA classification, and seeded-reuse selection are byte-equivalent to serial; run-directory collisions are prevented upstream by microsecond-suffixed run ids.

## Risk & Rollout

Presubmit-only shell. `EVAL_TASK_PARALLELISM=1` is the tested escape hatch and reproduces serial semantics through the same code path. The known risk is model-quota exhaustion at the default of 4: on a quota-constrained dev project that lost 6/16 units to 429s at P=4, one at P=2, and 7/16 at P=16 — with no wall-clock gain from 4 to 16, so over-provisioning lanes is all cost and no speed (each loss surfaced loudly as a graded transport failure, not a silent pass). The pool projects' quota is unmeasured from here — **this draft's own presubmit run is the measurement**, and the default should be settled from its log (429 count at P=4) before undrafting. Revert = one commit; no state outlives a run beyond the artifacts it always shipped.

### DIFF
--- a/hack/ci-eval-pr.sh
+++ b/hack/ci-eval-pr.sh
@@ -941,9 +941,13 @@ done
 # silently un-gate transcript-read checks.
 STATE_DIR="$(mktemp -d)"
 
-# mkdir is the mutex: atomic on every filesystem this runs on, and a crashed
-# holder is collected by `wait`, after which nothing contends. Two locks
-# serialize what genuinely cannot overlap while noop units fill the lanes:
+# mkdir is the mutex: atomic on every filesystem this runs on. A holder that
+# dies without releasing (an OOM-killed subshell releases nothing) would
+# otherwise strand every contender in a silent spin that `wait` can never
+# collect past, so acquisition carries a deadline: a unit that gives up fails
+# loudly and grades as MISSING, which is a diagnosis the gate already
+# reports. Two locks serialize what genuinely cannot overlap while noop
+# units fill the lanes:
 #   per task  -- repetitions of ONE task never overlap. Concurrent reps of a
 #                ledger-writing audit rewrite one shared ledger issue and
 #                grade each other's artifact; concurrent reps of the autoops
@@ -955,14 +959,36 @@ STATE_DIR="$(mktemp -d)"
 #                per-run isolation (own kubeconfig, gcloud config, tofu data
 #                dir) is off, and two concurrent tofu units would race the
 #                shared kubeconfig's current-context and their state locks.
-lock_acquire() { until mkdir "$1" 2>/dev/null; do sleep 3; done; }
+lock_acquire() { # <dir> [deadline-seconds]
+  local waited=0 limit="${2:-1800}"
+  until mkdir "$1" 2>/dev/null; do
+    sleep 3
+    waited=$((waited + 3))
+    if [ "${waited}" -ge "${limit}" ]; then
+      echo "ERROR: gave up on ${1} after ${limit}s; holder likely died without releasing" >&2
+      return 1
+    fi
+  done
+}
 lock_release() { rmdir "$1" 2>/dev/null || true; }
 
-run_one_unit() { # <task-path> <task-name> <rep> <reuse:true|empty> <has-stack:true|empty>
-  local task="$1" name="$2" rep="$3" reuse="$4" has_stack="$5"
+run_one_unit() { # <task-path> <task-name> <rep> <reuse:true|empty> <has-stack:true|empty> <seq>
+  local task="$1" name="$2" rep="$3" reuse="$4" has_stack="$5" seq="$6"
   local log="/tmp/eval_${name}_rep${rep}.log"
-  lock_acquire "${STATE_DIR}/lock-task-${name}"
-  [ -n "${has_stack}" ] && lock_acquire "${STATE_DIR}/lock-infra"
+  # A distinct local port per unit: the harness's port-forward is owned by
+  # the process that spawned it and its atexit teardown would drop a shared
+  # listener under every sibling mid-conversation. On its own port, each
+  # unit owns its own tunnel and keeps the harness's stale-tunnel recycling.
+  export AGENT_LOCAL_PORT=$((28642 + seq))
+  if ! lock_acquire "${STATE_DIR}/lock-task-${name}"; then
+    echo "<<< [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] ${name} rep ${rep} gave up on its task lock" >&2
+    return 0
+  fi
+  if [ -n "${has_stack}" ] && ! lock_acquire "${STATE_DIR}/lock-infra"; then
+    lock_release "${STATE_DIR}/lock-task-${name}"
+    echo "<<< [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] ${name} rep ${rep} gave up on the infra lock" >&2
+    return 0
+  fi
   if [ -n "${reuse}" ]; then
     export GKE_CLUSTER_NAME="${SEEDED_TASK_CLUSTER}" CLUSTER_NAME="${SEEDED_TASK_CLUSTER}"
     export TF_VAR_cluster_name="${SEEDED_TASK_CLUSTER}" GCP_LOCATION="${SEEDED_TASK_LOCATION}"
@@ -992,28 +1018,31 @@ run_one_unit() { # <task-path> <task-name> <rep> <reuse:true|empty> <has-stack:t
   echo "<<< [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] finished ${name} rep ${rep} in $(((end - start) / 1000))s"
 }
 
-# Cost-descending, then rep-ascending: repetitions of one task serialize on
-# the task lock, so adjacent same-task units would park a lane sleeping on
-# it -- rep-major interleaving keeps every lane doing work.
+# Rep-ascending FIRST, cost-descending within a rep: repetitions of one task
+# serialize on the task lock, so a same-task unit launched early just parks a
+# lane sleeping on it -- the pool run of 2026-08-31 (build 2094432646640701440)
+# spent two of four lanes that way for its first twelve minutes under the
+# cost-first ordering this replaces.
 UNIT_QUEUE="$(
   for REP in $(seq 1 "${EVAL_REPETITIONS}"); do
     i=0
     for TASK in "${TASKS[@]}"; do
-      printf '%s %s %s\n' "$(unit_cost_hint "${TASK_NAMES[i]}")" "${REP}" "$i"
+      printf '%s %s %s\n' "${REP}" "$(unit_cost_hint "${TASK_NAMES[i]}")" "$i"
       i=$((i + 1))
     done
-  done | sort -k1,1rn -k2,2n
+  done | sort -k1,1n -k2,2rn
 )"
 UNIT_TOTAL="$(printf '%s\n' "${UNIT_QUEUE}" | grep -c .)"
 
 profile_begin "task fan-out: ${UNIT_TOTAL} units, parallelism=${EVAL_TASK_PARALLELISM}"
-while read -r _COST REP IDX; do
+while read -r REP _COST IDX; do
   [ -n "${IDX:-}" ] || continue
   while [ "$(jobs -rp | wc -l | tr -d ' ')" -ge "${EVAL_TASK_PARALLELISM}" ]; do
     sleep 3
   done
   echo ">>> [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] launching ${TASK_NAMES[IDX]} rep ${REP}/${EVAL_REPETITIONS}"
-  run_one_unit "${TASKS[IDX]}" "${TASK_NAMES[IDX]}" "${REP}" "${TASK_REUSE[IDX]}" "${TASK_HAS_STACK[IDX]}" &
+  UNIT_SEQ=$((${UNIT_SEQ:-0} + 1))
+  run_one_unit "${TASKS[IDX]}" "${TASK_NAMES[IDX]}" "${REP}" "${TASK_REUSE[IDX]}" "${TASK_HAS_STACK[IDX]}" "${UNIT_SEQ}" &
   # Staggered, so N units do not open their first model call in the same
   # second -- burst 429s at the model quota are the fan-out's failure mode.
   sleep 5
--- a/hack/ci-eval-pr.sh
+++ b/hack/ci-eval-pr.sh
@@ -217,6 +217,10 @@ profile_and_dump_on_exit() {
   dump_prow_artifacts_on_failure
 }
 trap profile_and_dump_on_exit EXIT
+# A Prow deadline delivers SIGTERM, which does not run the EXIT trap on its
+# own; converting it to an exit is what lets the artifact collection above
+# fire on a deadline kill.
+trap 'exit 143' TERM INT
 
 START_TIME=$SECONDS
 echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Running PR Smoke Test Evaluation for PR #${PR_ID} in Namespace: ${TARGET_NAMESPACE} ==="
@@ -517,12 +521,10 @@ TASKS=(
   # crashloop 142s + the two incumbents, against the deadline the
   # 2026-08-26 run blew with full audits.
   #
-  # The six probes sit ahead of the rest on purpose. The loop below is
-  # sequential (one task at a time, no BENCH_PARALLEL), so the Prow deadline
-  # truncates the TAIL of this list; none of the probes has ever executed, so
-  # their cost is unmeasured and their signal is what this change exists to
-  # produce. Ordering the unmeasured work first means a timeout loses a
-  # measured repeat, not the new signal.
+  # This list is the gate's REPORTING order. Execution order is the
+  # fan-out's cost-hinted queue below (longest units first), so a Prow
+  # deadline kills whatever is still in flight rather than truncating this
+  # list's tail.
   "./tasks/reliability-pdb-probe/task.yaml"
   "./tasks/capacity-pinned-pool-probe/task.yaml"
   "./tasks/security-overgrant-probe/task.yaml"
@@ -542,8 +544,8 @@ TASKS=(
   # rca-remediation-pr -- remediation domain. Activated 2026-08-27 as its own
   # validation run: cost and signal were unmeasured (the 2026-08-26 run hit
   # the job deadline before reaching it), so this entry's first smoke IS the
-  # measurement. Placed after the six probes (proven, ~20 min together) and
-  # before the canary so a surprise here cannot starve the probes of budget.
+  # measurement. Launch priority lives in unit_cost_hint below, not in this
+  # list's position.
   # The one active task that WRITES: it files a remediation PR against the
   # leased project's throwaway GitOps repo via submit-suggestion.
   "./tasks/rca-remediation-pr/task.yaml"
@@ -559,15 +561,10 @@ TASKS=(
   # role to a kubeconfig. It is the cheapest task in this array (142s on the
   # 2026-08-25 run) and it proves the chain the probes above stand on.
   "./tasks/cluster-agent-crashloop-debug/task.yaml"
-  # Three more cluster-debugging cases in the same family, added by #982 and
-  # placed here rather than at the head of the array. The probes above go
-  # first because they are unmeasured and the sequential loop truncates the
-  # TAIL; these three are measured -- 190s, 142s and 220s on build
-  # 2092719124550520832 -- so ordering them first would protect the known at
-  # the expense of the unknown, which is backwards. What they do need is to
-  # stay AHEAD of gpu-stress-test-diagnosis below, the first of the array's two
-  # `deployer: tofu` entries, which spends minutes provisioning a cluster
-  # before it scores anything. All three are `deployer: noop`.
+  # Three more cluster-debugging cases in the same family, added by #982:
+  # measured 190s, 142s and 220s on build 2092719124550520832, all
+  # `deployer: noop`. Position here is reporting order only; execution
+  # order is unit_cost_hint's queue.
   #
   # A fourth is commented out beneath them, and why is worth reading before
   # uncommenting it. All four are read-only: no pull request, no ledger, so
@@ -735,9 +732,11 @@ export DETERMINISTIC_CORRECTNESS_FLOOR="${DETERMINISTIC_CORRECTNESS_FLOOR:-1.0}"
 
 # Repetitions per task. Three is what the collapse rule needs: a case reds the
 # job alone only by failing ALL of them. Two-of-three would fire 1.45 times per
-# pull request by chance at suite scale; three-of-three fires 0.03 times. The
-# loop is serial (BENCH_PARALLEL=false), so this multiplies wall-clock by three
-# -- how it scales past a handful of tasks is issue #902's lane, not this one.
+# pull request by chance at suite scale; three-of-three fires 0.03 times.
+# Each repetition is one unit of the parallel fan-out below, so at
+# parallelism P this multiplies wall-clock by roughly 3/P, not 3; scale past
+# that is issue #902's lane. The serial measurements kept below predate the
+# fan-out and are its baseline.
 #
 # FOURTEEN tasks at three repetitions is FORTY-TWO devops-bench invocations,
 # where the presubmit's budget was sized for two. This number is no longer an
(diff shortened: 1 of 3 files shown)


hack/ci-eval-pr.sh
```diff
@@ -879,91 +882,168 @@ ARTIFACT_DIR="${ARTIFACTS:-/tmp/artifacts}"
...
-  profile_begin "task ${TASK_NAME}: devops-bench run"
-  TASK_START=$SECONDS
-  echo ">>> [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Running Task: ${TASK_NAME} (${TASK}) x${EVAL_REPETITIONS} <<<"
+# ─── Parallel fan-out ─────────────────────────────────────────────────────────
+# The schedulable unit is one (task, repetition): every invocation is an
+# independent agent conversation, and the agent span is ~98% of its wall clock
+# (profiled 2026-08-28), so the matrix is embarrassingly parallel. The cap
+# bounds concurrent load on the one gateway, LiteLLM and the judge quota;
+# 1 reproduces serial behaviour through the same code path.
+EVAL_TASK_PARALLELISM="${EVAL_TASK_PARALLELISM:-4}"
```
jayantid, line 891: 🟡 **Not blocking the draft — but it should shape how you pick the default from the pool measurement.** Once the baseline store fills, PR runs and the evidence they are graded against will have been produced at different contention levels on one shared quota, and nothing in the comparison can tell.

The nightly recorder (oss-test-infra#2665) runs this same script solo: one job, P=4, ~4 concurrent conversations on the shared `kube-agents-gemini-api-key`. The presubmit runs at `max_concurrency: 6`, so busy hours put up to 6×P conversations on that same key. Your run E is the direct evidence this matters: past the quota knee, units 429 and the survivors slow 5–6×. The version key is five components — setupId, scoringVersion, judge, fleet, verifiers — and contention is not one of them, so once a case is admitted, collapse and rung 6 grade busy-hour PR runs against clean-context baseline evidence. Systematically, and always in the direction of redding the PR.

Dormant today (empty store, #2665 unprovisioned). Two asks before undrafting:
1. Pick the default P from the **aggregate** budget — 6 jobs × P under the shared key's knee — not from one job's speedup curve.
2. Record `EVAL_TASK_PARALLELISM` into the evidence line (or fold it into the version key), so like only ever compares with like.

hack/ci-eval-pr.sh
```diff
@@ -879,91 +882,168 @@ ARTIFACT_DIR="${ARTIFACTS:-/tmp/artifacts}"
...
+#                grade each other's artifact; concurrent reps of the autoops
+#                task plant simultaneous incidents with no card attribution;
+#                and same-task reps share a tofu stack directory and cluster
+#                name. Serial reps are also what keeps them comparable.
+#   infra     -- at most one stack-bearing (tofu) unit runs at a time,
+#                across tasks: BENCH_PARALLEL stays false, so devops-bench's
+#                per-run isolation (own kubeconfig, gcloud config, tofu data
+#                dir) is off, and two concurrent tofu units would race the
+#                shared kubeconfig's current-context and their state locks.
+lock_acquire() { until mkdir "$1" 2>/dev/null; do sleep 3; done; }
```
jayantid: 🟡 **Not merge-blocking; one line plus a comment fix.** A lock holder that dies without releasing strands every contender in a silent 3-second spin until the Prow deadline kills the job.

The comment above says "a crashed holder is collected by `wait`, after which nothing contends." It is the other way around: `wait` cannot return while a contender is still spinning in `until mkdir`, so one dead holder converts into the whole run riding to the 240m ceiling with no line in the log saying why. The realistic killer is not a script bug (the ||-true coverage between acquire and release is good) — it is the kernel: this pod requests 1Gi and a busy fan-out runs four `uv`/tofu stacks, so an OOM-killed unit subshell releases nothing.

```bash
lock_acquire() { # <dir> [deadline-seconds]
  local waited=0 limit="${2:-1800}"
  until mkdir "$1" 2>/dev/null; do
    sleep 3; waited=$((waited + 3))
    if [ "${waited}" -ge "${limit}" ]; then
      echo "ERROR: gave up on ${1} after ${limit}s; holder likely died without releasing" >&2
      return 1
    fi
  done
}
```

A unit that then fails loudly grades as MISSING — exactly the diagnosis the gate already knows how to report.

jayantid:
Measured assessment of this PR's first Prow run (build `2094432646640701440`, the only n so far — treat accordingly), against the sequential baselines:

**It delivers the operational goal:** the 16-case × 3-rep suite that twice died at the 240m deadline serially (builds 2094208…250304, 2094333…343488) finished **GREEN in 167 min** with 48/48 units complete and all artifacts shipped. That alone justifies the mechanism.

**But the realized speedup is ~1.2–1.5×, not ~÷P.** Realized parallelism was 2.95× (26,331s busy ÷ 8,920s span) — and contention gave most of it back: per-conversation time inflated ~2.4× under concurrent load vs the same cases sequential (e.g. crashloop-misleading rep at 1,664s vs 190s serial). Net normalized: 3.78 → 3.10 min per case-rep.

**The flake cost is real but currently absorbed:** transport events went 2 (sequential, 45 units) → **54** (parallel, 48 units), producing one INFRA rep (`compliance-rbac-overgrant` rep2, 4× 502). The 959/986/1063-era retries and INFRA classification kept the verdict unaffected — shared-case rep pass rate 88.6% vs 88.9% sequential, no judge-score degradation. Forward risk worth stating: once cases are *admitted* (collapsible), a concurrent-load INFRA on an admitted case's reps can red the gate — this is one bad gateway day from flipping.

Two actionable findings:

1. **Packing flaw, ~30–45 min recoverable:** the cost-sorted queue front-loads gpu+autoops, whose 6 units serialize on the task+infra mutexes while their parked siblings hold all 4 lanes — effective concurrency was **1** from 14:48 to ~15:24, and the 2,928s infra chain ran strictly serially. A better interleave (spread mutex-bound units across the schedule) recovers most of it.
2. **Consider `EVAL_TASK_PARALLELISM=2` as the default**: contention ate ~60% of P=4's theoretical gain here; the knee of the PR body's own curve is 2–4, and P=2 likely costs ~15–20 min of wall while cutting the 502 burst substantially. Either way, P=2 is the tested escape hatch if INFRA reps exceed 1/run.

Unrelated catch from the same run, routing separately: `autoops-warning-event-triage` failed 0/3 on its `triage-offers-lettered-options` phrase check (0.75 < 1.0 floor, identical all three reps) — a check-design item for that case, not a parallelism casualty.

(Assessment by agent under jayantid's direction; working evidence retained. n=1 caveat: the flake-rate and normalization numbers deserve 2–3 more parallel runs before being treated as steady state.)

hack/ci-eval-pr.sh
```diff
@@ -879,91 +882,168 @@ ARTIFACT_DIR="${ARTIFACTS:-/tmp/artifacts}"
...
+    export TF_VAR_cluster_name="${EVAL_CLUSTER_NAME}" GCP_LOCATION="${EVAL_DEFAULT_LOCATION}"
     unset TF_VAR_reuse_existing_cluster
   fi
+  export BENCH_NO_INFRA="false"
+  local start end dir
+  start="$(_now_ms)"
+  (cd "${BENCH_DIR}" && uv run devops-bench "${task}" --agent-type kubeagents 2>&1 | _ts_lines > "${log}") || true
+  end="$(_now_ms)"
+  [ -n "${has_stack}" ] && lock_release "${STATE_DIR}/lock-infra"
+  lock_release "${STATE_DIR}/lock-task-${name}"
```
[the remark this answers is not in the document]
haoxuw: Fixed in 9fd8e2b: every unit now exports its own `AGENT_LOCAL_PORT` (28642+seq), so each devops-bench process spawns, owns, and tears down only its own forward — no shared listener, no atexit teardown under a sibling, and the harness's stale-tunnel recycling stays intact per process. `scripts/test_ci_eval_fanout.py` (same commit) pins the scheduler pieces around it. Good catch on the opening-turn raise carrying no INFRA marker — that grading gap exists independently in the harness and is worth its own look.

hack/ci-eval-pr.sh
haoxuw: Fixed in 9fd8e2b with your function shape verbatim (default deadline 1800s): a unit that gives up logs the holder-likely-died line, releases whatever it already held (task lock released if the infra lock times out), writes no state, and grades as MISSING. The backwards comment about `wait` collecting crashed holders is corrected, and the deadline behaviour is pinned by scripts/test_ci_eval_fanout.py.

hack/ci-eval-pr.sh
haoxuw: Agreed on both, and this run's own log is your evidence: at P=4 in the pool the average unit ran 548s against the serial fleet average of ~216s — contention inflation is real and currently invisible to the version key. Ask 1 is folded into the PR body: the default should be chosen from the aggregate budget (max_concurrency × P on the shared key), and this run suggests the aggregate knee is at or below 24 concurrent conversations. Ask 2 (recording EVAL_TASK_PARALLELISM into the evidence line or version key) touches BaselineRecord/VersionKey and the store readers, so I have not smuggled it into this PR — I'd rather it land as its own change before EVAL_BASELINE_STORE turns on, since the store is dormant today and no evidence is being written at any contention level yet. Happy to write that follow-up if you want it keyed your way.

haoxuw:
Re the review's Medium (Self-Review premise): correct — `scripts/test_ci_eval_trap.py` is exactly the seam I claimed didn't exist, and the disposition was wrong. 9fd8e2b adds `scripts/test_ci_eval_fanout.py` in that file's lift-and-run style, pinning the lock deadline, mutual exclusion, queue order, and the run-directory recovery regex against the real script text; `scripts/test_*.py` is already in `PYTHON_TEST_DIRS`, so it gates every PR. Self-Review section updated to match.

haoxuw:
## Root-causing the pool run's 2.5× per-unit slowdown — it was mostly our own tunnel, not the key

Follow-up on the contention discussion with evidence from [build 2094432646640701440](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1057/pull-kube-agents-smoke-test/2094432646640701440) (started 2026-08-31 14:30Z, **passed after 2h46m57s**), which ran the pre-fix head `21ed298`:

**What model-quota throttling actually looks like** — from the dev-project runs (B/E in the table), where it is proven: the unit's record carries

```
litellm.RateLimitError: Vertex_ai_betaException - 429 "You exceeded your current quota"
```

and the unit dies in 11–14s. The pool run has **zero** of these across 48 units.

**What the pool run actually shows** — pulling the per-unit artifact logs: 15 of 48 units carry transport-failure markers, and the slow units' time sits in tunnel stalls, not slow model turns. The worst sample (`cluster-agent-crashloop-misleading-symptom` rep 1, 1664s vs 390s for rep 2):

```
error: unable to listen on any of the requested ports: [{8642 8643}]
   → 1011s stall → status turn failed (1/3): RemoteDisconnected
error: unable to listen on any of the requested ports: [{8642 8643}]
   → 300s stall
```

1311s of its 1664s — 79% — is two dead-transport stalls following port-bind collisions on the shared local forward. That is exactly the High finding in this review (owner's exit tears the shared listener down under siblings; losers of the re-bind race stall), already fixed in `9fd8e2b` by giving every unit its own `AGENT_LOCAL_PORT`.

**What this changes:** the run's 548s-vs-216s average was mostly self-inflicted transport churn plus 26% queue-packing loss (also fixed in `9fd8e2b`, rep-major ordering) — which is good news: more of the clean-room 3.17× should be recoverable on the pool than the contention framing suggested. The shared-key aggregate-budget concern (F1) stands as the design truth for the baseline store — the dev-project 429s prove the failure mode exists — but this particular run does not evidence the pool key throttling at P=4.

**Next presubmit run on the current head is the verification**: with per-unit ports and rep-major packing, the prediction is materially under this run's 149min eval span, with `unable to listen`/`RemoteDisconnected` markers gone from the unit logs.

| run | where | config | eval wall | quality |
| --- | --- | --- | ---: | --- |
| serial baseline | pool, 2026-08-27 | 39 units | 140.4m | green |
| fan-out `21ed298` | pool, 2026-08-31 | 48 units, P=4 | 149.4m (job 2h46m57s) | green; 15/48 units transport-marked |
| fan-out (clean A/B) | dev, 2026-08-31 | 16 units, P=4 | 17.0m vs 54.0m serial | 6 units 429'd (real quota, quoted above) |

haoxuw:
> Measured assessment of this PR's first Prow run (build `2094432646640701440`, the only n so far — treat accordingly), against the sequential baselines:
> 
> **It delivers the operational goal:** the 16-case × 3-rep suite that twice died at the 240m deadline serially (builds 2094208…250304, 2094333…343488) finished **GREEN in 167 min** with 48/48 units complete and all artifacts shipped. That alone justifies the mechanism.
> 
> **But the realized speedup is ~1.2–1.5×, not ~÷P.** Realized parallelism was 2.95× (26,331s busy ÷ 8,920s span) — and contention gave most of it back: per-conversation time inflated ~2.4× under concurrent load vs the same cases sequential (e.g. crashloop-misleading rep at 1,664s vs 190s serial). Net normalized: 3.78 → 3.10 min per case-rep.
> 
> **The flake cost is real but currently absorbed:** transport events went 2 (sequential, 45 units) → **54** (parallel, 48 units), producing one INFRA rep (`compliance-rbac-overgrant` rep2, 4× 502). The 959/986/1063-era retries and INFRA classification kept the verdict unaffected — shared-case rep pass rate 88.6% vs 88.9% sequential, no judge-score degradation. Forward risk worth stating: once cases are _admitted_ (collapsible), a concurrent-load INFRA on an admitted case's reps can red the gate — this is one bad gateway day from flipping.
> 
> Two actionable findings:
> 
> 1. **Packing flaw, ~30–45 min recoverable:** the cost-sorted queue front-loads gpu+autoops, whose 6 units serialize on the task+infra mutexes while their parked siblings hold all 4 lanes — effective concurrency was **1** from 14:48 to ~15:24, and the 2,928s infra chain ran strictly serially. A better interleave (spread mutex-bound units across the schedule) recovers most of it.
> 2. **Consider `EVAL_TASK_PARALLELISM=2` as the default**: contention ate ~60% of P=4's theoretical gain here; the knee of the PR body's own curve is 2–4, and P=2 likely costs ~15–20 min of wall while cutting the 502 burst substantially. Either way, P=2 is the tested escape hatch if INFRA reps exceed 1/run.
> 
> Unrelated catch from the same run, routing separately: `autoops-warning-event-triage` failed 0/3 on its `triage-offers-lettered-options` phrase check (0.75 < 1.0 floor, identical all three reps) — a check-design item for that case, not a parallelism casualty.
> 
> (Assessment by agent under jayantid's direction; working evidence retained. n=1 caveat: the flake-rate and normalization numbers deserve 2–3 more parallel runs before being treated as steady state.)

yea regarding "the realized speedup is ~1.2–1.5×, not ~÷P" there are several tests much longer than others, served as bottleneck. I think we should set EVAL_TASK_PARALLELISM=16, instead of 2, so that we can scale for more tests in the future

haoxuw:
all other comments addressed, please have another look

Fuxiao-Gao:
Follow-up measurement from three post-merge runs (2026-08-31): the transport-failure class this PR's Risk section budgeted (2 → 54 events, absorbed by repetitions) has a subset the repetitions cannot absorb — empty-trajectory/zero-token repetitions red jobs through rung 3, which sits above the collapse rule. Two of three runs that evening went RED on exactly that signature while the affected cases' sibling repetitions passed 2/2. Evidence and three candidate fixes: #1116.

PR #1058, opened 2026-08-31
feat(ci): run the nightly E2E matrix on its own environment and promote staging

### LINKED ISSUE #1016: e2e-nightly-matrix.yml and e2e-manual-runner.yml declare no environment, so every vars.* they read is empty
Both workflows target the RC cluster — they share `concurrency: group: rc-environment` with the RC pipeline — but neither job declares `environment: rc`. GitHub therefore resolves their `vars.*` at **repository** scope, and `gh api /repos/gke-labs/kube-agents/actions/variables` returns **no repository-level variables at all**.

Every `vars.*` in those two files is the empty string.

## It is already breaking the nightly

`e2e-nightly-matrix.yml` has failed three nights running — 33142570581, 33037281865, 32923009352 — all at `Authenticate to Google Cloud`:

```
##[error]google-github-actions/auth failed with: the GitHub Action workflow must specify
exactly one of "workload_identity_provider" or "credentials_json"!
```

because `vars.GCP_WORKLOAD_IDENTITY_PROVIDER` is empty.

## The part that will bite after the auth fix

The same emptiness makes the hardcoded fallbacks load-bearing rather than decorative:

```
e2e-nightly-matrix.yml:82   location: ${{ vars.GCP_REGION || 'us-central1' }}
e2e-nightly-matrix.yml:96   REGION="${GCP_REGION:-us-central1}"
e2e-manual-runner.yml:91    location: ${{ vars.GCP_REGION || 'us-central1' }}
e2e-manual-runner.yml:101   REGION="${GCP_REGION:-us-central1}"
```

Those fallbacks are the only region value those workflows ever see. The same is true of `vars.GCP_PROJECT_ID`, `vars.GKE_CLUSTER_NAME`, `vars.GH_ORG` and `vars.GH_REPO`, which have `|| 'test-org-kube-agent'`-style fallbacks or none at all.

Fixing the auth without also fixing this would swap one failure for another: the workflows would authenticate and then look for a cluster using whatever the fallbacks happen to say.

## Suggested resolution

Add `environment: rc` to the job in each workflow, which makes all of `vars.*` resolve the way the RC pipeline's jobs already do, and then remove the fallbacks — a fallback that silently substitutes a different cluster is worse than an empty value that fails loudly.

Worth checking at the same time whether these two should share the `rc` environment at all, given they hold the `rc-environment` concurrency lock and can therefore run against a cluster the RC pipeline is mid-way through provisioning.

Found while investigating the RC region migration for #1012.

mplakhtiy:
## Summary

The nightly E2E matrix gets a pipeline that owns its own infrastructure, and the rest of the nightly design lands with it. `nightly-pipeline.yml` resolves the newest `rc_*_validated` candidate, checks its images are still in GHCR, builds a whole environment at that commit in the `nightly` GitHub environment, runs the `nightly` matrix on it, tags the commit `staging_<ts>_<sha>` when the matrix passes, and tears the environment down. The staging tag is what `staging-redeploy-*.yml` deploy on, so a green nightly moves staging and a red one leaves it where it is.

Getting there needed the RC pipeline's deploy, E2E and teardown machinery to stop being RC-specific, which is most of the diff: three scripts renamed, two workflows parameterised on a `github_environment` input, and the RC pipeline's inline step 3 extracted into a reusable `e2e-run.yml` both pipelines call. On top of that the branch now carries the design's remaining phases rather than deferring them — the RC cron moves behind a scheduler so a quiet tick stops reporting green, `E2E_ENV` becomes `E2E_SUITE`, and the workflow names follow one convention. Nightly still ships `workflow_dispatch`-only; the cron is the one deliberate follow-up, and the reason is below.

## Why This Change

`e2e-nightly-matrix.yml` has never succeeded and could not. The job declared no `environment:`, so every `vars.*` it read resolved against the repository scope — which holds zero variables, all of them being environment-scoped — and `google-github-actions/auth` failed with `must specify exactly one of "workload_identity_provider" or "credentials_json"`. Its only scheduled run died there in 15 seconds. That is issue #1016.

Two further problems sat behind the auth one. It deployed nothing and asserted nothing about what was deployed, so it would have tested whatever happened to be on the cluster — which, since the RC pipeline destroys its environment on a passing run, is usually nothing. And it held `concurrency: rc-environment` at workflow level, contending with the three-hourly release pipeline for a cluster it never installed to.

Downstream of that, staging is manual and stale: one `staging/**` tag has ever been pushed, `staging/2026-07-23`, and that deploy failed on a rollout timeout. Nothing has promoted since.

Separately, and the reason the RC pipeline is touched here at all: **a skipped RC run and a passing one were the same green, and the skip won.** When step 1 found no new candidate it set `skip_rc=true`, every later job was skipped by its guard, and skipped jobs do not fail a run — so the run concluded `success`. Three quiet hours after a genuine failure therefore reported the pipeline healthy while the last run that did anything was still red. Skips are 23 of 68 candidates, so this is the common case rather than an edge.

## What Changed

**The nightly pipeline** — `nightly-pipeline.yml`, five jobs, `workflow_dispatch` only. Two skip decisions, deliberately separate: `skip_pipeline` when no validated candidate exists at all, and `skip_promotion` when the candidate already carries a staging tag — in which case the matrix still runs and nothing is pushed, which is what makes a re-run a no-op rather than a second tag.

**`github_environment`, required and defaultless** — on `deploy-environment.yml`, `teardown-environment.yml` and the new `e2e-run.yml`. It renders into the job's `environment:` and into its concurrency group, so `rc` still produces `rc-environment` and the RC pipeline's serialisation is unchanged. No default, because a nightly caller that omitted it would rebuild or destroy the RC environment.

**`e2e-run.yml`, extracted from the RC pipeline's inline step 3.** One `blocking_suite` that gates the run and an `optional_suites` **list** run under `continue-on-error`, driven by `run_optional_e2e_suites.sh`. Comma-separated, because `workflow_call` has no list input type; every suite runs whatever the ones before it did, and the script exits non-zero naming the failures so a tolerated red still says which suite produced it. Per-suite filters (`STOCKOUT_SCENARIOS`, `FLEET_AUDIT_STREAMS`) live in each suite's `env_vars` in `e2e_config.yaml` rather than as workflow inputs, since one input cannot carry a different value per suite in a list. It also takes **`alert_ingress_required`**, which decides whether a failed Pub/Sub alert-ingress install fails the job. It defaults false, reproducing the RC pipeline's previous behaviour, and nightly passes true — its blocking suite runs `test_stockout_investigation.py` with every scenario, so tolerating a missing adapter there would turn one clear failure into every scenario burning a 360s watch.

**The nightly staging gate is scoped to evidence.** `blocking_suite: rc` — Agent API health, the fleet-audit ledger across all streams, and stockout scenario 04 — is the subset the RC pipeline exercises every three hours and passes. The three files that have never completed a green run (`operator/agentplugins_e2e_test.py`, `gchat_agent_test.py`, and the stockout file across every scenario) run as `optional_suites: agent-plugin,gchat,stockout-full`, which between them cover every file the `nightly` suite holds. A night therefore produces the full matrix's signal without an unproven file deciding whether staging moves; the bar for promoting one into the gate is a green nightly run with it in the optional list.

**`rc-scheduler.yml`, and the RC pipeline loses its cron.** The scheduler holds `17 */3 * * *`, resolves the candidate with the pipeline's own `resolve_rc_tag.sh` — one implementation of the skip decision, not two — and dispatches `rc-release-pipeline.yml` only when there is something to test. The pipeline therefore has no quiet-tick path at all, and its last conclusion is its last real result. It dispatches with `GITHUB_TOKEN` and `actions: write` on the job: `workflow_dispatch` and `repository_dispatch` are the two events GitHub exempts from the rule that suppresses runs triggered by the default token, so no PAT is needed to start a workflow. The tag push in `nightly-pipeline.yml` keeps `RELEASE_BOT_TOKEN`, because that suppression is real for tags. There is no release cron in this repository yet, so nothing else needed the same treatment.

**Scripts lose their `rc_` prefixes** — `provision_environment.sh`, `teardown_environment.sh`, `teardown_common.sh`, and the `teardown_*` functions. **`RC_TEARDOWN_STRICT` loses its prefix too**, which the earlier revision of this branch deferred as a two-part change needing repository admin. It no longer is, and the settings half has since been done: `TEARDOWN_STRICT` is set on both `rc` and `nightly`, so `deploy-environment.yml` forwards that name alone. `teardown_common.sh` still reads the legacy name as a fallback for hand-runs against an unmigrated environment; deleting `RC_TEARDOWN_STRICT` from both settings pages is what retires it.

**The naming sweep.** `E2E_ENV` → `E2E_SUITE`, `environments:` → `suites:`, `default_environment` → `default_suite`, and suite values lose a `-e2e` suffix that was redundant on all six (`rc`, `nightly`, `audit`, `gchat`, `investigations`, `agent-plugin`). Every old spelling is still read for one release, with a deprecation warning, so nothing breaks mid-flight. `e2e-manual-runner.yml`'s `test_environment` input becomes `test_suite` — it named a suite and sat one letter from `github_environment`, which names something else entirely. `rc-deploy-environment.yml` and `rc-teardown-environment.yml` drop a prefix that stopped being true the moment they took a `github_environment` input, as do the two minter wiring tests, and every workflow display name follows `<Pipeline>: <Action>` with the `Step N` numbering dropped.

**`is_commit_already_validated` → `is_rc_candidate_commit_already_validated`**, glob anchored from `*_validated` to `rc_*_validated`. Unanchored, any tag family ending `_validated` answered for the RC skip gate.

**One tagger instead of four copies.** `tag_commit.sh` holds the banner-plus-`ensure_git_tag` body; `create_release_tag.sh`, `tag_validated_release.sh` and the new `tag_staging_promotion.sh` are wrappers, and `tag_ga_release.sh` hands it the final tag while keeping its pure-SemVer gate, its argument-swap handling and its stamped release commit.

**`staging-redeploy-*.yml` trigger on `staging_*`** rather than `staging/**`, matching the `rc_*` family the tag is derived from. The one historical `staging/2026-07-23` tag never re-triggers, so nothing needs migrating.

**The `GCP_SA_KEY` credential path is removed repo-wide.** The secret is set at no scope — not repository, not `rc`, not `nightly` — so all six references resolved to an empty string and Workload Identity Federation has always been the only auth path. Gone from `teardown-environment.yml` (the `secrets:` block and the `credentials_json` line), both pipelines' forwards, and the `|| ''` fallback in `e2e-manual-runner.yml` and `e2e-gchat-test.yml`. No behaviour change; one fewer credential path to keep in step.

**`e2e-nightly-matrix.yml` is deleted.** The pipeline is what its cron was supposed to do, and `e2e-manual-runner.yml` — which now takes the same `github_environment` input, fixing the other half of #1016 — covers dispatching the matrix by hand with a superset of its inputs.

**Two GitOps-repository fixes that the nightly matrix forced.** The `nightly` suite stopped pinning `test-org-kube-agent`, and `e2e-manual-runner.yml` stopped defaulting `GITHUB_ORG`/`GITHUB_REPO` to `vars.GH_ORG`/`vars.GH_REPO`: now that it binds an environment, that pair names the release repository, and handing it to a suite that mints a live GitHub App token would scope that token here.

## Context

- Implements [gke-agentic/kube-agents#166](https://github.com/gke-agentic/kube-agents/pull/166), the design, in full apart from the one item below.
- Closes #1016 (`e2e-nightly-matrix.yml` and `e2e-manual-runner.yml` declare no environment).
- **Phase 2 needed no code, and what is left of it is outside this diff.** The `kube-agents-nightly` GCP project, the `nightly` GitHub environment, its 23 variables and its `GH_APP_ID` and `GEMINI_API_KEY` secrets all exist. What this branch carries for it is documentation, since several places asserted the project and environment were still missing.

  **Two prerequisites remain, and both sit under the matrix's blocking half**, so the first dispatch fails on a fleet-audit or a Chat assertion rather than on anything it was meant to grade. Neither is a diff.

  1. **The token minter on `gke-agentic/kube-agents-nightly-infra`.** The repository now exists; the GitHub App still has to be installed on it and its private key imported into the nightly project's KMS key. `test_github_token_minting_and_connectivity` is blocking here where the RC gate tolerates it.
  2. **`E2E_CHAT_REFRESH_TOKEN` on `nightly`.** The environment overrides three of the four `E2E_CHAT_*` secrets and inherits the fourth from repository scope, so an August client id and secret pair with a July refresh token that was not issued against them. A refresh token does not exchange across clients. `gchat_agent_test.py` is in the blocking suite.
- **Deviation from the design, argued rather than skipped.** §6 asks for the tagger consolidation to land "as its own step rather than riding along inside the pipeline change", and it is inside this one. The reason is dependency rather than convenience: the nightly pipeline needs `tag_staging_promotion.sh`, §6 asks for it to be a thin wrapper over a shared tagger, and that shared tagger does not exist until the consolidation happens. Splitting it would have produced a stacked PR whose only consumer is this branch. `tag_ga_release.sh` is also not a thin wrapper — it keeps its SemVer validation, its swapped-argument handling and `create_stamped_release_commit` — because reducing it further meant moving stamping logic into a script three callers share, on the one rung where a mistake cannot be undone by deleting a tag.
- **One place the design and the implementation read differently, and the design is right in effect.** §3.4's table says nightly is not exposed to the skip-green problem "by construction", because §3.1 runs it every night whatever the candidate. The pipeline does have a `skip_pipeline` path, so read literally the row is inaccurate — but `resolve_promotion_candidate.sh` selects with `get_latest_validated_rc_tag`, which has no recency window, so the path fires only when the repository holds zero `rc_*_validated` tags. There are 45. §3.1's actual requirement — that the matrix runs every night so `operator/agentplugins_e2e_test.py` and `gchat_agent_test.py` keep their only scheduled coverage — holds. Recorded because Phase 5's scope was derived from that row, not because anything needs changing.
- **The cron is the one deliberate follow-up**, one line, once the pipeline has been dispatched by hand. That is the design's own Phase 4 decision and it still holds: a schedule cannot be tested from a branch, so the first real run is necessarily after merge, and a dispatch-only workflow cannot damage anything in the meantime. Turning it on deserves to be its own reviewable change rather than a line buried in this one.

## Testing

- `pytest tests/ --ignore=tests/e2e --ignore=tests/memory --ignore=tests/integration` — 662 passed, 581 subtests passed (610/574 before this work). `tests/integration/` fails on this machine with `session_kv_server exited early`; the same four tests fail identically on this branch's parent commit with the tree otherwise untouched, so it is a local sandbox limit, not this branch. CI covers them.
- `make docs-check` — clean, including the context budget.
- `prettier --check` (3.9.6, the version `prettier.yml` pins) on every changed Markdown, YAML and JSON file — clean.
- **`actionlint` could not be run here** and is not claimed. It exits 0 with no output on this machine even for `--version`, so a clean result from it would mean nothing. The changed workflows were parsed as YAML instead, and every input each caller passes to `e2e-run.yml` was checked against the inputs that workflow declares. CI's `actionlint.yml` is the real check.
- New tests: `tests/test_tag_commit.py`, `tests/test_resolve_promotion_candidate.py`, `tests/test_nightly_pipeline_wiring.py`, `tests/test_run_optional_e2e_suites.py` (9), `tests/test_record_nightly_candidate_summary.py` (7), `tests/test_dispatch_rc_pipeline.py` (8), and `tests/test_rc_scheduler_wiring.py` — the last pins that the cron sits on exactly one workflow, that the pipeline has none, that the scheduler binds `rc` and reuses `resolve_rc_tag.sh` rather than reimplementing the skip decision, that the dispatch uses the default token and the job carries the `actions: write` that token needs, and that every input it sends is one the pipeline accepts. It now reads across a script boundary rather than grepping the workflow for `gh workflow run`: the workflow owns when the dispatch runs and with which token, `dispatch_rc_pipeline.sh` owns what it sends, and the pipeline's declared inputs still have to accept the result. The three summary-and-dispatch blocks that moved out of the workflows have unit tests of their own, covering the annotation a failed dispatch raises, the refusal to call `gh` on a missing input, and the two nightly skips rendering differently.
- Extended: `tests/test_install_pubsub_platform.py` now pins that ingress tolerance follows `alert_ingress_required` and that the input is true exactly for a caller whose blocking suite reads an alert, derived from `e2e_config.yaml` rather than hardcoded per call site. `tests/test_provision_environment.py` covers each half of the `TEARDOWN_STRICT` rename separately, including the new name winning over a stale legacy one. `tests/test_execute_e2e_tests.py` covers the `E2E_ENV` alias, asserts no suite name still carries the suffix, and covers `canonical_suite_name` on both paths the legacy suffix reaches — the selector and a pre-rename config's own suite name.

### Live validation

**Not live-tested.** The blocker is narrower than it was — the `nightly` GCP project and GitHub environment now exist, so that half is no longer missing — but a `workflow_dispatch` button does not exist until the file is on the default branch, and a `schedule:` cannot fire from a branch either. There is no way to run either pipeline from here.

What that leaves unproven, stated plainly rather than implied away:

- **The whole nightly pipeline.** Never executed. The first run will be a manual dispatch after merge, and it is expected to be the thing that finds the remaining configuration gaps.
- **The RC pipeline changes, which are now larger than they were.** Steps 1–5 are rewired and, new in this revision, **the cron has moved off the pipeline onto `rc-scheduler.yml`**. If the scheduler fails to dispatch, the RC pipeline simply stops running — and it stops quietly, because a scheduler that dispatched nothing looks the same as one that decided not to. `tests/test_rc_scheduler_wiring.py` pins the mechanical parts, but the first scheduled tick after merge is the real test. Watch the scheduler's first run, not the pipeline's absence.
- **`github_environment: "rc"` reproducing the previous behaviour.** Argued from the diff and checked line by line against the inline step 3 it replaced, not observed.
- **The staging trigger.** `staging_*` versus `staging/**` is not observable until a promotion pushes a tag.
- **`TEARDOWN_STRICT`** is unit-tested against a stubbed `uninstall.sh`, not against a real failed teardown. The variable is now set on both environments and the workflow forwards only that name, so the first real teardown failure is what confirms the forward reaches the parser.
- **The `optional_suites` list.** `run_optional_e2e_suites.sh` is unit-tested against a mock runner — every suite runs after a failure, the exit status still reports it, the summary names which — but no pipeline run has driven it against real suites.

**Two things only a live run settles, and they are the two worth watching:**

1. **The scheduler dispatching the pipeline.** Dispatch `rc-scheduler.yml` by hand immediately after merge rather than waiting for a tick. The dispatch now uses `GITHUB_TOKEN` with `actions: write`, which is the documented path and needs no secret, but no run has exercised it — and if it fails, the RC pipeline stops running until someone notices the red scheduler.
2. **The two nightly prerequisites in **Context** above.** The minter App with its KMS key, and `E2E_CHAT_REFRESH_TOKEN`. Both fail inside the blocking suite rather than at provisioning, so the first dispatch spends a cluster before reporting them.

No test artifacts created, nothing to clean up: nothing here touched a cluster, a registry or a tag.

## Self-Review

**Context each pass ran in.** The adversarial and docs-drift passes ran in a **subagent that did not write the change**, handed the diff range and the design document and nothing else — not the plan, not the reasoning. It reviewed commit `30559d07`. Findings below are its, with dispositions.

**A third pass, in a third subagent, reviewed the whole diff again** at `9b5c01b7`, handed the diff range and the design and nothing else. It raised two HIGH, four MEDIUM and two LOW; all eight are dispositioned below, folded into the lists rather than appended. It verified against the live repository rather than reasoning from the diff — resolving both GitOps repositories over the API, reading every environment's variables and secrets with their creation timestamps, and checking the `GITHUB_TOKEN` claim against GitHub's own documentation. Two of its HIGHs are environment gaps this diff cannot fix and **Context** now names; the rest are fixed here.

**A second pass, in a second subagent, reviewed the commits that fix them** — the delta carrying most of Phases 3, 5 and 6 — again handed only the diff and the design. It raised one HIGH, four MEDIUM and two LOW, all now addressed and listed below alongside the first pass's. It verified rather than assumed where it could: it exercised `teardown_strict_source` directly against bash 3.2 with an empty-but-defined `TEARDOWN_STRICT` to confirm the fallback still reaches the legacy name, and it checked the live environment's variables and secrets against what the docs claim.

Between the two passes, one defect was caught by a test rather than a reviewer: the `<Pipeline>: <Action>` rename produced `name: RC: Release Pipeline`, which is invalid YAML and would have failed every workflow parse. The new scheduler test caught it immediately.

**A fourth round came from a human reviewer** (@eLeontev) rather than a self-review pass: twelve threads, dispositioned in the threads themselves rather than duplicated here. Two changed the shape of the change and belong in this section because they invalidate what it previously claimed:

- **The nightly gate was scoped to suites that had never passed.** Raised as a blocker, and correct. `blocking_suite: nightly` put five test files in front of the staging promotion when three of them had no green run behind them, so the first red night would not have distinguished a bad candidate from a bad suite. The gate is now the RC-proven subset with the rest as tolerated coverage — see **What Changed**. Nothing on the RC side became mandatory in this change either before or after; that split is unchanged from `main`.
- **A credential path that was dead everywhere.** `GCP_SA_KEY` was flagged on one line of the nightly pipeline; checking it showed the secret is set at no scope at all, so all six references across the repository were resolving to an empty string. Removed as a set rather than one line, since a fallback nothing populates is worse than no fallback.

The other ten were fallback removal, three inline shell blocks moving into scripts with tests, the completed `TEARDOWN_STRICT` rename, readable stockout scenario slugs, the `optional_suites` list, and the default suite. All are fixed at `55218b9e`; two are questions about scope (whether the manual runner should still target `rc`, and the blocker's disposition) that are answered in their threads and left open for the reviewer rather than resolved unilaterally.

Angles the pass ran, and the disposition of each finding:

- **Caller-specific behaviour hidden in a shared workflow.** *Found and fixed.* `e2e-run.yml` hardcoded `continue-on-error: true` on the alert-ingress install with a comment asserting the ingress feeds "the optional suite and nothing else" — true for the RC caller, false for nightly, whose blocking suite reads the alerts. Now an input, defaulting to the RC behaviour, with the comment and the test docstring corrected to match.
- **Configuration that cannot take effect.** *Found and fixed.* The `nightly` suite set a `CHAT_TOPIC_NAME` that was always overridden, because `e2e-run.yml` exports the variable unconditionally, an Actions `env:` key is defined even when its expression is empty, and `execute_e2e_tests.py` merges `**os.environ` after the config block. Removed, with the mechanism recorded in its place. Checked against the live environment afterwards: `nightly` does set the variable, so this could not have fired today — it was dead config that read as a fallback, not a live bug.
- **Claims in prose the diff does not support.** *Found and fixed.* The pipeline header said every job binds `nightly` and takes the environment lock; two of five do neither. The "one acquisition per run" comment named the wrong concurrency group — the environment lock is taken three times, with gaps. Both rewritten to describe the file.
- **Setup documentation that omits what the code reads.** *Found and fixed.* The nightly-environment checklist presented itself as "the ones the pipeline reads directly" while omitting `CHAT_TOPIC_NAME`, `GH_ORG`, `GH_REPO`, `GITOPS_ORG` and `GITOPS_REPO`, plus the `GH_APP_ID` secret.
- **Scope creep against the design.** *Reported, and answered rather than fixed.* §6 asked the tagger consolidation to land separately; the reason it did not is now in **Context** above. The reviewer's own note was that the silence was the problem, not the choice.
- **A design claim the implementation falsifies.** *Found and recorded.* §3.4 says nightly has no skip path "by construction"; it has one. In Context above.
- **Silent-skip chains.** *Found and fixed* in the earlier revision. The promotion job was first gated on the **job**, and step 5 `needs` it — so every night whose candidate was already promoted would have skipped teardown and left a GKE cluster billing under a green run. The condition moved onto the steps and `tests/test_nightly_pipeline_wiring.py` pins it.
- **Reusable-workflow input plumbing.** *Found and fixed* in the earlier revision. `e2e-run.yml` carried both `workflow_dispatch` and `workflow_call` with three inputs declared only under the latter, so a dispatch run would have rendered an empty `timeout-minutes:`. Dropped `workflow_dispatch`.
- **Argument handling in the new bash.** *Found and fixed* in the earlier revision. `tag_commit.sh`'s `--title`/`--detail` used a bare `shift 2`, which aborts under `set -e` with bash's own message and no mention of the flag.
- **Anything that could push a wrong tag.** *Checked, guards added.* `staging_tag_for_rc` refuses anything outside the `rc_` family; `tag_staging_promotion.sh` re-derives the tag, refuses a mismatched explicit one, and re-checks the prefix before the push; `resolve_promotion_candidate.sh` refuses a hand-passed tag whose commit carries no `rc_*_validated` marker. Each has a test.
- **Environment-secret resolution.** *Checked, no change.* Both `uses:` calls needing environment secrets pass `secrets: inherit` with `# zizmor: ignore[secrets-inherit]`, matching what #1029 established and for the same reason.
- **Renames leaving dangling references.** *Checked, swept twice* — once for the script renames and again for the workflow and suite renames. `make docs-check` and the test suite are the mechanical backstop.
Second pass, on the scheduler and the naming sweep:

- **Docs claiming an external prerequisite is satisfied when it is not.** *Found twice, and the second finding superseded the first.* The second pass caught the branch asserting nothing outside the repository was missing when `nightly` lacked `GH_APP_ID`. The third pass caught the correction going stale in turn: the secret was created nine minutes after the commit that documented its absence, so the branch then named a satisfied prerequisite as the single outstanding one — which is worse, because it reads as a closed checklist and hides the two gaps that are real. `testing-strategy.md`, `scripts/release/README.md` and **Context** above now name the minter App and the Chat refresh token instead, and the whole variable-and-secret comparison was re-run rather than the one line edited.
- **A credential assumption nothing in the repository establishes.** *Found, then resolved rather than mitigated.* The second pass reported that `RELEASE_BOT_TOKEN` had become the only thing starting the RC pipeline while its ability to dispatch could not be checked from source; the mitigation was an `::error` annotation. The third pass found the premise underneath it false: the comment, the test docstring and the body all said a `workflow_dispatch` made with `GITHUB_TOKEN` "returns 204 and starts nothing", and GitHub documents `workflow_dispatch` and `repository_dispatch` as the two events exempt from the suppression rule that claim generalises from. The unverified credential was buying protection from a failure mode that does not exist. The dispatch now uses `github.token` with `actions: write` on the job, and the annotation names the 403 and 404 cases that can actually occur. The tag push keeps the PAT, where the suppression is real.
- **A canonical comment describing the constant it replaced.** *Found and fixed.* `install_pubsub_platform.sh`'s FAILURE POLICY header is where the workflow comment and the README both defer to, and it still said the step carries `continue-on-error: true` while calling the alert-blocking caller hypothetical — a caller that shipped in the same diff.
- **Docs the rename swept past.** *Found and fixed.* `tests/e2e/README.md` still gave the RC pipeline a schedule of its own, and `docs/README.md`'s row for `testing-strategy.md` still summarised nightly as having no project. `make docs-check` cannot catch either: it enforces presence and alignment, not summary accuracy.
- **Superseded comments left stacked rather than deleted.** *Found and fixed.* Two blocks still argued the `RC_` prefix was permanent, a few lines above the ones saying it is not — which would have stopped the next maintainer doing the one thing that lets the fallback be deleted.
- **A fallback credited to code that cannot provide it.** *Found and fixed.* The `CHAT_TOPIC_NAME` deletion comment named `gchat_agent_test.py`'s default as the backstop; that default only fires when the key is absent, and the workflow always defines it.

Third pass, on the whole diff at `9b5c01b7`:

- **An environment variable naming a resource that does not exist.** *Found; the repository has since been created, the rest is outstanding.* `nightly`'s `GITOPS_REPO` named `gke-agentic/kube-agents-nightly-infra`, which 404'd on a token that resolved its `rc` sibling. That pair scopes the token minter and the blocking fleet-audit test, so the first dispatch would have provisioned a minter against a repository that did not exist and failed forty minutes later on an assertion that names neither. The repository now exists; installing the App and importing its key are in **Context**.
- **A secret cascade that resolves but does not work.** *Found; the fix is a settings change, named in **Context**.* `nightly` overrides three of the four `E2E_CHAT_*` secrets and inherits the refresh token from repository scope, and the branch's own checklist argued that cascade was fine. It is fine only for an environment that overrides none of them — a refresh token does not exchange against a client id and secret it was not issued for. `scripts/release/README.md` now says that, and says it as a rule about overriding as a set rather than as a note about `nightly`.
- **Compatibility claimed more broadly than it was implemented.** *Found and fixed.* The body said every old spelling still reads for one release. The variable, the flag and the config keys were aliased; the suite **values** were not, so `E2E_ENV=rc-e2e` — what `.env.example` recommended until this change — exited 1 as an unknown suite. Worse, a pre-rename config names the chat suite `gchat-e2e`, and that name reaches `_CHAT_ONLY_SUITE` directly rather than through the selector, so the one suite that deliberately needs no cluster demanded `GCP_PROJECT_ID`. `canonical_suite_name` now strips the suffix on both paths, with a deprecation warning on the selector, and four tests cover it.
- **A comment justifying a binding by a variable nobody sets.** *Found and fixed.* `nightly-pipeline.yml` said step 1 binds `nightly` "for one variable: REGISTRY_PREFIX". No environment defines it and the repository scope holds no variables at all, so it arrives empty and falls back to `DEFAULT_REGISTRY_PREFIX`. The binding is still required — `GH_ORG` and `GH_REPO` have no fallback — so the comment now says that, and the README marks `REGISTRY_PREFIX` optional instead of listing it among the values a new environment must set.
- **An environment diverging from the one it claims to mirror.** *Found; the variable is being aligned.* `nightly` set `MEMORY_PROVIDER=hindsight` where `rc` sets `multiuser_memory`, so the nightly install would bring up a Postgres-backed memory stack the release gate never validated the candidate against — against three separate statements that the environments mirror. Being corrected in the environment rather than in the diff, so the claim becomes true rather than qualified.
- **The word the rename was removing, left in the file that removes it.** *Found and fixed.* `.env.example` renamed the variable and its value and still called a suite an environment in the comment above them.

- **Two open questions the earlier revision flagged, now settled.** The `inputs` context *is* available to workflow-level `concurrency` on a `workflow_dispatch` (it is in GitHub's context-availability table alongside `github` and `vars`), and `github_environment: "rc"` does reproduce the previous `environment: rc` and `concurrency: rc-environment` at all three call sites.

## Risk & Rollout

The nightly pipeline is inert: dispatch-only, guarded by `if: github.repository == 'gke-labs/kube-agents'`, and nothing schedules it. Merging it cannot start anything.

**The risk is the RC cron move**, which is larger than anything else here. Before, the pipeline was self-triggering; now a scheduler has to dispatch it, and the failure mode is quiet — if `rc-scheduler.yml` errors or resolves wrongly, no pipeline runs and there is no red pipeline to notice. Mitigations are that the scheduler itself goes red when `resolve_rc_tag.sh` fails or the dispatch call does, that it reuses the pipeline's own resolver rather than a second copy, and that its decision is recorded in the job summary either way. The dispatch no longer depends on a PAT whose scope cannot be read from this repository — `actions: write` is declared in the file — which removes the largest unknown this section previously carried. Watch its first scheduled tick after merge.

Reverting is a `git revert` of the branch's commits; it restores the cron on the pipeline, the inline step 3 and the hardcoded `environment: rc` together, and no state outside the repository has to be undone. Reverting only the scheduler commit is also safe as long as the cron goes back on `rc-release-pipeline.yml` in the same change.

Two smaller externally visible changes. The staging trigger moves from `staging/**` to `staging_*`; until a promotion pushes a tag nothing fires either, so the window where it matters is after the cron lands. And the workflow renames start fresh entries in the Actions sidebar — the run history of `rc-deploy-environment.yml` and `rc-teardown-environment.yml` stays under the old names, and any bookmarked dispatch URL for them breaks.


.github/workflows/deploy-environment.yml
```diff
@@ -171,11 +197,17 @@ jobs:
...
+          # holds none of them.
           # `vars` resolves environment over repository over organization, so a
           # repository-level one works too and is the easier of the two to set
           # by accident and then not find again.
+          # Both spellings forwarded: teardown_common.sh prefers TEARDOWN_STRICT
+          # and falls back to the RC_-prefixed one the environment settings still
+          # carry, so the settings rename can happen on either side of this merge.
+          # Drop the second line once `rc` and `nightly` both define the new name.
+          TEARDOWN_STRICT: ${{ vars.TEARDOWN_STRICT }}
           RC_TEARDOWN_STRICT: ${{ vars.RC_TEARDOWN_STRICT }}
```
eLeontev: I would like to remove it in this PR, and just add in github env the new one and keep the old till PR is not merged.

.github/workflows/e2e-manual-runner.yml
```diff
@@ -108,9 +134,15 @@ jobs:
           GKE_CLUSTER_NAME: ${{ vars.GKE_CLUSTER_NAME }}
           AGENT_NAMESPACE: ${{ vars.AGENT_NAMESPACE || 'kubeagents-system' }}
           REGISTRY: ${{ vars.REGISTRY || vars.REGISTRY_PREFIX || (vars.GCP_PROJECT_ID && format('{0}-docker.pkg.dev/{1}/kube-agents', vars.GCP_REGION || 'us-east4', vars.GCP_PROJECT_ID)) || '' }}
```
eLeontev: please remove all fallbacks. The values should be taken only from github env.

.github/workflows/e2e-manual-runner.yml
```diff
@@ -38,8 +60,11 @@ on:
         default: "false"
         type: string
 
+# Follows the environment it was dispatched against. Hardcoding `rc-environment`
+# here used to make this workflow contend with the RC pipeline for a cluster it
+# does not deploy to, whichever environment it was actually pointed at.
 concurrency:
-  group: rc-environment
+  group: ${{ inputs.github_environment }}-environment
```
eLeontev, line 97: I'm a bit confused - do we really want to run e2e test on rc? I would like to do it on nightly stand only and do not allow on rc, or opposite. So tests are always should be run in the same env.

.github/workflows/e2e-manual-runner.yml
```diff
@@ -1,21 +1,43 @@
-name: Manual E2E Runner
+name: "E2E: Manual Runner"
 
 on:
   #checkov:skip=CKV_GHA_7:Manual E2E test runner requires environment and scenario filter parameters
   workflow_dispatch:
```
eLeontev, line 37: q: To run this action, the environment should be provisioned. I don't see any provision scripts, so in case if target gcp project has not clusters (e.g. after previous success run - teardown process) the test will always fail. 

How do we mange it?

.github/workflows/e2e-manual-runner.yml
```diff
@@ -121,9 +153,9 @@ jobs:
           E2E_CHAT_REFRESH_TOKEN: ${{ secrets.E2E_CHAT_REFRESH_TOKEN }}
           CHAT_SPACE_ID: ${{ secrets.E2E_CHAT_SPACE_ID }}
         run: |
-          TARGET_ENV="$TEST_ENVIRONMENT"
-          if [ "$TARGET_ENV" = "all" ]; then
-            python3 scripts/release/execute_e2e_tests.py --env all
+          TARGET_SUITE="$TEST_SUITE"
```
eLeontev: please move the scripts to the script files

.github/workflows/e2e-manual-runner.yml
```diff
@@ -108,9 +134,15 @@ jobs:
...
+          # pair names the RELEASE repository (gke-labs/kube-agents), which is
+          # what common.sh's get_target_repo resolves for tag operations —
+          # handing it to a suite that mints a GitHub App token would scope a
+          # live token at this repository. GITOPS_REPO is bare; the conftest
+          # github_repo fixture composes owner/repo from GITHUB_ORG.
+          GITHUB_ORG: ${{ vars.GITOPS_ORG }}
+          GITHUB_REPO: ${{ vars.GITOPS_REPO }}
+          TEST_SUITE: ${{ inputs.test_suite }}
           STOCKOUT_SCENARIOS: ${{ inputs.stockout_scenarios }}
           FLEET_AUDIT_STREAMS: ${{ inputs.fleet_audit_streams || 'all' }}
```
eLeontev: default already defined in the inputs. please remove it if redundant.

.github/workflows/nightly-pipeline.yml
```diff
@@ -0,0 +1,266 @@
...
+      - name: Record what this run will do
+        env:
+          COMMIT_SHA: ${{ steps.resolve.outputs.commit_sha }}
+          RC_TAG: ${{ steps.resolve.outputs.rc_tag }}
+          STAGING_TAG: ${{ steps.resolve.outputs.staging_tag }}
+          SKIP_PIPELINE: ${{ steps.resolve.outputs.skip_pipeline }}
+          SKIP_PROMOTION: ${{ steps.resolve.outputs.skip_promotion }}
+          SKIP_REASON: ${{ steps.resolve.outputs.skip_reason }}
+        run: |
+          {
```
eLeontev: please move to the script + test

.github/workflows/nightly-pipeline.yml
```diff
@@ -0,0 +1,266 @@
...
+      needs.step-1-resolve-candidate.outputs.skip_pipeline != 'true'
+    uses: ./.github/workflows/teardown-environment.yml
+    permissions:
+      contents: read
+      id-token: write
+    with:
+      github_environment: "nightly"
+      rc_tag: ${{ needs.step-1-resolve-candidate.outputs.commit_sha }}
+    secrets:
+      GCP_SA_KEY: ${{ secrets.GCP_SA_KEY }}
```
eLeontev: we don't have it in github env, I don't understand how we are planning to use it or why we define the secret which is always empty

.github/workflows/rc-release-pipeline.yml
```diff
@@ -72,118 +91,26 @@ jobs:
...
-          GITHUB_ORG: ${{ vars.GITOPS_ORG }}
-          GITHUB_REPO: ${{ vars.GITOPS_REPO }}
-        run: make test-e2e
+    with:
+      github_environment: "rc"
+      commit_sha: ${{ needs.step-1-create-tag.outputs.commit_sha }}
+      # The Chat gate blocks the candidate; the cluster and fleet-audit suite is
+      # coverage that must not. Same split as before this job became a call.
+      blocking_suite: "gchat"
+      optional_suite: "rc"
```
eLeontev: can we use list to use with `optionql_suite`? I would liket to have the option to run multiple suites to allow to fail.

.github/workflows/rc-release-pipeline.yml
```diff
@@ -72,118 +91,26 @@ jobs:
...
+      stockout_scenarios: "04"
```
eLeontev: Blocker: I don't agree to add any of the new tests as mandatory till they passed real pipeline runs e.g. rc or nightly (both independent) since the failures might be in different areas (tests/github configuration/install scripts/ gcp project setup).

In actual combination once we merge it I'm pretty sure that all pipelines will fail. Instead I recommend to use the same approach as we already applied - use only proved tests to run as blocking/mandatory and all the rest use as an optional (like `optional_suites: ['rc', '04']`), etc. 

As well there might be some additional refactoring to rename the suites to readable format to use  at least `04` -> `stockout_suite_04`.

Note: not sure if all not optional are mandatory, but I read it as mandatory.

.github/workflows/rc-scheduler.yml
```diff
@@ -0,0 +1,131 @@
...
+      # ran, which is precisely the invisible-skip problem this file exists to
+      # remove.
+      - name: Dispatch the release pipeline
+        if: steps.resolve.outputs.skip_rc != 'true'
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
+          COMMIT_SHA: ${{ steps.resolve.outputs.commit_sha }}
+          RC_TAG: ${{ steps.resolve.outputs.rc_tag }}
+        run: |
+          set -euo pipefail
```
eLeontev: can we use the rule - all scripts in script files with tests?

tests/e2e/e2e_config.yaml
```diff
@@ -11,21 +11,27 @@ defaults:
   region: "us-central1"
   namespace: "kubeagents-system"
   timeout_seconds: 600
-  default_environment: "investigations-e2e"
+  default_suite: "investigations"
```
eLeontev: I'm a bit curious about selected default. Is it better to use gchat, or create the default suite which is super simple/light dry run?

bradhoekstra:

/review all

.github/workflows/deploy-environment.yml
mplakhtiy: Agreed — done in 55218b9e, with the settings change first because the order matters.

`rc` and `nightly` both carried `RC_TEARDOWN_STRICT=true` and neither defined `TEARDOWN_STRICT`, so dropping the forward before the settings caught up would have left the parser reading an unset variable — and unset is "off", silently, with no error. `TEARDOWN_STRICT=true` is now set on both environments, the old variable is still there until this merges as you asked, and `deploy-environment.yml` forwards only the new name.

`teardown_common.sh` keeps reading the legacy name as a fallback. Nothing in CI reaches it any more; it is there for anyone running these scripts by hand against an environment nobody has migrated, and it goes when `RC_TEARDOWN_STRICT` is deleted from both settings pages. `scripts/release/README.md` now says exactly that instead of describing a migration in progress.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Removed, all of them, in 55218b9e.

`REGISTRY` is deleted outright rather than simplified: `execute_e2e_tests.py` already resolves it with the identical chain — `REGISTRY`, then `REGISTRY_PREFIX`, then `<region>-docker.pkg.dev/<project>/kube-agents` — so the workflow expression was a second copy of that logic in a second place to drift.

The rest now read the bound environment with no `||`: `AGENT_NAMESPACE`, `GCP_REGION`, `GCP_WORKLOAD_IDENTITY_PROVIDER`, `GCP_SERVICE_ACCOUNT`, and `fleet_audit_streams` from its input. The two `us-east4` literals went with them — both environments are `us-central1`, so those defaults were wrong as well as unreachable. `credentials_json: ${{ secrets.GCP_SA_KEY || '' }}` is gone too; see your thread on that one.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: This workflow does not decide where a suite runs in a pipeline — it is the by-hand route into whichever cluster is currently standing, and the unattended answer is fixed and unchanged by this PR: the RC pipeline runs `gchat` plus optional coverage on `rc`, the nightly pipeline runs its matrix on `nightly`, and neither touches the other's project.

What it is for is the failure case. Every pipeline's teardown job hangs off the implicit `success()` of the ones before it, so a red run deliberately keeps its cluster for diagnosis — and the cluster you need is whichever pipeline failed. Restricting this to `nightly` would mean no way to inspect a failed RC cluster by hand.

The `concurrency` edit here is a bug fix rather than a policy change: the group was hardcoded `rc-environment`, so a manual run pointed at `nightly` queued behind the RC pipeline while holding no lock on the cluster it was actually using.

If you would still rather the manual route be nightly-only, it is a one-line change to the choice list — say so and I will make it.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Correct, it provisions nothing, and that is deliberate — it is a debugging tool rather than a pipeline. With no cluster standing it fails at `get-gke-credentials`.

Both environments are ephemeral, so most of the time there is indeed nothing to run against: a pipeline that passes destroys its own cluster in its last step. What leaves one up is a failed run, which skips teardown on purpose.

To build one deliberately, dispatch the same three workflows the pipelines chain: `deploy-environment.yml` (it takes `github_environment` and a tag or commit SHA), then this, then `teardown-environment.yml`. Nothing reclaims a cluster created that way, so the third one matters.

None of that was written where somebody dispatching this would see it, which is the real gap your question found. 55218b9e adds it to the file header.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Agreed — though it collapses rather than moves, in 55218b9e.

Both arms of that `if` ran the same command: `--suite all` is precisely what `execute_e2e_tests.py` already expands into one child process per suite, so the branch was doing nothing. The step is now a single line calling `./scripts/release/execute_e2e_tests.sh`, the same wrapper `e2e-run.yml` and `make test-e2e` go through — no new script to add, and `tests/test_execute_e2e_tests.py` already covers `--suite all`.

The three inline blocks that were genuinely logic did move into scripts with tests; see the nightly-pipeline and rc-scheduler threads.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: Removed in 55218b9e. `fleet_audit_streams` already defaults to `all` on the input, so the `|| 'all'` could only ever fire if the input itself were deleted.

Swept the rest of the block for the same thing while I was in there — that is the fallback thread above.

.github/workflows/nightly-pipeline.yml
mplakhtiy: Agreed. In 55218b9e it is `scripts/release/record_nightly_candidate_summary.sh`, with `tests/test_record_nightly_candidate_summary.py` covering it in 7 tests.

The thing worth having a test for turned out to be the distinction the block encodes: `SKIP_PIPELINE` means no candidate and no run at all, `SKIP_PROMOTION` means the matrix does run but the commit already carries a staging tag so a pass pushes nothing. Rendering one as the other would report a night that did nothing as a night that tested something. The tests pin both renderings, that it appends rather than clobbers `GITHUB_STEP_SUMMARY`, and that it runs outside Actions without one.

.github/workflows/nightly-pipeline.yml
mplakhtiy: You are right, and it is worse than unused — I checked, and `GCP_SA_KEY` is set at no scope at all: not repository, not `rc`, not `nightly`. Every reference to it has resolved to an empty string on every run that has ever happened, and WIF has always been the only auth path.

It is pre-existing rather than new here: `main` forwards it from `rc-release-pipeline.yml` and declares it optional in `rc-teardown-environment.yml`, and the nightly file copied the pattern. Since it is dead everywhere, 55218b9e removes the whole chain rather than just this line — the `secrets:` block in `teardown-environment.yml` and the `credentials_json` line under it, the forward from both pipelines, and the same `|| ''` fallback in `e2e-manual-runner.yml` and `e2e-gchat-test.yml`. Six references, no behaviour change, one fewer credential path to keep in step.

.github/workflows/rc-release-pipeline.yml
mplakhtiy: Yes. In 55218b9e `optional_suite` becomes `optional_suites`, a comma-separated list — `workflow_call` has no list input type, arrays exist only for `workflow_dispatch` choices.

`scripts/release/run_optional_e2e_suites.sh` drives it, with `tests/test_run_optional_e2e_suites.py` (9 tests). The property worth pinning is that every suite runs whatever the ones before it did: a loop that stopped at the first failure would silently drop the coverage behind it and the step would look identical. The script exits non-zero naming the failures, which `continue-on-error` turns into a tolerated red with a per-suite table in the job summary rather than one opaque step.

.github/workflows/rc-release-pipeline.yml
mplakhtiy: Taking the blocker first, since it decides the rest.

**On the RC side nothing became mandatory in this change.** `blocking_suite: gchat` with `rc` optional is exactly the split `main` runs today — `E2E_ENV: gchat-e2e` as the gate, `E2E_ENV: rc-e2e` under `continue-on-error: true`. The job became a call to `e2e-run.yml`; the tolerance did not move. So merging this cannot turn the RC pipeline red by making something new mandatory there.

**Where your point lands is the nightly pipeline, and I agree with it.** `blocking_suite: nightly` put all five test files in front of the staging promotion, and three of them have never completed a green run — `e2e-nightly-matrix.yml`, the workflow this replaces, has never had a successful run at all. The first red night would have told you nothing about whether the candidate was bad or the suite was, and staging would have stopped moving either way.

55218b9e gates on evidence instead. `blocking_suite: rc` — Agent API health, the fleet-audit ledger across all streams, and stockout 04 — is the subset the RC pipeline exercises every three hours and passes; [run 33378383224](https://github.com/gke-labs/kube-agents/actions/runs/33378383224) is the most recent. The rest run as `optional_suites: agent-plugin,gchat,stockout-full`, which between them cover every file the `nightly` suite holds, so a night still produces the full matrix's worth of signal. The bar for promoting one into the gate is a green nightly run with it in the optional list, and the comment in the file says so.

**On naming:** `04` is not a suite, it is the `STOCKOUT_SCENARIOS` filter, and `test_stockout_investigation.py` already matches by prefix *or* exact slug — so the readable form works today. It is now `04-missing-zone-fallback` everywhere it is pinned. The filters themselves moved out of the workflow into each suite's `env_vars` in `e2e_config.yaml`, which the list made necessary anyway: one workflow input cannot carry a different scenario filter for `rc` and for `stockout-full`.

**On your last note** — yes, anything not in `optional_suites` is mandatory. Only `blocking_suite` fails the run.

.github/workflows/rc-scheduler.yml
mplakhtiy: Yes, and that is the rule the rest of `scripts/release/` already follows — these two steps were the exception. Both moved in 55218b9e:

- `scripts/release/dispatch_rc_pipeline.sh` — the `gh workflow run` call, its `::error` annotation, and the dispatched-summary table.
- `scripts/release/record_rc_scheduler_skip.sh` — the quiet-tick summary.

`tests/test_dispatch_rc_pipeline.py` covers both in 8 tests, including that a failed dispatch produces the annotation rather than a bare exit, and that a missing `COMMIT_SHA` or `RC_TAG` aborts before calling `gh`. `tests/test_rc_scheduler_wiring.py` used to find the dispatch by grepping the workflow for `gh workflow run`; it now reads across the seam instead — the workflow decides when to run the script and with which token, the script decides what it sends, and the pipeline's declared inputs still have to accept it.

One note on the hunk you are reading: it predates 3ba0d40a, which switched this step from `RELEASE_BOT_TOKEN` to `github.token`. `workflow_dispatch` is exempt from the recursion suppression, so the PAT was never needed here.

tests/e2e/e2e_config.yaml
mplakhtiy: Gone with `gchat` in 55218b9e.

Worth saying how narrow the blast radius is: `default_suite` only fires when nothing sets `E2E_SUITE`, and every CI path sets it explicitly — `e2e-run.yml` from its input, `e2e-manual-runner.yml` from its dispatch input. So this is the local-developer default and conftest's fixture fallback, never something a pipeline reaches.

Given that, `gchat` over a new light dry-run suite: it is the only suite with an unbroken record of passing on a real cluster, so it is the honest answer to "what runs if you name nothing". It does need the `E2E_CHAT_*` credentials, and failing loudly on a missing credential is a better default than a suite that half-runs and reports green.

.github/workflows/e2e-manual-runner.yml
eLeontev: I would like to discuss the concept - I don't think that we will run it in that scenario. So let's figure out why we need this and when we will run it?

.github/workflows/e2e-manual-runner.yml
eLeontev: TBD https://github.com/gke-labs/kube-agents/pull/1058#discussion_r3895473154

tests/e2e/e2e_config.yaml
```diff
@@ -11,34 +11,49 @@ defaults:
   region: "us-central1"
   namespace: "kubeagents-system"
   timeout_seconds: 600
```
eLeontev: nit: please take a look about timeout value set in nightly yaml - it looks like that nightly will fail due to shorter timeout - the right values should be exact the same or a bit larger to include provision/teardown.

.github/workflows/e2e-run.yml
```diff
@@ -0,0 +1,210 @@
...
+      # and carries no default.
+      github_environment:
+        description: "GitHub environment supplying the GCP project, region and cluster"
+        required: true
+        type: string
+      commit_sha:
+        description: "Commit SHA the environment was deployed at"
+        required: true
+        type: string
+      blocking_suite:
```
eLeontev, line 38: nit: can be implemented in a separate PR, but looks like blocking_suite need to be renamed to `blocking_suites` and accepts list of suits for better flexibility.

.github/workflows/e2e-manual-runner.yml
mplakhtiy: I will merge this with a comment to extend it to autopush and staging enviroment and will link an issues. and state that this works only with rc and nightly only on failures and until next run is executed and destroys existing clusters.

tests/e2e/e2e_config.yaml
mplakhtiy: Looked into this, and the value the comment lands on is inert: `defaults.timeout_seconds` is read by nothing. `execute_e2e_tests.py`'s `run_suite_tests` consults `defaults` only for `region`, `namespace` and `default_suite`, and `conftest.py` merges the block into its settings dict without ever looking that key up. It has been dead since before this branch. 742dd747 removes it and says what the block is actually for, because a knob nobody reads still reads as configuration — which is what drew the question.

On the real timing: the budgets are per job and they do not accumulate. Step 2 (provision) has 60 minutes of its own in `deploy-environment.yml`, step 5 (teardown) has 45 in `teardown-environment.yml`, and step 3's `timeout_minutes` covers only the E2E job — checkout, auth, readiness and the suites. So it does not need to be larger to "include provision/teardown"; those never run inside it.

The nightly passes 120 against the RC's default of 60, which is the right direction anyway: the same fixed readiness cost, but all five test files rather than one suite's worth.

.github/workflows/e2e-run.yml
mplakhtiy: Agreed, and taking the separate-PR route you offered: gke-labs/kube-agents#1073.

The asymmetry is not deliberate — the gate was a single suite when this workflow was factored out, and the list arrived later for the tolerated side. The nightly pipeline is the case that will want it first, since the plan there is to fold `agent-plugin`, `gchat` and `stockout-full` into the gate as each earns a green run, and the second one to be promoted needs `blocking_suites`.

The issue records what it involves: `run_optional_e2e_suites.sh` already runs a comma-separated list, keeps going past a failure, and exits non-zero naming the failures — blocking behaviour already, with `continue-on-error` the only difference — so the work is mostly renaming it and its `OPTIONAL_SUITES` variable to something neutral, plus the two call sites and the docs. The one open decision is whether a blocking list should stop at the first failure or run all of them.

Keeping it off this branch, which is already 68 files.

eLeontev (approved):

PR #1063, opened 2026-08-31
fix(ci): retry the agent connectivity check across fresh tunnels; capture envoy and kube-system evidence on failure

jayantid:
## Problem

The smoke harness's Platform Agent connectivity check (`hack/ci-deploy.sh`, section 7) starts `kubectl port-forward svc/platform-agent 8642:8642` in the background, waits with `nc -z`, then issues **one** `curl -s -X POST /v1/responses` — a real model inference — with no `--max-time`, no retry, and no port-forward restart. Any empty or non-matching response is a hard `exit 1`.

On 2026-08-31, 4 of 7 runs died exactly there, with an identical port-forward log each time: one `Handling connection for 8642` followed by `error: lost connection to pod` — while the gateway pod was 4/4 Running, 0 restarts, probes passing. The four deaths landed on four different pool projects and the same `base_sha` as interleaved passes (builds `2094332662654177280`, `2094347266671251456`, and two PR-1058 builds). Healthy checks complete in 37–42s; the deaths came 26s–5m30s in, on cold autoscaling pools where the API-server tunnel is least stable. This is a flake in the check, not a product regression: a single tunnel drop mid-inference is indistinguishable from a broken agent, and a dead `kubectl port-forward` never recovers on its own.

## Fix

- Wrap the check in a bounded retry (5 attempts). **Each attempt gets a fresh tunnel**: kill any existing port-forward, spawn a new one (pid tracked, `nc -z` wait bounded as before), then curl with `--max-time 120`. One-line log per failed attempt (`connectivity attempt 2/5 failed: empty response after port-forward drop; respawning tunnel`); hard-fail only after all attempts, with the port-forward log tail echoed (the log is now appended across attempts with per-attempt headers, so the failure dump shows every tunnel's fate).
- The success predicate, endpoint, and payload are unchanged — what "healthy" means is identical. The check still exits with the tunnel torn down and the plain `dump_prow_artifacts_on_failure` trap restored; nothing downstream reuses this tunnel (the bench harness manages its own per-task forwards in `bench/kube_agents_bench/harness.py`).

This converts all four 08-31 deaths to passes: the same clusters served the identical request successfully minutes later, so a second attempt on a fresh tunnel would have succeeded in each case — while a genuinely broken agent still fails all 5 attempts, now with better evidence.

## Failure-dump additions (`dump_prow_artifacts_on_failure`, `hack/ci-env.sh`)

Two captures the 08-31 investigation was missing, each `|| true`-guarded in the function's existing style:

- `kubectl logs deployment/platform-agent-gateway -c envoy-credential-proxy` — the existing gateway capture reads the pod's default container (platform-agent); a dropped stream is visible from the envoy sidecar's side, which is exactly the evidence we didn't have.
- `kubectl get events -n kube-system --sort-by=.lastTimestamp | tail -100` — konnectivity/tunnel churn shows up there, not in the agent namespace.

## Tests

`scripts/test_ci_connectivity_retry.py` (same lift-the-real-shell convention as `scripts/test_ci_eval_trap.py`, stubbed `kubectl`/`curl`/`nc`): first-attempt drop → retry on a fresh tunnel and exit 0 with exactly one respawn logged; all attempts exhausted → exit 1 with every attempt logged and the tunnel log tail echoed; the curl carries `--max-time`; the two new artifact captures run in the dumper and cannot fail it. `bash -n` clean on both scripts; shellcheck delta vs main is empty (only the pre-existing SC2034 at a shifted line); `scripts/test_task_registration.py` and `scripts/test_ci_eval_trap.py` still green.

Part of #897; evidence relates to #988's startup family and the 08-31 wave investigation.

### DIFF
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -306,37 +306,70 @@ STEP_START=$SECONDS
 echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Verifying Platform Agent API Connectivity ==="
 API_KEY="$(kubectl get secret platform-agent-secrets -n "${NAMESPACE}" -o jsonpath='{.data.API_SERVER_KEY}' | base64 --decode)"
 
-kubectl port-forward svc/platform-agent -n "${NAMESPACE}" 8642:8642 >/tmp/pf-8642.log 2>&1 &
-PF_PID=$!
+# On cold autoscaling pools the API-server tunnel behind `kubectl port-forward`
+# drops mid-request ("error: lost connection to pod" with the gateway pod
+# healthy throughout), and a dead port-forward never comes back on its own —
+# so every attempt gets a fresh tunnel, and only the response decides health.
+PF_PID=""
 cleanup_pf_and_dump() {
   kill "${PF_PID:-}" 2>/dev/null || true
   dump_prow_artifacts_on_failure
 }
 trap cleanup_pf_and_dump EXIT
 
-echo "Waiting for platform-agent port-forward on port 8642..."
-for i in {1..30}; do
-  if nc -z localhost 8642 2>/dev/null; then
+CONNECTIVITY_ATTEMPTS=5
+CONNECTIVITY_OK="false"
+: >/tmp/pf-8642.log
+for ((attempt = 1; attempt <= CONNECTIVITY_ATTEMPTS; attempt++)); do
+  # Kill any previous tunnel and start a fresh one; the log is appended so a
+  # failure dump shows every attempt, not just the last.
+  if [ -n "${PF_PID}" ]; then
+    kill "${PF_PID}" 2>/dev/null || true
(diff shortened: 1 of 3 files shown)


jayantid:
The smoke failure was infrastructure, not this PR: the Cloud Build image build died fetching `go.yaml.in/yaml/v2@v2.4.4` from proxy.golang.org (`stream error: INTERNAL_ERROR`) ~14 min in, before deploy — the connectivity check this PR changes never executed. Contributing factor: the run leased `kube-agents-evals-10`, a freshly onboarded pool project with no warm layer cache in its Artifact Registry yet, so the full image build ran and put the public module proxy back on the critical path — the same class oss-test-infra#2664 fixed for boskosctl, now via the image-build route on cold projects. Warm-cache seeding belongs on the pool-onboarding checklist (#994). Retriggered.

dshnayder (approved):

PR #1068, opened 2026-08-31
feat(plugins): publish plugin images and template AgentPlugin in Helm and Terraform

### LINKED ISSUE #1013: AgentPlugins cannot be installed by Terraform + Helm, so the stockout investigator ships with dead alert ingress
Neither plugin under `agentplugins/` can be installed by the project's own install engine. Terraform provisions the GCP half and the `kube-agents` chart provisions the Kubernetes half, but neither knows an `AgentPlugin` exists: the only way to get one onto a cluster is to run the plugin's own `install.sh` by hand, which builds and pushes an image from the source tree.

The user-visible consequence is that `gke-stockout-investigator` installs and reports healthy while doing nothing. It contributes route config under `platforms.pubsub.extra.subscriptions` and nothing else; the adapter that opens the listener is a *separate* plugin, `pubsub-platform`. Install the investigator without it and the gateway opens no listener at all, so alerts are published to a topic nobody is pulling. The failure is silence — no error in the CR, no error in the gateway log, an `AgentPlugin` in phase `Ready`. `agentplugins/gke-stockout-investigator/README.md` names the adapter as the ingress dependency, so the coupling is known; nothing enforces it.

This was found while diagnosing the RC promotion gate, where the same gap failed `test_stockout_ingress_alert_smoke` and `test_stockout_scenario[04]` on every run for a week. #1012 installs the adapter from `scripts/release/wait_for_gke_readiness.sh` so the gate is honest, and deliberately does not close this: the RC now validates a configuration no user install produces, which is its own problem.

## What is missing

- **The images are not published.** `images.json` — the inventory of every image an install pulls, read by `make mirror-images`, the kustomize targets and the docs generator — has no entry for either plugin. No workflow builds or pushes them; `agentplugins-test.yml` only runs their unit tests, and its own header records that "these ship as their own OCI images". Today the tag is derived from source content by `plugin_image_resolve` and built on the installing host, which is a different model from the four pinned release images.
- **The chart templates no `AgentPlugin`.** `charts/kube-agents/templates/` mentions the kind only in `operator-rbac.yaml`. There is no value to enable a plugin and no template to render one.
- **Terraform has no variable for it**, so `terraform/examples/full-install` cannot turn one on, and `install.sh` has no flag.

## Rough shape of a fix

1. Publish `pubsub-platform` (and then `gke-stockout-investigator`) as pinned release images, alongside the existing four, with `images.json` entries and a pin.
2. Template the `AgentPlugin` in the chart behind a value, defaulting off.
3. Thread one variable through `terraform/examples/full-install` and an `install.sh` flag.
4. Regenerate the docs the inventory feeds, and reconcile `agentplugins/README.md` and the two plugin READMEs to the new path.

Step 1 is the design decision worth settling first: a pinned, published image is a different model from the content-tag-and-build-locally scheme the plugin installers use today, and whichever wins should apply to both plugins rather than being decided per plugin.

## Worth deciding alongside

Whether a consumer plugin should be able to declare its ingress dependency, so that installing `gke-stockout-investigator` without `pubsub-platform` is a reconcile error rather than silence. The operator already reports `Degraded` for an unpullable plugin image, so there is somewhere for such a check to live.

Filed from the #1012 investigation. Written with the help of Claude.

sergiu1spiridon:
## Summary

Enables `AgentPlugin` deployment (`pubsub-platform` and `gke-stockout-investigator`) through the canonical `kube-agents` Helm chart, Terraform composition (`terraform/examples/full-install`), and installer scripts (`install.sh`). Publishes both plugins as pinned OCI release images in `images.json`, GHCR, and GCP workflows.

## Why This Change

Neither plugin under `agentplugins/` could be installed by the repository's install engine. While `gke-stockout-investigator` was packaged in the repo, it relies on the `pubsub-platform` adapter for Cloud Pub/Sub alert ingress. Without declarative templating in Helm, Terraform variables, and published release images, deploying the investigator resulted in dead ingress.

## What Changed

- **Release Image Publishing (`images.json`, `.github/workflows/`):** Registered `pubsub-platform` and `gke-stockout-investigator` in `images.json` as first-party release images, and added container build/push/cosign signing steps to GHCR and GCP release workflows.
- **Helm Chart (`charts/kube-agents/`):** Added `plugins.pubsubPlatform` and `plugins.stockoutInvestigator` configurations in `values.yaml` (defaulting to disabled) and added `charts/kube-agents/templates/plugins.yaml` rendering `AgentPlugin` CRs with image mirror registry rewriting support and dependency enforcement (`stockoutInvestigator` requires `pubsubPlatform`).
- **Terraform (`terraform/examples/full-install/`):** Exposed `enable_pubsub_platform`, `enable_stockout_investigator`, and Pub/Sub alert topic/subscription variables in `variables.tf`, forwarding them into the Helm release in `main.tf`.
- **Installer CLI (`install.sh`, `installer_common.sh`):** Added `--enable-pubsub-platform` and `--enable-stockout-investigator` CLI flags, updating state files and generating `terraform.tfvars`.
- **Automated Tests (`tests/test_helm_plugins.py`, `tests/test_installer_common.py`, `tests/test_install_script.py`):** Added unit tests covering Helm plugin rendering, dependency assertions, image mirror rewriting, CLI argument parsing, and tfvars generation.

## Context

Closes #1013

## Testing

- `python3 -m unittest tests/test_helm_plugins.py tests/test_install_script.py tests/test_installer_common.py tests/test_provision_rc_environment.py` (97 passed)
- `make docs-generate`
- `make docs-check` (passed, 0 errors)

### Live validation

- **Target environment:** GKE Autopilot cluster `platform-agent-host` in `us-east4` on project `sergiuspiridon-gkedemos`.
- **Workflow verification:** Executed RC deployment workflow with plugins enabled (`ENABLE_PUBSUB_PLATFORM=true`, `ENABLE_STOCKOUT_INVESTIGATOR=true`).
- **Observed:** Verified `AgentPlugin` CR reconciliation by operator, `pubsub-platform` container sidecar start, and `gke-stockout-investigator` container sidecar start in namespace `kubeagents-system`.

## Self-Review

- **Adversarial review angles:**
  - Verified dependency assertion in Helm chart when `stockoutInvestigator.enabled=true` and `pubsubPlatform.enabled=false`.
  - Verified default values ensure existing deployments without plugin flags remain 100% backward compatible (`enabled: false`).
  - Verified mirror registry rewriting with `global.imageRegistry`.
- **Docs-drift review:**
  - Ran `make docs-generate` and `make docs-check` verifying documentation table synchronization with `images.json`.

## Risk & Rollout

- **Risk:** Low. AgentPlugins are disabled by default in both Helm and Terraform; existing installations without explicit plugin enablement are unaffected.
- **Rollout:** Merges cleanly to `main` without runtime breaking changes.

---
Generated with the help of Antigravity.


terraform/examples/full-install/main.tf
```diff
@@ -310,6 +313,62 @@ module "github_minter" {
   depends_on = [google_project_service.required]
 }
 
+resource "google_pubsub_topic" "stockout_alerts" {
```
mplakhtiy, line 320: **BLOCKER — the composition creates Pub/Sub resources that already exist in the RC and nightly projects, and every deploy opts in by default.**

These three resources default to `gke-stockout-alerts-topic` / `-sub` / `-sink` (`variables.tf:459-472`) — the same three names `agentplugins/gke-stockout-investigator/install.sh:56-58` creates imperatively. That installer has already run against the RC and nightly projects: `tests/e2e/test_stockout_investigation.py:627-636` creates the topic with `gcloud` when it is absent, and `:663-680` runs `install.sh` whenever the `gkestockoutinvestigator` AgentPlugin is missing — the state of every environment provisioned before this PR. Nothing removes them afterwards; `teardown_environment.sh` runs `uninstall.sh`, which destroys the composition, and the composition never knew about them.

Terraform does not adopt pre-existing resources. `lifecycle.sh` imports only the KMS objects, and says why it has to (`lifecycle.sh:249`: "could not import $address ($id); the apply will fail with a 409").

Nobody has to opt in for this to fire. `scripts/release/provision_environment.sh:152,156` read `${ENABLE_PUBSUB_PLATFORM:-true}` and `${ENABLE_STOCKOUT_INVESTIGATOR:-true}`, and an unset repository variable renders as the empty string, so the default applies and both flags are passed. **Step 2 of the RC pipeline and the nightly pipeline both go red on the first run after merge**, and stay red until someone deletes the three objects by hand or imports them into state.

The same trap catches any user who installed the plugin the documented way (`agentplugins/README.md`) and then sets `enable_stockout_investigator = true`.

Fix: give the topic/subscription/sink `terraform import` handling in `lifecycle.sh` the way `adopt_kms` does, or default `enable_stockout_investigator` to `false` in `provision_environment.sh` (see the note on `:152`) and clean the three objects out of the RC and nightly projects before the first apply.

_Not verified against a real project — derived from the resource names, the e2e fixture, and `lifecycle.sh`'s import scope, not from an observed 409._

terraform/examples/full-install/main.tf
```diff
@@ -310,6 +313,62 @@ module "github_minter" {
...
+    ttl = ""
+  }
+}
+
+resource "google_logging_project_sink" "stockout_alerts" {
+  count       = var.enable_stockout_investigator ? 1 : 0
+  project     = var.project_id
+  name        = var.stockout_pubsub_sink
+  destination = "pubsub.googleapis.com/${google_pubsub_topic.stockout_alerts[0].id}"
+  filter      = "resource.type=\"k8s_cluster\" AND (log_id(\"test-stockout\") OR log_id(\"container.googleapis.com/cluster-autoscaler-visibility\")) AND (jsonPayload.messageId:(\"scale.up.error.out.of.resources\" OR \"scale.up.error.quota.exceeded\" OR \"scale.up.error.ip.space.exhausted\" OR \"scale.up.no.scale.up\") OR jsonPayload.noDecisionStatus.noScaleUp:* OR jsonPayload.resultInfo.results.errorMsg.messageId:(\"scale.up.error.out.of.resources\" OR \"scale.up.error.quota.exceeded\" OR \"scale.up.error.ip.space.exhausted\" OR \"scale.up.no.scale.up\")) AND (resource.labels.cluster_name=\"${module.gke_cluster.cluster_name}\" OR jsonPayload.resource.labels.cluster_name=\"${module.gke_cluster.cluster_name}\")"
```
mplakhtiy: **HIGH — the `resource.type="k8s_cluster"` conjunct drops every synthetic alert routed through Cloud Logging.**

The filter this replaces (`agentplugins/gke-stockout-investigator/install.sh:68`) has no such conjunct, and the comment immediately above it says why in as many words:

> Both spellings of the label are matched. Real autoscaler entries carry the monitored resource, so the cluster name is at `resource.labels`; the synthetic `test-stockout` entries the scenarios write are `global`-resourced, and carry it inside the payload.

`agentplugins/gke-stockout-investigator/scenarios/lib/common.sh:414` writes those entries with `gcloud logging write test-stockout "$inner" --payload-type=json`, which stamps `resource.type="global"`. Under this filter the sink matches none of them, so `--via-sink` (`scenarios/README.md:67`, "Route through Cloud Logging instead of publishing to the topic") — the one mode that exercises the production ingress path end to end — silently delivers nothing.

Real autoscaler events are `k8s_cluster`-resourced, so production alerting still works and the nightly suite (which publishes straight to the topic) still passes. That is what makes this quiet rather than loud, and it is the precise failure mode the sibling filter's comment exists to prevent.

Fix: drop `resource.type="k8s_cluster" AND ` so the filter matches `install.sh:68`, or state why the synthetic path is being given up.

mplakhtiy (commented):
Review of the full diff, run in a fresh context that did not write the change, following `.agents/skills/review-adversarial/SKILL.md`.

The pass started at `497124e` and you pushed `ceb0d04` and `ccfb436` while it was running. I re-checked every finding against the new head: **both of those commits land** — `ceb0d04` removes the `ALLOW_UNENCRYPTED_SECRETS` → `enable_database_encryption` widening, and `ccfb436` restores the `image_registry` fallback in `local.third_party_registry` so it matches the comment above it and the chart helper. That also closes the fourth open `kube-agents-bot` thread, which `04256b12` had not: that commit fixed the chart helper, a different consumer from the Terraform local. The other three bot threads are addressed by `5a29367`/`b3e6952`/`497124e`.

What remains is **1 blocker, 2 high, and 6 medium**, below. Four LOW findings (unread `plugins.*.image.pullPolicy` values key, `agentRef` defaulting to a literal rather than `platformAgent.name`, five stale doc lines, and the Self-Review/Live-validation sections) are held back to keep this readable — ask and I will post them.

Green CI is not evidence for most of these: they sit on paths CI does not exercise — a real `terraform apply` against a long-lived project, the `--via-sink` alert path, and a docker-daemon plugin build.

agentplugins/pubsub-platform/Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM scratch
+FROM alpine:3.24
```
mplakhtiy: **HIGH — `FROM scratch` → `FROM alpine:3.24` breaks the plugin-image contract and makes the two builders disagree under one content tag.**

Two documents state the contract this breaks, and neither is updated. `agentplugins/lib/plugin_image.sh:4`: "Every plugin image in this directory is `FROM scratch` plus a COPY of one tree, so the entire build is a single tar layer." `agentplugins/README.md:35` repeats it.

The consequence is not cosmetic. `plugin_image_crane_publish` (`plugin_image.sh:872-905`) assembles a single-layer image with `crane append` on an empty base — it never reads the Dockerfile — while `plugin_image_docker_publish` (`:861-868`) builds the Dockerfile and now produces an alpine rootfs. `plugin_image_build_digest` (`:385-394`) hashes the Dockerfile into the content tag, so both builders mint the **same** tag for materially different images, and `plugin_image_publish` (`:972-975`) skips the build when that tag already exists. The file's own comment at `:114-123` names this outcome: "Whichever builder reached that tag first would then define it forever, because publish skips a tag that already exists." `plugin_image_builder` picks crane whenever no Docker daemon answers, described there as "the common case on a laptop", so both branches are routinely taken.

The plugin image is mounted as an `ImageVolumeSource` at `$HERMES_HOME/plugins/<name>` (`platformagent_manifests.go:2028-2041`, `:967-971`), so with the alpine base that directory becomes a whole distribution root instead of the four plugin files, and the image inherits alpine's CVE surface.

The change also has no consumer left in this PR: it was introduced for the staging init container that ran `/bin/sh` inside the plugin image, and commit `b3e6952` reverted that container. The diff now contains no `k8s-operator/internal/` changes at all. Nothing here needs a shell.

Fix: revert both Dockerfiles to `FROM scratch`.

agentplugins/gke-stockout-investigator/Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM scratch
+FROM alpine:3.24
```
mplakhtiy: **HIGH (same finding as on `agentplugins/pubsub-platform/Dockerfile:1`)** — `FROM scratch` → `FROM alpine:3.24` breaks the single-tar-layer contract in `agentplugins/lib/plugin_image.sh:4` and `agentplugins/README.md:35`, and makes the crane and docker builders produce materially different images under the same content tag. The staging init container that motivated it was reverted in `b3e6952`. Revert to `FROM scratch`.

terraform/examples/full-install/main.tf
```diff
@@ -310,6 +313,62 @@ module "github_minter" {
...
+resource "google_pubsub_topic" "stockout_alerts" {
+  #checkov:skip=CKV_GCP_83:Stockout alert topic uses default Google-managed encryption keys
+  count   = var.enable_stockout_investigator ? 1 : 0
+  project = var.project_id
+  name    = var.stockout_pubsub_topic
+
+  depends_on = [google_project_service.required]
+}
+
+resource "google_pubsub_subscription" "stockout_alerts" {
```
mplakhtiy, line 329: **MEDIUM — these six resources reimplement `terraform/modules/drift-pubsub` inline, and drop two of its settings.**

`terraform/modules/drift-pubsub/main.tf:70-155` already has this exact shape — topic, subscription, logging sink with `unique_writer_identity`, the sink-writer `roles/pubsub.publisher` binding, and the consumer's `roles/pubsub.subscriber` + `roles/pubsub.viewer` pair — with the reasoning for each in comments (notably why `viewer` is needed alongside `subscriber`, at `:150-155`). AGENTS.md describes `drift-pubsub` as a reusable module that "is not yet part of that composition", and the composition's convention for this shape is a module call (`module.chat_pubsub`, `main.tf:277`).

Two settings the module carries do not survive the copy: `retry_policy` — its comment explains it keeps a nacked message from becoming a hot redelivery loop — and `message_retention_duration`.

The duplication is confirmed. The missing `retry_policy` is a **possible** runtime problem rather than a confirmed one: it depends on whether the pubsub adapter nacks unparseable payloads the way the drift detector does, which I did not establish.

charts/kube-agents/templates/plugins.yaml
```diff
@@ -0,0 +1,87 @@
+{{- if .Values.plugins }}
+{{- if and .Values.plugins.stockoutInvestigator .Values.plugins.stockoutInvestigator.enabled }}
+{{- if not (and .Values.plugins.pubsubPlatform .Values.plugins.pubsubPlatform.enabled) }}
+{{- fail "plugins.stockoutInvestigator is enabled, but plugins.pubsubPlatform is disabled: the GKE Stockout Investigator requires the pubsub-platform adapter for Cloud Pub/Sub alert ingress. Set plugins.pubsubPlatform.enabled=true." }}
+{{- end }}
+{{- end }}
+
+{{- if and .Values.plugins.pubsubPlatform .Values.plugins.pubsubPlatform.enabled }}
```
mplakhtiy, line 12: **MEDIUM — this template is a third copy of both plugins' AgentPlugin specs, with nothing keeping the copies in step.**

The `platform_toolsets` block is copied from `agentplugins/pubsub-platform/templates/agentplugin.yaml`, and the whole `platforms.pubsub.extra.subscriptions.gke_stockout_alerts` block — prompt, query, filter, `threshold_*`, `deduplicate_fields`, `skills`, `subscription`/`topic`/`sink` — from `agentplugins/gke-stockout-investigator/templates/agentplugin.yaml:43-133`, verbatim except that every explanatory comment is dropped. The dedup-field list, the "identity ONLY, deliberately no failure detail" reasoning, the "do not add `mcp-agent_common`" warning: all of it now has two homes and one of them says nothing.

AGENTS.md's install-engine rule is explicit that duplication needs a guard: "The two places manifests still exist twice on purpose … each have a check or a comment saying so." This third copy has neither. #798 (`sklodowska-dev`, open) is adding diagnostics to the same plugin, so the copies can start diverging immediately.

images.json
```diff
@@ -36,6 +36,24 @@
       "pulledBy": "The optional inference-replay integration.",
       "description": "Record/replay proxy for model traffic."
     },
+    {
+      "name": "pubsub-platform",
+      "origin": "first-party",
+      "repository": "ghcr.io/gke-labs/kube-agents/pubsub-platform",
+      "tagPolicy": "release",
+      "override": "PUBSUB_PLATFORM_IMAGE",
```
mplakhtiy: **MEDIUM — these two `override` keys name environment variables nothing reads, and the generated docs publish them.**

`PUBSUB_PLATFORM_IMAGE` and `STOCKOUT_INVESTIGATOR_IMAGE` appear nowhere else in the repository. Compare the entries beside them: `REPLAY_IMAGE` is consumed by `k8s-operator/Makefile:199-204`, `CREDENTIAL_PROXY_IMAGE` by `k8s-operator/internal/controller/manifest_helpers.go:60`.

The chart takes `plugins.<name>.image.repository` and the composition exposes no variable for it, so the "Override" column of the published table (`docs/site/src/content/docs/deploy/docker-images.md:36-37`) now names two environment variables a reader can set with no effect.

scripts/release/provision_environment.sh
```diff
@@ -149,6 +149,14 @@ if [ -n "${USER_PROFILE_ENABLED:-}" ]; then
   INSTALL_ARGS+=(--user-profile-enabled="${USER_PROFILE_ENABLED}")
 fi
 
+if [ "${ENABLE_PUBSUB_PLATFORM:-true}" = "true" ]; then
+  INSTALL_ARGS+=(--enable-pubsub-platform)
+fi
+
+if [ "${ENABLE_STOCKOUT_INVESTIGATOR:-true}" = "true" ]; then
+  INSTALL_ARGS+=(--enable-stockout-investigator)
+fi
```
mplakhtiy: **MEDIUM — this file defaults both plugin flags to true, alone among every default in the repository.**

`install.sh` defaults both to `false` (`install.sh:87-88`), the chart defaults both to `false` (`values.yaml:630,644`), and the Terraform variables default to `false` (`variables.tf:444,450`). This one file defaults them to `true`, and it is the only path any CI environment is provisioned through.

It is also the odd one out inside its own file: every other optional knob here is either `[ -n "${VAR:-}" ]` pass-through or, for `GOOGLE_CHAT_ENABLED`, `= "true"` defaulting to false. And the equality test is not the repo's `is_truthy`, so `ENABLE_PUBSUB_PLATFORM=1` or `TRUE` **disables** the plugin rather than enabling it.

The only trace of the decision is `validate_and_log_deploy_summary.sh:66-67` printing `<DEFAULT (true)>`. It is not in the PR description, and it is what makes the blocker above reachable without anyone opting in.

agents/platform/scripts/pr_triggers.py
```diff
@@ -251,11 +251,11 @@ def strip_markers(text: str) -> str:
     than GitHub's comment limit allows — converges in 2,602 passes and 0.08s.
     """
     text = normalise_newlines(text)
-    while True:
-        stripped = MARKER_RE.sub("", text)
-        if stripped == text:
-            return text.strip()
-        text = stripped
+    while "agent-" in text:
```
mplakhtiy: **MEDIUM — out of scope: this rewrite does not serve "publish plugin images and template AgentPlugin", and is not in the body's "What Changed".**

The rewrite itself is *correct* — every `MARKER_RE` match contains the literal `agent-` (`:126-128`), so the fast path is sound. The problem is its companion change at `agents/platform/scripts/test_pr_triggers.py:483`, which loosens the convergence assertion from `< 0.3s` to `< 0.5s`. A perf guard relaxed on a branch that is not about performance is the shape the adversarial-review method calls out by name, and the docstring still claims 0.08s.

This is one of three unrelated hunks left in the diff; see the notes on `validate_and_log_deploy_summary.sh:18` and `docker-publish-ghcr.yml:19`.

agents/platform/scripts/test_pr_triggers.py
```diff
@@ -480,7 +480,7 @@ def test_no_marker_survives_at_any_nesting_depth(self):
         elapsed = time.monotonic() - started
         self.assertEqual(out, "")
         self.assertEqual(pr_triggers.MARKER_RE.findall(out), [])
-        self.assertLess(elapsed, 0.3, f"took {elapsed:.3f}s")
+        self.assertLess(elapsed, 0.5, f"took {elapsed:.3f}s")
```
mplakhtiy: **MEDIUM (part of the out-of-scope finding on `pr_triggers.py:254`)** — a performance assertion loosened from `< 0.3s` to `< 0.5s` on a branch that is not about performance, with the docstring still claiming 0.08s. If the fast-path rewrite is correct, this bound should not have needed relaxing; if it did need relaxing, that is worth saying out loud.

scripts/release/validate_and_log_deploy_summary.sh
```diff
@@ -16,6 +15,10 @@ for var in "${REQUIRED_VARS[@]}"; do
   fi
 done
 
+if [ -z "${GCP_SERVICE_ACCOUNT:-}" ] && [ -z "${GCP_SA_KEY:-}" ]; then
```
mplakhtiy: **MEDIUM (part of the out-of-scope finding on `pr_triggers.py:254`)** — `GCP_SERVICE_ACCOUNT` moved out of `REQUIRED_VARS` and is now satisfied by `GCP_SA_KEY` instead.

`deploy-environment.yml` never exports `GCP_SA_KEY`, and `.github/workflows/e2e-gchat-test.yml:65` records that "GCP_SA_KEY is set at no scope" — so this branch is unreachable in this repository. It relaxes a required-variable gate in exchange for nothing.

Alongside it, the step rename from "Authenticate to GCP via Workload Identity Federation (WIF)" to "Authenticate to GCP" (`deploy-environment.yml:114`) reads like a fork-local accommodation.

.github/workflows/docker-publish-ghcr.yml
```diff
@@ -15,7 +16,7 @@ permissions:
 jobs:
   publish-ghcr:
     runs-on: ubuntu-latest
-    if: github.repository == 'gke-labs/kube-agents'
+    if: github.repository == 'gke-labs/kube-agents' || github.event_name == 'workflow_dispatch'
```
mplakhtiy: **MEDIUM (part of the out-of-scope finding on `pr_triggers.py:254`)** — the fork guard is widened with `|| github.event_name == 'workflow_dispatch'` here and in `docker-publish-k8s-operator.yml:19`.

This is probably safe: `IMAGE_REPOSITORY` is derived from `github.repository`, so a fork dispatch pushes to its own namespace and leaks nothing. But AGENTS.md states the guard rule for push-triggered credentialed workflows without this carve-out, so if it is wanted it wants a comment saying why and a mention in the PR body.

.github/workflows/docker-publish-ghcr.yml
```diff
@@ -85,6 +86,36 @@ jobs:
...
+        id: build-stockout-investigator
+        with:
+          context: agentplugins/gke-stockout-investigator
+          platforms: linux/amd64
+          push: true
+          tags: |
+            ghcr.io/${{ env.IMAGE_REPOSITORY }}/gke-stockout-investigator:latest
+            ghcr.io/${{ env.IMAGE_REPOSITORY }}/gke-stockout-investigator:${{ github.sha }}
+
+      - name: Mirror Third-Party Images to GHCR (cert-manager, litellm, fluent-bit)
```
mplakhtiy: **MEDIUM — this step publishes eleven third-party vendor images under the project's namespace on every push to `main`, and the body does not mention it.**

`MIRROR_PREFIX=ghcr.io/$IMAGE_REPOSITORY INCLUDE=third-party ./scripts/mirror_images.sh` copies all eleven third-party entries in `images.json` — cert-manager ×5, litellm, fluent-bit, `alpine/k8s`, the abcxyz token minter, and both Hindsight images — into `ghcr.io/gke-labs/kube-agents/*`. That is a redistribution decision.

Two mechanical problems on top of it:

1. `ubuntu-latest` carries neither `crane` nor `skopeo`, so `mirror_images.sh:118-137` falls through to the `docker` backend, whose own warning is that it "flattens multi-arch images to this host's architecture. That also changes their digests, so any reference pinned by digest … will not resolve against the copy" — and `hindsight-api` and `hindsight-postgresql` are both digest-pinned in `images.json`.
2. Four of the eleven come from `docker.io` with no Docker Hub login in this job, so the step is exposed to anonymous pull limits on the workflow that publishes every release image. (This second one is a **possible** problem, not one I confirmed.)

`INCLUDE=third-party` is a real `mirror_images.sh` input (`:47,80,234`) — the issue is the copy backend and the scope, not a bad flag.

agentplugins/pubsub-platform/install.sh
```diff
@@ -88,6 +88,17 @@ plugin_image_publish "$SCRIPT_DIR" "${SCRIPT_DIR}/files/platforms/pubsub"
...
+# update Helm ownership metadata so this standalone release can upgrade it without collision.
+if kubectl --context="$CONTEXT" get agentplugin pubsubplatform -n "$NAMESPACE" >/dev/null 2>&1; then
+    current_rel="$(kubectl --context="$CONTEXT" get agentplugin pubsubplatform -n "$NAMESPACE" -o jsonpath='{.metadata.annotations.meta\.helm\.sh/release-name}' 2>/dev/null || true)"
+    if [ -n "$current_rel" ] && [ "$current_rel" != "pubsubplatform" ]; then
+        echo "Adopting AgentPlugin pubsubplatform from release '${current_rel}' to 'pubsubplatform'..."
+        kubectl --context="$CONTEXT" annotate agentplugin pubsubplatform -n "$NAMESPACE" meta.helm.sh/release-name="pubsubplatform" --overwrite >/dev/null 2>&1 || true
+        kubectl --context="$CONTEXT" annotate agentplugin pubsubplatform -n "$NAMESPACE" meta.helm.sh/release-namespace="$NAMESPACE" --overwrite >/dev/null 2>&1 || true
+        kubectl --context="$CONTEXT" label agentplugin pubsubplatform -n "$NAMESPACE" app.kubernetes.io/managed-by="Helm" --overwrite >/dev/null 2>&1 || true
+    fi
+fi
```
mplakhtiy: **🔴 HIGH — two install engines now own the same `AgentPlugin`, and each steals ownership back from the other by hand.**

Commit `1193b170` deleted the #1013 stopgap `scripts/release/install_pubsub_platform.sh`; commit `6d850f4b` restored the script but left the documentation removals in place. At HEAD:

- `git ls-tree` shows the script at the same blob on `main` and on this branch — it is present, unchanged.
- `.github/workflows/e2e-run.yml:133` still runs it. That file is not in this diff.
- `tests/test_install_pubsub_platform.py` still asserts it is executable and ordered before `wait_for_gke_readiness.sh`.
- But the PR deletes its `scripts/release/README.md` bullet, rewrites the `common.sh:47-52` comment that named it, and rewrites `wait_for_gke_readiness.sh`'s header to "Infrastructure and plugins are deployed in the deploy step prior to waiting for readiness" — which is false for `e2e-manual-runner.yml`, which calls the script with no such step. Two references survive the deletion at `README.md:194` and `:323`, and `README.md:35` still says `release_resolve_target` "resolves the cluster the two kubectl-facing scripts below act on" when only one is now listed.

The runtime half is the more serious one. With the plugins enabled, `terraform apply` renders `AgentPlugin/pubsubplatform` owned by release `kube-agents`, pointing at `ghcr.io/gke-labs/kube-agents/pubsub-platform:<tag>`. Then `e2e-run.yml` runs the stopgap, which runs this script, whose **new hunk here** re-annotates the object to `meta.helm.sh/release-name: pubsubplatform` and applies the standalone chart's render — whose image is a content-tagged build pushed to the install's own Artifact Registry. The next `terraform apply` flips ownership back via the mirror-image hunk at `terraform/examples/full-install/lifecycle.sh:274-294`. Both directions are hard-coded, one in each engine, and neither release ever drops the object from its stored manifest — so `helm uninstall` on either side deletes a resource the other still believes it owns.

Consequence for this PR specifically: on the RC — the environment the Live validation section names — the plugin that actually runs is not the image this PR publishes.

AGENTS.md: "The install has one engine: Terraform + Helm … Do not add a second expression of an install step … The two places manifests still exist twice on purpose … each have a check or a comment saying so." Neither side of this pair carries such a note. #1013 is itself explicit that #1012's stopgap "deliberately does not close this: the RC now validates a configuration no user install produces, which is its own problem" — closing #1013 while that remains true is the mismatch.

Either resolution works; the current state does not. Delete the stopgap with its `e2e-run.yml` step and `tests/test_install_pubsub_platform.py` (what the script's own header and #1013 ask for, and what `1193b170` did) and restore nothing but the doc deletions — or keep the stopgap, revert the three doc deletions, and say why #1013 is not being closed.

*Verdict: CONFIRMED for the documentation contradiction and the two-owner design. The "RC runs the wrong image" consequence is derived from reading both installers, not observed on a cluster.*

terraform/examples/full-install/lifecycle.sh
```diff
@@ -250,9 +271,31 @@ adopt_kms() {
...
+        current_rel="$(kubectl get agentplugin "$plugin" -n "$namespace" -o jsonpath='{.metadata.annotations.meta\.helm\.sh/release-name}' 2>/dev/null || true)"
+        if [[ -n "$current_rel" && "$current_rel" != "kube-agents" ]]; then
+          log "adopting in-cluster AgentPlugin $plugin from release '$current_rel' into 'kube-agents'"
+          kubectl annotate agentplugin "$plugin" -n "$namespace" meta.helm.sh/release-name="kube-agents" --overwrite >/dev/null 2>&1 || true
+          kubectl annotate agentplugin "$plugin" -n "$namespace" meta.helm.sh/release-namespace="$namespace" --overwrite >/dev/null 2>&1 || true
+          kubectl label agentplugin "$plugin" -n "$namespace" app.kubernetes.io/managed-by="Helm" --overwrite >/dev/null 2>&1 || true
+        fi
+      fi
+    done
+  fi
```
mplakhtiy: **🟡 MEDIUM — this adoption block is silently skipped on zonal clusters, and it mutates the caller's kubeconfig.** Four problems in one block.

**Zonal clusters.** `location` was reassigned at `:198` with `sed -E 's/-[a-z]$//' <<<"$(tfvar location)"`, whose own comment says it does that because "KMS locations are regional". This block reuses that variable for `gcloud container clusters get-credentials --location "$location"`. On a zonal install (`location = us-central1-a`) the lookup asks for a regional cluster, fails, and — because everything is redirected to `/dev/null` — the whole block is skipped with no log line. The `helm_release.kube_agents` apply that follows then aborts with `invalid ownership metadata`, which is exactly the failure this block exists to prevent. Use `$(tfvar location)` here.

**Kubeconfig side effect.** `get-credentials` writes the cluster's context into the caller's kubeconfig and makes it current, and the three `kubectl` calls below carry no `--context`, so they depend on that. `agentplugins/*/install.sh` pass `--context="$CONTEXT"` for the same operations. Running `lifecycle.sh apply` now silently switches a developer's active kubectl context.

**Unconditional.** Every other target in `adopt_kms` is gated on its variable (`create_cluster`/`enable_database_encryption`, `enable_github_minter`, `enable_stockout_investigator`). This block runs on every apply, including installs that will never have a plugin.

**Accounting.** It sits inside a function called `adopt_kms`, and `$adopted` does not count it — so `log "resource adoption complete: $adopted imported"` reports `0` after an ownership adoption. Worth its own function.

*Verdict: CONFIRMED for all four; the zonal failure is read from the code, not observed.*

charts/kube-agents/values.yaml
```diff
@@ -624,3 +636,35 @@ platformAgent:
...
+plugins:
+  # Pub/Sub platform adapter AgentPlugin: opens the event listener for Cloud Pub/Sub subscriptions.
+  pubsubPlatform:
+    enabled: false
+    name: pubsubplatform
+    agentRef: platform-agent
+    image:
+      repository: ghcr.io/gke-labs/kube-agents/pubsub-platform
+      tag: "" # defaults to Chart.appVersion
+      pullPolicy: IfNotPresent
```
mplakhtiy, line 657: **🟡 MEDIUM — `pullPolicy` is a values key the template never reads.**

`plugins.pubsubPlatform.image.pullPolicy` (here) and `plugins.stockoutInvestigator.image.pullPolicy` (`:664`) are documented by their presence in `values.yaml`, but `templates/plugins.yaml` renders only `agentRef`, `image`, `targetProfile` and `config` — never `imagePullPolicy`.

It is not a field the CRD lacks: `charts/kube-agents/crds/kubeagents.x-k8s.io_agentplugins.yaml` lists `imagePullPolicy` among the six `spec` properties, and the operator honours it. Verified — `helm template … --set plugins.pubsubPlatform.image.pullPolicy=Always` produces a CR with no `imagePullPolicy` at all, so a user who sets it gets silence rather than an error.

Either render it in `templates/plugins.yaml:19` and `:65`, or drop the key from `values.yaml`. Unchanged from the previous review pass; no thread was opened on it.

charts/kube-agents/templates/plugins.yaml
```diff
@@ -0,0 +1,159 @@
...
+---
+apiVersion: kubeagents.x-k8s.io/v1alpha1
+kind: AgentPlugin
+metadata:
+  name: {{ .Values.plugins.pubsubPlatform.name | default "pubsubplatform" }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kube-agents.labels" . | nindent 4 }}
+spec:
+  agentRef: {{ .Values.plugins.pubsubPlatform.agentRef | default "platform-agent" | quote }}
```
mplakhtiy: **🟡 MEDIUM — `agentRef` defaults to a literal rather than to `platformAgent.name`.**

`default "platform-agent"` here and at `:64`, while the PlatformAgent this same chart renders is named `.Values.platformAgent.name` (`platform-agent-cr.yaml:5`, default `platform-agent`). The two agree today and the Terraform composition does not override the name, so nothing breaks now.

An install that sets `platformAgent.name` gets plugins pointing at an agent that does not exist — and per the comment at `agentplugins/gke-stockout-investigator/install.sh:288-291`, "the operator ignores a plugin whose agentRef names no PlatformAgent in the namespace". So the install reports success, the CR is created, and the plugin attaches to nothing. Defaulting to `.Values.platformAgent.name` makes the coupling explicit and removes the failure mode. Unchanged from the previous pass.

charts/kube-agents/templates/plugins.yaml
```diff
@@ -0,0 +1,159 @@
...
+        # Do not add `mcp-agent_common`: the `agent_common` server is no longer
+        # declared for any profile (see the note above `mcp_servers` in
+        # deploy/shared/defaults/config.yaml), so the grant would name a server that
+        # never starts. Delegation is kanban-only.
+        - mcp-router
+        - kanban
+        - memory
+        - hermes-cli
+        - hermes-api-server
+        - mcp-platform_control
```
mplakhtiy, line 49: **🟡 MEDIUM — this file is a second copy of both plugins' own chart templates, with nothing keeping the two in step.**

`templates/plugins.yaml:20-42` and `:68-159` render the same `AgentPlugin` spec as `agentplugins/pubsub-platform/templates/` and `agentplugins/gke-stockout-investigator/templates/agentplugin.yaml`. No check compares them, and no comment records that the duplication is deliberate.

The comments were largely carried across this time, which answers the "one copy says nothing" half of the earlier finding. Three blocks were not carried across, and each is a rationale a future editor of this copy would need:

- the "No `approvals:` block, deliberately" rationale (`agentplugins/gke-stockout-investigator/templates/agentplugin.yaml:22-42`)
- the "No `require_skills` here" note (`:65-71`)
- the comment on `sink:` explaining why the name has to come from the installer

AGENTS.md requires a duplicated manifest to carry "a check or a comment saying so" — the way the CRD/RBAC mirror has `make chart-check` and the kustomize manifests have their comment. This one has neither, so the next divergence is silent.

*Verdict: CONFIRMED as duplication.*

charts/kube-agents/templates/platform-agent-cr.yaml
```diff
@@ -37,6 +37,29 @@ spec:
...
+      {{- end }}
+      {{- with $tuning.platform }}
+      platform:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with $tuning.cluster }}
+      cluster:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- end }}
```
mplakhtiy: **🟡 MEDIUM — the auto-applied `harness.tuning` reaches every Cluster Agent, and the guards drop `0`.**

Applying tuning when the investigator is enabled is legitimate — it answers bot thread 23, and the plugin's own installer does patch these limits while the chart path did not. Three things about how it landed.

**It reaches the Cluster Agents, not just the platform profile.** Enabling one plugin sets `cluster.maxTurns: 150` and `cluster.apiMaxRetries: 8`, and the CRD's own documentation for that field says the overlay is "merged into each existing cluster profile at pod startup and into a new one when it is scaffolded". So every Cluster Agent in the fleet gets different execution limits because a Pub/Sub investigator was turned on. That may well be intended, but the only record of the decision is a comment in `values.yaml` — `charts/kube-agents/README.md` documents the chart's runtime knobs and says nothing about it.

**The guards drop `0`.** `:45-49` hand-rolls `if`/`with` instead of using `kube-agents.compactFields`, whose comment two blocks above in this same file says it exists precisely so that `false` and `0` survive while null and `""` do not. Verified — `--set platformAgent.harness.tuning.maxInProgress=0` with the plugin enabled renders `maxInProgress: 3`. The CRD's minimum is 1, so this is a convention break rather than a live defect today, but it is the failure the helper was written to prevent.

**`merge` mutates its first argument.** `{{- $tuning = merge $tuning $defaultStockoutTuning }}` at `:43` writes into `.Values.platformAgent.harness.tuning` whenever that key is non-empty. Harmless today because nothing else reads it afterwards; `deepCopy` before merging is the safe form.

*Verdict: CONFIRMED for the fleet-wide scope; the last two are convention findings, not live defects.*

scripts/release/common.sh
```diff
@@ -14,12 +14,14 @@ export DEFAULT_REGISTRY_PREFIX="ghcr.io/gke-labs/kube-agents"
...
 
-# Declarative registry of all 4 required container images
+# Declarative registry of all required release container images
 export REQUIRED_RELEASE_IMAGES=(
   "k8s-operator"
   "platform-agent"
   "credential-proxy"
   "replay-proxy"
+  "pubsub-platform"
+  "gke-stockout-investigator"
```
mplakhtiy, line 24: **🟡 MEDIUM — growing `REQUIRED_RELEASE_IMAGES` from four to six leaves three readers behind.**

The comment directly above this array **was** updated to say six. These were missed:

- `scripts/release/README.md:56` still says `verify_release_eligibility.sh` "verifies all 4 required container images", and `:37` still enumerates `verify_candidate_images.sh`'s set as "(`k8s-operator`, `platform-agent`, `credential-proxy`, `replay-proxy`)".
- `tests/testing/release.py:14` — `MOCK_REQUIRED_RELEASE_IMAGES` still lists four. Every release-script test (`test_release_common.py`, `test_promote_release_images.py`, `test_sign_release_images.py`) builds its registry fixture from that list, so promotion and signing are now exercised against a four-image world while the scripts iterate six. That is a test-fidelity gap, not just a doc one.

Two further docs are left behind by the additions elsewhere in this diff:

- **#1013 step 4 is not done** — "reconcile `agentplugins/README.md` and the two plugin READMEs to the new path". `agentplugins/README.md:19` still presents each plugin's `install.sh` as *the* way to install, and `:36` still states "Nothing here submits a Cloud Build", which `docker-publish-gcp.yml` now does for both plugins on every push to `main`. Two image models — a pinned release tag, and a content tag built on the installing host — now co-exist for the same plugin with nothing saying which applies when. The issue named that as "the design decision worth settling first".
- `terraform/examples/full-install/README.md` documents each optional feature (`enable_google_chat`, `enable_slack`, `enable_webhooks`) and gains nothing for the five new variables or the six GCP resources they create.

`make docs-check` passes on all of this — it checks generated regions, links, terminology, map coverage and the context budget, not whether prose still matches source.

*Verdict: CONFIRMED.*

mplakhtiy (commented):
Review of the full diff at `676bc69`, run in a context that did not write the change. Findings at MEDIUM and above are inline below; lower-severity items and notes on the PR description are omitted from this pass.

scripts/release/provision_environment.sh
```diff
@@ -149,6 +149,14 @@ if [ -n "${USER_PROFILE_ENABLED:-}" ]; then
   INSTALL_ARGS+=(--user-profile-enabled="${USER_PROFILE_ENABLED}")
 fi
 
+if [ "${ENABLE_PUBSUB_PLATFORM:-false}" = "true" ]; then
+  INSTALL_ARGS+=(--enable-pubsub-platform)
+fi
+
+if [ "${ENABLE_STOCKOUT_INVESTIGATOR:-false}" = "true" ]; then
+  INSTALL_ARGS+=(--enable-stockout-investigator)
+fi
```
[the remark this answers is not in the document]
sergiu1spiridon: scripts/release/install_pubsub_platform.sh was an explicit temporary stopgap until #1013 added native Terraform & Helm support for AgentPlugins (as noted in that script's header).

To make the environment requirements clear and prevent any gaps on nightly or rc:

Updated scripts/release/README.md to explicitly document ENABLE_PUBSUB_PLATFORM and ENABLE_STOCKOUT_INVESTIGATOR in both the required environment variables list and the nightly matrix integrations table.
Noted in the PR description that ENABLE_PUBSUB_PLATFORM=true and ENABLE_STOCKOUT_INVESTIGATOR=true should be configured on the nightly and rc GitHub environments.
The standalone plugin installer scripts (agentplugins/{pubsub-platform,gke-stockout-investigator}/install.sh) remain intact for direct/ad-hoc installations.

k8s-operator/internal/controller/platformagent_manifests.go
```diff
@@ -964,6 +969,36 @@ func pluginMountPath(homeDir string, plugin *agentv1alpha1.AgentPlugin) string {
...
+	stageScript := fmt.Sprintf("mkdir -p %s && (if [ -d /files ]; then cp -a /files/. %s/; else for item in /*; do case \"$item\" in /bin|/boot|/dev|/etc|/home|/lib*|/media|/mnt|/opt|/proc|/root|/run|/sbin|/srv|/sys|/tmp|/usr|/var) ;; *) cp -a \"$item\" %s/ ;; esac; done; fi); true",
+		mountPath, mountPath, mountPath)
+
+	return corev1.Container{
+		Name:            buildPluginStagingContainerName(plugin.Name),
+		Image:           plugin.Spec.Image,
+		ImagePullPolicy: pullPolicy,
+		SecurityContext: hardenedSecurityContext(),
+		Command: []string{
+			"/bin/sh",
```
mplakhtiy, line 1019: **🔴 Blocker — the staging init container runs `/bin/sh` from the plugin's own image, and it is now the default on every GKE cluster, so an existing `FROM scratch` plugin image turns the operator upgrade into an agent outage.**

Three changes compose into this:

- `platformagent_controller.go:2260-2268` — `clusterImageVolumeSupport` returns `(false, true)` whenever the API server's `GitVersion` contains `-gke`, **before** the version check. The comment names GKE Warden's `autopilot-volume-type-limitation`, which is an Autopilot policy, but the test is on the string `-gke`, and `v1.31.5-gke.1023000` matches as readily as an Autopilot version does. Every GKE install — Standard included — now takes the fallback path unless someone sets the `kubeagents.x-k8s.io/enable-image-volumes` annotation.
- This line — `Command: ["/bin/sh", "-c", stageScript]`, executed **inside the plugin's own image**.
- `manifest_helpers.go:428-431` — the gateway Deployment defaults to `Replicas: 1` and `Strategy: Recreate`.

A plugin image with no `/bin/sh` produces an init container the runtime cannot start; an init container that never completes blocks the Pod; and `Recreate` has already deleted the old one. Before this change the same cluster left the plugin unmounted and kept the agent running — the fail-closed reasoning still written at `platformagent_controller.go:2211-2216` ("leaving its plugins unloaded" is the *lesser* failure) is exactly inverted by it.

Who has such an image: everyone. `agentplugins/lib/plugin_image.sh:4` states the contract — "Every plugin image in this directory is `FROM scratch` plus a COPY of one tree" — and every plugin image published before this PR was built that way. The RC and nightly environments both have one, since `tests/e2e/test_stockout_investigation.py:663-690` runs `install.sh` whenever the AgentPlugin is absent. Nothing gates the init container on the plugin having been rebuilt.

`kube-agents-bot` raised this on 2026-09-01 against `platformagent_manifests.go:992` with the same mechanism. `b3e6952e` reverted the init container in response; `869e76c2` re-landed it. That thread is still open and has no reply.

agentplugins/pubsub-platform/Dockerfile
```diff
@@ -1,2 +1,3 @@
-FROM scratch
+FROM busybox:musl
```
mplakhtiy: **🔴 Blocker — `FROM busybox:musl` breaks the plugin-image contract, and the crane builder still produces the `FROM scratch` image under the same content tag.**

The two builders in `agentplugins/lib/plugin_image.sh` are not equivalent under a base image:

- `plugin_image_docker_publish` (`:861-868`) runs `docker build -f "${ctx}/Dockerfile"`, so it honours the busybox base and both `COPY` lines.
- `plugin_image_crane_publish` (`:871-906`) is "No daemon, no Dockerfile" — it tars `src_dir` and runs `crane append -f "$layer" -t "$staging"` with **no `-b` base**, then `crane mutate`. It produces a single-layer image holding the plugin tree at `/` and nothing else: no busybox userland, no `/bin/sh`, no `/files`.
- `plugin_image_builder` (`:505-539`) picks crane "whenever no Docker daemon answers", described there as "the common case on a laptop" — and `agentplugins/README.md` now presents the standalone installer as the recommended developer workflow.

So the crane path publishes exactly the image the init container above cannot start, and it publishes it under the same tag the docker path would use for a working one. `plugin_image_build_digest` hashes the Dockerfile, so the tag moves once — but it cannot distinguish the two builders, and `plugin_image_publish` skips a tag that already exists, so whichever builder reaches a given content tag first defines that image permanently. The file's own comments say this in as many words (`:81` "so the crane layer and `docker build` cannot ship different files"; `:113-122` on the silent-divergence failure mode the single-reader design exists to prevent). Neither comment nor `agentplugins/README.md:35` was updated.

This finding was already raised on this branch against both Dockerfiles when the base was `alpine:3.24`, with the same two citations. Both threads are open, neither has a reply, and the change returned as `busybox:musl`.

agentplugins/pubsub-platform/install.sh
```diff
@@ -88,6 +88,16 @@ plugin_image_publish "$SCRIPT_DIR" "${SCRIPT_DIR}/files/platforms/pubsub"
 # Deploy the chart directly. This used to call scripts/deploy_extension.sh, which is not
 # in this repository — the installer could never have run as written.
 echo "Deploying pubsub-platform extension via Helm chart..."
+# If the AgentPlugin is already deployed as part of the main kube-agents release,
```
mplakhtiy, line 91: **🟠 High — nothing adopts an AgentPlugin already owned by the standalone release, so the Helm upgrade fails on exactly the migration this PR recommends.**

Ownership is handled in one direction only. This block, and `agentplugins/gke-stockout-investigator/install.sh:294-323`, check `meta.helm.sh/release-name` and skip when the object already belongs to `kube-agents`. Nothing does the reverse: when the AgentPlugin exists and belongs to the standalone release (`pubsubplatform` / `gkestockoutinvestigator`), `helm_release.kube_agents` refuses it with `invalid ownership metadata`.

The `lifecycle.sh` block that re-annotated the object was added in `6d850f4b` and removed again in `6748bb64`. The removal correctly disposes of the earlier finding about its zonal `get-credentials` bug and its kubeconfig side effect, but it leaves the collision that block existed to prevent.

The affected population is not hypothetical: it is every install that followed `agentplugins/*/install.sh` — the only supported path until this PR — and then sets `enable_pubsub_platform = true`, which is precisely the migration `agentplugins/README.md:19-27` now recommends. The RC and nightly environments are in it.

Either direction is a defensible design; the gap is that the upgrade path is unhandled and undocumented. Neither `terraform/examples/full-install/README.md` nor `agentplugins/README.md` tells an operator to `helm uninstall pubsubplatform` first.

scripts/release/provision_environment.sh
```diff
@@ -149,6 +149,14 @@ if [ -n "${USER_PROFILE_ENABLED:-}" ]; then
   INSTALL_ARGS+=(--user-profile-enabled="${USER_PROFILE_ENABLED}")
 fi
 
+if [ "${ENABLE_PUBSUB_PLATFORM:-false}" = "true" ]; then
```
mplakhtiy, line 161: **🟠 High — the nightly's blocking stockout suite loses its ingress guardrail and now depends on two GitHub environment variables this PR cannot set.**

Deleting the `#1013` stopgap is right, and its `e2e-run.yml` step had to go with it. What went with them is the guardrail: `alert_ingress_required: true` existed so that a missing adapter failed at the install, "where the error names the cause", rather than 360 seconds later as an agent that never started an investigation. The deleted comment says so explicitly.

The replacement is `vars.ENABLE_PUBSUB_PLATFORM` / `vars.ENABLE_STOCKOUT_INVESTIGATOR` on the `rc` and `nightly` GitHub environments, which `scripts/release/README.md:194` now documents as "Required". Those are repository settings, not code — this PR cannot set them, and nothing in CI checks that they are. If they are unset, `deploy-environment.yml` passes an empty string, this test takes the else branch, no plugin is installed, and the nightly's **blocking** `test_stockout_investigation.py` fails at its timeout with the misleading symptom the deleted comment describes. It is silent in the other direction too: nothing warns that the ingress the suite needs was not installed.

Two things would close it: set the variables before merge and say so in the PR body, and have `validate_and_log_deploy_summary.sh` — which already echoes both values — fail rather than print `<DEFAULT (false)>` when the environment is one whose blocking suite reads an alert.

charts/kube-agents/values.yaml
```diff
@@ -631,3 +643,35 @@ platformAgent:
...
       # Render responses with Microsoft Adaptive Cards v1.5 (markdown fallback if false).
       adaptiveCards: true
+
+# AgentPlugins deployed into the cluster.
+plugins:
+  # Pub/Sub platform adapter AgentPlugin: opens the event listener for Cloud Pub/Sub subscriptions.
+  pubsubPlatform:
+    enabled: false
+    name: pubsubplatform
+    agentRef: platform-agent
```
mplakhtiy: **🟠 High — `agentRef` still resolves to the literal `platform-agent`, so a custom `platformAgent.name` produces an orphan plugin.**

`charts/kube-agents/templates/plugins.yaml:16` and `:46` were changed to `.Values.plugins.*.agentRef | default .Values.platformAgent.name`, which answers the earlier finding — except `values.yaml` still ships `agentRef: platform-agent` as a concrete default under both plugins, so the `default` pipeline never fires. Rendered:

    $ helm template t charts/kube-agents --set platformAgent.name=my-agent \
        --set plugins.pubsubPlatform.enabled=true
    kind: PlatformAgent
    metadata:
      name: my-agent
    ...
    kind: AgentPlugin
    spec:
      agentRef: "platform-agent"

The consequence is the one `agentplugins/gke-stockout-investigator/install.sh:288-291` spells out: "the operator ignores a plugin whose agentRef names no PlatformAgent in the namespace". The install reports success, the CR is created, the plugin attaches to nothing, and the only symptom is that no alert is ever investigated.

The same line needs the same fix at `:664` for `stockoutInvestigator`. The open thread on this is against the template, so a reviewer reading the template alone would conclude it was already fixed.

```suggestion
    agentRef: ""
```

k8s-operator/internal/controller/platformagent_manifests.go
```diff
@@ -964,6 +969,36 @@ func pluginMountPath(homeDir string, plugin *agentv1alpha1.AgentPlugin) string {
...
 
+// buildPluginStagingInitContainer builds an init container that extracts a plugin's container image
+// into an emptyDir volume on clusters where ImageVolumeSource is unsupported or restricted (e.g. GKE Autopilot).
+func buildPluginStagingInitContainer(homeDir string, plugin *agentv1alpha1.AgentPlugin) corev1.Container {
+	mountPath := pluginMountPath(homeDir, plugin)
+	pullPolicy := corev1.PullIfNotPresent
+	if plugin.Spec.ImagePullPolicy != nil {
+		pullPolicy = *plugin.Spec.ImagePullPolicy
+	}
+	stageScript := fmt.Sprintf("mkdir -p %s && (if [ -d /files ]; then cp -a /files/. %s/; else for item in /*; do case \"$item\" in /bin|/boot|/dev|/etc|/home|/lib*|/media|/mnt|/opt|/proc|/root|/run|/sbin|/srv|/sys|/tmp|/usr|/var) ;; *) cp -a \"$item\" %s/ ;; esac; done; fi); true",
```
mplakhtiy: **🟠 Medium — the staging script swallows every error, and the operator reports the plugin `Ready` regardless.**

The stage script ends `...); true`. Any failure inside the subshell — the `cp -a`, a full emptyDir, an unexpected image layout — is discarded, and the init container exits 0 with an empty plugin directory.

`evaluatePluginReadiness` (`platformagent_controller.go:2319-2334`) lost its `!imageVolumeSupported` → `Degraded` case in the same commit and now sets `Ready` with the message "Plugin successfully staged via init container". Nothing observes whether anything was actually staged. So a plugin that staged nothing reports `Ready`, and the agent comes up with an empty plugin directory — the "installs and reports healthy while doing nothing" failure mode `#1013` was filed about, reintroduced one layer down.

Dropping the `; true` and letting a failed copy fail the init container would at least surface it, at the cost of turning it into the outage in the blocker above. A `[ -n "$(ls -A $M)" ]` assertion plus a real exit code, and a condition reason the operator can set from the container's status, would surface it without that.

agentplugins/gke-stockout-investigator/Dockerfile
```diff
@@ -1,2 +1,3 @@
-FROM scratch
+FROM busybox:musl
```
mplakhtiy: **🟠 Medium — `busybox:musl` is an unpinned mutable base tag with no `images.json` entry.**

Both plugin images are now built on a floating Docker Hub tag. `docker-publish-ghcr.yml` and `docker-publish-gcp.yml` resolve it afresh on every push to `main`, so what the two release images contain changes without any commit in this repository.

AGENTS.md holds third-party GitHub Actions to a 40-character SHA for exactly this reason, and `images.json` is described as the inventory of "every container image an install pulls, with its upstream reference and pin". Whether a base image baked into a first-party image falls inside that wording is arguable — which is itself worth settling rather than leaving implicit.

Pin the base by digest (`busybox:musl@sha256:...`) at minimum, and say in `agentplugins/README.md` what the base is and why.

mplakhtiy (commented):
Adversarial review of the full diff at `8420fde`, run in a subagent that did not write the change. Posting the MEDIUM-and-above findings: 2 blocker, 3 high, 4 medium. Lower-severity items are held back.

Two of the blockers are changes that were each rejected once already on this branch and re-landed in the four commits after the last `/review` at 2026-09-01T16:17:22Z, which no reviewer has read.

#### 🟠 Medium — the AgentPlugin documentation describes behaviour the operator no longer has

`docs/site/src/content/docs/operator/agentplugin-crd.md:77`, `:79`, `:95`, `agentplugins/README.md:35`

- `:77` states that on clusters where OCI image volumes are unsupported "the operator skips OCI volume attachment ... The plugin status is updated to `Phase: Degraded` with condition `Reason: ImageVolumeUnsupported`." The operator now attaches an emptyDir and an init container, and `ImageVolumeUnsupported` was deleted from `evaluatePluginReadiness` — the string no longer exists in the Go source.
- `:95` — the condition-reason table still lists `ImageVolumeUnsupported`, a reason nothing can emit.
- Nothing records the three facts a plugin author now needs: that GKE takes the fallback path by default, that the fallback requires `/bin/sh` in the plugin image, and that the `kubeagents.x-k8s.io/enable-image-volumes` annotation is the way back to native image volumes.
- `agentplugins/README.md:35` — "A plugin image is `FROM scratch` plus a COPY of one directory, so the whole build is a single tar layer" is now false for both shipped plugins.

`make docs-check` passes on all of it, because it checks generated regions, links, terminology, map coverage and the context budget — not whether prose matches source.

agentplugins/pubsub-platform/Dockerfile
```diff
@@ -1,2 +1,3 @@
...
 COPY files/platforms/pubsub/ /
+COPY files/platforms/pubsub/ /files/
```
mplakhtiy: **🟠 Medium — on the ImageVolume path the plugin mount now carries the whole busybox rootfs.**

An `ImageVolumeSource` mounts the image's entire root filesystem at `$PLATFORM_AGENT_HOME/plugins/<name>` (or `/opt/agent-plugins/<profile>/<name>`). With a busybox base that is `bin/`, `etc/`, `lib/`, `usr/`, `var/`, a second copy of the plugin tree under `files/`, and a shell — all mounted into the agent container's plugin directory.

The emptyDir path built at `platformagent_manifests.go:2139-2144` copies only `/files`, so the two staging mechanisms now put materially different trees at the same path, and only one of them is what Hermes' plugin loader was written against. Non-GKE 1.35+ clusters and any install with `enable-image-volumes: "true"` take the ImageVolume path, so it is reachable rather than dead.

Related: this file carries both `COPY files/ /`-style copies, shipping the tree twice — the `/files/` copy is what the staging script reads, the root copy is what an ImageVolume mount exposes. Whether the extra top-level directories confuse `deploy/shared/profile_plugins.py` was not tested.

scripts/release/provision_environment.sh
sergiu1spiridon: We are adding these variables in the RC environment outside this PR

k8s-operator/internal/controller/platformagent_manifests.go
sergiu1spiridon: Scoped the `initContainer` plugin staging fallback strictly to GKE Autopilot clusters, leaving GKE Standard clusters on native Kubernetes `ImageVolumeSource`.

### What changed and why:

1. **GKE Standard uses native `ImageVolumeSource`:**
   The previous `strings.Contains(ver.GitVersion, "-gke")` check was overly broad because it matched GKE Standard alongside Autopilot. In `k8s-operator/internal/controller/platformagent_controller.go`, `clusterImageVolumeSupport` now checks for Autopilot-specific API groups (`auto.gke.io` / `warden.gke.io`). 
   On GKE Standard (and upstream K8s >= 1.35), `clusterImageVolumeSupport` returns `supported = true`. No init container is injected, native `ImageVolumeSource` mounts the image directly, and existing `FROM scratch` plugin images continue to work without executing `/bin/sh` or risking container startup failures.

2. **GKE Autopilot uses the `emptyDir` staging fallback:**
   On GKE Autopilot, GKE Warden's `autopilot-volume-type-limitation` policy rejects the `image` volume type at admission time. For Autopilot clusters, the operator falls back to staging the plugin via an init container into `emptyDir`.

3. **Plugin image base:**
   The plugin images in this repository (`agentplugins/gke-stockout-investigator` and `agentplugins/pubsub-platform`) are built `FROM busybox:musl`, providing `/bin/sh` and `/files/` so the same image functions under both models:
   - On **GKE Standard**, containerd mounts the image layers directly as a read-only volume (the shell is never executed).
   - On **GKE Autopilot**, the init container runs `/bin/sh` to copy the plugin assets into the `emptyDir`.

4. **Tests and docs:**
   - Updated `agentplugins/lib/plugin_image.sh` to document the `FROM busybox:musl` base for volume-restricted environments.
   - Added unit test cases in `platformagent_controller_test.go` verifying that GKE Standard >= 1.35 resolves `imageVolumeSupported = true` while GKE Autopilot resolves `imageVolumeSupported = false`.

agentplugins/pubsub-platform/Dockerfile
sergiu1spiridon: Resolved: aligned the `crane` builder with `docker build` to honour base images and staging paths, and bumped the recipe counter to invalidate previous tags.

### What changed:

1. **`crane append` honours the base image:**
   In `agentplugins/lib/plugin_image.sh:plugin_image_crane_publish`, the builder now parses the `FROM <image>` line from the plugin's `Dockerfile`. When a base image is specified (such as `busybox:musl`), `crane append` passes `-b "$base_image"`. For `FROM scratch` images, it continues to append to an empty base without `-b`.

2. **Filesystem layout parity:**
   If the `Dockerfile` contains a `COPY ... /files/` instruction, `plugin_image_crane_publish` stages the files into `$root/files/` in addition to `$root/`, ensuring the crane-assembled tarball contains both paths identically to `docker build`.

3. **Recipe counter bumped to 5 (`PLUGIN_IMAGE_RECIPE=5`):**
   Bumping the recipe moves the content tag digest, ensuring any previous single-layer `scratch` image published under recipe 4 is not reused.

4. **Tests and documentation:**
   - Updated `agentplugins/README.md` and the recipe history in `plugin_image.sh`.
   - Added unit tests in `agentplugins/lib/tests/test_plugin_image.py` asserting that `crane append` receives `-b busybox:musl` when `FROM busybox:musl` is in the `Dockerfile`, and omits `-b` when `FROM scratch` is used.
   - All 81 unit tests in `test_plugin_image.py` pass.

agentplugins/pubsub-platform/install.sh
sergiu1spiridon: Resolved: documented the migration requirement in both `agentplugins/README.md` and `terraform/examples/full-install/README.md`.

### Summary:
Because Helm strictly enforces resource ownership (`meta.helm.sh/release-name`), `helm_release.kube_agents` cannot adopt an `AgentPlugin` custom resource that is already owned by an existing standalone release (`pubsubplatform` or `gkestockoutinvestigator`). 

Rather than attempting implicit in-cluster resource mutations during `lifecycle.sh` (which caused side-effects on kubeconfig and zonal cluster credentials), the migration requirement is now explicitly documented:
- In `agentplugins/README.md § Installing`: Added a migration note instructing operators to run `helm uninstall pubsubplatform -n <namespace>` / `helm uninstall gkestockoutinvestigator -n <namespace>` before enabling declarative plugins in Terraform.
- In `terraform/examples/full-install/README.md § Pub/Sub Platform and Stockout Investigator Plugins`: Documented the uninstall step required prior to setting `enable_pubsub_platform = true` or `enable_stockout_investigator = true` on existing clusters that previously ran standalone installs.

charts/kube-agents/values.yaml
sergiu1spiridon: Resolved: changed default `agentRef` to `""` in `charts/kube-agents/values.yaml` under both plugins and added unit tests.

### What changed:
1. In `charts/kube-agents/values.yaml`, set `agentRef: ""` for both `pubsubPlatform` and `stockoutInvestigator`.
2. This allows the template's `.Values.plugins.*.agentRef | default .Values.platformAgent.name` fallback pipeline to evaluate `.Values.platformAgent.name` when `agentRef` is unset, while retaining `"platform-agent"` on standard installs and correctly inheriting custom `platformAgent.name` overrides.
3. Added `test_plugins_inherit_custom_platform_agent_name` to `tests/test_helm_plugins.py` verifying that custom agent names propagate to rendered `AgentPlugin` manifests.

k8s-operator/internal/controller/platformagent_manifests.go
sergiu1spiridon: Resolved: removed error suppression from the staging script, added a non-empty check, and surfaced init container staging failures in the AgentPlugin status condition.

### What changed:
1. **Staging script assertion (`platformagent_manifests.go`):**
   - Dropped the trailing `; true` so copy errors fail the init container.
   - Added `&& [ -n "$(ls -A <mountPath>)" ]` to the stage command; if an image layout produces an empty plugin directory, the command fails and the container terminates with non-zero exit code.

2. **Failure observation & condition update (`platformagent_controller.go`):**
   - Updated controller failure detection (`detectPluginFailures`) to inspect `pod.Status.InitContainerStatuses` for the staging init container (`stage-plugin-<name>`).
   - If the container terminates with a non-zero exit code or enters `CrashLoopBackOff`, the operator marks the `AgentPlugin` status as `Phase: "Degraded"` with `Reason: "StagingFailed"` and includes the exit code/message in the condition.

3. **Tests:**
   - Added `TestBuildPluginStagingInitContainer_AssertsNonEmptyAndFailsOnErrors` in `platformagent_manifests_test.go` asserting that `; true` is omitted and the non-empty check is present.
   - Added `TestUpdatePluginStatuses_StagingFailureIsReported` in `platformagent_controller_test.go` verifying that staging container failures mark the plugin `Degraded` with `Reason: "StagingFailed"`.

agentplugins/gke-stockout-investigator/Dockerfile
sergiu1spiridon: Resolved: pinned `busybox:musl` by its immutable multi-arch digest, added it to `images.json` under build-time base images, and documented its role in `agentplugins/README.md`.

### What changed:
1. **Pinned base image by digest:**
   In both `agentplugins/gke-stockout-investigator/Dockerfile` and `agentplugins/pubsub-platform/Dockerfile`:
   ```dockerfile
   FROM busybox:musl@sha256:32b5cdad7cce41dfd53d0ae06baebcf8357a147ee7694dc706911c373bc30c37

agentplugins/pubsub-platform/Dockerfile
sergiu1spiridon: Resolved: removed the duplicate `COPY ... /files/` from both Dockerfiles and verified `profile_plugins.py` and Hermes plugin discovery behavior when a base userland is present.

### Details:
1. **Single COPY in Dockerfiles:**
   Removed `COPY ... /files/` from `agentplugins/gke-stockout-investigator/Dockerfile` and `agentplugins/pubsub-platform/Dockerfile`. Files are now copied only once to `/`. On the init container staging path, the fallback loop in `platformagent_manifests.go` filters out rootfs directories (`/bin`, `/etc`, `/lib*`, `/usr`, etc.) and copies the plugin files into the target `emptyDir`.

2. **ImageVolume compatibility & `profile_plugins.py`:**
   - Added `test_plugin_with_busybox_rootfs_links_normally` to `tests/test_profile_plugins.py` confirming that `profile_plugins.py` symlinks the mounted plugin directory wholesale into `$HERMES_HOME/profiles/<profile>/plugins/<plugin>` without traversing or choking on rootfs directories (`bin/`, `etc/`).
   - Hermes resolves plugins by checking for `<plugin>/plugin.yaml` inside each directory under `plugins/`. This matches the existing E2E test plugin harness (`tests/e2e/operator/templates/plugin_src/Dockerfile`), which has been built `FROM alpine:3.24` + `COPY . /` and loads plugins and skills without collision.

mplakhtiy (approved):

PR #608, opened 2026-08-10
feat(inference): support custom OpenAI-compatible endpoints & standalone Gemma 4 vLLM recipe

fkc1e100:
## Summary

- **Generic Custom OpenAI-Compatible Provider Support (Addressing @mastersingh24):**
  - Kube-agents does not deploy models or model servers. In-tree model server kustomize overlay (`k8s-operator/config/integrations/vllm-gemma/`) and imperative GPU scripts have been completely removed.
  - Added first-class generic `model_provider: custom` (OpenAI-compatible) endpoint support across operator, Helm chart, installer, and Terraform (`variables.tf`, `main.tf`, `terraform.tfvars.example`), parameterized via `CUSTOM_API_BASE` and `CUSTOM_API_KEY`.
  - Added NetworkPolicy egress rules in the chart covering standard HTTP/HTTPS/service ports (8000, 8080, 80, 443) to both cluster-internal and external endpoints.
  - Provided `k8s-operator/config/integrations/litellm/overlays/custom/` overlay for dev/kustomize workflows.
- **Consolidated Standalone Reference Recipe in `examples/vllm-gemma/`:**
  - Standalone, unmanaged reference recipe for Gemma 4 on GKE with vLLM.
  - Container `securityContext` dropping all capabilities, digest pin, L4 `nodeSelector`, and native `bfloat16` (`--tensor-parallel-size=2`).
  - Sized for **Gemma 4-27B** (`google/gemma-4-27B-it`) and **Gemma 4-31B** (`google/gemma-4-31B-it`), explicitly documenting why smaller models (2B/4B/9B) lack the reasoning capacity for Kubernetes agentic operations.
  - Both declarative Terraform (`google_container_node_pool`) and 2-line imperative `gcloud` create/delete commands documented for testing.
- **GPU Architecture & Teardown Story (Addressing @mplakhtiy):**
  - Documented GPU node pool creation (`gcloud container node-pools create`) and deletion (`gcloud container node-pools delete`) commands in `docs/site/src/content/docs/concepts/inference-gateway.md` and `examples/vllm-gemma/README.md`.
[...]

PR #1090, opened 2026-09-01
feat(ci): assert static NetworkPolicy DNS egress parity across all copies

### LINKED ISSUE #747: NetworkPolicy: a parity guard for the eight static copies, and operator-owned litellm-policy (remainder of #676/#687)

shalinibhatiapl:
## Summary

Adds an automated parity checker (`scripts/check_iac_parity.py`), unit test suite (`scripts/test_check_iac_parity.py`), and Makefile target (`iac-parity-check`) to assert that all static NetworkPolicy copies in the repository maintain the required DNS egress peer shape (`10.96.0.10/32` literal plus `0.0.0.0/0` with RFC 1918 exceptions). Also adds repository-wide discovery to ensure no newly added DNS-bearing NetworkPolicy can escape the static roster.

## Why This Change

Eight static manifests hand-maintain port 53 DNS egress rules. In #687, an update had to touch all copies and initially missed `deployment.yaml.template` because standard `*.yaml` searches skipped it. In #608, a static policy was added omitting required DNS peers, which causes silent DNS resolution failures on GKE clusters where `kube-dns` is assigned a non-standard or public Service CIDR VIP (e.g. `34.118.224.0/20`). Nothing previously checked that static copies stay aligned or that new copies do not drift.

## What Changed

- Implemented `scripts/check_iac_parity.py` to parse static NetworkPolicy manifests, sanitize Go/Helm template tags offline, and verify that every port 53 egress rule contains `10.96.0.10/32` and an `0.0.0.0/0` peer with at least RFC 1918 private subnets (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`) in its `except` list. Accommodates the 5-entry house shape in `networkpolicy-core-egress.yaml`.
- Added `discover_dns_network_policies()` in `scripts/check_iac_parity.py` and `test_discovery_matches_roster_exactly` in `scripts/test_check_iac_parity.py` to walk the repository tree and assert that any manifest defining a port 53 egress rule is tracked in `STATIC_NETWORK_POLICIES` or explicitly listed with a reason in `EXCLUDED_NETPOL_MANIFESTS`.
- Added `make iac-parity-check` target in `Makefile` with a PyYAML preflight check for local developer convenience.
- Automated CI enforcement runs via `scripts/test_check_iac_parity.py` under `make test-python` (`.github/workflows/python-tests.yml`), leaving `.github/workflows/` untouched and avoiding runner environment package installation hazards.
- Added 66 unit tests in `scripts/test_check_iac_parity.py` covering live production manifests, discovery drift across all 6 extensions (`.yaml`, `.yml`, `.yaml.template`, `.yml.template`, `.yaml.tmpl`, `.yml.tmpl`), synthetic peer mutations, string port handling, multi-doc ingress skipping (omitted, empty, and non-empty egress), protocol TCP/UDP validation, independent wildcard exception validation, Helm template trailing comments/inline conditionals, link-local DNS literals, defensive scalar error handling without tracebacks, and CLI invocation modes.

## Context

Closes #747 (Item B5 and D1).
Addresses code review feedback from review pullrequestreview-5076622375.

## Testing

- Ran `python3 -m unittest scripts/test_check_iac_parity.py -v`: 66/66 tests pass in 0.591s.
- Ran `make iac-parity-check`: verified exit code 0 (`OK: Verified DNS egress rule parity across 8 static copies (8 rules checked).`).
- Ran `make docs-check`: all documentation checks, links, and context budgets pass with 0 errors.

### Live validation

Conducted live local mutation testing to verify that real-world drift is blocked:
1. **Classic DNS literal omission:**
   - Removed `10.96.0.10/32` from `deploy/kustomize/platform/networkpolicy-core-egress.yaml` -> `make iac-parity-check` failed immediately with exit code 1 and reported missing literal `10.96.0.10/32`. Restored file -> passed with exit code 0 (`OK`).
2. **RFC 1918 exception omission:**
   - Removed `10.0.0.0/8` from `k8s-operator/config/integrations/litellm/base/networkpolicy.yaml` -> `make iac-parity-check` failed immediately with exit code 1 and reported missing private CIDR `10.0.0.0/8`. Restored file -> passed with exit code 0 (`OK`).
3. **CI Execution:**
   - Executed under `python-tests.yml` via `make test-python` on every pull request, where `scripts/test_check_iac_parity.py` validates all live production static manifests and discovery completeness.

## Self-Review

Ran required pre-PR passes in isolated contexts and addressed subsequent review feedback:
- **`review-adversarial` pass (isolated subagent)**:
  - *Angle A (`IntOrString` port sensitivity)*: Identified that `ports[].port` specified as a string `"53"` would evaluate `!= 53`. **Disposition: Fixed.** Updated port matcher to accept `port in (53, "53")`.
  - *Angle C (Dead constant)*: Identified unused `HOUSE_SHAPE_EXCEPT` constant. **Disposition: Fixed.** Exported `is_house_shape()` helper and added test coverage.
  - *Angle A (Multi-doc Ingress policies)*: Flagged that an Ingress-only policy in a multi-document file would report missing port 53. **Disposition: Fixed.** Filtered out policies that do not define egress.
- **Review `pullrequestreview-5076622375` & follow-up passes**:
  - *MEDIUM (Hardcoded roster with no discovery-drift check)*: Flagged that a future unlisted manifest with port 53 would escape the guard. **Disposition: Fixed.** Implemented `discover_dns_network_policies()` and `test_discovery_matches_roster_exactly`.
  - *LOW (PEP-668 runner hazard in validate.yml)*: Flagged that `pip install` on Ubuntu 24.04 could fail without `--break-system-packages`. **Disposition: Fixed.** Reverted edits to `.github/workflows/validate.yml` completely. CI enforcement is handled purely through `python-tests.yml` via `scripts/test_check_iac_parity.py`.
  - *LOW (Whole-file parse coupling)*: Flagged that unparseable non-NetworkPolicy templates in multi-doc manifests could fail NetworkPolicy extraction. **Disposition: Fixed.** Refactored `load_network_policies()` to isolate and parse `kind: NetworkPolicy` chunks per-document.
  - *HIGH/MEDIUM (Wildcard exception pooling across multiple peers)*: Flagged that multiple `0.0.0.0/0` peers pooled exceptions, allowing split private CIDR exclusions to bypass validation. **Disposition: Fixed.** Validated every wildcard peer independently and rejected multiple `0.0.0.0/0` peers per rule.
  - *MEDIUM (Helm template trailing comments & inline controls)*: Flagged that trailing comments on control directives caused PyYAML parser errors. **Disposition: Fixed.** Stripped control tags before YAML parsing while preserving inline YAML payloads.
  - *MEDIUM (Ingress-only policy with explicit empty `egress: []`)*: Flagged that policies with `policyTypes: [Ingress]` and `egress: []` falsely failed DNS check. **Disposition: Fixed.** Differentiated explicit `policyTypes` from inferred types.
  - *LOW (Live validation exit code mismatch)*: Corrected live validation record to state exit code 1.
  - *LOW/TEST (Discovery extension table & Ingress coverage)*: Added table-driven discovery test covering `.yaml`, `.yml`, `.yaml.template`, `.yml.template`, `.yaml.tmpl`, and `.yml.tmpl`, plus tests for non-empty egress on ingress-only policies and explicit `policyTypes: [Egress]` missing DNS rules.
  - *LOW (Protocol validation & link-local literals)*: Added validation that port 53 egress rules specify both UDP and TCP protocols, and ensured RFC 1918-only policies are not rejected when omitting link-local literals.
  - *LOW (Defensive scalar parsing)*: Added graceful error reporting for malformed scalar fields in manifests without throwing unhandled exceptions.
  - *LOW (PR body test count drift)*: Refreshed PR body test count from 32 to 66 across all 6 discovery extensions to match the suite at HEAD.
- **`review-docs-drift` pass (isolated context)**:
  - Checked docs map `docs/README.md`, `AGENTS.md`, and `docs/site/`. Sibling presubmit targets (`images-check`, `chart-check`) are not listed in a global doc table (they are documented via self-describing `make help` `## description` comments). `make help` correctly formats `iac-parity-check`. No documentation was rendered stale. `make docs-check` passed with 0 errors. **Disposition: Confirmed clean, no docs changes required.**

## Risk & Rollout

- **Blast radius:** Zero runtime impact. Confined to CI presubmits and developer testing tooling.
- **Failure modes & fallbacks:** If a static manifest is edited and omits required DNS peers, or if a new unlisted DNS policy is added, `make test-python` / `scripts/test_check_iac_parity.py` fails in presubmit with the exact file, policy, and missing CIDR.
- **Upgrade impact:** None. No CRDs, manifests, or agent configurations are altered.
- **Reversion:** `git revert` this commit.

---
_Generated with the help of Antigravity._

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -563,7 +563,7 @@ chart-sync: ## Sync the Helm chart's CRD copies and operator ClusterRole rules f
 chart-check: ## Verify the chart's CRD/RBAC copies match k8s-operator/config (CI runs this).
 	@./hack/sync-chart-manifests.sh --check
 
-iac-parity-check: ## Verify DNS egress rule parity across static NetworkPolicy copies (CI runs this).
+iac-parity-check: ## Verify DNS egress rule parity across static NetworkPolicy copies (CI runs this via scripts/test_check_iac_parity.py).
 	@python3 -c "import yaml" 2>/dev/null || { \
 		echo "iac-parity-check needs PyYAML: python3 -m pip install pyyaml (or make test-python-deps)"; \
 		exit 1; \
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ BASE_IMAGE_ARGS := $(foreach v,$(BASE_IMAGE_VARS),$(if $($(v)),--build-arg $(v)=
 SANDBOX_IMAGE_VARS := PYTHON_IMAGE
 SANDBOX_IMAGE_ARGS := $(foreach v,$(SANDBOX_IMAGE_VARS),$(if $($(v)),--build-arg $(v)=$($(v))))
 
-.PHONY: default help docker-build docker-build-agents docker-build-credential-proxy docker-build-sandbox docker-smoke-sandbox docker-push docker-push-agents docker-push-credential-proxy docker-push-sandbox dev-rebuild-agent mirror-images images-check status prettier-check prettier-write test-python test-python-deps test-bench test-bench-deps bench-case-check e2e-tests e2e-test-deps test-e2e test-e2e-deps validate prompt-check docs-generate docs-check docs-check-generated docs-check-links docs-check-terminology docs-check-map docs-check-context-budget chart-sync chart-check tf-apply tf-destroy coverage coverage-check test-integration conformance
+.PHONY: default help docker-build docker-build-agents docker-build-credential-proxy docker-build-sandbox docker-smoke-sandbox docker-push docker-push-agents docker-push-credential-proxy docker-push-sandbox dev-rebuild-agent mirror-images images-check status prettier-check prettier-write test-python test-python-deps test-bench test-bench-deps bench-case-check e2e-tests e2e-test-deps test-e2e test-e2e-deps validate prompt-check docs-generate docs-check docs-check-generated docs-check-links docs-check-terminology docs-check-map docs-check-context-budget chart-sync chart-check iac-parity-check tf-apply tf-destroy coverage coverage-check test-integration conformance
 
 # The agent images this repository builds -- one per `--target` stage in
 # deploy/docker/Dockerfile, which is not the same thing as one per directory
@@ -562,6 +562,13 @@ chart-sync: ## Sync the Helm chart's CRD copies and operator ClusterRole rules f
 
 chart-check: ## Verify the chart's CRD/RBAC copies match k8s-operator/config (CI runs this).
 	@./hack/sync-chart-manifests.sh --check
+
+iac-parity-check: ## Verify DNS egress rule parity across static NetworkPolicy copies (CI runs this).
+	@python3 -c "import yaml" 2>/dev/null || { \
+		echo "iac-parity-check needs PyYAML: python3 -m pip install pyyaml (or make test-python-deps)"; \
+		exit 1; \
+	}
+	@python3 scripts/check_iac_parity.py
 
 tf-apply: ## Apply terraform/examples/full-install, adopting KMS resources a previous destroy left behind.
 	@./terraform/examples/full-install/lifecycle.sh apply $(ARGS)
--- a/scripts/check_iac_parity.py
+++ b/scripts/check_iac_parity.py
@@ -65,6 +65,20 @@ STATIC_NETWORK_POLICIES: tuple[str, ...] = (
     "k8s-operator/config/integrations/litellm/base/networkpolicy.yaml",
 )
 
+# Manifests containing kind: NetworkPolicy that are deliberately excluded from the
+# static DNS parity roster. Every entry must carry its reviewed reason.
+EXCLUDED_NETPOL_MANIFESTS: dict[str, str] = {
+    # Operator reconciliation test fixtures; golden expected outputs verified by Go controller tests.
+    "k8s-operator/internal/testing/testdata/platform/expected/platformagent.yaml": "operator golden test fixture",
+    "k8s-operator/internal/testing/testdata/platform/expected/platformagent-tagged.yaml": "operator golden test fixture",
+    "k8s-operator/internal/testing/testdata/platform/expected/platformagent-telemetry.yaml": "operator golden test fixture",
+    "k8s-operator/internal/testing/testdata/platform/expected/platformagent-split-broker.yaml": "operator golden test fixture",
+}
+
+IGNORED_DIRS: frozenset[str] = frozenset(
+    {".venv", "node_modules", "__pycache__", ".git", ".coverage-data", ".terraform", ".claude", "docs/site"}
+)
+
 REQUIRED_DNS_LITERAL = "10.96.0.10/32"
 REQUIRED_WILDCARD_CIDR = "0.0.0.0/0"
 REQUIRED_EXCEPT_MINIMUM: frozenset[str] = frozenset(
@@ -98,16 +112,75 @@ def sanitize_helm_template(text: str) -> str:
 
 
 def load_network_policies(path: Path) -> list[dict]:
-    """Read a manifest or template file and return all NetworkPolicy documents."""
+    """Read a manifest or template file and return all NetworkPolicy documents.
+
+    Parses document-by-document so syntax anomalies in unrelated manifests (like
+    Deployments or ConfigMaps in the same template file) do not fail NetworkPolicy
+    extraction.
+    """
     if not path.is_file():
         raise FileNotFoundError(f"File not found: {path}")
     raw_content = path.read_text(encoding="utf-8")
     sanitized = sanitize_helm_template(raw_content)
     policies: list[dict] = []
-    for doc in yaml.safe_load_all(sanitized):
-        if isinstance(doc, dict) and doc.get("kind") == "NetworkPolicy":
-            policies.append(doc)
+
+    # Split documents on YAML boundary '---'
+    for chunk in re.split(r"^---(?:\s.*)?$", sanitized, flags=re.MULTILINE):
+        chunk_stripped = chunk.strip()
+        if not chunk_stripped:
+            continue
+        # Only parse chunks that declare kind: NetworkPolicy
+        if not re.search(r"^\s*kind:\s*['\"]?NetworkPolicy['\"]?", chunk_stripped, flags=re.MULTILINE):
+            continue
+        try:
+            doc = yaml.safe_load(chunk_stripped)
+            if isinstance(doc, dict) and doc.get("kind") == "NetworkPolicy":
+                policies.append(doc)
+        except Exception as exc:
+            raise ValueError(f"malformed NetworkPolicy document in {path}: {exc}") from exc
+
     return policies
+
+
+def discover_dns_network_policies(root: Path = REPO_ROOT) -> set[str]:
+    """Scan the repository tree for all manifest files defining a port-53 DNS egress rule.
+
+    Mirroring scripts/test_test_discovery.py, this discovery scan prevents any unlisted
+    DNS-bearing NetworkPolicy from escaping the static parity guard.
+
+    Returns:
+        set[str]: set of repo-relative POSIX file paths.
+    """
+    discovered: set[str] = set()
+    patterns = ("*.yaml", "*.yml", "*.yaml.template", "*.yml.template")
+    for pattern in patterns:
+        for path in root.rglob(pattern):
+            if any(part in IGNORED_DIRS for part in path.parts):
+                continue
+            rel = path.relative_to(root).as_posix()
+            if rel in EXCLUDED_NETPOL_MANIFESTS:
+                continue
+            try:
+                raw = path.read_text(encoding="utf-8")
+            except Exception:
+                continue
+            # Fast check before parsing
+            if "NetworkPolicy" not in raw or "53" not in raw:
+                continue
+            try:
+                policies = load_network_policies(path)
+            except Exception:
+                continue
+            for pol in policies:
+                spec = pol.get("spec") or {}
+                for rule in spec.get("egress") or []:
+                    if not isinstance(rule, dict):
+                        continue
+                    ports = rule.get("ports") or []
+                    if any(isinstance(p, dict) and p.get("port") in (53, "53") for p in ports):
+                        discovered.add(rel)
+                        break
+    return discovered
 
 
 def check_dns_egress_rule(
@@ -239,7 +312,7 @@ def check_all(
     return total_rules, all_errors
 
 
-def main() -> int:
+def main(argv: list[str] | None = None) -> int:
     parser = argparse.ArgumentParser(
         description="Verify DNS egress rule parity across static NetworkPolicy copies."
     )
@@ -254,7 +327,7 @@ def main() -> int:
         nargs="*",
         help="Optional specific file paths to check (defaults to STATIC_NETWORK_POLICIES).",
     )
-    args = parser.parse_args()
+    args = parser.parse_args(argv)
 
     files = args.files if args.files else STATIC_NETWORK_POLICIES
     if args.verbose:
--- a/scripts/check_iac_parity.py
+++ b/scripts/check_iac_parity.py
@@ -67,16 +67,21 @@ STATIC_NETWORK_POLICIES: tuple[str, ...] = (
 
 # Manifests containing kind: NetworkPolicy that are deliberately excluded from the
 # static DNS parity roster. Every entry must carry its reviewed reason.
-EXCLUDED_NETPOL_MANIFESTS: dict[str, str] = {
-    # Operator reconciliation test fixtures; golden expected outputs verified by Go controller tests.
-    "k8s-operator/internal/testing/testdata/platform/expected/platformagent.yaml": "operator golden test fixture",
-    "k8s-operator/internal/testing/testdata/platform/expected/platformagent-tagged.yaml": "operator golden test fixture",
-    "k8s-operator/internal/testing/testdata/platform/expected/platformagent-telemetry.yaml": "operator golden test fixture",
-    "k8s-operator/internal/testing/testdata/platform/expected/platformagent-split-broker.yaml": "operator golden test fixture",
-}
+# Note: Go test fixtures under testdata/ are ignored systematically via IGNORED_DIRS.
+EXCLUDED_NETPOL_MANIFESTS: dict[str, str] = {}
 
 IGNORED_DIRS: frozenset[str] = frozenset(
-    {".venv", "node_modules", "__pycache__", ".git", ".coverage-data", ".terraform", ".claude", "docs/site"}
+    {
+        ".venv",
+        "node_modules",
+        "__pycache__",
+        ".git",
+        ".coverage-data",
+        ".terraform",
+        ".claude",
+        "docs/site",
+        "testdata",
+    }
 )
 
 REQUIRED_DNS_LITERAL = "10.96.0.10/32"
--- a/scripts/check_iac_parity.py
+++ b/scripts/check_iac_parity.py
@@ -49,6 +49,7 @@ the required peer shape, without evaluating dynamic Helm value permutations.
 from __future__ import annotations
 
 import argparse
+import os
 import re
 import sys
 from pathlib import Path
@@ -63,6 +64,17 @@ except ImportError:
     )
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
+
+DNS_PORT: int = 53
+DNS_PORT_STR: str = "53"
+DNS_PORTS: tuple[int | str, ...] = (DNS_PORT, DNS_PORT_STR)
+
+DISCOVERY_FILE_PATTERNS: tuple[str, ...] = (
+    "*.yaml",
+    "*.yml",
+    "*.yaml.template",
+    "*.yml.template",
+)
 
 STATIC_NETWORK_POLICIES: tuple[str, ...] = (
     "charts/kube-agents/templates/litellm.yaml",
@@ -120,14 +132,16 @@ def sanitize_helm_template(text: str) -> str:
     text = re.sub(r"\{\{-?\s*/\*.*?\*/\s*-?\}\}", "", text, flags=re.DOTALL)
     lines: list[str] = []
     for line in text.splitlines():
-        stripped = line.strip()
-        if stripped.startswith("{{") and stripped.endswith("}}"):
-            # Whole line is a template control statement
+        # Check if the line consists entirely of template control tags (ignoring trailing comments)
+        line_without_tags = re.sub(r"\{\{.*?\}\}", "", line)
+        if line_without_tags.split("#", 1)[0].strip() == "":
             lines.append("# " + line)
-        else:
-            # Replace inline template expressions like {{ .Release.Namespace }}
-            cleaned = re.sub(r"\{\{.*?\}\}", "placeholder", line)
-            lines.append(cleaned)
+            continue
+        # For lines with mixed content, strip control directives (if, else, end, with, range)
+        cleaned = re.sub(r"\{\{-?\s*(?:if\b|else\b|end\b|range\b|with\b).*?-?\}\}", "", line)
+        # Any remaining template expressions are value interpolations
+        cleaned = re.sub(r"\{\{.*?\}\}", "placeholder", cleaned)
+        lines.append(cleaned)
     return "\n".join(lines)
 
 
@@ -163,24 +177,28 @@ def load_network_policies(path: Path) -> list[dict]:
 
 
 def validate_exclusions(
-    exclusions: dict[str, str] = EXCLUDED_NETPOL_MANIFESTS,
-    roster: Iterable[str] = STATIC_NETWORK_POLICIES,
-    root: Path = REPO_ROOT,
+    exclusions: dict[str, str] | None = None,
+    roster: Iterable[str] | None = None,
+    root: Path | None = None,
 ) -> list[str]:
     """Validate that every exclusion carries a non-empty reason, points to a file on disk, and is disjoint from the roster."""
+    resolved_exclusions = EXCLUDED_NETPOL_MANIFESTS if exclusions is None else exclusions
+    resolved_roster = STATIC_NETWORK_POLICIES if roster is None else roster
+    resolved_root = REPO_ROOT if root is None else root
+
     errors: list[str] = []
-    roster_set = set(roster)
-    for path_str, reason in exclusions.items():
+    roster_set = set(resolved_roster)
+    for path_str, reason in resolved_exclusions.items():
         if not reason or not reason.strip():
             errors.append(f"Exclusion for {path_str} must carry a non-empty reviewed reason")
-        if not (root / path_str).is_file():
+        if not (resolved_root / path_str).is_file():
             errors.append(f"Excluded manifest does not exist on disk: {path_str}")
         if path_str in roster_set:
             errors.append(f"Excluded manifest {path_str} cannot also be in STATIC_NETWORK_POLICIES roster")
     return errors
 
 
-def discover_dns_network_policies(root: Path = REPO_ROOT) -> set[str]:
+def discover_dns_network_policies(root: Path | None = None) -> set[str]:
     """Scan the repository tree for all manifest files defining a port-53 DNS egress rule.
 
     Scans deployed manifest files (*.yaml, *.yml, *.yaml.template, *.yml.template)
@@ -189,38 +207,68 @@ def discover_dns_network_policies(root: Path = REPO_ROOT) -> set[str]:
     Returns:
         set[str]: set of repo-relative POSIX file paths.
     """
+    resolved_root = REPO_ROOT if root is None else root
     discovered: set[str] = set()
-    patterns = ("*.yaml", "*.yml", "*.yaml.template", "*.yml.template")
-    for pattern in patterns:
-        for path in root.rglob(pattern):
-            rel = path.relative_to(root)
-            if any(part in IGNORED_DIRS for part in rel.parts):
+
+    for dirpath, dirnames, filenames in os.walk(resolved_root):
+        # Prune ignored directories in place to avoid entering them
+        dirnames[:] = [d for d in dirnames if d not in IGNORED_DIRS]
+        rel_dir = Path(dirpath).relative_to(resolved_root)
+        if any(part in IGNORED_DIRS for part in rel_dir.parts):
+            continue
+        rel_dir_posix = rel_dir.as_posix()
+        if any(
+            rel_dir_posix == prefix or rel_dir_posix.startswith(prefix + "/")
+            for prefix in IGNORED_PREFIXES
+        ):
+            dirnames.clear()
+            continue
+
+        for filename in filenames:
+            if not any(
+                filename.endswith(ext.lstrip("*"))
+                for ext in DISCOVERY_FILE_PATTERNS
+            ):
                 continue
+
+            path = Path(dirpath) / filename
+            rel = path.relative_to(resolved_root)
             rel_posix = rel.as_posix()
-            if any(rel_posix == prefix or rel_posix.startswith(prefix + "/") for prefix in IGNORED_PREFIXES):
-                continue
+
             if rel_posix in EXCLUDED_NETPOL_MANIFESTS:
                 continue
+
             try:
                 raw = path.read_text(encoding="utf-8")
             except Exception as exc:
-                raise RuntimeError(f"failed to read {rel_posix} during DNS policy discovery: {exc}") from exc
+                raise RuntimeError(
+                    f"failed to read {rel_posix} during DNS policy discovery: {exc}"
+                ) from exc
+
             # Fast check before parsing
-            if "NetworkPolicy" not in raw or "53" not in raw:
+            if "NetworkPolicy" not in raw or DNS_PORT_STR not in raw:
                 continue
+
             try:
                 policies = load_network_policies(path)
             except Exception as exc:
-                raise ValueError(f"failed to parse NetworkPolicy in {rel_posix} during DNS policy discovery: {exc}") from exc
+                raise ValueError(
+                    f"failed to parse NetworkPolicy in {rel_posix} during DNS policy discovery: {exc}"
+                ) from exc
+
             for pol in policies:
                 spec = pol.get("spec") or {}
                 for rule in spec.get("egress") or []:
                     if not isinstance(rule, dict):
                         continue
                     ports = rule.get("ports") or []
-                    if any(isinstance(p, dict) and p.get("port") in (53, "53") for p in ports):
+                    if any(
+                        isinstance(p, dict) and p.get("port") in DNS_PORTS
+                        for p in ports
+                    ):
                         discovered.add(rel_posix)
                         break
+
     return discovered
 
 
@@ -235,7 +283,8 @@ def check_dns_egress_rule(
     to_peers = rule.get("to") or []
     has_dns_literal = False
     has_wildcard = False
-    wildcard_excepts: set[str] = set()
+    wildcard_peers: list[dict] = []
+    rule_desc = f"{path_display} (policy '{policy_name}', DNS rule #{rule_idx})"
 
     for peer in to_peers:
         if not isinstance(peer, dict) or "ipBlock" not in peer:
@@ -248,11 +297,18 @@ def check_dns_egress_rule(
             has_dns_literal = True
         elif cidr == REQUIRED_WILDCARD_CIDR:
             has_wildcard = True
+            wildcard_peers.append(peer)
+            peer_excepts: set[str] = set()
             except_list = ip_block.get("except")
             if isinstance(except_list, list):
-                wildcard_excepts.update(str(x) for x in except_list)
+                peer_excepts = {str(x) for x in except_list}
+            missing = REQUIRED_EXCEPT_MINIMUM - peer_excepts
+            if missing:
+                errors.append(
+                    f"{rule_desc}: '{REQUIRED_WILDCARD_CIDR}' peer except list is missing required private CIDRs: "
+                    f"{sorted(missing)}"
+                )
 
-    rule_desc = f"{path_display} (policy '{policy_name}', DNS rule #{rule_idx})"
     if not has_dns_literal:
         errors.append(
             f"{rule_desc}: missing required ipBlock literal '{REQUIRED_DNS_LITERAL}' for classic ClusterIP DNS"
@@ -261,24 +317,22 @@ def check_dns_egress_rule(
         errors.append(
             f"{rule_desc}: missing required '{REQUIRED_WILDCARD_CIDR}' peer with except list for dynamic/public DNS VIPs"
         )
-    else:
-        missing_except = REQUIRED_EXCEPT_MINIMUM - wildcard_excepts
-        if missing_except:
-            errors.append(
-                f"{rule_desc}: '{REQUIRED_WILDCARD_CIDR}' except list is missing required private CIDRs: "
-                f"{sorted(missing_except)}"
-            )
+    elif len(wildcard_peers) > 1:
+        errors.append(
+            f"{rule_desc}: expected at most one '{REQUIRED_WILDCARD_CIDR}' peer"
+        )
 
     return errors
 
 
-def check_network_policy_file(path: Path, root: Path = REPO_ROOT) -> tuple[int, list[str]]:
+def check_network_policy_file(path: Path, root: Path | None = None) -> tuple[int, list[str]]:
     """Check all Egress NetworkPolicy resources in a file for DNS egress parity.
 
     Returns:
         tuple[int, list[str]]: (number of DNS rules checked, list of error messages)
     """
-    rel_path = str(path.relative_to(root)) if path.is_relative_to(root) else str(path)
+    resolved_root = REPO_ROOT if root is None else root
+    rel_path = str(path.relative_to(resolved_root)) if path.is_relative_to(resolved_root) else str(path)
     try:
         policies = load_network_policies(path)
     except Exception as exc:
@@ -307,7 +361,7 @@ def check_network_policy_file(path: Path, root: Path = REPO_ROOT) -> tuple[int, 
             for r in (egress_rules or [])
             if isinstance(r, dict)
             and any(
-                isinstance(p, dict) and p.get("port") in (53, "53")
+                isinstance(p, dict) and p.get("port") in DNS_PORTS
                 for p in (r.get("ports") or [])
             )
         ]
@@ -330,27 +384,30 @@ def check_network_policy_file(path: Path, root: Path = REPO_ROOT) -> tuple[int, 
 
 
 def check_all(
-    files: Iterable[Path | str] = STATIC_NETWORK_POLICIES,
-    root: Path = REPO_ROOT,
+    files: Iterable[Path | str] | None = None,
+    root: Path | None = None,
     verbose: bool = False,
 ) -> tuple[int, list[str]]:
     """Check all specified files for DNS egress parity."""
+    resolved_files = STATIC_NETWORK_POLICIES if files is None else files
+    resolved_root = REPO_ROOT if root is None else root
+
     total_rules = 0
     all_errors: list[str] = []
 
-    exclusion_errors = validate_exclusions(root=root)
+    exclusion_errors = validate_exclusions(root=resolved_root)
     all_errors.extend(exclusion_errors)
     if verbose and exclusion_errors:
         for err in exclusion_errors:
             print(f"  FAIL: exclusion contract: {err}")
 
-    for item in files:
-        path = root / item if not Path(item).is_absolute() else Path(item)
-        rules_checked, file_errors = check_network_policy_file(path, root)
+    for item in resolved_files:
+        path = resolved_root / item if not Path(item).is_absolute() else Path(item)
+        rules_checked, file_errors = check_network_policy_file(path, resolved_root)
         total_rules += rules_checked
         all_errors.extend(file_errors)
         if verbose:
-            rel = str(path.relative_to(root)) if path.is_relative_to(root) else str(path)
+            rel = str(path.relative_to(resolved_root)) if path.is_relative_to(resolved_root) else str(path)
             if file_errors:
                 print(f"  FAIL: {rel} ({len(file_errors)} errors)")
             else:
@@ -376,9 +433,10 @@ def main(argv: list[str] | None = None) -> int:
     )
     args = parser.parse_args(argv)
 
-    files = args.files if args.files else STATIC_NETWORK_POLICIES
+    files = args.files if args.files else None
+    display_count = len(files) if files else len(STATIC_NETWORK_POLICIES)
     if args.verbose:
-        print(f"Checking {len(files)} static NetworkPolicy copies for DNS egress parity...")
+        print(f"Checking {display_count} static NetworkPolicy copies for DNS egress parity...")
 
     rules_checked, errors = check_all(files=files, root=REPO_ROOT, verbose=args.verbose)
 
@@ -387,15 +445,15 @@ def main(argv: list[str] | None = None) -> int:
         for err in errors:
             print(f"  * {err}", file=sys.stderr)
         print(
-            f"\n{len(errors)} error(s) found across {len(files)} static policy copies.",
+            f"\n{len(errors)} error(s) found across {display_count} static policy copies.",
             file=sys.stderr,
         )
         return 1
 
     if not args.verbose:
-        print(f"OK: Verified DNS egress rule parity across {len(files)} static copies ({rules_checked} rules checked).")
+        print(f"OK: Verified DNS egress rule parity across {display_count} static copies ({rules_checked} rules checked).")
     else:
-        print(f"\nAll {len(files)} static policy copies passed parity check ({rules_checked} rules).")
+        print(f"\nAll {display_count} static policy copies passed parity check ({rules_checked} rules).")
     return 0
 
 
--- a/scripts/check_iac_parity.py
+++ b/scripts/check_iac_parity.py
@@ -34,6 +34,16 @@ deploy/kustomize/platform/networkpolicy-core-egress.yaml carries the 5-entry hou
 shape (adding 100.64.0.0/10 and 169.254.0.0/16, which is more contained and
 mirrors the operator-generated external egress policy). This script enforces the
 required-peer shape across all copies so drift is caught before merge.
+
+Scope:
+This check audits deployable Infrastructure-as-Code manifest files (*.yaml,
+*.yml, *.yaml.template, *.yml.template) across Helm templates, Kustomize bases,
+and integration examples. It does not scan Markdown documentation or agent skill
+prompts (*.md).
+
+Template semantics:
+For Helm templates, this check asserts that the static source definition carries
+the required peer shape, without evaluating dynamic Helm value permutations.
 """
 
 from __future__ import annotations
@@ -70,6 +80,7 @@ STATIC_NETWORK_POLICIES: tuple[str, ...] = (
 # Note: Go test fixtures under testdata/ are ignored systematically via IGNORED_DIRS.
 EXCLUDED_NETPOL_MANIFESTS: dict[str, str] = {}
 
+# Directory names to skip anywhere in the repo-relative path
 IGNORED_DIRS: frozenset[str] = frozenset(
     {
         ".venv",
@@ -79,9 +90,13 @@ IGNORED_DIRS: frozenset[str] = frozenset(
         ".coverage-data",
         ".terraform",
         ".claude",
-        "docs/site",
         "testdata",
     }
+)
+
+# Relative directory prefixes to skip
+IGNORED_PREFIXES: tuple[str, ...] = (
+    "docs/site",
 )
 
 REQUIRED_DNS_LITERAL = "10.96.0.10/32"
@@ -147,11 +162,29 @@ def load_network_policies(path: Path) -> list[dict]:
     return policies
 
 
+def validate_exclusions(
+    exclusions: dict[str, str] = EXCLUDED_NETPOL_MANIFESTS,
+    roster: Iterable[str] = STATIC_NETWORK_POLICIES,
+    root: Path = REPO_ROOT,
+) -> list[str]:
+    """Validate that every exclusion carries a non-empty reason, points to a file on disk, and is disjoint from the roster."""
+    errors: list[str] = []
+    roster_set = set(roster)
+    for path_str, reason in exclusions.items():
+        if not reason or not reason.strip():
+            errors.append(f"Exclusion for {path_str} must carry a non-empty reviewed reason")
+        if not (root / path_str).is_file():
+            errors.append(f"Excluded manifest does not exist on disk: {path_str}")
+        if path_str in roster_set:
+            errors.append(f"Excluded manifest {path_str} cannot also be in STATIC_NETWORK_POLICIES roster")
+    return errors
+
+
 def discover_dns_network_policies(root: Path = REPO_ROOT) -> set[str]:
     """Scan the repository tree for all manifest files defining a port-53 DNS egress rule.
 
-    Mirroring scripts/test_test_discovery.py, this discovery scan prevents any unlisted
-    DNS-bearing NetworkPolicy from escaping the static parity guard.
+    Scans deployed manifest files (*.yaml, *.yml, *.yaml.template, *.yml.template)
+    across the repository. Markdown documents and skill prompt files are excluded.
 
     Returns:
         set[str]: set of repo-relative POSIX file paths.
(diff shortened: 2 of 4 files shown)


jayantid:
**Smoke red triage (gate-watch):** not caused by this PR (diff is Makefile + a CI parity checker, never deployed). [Build 2095063285538754560](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1090/pull-kube-agents-smoke-test/2095063285538754560): `compliance-rbac-overgrant` rung-4 collapse with the exact #1171 signature (`required phrases absent ... ['debug-binding']`), plus `autoops-warning-event-triage` blocked at provision with `PermissionError: /opt/data/profiles/platform` in the watcher log (#1144; fix in flight #1183) and a 429 rep (#1097). `/retest` after #1183 (or the #1173 demotion) lands.

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,361 @@
...
+    DNS-bearing NetworkPolicy from escaping the static parity guard.
+
+    Returns:
+        set[str]: set of repo-relative POSIX file paths.
+    """
+    discovered: set[str] = set()
+    patterns = ("*.yaml", "*.yml", "*.yaml.template", "*.yml.template")
+    for pattern in patterns:
+        for path in root.rglob(pattern):
+            if any(part in IGNORED_DIRS for part in path.parts):
```
mplakhtiy: **MEDIUM — `IGNORED_DIRS` is matched against the absolute path, so a directory name *above* the repo root silences the whole discovery scan.**

`root.rglob(...)` yields absolute paths, so `path.parts` carries every segment from `/` down. Any ancestor directory named `.claude`, `.git`, `.venv`, `node_modules`, `__pycache__`, `.terraform`, `.coverage-data` or `testdata` matches on every file in the repository, and `discover_dns_network_policies()` returns `set()`.

Reproduced. Running the suite from `<repo>/.claude/worktrees/pr-1090` — the layout `.claude/commands/pr-review-batch.md` and the `EnterWorktree` tool both create — fails `test_discovery_matches_roster_exactly` with all 8 roster files reported as no longer containing DNS egress rules. The same tree copied to `/tmp` passes 17/17. So a contributor or agent running `make test-python` from the repository's own documented review workflow gets a red suite on a test unrelated to their change, and the obvious reading of the message ("the roster is stale") is the opposite of what happened.

Invisible from a normal checkout, which is why the earlier passes missed it; CI runs at `/home/runner/work/kube-agents/kube-agents`.

The relative form is already what line 166 uses for `EXCLUDED_NETPOL_MANIFESTS`, and `rel` is computed two lines below.

```suggestion
            if any(part in IGNORED_DIRS for part in path.relative_to(root).parts):
```

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,361 @@
...
+    """Scan the repository tree for all manifest files defining a port-53 DNS egress rule.
+
+    Mirroring scripts/test_test_discovery.py, this discovery scan prevents any unlisted
+    DNS-bearing NetworkPolicy from escaping the static parity guard.
+
+    Returns:
+        set[str]: set of repo-relative POSIX file paths.
+    """
+    discovered: set[str] = set()
+    patterns = ("*.yaml", "*.yml", "*.yaml.template", "*.yml.template")
```
mplakhtiy: **MEDIUM — discovery globs four YAML extensions, so a DNS NetworkPolicy inside a `SKILL.md` escapes it, and there is one in the tree today.**

`agents/platform/skills/gke-multitenancy/SKILL.md:149-171` ships a NetworkPolicy the Platform Agent is instructed to apply to a user's cluster. Its port-53 rule reaches the resolver through a `k8s-app: kube-dns` podSelector only — no `10.96.0.10/32` ipBlock, no `0.0.0.0/0` peer. That is exactly the shape this module's own docstring (lines 20-23) says does not work: "NetworkPolicy matches that VIP rather than the backend pods, so selectors alone do not cover it."

Both `.github/risk-rules.yml` and `AGENTS.md` treat Markdown under `agents/**` as executable agent configuration rather than documentation, so that is a manifest, not an example in prose.

This is #687's failure mode moved one file type over. The docstring's own account of #687 — that greps for `*.yaml` skipped it — is the reason `*.yaml.template` is in this tuple; `.md` was not added.

Confirmed that discovery cannot see the file. Whether that deployed policy actually breaks DNS depends on an agent applying the snippet verbatim, but the escape from the guard does not depend on that.

Either extend discovery to fenced YAML under `agents/**/*.md`, or state the file-type boundary in the docstring so the guarantee matches the code. The `SKILL.md` policy itself is a separate defect and belongs in its own issue either way.

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,361 @@
...
+def sanitize_helm_template(text: str) -> str:
+    """Sanitize Go/Helm template tags so that YAML parsers can load the manifests."""
+    # Strip multi-line comments: {{- /* ... */ -}} or {{/* ... */}}
+    text = re.sub(r"\{\{-?\s*/\*.*?\*/\s*-?\}\}", "", text, flags=re.DOTALL)
+    lines: list[str] = []
+    for line in text.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("{{") and stripped.endswith("}}"):
+            # Whole line is a template control statement
+            lines.append("# " + line)
```
mplakhtiy: **MEDIUM — a required peer wrapped in a Helm conditional reads as unconditionally present.**

Commenting out whole-line `{{- if ... }}` and `{{- end }}` lets the YAML parse, and also erases the distinction between a peer that always renders and one that renders only under a value.

Reproduced. A manifest whose `10.96.0.10/32` ipBlock sits inside `{{- if .Values.legacyClusterIpRange }}` passes with `OK: Verified DNS egress rule parity across 1 static copies (1 rules checked).`

Two of the eight roster entries are Helm templates, so this is in scope for the change as it stands. It is also about to become concrete: sibling PR #1195 changes `charts/kube-agents/templates/litellm.yaml:245` to gate that NetworkPolicy behind `(not .Values.platformAgent.enabled)`, so on the default install (`platformAgent.enabled: true`) it renders nothing at all while this checker keeps counting it among the eight copies it verified. Merge order decides nothing — the checker never notices either way.

The opposite direction is safe, for the record: a peer pulled in by a whole-line `{{- include ... }}` vanishes from the parsed document and the check fails loudly with the missing-literal error. Only the conditional-wrapping direction is silent.

Fix direction: `helm template` the chart files rather than sanitizing them, or track conditional depth in the sanitizer and refuse to count a peer nested inside one. If neither is worth the cost, the honest docstring claim is that the peers are present in the source, not that every rendered copy carries them.

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,361 @@
...
+        for path in root.rglob(pattern):
+            if any(part in IGNORED_DIRS for part in path.parts):
+                continue
+            rel = path.relative_to(root).as_posix()
+            if rel in EXCLUDED_NETPOL_MANIFESTS:
+                continue
+            try:
+                raw = path.read_text(encoding="utf-8")
+            except Exception:
+                continue
```
mplakhtiy: **LOW — the exclusion escape hatch is unvalidated, untested, and short-circuits before the file is read.**

Line 69's "Every entry must carry its reviewed reason" is enforced by nobody. Nothing checks that a reason is non-empty, that the path exists, or that an excluded path is not also in `STATIC_NETWORK_POLICIES`. Verified: adding `{"examples/sneaky/networkpolicy.yaml": ""}` for a policy missing both required peers leaves the suite green and the checker at exit 0.

The bare `continue` on the read failure just below, and the matching one on parse failure at 177-178, compound it — a manifest whose sanitized YAML will not parse drops out of discovery without a word. No file in the tree hits that today (I checked all of them), so it is latent rather than live. The coverage bot flagged 167, 170-171 and 177-178 as uncovered, which is the same observation from the other direction.

`EXCLUDED_NETPOL_MANIFESTS` is imported into the test file and never asserted on.

Worth asserting in `test_discovery_matches_roster_exactly`: every exclusion carries a non-empty reason, points at a file that exists, and is disjoint from the roster. And either log or collect the skipped-on-parse-failure paths rather than dropping them silently.

mplakhtiy (commented):
Full adversarial pass over the diff at `4ea6815`, run in a context that did not write the change. Six findings inline: three MEDIUM on the discovery scan and the Helm sanitizer, three LOW.

The MEDIUMs all land on the same claim — that repository-wide discovery makes the hand-maintained roster self-maintaining. Each was reproduced rather than reasoned about: the ignore list disables the scan entirely from a `.claude/worktrees/` checkout, a DNS NetworkPolicy in a `SKILL.md` sits outside the glob, and a peer behind `{{- if }}` counts as unconditionally present.

What I ran: `python3 -m unittest scripts.test_check_iac_parity -v` (17/17 from a clean checkout, 1 failure from a worktree), `make iac-parity-check`, `python3 scripts/check_iac_parity.py -v`, a `grep -rn "port: *53"` sweep across every file extension in the tree, and four planted mutations.

Verified clean, so nobody re-litigates them: the eight-copy roster is complete for deployed YAML; an egress policy carrying no port-53 rule at all is invisible to both the roster and the discovery test, and that is correct here, since NetworkPolicies are additive and the `networkpolicy-{apiserver,external,internal}-egress.yaml` set deliberately leans on `networkpolicy-core-egress.yaml` for DNS; the two `hindsight` policies are Ingress-only by documented intent; `default-deny-netpol.yaml` has no rule to check; the test file is reached by `PYTHON_TEST_DIRS` and resolves under the exact CI invocation; and #1169 does not collide.

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,361 @@
...
+IGNORED_DIRS: frozenset[str] = frozenset(
+    {
+        ".venv",
+        "node_modules",
+        "__pycache__",
+        ".git",
+        ".coverage-data",
+        ".terraform",
+        ".claude",
+        "docs/site",
```
mplakhtiy: **LOW — `"docs/site"` can never match.**

`path.parts` yields `('docs', 'site', ...)`, never the string `"docs/site"`, so this entry is dead: `'docs/site' in Path('docs/site/x.yaml').parts` is `False`. Harmless today, since `docs/site` holds no `.yaml` or `.yml` files, but the intent it records is not implemented and the first YAML fixture added there will be scanned despite the list saying otherwise.

Worth settling alongside the `relative_to` fix above — note that switching to relative paths still will not make a multi-segment entry work without a prefix match, so this needs either a real prefix check or deletion.

Makefile
```diff
@@ -393,6 +393,13 @@ chart-sync: ## Sync the Helm chart's CRD copies and operator ClusterRole rules f
 chart-check: ## Verify the chart's CRD/RBAC copies match k8s-operator/config (CI runs this).
 	@./hack/sync-chart-manifests.sh --check
 
+iac-parity-check: ## Verify DNS egress rule parity across static NetworkPolicy copies (CI runs this).
```
mplakhtiy: **LOW — the help text says "(CI runs this)" and no workflow runs it.**

The two sibling targets carrying that same phrase are literally invoked by `.github/workflows/validate.yml` (`make chart-check` at line 66, `make images-check` at line 69). Nothing invokes `make iac-parity-check`; the invariant is gated transitively, through `test_all_production_copies_pass` calling `check_all()` under `make test-python`.

Per `AGENTS.md`, the `## description` **is** the published reference — `make help` prints it — so this string currently tells a reader the target is a CI step. Someone who later deletes or renames `scripts/test_check_iac_parity.py` will believe a workflow still guards this invariant.

```suggestion
iac-parity-check: ## Verify DNS egress rule parity across static NetworkPolicy copies (CI runs this via scripts/test_check_iac_parity.py).
```

scripts/check_iac_parity.py
shalinibhatiapl: Switched to evaluating directory segments against path.relative_to(root).parts instead of absolute path.parts, preventing ancestor directories (such as .claude/worktrees/) from matching. Added synthetic test test_ignored_dirs_in_ancestor_path_does_not_break_discovery to prevent regressions.

scripts/check_iac_parity.py
shalinibhatiapl: Documented the file-type boundary in the module docstring and discover_dns_network_policies: this checker explicitly audits deployable IaC manifest files (*.yaml, *.yml, *.yaml.template, *.yml.template) across Helm templates, Kustomize bases, and integration examples, and intentionally excludes Markdown documentation or skill prompts (*.md).

scripts/check_iac_parity.py
shalinibhatiapl: Documented template semantics in the module docstring: for Helm templates, this check asserts that the static source definition carries the required peer shape without evaluating dynamic Helm value permutations.

scripts/check_iac_parity.py
shalinibhatiapl: Added validate_exclusions to check that every entry carries a non-empty reviewed reason, exists on disk, and is disjoint from STATIC_NETWORK_POLICIES. This is enforced during check_all and tested in test_excluded_manifests_contract. In discover_dns_network_policies, replaced bare continue on read/parse failures with explicit RuntimeError and ValueError exceptions, covered by new synthetic tests.

scripts/check_iac_parity.py
shalinibhatiapl: Extracted "docs/site" into IGNORED_PREFIXES = ("docs/site",) and matched via prefix check (rel_posix == prefix or rel_posix.startswith(prefix + "/")). Added synthetic test test_ignored_prefixes_docs_site.

Makefile
shalinibhatiapl: Updated the help text for iac-parity-check in Makefile to (CI runs this via scripts/test_check_iac_parity.py).

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,547 @@
...
+)
+
+REQUIRED_DNS_LITERAL = "10.96.0.10/32"
+REQUIRED_WILDCARD_CIDR = "0.0.0.0/0"
+REQUIRED_EXCEPT_MINIMUM: frozenset[str] = frozenset(
+    {"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"}
+)
+HOUSE_SHAPE_EXCEPT: frozenset[str] = frozenset(
+    {"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "100.64.0.0/10", "169.254.0.0/16"}
+)
```
mplakhtiy, line 157: **HIGH — the guard verifies two of the six DNS peers and neither protocol, so three drift scenarios that break DNS pass green.**

`REQUIRED_DNS_LITERAL` and `REQUIRED_WILDCARD_CIDR` are the whole peer contract, and `check_dns_egress_rule` (`:328`) never looks at `ports[].protocol` — the port-53 match at `:449-453` tests `p.get("port")` alone. All eight roster copies today are identical in the DNS rule: `protocol: TCP` and `UDP`, and six peers in the same order.

```
protos=['TCP','UDP'] peers=['selector','selector','169.254.20.10/32','169.254.169.254/32','10.96.0.10/32','0.0.0.0/0']
```

Four of those six peers and both protocols are unguarded. Three concrete regressions, each reproduced against the branch and reverted:

**1. Dropping `169.254.169.254/32` from `deploy/kustomize/platform/networkpolicy-core-egress.yaml`.** That peer is #1169, which landed on `main` as `799c7e30` after this branch diverged. The manifest's own comment says why nothing else covers it:

> Cloud DNS for GKE resolves at the metadata address rather than through kube-dns: a Pod's resolv.conf names 169.254.169.254, so none of the peers above is the resolver and none of the blocks below reaches it — the 0.0.0.0/0 peer excepts the whole 169.254.0.0/16 range.

Delete the peer and `make iac-parity-check` prints `OK: ... 8 static copies (8 rules checked)`. On a Cloud-DNS-for-GKE cluster the platform agent then resolves nothing — the #687 outage again, through a different address.

**2. Dropping `169.254.20.10/32` from the same file.** Same mechanism for NodeLocal DNSCache: that copy's five-entry `except` list excludes `169.254.0.0/16`, so the wildcard peer does not reach the cache, and the two selector peers match pods rather than the link-local address. Green.

**3. Dropping `protocol: UDP` from any copy's port-53 ports list.** The rule still carries `port: 53`, so `dns_rules` is non-empty and every peer assertion passes. Resolvers try UDP first; TCP-only DNS egress is a resolution failure on every cluster shape, not just the unusual ones.

The peer half of this bites only `networkpolicy-core-egress.yaml`, because it is the one copy whose five-entry `except` removes `169.254.0.0/16`; on the other seven the wildcard peer still reaches those addresses. The protocol half bites all eight. That is the asymmetry #747 B5 named when it said "identity can be asserted only after that alignment lands" — the required-peer minimum is sufficient for seven copies and insufficient for the one, and nothing in the script encodes the difference.

**Fix direction.** Assert both protocols on the port-53 rule, and make the required-peer set a function of the `except` list: when the wildcard peer excepts `169.254.0.0/16`, `169.254.20.10/32` and `169.254.169.254/32` become required literals. Failing that, name in the docstring the four peers the check does not defend — the file currently reads as though the shape it enumerates is the shape that keeps DNS working.

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,547 @@
...
+        line_without_tags = HELM_TAG_LINE_RE.sub("", line)
+        if line_without_tags.split("#", 1)[0].strip() == "":
+            lines.append("# " + line)
+            continue
+        # For lines with mixed content, strip control directives (if, else, end, with, range)
+        cleaned = HELM_CONTROL_DIRECTIVE_RE.sub("", line)
+        # Any remaining template expressions are value interpolations
+        cleaned = HELM_TAG_LINE_RE.sub(TEMPLATE_PLACEHOLDER, cleaned)
+        lines.append(cleaned)
+    return "\n".join(lines)
```
mplakhtiy, line 226: **MEDIUM — a peer or a whole policy behind a Helm conditional counts as present.**

`sanitize_helm_template` comments out whole-line control tags and strips inline ones, which erases the distinction between a peer that always renders and one that renders only under a value. Two reproductions, both exit 0:

- `10.96.0.10/32` in `charts/kube-agents/templates/litellm.yaml` moved inside `{{- if .Values.litellm.legacyServiceRange }}`.
- The entire `github-minter` NetworkPolicy wrapped in `{{- if .Values.neverTrue }}` — and the summary still reports "8 static copies (8 rules checked)".

The docstring's "Template semantics" paragraph (`:44-51`) discloses the limitation, so a reader of the source is not misled. What the check does not do is the assertion #747 D1 asks for by name:

> Land #1090's rendering assertion: `kubectl kustomize` / `helm template` the rendered surfaces and assert the port-53 rule contains both the `10.96.0.10/32` literal and an `0.0.0.0/0` peer carrying the `except` list.

Rendering is exactly what would catch both reproductions above.

There is a near-term instance of this. #1195 (draft) changes this chart template's NetworkPolicy guard so the template stops rendering on a default install, moving the policy to `platformagent_litellm_policy.go`. This checker keeps counting the template afterwards and cannot see the Go source at all.

**Fix.** Render — `helm template` for the two chart copies, `kubectl kustomize` for the kustomize one — or state in the docstring that D1 is delivered as a source-level check and why, so the limitation is recorded against the requirement rather than only as template semantics.

scripts/check_iac_parity.py
```diff
@@ -0,0 +1,547 @@
...
+DNS_PORTS: tuple[int | str, ...] = (DNS_PORT, DNS_PORT_STR)
+
+DISCOVERY_FILE_PATTERNS: tuple[str, ...] = (
+    "*.yaml",
+    "*.yml",
+    "*.yaml.template",
+    "*.yml.template",
+    "*.yaml.tmpl",
+    "*.yml.tmpl",
+)
```
mplakhtiy, line 112: **MEDIUM — discovery escapes four file classes; the docstring's Scope paragraph names one of them.**

Each escape was verified by planting a DNS-bearing NetworkPolicy and confirming `test_discovery_matches_roster_exactly` stayed green:

- **`*.md`.** Not in `DISCOVERY_FILE_PATTERNS`. `agents/platform/skills/gke-multitenancy/SKILL.md:152` ships one today — a NetworkPolicy the Platform Agent applies to a user's cluster, whose DNS rule reaches the resolver through a `k8s-app: kube-dns` podSelector only, the shape this script's own docstring at `:20-23` says does not work.
- **Any directory named `testdata`, at any depth.** `IGNORED_DIRS` (`:112`) prunes by directory name, not by path, so the prune is repository-wide rather than scoped to the Go fixtures the comment at `:99` cites.
- **`docs/site/**`,** via `IGNORED_PREFIXES` (`:117-119`).
- **Any extension outside the six patterns above,** including a plain `.tpl` file inside `charts/kube-agents/templates/`.

The Scope paragraph (`:38-42`) names the `.md` boundary. The `testdata` comment at `:99` says the opposite of what the code does, and the other two escapes are implied by the constants rather than stated.

**Fix.** Correct the `:99` comment to say the `testdata` prune applies at any depth repository-wide, and extend Scope to name the directory prunes and the extension boundary. The `gke-multitenancy/SKILL.md` policy is a separate defect and wants its own issue either way.

mplakhtiy (commented):
Three findings from a full pass over the branch: one HIGH on what the guard does not catch, two MEDIUMs on the gap between the checker's stated coverage and its behaviour.

scripts/check_iac_parity.py
shalinibhatiapl: Fixed.
- Port-53 egress rules now require both `UDP` and `TCP` protocols declared on port 53.
- When the wildcard `0.0.0.0/0` peer excepts `169.254.0.0/16` (the 5-entry house shape), explicit `169.254.20.10/32` (NodeLocal DNSCache) and `169.254.169.254/32` (Cloud DNS for GKE) literals are required peers.
- Rebased onto `upstream/main` to incorporate #1169 (`799c7e30`).
- Added 5 regression tests in `scripts/test_check_iac_parity.py` covering UDP/TCP requirements, house shape link-local literal enforcement, and RFC 1918-only passes.

scripts/check_iac_parity.py
shalinibhatiapl: Fixed. Updated the module docstring in `scripts/check_iac_parity.py` to record that #747 D1 is delivered as a static source-level audit rather than engine rendering (`helm template` / `kubectl kustomize`), and documented the rationale:
- Heterogeneity across static surfaces (`.yaml`, `.yaml.template` with shell variables, examples) that cannot all be rendered by Helm/Kustomize.
- Zero external tool dependencies and portability for offline execution under standard Python unit tests (`make test-python`, `python-tests.yml`).
- Baseline authoring guard across checked-in static files.

Also recorded the trade-offs and limitations against D1: conditional presence in Helm templates where stripped directives cause peers behind values to evaluate as present in source, and operator Go-generated policies (such as #1195) being outside the source scanner's scope.

scripts/check_iac_parity.py
shalinibhatiapl: Fixed.
- Extended the `Scope:` section in `scripts/check_iac_parity.py` to explicitly document the four escaped file classes: `*.md` files (noting the separate defect in `gke-multitenancy/SKILL.md`), unmatched extensions outside `DISCOVERY_FILE_PATTERNS`, directory prunes at any depth via `IGNORED_DIRS`, and prefix prunes via `IGNORED_PREFIXES`.
- Corrected the comment on `IGNORED_DIRS` to state that any directory named `testdata/` is pruned systematically at any depth repository-wide.
- Added test coverage in `scripts/test_check_iac_parity.py` (`test_discovery_ignores_testdata_at_any_depth` and `test_discovery_ignores_unmatched_extensions`).

jayantid:
Red-sweep classification: stale false red — this run (head 9dfe19b, pushed after the 09-04 02:01Z green) sat in the documented 2026-09-04 environmental window (#1249: impersonation PERMISSION_DENIED + Boskos 401s + infra-lost reps); the diff is a CI parity-check script and cannot cause this shape. Retesting.

/retest

mplakhtiy (approved):

scripts/test_check_iac_parity.py
```diff
@@ -0,0 +1,1617 @@
...
+        buf_err = io.StringIO()
+        with contextlib.redirect_stdout(buf_out), contextlib.redirect_stderr(buf_err):
+            rc = main(["-v", str(p)])
+        self.assertEqual(rc, 1)
+        self.assertIn(f"FAIL: {p}", buf_out.getvalue())
+        self.assertIn("ERROR: Static NetworkPolicy DNS egress parity check failed", buf_err.getvalue())
+
+
+if __name__ == "__main__":
+    unittest.main()
```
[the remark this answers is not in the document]
shalinibhatiapl: Refreshed the PR description to match the suite at HEAD: updated unit test count to 66/66 (0.591s) across all 6 discovery extensions, reflecting the commits that landed through subsequent review rounds.

PR #1092, opened 2026-09-01
feat(release)!: gate the GA publish on the staging promotion tag and an unattended-release decision

mplakhtiy:
## Summary

Two changes to the GA release path, which together are Phases R0–R2 of the weekly-release-promotion plan.

**The gate moved up a rung.** `verify_release_eligibility.sh` now requires a shape-matched `staging_<ts>_<sha>` tag instead of `rc_*_validated`. A staging tag means the full nightly E2E matrix passed on the commit; an `rc_*_validated` tag means only the narrow three-hourly suite did, which was never the bar a GA release was meant to clear.

**`release-publish.yml` gained an `evaluate-schedule` job** that decides whether an unattended run should publish, and the publishing job is conditioned on its verdict. The decision is three conditions over the tag graph: a candidate carries a shape-valid staging tag, commits exist between the newest GA tag and that candidate, and nothing in the range is a breaking change. The first two failing are skips that exit 0; a breaking change fails the job, because it recurs every run until somebody publishes by hand.

**No `schedule:` is added.** The gate reads a tag family only `nightly-pipeline.yml` produces, and that pipeline is dispatch-only too, so a weekly cron would skip green every Thursday and demonstrate nothing. `scripts/release/README.md` holds the ordering: nightly cron first, then `dry-run`, then `evaluate`, then the cron here.

## ⚠️ Read before merging: this blocks GA releases until the nightly runs

**There are zero `staging_*` tags in this repository today, and this branch makes one a hard requirement.** The moment it lands, every release dispatch — including the ordinary `bypass` one that last succeeded on 2026-08-28 — exits 1 two steps in:

```
❌ BLOCKED: Commit <sha> has NOT been promoted to staging!
   No tag matching 'staging_<ts>_<sha>' points to this commit.
```

That is the retarget working as designed, not a defect, but it is a release outage until `nightly-pipeline.yml` is dispatched by hand and pushes its first staging tag. Its four runs to date (all 2026-08-31) failed, so that is real work rather than a formality. The only other way through is `skip_staging_validation` with an audit reason, which is the emergency override for hotfixes and not a way to cut an ordinary release.

The plan sequences R1 after "the nightly pipeline is producing staging tags reliably". **This PR deliberately runs ahead of that sequencing** at the requester's direction, so the precondition is not met and the consequence above is the cost. `scripts/release/README.md` leads its rollout section with it, and step 1 of that section is getting the nightly green.

**Merging ahead of the tag is the deliberate choice, not an oversight**, and the outage window above starts the moment this lands. Clearing it: get the nightly green so it pushes a real `staging_<ts>_<sha>`; if it has not by Thursday 2026-09-03, the tag is created by hand from a fresh `rc_*_validated`. Hand-creating it costs something worth stating — the tag asserts the full matrix passed on that commit when it did not, it is a live deploy trigger (`staging-redeploy-*.yml` fire on the push), and it needs `RELEASE_BOT_TOKEN` or nothing fires at all. Either route unblocks the *gate*, not the first release: #865's `feat(install)!:` is inside the range, so the first release out still goes by hand with `schedule_gate: bypass`.

## Why This Change

A GA release happens today only when a human opens the workflow and clicks "Run workflow", and what they are asked to vouch for is a candidate that passed the three-hourly suite. Two separable problems: what gates a release, and when one happens.

**On the gate.** The `rc_*_validated` tag is the wrong evidence for an immutable, published artifact — it is the narrow suite, and the full matrix result already exists one rung up. The `rc_*_validated` condition is dropped rather than kept beside the staging one, because it is implied: `tag_staging_promotion.sh` refuses to promote a commit that does not already carry it. Requiring both would re-check a property the first guarantees and leave two gates to keep in step.

What replaces it as the defence against a fabricated tag is the **shape**. `staging-redeploy-*.yml` triggers on the bare `staging_` prefix, so a hand-typed `staging_hotfix` is a supported way to redeploy staging and must not read back as "the matrix passed here". `STAGING_TAG_SHAPE_REGEX` matches the timestamp and short SHA in the positions `staging_tag_for_rc` puts them. It stops an accident, not an attacker — nothing ties the 7-hex field to the commit the tag points at — and `common.sh` says so rather than overclaiming.

**On the schedule.** Automating the decision needs something to answer the question the human was answering, and the publishing path cannot answer it alone. It is worth being precise, because the obvious justification is wrong: on an ordinary quiet week the publishing path is already green — `calculate_next_version.sh` exits 0 with `has_changes=false`, and `verify_release_eligibility.sh` takes its idempotent-skip branch. Three narrower things are left:

- **Nothing stops for a breaking change.** An unattended run is exactly where one should not go out unwatched.
- **Two shapes do exit 1 with nothing to ship.** No staging tag anywhere in history; and a GA tag on a commit that is not the gated candidate's stamped child — what an emergency release leaves behind — which trips the "tag already exists on a different commit" collision. Condition 2 covers the second, which is the one that recurs. A workflow that is red on weeks with nothing to ship is one nobody reads.
- **The verdict is made in one place, on the tag graph**, rather than through a `gh release view` call and a commit-shape heuristic several scripts deep, and it is reported in the job summary.

There is deliberately no weekday or elapsed-time logic in the resolver: the cron is the cadence. "Has this already been released?" needs no condition either — a GA tag on the gated commit empties condition 2's range.

## What Changed

- `scripts/release/common.sh` — `STAGING_TAG_SHAPE_REGEX`, `get_latest_staging_tag` and `staging_promotion_tags_at_commit`, all shape-anchored. `get_existing_staging_tag` is now shape-anchored too, and has to be: it sets `skip_promotion` in `resolve_promotion_candidate.sh`, so a prefix match there told the nightly a `staging_hotfix`-carrying commit was already promoted while the release gate read the same commit as unreleasable. Also `release_read_commit_range` and `commit_messages_have_breaking_change`, shared with `calculate_next_version.sh` so the bump and the halt cannot be scoped to different commits or disagree about what "breaking" means.
- `scripts/release/verify_release_eligibility.sh` — the gate itself, retargeted.
- `scripts/release/calculate_next_version.sh` — auto-resolves the same staging tag, so the version is not computed for a different commit than the one published.
- `scripts/release/resolve_scheduled_release.sh` — the three conditions, emitting `should_release`, `release_commit`, `gate_tag` and `skip_reason`. Skips exit 0; a halt raises `::error` and exits non-zero.
- `scripts/release/decide_release_gate.sh` — which way into the workflow a run is taking. A `schedule` event always evaluates. A dispatch reads a new `schedule_gate` input: `bypass` (the default, and what every dispatch did before, emergency path included), `dry-run` (report the verdict, publish nothing) and `evaluate` (act on it, as a cron tick would).
- `.github/workflows/release-publish.yml` — the gate job, the verdict on the publish job's `if:`, and a `TARGET_COMMIT` fallback so an unattended run releases the commit the gate passed.
- `decide_release_gate.sh` also refuses `dry-run` or `evaluate` dispatched alongside a `target_commit`. Those two modes let the resolver pick the commit from the tag graph, while the publish job's `TARGET_COMMIT` prefers the input — so honouring both evaluated one commit's range, breaking-change halt included, and published a different one. Naming a commit is `bypass`, which is the default and the emergency path.
- Docs: `scripts/release/README.md` is canonical for the gate, the outage window and the rollout order; the site page, the SemVer design doc's flow diagram, the documentation map and the identifier-sources table are reconciled to it.

**BREAKING CHANGE:** the `skip_rc_validation` dispatch input is renamed `skip_staging_validation`, and `SKIP_RC_VALIDATION` renamed `SKIP_STAGING_VALIDATION`, to name the gate they now bypass.

## Context

Implements Phases R0, R1 and R2 of the weekly-release-promotion plan, less the cron itself — see the warning section above for why R1 is here rather than sequenced after the nightly, and why the cron is not.

Supersedes #970, which was closed. Two of its choices are deliberately not carried over: the weekday-anchor epoch arithmetic with an injectable clock, and treating a breaking change as a green skip.

**#1065 has merged, and this branch is rebased onto it.** It rewrote `docs/site/src/content/docs/deploy/release-versioning.md` wholesale around the pre-retarget contract — `rc_*_validated` as the gate, and two copy-pasteable `gh workflow run … -f skip_rc_validation=true` invocations in a new emergency runbook, which are the commands someone reaches for mid-incident and which would fail with an unknown-input error after this lands. The rebase reconciled the page rather than resolving it as text: a **Staging Promoted** row added to the taxonomy, the SemVer-calculation prose and release prerequisite retargeted, both runbook invocations renamed to `skip_staging_validation`, and the eligibility criterion rewritten. Every statement on the page was re-checked against source on this branch.

#1068 also touches `scripts/release/README.md`'s Workflow Mapping table; that one is a plain conflict. #1093 (`eLeontev`) touches `common.sh`, `tests/test_release_common.py` and `tests/testing/release.py`; its `common.sh` change is additive and adds no tag helpers, so it is a merge-conflict warning rather than a divergence.

## Testing

- New: `tests/test_resolve_scheduled_release.py` (17), `tests/test_decide_release_gate.py` (13), `tests/test_release_publish_workflow.py` (13), plus additions to `tests/test_release_common.py`, `tests/test_verify_release_eligibility.py` and `tests/test_calculate_next_version.py`.
- Full unit suite: **841 passed, 3 skipped, 626 subtests** (`pytest tests/ --ignore=tests/e2e --ignore=tests/integration`). `tests/integration/` is excluded because it fails locally on a missing `uvicorn`, unrelated to this branch.
- **Mutation-checked, not just green.** Five of the new tests were confirmed to fail on the regression they exist for, by making the change and running them: reverting `get_existing_staging_tag` to a prefix match; reverting each of `verify_release_eligibility.sh`'s two staging lookups to a prefix match; deleting the `target_commit` refusal from `decide_release_gate.sh` (2 failures); and reverting the resolver's blob-tag path to a bare `exit 1` (1 failure). Every mutation was reverted and the tree re-verified clean.
- `make docs-check` clean. `prettier@3.9.6 --check` clean on all changed Markdown/YAML. `shellcheck -x` clean on both new scripts; `common.sh` and `calculate_next_version.sh` carry only pre-existing SC2155/SC2129 warnings on lines this branch did not author.
- **actionlint now has run — in CI, and it passes**, along with `checkov`, `semgrep`, `zizmor`, `prettier` and the Python suite on the current head. It is still SIGKILLed locally (exit 137), so CI remains the first mechanical check of the step expressions and the `needs` graph; the earlier caveat that they were unchecked no longer holds.

### Live validation

**Not live-tested against a running installation, and it cannot be.** This is a CI workflow and the scripts behind it; nothing here reaches a cluster, an operator, or an agent pod — there is no CR status, Deployment env or in-pod file for it to change. Scheduled workflows also only run from the default branch, so the cron path is not exercisable from a branch at all.

What was exercised instead, against real `git` rather than mocks of it:

- Both new scripts run end to end against purpose-built repositories for every branch: no candidate, a candidate with nothing new, a candidate with new commits, a breaking change in the range, a breaking change already behind the last GA tag, and a repository with no GA tag at all.
- **The release-outage claim in the warning section was reproduced, not reasoned about**, in a scratch repository matching the upstream tag graph (a GA tag, an `rc_*_validated` candidate, no staging tag): on `main` the eligibility check prints `✅ ELIGIBLE: Found validated RC tag(s)` and proceeds; on this branch it exits 1 with `❌ BLOCKED … has NOT been promoted to staging`.
- The quiet-week and emergency-leftover claims were reproduced against the unmodified `verify_release_eligibility.sh`: the quiet week exits 0 on the idempotent skip, and the emergency-leftover shape exits 1 with "tag already exists in git repository on a different commit".
- The `pipefail` hazards were reproduced before and after: `echo … | grep -q` returns 141 on a 400 KB matching corpus and reads as "no match", while the herestring form matches.
- The stderr-capture bug was reproduced with a branch sharing a GA tag's name: `git log 0.1.0..HEAD --format=%s 2>&1` over an **empty** range returns `warning: refname '0.1.0' is ambiguous.`, which read as "there are commits to ship".

No test artifacts were left behind; all scratch repositories were created outside the checkout and are disposable.

## Self-Review

Four rounds, each in a fresh subagent that did not write the change — handed the diff range, the skill, and the `make docs-check` output, with the plan, reasoning and commit messages withheld. Rounds 1 and 2 covered the R0 gate against `rc_*_validated`; round 3 covered the staging retarget; **round 4 re-read the whole diff after the rebase onto #1065**, which is the state the branch is in now. Every finding was verified against source before being acted on, and two of round 4's fixes were mutation-checked by running the tests against the un-fixed code.

Round 4 additionally checked the rebase reconciliation of `release-versioning.md` statement by statement against source, and swept the repository for anything still depending on the pre-rename input.

**Angles covered** (adversarial): line-by-line error handling and fallback values; removed behaviour and weakened tests; cross-file tracing into `calculate_next_version.sh`, `verify_release_eligibility.sh`, `resolve_promotion_candidate.sh` and `common.sh`; permissions, action pinning, fork guard and the blast radius of an unattended publish; reuse and duplication; shell correctness under `set -euo pipefail` (SIGPIPE/141, stderr capture, `mktemp` cleanup on error paths); conventions and docs; test coverage, including mutation-checking the new tests; sibling/in-flight PRs. **Docs-drift**: every identifier in the new prose traced to source, skip-vs-halt semantics, generated regions, map coverage, and whether any other document states a fact this change invalidates.

Fixed, from round 4:

- **`schedule_gate: evaluate` with an explicit `target_commit` published a commit the gate never evaluated.** The publish job resolves `TARGET_COMMIT` as `inputs.target_commit || needs.evaluate-schedule.outputs.release_commit`, so the input wins — meaning the three conditions ran against the newest staging tag's commit and a different one shipped, with condition 3's breaking-change halt computed over a range that did not describe what went out. Both resolver-consulting modes refuse the combination now; `bypass` still takes a named commit, so the emergency path is untouched. Three tests, and deleting the guard fails two of them.
- **The rollout the README prescribes goes red at step 2 and publishes nothing at step 3.** #865's `feat(install)!:` is already on `main` and inside the range any first staging tag produces, so condition 3 halts both steps. The README's step 3 now says so and names the way out; the body's "once it merges" below is corrected to the present tense.
- **The resolver's two machinery failures skipped the outputs the gate job declares.** A bare `exit 1` left `gate_tag` and `skip_reason` empty for the one reader who needs them — the person asking why the release did not happen. Both leave through `emit_and_exit` now.
- **`common.sh`'s rc-family comment still named `verify_release_eligibility.sh` as anchoring there.** After the retarget it does not read that family at all.
- **`get_existing_staging_tag` reintroduced `| head -n 1`** four lines under the comment explaining why that shape is wrong under `pipefail`.
- `docs/README.md` still listed "RC validation gate" as a key concept of the page this branch retargeted; and a link the rebase added was labelled `nightly-pipeline.yml` while pointing at the release-scripts directory.

Fixed, from round 3:

- **The nightly's "already promoted?" lookup and the release gate disagreed.** A hand-pushed `staging_hotfix` made `get_existing_staging_tag` report the commit promoted, so `resolve_promotion_candidate.sh` set `skip_promotion` and the real staging tag was never pushed — while the gate, matching on shape, read that same commit as unreleasable. Silently unshippable candidate, nothing red. Both lookups are shape-matched now.
- **The emergency-override path of `calculate_next_version.sh` had no live coverage.** Its test still set the pre-rename `SKIP_RC_VALIDATION`, and passed anyway because the fixture had no staging tag, so the auto-resolve branch also fell through to `HEAD`. Fixed the variable, gave the fixture a promoted commit so the branches differ, and added the opposite case.
- **`verify_release_eligibility.sh`'s shape check had no test.** Swapping it back to a prefix match left the whole suite green while a hand-pushed deploy-trigger tag became release evidence. Two tests now, one per entry point.
- **`scripts/release/README.md` said a manual release "behaves as it always has".** It does not. The rollout section now leads with the outage and makes the nightly step 1.
- **The shape check was described as proving more than it does** — "a hand-made tag that does match names a real candidate" is false, since nothing ties the 7-hex field to the commit. Corrected in `common.sh` and the README.
- **The PR title carried no `!`.** This repository squash-merges with the PR title as the subject and an empty body, so the `!` on the commit would never have reached `main` — and the gate this PR adds would not have halted on this very change. Retitled.
- **The PR body described R1 as "deliberately not here".** Rewritten; this document is the corrected version.
- Two SC2034 warnings new in this diff (not pre-existing, as an earlier draft claimed); an `echo … | head -n 1` surviving in the file this branch rewrote for that hazard; and the resolver's condition-2 comment still restating an argument the README retracts.

Fixed, from rounds 1 and 2 (all still holding at the current head):

- **The change's central justification was contradicted by the code it cited.** All three copies claimed the publishing path goes red on a quiet week. It does not — it takes the idempotent-skip branch and goes green.
- **`$(git log … 2>&1)` captured stderr into the commit subjects.** git warns on *success* for an ambiguous refname, so an empty range read as non-empty and an unattended run would have published a release for a week with nothing in it. Both this gate and `calculate_next_version.sh` carried it; both now use `release_read_commit_range`.
- **The breaking-change regexes, and the range they applied to, were duplicated** between the gate and the version calculator, with a comment asserting agreement and nothing enforcing it. Both are single definitions in `common.sh` now, and a test fails if either script grows a copy back.
- **Two `echo | grep -q` constructs under `pipefail`** would report 141 — read as "no match" — on a large enough corpus, waving through a breaking change. Both use herestrings now.
- **With no GA tag the halt scanned all of history**, so any long-shipped `feat!:` wedged the gate permanently. A never-released repository now skips both remaining conditions.
- **The `TARGET_COMMIT` test pinned presence, not order**, so swapping the operands would have passed while letting the gate override an explicitly dispatched commit.
- **`gate_tag` and `skip_reason` were written but unreachable** — now job outputs.

Deliberately not fixed:

- **The wider `common.sh` lookup duplication.** The file now holds six near-identical "list tags by glob, filter, take one" bodies, and two of the older ones (`get_latest_validated_rc_tag`, `get_existing_rc_tag`) still use the `| grep | head -n 1` pipeline this PR's own comment condemns. On `get_latest_validated_rc_tag` the corpus grows monotonically, so it is a real if not-yet-reachable hazard. Collapsing all six behind a shared helper would widen this PR into code it does not otherwise touch; the one instance this branch authored is fixed. Worth its own change.
- **The release outage on merge.** Real, reproduced, and the largest thing about this change — see the warning section. It is a sequencing decision taken deliberately, not an oversight, and the fix is to run the nightly before merging rather than to change code here.
- **The gate job checks out full history even on `bypass`, which reads no git state.** Conditioning the checkout would leave the step below it invoking a script that is not on disk.
- **`AGENTS.md`'s fork-guard bullet names `rc-deploy-environment.yml`, which does not exist** (it is `deploy-environment.yml`). Real, pre-existing, and a one-word change to an always-loaded instruction file 0.1k under its context budget. Worth its own issue; not opened unilaterally.
- Pre-existing `SC2155`/`SC2129` warnings on lines this branch did not author. CI does not run shellcheck over `scripts/release/`.

Reported, not acted on:

- **No real Actions run.** actionlint now passes in CI, which retires the expression-syntax and `needs`-graph half of this caveat. What remains unobserved is runtime semantics: that `inputs.schedule_gate` renders empty on a `schedule` event, that `github.ref` is the default branch on a cron tick, and that job outputs propagate from a child process's `$GITHUB_OUTPUT` writes are reasoned from documented behaviour. `decide_release_gate.sh` pins the mode to `evaluate` on a schedule event regardless, and that is tested.
- **A hand-named RC tag (`rc_hotfix`, reachable through `rc-create-tag.yml`'s `rc_tag` input) maps to `staging_hotfix`, which the shape gate rejects** — leaving a genuinely promoted candidate unreleasable, with the resolver reporting "no candidate has passed the gate". Two deliberate manual choices are needed to reach it and the outcome is a visible skip rather than a bad release, so it is recorded rather than fixed.

## Risk & Rollout

**Not low, and the risk is the gate rather than the scheduler.** There is no `schedule:`, so no unattended run can occur, and `schedule_gate` defaults to `bypass`, so the gate job is a no-op on every existing dispatch. But the eligibility retarget applies to every path: until a staging tag exists, no GA release can be published except through the emergency override. See the warning section.

Reverting is deleting the two new scripts, restoring the workflow's single job, and restoring the gate in `verify_release_eligibility.sh`, `calculate_next_version.sh` and `common.sh` along with the input names.

Rollout, in order: dispatch `nightly-pipeline.yml` and get a real `staging_<ts>_<sha>` tag pushed; dispatch `release-publish.yml` with `schedule_gate: dry-run` and read the verdict in the job summary; dispatch with `evaluate`, which publishes a real GA release if the verdict says so; then add the cron. `scripts/release/README.md` holds the cron value and the reasoning.

Worth knowing before the cron goes on, and before the rollout steps above: **#865 is a `feat(install)!:` and it is already on `main`**, inside the range any first staging tag will produce. Condition 3 therefore halts steps 2 and 3 on their first run, and keeps halting every run until someone publishes that release by hand with `schedule_gate: bypass`. That is the designed behaviour, but it is the first thing anyone running the rollout will see rather than a rare event.

---


scripts/release/resolve_scheduled_release.sh
```diff
@@ -0,0 +1,189 @@
...
+fi
+
+if [ -z "${RELEASE_RANGE_SUBJECTS}" ]; then
+  # Two shapes reach here and both are "nothing to ship". The ordinary quiet week
+  # would be handled without this — verify_release_eligibility.sh recognises the
+  # GA tag as the stamped child of this candidate and skips green on its own — so
+  # what this saves there is a wasted checkout, version calculation and four
+  # registry inspections rather than a red run.
+  #
+  # The other shape is the one that goes red today: an emergency release put the
```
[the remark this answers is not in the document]
mplakhtiy: Confirmed and fixed, in two passes.

`64553362` removed "four registry inspections", which was the wrong half of the claim — you are right that both the idempotent skip and the emergency-leftover collision return before `check_commit_images_exist`.

You were also right that removing that was not enough. What `64553362` left behind still claimed the skip saves "the publish job's checkout", which the README retracts two paragraphs on ("not a checkout on balance either, since the gate job pays its own `fetch-depth: 0` on every run") — so the comment was still a restatement, and still a wrong one. `aa3b47d8` cuts the accounting entirely: the comment now says what condition 2 is *for* (the emergency-leftover shape, the one that goes red and recurs), notes that the quiet week reaches the same branch without needing it, and points at the README for what the skip is and is not worth. One home for the argument, which is what the file header asks for.

Two other things you flagged in passing are also fixed in `aa3b47d8`: the `echo … | head -n 1` at `verify_release_eligibility.sh:184`, and the two `SC2034` warnings, which were new in this diff rather than pre-existing as the body claimed.

Worth flagging since it postdates your read: the branch now also retargets the release gate from `rc_*_validated` to shape-matched `staging_<ts>_<sha>` tags and renames `skip_rc_validation` to `skip_staging_validation`. You reviewed `ada3802`, three commits back, so none of that has been read by you — `/review all` when convenient.

.github/workflows/release-publish.yml
```diff
@@ -1,24 +1,52 @@
 name: Release & Publish (GA)
 
+# Promotes a gate-passing commit to a GA release: SemVer from Conventional
```
eLeontev, line 3: Blocker: Do not merge till the first available `staging_xxxx` is created. 

Alter option (recommended): manually create a new staging tag on fresh `rc_**_valdiated` to not be blocked.

scripts/release/common.sh
```diff
@@ -289,11 +360,67 @@ staging_tag_for_rc() {
...
+# the tag list outgrows a pipe buffer.
+get_latest_staging_tag() {
+  local tags
+  tags="$(git tag -l --sort=-v:refname "${STAGING_TAG_PREFIX}*" 2>/dev/null || true)"
+  grep -m1 -E "${STAGING_TAG_SHAPE_REGEX}" <<<"${tags}" || true
+}
+
+# Lists the shape-valid staging promotion tags pointing at a commit, one per
+# line. Empty output means this commit has not passed the nightly matrix.
+staging_promotion_tags_at_commit() {
```
eLeontev, line 402: I was sure that we already have the tag function somewhere, please reuse  if valid.

eLeontev:
One additional point: can you validate the changes on your own fork to try the different scenarios in isolation (comment out all release related scripts and keep only infra ones) to check that new infra changes are properly work.

.github/workflows/release-publish.yml
mplakhtiy: Agreed, and it is the blocker. Not merging until a real `staging_<ts>_<sha>` tag exists.

Reproduced rather than reasoned about: `git tag -l 'staging_*'` is empty upstream, and 25 commits sit between `0.3.0` and the newest validated candidate. The first dispatch after merge — the ordinary `bypass` one — exits 1 two steps in.

**Plan: the nightly first.** Dispatch `nightly-pipeline.yml` and let a green matrix push the tag, so the tag means what the gate reads it as. If it is still not producing one by Thursday, the tag gets created by hand from a fresh `rc_*_validated` — your alternative. It works today: `a00e831`'s tree already declares the flat `staging_*` trigger, so `tag_staging_promotion.sh` passes all four of its guards on `rc_2609011232_a00e831_validated`. Two things to know if we end up there — the tag is a live deploy trigger, so `staging-redeploy-{agent,controller,integrations}.yml` fire on the push, and it needs `RELEASE_BOT_TOKEN` or nothing fires at all.

One thing your comment does not anticipate, and it applies either way: **a staging tag unblocks the gate, not the first release.** `0.3.0..a00e831` contains #865's `feat(install)!:`, so the first `dry-run` and `evaluate` still halt red on condition 3. That release goes out by hand with `schedule_gate: bypass`; the next GA tag empties the range and the gate behaves as documented. `scripts/release/README.md`'s rollout says so now, at `af0feb38` — it previously implied step 3 would publish.

scripts/release/common.sh
mplakhtiy: Checked, and there is nothing to reuse — but you were right that the file has a duplication problem, so I fixed the part this branch authored.

Searched `scripts/release/`, `scripts/`, `k8s-operator/scripts/` and the workflows. No pre-existing function answers either question these do:

- `get_latest_staging_tag` — "the newest shape-valid staging tag anywhere". Nearest relative is `get_latest_validated_rc_tag`, which is the same idea for the rc family.
- `staging_promotion_tags_at_commit` — "the shape-valid staging tags at a commit". Nearest relatives are `get_existing_rc_tag` and `is_rc_candidate_commit_already_validated`, both rc-family.

`staging_tag_for_rc` and `get_existing_staging_tag` are pre-existing, and `get_existing_staging_tag` is modified here to delegate to the new helper rather than carry its own lookup — reuse in the direction you asked for. You may be thinking of `staging_trigger_matches_at_commit`: staging-tag-shaped and sitting right there, but it answers whether a candidate's tree declares a trigger the tag would match, not which tags exist.

**Where the instinct lands.** `common.sh` now holds six near-identical "list tags by glob, filter, take one" bodies, and two of the older ones — `get_latest_validated_rc_tag` and `get_existing_rc_tag` — still use the `| grep | head -n 1` pipeline the comment on `get_latest_staging_tag` explains is wrong under `pipefail`. On `get_latest_validated_rc_tag` the corpus is every `rc_*_validated` tag in the repository and grows monotonically, so that one is a real hazard rather than a style point: at pipe-buffer size grep exits 141, the `|| echo ""` reads it as "no candidate", and the nightly skips green.

Fixed at `af0feb38`: `get_existing_staging_tag` no longer pipes into `head -n 1` — it was doing exactly that four lines under the comment condemning it. Left alone: collapsing all six behind a shared `_latest_tag_matching` / `_tags_at_commit_matching` pair. That is the right end state and it closes the hazard on the two rc-family ones, but it rewrites functions this PR does not otherwise touch, on the branch that retargets the release gate. Worth its own change — say the word and I will open it.

.github/workflows/release-publish.yml
mplakhtiy: Correcting my reply above: I misstated the decision. **This is merging now, ahead of the first staging tag** — not held until one exists.

The consequence is unchanged and is the one you flagged: from the moment it lands, no ordinary GA release can be cut until a `staging_<ts>_<sha>` tag exists, and the only way past in the meantime is `skip_staging_validation` with an audit reason. That is accepted deliberately rather than overlooked; the PR body's warning section leads with it.

The plan for clearing it is the one in my reply above, just not as a merge precondition: get `nightly-pipeline.yml` green and let it push a real tag. **If it still has not by Thursday, the tag gets created by hand from a fresh `rc_*_validated`** — your recommended alternative — with the two caveats I noted there (the tag is a live deploy trigger, and it needs `RELEASE_BOT_TOKEN` or nothing fires).

Everything else in my reply above stands, including that the first `dry-run`/`evaluate` still halts red on #865's `feat(install)!:` regardless of which route produces the tag.

eLeontev (approved):

Issue #1097, opened 2026-09-01
Agent endpoint slot saturation: ~6 of 10 concurrent-run slots held by unseen runs during smoke


jayantid:
Caught-by: evals

During PR #1089's smoke run (build 2094714569262895104, kube-agents-evals-5, 2026-09-01), the agent endpoint rejected every new run with `HTTP 429: Too many concurrent runs (max 10)` for a ~4-minute window (~10:40–10:44Z). The eval harness's task parallelism was 4, so at least ~6 slots were held by runs the harness did not start or had already given up on — the build log shows several 600–950s agent executions in flight in that window, consistent with leaked or stuck server-side runs holding slots past their useful life.

Questions to answer:
- What holds a slot, and when is it released — on terminal card state, on HTTP disconnect, on a timeout? Is there a slot reaper?
- Do delegation-heavy tasks (rca, gpu-stress, autoops triage) hold a slot for the whole card lifetime even after the harness stops polling?
- Is `max 10` per endpoint pod or global, and is it sized for 3-repetition eval traffic plus real users?

Grading impact is already handled separately: kube-agents#1095 classifies terminal 429s as infra so this cannot red a PR. This issue is the server side — saturation still costs a repetition of real signal every time it happens, and in production it would cost a user their request.

Evidence: seven rep-3 records with literal output `Error: HTTP 429 from agent endpoint: Too many concurrent runs (max 10)`, agent execution ~1.2s each; two sibling repetitions in the same run correctly tagged KUBE_AGENTS_INFRA_FAILURE via the retry path.

Fuxiao-Gao:
Three more instances of this saturation class from the #1023 lane's runs (2026-08-31, builds `2094518651444006912` and `2094522220385669120`, evals-4 and the 09-01 evals-5 re-run `2094787420116488192`): the failing repetitions' outputs read `API call failed after 3 retries: HTTP 429`, redding those jobs through rung 3 since #1095 hasn't merged. Evidence trails and the non-429 residue are in #1116, now cross-scoped to avoid duplicating this issue.

Fuxiao-Gao:
Disambiguation for future triage: two different resources have produced 429 storms in this ledger, and this issue owns only one of them.

[...]

Issue #1010, opened 2026-08-28
Platform worker marks its delegation card complete on fan-out: "Awaiting synthesis" delivered as the final answer


jayantid:
The platform worker can mark its delegation card **complete** when all it has done is fan work out to the cluster agents — delivering an intermediate status line as its final answer. First cleanly captured by the eval gate on 2026-08-27.

## The reproduction

Build [`2093054394793725952`](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/998/pull-kube-agents-smoke-test/2093054394793725952) (project `kube-agents-evals-2`), task `capacity-pinned-pool-probe`, card `t_470a97c5`. The user-visible answer, in full:

> Task `t_470a97c5` has finished. Here is the result:
> **Fanned out cluster investigation tasks for `inference-server` to each cluster agent. Awaiting synthesis.**

The card reached a terminal status ("finished") 129 seconds in, the harness delivered the result exactly as the delegation protocol promises — and the result is a progress note. "Awaiting synthesis" is by its own words not an answer, yet the worker completed its card on it. The question (name the pool, what limits scaling, where the HPA ceiling sits) was never answered; the eval's checks failed the run correctly.

## Why this is an agent defect, not an eval artifact

- A human asking the same question in chat gets the same experience: a confident "finished" hand-back containing no answer. The eval just made it measurable.
- The delegation-wait machinery is exonerated by the same run: other delegating tasks on the identical build settled with real folded results (`rca-remediation-pr` delivered a grounded RCA + PR link and passed).
- Contrast with slowness: in the same run the compliance audit worker ran past the harness's wait but **kept its card open** until it had a real result — the correct behavior under multi-level delegation. This card closed early.

## Shape of the bug

Multi-level delegation: platform worker → cluster agents. The worker treats "I dispatched the sub-tasks" as task completion instead of awaiting and synthesizing the sub-results. Whether that is a SOUL/prompt gap ("a card is complete when the *answer* exists, not when the work is *delegated onward*"), a kanban state-machine gap (no distinct state for awaiting-children), or a race (worker's completion check saw its own dispatch as progress) needs an owner who knows the worker loop.

## Frequency

[...]

PR #1112, opened 2026-09-01
perf(eval): widen the step-0 inert list to tests/ and terraform/

haoxuw:
## Summary

One line. `hack/ci-eval-pr.sh` keeps a list of paths that cannot affect an eval result; step 0 reuses a pull request's previous passing verdict when *every* file changed since the last green is on that list. The list had four entries. This adds two directories and three root config files:

```diff
-readonly REVALIDATION_INERT_PATHS='^((docs|\.github|examples)/|…)'
+readonly REVALIDATION_INERT_PATHS='^((docs|\.github|examples|tests|terraform)/|…|\.gitignore|\.gitattributes|\.prettierignore)$)'
```

Nothing else in the script changes. The predicate, the evidence checks and the log lines are untouched, and the existing 22 tests pass unmodified.

## Why only this

This began as an attempt to cut the presubmit's cost meaningfully, by caching model responses across the three repetitions each case runs. That work is closed (#1285) and this is what survived it. The short version of why:

**A cached response only helps if the same question is asked twice, and the eval never asks the same question twice.** Two repetitions of one case diverge in shape — a different number of tool calls, a different number of status polls — and the cache keys on the whole conversation, so one divergence early in a run costs every turn after it. Measured directly: an identical conversation replays 100%, but inserting a single extra turn near the start drops it to 11%, which is what the real suite produces.

**That divergence is not a bug to fix, because the model is deliberately not deterministic.** The agent's calls carry no temperature, so they run at the provider default, and three identical requests return three different answers. Setting `temperature=0` narrows that but does not close it, and it is not something we want: a nonzero temperature is what gives the agent its usable behaviour. The eval is measuring an agent that is allowed to vary, so replaying its answers was always going to be the wrong tool.

Re-keying 5,294 recorded turns under progressively looser matching confirmed there was nothing left to win there: even masking *every number* — which would be unsafe — merged 0.4% of them.

So the remaining saving is not "run the suite more cheaply", it is "do not run it when nothing could have changed". That is what step 0 already does, and widening its list by two directories is the whole of this change.

## Why `tests/` and `terraform/`, and nothing else

A directory earns a place by being unable to reach a run: **no `COPY` in either root-context Dockerfile draws from it, and no `hack/ci-*.sh` reads it.** `tests/` is the repository's own suite; `terraform/` provisions the projects a run leases rather than taking part in one. Neither ships in the agent image.

`scripts/`, `.agents/` and `admin_console/` are deliberately absent, and this is the part worth arguing with if you disagree. Each would add a little more saving. Each would also be a bet on what a directory will *hold* rather than a statement about what it *is* — a Python file added under any of them could later join the eval path, and the list would keep skipping it. A wrong entry here reuses a stale verdict silently; a missing one only costs an unnecessary run, so the list stays on the safe side of that trade.

An earlier draft inverted the whole thing — enumerate what the eval *measures* and skip everything else. It skipped considerably more and was abandoned because it fails **open**: a new directory nobody classified looks harmless and gets skipped.

## What it is worth

Small, and worth stating honestly rather than quoting the flattering figure.

Step 0 only runs inside a job Prow has already started, and Prow's `skip_if_only_changed` already skips the whole job for entirely-docs pull requests — 20.5% of recent merges. Step 0's remaining job is the case Prow cannot see: a pull request that *does* contain code, so the job starts on every push, but whose latest push does not.

```
Merged pull requests, >1 commit, that Prow does not skip : 33
  final push can revalidate — before                     :  2  (6%)
  final push can revalidate — after                      :  3  (9%)
```

One pull request in thirty-three. A hit saves the eval matrix; the job still pays its cluster lease, image build, deploy and teardown first.

The two larger levers are both outside this PR. Widening Prow's `skip_if_only_changed` would skip whole jobs rather than just matrices. And 59% of the eval's prompt tokens are already served from Gemini's prefix cache, with room to reach roughly 90% by moving from implicit to explicit caching.

## This PR runs the full eval, on purpose

It changes `hack/ci-eval-pr.sh`, and `hack/` is not on the inert list — so step 0 refuses to revalidate this pull request and the matrix runs.

That is the trust property working, not an oversight. The script's own header states it: a pull request that edits the revalidation logic touches `hack/`, so its own run goes full. If `hack/` were skippable, a change to the skip logic could skip its own verification, which is the one hole that must not exist. The run gives no signal about this change — the unit tests do that — but it is the right price for the property.

## Testing

- `python3 -m pytest tests/test_ci_eval_revalidation.py -q` → 23 passed. The 22 existing tests are unmodified: the predicate's semantics did not change, only the list it reads.
- The new test is the guard. It parses every `COPY` source out of both root-context Dockerfiles and every `scripts/` reference out of the eval drivers, then asserts none of them matches the inert list — using `grep -E`, the same tool the gate uses. It fails if a Dockerfile starts copying out of `tests/`, or if `agents/` is ever added to the list.
- shellcheck (`--severity=warning`), `bash -n`, ruff on the repository's rule set → clean.
- Pre-existing and unrelated: `tests/test_ci_eval_revalidation.py` cannot be imported by bare per-directory `unittest discover`, because it imports `tests.testing.common`. Identical on unmodified `main`.

### Live validation

None owed: `.agents/rules/eval_driven_development.md` exempts `hack/ci-eval-pr.sh` by name, and this changes only whether the matrix runs, never what the agent does. This pull request's own full eval run is the live exercise.

## Self-Review

Earlier heads of this branch carried a much larger change, and both required passes ran against it in subagents given only the diff; their findings are in this PR's comment history. What remains is a subset, minus the parts those findings and later measurement argued against:

- The inverted allowlist is gone — it failed open, and the guard test written to compensate could not close the gap for a new top-level directory.
- The blocker one pass found — a guard test written for pytest in a directory swept by `unittest discover`, which collected **0 tests while appearing to pass** — no longer applies; the remaining guard lives in the existing suite.
- A duplicate `export BOOTSTRAP_ADMITTED` that an earlier rebase introduced, which would have silently overridden `hack/eval/blocking-roster.txt` with a stale hardcoded list, is gone.

Claims I made in this thread and corrected after re-measuring: the saving was first quoted without accounting for Prow's job-level skip; the old list was described as never firing when it fires on 6% of pushes; and a 21% figure belonged to the fail-open shape, not to this one.

### DIFF
--- a/bench/kube_agents_bench/gate.py
+++ b/bench/kube_agents_bench/gate.py
@@ -411,9 +411,42 @@ def _cmd_suite(args: argparse.Namespace) -> int:
     #
     # A warning rather than a red. The name might belong to a case that is
     # legitimately absent from this run -- commented out of TASKS, or filtered
-    # -- and redding the job for naming a case it did not run would make the
-    # variable unusable for the transition it exists to cover.
+    # by hand -- and redding the job for naming a case it did not run would
+    # make the variable unusable for the transition it exists to cover. The
+    # one exception is automatic selection, handled below.
     unknown = sorted(_bootstrap_admitted() - {str(c.get("case")) for c in cases})
+
+    # Under EVAL_TASK_SELECTION=auto the "legitimately filtered" escape does
+    # not hold: the filter is a program running on every pull request, and a
+    # subset that dropped every case the gate arms on grades a suite in which
+    # nothing can reach a blocking rung -- a disarmed gate reporting green.
+    # The selection floor (EVAL_ADMITTED_CASES in hack/ci-eval-pr.sh) exists
+    # to keep an admitted case in every subset; this is the backstop that
+    # reds the run if that wiring ever drops one. The trigger is `unknown`
+    # -- an armed case selection FAILED TO GRADE -- not "zero admitted",
+    # which a store outage or stale evidence also produces with the floor
+    # intact and which must stay advisory for the reasons _load_store gives.
+    # When the baseline store replaces BOOTSTRAP_ADMITTED as the source of
+    # admission, extend `unknown` with the store's admitted ids.
+    disarmed = (
+        os.environ.get("EVAL_TASK_SELECTION", "") == "auto"
+        and bool(unknown)
+        and not any(c.get("admitted") for c in cases)
+    )
+    if disarmed:
+        verdict.green = False
+        verdict.reasons.append(
+            "automatic task selection dropped every case the gate arms on "
+            f"({', '.join(unknown)}); nothing this run graded could have "
+            "blocked the merge. The selection floor must carry the admitted "
+            "cases (EVAL_ADMITTED_CASES in hack/ci-eval-pr.sh)."
+        )
+        print(
+            "::error::EVAL_TASK_SELECTION=auto graded no admitted case: "
+            f"selection dropped {', '.join(unknown)}. The selection floor "
+            "must carry the admitted cases (EVAL_ADMITTED_CASES).",
+            file=sys.stderr,
+        )
 
     # Per-case notes, deduplicated. A misspelled EVAL_JUDGED_METRICS name is
     # one configuration mistake, not one per case, and repeating it fourteen
@@ -425,7 +458,7 @@ def _cmd_suite(args: argparse.Namespace) -> int:
     # silently loosens the gate, and the one thing that must not happen is a
     # green nobody knows was measured against nothing.
     banners = []
-    if unknown:
+    if unknown and not disarmed:
         print(
             f"WARNING: BOOTSTRAP_ADMITTED names no graded case: {unknown}",
             file=sys.stderr,
--- a/hack/ci-eval-pr.sh
+++ b/hack/ci-eval-pr.sh
@@ -1068,6 +1068,56 @@ TASKS=(
   # "./tasks/cluster-agent-crashloop-fix-request/task.yaml"
 )
 
+# ─── Change-based task selection ──────────────────────────────────────────────
+# hack/eval_triggers.yaml maps changed files to the tasks that must run;
+# anything unmapped, and any selector failure, keeps the full matrix.
+# Modes: shadow (default) only logs what auto would keep; auto filters TASKS;
+# off skips the call. NONE never shrinks TASKS: whole-job skips are Prow's
+# skip_if_only_changed, so NONE inside a running job reads as a disagreement
+# between the two layers, and a disagreement gets the full matrix.
+EVAL_TASK_SELECTION="${EVAL_TASK_SELECTION:-shadow}"
+case "${EVAL_TASK_SELECTION}" in shadow | auto | off) ;; *)
+  echo "ERROR: EVAL_TASK_SELECTION must be shadow, auto or off, got '${EVAL_TASK_SELECTION}'." >&2
+  echo "A typo silently behaving as shadow is a misconfiguration on a knob meant to be flipped remotely -- refusing." >&2
+  exit 1
+  ;;
+esac
+if [ "${EVAL_TASK_SELECTION}" != "off" ] && [ -n "${PULL_BASE_SHA:-}" ]; then
+  SELECT_ACTIVE=()
+  for TASK in "${TASKS[@]}"; do
+    SELECT_ACTIVE+=("$(basename "$(dirname "${TASK}")")")
+  done
+  # Three-dot: what the PR changes, not what main gained since it branched.
+  if SELECTION="$(git -C "${SCRIPT_DIR}/.." diff --name-only "${PULL_BASE_SHA}...${PULL_PULL_SHA:-HEAD}" \
+    | python3 "${SCRIPT_DIR}/eval_triggers.py" "${SELECT_ACTIVE[@]}")"; then
+    SELECTION_KIND="$(printf '%s\n' "${SELECTION}" | head -n 1)"
+    case "${SELECTION_KIND}" in
+      SUBSET)
+        SELECTED_NAMES="$(printf '%s\n' "${SELECTION}" | tail -n +2)"
+        KEPT="$(printf '%s\n' "${SELECTED_NAMES}" | grep -c . || true)"
+        echo "task selection (${EVAL_TASK_SELECTION}): ${KEPT} of ${#TASKS[@]} tasks: $(printf '%s' "${SELECTED_NAMES}" | tr '\n' ' ')"
+        if [ "${EVAL_TASK_SELECTION}" = "auto" ] && [ "${KEPT}" -ge 1 ]; then
+          SELECTED_TASKS=()
+          for IDX in "${!TASKS[@]}"; do
+            if printf '%s\n' "${SELECTED_NAMES}" | grep -Fqx "${SELECT_ACTIVE[IDX]}"; then
+              SELECTED_TASKS+=("${TASKS[IDX]}")
+            fi
+          done
+          TASKS=("${SELECTED_TASKS[@]}")
+        fi
+        ;;
+      NONE)
+        echo "task selection (${EVAL_TASK_SELECTION}): no task triggered; running the full matrix (job-level skip is Prow's skip_if_only_changed)"
+        ;;
+      *)
+        echo "task selection (${EVAL_TASK_SELECTION}): full matrix"
+        ;;
+    esac
+  else
+    echo "task selection: selector failed; keeping the full matrix (fail closed)"
+  fi
+fi
+
 # Floor for VerificationCorrectness on a repetition of a task that declares a
 # verification_spec. 1.0 while every declared objective is meant to hold
 # outright. Exported: bench-gate reads it, so it is a starting point to tune
--- a/hack/ci-eval-pr.sh
+++ b/hack/ci-eval-pr.sh
@@ -1068,6 +1068,56 @@ TASKS=(
   # "./tasks/cluster-agent-crashloop-fix-request/task.yaml"
 )
 
+# The transition bridge. bench/baselines/ ships EMPTY, so no case is admitted
+# and nothing can reach the collapse rung -- which would mean the presubmit
+# blocks on nothing for as long as screening takes. Cases named here keep their
+# old blocking behaviour meanwhile.
+#
+# It is a bridge and not a destination: a bootstrap-admitted case has no
+# measured evidence, so it arms rung 4 but leaves rung 6 quiet and contributes
+# nothing to main's side of the aggregate. Screening replaces it.
+#
+# This roster is what blocks a pull request once the Prow job stops being
+# optional. Thirteen of the seventeen active cases are admitted: the ones
+# whose recent record shows failures only on their own regressions or on
+# infra classes the harness already excludes from the verdict. Four are
+# held out -- they still run and report on every pull request, and they
+# cannot red one on a GRADED failure. The scope of that promise is rungs
+# 4 and 6: rungs 1-3 (a forbidden mutation, an erroring check, a record
+# that is not a real run) stay blocking for every case by design,
+# admitted or not -- see grade_case, which evaluates them before it reads
+# admission. security-overgrant-remediation-proposal (#1066) is simply
+# new: it earns its record like any case, then enters. The other three
+# each have a filed issue naming the exit condition:
+#
+#   capacity-pinned-pool-probe            -- #1010: worker completes its
+#     card at fan-out ("Awaiting synthesis" as the final answer). The
+#     failure is correlated across repetitions when the agent chooses to
+#     fan out, so the collapse rule does not absorb it. Enters when the
+#     fix merges.
+#   cluster-agent-healthy-workload-no-finding -- #1100: the agent invents
+#     a finding on a healthy workload ~1 run in 8. Main's own trait, so a
+#     collapse would tax an innocent PR. Enters when the false-positive
+#     rate drops or when rung-6 screening can compare against main.
+#   autoops-warning-event-triage          -- #1101: 0/5 graded repetitions
+#     on record; admitting it reds every pull request today. Enters when
+#     the lettered-options bar is settled and it has a clean record.
+#
+# If an admitted case reds a pull request its diff cannot explain on a
+# graded failure, demote it here and reference its issue. Demotion is a
+# one-line same-day edit to this list -- this file, not the Prow config,
+# is deliberately the fast lever. It is the lever for rung-4 reds ONLY: a
+# rung-1-3 red (mutation, erroring verifier, an empty record on a task
+# that provisions nothing -- a record whose deployer died before any
+# agent ran grades INFRA and reds nobody) does not stop when its case
+# leaves this list, because those classes signal a broken case or
+# install, not flake, and the fix is on that side.
+#
+# agent-kanban-smoke earned its seat back after the 08-27 redesign (a real
+# SRE question graded on kanban_create plus cluster names); the reds that
+# once argued for un-arming it belonged to the old vocabulary check.
+export BOOTSTRAP_ADMITTED="${BOOTSTRAP_ADMITTED:-reliability-pdb-probe,security-overgrant-probe,upgrades-lagging-master-probe,consistency-authorized-networks-probe,cost-idle-pool-probe,obtainability-remediation-proposal,rca-remediation-pr,compliance-rbac-overgrant,cluster-agent-crashloop-debug,cluster-agent-crashloop-misleading-symptom,cluster-agent-crashloop-evidence-chain,gpu-stress-test-diagnosis,agent-kanban-smoke}"
+
 # ─── Change-based task selection ──────────────────────────────────────────────
 # hack/eval_triggers.yaml maps changed files to the tasks that must run;
 # anything unmapped, and any selector failure, keeps the full matrix.
@@ -1088,8 +1138,10 @@ if [ "${EVAL_TASK_SELECTION}" != "off" ] && [ -n "${PULL_BASE_SHA:-}" ]; then
     SELECT_ACTIVE+=("$(basename "$(dirname "${TASK}")")")
   done
   # Three-dot: what the PR changes, not what main gained since it branched.
+  # Admitted cases ride along so they survive any subset: the gate's
+  # blocking rungs arm on admitted cases alone.
   if SELECTION="$(git -C "${SCRIPT_DIR}/.." diff --name-only "${PULL_BASE_SHA}...${PULL_PULL_SHA:-HEAD}" \
-    | python3 "${SCRIPT_DIR}/eval_triggers.py" "${SELECT_ACTIVE[@]}")"; then
+    | EVAL_ADMITTED_CASES="${BOOTSTRAP_ADMITTED:-}" python3 "${SCRIPT_DIR}/eval_triggers.py" "${SELECT_ACTIVE[@]}")"; then
     SELECTION_KIND="$(printf '%s\n' "${SELECTION}" | head -n 1)"
     case "${SELECTION_KIND}" in
       SUBSET)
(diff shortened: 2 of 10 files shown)


hack/eval_triggers.yaml
```diff
@@ -0,0 +1,61 @@
...
+# `extensions` limits what a bucket's wildcard patterns may claim; other
+# file types fall through. Literal patterns (no wildcard) skip the gate.
+# `**` crosses slashes; `*` and `?` do not.
+
+# Added to every non-empty selection: the cheapest task that proves the
+# platform->cluster-agent chain.
+floor:
+  - cluster-agent-crashloop-debug
+
+buckets:
```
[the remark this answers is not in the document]
haoxuw: Fixed in 766f1fa: `gpu-stress-test-diagnosis` is now in `floor`, so every non-empty selection carries the bootstrap-admitted case and the blocking rungs stay armed on filtered runs. The new test `test_the_floor_carries_every_bootstrap_admitted_case` parses `BOOTSTRAP_ADMITTED`'s default out of `ci-eval-pr.sh` and asserts each named case is in the floor, so a change to either default breaks loudly rather than silently disarming the gate. Good catch — the shadow-log evidence plan indeed could not have surfaced this.

haoxuw:
Holding the PR for now -- the concept is good, but may introduce too much noise in the current testing framework and perf measurements.

Will wait for the testing framework to finalize before action on this

bench/kube_agents_bench/gate.py
```diff
@@ -425,7 +458,7 @@ def _cmd_suite(args: argparse.Namespace) -> int:
     # silently loosens the gate, and the one thing that must not happen is a
     # green nobody knows was measured against nothing.
     banners = []
-    if unknown:
+    if unknown and not disarmed:
```
jayantid, line 461: 🟡 **Not merge-blocking — dormant until the Prow-side flip to `auto`, but fix it before that flip.**

`gate.py:461` still raises the "`BOOTSTRAP_ADMITTED` names no graded case" banner whenever `unknown` is non-empty and the run is not disarmed. Under `auto`, selection drops most of the 13-name roster by design, so `unknown` is non-empty on essentially every subset run.

Once `auto` flips, every selected pull request's verdict markdown opens with a banner naming ~11 admitted cases and saying "if that is a typo, the case it was meant to keep blocking is not blocking". Reproduced at this head: `auto` plus graded `cluster-agent-crashloop-debug` and `gpu-stress-test-diagnosis` prints an 11-name typo warning above a GREEN verdict. This file's own comments say a repeated banner teaches readers to skip banners — and the disarmed red would then land exactly where readers have learned not to look.

Under `auto`, demote the banner to a one-line stderr note; a graded admitted case already proves the wiring works:

```python
if unknown and not disarmed:
    if os.environ.get("EVAL_TASK_SELECTION", "") == "auto":
        print(f"note: admitted cases not selected this run: {unknown}", file=sys.stderr)
    else:
        ...  # existing warning + markdown banner
```

jayantid:
**Smoke red triage (gate-watch):** not caused by this PR (diff is CI eval-selection only — `gate.py`, `ci-eval-pr.sh`, triggers — it cannot alter agent behavior). [Build 2095167118659031040](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1112/pull-kube-agents-smoke-test/2095167118659031040): 5 cases NOT_A_REAL_RUN on Vertex 429s (#1097/#1184), `compliance-rbac-overgrant` collapse with the #1171 `['debug-binding']` signature, and an `rca-remediation-pr` collapse where the agent answered with a delegation notice + `kanban_create` instead of executing (same shape hit PR 1089's run an hour later — being tracked under the #1010 family, noted there). `/retest` after #1183 lands, or sooner if you're feeling lucky — some afternoon runs did pass.

jayantid (changes requested):
I think this is a good improvement. Lets hold off until gate is a solid looking state, then we can add this. Wdyt?

jayantid:
Red-sweep classification: stale false red — the 09-02 run red'd on the pre-#1191 empty-record class (#1184) plus the compliance-rbac-overgrant collapse documented environment-wide in #1171. Not this diff — but the branch now conflicts with main, so prow cannot run a retest until it is rebased.

haoxuw:
## Combined with #1771, and what the comparison showed

Rebased this branch onto `main` (314 commits) and folded in #1771's Step 0 change. Head is now `7d24627a`; #1771 is closed as superseded.

### The two approaches answer different questions

| | #1112 (this branch) | #1771 (folded in) |
|---|---|---|
| question | given we are running, **which tasks**? | do we need to run **at all**? |
| mechanism | changed files → task subset | delta since last green touches nothing the eval measures → reuse that green |
| default on an unknown path | full matrix (fail-closed) | skip (fail-open, guarded by a test) |
| active when merged | shadow only, needs a Prow-side flip | immediately |

They stack rather than compete: Step 0 exits before any cluster work, and selection narrows whatever survives it.

### Measured over the same 200 first-parent merges to `main`

```
                              #1112 says ALL     NONE   SUBSET
  #1771 skips whole matrix                39        7        0
  #1771 runs                             154        0        0

  #1771 alone skips the matrix entirely : 46 (23.0%)
  #1112 alone narrows to a SUBSET       :  0 (0.0%)
  #1112 alone says NONE                 :  7  (which by its own rule is the full matrix)
```

The selector never narrowed on real traffic. That is not a defect in it — I verified it behaves exactly as designed on synthetic inputs (`bench/tasks/x/task.yaml` → SUBSET, `docs/x.md` → NONE, `agents/…` → ALL). It is a fact about how merges land: in the last 200, none touched *only* `bench/tasks/<name>/**` or *only* `bench/tf/prebuilt/<stack>/**`. Merges bundle a task edit with code, docs or tests, and one unowned path widens to ALL.

So on today's traffic the saving comes entirely from Step 0. The selector's value is prospective — it pays off for task-development PRs, which is exactly the traffic shadow mode exists to measure. Worth keeping and worth measuring; not worth waiting for.

### The "two contradictory path tables" concern is narrower than it looked

An earlier review flagged that both changes maintain a path table in this script with opposite defaults. Having now read both: they are answering different questions, so the opposite defaults are correct, not contradictory. `eval_triggers.yaml` asks "which tasks does this file affect?" — unknown means run everything. `REVALIDATION_RUNTIME_PATHS` asks "can this file change the outcome?" — not on the list means no.

I considered unifying them into one table and decided against it, on the evidence: unification would force one of the two to take the wrong default, and it would make Step 0 depend on Python and PyYAML where it currently needs only `grep` — a robustness regression in the cheapest gate, buying 0% on the measured traffic. Each table now carries a comment pointing at the other.

### What the rebase actually required

More than a replay. `main` moved registration out of the `TASKS` array into `hack/eval/presubmit-cases.txt`, and rewrote the admission block this branch was relocating:

- `docs/designs/bench-case-format.md` — took `main`'s registration text and re-applied this branch's selector sentence.
- `hack/ci-eval-pr.sh` — `main` now derives and exports `BOOTSTRAP_ADMITTED` from `hack/eval/blocking-roster.txt`, which is what `41c4e60f` was hand-arranging. Kept `main`'s; the selector still reads it via `EVAL_ADMITTED_CASES`.
- `bench/README.md` — `main`'s copy had diverged wholesale (whole new sections); took it and re-applied only the `EVAL_TASK_SELECTION` inventory entry.
- `bench/kube_agents_bench/gate.py` — comment-only; merged both wordings.
- **`scripts/test_ci_eval_triggers.py` — a real break, not a textual one.** Its floor check asserted the floor task appears as a literal `"./tasks/<id>/task.yaml"` line in a `TASKS` array that no longer exists, so it failed on every floor entry after the rebase. Retargeted to `presubmit-cases.txt` in `ed84d2a1`.

### Verification on the combined head

- `scripts/test_ci_eval_triggers.py` 38 passed; `tests/test_ci_eval_revalidation.py` + `tests/test_eval_runtime_paths.py` 27 passed / 16 subtests — 65 together.
- `bash -n` and shellcheck (`--severity=warning`) clean on `hack/ci-eval-pr.sh`.
- The two layers do not interact in the diff: #1771's commit cherry-picked onto this branch with no conflict.
- `bench/tests/test_gate.py` could not run here — it fails collection with a `UnicodeEncodeError` on my machine, identically on unmodified `origin/main`, so it is a local locale problem rather than anything this branch did. It wants running in CI.

### Left for you

The **Self-Review** and **Live validation** sections of this PR body are now stale: they describe the pre-rebase branch, and the shadow-mode log line they promise to confirm has not been re-checked on this head. Both want rewriting before review. I have not touched the body.

haoxuw:
## Correction: I overstated the saving, and #1813 is why

#1813 (docs-only) never ran the smoke job at all — Prow's `skip_if_only_changed` skipped it at the job level, so Step 0 never executed. That prompted me to re-measure against Prow instead of in isolation, and the earlier figures in this thread were wrong.

### What I said before, and what is actually true

I reported the predicate inversion taking revalidation from 7.0% to 23.0% of merges. That measured the predicate **on its own**, as though every merge reached Step 0. Most do not: Prow decides first, and for a docs-only diff the job never starts.

Measured over the same 200 first-parent merges, asking who actually prevents the matrix:

```
  Prow skips the whole job (Step 0 never runs) :  41 (20.5%)
  Job runs; OLD Step 0 revalidated             :   0 (0.0%)
  Job runs; NEW Step 0 revalidates             :  18 (9.0%)
```

The old inert-path denylist was a strict subset of Prow's filter, so **its incremental value was zero** — every merge it would have revalidated, Prow had already skipped. That is worth knowing on its own.

### But merges are the wrong unit, and that cuts the other way

A merge's file list is the whole pull-request diff — exactly what Prow already sees. Step 0 exists for a different event: a **push onto an open pull request**, where the PR as a whole contains code (so Prow runs the job on every push) but the latest push does not. Merge history cannot show that, so the 9.0% above still measures the wrong thing.

Measuring the right one, over merged pull requests with more than one commit that Prow would not skip:

```
  eligible pull requests                                 : 25
  whose final push touched nothing the eval measures     :  4 = 16%
      #1787  docs/ci-health.md
      #1748  tests/conformance/...
      #1691  docs/site/...
      #1688  .agents/..., install.sh, tests/test_install_script.py
```

Each of those pushes costs a full ~2h matrix re-run today. That is the case this change makes reachable, and the old denylist could not reach it either — `tests/`, `.agents/` and `install.sh` are not on it.

(One caveat on that 16%: it counts only each PR's *final* push. Intermediate pushes with the same shape would also fire, so it is a floor, not the whole figure. I did not enumerate them.)

### Where the bigger lever actually is

Prow's filter skips the **whole job** — lease, image build, deploy, teardown and matrix. Step 0 skips only the matrix; the surrounding ~20–30 minutes are already spent by the time it runs. So widening `skip_if_only_changed` in `oss-test-infra` to cover `terraform/`, `.agents/`, `admin_console/` and the rest of `scripts/` would save strictly more per merge than anything this repository can do — it is simply not a change that can land here. Worth raising separately with whoever owns that config.

### What is in this branch

Unchanged in substance — 12 lines of behaviour, the allowlist plus `grep -Ev` → `grep -E`. Only the commit message moved, to state the saving against Prow rather than in isolation. The guard test and its derivation are as described in the previous comment.

haoxuw:
## Reduced to the minimal change, and the issues found are fixed

Head is now `35e6d07c`: `main` plus one commit, **232 insertions across 5 files**, of which 12 lines are behaviour. The task-selection machinery is no longer on this branch.

### Why the selector came off

Not because it is wrong — because measurement said it buys nothing yet and carrying it cost real safety:

- **0% on real traffic.** Over the 200 most recent first-parent merges it narrowed to a SUBSET zero times. Not a defect in it: I verified it behaves exactly as designed on synthetic inputs (`bench/tasks/x/task.yaml` → SUBSET, `docs/x.md` → NONE, `agents/…` → ALL). Merges bundle a task edit with code or docs, and one unowned path widens to ALL.
- **It is shadow-mode**, so it changes nothing until a separate Prow-side flip.
- **Rebasing it 314 commits introduced a real bug** (below) and required retargeting one of its tests, because `main` moved registration out of the `TASKS` array into `hack/eval/presubmit-cases.txt`.

Nothing is lost. Those commits are preserved at `7b8d904f`, rebased and green, and can be re-proposed on their own branch once shadow logs show the traffic they are for. Splitting them out also removes the two-path-tables concern entirely.

### Issues found and fixed

1. **A duplicate `export BOOTSTRAP_ADMITTED`, introduced by my own rebase.** The selector commits relocated a hardcoded admitted list above the selection block; `main` has since derived the same variable from `hack/eval/blocking-roster.txt`. Both survived, and the hardcoded one came first — so `${BOOTSTRAP_ADMITTED:-…}` at the roster site saw it already set and kept the stale value. The roster file would have been read, validated, and then **silently ignored**, with a 13-case list from an older `main` deciding what blocks instead. Gone with the selector commits; the tree now has exactly one export, the roster-derived one.
2. **The selector's floor check asserted a `TASKS` array that no longer exists** — it failed on every floor entry after the rebase. Fixed in `ed84d2a1`, which rides with the preserved selector branch rather than here.
3. **Stale figures in this thread.** Corrected in the previous comment: measured against Prow rather than in isolation.

### What this branch delivers, stated against what already exists

```
200 merges: Prow skips the whole job on 41 (20.5%); Step 0 adds 18 (9.0%)
```

And on the event Step 0 actually exists for — a push onto an open pull request whose overall diff contains code, so Prow runs the job every time — 4 of 25 eligible merged pull requests had a final push touching nothing the eval measures (#1787, #1748, #1691, #1688). Each costs a full matrix re-run today.

Worth being plain about the baseline: the inert-path denylist this replaces is a strict subset of Prow's `skip_if_only_changed`, so **its incremental value is zero**. This change is what makes the gate do anything at all.

### Robustness

The allowlist fails open where the denylist failed closed, so `tests/test_eval_runtime_paths.py` derives from the sources what belongs on it — `COPY` lines in both root-context Dockerfiles, and `scripts/` references parsed out of the eval drivers — and reds when the list misses one. Derived rather than listed, so it does not rot on a rename; an earlier draft carried a 34-entry hand-maintained roster and was dropped for exactly that reason. Writing it found two paths a hand-built list had missed: `tags.env` and `scripts/eval_rosters.py`.

Known limits, stated in the test's docstring rather than implied: a new top-level directory reaching a run *without* a Dockerfile `COPY` or an eval-driver reference is not caught, and `k8s-operator/Dockerfile` is out of scope because its sub-context makes its sources non-repository-relative.

### Verification on this head

`bash -n` and shellcheck clean; ruff clean on the repo's rule set; `make docs-check` clean; 27 passed / 16 subtests; the guard runs 4 tests under `unittest discover`, the runner that owns `tests/`. 100 generated pull-request scenarios driven through the real extracted gate: 100/100 matched expectation, 76 revalidated, **zero logic changes wrongly skipped**.

### The larger lever is not here

Prow's filter skips the whole job — lease, build, deploy, teardown and matrix. Step 0 skips only the matrix, so ~20–30 minutes are already spent by the time it runs. Widening `skip_if_only_changed` in `oss-test-infra` to cover `terraform/`, `.agents/`, `admin_console/` and the rest of `scripts/` would save strictly more per merge than anything this repository can do. Happy to draft that regex and its evidence separately.

The PR title and body still describe the selector and want rewriting for what the branch now contains.

haoxuw:
## Before/after, measured — and a correction to my own wording

I said in the previous comment that the old denylist's "incremental value is zero" and that this change "is what makes the gate do anything at all." The first is accurate only in the narrow sense I measured it — over *merge diffs*, against Prow. The second is too strong, and the push-level numbers show why.

Measured on the event Step 0 actually serves: a push onto an open pull request whose overall diff contains code, so Prow runs the job every time.

```
eligible pull requests (multi-commit, Prow would NOT skip)  : 33

  final push revalidated by the OLD denylist : 2  ( 6%)
  final push revalidated by the NEW allowlist: 7  (21%)
  incremental                                : 5 pull requests
```

So the old gate does fire occasionally — on a docs-only final push — just rarely, because `docs/` and root markdown are all it recognises. The five it misses and this change catches:

| PR | final push |
|---|---|
| #1748 | `tests/conformance/…` |
| #1688 | `.agents/…`, `install.sh`, `tests/test_install_script.py` |
| #1651 | `Makefile`, `docs/…`, `tests/test_make_help_targets.py` |
| #1642 | `scripts/pool_pressure.py`, `scripts/test_pool_pressure.py` |
| #1631 | `.agents/rules/…`, `.agents/skills/…` |

Every one is a follow-up push that cannot change what the agent does, and every one costs a full eval matrix today.

### Both statements, stated precisely

- **Over merge diffs, against Prow:** the old denylist adds nothing — it is a strict subset of `skip_if_only_changed`, so Prow had already skipped the job for everything it would have caught. The new allowlist adds 18 of 200 (9.0%).
- **Over push deltas:** old 6%, new 21% of eligible pull requests. This is the number that matters, because it is the event Step 0 exists for and the one Prow structurally cannot see.

### What is not measured

No CI run has exercised this yet; everything above is replayed against history offline. The per-fire saving is the eval matrix — the job still pays its lease, image build, deploy and teardown first — and I am taking the matrix's share of a 1.5–3.5 hour job from this repository's own documentation rather than timing it myself. The push counts include only each pull request's *final* push, so intermediate pushes of the same shape would add to them: 21% is a floor, not a ceiling. Sample is 33 eligible pull requests.

haoxuw:
To summarize: Tried MANY things. The core blocker is that our model's temperature has to be > 0 to ensure better UX. Eventually only one line was useful -- which is to expand a little bit on the ignore-test regex pattern

bench/kube_agents_bench/gate.py
haoxuw: No longer applicable to this PR: `bench/kube_agents_bench/gate.py` is not in the diff any more, and neither is the selector it was gating.

The branch was reduced to a single change — widening step 0's inert-path list by `tests/` and `terraform/`. The change-based task selection, its `eval_triggers` config, and this gate backstop were all split out and are preserved at `7b8d904f`. Your point stands for that work and should be settled there before any flip to `auto`: the reproduction you gave — an 11-name typo banner sitting above a GREEN verdict — is exactly the kind of thing worth fixing while the code is in front of someone.

Resolving here because the file this thread anchors to is no longer part of this pull request, not because the finding has been answered.

haoxuw:
@jayantid — your request has effectively been taken, just more bluntly than you proposed.

You asked to hold this off until the gate is in a solid-looking state. The change-based task selection, its `eval_triggers` config and the gate backstop are no longer on this branch at all. They are preserved at `7b8d904f` — rebased onto current `main` and green — and belong in their own pull request once the gate is where you want it. Nothing about them needs deciding here.

What is left is one line, and it does not touch the gate:

```diff
-readonly REVALIDATION_INERT_PATHS='^((docs|\.github|examples)/|…)'
+readonly REVALIDATION_INERT_PATHS='^((docs|\.github|examples|tests|terraform)/|…)'
```

Step 0 already reuses a pull request's previous green verdict when every file changed since is on that list. This adds `tests/` and `terraform/` (plus three root dotfiles), because neither can reach a run: no `COPY` in either root-context Dockerfile draws from them, and no `hack/ci-*.sh` reads them. The predicate, the admission path and the gate are untouched, and the 22 existing step-0 tests pass unmodified.

Worth knowing before you re-review, since it changes what this is worth:

- It is small. Measured against merged pull requests with more than one commit that Prow does not already skip outright, the share whose final push can revalidate goes from 6% to 9% — about one in thirty-three. Prow's own `skip_if_only_changed` already handles the entirely-docs case (20.5% of recent merges).
- It cannot disarm anything. The list only decides whether the *matrix runs at all* on a re-push; it has no bearing on which cases are admitted or on any rung.
- `scripts/`, `.agents/` and `admin_console/` were considered and deliberately left off. They would add more saving, but each is a bet on what a directory will later hold rather than what it is, and a wrong entry reuses a stale verdict silently.

Two review comments on the older content were addressed: the bot's floor finding was fixed at the time, and yours on `gate.py:461` is answered in its thread — the file is no longer in this diff, and your point still stands for the selector PR.

Happy to park this too if you would rather touch nothing near the eval until the gate settles. It is one line and it will keep.

hack/ci-eval-pr.sh
```diff
@@ -104,7 +113,7 @@ readonly EVAL_VERDICT_OUTCOME_NOT_EVALUATED="not_evaluated"
 # docs/ branch, `bench/OWNERS` must not match the OWNERS one, and a .md file
 # below the root (agents/**/*.md is prompt content shipped in the image)
 # must still run the eval.
-readonly REVALIDATION_INERT_PATHS='^((docs|\.github|examples)/|[^/]+\.md$|(LICENSE|OWNERS|OWNERS_ALIASES)$)'
+readonly REVALIDATION_INERT_PATHS='^((docs|\.github|examples|tests|terraform)/|[^/]+\.md$|(LICENSE|OWNERS|OWNERS_ALIASES|\.gitignore|\.gitattributes|\.prettierignore)$)'
```
jayantid, line 116: 🟡 **Technical design.** Three of the five new entries make this list looser than Prow's `skip_if_only_changed`, and the header still promises it is only ever stricter.

Line 107 says the list may be stricter than Prow's filter, so the worst divergence is an unnecessary full run. That still holds for `tests/` and `.gitignore`, which Prow already skips (oss-test-infra#2686). It does not hold for `terraform/`, `.gitattributes` or `.prettierignore`. None is in Prow's filter today, and the comment at line 98 gives evidence for the two directories only.

After merge, Prow runs the job for a terraform-only or `.gitattributes`-only push, and step 0 reuses a green for that same push. `.gitattributes` is the one that can bite. It governs how every other file is materialised at checkout, so it can change bytes in the image without appearing in any `COPY`, and the new guard test cannot see it.

Either land the three in Prow's filter first, the #2686 route, and keep the subset property, or ship `tests/` and `.gitignore` now and drop the three. Rewrite the header sentence either way.

fkc1e100:
**Triage note:** This PR has been on `do-not-merge/hold` for 22 days. 

@haoxuw — Is this hold still needed? If so, please add a comment explaining why. If not, please `/hold cancel` to release it to the merge pool.

haoxuw:
not very valuable anymore, cancelled

PR #1134, opened 2026-09-01
fix(chat): deliver card artifacts the sandbox hides and Chat cannot attach

### LINKED ISSUE #999: Google Chat deployments behind the credential proxy cannot deliver task file attachments
## What happens

On a Google Chat deployment that reaches Chat through the credential proxy, a specialist that completes a kanban task with a deliverable file cannot get that file to the person who asked. The gateway notifier posts the completion summary and then:

> ⚠️ Couldn't attach **<file>.md**.
> File attachments are unavailable on this deployment: uploading to Google Chat needs a per-user OAuth token, and this install reaches Chat through the credential proxy, which holds no user credentials.
> Ask me to paste the contents here if you need them in chat.

The document itself is left on the agent host under `/opt/data/kanban/attachments/<task_id>/`, where a chat user cannot reach it.

## Why this matters

`kanban_complete`'s own schema tells workers that files listed in `artifacts` "will upload … as native attachments to the human who subscribed to the task, so the deliverable lands in their chat alongside the summary instead of being a path they have to fetch by hand." Workers follow that contract — they produce the document, list it, and reference it in their summary ("Full details in the attached proposal document") — and the promise then fails at delivery. The fallback puts the burden on the user to notice the warning and ask for a paste.

## Reproduction

1. Install kube-agents with the Google Chat integration on a deployment where outbound Chat traffic goes through the credential proxy (no per-user OAuth token available to the gateway).
2. In a Chat thread, ask the agent for any analysis that naturally produces a document, e.g. "Write a design assessment and attach it as a markdown file."
3. The delegated task completes; the notifier posts the summary, then the "Couldn't attach" warning; the file remains at `/opt/data/kanban/attachments/<task_id>/` on the host.

## Expected behavior

One of, in order of preference:

1. The attachment uploads through credentials the deployment actually holds (e.g. app credentials where the Chat API permits media upload), so the schema's promise is kept.
2. When upload is impossible, the notifier automatically inlines the content below a size threshold (or posts it as a follow-up message), instead of a warning that asks the user to request a paste.
3. At minimum, the `kanban_complete` schema wording is deployment-aware, so workers on such installs are not instructed to promise attachments that cannot be delivered.

dshnayder:
## Summary

A card that produces a file does not deliver it on a Google Chat install, for two independent reasons, and this PR fixes both. The agent's shell now runs in a separate sandbox pod, so the gateway cannot see the file at all and drops it without a word; and even when it can see it, Chat's `media.upload` refuses app authentication, so there is no attachment to send. The fix stages each declared artifact out of the sandbox for the length of the delivery, then pastes text deliverables into the thread and names the agent's own path for everything else.

## Why This Change

Issue #999 asks for one of three things, in preference order. Taking them in turn is what determined the scope here.

**Option 1 — upload through credentials the deployment holds — is not reachable.** Chat's media upload is user-auth only, by API design: _"This method doesn't support app authentication with a service account. Authenticate with a user account."_ The escape hatches are per-user OAuth held by the credential proxy (feasible; the paste-the-code flow needs no public callback, so this is not blocked on anything architectural — but it is a `client_secret` Secret, durable refresh-token storage, new proxy endpoints, and unstubbing `_handle_setup_files_command`, which is roughly a week and a new class of stored credential) or domain-wide delegation (small code change, large blast radius). Neither belongs in a fix for this issue.

**Option 2 — inline the content below a size threshold — is what this PR does.**

**Option 3 — make the `kanban_complete` schema wording deployment-aware — is not needed.** It existed to push workers onto a different channel. Staging recovers the artifacts workers already declare, so nothing about worker behaviour has to change. It stands or falls on its own now.

Without this, a worker that produces a report on a Chat install delivers nothing — or, before the sandbox, a path on a pod filesystem the person in the thread has no way to read.

## What Changed

**Staging the file out of the sandbox.**

- `agents/platform/scripts/sandbox_artifact_patch.py` — new. Wraps `GatewayKanbanWatchersMixin._deliver_kanban_artifacts`. When the sandbox is on, it reads each declared artifact out of `platform-agent-shell-0` over `sandbox_exec.read_bytes()` as the `agent` login, writes it to a temp directory in the gateway, hands the original method the local paths, and deletes them in a `finally`. `original_path()` maps a staged copy back to the path the agent wrote, so nothing user-facing names a temp file that no longer exists.
- `agents/platform/scripts/sandbox_exec.py` — `read_bytes()`, and the `_READ_INCOMPLETE` exit code that keeps a failed far-side read from arriving as a successful zero-byte file.
- `agents/platform/scripts/sitecustomize.py` — `UNCONDITIONAL_PATCHES`, deferred the same way as the relay patches but applied whatever the chat wiring is. The gateway drops sandbox artifacts for every adapter, so gating this on a relay URL would leave a direct-token install with the same silent loss.

Four ceilings bound the copy: 8 MiB per file, 16 files per card, 16 MiB across the card, and a 120-second deadline for the lot. Over any of them the file is skipped with a warning and the rest of the card still delivers. Paths on the gateway's own media-delivery denylist are refused before anything is read, and strict media-delivery mode refuses staging outright.

**Pasting it into the thread.**

- `agents/platform/scripts/google_chat_relay_patch.py` — `patch_adapter_class` now also replaces `_post_attachment_fallback`. Both of `_send_file`'s give-up paths (no user OAuth token, and a token the API refused) route through it, which on this deployment is every attempt.
- `deploy/docker/patches/apply_google_chat_attachment_notice.py` and its test, plus the `Dockerfile` comment — the notice is now the residual path, not the universal one; its relay branch no longer offers a paste the code has just declined; and it resolves a staged path through `original_path()` so a direct-token install names the agent's path too.
- `deploy/docker/patches/verify_slack_relay_registry_contract.py` — pins `_post_attachment_fallback`'s and `_deliver_kanban_artifacts`' upstream signatures, the two `filter_local_delivery_paths` helpers the staging screen imports, and `MESSAGE_CHAR_CAP` against the shipped adapter's real `_MAX_TEXT_LENGTH`, all at image-build time. It reads `_deliver_kanban_artifacts` through `inspect.unwrap`: the check runs with `/opt/defaults/scripts` on `PYTHONPATH`, so `sitecustomize` arms its finder in the verifier's own interpreter and the staging patch is installed before the attribute can be read. The patch therefore carries `functools.wraps`, and the verifier fails if the shim it finds has no `__wrapped__` to follow — otherwise the pin compares the shim against itself and guards nothing.
- `docs/site/src/content/docs/concepts/chatops.md`, `docs/README.md` — the end-to-end walkthrough now says what a thread does with a file deliverable and what the declaration and the ceilings are, and the identifier-source table points at both modules that own a limit.
- `docs/designs/agent-shell-sandboxing.md` — the sandbox design document owns the declared-writeback decision, and this change answers part of it, so it now records what crosses the pod boundary and why. Its `kanban_complete(artifacts=[...])` section splits the defect into the half that raises (managed scratch, still open, `kanban_attach` still the workaround) and the half that was dropped in silence (everything else, closed here), the five open questions under "declared writeback" are answered for the card caller with the limits and the cleanup owner, and the surrounding claims of "no workaround" and "nothing crosses" are narrowed to what is still true.

The paste behaviour: `.md`, `.txt`, `.json`, `.yaml`, `.yml`, `.csv`, `.log` up to 32 KiB are posted as message text. Anything else — a PDF, an image, an oversize file, bytes that turn out not to be UTF-8, a file that cannot be read at all — falls through to the existing notice.

Implementation decisions worth a reviewer's attention:

**Runtime patches, not build-time ones.** The obvious place for the paste is beside `deploy/docker/patches/apply_google_chat_attachment_notice.py`, which already rewrites this same method. Both went in runtime patches instead. A build-time patch needs a new `patchlib` anchor asserted by the Dockerfile, and every anchor is another way a base-image bump breaks the build rather than degrading. The relay patch also scopes the paste correctly for free: it only runs where `GOOGLE_CHAT_RELAY_URL` is set, which is exactly the set of installs that cannot attach. What that costs is a signature the tests cannot pin, so both wrapped methods are in the build-time signature gate.

**Staging is unconditional; the paste is not.** The two halves fix different bugs for different populations. Everything the sandbox hides is hidden from Slack and from a direct-token Chat install as well, so the staging patch registers whatever the chat wiring is and decides for itself, per delivery, whether the sandbox is on — memoised after the first answer, because the check reads the terminal config.

**Chunking is done in the relay patch rather than handed to `send()`.** The adapter chunks anything over its 4000-character cap itself, so passing it the whole report would work — except for fencing. A fence opened in the first message and closed in the last leaves every message between them rendering as prose, and for a JSON blob or a log line that is worse than useless, because Chat's markdown eats the underscores and asterisks. So the split happens before the fence goes on, on a line boundary where there is one. Markdown reports are not fenced — they are meant to be rendered, and read worse in a box.

**Every budget is derived from the adapter's cap, not written down beside it.** `_payload_budget()` subtracts a bounded header reserve (and the fence, when fenced) from `MESSAGE_CHAR_CAP`, so what goes on the wire is what the cap allows. The filename inside the header is middle-truncated at 80 characters, because a reserve is only sound if what it reserves for is bounded and an agent-chosen filename is not. A caption is unbounded prose and cannot ride on a reserve at all, so it leads in its own message whenever it would not fit.

**It returns `success=True` when the content lands, and hands back to the notice when it does not.** A refusal on any message — including a `send` that returns `None` or raises — stops the rest and calls the original notice, so the thread is left with the agent's path rather than a fragment and silence.

**Neither patch can take the gateway down by failing.** Looking up the original method is `getattr(..., None)`: a base image that renamed `_post_attachment_fallback` would otherwise lose `connect`, the relay loop, and Chat entirely, over a cosmetic feature. `_stage` never raises — a malformed path, a full `/tmp`, an unreachable sandbox and a read that dies mid-file each drop that one artifact and leave the rest of the delivery alone — because the notifier writes its cursor after delivery, so an exception out of the coroutine makes that subscription retry the same card forever.

## Context

Closes #999.

#913 (agent shell sandboxing) merged on 2026-09-05 as `f66d1957`; `main` is merged into this branch and the staging work is in this PR rather than a follow-up. Two things had to be true for that, and both are now: `sandbox_exec.py` exists on `main`, and the sandbox's effect on this path has been measured rather than inferred.

**What the sandbox does to artifact delivery.** With `terminal.backend: ssh` the agent's file tools run in `platform-agent-shell-0`, against that pod's own RWO volume, so a file a worker writes is not visible to the gateway process. The drop happens a layer above the adapter: `GatewayKanbanWatchersMixin._deliver_kanban_artifacts` screens every candidate path with `os.path.isfile` before the adapter is involved at all, that call is `False` for every sandbox-written file, and the method returns on the branch whose comment reads "the path may have been mentioned for reference only". No upload, no upload error, and no notice — for every artifact type and every adapter, Slack and direct-token installs included. This is design-doc finding T10.3 in `docs/designs/agent-shell-sandboxing.md`, deferred out of #913.

Measured on `kage-management` before the fix: a stub adapter installed into the running gateway received zero calls for a card that declared a sandbox-written artifact, and `os.path.isfile` on that path returned `False` in the gateway process. See **Live validation**.

**What `kanban_complete` does with a declared sandbox path**, since the design doc says it rejects the call outright and that would make this patch inert. It does not, except in one case. `kanban_db.py:5700-5716` validates only artifacts under the card's *managed scratch* `workspace_root`; anything outside takes `persisted.append(artifact); continue` with no `is_file()` check. Probed in the running gateway:

| Declared path | Managed scratch? | Outcome under the sandbox |
| --- | --- | --- |
| `/opt/data/INVENTORY.md` | no | passes `kanban_complete`, silently dropped by the notifier → **this fixes it** |
| `/opt/data/scratch/x/report.md` | no | same |
| `/opt/data/kanban/workspaces/t_abc/report.md` | yes | `kanban_complete` raises `ArtifactPreservationError` → out of scope, and named as such in the module docstring |

`agents/platform/SOUL.md:23` tells workers to declare the first shape, so the common case is the one this recovers. The design document was wrong about this and about four other things this change settles; it is corrected in this PR — see **What Changed** and **Self-Review**.

**What this does not recover.** `_deliver_kanban_artifacts` has three artifact sources, and only the first, `event_payload["artifacts"]`, is an explicit list of paths. The other two hand a block of the model's prose to `BasePlatformAdapter.extract_local_files`, which applies its own `os.path.isfile` *inside* the adapter — so those paths are gone before any patch outside it can see them. A path recovered from prose is dropped exactly as it is dropped today.

## Testing

- `agents/platform/scripts/sandbox_artifact_patch.py` — 33 tests in `test_sandbox_artifact_patch.py`. Cover: the wrapper leads back through `inspect.unwrap` to the method it replaced, with the same signature, which is what the build-time pin depends on; staging replaces the declared list and the caller's payload is not mutated; the sandbox-off path is a straight passthrough; a `SandboxUnavailable`, a `SandboxMisconfigured`, a `None` read, a NUL in the path, a full temp directory and an oversize file each drop one artifact and leave the rest; the per-card count cap, the per-file cap, the aggregate byte budget and the deadline; `~` expands against the sandbox login's home, not the gateway's; the denylist screen fails closed when the gateway's helpers cannot be imported and refuses everything under strict media-delivery mode; the staged copies and their `_ORIGINALS` rows are gone after the delivery, including when it raises; two files with the same basename do not collide; and the stale-directory sweep at install.
- `agents/platform/scripts/sandbox_exec.py` — 41 tests in `test_sandbox_exec.py`, including a `ReadScriptTestCase` that runs the real generated shell program against real files: a readable file round-trips, a missing and an unreadable one give the `_READ_UNREADABLE` code, and a read that fails partway gives `_READ_INCOMPLETE` rather than an empty success.
- `agents/platform/scripts/google_chat_relay_patch.py` — 35 tests in `test_google_chat_relay_patch.py`. They drive `install()` and the registry wrapper rather than calling `patch_adapter_class` directly, so the registration path is the one production takes. Cover: a markdown report inlines and the notice does not also post; the header carries filename and size; the paste is threaded; a caption leads; a binary, an oversize file, a non-UTF-8 file, a missing file and an empty file each defer to the notice; structured formats get a fence per chunk; every part of a multi-message report is marked; no message exceeds the cap however long the caption; a refusal, a `None` return or a raise each leaves the notice and the agent's path; an adapter missing the hook still gets the transport; and the notice names the agent's path rather than the staged copy.
- `deploy/docker/patches/test_google_chat_attachment_notice.py` — 31 tests, including the build-time notice's own staged-path translation and the case where `sandbox_artifact_patch` is not importable at all, which is what the stock upstream image the applier runs against looks like.
- The relay suite is a genuine regression test, not a shape check: run out-of-tree against the pre-fix module it goes red (5 failures, 8 errors), green against this head.
- `make test-python` — every directory green except `tests/`, which is red for two modules this diff does not touch: `test_install_script.py` (KMS key import) and `test_generate_release_sbom.py` (a Linux-only CI swap branch). A clean `upstream/main` worktree is red in the same two modules with the same errors. Per-directory: `agents/platform/scripts` 1743 OK, `deploy/docker/patches` 986 OK, `tests/integration` 31 OK, `agents/chat/scripts` 240 OK.
- `make docs-check` — clean. `prettier --check` clean on the three Markdown files in the diff.

### Live validation

Three runs on `kage-management`, the one install here with a real Google Chat connection. Each restored the install afterwards and deleted its scratch tags from Artifact Registry. Run 3 is the one that reaches this branch's head; runs 1 and 2 cover ground it does not, and are kept for that reason.

**Run 1 — the paste, 2026-09-02.** Image `us-central1-docker.pkg.dev/agentic-harness-demo/demo/platform-agent:ka999-20260902a` (and the matching `credential-proxy` tag, since the operator derives the sidecar from the agent image path). The operator was not part of the swap and stayed on `k8s-operator:dev-737m-20260901a`. Restored to `dev-737m-20260901a` afterwards, with Chat reconnected on the restored image (`✓ google_chat connected`).

A Vertex AI outage during the window — 503s that predate this image by a day and hit the unrelated cron watcher thread too — kept the model from driving a task to `kanban_complete`, so the notifier was driven directly instead of through a model turn: `_deliver_kanban_artifacts` called inside the running gateway pod, against the real adapter class the gateway had already patched, over the real credential proxy, into `spaces/6_pNByAAAAE`. The adapter was wired without calling `connect()` on purpose — its relay long-poll would have competed with the live gateway for `/v1/chat/events` on a shared install — so the transport is the same `RemoteResource` facade the live adapter uses, taken from `connect`'s closure. The credential proxy logged 7 consecutive `POST /v1/chat/api` → 200 at 13:13:32–13:13:36 Toronto, one per message the run sent, into thread `spaces/6_pNByAAAAE/threads/9UiA9ATNI0s`.

What rendered in the thread, confirmed by a human reading it:

| Artifact | Size | Result |
| -------- | ---- | ------ |
| `pool-utilisation.md` | 528 B | 1 message, 553 chars — inlined, Chat rendered the Markdown |
| `pool-metrics.json` | 370 B | 1 message, 409 chars — fenced, rendered as a Chat code block |
| `fleet-audit.md` | 9,560 B | 3 messages (3864 / 3808 / 1891 chars) headed `📄 fleet-audit.md (9.3 KB · 1 of 3)`, `(2 of 3)`, `(3 of 3)`, reading continuously across the split |
| `topology.pdf` | 193 B | the notice, naming `/opt/data/scratch/ka999/topology.pdf` |

chat-thread-inline-md-json

_Published at commit 5716e3f84919 — pre-captured image (ka999-chat-1.png)._

The top of that thread is the first two rows: `pool-utilisation.md` inlined with Chat rendering the Markdown — the bold headings and the bullet list are Chat's, not literal asterisks — and `pool-metrics.json` opening as a Chat code block rather than prose, which is the fencing decision under **What Changed**. Google Chat needs an authenticated browser session, so this is a `screencapture` published through the script's `--file` form rather than a headless capture.

chat-thread-pdf-notice

_Published at commit 5716e3f84919 — pre-captured image (ka999-chat-2.png)._

The bottom of the same thread is the tail of `fleet-audit.md`'s third message running into the `topology.pdf` notice, which is the fourth row and the case that proves the change did not simply inline everything. It also shows the rewritten relay wording — the notice says why the file could not be pasted instead of inviting a paste the code had already declined — and the pre-existing `**topology.pdf**` asterisks described below.

chat-thread-part-marker

_Published at commit 5716e3f84919 — pre-captured image (ka999-chat-3.png)._

The third capture is the seam between messages 2 and 3 of `fleet-audit.md`, which is where the part marker lives: the second message ends on the `Finding 37: cluster-037` heading and the third opens with `📄 fleet-audit.md (3 of 3)` and then the body text that heading introduced. The size rides on the first part only. The split landing between a heading and its paragraph is the line-boundary rule doing its job — no message ends mid-word and no fence is opened in one message and closed in another.

The `.pdf` case is the one that matters alongside the others: a change that inlined everything would look identical in the first three rows. The multi-part case covers the part markers and the per-chunk fence, which is the part no unit test can show rendered.

**Run 2 — the staging, 2026-09-04.** Same install, image `…/demo/platform-agent:dev-999stage-102d8e92` — this branch's commits cherry-picked onto #913's head with the staging patch on top. `kage-management` is now registered in `~/.config/kube-agents/live-test-envs.json` and this run was taken under a real lease (`scripts/live_test_lease.py acquire` … `release`), closing the gap run 1 had.

- **Negative control — same image, staging disabled.** A stub adapter installed into the running gateway received zero calls for a card whose completion declared a sandbox-written artifact, and `os.path.isfile` on that path returned `False` in the gateway process. That is the measurement behind **Context**.
- **With staging on.** The thread received the deliverable's contents under a `node-pool-review.md (539 B)` header, byte-identical to the file in `platform-agent-shell-0`, ending on the line "This file was written inside platform-agent-shell-0, not in the gateway pod." No staging warnings and no "Skipping unsafe local file path" in the gateway log, and no leftover staging directory in `/tmp` afterwards.

Afterwards the install was restored to `platform-agent:3d604b09…` with all three images verified back on that tag and the gateway 3/3 Running, the scratch tags were deleted, and the lease was released.

**Run 3 — the branch head, 2026-09-05.** Same install, under a lease, on an image built from this branch: `…/demo/platform-agent:pr1134-55f0209c` and the matching `credential-proxy` tag. The operator was not swapped, and neither was the sandbox StatefulSet — `resolveShellSandboxImage` in `k8s-operator/internal/controller/shell_sandbox_manifests.go` does not derive the sandbox image from `spec.deployment.tag`, so `platform-agent-shell-0` kept running the install's own image and the two-pod split under test was the real one. `55f0209c` and the current head `17a489a4` differ only in `docs/designs/agent-shell-sandboxing.md`; no code moved after this run.

The build is itself the evidence for the verifier fix, since the check only runs inside the image: `verify_slack_relay_registry_contract [unwired]: all checks passed` and `[wired]: all checks passed`, where the previous head reported `_deliver_kanban_artifacts: already wrapped before its signature could be read` eleven times.

Then, inside the running gateway pod:

- **`sandbox_enabled()` is `True`**, and a file written in `platform-agent-shell-0` as the `agent` login (uid 1000) is `visible to the gateway: False`. The premise holds on this install, not just in the probe from run 2.
- **`sitecustomize` installs the patch with nobody calling `install()`.** `PATCH_FLAG` on the mixin reads `False` before `import gateway.platform_registry` and `True` after it. That is the wiring `UNCONDITIONAL_PATCHES` adds, exercised in the real interpreter rather than in a test harness.
- **`inspect.unwrap` recovers the upstream method through the installed shim**, and its signature is the one the build-time gate pins. Same mechanism as the build check, observed in the process that actually serves chat.
- **A declared sandbox artifact is staged and arrives byte-identical.** `_deliver_kanban_artifacts` was driven for a card declaring `/opt/data/ka999-live.md`; the local copy handed to the original method matched the sandbox file byte for byte.
- **The denylist screen refuses a credential.** `/home/agent/.ssh/id_ed25519` was dropped before any read, with the warning, while the same card's legitimate artifact still delivered.
- **Cleanup is real.** After the delivery the staged file was gone and `original_path()` no longer knew the translation; no `sandbox-artifact-*` directory was left in `/tmp`.
- **`original_path()` names the agent's path during delivery**, sampled inside the recorder rather than after cleanup: `/opt/data/ka999-live.md`, not the `/tmp/sandbox-artifact-…` copy. This is the fix for A11/D5, measured on the path a user would read.

One thing worth reporting that is not a pass or a fail: the gateway pod took about 11 minutes and one startup-probe restart to reach 3/3 on the test image, sitting on `Mirroring the profile layout into the shell sandbox...`, against about 60 seconds on the rollback. That is #913's mirror step, which this PR does not touch, and it is slow against a freshly-pulled sandbox volume rather than a warm one — but a reviewer seeing a restart during a rollout should know it was seen here too.

Afterwards the install was restored to `platform-agent:3d604b09…` with the gateway 3/3 Running on that tag, both `pr1134-55f0209c` tags were deleted from Artifact Registry, the probe file was removed from the sandbox, and the lease was released (`status` reads `FREE`).

**What none of the three runs reaches.** The Google Chat leg was not exercised end-to-end on this head. Run 1 posted into a real thread and a human read the rendering, but that was the pre-review notice wording on the pre-staging image; driving a card all the way from the model to a rendered message on this head needs messages sent into and read out of a live Chat space, which I do not do without asking. Runs 1 and 2 also predate the round-2 rewrites — run 1's notice is one clause short of what ships, and run 2's staging module has since moved its reads off the event loop, grown a timeout and a deadline, lost the stale-copy shortcut, and gained the denylist screen, the byte budget and the startup sweep. Run 3 covers all of those except the rendering.

## Self-Review

Two rounds of both required passes, each pass in its own subagent handed the diff range and nothing else — no context that wrote the change reviewed it. Neither pass was skipped in either round. Two changes landed after round 2 and are described at the end of this section; they have not had a pass of their own.

**Round 1 (the paste, pre-#913-merge).** 17 distinct findings from 11 adversarial + 8 docs-drift, 3 overlaps collapsed. Angles B, E, G, I came back clean on the adversarial pass, which also verified the new suite red-then-green out of tree.

Blocking and High, all fixed:

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | **HIGH.** The flat 3500-character payload budget bounded the chunk, not the message. Caption + header + fence + chunk went over the adapter's 4000 cap at a 455-character caption; a 518-character one measured 4064, and `send()` then re-split it through the middle of the fence — the exact outcome the budget existed to prevent. | Fixed. `_payload_budget()` derives from `MESSAGE_CHAR_CAP`, the filename is bounded, and an oversize caption leads in its own message. |
| 2 | `_post_attachment_fallback` was not in the build-time signature gate, whose docstring claims to cover every callable the relay replaces. | Fixed. Pinned in `verify_slack_relay_registry_contract.py`. |
| 3 | Two docs said the notice fires on _every_ attachment. Its own motivating example, a 9,867-byte `.md`, is now inlined. | Fixed. The patch docstring and the `Dockerfile` comment describe the notice as the residual path. |
| 4 | The same docstring called inlining unbuilt future work, one directory over from the code. | Fixed in the same rewrite. |

Medium, all confirmed and fixed: only parts 2..N were marked, so a report whose second message was refused read as complete (every part now carries its marker); a refused `send` returned without calling the original notice, so the user got nothing at all — a regression against pre-change behaviour (every refusal now returns through the notice); the notice's relay branch invited a paste in precisely the cases the new code had just declined to paste (rewritten to say why); nothing pinned the chunk budget against the adapter's real `_MAX_TEXT_LENGTH` (the build-time verifier now parses it out of the shipped adapter); a `kib = 1024` magic constant (named); `_inline_text` discarded the suffix and byte count, so the caller re-split and re-encoded to render a header (it returns a `_Deliverable` now); and three documentation gaps.

Settled by reading the base image out of a running gateway pod, since `send`/`_create_message`/`_send_file` are not in this repository: `send()` raises rather than returning on a 429 and on any status it has no branch for, and nothing between there and the notifier catches it — **a real regression, fixed**, the paste is wrapped and a raise falls back to the notice. `send()` never returns `None`; that branch stays as belt-and-braces against a base-image change. The sequential round-trips the paste adds are the ones the adapter already makes for long text — no change.

Also corrected by reading the running pod: the relay patch attaches to `hermes_plugins.google_chat_platform.adapter.GoogleChatAdapter`, not the `plugins.platforms.google_chat.adapter` the verifier imports. Those are the *same file* loaded under two module names, so the pinned signature and the `_MAX_TEXT_LENGTH=4000` the verifier reads are the runtime ones — but the two class objects are distinct, which is why `_credential_proxy_relay_patched` reads `False` on the one the verifier holds. The gate is sound; the aliasing is a trap for the next reader.

**Round 2 (the staging, after the merge).** 15 findings from the adversarial pass, 13 from docs-drift, overlaps collapsed. Everything below is fixed unless marked otherwise.

Blocking and High:

| # | Finding | Disposition |
| --- | --- | --- |
| A2 | **Up to 16 blocking SSH calls on the gateway's event loop with no timeout.** `sandbox_exec.py`'s own docstring says the model owns the `agent` login's `~/.bashrc`, so `sleep infinity` there freezes chat for every adapter with no way to cancel; without a hostile file, a sandbox rollout stalls chat ~45 s per artifact. | Fixed. `asyncio.to_thread` plus a 30-second per-read timeout and a 120-second deadline for the whole card. |
| A3 | **A stale gateway-side file shadowed the sandbox artifact and the wrong report was delivered, silently.** `sandbox_mirror.py` copies the agent pod's directories into the sandbox at the *same absolute paths* and never deletes the source, so a card regenerating `/opt/data/scratch/audit.csv` in the sandbox found the pre-migration copy locally, skipped staging, and delivered last month's audit as this month's. | Fixed. Staging now reads from the sandbox whenever the sandbox is on; the local file is no longer a reason to skip. |
| A4 | **Nothing contained an exception out of `_stage`.** A NUL in a model-composed path reaches `subprocess.run` and raises out of the coroutine; the notifier writes its cursor after delivery, so that card retries forever and the subscription stops notifying permanently. `mkdtemp` on a full `/tmp` does the same. | Fixed. `_stage` cannot raise: the temp directory is created inside the guard and every per-path failure drops that artifact only. |
| A5 | **8 MiB × 16 = 128 MiB into a 2 Gi `emptyDir` whose overflow is a pod eviction, not `ENOSPC`.** No aggregate cap, and no sweeper for directories left by a delivery killed mid-flight. | Fixed. `STAGE_TOTAL_MAX_BYTES` bounds the card, and `install()` sweeps `sandbox-artifact-*` directories older than an hour. |
| A6 | `_deliver_kanban_artifacts`' signature was restated in the wrapper with no build-time pin — the exact failure `verify_slack_relay_registry_contract.py` was written after (#718's keyword-only `scope` broke every platform plugin with green tests). | Fixed. Pinned, along with the three `filter_local_delivery_paths` helpers. The verifier runs in an interpreter where `sitecustomize` has already installed the shim, so it reads the method through `inspect.unwrap` and fails if there is no `__wrapped__` to follow — see the last part of this section for why the first attempt at this was wrong. |
| A7 | **A failed far-side read arrived as a successful empty file.** `head -c N \| base64` makes the pipeline's status `base64`'s, which exits 0 on the empty input a failed `head` hands it — so a file removed between the readability test and the read was staged as a 0-byte report. | Fixed. The read lands in a scratch file first and a failure exits `_READ_INCOMPLETE`. A test runs the real shell program. |
| A9 | **Staging launders every path past the gateway's delivery filter**, because the path it inspects is now always a fresh `/tmp` directory. | Fixed, but not as proposed. Screening the *originals* through `filter_local_delivery_paths` cannot work: reading the base image showed `validate_media_delivery_path` calls `resolve(strict=True)` and `is_file()`, which reject every sandbox path by construction. Instead the denylist half is applied to the original path — `_path_under_denied_prefix`, `_MEDIA_DELIVERY_DENIED_PREFIXES`, `_MEDIA_DELIVERY_DENIED_HOME_SUBPATHS`, imported by name so the values cannot drift — the screen fails closed if any of them cannot be imported, and strict media-delivery mode refuses staging entirely. |
| A10 | `~` expanded against the *gateway's* HOME for a path read in the sandbox, so `~/report.md` read the wrong file or none. | Fixed. Expanded against the sandbox login's home. |
| A11 / D5 | `original_path()` had no test, and on a direct-token install nothing called it at all — the build-time notice rendered the staged `/tmp` path that `_cleanup` had already deleted. | Fixed. The build-time patch resolves through `original_path()` behind a guarded import, with tests on both halves and on the image where the module is absent. |
| D1–D4, D13, A13, A14 | Six stale docstrings and comments — `sitecustomize`'s "both relay patches", its false claim that the sandbox decision happens at gateway-import time, `TERMINAL_PRINCIPAL`'s "two callers … neither consumes the output" rule that `read_bytes` breaks, `run()`'s "two callers", `google_chat_relay_patch`'s "already on the agent's own disk", and `test_sitecustomize`'s summary line — plus INFO→WARNING on the import failure and memoising `_sandbox_on()`. | Fixed. |
| D8 / A15, D9 | `chatops.md` understated the constraints: declaration is now required, the ceilings were unmentioned, and "the agent host" now denotes the sandbox pod. | Fixed, with an identifier-source row for the second module. |

**Rejected, with evidence.** A1, the adversarial pass's blocking premise: that `kanban_complete` rejects sandbox artifacts outright, which would make this patch inert. It does not, except for managed-scratch paths — see the table in **Context**, probed in the running gateway. The finding is correct *about the design document*, which is D6/A12 below and the strongest argument for fixing it. A13's circular-import concern was also refuted against the image: `gateway/kanban_watchers.py` and `gateway/platform_registry` import no part of each other.

**Accepted, not fixed:**

- **A8 — unbounded stdout buffering.** `run()` captures the far side's output whole, so a `~/.bashrc` printing without limit is an OOM of the gateway. This is a property of every `sandbox_exec.run()` call rather than of this caller, and the agent pod trusts the sandbox it dialled. Documented as a known limit in `run()`'s docstring; capping it is a hardening change of its own.
- **Round 1, finding 12 — deliverable contents now enter Google Workspace retention and Vault unredacted**, where before only a filename and a host path did, and the allowlist includes `.log`, `.yaml` and `.json`. The mechanism is certain; the exposure depends on what a deliverable contains. Accepted as the cost of the fix — the alternative is not delivering the report — and written up under **Risk & Rollout**. It is not new in kind: the `result` field of every completed card already posts to the thread verbatim.

**Review findings from @jayantid, both marked not merge-blocking:**

- **The relayed notice under-described its own trigger set — confirmed, fixed.** It gave three reasons a paste was missing (no text form, too large, unreadable), but a refused or raising `send` routes to that notice too, and on that path the file is perfectly pasteable. A 429 at part 2 of a valid 10 KiB `.md` left the reader three reasons that were all false for their file. The line now names a failed post as a fourth reason, with a test on it.
- **That the `/hold` is unnecessary.** My first reply to this was wrong — I said that under #913 this code "degrades to today's behaviour without help (`open()` raises, `_inline_text` returns `None`, the notice posts)". It does not; nothing in the PR was reached, because the `isfile` screen drops the path before the adapter is involved. Corrected in the thread with the measurement. The finding's substance is now moot: #913 has merged, the ordering constraint is gone, and the staging fix is in this PR. @dmitryshnayder has lifted the hold.

**What the passes could not cover, and what a reviewer should not assume was checked:**

- **No mutation-based verification** — the passes shared one checkout, so neither was allowed to edit.
- **The base image** was unreadable from the round-2 pass's machine, so every claim it made about upstream came from this repository's documentation of it — which is how A1 happened. I settled A1, A9 and A13 afterwards by pulling `nousresearch/hermes-agent@sha256:3811ed13…` and reading the files.
- **The sibling-PR scan (angle J) did not run inside either pass**, because `gh` cannot execute on the machine they ran on. Run by hand: no open PR touches `google_chat_relay_patch.py`, `sandbox_artifact_patch.py`, `sandbox_exec.py`, `apply_google_chat_attachment_notice.py`, or `verify_slack_relay_registry_contract.py`. Overlap is confined to high-traffic shared files — `docs/README.md` and `docs/site/src/content/docs/concepts/chatops.md`. That is merge-conflict risk, not competing work.
- **The Google Chat rendering was not re-observed on this head** — see the last paragraph of **Live validation**. Everything else in that section was.

**Two changes since round 2, neither of which has had a pass of its own.**

- **The build-time gate the round-2 fix added did not work, and CI is what found it.** A6's fix pinned `_deliver_kanban_artifacts` and tried to keep `sitecustomize` from arming its finder by resolving the mixin before importing `gateway.platform_registry`. That is not winnable: `unwired()` imports the registry on its own first line, so the patch is always installed before `upstream_callables()` runs, and the `PATCH_FLAG` guard turned the whole check red — eleven times on `b9887892`. The fix stops racing import order and reads through it instead: `functools.wraps` on the shim, `inspect.unwrap` in the verifier, and a failure if the shim it finds has no `__wrapped__`. The two paragraphs above the code say why, because the next reader who deletes `@functools.wraps` as cosmetic will not get a red build — they will get a gate that passes whatever upstream does. A unit test now holds the `unwrap` identity, and the image build reports both phases green. Worth a reviewer's eye: the guard that fails when the shim is unwrappable is the only thing standing between this and a silently useless check.
- **`docs/designs/agent-shell-sandboxing.md` is updated in this PR** rather than left to a follow-up. Round 2 accepted this as not-fixed on the grounds that it is #913's file and its author is this PR's reviewer; @dmitryshnayder has since asked for it to be corrected here. What went in is under **What Changed**. It closes A12/D6 and D10–D12: `:1985`'s "no workaround in this design" is narrowed, the five open questions at `:1998-2003` are answered for the card caller with an explicit warning against reading them as the general answer, and `:2652`'s `kanban_attach` workaround is kept but scoped to the half that still raises. The two pre-existing errors round 2 noticed alongside them (`:1734`'s "the one place in the repository", `:2613`'s hard-coded call-site count) are **left alone** — they are not this change's, and correcting them would put edits in that file that nothing in this diff motivates.

## Risk & Rollout

Two blast radii, and they are different.

**The staging patch is on for every install with the sandbox enabled**, whatever the chat platform. Within that it touches only the path that is already broken — an artifact the gateway cannot see, which today is dropped in silence. Nothing that delivers today changes: a gateway-visible file takes the same code path it always did, and a card that declares nothing never reaches the staging branch. Its failure modes are all bounded to one artifact: a read that fails, times out, exceeds a ceiling, or names a denied path drops that file with a warning and delivers the rest. The one resource it consumes is up to 16 MiB of the agent pod's `/tmp` for the length of a delivery, deleted in a `finally` and swept at startup if a pod died holding it.

**The paste is contained to Google Chat installs that use the credential proxy relay**, and within those, to the path that was already failing. A file outside the allowlist or over the cap still gets exactly the notice it gets today, and a refusal partway through a paste now also gets that notice.

**One thing an operator should decide before enabling this:** the contents of a deliverable now go into a Google Chat message, which means Workspace retention, Vault, eDiscovery, and export — unredacted, and for `.log`, `.yaml` and `.json` as well as prose. Before this change the same thread carried only the filename and a host path. This is not a new kind of exposure for the product (a completed card's `result` field already posts verbatim), but it is a new volume of it, and an install whose agents write logs or manifests containing secrets is putting those in the retention corpus. There is no per-deployment switch for it in this PR; disabling it means not running the relay patch, which means not having Chat.

The worst new _functional_ failure mode is a report that is unpleasant to read in a thread — many messages, or fencing that suits the content badly — not a lost or corrupted one.

Revert is the commit; there is no state, no migration, and no config to unwind. Both patches are runtime-applied inside the agent image, so rolling back means rolling back the image tag.

One pre-existing rendering bug is visible in the captures and is **not** from this branch: the notice reads `⚠️ Couldn't attach **topology.pdf**.` with literal asterisks, because Chat's markup uses `*single*` for bold. That text comes from `deploy/docker/patches/apply_google_chat_attachment_notice.py` (#825, `68e6735e`) and is out of scope here.

---


agents/platform/scripts/google_chat_relay_patch.py
```diff
@@ -1,4 +1,5 @@
-"""Credential-free Google Chat transport for Hermes' bundled adapter."""
+"""Credential-free Google Chat transport for Hermes' bundled adapter, and the
```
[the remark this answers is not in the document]
dshnayder: All four points addressed; PR body updated.

**Screenshots.** Fair — the section had none and did not say why. Three now, published through `scripts/pr_evidence_screenshot.sh --file` and stamped at `5716e3f84919`: the top of the thread (`pool-utilisation.md` inlined with Chat rendering the Markdown, `pool-metrics.json` opening as a code block), the seam between messages 2 and 3 of `fleet-audit.md` showing the `📄 fleet-audit.md (3 of 3)` part marker, and the bottom (the `topology.pdf` notice, which is also the case proving the change did not just inline everything). The `--file` form rather than a headless capture because Google Chat needs an authenticated browser session; the body says that.

**Operator version.** Added. It was never part of the swap — `kubeagents-system/kubeagents-controller-manager` stayed on `k8s-operator:dev-737m-20260901a` throughout, and only the agent and credential-proxy tags moved.

**The lease.** Not taken, and the body now says so plainly instead of leaving it to inference. `kage-management` is not registered in `~/.config/kube-agents/live-test-envs.json` and holds no lease ConfigMap in `kubeagents-system`, so `live_test_lease.py` was enforcing nothing and nobody else held it either. What stood in for it: the install's owner granted the window explicitly, the prior image tags were recorded before the patch and restored after, and `connect()` was deliberately not called so its relay long-poll could not compete with the live gateway for `/v1/chat/events`. That is weaker than the lease, and the body says that too.

**`a89275cb`.** Correct — a pre-rebase SHA the rebase erased. Both tables now read `4b1b3d91`, with finding 16 attributed to `59155cba`, which is where the `chatops.md` fix actually landed.

dshnayder:
/hold

Holding until #913 (agent shell sandboxing) merges — see **Context** in the description for why the ordering matters. Review is welcome now; this only blocks the merge. Remove with `/unhold` once #913 is in.

deploy/docker/patches/apply_google_chat_attachment_notice.py
```diff
@@ -159,7 +171,9 @@
                 "which holds no user credentials."
             )
             lines.append(
-                "Ask me to paste the contents here if you need them in chat."
+                "This one could not be pasted into the thread either: it "
```
jayantid, line 174: 🟡 **Not merge-blocking.** A user whose paste died on a transient API error reads three reasons that are all false for their file, and concludes a retry is pointless.

The relayed notice explains a missing paste with exactly three reasons: no text form, too large, or unreadable. But the relay patch also routes every refused or raising `send` to this same notice — `agents/platform/scripts/google_chat_relay_patch.py:451` and `:471` — and on that path the file is a perfectly pasteable text file. This module's own docstring (line 42) says the text is "written for that set", a set it defines as including "any paste the API refuses partway through". The text does not cover that member.

Concretely: a 429 at part 2 of a valid 10 KB `.md` leaves the thread with this notice. Every stated reason is false, and asking again — which would likely succeed — reads as futile.

Add the fourth reason:

```python
lines.append(
    "This one could not be pasted into the thread either: it "
    "has no text form, is too large to read here, could not "
    "be read at all, or posting it failed partway through."
)
```

agents/platform/scripts/google_chat_relay_patch.py
```diff
@@ -8,13 +9,188 @@
...
+    """
+    suffix = os.path.splitext(path)[1].lower()
+    if suffix not in INLINE_SUFFIXES:
+        return None
+    try:
+        with open(path, "rb") as handle:
+            # One byte past the cap, so a file that grew between a stat and
+            # this read is refused rather than silently truncated into chat.
+            raw = handle.read(INLINE_MAX_BYTES + 1)
+    except OSError:
```
jayantid, line 162: 🟡 **Not merge-blocking — the opposite: the `/hold` looks unnecessary, and it parks this fix behind a 186-file PR.**

The hold says #913 must merge first. But merge order cannot change the merged end state, and neither PR contains the follow-up pieces (the notifier reading `task_attachments`, the worker guidance). The only difference between the two orders is the interim, and it favors this PR: merged first, non-sandbox installs — including the install that reported #999 — get the fix sooner.

Under #913's `terminal.backend: ssh` this code already degrades to today's behavior on its own: the file is not visible to the gateway, `open()` raises `OSError`, `_inline_text` returns `None` (this line), and the notice posts exactly as it does now. The T10.3 quiet-drop symptom arrives with #913's rollout whichever PR merges first. The description makes this argument itself: "the install that reported #999 is not a sandbox install."

Unless the hold protects something the description doesn't name, `/unhold` and let this land on its own timeline.

jayantid:
**Smoke red triage (gate-watch):** not caused by this PR (Google Chat relay/docs diff). [Build 2095204541652799488](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1134/pull-kube-agents-smoke-test/2095204541652799488) (evals-2) is the fleet-wide Vertex 429 storm (#1097/#1184): rung-3 reps are empty-trajectory 429s, and the `rca-remediation-pr`/`compliance` rung-4 collapses are delegated work dying downstream on the same 429s. This run started pre-18:51Z, so `compliance-rbac-overgrant` still gated it — #1173 (merged 18:51Z) removes that for the next run. `/retest` once the quota window calms; PR 1177 already went green on retest at ~20:00Z, so the storm is intermittent.

haoxuw (approved):
LGTM, should wait for unhold

jayantid (approved):

deploy/docker/patches/apply_google_chat_attachment_notice.py
dshnayder: Confirmed and fixed in `6989f565`. You're right that the module counted the case and the text didn't: the docstring at line 42 lists "any paste the API refuses partway through" in the residual set, and both `google_chat_relay_patch.py:451` (refusal) and `:471` (raise) return through `notice()`, so a valid `.md` that died on a 429 at part 2 reached three reasons that were all false for it.

Took your wording verbatim — the line now ends `…could not be read at all, or posting it failed partway through.` Two things alongside it: a test asserting the fourth reason is present on the relayed branch, named for why it matters rather than for the string, and the sibling test's docstring extended, since it enumerated the same three-member set in prose.

`deploy/docker/patches` is 936 tests, OK; the relay suite is 33, OK; `make docs-check` clean.

agents/platform/scripts/google_chat_relay_patch.py
dshnayder: Your technical argument holds, and I'll say where: under `terminal.backend: ssh` this code does degrade to today's behaviour on its own — `open()` raises `OSError`, `_inline_text` returns `None`, the notice posts. I checked that path when writing the ordering note and it is why the note says "should not merge before", not "will break if it does". The T10.3 quiet drop is #913's rollout, not this merge, and you're right that the merged end state is identical either way.

I can't lift it, though. The hold is the repository owner's explicit instruction, given after reading the ordering argument in the description: they asked for a gate rather than a written request precisely because an approved PR merges on its own. So `/unhold` is theirs, not mine — I've put your argument in front of them and will act on whatever they decide.

Leaving this open rather than resolving it, since it's a live disagreement about what should happen next rather than something I've addressed.

dshnayder: **Correction — my reply above was wrong about the mechanism, and I have now measured it.**

I wrote that under `terminal.backend: ssh` "the notice posts". It does not. Nothing posts.

`_inline_text` is never reached, and neither is `_send_file` nor `_post_attachment_fallback`. The drop happens a layer up, in the notifier. `GatewayKanbanWatchersMixin._deliver_kanban_artifacts` screens every candidate path with `os.path.isfile` *before* the adapter is involved at all. Under the sandbox the agent's file tools run in `platform-agent-shell-0` against that pod's own RWO volume, so in the gateway process that call is `False` for every file the agent wrote: `candidates` comes back empty and the method returns on the branch whose comment reads "the path may have been mentioned for reference only". No upload, no upload error, and no notice — for every artifact type, and for every adapter, Slack and direct-token installs included.

Measured on `kage-management` today, on an image built from #913's head plus this branch's commits: a stub adapter installed into the running gateway received **zero** calls, and `os.path.isfile` on the sandbox-written path returned `False` in the gateway process. This does not need this PR's feature to be present in order to happen — it is a #913 regression that is both wider than this change and older than it.

What that does and does not do to your argument:

- **Still right.** The merged end state does not depend on the order, and non-sandbox installs — including the one that reported #999 — get the fix sooner if this lands first. Nothing above touches that.
- **Wrong in my reply.** The interim is not "this feature degrades to today's behaviour on its own". Under #913 the feature is not reached, and neither is the fallback it was written to defer to.
- **Understated in the description.** The ordering note calls the sandbox symptom "worse and quieter" and attributes it to T10.3 as a #913 rollout matter. It is worse and quieter than I described it, and it is total: a completed card that produced a report delivers no report and no explanation.

The fix for it exists and is verified, and it is not this PR. Stage the declared artifact out of the sandbox over `sandbox_exec.read_bytes()` as the `agent` login, ahead of the `isfile` screen, and hand the original method a local path. It cannot ride here — `sandbox_exec.py` does not exist on `main`, it arrives with #913 — so it goes up as its own PR against its own issue once #913 merges. A live run on `kage-management` with #913 plus that patch put the deliverable's contents into the Chat thread under a `node-pool-review.md (539 B)` header, byte-exact, alongside a negative control on the same image showing the silent drop.

Leaving this open. It asks for `/unhold`, which is still the repository owner's to give, and the measurement points both ways for them: the interim under this-PR-first is worse than I told them, and the fix for that interim is queued behind #913 either way.

dshnayder:
/retest

**Smoke red triage:** not caused by this PR, and one retest on a suspected transient.

The same code passed on this branch 8.5 hours earlier, on the same base:

| Head | `pull-kube-agents-smoke-test` | BaseSHA | When |
| --- | --- | --- | --- |
| `5716e3f8` | success | `32c5538b` | 2026-09-04 10:34Z |
| `6989f565` | failure | `32c5538b` | 2026-09-04 19:11Z |

`main` did not move underneath it — same `BaseSHA`, pinned by `smoke-test-sticky`. The entire delta between the green and the red is `5716e3f8..6989f565`: four words inside the `PATCHED` replacement string of `deploy/docker/patches/apply_google_chat_attachment_notice.py`, plus its unit test. That is the fix for [this thread](https://github.com/gke-labs/kube-agents/pull/1134#discussion_r3917548060). The patchlib anchor the Dockerfile asserts is not touched, `build` passed on the failing head (which is what applies that patch into the agent image), and nothing in the eval matrix exercises a Google Chat attachment notice. The push is also what re-ran the job, which is #1179.

@jayantid triaged this PR's previous smoke red the same way on 2026-09-02.

I could not name the failing case from the record: the job pod is GC'd (`Log not found`) and `gs://kube-agents-prow` refuses anonymous reads, so the classification above is from the diff rather than from the artifacts. If gate-watch can see which case and rung it hit and it turns out to point here, say so and I will take it.

agents/platform/scripts/google_chat_relay_patch.py
dshnayder: #913 merged this morning as `f66d1957`, so the premise of the hold is gone. `main` is merged into this branch and the staging fix that #913 made necessary is now in this PR rather than a follow-up — `sandbox_exec.py` only existed on `main` once #913 landed, which is what had blocked folding it in.

That resolves the half of your finding that was always right: the merged end state does not depend on the order. It also resolves the half I got wrong. My earlier reply claimed this code "degrades to today's behaviour without help (`open()` raises, `_inline_text` returns `None`, the notice posts)". It does not — nothing in the PR was reached at all, because `_deliver_kanban_artifacts`' `os.path.isfile` screen drops the path before the adapter is involved. Measured on `kage-management`: a stub adapter got zero calls for a card declaring a sandbox-written artifact.

What is left on this thread is the hold itself, which is @dmitryshnayder's to lift rather than mine — leaving it open for them.

dshnayder: Closing this out. @dmitryshnayder has lifted the hold, so the one thing this thread was still holding open is gone.

For the record of what the finding produced: the notice wording is fixed and tested, my first reply's reasoning about #913 was wrong and is corrected above with the measurement, and the ordering problem the hold guarded against no longer exists — #913 merged as `f66d1957`, `main` is merged in, and the staging fix that makes a sandbox-written artifact reach the thread is in this PR rather than a follow-up. The branch head has since been live-validated against `kage-management` on this head; **Live validation** in the body has run 3.

Resolving. Reopen it if you disagree with any of that.

PR #1136, opened 2026-09-01
fix(docker): state the mode on the hermes_cli patch files

bradhoekstra:
## Summary

Three of the kanban patch modules land in `/opt/hermes/hermes_cli` without an explicit mode. `cp` gives a newly created destination the source's mode masked by the umask in force, and an unmoded `COPY` carries the build context's mode straight into the layer — so the mode those files end up with is the mode they had in whoever's working tree built the image. Git tracks only the executable bit, so a tree cloned under `umask 077` carries 0600 throughout, and the agent, which runs as uid 10000, cannot read its own modules. This sets the mode at the three write sites and adds three guards to the Dockerfile contract suite.

## Why This Change

Built under `umask 022` — which is what CI does — the files land 0644 by luck and everything is fine. Built under `umask 077`, the identical Dockerfile produces an image that dies on its first import:

```
PermissionError: [Errno 13] Permission denied:
  '/opt/hermes/hermes_cli/kanban_scheduling.py'
```

Precompiled bytecode doesn't paper over it: CPython derives a `.pyc`'s mode from its source's, so an unreadable source has an unreadable cache. Without this, the repo builds correctly only for people whose umask is permissive, and no gate anywhere reports the difference.

`agentplugins/lib/plugin_image.sh:798` already stages the plugin image against exactly this failure. The agent image reaches it by a different route.

## What Changed

- **`deploy/docker/Dockerfile`** — `--chmod=0644` on the two COPYs into `/opt/hermes/hermes_cli`, and `install -m 0644` in place of `cp` for the third. `hermes_cli` is the only patch destination no normalization loop walks; every other one is swept by the `platform` stage's loop. No instruction added or removed, so the chain stays at 114 of the 120 layer budget.
- **`deploy/docker/test_image_startup_contract.py`** — three guards, plus one shared `joined_instructions()` helper that replaces two hand-rolled copies of the continuation-joining loop already in that file.

Stated at the write site rather than by adding `hermes_cli` to the loop, because the mode is wrong in `agent-base` and the loop runs three stages later. **Layer cost is not the trade-off it appears to be and is not the reason:** measured against this base, a `find … -exec chmod` over a tree whose modes are already correct exports a ~4kB layer — identical to `RUN true` — because BuildKit diffs the result, not the syscall.

## Context

`hermes_cli` is the one patch destination outside the `platform` stage's `find … -exec chmod` normalization loop over `/opt/hermes/{plugins,gateway,tools}`, which is why it is the only tree where an unmoded write survives to runtime.

## Testing

`make docs-check` clean. All 50 tests in `deploy/docker/test_image_startup_contract.py` pass; the two new write-site guards are red against this PR's merge base and green at HEAD, and the loop-membership guard goes red when a tree is dropped from the loop. Also verified the guards reject `chmod 600` as a "fix", accept `chmod -R a+r` and `install --mode=`, and catch a lowercase `copy`.

### Live validation

Against a real install — GKE Standard, `us-central1`, namespace `kubeagents-system`, operator-managed `PlatformAgent` CR, agent running as uid 10000. Both images built **from a working tree with the three patch sources at 0600**, since building from a normal 0644 checkout produces a correct image with or without the fix and proves nothing.

Rolled by patching `.spec.deployment.tag`; the operator reconciled it through to the Deployment.

```
CR:          .spec.deployment.tag = hermescli-modes-1ab5b71a   .status.phase = Ready
Deployment:  platform-agent = …/platform-agent:hermescli-modes-1ab5b71a
Pod:         3/3 Running

$ kubectl exec … -c platform-agent -- sh -c 'id -u; ls -l …; python3 -c "import …"'
uid=10000 gid=10000
-rw-r--r-- root /opt/hermes/hermes_cli/kanban_guardrail_exit.py
-rw-r--r-- root /opt/hermes/hermes_cli/kanban_scheduling.py
-rw-r--r-- root /opt/hermes/hermes_cli/kanban_wake_nudge.py
-rw-r--r-- /opt/hermes/hermes_cli/__pycache__/kanban_scheduling.cpython-313.pyc
-rw-r--r-- /opt/hermes/hermes_cli/__pycache__/kanban_wake_nudge.cpython-313.pyc
import OK
```

**Negative control, on the same install.** The merge-base Dockerfile, same 0600 context, same base digest, run as a throwaway pod so the shared agent was never deliberately broken:

```
uid=10000
-rw------- /opt/hermes/hermes_cli/kanban_guardrail_exit.py
-rw------- /opt/hermes/hermes_cli/kanban_scheduling.py
-rw------- /opt/hermes/hermes_cli/kanban_wake_nudge.py
PermissionError: [Errno 13] Permission denied: '/opt/hermes/hermes_cli/kanban_scheduling.py'
```

Same at the image layer, two `--target agent-base` builds differing only in the Dockerfile: 0600 + `PermissionError` before, 0644 + `import OK` after.

**Not covered:** the agent's own logs show an unrelated pre-existing `HTTP Error 503` from `google_chat_relay_patch.py` polling a chat relay that isn't up on this install — present before and after, nothing to do with file modes. I did not exercise a full kanban work cycle, only module import and the patched-behaviour grep (`_kanban_record_wake(conn)` → 3).

**Cleanup:** install restored to its prior tag and `Ready` (confirmed), throwaway pod deleted, the deliberately broken negative-control image deleted from the registry.

## Self-Review

Adversarial and docs-drift, both run twice, each time in a subagent handed only the diff range. The second round changed the design, so only it is reported here.

- **The layer-cost argument in my first version was false.** I had claimed extending the normalization loop to `hermes_cli` would copy 21MB into that layer. Both passes challenged it; measuring says a no-op `chmod` over that tree exports 4.1kB, and over all of `/opt/hermes` 45.1kB. The size figures were wrong too (~12MB in the pinned base, not 21MB — the 21MB is post-`compileall`; `web_dist` is 25%, not "nearly all"). **Fixed by deleting the argument** and stating the measurement instead. The approach survives on a different rationale; the old one didn't survive checking.
- **My second justification was also wrong.** I'd argued `credential-proxy` needs the write-site fix because it never runs the loop. It runs only `credential_proxy.py`, which imports nothing from those trees. **Fixed**; the comment now gives only the reason that holds.
- **A hunk I'd added introduced a hazard this Dockerfile already documents.** I had put `--chmod=0644` on the staging `COPY` into `/tmp/latency-patches`; `--chmod` propagates to parent directories BuildKit creates, making that directory 0644 — the exact trap called out at `Dockerfile:1591`. Confirmed by building it. **Reverted.** The two direct COPYs are unaffected: their parent already exists and stays 755.
- **The loop-membership guard was vacuous.** It substring-matched the joined text of every `find … chmod` RUN, and `/opt/hermes/plugins` appears in two others — deleting it from the loop left the test green. **Fixed**: it now parses the `for d in …` list and is scoped to the `platform` stage. Verified the mutation fails.
- **The suite was a third copy of a Dockerfile parser, in the wrong directory.** Moved from `tests/` into `deploy/docker/test_image_startup_contract.py` — the Dockerfile-contract suite, which already carries `stage_body`/`platform_stage` and already asserts on build-time modes — and the two existing hand-rolled join loops now call the shared helper. The two passes disagreed here (one read `AGENTS.md` as pointing to `deploy/docker/`, the other read `docs/testing-map.md` as pointing to `tests/`); I went with the existing suite.
- **Mode-parsing was both too lax and too strict.** It credited a `chmod` to *any* mode, so `cp … && chmod 600 …` passed; and it rejected `ugo+r`, `chmod -R`, and `install --mode=`, which would go red on a correct Dockerfile. Both fixed and checked across 15 mode forms.
- **Not fixed, deliberately.** Seven other unmoded COPYs into `/opt/hermes/{tools,gateway}` in `agent-base` reach the `credential-proxy` image with no loop behind them. That is the same latent defect, but it isn't live — the sidecar imports none of those modules — and widening this PR to cover it would change an image whose behaviour is not otherwise in question. The test docstring names the gap rather than hiding it.
- **Reported, not fixed (pre-existing, unrelated):** `docs/testing-map.md` has no row for the `deploy/docker/test_*.py` home, and line 11 cites `Makefile:129-142` where the glob block now runs to 144.
- **Known limits of the guards:** they read the Dockerfile, not a built image, and see only `COPY` and `cp`/`mv`/`install`. A module arriving via `ADD`, `tar -x`, or an applier using `shutil.copy` (which also preserves the source mode) would not be caught. Stated in the docstring.

## Risk & Rollout

Low. Three tokens in one build stage; no runtime code path, no interface, no manifest, no IAM. The modes it sets are what the surrounding tree already has, and all 620 files in `/opt/hermes/hermes_cli` in the base image are already 0644, so nothing is loosened relative to a correct build — it only stops a hardened builder from producing a broken image. No layer added; 114 of 120. Revert is the commit.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -820,7 +820,7 @@ RUN /opt/hermes/.venv/bin/python3 /tmp/kanban-result/apply_kanban_report_format.
 # that rather than leaving it to a reader.
 # See the module docstring in deploy/docker/patches/kanban_scheduling.py and the
 # applier docstring in deploy/docker/patches/apply_kanban_scheduling.py.
-COPY deploy/docker/patches/kanban_scheduling.py /opt/hermes/hermes_cli/kanban_scheduling.py
+COPY --chmod=0644 deploy/docker/patches/kanban_scheduling.py /opt/hermes/hermes_cli/kanban_scheduling.py
 COPY deploy/docker/patches/apply_kanban_scheduling.py \
      deploy/docker/patches/verify_kanban_scheduling.py \
      deploy/docker/patches/patchlib.py \
@@ -886,7 +886,7 @@ RUN /opt/hermes/.venv/bin/python3 /tmp/kanban-scheduling/apply_kanban_scheduling
 # further down. It must not ship alone: raising the cap only moves when the bug
 # fires.
 # See the module docstring in deploy/docker/patches/kanban_guardrail_exit.py.
-COPY deploy/docker/patches/kanban_guardrail_exit.py /opt/hermes/hermes_cli/kanban_guardrail_exit.py
+COPY --chmod=0644 deploy/docker/patches/kanban_guardrail_exit.py /opt/hermes/hermes_cli/kanban_guardrail_exit.py
 COPY deploy/docker/patches/apply_kanban_guardrail_exit.py \
      deploy/docker/patches/verify_kanban_guardrail_exit.py \
      deploy/docker/patches/patchlib.py \
@@ -1013,7 +1013,18 @@ COPY deploy/docker/patches/kanban_wake_nudge.py \
      deploy/docker/patches/verify_kanban_event_routing.py \
      deploy/docker/patches/patchlib.py \
      /tmp/latency-patches/
-RUN cp /tmp/latency-patches/kanban_wake_nudge.py /opt/hermes/hermes_cli/kanban_wake_nudge.py && \
+# `install -m 0644`, not `cp`, for the first line. `cp` gives a newly created
+# destination the source's mode masked by the umask in force, and the source
+# here is whatever the build context held: git tracks one permission bit, so a
+# tree checked out under `umask 077` carries 0600 throughout and `cp`
+# propagates it rather than repairing it. The agent runs as uid 10000 and
+# cannot read the result. Same failure agentplugins/lib/plugin_image.sh:798
+# stages against for the plugin image, same cause.
+#
+# The other four land in /opt/hermes/tools and /opt/hermes/gateway, which the
+# `platform` stage normalizes; /opt/hermes/hermes_cli is the one patch
+# destination no loop walks, so it is stated here at the write site.
+RUN install -m 0644 /tmp/latency-patches/kanban_wake_nudge.py /opt/hermes/hermes_cli/kanban_wake_nudge.py && \
     cp /tmp/latency-patches/kanban_auto_subscribe.py /opt/hermes/tools/kanban_auto_subscribe.py && \
     cp /tmp/latency-patches/kanban_comment_status.py /opt/hermes/tools/kanban_comment_status.py && \
     cp /tmp/latency-patches/kanban_notify_delivery.py /opt/hermes/gateway/kanban_notify_delivery.py && \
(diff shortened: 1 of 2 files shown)


jayantid (approved):

bradhoekstra:
/hold cancel

#1173 merged at 18:51Z, so the pool is clear. Un-holding.

Issue #1054, opened 2026-08-28
The catch ledger: real product bugs surfaced by the eval gate


jayantid:
**Real product defects surfaced by the eval gate.** The criterion: a defect in the *product* — agent behavior, SOP, runtime, publish path — that an eval failure surfaced. Check-design and harness fixes deliberately do not count; those are the gate improving itself, and mixing them in would inflate the number.

The machine-readable convention: qualifying issues carry the literal line `Caught-by: evals` in their body (`gh search issues --repo gke-labs/kube-agents "Caught-by: evals" in:body`). This issue is the human-readable roster and holds any catches whose authors have not added the marker.

| # | Caught | Defect | Introduced by | Status |
| --- | --- | --- | --- | --- |
| #1010 | 2026-08-27, `capacity-pinned-pool-probe` | Platform worker marks its delegation card complete on fan-out — "Awaiting synthesis" delivered as the final answer | pre-existing | open, agent team |
| #1030 | 2026-08-27/28, `compliance-rbac-overgrant` (canary) | Ledger publish path broken across the #955 sandbox/proxy uid split — audits computed correct findings that never reached GitHub | #955 (2026-08-26) | **fixed** — #1053 merged 08-29; canary 2/3 reps on the fix PR itself, 14/17 green wave post-merge |
| #988 | 2026-08-26/27, multiple tasks | Platform-agent front door dies mid-run (gateway stall + pod bounce, two signatures) — HTTP path dead while workers keep working | pre-existing, exposed by sustained eval load | open, proxy owner |
| #1097 | 2026-09-01, seven tasks at once (PR #1089 run) | Agent endpoint slot saturation — ~6 of 10 concurrent-run slots held by unseen/stuck server-side runs, rejecting real work with HTTP 429 for a 4-minute window | pre-existing, exposed by 3-repetition eval load | open, needs endpoint owner |

[...]

PR #1145, opened 2026-09-02
feat(dashboard): redesign the eval dashboard to the two-band agent/gate layout

jayantid:
## Summary

Redesigns the eval dashboard from a single suite table into a two-band story page. Band 1, "The agent", answers whether the agent is getting better on the cohort where the codebase is what actually shipped — the final `pr_merged` run of each merged PR: a weekly pass-rate tile with a prior-week delta, the human-annotated catch counts, domain coverage, and a by-day pass-fraction trend with named event markers. Band 2, "The gate", answers whether the gate is trustworthy across all PR runs: false-red / infra-rep-rate / wall-clock / queue-wait tiles, a case × run outcome matrix over the last 30 measured runs with rep-level cell states, and a Pareto of failure signatures grouped by a small documented reason-normalization map. Every fragment builder in `render.py` has its JS mirror in the template, as before; the JS-rendered page is pixel-identical to the baked one.

## Why This Change

The current page answers "what happened in the latest run" but not the two questions people actually bring to it: is the agent improving where it matters (merged code, not every red experiment), and can the gate's verdicts be trusted (which failures are flake, which are infra, what a run costs). Without this change the trend chart keeps charting every PR run on a per-PR-number axis — so one broken PR's run drags the line down and reads as an agent regression — and there is nowhere to see per-case flake patterns across runs at all.

## What Changed

- `scripts/eval_dashboard/render.py` — band builders replace the hero/suite/trends builders. New shared vocabulary: `task_reps()` (reads the additive `tasks[].reps` field and falls back to the task's single `result`), `cell_state()` (pass / partial / fail-all / infra-excluded / not-run), `is_run_event()` (a run where ≥80% of graded tasks failed is charged to the run, and its failures are excluded from per-case stats — the rule is stated verbatim in the matrix caption and the footer), `merged_final_runs()` (band-1 cohort), and the failure-signature normalization + classification maps (small and documented at the top of the file; anything unmatched groups by its first 60 chars and renders a neutral bar).
- `scripts/eval_dashboard/template/index.html.tmpl` — the JS mirror of all of the above, plus the Mock A band/matrix/Pareto CSS mapped onto the existing theme variables. The freshness badge, 60-second poll, `__APP__` scoping and `<` bootstrap escaping are unchanged.
- `scripts/eval_dashboard/events.yaml` — new optional annotation file beside `case-notes.yaml`: dated `{date, label}` markers (rendered on the trend and the matrix header) plus the two human-judgment numbers the tiles need (`catches`, `false_reds_7d`). Documented by its own header comment, the same way `case-notes.yaml` documents itself. Absent file → no markers and em-dash tiles.
- `scripts/eval_dashboard/case-notes.yaml` — entries gain an optional `badge:` key (`held-out` / `new`) that the matrix rows render; seeded for the three held-out cases and the new remediation case.
- Removed: the "Latest run" hero tile, the per-PR-number trend chart, the single-case judge chart, the per-case suite table (superseded by the matrix), and the "Median case cost" tile. The old "Evidence on record" table is kept in slimmed form at the bottom — it is the only home for admission-progress depth (`runs_on_record` vs the 20-run yardstick), the IN PRESUBMIT state, and the case-notes annotations with issue links; its pass-rate column is dropped because the matrix now shows per-case outcomes directly.
- The queue-wait tile always renders "—": `data.json` carries no queued-at timestamp, so the number is not computable, and the tile says so instead of inventing one.

## Context

- **Depends on (ordering only, not correctness):** the rep-schema collector PR (branch `feat/dashboard-rep-schema`, not yet open at time of writing) adds `tasks[].reps` and `runs[].pr_merged` to `data.json`. This PR is built against that documented contract with its own fixtures and **degrades safely if deployed first**: with today's production data (neither field present) the matrix falls back to single task results, the merged-PR cohort is empty ("no merged-PR runs on record", "not enough merged-PR days yet"), and the infra tile labels itself "task-level fallback". Merge order therefore does not matter for safety; the agent band only lights up once the collector PR lands.
- Duplicate-work scan (AGENTS.md "Before Starting a Task"): #1141 (same author) touches `collect.py`/`SCHEMA.md` for the hourly refresh periodic, and draft #1082 touches `test_eval_dashboard_collect.py` — neither touches `render.py`, the template, or `test_eval_dashboard.py`, so this is merge-conflict-free overlap, not duplication. No open PR or issue covers the renderer redesign itself; it is lane work under #1022 / #1020.
- `collect.py` is deliberately untouched (owned by #1141 and the rep-schema branch).

## Testing

- `cd scripts && python3 -m unittest discover -p "test_*.py"` — 799 tests, green (56 in `test_eval_dashboard.py`, rewritten for the new layout).
- New coverage: reps-present cell states, reps-absent fallback (today's production shape), run-level-event exclusion from the Pareto, cohort selection (`pr_merged` true/false/null and final-run-per-PR), week-boundary pass-rate math (exact 7-day boundary), empty and absent `events.yaml`, badge rendering from notes, hostile `reps[].reason` escaping, JS-vs-Python rounding parity (12.5% → 13%), and the existing bootstrap `<` escaping tripwires.
- `prettier` could not run locally (no node/npm on this machine; AGENTS.md notes `npx prettier` is unreliable here anyway) — the changed YAML follows the existing files' formatting and CI's prettier check will verify.

### Live validation

Not live-tested against a running installation — this change never touches a cluster: it is the static renderer for the published dashboard page, exercised end to end locally instead:

- Rendered a rich 22-run / 17-case fixture (reps, `pr_merged`, a run-level-event run, a 429 infra-wave run) through the real CLI: `python3 scripts/eval_dashboard/render.py --data demo.json --out-dir out/`.
- Screenshot of that render (headless Chromium, dark theme):

eval-dashboard-mock-a

_Published 20260902T020156Z at commit 39e71f1c2053 — pre-captured image (dash-server.png)._

- JS-mirror proof: a second copy of the rendered page with the baked `__APP__` fragment replaced by a sentinel (so everything visible is painted by the template's JS from the bootstrap data) screenshots **byte-identical** to the server-rendered page — the mirror executes and produces the same DOM. A JS syntax error would have left the sentinel on screen.
- Degradation check: the current production `data.json` shape (no `reps`, no `pr_merged`) renders the fallback states described under Context (covered by `test_todays_production_shape_degrades_gracefully`, plus a manual render).
- Could not cover: the live 60-second poll loop against a real bucket (file:// preview deliberately skips fetching), and light-theme rendering was only spot-checked — the published page pins `data-theme="dark"`.

## Self-Review

Both required passes ran in contexts that did not write the change — two fresh subagents, each handed the checkout, the diff range (`upstream/main...HEAD`) and the relevant skill, nothing else.

**Adversarial pass** (subagent, fresh context): traced every interpolation site in both builders (element text, `title` attributes, Pareto labels, trend `data-l`, badges, notes, footer), the bootstrap `<`-escaping, week-boundary and threshold math, every division, the run-event exclusion scoping, and Python-vs-JS parity function by function. Findings and dispositions:

- Evidence table crashed on malformed `cases[]` entries (non-dict entry; string/bool `runs_on_record`) where every sibling section degrades — **fixed**: `isinstance` filter plus `is_count` guards in the row and the sort key, mirrored in JS.
- Re-typed `coverage` (non-dict) crashed the Python renderer, and a string `uncovered` iterated character-wise in Python while throwing in JS — **fixed** on both sides (`isinstance` / `Array.isArray` guards); tolerance tests added.
- `load_notes` broke its own "malformed degrades to no note" promise on a top-level list, a list under `notes:`, or scalar `issues` — **fixed** with the guards `load_events` already had; test added.
- Parity: JS `Date.parse` reads a timezone-naive ISO stamp as *local* time where Python assumes UTC, shifting every day bucket and week window for a non-UTC viewer — **fixed**: the JS mirror appends `Z` to naive stamps; a script-source tripwire test pins it.
- Five smaller mirror drifts reachable only from off-contract data (`source: null` rendering, scenario counts over non-dict entries, boolean `runs_on_record`, UTF-16 vs code-point snippet slicing, `localeCompare` vs code-point sort) — **all fixed** on whichever side diverged.
- Matrix fallback said "no measured runs yet" even when the missing input was the case list — **fixed**: split messages.
- Question: run-level-event failures still count in band 1's pass rate and trend. **Deliberate, not fixed**: the rule is scoped to per-case aggregates; band 1's cohort is the final run of each *merged* PR, and a merged PR's own failures are that cohort's signal. Now stated in the module docstring so the next reviewer does not have to ask.
- Question: `false_reds_7d` is a rolling-window claim with no as-of date and nothing to detect staleness. **Deliberately not fixed**: the file header says a stale line is deletable without breaking a render, the tile is labeled "human-classified · events.yaml", and freshness machinery for an advisory hand count is more mechanism than the number warrants.
- Its named residual gap — "the JS mirror was never executed in this pass" — is closed by the Live validation sentinel check above (the JS-rendered page screenshots byte-identical to the baked one in headless Chromium).

**Docs-drift pass** (second subagent, fresh context): checked SCHEMA.md, the docs map, README/docstring prose, and ran `make docs-check` (all five checks green). Findings and dispositions:

- SCHEMA.md did not define `tasks[].reps` / `runs[].pr_merged`, the fields this renderer now reads — **fixed**: documented as optional additive fields in the existing `stale_after_s` style ("no collector version emits them yet").
- `scripts/eval_dashboard/__init__.py` still said "the one optional extra input is case-notes.yaml" — **fixed**: names both annotation files and the badge key.
- Present-tense "a newer collector emits" in two docstrings read as describing a collector that is not in the tree — **fixed**: rephrased to match SCHEMA.md's wording.
- Hand-maintained counts in events.yaml are a drift magnet — **accepted by design**; that is the file's stated purpose, receipts live in the #1054 ledger.

All findings above were re-verified against the final head: full suite green (799 tests), fresh render, and the byte-identical JS/baked screenshot comparison repeated after the fixes.

## Risk & Rollout

Renderer only — no runtime agent code paths, no CI verdict logic, no collector changes. The 15-minute refresh periodic publishes whatever is on `main`, so this layout goes live on merge; a bad render is a display problem, not a gating one (the publish hook is fail-safe by contract — `test_eval_dashboard_publish.py`). New failure mode: a malformed `events.yaml` on `main` — mitigated by tolerant parsing (malformed entries drop, absent file renders em-dashes) and tests. Rollback: revert this commit; the page returns to the current layout on the next publish.

Generated with the help of Claude Code (model: Fable 5).

### DIFF
--- a/scripts/eval_dashboard/SCHEMA.md
+++ b/scripts/eval_dashboard/SCHEMA.md
@@ -124,9 +124,9 @@ collector version in this tree emits them yet; the renderer already reads
 both.
 
 - `runs[].pr_merged` — `true` | `false` | `null`: whether the run's PR has
-  merged. The renderer's "agent" band charts only runs with `pr_merged:
-  true` (the final such run per PR); absent or `null` keeps a run out of
-  that cohort without any other effect.
+  merged. The renderer's "agent" band charts only runs where it is `true`
+  (the final such run per PR); absent or `null` keeps a run out of that
+  cohort without any other effect.
 - `runs[].tasks[].reps` — the task's individual repetitions, in order:
   `[{"n": 1, "result": "pass"|"fail"|"infra", "reason": "<string>"|null}]`.
   `reason` is free-form log text (renderers must escape it). `infra` reps
--- a/scripts/eval_dashboard/SCHEMA.md
+++ b/scripts/eval_dashboard/SCHEMA.md
@@ -117,6 +117,23 @@ Additive, optional, and safe to omit — consumers must default them.
   page labels itself `STALE`. No collector version emits it yet; the
   renderer defaults to `7200` when it is absent.
 
+### Optional run and task fields
+
+Additive, optional, and safe to omit — consumers must default them. No
+collector version in this tree emits them yet; the renderer already reads
+both.
+
+- `runs[].pr_merged` — `true` | `false` | `null`: whether the run's PR has
+  merged. The renderer's "agent" band charts only runs with `pr_merged:
+  true` (the final such run per PR); absent or `null` keeps a run out of
+  that cohort without any other effect.
+- `runs[].tasks[].reps` — the task's individual repetitions, in order:
+  `[{"n": 1, "result": "pass"|"fail"|"infra", "reason": "<string>"|null}]`.
+  `reason` is free-form log text (renderers must escape it). `infra` reps
+  are excluded from every pass-fraction denominator, exactly like `infra`
+  task results. When `reps` is absent the task's single `result` stands in
+  for one rep.
+
 ### `coverage` — from `docs/designs/domains.yaml`
 
 - `domains_total` — number of entries under `domains:`.
--- a/scripts/eval_dashboard/__init__.py
+++ b/scripts/eval_dashboard/__init__.py
@@ -4,7 +4,9 @@ The collector (`collect.py`) writes one data.json; its schema is a contract
 shared with the renderer and the publisher -- see SCHEMA.md in this directory
 before changing any field. `render.py` turns one data.json (schema_version 1)
 into `out/index.html` plus a copy of the data file; `publish.py` ships an
-out-dir to its serving location. Everything on the page is computed from
-data.json alone -- the one optional extra input is `case-notes.yaml`, which
-carries human one-line annotations and issue links per case.
+out-dir to its serving location. Everything measurable on the page is
+computed from data.json alone -- the two optional extra inputs are
+`case-notes.yaml` (human one-line annotations, issue links and badges per
+case) and `events.yaml` (dated event markers plus the human-classified
+catch and false-red counts); render.py's docstring owns the details.
 """
--- a/scripts/eval_dashboard/case-notes.yaml
+++ b/scripts/eval_dashboard/case-notes.yaml
@@ -12,13 +12,21 @@
 #     <case-name>:
 #       note: one-line human context, shown under the case name
 #       issues: ["#123", "#456"]   # rendered as links to this repo's issues
+#       badge: held-out            # matrix row pill: "held-out" or "new"
 
 notes:
   agent-kanban-smoke:
     note: redesigned 08-27
+  autoops-warning-event-triage:
+    badge: held-out
   capacity-pinned-pool-probe:
     issues: ["#1010"]
+    badge: held-out
+  cluster-agent-healthy-workload-no-finding:
+    badge: held-out
   compliance-rbac-overgrant:
     issues: ["#998", "#985"]
   gpu-stress-test-diagnosis:
     note: reuse regression under investigation
+  security-overgrant-remediation-proposal:
+    badge: new
--- a/scripts/eval_dashboard/events.yaml
+++ b/scripts/eval_dashboard/events.yaml
@@ -1,0 +1,34 @@
+# Optional human annotations for the eval dashboard that no log line can
+# carry: dated event markers, and the few human-judgment counts the tiles
+# show. Everything measurable comes from data.json; this file is only for
+# what a human classified or decided. render.py tolerates an absent or
+# empty file (markers disappear, the annotated tiles render an em-dash), so
+# deleting a stale line can never break a render.
+#
+# File shape (all keys optional):
+#
+#   events:                       # vertical markers on the day trend and the matrix
+#     - date: "2026-08-29"        # UTC day, YYYY-MM-DD
+#       label: "#1053 publish fix"
+#   catches:                      # band 1 "Product bugs caught" tile
+#     product_bugs: 4             # real product bugs the suite caught
+#     prs_blocked: 2              # PRs the gate stopped from merging
+#     ledger: "#1054"             # where the receipts live
+#   false_reds_7d: 1              # band 2 tile: human-classified false reds
+
+events:
+  - date: "2026-08-27"
+    label: "kanban redesign"
+  - date: "2026-08-29"
+    label: "#1053 publish fix"
+  - date: "2026-08-31"
+    label: "#1063 + 360m timeout"
+  - date: "2026-09-01"
+    label: "#1095 429→infra + #1096 roster"
+
+catches:
+  product_bugs: 4
+  prs_blocked: 2
+  ledger: "#1054"
+
+false_reds_7d: 1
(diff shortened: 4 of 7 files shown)


jayantid:
Smoke red classified: not this PR. The only failing case is `autoops-warning-event-triage` — all 3 repetitions "results.json carries a record with no scores map: the run or its scoring pass crashed" (rung 2, CHECK_DID_NOT_RUN). Evidence it is pre-existing harness state, not this diff:

- This PR changes only the dashboard renderer, and this job never executes it — the run's own log says `eval-dashboard publish skipped: not a main-branch run (JOB_TYPE=presubmit)` (build-log line 3097).
- PR #1137's smoke run (build 2094904140260970496, finished 2026-09-01T23:17Z, before this PR existed) failed with the identical signature on the identical case.
- The case is the known-immature one tracked in #1101 (calibrate the judged checks and build a blocking-grade record); every other case passed here (admitted-case pass rate 89.7%).

/retest

jayantid:
Second smoke red classified: again not this PR, and a different case than the first red — `autoops-warning-event-triage` passed all 3 repetitions this time, confirming the first red was flake. This run's only red is `compliance-rbac-overgrant` (rung 4, all 3 reps): every check found only a ledger generated 10h before the run started ("it is a previous run's ledger, so this run published nothing") — the audit-SOP ledger-publish latency tracked in #985, which is exactly the issue `case-notes.yaml` already annotates this case with. The dashboard hook again logged its presubmit skip, so nothing in this diff executed in the job.

/retest

jayantid:
Third smoke red classified: pure infra, no eval task ever ran. The job died at chart deploy time on leased project `kube-agents-evals-4` with `Error: UPGRADE FAILED: "kube-agents" has no deployed releases` — a dirty/stuck helm release state on the pool project (the teardown's `failed to delete release` confirms it); the log has zero `Task ... Result:` lines. Note the three reds so far are three different, unrelated signatures (autoops scoring crash → known-flaky case, passed 3/3 on rerun; #985 ledger latency on compliance-rbac-overgrant; now a dirty leased project), which is pool instability, not this renderer-only diff.

/retest

jayantid:
Fourth smoke red classified: infra again — four cases hit rung 3 NOT_A_REAL_RUN ("the trajectory is empty ... tokens.total is 0: no model call was billed"), the endpoint slot-saturation signature tracked in #1097 / #1116, plus the recurring #985 stale-ledger failure on `compliance-rbac-overgrant`. Every failing signature across the four reds has an open tracking issue, and none of them execute anything from this renderer-only diff. The job does pass intermittently right now — #1141 went green at 06:27Z and #1135 at 10:37Z today — so one more rerun is warranted rather than noise.

/retest

dshnayder (approved):

PR #1118, opened 2026-09-01
feat(platform): add an update-pr skill and a pr_updates watcher sweep

AntonTyb:
## Summary

Adds an `update-pr` skill to the Platform Agent and a `pr_updates` sweep to the repository watcher, so a pull request the agent authored stops needing a human to notice it has gone stale. The sweep runs on the existing `github-repo-watcher` tick, finds the agent's own open pull requests that are conflicted or red, and files one kanban card per pull request. The card wakes the agent into the skill, which works the pull request in a fixed order — merge conflicts first, then reviewer comments, then CI failures — and pushes one commit per stage rather than one per fix. Stage two is not new logic: it delegates to the existing `pr-conversation` skill, which already owns reviewer comments.

## Why This Change

The agent opens pull requests and then loses track of them. A branch that conflicts with `main`, or whose CI went red after a base change, sits until a person happens to look; the existing watcher only reacts to a reviewer typing a comment, so a pull request nobody has commented on is invisible to it no matter how unmergeable it has become. The result is a queue of the agent's own work that only a human can unstick, which is the opposite of what the watcher is for.

The order matters as much as the trigger. Fixing CI on a conflicted branch produces a commit that will be re-resolved by the merge; answering a reviewer before the merge means answering against a diff that is about to change. Conflicts, then comments, then CI is the only order where each stage's work survives the next.

## What Changed

- `agents/platform/skills/update-pr/` — the new skill. `SKILL.md` is the procedure; `scripts/update_pr.py` has two verbs, `poll` (what state is this pull request in, and is there work) and `record` (write the attempt marker and the outcome comment). The model does the fixing between them; the script does the state reads and the bookkeeping, so the marker cannot be written by a model that decided it was finished.
- `agents/platform/scripts/github_scan_gate.py` — the `pr_updates` sweep, registered above `pr_comments` in `SWEEPS`. Registration order is run order, and this one has to run first because it **claims** the pull requests it cards; `pr_comments` skips a claimed pull request so one tick cannot dispatch two workers onto the same branch. The claim set is now `set[tuple[str, int]]` rather than bare numbers.
- `agents/platform/scripts/forge.py` — two reads added to the `ForgeProvider` protocol, `conflict_state` and `failing_checks`, plus a frozen `CheckRun`. `failing_checks` reads **two** registers — check runs and legacy commit statuses — because Prow reports only in the latter and neither register sees the other's results. Three API calls per healthy pull request per tick: one for merge state, two for CI.
[...]

Issue #1113, opened 2026-09-01
Follow-up from #504: multi-repo support — carried-over and new findings


mplakhtiy:
Follow-up work for #504 (`feat(agent): add support for multiple github repos`), which is being
merged with these findings knowingly outstanding. #504 moves the agent's repository target from the
`Git Repo:` line of `SETTINGS.md` to a `<agent-name>-gitops-state` ConfigMap holding a JSON list of
managed repositories, seeded by the operator from `spec.integration.github.gitRepo` and extended by
an administrator editing the ConfigMap.

This issue is the disposition record for everything raised against that branch and not fixed on it.
It covers a final adversarial pass at head `8ae1369a` plus a re-check of all 38 review threads left
unresolved on the pull request. All of it is follow-up; none of it was treated as blocking.

Sections C and D exist so nobody re-files work that is already done — several of the open threads on
#504 read as live findings but are answered in the code, and eleven more point at files the branch
stopped touching after it was trimmed.

---

## Status — updated 2026-09-03 (verified against main `b0a4d7d1`)

Fifteen of the seventeen items still reproduce. One §B8 sub-item is resolved and is marked as such
below; A6 is a record of inaccuracies in a merged pull request's description and was never
actionable, leaving fifteen open items and one closed.

#1151 has been closed as a duplicate of §B8 and folded into it. Its three line references are now
carried in §B8, along with a fourth line at the same drift and one page checked and found accurate.

#1150 — the pull request that covers the Python half of B4, A7 and B2, described in the comments
below — is open and unmerged. Those halves are pending, not landed.

---

### A1 — Minty never reads `default.yaml`, so the chart ships no usable policy for any repository

`charts/kube-agents/templates/github-minter.yaml:61,161`;
`k8s-operator/config/integrations/github/configmap.yaml.template:6`,
`deployment.yaml.template:83`; `syncGithubTokenMinterConfigMap`.

The minter policy ConfigMap key changed from `<org>-<repo>.yaml` to `default.yaml`, and the mount
changed from a `subPath` file at `/etc/minty/<org>/<repo>.yaml` to the whole ConfigMap directory at
`/etc/minty/<org>`. The pinned Minty (`v2.7.1`, `images.json`) resolves a policy at exactly
`<configDir>/<org>/<repo>.yaml` and has no `default` fallback —
`pkg/config/config_loader.go:114`, `name := fmt.Sprintf("%s/%s/%s.yaml", l.configDir, org, repo)`.
[...]

PR #1150, opened 2026-09-02
refactor(platform): parse repository identity in one place

AntonTyb:
## Summary

Every Python assertion of "is this a repository, and is it one I may act on" now runs through one parser, `agents/platform/scripts/repo_ref.py`. It returns a `RepoRef` carrying a host and an opaque path of arbitrary depth, and the two-segment `owner/repo` rule becomes a per-provider check on GitHub applied to a ref whose host has already been parsed — rather than an invariant of the whole stack, expressed six times in six regexes that had quietly drifted apart. `forge.provider_for` dispatches on that parsed host and raises on one it has no provider for, where before anything unrecognised fell through to `GitHubProvider`. This is step 1 of `docs/designs/multi-forge-support.md` §9, the first implementation PR after the design landed in #1133.

## Why This Change

`owner/repo` was asserted in seven places across five modules and two languages, and no module could see what another was asserting. That produced two concrete problems.

The shapes had drifted. `credential_proxy.is_valid_repository` — the validator on the credential sidecar's trust boundary — accepted `acme/..` and `acme/-x`, which every other copy rejected. `audit_report.py` matched the bare regex directly instead of the validator wrapping it. Two of the six regex copies were already dead code.

And the host was being discarded before the slashes were counted, so `provider_for` selected GitHub for anything it did not recognise. A GitLab SCP remote and `https://example.invalid/github.com/o/r` both got `GitHubProvider` and a GitHub token.

Beyond the defects, the shape is what blocks a second forge: a GitLab project at `group/subgroup/project` is ordinary, and six independent copies of "exactly one slash" is six places that have to learn otherwise before any of it can be attempted.

## What Changed

- **`agents/platform/scripts/repo_ref.py`** (new). Standard-library-only, because `credential_proxy.py` imports it and runs in the credential-holding container under a different UID — a validator on that boundary must not pull in `gitops_workspace`, which shells `kubectl`. A host is read only where the syntax states one (a scheme, or the scp `host:path` form), with one exception for a schemeless value whose first segment is a known forge host; inferring one otherwise would read `my.org/repo`, a legal bare slug, as a host and a one-segment path.
- **`forge.provider_for`** parses the host and raises `UnknownForgeHost` (`FORGE_HOST_UNSUPPORTED`) or `RepoUnparseable` instead of falling back. `forge._parse_repo` and its `SETTINGS.md` constants are deleted — #504 removed the path that called them.
- **Six copies of `BARE_REPO_RE`** and two of `_valid_repo_component` are gone, along with `SETTINGS_REPO_RE`, `REPO_URL_RE`, and `github_token_refresh`'s dead `GITHUB_HOSTS` re-export.
- **`get_managed_github_repos()`** logs why it skipped an entry — an unsupported `type`, or a URL it could not resolve — instead of dropping it silently. The design's §9 step 1 originally called for this to become a dispatch on `type`; there is one provider and nothing to dispatch to, so the skip is made loud here and the dispatch moves to step 5, alongside the provider it would dispatch to. The design doc is amended to match.
- **Both `provider_for` call sites in `pr_conversation.py`** moved inside their `try` blocks. They sat above the `except forge.ForgeError` guard because `provider_for` could not raise; now it can, and an unknown host would have been a traceback rather than the reason code the SKILL tells the model to read.

Three behaviour changes worth naming, all narrowings on the credential plane:

- `is_valid_repository` now refuses traversal segments, leading dashes, host-carrying values, and any value it would have to normalise. Strictly narrower — I enumerated the old accept set against the new one and nothing is newly admitted.
- The predicate requires the value to already *be* the slug, not merely normalise to one. `parse` strips whitespace, surrounding slashes, and a trailing `.git`, but every caller checks and then passes the **raw** string on. `credential_proxy` admits it and its caller execs `github_token_refresh.py` with the original, which splits on `/` and posts the left half to Minty as an org name — so `" acme/toolkit\n"` would have reached the broker as `"  acme"`.
- The owner slot may not be a spelling of GitHub's own host. `github.com/acme` fails on depth once the host is lifted, but `www.github.com/acme` and `ssh.github.com/acme` are outside `KNOWN_HOSTS` by design and would read as an owner — one GitHub cannot issue, since a namespace may not contain a dot.

## Context

Implements step 1 of `docs/designs/multi-forge-support.md` §9. PR 2 of 8; the design landed as #1133.

Related open work:

- **#1113** (Follow-up from #504, assigned to @kirill-kosov) — this PR covers the Python half of **B3** (an SCP-form remote on any host silently rewritten to a GitHub repository; `provider_for` now refuses it, but the Go `CleanRepoSlugWithOrg` rewrite remains and is step 2), the Python half of **A7** (uppercase URL scheme — `HTTPS://github.com/acme/toolkit` now resolves), and part of **B2** (a malformed `managed_repos` entry is now logged rather than read back as "nothing registered"). Commented there.
- **#1118** (open, ready) touches `forge.py`, `test_forge.py`, `pr_conversation.py`, `docs/designs/pr-comment-conversation.md` and `docs/README.md`. Two kinds of overlap: a textual conflict on the `BARE_REPO_RE` region of `pr_conversation.py` that this PR deletes, and two new `forge.provider_for(repo=...)` calls in `update_pr.py` that sit outside their `try` guards — the shape this PR corrects, which git will not flag because the files differ. Neither is reachable today (both callers produce bare slugs, which select GitHub without a host lookup), so it is a latent divergence rather than a bug. Deliberately not fixed here; commenting on #1118 instead so whoever rebases second folds it in.

Follow-up filed as #1151: three site pages still describe the `SETTINGS.md` / `Git Repo:` path #504 removed. Pre-existing, surfaced by this PR's prose, deliberately out of scope.

## Testing

- `agents/platform/scripts`: 1274 tests. One pre-existing loader error, `test_session_kv_server`, because `fastapi` is absent from this environment — unrelated and present on the base.
- `fleet-audit` 547, `github-issue-resolver` 61, `pr-conversation` 82 — all green.
- `make docs-check` passes; `prettier@3.9.6 --check` clean on the three changed Markdown files.
- **Mutation-checked the three new gates**, because two of them were added precisely because nothing pinned the behaviour: reverting the slug gate to the old `count("/") != 1` fails 6 subtests; dropping the owner-slot host check fails 5; moving either `provider_for` call back above its guard produces 3 errors. Before these tests, all three reverts left every suite green.

### Live validation

**Not live-tested against a running installation.** The `kagent1` sandbox cluster's control plane is a private endpoint (172.16.0.2, no route from here), and the other two contexts available to me have no `PlatformAgent` CRD. Nothing here observed the sidecar importing `repo_ref` under UID 10001, an operator-authored ConfigMap flowing through the new warning path, or a real `/v1/github/refresh` round trip.

Substituted a differential harness against the base tree, which is weaker but is not nothing: every changed entry point run over a real corpus, then the same corpus against the pre-change code extracted with `git archive refs/kube-agents-base/main | tar -x -C /tmp/rr-base`. Inputs included a real git repository driven with `git remote set-url` and a real state file, not only string literals.

Confirmed deltas, base → head:

```
provider_for("git@gitlab.com:group/project")          GitHubProvider -> UnknownForgeHost(FORGE_HOST_UNSUPPORTED)
provider_for("https://example.invalid/github.com/o/r") GitHubProvider -> UnknownForgeHost(FORGE_HOST_UNSUPPORTED)
is_valid_repository("acme/..")                                  True -> False
is_valid_repository("acme/-x")                                  True -> False
is_valid_repository("github.com/o")                             True -> False
```

I also used the harness to close the one inversion risk the design warns about in §7 — a CR that Go admission accepts and the pod now rejects. 31 Go-admissible URL shapes through `extract_github_slug`, base versus head: 5 deltas, all *widenings* (`:443`, userinfo, `?query`, `#fragment`, `ssh://`), every one still naming `github.com`. No narrowing on that path. Traversal shapes such as `https://github.com/../evil`, which `CleanRepoSlugWithOrg` does admit, were already rejected on the base.

No test artifacts to clean up — nothing was created outside `/tmp`.

## Self-Review

Two passes, `review-adversarial` and `review-docs-drift`, each run twice: once against the first cut and once against the fixed branch. Every run was a fresh subagent that did not write the change, was handed the diff range and nothing else, and — for the re-runs — was not shown the earlier round's findings. Both re-runs are against the current head; the superseded round is dropped, its dispositions folded in below.

**Angles run.** Adversarial: confused-deputy resistance on the URL parser, the accept-set delta of every changed validator, regex backtracking, dead-symbol sweep, packaging (does `repo_ref` actually reach the sidecar image), test-quality-by-mutation, and cross-branch conflict with open PRs. Docs-drift: code→docs (every deleted identifier grepped across all doc and config file types), docs→source (each testable assertion in the two edited design docs checked against code at head, including the Go side, the Dockerfile, and the operator's ConfigMap builder), map coverage and churn, and the mechanical gates.

**Findings and disposition:**

- **MEDIUM, confirmed** — the new slug gate in `refresh_git_credentials` had no test; reverting it left every suite green. *Fixed*, with the mutation check above.
- **Blocking (docs-drift)** — `docs/README.md`'s status cell still read "NOT yet implemented", contradicting the banner this PR rewrote. `docs-check-map` validates presence and alignment only, so nothing caught it. *Fixed.*
- **LOW, confirmed** — the two `provider_for` calls moved inside their guards were never exercised raising; `run_helper` patched `provider_for` with `return_value` in all 80 tests. *Fixed*: added a `provider_error` seam and one test per site.
- **LOW, confirmed** — `RepoRef.__str__` hardcoded `/` while the module declares `PATH_SEPARATOR`. *Fixed.*
- **LOW, confirmed** — #1118 reintroduces the unguarded `provider_for` shape. *Not fixed*: pre-emptively editing a file this PR does not otherwise touch would guarantee a conflict with an open PR to fix something unreachable today. Commented on #1118 instead.
- **Advisory (docs-drift)** ×4 — §3 undercounted the callers that had component rules (three, not two, and `credential_proxy` was the one without); `pr-comment-conversation.md` §3 described selection as raising without mentioning the hostless default, which is the only shape live callers produce; §1's "worth changing regardless" paragraph still read as pending for layer 2; and a source comment claimed a "documented" unregistration path documented only in a Go comment. *All four fixed.*
- **Advisory, not caused by this PR** — three site pages still describe the removed `SETTINGS.md` / `Git Repo:` path. *Not fixed here*: the drift predates this change and correcting three site pages widens the review surface of a refactor. Filed as #1151.

**Reviewer finding (mplakhtiy), confirmed and fixed in `cb7840b0`** — the `KNOWN_HOSTS` shorthand lift also reached the credential plane, so a schemeless `github.com/acme/toolkit` remote passed `github_repo_from_remote` where the base refused it. Reproduced against both heads before fixing. The lift is right for a repository someone *registered* — `github.com/o/r` is the shorthand a person types — and wrong for a git remote, because git emits no such value: `git remote add origin github.com/acme/toolkit` succeeds, but git reads it as a relative local path, so a directory name in a `.git/config` the sandbox controls could mint an installation token for whichever org it named. Rather than dropping the lift, `RepoRef` now records which of the two produced the host: `host_inferred`, set only by the lift, and `host_stated` for a host the value's own syntax named. The credential path gates on `host_stated`; `provider_for` and `extract_github_slug` keep taking the shorthand. Mutation-checked both ways (guard reading `.host` → 2 failures; lift not setting `host_inferred` → 3 failures).

**Reviewer finding (jayantid), confirmed and fixed in `f20bea0f`** — the `KNOWN_HOSTS` shorthand lift also ran on URL-parsed results, so a URL with an empty authority got a host its syntax never stated: `file:///github.com/o/r` read as a GitHub remote, and `file:///gitlab.com/g/p` yielded a hostless ref that `forge.provider_for` maps to `GitHubProvider` by default — the silent fallback this module exists to remove. A URL that parses to no host now raises `RepoRefError` before the lift; the lift is left for the schemeless branch, which is the only one that produces a bare `github.com/o/r`. Both URLs are in `test_repo_ref.py`.

**Rebase onto current `main`** (2026-09-08, `b1658753` → `dff63550`, folding in the earlier 2026-09-07 round). `forge.py`'s "On the repository parser" docstring conflicted both times and was resolved by hand. The second rebase crossed #1200, "harden repository validation", which touched six of the same files; git auto-merged five of them, so rather than trust the absence of conflict markers I read #1200's diff per file and checked each piece of hardening survived — all three `validate_repo_org` call sites, all three new tests, and both `_parse_managed_repos_json` warnings. That audit found one defect the clean merge had introduced: `gitops_workspace.py` ended up with two module loggers, #1200's `logger` wedged into the middle of the import block and this branch's `LOGGER`, with the two warnings bound to the dead one. *Fixed* — consolidated on `LOGGER`, one definition and four call sites. The gates were re-run on the new head.

**Found while fixing, reported by neither pass.** Writing the test for the first finding showed `www.github.com/acme` still passing the gate — the same class of bug it claims to fix, in a spelling `KNOWN_HOSTS` does not carry. The owner-slot host check is the fix, and it is a genuine narrowing relative to both the base and my first cut.

**What the passes could not cover.** No live installation and no Docker build, so the packaging claim rests on reading `deploy/docker/Dockerfile:1698` plus `test_image_startup_contract` rather than on running anything. The Go side is deliberately unread (step 2 moves it); the inversion risk that creates is the one gap the adversarial pass most wanted a second reader on, and the differential sweep under Live validation is my attempt at it rather than a substitute for review. Coverage floor not measured. `make prettier-check` fails on 19 pre-existing `.agents/skills/` files this diff does not touch.

## Risk & Rollout

Moderate blast radius, narrow failure mode. Nothing changes for a correctly-configured GitHub install: `acme/toolkit` and `https://github.com/acme/toolkit` parse exactly as before, and the widenings on the registration path all still name `github.com`.

The one new failure mode is the narrowing on the credential plane. A `managed_repos` entry or a CR carrying a repository value that only *normalised* to a slug — a trailing `.git`, surrounding whitespace, a `github.com/` prefix on a value that was already being treated as bare — now fails validation where it previously passed. That is the point of the change, but on an install with such an entry it presents as a repository the agent stops acting on. It will say so: `get_managed_github_repos()` now logs the entry and the reason, which is the other half of this PR.

Revert is a clean `git revert` — no migration, no persisted state, no CRD or chart change.

---


agents/platform/scripts/repo_ref.py
```diff
@@ -0,0 +1,273 @@
...
+            split = urlsplit(text)
+            host, path = (split.hostname or ""), split.path
+        except ValueError as error:
+            raise RepoRefError(value) from error
+    else:
+        scp = SCP_REMOTE_RE.match(text)
+        host, path = (scp.group("host"), scp.group("path")) if scp else ("", text)
+
+    path = _trim(path)
+    if not host and PATH_SEPARATOR in path:
```
jayantid, line 221: 🟡 **Not merge-blocking.** The `KNOWN_HOSTS` lift also runs on URL-parsed results, so a scheme with an empty authority gets a host the syntax never stated.

The module contract ("a host is read only where the syntax states one … the one exception is a schemeless value") and the amended §3 of `multi-forge-support.md` both exclude this branch, but `parse("file:///github.com/o/r")` returns host `github.com`, and `parse("file:///gitlab.com/g/p")` returns a hostless ref that `provider_for` silently maps to `GitHubProvider` — the fallback shape this PR removes elsewhere. Concretely, `github_repo_from_remote` and `extract_github_slug` both accept `file:///github.com/o/r` where base refused it. No live caller produces these shapes today, so this arms when callers start passing registered URLs (§9 step 3/4) or an administrator hand-edits a `managed_repos` entry. The credential boundary is unaffected: `is_github_slug` still refuses every URL form.

Refusing an empty authority after a scheme makes the docstring true and costs no legitimate input:

```python
        try:
            split = urlsplit(text)
            host, path = (split.hostname or ""), split.path
        except ValueError as error:
            raise RepoRefError(value) from error
        if not host:
            raise RepoRefError(value)
```

jayantid (approved):

jayantid:
**Smoke red triage (gate-watch):** not caused by this PR. [Build 2095075912293814272](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1150/pull-kube-agents-smoke-test/2095075912293814272) never reached the evals: deploy on pool project `kube-agents-evals-4` failed with `Error: UPGRADE FAILED: "kube-agents" has no deployed releases` — the stuck Helm release record tracked in #1172 (evals-4 is currently poisoned; the structural heal is in flight in #1180). `/retest` — a fresh lease will very likely land on a healthy project.

jayantid:
Red-sweep classification: stale false red — 87.9% pass rate, red on a single rung-3 empty-trajectory rep (tokens.total=0; the #1184 class), run started 09-03 01:25Z, before the #1191 fix landed. Not this diff — but the branch now conflicts with main, so prow cannot run a retest until it is rebased.

agents/platform/scripts/repo_ref.py
AntonTyb: Fixed in f20bea0f. A URL that parsed to an empty authority now raises `RepoRefError` before the shorthand lift runs, so `file:///github.com/o/r` is refused rather than read as a GitHub remote — and `file:///gitlab.com/g/p` no longer reaches `forge.provider_for` hostless, where the GitHub default would have taken it. The lift is left for a bare `github.com/o/r`, which only the schemeless branch produces. Comment at `repo_ref.py:180-189` says so, and `test_repo_ref.py` covers both URLs.

The branch was rebased onto current `main` in the same push, so the line numbers moved.

AntonTyb:
/hold

Holding pending the design collision raised on #1246. That PR renames away `docs/designs/multi-forge-support.md`, which this PR amends, and #1253 adds `providers/identity.py`, which answers the same question as this PR's `repo_ref.py`. This PR carries both `approved` and `lgtm` and would otherwise merge as soon as the gate is green, which is the one outcome worth avoiding while who-lands-what is unsettled.

Nothing here is withdrawn and no rework has been done. `/unhold` once #1246 is answered.

deploy/sandbox/Dockerfile
```diff
@@ -295,6 +300,7 @@ COPY agents/platform/governance/ /opt/defaults/governance/
 # Hermes home. Leaving the paths empty reads as a broken image; the stub says what
 # is actually true and exits non-zero.
 COPY agents/platform/scripts/sandbox_exec.py \
+     agents/platform/scripts/repo_ref.py \
```
jayantid: 🟡 **Not merge-blocking.** The PR body never mentions this file, so its one packaging fix is invisible to a reader of the body.

This COPY line is load-bearing: `forge.py` and `gitops_workspace.py` are copied into this image and both `import repo_ref` at module level, so without it the sandbox image fails `smoke-test.sh:272`. The body's What Changed list omits `deploy/sandbox/Dockerfile`, and the Self-Review's packaging note says the packaging claim rests on `deploy/docker/Dockerfile:1698` plus `test_image_startup_contract` — the sidecar image only.

A reader auditing packaging from the body checks the sidecar and stops. The image that actually needed a fix is named nowhere but the diff. The body is otherwise exhaustive, which invites "not listed → not changed".

Fix: one line in What Changed naming this file, and widen the packaging sentence to say the sandbox image is covered by `docker-build.yml`'s presubmit smoke test.

jayantid:
Stale red from the seeded-fleet fixture outage (#1278): the last run (build 2097213176524312576, leased kube-agents-evals-27) collapsed only on the crashloop trio with `required phrases absent: ['OOMKilled']` — the seeded-a fixture is repaired on all 30 pool projects, retesting.

/retest

jayantid (approved):

mplakhtiy (commented):
Reviewed the full diff at `dff63550` in a clean context that did not write the change; merged `upstream/main` cleanly first. One Medium; the Lows are in my notes.

**Coverage note worth stating:** the bot's "No findings" was a wide pass, but its commit `b165875` is not on this branch at all — the branch is now a single commit, so that review read code that no longer exists and is not coverage of this head. jayantid's second approval *is* at the current head.

**The refactor holds where it matters.** I enumerated `credential_proxy.is_valid_repository`'s accept set against the base independently: strictly narrower, six deltas, all narrowings. `provider_for`'s substring-host defect is genuinely gone. The one finding below is a behaviour change that goes the other way.

agents/platform/scripts/github_token_refresh.py
```diff
@@ -171,77 +171,35 @@ def refresh_credentials_once(
...
+    if ref is None:
+        log(f"Ignoring git remote: '{url}' is not a repository URL.")
         return None
-    if host.lower() not in GITHUB_HOSTS:
-        # The host, never the URL: a remote can carry `user:token@` in front of
-        # it. Without this an operator whose clone uses an SSH host alias sees
-        # only the caller's "Could not identify target repository 'None'".
-        log(f"Ignoring git remote: host '{host}' is not a GitHub host.")
+    if not ref.host:
+        log(f"Ignoring git remote: '{url}' names no host.")
```
mplakhtiy: **🟠 Medium — a hostless `github.com/owner/repo` remote now passes this check, on the credential plane, where the base refused it.**

`repo_ref.parse` lifts a schemeless first segment out of the path when it is in `KNOWN_HOSTS`, so `github.com/acme/toolkit` comes back with `host="github.com"` — synthesised, not stated. This guard and `is_github` below it both pass, and the function returns `acme/toolkit`. The base had no such branch: with no `://` it required an scp match, which needs a colon, so the value returned `None`.

Reproduced against a real repository rather than a string literal:

```
$ git init -q . && git remote add origin github.com/acme/toolkit
HEAD get_current_git_repo -> acme/toolkit
BASE get_current_git_repo -> None
```

**Failure scenario.** `git remote add origin github.com/acme/toolkit` is accepted by git — it is a relative local path, not a GitHub URL. On the base, `refresh_git_credentials()` fails with `Could not identify target repository 'None'`. On the head it mints a github.com installation token scoped to org `acme`, runs `gh auth login --with-token` and `gh auth setup-git` with it, and `gitops_workspace.resolve_repo` reports the workspace as `acme/toolkit`. The token stays scoped to the org named in the value, so this is not a cross-org escalation — it is a credential the agent acquires on the strength of a directory path.

**It contradicts three statements the change makes about itself:**

- this function's own new docstring — *"A remote always names a host, so the bare shorthand is refused here even though `repo_ref` parses it: git cannot produce an `origin` of `acme/repo`"*. Git cannot produce an `origin` of `github.com/acme/repo` either, and that one is admitted;
- the PR body's *"Three behaviour changes worth naming, **all narrowings** on the credential plane"*;
- the body's *"the widenings on the registration path all still name `github.com`"*, which locates every widening on `extract_github_slug`. This one is on the remote path.

It is also the same class of defect jayantid raised at `repo_ref.py:195` and you fixed in `f20bea0f` — a host the syntax never stated — reached through the schemeless branch of the lift rather than the URL branch. That fix deliberately "left the lift for the schemeless branch"; this is what the schemeless branch does with it.

**No test covers the shape.** `test_github_token_refresh.py` exercises this through `get_current_git_repo` for https, scp, ssh-over-443, `www.`, lookalike hosts, deep links, traversal, a gitlab remote and `/srv/git/kube-agents.git` — but no fixture is a schemeless `github.com/...`. Change `test_get_current_git_repo_local_path_returns_none`'s fixture to `github.com/gke-labs/kube-agents` and it fails at head, passes at base.

**Suggested fix.** `require_host=True` on `github_slug` does not help — the lift synthesises the host before the check sees it. Either give `parse` a keyword that suppresses the lift for callers reading a remote, or have `RepoRef` record whether the host was stated or inferred so this function can refuse an inferred one. The second generalises the mechanism that has now produced two findings on this branch, rather than patching the second site.

Not merge-blocking.

AntonTyb: Confirmed and fixed in cb7840b0.

I reproduced it exactly as you described: on `dff63550`, `github.com/acme/toolkit` and `GitHub.com/acme/toolkit` both returned `acme/toolkit` from `get_current_git_repo()`, where base returned `None` for each. Reachable through `gitops_workspace.py:864,881` against a workspace the sandbox writes.

The lift itself is right for the caller it was written for — `github.com/o/r` is the shorthand a person types when registering a repository, and `provider_for` and `extract_github_slug` should keep taking it. It is wrong for a git remote, because git emits no such value: `git remote add origin github.com/acme/toolkit` succeeds, but git reads it as a relative local path. So rather than dropping the lift I made the ref record which of the two produced the host. `RepoRef` gains `host_inferred`, set only by the lift, and `host_stated` — a host the value's own syntax named, via a scheme or an scp remote. `github_repo_from_remote` now gates on `host_stated` instead of `host`.

After the fix, `github.com/acme/toolkit`, `GitHub.com/acme/toolkit`, `acme/toolkit` and `/srv/git/x.git` all return `None`; `https://`, scp, ssh-over-443 and the `www.` alias still resolve; `extract_github_slug` is unchanged. Coverage is in `test_get_current_git_repo_local_path_returns_none` plus two new `test_repo_ref` cases, mutation-checked both ways (guard reading `.host` → 2 failures; lift not setting `host_inferred` → 3 failures).

Worth flagging: #1253's `providers/identity.py` has the same behaviour — its `repository_host` returns `github.com` for a schemeless `github.com/acme/toolkit` through the `"." in head` branch. If that design is the one that lands, the distinction needs to go there too.

dshnayder:
> /hold
> 
> Holding pending the design collision raised on #1246. That PR renames away `docs/designs/multi-forge-support.md`, which this PR amends, and #1253 adds `providers/identity.py`, which answers the same question as this PR's `repo_ref.py`. This PR carries both `approved` and `lgtm` and would otherwise merge as soon as the gate is green, which is the one outcome worth avoiding while who-lands-what is unsettled.
> 
> Nothing here is withdrawn and no rework has been done. `/unhold` once #1246 is answered.

@AntonTyb #1246 is answered. This #1150 merges first, #1246 is next.

PR #1153, opened 2026-09-02
fix(auth): add credential refresh and transient retries to forge.py

### LINKED ISSUE #1091: fix(auth): forge.py lacks credential refresh — token expiry falsely reports GITHUB_AUTH_NOT_CONFIGURED
## Summary

`agents/platform/scripts/forge.py` (used by `github_scan_gate.py` in the `pr_comments` sweep) has no credential refresh or retry logic. When the 1-hour GitHub App installation token expires, `gh_preflight()` executes `gh auth status`, which exits non-zero and unconditionally raises `ForgeError("GITHUB_AUTH_NOT_CONFIGURED")`.

This produces false-alarm scheduled warnings in chat:
`⚠️ **GitHub PR watcher is not running:** GITHUB_AUTH_NOT_CONFIGURED`
even though GitHub authentication and the token broker are fully configured and functional.

This structural gap was previously noted as Out-of-Scope in #875:
> "`forge.py` never mints at all, so `sweep_pr_comments` depends on the `issues` sweep having refreshed earlier in the same tick. Related, different fix, separate issue."

## Root Cause & Timing Gap

Tokens minted by `github-token-minter` have a 1-hour (3600s) lifetime.
`github-repo-watcher` runs every 10 minutes (`*/10 * * * *`) and executes sweeps sequentially:
1. `issues` (`resolver.py poll`)
2. `pr_comments` (`forge.py`)

In `resolver.py`, `run_gh()` inspects auth failures via `_looks_like_auth_failure()` and triggers `_refresh_credentials_once()` -> `POST /v1/github/refresh` to the sidecar before retrying (#783).

In `forge.py`, `gh_preflight()` does:
```python
result = run(["auth", "status"])
if result.returncode == 0:
    return
raise ForgeError(
    "GH_CLI_NOT_FOUND"
    if result.returncode == GH_MISSING_RC
    else "GITHUB_AUTH_NOT_CONFIGURED"
)
```

### Expiry Boundary Condition:
Because `forge.py` does not perform token refresh, it relies entirely on `resolver.py` having refreshed the token earlier in the same tick. This creates an hourly boundary condition:
1. **T+0**: Token is minted/refreshed (valid for 3600s).
2. **T+3598s** (tick at minute 60):
   - `sweep_issues` runs `resolver.py poll` -> `gh auth status` returns 0 because the token has ~2 seconds of validity left. `resolver.py` does not refresh the token and exits cleanly.
   - Milliseconds later at **T+3600s**, `sweep_pr_comments` runs `provider.preflight()` -> `gh auth status`. At this exact moment, the token expires. `gh auth status` exits with returncode 1.
   - `forge.py` unconditionally raises `ForgeError("GITHUB_AUTH_NOT_CONFIGURED")` and prints a warning to chat.
3. **T+4200s** (next tick, 10 minutes later):
   - `sweep_issues` runs `resolver.py poll`, sees the expired token, refreshes credentials via the sidecar, and succeeds.
   - `sweep_pr_comments` then passes silently with the refreshed token, causing the warning to disappear on its own.

Additionally, if `GITHUB_WATCHER_SWEEPS="pr_comments"` is configured alone, the PR watcher permanently fails after the initial token expires because `resolver.py` never runs to perform the refresh.

## Proposed Fix

1. **Add refresh capability to `forge.py`**:
   - Call `refresh_git_credentials(repo)` (or delegate via `github_token_refresh`) when `gh auth status` or `GitHubProvider._call` encounters an authentication failure.
   - Retry the command once behind a freshly minted token, matching `resolver.py`.
   - Distinguish `GITHUB_TOKEN_REFRESH_FAILED` from `GITHUB_AUTH_NOT_CONFIGURED` so a broker outage is reported accurately rather than misdiagnosed as missing configuration.
2. **Transient retry for PR conversation fetches**:
   - In `GitHubProvider._call`, consider a single bounded retry for transient network/API timeouts before marking pull requests as `unreadable`.

## Related
- #716 — PR introducing `forge.py`
- #783 — re-mint GitHub token in `resolver.py` instead of reporting unconfigured
- #875 — GitHub token refresh retry (identified this limitation as out-of-scope)

kirill-kosov:
## Summary

This change refactors GitHub App installation token refresh logic into `agents/platform/scripts/github_token_refresh.py` and adds credential re-minting and transient read retries to `forge.py`.

## Why This Change

`forge.py` previously contained no credential refresh or retry logic. When the 1-hour GitHub App installation token expired between cron ticks, `gh_preflight()` executed `gh auth status`, which failed and unconditionally raised `ForgeError("GITHUB_AUTH_NOT_CONFIGURED")`. This surfaced false-alarm warnings in chat indicating the PR watcher was unconfigured when the token had simply expired. Additionally, transient network errors on PR comment fetches caused open PRs to be skipped and reported as unreadable.

## What Changed

- **`agents/platform/scripts/github_token_refresh.py`**: Shared credential refresh helpers (`refresh_credentials_once`, `looks_like_auth_failure`, `is_refresh_failed`, `reset_refresh_state`, `GH_MISSING_RC`, `GH_TIMEOUT_RC`).
- **`agents/platform/scripts/forge.py`**: Updated `run_gh()` to attempt re-minting on authentication failure. Added `retry_transient` to `_call` to retry read operations (`list_open_prs`, `list_comments`, `list_commits`) once on transient errors, while keeping mutating calls (`post_comment`, `acknowledge`) un-retried to avoid duplicate posts.
- **`agents/platform/skills/github-issue-resolver/scripts/resolver.py`**: Removed duplicate private functions (`_refresh_credentials_once`, `_looks_like_auth_failure`, `refresh_credentials`) and migrated to the shared `github_token_refresh` module.
- **`agents/platform/scripts/github_scan_gate.py`**: Updated `_resolver_path()` to search across candidate paths (`profiles/platform`, `hermes_home`, `/opt/platform-template`).
- **Unit Tests**: Updated test suites in `test_forge.py`, `test_github_token_refresh.py`, `test_github_scan_gate.py`, and `test_resolver.py`.

## Context

Closes #1091

## Testing

Automated test suites passed: `test_forge.py`, `test_github_token_refresh.py`, `test_github_scan_gate.py`, and `test_resolver.py` (274 unit tests passed in 0.35s).

### Live validation

Not live-tested on a cluster; validated locally using comprehensive unit test coverage and mock command execution across all provider endpoints and token failure modes.

## Self-Review

Passes executed in clean context:
- **Adversarial pass**:
  - Checked whether `retry_transient` could double-post comments or reactions: confirmed mutating calls (`post_comment`, `acknowledge`) default to `retry_transient=False` and `GH_TIMEOUT_RC` is excluded from auth retries.
- **Docs-drift pass**:
  - Verified module docstrings and constant descriptions across `github_token_refresh.py`, `forge.py`, and `resolver.py`.

## Risk & Rollout

Low risk. Reverting the commit restores independent per-script execution without affecting sidecar authentication endpoints.


agents/platform/scripts/forge.py
```diff
@@ -385,6 +381,22 @@ def run_gh(argv: Sequence[str]) -> subprocess.CompletedProcess:
...
+    A missing binary is reported as `GH_MISSING_RC` so it stays distinguishable
+    from a command that ran and failed.
+    """
+    result = run_gh_once(argv)
+    if looks_like_auth_failure(argv, result) and refresh_credentials_once(argv):
+        result = run_gh_once(argv)
+    return result
+
+
 def gh_preflight(run: Callable[[Sequence[str]], subprocess.CompletedProcess] = run_gh):
```
mplakhtiy, line 423: **HIGH — a refused mint is still reported as `GITHUB_AUTH_NOT_CONFIGURED`.**

`gh_preflight` (the `raise` at lines 410-414, just below this diff hunk) still branches two ways: `GH_CLI_NOT_FOUND` on `GH_MISSING_RC`, `GITHUB_AUTH_NOT_CONFIGURED` on everything else. It never consults `is_refresh_failed()` — the import at line 354 pulls in `GH_MISSING_RC`, `looks_like_auth_failure`, and `refresh_credentials_once`, but not that one. The only consumers of `is_refresh_failed` are `resolver.py:33` and `resolver.py:471`.

**Failure scenario.** The installation token has expired and the credential sidecar or Minty is down. `sweep_pr_comments` calls `provider.preflight()` → `gh auth status` exits non-zero → `looks_like_auth_failure` returns True on its `auth status` short-circuit → `refresh_credentials_once` raises inside `refresh_git_credentials`, sets `_refresh_failed = True`, returns False → no retry → `gh_preflight` raises `ForgeError("GITHUB_AUTH_NOT_CONFIGURED")` → `github_scan_gate._forge_warning` (line 387) renders it verbatim and `main()` prints it to stdout, the delivery channel. The room gets `⚠️ **GitHub PR watcher is not running:** GITHUB_AUTH_NOT_CONFIGURED` — a broker outage reported as missing configuration.

Three sources say that string is wrong here:

- #1091's Proposed Fix item 1: "Distinguish `GITHUB_TOKEN_REFRESH_FAILED` from `GITHUB_AUTH_NOT_CONFIGURED` so a broker outage is reported accurately rather than misdiagnosed as missing configuration." This PR says `Closes #1091`.
- `docs/site/src/content/docs/overview/architecture.mdx:126`: "`github-repo-watcher` has no ladder to route through … so a broker that refuses surfaces in the room as `GITHUB_TOKEN_REFRESH_FAILED`". `github-repo-watcher` runs both sweeps, so after this change the `pr_comments` half contradicts that sentence — and it is this change that makes the sentence reachable at all, since forge previously never minted.
- `github_scan_gate.py:44-52` lists `GITHUB_TOKEN_REFRESH_FAILED` among the codes.

The fix is the three lines `resolver.py:471-476` already carries. No test covers the path either way: `PreflightTest` injects a fake `run` and never reaches `run_gh`, and `test_auth_failure_retries_after_credential_refresh` mocks `forge.refresh_credentials_once` to return `True`, so the broker-refused branch is unexercised in both directions.

agents/platform/scripts/forge.py
```diff
@@ -420,15 +432,28 @@ def __init__(self, run: Optional[Callable] = None):
...
         a `gh` failure that survived the preflight: the credential works
         somewhere, just not here.
+
+        When `retry_transient=True` (for read-only queries), a non-zero exit gets
+        a single bounded retry before raising `REPO_UNREACHABLE`. Mutating calls
+        (e.g., post_comment, acknowledge) must leave `retry_transient=False` to
+        avoid double-posting on a sidecar timeout.
         """
         result = self._run(list(argv))
+        if result.returncode != 0 and retry_transient:
```
mplakhtiy: **MEDIUM — `retry_transient` retries every non-zero exit, and multiplies the sweep's worst case.**

There is no classification and no delay. The parameter is named for transient faults and the docstring above says "read-only queries", but the condition is `returncode != 0` — which covers a 404 on a repository outside the installation, a 403, a rate limit, and a hung proxy equally.

**Failure scenario.** `run_gh_once` caps each call at `GH_TIMEOUT_S = 60`, and that constant's own comment (lines 89-91) says why: "A hung proxy must not hold the cron tick's per-job lock open indefinitely." With this retry one `_call` can spend 120s on a hung proxy, or 180s when `run_gh`'s auth retry also fires. `sweep_pr_comments` makes one `list_open_prs` per managed repo and, per agent pull request, two `_collect` calls plus `list_commits` — all four sites pass `retry_transient=True` (lines 523, 642, 728). Against a wedged sidecar the sweep's ceiling grows 2-3x at a `*/10` schedule with a per-job lock, which is the condition the timeout was chosen to avoid. On a permanent 404 the cost is smaller but still doubled API traffic on every tick, forever.

Narrower form: skip the retry when stderr carries a definitive HTTP status (401/403/404) or when the return code is the sidecar's `GH_TIMEOUT_RC` — the same shape `looks_like_auth_failure` already uses to decide what is worth a second attempt.

agents/platform/scripts/github_scan_gate.py
```diff
@@ -195,7 +197,17 @@ def selected_sweeps() -> tuple[tuple[str, ...], list[str]]:
...
+    candidates = (
+        home / PLATFORM_PROFILE_DIR / RESOLVER_REL,
+        home / RESOLVER_REL,
+        Path(PLATFORM_TEMPLATE_DIR) / RESOLVER_REL,
+        Path(__file__).resolve().parent.parent / RESOLVER_REL,
+    )
+    for candidate in candidates:
+        if candidate.is_file():
+            return candidate
+    return home / PLATFORM_PROFILE_DIR / RESOLVER_REL
```
mplakhtiy: **MEDIUM — the fallback names a path that cannot exist on the platform profile.**

The platform cron runs with `HERMES_HOME=/opt/data/profiles/platform` (`deploy/docker/Dockerfile:1262`; `deploy/docker/plugins/chat/test_chat_relay.py:116` pins the same value; `deploy/shared/docker-entrypoint.sh:885` symlinks `profiles/platform/scripts` → `scripts`). Against that home:

- candidate 1 → `/opt/data/profiles/platform/profiles/platform/…`, never exists;
- candidate 2 → the old path, and the one that resolves;
- candidate 3 → `/opt/platform-template/skills/…`, present in the image;
- candidate 4 → `Path(__file__).resolve()` follows the symlink to `/opt/data`, so it probes `/opt/data/skills/…`, which does not exist.

So on the platform profile the change is inert except for three extra `stat` calls — but the fallback at line 210 is candidate 1, not the old `home / RESOLVER_REL`. When the resolver genuinely is missing, `run_resolver_poll` now raises `FileNotFoundError(f"resolver not found at {script}")` naming `/opt/data/profiles/platform/profiles/platform/skills/github-issue-resolver/scripts/resolver.py`, a directory no install has, where the previous message named the correct expected location. That turns the one diagnostic an operator gets for a broken scaffold into a wrong one.

If the motivation is a scan gate running with `HERMES_HOME` unset or pointed at `/opt/data`, the candidate order is right for that case and only the fallback needs to change.

agents/platform/skills/github-issue-resolver/scripts/resolver.py
```diff
@@ -28,39 +28,13 @@
...
-    r"|not logged in"
-    r"|token is invalid"
-    r"|invalid token",
-    re.IGNORECASE,
+from github_token_refresh import (
+    GH_MISSING_RC,
+    is_refresh_failed,
+    looks_like_auth_failure,
+    refresh_credentials_once,
 )
```
mplakhtiy, line 30: **MEDIUM — the `sys.path` comment above now states the opposite of what the code does.**

Lines 19-20 still read: "The import itself is lazy, in refresh_credentials below, so this module still imports on a dev machine with nothing staged under /opt." `refresh_credentials` was deleted by this diff and the import is now module-level, right here — so both sentences are false, and the property the second one documents (resolver.py importing with nothing staged under `/opt`) is what this change removed.

Not a live break: the three `sys.path.append` entries cover the container and the source checkout, and the suite passes. The cost is that the next reader is told the module has a property it no longer has, on the line they would consult before adding a fourth caller.

Separately, this import sits after `SCRATCH_DIR = "/opt/data/scratch"` rather than in the import block above it.

mplakhtiy (commented):
Adversarial review pass over the branch diff. Findings only — seven inline below, one coordination note here.

#### #1150 rewrites the same regions of the same three files

#1150 (`refactor(platform): parse repository identity in one place`) touches `forge.py`, `github_token_refresh.py`, `resolver.py`, `test_forge.py`, and `test_github_token_refresh.py`. Its hunk headers include `github_token_refresh.py @@ -39,77 +39,35 @@` — the exact span this PR inserts 112 lines into — and `resolver.py @@ -61,14 +61,6 @@`, inside the block this PR rewrote. A textual conflict is near-certain whichever lands second.

The substantive overlap: #1150 centralises repository parsing, and finding 5 below is a new ad-hoc `-R`/`--repo` parse added here. Whoever rebases should use #1150's helper rather than carry a fourth parser forward. #1118 (`pr_updates` sweep) also touches `forge.py` and `github_scan_gate.py`; its regions look further apart but will need the same attention.

agents/platform/scripts/github_token_refresh.py
```diff
@@ -38,6 +39,117 @@ def log(msg: str):
...
+                    repo = argv_list[argv_list.index(flag) + 1]
+                    break
+                except (ValueError, IndexError):
+                    pass
+
+    if not repo:
+        try:
+            from gitops_workspace import get_managed_github_repos
+            managed = get_managed_github_repos()
+            repo = managed[0] if managed else None
```
mplakhtiy, line 145: **MEDIUM — forge always scopes its mint from `managed[0]`, never the repo whose call failed.**

`refresh_credentials_once` derives the repository from `-R`/`--repo` in the argv, then falls back here. Every forge call is `gh api repos/{repo}/…` (`forge.py` lines 517-523, 642, 673, 693, 722-728) — the flag never appears — so forge always takes this fallback. Meanwhile `sweep_pr_comments` iterates every managed repository (`github_scan_gate.py:575`).

**Failure scenario.** On an install whose managed repositories span two organisations, a 401 while listing `orgB/y` mints a token scoped from `orgA/x`. `refresh_git_credentials` expands `repositories_to_scope` only to repositories whose org matches (the `m_org.lower() == org_name.lower()` guard), and the sidecar path re-runs the same helper with the single repository it was handed (`credential_proxy.py:3226-3228`), so the new token does not cover `orgB`. The retry fails identically, the at-most-once guard blocks a second mint for the rest of the process, and the sweep reports `REPO_UNREACHABLE` for every remaining `orgB` repository on that tick. Same-org multi-repo is unaffected.

`resolver.py` shares the fallback but passes `-R` on its calls, so it hits the right repository; forge is the case where the fallback is the only path. Passing the repo through — `_call` knows it, and `refresh_credentials_once` already takes a `repo=` keyword — closes it.

agents/platform/scripts/forge.py
```diff
@@ -355,15 +351,15 @@ def _parse_repo(configured: str) -> str:
     return repo
 
 
-def run_gh(argv: Sequence[str]) -> subprocess.CompletedProcess:
-    """One `gh` invocation, never raising for a non-zero exit.
+from github_token_refresh import (
+    GH_MISSING_RC,
+    looks_like_auth_failure,
+    refresh_credentials_once,
+)
```
mplakhtiy: **LOW — the shared import sits mid-file, 270 lines below the other imports.**

Beyond placement, it gives a module documented as importable by "every skill script on the Platform Agent's `sys.path`" (lines 3-5) a transitive import-time dependency on `github_token_refresh` → `credential_proxy_client`, and `forge.py` does no `sys.path` work of its own. Five modules import forge; `pr_triggers.py:60` and `github_scan_gate.py:72` do no path setup and rely purely on co-location in `$HERMES_HOME/scripts`, which holds today. `github_token_refresh` also mutates `sys.path` at import (three `append`s and a `sys.path.insert(0, …)`), so `import forge` now carries that side effect in every process that does it.

Moving this to the import block at line 86 costs nothing and makes the dependency visible where a reader looks for it.

agents/platform/scripts/forge.py
```diff
@@ -385,6 +381,22 @@ def run_gh(argv: Sequence[str]) -> subprocess.CompletedProcess:
...
+
+    A failed call gets one retry behind a freshly minted token on auth failures.
+    Callers here always need the reason code more than the exception: a token
+    without scope for this repository and a repository that 404s both exit
+    non-zero with usable stderr, and turning that into a traceback loses it.
+    A missing binary is reported as `GH_MISSING_RC` so it stays distinguishable
+    from a command that ran and failed.
+    """
+    result = run_gh_once(argv)
+    if looks_like_auth_failure(argv, result) and refresh_credentials_once(argv):
```
mplakhtiy: **LOW — forge's own timeout escapes the `GH_TIMEOUT_RC` exclusion.**

`run_gh_once` maps `subprocess.TimeoutExpired` onto return code **1**, not `GH_TIMEOUT_RC` (124) — see lines 375-381. `looks_like_auth_failure` excludes 124, but for `gh auth status` it short-circuits to `True` on any non-zero code, so a preflight that hangs for the full 60s spends a mint, which is the case the shared helper's comment says a timeout should not buy.

Harmless in effect (`auth status` is idempotent and the mint is capped at one), but the exclusion only ever covers the sidecar's own 124 here, never forge's local timeout. Mapping the timeout onto `GH_TIMEOUT_RC` would make it hold.

kirill-kosov: Fixed

agents/platform/scripts/forge.py
kirill-kosov: Fixed

agents/platform/scripts/github_token_refresh.py
kirill-kosov: Fixed

agents/platform/skills/github-issue-resolver/scripts/resolver.py
kirill-kosov: Fixed

agents/platform/scripts/github_scan_gate.py
kirill-kosov: Fixed

agents/platform/scripts/forge.py
kirill-kosov: Fixed

agents/platform/scripts/forge.py
kirill-kosov: Fixed

agents/platform/scripts/forge.py
```diff
@@ -420,15 +457,39 @@ def __init__(self, run: Optional[Callable] = None):
...
+            result = self._run(list(argv))
+        if (
+            result.returncode != 0
+            and retry_transient
+            and _should_retry_transient(result)
+        ):
+            try:
+                result = self._run(list(argv), repo=repo)
+            except TypeError:
+                result = self._run(list(argv))
```
mplakhtiy: **MEDIUM — this `except TypeError` probe can silently re-invoke a mutating call.**

The `except` is not narrowed to the call's own signature mismatch. A `TypeError` raised anywhere inside the runner's body — after it has already shelled out — is caught here and the runner is invoked a second time. `post_comment` (`forge.py:716`) and `acknowledge` (`forge.py:737`) both route through `_call` and both now pass `repo=`, so this path sits directly under the guarantee the docstring three lines above makes: "Mutating calls … must leave `retry_transient=False` to avoid double-posting." `retry_transient` gates the transient retry at line 484. It does not gate this one.

Not reachable today, and worth being precise about that: `run_gh`'s post-subprocess path is TypeError-proof by accident — `refresh_credentials_once` wraps its `gitops_workspace` import and its `refresh_git_credentials` call in `except Exception` (`github_token_refresh.py:130-145`), and `looks_like_auth_failure` does nothing that can raise. So this is latent. It becomes live the first time anyone adds a line to `run_gh` between the subprocess returning and `_call` receiving the result, and the symptom then is a duplicate comment on a customer's pull request, attributed to nothing in the diff that introduced it.

What makes it worth closing now rather than noting is that the shim exists only to keep test doubles working, and this same commit demonstrates the alternative. `FakeGh.__call__` was given `**_kwargs` (`test_forge.py:62`); the five ad-hoc `def run(argv)` closures at `test_forge.py:403, 415, 429, 441, 453` were left alone and production code was widened around them instead. Give those five the same treatment and both `try` blocks go away.

Two smaller things in the same place. `_call`'s docstring documents `expect_json` and `retry_transient` but not `repo`, and `GitHubProvider.__init__`'s `run: Optional[Callable]` (`forge.py:452`) still says nothing about the new keyword — so the seam contract changed and is now discoverable only by reading the exception handler. And nothing tests the shim's own failure mode: there is no case asserting that a `TypeError` from inside a runner is *not* swallowed.

agents/platform/scripts/github_token_refresh.py
mplakhtiy, line 145: **MEDIUM — the new `repo=` plumbing does not reach the multi-org case #1091 describes.**

`sweep_pr_comments` calls `provider.preflight()` (`github_scan_gate.py:561`) before the repo loop (`github_scan_gate.py:575`) — both outside this diff. On the exact scenario the PR targets, an installation token that expired between ticks, `gh auth status` fails there first. `looks_like_auth_failure` short-circuits to `True` on `auth status`, `run_gh` calls `refresh_credentials_once(["auth", "status"], repo=None)`, no `-R` is present in that argv, and the helper falls through to `managed[0]` at line 138. The mint is spent, scoped to whatever organisation the first managed repository belongs to. The `_refresh_attempted` guard at line 120 then makes every later `refresh_credentials_once(argv, repo="orgB/y")` return `False` **before it ever reads `repo`**.

**Failure scenario.** An install whose managed repositories span two organisations, token expired. Preflight mints for `orgA` and passes. The loop reaches `orgB/y`, gets a 401, cannot mint again, and — because the whole loop sits inside one `try` (`github_scan_gate.py:579`) — the sweep aborts with `REPO_UNREACHABLE` rather than skipping that one repository. That is the behaviour on `main` today.

The keyword does help in a narrower case: preflight passes because the token is still valid for `orgA`, and a repository in `orgB` added since the last mint 401s inside the loop. There the mint is now correctly scoped, and that is a real improvement worth keeping.

Not a regression. The ask is only that multi-org not be left looking closed — doing that properly means either scoping the preflight mint (which has no single repository to name) or lifting the at-most-once guard to once-per-organisation, and both are larger than this change. A follow-up issue would cover it.

mplakhtiy (commented):
Two MEDIUM findings, anchored inline below. Full-diff review at `522aa14` against `main`@`667565d5`.

I also re-verified the seven round-1 findings against the code at head rather than against the replies — all seven are fixed, and the refused-mint path at `forge.py:433` now branches in the same order `resolver.py:468` uses, with tests covering both the injected-flag and real-`run_gh` paths.

jayantid:
Your smoke red is **not caused by this PR**. `compliance-rbac-overgrant` — the only FAILED case in your build — regressed environment-side early today (credential-proxy reachability to the seeded clusters' control planes; evidence in #1171) and it failed the same way across six unrelated PRs this morning. Per the gate's demotion protocol it's being removed from the blocking roster in #1173; once that merges, a `/retest` should go green barring genuine issues in your diff. Sorry for the noise — this case is held out until #1171's re-admission bar is met. (Your 14:31Z rerun may also clear it if it drew a healthy eval project.)

agents/platform/scripts/forge.py
kirill-kosov: Fixed

agents/platform/scripts/github_token_refresh.py
kirill-kosov: Documented Multi-Org Limitation

mplakhtiy (approved):

mplakhtiy (approved):

jayantid:
**Smoke red triage (gate-watch), update for the new run:** the fresh red ([build 2095192734196305920](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1153/pull-kube-agents-smoke-test/2095192734196305920)) is a *different* environmental cause than this morning's: the fleet-wide Vertex 429 window ~18:00-19:25Z (#1097/#1184) — `rca-remediation-pr` collapsed with the platform agent crashing on 429s after delegation. Still not this PR. Good news: #1173 merged at 18:51Z so `compliance-rbac-overgrant` no longer gates. `/retest` once the quota window calms (or after #1183).

PR #1159, opened 2026-09-02
feat(operator)!: declare the forge in spec.integration.git

### LINKED ISSUE #1085: Repository resolution from SETTINGS.md accepts any host, so a PlatformAgent can point the token refresher at another org's repo
`gitops_workspace.resolve_repo` tries `repo_from_settings(SETTINGS.md)` before falling back
to `github_token_refresh.get_current_git_repo()`, and on the audit-cron path the git remote
is not available at all — so `repo_from_settings` is the parser that decides which repository
the agent asks Minty for.

That parser does no host check. It strips a leading `https://(www.)?github.com/` if present
and otherwise takes the last two path segments:

```
'https://evil.example/gke-labs/kube-agents.git' -> 'gke-labs/kube-agents'
'https://github.com.evil.example/o/r.git'       -> 'o/r'
'https://evil.example/github.com/o/r.git'       -> 'o/r'
'git@evil.example:o/r.git'                      -> 'git@evil.example:o/r'
```

The operator does not close it upstream either: `ValidateGitRepoURL`
(`k8s-operator/api/v1alpha1/common_types.go`) checks length, control characters and scheme,
and requires only `u.Host != ""`. A `PlatformAgent` with
`gitRepo: https://evil.example/victim-org/victim-repo.git` is therefore admitted, written into
`SETTINGS.md`, and resolves to `victim-org/victim-repo`.

Minty's CEL rules are the thing that actually decides whether a token is issued, so this is a
malformed request rather than a guaranteed cross-org token — but the local check that is
supposed to stop it is absent on the path that wins.

`forge.py`'s "On the repository parser" note already records this parser as knowingly loose
("out of scope here and noted rather than fixed").

### Suggested fix

Give `repo_from_settings` the same parsed-host check that `github_repo_from_remote` in
`agents/platform/scripts/github_token_refresh.py` now has, and decide separately whether
`ValidateGitRepoURL` should reject a non-GitHub host at admission.

Found by the adversarial preflight pass on the CodeQL alert 20 fix (`github_token_refresh.py`),
which hardened the fallback parser only.

AntonTyb:
## Summary

`spec.integration.git` is now the PlatformAgent's forge declaration: a provider, an
optional host, a repository, and an optional namespace. `spec.integration.github` stays
as a deprecated alias for `provider: github`, and every existing values file, installer
output, and hand-written CR keeps working unchanged. Behind the field, validation stops
being a set of GitHub-shaped string checks and becomes a lookup in a provider registry —
each provider states its hosts, its namespace grammar, and its path depth, and the
webhook dispatches to whichever one the CR declared. Only `github` is registered, so no
CR can name a forge the agent cannot talk to.

This is step 2 of `docs/designs/multi-forge-support.md` §9, stacked on #1150.

## Why This Change

`CleanRepoSlugWithOrg` counted path slashes *after* discarding the host, so
`git@gitlab.com:group/project` was admitted, reduced to `group/project`, and written into
the GitOps state ConfigMap as `https://github.com/group/project` labelled
`"type":"github"` — a different, real repository that the agent would then attempt to
push to. A second path did the same thing more quietly: `CleanRepoURLWithOrg` returned
http(s) URLs verbatim, so `https://gitlab.com/o/r` seeded an entry whose `type` and `url`
named different forges. A test row pinned that second behaviour, so it was load-bearing
rather than accidental.

Both are the same missing concept — nothing in the operator knew which forge a repository
belonged to — which is also why there is nowhere to put GitLab support. Closes #1085.

## What Changed

- `api/v1alpha1/gitprovider.go` — the `GitProvider` type and its registry. A provider
  states four things (hosts, default host, namespace grammar, path depth bounds) and
  `Resolve` turns a declared host, repository, and namespace into a canonical `RepoRef`
  or an error. `lookupGitProvider` is parameterised over the table so tests can register
  a second provider without shipping a CRD enum value the agent cannot honour.
- `api/v1alpha1/common_types.go` — `GitSpec`, and `CleanRepoSlugWithOrg` /
  `CleanRepoURLWithOrg` marked deprecated with the defect they carried spelled out. The
  new fields carry schema patterns as well as webhook validation: `ENABLE_WEBHOOKS=false`
  is a supported mode, so the API server has to keep refusing whitespace and control
  characters on its own.
- `internal/webhook/platformagent_webhook.go` — provider-dispatched validation, reporting
  against the field the CR actually wrote (`spec.integration.git.repository` or
  `spec.integration.github.gitRepo`) rather than a fixed path.
- `internal/controller/platformagent_controller.go` — `ManagedRepoEntry.Type` now carries
  the declared provider instead of a hardcoded `"github"`, and the seed dedup compares
  resolved identity rather than URL strings.
- `charts/kube-agents` — renders the `git` block, refuses both spellings at once at
  `helm install` where the administrator can still see which values file set what, and
  refuses `githubMinter.enabled` alongside a non-GitHub provider.

## Context

- Design: `docs/designs/multi-forge-support.md` (merged in #1133).
- Stacked on #1150 (`refactor(platform): parse repository identity in one place`).
  **Review commit `11a46c1c` only** — GitHub will not let a PR base on a branch outside
  the base repository, so this opens against `main` as a draft and is rebased once #1150
  merges.
- Closes #1085.
- Deferred to §9 step 5, and recorded there: making `install.sh` and
  `terraform/examples/full-install`'s GitHub App inputs provider-conditional. A Terraform
  variable whose only legal value is its default is configuration nobody can set, so it
  lands with the second provider rather than ahead of it. §6 of the design says the same.

## Testing

- `go build ./... && go vet ./... && go test ./...` in `k8s-operator/` — green.
- `python3 -m unittest tests.test_chart_git_integration` — 9 tests, green. New file.
- `make generate manifests`, `make chart-sync`, `make chart-check` — clean.
- `make docs-check` — clean. `prettier@3.9.6 --write` on every changed Markdown and YAML
  file reported all unchanged.

### Live validation

**Not live-tested.** I have no reachable kube-agents installation, and this stack has
none available to it — the same constraint recorded on #1150.

In its place, a differential harness: `git archive` of the merge base and of this commit
extracted to two throwaway trees, with the same probe compiled against each, so the
before/after is the real code rather than a description of it.

```
BASE git@gitlab.com:group/project.git              -> https://github.com/group/project   err=<nil>
BASE https://gitlab.com/group/project.git          -> https://gitlab.com/group/project   err=<nil>
BASE https://bitbucket.org/team/repo               -> https://bitbucket.org/team/repo    err=<nil>
BASE ssh://git@github.com:gke-labs/kube-agents.git -> https://github.com/gke-labs/kube-agents

HEAD git@gitlab.com:group/project.git              -> refused: names host "gitlab.com", which is not a github host
HEAD https://gitlab.com/group/project.git          -> refused: names host "gitlab.com", which is not a github host
HEAD https://bitbucket.org/team/repo               -> refused: names host "bitbucket.org", which is not a github host
HEAD ssh://git@github.com:gke-labs/kube-agents.git -> https://github.com/gke-labs/kube-agents (unchanged)
```

The first BASE line is the defect: another forge's repository silently became a GitHub
one. The last line is the control — a value that was already correct still resolves the
same way, so the refusals are the new host check rather than the parser having become
stricter about everything.

The chart half was exercised with real `helm template` renders rather than assertions
about the templates (`tests/test_chart_git_integration.py`), covering the `git` block, the
alias, both-at-once, an unregistered provider, the `None` sentinel, and the minter guard.

What this does not cover: whether the operator reconciles the new field in a live
cluster, whether the webhook rejects a bad CR at `kubectl apply`, and whether the agent
pod reads the resulting ConfigMap. Envtest covers the webhook wiring; nothing here covers
the pod. Both need an install.

The throwaway trees were under `/tmp` and are not in the diff.

## Self-Review

Two passes, each in a subagent handed only the diff range — neither had written the
change or seen the reasoning behind it (`/pr-preflight`).

**Adversarial pass.** Looked for: values the new validation admits that it should not,
values it now refuses that used to work, divergence between the Go parser and
`repo_ref.py`, the chart rendering a CR the operator rejects, and behaviour that depends
on the webhook being enabled. Eight findings, all fixed:

1. `Resolve` validated the repository's own segments but not the ones a declared
   `namespace` contributed, so `namespace: ..` reached the state ConfigMap as a traversal
   the parser had already been asked to refuse. Now every segment goes through
   `safeRepoSegment`. Test: `TestResolveValidatesEveryPathSegment`.
2. `GitSpec.Host` and `.Namespace` had no schema pattern, so under
   `ENABLE_WEBHOOKS=false` — documented and supported — the API server accepted
   whitespace and control characters. Both now carry a provider-independent pattern, with
   a comment saying why it is deliberately looser than the provider's.
3. The seed dedup compared `ManagedRepoEntry.URL` as a string, so
   `git@github.com:o/r.git` and `https://www.github.com/o/r` became two entries for one
   repository. Now `sameManagedRepo` compares resolved identity. Test:
   `TestSameManagedRepoComparesIdentityNotSpelling`.
4. A host declared in `spec.integration.git.host` was validated but not canonicalised, so
   `host: ssh.github.com` seeded an entry the agent's own registration check refuses on a
   spelling the operator accepted. Test: `TestResolveCanonicalisesADeclaredHost`.
5. `splitAuthority` fell back to reading a non-numeric port as an scp path under *any*
   scheme, so `https://github.com:evil/owner` resolved to a repository nobody wrote. The
   fallback is now confined to `git` and `ssh`, where it is git's own syntax. The same
   finding turned up a real Go/Python divergence in `ssh://git@github.com:owner/repo`;
   Go's reading is git's, Python's is wrong, and `repo_ref.go`'s header now records that
   with the step (§9 step 3) that corrects the Python side. Test:
   `TestNonNumericPortIsAnScpPathOnlyUnderGitAndSsh`.
6. The chart passed any `provider` string through to the CR, so a typo surfaced as an
   API-server enum rejection naming no values key. The helper now checks against the
   registered list. Test: `test_an_unregistered_provider_fails_the_render`.
7. `gitRepo: None` — the "no GitOps repository" sentinel — counted as a `github`
   declaration in the chart helper, so an install that had opted out could not adopt the
   `git` block. Fixed, and tested. The same finding caught a comment on
   `NoRepositorySentinel` claiming `install.sh` emits the value; greps show nothing in the
   repository writes it (`hack/ci-deploy.sh` opts out with an empty string), so the
   comment now says what is true.
8. The Degraded condition reported "Invalid GitHub repository URL" for a CR that had
   declared no GitHub anything. Message widened; the `InvalidGitRepoURL` reason string
   deliberately stays, because it is machine-readable and changing it would break anyone
   matching on it.

`Resolve` also now lifts only the canonical host out of a schemeless path, matching
`repo_ref.py`'s `KNOWN_HOSTS`; lifting every alias would have made `www.github.com/o/r` a
two-segment path. Test: `TestResolveLiftsOnlyTheDefaultHostFromASchemelessPath`.

**Docs-drift pass.** Looked for prose describing the GitHub-only shape as current. Found
the design's own §1/§2/§3/§6 still written in the future tense about what this commit
lands, the CRD reference page and operator index describing only `github`, the chart
README's GitHub bullet, and `docs/README.md`'s status cell. All updated. No Blocking
finding left open.

**Rebase onto current `main`** (2026-09-08, `11a46c1c` → `3c7eafa9`, folding in the
earlier 2026-09-07 round, whose conflicts were confined to `docs/README.md`'s map rows).
The second rebase crossed #1200, which hardened `CleanRepoSlugWithOrg`'s hand-rolled
parser in place while this branch replaces that body with a `provider.Resolve` call.
Taking either side wholesale would have dropped one of them, so before keeping the
delegation I checked `Resolve` is a strict superset of what #1200 added: the segment
charset regex, the `.`/`..` traversal refusal, leading-dash flag injection, and
non-GitHub host rejection in both the scp and `host/owner/repo` forms. It is, and this
branch never carried the `https://` short-circuit #1200 removed. Three consequences:

- `common_types_test.go`'s two GitLab rows now expect rejection on both sides. #1200
  flipped them by hardening the parser; here both helpers resolve through the declared
  provider and refuse together. The comment credits #1200 rather than describing a defect
  that no longer exists on either side.
- `validRepoSlugPart` was left dead once the inline body it served was gone. Found by
  grepping after resolving, not by the compiler — an unused package-level `var` builds
  fine. Removed.
- The Degraded branch merged #1200's named constants and `CorruptManagedRepos` case with
  this branch's widened message, keeping #1200's admission-webhook clause because
  `platformagent_controller_test.go:857` and `:929` assert on it.

`platformagent-crd.md` conflicted on the same paragraph both sides rewrote; #1200's
upgrade-note caution is kept, restated against `spec.integration.git.repository` with the
deprecated spelling named, and the `Degraded` row now points at `git` rather than the
alias. Gates re-run on the new head.

**One thing to confirm rather than a finding.** The commit is `feat(operator)!:` with the
breaking marker, because a CR naming another forge's host was accepted before and is
rejected now. That is the defect being fixed, so the argument for dropping the `!` is
that nobody was relying on it deliberately; the argument for keeping it is that a CR that
applied yesterday may not today. I kept it — say the word and I will drop it, since the
release classification is yours.

## Risk & Rollout

The blast radius is CR admission and the GitOps state ConfigMap. The new failure mode is
a PlatformAgent that used to apply and now does not: specifically, one whose `gitRepo`
names a non-GitHub host. Those installs were already broken — the agent was pushing to,
or failing against, a GitHub repository of the same path — but they were broken silently
and are now loud, which will look like a regression to whoever hits it. The Degraded
condition names the host and the field.

Reverting is a straight revert: the alias path is unchanged, so no CR written against the
old shape needs migrating back. Nothing has to happen at merge time beyond rebasing onto
#1150.


fkc1e100:
Closing as stale. This forge integration PR has been open for 21 days with unresolved conflicts.

This is part of the multi-forge work. If still needed, please re-evaluate against current architecture.

PR #1161, opened 2026-09-02
refactor(agents): open pull requests through the forge provider

AntonTyb:
`submit_suggestion.py` opens pull requests through `forge.ForgeProvider` instead of shelling `gh` at three call sites of its own. This is step 3 of [`docs/designs/multi-forge-support.md`](https://github.com/gke-labs/kube-agents/blob/main/docs/designs/multi-forge-support.md) §4, for one of the three consumers that step covers.

## Context

**Stacked.** This branch sits on top of #1159 (step 2), which sits on #1150 (step 1). GitHub will not take a base branch that lives on a fork, so it is opened as a draft against `main` and its diff shows all three. **Review commits `0c18413a` and `6e923dd2` only** — everything before them belongs to the parents. It comes off draft when #1159 merges.

**Scope, and why the other two consumers are not here.** §4 names three scripts to migrate. `audit_report.py` has twenty `gh` call sites spanning labels, issues, and pull requests, which is its own change rather than a paragraph of this one. `resolver.py`'s runner is being rewritten by #1153 (kirill-kosov), so migrating it now would be two people editing the same function. This PR touches neither file, and touches only the protocol surface and the new methods of `forge.py` — not the `run_gh`/`_call` internals #1153 is changing. The design doc and §9's step list are updated to say which third of step 3 has landed.

**Overlap to be aware of, not a blocker.** #813 (Fuxiao-Gao) normalises escaped newlines in `handle_submit` before `create_pull_request` is called, and #565 (fkc1e100) appends metrics to the PR description. Both are upstream of the seam this PR moves, so neither conflicts semantically; #813 is already marked conflicting against `main` independently of this branch.

## What changed

`ForgeProvider` gains three methods — `create_pull_request`, `update_pull_request`, `pull_request_url` — and `GitHubProvider` implements them. The protocol grew by exactly what one consumer needed, so nothing ships as dead code.

Three things about the implementation are worth a reviewer's attention:

- **`PullRequestExists` is a distinct exception type.** `gh pr create` fails when the branch already has an open pull request, and that refusal arrives *after* the push has landed. `submit-suggestion` treats it as the success case of a resubmission; another caller opening a change for the first time would read the same refusal as a genuine conflict. Recognising GitHub's phrasing belongs to the provider; deciding it is not a failure belongs to the caller, and the exception is the seam between them.
- **Bodies travel on stdin, never as a path.** Each method takes `body: str` and passes `--body-file -` with the text on fd 0. Since #913 the caller, the sandbox and the broker are three containers with three filesystems, so `--body-file <path>` names a file only the writer can open — and in content mode there is no workspace directory in this container to write it to at all. `credential_proxy_client.STDIN_FILE_FLAGS` is what forwards fd 0 across both hops when a flag in that set is followed by a bare `-`; `forge.BODY_STDIN` is the matching spelling.
- **The three calls pass no `cwd`.** They pass `-R <repo>`, so `gh` never consults the local clone, and `credential_proxy.git_lease_violation` returns immediately for any argv whose program is not `git` — the calls were never lease-gated.

## Testing

`agents/platform/scripts/` — the full suite passes on the rebased head (`ce94d6f7`), the one exception being `test_session_kv_server`, which fails to import for a missing dependency on `main` as well. `fleet-audit` 547, `pr-conversation` 82, `github-issue-resolver` 61, `tests/test_shipped_command_policy.py` 38 — all pass. `make docs-check` clean; `prettier@3.9.6 --check` clean on every changed Markdown file.

New coverage: `CreatePullRequestTest` (7), `UpdatePullRequestTest` (2), `PullRequestUrlTest` (3), `test_the_pull_request_body_travels_on_stdin`, and `ProtocolConformanceTest` extended to all ten methods in both directions.

### Live validation

**Not live-tested — there is no reachable kube-agents installation available to this session.** Nothing below substitutes for one, and the first finding under Self-Review is exactly the class of defect only a real install settles.

What was done instead: a differential harness that extracts the parent commit's tree with `git archive`, stubs the boundary each tree actually uses (BASE shells `gh` through `subprocess` inside the skill; HEAD goes through `forge`'s injected runner), and records every argv that would reach the credential proxy.

The happy path, unchanged in effect and changed in shape:

```
BASE  gh pr create --repo o/r --title 'chore: demo' --body 'BODY\nBODY\nBODY\n' --base main --head platform-agent/demo   (cwd=/tmp/pr4ws)
HEAD  gh pr create -R o/r --base main --head platform-agent/demo --title 'chore: demo' --body-file /tmp/pr4ws/tmpXXXX.md  (cwd=None)
```

The resubmission path, with `gh pr create` scripted to return the real "already exists" refusal — same three calls, same order, same returned URL, and no body file left behind:

```
BASE  pr create -> pr edit platform-agent/demo -> pr view platform-agent/demo    files left in workspace: []
HEAD  pr create -> pr edit platform-agent/demo -> pr view platform-agent/demo    files left in workspace: []
```

Those two harness runs predate the rebase and no longer reach the head: the `--body-file /tmp/.../tmpXXXX.md` they recorded is now `--body-file -` with the body on fd 0. The call sequence and the returned URL they establish are unchanged, and the argv shape they show is not.

What was re-run against the head instead: deleting the stdin forwarding and re-running the suite. Eleven tests fail and two error, and all pass again when it is restored — so the coverage catches a return to a path-based body rather than recording whichever form is in the tree.

## Self-Review

Both required pre-PR passes ran in subagent contexts that did not write the change, each given the commit range and nothing else.

**Adversarial pass** (`review-adversarial`) — seven findings, all confirmed, all fixed in `6e923dd2`:

1. **Blocker: the body file was 0600 and the sidecar's `gh` could not read it.** Fixed at the time by extracting `forge.body_file` and staging 0664. **Superseded by the rebase** — see the rebase findings below; there is no body file any more.
2. **A post-push refusal was reported as `REPO_UNREACHABLE`, truncated to 200 characters of stderr, with stdout and the exit code dropped.** Reachability is the one explanation already ruled out by the push that just succeeded. Now `PULL_REQUEST_REFUSED` carrying the exit code and both streams, with a handler in `main()` so the diagnosis is not one line.
3. **A third copy of the body-staging helper, and the copy that dropped the safety bit.** Resolved by (1)'s extraction; also superseded — the helper is gone and the two older copies had already moved to stdin on `main`.
4. **`create_pull_request` bypassed `_call` while the module docstring claimed nothing does.** A provider replacing `_call` to reach a `/v1/<forge>/…` route would have inherited a method still shelling `gh`. Now `_call(check=False)`, with a test that a subclass overriding `_call` sees the call.
5. **`pull_request_url`'s documented `""` return was unreachable, and a failed read-back turned a successful create into a reported failure.** Docstring corrected; the fallback catches and returns `""`, and the caller raises with a message that says the change landed and not to resubmit.
6. **Magic constant `200`, twice.** Named `_DETAIL_CHARS`.
7. **A comment describing code the commit deleted** (`tests/test_shipped_command_policy.py`). Corrected.

The pass also recorded five things it checked and did not find fault with: dropping `cwd`, the lease check, the command-policy match text, `gh pr create` outside a clone, and `import forge` resolving in the pod. It did not read #1153's full diff, only where it hooks the runner.

**Rebase onto current `main`** (2026-09-08, `b1658753` → `33ccddde`, folding in the earlier 2026-09-07 round). The first round hit `main` moving past this branch's central assumption while it sat, and its conflict resolution is a substantive API change rather than a mechanical merge:

- **The `body_file: str` provider API was wrong against post-#913 `main`.** `forge.post_comment`'s own docstring on `main` is the statement of it: with the caller, the sandbox and the broker in three containers, "there is no path all three processes can see". `main` had already moved its own direct `gh` calls to `--body-file -`. This branch would have reintroduced a written temp file, and in content mode — where `workspace` is `None` — there would have been nowhere to write it. All the body-carrying provider methods now take `body: str` and send it on fd 0; `forge.body_file`, `_BODY_FILE_MODE`, `_BODY_FILE_SUFFIX` and `BodyFileTest` are deleted.
- **`test_the_pull_request_body_travels_on_stdin` on `main` unpacked `(argv, cwd)` tuples** the skill no longer records, having been written when the skill shelled `gh` itself. Rewritten against `forge.run_gh`.
- Coverage for the new shape was checked by reverting it, not assumed: 11 failures and 2 errors.
- The second round applied without conflict, which is not on its own evidence that nothing was dropped. #1200 touches two of this branch's files, so I diffed its changes against the rebased tree: the `validate_repo_org` call in `submit_suggestion.validate_repo` and its rewrite of `forge.py`'s repository-parser docstring both survive, the latter in the form #1150 gave it. Gates re-run on the new head.

**Docs-drift pass** (`review-docs-drift`) — ten blocking findings, all fixed: the `multi-forge-support.md` status banner, §1's consumer count, §2's "seven operations", §4's opener and its "three remaining", §9's step 3 marker, `gitops-workspace-leases.md`'s layout comment and its "`cwd` on every subprocess" sentence, and the `docs/README.md` map row that tracks the banner. Of five non-blocking findings, four were taken (the `pr-comment-conversation.md` protocol-subset caveat, its scope list, the base-branch wording in `declarative-workflow.md`). One was left: a line-number citation in that document's live-validation appendix, which is a record of what was observed at the time and not a claim about `main`.

---


fkc1e100:
Closing as stale. This forge provider PR has been open for 21 days with unresolved conflicts.

This is part of the multi-forge work. If still needed, please re-evaluate against current architecture.

PR #1164, opened 2026-09-02
feat(gate): red the suite when automatic selection drops every armed case

haoxuw:
## Summary

`bench-gate suite` now folds a red into its verdict when automatic task selection (`EVAL_TASK_SELECTION=auto`) dropped every case the gate arms on and nothing graded was admitted. It is the backstop for the selection floor introduced in #1112: with it, a selection bug can cost a red run that names its cause; without it, the same bug produces a green run in which no case could ever have blocked the merge.

## Why This Change

The gate deliberately downgrades "`BOOTSTRAP_ADMITTED` names no graded case" to a warning, because the name may belong to a case a human commented out. Automatic selection breaks that reasoning: the filter is a program running on every pull request, and the escape hatch would fire routinely on a green banner nobody reads — precisely on eval-case-only PRs, where the gate matters most. This was found by `kube-agents-bot`'s review of #1112; the floor fix landed there, and this is the layer that survives the transition to store-based admission (the selector runs pre-install and can never read the store; the gate already computes admission from whichever source is live).

## What Changed

- `bench/kube_agents_bench/gate.py` — in `_cmd_suite`, when `EVAL_TASK_SELECTION=auto`, an armed case went ungraded (`unknown`), and zero graded cases are admitted, the verdict itself turns red: exit code, markdown, and the JSON artifact agree. The trigger is the *ungraded armed case*, not "zero admitted" alone — a store outage or stale evidence de-admits everything with the floor intact, and those keep degrading to advisory per `_load_store`'s own rationale. Outside `auto`, warning-only behaviour is unchanged.
- `bench/tests/test_gate.py` — four new tests: the drop reds (including the JSON artifact asserting `green: false`), an admitted case grades normally, the store-outage shape (zero admitted, armed case graded) stays advisory, and shadow/unset keeps the warning. `EVAL_TASK_SELECTION` joins the autouse env-cleaning fixture.
- `docs/designs/eval-scorer.md` (Bootstrap section), `bench/README.md` (gate env-variable inventory), `bench/baselines/README.md` (the "empty store is a green" claim gains its one exception) — prose kept in step.

## Context

Depends on #1112 (the selector, the floor, and the `EVAL_ADMITTED_CASES` wiring this backstops) — merge after it, and before any Prow-side flip to `EVAL_TASK_SELECTION=auto`. #1112 also carries the `export EVAL_TASK_SELECTION` that makes the variable visible to `bench-gate` as a child process. Sequencing: #1112 (shadow) → this → the flip.

## Testing

- `bench/tests/test_gate.py`: 64 passed (60 existing + 4 new) via `uv run --with pytest python -m pytest tests/test_gate.py` in `bench/`. Coverage on `kube_agents_bench/gate.py`: 96% lines.
- The red-then-green direction was verified by the adversarial pass executing the new tests against the merge base in a throwaway worktree: the drop test fails there (exit 0 pre-change) and passes with the change; the no-misfire tests pass on base too, pinning pre-existing behaviour.
- `make docs-check` clean; pinned `prettier@3.9.6 --check` clean on the three changed Markdown files.

### Live validation

Not live-tested against a running installation: the change is a verdict rule inside `bench-gate suite`, and the red path is unreachable in any live run today — nothing sets `EVAL_TASK_SELECTION=auto` until the Prow-side flip this PR is a prerequisite for. The unit tests exercise the full CLI surface (exit code, stderr, markdown, JSON artifact) that a Prow run consumes.

## Self-Review

Both passes ran as subagents handed the diff range (`refs/kube-agents-base/main...HEAD`), the skill path, and the gate output, instructed to derive intent from the diff and read no commit messages. `make docs-check`, pinned prettier, and the bench suite ran in the main loop. Merged dispositions:

1. **MEDIUM, CONFIRMED (adversarial) — the original trigger ("zero admitted" alone) also fired on store outages and stale evidence, states the gate deliberately keeps advisory, with a misdiagnosing error message.** Fixed: the trigger now additionally requires an armed case selection failed to grade, and a new test pins the store-outage shape as staying green.
2. **MEDIUM, CONFIRMED (adversarial) — the override changed only the exit code; the markdown said GREEN and the JSON artifact recorded `green: true` on a failed job.** Fixed at the altitude the finding suggested: the red is folded into the `SuiteVerdict` itself, so all three surfaces agree; the JSON is now asserted in the test.
3. **MEDIUM, CONFIRMED (adversarial) / Blocking (docs-drift) — the doc paragraph and error text referenced `hack/eval_triggers.yaml` and `EVAL_ADMITTED_CASES`, which exist only in open PR #1112.** Addressed as a stated sequencing dependency (Context above): this PR is coherent only merged after #1112, and the reviewer request is to hold it until then. The error text now points at `EVAL_ADMITTED_CASES in hack/ci-eval-pr.sh` rather than the yaml file.
4. **MEDIUM, PLAUSIBLE (adversarial) — nothing exported `EVAL_TASK_SELECTION` into `bench-gate`'s environment, so a script-default flip to `auto` would prune the matrix with the backstop dark.** Fixed in #1112 (an `export` after the mode validation), where the variable lives; the two PRs now agree on the transport.
5. **LOW, CONFIRMED (both passes) — an unrelated formatting hunk in `eval-scorer.md`.** Fixed: caused by running `prettier@3.6.2` instead of the pinned `3.9.6`; the file was reset to base and only the intended paragraph re-applied (branch diff is 11 added lines).
6. **LOW, CONFIRMED (adversarial) — `bench/README.md`'s gate environment-variable inventory did not gain the new variable.** Fixed.
7. **Advisory (docs-drift) — the unconditional "empty store is a legitimate green" claims in `bench/baselines/README.md` and the `ci-eval-pr.sh` comment gain an undocumented exception.** Fixed in `bench/baselines/README.md` with a one-line cross-reference; the `ci-eval-pr.sh` comment belongs to #1112's file and is accurate until the flip.
8. **Advisory (docs-drift) — `docs/designs/testing-strategy.md` has no task-selection concept.** Deferred with reason: selection ships shadow-only in #1112; the strategy doc gets its §4.2 line when the flip to `auto` is proposed, which is also when this backstop becomes load-bearing.

Angles run clean (adversarial): line-by-line logic, removed-behaviour/CI-weakening audit, operations/security surface, reuse/simplification. Not covered: no live Prow run; the oss-test-infra job config (other repository) is where `auto` would be set.

## Risk & Rollout

Unreachable until armed: no environment sets `EVAL_TASK_SELECTION=auto` today, so merging this changes no run's verdict. When armed, the new failure mode is a red naming the dropped cases — the intended behaviour — and the no-misfire tests pin that outages and stale evidence keep their advisory degradation. Revert is a clean `git revert`.

---

Generated with the help of Claude Code (Fable 5).

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -60,7 +60,7 @@ That evidence is what `baselines/` holds, and admission is computed from it rath
 
 Two speeds, deliberately: the deterministic `Verification*` scores decide whether a repetition passed, and no judged score can fail a repetition on its own. The captured fixtures are the argument — three byte-identical failing runs scored `OutcomeValidity` 0.9, 1.0 and 0.2 while `VerificationCorrectness` held at 0.5 on all three. The judge is given exactly one job (rung 6): catching a **collapse** in judged quality against main's mean at the same version key, at a margin of two standard errors of that measured spread. At three repetitions it cannot see drift, and widening it is a matter of more repetitions or a less variable metric, not a smaller number.
 
-Thresholds are named environment variables, all with the documented default: `EVAL_REPETITIONS`, `DETERMINISTIC_CORRECTNESS_FLOOR`, `EVAL_ADMISSION_RATE`, `EVAL_ADMISSION_MIN_RUNS`, `EVAL_AGGREGATE_MARGIN`, `EVAL_JUDGED_MARGIN`, `EVAL_JUDGED_METRICS`, `EVAL_BASELINE_STORE`, `EVAL_BASELINE_MAX_OBJECTS`, and `BOOTSTRAP_ADMITTED` (the transition bridge — cases that keep blocking before any screening exists).
+Thresholds are named environment variables, all with the documented default: `EVAL_REPETITIONS`, `DETERMINISTIC_CORRECTNESS_FLOOR`, `EVAL_ADMISSION_RATE`, `EVAL_ADMISSION_MIN_RUNS`, `EVAL_AGGREGATE_MARGIN`, `EVAL_JUDGED_MARGIN`, `EVAL_JUDGED_METRICS`, `EVAL_BASELINE_STORE`, `EVAL_BASELINE_MAX_OBJECTS`, and `BOOTSTRAP_ADMITTED` (the transition bridge — cases that keep blocking before any screening exists). `EVAL_TASK_SELECTION` is read too: when it is `auto` and selection dropped every armed case, the suite verdict reds instead of warning.
 
 ## Portal CUJ evaluations
 
--- a/bench/baselines/README.md
+++ b/bench/baselines/README.md
@@ -95,7 +95,9 @@ edit away from being gone.
 
 A store that will not parse is an **error**, never an empty store. Empty means
 "nothing admitted, aggregate advisory", which is a green; a corrupt file
-reaching that state would silently disarm the gate.
+reaching that state would silently disarm the gate. (One exception to that
+green: under `EVAL_TASK_SELECTION=auto`, a run that also failed to grade an
+armed case reds — see the Bootstrap section of `docs/designs/eval-scorer.md`.)
 
 A leftover `<case-id>.json` from the pre-JSONL format is an **error**, not a
 file to skip: skipping it would read as "never screened" and silently de-admit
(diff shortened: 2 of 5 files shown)


haoxuw:
Folded into #1112 at the author's direction (commit 8298ed5 there, cherry-picked unchanged). Landing the backstop with the selector also resolves this PR's sequencing findings: the doc paragraph and error text now merge together with the files they name.

PR #1142, opened 2026-09-02
feat(eval-dashboard): collect per-rep detail and merged-PR state

jayantid:
## Summary

The eval-dashboard collector keeps one result per task per run and discards the per-repetition grading the smoke test has printed since the multi-repetition eval landed. This PR teaches `collect.py` the repetition-era log format: each task gains an optional `reps` list (`{n, result, reason}` per graded repetition), each run gains an optional `pr_merged` flag resolved best-effort through `gh`, and the `[UNSTABLE]` task verdict the rep era introduced now parses instead of silently dropping the task. `schema_version` stays 1 — both fields are additive and documented in SCHEMA.md with explicit omission semantics.

## Why This Change

This is PR A of the dashboard redesign (lane #1022): the renderer needs to distinguish a flaky case (2/3 reps) from a hard fail (0/3) from an infra-excluded rep, and needs a merged-PR cohort split so main-adjacent runs can be read separately from arbitrary presubmit churn. Neither is derivable from today's one-result-per-task schema. There is also an active data loss: rep-era logs grade most non-green cases `[UNSTABLE]`, which the current `_TASK_LINE` regex does not match, so those tasks vanish from `tasks[]` entirely — every rep-era dashboard tick has been under-reporting exactly the cases worth looking at.

## What Changed

- `parse_build_log` reads the per-repetition grading block (identical in serial and parallel fan-out logs): `rep N: <verdict> -- <text>` lines attach to the `Task <name> Result:` line they follow, and a `>>>` section header closes the block so a stray rep-shaped line cannot attach across tasks. Rep verdicts map `pass`/`infra` verbatim and everything else (`fail`, `blocked`, unknown tokens) to `fail`; a rep carrying the literal `KUBE_AGENTS_INFRA_FAILURE` marker is `infra` regardless. `reason` is the text after the **first** ` -- ` (fail reasons contain the delimiter themselves), with the trailing `[OutcomeScore=…]` metrics dump stripped, capped at 300 chars, `null` for passes. Tasks with no grading lines get **no** `reps` key — absence means unknown, never fabricated.
- `[UNSTABLE]` joins `_TASK_LINE` and grades as `fail` (passed some but not all reps is not a clean pass; `reps` carries the split, and the schema's `pass|fail|infra` vocabulary stays closed).
[...]

PR #1165, opened 2026-09-02
fix(dashboard): the domains tile said "blocking" while counting "in presubmit"

jayantid:
## Summary

The "Domains covered" tile's detail line read **"28 scenarios · 19 blocking"** on the live dashboard, but only the 13 cases in `BOOTSTRAP_ADMITTED` can red a PR. `cases[].active` means "named in the TASKS array" — runs in presubmit — and the renderer was labeling that count *blocking*, hold-outs included. This changes the label to what the number actually is: **"in presubmit"**.

## Why This Change

The gate went merge-blocking today; the first question anyone asks the dashboard is "what can red my PR?" and the tile answered it wrong by 6 cases (including all three documented hold-outs). Spotted by Jayanti reading the live page against the roster.

## What Changed

- `scripts/eval_dashboard/render.py` and `scripts/eval_dashboard/template/index.html.tmpl`: label + variable rename, with a comment stating the active-vs-admitted distinction. Both sides changed per the Python/JS parity contract.
- `scripts/test_eval_dashboard.py`: the two assertions on the detail string updated.

## Context

Proper fix is a `cases[].admitted` field (collector parses the `BOOTSTRAP_ADMITTED` default with the same whitespace-or-comma contract `gate.py` uses) so the tile can say "13 blocking · 20 in presubmit" — deferred because `collect.py` is contested by #1142's in-flight conflict resolution; will follow it.

## Testing

- `python3 -m unittest discover -s scripts` — full suite green (dashboard files: 89).

### Live validation

Not live-tested: label-only change; the next dashboard refresh publishes it. The wrong label is live right now, which is the urgency.

## Self-Review

Hostile re-read: checked every other occurrence of "blocking" in the renderer — the page intro ("judged scores are recorded, never blocking") is accurate and untouched; no other tile derives from `active`. JS and Python outputs verified string-identical for the changed line via the existing parity tests.

## Risk & Rollout

Two-word label change; revert is trivial.

### DIFF
--- a/scripts/eval_dashboard/render.py
+++ b/scripts/eval_dashboard/render.py
@@ -646,11 +646,15 @@ def agent_tiles_html(data: dict, events: dict) -> str:
             else delta_chip("flat", f"{total - covered} open")
         )
         cases = [c for c in data.get("cases") or [] if isinstance(c, dict)]
-        blocking = sum(1 for c in cases if c.get("active"))
+        # ``active`` means "runs in presubmit" (named in the TASKS array), NOT
+        # "can red a PR" -- admission is BOOTSTRAP_ADMITTED's roster, which the
+        # collector does not read yet. Until a cases[].admitted field exists,
+        # calling this count "blocking" overstates the gate.
+        in_presubmit = sum(1 for c in cases if c.get("active"))
         if uncovered:
             detail = f"uncovered: {', '.join(uncovered)}"
         else:
-            detail = f"{len(cases)} scenarios · {blocking} blocking"
+            detail = f"{len(cases)} scenarios · {in_presubmit} in presubmit"
         tiles.append(tile("Domains covered", f"{covered}<small>/ {total}</small>", chip, detail))
     else:
         tiles.append(tile("Domains covered", "—", "", "not reported"))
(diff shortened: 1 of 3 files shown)


jayantid:
**Smoke red triage (gate-watch, self-note):** dashboard-only diff, red is environmental: `compliance-rbac-overgrant` rung-4 collapse (#1171 — no ledger published; proxy fix #1183 in flight, demotion #1173 pending) plus Vertex 429s (#1097) on the NOT_A_REAL_RUN reps in [build 2095156291314913280](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1165/pull-kube-agents-smoke-test/2095156291314913280). Will `/retest` once #1183 or #1173 lands.

jayantid:
Red-sweep classification: false red — 100% admitted-case pass rate; red solely on compliance-rbac-overgrant rung 2, where the verifier check errored rather than ran ('an-rbac-overgrant-finding-was-filed', VerificationCoverage=0.75). That case is the known degraded fleet-audits canary (#1171), and a dashboard-label diff cannot affect it. Retesting.

/retest

PR #1135, opened 2026-09-01
feat(cron): enforce per-job risk tiers and harden cron approval gates (#993)

shalinibhatiapl:
## Summary

Enforces declared per-job risk tiers (`low`, `high`) across Platform and Planning Agent cron rosters and wires autonomous security gates into tool execution. Autonomous cron runs unconditionally refuse `execute_code`, block raw terminal escape and C0/C1 control sequence injection, reject lookalike domain evasions, and enforce a fail-closed segment-aware read-only command policy allowlist (`cron_command_policy_block`) for `risk: "high"` and unannotated dispatches, refusing state mutations while allowing inspection commands to execute. Legacy unannotated custom jobs on persistent volume storage are automatically backfilled with `low` risk during upgrade profile scaffolding to ensure uninterrupted Day-2 audit operations.

## Why This Change

Addresses Issue #993 (`THREAT-002` and `SKILL-002`):
- Autonomous watchdogs running under ambient cron approval modes could execute unvetted shell scripts via `execute_code`.
- Command guard heuristics were vulnerable to raw terminal escape sequence injection (C0 and 8-bit C1 controls) that conceals payloads from pattern matching, as well as lookalike domain/TLD evasions targeting trusted infrastructure domains (`kubernetes.io`, `googleapis.com`, `github.com`, `x-k8s.io`).
- `risk: "high"` jobs previously escalated to `deny` approval mode, causing a complete execution blackout for high-risk jobs. This PR replaces `high -> deny` with a survivable segment-aware read-only command allowlist that refuses state mutations while permitting read inspection commands.
- Unconditional `THREAT-002` gates (`execute_code`, lookalikes, terminal escapes) apply to every cron run; `approvals.cron_scan: false` opts out of the Tirith AST scan only.
- Unannotated cron runs or unspecified dispatches default fail-closed to `risk="high"` (read-only policy) at runtime while preserving existing custom volume jobs on cluster upgrades.

## What Changed

- **Per-Job Risk Schema (`SKILL-002`) & Migration**:
  - Added explicit `"risk": "low" | "high"` metadata to all jobs in `agents/platform/cron/jobs.json` (including newly merged `findings-morning-nudge` annotated as `"risk": "low"`) and `agents/chat/defaults/cron/jobs.json`.
  - Added `check_cron_risk()` validation in `scripts/check_prompt_assets.py` to gate rosters at presubmit.
[...]

PR #1167, opened 2026-09-02
fix(slack): request the reactions:write scope the adapter already uses

bradhoekstra:
## Summary

`hermes slack manifest` generates the Slack app definition an installer pastes into the Slack console, and its `bot_scopes` list asks for `reactions:read` but not `reactions:write`. The adapter spends `reactions:write` on every turn — 👀 when it picks a message up, ✅ or ❌ when the turn ends — so Slack rejects each of those calls with `missing_scope` and no kube-agents install has ever shown a reaction. This adds the scope through the repository's normal image-patch route, gates it with a build-time check that reads the scope out of the manifest the CLI actually prints, and brings the two hand-maintained scope lists in `INSTALL.md` and the installer's Slack checklist back in step with it.

## Why This Change

The generated manifest is self-inconsistent rather than merely short: it already subscribes to `reaction_added` and `reaction_removed`, so the app it describes is built to hear reactions and not to make any.

Nothing surfaces the failure. `_add_reaction` swallows the rejection in an `except` whose own comment names this case — `# Don't log as error — may fail if already reacted or missing scope` — logs at `debug`, and neither caller reads the `False` it returns. The turn otherwise succeeds, so a dead feature is indistinguishable from a slow agent. `SLACK_REACTIONS` defaults to `"true"`, which means the feature is enabled on every install and works on none of them.

## What Changed

- `deploy/docker/patches/apply_slack_reactions_scope.py` — inserts `reactions:write` into `bot_scopes`. The list is located with `find_assign` rather than pinned by an anchor spelling out its contents, because upstream edits it: `reactions:read` itself is absent in v2026.7.20 and present from v2026.8.3, and a contents anchor would fail the build on any unrelated scope change. What is asserted instead is what the patch reasons about — that `bot_scopes` is still a list literal and still carries `reactions:read`. If upstream drops the read scope it has stopped supporting reactions, and this should fail loudly rather than quietly grant write to an app that no longer reacts.
- `deploy/docker/patches/verify_slack_reactions_scope.py` — the build-time gate. A grep for the inserted line would not establish the claim: `_build_full_manifest` appends to `bot_scopes` on two of its three `messaging_experience` branches and sorts before emitting, so the element being in the source says nothing about the JSON an operator pastes. The gate calls the real function once per branch and reads the scope out of what it returns. It also fails the other way — if the adapter stops calling the reaction helpers, the patch is widening the app's permissions for nothing and should be deleted.
- `deploy/docker/patches/test_slack_reactions_scope.py` — 8 host tests, mostly the drift cases a healthy image never reaches: the list renamed, `reactions:read` dropped, the list collapsed onto one line, `bot_scopes` not a list literal. Each asserts nothing is written on refusal.
- `deploy/docker/Dockerfile` — rides the existing `message-surface` `COPY`/`RUN` group, so it adds no image layer.
- `INSTALL.md` and `scripts/installer/print_instructions_slack.sh` — both hand-maintain a bot-scope list for operators who skip the *optional* manifest step. Both gain `reactions:write` and a note on how it fails, alongside the existing `files:write` one.
- `docs/README.md` — an identifier-source row for the Slack scope list. Its absence is what let one prose copy drift from the other.

## Context

No open issue or PR covers this. #1135 and #1136 also touch `deploy/docker/Dockerfile` and `deploy/docker/patches/`, in unrelated regions — merge-conflict risk only, nothing shared.

Two adjacent things this change deliberately does not do:

- **Upstream is not notified.** `deploy/docker/patches/` is this repository's normal route for a Hermes fix, per the existing `apply_slack_code_emphasis.py`.
- **No off-switch is documented.** `SLACK_REACTIONS=false` disables the feature inside Hermes, but there is no supported operator-facing route to set it: `safeSandboxEnvOverrides` is an allowlist that does not include it (and strips `ValueFrom`), the operator hard-codes `Platforms.Slack.Extra = {"rich_blocks": true}`, and the chart's `extraEnv` belongs to the controller manager. An earlier draft of this PR documented a route; it did not work, and documenting a fake off-switch in a paragraph about silent failure seemed worse than documenting none. Giving operators a real one is follow-up work, not this change.

Rebased across #1130, which moved `k8s-operator/scripts/` to `scripts/installer/`.

## Testing

- 943 tests in `deploy/docker/patches/`, `OK (skipped=1)`.
- `make docs-check` clean.
- `prettier@3.9.6 --check` (the repo's pinned version) clean on the changed Markdown.
- `bash -n` and `shellcheck` on `print_instructions_slack.sh` — only the pre-existing SC1091 about sourcing `common.sh`.
- Full agent image build succeeds. `scripts/check_image_layers.py`: `115 layers, 5 under the 120 budget` — the change adds none.
- Applier run against both live base tags, v2026.8.13 (main's pin) and v2026.8.19 (#1132's), so it lands whichever merges first. Verifier green after, and a second run refused with the first run's result intact.

### Live validation

Exercised against the `bhoekstra-gkedemos` install — cluster `kube-agents-cluster` (regional `us-central1`), namespace `kubeagents-system`, operator `kubeagents-controller-manager`, `PlatformAgent/platform-agent`. The agent image is pinned by the CR's `spec.deployment.tag`, which is the lever used below. The shared-install lease was held for the duration (`live_test_lease` → `HELD … PR #1167`) and released at the end.

**1. The patched image, running on the install, generates the right manifest.** Pinned the CR to `platform-agent:reactions-scope3` — the image built from this branch's head — and waited out the `Recreate` rollout.

```
CR .status.phase                      Ready
pod platform-agent-gateway-fb78cbc99-nhjpr   3/3 Running
  image  …/platform-agent:reactions-scope3

# the patched source, inside the running pod
$ kubectl exec … -- grep -n -A1 '"reactions:read"' /opt/hermes/hermes_cli/slack_cli.py
94:        "reactions:read",
95-        "reactions:write",

# the Slack relay came up on it
[Slack] Socket Mode connected (1 workspace(s))
gateway.run: ✓ slack connected
```

`hermes slack manifest`, run inside that pod, emits **18 bot scopes including `reactions:write`** — and they are an exact set match for the manifest the live Slack app is running, nothing only-in-pod and nothing only-in-applied. The manifest also subscribes `reaction_added`/`reaction_removed`, so the self-inconsistency described above is resolved: the app can now both hear and make reactions.

**2. The mechanism, not a coincidence.** Reverted the CR to its original tag and ran the same command on the same install:

```
unpatched pod (dns-endpoint-7dac349): 17 scopes, reactions:write present: False
patched   pod (reactions-scope3)    : 18 scopes, reactions:write present: True

delta: ['reactions:write'] added; [] removed
```

Same cluster, same operator, same CR — one scope of difference, in the direction claimed.

**3. The scope is what was actually broken, proven with the image held constant.** Separately and before the above, the Slack app's manifest was updated and the app reinstalled while the install ran the *unpatched* `dns-endpoint-7dac349` image throughout. `apps.manifest.update` returned `ok` / `permissions_updated: True`; `auth.test`'s `X-OAuth-Scopes` header went from 17 scopes to 18, gaining `reactions:write`. The bot token string was unchanged — same fingerprint before and after — because with `token_rotation_enabled: false` Slack widens the existing token in place, so no Secret edit and no restart were needed and the running pod's in-memory token became the wider one. A real Slack DM to the bot then showed 👀 on pickup and ✅ on completion, where no reaction had ever appeared before.

**4. The build-time gate was tested red as well as green**, inside the pinned base image `nousresearch/hermes-agent:v2026.8.19@sha256:3811ed13…`:

```
GREEN:
slack_reactions_scope verify: reactions:write present in all 3 emitted manifests;
  adapter still calls _add_reaction, _remove_reaction

RED (lifecycle call sites deleted, helpers left defined):
slack_reactions_scope verify: plugins/platforms/slack/adapter.py defines but no
  longer calls _add_reaction, _remove_reaction, so reactions:write is being
  granted for nothing — drop this patch instead of widening the app's permissions
```

**Not covered.** The 👀/✅ observation in step 3 is a person's direct report, not a log line: the gateway pod was replaced mid-test by an unrelated deploy, which took its logs with it, and reactions log nothing on success in any case — `_add_reaction` swallows failures at `debug`. Steps 1 and 2 were not re-run through a live Slack message, because the manifest the patched pod emits is provably the one Slack is already running, so a second message would exercise the same granted scope.

**Cleanup.** The CR is back on `dns-endpoint-7dac349`, pod `platform-agent-gateway-f9db8f877-w28fg` 3/3 Running, `.status.phase` `Ready`; the lease is released. The `reactions-scope3` tag remains in the project's Artifact Registry as a build artifact. The Slack scope grant is left in place deliberately — it is the fix, and that workspace is a permanent configuration rather than a test artifact.

## Self-Review

Both required passes ran via `/pr-preflight`, each in its own subagent handed only the diff range and its skill — neither held the context that wrote the change. `make docs-check` ran in the main loop. The range was `refs/kube-agents-base/main...HEAD` against a freshly fetched base.

The drift check before the passes caught the most consequential problem: `main` had moved 21 commits and #1130 renamed `k8s-operator/scripts/` to `scripts/installer/`, so the installer edit targeted a path that no longer existed. Rebased before reviewing.

`review-adversarial` worked its full angle set and separately confirmed the credential relay is not the cause — kube-agents routes every Slack call through `RemoteSlackClient`, and neither the shim nor the proxy allowlists method names, so the scope really is the only thing in the way. `review-docs-drift` checked generated regions, map coverage, the identifier-source tables, and whether any doc prose describes reactions.

| Finding | Pass | Disposition |
| --- | --- | --- |
| `INSTALL.md` still listed the old 8 scopes — the canonical install procedure, which two other docs defer to, and step 4's `--slashes-only` emits no `oauth_config` at all | both (HIGH/CONFIRMED, Blocking) | Fixed |
| Docstring claimed `files:write` arrived in v2026.8.19 — false; it is present in every pinned tag back to v2026.7.7.2 | docs-drift (Blocking) | Fixed. Verified against five pinned digests; replaced with `reactions:read`, which makes the same point and is true |
| Verifier's comment claimed path-loading insulates it from an unimportable package `__init__` — it does not, `_build_full_manifest` imports `hermes_cli.commands` in its body | adversarial (MEDIUM/CONFIRMED) | Fixed — comment now states what path-loading actually buys |
| The "drop this patch" guard matched `.reactions_add(`, which occurs only inside the private helpers. Upstream deleting the lifecycle hooks — the realistic way this feature dies — would leave the helpers defined and the gate passing | adversarial (MEDIUM/CONFIRMED) | Fixed — anchors on the `self._add_reaction(` call sites. Red/green transcript above; the old anchors passed this case |
| `SLACK_REACTIONS=false` named as an off-switch with no documented route | both (LOW/CONFIRMED, Advisory) | Fixed by deletion. The two passes proposed different routes and, on reading the source, neither works — see Context. Removed the sentence rather than ship a fake off-switch |
| Verifier's `main()` ignored `argv` while both siblings in the same `COPY` accept a root | adversarial (LOW/CONFIRMED) | Fixed |
| `INSTALL.md` twice told the reader to run the printers from `k8s-operator/` — dead since #1130 | docs-drift (Advisory) | Fixed. Pre-existing, but 20 lines from an edit this PR had to make anyway |
| `docs/designs/pr-comment-conversation.md` hard-coded "nineteenth anchored substitution"; there are 21 | docs-drift (Advisory) | Fixed by de-counting. Already stale, and this PR nudges it further |
| No identifier-source row owned the Slack scope list, which is what let one copy drift | docs-drift (Advisory) | Fixed — one row added |
| If upstream adds `reactions:write` itself, `refuse_if_patched` fails with "already patched", sending the reader after a double-application instead of telling them to delete the patch | adversarial (LOW/PLAUSIBLE) | **Not actioned.** Loud rather than silent, the message lives in shared `patchlib.py`, and every sibling applier has the same shape — changing it here alone would make this patch inconsistent for a case that has not happened. Flagging rather than fixing |

**What the passes could not cover.** Neither reached a live install; that is the section above, and it was not done by them. Docs-drift's version history bottoms out at v2026.7.7.2, so "`files:write` predates v2026.8.19" is established but the release that first added it is not — the replacement docstring avoids the claim. `scripts/installer/print_instructions_slack.sh` has no test anywhere and CI shellchecks only `install.sh`/`uninstall.sh`/`upgrade.sh`, so the drift fixed here has no automated backstop in either direction.

## Risk & Rollout

Low. The runtime effect is one additional string in a generated JSON document; no Go code, no chart template, no operator behaviour changes, and no image layer is added. An install that never pastes the manifest is unaffected. The new permission is the narrowest one that covers the calls the adapter already makes, and it does not widen what the agent can read.

The build is where this fails if it fails: every anchor is asserted, and a base-image bump that moves `bot_scopes` fails the image build loudly rather than shipping a silently unpatched CLI. Revert is deleting the three patch files and their `COPY`/`RUN` lines.

Existing installs need the Slack app manifest re-pasted and the app reinstalled to pick the scope up — the image alone cannot grant it, since the grant lives in Slack. Nothing breaks if they do not; reactions simply keep not appearing.

---


dshnayder (approved):

bradhoekstra:
/hold

Temporary: holding my own PR to drop it from the pending 7-PR batch so Tide merges the overridden singles instead. No change requested. Will `/hold cancel` shortly.

PR #1039, opened 2026-08-28
feat(crd,operator): add Microsoft Teams integration to PlatformAgent CRD and controller

fkc1e100:
## Summary

Adds native Microsoft Teams chat integration to the `PlatformAgent` Custom Resource Definition and operator controller. Reconciles Teams bot credentials, single-tenant security controls, user access controls, and Bot Framework network egress policies into the Platform Agent deployment, managed configuration, and network policies.

## Why This Change

Enterprise environments using Microsoft Teams as their primary collaboration platform need ChatOps access to the Kubernetes Agentic Harness without requiring external gateway proxies or manual manifest editing.

## What Changed

- **CRD Specification (`k8s-operator/api/v1alpha1/platformagent_types.go`)**:
  - Added `Teams *TeamsSpec` to `PlatformAgentIntegrationSpec`.
  - Added `TeamsSpec` struct supporting `Enabled`, `AppIdSecretRef` (SecretKeySelector), `AppPasswordSecretRef` (SecretKeySelector), `TenantId` (string Entra tenant GUID), `AllowedUsers`, `AllowAllUsers`, `HomeChannel`, and `AdaptiveCards`.
  - Added CEL validation rule enforcing that `appIdSecretRef` and `appPasswordSecretRef` are configured when `teams.enabled == true`.
  - Generated CRD YAML manifests and DeepCopy methods.
- **Controller Manifest Synthesis (`k8s-operator/internal/controller/platformagent_manifests.go`)**:
  - Injects `TEAMS_APP_ID` and `TEAMS_APP_PASSWORD` secret references (`valueFrom.secretKeyRef`), and `TEAMS_TENANT_ID` string value, into the container and credential proxy sidecar.
  - Exports `TEAMS_RELAY_URL`, `TEAMS_ALLOWED_USERS`, `TEAMS_ALLOW_ALL_USERS`, and `TEAMS_HOME_CHANNEL` into container environment variables.
  - Generates `managed-config.yaml` with `platforms.teams` configuration including adaptive card support and bot typing indicators.
  - Synthesizes FQDN egress network policies permitting traffic to Microsoft Entra ID (`login.microsoftonline.com`) and Bot Framework (`*.botframework.com`).
- **Testing & Verification**:
[...]

PR #1169, opened 2026-09-02
fix(operator): permit DNS to the metadata address for Cloud DNS for GKE

### LINKED ISSUE #1140: fix(operator): gateway NetworkPolicy blocks port 53 to 169.254.169.254, breaking DNS on Cloud DNS for GKE clusters
## Summary

The operator-generated `<name>-gateway-netpol` permits port 53 only to the kube-dns/node-local-dns
pod selectors, `169.254.20.10/32`, and the discovered kube-dns Service ClusterIP. On a cluster
running **Cloud DNS for GKE**, none of those is the resolver: pods' `resolv.conf` points at
`169.254.169.254`, and the policy drops every query to it. The gateway Pod loses name resolution
outright.

Reported from a live install: the `envoy-credential-proxy` sidecar could not resolve
`googleapis.com` and failed to bootstrap. Deleting the NetworkPolicy restored it.

This is the same shape as bug 3 in #747 — a policy naming a DNS peer that is correct on the
author's cluster and wrong on the reporter's — one rung further out. #687 fixed the Service-VIP
case; the Cloud DNS case is still open.

## Where

`buildNetworkPolicy`, rule 1, at
[`k8s-operator/internal/controller/platformagent_manifests.go:4144-4196`](https://github.com/gke-labs/kube-agents/blob/main/k8s-operator/internal/controller/platformagent_manifests.go#L4144-L4196):

```go
dnsPeers := []networkingv1.NetworkPolicyPeer{
    // kube-system/k8s-app=kube-dns
    // kube-system/k8s-app=node-local-dns
    {IPBlock: &networkingv1.IPBlock{CIDR: "169.254.20.10/32"}},
}
// + formatCIDRPeers(profile.DNSClusterIPs)  -> discovered kube-dns ClusterIP, or 10.96.0.10
```

Rule 2 below it already permits `169.254.169.254/32`, but only on TCP 80. Nothing in the policy
permits 53 — UDP or TCP — to that address.

The DNS ladder in
[`netpolprofile.go:129-157`](https://github.com/gke-labs/kube-agents/blob/main/k8s-operator/internal/controller/netpolprofile.go#L129-L157)
cannot rescue this: it discovers the `kube-system/kube-dns` Service ClusterIP, and falls back to
`10.96.0.10`. Under Cloud DNS neither address answers, so discovery succeeding is not the same as
discovery being right.

## The static copies already allow it; the operator-rendered ones do not

Parsing every `kind: NetworkPolicy` in the tree for its port-53 egress peers:

| Policy | Source | `169.254.169.254` on 53 |
| --- | --- | --- |
| `litellm-policy` | `charts/kube-agents/templates/litellm.yaml` | yes |
| `github-token-minter-policy` | `charts/kube-agents/templates/github-minter.yaml` | yes |
| `litellm-policy` | `k8s-operator/config/integrations/litellm/base/networkpolicy.yaml` | yes |
| `github-token-minter-policy` | `k8s-operator/config/integrations/github/deployment.yaml.template` | yes |
| `litellm-policy` / `vllm-policy` | the three `examples/*/networkpolicy.yaml` | yes |
| `platform-agent-core-egress` | `deploy/kustomize/platform/networkpolicy-core-egress.yaml` | **no** — the `0.0.0.0/0` peer excepts `169.254.0.0/16` |
| `<name>-gateway-netpol` | `buildNetworkPolicy` | **no** |
| `<name>-sandbox-metadata-deny` | `buildAgentEgressPolicy` | **no**, and by design (below) |

So a Cloud DNS install has a working LiteLLM and a working minter beside a gateway Pod that cannot
resolve. That asymmetry is the argument that the gateway policy is the outlier rather than the
static copies being too loose.

## The credential-isolation policy makes this harder than a one-line fix

`buildAgentEgressPolicy` renders `<name>-sandbox-metadata-deny` when
`spec.security.egressPolicy: Allowlist`, and its one stated invariant is that no rendered rule
permits a metadata address. It enforces that on the DNS rule specifically —
[`platformagent_egress_policy.go:245-251`](https://github.com/gke-labs/kube-agents/blob/main/k8s-operator/internal/controller/platformagent_egress_policy.go#L245-L251)
filters resolved DNS IPs through `isMetadataAddress` and drops any that match.

Under Cloud DNS that invariant and working DNS are mutually exclusive for the Pod that policy
selects. The two are not actually the same grant — DNS needs 53, the credential path needs 80/988 —
so the resolution is probably to allow `169.254.169.254/32` on 53 only, and to say in the file why
port-scoping is enough. But it is a deliberate design constraint, not an oversight, so it needs a
decision rather than a patch.

## Suggested fix

1. Add `169.254.169.254/32` (and `fd20:ce::254/128` for dual-stack) to the port-53 peers in
   `buildNetworkPolicy`, unconditionally, alongside the existing `169.254.20.10/32`.
2. Decide the `sandbox-metadata-deny` question above; if the answer is yes, narrow
   `isMetadataAddress` filtering to the non-DNS rules rather than dropping the peer.
3. Add `169.254.169.254/32` to the port-53 rule in `networkpolicy-core-egress.yaml`, which today
   excepts the whole link-local range.
4. Extend the parity checker in #1090 to assert the metadata address is present on port 53, so the
   copies cannot drift apart again.

Alternative worth considering instead of (1): detect Cloud DNS at reconcile and add the peer only
then. Against it — the peer is harmless on a kube-dns cluster (nothing listens on 53 there), and
"detect the dataplane" is the machinery #747 Part A is trying to remove, not add to.

## Workaround

Better than deleting the policy, on installs not using `egressPolicy: Allowlist`:

```yaml
spec:
  networkPolicy:
    dnsClusterIPs: ["169.254.169.254"]
```

or the `kubeagents.x-k8s.io/dns-cluster-ip` annotation. `normalizeCIDRTarget` does not filter
link-local, so this renders `169.254.169.254/32` into the port-53 rule of the gateway policy. It
replaces the discovered ClusterIP rather than adding to it, which is correct under Cloud DNS but
means the value has to be revisited if the cluster ever moves back to kube-dns.

On an `Allowlist` install this escape hatch does nothing for `sandbox-metadata-deny` — the
`isMetadataAddress` filter strips it — so there deleting or disabling the policy is currently the
only option.

## Environment

GKE with Cloud DNS for GKE enabled. Please add the cluster's dataplane (V1/V2), GKE version, Cloud
DNS scope (cluster or VPC), and whether `spec.security.egressPolicy` is set, since the fix for
`sandbox-metadata-deny` depends on the last one.

## Related

- #747 — NetworkPolicy correctness debt; this is the same failure family (bug 3), one rung further out
- #687 — fixed the kube-dns Service-VIP case
- #1090 — static NetworkPolicy DNS egress parity checker, open

bradhoekstra:
## Summary

The operator's `NetworkPolicy` never granted the agent Pod port 53 to `169.254.169.254`, which is the resolver on a [Cloud DNS for GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-dns) cluster. Every Pod's `resolv.conf` names that address there, so the policy took the agent's name resolution away entirely: the `envoy-credential-proxy` sidecar could not resolve `googleapis.com` and never finished bootstrapping. This adds the peer, on port 53 alone, to the three policies that render a DNS rule over the agent Pod — the operator's gateway policy, the `egressPolicy: Allowlist` sandbox policy, and the static `platform-agent-core-egress` manifest — and narrows the machine-checked invariant from "no rendered rule names a metadata address" to "no rendered rule names one on a credential port".

## Why This Change

Reported from the field: on a Cloud DNS cluster the install comes up broken and the only workaround is deleting the NetworkPolicy, which removes the whole egress boundary to fix a DNS rule. Nothing detects the condition today, so the peer is granted unconditionally. That costs one port-53 rule on clusters not using Cloud DNS and is the difference between working and not on clusters that are.

Port 53 to that address is not a credential path. GKE's metadata interception rewrites TCP 80 and 8080; the token API is 80 pre-NAT and 988 post-NAT, and no rule permits this address on either. The `<name>-sandbox-metadata-deny` policy still withholds the credential API — that is what it exists for — and the tests now hold that property at port granularity rather than address granularity.

## What Changed

- `platformagent_manifests.go` and `platformagent_egress_policy.go` add the resolver to both builders' port-53 peer sets, and a `peersNotAlreadyPresent` helper stops the peer being rendered twice when an operator sets `spec.networkPolicy.dnsClusterIPs` to an address the rule already carries.
- `deploy/kustomize/platform/networkpolicy-core-egress.yaml` was the last static copy missing the peer; the chart's `litellm.yaml` and `github-minter.yaml` already had it. A new test compares the static file's port-53 `ipBlock` peers against the operator's in both directions, so the two cannot drift again.
- `docs/site/src/content/docs/deploy/kustomize.md` — the documented patch for a custom kube-dns ClusterIP addressed `/spec/egress/0/to/3`, which the new peer displaces. It now appends to `/spec/egress/0/to/-`. Left as-is, the documented patch would have overwritten the resolver peer with the user's ClusterIP and reintroduced this exact outage.
- The CRD description for `egressPolicy` and four documents said the metadata *server* is left off an Allowlist policy. It is the credential *API* that is left off; the address is on the DNS rule. Both CRD copies regenerated.
- `.agents/skills/review-security-k8s-agents-firewall/SKILL.md` told an auditor to flag `169.254.169.254` with no port carve-out, so the repository's own audit skill would have failed the operator's own policy.

## Context

Closes #1140.

Three open pull requests touch adjacent ground. None conflicts on content; two conflict on files:

- **#1090** (`feat(ci): assert static NetworkPolicy DNS egress parity across all copies`) adds a Python parity checker over the eight static DNS-bearing manifests. It passes against this branch — it asserts peer contents, not peer count. Its module docstring records the pre-change world, noting that `networkpolicy-core-egress.yaml` lacks the resolver; whichever merges second should reconcile that line. The two guards are complementary and leave a gap: neither checks the resolver across all eight static copies.
- **#913** (`feat(operator): run the agent shell in a sandbox pod`, draft) adds `buildShellSandboxNetworkPolicy`, whose DNS rule names only the `k8s-app: kube-dns` Pod selector. It reproduces this outage on a Cloud DNS cluster, in a function this branch does not touch. Merge order, not a conflict: whichever lands second ships the fix only for its own copy.
- **#1039** and **#913** both edit all six `internal/testing/testdata/platform/expected/*.yaml` goldens this branch edits, and #913 also edits `credential-isolation.md`.

## Testing

- `go build ./...`, `go vet ./internal/...`, and `go test ./...` in `k8s-operator/` — green.
- `make docs-check` (all five checks), `make chart-check`, and `make -C k8s-operator manifests` with no drift.
- `npx prettier@3.9.6 --check` over the same changed-file list `.github/workflows/prettier.yml` builds — all 16 `.md`/`.yaml` files pass. Five of those hunks are whitespace in `review-security-k8s-agents-firewall/SKILL.md`: the file's misformatting predates this branch, but the Prettier job only checks files a PR touches, so this is the first PR that would have gone red on it.
- Every new test was watched fail against the un-fixed code before being kept. Removing the peer from both builders fails seven tests; reverting the dedup fails two; removing the peer from the static manifest alone fails the parity test with the original bug's signature.
- Not run: the Python suites, `tests/integration/`, and `bench/` — no Python test reads these manifests.

### Live validation

Exercised on the `bhoekstra-gkedemos` install (cluster `kube-agents-cluster`, `us-central1`, namespace `kubeagents-system`, permissions `read-only`, model served via Vertex through LiteLLM), with the operator built from this branch and pushed as `clouddns-dns53` (digest `sha256:d2325104…`).

Two-way A/B on the rendered gateway policy, because a peer that was always going to be there proves nothing:

| Operator image | Port-53 peers on `platform-agent-gateway-netpol` | `resourceVersion` |
| --- | --- | --- |
| pre-change | 4, no metadata address | 1787154036729007015 |
| `clouddns-dns53` | 5, including `169.254.169.254/32` | 1788356908939999015 |
| reverted | 4 again | 1788357075489519015 |

The `resourceVersion` advancing at each transition is what rules out a coincidental read: the operator rewrote the object both times. The policy carries an `ownerReference` to the `PlatformAgent`, so it is the reconciled object and not a hand-applied one.

Nothing broke with the peer present: the gateway Pod's `resolv.conf` still names `34.118.224.10`, both external and in-cluster names resolved from inside the pod, and all 4 containers stayed Ready with 0 restarts. The `envoy-credential-proxy` sidecar — the component the report says fails — is in the Pod this policy selects, so the gateway fix addresses the reported symptom directly.

**Still current at the head.** The only production changes in the three commits after the A/B are the `peersNotAlreadyPresent` helper and the swap to the `metadataResolverCIDR` constant. The helper is a no-op unless `dnsClusterIPs` overlaps a fixed peer, and this install resolves `34.118.224.10`, which is not one; the constant has the same value, and all six goldens are untouched since the A/B commit, so the rendered object is byte-identical to the one tested.

**What this did not cover, and why.**

- **No Cloud DNS cluster.** gkedemos runs kube-dns. The premise that a Cloud DNS Pod's `resolv.conf` names `169.254.169.254` comes from GKE's documentation and from the field report this fixes, not from a cluster I ran. The same goes for the comments' claim that the address answers 53 on a GCE node with no metadata interception.
- **The `<name>-sandbox-metadata-deny` policy was not exercised.** `egressPolicy: Allowlist` is refused without `splitCredentialBrokerPod: true`, which restructures the Deployment and has a documented three-step revert order that strands the re-merged broker if interrupted. Not something to do to a shared install for this change. Its rendering is covered by unit tests and goldens only.
- **The Kustomize path was verified by rendering, not by installing.** `kubectl kustomize` over a throwaway overlay carrying the newly documented patch: the custom ClusterIP appends and the resolver peer survives, which the old `to/3` patch would not have done.

An unprefixed `manager-role` ClusterRole created while fixing unrelated RBAC drift on the test cluster has been deleted, and the throwaway kustomize overlay was removed. Nothing else was left behind.

Separately, and not caused by this change: the live install's `kubeagents-manager-role` had drifted three rules behind `main`, which aborted every reconcile at `reconcilePodDisruptionBudget` before it reached the NetworkPolicy. Re-applying the kustomize-rendered role fixed it. Worth knowing if another install shows a policy that will not update.

## Self-Review

Three rounds of `review-adversarial` and `review-docs-drift`, each pass in its own subagent handed the diff range and nothing else — no round reviewed a diff in the context that wrote it. Rounds two and three were re-runs after fixes, per `review-preflight` §7; the list below is the current state, not the history.

**Angles covered:** whether port 53 to a metadata address reaches a token on either dataplane; whether the relaxed test invariant leaves the other two metadata addresses unguarded; peer duplication and `append` aliasing; whether every static copy of the DNS rule was found; whether doc prose still matches source after the CRD description changed; positional patch paths in documented Kustomize overlays; generated-region and docs-map obligations; sibling-PR divergence; and the mechanical gates CI runs.

Fixed:

1. **Blocking** (docs-drift) — the documented Kustomize patch addressed `/spec/egress/0/to/3`, which the new peer displaced; it would have overwritten the resolver and reintroduced the outage. Now appends. Verified by rendering.
2. **HIGH** (adversarial, and independently docs-drift) — the Prettier job checks changed files, so touching the firewall skill would have gone red on its pre-existing misformatting. Formatted.
3. **MEDIUM** (adversarial) — `permitsBeyondDNS` decided the DNS-rule exemption on ports alone, so it dropped that rule from the assertion for *all three* metadata addresses. A later edit putting `169.254.169.252` or `fd20:ce::254` on the DNS rule would have passed every metadata test in the file. The exemption is now granted to the resolver address only; confirmed by planting the daemon address on the DNS rule and watching the assertion fail.
4. **MEDIUM** (adversarial) — the static YAML's comment said the peer is "inert on a kube-dns cluster, where nothing listens on 53", contradicting the Go comment added in the same change. The Go version is right; the YAML now matches it.
5. **MEDIUM** (adversarial) — the docs made only the affirmative half of the safety case. Where an administrator has removed CoreDNS's external forwarding to close DNS exfiltration, this peer reopens it. That case is now stated on the canonical page.
6. **LOW** (adversarial) — the dedup helper was applied to the gateway builder and not its sibling, which still rendered `169.254.20.10/32` twice given that input. Fixed, with a test that reproduced the duplicate first.
7. **LOW** (adversarial) — the parity test compared the operator's *first* port-53 rule against every one of the static file's. It now aggregates both sides.
8. **LOW** (adversarial) — `metadataResolverCIDR` was used by one of the two builders granting the peer, so a grep for it found half the grant sites. Both use it now; goldens unchanged, which proves the value is identical.
9. **LOW** (adversarial) — the IPv4-only rationale claimed a difference from `linkLocalPeers` that the code does not have. Reworded to say what is true: the expressions are identical today and kept separate so they can diverge.
10. **LOW** (adversarial) — a `dnsPort` constant shadowed a package-scope one of the same name, in a change about which port is which.
11. **LOW** (adversarial) — the repository's own firewall-audit skill would have flagged this policy. Carved out for port 53.
12. **Advisory** (docs-drift) — the "what the gateway policy already permits" enumeration in `security-requirements.md` missed the port-53 grant its two sibling sentences gained. That paragraph's argument is the exhaustive list, so incomplete is close to wrong.
13. **Advisory** (docs-drift) — `security-and-iam.md` said the resolver "is granted on port 53 alone" in a bullet that then lists the same address on TCP 80 and 988, and attributed a `0.0.0.0/0`-with-excepts peer to both policies when only the static manifest has it. Both corrected, and the Cloud DNS rationale now links to the canonical page rather than restating it.
14. **Advisory** (docs-drift, twice) — line-wrap damage in `credential-isolation-design.md` from my own edit.
15. **Note** (adversarial) — a test row labelled "a named port cannot be read as 53" supplied a nil port. Relabelled, and a real named-port case added.

Not fixed, with reasons:

16. **Advisory** (docs-drift) — the `## Denying the sandbox the metadata server` heading still says "metadata server" while its prose now says "credential API". Four documents anchor to it; renaming breaks them to fix a wording mismatch that misleads nobody.
17. **Advisory** (docs-drift) — the full port-53 argument lives on the site page while `credential-isolation-design.md`, which the site page names as canonical, gets the summary. That inversion predates this branch and fixing it means moving prose this change did not write.
18. **LOW** (adversarial) — #1090's docstring and #913's sandbox policy are sibling work, recorded under Context rather than changed here.

Rejected as factually wrong:

19. **Advisory** (docs-drift, round three) — claimed the chart's `litellm.yaml` and `github-minter.yaml` DNS rules carry only selector peers, leaving those Pods broken under Cloud DNS. They already carry `169.254.169.254/32` on the port-53 rule at `litellm.yaml:281` and `github-minter.yaml:224`. Round one's passes both got this right; round three's did not.

**What the passes could not cover:** neither touched a live cluster, so every claim about GKE runtime behaviour in this change is asserted rather than observed — see Live validation. Neither ran the Python or integration suites. #1090 and #913 were read from their diffs, not from checkouts.

## Risk & Rollout

Adds one peer to a DNS rule on three policies; removes nothing. NetworkPolicy is additive, so the worst case on a cluster not using Cloud DNS is one unused peer — with the caveat under Self-Review item 5, which is now documented. No API, CRD schema, or reconcile-loop change; the CRD diff is description text only. Reverting is reverting the commits: the operator rewrites the policy on the next reconcile, which the live A/B demonstrates in both directions.

Merge-order note for whoever lands this: #1039 and #913 both edit the same six goldens, and #1090's docstring wants a line reconciled.

<sub>Generated with the help of Claude Opus 5 in Claude Code.</sub>


jayantid:
**Smoke red triage (gate-watch):** not caused by this PR. Every gating rep in [build 2095170383677231104](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1169/pull-kube-agents-smoke-test/2095170383677231104) traces to Vertex 429s (`litellm.RateLimitError ... "code": 429`) — known class #1097; #1184 tracks why an infra-blocked rep grades as a case failure. Your diff (operator egress) is unrelated. `/retest` is reasonable — runs did go green at 17:17Z — but 429 saturation was still intermittent this afternoon.

dshnayder (approved):

Issue #1025, opened 2026-08-28
Lane: presubmit under 3 hours, 3-repetition tolerance intact


jayantid:
**Goal this week: presubmit p50 ≤ 3 hours at the grown suite (~25 cases), with 3-repetition flake tolerance fully intact.**

Rejected on the record (2026-08-31): cutting repetitions. The collapse rule (red only on 3-of-3 failure) IS the flake tolerance. The clock comes down by engineering.

## This week
- **Adaptive repetitions** — the key item: a pass on repetition 1 ends the case (a pass cannot collapse); a failure runs the remaining repetitions. Same verdicts, green path ~1.7–2h at 25 cases.
- Async judging (the judged score is advisory; it doesn't belong on the critical path).
- A task-selector env (`EVAL_TASK_FILTER` or similar) — needed by three things: shards, the release gate's full-matrix run, the dashboard's failures-only rerun.
- Explore, don't build: a sharding design note + a manual two-shard spike to price the overhead. Details and measured baselines in this issue's earlier comments.

## Constraints
- Repetitions triggered by a failure must not feed pass-rate statistics or the baseline store (rates come from the nightly).
- In-run parallel lanes stay gated on #988.

**Done when:** adaptive reps merged and measured under 3h p50; the selector env exists; the shard spike is priced.

Part of #1020 (goal 3) · #897.

jayantid:
Re-chartered for the week of Sep 1 (epic #1020): target is now **presubmit p50 ≤ 3 hours at ~25 cases with 3-repetition flake tolerance intact**. The reps-to-1 idea is dead (rejected 08-31 — the collapse rule IS the flake tolerance; oss-test-infra#2676 became the 360m deadline raise instead). The load-bearing item is **adaptive repetitions** in `hack/ci-eval-pr.sh`: pass on rep 1 → done (a pass cannot collapse); fail → reps 2–3 run — identical collapse semantics, green path ~1.7–2h at 25 cases. Second lever: async judging (advisory score off the critical path, ~35–60 min at the grown suite). Retried reps must never feed the rate rung or the baseline store (925-review F2 caveat).

jayantid:
Charter expanded (jayantid, 2026-08-31): the lane now carries **two tracks**.

[...]

Issue #1024, opened 2026-08-28
Lane: release gate runs the full suite — non-gating, reported


jayantid:
**Goal this week: every RC runs the full eval suite, non-gating, with results anyone can find.**

The shape: the nightly's configuration pointed at the RC image (one job, two invocations). Blocking comes later, when baselines can arm the comparison honestly — this week is run-and-report only.

## This week
- An RC cut triggers a full-suite run against the RC image (deploys the RC's built images; no rebuild).
- Results visible without asking anyone: an optional status on the RC, the verdict in run artifacts, and the dashboard's Releases section.
- The non-inferiority number renders as advisory with an honest "baselines maturing" label.

## Constraints
- Nothing this lane ships may block a release this week.
- Don't widen job credentials just to auto-post comments — a pasted artifact link is acceptable.

**Done when:** one real RC has been through it end to end, all reporting layers populated.

Part of #1020 (goal 4) · #897. Depends on #1021's job; shares the task-selector with #1025.

jayantid:
Re-chartered for the week of Sep 1 (epic #1020): upgraded from "scaffold, don't arm" to **the RC checklist actually executes the suite against the RC image**, with the non-inferiority comparison running **advisory** until the nightly (#1021) holds enough baselines to arm it honestly. Running ≠ blocking in week one of baselines — the checklist says which gates are live and which are pending, per the dashboard's empty-state design.

stalhaali:
Five things I want to settle before writing code, because each one changes the shape of the change. 

**1. Cadence.** The issue says "every RC," and the shape is "the nightly's configuration pointed at the RC image (one job, two invocations)." RC tags cut 4-6/day, so a nightly invocation measures roughly one in five. Which did you mean? "Done when: one real RC has been through it end to end" is satisfied either way, and nightly is ~1/5 the pool cost against a pool already at max_concurrency: 15. I'd default to nightly unless you want the per-tag trigger — that one needs a postsubmit on refs/tags/rc_*, which I haven't confirmed Prow supports here. 

**2. EVAL_TIER value.** #1175's switch takes `presubmit` or `nightly` and exits 1 on anything else, so `EVAL_TIER=rc` fails today. Does the RC lane export `nightly` and lean on RC_COMMIT_SHA to distinguish itself, or should #1175 grow a third value? Your call — it's your diff. 

[...]

PR #1170, opened 2026-09-02
feat(ci): let the eval pipeline install a release candidate's published images

stalhaali:
## Summary

Setting `RC_COMMIT_SHA` switches `hack/ci-deploy.sh` from building the pull request's images into a leased project's Artifact Registry to installing the images already published at that commit, and stops the eval run that follows from filing its numbers as `main`'s. A new `hack/resolve-rc-target.sh` names the candidate. Nothing sets `RC_COMMIT_SHA` yet, so on merge this changes no run: it is the first slice of #1024, and the Prow job that turns it on is the next one.

## Why This Change

Issue #1024 wants the full eval suite run against a release candidate. The suite already runs on presubmits, so the obvious move is to point that machinery at an RC — but two things stop it from working as-is.

The first is that `hack/ci-deploy.sh` builds what it installs. A rebuild from a candidate's source is not the candidate's artefact: different base-image digests, a different `_KUBE_AGENTS_VERSION` baked in as the remote-MCP User-Agent. An eval that graded a rebuild would not be grading what the release ships, which is the whole point of the exercise.

The second is worse, and it is why the baseline guards are in this PR rather than a later one. An RC eval is a Prow periodic sitting on a commit that is on `main`, with no `PULL_NUMBER` — which satisfies both of the recorder's existing conditions exactly, so it would append the candidate's results to `main`'s baseline store. That is not correctable afterwards. `VersionKey` in `bench/kube_agents_bench/baselines.py` is `(setup_id, scoring_version, judge_model, fleet, verifiers)`, with no field naming the build a sample came from, so an RC record and a `main` record are the same record once written. The candidate would then be measured for non-inferiority against a window it had just moved. The guard has to land before the job that would start writing, not after.

## What Changed

- **`hack/ci-deploy.sh`** — a new section 2a decides where images come from. `RC_COMMIT_SHA` unset is the presubmit and is unchanged. Set, it skips section 4's `gcloud builds submit` entirely and installs the published images. The two registries differ in image *name* as well as host — Artifact Registry carries `kube-agents-operator`, the published one is `k8s-operator` — so the RC path drops the repository overrides and names only tags. That drop is also what carries the credential-proxy sidecar, which is not a chart value at all: the operator derives it from the agent image by rewriting the trailing path element and keeping the tag (`resolveCredentialProxyImage`), so it follows the agent's repository as long as nothing overrides it. The RC branch also refuses to run unless the tree is checked out at `RC_COMMIT_SHA`, since everything that is not an image comes from the tree.
- **`hack/resolve-rc-target.sh`** (new) — newest `rc_*` tag, resolved to a commit, refused unless all six `REQUIRED_RELEASE_IMAGES` are published there. Stdout is the SHA alone so a caller can use `RC_COMMIT_SHA="$(hack/resolve-rc-target.sh)"`; every diagnostic goes to stderr. Checking the tree out at that commit is deliberately the *caller's* job — bash reads a script incrementally, so a script cannot check out a different revision of itself, and `deploy-environment.yml` already settles this with a second checkout step.
- **`hack/ci-eval-pr.sh`** — declines to record a baseline and declines to publish the dashboard when `RC_COMMIT_SHA` is set.
- **`bench/kube_agents_bench/gate.py`** — `bench-gate record` refuses at the entry point too, for a caller that reaches it another way. Same invariant, second way in, matching the existing `PULL_NUMBER` refusal.
- **`scripts/release/common.sh`** — `check_commit_images_exist` called `docker manifest inspect` unguarded, so on a machine with no docker it failed for every image and the caller reported a publish outage when the real problem was a missing binary. It now probes through a new `registry_image_exists`, which prefers docker and falls back to the GHCR registry API over curl.
- **Docs** — `bench/baselines/README.md` and `docs/designs/eval-scorer.md` both documented the baseline-write rule as two conditions, both about pull requests; both now carry the third. `scripts/release/README.md` notes that the docker-free probe fallback covers `ghcr.io` only.

Two comments in `hack/ci-deploy.sh` claimed "everything this job deploys — chart, operator, agent — is built from the pull request". That stops being true on the RC path, so both now say which path they describe. The correctness argument they were making survives: the chart still comes from the checkout on both paths, which is what makes the caller's checkout load-bearing on the RC path.

## Context

- Part of #1024. Follow-ups, in order: the oss-test-infra Prow job that calls the resolver, checks out the SHA and runs `ci-deploy`; the full-matrix and run-identity work in `hack/ci-eval-pr.sh`; the pipeline trigger and RC commit status; the dashboard Releases section.
- `hack/ci-eval-pr.sh` overlaps with haoxuw's #1025 / PR #1112, and `scripts/eval_dashboard/` with jayantid's #1141, #1142, #1145. The edits here are three small additions to `ci-eval-pr.sh` and none to `scripts/eval_dashboard/`, so this is a merge-conflict note rather than duplicated work.
- Depends on nothing. #1068 is already merged and is why `REQUIRED_RELEASE_IMAGES` has six entries rather than four.

## Testing

- `tests/` (excluding `tests/integration/` and `tests/e2e/`): 975 pass, 16 skipped, 707 subtests pass. Three pre-existing failures in `tests/test_eval_dashboard_refresh.py` are not from this branch — they reproduce with the branch stashed, on the merge base.
- New coverage: `tests/test_ci_deploy_rc_images.py` (10 cases) and `tests/test_resolve_rc_target.py` (9 cases), both new files, plus 9 cases in `tests/test_release_common.py`'s new `RegistryImageProbeTest` and 1 in `scripts/test_eval_dashboard_publish.py`.
- `bench/tests/test_gate.py`: 61 pass, including the new `test_record_refuses_to_run_on_a_release_candidate`.
- `make docs-check`: passes.
- `prettier@3.9.6 --write` on the three changed Markdown files: no reformatting needed.
- `bash -n` on all four changed shell scripts.
- The two `ci-deploy.sh` tests lift sections 2a and 4 out of the real file by regex and execute them against stubs, rather than grepping for flags — a guard that greps passes for a section somebody has commented out. Most of them pin the `RC_COMMIT_SHA`-unset path, because every pull request in the repository runs it and a regression there is a repository-wide outage.
- Not run locally: `shellcheck` (not installed on this machine; CI runs it) and `tests/integration/` (its seam servers need `uvicorn`, absent here). The diff touches neither. No Go or Dockerfile changes, so no `go build` or image-layer budget check.

### Live validation

Exercised against the real repository and real `ghcr.io`, from a machine with **no docker on `PATH`** — which is the Prow job image's shape, and precisely the false negative the `registry_image_exists` guard fixes. No cluster was touched and nothing was deployed, so there is nothing to clean up.

`hack/resolve-rc-target.sh` against the live tag graph, picking the newest candidate on its own:

```
$ hack/resolve-rc-target.sh
ℹ️ Note: rc_2609021231_fdba3a7 (fdba3a7) already carries a validation marker; measuring it again.
❌ ERROR: ghcr.io/gke-labs/kube-agents is missing at least one of the required images at fdba3a7:
   ✓ k8s-operator
   ✓ platform-agent
   ✓ credential-proxy
   ✓ replay-proxy
   ✗ pubsub-platform
   ✗ gke-stockout-investigator
$ echo $?
1
```

That refusal is correct rather than a bug, and I proved the mechanism rather than accepting the coincidence. Probing each of the six packages directly, `pubsub-platform` and `gke-stockout-investigator` return HTTP 200 at `latest` and 404 at `fdba3a7` — the packages exist and are public, the images at that commit do not. The reason is ancestry: `git merge-base --is-ancestor` puts `fdba3a7` 11 commits *behind* `b4ee5f3e` (#1068), which is the commit that added those two build steps to `docker-publish-ghcr.yml` and those two entries to `REQUIRED_RELEASE_IMAGES`. All six return 200 at `b4ee5f3e` and at today's `main` head `667565d5`.

The success path, forced onto a post-#1068 commit with the `RC_TAG` override:

```
$ RC_TAG=b4ee5f3e... RC_TARGET_OUTPUT=/tmp/rc-target.env hack/resolve-rc-target.sh
======================================================================
🏷️ RELEASE CANDIDATE EVAL TARGET
Release Tag:    b4ee5f3eb9c2aceb7f03460d2e573278ab9483fe
Commit SHA:     b4ee5f3eb9c2aceb7f03460d2e573278ab9483fe
Images:         ghcr.io/gke-labs/kube-agents/<image>:b4ee5f3e...
======================================================================
$ echo $?
0
$ od -c /tmp/rc-sha.txt | tail -2      # stdout is the SHA and a newline, nothing else
0000040    a   b   9   4   8   3   f   e  \n
$ cat /tmp/rc-target.env
rc_tag=b4ee5f3eb9c2aceb7f03460d2e573278ab9483fe
rc_commit_sha=b4ee5f3eb9c2aceb7f03460d2e573278ab9483fe
```

**Worth knowing:** until `rc-scheduler.yml` cuts a candidate from a post-#1068 commit, the resolver refuses every RC in the tag graph. That state clears on its own with the next scheduled cut and blocks nothing today, since no job calls this yet — but the follow-up Prow job should not be armed before a candidate with all six images exists.

**What I could not cover.** The deploy itself. `hack/ci-deploy.sh`'s RC branch needs a leased eval project and a GKE cluster, which I have neither of, so it is covered by lifting sections 2a and 4 out of the real script and running them against stubs — real code, stubbed registry and `gcloud`. Unverified against a live install: that a leased eval project's GKE cluster can actually pull from GHCR. The four images the chart renders are public (200 anonymously, shown above), so this should be free, but `ci-deploy.sh` has only ever reached Artifact Registry and I would not call it proven until the follow-up job runs. If it turns out a pull secret is needed, that is a change to the follow-up, not to this diff.

## Self-Review

Both required passes ran in subagents handed the diff range `b4ee5f3e..HEAD` and nothing else — neither held the reasoning that produced the change.

**Docs-drift pass.** Looked for prose anywhere in the repository made false or incomplete by the diff, and for map/testing-map coverage of the new files.

- *Blocking, fixed.* `bench/baselines/README.md` documented the baseline-write rule as "enforced twice", both conditions about pull requests. An RC run is a periodic on a `main` commit with no `PULL_NUMBER`, so a reader concludes it appends — and it does not. Now carries the third condition and why it is not correctable afterwards. The by-hand `bench-gate record` paragraph was stale the same way and now names both variables and `--force`.
- *Fixed, though the pass raised it as non-blocking.* `docs/designs/eval-scorer.md` is the design record for the baseline gate, and a deliberately read-only class of `main`-commit run is a design decision with no home in it. Added a paragraph beside the existing pair.
- *Fixed, non-blocking.* `scripts/release/README.md`'s `REGISTRY_PREFIX` paragraph invites pointing an environment at another registry; the new docker-free fallback covers `ghcr.io` only. Latent rather than currently false — every release-path caller is a GitHub Actions runner with docker — but worth the clause.
- *Not fixed, deliberately.* The pass noted `registry_image_exists` is absent from `scripts/release/README.md:35`'s helper enumeration. That enumeration is already non-exhaustive — `check_commit_images_exist` itself is not in it, appearing only later at line 249 — so adding one helper to a list that does not claim completeness makes the list no more trustworthy. The behaviour that matters to a reader of that file is documented in the `REGISTRY_PREFIX` clause above instead.
- *Not fixed, deliberately.* `bench/tasks/DRAFTS.md:43` says `ci-deploy.sh` installs the PR's own images "on each run". The paragraph's subject is eval-project provisioning on the presubmit path, where it stays exactly true; qualifying it would add a caveat about a path that paragraph is not about.
- *Clean, checked:* `docs/testing-map.md` and AGENTS.md "Where Tests Go" (the two new files land in `tests/`, which `Makefile:142`'s glob already reaches — no new test directory, so the `PYTHON_TEST_DIRS` rule is not engaged); `docs/README.md` map (the diff adds no Markdown document, and there is no `hack/README.md`); the identifier-sources table; `ci-pool-projects.md`, whose every claim about `ci-deploy.sh` survives untouched.

**Adversarial pass.** Looked for ways the RC path can install or grade the wrong thing, ways the presubmit path can regress, tests that pass for the wrong reason, and error paths that mislead. Seven findings; five fixed, two answered.

- *Fixed — the one that could have graded the wrong commit.* `ci-deploy.sh` trusted that the tree had been checked out at `RC_COMMIT_SHA`. Only the images come from the candidate; the chart, the CRDs, and `bench/tasks` come from whatever tree the job sits on, so a caller who set the variable and skipped the checkout would grade the candidate's images against another commit's everything-else, silently. Section 2a now compares `RC_COMMIT_SHA` against `git rev-parse HEAD` and fails with the `git checkout --detach` command, rather than pointing the reader back at a header comment. Covered by `test_a_tree_at_another_commit_stops_the_deploy`.
- *Fixed — two mutations the tests could not see.* The pass demonstrated that deleting the `Accept:` header from the GHCR probe, and rewriting the manifest URL path to `/v2/<repo>/BROKEN/<ref>`, both left the entire suite green. Both break every real probe: GHCR answers 404, not 401, for a manifest request without the OCI media types, which is a false "image missing" indistinguishable from the real one. The mock curl is now shared (`tests/testing/release.py`), asserts the URL, the bearer token and the header before answering, and logs the request so three cases can read it back. Both demonstrated mutations now fail, as does dropping digest-reference handling.
- *Fixed.* `registry_image_exists` ended in a bare `docker manifest inspect …` whose exit status is only safe while every caller keeps the function in a condition context. They all do today; the next caller written as a plain statement would kill the script under `errexit` instead of getting a 1 back. Spelled out as an explicit `return 0` / `return 1`.
- *Fixed.* The same function split image references on the last colon, so `<repo>@sha256:…` would have asked for a repository named `…@sha256`, and a reference-less image for an empty reference. Neither shape reaches it from this diff's callers — every one passes `<repo>:<sha>` — but the docker branch accepts both, and a fallback that accepts less than the primary is a trap for the caller that finds it. Now splits on `@` first, then a colon in the last path segment, defaulting to `latest`. Three new cases.
- *Fixed.* `resolve-rc-target.sh` resolved a tag once and, on failure, reported "cannot resolve a commit" without trying a fetch — a stale clone reads as a bad tag. It now retries after `release_fetch_tags` on the explicit-`RC_TAG` path, which is the path that skips the fetch at the top.
- *Answered, documented rather than changed.* The resolver demands all six `REQUIRED_RELEASE_IMAGES` though an eval install renders only four; both plugin images default to `enabled=false`. A publish run that dropped only a plugin image would block an eval that would not have installed it. Keeping the six is deliberate: the gate's question is "is this commit published", and a shorter list here would be a second definition of a published commit, disagreeing with `verify_release_eligibility.sh` about which candidates exist. Now said in a comment, with what to weigh before shortening it.
- *Answered, not changed.* Section 2a's RC branch exports `IMG`, `AGENT_IMAGE`, `AGENT_TAG` and `IMAGE_TAG`, and the pass observed nothing reads them — `helm` takes the array. Confirmed: the only reader of `IMAGE_TAG` anywhere is `install.sh`, which `ci-deploy.sh` does not call. They are inert on the presubmit path too, and mirroring its shape is what makes the two branches diffable. Deleting pre-existing inert exports is a separate change.
- *Answered — an open question the pass raised rather than a defect.* Whether RC eval artifacts could be folded into the existing dashboard. They cannot, as the code stands: `hack/ci-eval-pr.sh` passes `--pr-glob gs://kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/*/pull-kube-agents-smoke-test/*`, scoped to both `pr-logs/pull/` and the presubmit job name, while a periodic writes under `logs/<job>/`. Skipping the publish is therefore right for this slice; the Releases section is follow-up work, and `scripts/eval_dashboard/` is jayantid's area.
- *What the pass could not cover, in its own words:* it did not execute `bench/tests/test_gate.py`, had no `shellcheck`, ran no `gh` query for overlapping work, and reached no Prow or GKE. The first three are covered above and in Testing; the last is the gap **Live validation** names.

**Found by CI, after opening.** `Run Agent Startup Tests` and `Run Python Unit Tests` both failed on the first commit, and both were the same defect in my test setup rather than in the change. `RegistryImageProbeTest` pinned `PATH` to `{bin_dir}:/usr/bin:/bin`, and `ubuntu-latest` ships docker at `/usr/bin/docker` — so six cases whose subject is what the probe does when docker is *absent* found the real binary and probed ghcr.io for a fixture SHA. They passed locally because the machine I wrote them on has no docker at all, which is the same blind spot in reverse. `PATH` is now a `create_minimal_tools_bin` directory and nothing else, as `tests/test_resolve_rc_target.py` already did; verified by putting a docker stub where the test process can see it and confirming the probe subprocess still reports none (2d6ca577).

**Found while verifying, before either pass.** Two defects I caught myself and fixed, recorded because a reviewer should know the tests were wrong once:

- `tests/test_resolve_rc_target.py` littered the repository root, one fixture directory per test. `create_mock_git_repo` takes `temp_dir.name` from anything that has one, and a `pathlib.Path` has `.name` — the last component. Handing it a `Path` built the mock repo at a *relative* `tmpXXXXXXXX/repo`. Now passes a `str`, with a comment saying why.
- A comment in `hack/resolve-rc-target.sh` claimed the default target is "the newest candidate still without a validation marker". It is not, and cannot be: `rc-tag-validated.yml` writes `_validated` as a *sibling* tag on the same commit, so `rc_2609021231_fdba3a7` and `rc_2609021231_fdba3a7_validated` both exist and the `grep -v '_validated$'` filter never excludes a validated candidate. The code's actual behaviour is right — walking back to an older unvalidated candidate would grade a commit the pipeline has already superseded — so the comment was corrected rather than the code, the marker check now goes through the canonical `is_rc_candidate_commit_already_validated`, and the fixture was rebuilt to model the marker as a sibling. My original fixture put the marker on its own commit with no bare tag, which is what let the wrong claim look tested. Two cases were added for the corrected behaviour.

**One deliberate deviation from the engineering rules.** `.agents/rules/core_engineering.md` asks for `readonly NAME=...` for Bash constants. The two new constants in `scripts/release/common.sh` use `export` instead, because that file has zero `readonly` today and is sourced by many scripts — a re-source would hard-error under `set -e`. The new `hack/resolve-rc-target.sh` is not sourced by anything and uses `readonly` as the rule asks.

## Risk & Rollout

Nothing in the repository sets `RC_COMMIT_SHA`, so on merge every path behaves as it does today and no release this week is affected. Reverting is reverting the commit.

The one change that is *not* dormant is `check_commit_images_exist`'s internals, which the release path does call — `verify_release_eligibility.sh`, from `release-publish.yml`. That job runs on `ubuntu-latest`, where docker is preinstalled, so it takes the `docker manifest inspect` branch and behaves exactly as before; the new fallback is reachable only where the old code was already broken. The new failure mode it introduces is the inverse of the one it fixes: a non-GHCR registry probed from a machine without docker now warns and answers "missing" rather than erroring, which is the conservative direction and is documented.

---

### DIFF
--- a/bench/baselines/README.md
+++ b/bench/baselines/README.md
@@ -93,6 +93,16 @@ be judged against. That is enforced twice — the `JOB_TYPE` condition in
 `PULL_NUMBER` is set — because a guard living only in shell is one careless
 edit away from being gone.
 
+A release-candidate run does not append either, and it is the case the sentence
+above does not cover: the candidate is a commit on `main`, and the run measuring
+it is a periodic with no `PULL_NUMBER`, so it satisfies both conditions exactly.
+`RC_COMMIT_SHA` being set is the third condition, enforced in the same two
+places. It has to be, because the mistake is not correctable afterwards:
+`VersionKey` names the setup, the scoring version, the judge and the two content
+versions, and nothing about which build produced a sample, so a candidate's
+record and `main`'s are the same record once written — and the candidate would
+then be judged non-inferior to a window it had just moved.
+
 A store that will not parse is an **error**, never an empty store. Empty means
 "nothing admitted, aggregate advisory", which is a green; a corrupt file
 reaching that state would silently disarm the gate.
@@ -241,8 +251,9 @@ uv run bench-gate record \
   --lines-out /tmp/appended.jsonl
 ```
 
-It refuses to run with `PULL_NUMBER` set. It is also unconditional on the
-verdict, deliberately — see above.
+It refuses to run with `PULL_NUMBER` or `RC_COMMIT_SHA` set; `--force` overrides
+both, for tests and local screening. It is also unconditional on the verdict,
+deliberately — see above.
 
 With `EVAL_BASELINE_STORE` pointing at a bucket the append lands and the loop
 closes. Against the default local store it does not: the store lives in git,
--- a/bench/kube_agents_bench/gate.py
+++ b/bench/kube_agents_bench/gate.py
@@ -525,6 +525,21 @@ def _cmd_record(args: argparse.Namespace) -> int:
         )
         return 2
 
+    if os.environ.get("RC_COMMIT_SHA") and not args.force:
+        # Same invariant, second way in. A release-candidate eval is a periodic
+        # with no PULL_NUMBER, so the check above lets it through, and a record
+        # written from one is indistinguishable afterwards: VersionKey names the
+        # setup, the scoring version, the judge and the two content versions,
+        # and nothing about which build produced the sample. The candidate would
+        # then be judged non-inferior to a window it had just moved.
+        print(
+            "::error::refusing to record a baseline with RC_COMMIT_SHA set "
+            f"({os.environ['RC_COMMIT_SHA']}): a release candidate is measured "
+            "against main's window, not added to it",
+            file=sys.stderr,
+        )
+        return 2
+
     cases = _read_case_results(args.case_result)
     if isinstance(cases, str):
         print(f"::error::{cases}", file=sys.stderr)
@@ -709,7 +724,7 @@ def main(argv: Sequence[str] | None = None) -> int:
     record.add_argument(
         "--force",
         action="store_true",
-        help="append even with PULL_NUMBER set; for tests and local screening",
+        help="append even with PULL_NUMBER or RC_COMMIT_SHA set; for tests and local screening",
     )
     record.set_defaults(func=_cmd_record)
 
--- a/bench/tests/test_gate.py
+++ b/bench/tests/test_gate.py
@@ -74,6 +74,7 @@ def _clean_env(monkeypatch):
         "EVAL_JUDGED_MARGIN",
         "EVAL_JUDGED_METRICS",
         "PULL_NUMBER",
+        "RC_COMMIT_SHA",
         "PULL_BASE_SHA",
         "GIT_COMMIT",
         "EVAL_BASELINE_STORE",
@@ -503,6 +504,18 @@ def test_record_refuses_to_run_on_a_pull_request(tmp_path, monkeypatch, capsys):
     store = store_with(tmp_path)
     assert run_record(store, case_file(tmp_path, "a")) == 2
     assert "only runs on main" in capsys.readouterr().err
+    assert list(store.glob("*.jsonl")) == []
+
+
+def test_record_refuses_to_run_on_a_release_candidate(tmp_path, monkeypatch, capsys):
+    """The other way into the same invariant. An RC eval is a periodic with no
+    PULL_NUMBER, so the check above lets it through — and a record written from
+    one cannot be told apart later, because VersionKey carries no field naming
+    the build a sample came from."""
+    monkeypatch.setenv("RC_COMMIT_SHA", "b4ee5f3eb9c2aceb7f03460d2e573278ab9483fe")
+    store = store_with(tmp_path)
+    assert run_record(store, case_file(tmp_path, "a")) == 2
+    assert "release candidate is measured against" in capsys.readouterr().err
     assert list(store.glob("*.jsonl")) == []
 
 
--- a/docs/designs/eval-scorer.md
+++ b/docs/designs/eval-scorer.md
@@ -663,6 +663,15 @@ bar upward until nothing could clear it and nothing could fall back below it.
 `hack/ci-eval-pr.sh`, and an independent refusal inside `bench-gate record` when `PULL_NUMBER` is
 set. A guard living only in shell is one careless edit from being gone.
 
+**Nor does a release-candidate run**, and that is a class of `main`-commit run that is deliberately
+read-only rather than a pull request in disguise. An RC eval is a periodic sitting on a commit that
+is on `main` and carries no `PULL_NUMBER`, so both conditions above pass it through. `RC_COMMIT_SHA`
+is the third condition, enforced in the same two places, and it exists because `VersionKey` carries
+no field naming the build a sample came from: once written, a candidate's record and `main`'s are
+indistinguishable, and the candidate is then measured for non-inferiority against a window it just
+moved. `bench/baselines/README.md` is canonical for the rule; this paragraph records why the
+read-only class exists.
+
 Two independent reasons, and the weaker one is the one usually cited. The narrow reason is
 self-admission: a case that wrote its own evidence could be admitted by the very diff that makes it
 pass. The broader reason is that **a branch is not `main`.** The baseline answers "how does this
--- a/hack/ci-deploy.sh
+++ b/hack/ci-deploy.sh
@@ -4,6 +4,11 @@
 # ==============================================================================
 # The evaluation cluster and its IAM are pre-configured; this script builds
 # the PR's images and deploys the kube-agents chart onto that cluster.
+#
+# Setting RC_COMMIT_SHA switches it to the release-candidate path: no build at
+# all, and the chart's published GHCR images at that commit instead. Section 2a
+# is the whole of the difference, and with the variable unset nothing below
+# behaves differently from the day this line was added.
 # ==============================================================================
 
 set -euo pipefail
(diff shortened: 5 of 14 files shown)


dshnayder (approved):

bradhoekstra:
/hold

Temporary, no change requested: parking this PR out of the Tide pool for a few minutes so Tide merges overridden PRs one at a time instead of forming another multi-hour batch. bradhoekstra will `/hold cancel` and re-override shortly.

Issue #1023, opened 2026-08-28
Coverage: a blocking case in every domain


jayantid:
**Goal this week: every domain has at least 2 active cases.**

Nine domains have exactly one case today (capacity, security, upgrades, consistency, cost, remediation, fleet-audits, chat-and-routing, incident-triage) → **9 new cases**. How each is designed is the author's call — the probe pattern (~3.5 min, planted defect, artifact-graded) is proven, not mandated.

## This week
- 9 new cases merged and active, one per single-case domain. Community authorship welcome (#982, #1045 showed it works).
- Carried from before: the rca repo-side PR verifier (its link check is hallucination-passable) and the safeguard-can-fail proof on an unregistered pool project (closes #898).

## Constraints
- Entry bar for any case to *block*: one clean measured run AND one provoked failure. Running non-blocking needs neither.
- Grade artifacts, not prose about them (the week-1 lesson; see kanban's history).

**Done when:** `gh` says every domain ≥2 active cases; the two carried items closed.

Part of #1020 (goal 2) · #897.

**One tracker for coverage.** The per-domain table is in the comments (2026-09-14). The refusal and clean-fleet-silence rows moved to Evals v2 (#1560, item 5) on 2026-09-14.

jayantid:
Re-chartered for the week of Sep 1 (epic #1020): the goal is **at least 2 cases per domain** — nine new cases (one more for each single-case domain: capacity, security, upgrades, consistency, cost, remediation, fleet-audits, chat-and-routing, incident-triage). The probe template is the pattern; entry bar for the blocking roster stays: one clean measured run AND one provoked failure. The original close-out items (rca PR verifier, safeguard-can-fail proof on evals-4) remain in scope.

Fuxiao-Gao:
Plan for this lane, given #1057 (presubmit fan-out) is in flight and the serial matrix already runs over the job deadline:

1. **Author second cases two per PR, merged registered-but-commented-out.** Each authoring PR runs its draft-phase presubmit with the new tasks uncommented at the head of `TASKS` (the unmeasured-first rule), which produces the clean measured run; the provoked-failure half is the mutated-fixture verifier drive (#982's pattern). The final commit comments the entries back out and records build IDs and measurements in `DRAFTS.md` and the task headers. A commented entry costs the merged matrix nothing.
[...]

PR #1175, opened 2026-09-02
feat(ci): add the nightly eval tier -- EVAL_TIER=nightly runs the full catalog

jayantid:
## Summary

`hack/ci-eval-pr.sh` gains a nightly tier: a `NIGHTLY_TASKS` array holding the seven cases recast out of the presubmit for cost, redundancy, or tofu wall clock (the five originally recast plus the two tofu incumbents #1218 moved out), and an `EVAL_TIER` switch that appends it to `TASKS` when a job exports `EVAL_TIER=nightly`. Unset (every existing job) the matrix is byte-for-byte the presubmit's, so this change is dormant until the companion oss-test-infra periodic exports the value — the same dormant-until-armed shape as `EVAL_DASHBOARD_TARGET`. The registration lint now reads both arrays; the domain-coverage lint deliberately still reads only `TASKS`, so a nightly entry counts as presubmit coverage of nothing.

## Why This Change

The nightly lane (#1021) wants a catch-all run of the full eval catalog against `main` every night — every active presubmit case plus the cases that don't fit a presubmit — feeding the baseline evidence store so rung-6 screening and admission (#925's gate machinery) can arm. The catalog side of that job did not exist: the four full audits sat registered-but-commented-out with "recast to the nightly tier" comments pointing at a tier nothing implemented, and `scripts/test_task_registration.py`'s own docstring recorded the debt ("no nightly runner exists anywhere in the tree ... It returns with the nightly tier"). Without this, the nightly periodic could only rerun the presubmit matrix, and the four audits would never run anywhere.

## What Changed

- `hack/ci-eval-pr.sh`: `NIGHTLY_TASKS` (seven entries, each with its one-line reason) and the `EVAL_TIER` switch — `presubmit`/unset is exactly `TASKS`, `nightly` appends `NIGHTLY_TASKS` so the nightly is a superset by construction and shared cases run identically in both tiers; anything else exits 1 before spending a cluster. The four audit-shaped cases and the two tofu incumbents get 900s `unit_cost_hint` entries so a nightly launches them first. The nightly-only cases, one-line reason each:
  - `obtainability-planted-pdb` — full-audit shape, measured PASSED at 962s/rep (2026-08-26): too slow for a presubmit seat; its domain's probe holds that seat.
  - `stockout-pinned-pool` — same shape and cost band; its 2026-08-26 failures were agent-endpoint 502s, not scenario bugs.
  - `upgrade-readiness-lagging-cluster` — same shape and transport failures; its delegated worker filed its ledger issue even so.
  - `consistency-drift-outlier` — same shape; its probe measured 233s/rep, which is why the probe kept the presubmit seat.
  - `obtainability-direct-query` — superseded in presubmit by `reliability-pdb-probe` (same defect, same question); 1.0 on #984's live validation.
  - `gpu-stress-test-diagnosis` — moved out of presubmit by #1218 (tofu provisioning serializes on the infra lock, ~20 min/rep); gates here via the nightly rather than in `BOOTSTRAP_ADMITTED`.
  - `autoops-warning-event-triage` — moved out of presubmit by #1218 (same tofu serialization, ~15 min/rep); accrues its #1101 admission record here.
  - Held OUT of the nightly, because their headers warn they are broken, unvalidated, or fail on a correct agent: `obtainability-refusal-direct-mutation` (objective 0.0 today), `cluster-agent-pending-replicas-capped-pool` (agent cannot diagnose the scenario yet), `cluster-agent-crashloop-fix-request` (`validated: false`, never watched to pass or fail), `chat-routing-fleet-question` (A2 harness blocker — fails on a correct system), `fleet-cost-idle-pool` (A3 calendar gate).
- `scripts/validate_bench_cases.py`: `registered_cases()` reads `NIGHTLY_TASKS` too (`TASKS` missing stays a loud `None`; the nightly array missing surfaces as named orphans, the better failure). The orphan message names the third registration home.
- `scripts/test_ci_eval_nightly.py` (new): lifts the real array and switch text and runs them — dormancy, the append order, the loud typo failure, array disjointness (an entry in both would double-run per nightly), entry existence, and the audit-band cost hints.
- `scripts/test_task_registration.py`: nightly entries count as registered; `NIGHTLY_TASKS` declared exactly once; docstring updated off the "no nightly runner exists" era.
- `EVAL_BASELINE_STORE` stays exactly as #925 designed it — read and write through the one variable, defaulting to unset, no bucket name introduced. Its comment now names the two in-flight oss-test-infra periodics (see Context) instead of implying an armed job exists.
- Docs brought in line with the tier existing: `bench/tasks/DRAFTS.md` (status cells and prose), `bench/CUSTOM-TASKS.md` and `docs/designs/bench-case-format.md` (registration contract), `docs/designs/testing-strategy.md` (§4.4/§4.5 pre-committed to these updates when the eval tier got built), `docs/designs/eval-scorer.md`, `docs/designs/domains.yaml`, `docs/testing-map.md`, `docs/README.md` (one map row), and the five task headers. Table re-padding in `DRAFTS.md`/`docs/testing-map.md` is prettier's own normalization of the changed cells, not hand re-alignment.

The #1043 dashboard publish hook is untouched: its guard keys on `JOB_TYPE`/`PULL_NUMBER`/`EVAL_DASHBOARD_TARGET`, none of which this change reads or writes, and `scripts/test_eval_dashboard_publish.py` stays green.

## Context

- Lane: #1021 (nightly as the catch-all), part of epic #1020; consumers #1023/#1024; store contract #925; dashboard hook #1043.
- Companion (must merge after this): [oss-test-infra#2682](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2682), a DRAFT adding the `ci-kube-agents-eval-nightly` periodic that exports `EVAL_TIER=nightly` — same contract pattern as #2679/#1141. Note the adjacent open work in that repo: [oss-test-infra#2665](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2665) proposes `periodic-kube-agents-eval-baseline` (exports `EVAL_BASELINE_STORE`, does not export `EVAL_TIER`, so as drafted it records the presubmit matrix only), and [oss-test-infra#2679](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2679) creates the same periodics file for the dashboard refresh. The two nightly proposals should converge to one writer; this change works with either, since the tier is one export.
- Tier-decision overlap to resolve deliberately, not by merge order: #1082 activates `stockout-pinned-pool` and `consistency-drift-outlier` as presubmit `TASKS` entries, and #1083 does the same for `upgrade-readiness-lagging-cluster` — the opposite tier call for three of the cases here. If those land, the entries must come OUT of `NIGHTLY_TASKS` in the same change; the new disjointness test reds a double listing rather than letting a case run twice per nightly.
- No open PR implements the nightly tier itself (searched "nightly" across open PRs/issues; #1021 is unassigned).

## Testing

- `bash -n hack/ci-eval-pr.sh` clean; the tier switch also exercised under bash 3.2 (macOS) via the lifted-text tests.
- Full `scripts/` suite: 995 tests OK (re-run after merging main at #1218) (includes the new `test_ci_eval_nightly.py` and the extended registration lint).
- `tests/` suite: green except four pre-existing `test_install_script.ImportGithubPemKmsKeyTest` failures that reproduce on an unmodified checkout (no Go toolchain on this machine); unrelated to this change.
- `python3 scripts/validate_bench_cases.py`: 31 cases OK.
- `make docs-check`: all five checks pass.
- `prettier` could not run locally (no Node runtime on this machine); the two re-padded tables were normalized programmatically to prettier's table algorithm and verified pipe-aligned. CI's prettier job is the authoritative check.

### Live validation

Not live-tested, deliberately: with `EVAL_TIER` unset — every job that exists today — the matrix is byte-for-byte the presubmit's, which the dormancy tests assert against the real script text, so there is no live behavior to observe until a job exports `EVAL_TIER=nightly`. No such job exists yet (the companion periodic is a draft), and running the full nightly matrix by hand would spend a Boskos-pooled project and hours of shared quota to observe what the first scheduled run will measure anyway. The first scheduled nightly is the field test; dormancy without the env is the safety.

## Self-Review

Both required passes ran in clean subagent contexts handed only the diff range (adversarial and docs-drift), before this PR was opened. Findings and dispositions, one list:

- **Fixed (adversarial, blocking):** the `EVAL_BASELINE_STORE` comment claimed the consuming periodic "exists now" under a name nothing uses — every clause verified false against oss-test-infra master. Rewritten to name what is actually true: two in-flight periodics (#2665 and this change's companion), neither merged, converging owed before arming.
- **Fixed (adversarial):** the nightly budget arithmetic asserted a 360m/3-rep job shape as fact; no merged job has any shape yet (and #2665 chose 7h/5 reps). The comment now states the assumption explicitly and how the requirement scales with repetitions.
- **Reported, not fixed here (adversarial):** #1082/#1083 embody the opposite tier decision for three of the five nightly cases — called out in Context above for a deliberate resolution; the disjointness test is the backstop either way.
- **Fixed (docs-drift, blocking):** nine documents/headers still described the "waiting for a nightly tier" state — DRAFTS.md, CUSTOM-TASKS.md, bench-case-format.md, testing-strategy.md (§ banner, §4.4, §4.5 — which pre-committed to exactly this update), eval-scorer.md, domains.yaml, testing-map.md, docs/README.md's map row, and the five task headers. All updated in this PR.
- **Deliberately not fixed:** `AGENTS.md`'s "an eval ... runs in the Prow presubmit" sentence (docs-drift, advisory) — editing the always-loaded instruction file for a tier nuance `docs/testing-map.md` now carries costs context budget for marginal precision; testing-map is the canonical mechanics page and was updated. The eval-scorer.md passages describing #2665 as "a draft pull request" with 10 repetitions stay: they describe that draft, which still exists in that state, and reconciling the two periodic proposals is that convergence's change, not this one's. Probe headers' "the future nightly tier" phrasing (advisory): stale-adjacent but not falsified — left for a sweep once the periodic actually runs.
- **Adversarial pass verified clean:** presubmit dormancy by execution under bash 3.2 with `set -euo pipefail`; every consumer lint re-parses the reshaped file; mutation-testing showed the new tests fail when append order, the default tier, or the validator's nightly parse breaks; the `set -u` empty-array trap on bash < 4.4 is unreachable while the non-empty test holds.

## Risk & Rollout

Dormant by construction: no existing job sets `EVAL_TIER`, and the dormancy tests pin the unset path to the exact presubmit matrix, so the blast radius before the periodic merges is zero. When the periodic arms the tier, the new failure mode is a nightly that blows its deadline — survivable (cost-hinted queue truncates in-flight units; the EXIT trap still records completed ones) and tunable from the periodic's timeout. Revert is a single-commit revert; nothing external references `NIGHTLY_TASKS` until the periodic lands, which is why that PR is a draft gated on this one.


jayantid:
**Smoke red triage (gate-watch, self-note):** environmental — same 18:00-19:25Z Vertex 429 storm as PRs 1177/1153/1142 ([build 2095189652142034944](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1175/pull-kube-agents-smoke-test/2095189652142034944), project evals-6): rung-3 429 reps plus `rca-remediation-pr`/`compliance` collapses via delegation dying on 429s and `kubeconfig is unreadable` (#1144/#1183). Will `/retest` after #1183.

dshnayder (approved):

hack/ci-eval-pr.sh
```diff
@@ -1080,6 +1075,88 @@ TASKS=(
...
+# unadmitted while spending ~10 minutes of matrix a night doing it.
+#
+# Selection is EVAL_TIER below, exported by the Prow periodic and nothing
+# else. The registration lint (scripts/validate_bench_cases.py) reads this
+# array as well as TASKS; the domain-coverage lint deliberately does not, so
+# an entry here satisfies presubmit domain coverage exactly as much as a
+# commented one -- not at all.
+#
+# Budget, priced the way EVAL_REPETITIONS' comment prices the presubmit: the
+# twenty-task presubmit matrix measured ~317min SERIAL, and these five add
```
jayantid: 🟡 **Not merge-blocking.** This budget paragraph went stale when the #1218 merge landed on this branch, and it misleads the next person pricing the nightly.

The comment prices the split as "the twenty-task presubmit matrix measured ~317min SERIAL, and these five add ... ~190min". After #1218 the presubmit is eighteen tasks and this array holds seven, not five: the two tofu incumbents (~106min serial of that ~317min) moved here. The ~507min nightly total is unchanged, but anyone recounting from this text gets the wrong presubmit baseline and the wrong per-array arithmetic. The PR body's "four audit-shaped cases get 900s unit_cost_hint entries" bullet has the same staleness: six entries now carry the 900 hint.

Fix: restate the decomposition as eighteen + seven with the tofu ~106min called out, and update the body bullet.

jayantid: Fixed in c157d2dd: the paragraph now prices the split as eighteen presubmit + seven nightly, with the two tofu incumbents' ~106min serial called out and the ~507min total noted as unchanged. The PR body's cost-hint bullet now reads "the four audit-shaped cases and the two tofu incumbents get 900s unit_cost_hint entries". bash -n, validate_bench_cases (31 OK), and the nightly/registration modules (68 tests) re-ran green at that commit.

jayantid:
/retest

Build 2095881657054662656 red on environment, not this diff: seeded-fleet-reader impersonation PERMISSION_DENIED on leased kube-agents-evals-27, Boskos heartbeat/release 401s, reps pinned at the 45-min delegation timeout, exit 143 at the 360m ceiling. Evidence in #1249 (with #1214 the likely storm half). Single retest per the AGENTS.md protocol.

PR #1173, opened 2026-09-02
ci(gate): demote compliance-rbac-overgrant from BOOTSTRAP_ADMITTED to held-out (#1171)

jayantid:
## Summary

Demotes `compliance-rbac-overgrant` from `BOOTSTRAP_ADMITTED` to the held-out list, per the demote-same-day protocol written above the roster in `hack/ci-eval-pr.sh`. The case keeps running and reporting on every pull request, but can no longer red one on a graded (rung-4) failure. Tracking issue #1171 carries the failure evidence and the re-admission bar; the dashboard case-notes gain the `held-out` badge and the issue reference so the published dashboard says the same thing the gate does.

## Why This Change

The gate went merge-blocking today (~14:00Z), and this case — the fleet-audits canary — is collapsing on pull requests whose diffs cannot explain it: five all-reps FAILED verdicts across unrelated PRs on 2026-09-02 (#1148, #1090, #1145, #1089, #1153 — builds and rep-level reasons in #1171). **#1153 is red solely on this case**; its diff is auth-retry changes in `forge.py`, nowhere near the bench or the audit path. The collapse is environment-side, not diff-side: delegated audits are losing control-plane reachability to the seeded clusters (partial audits that skip check 2.4 — the check that finds the planted `debug-binding` — with the agent reporting "access limitations" and falling back to raw `curl`), and some runs never publish a ledger at all (evals-9 publish gap 03:33Z–11:28Z, evals-10 03:23Z–13:51Z). Without this change, every PR that draws a degraded eval project stays blocked on a red it did not cause.

## What Changed

- `hack/ci-eval-pr.sh`: `compliance-rbac-overgrant` removed from the `BOOTSTRAP_ADMITTED` default (the one-line fast lever the roster comment prescribes); held-out entry added to the comment block referencing #1171 and its exit condition; admitted/held-out counts updated (twelve admitted, five held out).
- `scripts/eval_dashboard/case-notes.yaml`: `badge: held-out`, `#1171` added to the case's issues.
- `scripts/test_eval_dashboard.py`: expected issues list updated to match.

Deliberately NOT changed: rungs 1–3 stay blocking for this case (mutations, erroring verifiers, empty records still red a PR — the "no agent ran at all" mode seen today is rung 1-3 and demotion correctly does not silence it), and the case itself, its verifier, and the SOP are untouched — the defect is in the environment/delegation chain, and #1171 tracks that fix separately.

## Context

[...]

Issue #1144, opened 2026-09-02
Eval installs: credential proxy gets permission denied on /opt/data/profiles, so every kubectl fails and agents return blocked-preflight notices


stalhaali:
## In one paragraph

On the eval clusters, the credential proxy cannot read the folder `/opt/data/profiles/`. That folder
holds the kubeconfig files agents use, so every `kubectl` command the proxy is asked to run fails
with "permission denied". The visible symptom is confusing: an agent replies "I am blocked, I cannot
read my kubeconfig" instead of doing its job, and the eval then grades that reply as a wrong answer.
Seen on two different clusters, twelve hours apart, on two unrelated pull requests. **We do not know
the cause** — this issue is the symptom plus the evidence, for someone who can get onto a live pod.

## The symptom

From PR #1122's smoke run ([build 2094905948941324288](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1122/pull-kube-agents-smoke-test/2094905948941324288), project `kube-agents-evals`), the
proxy returns HTTP 500 for every exec request, several times a minute:

```
2026-09-01 22:17:07 INFO  credential-proxy exec request_id=... principal=unauthenticated executable=kubectl
2026-09-01 22:17:07 ERROR credential-proxy command failed request_id=... type=PermissionError
Traceback (most recent call last):
  File "/opt/defaults/scripts/credential_proxy.py", line 3037, in do_POST
    result = self.executor.execute(
...
PermissionError: [Errno 13] Permission denied: PosixPath('/opt/data/profiles/platform')
2026-09-01 22:17:01 INFO  credential-proxy http "POST /v1/exec HTTP/1.1" 500 -
```

`Errno 13` is the OS refusing a read. The path exists; the process is not allowed to open it.

## What it looks like from the outside

This is the part worth knowing, because the symptom does not look like a permissions bug.

On PR #1089's smoke run ([build 2094714569262895104](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1089/pull-kube-agents-smoke-test/2094714569262895104), project `kube-agents-evals-5`,
about twelve hours earlier), the same fault hit a *cluster* profile instead of the platform one. The
eval case `autoops-warning-event-triage` asked an agent to triage an incident, and the answer it
graded was:

> 🔀 Delegated to the **cluster-kube-agents-evals-5-…** agent
>
[...]

PR #1177, opened 2026-09-02
perf(test): warm the Go build cache and sweep the Python test directories in parallel

bradhoekstra:
## Summary

`Python Unit Tests` is the longest workflow on a pull request here, and its wall-clock is set by whichever of `make test-python` and `make coverage` finishes last. Two changes cut both: the injector seam test stops throwing away the Go build cache CI restored a moment earlier, and the 25 directories in `PYTHON_TEST_DIRS` run concurrently instead of one at a time.

Measured against this branch's merge base on an 8-core box: `make test-python` 5m42s → 1m21s, `make coverage` 8m31s → 2m57s. Coverage is the one that matters, since it is the longer job.

## Why This Change

`tests/integration/test_seam_injector_kv.py` shells out to `go test`, which compiles the whole operator dependency tree — controller-runtime and client-go. It set `GOCACHE` to a fresh temp directory, so that compile started cold on every run: 88.4s against 3.3s into a warm cache. Both CI jobs run `actions/setup-go` with `cache-dependency-path: k8s-operator/go.sum`, which restores roughly 413 MB into `~/.cache/go-build` immediately before `make` starts, and the temp directory discarded all of it. One file was ~40% of `make test-python`.

The sweep over `PYTHON_TEST_DIRS` was sequential, so the two targets cost the sum of 25 directories rather than the slowest one. Four directories are most of that sum.

## What Changed

- **`tests/integration/test_seam_injector_kv.py`** — `_go_cache_dir()` probes `go env GOCACHE` and uses it when it is an absolute path the process can write to, keeping a temp directory as the fallback. The file drops from 1m42s to 25s.
- **`Makefile`** — `PYTHON_TEST_JOBS` (default `nproc`) and a `sweep_python_test_dirs` macro that both `test-python` and `coverage` call.
- **`tests/test_python_test_sweep.py`** (new) — asserts a failing directory still fails the target, at both job counts.
- **`agents/platform/scripts/test_pr_triggers.py`** — a wall-clock assertion that concurrency showed to be flaky, replaced with a substitution-pass count plus a loose ceiling.
- **`agents/platform/scripts/pr_triggers.py`**, **`docs/testing-map.md`** — two prose corrections the passes below turned up.

Three things about the shape:

**The absoluteness check on `go env GOCACHE` is load-bearing.** With `HOME` unset the command prints the sentinel `off` and exits 0. Taking that at its word both writes a stray `off/` directory into the current working directory and hands `go test` a disabled cache, which it refuses: `build cache is disabled by GOCACHE=off, but required as of Go 1.12`. Rejecting anything non-absolute sends that case to the fallback where it belongs.

**One macro, not two loops.** `coverage` used to carry its own copy of the loop, with a comment asserting the two matched. A coverage target that walks the directories differently measures a different suite than the one that gates, and nothing would say so. Sharing the sweep makes the mirroring structural rather than asserted.

**The verdict channel is fail-closed.** A directory's result can no longer be a shell variable, because a subprocess cannot assign one the parent will see. Each worker writes its exit status to a per-directory file and the parent treats missing-or-non-zero as failure. A single shared append-only file would be shorter, but a failed append — a full `TMPDIR`, likelier with 25 concurrent suites than with one — would silently drop a red directory and let the gate pass.

## Context

No open pull request touches this work. [#1118](https://github.com/gke-labs/kube-agents/pull/1118) (AntonTyb) edits the same import block in `test_pr_triggers.py` at `@@ -40,6 +40,7 @@` and widens `MARKER_RE` to `agent-(answered|refused|updated)`. That is a textual conflict for whoever lands second and nothing more: the extra alternative does not change the per-pass match count on an `agent-answered` nest, so the pass-count assertion here survives the merge either way. Six other open pull requests touch the `Makefile` (#1124, #1121, #1090, #965, #913, #498); none touch the sweep region.

This is the second of the CI throughput items measured alongside [#1163](https://github.com/gke-labs/kube-agents/pull/1163) (Docker layer cache). Two remain and are not attempted here: path-filtering the advisory workflows, and consolidating the sub-minute workflows so they stop paying a runner allocation each.

## Testing

- `make docs-check` — passes. Generated regions current, 282 files link-checked, map covers 249 tracked docs, context budget 35.4k/38k.
- `prettier@3.9.6 --check` on the changed Markdown — clean.
- `make test-python` — 1m21s, the same two red directories as at the merge base (`admin_console/tests/`, `agents/platform/scripts/`, 4 ERRORs), all of them pre-existing local dependency gaps confirmed by re-running on a stashed tree. Green in CI.
- New test mutation-checked: deleting the parent loop's verdict check reds three of `tests/test_python_test_sweep.py`'s four cases.

### Live validation

A `Makefile` and a test-runner change cannot reach a kube-agents installation — nothing here renders a manifest, reconciles a CR, or runs in a pod. Per `contributing.md` the substitute is exercising it where it actually runs, so the evidence below is real runs of the two targets rather than reasoning about them.

**Equivalence before speed.** A faster suite that runs a different set of tests is not a faster suite.

| | Sequential | Parallel |
| --- | --- | --- |
| Directories swept | 25 | 25 |
| Tests run | 5638 | 5638 |
| Per-directory test counts | — | byte-identical (`diff` clean) |
| Coverage total | 26665 stmts, 7311 missed, 72.6% | 26665 stmts, 7311 missed, 72.6% |
| Red directories | 2 (pre-existing) | 2 (same two) |

The coverage row is the strongest of these: the merge base and this branch, run through entirely different sweeps, report the same three numbers.

**Wall-clock**, on an 8-core box, `PYTHON_TEST_JOBS` at its `nproc` default. "Before" is `b4ee5f3e`, the merge base, measured in a detached worktree rather than inferred:

| | Before | After |
| --- | --- | --- |
| `make test-python` | 5m42s | **1m21s** |
| `make coverage` | 8m31s | **2m57s** |
| `test_seam_injector_kv.py` alone | 1m42s | **25s** |
| `go test ./cmd/k8s-event-watcher/`, cold / warm GOCACHE | 88.4s | 3.3s |

**Proving the mechanism rather than a coincidence.** Each half was measured on its own rather than credited from the joint total. The cache change: the seam file alone, 1m42s at the merge base against 25s here, and the `go test` it drives at 88.4s into an empty cache against 3.3s into a warm one. The sweep: this branch at `PYTHON_TEST_JOBS=1` runs `make coverage` in 5m40s and at the `nproc` default in 2m57s, same commit, same suite, only the concurrency differing.

The two do not add up to the joint figure and I am not claiming they do — run-to-run spread on this box is tens of seconds, and the intermediate datapoints were not taken back-to-back with the rest. What the separate measurements establish is direction and rough magnitude for each half independently, which is what "not a coincidence" needs.

All four branches of `_go_cache_dir` were exercised — normal (returns `~/.cache/go-build`), uncreatable path, read-only parent (`HOME=/tmp/rohome`, mode 555, `mkdir` raises and the fallback is taken), and `HOME` unset (returns the temp fallback, and leaves no `off/` directory behind, which the pre-fix code did). Stability: three consecutive parallel runs of `test-python`, no flakes and identical results.

**Not covered:** BSD `xargs` — `-P` with `-I` is the one construct not exercised off GNU findutils, and the `nproc` fallback implies macOS is a supported dev platform, so it is worth someone running `make test-python` there once. Also uncovered is whether the independence claim holds on a 2-core runner, where the interleaving differs; the CI run on this pull request is the first real evidence and its timings are in the thread below once it lands.

**Baseline for the CI comparison**, four recent `Python Unit Tests` runs on `main`-targeting pull requests: `Run Python Unit Tests` 278–332s, `Coverage report` 335–358s, workflow ~361s.

## Self-Review

Both required passes ran in **subagents that did not write the change**, each handed only the diff range.

**Adversarial pass** — attacked the macro's quoting and `$(call)` expansion, whether the directories are genuinely independent, whether concurrent `coverage run` is safe, the GOCACHE probe's failure modes, and whether the rewritten assertion catches what the old one did. Ten findings:

- **Fixed (HIGH): `_go_cache_dir` returned the literal `"off"` when `HOME` is unset.** Confirmed by running it: `env -u HOME go env GOCACHE` prints `off` and exits 0, the function `mkdir`'d a relative `off/` in the cwd — `tests/integration/` under the sweep — and handed `go test` a disabled cache, turning a slow-but-green run into a hard failure plus a write inside the checkout. The pre-change code was immune because it never consulted `go env`. Fixed by requiring an absolute path; re-verified both directions.
- **Fixed (MEDIUM): the diff moved `verify:` 65 lines and stale-dated a line citation.** `docs/testing-map.md` said `Makefile:173`, which at HEAD lands in the middle of the new macro's comment. `make docs-check` does not validate `Makefile:NNN` references. Replaced with the target name so it cannot rot again. Three other `Makefile:NNN` citations elsewhere are off by one and pre-date this change; left alone as out of scope.
- **Fixed (MEDIUM): the `PYTHON_TEST_JOBS=1` comment sold a property the sweep does not have.** It claimed serialising gets you output arriving a directory at a time; the capture is unconditional, so `-P 1` changes execution order and nothing else. Comment rewritten to say what serialising is actually for.
- **Fixed (MEDIUM): two docstrings gave contradicting measured costs.** My new one said ~0.4s, `strip_markers`' said 0.08s; re-measured five times here at 0.267–0.274s. The conclusion held — a 0.3s bound has 10% of headroom and any load closes it — but the stated reason was wrong, so both are corrected to that number and the module docstring now leads with the pass count instead of a wall-clock.
- **Fixed (LOW-MEDIUM): the new pass-count assertion passed vacuously at zero.** `assertLessEqual(sub_calls, depth + 2)` is satisfied by `sub_calls == 0`, which is what a rewrite reaching for `finditer` or a differently named pattern leaves behind — the wrapper uncalled, the assertion guarding nothing while still reading like a cost bound. Now `assertEqual`, the value being exact and deterministic. The same finding noted that dropping wall-clock entirely means a backtracking regression would make each of the 2,602 passes slow and still pass; a 10s ceiling is back for that, ~40x the measured cost and unreachable by load.
- **Fixed (LOW): `$(call)` splits arguments on commas, and the comment warned only about single quotes.** A future third caller like `python3 -c "import os, sys"` would arrive silently truncated. Documented alongside the quote rule.
- **Fixed (LOW): a lost append to the shared `failed` file would let a red directory ship green.** Narrow trigger (a `TMPDIR` filling mid-run) and not reproduced, but the mechanism was in the code. Replaced with per-directory verdict files read fail-closed, so absence means failure.
- **Fixed (LOW): nothing tested the new failure-propagation path.** `tests/test_python_test_sweep.py` added, covering both job counts and both directions, plus one end-to-end case that the target converts `$failed` into a non-zero exit. Mutation-checked.
- **Accepted (LOW): a run killed mid-sweep loses the captured output.** `trap … TERM` fires on a CI cancellation or step timeout and the per-directory logs only print at the end, so a cancelled run shows the progress lines and nothing else. Under the old loop the output streamed. This is the cost of not interleaving 25 writers mid-line; the per-worker progress line is the mitigation, and it is now written down in the macro's comment rather than discovered on the first CI hang.
- **Noted, no change: #1118 conflicts textually.** Covered under Context above.

Confirmed as *not* findings, each checked by running rather than reading: the 25 directories are genuinely independent (four runs, identical per-directory pass/fail sets and test counts, `git status --porcelain` unchanged; every server in the sweep binds port 0); concurrent `coverage run` under `parallel = True` produces identical totals to serial; `$$` escaping and quoting are correct throughout both expansions; the 25 log-file names do not collide; `GO is None` cannot reach `_go_cache_dir` because `skipUnless` sits on the class; `os.access` is not racy here because `mkdir` runs first and returns `EROFS` even as root; concurrent `go test` on a shared `GOCACHE` is safe by design and does not arise anyway; the missing `-count=1` in one Go invocation cannot hit the test cache because `go test` caches only passing runs; and no workflow, script, or skill parses the sweep's output.

**Docs-drift pass** — one Blocking finding, the `docs/testing-map.md` citation above, now fixed. It verified that no document anywhere describes the sweep's ordering (grepped every `.md`/`.mdx` for sequential/serial/one-at-a-time/parallel phrasing — the docs say *which* directories run and *what* runs them, never in what order), that `PYTHON_TEST_JOBS` is owed no documentation because it is a variable rather than a target and `COVERAGE_SKIP_GO` sets that precedent, that `docs/README.md` owes no row since the diff adds no Markdown, that `make help` output is unchanged, and that no document states timings for the Python suite. Its advisory finding about the three surviving timing bounds — one guards anchoring, not backtracking — is fixed in the docstring.

## Risk & Rollout

The blast radius is CI and local development; nothing here ships in an image or reaches a cluster. The new failure mode is a test that turns out to depend on running alone, which would surface as a flake rather than as a wrong result — `PYTHON_TEST_JOBS=1` is the immediate mitigation and needs no revert, and the equivalence table above is the evidence that none does today. The `GOCACHE` change means the seam tier now writes to the developer's real build cache instead of a temp directory it deletes; that is the point, and it is what CI already assumed.

Reverting is `git revert`; nothing persists outside the run.

### DIFF
--- a/tests/test_python_test_sweep.py
+++ b/tests/test_python_test_sweep.py
@@ -1,0 +1,135 @@
+"""A failing directory in the `make test-python` sweep still fails the build.
+
+The sweep runs PYTHON_TEST_DIRS concurrently, so a directory's verdict can no
+longer be a shell variable -- a subprocess cannot assign one the parent will
+see. It travels through a per-directory file in a temp directory instead, and
+that is a channel with ways to go quiet: a write that fails, a name that
+collides, a parent loop that reads the wrong path. Every one of them looks the
+same from outside -- `make test-python` exits 0 with a red directory in the run
+-- which is the failure this repository's suite most needs not to have.
+
+Nothing else covers it. `scripts/test_test_discovery.py` checks which
+directories the sweep *reaches*; this checks what its verdict does to the exit
+status. Both job counts run, because serial and concurrent are separate paths
+through the same macro and only the concurrent one is new.
+
+Most of this drives `sweep_python_test_dirs` directly through a wrapper
+makefile rather than through `make test-python`, which is worth the indirection
+purely for time: the target's missing-import preflight starts twenty Python
+interpreters, and paying that six times would add half a minute to the suite
+this sweep exists to shorten. One case does go through the real target, since
+what the macro leaves in `$failed` matters only if the caller turns it into a
+non-zero exit.
+"""
+
+import os
+import pathlib
+import subprocess
+import tempfile
+import unittest
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parent.parent
+
+#: A real directory with no test_*.py in it. Discovery there finds nothing and
+#: succeeds, which is the "green" half of each case below at near-zero cost.
+EMPTY_DIR = "docs/"
+#: A directory that does not exist, so the sweep's `cd` fails before the
+#: per-directory command runs at all. Cheaper than a directory holding a
+#: deliberately failing test, and it exercises the same path out of the worker.
+MISSING_DIR = "nosuchdir/"
+#: Serial and concurrent are separate paths through sweep_python_test_dirs.
+JOB_COUNTS = (1, 2)
+#: Printed by the probe target below so the test can read `$failed` back.
+FAILED_MARKER = "SWEEP-FAILED:"
+SWEEP_TIMEOUT_SECONDS = 300
+
+#: A target that runs the sweep over a trivial command and reports the one
+#: thing the macro promises its callers: what `$failed` holds afterwards.
+PROBE_MAKEFILE = f"""\
+include Makefile
+sweep-probe:
+\t@$(call sweep_python_test_dirs,true) >/dev/null; echo "{FAILED_MARKER}[$$failed]"
+"""
+
+
+def _run_make(args):
+    env = dict(os.environ)
+    # This test may itself be running inside the sweep, and an inherited
+    # jobserver or MAKELEVEL would make the nested make behave unlike the one a
+    # developer runs by hand.
+    env.pop("MAKEFLAGS", None)
+    env.pop("MAKELEVEL", None)
+    return subprocess.run(
+        ["make", *args],
+        cwd=REPO_ROOT,
+        env=env,
+        capture_output=True,
+        text=True,
+        timeout=SWEEP_TIMEOUT_SECONDS,
+    )
+
+
+def sweep(dirs, jobs):
+    """Run the macro over `dirs`, returning (completed process, `$failed`)."""
+    with tempfile.NamedTemporaryFile("w", suffix=".mk", delete=False) as wrapper:
+        wrapper.write(PROBE_MAKEFILE)
+        wrapper_path = wrapper.name
+    try:
+        done = _run_make(
+            [
+                "-f",
+                wrapper_path,
+                "sweep-probe",
+                f"PYTHON_TEST_DIRS={' '.join(dirs)}",
+                f"PYTHON_TEST_JOBS={jobs}",
+            ]
+        )
+    finally:
+        os.unlink(wrapper_path)
+    marker = [ln for ln in done.stdout.splitlines() if ln.startswith(FAILED_MARKER)]
+    failed = marker[-1][len(FAILED_MARKER) :].strip("[]") if marker else None
+    return done, failed
+
+
+class SweepVerdictTest(unittest.TestCase):
+    def test_a_failing_directory_is_named_in_failed(self):
+        for jobs in JOB_COUNTS:
+            with self.subTest(jobs=jobs):
+                done, failed = sweep([EMPTY_DIR, MISSING_DIR], jobs)
+                self.assertEqual(MISSING_DIR, failed, done.stdout + done.stderr)
+
+    def test_a_clean_sweep_leaves_failed_empty(self):
+        # The other direction, so the test above cannot pass by the macro
+        # reporting everything as failed.
+        for jobs in JOB_COUNTS:
+            with self.subTest(jobs=jobs):
+                done, failed = sweep([EMPTY_DIR], jobs)
+                self.assertEqual("", failed, done.stdout + done.stderr)
+
+    def test_every_directory_runs_even_after_one_fails(self):
+        # The property the sequential loop had and the sweep had to re-earn: one
+        # failure must not stop the directories after it. A `set -e` regression
+        # here would hide whole suites behind a familiar-looking red run.
+        done, _ = sweep([MISSING_DIR, EMPTY_DIR], max(JOB_COUNTS))
+        self.assertIn(f"==> {EMPTY_DIR}", done.stdout)
+        self.assertIn(f"==> {MISSING_DIR}", done.stdout)
+
+
+class TestPythonExitStatusTest(unittest.TestCase):
+    def test_the_target_exits_non_zero_when_a_directory_fails(self):
+        # The one end-to-end case: `$failed` is only worth setting if the caller
+        # acts on it, and a sweep that reports correctly into a target that
+        # swallows the result is the same green-on-red as no sweep at all.
+        done = _run_make(
+            [
+                "test-python",
+                f"PYTHON_TEST_DIRS={EMPTY_DIR} {MISSING_DIR}",
+                f"PYTHON_TEST_JOBS={max(JOB_COUNTS)}",
+            ]
+        )
+        self.assertNotEqual(0, done.returncode, done.stdout + done.stderr)
+        self.assertIn(MISSING_DIR, done.stdout.split("Failing test directories:")[-1])
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,82 @@ PYTHON_TEST_DIRS := $(sort $(dir \
 	$(wildcard tests/test_*.py) \
 	$(wildcard tests/memory/test_*.py)))
 
+# How many of those directories `test-python` and `coverage` run at once. They
+# are separate `python3` processes that share nothing -- each cd's into its own
+# directory, servers in the seam tier bind port 0, and fixtures go through
+# tempfile -- so the sweep costs its slowest single directory rather than the
+# sum of all of them. Four directories are most of that sum, so the win is
+# large and then flat: raising this past a handful buys nothing.
+#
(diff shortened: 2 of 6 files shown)


tests/test_python_test_sweep.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import pathlib
+import subprocess
+import tempfile
+import unittest
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parent.parent
+
+#: A real directory with no test_*.py in it. Discovery there finds nothing and
+#: succeeds, which is the "green" half of each case below at near-zero cost.
```
jayantid: 🟡 **Not merge-blocking** — every assertion tolerates this; the comment is wrong, not the code.

Line 34 says discovery in `docs/` "finds nothing and succeeds". On Python 3.12, the version `.python-version` pins, `python3 -m unittest discover` exits 5 when no tests run, so `docs/` fails.

The claim holds for the three probe cases, whose command is `true`. It breaks in the end-to-end case: `make test-python` runs real discovery, `docs/` lands in `$failed` beside `nosuchdir/`, and the "green half" that case was designed around does not exist on the pinned interpreter. The test still passes, because it asserts only a non-zero exit and that `nosuchdir/` is named. But the target's output (`Failing test directories: docs/ nosuchdir/`) contradicts the comment, and anyone strengthening the assertion to keep `docs/` out of `$failed` reds CI immediately.

Scope the claim to the probe cases:

```python
#: A real directory with no test_*.py in it. Under the probe target's `true`
#: this is the "green" half of each case at near-zero cost. The end-to-end
#: case is the exception: Python 3.12+ exits 5 when discovery finds no tests,
#: so there docs/ fails too, and the assertions rely only on nosuchdir/.
EMPTY_DIR = "docs/"
```

jayantid:
**Smoke red triage (gate-watch):** not caused by this PR (diff is Makefile/CI/tests). [Build 2095190038248689664](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1177/pull-kube-agents-smoke-test/2095190038248689664) fell in the fleet-wide Vertex 429 window (~18:00-19:25Z, hit 4 unrelated PRs on 4 different pool projects) — #1097/#1184: rung-3 reps are empty-trajectory 429s, and the `rca-remediation-pr`/`compliance-rbac-overgrant` collapses are delegated work dying downstream on the same 429s + the credential-proxy `kubeconfig is unreadable` rejections (#1144, fix in flight #1183). Note #1173 merged 18:51Z, so compliance no longer gates. `/retest` once the quota window calms.

jayantid (approved):

PR #1107, opened 2026-09-01
chore(deps): bump LiteLLM from v1.96.2 to v1.98.0

bradhoekstra:
## Summary

Moves the LiteLLM pin from `v1.96.2` to `v1.98.0`, two minor releases forward. LiteLLM is the model gateway every agent call goes through, so this is the largest jump in the batch by release distance. `v1.98.0` is the latest stable — the newer tags upstream carries are prereleases, and this deliberately skips them.

## Why This Change

Two minors of upstream fixes on the component that brokers every model call. Nothing in the diff is driven by a specific CVE; the reason is that LiteLLM moves fast and a gateway two minors behind accumulates provider-compatibility bugs that surface as a single provider quietly failing rather than as an outage.

The version is stated in five places besides the inventory — the chart, two `examples/` deployments, and a staging workload bundle — so the bump is also about keeping those from diverging.

## What Changed

- `images.json` — the `litellm` entry, source for the rest.
- `charts/kube-agents/values.yaml` — the chart's pin, cross-checked by `make images-check` on both a default and a mirrored render.
- `examples/litellm-gemini/deployment.yaml` and `examples/litellm-chatgpt-subscription/deployment.yaml` — the two worked provider examples, covered by inventory check 4.
- `.../testing/staging_workloads/charts/workload-bundle/values.yaml` — the staging bundle's copy.
- `docs/site/src/content/docs/deploy/docker-images.md` — regenerated `<!-- BEGIN GENERATED -->` region.

## Context

One of seven single-image bumps opened together, each independently revertable: LiteLLM, Envoy, fluent-bit, alpine/k8s, hindsight-api, the Go toolchain, and the replay-proxy Python base. The Hermes agent pin is deliberately not in this batch — it lands separately and last. No open issue or pull request touches these files.

Version choice: upstream had tags newer than `v1.98.0` at the time of writing, all prereleases. This takes the latest stable.

## Testing

- `make images-check` — passes; all five checks, including the two `examples/` manifests.
- `make docs-check` — passes.
- `prettier --check` on the changed Markdown and YAML — clean.

### Live validation

`kube-agents-cluster` (regional `us-central1`, project `bhoekstra-gkedemos`), namespace `kubeagents-system`. Live-test lease held throughout. Exercised as part of one batched run covering all six non-Hermes bumps.

**Partial, and the gap is the important part of this section.**

[...]

PR #1182, opened 2026-09-02
chore(images): move hindsight-postgresql to the maintained pgvector image

bradhoekstra:
## Summary

`hindsight-postgresql` — the Postgres-with-pgvector image behind the Hindsight memory store — moves from `docker.io/ankane/pgvector:latest@sha256:956744bd…` to `docker.io/pgvector/pgvector:0.8.6-pg15@sha256:a947c45c…`. This is a rename to follow rather than a replacement: `ankane` is the pgvector author's personal Docker Hub namespace and the project now publishes under its own org. In practice it moves Postgres 15.4 → 15.19 and pgvector 0.5.1 → 0.8.6, staying on the pg15 line because the StatefulSet keeps its data on a PVC. Twelve other files named the old image, and five comments justified the digest pin with "upstream publishes only a floating `latest`", which the new tag makes false — so the rationale is rewritten rather than the string alone.

## Why This Change

The pinned image was last published on 2023-10-11. Every install has therefore been running about two years of unpatched Postgres and Debian 12 packages under a name nobody upstream maintains. Nothing breaks today, and nothing will announce it when something does: `ankane/pgvector` receives no CVE rebuilds, so the pin ages quietly and a future advisory against Postgres 15 or the base image has no upstream fix to move to.

## What Changed

- `images.json` is the source of truth every other consumer resolves from — the chart, `make mirror-images`, the kustomize deploy path, and the generated image table in the docs all follow it. The pg15 constraint and the November 2027 Postgres 15 EOL now live in its `description`, and the three places that used to restate them point there instead.
- `charts/kube-agents/values.yaml` and `k8s-operator/config/integrations/hindsight/postgresql.yaml` carry the two rendered references.
- `scripts/mirror_images.sh:140` said three inventory entries pin with `tag@sha256`; it is four — `hindsight-api`, `hindsight-postgresql`, `busybox`, and the Hermes base through `tags.env`. Pre-existing, corrected here because the comment sits in a file this change already edits.
- `docs/designs/memory.md` gains the two things a bump like this does not make obvious: an existing database keeps its extension at the version that created it, and the shared library moves anyway — which is what turns iterative ANN scans on.

## Context

No open issue or pull request covers this — nothing open touches the pgvector image or `images.json`'s Hindsight entries, and the nearest neighbour is the merged hindsight-api 0.9.1 → 0.9.2 bump, which this does not depend on.

Seven open pull requests do touch the same files in other places, so this is a merge-conflict warning rather than overlapping work: #1174 (`memory.md`), #1161 and #1159 (`values.yaml`, both draft), #1107 (`values.yaml` and `images.json`, the LiteLLM entry), #965 and #913 (`values.yaml`, `images.json`, `hack/check-image-inventory.sh`), and #608 (`images.json`). Whichever lands second rebases.

## Testing

- `make docs-check`, `make chart-check`, `hack/check-image-inventory.sh` — all pass.
- `prettier@3.9.6 --check` on every changed Markdown, YAML, and JSON file — clean.
- 78 passed, 3 skipped in the image/inventory/mirror/hindsight test selection.
- `shellcheck hack/check-image-inventory.sh` reports two SC2317 informationals, both present on `main` before this change.

### Live validation

gkedemos (`bhoekstra-gkedemos` / `kube-agents-cluster`, `us-central1`), under the live-test lease, in a scratch namespace `pgvector-livetest`. The shared install has memory disabled and no Hindsight deployed, so there was nothing in it to exercise; the test deployed the real thing through `make -C k8s-operator deploy-hindsight` with `hindsight-api:0.9.2`.

**An existing PVC upgrades in place.** Postgres came up on the *old* pin first, the API ran its migrations (23 tables, alembic `d1e2f3a4b5c6`, a real `hnsw (vector_cosine_ops)` index on `vector(384)`), and 500 embeddings were written. Redeploying with only the image changed reported `deployment.apps/hindsight-api unchanged` and `statefulset.apps/hindsight-postgresql configured`. The pod returned Running in 29s on the same PV `pvc-b94ed306`: Postgres 15.4 → 15.19, 500 rows, embedding checksum `d972db14…` byte-identical, alembic head unchanged, and the HNSW index built under 0.5.1 still scanned. The extension stayed at 0.5.1, as expected — the image supplies the `.so`, the database keeps its creating version. `ALTER EXTENSION vector UPDATE` then walked it to 0.8.6 with data and index intact and `halfvec` available.

**A fresh install gets 0.8.6.** With the StatefulSet and PVC deleted and redeployed: a new database at pgvector 0.8.6 automatically, the same 23 tables and alembic head, the HNSW index created, and `PGDATA` sitting beside `lost+found` in the subdirectory the manifest comment exists to explain. `/health` reported `database: connected`, and a recall through `/v1/default/banks/…/memories/recall` returned 200.

**Index correctness, not just query success.** Rather than treating a query that ran as a query that was right, the forced HNSW index scan was compared against a forced exact sequential scan over the same data — identical top-5 with distinct distances (`m-11:0.000000 m-72:0.000002 m-340:0.004358 …`), on the 0.5.1-built index and on a 0.8.6-built one.

**The iterative-scan change, proved by making it differ.** This surfaced in review after the cluster test, and it does not reproduce on a store as small as the one above, so it was exercised against a 20,000-row table in throwaway containers on the same two images. A `LIMIT 500` query at `hnsw.ef_search=40`, on the same data and the same forced HNSW index scan, returned **391 rows with `hnsw.iterative_scan=off`** and **500 rows with `strict_order`** — set distinctly rather than taken as a default, and reverted to confirm it goes back. That the GUC is genuinely registered rather than an accepted placeholder was checked separately: on 0.8.6, `SET hnsw.iterative_scan='bogus'` fails with `HINT: Available values: off, relaxed_order, strict_order`, and it does so on a database whose extension is still 0.5.1.

`EXPLAIN (ANALYZE)` on the `off` case, with the library loaded before the settings so nothing could be a placeholder, confirms the shape rather than assuming it:

```
 Limit (actual rows=391 loops=1)
   ->  Index Scan using t_v_idx on t (actual rows=391 loops=1)
         Order By: (v <=> '[0.5,...]'::vector)
```

with `SHOW hnsw.ef_search` reading `40` in the same session. That is what corrected the description: 391 is neither the `LIMIT` nor `ef_search`, so a scan with iterative mode off is not "capped at `ef_search` rows" — it makes one ground-layer pass and stops when that pass is exhausted, and how far it gets is a property of the graph.

**Why no API restart is needed — checked, because a review finding said the opposite.** `hindsight_api/engine/db/postgresql.py` caches any GUC the server rejects in a process-wide set whose comment says it is never cleared, which would mean an API pod that had already probed against 0.5.1 would skip the setting forever and the bump would do nothing until the Deployment rolled. It does not apply here, and the reason is testable: pgvector 0.5.1 does **not** reserve the `hnsw.` prefix, so `SELECT set_config('hnsw.iterative_scan', 'strict_order', false)` succeeds against it — verified against the old image, including after a vector operation forced the library to load, where `hnsw.ef_search` becomes visible at its 0.5.1 default of 40 while `hnsw.iterative_scan` and even a nonsense `hnsw.` name are still accepted. No `UndefinedObjectError` is ever raised, so nothing is cached. Postgres then applies the placeholder's value to the real variable when the defining library loads: issuing the three `set_config` calls first and the query afterwards — the pool's exact order, with no vector operation in between — gives 500 rows against 391 with the setting left alone.

**What this did not cover.** The upgrade half ran against a database whose schema the API built but which held synthetic vectors, not memories produced by real extraction — that needs an LLM-backed ingest that was not set up. The iterative-scan work was proved at the pgvector layer, in containers, and against Hindsight's source; it was not observed as a recall on a running install, because the gkedemos bank is empty and a recall against it would show nothing either way. So the retrieval-quality and latency consequences on a populated bank are argued, not measured. No mirrored install pulled the copy for real; the mirrored reference was verified by chart render and `mirror_images.sh --dry-run` only. One incidental observation: the API pod restarted once during the StatefulSet roll, a clean SIGTERM while the database was down, which would happen on any Postgres restart and is not caused by this change — confirmed from `--previous` logs rather than assumed.

Cleanup: namespace deleted, both test PVs reclaimed, the backing GCE disks gone, the shared install's pods and memory provider unchanged, lease released. The throwaway containers and Docker volumes were removed.

## Self-Review

`review-adversarial` ran once and `review-docs-drift` four times, each in a subagent that did not write the change, was handed only the diff range, and was told to derive intent from the diff and read no commit body. The mechanical gate ran separately and was handed to them as already-clean. Angles covered: image provenance and digest correctness against the live registry; whether pg15 → pg15 actually holds (the unstated precondition of the whole change); every rewritten rationale against a real upstream source; whether the rename breaks the mirror-naming invariant; whether any consumer still resolves the old name; and what the bump changes at runtime that the diff does not say. The last of those produced most of the list. Two successive passes found Blocking problems in prose the round before had asked for, which is worth saying plainly: the pin itself was right from the first commit, and everything since has been getting the explanation of it right. The last round's findings turned on a mental model — "an HNSW scan is capped at `ef_search` rows" — that Hindsight's own source comments state and that measurement does not support.

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | The swap moves the pgvector *library* to 0.8.6, but an existing database keeps extension 0.5.1 until `ALTER EXTENSION vector UPDATE` — nothing in the tree said so | **Fixed.** Documented in `memory.md`. Reached independently by both passes and settled empirically by the live test. |
| 2 | The library moves even when the extension does not, and Hindsight 0.9 defaults `HINDSIGHT_API_ANN_ITERATIVE_SCAN=true` — so the bump switches iterative ANN scans on for existing installs, silently | **Fixed and measured.** The most consequential finding here. `hindsight_api/config.py` sets `DEFAULT_ANN_ITERATIVE_SCAN = True`, and `engine/db/postgresql.py` probes for support by attempting the `SET` and catching `UndefinedObjectError` — it never reads the extension version, so support tracks the library. Documented, and demonstrated at 391 vs 500 rows above. |
| 3 | `hack/check-image-inventory.sh` still called `hindsight-postgresql` a "digest-only upstream"; the diff fixed the twin comment and missed this one | **Fixed.** |
| 4 | The pg15 rationale was stated in five places, and two pointers disagreed about which was canonical | **Fixed.** One home (`images.json`, per AGENTS.md's canonical-home table), everything else points at it. |
| 5 | `mirror_images.sh` says three entries pin `tag@sha256`; it is four | **Fixed**, pre-existing — see What Changed. |
| 6 | `memory.md` renamed the image inside a benchmark narrative measured on the old one | **Fixed**, but not as reported: the two passes disagreed, and reading it both were right — `max_connections = 100` is stock Postgres, not a property of either image, so naming any image there was the defect. |
| 7 | The recall-cost measurements in `memory.md` were taken under the 0.5.1 library, which could not run iterative scans | **Fixed**, though the first attempt at the caveat was itself wrong — see #12 for what it now says. |
| 8 | "silently change the database engine" is `latest`-era phrasing carried onto a version tag — the drift is now bounded to a Postgres minor | **Fixed** in four places. |
| 9 | `memory.md` hard-coded `0.8.6` twice, so the next patch bump would stale the design doc | **Fixed** — it names the pg15 line and defers the version to `images.json`. |
| 10 | `memory.md` could leave an operator thinking `ALTER EXTENSION vector UPDATE` is never appropriate, when pgvector's own upgrade procedure says to run it | **Fixed**, one clause. |
| 11 | Nothing in CI exercises the new image against a PVC written by the old one | **Not fixed**, deliberately. AGENTS.md puts "checks an install you already have" in `bench/cuj/`, manual by design because CI has no deployment to point at; a test elsewhere would either not run or not be testing this. Covered by Live validation instead. |
| 12 | **Blocking** — the caveat added for #7 claimed the new library hands the reranker more candidates and makes the recall-cost numbers "a floor", contradicting the four paragraphs beneath it, which establish that the reranker's input is bounded by the recall budget and that this bank already saturates it fivefold | **Fixed.** The caveat now says what actually changes — which units the retrieval stage returns, not how many pairs are scored — and stops claiming a direction for the latency. The finding was right and my clause was wrong. |
| 13 | **Blocking** — the API caches a rejected GUC in a process-wide set that is never cleared, so an existing install would need the API pod restarted before the bump took effect, making "on existing installs, no config change" false | **Rejected, with the reason tested, and the next pass agreed.** The cache is real and the code comment is accurate, but the path is never taken: pgvector only began reserving the `hnsw.` prefix in 0.6.0, so on 0.5.1 `set_config('hnsw.iterative_scan', …)` succeeds instead of raising, nothing is cached, and Postgres applies the placeholder to the real variable when the new library loads. Verified on both images in the pool's ordering; a later pass independently confirmed the 0.6.0 boundary in pgvector's source and PostgreSQL's `define_custom_variable`. Upstream's own comment on that `except` branch says an older pgvector "rejects it rather than accepting a placeholder" — true from 0.6.0, not of the 0.5.1 this install is on — so the bullet names the boundary to stop the next reader "correcting" it back. |
| 14 | `hnsw.max_scan_tuples` "(4,000)" reads as pgvector's default, which is 20,000 — 4,000 is Hindsight's | **Fixed.** Attributed, with both numbers and the fact that it is tunable. |
| 15 | The illustrative measurement was taken at `ef_search=40`, pgvector's default, when the shipped recall path uses 200 and `strict_order` | **Fixed.** The doc and this PR both now say the measurement is a synthetic bench at pgvector defaults, not the shipped configuration. |
| 16 | Iterative scan needs pgvector 0.8.0, a different floor from the 0.5.0 stated one paragraph earlier, and the two were easy to merge | **Fixed**, one clause naming both. |
| 17 | "an un-updated extension keeps working" was justified by Hindsight's feature floor, which does not address whether a 0.8.6 library can serve SQL objects declared at 0.5.1 | **Fixed.** The bullet now gives the reason that actually supports the conclusion. |
| 18 | The pin sits at Postgres 15, which is also Hindsight's minimum supported major, so a Hindsight release raising that floor forces the `pg_upgrade` the entry is written to defer | **Fixed** — added to the `images.json` rationale. Not drift; a gap in the reasoning. |
| 19 | **Blocking** — "the update gates SQL-level additions … which Hindsight does not use" is false: `vector--0.5.1--0.6.0.sql` sets `STORAGE = external` on the `vector` type, which applies to the column Hindsight writes | **Fixed.** The bullet now names it and sizes it honestly — at 384 dimensions the row rarely crosses the TOAST threshold, so the effect is small, but "nothing this deployment needs" was too strong. |
| 20 | **Blocking ×2** — with iterative scan off, `ef_search` is the ceiling on a single index scan, so the 391-row measurement at `ef_search=40` cannot be the shape described; and the same premise makes the recall-cost trace's `semantic=300` irreconcilable with a shipped `ef_search=200` | **Both fixed, by correcting my wording rather than the number.** `EXPLAIN (ANALYZE)` shows a genuine `Index Scan using t_v_idx` returning 391 rows with `SHOW hnsw.ef_search` reading `40` in the same session — so `ef_search` is not a row ceiling. The scan makes one ground-layer pass and ends when it drains. With that corrected, 300 rows at `ef_search=200` is unremarkable and the caveat no longer contradicts the trace beneath it. Hindsight's own comments use the "capped at `ef_search`" phrasing, so the doc now flags it as the wrong mental model. |
| 21 | "GUCs appear as soon as the image changes" is loose — they appear once a backend loads the new library | **Fixed**, naming the StatefulSet restart as what guarantees it. |

What the passes could not cover: none held cluster access or the lease, so all live testing was done separately in this context, and the last pass could not reproduce the 391/500 measurement for that reason; the mirrored path was checked by render and dry-run, never against a real registry; and pgvector's documentation says nothing either way about running an older extension version under a newer shared library, so that half of #1 rests on the CHANGELOG showing no breaking catalog change across the range plus the live observation, not on a statement from upstream. Two things raised and left open. Whether the gold-recall numbers in `## The experiment` need the same 0.5.1 caveat as the recall-cost section: they were taken on the same install, but whether iterative scan moves them depends on whether the planner used the HNSW index at all on a 1,414-document corpus, which needs an `EXPLAIN` on that install — so they are left uncaveated rather than given a caveat I cannot support. And a pre-existing docs defect this change did not cause and does not touch: the `docs/README.md` row for `memory.md` carries "#111-#116 open", which is the pull-request status the Documentation Guidelines forbid in docs. Flagging rather than fixing, because the map is edited from several branches a week and AGENTS.md is explicit about not rewriting rows a change did not author.

One structural suggestion not taken: the last pass observed that the two new bullets are long for a list whose other entries describe the StatefulSet's spec, and proposed a subsection. The bullets were tightened instead — a subsection would put upgrade semantics a scroll away from the image pin they qualify, which is where someone changing the tag will be looking.

## Risk & Rollout

The blast radius is the Hindsight Postgres pod, on installs that have memory enabled. Two changes arrive together and only one is visible in the diff. The first is the engine: Postgres 15.4 → 15.19 within the major, applied on the existing data directory, which the live test exercised end to end. The second is retrieval depth — the new library defines `hnsw.iterative_scan`, Hindsight is already asking for it on every pooled connection, and an ANN scan that would have stopped when its single ground-layer pass drained now resumes until the query's `LIMIT` is met. This takes effect on existing installs with no `ALTER EXTENSION`, no config change, and no API restart; the mechanism is in Live validation, because it is not obvious and one review pass concluded the opposite. The expected effect is on which units a recall retrieves rather than on its headline latency, since the reranker that dominates a recall is bounded by the recall budget and the design doc's own numbers show a bank that already saturates it — but that reasoning is from the source and the doc, not from a measured recall on a populated bank, which nothing available here could provide. `HINDSIGHT_API_ANN_ITERATIVE_SCAN=false` is upstream's kill switch if an install wants the old behaviour.

Reverting is a revert of this commit: pg15 → pg15 in both directions, and the extension is left at whatever version created the database unless an operator has run `ALTER EXTENSION vector UPDATE` — after which reverting the image would put a 0.8.6 catalog under the 0.5.1 library, which is not supported. Nothing in this repository runs that statement. Nothing has to happen at merge time; mirrored installs re-run `make mirror-images` as they would for any pin change.


dshnayder (approved):

bradhoekstra:
/hold

Temporary, no change requested: parking this PR out of the Tide pool for a few minutes so Tide merges overridden PRs one at a time instead of forming another multi-hour batch. Will `/hold cancel` and re-override shortly.

Issue #1171, opened 2026-09-02
compliance-rbac-overgrant: rung-4 collapse on unrelated PRs since 2026-09-02 ~02:00Z (fleet-audit delegation degraded) — demoted from BOOTSTRAP_ADMITTED


jayantid:
Caught-by: evals

## What is happening

`compliance-rbac-overgrant` — the fleet-audits canary (cron/SOP dispatch → delegation → token minting → gh ledger write) — is collapsing at rung 4 (all graded repetitions failing) on pull requests whose diffs cannot explain it. Six consecutive smoke runs across five unrelated PRs on 2026-09-02:

| PR | build (gs://kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/…/pull-kube-agents-smoke-test/…) | started (UTC) | verdict |
|----|----|----|----|
| #1148 | 1148/…/2095051967679172608 | 07:31 | **FAILED** (all 3 reps) |
| #1090 | 1090/…/2095063285538754560 | 08:16 | **FAILED** (all 3 reps) |
| #1145 | 1145/…/2095068341315047424 | 08:36 | **FAILED** (all 3 reps) |
| #1089 | 1089/…/2095070840138764288 | 08:46 | **FAILED** (rep 3 = empty record, rung 1-3 class) |
| #1148 | 1148/…/2095078977927385088 | 09:18 | UNSTABLE (1/3) |
| #1153 | 1153/…/2095101308846477312 | 10:47 | **FAILED** (all 3 reps) |
| #1148 | 1148/…/2095104535600041984 | 11:00 | UNSTABLE (2/3) |

**#1153 is red solely because of this case** (every other case in that build is UNSTABLE or PASSED). Its diff is `forge.py` auth retries — unrelated. With the presubmit gate merge-blocking as of today ~14:00Z, that is an innocent PR blocked by a case-level regression, which is exactly the demote-same-day condition written above `BOOTSTRAP_ADMITTED` in `hack/ci-eval-pr.sh`.

## Failure modes (rep-level evidence)

Three distinct modes, all downstream of the same degradation in the fleet-audit delegation chain:

**1. Degraded audit — “access limitations”, planted finding never checked.**
The delegated audit cannot reach the seeded clusters’ control planes with kubectl, falls back to raw `curl` against the public endpoint, and publishes a partial ledger in which check 2.4 (`cluster-admin` bindings) never ran — so the planted `debug-binding` finding is absent and the verifier reds:
[...]

PR #1183, opened 2026-09-02
fix(proxy): survive profile homes the credential sidecar cannot write, and widen them at scaffold

Fuxiao-Gao:
## Summary

Root-cause fix for the fleet-audit delegation degradation tracked in #1171. `hermes profile create` leaves a profile home 0700 and sandbox-owned; since the #955 UID split the credential sidecar reaches the data PVC only through the shared fsGroup, so it gets EACCES writing the kubeconfig pin into cluster profile homes and stat-ing `profiles/platform` for cwd containment. The proxy then turned the failed pin copy into a failed command even though `gcloud container clusters get-credentials` had succeeded, and every fleet audit degraded from there. This PR widens profile homes for the sidecar at scaffold time, repairs homes already sitting 0700 on a PVC at startup, and makes the pin copy-out non-fatal.

## Why This Change

Since ~2026-09-01, `compliance-rbac-overgrant` collapsed at rung 4 on PRs whose diffs cannot explain it, and #1173 demoted it from `BOOTSTRAP_ADMITTED`. The `envoy-credential-proxy.log` artifacts in the failing builds show the failure was not control-plane reachability: the credential fetch succeeded and the request died on `PermissionError` at the pin copy-out in `_execute_get_credentials` — build `2095068341315047424` shows it for all four clusters including `platform-agent-host`, and build `2095070840138764288` shows 21× EACCES on `/opt/data/profiles/platform` itself. Workers then fell back to `kubectl config use-context` (always refused by the read-only policy) and raw `curl`, and audits published partial ledgers or nothing. Without this fix the degraded eval projects keep failing every audit-dependent case and #1171's re-admission bar cannot be met.

## What Changed

- `agents/platform/scripts/credential_proxy.py` — an `OSError` writing the pin copy is logged and the successful result returned; the pin is an inspection artefact, and the managed kubeconfig commands actually run against is already filed. The warning states what still works and what does not.
- `agents/platform/scripts/profile_scaffold.py` — `ensure_profile` widens the home to `0o2770` on every call (group rwx for the fsGroup, setgid, others still excluded), repairing pre-fix homes whenever a scaffold revisits them.
- `deploy/shared/docker-entrypoint.sh` — startup repairs homes already 0700 on the PVC, since the platform scaffold is gated on `profile.yaml` and never reruns.
- Docs: the credential-isolation design doc and its site summary now state the copy-out is best-effort, and the design doc's UID-split bullet names the widening beside the `umask 0002` discipline it backstops.
- Five new tests across the three touched suites.

## Context

- #1171 carries the failure evidence and the re-admission bar; this PR addresses its "root-caused and fixed" condition but does not close it (re-admission of the case is a separate step). #1173 is the same-day demotion.
- #955 introduced the UID split; b3c3cca (#1053) fixed the same failure class for the ledger body file. #871 and #966 put profile homes on the audit hot path, which matches the onset timing.
- The agent-filed netpol egress diagnosis (gke-agentic/kube-agents-evals-13-infra#4) was a misdiagnosis: the fetches reached the GKE API fine and failed on the local write.
- Open #913 reworks the sandbox/sidecar arrangement and touches both `docker-entrypoint.sh` and `credential_proxy.py`; whichever merges second should re-validate the widening (noted under Risk).

## Testing

- `test_credential_proxy` (220), `test_profile_scaffold` (31), `tests.test_docker_entrypoint` (60) all pass at HEAD.
- Red-then-green: with the three source hunks reverted in a throwaway worktree, the new tests fail (1 failure + 3 errors in the scripts suites, 1 failure in the entrypoint suite). Run once by the adversarial pass and re-run by the author.
- `make docs-check` passes; prettier 3.9.6 over the changed Markdown made no changes; `sh -n` on the entrypoint.

### Live validation

Not live-tested against a UID-split install: the only installation available to the author predates #955 — `kubectl exec ... id` shows its sidecar also runs as uid 10000, so the EACCES cannot reproduce there. What that install did provide: `stat` on its data PVC shows every profile home is `drwx------ 10000:10000`, confirming the 0700 mode this PR widens comes from `hermes profile create` (no repo script sets it). The cross-UID failure itself is evidenced by the failing builds' proxy-log tracebacks through `_execute_get_credentials`. This PR's own smoke run is the end-to-end check: it deploys this branch into a leased eval project, and the degraded projects' PVCs still carry 0700 homes the entrypoint repair should heal — the observables are no `PermissionError` in `envoy-credential-proxy.log` and `compliance-rbac-overgrant` (held out, still reporting) completing its checks. Nothing was mutated on the inspected install and no artifacts were left behind.

## Self-Review

Both passes ran in fresh subagent contexts handed the diff range and the skill file only, not the author's session or reasoning.

Adversarial pass (all ten angles):

1. **Fixed** — the copy-out's comment and warning overclaimed ("commands that follow still reach the cluster"): `_resolve_kubeconfig` reads the pin to learn the target, so a Cluster Agent profile whose home refuses even the chmod comes up with `.env` pinned at a missing file, surfaced only by the preflight's missing-pin check. Both now state what works (the fetch, the managed copy) and what does not (resolution through this pin). Open question deliberately left rather than coded around: whether the scaffold path should hard-fail when the pin cannot be written. After the other two hunks the trigger is a home that refuses the chmod itself, the preflight does surface it, and a hard-fail belongs with rethinking that path, not with this fix.
2. **Fixed** — `PROFILE_HOME_MODE` was declared mid-file against the no-magic-constants placement rule; moved to the top of the file.

Reviewed and left alone, for the record: 0700→2770 widens nothing beyond the pod (both containers already run `RunAsGroup = FSGroup = 10000`; others stay excluded); the entrypoint's literal `chmod 2770` matches the pre-existing scratch repair and shell cannot import the Python constant; the root-skip guard on the EACCES test is required because root ignores mode bits. Stated gaps of the pass: cross-UID behavior is not exercisable in single-uid unit tests (hence the smoke run above); whether hermes itself tolerates a setgid group-writable home; only the home's top level is widened — nothing in #1171 cites deeper 0700 subdirectories.

Docs-drift pass: one blocking finding, **fixed** — the credential-isolation design doc and site page promised the pin copy "still exists" unconditionally; both now say the copy-out is best-effort. Two advisories, **fixed** — the design doc's UID-split bullet now names the widening, and `docker-images.md`'s claim that host-side entrypoint tests "can only prove which branch was taken" now excepts the volume-repair steps. `make docs-check` passes.

## Risk & Rollout

The behavior change is confined to the failure path: a pin copy-out that previously failed the request now logs and succeeds, and profile homes gain group access only within the pod's own fsGroup. New failure mode to know about: a profile whose pin cannot be written is now discovered at Cluster Agent preflight instead of at scaffold time, with the sidecar warning as the record. Revert is a clean revert of one commit; nothing has to happen at merge time. Merge-order note: #913 touches the same arrangement, and whichever lands second should re-run the entrypoint and proxy suites.


dshnayder (approved):
PR#913 will rework the credentials proxy. Approving this one as an intermediate fix.

Fuxiao-Gao:
Thanks for the review and the heads-up. Agreed on the framing — this is the intermediate shape: it stops the eval-project bleeding now, and #913 decides the final arrangement (the overlap is also noted in the Risk section).

Two facts worth carrying into the rework, since they outlive this code: `hermes profile create` makes a profile home 0700 regardless of the entrypoint's umask, and the eval-project PVCs hold poisoned 0700 homes until a deploy with the startup repair reaches them. If the new arrangement keeps any cross-UID write into a profile home, the five tests added here (or equivalents) are the regression net — happy to help port them when #913 rebases over this.

Issue #1179, opened 2026-09-02
eval presubmit: a comment-only push re-runs the 2-hour merge-blocking job and aborts the one in flight


stalhaali:
A push that changes only a comment re-runs `pull-kube-agents-smoke-test` in full, and aborts the run already in flight. The job is merge-blocking and took 123 minutes on its last clean pass, so the cost of editing a comment on a reviewed pull request is a two-hour round trip plus one of the fifteen Boskos leases for the duration.

## Evidence

[#1127](https://github.com/gke-labs/kube-agents/pull/1127) — the change under review is a bench phrase list, a test, and documentation. Its fourth commit, [`9715b3d`](https://github.com/gke-labs/kube-agents/commit/9715b3df), was written in response to a review finding and touches nothing but a comment block in `hack/ci-eval-pr.sh`: it repoints a held-out case's roster entry from #1100 to #1010. `git diff --numstat` is `7 4 hack/ci-eval-pr.sh`, and filtering the hunk to lines that do not begin with `#` leaves **zero**: not one shell statement changes.

Four builds of the job, from the status history on the pull request's two live heads:

| build | head | base SHA | outcome | wall clock |
| --- | --- | --- | --- | --- |
| [2094891450637488128](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1127/pull-kube-agents-smoke-test/2094891450637488128) | `fa02bfb4` | — | infra failure, zero repetitions ran (#1006) | — |
| [2095140696485793792](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1127/pull-kube-agents-smoke-test/2095140696485793792) | `50e0f44f` | `42c6e486` | **success** | 13:24:03 → 15:27:23, **123 min** |
| [2095196649893662720](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1127/pull-kube-agents-smoke-test/2095196649893662720) | `50e0f44f` | `ae55c6bd` | `failure` — "Aborted by trigger plugin" | 17:06:23 → 17:12:25, killed at 6 min |
[...]

PR #1186, opened 2026-09-02
feat(ci): reuse a PR's green eval verdict when only inert paths changed since

jayantid:
## Summary

Two changes that compose, both kube-agents-side only. First, `hack/ci-eval-pr.sh` gains a step 0: before any cluster work, it looks up this pull request's newest green `pull-kube-agents-smoke-test` build in the GCS job history, recovers that build's head and base SHAs from its `started.json` clone record, and — when that build is attested by its Prow-posted GitHub success status and everything that changed since, on the PR side and on main's side, matches a script-owned inert-path list — reuses the green verdict and exits 0 instead of re-running the eval matrix. Second, the long-form `BOOTSTRAP_ADMITTED` roster prose moves out of the script into `docs/eval-gate-roster.md`, so the next review finding against that prose is itself an inert-path edit.

## Why This Change

#1179 measured the cost: a push touching nothing but a comment block in `hack/ci-eval-pr.sh` (#1127's `9715b3d`, zero shell statements changed) aborted the run in flight and re-ran the job in full — the prior clean pass took 123 minutes, holding one of the fifteen Boskos leases the pool is saturated on. The same table shows a second waste class: an identical tree re-graded because `main` moved by unrelated commits. Prow's `skip_if_only_changed` cannot catch either, because it sees the PR's whole diff against the base, never the delta since the last green run. And the roster prose in the script is why the recurring trigger exists at all: it is review surface, and answering a finding against it cost a full run.

## What Changed

- **Step 0 in `hack/ci-eval-pr.sh`.** Reads `PULL_NUMBER` / `PULL_PULL_SHA` / `PULL_BASE_SHA` / `PULL_BASE_REF` from the Prow decoration; finds the newest `passed: true` build under `gs://kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/<PR>/pull-kube-agents-smoke-test/*/finished.json`; parses that build's `started.json` `repos` record (`main:<base_sha>,<pr>:<head_sha>`); diffs `prev_head..cur_head` and `prev_base..cur_base` with `--no-renames` (so a rename out of a gated path cannot hide its deletion); and skips only when every file in both deltas matches `REVALIDATION_INERT_PATHS`. Fail-closed on every doubt — no history, unreadable GCS, unparsable records, a commit the checkout lacks, one escaping file — each is one logged reason and a full run. First run on a PR is always full. `EVAL_SKIP_REVALIDATION=1` forces a full run for debugging. On reuse it logs `REVALIDATED against green build <id>`, the Spyglass URL, both delta file lists, and the predicate.
- **The inert list is a script-owned constant** seeded to today's `skip_if_only_changed` (`^docs/`, `^\.github/`, `^examples/`, root-level `*.md`, `LICENSE`/`OWNERS`/`OWNERS_ALIASES`, all root-anchored). It may drift STRICTER than the yaml's and deliberately lives here: the worst case of divergence is an unnecessary full run, never a wrongly skipped one. Cross-referenced to `prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml` in oss-test-infra.
- **What it saves, honestly:** in the current job the Boskos lease and `ci-deploy.sh` run in the Prow wrapper BEFORE this script, so a revalidated run still pays lease + image build + deploy + teardown (~20–30 min of pool time) — what it skips is the ~2 h eval span that dominates. Hoisting the check ahead of the lease would be an oss-test-infra change; #1179's comment scoped this to kube-agents only.
- **`docs/eval-gate-roster.md`** now carries the roster prose (admission bar, hold-outs with exit issues, rung scoping, demotion protocol), registered in `docs/README.md`'s map and tree. The script keeps a short comment: what the variable is, the comma-or-whitespace parsing contract (`_bootstrap_admitted()` in `bench/kube_agents_bench/gate.py`), the pointer, and the one-line demotion instruction. The moved prose was also corrected: the old comment said "thirteen of the seventeen active cases"; `TASKS` holds eighteen uncommented entries (post-#1218, which moved the two tofu cases to the nightly tier), and after the #1173, #1190, and #1189 demotions (all merged on main and reconciled into this branch) ten are admitted. `AGENTS.md`, `docs/designs/testing-strategy.md`, `bench/README.md`, and `scripts/eval_dashboard/SCHEMA.md` were repointed or given a one-clause step-0 exception where their descriptions went conditionally false.
- **Dashboard collector:** a revalidated run's build log carries no `Task ... Result:` lines and no final-verdict line. `scripts/eval_dashboard/collect.py` already tolerates that shape (aborted runs produce taskless builds today; `parse_build_log` never raises, `duration_s` falls back to `finished − started`) — verified by reading the parser, and SCHEMA.md now names the second cause of a verdict-less `SUCCESS` build.

## Context

- Implements the self-revalidation option proposed on #1179 (the issue asked the config owners to choose among options, so this does not claim to close it — it implements the fifth shape from the issue's comment thread).
- Competing-work check (2026-09-02): no open PR implements #1179 — searched open PRs for revalidation/#1179 references; none found. Related but different: #1112 selects eval tasks from changed files with a gate backstop — it shrinks the matrix per-PR, this change skips re-runs against a PR's own green history; they compose, but both edit `hack/ci-eval-pr.sh` and each carries a path-relevance list, so whoever merges second reconciles (converging the two lists into one source is a reasonable follow-up). #1180 touches `ci-deploy.sh`/`ci-teardown.sh` only — no overlap. #1173 and #1190 touched the roster block this PR moves; both merged first and this branch has merged and reconciled them (the doc carries the compliance-rbac-overgrant demotion, #1171, and the rca-remediation-pr demotion, #1189).
- Trust surface, stated precisely. Self case: subsumption — a PR's own code already IS the job (it can force its own context green by editing this script to `exit 0`), so revalidation adds no capability there, and editing the revalidation logic touches `hack/`, which is not inert, so that PR's own run goes full. Cross-PR case (kube-agents-bot's review finding, fixed in 3ac1581c): "a PR has no write path to the bucket" is NOT claimed — the test container may share the artifact-uploading pod identity, and the bucket ACL is not measurable from here — so the GCS records are never trusted alone. A candidate green build counts only when GitHub holds a `pull-kube-agents-smoke-test` success status event on the recovered head whose `target_url` names that same build id: posted by `google-oss-prow[bot]` with repository write permission no pull request holds, append-only per build (verified on #1127's head `50e0f44f`, where the earlier build's success event survives the later aborted failure). `finished.json`'s `revision` must also equal the `started.json` head, and both SHAs must be 40-hex before any git command sees them.

## Testing

- `tests/test_ci_eval_revalidation.py` — 22 tests, following the `scripts/test_ci_eval_trap.py` / `tests/test_ci_eval_ledger_mint.py` pattern (function and constants extracted from the shipped script, executed under `set -euo pipefail` with a stub `gsutil` and a fixture git repository standing in for the decorated checkout). The reuse path: green history + inert deltas fires; identical SHAs (empty deltas) are trivially inert; newest green wins and the build-id sort is proven numeric, not lexicographic; the `REVALIDATED against green build <id>` line shape and Spyglass URL are pinned; the gsutil call log pins that the history read is scoped to this PR's own job path. The fall-through matrix — every path runs full: no history; no green build; unparsable `finished.json`; missing `started.json`; `started.json` without the SHAs; non-inert file in the head delta; non-inert file in the base delta; rename of a non-inert file to an inert destination (`git mv code.py docs/moved.md`); root-anchoring traps (`docs-evil.go`, `sub/notes.md`, `bench/OWNERS`); a forged green record pair with no GitHub status behind it; a success status whose `target_url` names a different build; a malformed (non-40-hex) SHA in `started.json`; a `finished.json` revision disagreeing with the `started.json` head; missing git object; `EVAL_SKIP_REVALIDATION=1`; not a decorated presubmit. Plus a placement test pinning that step 0 precedes `ci-env.sh` sourcing, the EXIT trap, and cluster auth.
- `bash -n hack/ci-eval-pr.sh` clean.
- `scripts/` discovery suite: 1097 tests OK (after the 2026-09-08 merge of `main`). `tests/test_ci_eval*.py`: 54 tests OK, including this PR's 22 in `test_ci_eval_revalidation.py` and #1280's 7 in `test_ci_eval_seeded_a_heal.py` side by side — the two lift different blocks out of `hack/ci-eval-pr.sh` (function-name regexes here, the `# ─── 2c.` section marker there), so neither extraction reaches into the other's code. `scripts/validate_bench_cases.py`: 31 cases OK.
- `make docs-check` green (generated regions, links, terminology, map, context budget). `main` now sits within 5 chars of the 38,000-char budget, so the AGENTS.md step-0 clause is the shortest form that still names the mechanism — "a push restarts it unless only inert paths changed (step 0)" — the roster pointer is a bare path, and the redundant "rather than stacking pushes" tail was dropped; 10 chars under after the 2026-09-08 merge.
- Prettier could not run locally (no node on this machine); the new/edited Markdown follows the repo's conventions and CI's `prettier` check is the verifier.

### Live validation

Not live-tested, and here is exactly what that means. This change cannot reach a running kube-agents installation — it never touches a cluster; its entire surface is the Prow pod's environment, GCS records, and the decorated checkout. The unit tests stub GCS and git. What WAS verified against production artifacts: the record shapes and env contract were read from a real green build (PR #1127, build 2095140696485793792) — `finished.json`'s `passed`/`revision` fields, `started.json`'s `repos["gke-labs/kube-agents"]` format, and `podinfo.json`'s container env confirming `PULL_NUMBER`, `PULL_PULL_SHA`, `PULL_BASE_SHA`, `PULL_BASE_REF` are exported to the test container; the live Prow yaml was fetched to confirm the wrapper's lease/deploy ordering and that the inert list matches today's `skip_if_only_changed` character-for-character. The field test is structurally the first comment-only push to an already-green PR after this merges: its build log either says `REVALIDATED against green build <id>` and the job ends in minutes, or says `Step 0: full run: <reason>` and nothing regressed. Rollback short of revert is `EVAL_SKIP_REVALIDATION=1` in the job env.

## Self-Review

Both required passes ran in fresh subagent contexts handed the diff range and nothing else (neither wrote the change): the adversarial pass per `.agents/skills/review-adversarial/SKILL.md`, the docs-drift pass per `.agents/skills/review-docs-drift/SKILL.md`. Findings and dispositions, one list:

- **Rename hole (adversarial, blocking — confirmed by reproduction): FIXED.** `git diff --name-only` detects renames by default, so `git mv hack/tool.sh docs/moved.md` listed only the inert destination and the non-inert deletion rode a reused green. Both diffs now pass `--no-renames`, with a comment and a regression test (`test_a_rename_to_an_inert_path_is_a_full_run`).
- **gsutil stub never checked which URL was read (adversarial, should-fix): FIXED.** A wrong PR number or job name — reusing a foreign verdict — would have passed every test. The stub now logs each call and `test_the_history_read_is_scoped_to_this_prs_own_job` pins the exact `gs://…/pull/gke-labs_kube-agents/<PR>/pull-kube-agents-smoke-test/…` paths.
- **Phantom consumer claim (adversarial, should-fix): FIXED.** The test docstring said the dashboard collector greps the REVALIDATED line; `collect.py` reads nothing from it today. Docstrings now say the line is pinned for humans and for future collector support.
- **Newest-green-wins overrides a newer red at inert distance (adversarial, question): DOCUMENTED as deliberate.** An inert delta cannot feed the eval, so the reuse renders the same judgement a passing `/retest` would; the header comment now states the asymmetry and the two levers (non-inert push, `EVAL_SKIP_REVALIDATION=1`).
- **#1112 overlap (adversarial, question): REPORTED in Context**, not resolved here — merge-order reconciliation plus a possible follow-up converging the two path lists.
- **python3 not preflighted / `main` fallback literal (adversarial, nits): FIXED** — preflight with its own reason line; `REVALIDATION_DEFAULT_BASE_REF` named per the no-magic-constants rule.
- **Cross-PR record forgery (kube-agents-bot, high — confirmed): FIXED in 3ac1581c.** The bot showed the subsumption argument covers only the self case: a hostile PR's run could plant a fabricated green record pair under a victim PR's history path and suppress the victim's gate on its next inert push. The bucket ACL could not be measured from here (`getIamPolicy` denied), so the fix takes the stronger close the bot named: reuse now requires the Prow-posted GitHub success status event binding head and build id (which no PR can post), plus the record-consistency and 40-hex bindings, all fail-closed, all regression-tested with the bot's attack modelled directly.
- **Empty-delta edge and chaining (self-directed + adversarial): verified sound.** Identical SHAs mean a byte-identical tree to the one graded — correct reuse, tested. A revalidated run writes its own `passed: true` with its own `started.json`, and transitive reuse is sound because a file differing across a chain of inert links must differ in some link; the rename fix closes the one hole the pass found in that argument.
- **Verified-clean angles (adversarial):** regex root-anchoring (by execution: `docs-evil.go`, `sub/notes.md`, `bench/OWNERS`, `examplesX/foo` all fall through); errexit-suppressed control flow has no silently-fail-open path; force-push, retargeted-PR, and Tide-batch scenarios fail closed (batch runs log under `pr-logs/pull/batch` and lack `PULL_NUMBER`); red and aborted runs write `passed: false`.
- **Stale job descriptions (docs-drift, blocking): FIXED.** AGENTS.md's gate section and testing-strategy §4.2 said the job always runs the full matrix; both now carry the step-0 exception clause. `bench/README.md` and `scripts/eval_dashboard/SCHEMA.md` got the same one-clause treatment (advisory findings).
- **Map tree line (docs-drift, blocking): FIXED** — `eval-gate-roster.md` added to the standalone-docs tree entry. Deliberately NOT fixed: the same line's pre-existing omissions (`testing-map.md`, `chatops/`, `samples/`) — the map's own rule is one line per PR and no rewriting of rows this PR did not author.
- **Roster doc asserted the Prow filter's behaviour as fact (docs-drift, advisory): FIXED** with a hedge matching the script's "may be stricter" framing — that config lives in another repository.
- **Duplicated count in testing-strategy (docs-drift, advisory): KEPT** — the sentence pre-dates this PR; it was corrected to twelve (post-#1173), then eleven (post-#1190), now ten of eighteen (post-#1189/#1218), since this PR already edits it. The roster doc's copy carries an explicit recount instruction. `docs/designs/eval-scorer.md`'s "seventeen tasks" cost arithmetic is pre-existing historical pricing with its own recount warnings — untouched.
- **Docs accuracy (docs-drift, verified clean):** the doc's counts, hold-out set, and issue numbers were re-derived from the branch's `TASKS` array and roster (20 active / 12 admitted at review time, 11 after the #1190 merge / the complement enumerated), `_bootstrap_admitted()`'s parsing contract and `grade_case`'s admission-blind rungs 1–3 confirmed against source, all links resolve, and no other reference to the roster prose living in the script survives anywhere in the tree.

## Risk & Rollout

Blast radius is the presubmit job only — no runtime, chart, or operator paths. The design is fail-closed, so the cheap failure mode is a full run that could have been skipped (the status quo). The expensive failure mode — a wrong skip — is bounded by the predicate (both deltas, root-anchored list, `--no-renames`, fail-closed parsing) and is what the 18-test matrix pins. Revert is one squash commit; `EVAL_SKIP_REVALIDATION=1` in the job env disables the behaviour without a code change. Merge-order note: this file is hot this week — #1173, #1190, #1218 and #1280 (section 2c, the seeded-a pool heal) are already reconciled into this branch via merge commits, and if #1112 merges first the step-0/trigger hunks will need a manual merge (and vice versa); whoever merges second reconciles.

Generated with the help of Claude Code (Fable 5).


hack/ci-eval-pr.sh
```diff
@@ -22,6 +22,214 @@
...
+      break
+    fi
+  done <<EOF_REVALIDATION_CANDIDATES
+${candidates}
+EOF_REVALIDATION_CANDIDATES
+  if [ -z "${prev_green}" ]; then
+    echo "Step 0: full run: no green build among the newest ${REVALIDATION_HISTORY_LIMIT} ${REVALIDATION_JOB_NAME} builds for PR #${PULL_NUMBER}"
+    return 1
+  fi
+
```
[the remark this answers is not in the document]
jayantid: Fixed in 3ac1581c, taking the stronger of your two closes since the ACL is not measurable from this side (`storage.buckets.getIamPolicy` denied; nothing in-repo states what the pod utilities upload as). The GCS records are no longer trusted alone: a candidate build counts only when GitHub holds a `pull-kube-agents-smoke-test` **success status event on the recovered head whose `target_url` names that same build id**. Those events are posted by `google-oss-prow[bot]` with repository write permission no pull request holds, and they are append-only per build — verified against #1127's head `50e0f44f`, where the earlier build's success event survives the later aborted failure, so the row-2/3 reuse case still works. Your minimum bindings are in too: `finished.json` `revision` must equal the `started.json` head, and both SHAs must be 40-hex before any git command sees them (which also closes argument smuggling via a forged record). The attestation read fails closed — unreadable statuses endpoint, empty list, wrong build in the `target_url` all log a reason and run full. Regression tests model your attack directly: a fabricated green pair under the victim path with no status behind it (`test_a_forged_green_record_without_a_github_status_is_a_full_run`), a success status borrowed from a different build, the malformed-SHA record, and the disagreeing record pair. The trust-surface comment in the script and the PR body now state the cross-PR case explicitly instead of resting on subsumption.

dshnayder (approved):

bradhoekstra:
Heads-up on a number this PR carries forward. After #1190 merged tonight (21:41Z, removing `rca-remediation-pr` from `BOOTSTRAP_ADMITTED`) the roster is **eleven**, not twelve.

This PR is based on `3c0b747e`, which is pre-#1190, so "Twelve" was correct when you wrote it. Two places will land one off after a rebase:

- `hack/ci-eval-pr.sh`, the new "The admission bar, and who clears it" section — "Twelve of the twenty active cases are admitted".
- `docs/designs/testing-strategy.md` — "Twelve of the twenty are bootstrap-admitted" (this PR already corrects it from the older "Thirteen").

Current `main` parses to eleven:

```
$ git show upstream/main:hack/ci-eval-pr.sh \
    | grep -E '^export BOOTSTRAP_ADMITTED=' | tr ',' '\n' | grep -c .
11
```

Prose only — this PR doesn't touch the `export BOOTSTRAP_ADMITTED` line itself.

Your parenthetical is the right instinct, incidentally: "recount the uncommented entries in the script's `TASKS` array rather than trusting this sentence — an earlier copy of it miscounted twice". This would be the third time. #1112 takes the other route and parses the roster programmatically; if the sentence keeps drifting, dropping the absolute count in favour of the pointer may be worth doing here.

Flagging rather than opening a competing one-line PR, since you're already rewriting both lines.

bradhoekstra:
/hold

Temporary, no change requested: parking this PR out of the Tide pool for about 30 minutes so Tide merges overridden PRs one at a time instead of forming another multi-hour batch. bradhoekstra will `/hold cancel` shortly.

bradhoekstra:
/hold cancel

Done: the overridden PRs have merged. Thanks for the patience.

AGENTS.md
```diff
@@ -354,7 +354,8 @@ three repetitions each — and has been merge-blocking since 2026-09-02
 against a 360-minute ceiling — and a new push restarts it, so open the pull request early and
 batch changes rather than stacking pushes. Another pull request merging usually does not — the
 green status is re-pinned to `main`'s new head
-([how a change merges](docs/pull-request-workflow.md#how-a-change-merges)).
+([how a change merges](docs/pull-request-workflow.md#how-a-change-merges)); a push whose whole
+delta is inert paths (`REVALIDATION_INERT_PATHS`) reuses the prior green via step 0.
```
jayantid: 🟡 **Not merge-blocking — a merge-order hazard to know about.** This PR's AGENTS.md addition leaves 41 characters of context-budget headroom, so any AGENTS.md growth on main before this merges reds `docs-check-context-budget` here again.

The #1224 headroom bump (39k) closed unmerged; #1222 fixed main by trimming instead, and main now sits ~200 chars under the 38,000 budget. With this sentence added the branch measures 37,959 chars. If anything lands on main that adds even a short clause to AGENTS.md, this PR needs another merge-and-trim round before it can go green.

No change requested here — the sentence is already at its terse form ("via step 0", no "the job's"). The durable fix is #1223's structural question. If this sits unmergeable behind another budget red, that discussion is the place to push.

jayantid: Disposition: informational, no code change. The clause is already minimal and the budget check passes at dc6d54b8 (37,959/38,000 chars, verified with make docs-check-context-budget, exit 0). Watch item for whoever merges: if AGENTS.md grows on main first, this branch needs one more merge before lgtm.

jayantid:
/retest

Build 2095881261376606208 red on environment, not this diff: seeded-fleet-reader impersonation PERMISSION_DENIED on leased kube-agents-evals-9, Boskos heartbeat/release 401s, reps pinned at the 45-min delegation timeout, exit 143 at the 360m ceiling. (Step 0 behaved correctly on this build: it logged the non-inert delta and chose a full run.) Evidence in #1249. Single retest per the AGENTS.md protocol.

PR #1089, opened 2026-09-01
fix(chat): validate and sanitize multiuser memory entries against prompt injection

sergiu1spiridon:
## Summary

Adds input validation, token neutralization, control/zero-width character stripping, entry size/count limits, and index-based replace/remove targeting to `MultiUserFileMemoryProvider` while keeping storage on disk non-destructive and preserving markdown headings during read operations.

## Why This Change

`MultiUserFileMemoryProvider` previously accepted raw strings for `add` and `replace` without validating length or neutralizing special tokens. This exposed the memory store to prompt injection markers (e.g., `<|im_start|>`, `[INST]`, `<<SYS>>`, `### System:`), internal delimiter smuggling (`\n§\n`), ANSI escape sequence corruption, and unbounded memory file growth. Furthermore, `action: "read"` stripped markdown headings, causing models executing read-modify-replace workflows to destroy runbook structure on disk.

## What Changed

- [agents/chat/plugins/memory/multiuser_memory/__init__.py](file:///usr/local/google/home/sergiuspiridon/kube-agents/agents/chat/plugins/memory/multiuser_memory/__init__.py):
  - Added `sanitize_memory_entry` (for new entry validation on storage) and `sanitize_for_prompt` (in-memory prompt and read tool rendering).
  - Stripped C0/C1 control characters, ANSI escape codes, zero-width characters (e.g. U+200B, U+FEFF), and bidi override controls (U+202A–U+202E).
  - Neutralized prompt injection delimiters (`<system>`, `<instruction>`, `<prompt>`, etc.), LLM special tokens (`<|im_start|>`, `<|im_end|>`, `[INST]`, `<<SYS>>`), and persistence delimiter smuggling (`\n§\n` -> `\n;\n`).
  - Generalized boundary tag neutralization to defuse closing (`</ system>`, `< / system>`), self-closing (`<system/>`, `<system />`), spaced (`< system>`), and attribute-bearing (`< system role="admin">`) variants with `<`-excluding scan classes (`(?:[^><\n]|<(?![^\S\n]*/?[^\S\n]*(?:system|instruction|prompt|admin|untrusted_)))*>`), stopping each candidate at the next tag candidate to guarantee linear evaluation without capping tag length or failing on embedded brackets (`<system role="<">`).
  - Replaced ASCII `[ \t]` with `[^\S\n]` for intra-tag whitespace, defusing tags containing non-ASCII horizontal whitespace (e.g. U+00A0 no-break space, U+3000 ideographic space) without crossing newlines.
[...]

PR #1196, opened 2026-09-03
fix(ci): heartbeat the Boskos lease through teardown, and report teardown step failures truthfully

haoxuw:
## Summary

`hack/ci-teardown.sh` now keeps the job's Boskos lease alive for exactly as long as it runs, and reports each step's real outcome — a failed step prints `✗` and lands in a final `FINISHED WITH N FAILED STEP(S)` summary instead of being masked by an unconditional ✓. A new `hack/boskos_heartbeat.sh` daemon carries the liveness signal on its own channel (the Boskos `/update` call), adding ~3 lines to the job log instead of one per beat. The branch is rebased over #1185 and layers onto its release-record fallback and gated CRD delete.

## Why This Change

The Prow wrapper kills its `boskosctl heartbeat` before invoking teardown, so every teardown runs lease-blind against Boskos's ~5-minute reaper. On 2026-09-01, a 546s CRD delete against an unreachable control plane on `kube-agents-evals-4` outlived that window: the reaper reclaimed the project mid-teardown, the pool re-leased it to a concurrent build (build 2094805894989615104 leased and released it while the first job was still deleting), and the original job's release failed with `401 Unauthorized` (Boskos `OwnerNotMatch`). The same teardown printed `✓ CRD deletion finished in 546s` over the failure and silently skipped the ValidatingAdmissionPolicy sweep when API discovery failed — the #1006 leftover-resources failure mode with a green checkmark on it.

## What Changed

- `hack/boskos_heartbeat.sh` (new): POSTs the Boskos client's `/update` (same `name`/`owner`/`state` query `boskosctl` sends, verified against `sigs.k8s.io/boskos` client source) every 30s. Job log gets start/transition/stop lines only; per-beat detail goes to `${ARTIFACTS}/boskos-heartbeat.log`. A 401 on a beat is reported once per transition — the lease-lost signal that previously surfaced only as a failed release at job end. At the ~5m reaper window, 30s beats tolerate 9 consecutive misses.
- `hack/ci-teardown.sh`: starts the daemon for the script's lifetime and kills it from the EXIT trap **before** the wrapper's release runs, so the orphaned-heartbeat problem the wrapper's `cleanup()` defends against cannot return. Inside Prow the lease identity is derived from what the pod already has — `PROJECT_ID` (exported by the wrapper), owner `${JOB_NAME}-${BUILD_ID}` (the wrapper's own convention), and the fixed in-cluster Boskos URL — so no oss-test-infra change is needed to arm it; explicit `BOSKOS_*` env overrides all three, and outside Prow the daemon disables itself with one line.
- `hack/ci-teardown.sh`: truthful step accounting via `finish_step` — layered over #1185's steps without changing their logic: Step 1's uninstall-with-fallback reports ✗ when the uninstall failed (and gains `--ignore-not-found` so the job-start sweep with no release is not a failure), the gated CRD step reports ✗ on a timed-out delete or unreadable record probe (designed skips stay ✓), and both sweeps report `N/M kinds failed, K absent` — a kind whose API group the cluster does not serve counts absent, not failed, so healthy clusters do not accumulate false ✗. Exit code stays 0 unless `CI_TEARDOWN_STRICT=1`. Sweep deletes gain `--timeout=60s --request-timeout=30s`; Step 1/2 keep #1185's 10m/5m budgets, so fitting a full abort teardown inside the wrapper's 5m `grace_period` still requires the oss-test-infra `grace_period` raise (see Context).
- `tests/test_ci_teardown_status.py`, `tests/test_boskos_heartbeat.py` (new, 11 tests): run the real scripts; covered by the existing `tests/test_*.py` Makefile glob. Heartbeat tests are condition-based (wait-until with generous deadlines), not timing-based, so slow or loaded machines cannot fail a healthy daemon.

## Context

- Root-cause investigation of the 2026-09-01 incident (Prow build logs in `gs://kube-agents-prow` for PR #1089 builds around 15:13Z show the lease changing hands mid-teardown) and of the eval-slot saturation in #1097 — this PR fixes the teardown/lease half; the endpoint-slot half lives in the Hermes runtime, not this repo.
- Rebased over #1185 (merged) after its landing conflicted with the first push; #1180 (open) also touches this script for a different purpose — whichever merges second rebases.
- Out-of-repo follow-ups: raising the job's `grace_period` in oss-test-infra so even an everything-hangs abort teardown reaches its release, and the Boskos pool expiry value (gke-internal/test-infra) if a wider window is wanted.

## Testing

- 41 tests pass over the merged script: the 11 added here plus #1185's 30 (`test_ci_teardown_release_record`, `test_ci_teardown_sweep`, `test_ci_deploy_release_guard`), locally and in the `Run Python Unit Tests` check.
- The incident replay: stub kubectl failing with the incident's exact errors — the pre-#1185 script printed all ✓ and exited 0; this branch reports the failed CRD step and the absent VAP kind, and summarizes.
- The heartbeat suite ran 5x consecutively without flake after the condition-based rewrite; the teardown suite also passes with `JOB_NAME`/`BUILD_ID` set, proving the tests scrub inherited CI identity rather than heartbeating the real Boskos from inside a presubmit.
- `make docs-check` green; `bash -n` on all touched scripts.

### Live validation

This change's production surface is the Prow presubmit, which cannot run outside CI. It was exercised end-to-end against a real stack standing in for it: a kind cluster with the real `kube-agents` chart installed (operator, LiteLLM, CRDs; `podMonitoring` off — no GKE-managed CRDs off-GKE), the real upstream Boskos server backed by that cluster's CRD storage, the real Boskos reaper (60s expiry, 1-minute sweep — scaled from prod's ~5m), and the real `boskosctl` acquiring `busy` leases under the production owner convention. Only `gcloud` was stubbed (no GKE control plane to `get-credentials` against). A CR with an uncleared finalizer made CRD deletion stall, reproducing the incident's slow-teardown shape. The lease lifecycle mirrored production each time: acquire → real `boskosctl heartbeat` → kill it (as `cleanup()` does) → teardown → release.

Control (old script): the reaper reclaimed the lease mid-teardown, the script printed `✓ CRD deletion finished in 81s` + `Cleanup Complete` over it, and the release failed with the incident's exact error: `failed to release resource "kube-agents-evals-test": status 401 Unauthorized`.

This branch, at the merged head, same conditions: the daemon's beats (`ok http=200`, 11/11 at 30s cadence in the detail log) rode through a 308s teardown including the CRD delete hitting #1185's 5m bound; the lease was still `busy/pull-kube-agents-smoke-test-9999` when teardown ended; the release succeeded (`released resource "kube-agents-evals-test"`, exit 0); the log truthfully reported `✗ CRD step (delete timed out or failed after 5m) FAILED (status 1) after 301s` and the final summary; the daemon added 3 lines to the job log and did not outlive the script. Two earlier rerun attempts were invalidated by the host machine suspending mid-experiment (beats gapped for hours; the reaper correctly reclaimed) — an artifact of a sleeping laptop, and incidentally a live demonstration of silence-means-reclaim behaving as designed.

Not covered: the real in-cluster Boskos URL and the exact production reaper expiry (pool config lives in gke-internal/test-infra); the derived owner convention is taken from the wrapper's source in oss-test-infra and will 401 loudly-but-harmlessly if that convention ever changes. Test artifacts (kind cluster, local Boskos, stub dir) were torn down.

## Self-Review

Adversarial and docs-drift passes ran in clean subagent contexts (fresh contexts handed the diff range and the verified wrapper facts, not the authoring conversation); the automated review's findings after opening are folded in below.

Adversarial pass:
- **Fixed (Blocking):** the teardown test suite inherited `os.environ`, so inside Prow it would have armed the heartbeat against the real in-cluster Boskos. `_run` scrubs the five identity variables; verified by running the suite with them set.
- **Fixed:** kubectl `--timeout` bounds only the deletion wait; sweep deletes and the CRD delete gained `--request-timeout=30s` for the connect/discovery phase a dead control plane hangs.
- **Fixed:** curl's `-w` plus a fallback echo could log `http=000000`; beats normalise to the last three digits.
- **Fixed (cosmetic):** counts printed as `(exit N)`; now `(status N)`.
- Verified clean by the pass: quoting/word-splitting, `set -u` holes, EXIT-trap daemon lifetime (no leak, confirmed by execution), GNU/bash portability for the kubekins image.

Docs-drift pass:
- **Fixed:** the daemon header and test constants said 10-minute expiry while the teardown comment said 5; aligned to the ~5m window the wrapper's comments name (the automated review caught two further stale references in the test module docstring — also fixed).
- No repo doc describes the old teardown contract, so nothing went stale; no new Markdown, so no docs-map entry owed.

Automated review (kube-agents-bot) findings:
- **Fixed (merge-blocking):** heartbeat tests asserted beat counts against wall-clock windows and flaked on a stock machine; rewritten to condition-based waits (`wait_until` with a 30s ceiling) that pin behavior — beats repeat, identity correct, one transition line per 401, beats resume after a SIGSTOP hang — with no timing arithmetic. 5 consecutive clean runs.
- **Fixed (description drift):** three claims the rebase resolution had dropped from the code ("helm 3m", `--ignore-not-found`, the grace-period budget comment). `--ignore-not-found` is re-added (it also fixes the job-start sweep reporting a false ✗); the helm/CRD budgets deliberately stay #1185's 10m/5m readonly values, and every body claim about budgets now matches the code.
- **Fixed:** the two stale 10-minute comments in the test module.

Post-rebase additions this section still owes a reviewer: the `finish_step` layering over #1185's steps is covered by #1185's own 30 tests passing unmodified except for none — no #1185 test was edited; the absent-vs-failed sweep calibration is new logic added after live validation showed 3 false ✗ on a cluster without cert-manager/managed-collection groups, and is pinned by the incident test.

## Risk & Rollout

The daemon is fail-soft: missing env disables it with one line, an unreachable or disagreeing Boskos yields logged-once failed beats, and it cannot outlive the script — worst case is today's behavior plus a few log lines. Behavior changes with real blast radius: sweep deletes now carry 60s/30s bounds (healthy sweeps run in seconds; a hung one now surfaces as ✗ instead of hanging the job), and `CI_TEARDOWN_STRICT` defaults off so no caller's exit-code contract changes at merge. Revert is a single commit revert with no persistent state. The derived owner convention couples this repo to the wrapper's naming; if oss-test-infra changes it, beats 401 visibly and teardown is unaffected.

---

### DIFF
--- a/hack/boskos_heartbeat.sh
+++ b/hack/boskos_heartbeat.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+# ==============================================================================
+# Boskos lease heartbeat daemon
+# ==============================================================================
+# POSTs the Boskos client's /update call every
+# BOSKOS_HEARTBEAT_INTERVAL_SECONDS so the lease's LastUpdate stays fresh and
+# the reaper (ranch.Reset, ~5m window) does not reclaim the project while a
+# CI phase still runs. Covers what the Prow wrapper's own heartbeat does not
+# — teardown, where a slow sweep outliving the lease turned the final
+# release into a 401 (2026-09-01, kube-agents-evals-4).
+#
+# Liveness travels on its own channel: per-beat lines go to
+# BOSKOS_HEARTBEAT_LOG, stdout carries only start, ok<->fail transitions,
+# and a stop summary. At the ~5m window a 30s cadence tolerates 9 missed
+# beats, so a 3-minute hang (6 beats) keeps the lease; a pod frozen past the
+# window loses it, which is the reclaim working as intended.
+#
+# Usage (backgrounded, killed by the caller's EXIT trap):
+#   ./hack/boskos_heartbeat.sh & HEARTBEAT_PID=$!
+#   trap 'kill "${HEARTBEAT_PID}" 2>/dev/null || true' EXIT
+#
+# Disabled (single notice, exit 0) unless BOSKOS_HOST, BOSKOS_RESOURCE_NAME,
+# and BOSKOS_OWNER_NAME are all set. Pool resource names are DNS-safe, so
+# the query string needs no URL encoding.
+
+set -uo pipefail
+
+BOSKOS_HEARTBEAT_INTERVAL_SECONDS="${BOSKOS_HEARTBEAT_INTERVAL_SECONDS:-30}"
+# State every kube-agents lease is held in while a job owns it; /update 401s
+# on an owner mismatch and 409s on a state mismatch, so both must match the
+# acquire call the Prow wrapper made.
+BOSKOS_RESOURCE_STATE="${BOSKOS_RESOURCE_STATE:-busy}"
+# Per-beat detail lands here, off the job log. ARTIFACTS is set by Prow.
+BOSKOS_HEARTBEAT_LOG="${BOSKOS_HEARTBEAT_LOG:-${ARTIFACTS:-/tmp}/boskos-heartbeat.log}"
+# A beat must never wedge the loop behind a slow server: cap each call well
+# under the interval so a timed-out beat still leaves room for the next one.
+CURL_MAX_TIME_SECONDS=10
+LOG_PREFIX="boskos-heartbeat:"
+
+if [ -z "${BOSKOS_HOST:-}" ] || [ -z "${BOSKOS_RESOURCE_NAME:-}" ] || [ -z "${BOSKOS_OWNER_NAME:-}" ]; then
+  echo "${LOG_PREFIX} disabled (BOSKOS_HOST/BOSKOS_RESOURCE_NAME/BOSKOS_OWNER_NAME not all set)"
+  exit 0
+fi
+
+UPDATE_URL="${BOSKOS_HOST}/update?name=${BOSKOS_RESOURCE_NAME}&owner=${BOSKOS_OWNER_NAME}&state=${BOSKOS_RESOURCE_STATE}"
+
+beats_sent=0
+beats_failed=0
+# "" until the first beat resolves, then ok|fail; transitions are the only
+# per-beat events worth a line on the job log.
+last_status=""
+
+summary() {
+  echo "${LOG_PREFIX} stopping for ${BOSKOS_RESOURCE_NAME}: ${beats_sent} beats sent, ${beats_failed} failed (detail: ${BOSKOS_HEARTBEAT_LOG})"
+  exit 0
+}
+trap summary TERM INT
+
+mkdir -p "$(dirname "${BOSKOS_HEARTBEAT_LOG}")" 2>/dev/null || true
+echo "${LOG_PREFIX} started for ${BOSKOS_RESOURCE_NAME} (owner ${BOSKOS_OWNER_NAME}, every ${BOSKOS_HEARTBEAT_INTERVAL_SECONDS}s, detail: ${BOSKOS_HEARTBEAT_LOG})"
+
+while true; do
+  # curl -w emits a code even on failure ("000", or "200000" when the
+  # connection dies after headers), so normalise to the LAST three digits
+  # rather than appending a fallback that doubles it up.
+  http_code="$(curl -sS -o /dev/null -w '%{http_code}' --max-time "${CURL_MAX_TIME_SECONDS}" \
+    -X POST "${UPDATE_URL}" 2>>"${BOSKOS_HEARTBEAT_LOG}")" || true
+  http_code="${http_code:(-3)}"
+  [ -n "${http_code}" ] || http_code="000"
+  beats_sent=$((beats_sent + 1))
+  if [ "${http_code}" = "200" ]; then
+    status="ok"
+  else
+    status="fail"
+    beats_failed=$((beats_failed + 1))
+  fi
+  echo "$(date -u +'%Y-%m-%dT%H:%M:%SZ') ${status} http=${http_code}" >>"${BOSKOS_HEARTBEAT_LOG}"
+  if [ "${status}" != "${last_status}" ]; then
+    if [ "${status}" = "fail" ]; then
+      # 401 here means the lease is already lost (owner mismatch) — the exact
+      # signal that used to surface only as a failed release at job end.
+      echo "${LOG_PREFIX} beat FAILED for ${BOSKOS_RESOURCE_NAME} (http=${http_code}); continuing"
+    elif [ -n "${last_status}" ]; then
+      echo "${LOG_PREFIX} recovered for ${BOSKOS_RESOURCE_NAME}"
+    fi
+    last_status="${status}"
+  fi
+  sleep "${BOSKOS_HEARTBEAT_INTERVAL_SECONDS}"
+done
--- a/hack/ci-teardown.sh
+++ b/hack/ci-teardown.sh
@@ -46,6 +46,26 @@ readonly SECRET_NAME_PREFIX_RE='^secret/'
 readonly CHART_CRD_DIR="charts/kube-agents/crds/"
 readonly CRD_DELETE_TIMEOUT="5m"
 
+# Exit nonzero when any step failed; default off so the Prow wrapper still
+# reaches its Boskos release — the honest signal is the ✗ lines and the final
+# summary. `readonly VAR="${VAR:-default}"` throughout this block: env still
+# overrides, and the step tests lift `readonly ` head lines along.
+readonly CI_TEARDOWN_STRICT="${CI_TEARDOWN_STRICT:-0}"
+
+# Bounds for the sweep deletes in Steps 3 and 4 (Steps 1 and 2 carry their
+# own budgets above). --timeout bounds the wait-for-deletion phase; the
+# request timeout bounds each API call, the phase an unreachable control
+# plane hangs — where the 2026-09-01 incident spent 546s.
+readonly SWEEP_DELETE_TIMEOUT="${CI_TEARDOWN_SWEEP_TIMEOUT:-60s}"
+readonly KUBECTL_REQUEST_TIMEOUT="${CI_TEARDOWN_KUBECTL_REQUEST_TIMEOUT:-30s}"
+
+# Endpoint and owner convention the Prow wrapper leases with
+# (BOSKOS_OWNER="${JOB_NAME}-${BUILD_ID}", oss-test-infra
+# prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml), so the
+# heartbeat below arms itself with no wrapper change. If the convention ever
+# changes, beats 401 — one loud daemon line, harmless to the teardown.
+readonly PROW_BOSKOS_DEFAULT_HOST="http://boskos.boskos.svc.cluster.local"
+
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "${REPO_ROOT}"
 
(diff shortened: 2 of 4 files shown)


tests/test_boskos_heartbeat.py
```diff
@@ -0,0 +1,176 @@
...
+        self.assertGreaterEqual(len(beats), 5, stdout)
+        for _, name, owner, state in beats:
+            self.assertEqual((name, owner, state), (LEASE_NAME, LEASE_OWNER, LEASE_STATE))
+
+    def test_stdout_does_not_flood_but_detail_log_has_every_beat(self):
+        proc = self._spawn()
+        time.sleep(3)  # ~15 beats
+        stdout = self._stop(proc)
+        beat_count = len(_FakeBoskos.updates)
+        self.assertGreaterEqual(beat_count, 10)
```
jayantid: 🔴 **Merge-blocking until the thresholds get headroom.** Two to three of the five heartbeat tests fail nondeterministically on a stock dev machine.

The beat-count assertions assume the loop's period equals the 0.2s interval, but each iteration is curl spawn + HTTP round-trip + `sleep`. Line 121 demands 10 beats in 3s (period ≤ 0.3s, so ≤ 0.1s of overhead per beat); lines 118 and 135 demand 5 in 2s.

Run twice at `16744a2` on an idle macOS machine: first run 3 failures, second run 2, with different failing sets (`4 not greater than or equal to 5`, `7 not greater than or equal to 10`). Once merged, the `tests/test_*.py` Makefile glob puts these in every `make test`: Mac contributors get a red suite, and a loaded presubmit node has ~0.1s of headroom before it flakes.

Poll to a deadline instead of sleeping a fixed time and asserting a count:

```python
deadline = time.monotonic() + 10
while len(_FakeBoskos.updates) < 10 and time.monotonic() < deadline:
    time.sleep(0.05)
beat_count = len(_FakeBoskos.updates)
```

hack/ci-teardown.sh
```diff
@@ -46,6 +46,29 @@ readonly SECRET_NAME_PREFIX_RE='^secret/'
...
 
+# Exit nonzero when any step failed. Default off: today's callers treat
+# teardown as best-effort and the Prow wrapper must still reach its Boskos
+# release, so the honest signal is the ✗ lines and the final summary until
+# every caller is ready for a hard failure. `readonly VAR="${VAR:-default}"`
+# throughout this block: env still overrides, and the tests that lift the
+# steps out of this file carry `readonly ` head lines along.
+readonly CI_TEARDOWN_STRICT="${CI_TEARDOWN_STRICT:-0}"
+
+# Bounds for the sweep deletes in Steps 3 and 4 (Steps 1 and 2 carry their
```
jayantid, line 55: 🟡 **Not merge-blocking by itself, but sync the description before sign-off: three changes it claims are not in this diff.**

At `16744a2` the code lacks three things the PR description and self-review assert. `RELEASE_UNINSTALL_TIMEOUT` is still `10m` (hack/ci-teardown.sh:32) while What Changed and Risk & Rollout both say "helm 3m (was 10m)". No `helm uninstall --ignore-not-found` exists in the script (self-review says "Fixed"). No comment about grace_period or the single-stuck-step budget exists (self-review says "reworded rather than fixed").

The Summary's abort-path claim — steps bounded so an aborted job reaches its Boskos release inside the 5-minute grace period — is therefore not delivered. A stuck `helm uninstall --wait` still holds teardown for up to 10m and the job is SIGKILLed before its release, exactly as before this PR.

Push the missing commit if it exists, or land the timeout and reword:

```bash
readonly RELEASE_UNINSTALL_TIMEOUT="${CI_TEARDOWN_RELEASE_UNINSTALL_TIMEOUT:-3m}"
```

tests/test_boskos_heartbeat.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def test_hang_recovers_within_expiry_budget(self):
+        interval = float(TEST_INTERVAL_SECONDS)
+        # The scenario from the incident discussion: the job hangs for 3
+        # minutes (6 intervals), then resumes; the next beat must land long
+        # before the expiry (20 intervals at production ratio).
+        hang_intervals = 6
+        expiry_intervals = (
+            PRODUCTION_EXPIRY_SECONDS // PRODUCTION_INTERVAL_SECONDS
+        )  # = 20
```
jayantid: 🟡 **Not merge-blocking.** Two comments in this file still describe the 10-minute expiry the self-review says was aligned to ~5.

`PRODUCTION_EXPIRY_SECONDS = 5 * 60` (line 33) makes `expiry_intervals` equal 10, but the inline comment here says `# = 20`, and the module docstring (line 10) says "30s interval vs 10min expiry = 20 beats".

A reader tuning the heartbeat interval against these comments budgets twice the real tolerance: 20 missed beats when the ~5m reaper window allows 10. The daemon header and the teardown comment were aligned; this file's two prose references were missed.

```python
-  inside the expiry budget (30s interval vs 10min expiry = 20 beats);
+  inside the expiry budget (30s interval vs ~5min expiry = 10 beats);
...
-        )  # = 20
+        )  # = 10
```

tests/test_boskos_heartbeat.py
haoxuw: Oh good catch!

haoxuw: Fixed in 9c3985c. The suite is rewritten to be condition-based rather than timing-based: every assertion now waits for a condition (`wait_until`, 30s ceiling, 50ms poll) instead of demanding N beats inside a wall-clock window, so loop overhead per beat no longer matters. What is pinned is behavior only — beats repeat with the right name/owner/state, stdout stays ≤3 lines while the detail log records, a 401 produces exactly one transition line and the loop survives, and a beat lands after a SIGSTOP/SIGCONT hang. The 30s/5m production ratio remains asserted as pure arithmetic. Five consecutive local runs clean; the suite also got faster (~5s) since waits end when conditions hold.

hack/ci-teardown.sh
haoxuw: Both done as of 9c3985c. Code side: `--ignore-not-found` is re-added to the Step 1 uninstall (it had been dropped in the #1185 rebase resolution; it also stops the job-start sweep from reporting a false ✗ when no release exists). Description side: the body is rewritten to match the merged head — the helm/CRD budgets are stated as #1185's readonly 10m/5m (deliberately kept; the abort-grace fit is explicitly deferred to the oss-test-infra grace_period raise, named in Context), and only the sweep deletes carry this PR's 60s/30s bounds. The Self-Review section now carries these dispositions.

tests/test_boskos_heartbeat.py
haoxuw: Fixed in 9c3985c — the module docstring and the inline `# = 20` were the last two 10-minute references; the rewrite that replaced the timing assertions removed both, and the production ratio now appears once, as arithmetic (`PRODUCTION_EXPIRY_SECONDS = 5 * 60`, 10-beat budget) with the hang test asserting a 3-minute hang fits it.

hack/boskos_heartbeat.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
haoxuw: PR body rewritten at 9c3985c to match the merged head: the budget claims now describe #1185's 10m/5m as kept, `--ignore-not-found` is back in the code, and the Summary's abort-grace claim is replaced with the honest statement that the full fit depends on the oss-test-infra grace_period raise (Context section). Live validation was re-run at the merged head — 11/11 beats through a 308s teardown, release exit 0.

jayantid (approved):

bradhoekstra:
/hold

Temporary, no change requested: parking this PR out of the Tide pool for a few minutes so Tide merges overridden PRs one at a time instead of forming another multi-hour batch. bradhoekstra will `/hold cancel` and re-override shortly.

PR #1197, opened 2026-09-03
fix(stockout): wait for agent availability and clean up kanban tasks

sergiu1spiridon:
## Summary

Adds Platform Agent availability and readiness waits across the stockout E2E test suite, verification scripts, and scenario harness, and reclaims/archives stale kanban tasks between scenario runs to prevent concurrency slot starvation (`max_in_progress = 2`).

## Why This Change

During GKE Autopilot node scale-down and bin-packing compaction (e.g. after earlier scenarios delete large test workloads), the Platform Agent gateway pod can be evicted and rescheduled midway through the test suite (as observed in GitHub Actions run 33723476653). The scenario runner cached dead pod references and lacked polling for pod readiness, while orphaned foreign claims and completed tasks filled the kanban board, causing concurrency slot starvation for subsequent alerts.

## What Changed

- **`tests/e2e/test_stockout_investigation.py`**:
  - `_wait_for_agent_available`: waits for gateway workload rollout, recomputes the deadline after rollout status returns so rollout and pod readiness probes have independent timeout budgets, and probes for pod running, non-terminating status, container readiness, and `SKILL.md` mount.
  - `_clean_stale_kanban_tasks`: reclaims and archives tasks for the route, checks return codes, verifies that 0 active tasks remain for the route on the board, surfaces non-zero errors via `RuntimeError`, and returns the archived count.
  - Added explicit timeouts on `subprocess.run` for both smoke test (`_SMOKE_VERIFY_TIMEOUT_SECONDS = 300`) and scenario runner (`watch_timeout + _SCENARIO_RUN_HEADROOM_SECONDS = 300`), with comments documenting headroom sizing against preflight and cleanup.
- **`agentplugins/gke-stockout-investigator/scenarios/lib/common.sh`**:
  - Introduced `SKILL_MOUNT_TIMEOUT="${SKILL_MOUNT_TIMEOUT:-30}"` so transient skill mount lag resolves quickly and permanent failures do not delay execution or exhaust the 300s headroom.
  - Updated `cleanup_kanban` and `clear_dedup` to check exit codes, count and log archived tasks (`ok "archived N stale kanban task(s)..."` or `dim "kanban board clean..."`), verify board state, and log errors on failure rather than swallowing them.
  - Replaced static pod resolution with `_wait_for_platform_pod`, writing the resolved pod name to `${SCENARIO_RUN_DIR}/platform_pod` and syncing via `_sync_platform_pod` so subshell calls update `PLATFORM_POD` in parent contexts.
  - Redirected progress output in `_wait_for_platform_pod` to stderr (`>&2`) so diagnostic text is never captured as stdout or parsed as a completed task.
  - Capped in-loop pod re-resolution during watch loop failures to 15s (`IN_LOOP_POD_RESOLVE_TIMEOUT`), returning 1 without dying so subshells don't abort with `set -e`.
  - Converted `watch_investigation` to bound against a real wall-clock deadline (`start_time + WATCH_TIMEOUT`), tracking true elapsed time.
- **`agentplugins/gke-stockout-investigator/verify.sh`**: Added Step 0 checking gateway deployment rollout readiness before publishing test alerts.
- **`tests/e2e/test_agent_fleet_audit.py`**: Added a polling wait loop for running, ready platform-agent pods in `test_github_token_minting_and_connectivity`.
- **`tests/test_stockout_fixture_helpers.py`**: Added unit test coverage for `_wait_for_agent_available` (including independent probe budget after slow rollout) and `_clean_stale_kanban_tasks` (count return and error raising).

## Context

Addresses stockout scenario investigation timeouts and kanban slot starvation caused by gateway pod rescheduling during E2E runs (e.g. GitHub Actions run 33723476653).

## Testing

- `python3 -m unittest discover -s tests -p "test_stockout_fixture_helpers.py"`: 43/43 tests passing.
- `make docs-check`: clean (relative links, terminology, context budget, and documentation map all passing).

### Live validation

Exercised against running GKE Autopilot cluster in `rc` environment:
- **Cluster**: `platform-agent-host` (`us-east4`, project `sergiuspiridon-gkedemos`, namespace `kubeagents-system`)
- **Images & Operator**:
  - `k8s-operator`: `ghcr.io/sergiu1spiridon/kube-agents-sergiu/k8s-operator:4044928181fb535865c24a592959b686af927719`
  - `platform-agent`: `ghcr.io/sergiu1spiridon/kube-agents-sergiu/platform-agent:4044928181fb535865c24a592959b686af927719`
  - Gateway pod: `platform-agent-gateway-787c6d5d9f-hbfbm`
- **E2E Test Suite Runs**:
  - Run [#33749373944](https://github.com/sergiu1spiridon/kube-agents-sergiu/actions/runs/33749373944):
    - `test_stockout_ingress_alert_smoke`: **PASSED**
    - `01-gpu-regional-scarcity`: **SKIPPED** (no GPUs on cluster)
    - `02-gpu-quota-exceeded`: **SKIPPED** (no GPUs on cluster)
    - `03-large-vm-shape-scarcity`: **PASSED**
    - `04-missing-zone-fallback`: **PASSED**
    - `05-missing-ondemand-floor`: **PASSED** (successfully waited for gateway readiness and investigated alert without timing out on evicted pod references)
    - `06-stateful-disk-generation-mix`: **PASSED**
  - Run [#33758170112](https://github.com/sergiu1spiridon/kube-agents-sergiu/actions/runs/33758170112):
    - `test_stockout_ingress_alert_smoke`: **PASSED**
    - `07-hyperdisk-incompatibility`: **PASSED** (previously failed due to kanban concurrency starvation `max_in_progress=2`; verified clean pre-run archiving and instant claim)
    - `08-ccc-priority-starvation`: **PASSED** / claimed and executing cleanly.
- **Kanban Board Verification**:
  - Verified via `hermes kanban ls --json` inside `platform-agent-gateway-787c6d5d9f-hbfbm` that route tasks are archived before each scenario, ensuring 0 stale running tasks and freeing slots for new alerts.

### Self-Review

- Passes run:
  - `review-adversarial` run in dedicated subagent (`f1aa9165-ca7b-435e-b792-1b4296de1db2`).
  - `review-docs-drift` run in dedicated subagent (`f3772aa8-f931-4cfb-aca7-ed48765eb9c4`).
  - Automated review by `kube-agents-bot` (3ce6d33).
  - Maintainer review by `@mplakhtiy` on PR #1197.
- Findings and dispositions:
  - `mplakhtiy` Finding 1 (🔴 High - `deploy/docker/patches/kanban_scheduling.py` liveness bypass): Addressed. Reverted `kanban_scheduling.py` back to `upstream/main`, keeping this PR strictly 100% test scaffolding and scenario fixtures.
  - `mplakhtiy` Finding 2 (🟠 Medium - `_wait_for_agent_available` shared deadline): Addressed. Recomputed deadline after `rollout status` succeeds so rollout and pod probe loop have separate budgets. Added unit test `test_slow_rollout_does_not_consume_probe_budget`.
  - `mplakhtiy` Finding 3 & 4 (🟠 Medium - preflight worst case & skill mount retry): Addressed. Added `SKILL_MOUNT_TIMEOUT=30` in `common.sh`. Preflight worst-case duration is now 180s + 30s = 210s, safely within the 300s headroom (`_SCENARIO_RUN_HEADROOM_SECONDS`). Permanent missing skill failures exit in 30s instead of 180s.
  - `mplakhtiy` Finding 5 (🟠 Medium - kanban cleanup reporting & error swallowing): Addressed. Un-swallowed errors, verified board state via post-cleanup `hermes kanban ls`, logged archived task counts, and added unit tests in `CleanStaleKanbanTasksTest`.
  - `review-docs-drift`: No Blocking or Advisory findings.
  - `kube-agents-bot` Findings 1-4: Addressed (wall-clock deadline, progress output to stderr, subshell latch, live validation details).

## Risk & Rollout

Low risk. Changes are strictly scoped to test scaffolding and scenario execution scripts. No container image patches, customer-facing CRDs, or production runtime APIs are altered.

---
Generated with the help of Antigravity.

### DIFF
--- a/agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
+++ b/agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
@@ -373,62 +373,26 @@ preflight() {
 cleanup_kanban() {
     _sync_platform_pod
     [ -n "$PLATFORM_POD" ] || return 0
+    local script_dir clean_script
+    script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    clean_script="${script_dir}/clean_stale_kanban_tasks.py"
+    if [ ! -f "$clean_script" ]; then
+        warn "kanban cleanup script missing: ${clean_script}"
+        return 1
+    fi
     local out
     if out="$(kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
-        python3 -c '
-import json, os, subprocess, sys
-
-route_name = sys.argv[1] if len(sys.argv) > 1 else ""
-cmd_env = dict(os.environ)
-cmd_env["HOME"] = "/tmp"
-cmd_env.setdefault("HERMES_HOME", "/opt/data")
-try:
-    res = subprocess.run(["hermes", "kanban", "ls", "--json"], env=cmd_env, capture_output=True, text=True, check=True)
-    data = json.loads(res.stdout)
-except Exception as e:
-    sys.stderr.write(f"kanban ls failed: {e}\n")
-    sys.exit(1)
-
-tasks = data.get("tasks") if isinstance(data, dict) else data
-archived = 0
-errors = []
-for t in (tasks or []):
-    tid = t.get("id")
-    title = str(t.get("title", ""))
-    status = t.get("status")
-    if tid and title.startswith(route_name) and status in ("running", "claimed", "ready", "blocked", "todo"):
-        if status in ("running", "claimed"):
-            rec = subprocess.run(["hermes", "kanban", "reclaim", tid], env=cmd_env, capture_output=True, text=True)
-            if rec.returncode != 0:
-                errors.append(f"reclaim {tid}: {rec.stderr.strip()}")
-        arc = subprocess.run(["hermes", "kanban", "archive", tid], env=cmd_env, capture_output=True, text=True)
-        if arc.returncode != 0:
-            errors.append(f"archive {tid}: {arc.stderr.strip()}")
-        else:
-            archived += 1
-
-verify = subprocess.run(["hermes", "kanban", "ls", "--json"], env=cmd_env, capture_output=True, text=True)
-if verify.returncode == 0:
-    vdata = json.loads(verify.stdout)
-    vtasks = vdata.get("tasks") if isinstance(vdata, dict) else vdata
-    lingering = [t.get("id") for t in (vtasks or []) if str(t.get("title", "")).startswith(route_name) and t.get("status") in ("running", "claimed", "ready", "blocked", "todo")]
-    if lingering:
-        errors.append(f"{len(lingering)} tasks still active: {lingering}")
-
-if errors:
-    sys.stderr.write("; ".join(errors) + "\n")
-    sys.exit(1)
-
-print(archived)
-' "$ROUTE_NAME" 2>&1)"; then
+        python3 -c "$(< "$clean_script")" "$ROUTE_NAME" 2>&1)"; then
         local count="${out##*$'\n'}"
         if [ "$count" -gt 0 ] 2>/dev/null; then
             ok "archived ${count} stale kanban task(s) for ${ROUTE_NAME}"
         else
             dim "kanban board clean for ${ROUTE_NAME} (0 active tasks)"
         fi
+        return 0
     else
         warn "kanban cleanup encountered an error: ${out}"
+        return 1
     fi
 }
 
@@ -440,7 +404,9 @@ clear_dedup() {
     kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
         rm -f /opt/data/pubsub_registry.json >/dev/null 2>&1 || true
     ok "cleared the adapter dedup registry"
-    cleanup_kanban
+    # Board cleanup is best-effort hygiene before starting a scenario run; if it fails,
+    # warn but proceed so a transient Hermes error does not abort the scenario setup.
+    cleanup_kanban || warn "proceeding despite kanban cleanup failure; slots may remain occupied"
 }
 
 # ------------------------------------------------------------------- the payload
@@ -695,7 +661,9 @@ apply_workload() {
 
 cleanup_workload() {
     step "Cleaning up"
-    cleanup_kanban
+    # Reclaim and archive any active tasks created during this run. Non-fatal so that
+    # k8s workload deletion and PVC cleanup below proceed even if Hermes encounters an error.
+    cleanup_kanban || true
     if ! declare -F scenario_manifest >/dev/null; then
         dim "nothing to remove"; return 0
     fi
--- a/agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
+++ b/agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
@@ -111,6 +111,8 @@ KEEP_DEDUP=0
 DRY_RUN=0
 EMIT_MANIFEST=""
 WATCH_TIMEOUT="${WATCH_TIMEOUT:-600}"
+AGENT_AVAILABILITY_TIMEOUT="${AGENT_AVAILABILITY_TIMEOUT:-180}"
+AGENT_POLL_INTERVAL=5
 
 SCENARIO_SLUG="$(basename "${BASH_SOURCE[1]:-scenario}" .sh)"
 RUN_ID=""
@@ -245,12 +247,40 @@ _parse_args() {
 
 # -------------------------------------------------------------------- preflight
 
-# Resolve the gateway pod once; several steps need to exec into it.
+# Wait for a ready, non-terminating gateway pod with container platform-agent ready.
+_wait_for_platform_pod() {
+    local waited=0 pod=""
+    dim "waiting up to ${AGENT_AVAILABILITY_TIMEOUT}s for a ready platform-agent-gateway pod"
+    while [ "$waited" -lt "$AGENT_AVAILABILITY_TIMEOUT" ]; do
+        pod="$(kmgmt get pods -n "$AGENT_NAMESPACE" -l app=platform-agent-gateway -o json 2>/dev/null | python3 -c '
+import sys, json
+try:
+    data = json.load(sys.stdin)
+    for p in data.get("items", []):
+        meta = p.get("metadata", {})
+        if meta.get("deletionTimestamp"):
+            continue
+        if p.get("status", {}).get("phase") != "Running":
+            continue
+        statuses = {c["name"]: c.get("ready", False) for c in p.get("status", {}).get("containerStatuses", [])}
+        if statuses.get("platform-agent", False):
+            print(meta.get("name", ""))
+            break
+except Exception:
+    pass
+' || true)"
+        if [ -n "$pod" ]; then
+            PLATFORM_POD="$pod"
+            return 0
+        fi
+        sleep "$AGENT_POLL_INTERVAL"
+        waited=$((waited + AGENT_POLL_INTERVAL))
+    done
+    die "no ready platform-agent-gateway pod in ${AGENT_NAMESPACE} after ${AGENT_AVAILABILITY_TIMEOUT}s"
+}
+
 _resolve_platform_pod() {
-    PLATFORM_POD="$(kmgmt get pods -n "$AGENT_NAMESPACE" \
-        -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null \
-        | grep '^platform-agent-gateway-' | head -1 || true)"
-    [ -n "$PLATFORM_POD" ] || die "no platform-agent-gateway pod in ${AGENT_NAMESPACE} on ${MGMT_CONTEXT}"
+    _wait_for_platform_pod
 }
 
 # The failure modes below are the ones that have actually cost time: an alert that is
@@ -291,14 +321,21 @@ preflight() {
     _resolve_platform_pod
     dim "gateway pod ${PLATFORM_POD}"
 
-    # $HERMES_HOME is /opt/data in the shipped image, but read it rather than assume:
-    # a plugin that mounted at the wrong path is exactly the failure this catches.
-    if kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
-        sh -c 'test -f "${HERMES_HOME:-/opt/data}/profiles/platform/plugins/'"${PLUGIN_NAME}"'/skills/gke-stockout-investigator/SKILL.md"' 2>/dev/null
-    then
+    # Wait up to AGENT_AVAILABILITY_TIMEOUT for the skill to be readable in the resolved pod
+    local skill_waited=0 skill_found=0
+    while [ "$skill_waited" -lt "$AGENT_AVAILABILITY_TIMEOUT" ]; do
+        if kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
+            sh -c 'test -f "${HERMES_HOME:-/opt/data}/profiles/platform/plugins/'"${PLUGIN_NAME}"'/skills/gke-stockout-investigator/SKILL.md"' 2>/dev/null; then
+            skill_found=1
+            break
+        fi
+        sleep "$AGENT_POLL_INTERVAL"
+        skill_waited=$((skill_waited + AGENT_POLL_INTERVAL))
+    done
+    if [ "$skill_found" -eq 1 ]; then
         ok "skill resolves in the platform profile"
     else
-        warn "SKILL.md not found under the platform profile's plugins directory"
+        warn "SKILL.md not found under the platform profile's plugins directory after ${skill_waited}s"
         warn "the investigation would run without it — check the plugin mount"
     fi
 
@@ -313,13 +350,39 @@ preflight() {
     fi
 }
 
+# Reclaim and archive any active, blocked, or stuck tasks on the kanban board for this route
+# so concurrency slots (max_in_progress = 2) are not starved.
+cleanup_kanban() {
+    [ -n "$PLATFORM_POD" ] || return 0
+    kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
+        python3 -c "
+import subprocess, json
+try:
+    cmd_env = {'HOME': '/tmp', 'PATH': '/opt/hermes/.venv/bin:/usr/local/bin:/usr/bin:/bin'}
+    out = subprocess.check_output(['hermes', 'kanban', 'ls', '--json'], env=cmd_env)
+    data = json.loads(out)
+    tasks = data.get('tasks') if isinstance(data, dict) else data
+    for t in (tasks or []):
+        tid = t.get('id')
+        title = str(t.get('title', ''))
+        status = t.get('status')
+        if tid and title.startswith('${ROUTE_NAME}') and status in ('running', 'claimed', 'ready', 'blocked', 'todo'):
+            if status in ('running', 'claimed'):
+                subprocess.run(['hermes', 'kanban', 'reclaim', tid], env=cmd_env, capture_output=True)
+            subprocess.run(['hermes', 'kanban', 'archive', tid], env=cmd_env, capture_output=True)
+except Exception:
+    pass
+" >/dev/null 2>&1 || true
+}
+
 # The adapter dedups on cluster + namespace + controller for 24h. Re-running a
 # scenario is the normal case here, so clear the registry unless asked not to.
 clear_dedup() {
     [ "$KEEP_DEDUP" -eq 1 ] && { dim "keeping dedup registry (--keep-dedup)"; return 0; }
     kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
         rm -f /opt/data/pubsub_registry.json >/dev/null 2>&1 || true
-    ok "cleared the adapter dedup registry"
+    cleanup_kanban
+    ok "cleared the adapter dedup registry and kanban route tasks"
 }
 
 # ------------------------------------------------------------------- the payload
@@ -574,6 +637,7 @@ apply_workload() {
 
 cleanup_workload() {
     step "Cleaning up"
+    cleanup_kanban
     if ! declare -F scenario_manifest >/dev/null; then
         dim "nothing to remove"; return 0
     fi
@@ -675,6 +739,7 @@ print(r.text)
 " 2>/dev/null)"; then
         printf '%s\n' "$out"
     else
+        _wait_for_platform_pod || true
         warn_once sessions "could not read the agent's session list; treating it as empty — ${out:-no response} (issue #786)"
         echo '{}'
     fi
@@ -706,6 +771,9 @@ _task_after() {
         env HOME=/tmp hermes kanban ls --json 2>"$err")"; then
         errtext="$(head -c 200 "$err")"
         rm -f "$err"
+        if [[ "$errtext" == *"not found"* || "$errtext" == *"NotFound"* || "$errtext" == *"terminating"* || "$errtext" == *"closed before"* ]]; then
+            _wait_for_platform_pod || true
+        fi
         warn_once tasks "could not read the agent's kanban board; treating it as empty — ${errtext:-${out:-no response}}"
         return 0
     fi
@@ -713,11 +781,19 @@ _task_after() {
     printf '%s\n' "$out" | python3 -c "
 import json, sys
 since = float('$since')
+raw = sys.stdin.read()
+data = None
 try:
-    data = json.load(sys.stdin)
+    data = json.loads(raw)
 except Exception:
-    # 3, not 0: the board answered with something that is not JSON, which is a broken
-    # read and not an empty one. The caller turns this into one warning per run.
+    start = min([pos for pos in (raw.find('['), raw.find('{')) if pos != -1], default=-1)
+    end = max([pos for pos in (raw.rfind(']'), raw.rfind('}')) if pos != -1], default=-1)
+    if start != -1 and end != -1 and end > start:
+        try:
+            data = json.loads(raw[start:end+1])
+        except Exception:
+            pass
+if data is None:
     sys.exit(3)
 tasks = data.get('tasks') if isinstance(data, dict) else data
 for t in sorted(tasks or [], key=lambda x: x.get('created_at') or 0, reverse=True):
--- a/agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
+++ b/agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
@@ -213,7 +213,7 @@ Usage: ./${SCENARIO_SLUG}.sh [options]
                     cluster+namespace+controller is otherwise silently dropped.
   --watch-timeout N Seconds to watch for the investigation (default ${WATCH_TIMEOUT}).
   --emit-manifest F Write the scenario's manifest to F ("-" for stdout) and exit, ready
-                    for `kubectl apply -f`. Includes the namespace and pins the workload
+                    for \`kubectl apply -f\`. Includes the namespace and pins the workload
                     to it, so the file stands alone — useful for launching a scenario by
                     hand, or for reading what a scenario actually deploys.
   --dry-run         Print the manifest and the alert payload; change nothing.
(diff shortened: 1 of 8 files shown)


agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -706,18 +771,29 @@ _task_after() {
         env HOME=/tmp hermes kanban ls --json 2>"$err")"; then
         errtext="$(head -c 200 "$err")"
         rm -f "$err"
+        if [[ "$errtext" == *"not found"* || "$errtext" == *"NotFound"* || "$errtext" == *"terminating"* || "$errtext" == *"closed before"* ]]; then
+            _wait_for_platform_pod || true
+        fi
         warn_once tasks "could not read the agent's kanban board; treating it as empty — ${errtext:-${out:-no response}}"
         return 0
     fi
```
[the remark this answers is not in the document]
sergiu1spiridon: Fixed in commit `c1d4593c`. Progress output in `_wait_for_platform_pod` is now explicitly redirected to stderr (`dim ... >&2`), ensuring stdout remains completely clean and diagnostic text is never captured into `line` or misinterpreted as task completion.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -675,6 +739,7 @@ print(r.text)
 " 2>/dev/null)"; then
         printf '%s\n' "$out"
     else
+        _wait_for_platform_pod || true
         warn_once sessions "could not read the agent's session list; treating it as empty — ${out:-no response} (issue #786)"
         echo '{}'
     fi
```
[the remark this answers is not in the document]
sergiu1spiridon: Fixed in commit `c1d4593c`. `_wait_for_platform_pod` now persists the resolved pod name to `${SCENARIO_RUN_DIR}/platform_pod` latch, and `_sync_platform_pod` synchronizes `PLATFORM_POD` from the latch across subshells and parent commands. In-loop re-resolution is also capped to 15s (`IN_LOOP_POD_RESOLVE_TIMEOUT`) and returns 1 rather than calling `die` to prevent premature `set -e` aborts.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -111,6 +111,8 @@ KEEP_DEDUP=0
 DRY_RUN=0
```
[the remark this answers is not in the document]
sergiu1spiridon: Updated the PR description with live validation details from GKE Autopilot cluster `platform-agent-host` (`us-east4`, namespace `kubeagents-system`), container images (`4044928181fb535865c24a592959b686af927719`), gateway pod `platform-agent-gateway-787c6d5d9f-hbfbm`, live E2E run links (#33749373944 and #33758170112), and confirmed pass of scenarios 03, 04, 05, 06, and 07.

deploy/docker/patches/kanban_scheduling.py
```diff
@@ -756,13 +756,12 @@ def release_dead_foreign_claims(
...
                     # ``detect_crashed_workers``.
                     continue
-        if pid_alive(pid):
-            # Either a real live worker or a PID collision. The TTL remains the
-            # backstop; never take a card away from something that might be
-            # running it.
-            continue
-
         fate = classify_reclaim(lock, claimer_id, row["claimed_at"], process_start)
+        if fate != POD_REPLACED and pid_alive(pid):
```
mplakhtiy: **HIGH — the liveness bypass is untested, and the docstring it contradicts still says the opposite.**

Behaviour delta, measured rather than reasoned about — same board, same inputs, base module against PR module:

```
board row: ("gw", "running", pid=2001, claim_lock=OTHER_POD, claimed_at=PRE_BOOT)
sweep:     release_dead_foreign_claims(conn, NEW, ALIVE)     # pid_alive always True

BASE   released: []       status: running
PR     released: ['gw']   status: ready       (fate: pod_replaced)
```

A card held by a foreign pod whose worker PID is **alive** is now taken away. That is the outcome this module's own prose rules out. `kanban_scheduling.py:232-237`, unchanged by this diff:

> The liveness test is what makes the sweep safe in every caller, not just the gateway. A CLI invocation has its own `_claimer_id()` and would otherwise consider the live gateway's claims foreign — but the gateway's workers have live PIDs, so they are left strictly alone.

That paragraph is now false for the `POD_REPLACED` arm. The file states the property as load-bearing ("what makes the sweep safe in **every caller**") and the diff narrows it without touching a word of the docstring.

**Why the suite did not catch it.** `test_a_live_foreign_worker_is_left_to_the_ttl` (`test_kanban_scheduling.py:805`) is the test that documents this property — *"This is what makes the sweep safe from a CLI process."* It passes `fence_board([("gw", "running", 2001, OLD)])` with no `claimed_at`, and `fence_board` defaults `claimed_at` to NULL. NULL routes to `CLAIM_TIME_UNKNOWN`, not `POD_REPLACED`, so the one test guarding the invariant never reaches the line that changed. `test_rolled_pods_claims_come_back_without_waiting_out_the_ttl:816` does set `PRE_BOOT`, but passes `DEAD`. No test in the file combines `POD_REPLACED` with a live PID — the entire behaviour delta.

The PR body offers "108/108 tests passing" for this change. Base is also 108: the diff touches no test file under `deploy/docker/patches/`. The number says the change broke nothing; it says nothing about whether the change works.

**Failure scenario.** Any caller opening the board from a different pod name whose own process started after the claim was made. The exposure is narrow — `kubectl exec` into the gateway shares the hostname, so `claim_host` matches and the arm is `PROCESS_DIED_IN_PLACE` with `pid_alive` intact; and `useStatefulSet` (`platformagent_manifests.go:1570-1576`) gives each replica its own RWO PVC, hence its own board. The reachable case is a multi-replica **Deployment** on shared RWX storage, where each replica's sweep would hand itself cards its live siblings are working, uncharged.

The change is well-motivated — the docstring itself laments that PIDs restart from 1 in a new container, so `pid_alive` in a replacement pod is close to a coin flip. But narrowing a stated safety invariant needs the test that pins the new boundary (`POD_REPLACED` + `ALIVE` releases; `PROCESS_DIED_IN_PLACE` and `CONCURRENT_OWNER` + `ALIVE` still hold) and the docstring edit at lines 232-237 that stops the file asserting an unconditional liveness test. Both are absent.

Worth noting this is the only production line in a PR whose title, body and ~95% of the diff are test scaffolding. It may deserve its own PR.

tests/e2e/test_stockout_investigation.py
```diff
@@ -524,6 +530,73 @@ def _verify_skill_mounted(
...
+) -> str:
+    """Waits until the platform-agent gateway workload is rolled out, ready, and responsive."""
+    kind, workload_name = _gateway_workload(agent_ref, namespace)
+    if kind is None:
+        pytest.fail(
+            f"No Deployment or StatefulSet '{workload_name}' in namespace '{namespace}'; "
+            f"PlatformAgent '{agent_ref}' has no gateway workload."
+        )
+    target = f"{kind}/{workload_name}"
+    deadline = time.time() + timeout
```
mplakhtiy: **MEDIUM — this deadline is shared with the rollout wait that runs before the loop it bounds.**

`deadline` is fixed here, then the `rollout status` call on the next line is allowed to consume all of `timeout`. A rollout that returns success near the boundary leaves the pod-probe loop at line 577 zero or one iteration, and the test fails with *"Platform Agent pod for … not available and ready within 180s"* — a message about the pod when the time actually went to the rollout.

Either budget the two separately, or recompute the deadline after `rollout status` returns.

The added `test_unready_pod_raises_when_timeout_exhausted` passes `timeout=0`, so it exercises the zero-iteration path but not the interaction that produces it in the field.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -291,14 +335,22 @@ preflight() {
     _resolve_platform_pod
```
mplakhtiy, line 336: **MEDIUM — preflight's worst case is now 360s, against 300s of headroom.**

`_resolve_platform_pod` here is bounded by `AGENT_AVAILABILITY_TIMEOUT=180`, and the new skill-mount loop below it is bounded by the same 180. They are sequential, so preflight's worst case is 360s. The scenario subprocess is capped at `watch_timeout + _SCENARIO_RUN_HEADROOM_SECONDS` = `360 + 300` = 660s (`test_stockout_investigation.py:912`). In the worst case the cap fires during preflight and the failure surfaces as *"Stockout Scenario … timed out"*, pointing at the investigation rather than at the wait that consumed the budget.

In practice `_wait_for_agent_available` runs in-process first, so by the time the script starts the pod is ready and both loops return on their first iteration. Nothing enforces that ordering and it is not stated anywhere. Sizing the headroom against the waits it now has to cover, or a comment recording why 300 is enough, would close it.

mplakhtiy (commented):
Adversarial review of the full diff at `c1d4593c`, run in a subagent that did not write the change, against `main` (merged clean, no conflicts).

Five findings, Medium and above, all anchored inline below. One HIGH.

For the record, three things I checked and found **fine**: the `>&2` redirect in `_wait_for_platform_pod` (`dim` does print to stdout at `common.sh:132`), the `${SCENARIO_RUN_DIR}/platform_pod` latch with `_sync_platform_pod` at every parent call site, and the `start_time + WATCH_TIMEOUT` wall-clock deadline. The code half of all three replies to the bot holds up. Also clean: the `-l app=platform-agent-gateway` selector matches what the operator sets, `verify.sh` Step 0's hardcoded `deployment/` is not a new constraint, the `startswith(ROUTE_NAME)` filter matches the convention `_task_after` already used, and there is no `set -euo pipefail` hazard in the new shell.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -291,14 +335,22 @@ preflight() {
...
 
-    # $HERMES_HOME is /opt/data in the shipped image, but read it rather than assume:
-    # a plugin that mounted at the wrong path is exactly the failure this catches.
-    if kmgmt exec -i=false -n "$AGENT_NAMESPACE" "$PLATFORM_POD" -c platform-agent -- \
-        sh -c 'test -f "${HERMES_HOME:-/opt/data}/profiles/platform/plugins/'"${PLUGIN_NAME}"'/skills/gke-stockout-investigator/SKILL.md"' 2>/dev/null
-    then
+    # Wait up to AGENT_AVAILABILITY_TIMEOUT for the skill to be readable in the resolved pod
+    _sync_platform_pod
+    local skill_waited=0 skill_found=0
+    while [ "$skill_waited" -lt "$AGENT_AVAILABILITY_TIMEOUT" ]; do
```
mplakhtiy: **MEDIUM — this retry adds 180s to the permanent-failure path without changing its outcome.**

Previously one probe, then `warn`. Now up to 180s of polling, then the same `warn` at line 353, then the run continues to a watch that cannot succeed because the skill is not there. A genuinely unmounted plugin now costs three extra minutes per scenario to reach the same verdict.

The retry helps the *transient* case — pod just came up, mount lagging — which is the point. But that case resolves in seconds, so a separate, much shorter constant would buy it without paying 180s for the permanent case.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -313,13 +365,41 @@ preflight() {
...
+        tid = t.get('id')
+        title = str(t.get('title', ''))
+        status = t.get('status')
+        if tid and title.startswith('${ROUTE_NAME}') and status in ('running', 'claimed', 'ready', 'blocked', 'todo'):
+            if status in ('running', 'claimed'):
+                subprocess.run(['hermes', 'kanban', 'reclaim', tid], env=cmd_env, capture_output=True)
+            subprocess.run(['hermes', 'kanban', 'archive', tid], env=cmd_env, capture_output=True)
+except Exception:
+    pass
+" >/dev/null 2>&1 || true
```
mplakhtiy: **MEDIUM — the kanban cleanup reports success even when it did nothing.**

Failures are swallowed at three levels: the in-pod `except Exception: pass` just above, the `subprocess.run(..., capture_output=True)` calls at lines 387-388 whose return codes are never read, and this outer `>/dev/null 2>&1 || true`. Then `clear_dedup` prints `ok "cleared the adapter dedup registry and kanban route tasks"` unconditionally at line 402. `_clean_stale_kanban_tasks` (`test_stockout_investigation.py:533-553`) has the same shape — its `_kubectl` result is discarded.

So if `hermes kanban reclaim` is not a subcommand, or `archive` refuses a `running` card, the slot starvation this PR exists to fix persists and every transcript says it was fixed. The board state is the one thing worth asserting here, and nothing asserts it. Checking `hermes kanban ls --json` by hand is evidence the happy path works, but that is a human looking once, not the script noticing.

At minimum, count what was archived and say so, or let a non-zero exit surface.

deploy/docker/patches/kanban_scheduling.py
sergiu1spiridon: Reverted `deploy/docker/patches/kanban_scheduling.py` back to `upstream/main`. This removes the runtime patch completely so this PR is strictly scoped (100%) to test harness and scenario fixtures. The active card reclamation and archiving performed during scenario preflight/cleanup in `common.sh` and `test_stockout_investigation.py` already ensures concurrency slots are clear without modifying the container image logic here.

tests/e2e/test_stockout_investigation.py
sergiu1spiridon: Recomputed `deadline = time.time() + timeout` immediately after `_kubectl("rollout", "status", ...)` returns success. The rollout status check and the pod readiness probe loop now have completely independent budgets, ensuring that a slow rollout does not starve the pod probe loop or misattribute timeout failures. Added `test_slow_rollout_does_not_consume_probe_budget` in `test_stockout_fixture_helpers.py` to verify this behavior.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
sergiu1spiridon: Preflight wait budgets have been restructured. Added `SKILL_MOUNT_TIMEOUT="${SKILL_MOUNT_TIMEOUT:-30}"` so the skill-mount check uses at most 30s instead of 180s. Total worst-case preflight duration is now 180s (`_resolve_platform_pod`) + 30s (`SKILL_MOUNT_TIMEOUT`) = 210s, which is well within the 300s headroom (`_SCENARIO_RUN_HEADROOM_SECONDS`). Documented the headroom sizing in `test_stockout_investigation.py`.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
sergiu1spiridon: Introduced dedicated `SKILL_MOUNT_TIMEOUT="${SKILL_MOUNT_TIMEOUT:-30}"` for the skill mount loop in `common.sh`. In transient cases (such as just after pod rollout) the mount resolves in seconds, while permanent failure paths now exit and warn after 30s instead of consuming 180s.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
sergiu1spiridon: Updated both `cleanup_kanban` in `common.sh` and `_clean_stale_kanban_tasks` in `test_stockout_investigation.py`:
- `hermes kanban ls` failures and non-zero exits from `reclaim`/`archive` are no longer swallowed; errors surface to stderr and fail with non-zero exit / raise `RuntimeError`.
- Runs a post-cleanup `hermes kanban ls` verification to assert 0 active tasks remain for the route on the board.
- Counts and logs the number of tasks archived (`ok "archived N stale kanban task(s) for ${ROUTE_NAME}"` or `dim "kanban board clean for ${ROUTE_NAME} (0 active tasks)"`).
- Added unit tests in `test_stockout_fixture_helpers.py` (`CleanStaleKanbanTasksTest`) validating count extraction and error raising.

agentplugins/gke-stockout-investigator/scenarios/lib/common.sh
```diff
@@ -313,13 +368,79 @@ preflight() {
...
+' "$ROUTE_NAME" 2>&1)"; then
+        local count="${out##*$'\n'}"
+        if [ "$count" -gt 0 ] 2>/dev/null; then
+            ok "archived ${count} stale kanban task(s) for ${ROUTE_NAME}"
+        else
+            dim "kanban board clean for ${ROUTE_NAME} (0 active tasks)"
+        fi
+    else
+        warn "kanban cleanup encountered an error: ${out}"
+    fi
```
mplakhtiy, line 396: **MEDIUM — `cleanup_kanban` swallows every error and cannot fail.**

The in-pod program exits 1 on a reclaim failure, an archive failure, or a task still active after the verify pass. The wrapper turns all of that into a `warn` and returns 0. Nothing downstream can distinguish a clean board from a cleanup that did nothing.

That includes the pre-run path: `clear_dedup` calls `cleanup_kanban` at `:443`, which is exactly where a swallowed failure reproduces the `max_in_progress = 2` slot starvation this change exists to fix. The scenario then runs against a board whose slots are already consumed, and the only trace is a `warn` line.

The two call sites added on the Python side do the same thing more explicitly — `_clean_stale_kanban_tasks` is wrapped in `except Exception: pass` at `tests/e2e/test_stockout_investigation.py:886-890` and `:970-974`.

Fix: propagate a non-zero return from `cleanup_kanban` and let the caller decide, or state in a comment why board-hygiene failure is non-fatal on this path.

tests/e2e/test_stockout_investigation.py
```diff
@@ -524,6 +531,110 @@ def _verify_skill_mounted(
     )
 
 
+def _clean_stale_kanban_tasks(pod: str, namespace: str, route_name: str) -> int:
+    """Reclaims and archives any lingering kanban tasks for route_name so slots are open.
+
+    Returns the count of tasks archived, or raises RuntimeError if cleanup or verification fails.
+    """
```
mplakhtiy: **MEDIUM — the kanban cleanup program now exists twice, hand-transcribed, and nothing tests one copy.**

`agentplugins/gke-stockout-investigator/scenarios/lib/common.sh:377-421` is the same ~35-line program as `:539-579` here: same statuses, same `startswith(route_name)` filter, same reclaim-then-archive, same verify pass — transcribed into a second quoting regime, a single-quoted shell literal against a Python string of `"…\n"` fragments.

`9f66d929` is the evidence rather than the hypothetical: it makes the identical `cmd_env = dict(os.environ)` / `setdefault("HERMES_HOME", …)` fix in both copies, in one commit, because the same bug was in both.

No test reaches the shell copy — nothing under `tests/` references `scenarios/lib/common.sh`. The Python copy has three unit tests and all three mock `_kubectl`, so they cover the wrapper's return and raise, not the program. The shell copy is the one the RC promotion gate runs.

**Failure scenario.** The next fix lands in the Python copy, the shell copy keeps the bug, and the scenario harness — the path the gate exercises — quietly stops archiving. Nothing fails; slots starve again.

Extracting the program to one file the pod can be handed, or having the shell call the same helper, closes it. Short of that, a comment in each copy naming the other is the minimum.

tests/e2e/test_stockout_investigation.py
```diff
@@ -714,6 +825,7 @@ def ensure_stockout_plugin_installed(
         workload_name,
         budget_deadline,
     )
+    _clean_stale_kanban_tasks(pod, agent_namespace, os.environ.get("STOCKOUT_ROUTE", _DEFAULT_ROUTE_NAME))
```
mplakhtiy: **MEDIUM — making the cleanup fatal turns a board race into a whole-suite error.**

`_clean_stale_kanban_tasks` raises `RuntimeError` when `archive` returns non-zero, or when the post-cleanup `ls` still shows any route task in `running|claimed|ready|blocked|todo`. This call sits in the session-scoped fixture, so the raise errors setup for every test using `ensure_stockout_plugin_installed`.

A card is reclaimed and then archived as two separate `hermes` invocations. If the scheduler re-claims it in between — which is what a scheduler does with a card that just returned to `ready` — `archive` fails, `errors` is non-empty, the in-pod program exits 1, and the suite dies at setup with "Kanban cleanup failed on pod …". The verify `ls` has the same shape: an alert arriving during it is indistinguishable from a stale card. Nothing retries either.

Before this change there was no way for board hygiene to fail a test. The inconsistency is the tell — the same call is fatal here and `except Exception: pass` at the two `finally` sites (`:886-890`, `:970-974`), with no comment saying why the two differ.

A bounded retry around reclaim→archive, or treating the verify pass as a warning rather than a raise, keeps the diagnostic without handing the suite a new way to go red.

tests/e2e/test_stockout_investigation.py
```diff
@@ -524,6 +531,110 @@ def _verify_skill_mounted(
...
+    # Recompute the deadline so pod readiness and skill mount have their own dedicated budget,
+    # rather than sharing a deadline that a slow rollout could exhaust before the probe loop starts.
+    deadline = time.time() + timeout
+    ready_pod = None
+    while time.time() < deadline:
+        revision = _current_revision_selector(kind, workload_name, namespace)
+        pod, _ = _gateway_pod(agent_ref, namespace, revision)
+        if pod:
+            home = _agent_home(agent_ref, namespace)
+            skill_path = f"{home}/profiles/platform/plugins/{_PLUGIN_NAME}/skills/{_PLUGIN_SKILL_NAME}/SKILL.md"
```
mplakhtiy: **MEDIUM — `_wait_for_agent_available` re-implements the skill probe and hardcodes the profile.**

`_verify_skill_mounted` in this file (`:465-476`) derives the same path from the AgentPlugin CR's `targetProfile`, and falls back to `{home}/plugins/{_PLUGIN_NAME}` when it is empty, with a comment citing `deploy/shared/profile_plugins.py` for why the two layouts differ. `targetProfile` is a Helm value (`agentplugins/gke-stockout-investigator/values.yaml:7`), so it is overridable.

**Failure scenario.** An install that sets `targetProfile` to anything but `platform`, or leaves it empty: this wait probes a path that never exists, burns the full 180s, and fails with "Platform Agent pod for … not available and ready within 180s" — a message about pod readiness for a path problem — minutes after `_verify_skill_mounted` passed on the same pod for the same skill. The suite contradicts itself and blames the wrong layer.

Latent rather than live: the shipped default is `platform`, so it works today. Calling `_verify_skill_mounted`'s path logic instead of restating it removes the drift.

mplakhtiy (commented):
Full review of `9f66d929`. Four findings, all MEDIUM and all inline: the shell cleanup wrapper cannot fail, the cleanup program exists in two hand-transcribed copies with one of them untested, the Python cleanup is fatal inside the session fixture, and the new availability wait hardcodes the profile path.

mplakhtiy (commented):
One finding on the availability-probe path introduced by `32f9dfbc`, at MEDIUM. The four findings from the previous pass are all fixed on this head; verified against source.

tests/e2e/test_stockout_investigation.py
```diff
@@ -436,6 +445,38 @@ def _agent_home(agent_ref: str, namespace: str) -> str:
     return home or "/opt/data"
 
 
+def _plugin_target_profile(namespace: str) -> str:
+    """Reads targetProfile from the AgentPlugin CR, or empty string if unset."""
+    res = _kubectl(
+        "get", "agentplugins", _PLUGIN_NAME, "-n", namespace,
+        "-o", "jsonpath={.spec.targetProfile}",
+    )
+    return res.stdout.strip() if res.returncode == 0 else ""
```
mplakhtiy, line 454: An empty return here means "the CR has no `targetProfile`", and `_plugin_skill_path` acts on that by switching to the default-profile layout `{home}/plugins/{plugin}` instead of `{home}/profiles/{profile}/plugins/{plugin}`.

A failed `kubectl get agentplugins` returns the same empty string. That includes the `_KUBECTL_TIMEOUT_RC` case, because `_kubectl` is called here without `fail_on_timeout` — so a 30s API stall is indistinguishable from an unset field.

`_plugin_skill_path` is now called once, before the probe loop (`:617`), so a single unlucky moment poisons the whole 180s of probing rather than one iteration of it.

**Failure scenario.** The API server is slow for 30s while the availability probe starts. `_plugin_target_profile` times out and returns `""`. The default-profile path is computed and reused for every subsequent probe, none of which can match, and the run fails after 180s with *"Platform Agent pod for … not available and ready within 180s"* — a message about pod readiness for what is actually a path problem.

`_verify_skill_mounted` does not share this: its `target_profile` comes from the plugin object `_wait_for_plugin_ready` has already validated.

**Fix.** Either pass `fail_on_timeout=True` on the `_kubectl` call above, or distinguish "read failed" from "field is empty" and let the former propagate. `_kubectl`'s own docstring already argues for the distinction: *"a returned failure is worse than an exception … Both are false statements about the cluster derived from a slow API server."*

mplakhtiy (approved):

jayantid:
Same situation as #608 (see the note there): GitHub reports this PR `BLOCKED` while it sits in tide's merge batch — the batch can't complete and ~13 other passing PRs are capture-blocked behind it (#1202 mechanics). Check the Checks tab for missing/never-ran required contexts on this head; a push or rerun that produces them — or a brief `/hold` — releases the queue.

Issue #1117, opened 2026-09-01
ci: autopush and staging only ever get new images — Terraform never re-applies, so their infrastructure drifts from main indefinitely


mplakhtiy:
> **Blocked on #1081** for the automation half. The drift-visibility step (§8) and the Terraform
> idempotency fixes (§4) are actionable now; everything that runs the installer unattended against
> a long-lived environment waits for `install.env` to become an input. See
> [Proposed sequencing](#proposed-sequencing).

## Summary

`autopush` and `staging` are updated by **image tag only**. Nothing ever runs Terraform against
them, so every infrastructure change that lands on `main` — IAM roles, Pub/Sub, cluster settings,
node pools, the minter's GCP side, and the chart values the composition renders — is invisible in
both environments until somebody re-applies by hand.

`rc` and `nightly` do not have this problem: both are torn down and rebuilt from
`terraform/examples/full-install` on every pipeline run, so they always run today's composition.
The two long-lived environments run *last month's* composition with today's images.

`#504` (multi-repo GitHub support, merged 2026-09-01) is the trigger for filing this: it changed
`terraform/examples/full-install/main.tf`, `charts/kube-agents/values.yaml`,
`charts/kube-agents/templates/github-minter.yaml` and the CRD. The CRD and the chart templates will
reach autopush. The composition change will not.

## Evidence

Last successful `terraform apply`, read from the state objects:

| Environment | Project | State object last written | Age |
| --- | --- | --- | --- |
| `rc` | `kube-agents-rc` | 2026-09-01T15:47Z | minutes — reconciled every pipeline run |
| `autopush` | `kube-agents-autopush` | 2026-08-20T13:18Z | **12 days** |
| `staging` | `kube-agents-gkedemos` | 2026-08-21T12:18Z | **11 days** |

```
$ gcloud storage ls -l gs://kube-agents-autopush-kube-agents-tfstate/kube-agents/platform-agent-host/
    148920  2026-08-20T13:18:45Z  .../default.tfstate

$ gcloud storage ls -l gs://kube-agents-gkedemos-kube-agents-tfstate/kube-agents/platform-agent-host/
    148645  2026-08-21T12:18:06Z  .../default.tfstate

$ gcloud storage ls -l gs://kube-agents-rc-kube-agents-tfstate/kube-agents/platform-agent-host/
     38315  2026-09-01T15:47:17Z  .../default.tfstate
```

[...]

PR #1199, opened 2026-09-03
feat(ci): unify image publishing, consolidate autopush/staging atomic deployments, and protect plugin infrastructure

### LINKED ISSUE #1017: reusable-deploy-integrations.yml would scope a live GitHub App token at the release repo if ever called for rc or nightly

### LINKED ISSUE #1198: ci(deploy): six image-only redeploy workflows still bypass upgrade.sh, and nothing clears a wedged Helm release
Follow-up from #1117.

#1117 closed the Terraform half: `nightly-pipeline.yml` now applies the composition to both
long-lived environments once the E2E matrix is green (`step-4-reconcile-staging` at
`.github/workflows/nightly-pipeline.yml:166`, `step-6-reconcile-autopush` at `:266`), and
`drift-detect.yml` reports a non-empty plan daily. The image half is unchanged. Six workflows still
roll images onto `autopush` and `staging` with `helm upgrade` alone, on triggers that know nothing
about the reconcile, and CI still does not use the deploy path shipped to users.

This is the remainder of the plan in
https://github.com/gke-labs/kube-agents/issues/1117#issuecomment-5508690912 — its PRs 2, 3 and 4.
PR 1 of that plan landed as #1152, which is what closed #1117; PRs 2–4 were never filed anywhere.
Every line below was checked against `main` at `b0a4d7d1`.

## 1. Six redeploy workflows deploy outside `upgrade.sh`

Entry points, all still present:

- `.github/workflows/autopush-redeploy-agent.yml:3-8` — `workflow_run` on `Publish to GHCR`
- `.github/workflows/autopush-redeploy-controller.yml:3-8` — `workflow_run` on `Publish Controller to GHCR`
- `.github/workflows/autopush-redeploy-integrations.yml:3-18` — `push` to `main` on chart and `images.json` paths
- `.github/workflows/staging-redeploy-agent.yml:3-8`, `staging-redeploy-controller.yml:3-8`, `staging-redeploy-integrations.yml:3-8` — `push` on `staging_*` tags

All six call three reusable workflows whose payload is a bare `helm upgrade
--reset-then-reuse-values` on the release `helm_release.kube_agents` owns:
`.github/workflows/reusable-deploy-agent.yml:127-128`,
`.github/workflows/reusable-deploy-controller.yml:114`, and
`.github/workflows/reusable-deploy-integrations.yml:161`, `:227`, `:390`.

`upgrade.sh --upgrade-mode=full` (`upgrade.sh:106`) is the path users and
`.agents/skills/upgrade-kube-agents/SKILL.md` run for the same job. CI reaches it only through
`reconcile_environment.sh`, on the nightly and on manual dispatch, never on a deploy.

Consequences that hold today:

- The three components deploy on independent triggers, so `autopush` routinely runs an agent image
  from one commit and an operator image from another.
- A staging promotion is two operations against one release: the nightly applies the composition
  (`nightly-pipeline.yml:166`) and then pushes the tag that starts three more `helm upgrade`s. The
  ordering comment at `nightly-pipeline.yml:176-182` exists because they contend.
- The redeploy path takes no live-test lease. `scripts/live_test_lease.py` is consulted only by
  `scripts/release/reconcile_environment.sh:201` and `deploy-environment.yml`, so an image roll
  restarts pods under an agent that holds the lease.

## 2. Autopush redeploys on every image publish, and can starve its own reconcile

`autopush-redeploy-{agent,controller}.yml` fire on every GHCR publish, which is every push to
`main`. `scripts/release/reconcile_environment.sh:131-163` waits out in-flight redeploys before an
apply because both touch the same release; the budget is `REDEPLOY_WAIT_SECONDS=900`
(`reconcile_environment.sh:36`), and past it the reconcile emits a warning and exits `deferred`
without applying (`:161`, `:165-173`). On a busy day the deploy rate is what decides whether the
infrastructure fix from #1117 runs at all.

The comment proposes a candidate-driven schedule instead: `find_latest_built_commit`
(`scripts/release/common.sh:620`) picks the newest commit whose required images are all in GHCR, a
no-op check skips when that already matches what is deployed, the lease is checked, and one atomic
`upgrade.sh` invocation deploys images, chart and composition from that single commit.

## 3. A wedged Helm release is still cleared by hand

`scripts/helm_upgrade_retry.sh` polls the release lock every 30s for a 2100s budget
(`:28-29`, `:34-60`), then fails and prints the `helm rollback` command for a human to run
(`:62-66`). Nothing anywhere calls `helm rollback`; there is no `ensure_clean_helm_release` in
`upgrade.sh` or `scripts/installer/`. The script's own comment gives the reason it does not roll
back — from inside the loop, a lock held by a healthy deploy is indistinguishable from a wedged
release — which is an argument for a pre-flight check that reads `helm status` before starting,
not against clearing the lock at all.

## Scope

1. Replace the three `staging-redeploy-*.yml` with one workflow on `push: tags: staging_*` plus
   `workflow_dispatch`, running `upgrade.sh --non-interactive --upgrade-mode=full
   --image-tag=<peeled commit sha>`.
2. Add pre-flight recovery for a `pending-*` release to `upgrade.sh` / `scripts/installer/`:
   read `helm status`, roll back to the last `deployed` revision, then upgrade.
3. Replace the three `autopush-redeploy-*.yml` with one scheduled workflow plus
   `workflow_dispatch`, wiring candidate discovery, the no-op check, the lease check and the
   recovery above.
4. Delete `reusable-deploy-{agent,controller,integrations}.yml` and
   `scripts/helm_upgrade_retry.sh` once nothing calls them, and drop the redeploy-waiting block in
   `reconcile_environment.sh:131-163` that exists only to work around them.

The cadence in step 3 is a decision, not a given. The comment argues for `0 */2 * * *`; the
alternative is to keep deploying per publish and accept the contention.

## Related

- #1117 — the parent issue, closed by #1152
- #911 — Create Drift Detector
- #1017 — `reusable-deploy-integrations.yml` would scope a live GitHub App token at the release repo if called for `rc` or `nightly`; step 4 above deletes that file

eLeontev:
## Summary

This pull request completes the CI/CD and deployment infrastructure stabilization roadmap outlined in #1117 and #1198 (Scopes PR2, PR3, PR4), consolidating fragmented image publishing and redeployment pipelines into unified, atomic workflows driving `./upgrade.sh --upgrade-mode=full`, adding pre-deployment image verification, implementing safe Helm self-healing, and protecting Cloud Pub/Sub and Logging plugin infrastructure during environment reconciliation.

## Why This Change

Following #1117 and #1152 (which closed PR1 of the infra stabilization plan), the deployment and image roll pipelines remained fragmented:
1. Six separate redeploy workflows rolled controller, agent, and integrations images onto `autopush` and `staging` independently via bare `helm upgrade` commands rather than the production `./upgrade.sh --upgrade-mode=full` entry point. This caused component version skew, lease contention between image rolls and scheduled reconciliations, and deadlocks on Helm release locks.
2. In PR #1068, first-party plugin OCI images (`pubsub-platform` and `gke-stockout-investigator`) were introduced, but `ENABLE_PUBSUB_PLATFORM` and `ENABLE_STOCKOUT_INVESTIGATOR` were not mapped in `render_install_env.sh` and `reconcile-environment.yml`. As a result, unattended environment reconciliations and daily drift detections planned destruction of Cloud Pub/Sub topics, subscriptions, and Cloud Logging sinks.
3. Issue #1017 identified an over-privileged minter scope in `reusable-deploy-integrations.yml`, which is safely retired and deleted with the consolidation into atomic workflows.

## What Changed

- **Unified Container Image Publishing (`.github/workflows/docker-publish-ghcr.yml`)**:
  - Consolidated standalone operator and agent publishing workflows into a single authoritative workflow.
  - Builds, tags with immutable commit SHA, and cosign-signs all 7 first-party OCI images: `k8s-operator`, `platform-agent`, `credential-proxy`, `agent-sandbox`, `replay-proxy`, `pubsub-platform`, and `gke-stockout-investigator`.
  - Harmonized tagging across the entire suite (commit SHA + `latest`).
- **Atomic Deployments for Long-Lived Environments (`autopush-deploy.yml` & `staging-deploy.yml`)**:
  - Replaced 9 legacy fragmented workflows (`{autopush,staging}-redeploy-{agent,controller,integrations}.yml` and `reusable-deploy-*.yml`) with two unified deployment workflows.
  - Executes `./upgrade.sh --upgrade-mode=full` through `reconcile-environment.yml`, ensuring Terraform composition, Helm values, and container images deploy atomically from a single verified commit SHA.
  - Decoupled workflow-level concurrency from job concurrency to prevent dispatch deadlocks.
  - Bound `environment: autopush` and `environment: staging` to candidate resolution jobs to correctly scope `vars.REGISTRY_PREFIX`.
  - Cleaned up obsolete `force_reconcile` input and unreachable no-op branches.
- **Pre-Deployment Image Verification & Candidate Resolution**:
  - Introduced `scripts/release/resolve_deploy.sh` to resolve target candidate commits deterministically without ambiguous fallbacks.
  - Added `scripts/release/verify_candidate_images.sh` to verify all 7 required release images in GHCR prior to environment reconciliation.
- **Decoupled Nightly Staging Promotion Tagging (`.github/workflows/nightly-pipeline.yml`)**:
  - Step 4 derives and pushes an annotated `staging_<ts>_<sha>` promotion tag via `RELEASE_BOT_TOKEN` only after a green E2E test matrix, triggering `staging-deploy.yml` asynchronously.
- **Protection of Plugin Infrastructure during Reconciliation**:
  - Added `ENABLE_PUBSUB_PLATFORM` and `ENABLE_STOCKOUT_INVESTIGATOR` to `MAPPING` and `REQUIRED_STRICT` in `scripts/release/render_install_env.sh` and exposed them in `reconcile-environment.yml`.
- **Safe Helm Pre-Flight Self-Healing (`scripts/installer/installer_common.sh`)**:
  - `ensure_clean_helm_release` detects active in-flight operations, waits out active deployments with polling, recovers wedged locks via rollback to the last known good revision, and strictly refuses destructive automatic `helm uninstall` without explicit opt-in (`ALLOW_UNINSTALL_PENDING_RELEASE=true`).
- **Post-Upgrade Gate Scoping (`upgrade.sh`)**:
  - Scoped `confirm_agent_image.sh` to `harness` and `full` upgrade modes, guarded by `deployment/platform-agent-gateway` existence, preventing spurious timeouts in `operator` mode.
- **Signal & Exit Status Preservation (`scripts/release/reconcile_environment.sh`)**:
  - Replaced catch-all traps with dedicated signal handlers (INT, TERM, HUP) that release the lease and terminate immediately, preserving actual non-zero exit codes.
- **Conformance Test Harness & Invariant Support**:
  - Updated `tests/conformance/_harness.py` and `tests/conformance/test_B_write_path.py` (Invariant B4) to evaluate `needs` prerequisite job dependency chains for workflow gating and updated the autopush workflow reference.

## Context

- Resolves #1198
- Resolves #1017
- Follow-up to #1117 and #1152
- Incorporates plugin image lifecycle from #1068 and agent sandbox from #913

## Testing

- Ran full test suite across the repository: unit tests in `tests/` passing cleanly.
- Verified test suites for affected components:
  - `tests/test_nightly_pipeline_wiring.py` (OK)
  - `tests/test_reconcile_environment.py` (OK, 49 tests)
  - `tests/test_upgrade_script.py` (OK, 20 tests)
  - `tests/test_deploy_image_confirmation.py` (OK, 24 tests)
  - `tests/test_installer_common.py` (OK, 57 tests)
  - `python3 tests/conformance/run.py` (OK, 108 tests passing, 12 expected failures)
  - `./hack/check-image-inventory.sh` (OK)
- Formatted YAML workflows with Prettier and validated GitHub Actions inputs with Checkov annotations.

### Live validation

Not live-tested against a live cluster. This pull request modifies CI/CD deployment pipelines, pre-deployment validation scripts, and packaging workflows. Per `.github/PULL_REQUEST_TEMPLATE.md` (lines 56–58), live validation against long-lived autopush/staging clusters is reserved for CI workflow execution once merged or dispatched through the workflow test matrix. All script logic, parameter passing, error conditions, and Helm status state machines have been verified using hermetic subshell test harnesses with mocked commands.

## Self-Review

- **Pass 1 (Adversarial review - concurrency & deadlocks)**: Inspected workflow and job concurrency declarations in `autopush-deploy.yml` and `staging-deploy.yml`. Found collision between workflow-level and job-level concurrency groups (`autopush-deploy.yml`), which would deadlock queued jobs against the workflow itself. Resolved by separating workflow-level concurrency naming.
- **Pass 2 (Adversarial review - error handling & traps)**: Inspected `scripts/release/reconcile_environment.sh`. Found that manual `release_lease` invocation swallowed non-zero exit codes before `exit $status`. Replaced combined traps with dedicated signal handlers (INT, TERM, HUP) that release the lease and terminate immediately with standard exit codes.
- **Pass 3 (Adversarial review - redundant wait blocks)**: Found that `await_redeploys` and the fast no-op exit in `reconcile_environment.sh` were obsolete following atomic workflow consolidation and bypassed complete Terraform drift remediation. Removed both sections cleanly, along with obsolete `force_reconcile` inputs.
- **Pass 4 (Adversarial review - post-upgrade verification)**: Identified that `confirm_agent_image.sh` in `upgrade.sh` ran during `--upgrade-mode=operator` where gateway images are not updated, causing 5-minute timeouts. Confined verification to `harness` and `full` modes and guarded with deployment existence check.
- **Pass 5 (Adversarial review - Helm self-healing & secret preservation)**: Identified that `ensure_clean_helm_release` raced with active in-flight deployments and unconditionally ran `helm uninstall` on `pending-install` or missing history, destroying `PlatformAgent` and `platform-agent-secrets`. Added in-flight age detection with wait-out polling, and blocked automatic uninstall by default (`ALLOW_UNINSTALL_PENDING_RELEASE=true` required).
- **Pass 6 (Rebase & image alignment)**: Cleanly rebased onto `upstream/main` after PR #913 (`agent-sandbox`) and PR #1197. Harmonized `docker-publish-ghcr.yml` tags across all 7 images and updated the conformance harness for B4 invariant verification.

## Risk & Rollout

Low to medium risk. Changes consolidate existing CI deploy workflows into unified atomic workflows without modifying core agent or operator runtime application code. The new atomic deploy paths use the same standard `upgrade.sh --upgrade-mode=full` entry point recommended for production installations. Rollback: revert this PR to restore the previous independent redeploy workflows if unexpected CI promotion issues arise.


mplakhtiy:
`tests/test_gateway_rollout_budgets.py` comes out of this PR covering fewer callers than it went in with, and the two it drops are the ones that run on every redeploy.

That file exists to hold `startupProbe budget < rollout gate < progressDeadlineSeconds` for every caller that waits on these Deployments. It read four sources before this change and reads three after, and the swap is not one-for-one:

- **Gained:** `upgrade.sh` — the gateway gate through `GatewayRolloutBudgetTest`, and the controller-manager gate through the new `UpgradeRolloutGateTest`.
- **Lost:** `reusable-deploy-agent.yml`'s 900s gateway gate — `_AGENT_WORKFLOW` is deleted and `setUp` repointed at `_UPGRADE_SCRIPT` — and both of `reusable-deploy-integrations.yml`'s 420s gates for `litellm` and `github-token-minter`, since `IntegrationsRolloutGateTest` is removed outright.

Those two workflows are what `autopush-redeploy-*` and `staging-redeploy-*` call, so once this merges nothing fails if someone raises `agentAPIProbe`'s failure threshold or lowers `gatewayProgressDeadlineSeconds` and leaves the workflow gates behind. That is the regression the file's own docstring says it was written for, citing `wait_for_gke_readiness.sh` having "sat outside this file's scope and ran a single 300s gate for both".

Two smaller things follow from the same hunk:

- The docstring still introduces the file as the companion "for the Deployment the redeploy workflows actually gate on", which is now the caller it no longer reads.
- The comment above `_DEFAULT_PROGRESS_DEADLINE_SECONDS` reads "The integrations Deployments below rely on it rather than setting their own", and there are no integrations Deployments below it any more.

Keeping `_AGENT_WORKFLOW` and `_INTEGRATIONS_WORKFLOW` alongside `_UPGRADE_SCRIPT`, and parameterising `GatewayRolloutBudgetTest` over both gateway callers, would cover all four without giving anything up.

For context on where this came from: I hit the 120s gateway gate failing a nightly reconcile on both `autopush` and `staging` today, wrote the same `upgrade.sh` 120s→900s change on a separate branch, and found this PR during the pre-PR review. I dropped that branch in favour of this one — it carries the `confirm_agent_image.sh` read-back guard mine did not. This is the one finding from that review that applies here.

.github/workflows/autopush-deploy.yml
```diff
@@ -0,0 +1,101 @@
...
+        type: boolean
+
+permissions:
+  contents: read
+
+concurrency:
+  group: autopush-environment
+  cancel-in-progress: false
+
+jobs:
```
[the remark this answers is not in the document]
eLeontev: Resolved in 616436ac: Renamed workflow-level concurrency groups to `autopush-deploy` and `staging-deploy`, decoupling them from job-level concurrency groups so queued runs no longer deadlock.

.github/workflows/staging-deploy.yml
```diff
@@ -0,0 +1,91 @@
...
+        type: boolean
+
+permissions:
+  contents: read
+
+concurrency:
+  group: staging-environment
+  cancel-in-progress: false
+
+jobs:
```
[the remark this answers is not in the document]
eLeontev: Resolved in 616436ac: Renamed workflow-level concurrency groups to `autopush-deploy` and `staging-deploy`, decoupling them from job-level concurrency groups so queued runs no longer deadlock.

scripts/release/reconcile_environment.sh
```diff
@@ -116,6 +116,31 @@ fi
...
+  if [ -n "${running_tag}" ] && [ "${running_tag}" = "${IMAGE_TAG}" ]; then
+    echo "==> '${ENV_NAME}' is already running ${IMAGE_TAG}. No changes to apply."
+    summary "### No-op — \`${ENV_NAME}\`"
+    summary ""
+    summary "\`${ENV_NAME}\` is already running \`${IMAGE_TAG}\`. No deployment required."
+    output "result" "noop"
+    exit 0
+  fi
+fi
+
```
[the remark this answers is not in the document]
eLeontev: Resolved in `6f4a4334`: Removed the premature Section 2b fast no-op image tag exit, ensuring full Terraform reconciliation runs so partial or drifted configurations are always caught and reconciled.

scripts/release/reconcile_environment.sh
```diff
@@ -135,23 +160,19 @@ await_redeploys() {
...
+        jq --arg current_id "${GITHUB_RUN_ID:-}" \
+        '[.[] | select(.status == "in_progress" or .status == "queued" or .status == "waiting") | select((.databaseId | tostring) != $current_id)] | length' 2>/dev/null)"; then
+        n="${query_out:-0}"
+      fi
+      if [ -n "$n" ] && [ "$n" != "0" ]; then
+        running="${running} ${wf}(${n})"
+      fi
     done
     [ -n "$running" ] || return 0
     echo "==> Waiting for in-flight redeploys:${running}"
```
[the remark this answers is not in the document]
eLeontev: Resolved in `6f4a4334`: Dropped obsolete `await_redeploys` and `REDEPLOY_*` constants following consolidation into atomic deploy workflows (#1198 Scope 4), eliminating deadlocks against queued sibling deploys.

upgrade.sh
```diff
@@ -838,8 +848,11 @@ main() {
...
     kubectl rollout status deployment/kube-agents-controller-manager -n kubeagents-system --timeout=120s
   fi
+  if [ -n "$PARAM_IMAGE_TAG" ] && [ -f "${repo_dir}/scripts/confirm_agent_image.sh" ]; then
+    "${repo_dir}/scripts/confirm_agent_image.sh" "$target_namespace" platform-agent-gateway "$PARAM_IMAGE_TAG"
+  fi
   if [ "$PARAM_UPGRADE_MODE" = "harness" ] || [ "$PARAM_UPGRADE_MODE" = "full" ] || [ "$restarted_agent" = "true" ]; then
-    kubectl rollout status deployment/platform-agent-gateway -n kubeagents-system --timeout=120s
+    kubectl rollout status deployment/platform-agent-gateway -n kubeagents-system --timeout=900s
   fi
   print_success "Upgraded deployments verified healthy."
```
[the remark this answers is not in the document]
eLeontev: Resolved in `f19774e0`: Scoped `confirm_agent_image.sh` verification to `harness` and `full` upgrade modes (the modes that actually retag the agent), and guarded the call with a check for `deployment/platform-agent-gateway` existence.

upgrade.sh
```diff
@@ -738,7 +738,17 @@ main() {
...
+    # In plan mode, do not mutate state with a rollback, but warn if release is stuck
+    local current_helm_status
+    current_helm_status="$(helm_release_status kube-agents "$target_namespace")"
+    if [[ "$current_helm_status" =~ ^pending- ]]; then
+      print_warning "Helm release 'kube-agents' is currently in '${current_helm_status}'. Note: rollback is skipped in plan mode."
+    fi
+  fi
   # NAMESPACE steers the generator's Secret-recovery reads (install.env omits
   # credentials when PERSIST_SECRETS_ON_DISK=false; the live Secret has them).
   NAMESPACE="$target_namespace" \
```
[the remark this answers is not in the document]
eLeontev: Resolved in `17fa2e52`: Added in-flight operation age detection with wait-out polling, and blocked automatic `helm uninstall` by default to protect `PlatformAgent` and `platform-agent-secrets`, requiring explicit `ALLOW_UNINSTALL_PENDING_RELEASE=true` opt-in.

.agents/rules/github_actions.md
```diff
@@ -20,7 +20,7 @@ mails the fork owner. Put the guard on each job rather than trusting the skip to
 `needs`; an `always()` added later removes the implicit `success()` and the job runs anyway.
```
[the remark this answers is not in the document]
eLeontev: Resolved: Updated the pull request description to strictly adhere to `.github/PULL_REQUEST_TEMPLATE.md` with all required sections (Why This Change, What Changed, Context, Live validation, Self-Review, Risk & Rollout) fully populated.

eLeontev:
Hi @mplakhtiy! Thank you for the detailed review and context!

Regarding `_AGENT_WORKFLOW` (`reusable-deploy-agent.yml`) and `_INTEGRATIONS_WORKFLOW` (`reusable-deploy-integrations.yml`):
As part of the infrastructure stabilization roadmap in #1198 (Scopes 1, 3, and 4), this PR completely retires and deletes all 6 standalone redeploy workflows (`{autopush,staging}-redeploy-{agent,controller,integrations}.yml`) and the three reusable workflows (`reusable-deploy-{agent,controller,integrations}.yml`).

They are replaced by the two atomic workflows (`autopush-deploy.yml` and `staging-deploy.yml`), which invoke `reconcile-environment.yml` and `./upgrade.sh --upgrade-mode=full`. Therefore, `reusable-deploy-*.yml` no longer exist in the repository or run on redeploys.

Because `upgrade.sh` is now the single authoritative deploy entry point for both human operators and CI deploys across long-lived environments, `test_gateway_rollout_budgets.py` was repointed directly at `upgrade.sh`.

Good catch on the leftover docstring and comment references to the retired redeploy workflows and integrations Deployments — in commit `971416f6` we've updated the module docstring and the comment above `_DEFAULT_PROGRESS_DEADLINE_SECONDS` to accurately describe `upgrade.sh` and `wait_for_gke_readiness.sh`.

scripts/installer/installer_common.sh
```diff
@@ -669,6 +669,134 @@ if isinstance(value, str) and value:
...
+        fi
+      else
+        if type print_warning >/dev/null 2>&1; then
+          print_warning "Helm release '${release_name}' is in '${release_status}', but no previous deployed revision exists in history. Uninstalling to recover..."
+        else
+          echo "⚠️ WARNING: Helm release '${release_name}' is in '${release_status}', but no previous deployed revision exists in history. Uninstalling to recover..." >&2
+        fi
+        helm uninstall "${release_name}" -n "${namespace}" --wait
+        return 0
+      fi
```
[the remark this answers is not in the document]
eLeontev: Resolved in `17fa2e52`: Blocked automatic `helm uninstall` by default to protect `PlatformAgent` and `platform-agent-secrets`, requiring explicit `ALLOW_UNINSTALL_PENDING_RELEASE=true` opt-in. Stuck releases without a deployed revision now exit 1 with a clear error rather than destroying release secrets.

upgrade.sh
```diff
@@ -838,8 +848,11 @@ main() {
     # operator Deployment with the release name.
     kubectl rollout status deployment/kube-agents-controller-manager -n kubeagents-system --timeout=120s
   fi
+  if [ -n "$PARAM_IMAGE_TAG" ] && [ -f "${repo_dir}/scripts/confirm_agent_image.sh" ]; then
+    "${repo_dir}/scripts/confirm_agent_image.sh" "$target_namespace" platform-agent-gateway "$PARAM_IMAGE_TAG"
+  fi
   if [ "$PARAM_UPGRADE_MODE" = "harness" ] || [ "$PARAM_UPGRADE_MODE" = "full" ] || [ "$restarted_agent" = "true" ]; then
-    kubectl rollout status deployment/platform-agent-gateway -n kubeagents-system --timeout=120s
+    kubectl rollout status deployment/platform-agent-gateway -n kubeagents-system --timeout=900s
   fi
```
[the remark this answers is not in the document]
eLeontev: Resolved in `f19774e0`: Scoped `confirm_agent_image.sh` verification to `harness` and `full` upgrade modes (the modes that actually retag the gateway image), and guarded the call with a check for `deployment/platform-agent-gateway` existence.

.github/workflows/autopush-deploy.yml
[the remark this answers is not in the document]
eLeontev: Resolved in `616436ac`: Renamed workflow-level concurrency groups to `autopush-deploy` and `staging-deploy`, decoupling them from job-level concurrency groups so queued runs no longer deadlock.

.agents/rules/github_actions.md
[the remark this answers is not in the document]
eLeontev: Resolved: Updated the pull request description to strictly adhere to `.github/PULL_REQUEST_TEMPLATE.md` with all required sections (Why This Change, What Changed, Context, Live validation, Self-Review, Risk & Rollout) fully populated.

scripts/installer/installer_common.sh
```diff
@@ -669,6 +669,134 @@ if isinstance(value, str) and value:
...
+        return 1
+      fi
+      ;;
+    pending-upgrade|pending-rollback)
+      if type print_warning >/dev/null 2>&1; then
+        print_warning "Helm release '${release_name}' in namespace '${namespace}' is stuck in '${release_status}'. Attempting automatic rollback..."
+      else
+        echo "⚠️ WARNING: Helm release '${release_name}' in namespace '${namespace}' is stuck in '${release_status}'. Attempting automatic rollback..." >&2
+      fi
+
```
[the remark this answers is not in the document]
eLeontev: Resolved in `17fa2e52`: Added in-flight operation age detection with wait-out polling, ensuring active deployments are allowed to complete rather than raced with rollback.

upgrade.sh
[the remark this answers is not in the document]
eLeontev: Resolved in `f19774e0`: Scoped `confirm_agent_image.sh` verification to `harness` and `full` upgrade modes (the modes that actually retag the gateway image), and guarded the call with a check for `deployment/platform-agent-gateway` existence.

.agents/rules/github_actions.md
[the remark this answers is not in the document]
eLeontev: Resolved: Updated the pull request description to strictly adhere to `.github/PULL_REQUEST_TEMPLATE.md` with all required sections (Why This Change, What Changed, Context, Live validation, Self-Review, Risk & Rollout) fully populated.

.github/workflows/staging-deploy.yml
```diff
@@ -0,0 +1,92 @@
...
+      id-token: write
+      actions: read
+    uses: ./.github/workflows/reconcile-environment.yml # zizmor: ignore[secrets-inherit]
+    with:
+      github_environment: "staging"
+      mode: "apply"
+      image_tag: ${{ needs.resolve-commit.outputs.commit_sha }}
+      lease_policy: ${{ needs.resolve-commit.outputs.lease_policy }}
+      force_reconcile: ${{ needs.resolve-commit.outputs.force_reconcile == 'true' }}
+    secrets: inherit
```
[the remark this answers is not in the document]
eLeontev: Resolved in `a4182968`:
1. In `scripts/release/resolve_deploy.sh`, changed the default `lease_policy` from `defer` to `fail` for automated triggers (tag push for staging and image publish for autopush). If a live-test lease is held, the deploy step now fails loudly with `exit 1` instead of silently deferring with exit 0 and dropping the promotion.
2. In `scripts/release/reconcile_environment.sh`, updated `lease_policy == "fail"` to set `output "result" "failed"` before exiting 1.
3. Created reusable verifier `scripts/release/verify_deploy_result.sh` invoked by both `staging-deploy.yml` and `autopush-deploy.yml` in a dedicated `verify-deploy` job, failing loudly if the deploy result is deferred, failed, or missing.
4. Defaulted `workflow_dispatch` input `lease_policy` to `fail` across deploy workflows.
5. Added 11 unit tests in `tests/test_verify_deploy_result.py` and updated wiring tests.

eLeontev:
All review findings addressed in `a4182968`:

- **🔴 High (Staging promotion silently dropped on lease)**: Defaulted automated deploy `lease_policy` to `fail` across `resolve_deploy.sh`, `staging-deploy.yml`, and `autopush-deploy.yml`. Added reusable `scripts/release/verify_deploy_result.sh` in a dedicated `verify-deploy` job to verify deployment outcomes and fail loudly on deferrals or errors.
- **🟡 Low (Magic constants in `installer_common.sh`)**: Hoisted `HELM_OPERATION_TIMEOUT_DEFAULT=600`, `HELM_PENDING_WAIT_MAX_DEFAULT=300`, `HELM_LOCK_POLL_INTERVAL_DEFAULT=10`, and `HELM_ROLLBACK_TIMEOUT_DEFAULT="2m"` as file-scope `readonly` constants per AGENTS.md rules.
- **🟡 Low (Docs drift)**: Cleaned up stale references to deleted reconcile jobs across `nightly-pipeline.yml`, `drift-detect.yml`, `scripts/release/README.md`, and `environment-reconcile.md`.
- Rebased cleanly onto latest `upstream/main`.

scripts/installer/installer_common.sh
```diff
@@ -669,6 +679,235 @@ if isinstance(value, str) and value:
...
+  # in-flight deployment rather than a wedged zombie lock.
+  # Helm's standard release timeout across deploy workflows is 10m (600s).
+  local operation_timeout="${HELM_OPERATION_TIMEOUT:-$HELM_OPERATION_TIMEOUT_DEFAULT}"
+  local pending_age=""
+  if command -v kubectl >/dev/null 2>&1; then
+    local creation_ts
+    creation_ts="$(kubectl get secret -n "${namespace}" -l "owner=helm,name=${release_name},status=${release_status}" --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.creationTimestamp}' 2>/dev/null || true)"
+    if [ -n "${creation_ts}" ]; then
+      local created_epoch
+      created_epoch="$(date -d "${creation_ts}" +%s 2>/dev/null || echo "")"
```
mplakhtiy: **`date -d` is GNU-only, so the in-flight wait silently does not exist on macOS**

```bash
created_epoch="$(date -d "${creation_ts}" +%s 2>/dev/null || echo "")"
```

BSD `date` has no `-d`. Verified on a darwin host: `date -d "2026-09-04T10:00:00Z" +%s` -> `date: illegal option -- d`, rc 1. The `2>/dev/null || echo ""` swallows it, so `created_epoch` is empty, `pending_age` stays empty, `wait_budget` stays `0`, the whole wait block is skipped, and control falls straight into the recovery `case` -- `helm rollback` for `pending-upgrade`/`pending-rollback`, or a hard `return 1` for `pending-install`.

macOS is a supported host for this code path: `.github/workflows/installer-matrix-test.yml:40` runs the installer matrix on `[ubuntu-latest, macos-latest]`, and `upgrade.sh` is a documented front door for operators. That job is green only because it never reaches a pending release.

Failure scenario: an operator on a Mac runs `./upgrade.sh` against `autopush` while an `autopush-deploy` run is mid-`helm upgrade`. On Linux the function would wait it out; on macOS it immediately attempts `helm rollback` against the live operation. Helm v3 will usually refuse ("another operation in progress"), so the realistic outcome is `return 1` and an aborted upgrade rather than destruction -- but the wait added specifically to prevent racing an active deployment is not running at all, and nothing says so.

Also untested. The only test that exercises the wait, `tests/test_installer_common.py:963` (`test_pending_upgrade_in_flight_waits_and_succeeds_when_deployed`), forces `HELM_LOCK_WAIT_TIMEOUT=5`, which takes the `if [ -n "${HELM_LOCK_WAIT_TIMEOUT:-}" ]` branch and bypasses `pending_age` entirely. The `date -d` path has no coverage on any platform.

Suggested fix: parse the RFC3339 timestamp portably -- `python3` is already a dependency of these scripts (`reconcile_environment.sh` shells to it for the lease, and `installer_common.sh` already carries an inline `python3` block). Failing that, at minimum make "the timestamp could not be parsed" distinguishable from "the operation is old", the way `helm_release_status` deliberately carries a `sed` fallback for a missing `jq`.

mplakhtiy (commented):
Findings from a full review of this branch at `4798bcaa`. Two High, eight Medium; Low findings are omitted.

**`environment-reconcile.md` now contradicts itself: the first half still describes the deleted redeploy workflows as the mechanism** -- `docs/site/src/content/docs/deploy/environment-reconcile.md:23-29` and `:45-56`, both outside this diff's hunks.

The PR rewrote the "The nightly reconcile" section into "Deploying and reconciling long-lived environments", which is right. The two passages above it were not touched and are now false:

> The redeploy workflows that move them are `helm upgrade` on a pre-existing release and nothing more, so on their own they carry images and no infrastructure ... A green redeploy says the images rolled, which reads as "main is deployed" and is only half of it.

> ... the redeploy workflows move the running tag with `helm upgrade` and never run Terraform, so planning at the tag the cluster is _serving_ would show every redeploy since the last apply as a pending change to `helm_release.kube_agents`.

After this PR every deploy to a long-lived environment *is* a Terraform apply that records the tag in state, which is the opposite of both. "Three things close that gap" is a framing built on the first passage.

Fix it: this is the canonical page for this path (`scripts/release/README.md:140` says so), and "the page must not say two different things after the merge" is one of `review-docs-drift`'s checks.

scripts/installer/installer_common.sh
```diff
@@ -669,6 +679,235 @@ if isinstance(value, str) and value:
...
+
+  # Calculate wait budget if an in-flight operation might still be running
+  local wait_budget=0
+  if [ -n "${HELM_LOCK_WAIT_TIMEOUT:-}" ]; then
+    wait_budget="${HELM_LOCK_WAIT_TIMEOUT}"
+  elif [ -n "${pending_age}" ] && [ "${pending_age}" -ge 0 ] && [ "${pending_age}" -lt "${operation_timeout}" ]; then
+    wait_budget=$(( operation_timeout - pending_age ))
+    local max_wait="${HELM_PENDING_WAIT_MAX:-$HELM_PENDING_WAIT_MAX_DEFAULT}"
+    if [ "${wait_budget}" -gt "${max_wait}" ]; then
+      wait_budget="${max_wait}"
```
mplakhtiy: **The 300s wait cap is below the 600s operation timeout it is derived from, so a healthy in-flight operation is declared wedged**

`HELM_OPERATION_TIMEOUT_DEFAULT=600` is correct -- it matches `terraform/examples/full-install/main.tf:453`, where `helm_release.kube_agents` sets `timeout = 600`. The arithmetic on line 766 is correct too: an operation that started `pending_age` seconds ago has `600 - pending_age` seconds left before Helm itself gives up.

Then `max_wait` (300) throws half of that away. A release that went `pending-upgrade` 30 seconds ago has 570s legitimately remaining; the cap waits 300. At t~330s the poll loop expires, the code prints "Timed out waiting for in-flight Helm operation", falls through, and runs `helm rollback` against an operation that still has 270 seconds of its sanctioned budget left.

Failure scenario: `terraform apply` inside a scheduled `autopush-deploy` run is 30s into a `helm upgrade` that will take 8 minutes (`wait = true`, and this PR raises the gateway rollout gate to 900s precisely because these rollouts are slow). A second `upgrade.sh` -- a manual reconcile, a `workflow_dispatch` with `lease_policy: ignore`, an operator at a terminal -- enters `ensure_clean_helm_release`, waits the capped 300s, decides the release is wedged, and issues `helm rollback`. Helm v3 usually rejects a rollback during a pending operation, so the common outcome is `return 1` and an aborted second run; but if the first operation completes in the window between the last status poll and the rollback call, the rollback succeeds and reverts an upgrade that had just landed.

This is the same guarantee H1 breaks, by a different route.

Suggested fix: either drop `HELM_PENDING_WAIT_MAX` (the `600 - pending_age` expression is already bounded by the operation timeout) or, if a hard ceiling is wanted for CI wall-clock reasons, treat hitting it as "still in flight, do not recover" -- exit non-zero with that message rather than falling through into the rollback branch.

scripts/release/render_install_env.sh
```diff
@@ -122,6 +124,8 @@ ENABLE_GVISOR
 MEMORY_PROVIDER
 USER_PROFILE_ENABLED
 ENABLE_GKE_BACKUP_PLAN
+ENABLE_PUBSUB_PLATFORM
+ENABLE_STOCKOUT_INVESTIGATOR
```
mplakhtiy, line 128: **`REQUIRED_STRICT` gains two variables that must exist in GitHub environment settings before this merges**

`ENABLE_PUBSUB_PLATFORM` and `ENABLE_STOCKOUT_INVESTIGATOR` are added to `REQUIRED_STRICT`, and `--strict` is what every job touching a long-lived environment runs (`reconcile_environment.sh:75`). The mapping itself is right -- the doc table this PR adds says absent destroys the Pub/Sub topic and the log sink, which is exactly why they belong in the strict list.

But `render_install_env.sh` exits 1 naming the missing variables, and it runs before anything else in the reconcile. So from the first push to `main` after this merges:

- `autopush-deploy.yml` fails on every GHCR publish;
- `staging-deploy.yml` fails on the next promotion tag;
- `drift-detect.yml` goes red daily on both environments -- the doc this PR edits says as much ("Required for a **plan** as much as for an apply ... until an environment carries all twelve the daily drift report goes red on it").

`rc` and `nightly` are unaffected: `deploy-environment.yml:201` renders without `--strict`.

The remedy is settings work rather than code work. A GitHub-environment mismatch fails jobs with zero steps run, so it will not be obvious from the log what happened. Confirm both variables are set on the `autopush` and `staging` environments before merging.

.github/workflows/autopush-deploy.yml
```diff
@@ -0,0 +1,121 @@
...
+        description: "What to do when somebody is live-testing against autopush"
+        required: false
+        default: "fail"
+        type: choice
+        options:
+          - "fail"
+          - "ignore"
+          - "defer"
+      force_reconcile:
+        description: "Force apply even if autopush is already running the candidate commit"
```
mplakhtiy: **`force_reconcile` is wired through three workflows and a script, and nothing reads it; `noop` is a documented result no code produces**

Other sites: `.github/workflows/reconcile-environment.yml:39,85,176`; `.github/workflows/autopush-deploy.yml:63,77,101`; `.github/workflows/staging-deploy.yml:33,55,69,91`; `scripts/release/resolve_deploy.sh:82-99`; `scripts/release/verify_deploy_result.sh:38-40`.

`grep -rn "FORCE_RECONCILE\|force_reconcile"` over every `*.sh`, `*.yml` and `*.py` in the tree returns the three workflows, `resolve_deploy.sh`, and tests -- and nothing else. `reconcile_environment.sh` never reads `FORCE_RECONCILE`; neither does `upgrade.sh`. There is no "is the install already running this commit?" check anywhere for the flag to override.

Three consequences travel together:

1. The `workflow_dispatch` input "Force apply even if autopush is already running the candidate commit" describes behaviour that does not exist. Setting it changes nothing; leaving it false changes nothing either, because the skip it would override was never implemented.
2. `reconcile-environment.yml:46` now documents `result` as "planned, applied, deferred, **noop** or failed". `grep -rn 'output "result" "noop"'` finds no producer in the tree.
3. `verify_deploy_result.sh:38-40` has a `noop)` branch that can never be reached.

No test asserts the flag reaches a consumer -- `tests/test_resolve_deploy_scripts.py` only checks that `resolve_deploy.sh` echoes it back out -- which is why the suite is green around it.

Suggested fix: either implement the skip (and emit `noop`), or delete the input, the `FORCE_RECONCILE` env line, the `noop` output documentation and the `noop` branch. A dispatch control that reads as a safety override and does nothing is worse than its absence.

.github/workflows/autopush-deploy.yml
```diff
@@ -0,0 +1,121 @@
...
+          INPUT_LEASE_POLICY: ${{ inputs.lease_policy }}
+          INPUT_FORCE_RECONCILE: ${{ inputs.force_reconcile }}
+          EVENT_NAME: ${{ github.event_name }}
+          RUN_HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+        run: ./scripts/release/resolve_deploy.sh
+
+      - name: Verify candidate images exist in GHCR
+        env:
+          COMMIT_SHA: ${{ steps.resolve.outputs.commit_sha }}
+          REGISTRY_PREFIX: ${{ vars.REGISTRY_PREFIX }}
```
mplakhtiy, line 79: **The two new resolve jobs read `vars.REGISTRY_PREFIX` without binding an environment, which the sibling nightly workflow has a comment and a test forbidding**

Same shape at `.github/workflows/staging-deploy.yml:75` (`resolve-commit`).

Both jobs run `verify_candidate_images.sh` with `REGISTRY_PREFIX: ${{ vars.REGISTRY_PREFIX }}` and neither declares `environment:`. The `deploy` job they feed resolves the same variable from the bound `autopush` / `staging` environment (`reconcile-environment.yml:92`, `:182`).

`nightly-pipeline.yml:54-64` states the rule and the reason outright, for the identical pairing:

> Bound for GH_ORG, GH_REPO and REGISTRY_PREFIX ... Resolving those from a different scope than the deploy job would verify one registry and install from another. An unbound job resolves every `vars.*` to empty in silence.

and `tests/test_nightly_pipeline_wiring.py:70` pins it (`test_the_resolve_job_binds_the_nightly_environment`). The new workflows opt out of both, and the new tests added for them do not carry the equivalent assertion.

Failure scenario: latent today. That same nightly comment records that `REGISTRY_PREFIX` "is set at no scope", so both jobs currently fall through `get_registry_prefix` -> `DEFAULT_REGISTRY_PREFIX` = `ghcr.io/gke-labs/kube-agents`, which is also what the deploy resolves. The moment either environment sets `REGISTRY_PREFIX` -- an Artifact Registry mirror is the documented use case -- the gate verifies six images in GHCR and the install pulls from the mirror, and the pre-deploy verification this PR exists to add is checking the wrong registry.

Suggested fix: add `environment: autopush` / `environment: staging` to the two resolve jobs, and mirror the nightly's test assertion.

scripts/installer/installer_common.sh
```diff
@@ -669,6 +679,235 @@ if isinstance(value, str) and value:
...
+  if ! image="$(kubectl get deployment platform-agent-gateway -n "${namespace}" \
+    -o jsonpath='{.spec.template.spec.containers[?(@.name=="platform-agent")].image}' 2>/dev/null)"; then
+    return 0
+  fi
+  [ -n "${image}" ] || return 0
+  case "${image##*:}" in
+    */*) return 0 ;;
+    *) printf '%s\n' "${image##*:}" ;;
+  esac
+}
```
mplakhtiy, line 705: **`running_image_tag` is now defined twice, and this sourced copy silently shadows the documented one**

The other definition is `upgrade.sh:140-162`. `upgrade.sh` defines `running_image_tag` at file scope (line 140), then `main()` sources `installer_common.sh` at line 538, then calls `running_image_tag "$target_namespace"` at line 685. Source order decides: this `installer_common.sh` copy wins, and `upgrade.sh`'s is dead.

The dead one is the one carrying the explanation -- a nine-line comment on why the container is selected by name rather than index, ending "a positional read is one reordering away from pinning the composition's image_tag to a sidecar's version -- on a scheduled apply, silently". The live copy has a one-line comment and none of that. `grep` finds no caller of the new copy other than `upgrade.sh:685`, so the duplication buys nothing today.

Failure scenario: someone hardens the container selector or the tag-parsing `case` in the copy they can find (`upgrade.sh`, where the reasoning is), the behaviour does not change, and the bug they were fixing is still live in the shadowing copy.

Suggested fix: keep one. If the helper belongs in `installer_common.sh`, delete `upgrade.sh:140-162` and move its comment across.

tests/test_deploy_image_confirmation.py
```diff
@@ -34,52 +34,28 @@
...
-            "continue-on-error on the confirmation step lets an ignored tag deploy green",
-        )
-        self.assertNotRegex(
-            step["run"],
-            r"(\|\|\s*true|;\s*exit\s+0)\s*$",
-            "the confirmation's exit status must reach the job",
-        )
+        # A guard whose failure is swallowed is worse than no guard.
+        self.assertNotIn(f"{_SCRIPT.name}\" || true", self.text)
+        self.assertNotIn(f"{_SCRIPT.name}\" ; exit 0", self.text)
```
mplakhtiy: **The confirmation guard test no longer enforces the property its name claims**

The old test read the workflow as YAML and checked two structural things: that the step carried no `continue-on-error`, and that its `run` did not end in `|| true` or `; exit 0`, with a docstring explaining that "both routes to that are cheap to add later and invisible in review, so pin them". It is now two literal substring checks against `upgrade.sh`.

Neither literal can ever appear. The actual invocation is

```
"${repo_dir}/scripts/confirm_agent_image.sh" "$target_namespace" platform-agent-gateway "$PARAM_IMAGE_TAG"
```

so the script name is never the token immediately before `|| true`.

Demonstrated: appending `|| true` to that invocation and re-running `tests/test_deploy_image_confirmation.py` gives `24 passed`. The test cannot fail for the change it exists to prevent.

Suggested fix: match on the whole invocation line -- find the line containing `confirm_agent_image.sh` and assert it does not end in `|| true` or `; exit 0`.

scripts/release/reconcile_environment.sh
```diff
@@ -188,11 +121,14 @@ fi
...
-  LEASE_HELD="false"
+  if [ "$LEASE_HELD" = "true" ]; then
+    if ! python3 "${REPO_ROOT}/scripts/live_test_lease.py" release >/dev/null 2>&1; then
+      echo "⚠️ Warning: Failed to release live-test lease." >&2
+    fi
+    LEASE_HELD="false"
+  fi
 }
-trap release_lease EXIT
+trap release_lease EXIT INT TERM HUP
```
mplakhtiy: **`trap release_lease EXIT INT TERM HUP` releases the lease on cancellation and then keeps running**

Adding `INT TERM HUP` to the trap is the right instinct -- an untrapped fatal signal skips the EXIT trap and strands the lease for its 90-minute TTL. But the handler does not exit, and bash resumes execution after a trapped signal rather than terminating.

Demonstrated with a reduced repro of this exact handler shape (`trap f EXIT INT TERM HUP` where `f` does not `exit`): `kill -TERM` printed the release message, execution continued past the signal, and the script exited 0.

Failure scenario: someone cancels an `autopush-deploy` run. The runner sends SIGINT, then SIGTERM. The handler releases the lease, the script carries on into or through the `terraform apply` / `helm upgrade`, and the environment is now being mutated by a run that has told everyone else the lease is free -- including a queued deploy or an agent who was waiting for it. The old `trap release_lease EXIT` did not have this shape.

Suggested fix: `trap 'release_lease; exit 130' INT` and `exit 143` for TERM (or a shared handler that re-raises with the default disposition), keeping the bare EXIT trap for the normal path.

scripts/release/resolve_deploy.sh
```diff
@@ -0,0 +1,100 @@
...
+elif [ -n "${GITHUB_OUTPUT:-}" ]; then
+  echo "commit_sha=${target}" >> "${GITHUB_OUTPUT}"
+fi
+
+lease_policy="${INPUT_LEASE_POLICY:-}"
+if [ -z "${lease_policy}" ]; then
+  # Automated deploy triggers (such as tag pushes for staging or GHCR publishes for autopush)
+  # do not run on a recurring schedule. Silently deferring with exit 0 would drop the release
+  # while falsely reporting success. If a live-test lease is held, refuse to apply and fail loudly.
+  lease_policy="fail"
```
mplakhtiy: **Defaulting `lease_policy` to `fail` is right for staging and wrong for autopush, and the comment justifying it is inaccurate**

True for staging: the `staging_*` tag fires once, `resolve_promotion_candidate.sh` then marks that candidate promoted, and nothing retries.

Not true for autopush: `autopush-deploy.yml` starts on every successful GHCR publish, which is every push to `main`. That *is* a recurring retry, at roughly ten runs a day, and the next one deploys a strictly newer commit -- so nothing is dropped by deferring.

What `fail` buys instead: `reconcile_environment.sh` is the same file whose comment (lines 111-116) says "autopush is that install for most of this repository's agents", and AGENTS.md requires every pull request to be live-tested against a running install. So the expected steady state -- an agent holding the autopush lease -- now turns a push to `main` into a red workflow run that the pusher cannot act on. The `verify-deploy` job already guarantees a deferral is never mistaken for an apply, so the loud-failure property this change was after is available without making the routine case red.

Worth reconsidering -- `defer` for autopush, `fail` for staging -- but this is a judgment call and reasonable people could keep it uniform. At minimum the comment should stop asserting something false about autopush's trigger cadence.

scripts/installer/installer_common.sh
eLeontev: Resolved in commit fdad6072: Added portable date parsing via `parse_rfc3339_epoch` supporting GNU `date -d` and BSD `date -u -j -f`, failing fast with a non-zero exit on unparseable input.

scripts/installer/installer_common.sh
eLeontev: Resolved in commit aabdcea2: Removed the 300s sub-budget so the in-flight wait uses the full operation timeout minus pending age (600s). Added a guard preventing rollback if the wait exits while the release is still within its valid operation window.

scripts/installer/installer_common.sh
eLeontev: Resolved in commit 9168f218: Removed duplicate running_image_tag from upgrade.sh, consolidating on the shared implementation in scripts/installer/installer_common.sh with full docstrings and tests.

.github/workflows/autopush-deploy.yml
eLeontev: Resolved in commit 5b819ba6: Added `environment: autopush` to resolve-candidate in autopush-deploy.yml and `environment: staging` to resolve-commit in staging-deploy.yml to correctly scope vars.REGISTRY_PREFIX.

.github/workflows/autopush-deploy.yml
eLeontev: Resolved in commit 0bba2c64: Fully removed force_reconcile and the unreachable noop status from autopush-deploy.yml, staging-deploy.yml, reconcile-environment.yml, resolve_deploy.sh, verify_deploy_result.sh, release docs, and unit tests.

scripts/release/reconcile_environment.sh
eLeontev: Resolved in commit 62d174a8: Replaced the shared EXIT/signal trap with dedicated INT (exit 130), TERM (exit 143), and HUP (exit 129) traps alongside a separate EXIT trap, terminating the process immediately upon signal cancellation rather than resuming execution.

scripts/release/resolve_deploy.sh
eLeontev: Resolved in commit 4a24f232: Updated the comment in resolve_deploy.sh to accurately contrast the trigger cadence of staging (one-shot promotion tag) and autopush (recurrent GHCR publish on push to main) and document why automated release deploys default to failing loudly rather than silently dropping candidate commits.

tests/test_deploy_image_confirmation.py
eLeontev: Resolved in commit 5c248e97: Updated test_the_confirmation_can_still_fail_the_job in tests/test_deploy_image_confirmation.py to match the whole line invoking confirm_agent_image.sh and assert that no error-masking constructs (|| true, || :, || exit 0, ; exit 0) terminate the command.

scripts/release/render_install_env.sh
eLeontev: Resolved: Configured `ENABLE_PUBSUB_PLATFORM=true` and `ENABLE_STOCKOUT_INVESTIGATOR=true` in GitHub repository environment settings for both `autopush` and `staging`.

eLeontev:
All 10 findings from @mplakhtiy's review have been addressed, verified with tests, and pushed:

1. **Portable date parsing** (`scripts/installer/installer_common.sh`): Implemented portable RFC3339 timestamp parsing via `parse_rfc3339_epoch` supporting GNU `date -d` and BSD `date -u -j -f`, failing fast with non-zero exit on unparseable inputs (`fdad6072`).
2. **In-flight release wait budget & rollback guard** (`scripts/installer/installer_common.sh`): Removed arbitrary 300s sub-budget so the loop utilizes the full operation timeout minus pending age (600s). Added guard strictly blocking rollback while the release is within its valid operation window (`aabdcea2`).
3. **Consolidate `running_image_tag`** (`upgrade.sh`, `scripts/installer/installer_common.sh`): Removed duplicate implementation from `upgrade.sh` and consolidated on the shared helper in `installer_common.sh` with full docstrings and tests (`9168f218`).
4. **Target environment scoping for `REGISTRY_PREFIX`** (`.github/workflows/autopush-deploy.yml`, `.github/workflows/staging-deploy.yml`): Added `environment: autopush` to `resolve-candidate` and `environment: staging` to `resolve-commit` to prevent drift between resolve and deploy steps (`5b819ba6`).
5. **Remove `force_reconcile` and `noop`** (`.github/workflows/*`, `scripts/release/*`): Fully deleted unused `force_reconcile` inputs/outputs and unreachable `noop` deploy result branch across workflows, scripts, docs, and tests (`0bba2c64`).
6. **Signal trap handling in `reconcile_environment.sh`** (`scripts/release/reconcile_environment.sh`): Replaced shared signal trap with dedicated `INT` (exit 130), `TERM` (exit 143), and `HUP` (exit 129) traps alongside `EXIT`, terminating immediately upon workflow cancellation (`62d174a8`).
7. **Accurate `lease_policy` default rationale** (`scripts/release/resolve_deploy.sh`): Updated comment to contrast staging and autopush trigger models and document why automated release deploys default to failing loudly upon contention (`4a24f232`).
8. **Confirmation error-masking guard test** (`tests/test_deploy_image_confirmation.py`): Updated `test_the_confirmation_can_still_fail_the_job` to match the whole line invoking `confirm_agent_image.sh` and assert against error-masking constructs (`5c248e97`).
9. **Environment settings for `REQUIRED_STRICT`**: Configured `ENABLE_PUBSUB_PLATFORM=true` and `ENABLE_STOCKOUT_INVESTIGATOR=true` in GitHub environment settings for both `autopush` and `staging`.
10. **Documentation drift** (`docs/site/src/content/docs/deploy/environment-reconcile.md`): Replaced obsolete descriptions referencing deleted standalone redeploy workflows with accurate documentation of unified atomic reconciliations and Terraform state tag isolation (`0b79c852`).

All 184 unit tests and `make docs-check` pass cleanly.

mplakhtiy (approved):

jayantid:
Red-sweep classification: stale false red — run sat in the documented 2026-09-04 environmental window (#1249: seeded-fleet-reader impersonation PERMISSION_DENIED + Boskos 401s + infra-lost reps); retests after the window went green elsewhere. Not this diff. Retesting.

/retest

PR #1200, opened 2026-09-03
fix(platform): harden repository validation, token minter policy, and state reconciliation

### LINKED ISSUE #1113: Follow-up from #504: multi-repo support — carried-over and new findings

kirill-kosov:
## Summary

This pull request consolidates follow-up fixes for multi-repository configuration across the Go operator, minty token minter policies, Python agent runtime scripts, and documentation. It ensures pre-rendered minter policy keys are properly adopted, hardens repository slug and scheme validation, prevents operator reconciliation loops from wedging on malformed state, and updates documentation and test fixtures to reflect `gitops-state` ConfigMap seeding.

## Why This Change

Without these fixes, registering multiple repositories or upgrading minty token minter rules could fail due to missing initial `<repo>.yaml` policy keys at install time. In addition, malformed `managed_repos` state could cause operator reconciliation loops to error out repeatedly, and unanchored or case-variant repository strings could bypass validation or fail allowlist checks. Updating these paths guarantees robust state reconciliation and consistent multi-repo handling across all platform components.

## What Changed

- **Token Minter Policy & Helm**: Pre-rendered `<repo>.yaml` policy keys alongside `default.yaml` in Helm chart and Kustomize templates for `github-token-minter-config`, added defensive `$repo` sanitization in the Helm template, and updated operator adoption logic to track pre-rendered keys.
- **Operator Validation & Robustness**: Updated `reconcileGitopsStateConfigMap` and `syncGithubTokenMinterConfigMap` to log warning messages and degrade gracefully without wedging the reconciler on invalid JSON. Host-anchored SCP and URL remotes in `CleanRepoSlugWithOrg`, stripped schemes case-insensitively (`HTTPS://`), and validated slug components against `^[a-zA-Z0-9_.-]+$`.
- **Agent Runtime & Gates**: Surfaced `unreachable_repos` warnings in `github_scan_gate.py`, dynamically scaled `RESOLVER_TIMEOUT_S` based on managed repository count, logged warnings on `JSONDecodeError` while preserving `[]` fallback contracts in `gitops_workspace.py`, aligned empty allowlist behavior across skills, and deduplicated `over_budget` log emissions.
- **Refactoring & Documentation**: Extracted `validate_repo_org` into `gitops_workspace.py` for DRY primary organization validation, cleaned up stale `forge.py` docstrings, and updated `ci-pool-projects.md`, `platformagent-crd.md`, `conftest.py`, and `ci-deploy.sh` to reference `gitops-state` `managed_repos` instead of stale `SETTINGS.md` entries.

## Context

Closes #1113

## Testing

- **Go Operator Unit Tests**: Ran `go test ./...` in `k8s-operator/` — 100% PASS.
- **Python Unit Tests**: Ran `python3 -m unittest` across platform scripts and skills (`test_github_scan_gate.py`, `test_gitops_workspace.py`, `test_submit_suggestion.py`, `test_forge.py`, `test_pr_triggers.py`, `test_minter_repo_wiring.py`) — 362/362 PASS (0 failures).
- **Documentation Checks**: Executed `make docs-check` — 100% PASS (283 files checked, 0 broken links, terminology and context budget verified).

### Live validation

Not live-tested against a running cluster in this local environment; verified via unit tests and test suites covering operator reconcilers, template rendering, and python script gates.

## Self-Review

Passes executed in clean test subagents:
1. **Adversarial Pass**: Verified scheme stripping against mixed case (`HTTPS://`, `Git@`), host anchoring against non-GitHub SCP remotes (`git@gitlab.com:...`), and ConfigMap key slug character validation against path traversal (`..`) and special characters. Confirmed graceful degradation on malformed JSON without operator panic.
2. **Docs-Drift Pass**: Verified updated CRD documentation, deployment guides (`ci-pool-projects.md`), script docstrings (`forge.py`), and test fixtures (`conftest.py`, `ci-deploy.sh`) against codebase references. Confirmed `make docs-check` passes with zero warnings or errors.

## Risk & Rollout

Low risk. Changes preserve existing API contracts and allowlist defaults while adding defensive validation, graceful degradation, and ConfigMap key adoption logic.


k8s-operator/api/v1alpha1/common_types.go
```diff
@@ -1225,10 +1240,20 @@ func CleanRepoSlugWithOrg(rawURL, org string) (string, error) {
...
+	// Basic verification of owner/repo structure and component character set
+	parts := strings.Split(cleaned, "/")
+	if len(parts) != 2 {
 		return "", fmt.Errorf("invalid repository format")
 	}
+
+	owner, repo := parts[0], parts[1]
+	if owner == "" || repo == "" || repo == "." || repo == ".." {
+		return "", fmt.Errorf("invalid repository slug %q", cleaned)
+	}
```
mplakhtiy, line 1334: **HIGH** — the new slug component check guards the repo half only, so the Go validator is still weaker than the Python one it mirrors.

`validRepoSlugPart` is `^[a-zA-Z0-9_.-]+$`, which admits `.`, `..`, and a leading `-`. The explicit guard that rejects `.`/`..` is applied to `repo` and not to `owner`, and the leading-dash case is rejected for neither.

Re-implementing this function's exact branch sequence in Python and running 21 inputs through it, four are accepted that the Python counterpart rejects:

| input | Go result | `gitops_workspace.is_valid_repo_slug` |
| --- | --- | --- |
| `../repo` | accepted, slug `../repo` -> `https://github.com/../repo` | rejected |
| `./repo` | accepted, slug `./repo` | rejected |
| `-x/repo` | accepted | rejected |
| `acme/-x` | accepted | rejected |

`gitops_workspace.py:631` states the reasoning and applies it to both components:

```python
def _valid_repo_component(part: str) -> bool:
    """Reject path components unsafe to hand to `gh -R` or use as a filesystem path.

    The slug pattern permits "." and "-", so it happily produces "../..", and a
    leading dash is parsed by `gh` as a flag. Neither is a shape the regex can
    express.
    """
    return bool(part) and part not in (".", "..") and not part.startswith("-")
```

The cases this branch's own tests pin — `gke-labs/..`, `gke-labs/re'po`, `gke-labs/repo?tab=readme` — all sit on the repo half, which is why the gap does not show.

**Failure scenario:** `spec.integration.github.gitRepo: ../repo` passes admission and lands in `managed_repos` as `https://github.com/../repo`; the agent then hands that owner to clone and `gh` calls. `acme/-x` reaches `gh -R` with a component the Python layer exists to keep out. The agent-side consumers re-validate with `is_valid_repo_slug`, so neither is a live escalation on its own — but this is the producer, and a producer weaker than its consumers is the shape the validation is meant to close.

Apply the `.`/`..`/leading-dash rejection to both components:

```suggestion
	owner, repo := parts[0], parts[1]
	for _, part := range []string{owner, repo} {
		if part == "" || part == "." || part == ".." || strings.HasPrefix(part, "-") {
			return "", fmt.Errorf("invalid repository slug %q", cleaned)
		}
	}
```

k8s-operator/api/v1alpha1/common_types.go
```diff
@@ -1238,11 +1263,6 @@ func CleanRepoURLWithOrg(rawURL, org string) (string, error) {
 	if trimmed == "" || trimmed == "None" {
 		return "", fmt.Errorf("empty repository")
 	}
-	if strings.HasPrefix(trimmed, "https://") || strings.HasPrefix(trimmed, "http://") {
-		u := strings.TrimSuffix(trimmed, ".git")
-		u = strings.TrimSuffix(u, "/")
-		return u, nil
-	}
 	cleanedSlug, err := CleanRepoSlugWithOrg(rawURL, org)
```
mplakhtiy, line 1349: **HIGH** — deleting the `https://`/`http://` passthrough makes an existing CR with a non-GitHub `gitRepo` un-updatable.

`CleanRepoURLWithOrg` used to return an `https://`/`http://` URL verbatim. With that branch gone it falls through to `CleanRepoSlugWithOrg`, which now rejects any host that is not `github.com`/`www.github.com`. The branch's own test change is the proof:

```diff
-{"https://gitlab.com/gke-labs/kube-agents.git", "", "https://gitlab.com/gke-labs/kube-agents", false},
+{"https://gitlab.com/gke-labs/kube-agents.git", "", "", true},
```

This function is reached from the validating webhook (`platformagent_webhook.go:261`, via `ValidateGitRepoURLWithOrg`), which the CRD reference documents as `failurePolicy: Fail`.

**Failure scenario:** a PlatformAgent installed before this change carries a non-GitHub `gitRepo` — accepted at the time. After the operator upgrade, any update to that CR is rejected at admission until the field is corrected, including updates that have nothing to do with `gitRepo`.

The rejection itself is correct and intended. What is missing is anywhere that says so: no upgrade note in the CRD reference or release notes, and no operator-side signal telling an administrator which CRs are now un-updatable.

k8s-operator/internal/controller/platformagent_controller.go
```diff
@@ -852,7 +855,8 @@ func (r *PlatformAgentReconciler) syncGithubTokenMinterConfigMap(ctx context.Con
 
 	repos, err := parseManagedRepos(managedReposStr)
 	if err != nil {
-		return fmt.Errorf("failed to parse managed_repos for minter policy sync: %w", err)
+		logger.Info("skipping minter policy sync due to unparseable managed_repos in ConfigMap", "error", err)
```
mplakhtiy: **MEDIUM** — a corrupt `managed_repos` is now silent end to end.

Removing the wedge is right and the fix is correct: `reconcileGitopsStateConfigMap` no longer returns an error (see also `:714` and `:719`), so `Reconcile` proceeds to the agent Deployment. What it does not add is any signal a user can see.

The operator emits `logger.Info(...)` — not `Error`, not an Event — and sets no condition. On the agent side, `_parse_managed_repos_json` still returns `[]`, so `resolver.handle_poll` prints `NOT_CONFIGURED`, which `github-issue-resolver/SKILL.md` tells the agent is "a supported state, not a fault". The CR keeps reporting `Ready`.

**Failure scenario:** an administrator hand-edits `managed_repos` — the documented unregistration path — and leaves a trailing comma. Before this change, GitHub work stopped and the operator log screamed. After it, GitHub work stops and one Info line per reconcile is the only trace.

Suggested fix: `logger.Error` plus a `Degraded` condition (or an Event) on the CR. The `InvalidGitRepoURL` condition machinery at `platformagent_controller.go:1867,1944` is the pattern to copy.

k8s-operator/internal/controller/platformagent_controller_test.go
```diff
@@ -4539,6 +4539,61 @@ func TestSyncGithubTokenMinterConfigMap(t *testing.T) {
 	}
 }
 
+func TestSyncGithubTokenMinterConfigMap_AdoptsPreRenderedKeys(t *testing.T) {
```
mplakhtiy, line 4773: **MEDIUM** — the chart-key/mount-path coupling this PR introduces is not pinned by any test.

`TestSyncGithubTokenMinterConfigMap_AdoptsPreRenderedKeys` hand-writes a ConfigMap with a `repo-1.yaml` key and asserts the operator adopts it. That tests the operator, not the coupling. Nothing renders `charts/kube-agents/templates/github-minter.yaml` or `k8s-operator/config/integrations/github/configmap.yaml.template` and asserts the key equals `<bareRepo>.yaml` under `/etc/minty/<org>`. Grepping `tests/`: the only file mentioning the minter is `tests/e2e/test_agent_fleet_audit.py`; `tests/test_minter_repo_wiring.py` is about workflow variables, not the ConfigMap.

**Failure scenario:** someone edits `$repo` in the chart (this PR just did) or `${GITHUB_REPO}` in the kustomize template, the key stops matching `<configDir>/<org>/<repo>.yaml`, and the only symptom is a Minty 403 during the window before the operator's first reconcile — which no test tier covers.

Suggested fix: a `helm template` assertion in `tests/` that the ConfigMap carries a key equal to `<githubMinter.repo>.yaml`, and that the Deployment mounts at `/etc/minty/<githubMinter.org>`.

k8s-operator/internal/controller/platformagent_controller.go
```diff
@@ -890,13 +894,10 @@ func (r *PlatformAgentReconciler) syncGithubTokenMinterConfigMap(ctx context.Con
 	for key := range activeKeys {
 		currentVal, exists := minterCM.Data[key]
 		_, managed := operatorManagedKeys[key]
-		if !exists {
+		if !exists || !managed || currentVal != expectedContent {
```
mplakhtiy, line 1065: **MEDIUM** — adoption is wider than "adopt the chart's key": any unmanaged `<repo>.yaml` is now overwritten, and later deleted.

`!managed` makes the operator claim and overwrite **any** pre-existing key matching an active repo, whatever wrote it — then the prune loop deletes it when that repo is unregistered. The comment 60 lines above states the invariant that just went away: "An empty managed_repos with no previously operator-managed keys is a no-op to avoid wiping unmanaged keys."

**Failure scenario:** an operator hand-narrows `acme-repo.yaml` in `github-token-minter-config` to `contents: read`. The next reconcile replaces it with the `default.yaml` render (`contents`/`pull_requests`/`issues: write`) — a permission widening the operator performs silently — and unregistering that repo later deletes the key rather than restoring the hand-written one.

Adopting the chart's pre-rendered key is the intent and is correct. Claiming everything is broader than that. Worth a decision, not necessarily a change: either restrict adoption to keys whose content matches what the chart renders, or say in the comment that the operator owns every `<repo>.yaml` key for a managed repository and hand-editing them is unsupported.

agents/platform/scripts/github_scan_gate.py
```diff
@@ -226,11 +227,17 @@ def run_resolver_poll() -> dict:
     if not script.is_file():
         raise FileNotFoundError(f"resolver not found at {script}")
 
+    try:
+        managed_count = len(gitops_workspace.get_managed_github_repos())
+    except Exception:
+        managed_count = 1
+    timeout = max(RESOLVER_TIMEOUT_S, managed_count * 180)
```
mplakhtiy: **MEDIUM** — `max(300, n*180)` gives two repositories less than the reasoning behind 300 says one needs, times two.

`timeout = max(RESOLVER_TIMEOUT_S, managed_count * 180)` yields 300s for one repo, 360s for two, 540s for three. The comment above `RESOLVER_TIMEOUT_S = 300` reasons that budget for a *single* repository: "`resolver.py poll` sweeps stale issues before it queries, so it is not a read-only call and its runtime is not bounded by a single request". The per-repo work is `sweep_stale_issues` plus a 100-issue `gh issue list`. If 300 was right for one, n repos need roughly `n * 300`; `max` is the wrong operator and 180 is below the per-repo cost the base implies. From n=2 upward the budget is under-provisioned, and the failure mode is the poll being killed and the watcher silently not filing cards.

Separately, the comment at `github_scan_gate.py:96-98` still reasons the value for one repository and no longer describes what the code does.

Suggested fix: `RESOLVER_TIMEOUT_S + (n-1) * PER_REPO_TIMEOUT_S`, or `n * RESOLVER_TIMEOUT_S`, with the comment updated to say why — and `180` as a named module-level constant either way.

docs/site/src/content/docs/operator/platformagent-crd.md
```diff
@@ -470,7 +470,7 @@ Enables external integrations. Only the enabled ones need to be present.
...
+- **`github`** — `org` (optional target GitHub organization/user, up to 39 characters) and `gitRepo` (optional target GitOps repository URL or `owner/repo` shorthand, up to 2048 characters). Supports HTTPS/HTTP (`https://`, `http://`), SCP-style SSH (`git@github.com:owner/repo`), SSH/Git protocols (`ssh://`, `git://`), and bare `owner/repo` shorthand. Rejects non-GitHub hosts and URLs containing whitespace or invalid syntax. On reconcile, `gitRepo` is appended to the `gitops-state` ConfigMap (`managed_repos`) if absent; removing a repository configured via `gitRepo` requires clearing or updating `spec.integration.github.gitRepo` on the CR in addition to editing the ConfigMap. Populated by the installer when a GitOps repository is connected.
```
mplakhtiy: **MEDIUM** — the rewrite dropped the accurate half of the sentence along with the stale half.

The old sentence was half wrong (`SETTINGS.md`) and half right. The rewrite dropped both halves plus the `at admission (failurePolicy: Fail)` clause. Both dropped facts are still true in the tree:

- `k8s-operator/internal/controller/platformagent_controller.go:1867` sets `condReason = "InvalidGitRepoURL"`, and `:1944` writes the `Degraded` condition.
- `k8s-operator/api/v1alpha1/platformagent_webhook.go:261` calls `ValidateGitRepoURLWithOrg` at admission.

Documenting `org` and the `managed_repos` seeding is the right addition; deleting accurate reference material alongside it is a docs regression on the page a reader actually consults for this field.

Suggested fix: restore the `Degraded` / `InvalidGitRepoURL` sentence and the admission clause alongside the new prose.

mplakhtiy (commented):
Seven findings from a full review of the branch diff: two HIGH (repository slug validation, non-GitHub host rejection) and five MEDIUM. All are anchored inline.

docs/site/src/content/docs/operator/platformagent-crd.md
kirill-kosov: Fixed

agents/platform/scripts/github_scan_gate.py
kirill-kosov: Fixed

k8s-operator/internal/controller/platformagent_controller.go
kirill-kosov: Fixed

k8s-operator/internal/controller/platformagent_controller_test.go
kirill-kosov: Fixed

k8s-operator/internal/controller/platformagent_controller.go
kirill-kosov: Fixed

k8s-operator/api/v1alpha1/common_types.go
kirill-kosov: Fixed

k8s-operator/api/v1alpha1/common_types.go
kirill-kosov: Fixed

mplakhtiy (approved):

jayantid:
Red-sweep classification: stale false red — run sat in the documented 2026-09-04 environmental window (#1249: seeded-fleet-reader impersonation PERMISSION_DENIED + Boskos 401s + infra-lost reps); retests after the window went green elsewhere. Not this diff. Retesting.

/retest

PR #1204, opened 2026-09-03
ci: pin the presubmit to kube-agents-evals-16 (DO NOT MERGE)

### LINKED ISSUE #1051: seeded-fleet-reader is granted to nobody: every eval run falls back to the Prow runner's write credential
## Summary

`seeded-fleet-reader@<project>.iam.gserviceaccount.com` is provisioned on every pool project and impersonated on none of them. Every eval run attempts the impersonation, is denied, warns, and falls back to the Prow runner's own read-write credential. This is a standing condition on all ten pool projects, not a per-project misconfiguration.

It is the residual of A5, and it is the one thing standing between #898's third done-when — "The credential handed to a run is read-only" — and being true. That lane closed as completed on 2026-08-28 with this unmet, so nothing now tracks it.

## Why it happens on every project

Three facts compose:

1. `bench/tf/fleet/variables.tf:33` — `fleet_reader_token_creators` defaults to `[]`, and `bench/tf/fleet/main.tf:138-139` does `for_each = toset(var.fleet_reader_token_creators)`. An empty list creates zero `roles/iam.serviceAccountTokenCreator` bindings.
2. `scripts/provision_ci_pool_project.sh:436` — the fleet apply passes only `-var="project_id=${PROJECT_ID}"`. No tfvars file exists for the stack, and nothing in the repository assigns the variable a value.
3. `hack/ci-eval-pr.sh:260` — `FLEET_READONLY_SA` is always exported, defaulted from the leased `PROJECT_ID`.

Always attempted, never granted. `hack/fleet-kubeconfigs.sh:399-400` then warns once per cluster:

```
WARNING: <cluster> kubeconfig keeps the runner's own credential;
         FLEET_READONLY_SA=seeded-fleet-reader@<project>.iam.gserviceaccount.com
         could not be used
```

Measured across the 30 most recent `pull-kube-agents-smoke-test` builds (2026-08-28, 14:05Z–17:48Z). Thirteen of them reached `hack/fleet-kubeconfigs.sh`; **all thirteen warned, three times each — once per seeded cluster. None reached the step and skipped the warning.** The other seventeen died earlier, so they are not evidence either way — which is why the denominator is runs
that exported `FLEET_READONLY_SA` at all, not runs that finished without a warning.

Six distinct projects are covered: `kube-agents-evals`, `-2`, `-3`, `-4`, `-5` and `-6`.

`-7` through `-10` are not in that sample — the two runs that leased one died before the step. For `-7` the evidence is the pinned smoke test on #995, three warnings. `-8`, `-9` and `-10` were provisioned by the same script on 2026-08-27, with the same empty variable, so they behave identically by construction; that half is an inference, not a measurement.

## How it got here

No single change is at fault. An assumption held across three files and expired without anyone noticing.

- **2026-08-25, #938 (`f3565976`).** Created `seeded-fleet-reader` and the empty-default variable, and recorded the gap as open in `hack/ci-eval-pr.sh`'s own comment: "What is still missing is `FLEET_READONLY_SA`. The run holds a cluster-admin credential on a fleet every open PR shares until it is."
- **2026-08-26, #979 (`e38aefff`).** The provisioning script left the variable at its default, with a comment giving the reason: `hack/fleet-kubeconfigs.sh` only uses the reader account when `FLEET_READONLY_SA` is exported, so "until then the binding grants nothing that gets used." True on the day it was written.
- **2026-08-27, #956 (`22e4e980`).** One line made the export unconditional: `export FLEET_READONLY_SA="${FLEET_READONLY_SA:-seeded-fleet-reader@${PROJECT_ID}...}"`. This is the moment the dormant grant became an attempted-and-denied impersonation on every run, and the moment #979's comment stopped being true — without #979 changing.

So the fix is not a forgotten step to find in the provisioning script. It is a grant that was deliberately deferred and then needed.

## What this is and is not

**Not a functional block.** The kubeconfigs are still written, still point at the right seeded cluster, and every check still grades the right object. Five task files say so in their headers (`cluster-agent-crashloop-evidence-chain`, `-fix-request`, `-misleading-symptom`, `cluster-agent-healthy-workload-no-finding`, `cluster-agent-pending-replicas-capped-pool`).

**A privilege gap on shared state.** The fleet is shared by every open pull request. A run that holds a write-capable credential against it can spoil a fixture for another run, non-deterministically, minutes later. `roles/container.viewer` — what the reader account holds and nothing else — was checked live and grants no `container.secrets.*` and no create/update/delete/patch.

## Closing it

The variable's own comment says it is "empty by default so a project can adopt the stack before its runner identity is known." That rationale holds for an outside adopter and does not hold for this pool: the identity is a constant, not per-project, and it is already in the provisioning script as `PROW_RUNNER_SA` at `scripts/provision_ci_pool_project.sh:266` — `serviceAccount:prowjob-default-sa@kube-agents-prow.iam.gserviceaccount.com`, in the prefixed form the variable's validation requires, defined 170 lines above the fleet apply. Nothing needs discovering. Three changes:

**1. Pass it in the fleet apply** (`scripts/provision_ci_pool_project.sh:436`), and delete the comment above it, which is the stale claim the Done-when below already lists:

```bash
    tofu apply -auto-approve \
      -var="project_id=${PROJECT_ID}" \
      -var="fleet_reader_token_creators=[\"${PROW_RUNNER_SA}\"]"
```

Every project provisioned from here on is correct. Do **not** instead set the default in `bench/tf/fleet/variables.tf` — that would grant this pool's Prow account on any project that adopts the stack, which is what the empty default exists to prevent.

**2. Backfill the ten that already exist.** A script change reaches nothing already applied: those projects hold zero token-creator bindings today and will keep holding zero. One apply each, from `bench/tf/fleet` with that project's state backend:

```bash
tofu apply -var="project_id=<project>" \
  -var="fleet_reader_token_creators=[\"serviceAccount:prowjob-default-sa@kube-agents-prow.iam.gserviceaccount.com\"]"
```

No `-target` needed and no replant risk. `min_master_version` is not `ForceNew` (`bench/tf/fleet/main.tf:156-166`), so a version reconcile upgrades `seeded-b`'s master in place and leaves `createTime` alone — the A3 date gate does not move. This stack is re-applied on a schedule anyway (`main.tf:18-21`), so a re-apply is its normal operating mode.

**3. Pin it in the verifier.** `scripts/verify_ci_pool_project.py` has no check for this binding, so a project can pass all eight checks with the gap wide open — which is how ten of them did. Read `seeded-fleet-reader`'s service-account-level IAM policy and assert `PROW_RUNNER_MEMBER` holds `roles/iam.serviceAccountTokenCreator`. Without it this closes now and silently reopens on the
next project provisioned with `--skip-fleet`.

1 and 3 are ordinary pull-request work. 2 is ten mutating applies against live projects, so it needs a human, per #903. If 3 lands before 2, all ten projects report the new check as failing — harmless, since the verifier gates nothing at runtime, but expect it.

## Done when

- `gcloud auth print-access-token --impersonate-service-account=seeded-fleet-reader@<project>.iam.gserviceaccount.com`
  succeeds on all ten pool projects.
- A full eval run emits no `kubeconfig keeps the runner's own credential` warning.
- The five task-file header comments and the A5 residual note at `bench/tasks/DRAFTS.md:99` are updated or removed.
- `scripts/provision_ci_pool_project.sh:431-435` — the comment there says the binding "grants nothing that gets used" until `FLEET_READONLY_SA` is exported. That was true when written; `hack/ci-eval-pr.sh:260` now always exports it, so the comment is stale on all ten projects.

## References

- `bench/tasks/DRAFTS.md:99` — the A5 residual, described as "the half a human still has to do"
- `hack/ci-eval-pr.sh:246-260` — the gap written out in full, in a comment
- `bench/tf/fleet/README.md`, "A read-only credential for evaluations"
- #898 — closed completed 2026-08-28; its third done-when is this
- #901 — the 2026-08-25 comment stating the same gap, which is where it has been recorded until now
- #1005 — the other open A5 residual. Different problem: that one is the verifier sweep §A5 asks
  contributors to run and they cannot. This one is a binding nobody applied.
- #995 — merged as `473f67f7`; it provisioned `-8`/`-9`/`-10`, which is how the pool-wide scope
  was confirmed. All four verify clean (exit 0) as of 2026-08-28.

Owning lane: #903.

---

lapis2002:
## Summary

**DO NOT MERGE.** Pins this pull request's presubmit to `kube-agents-evals-16` so one real run can prove #1051's fix end to end. The binding that closes #1051 has been applied to `-16` by hand; a run against it must not emit the `kubeconfig keeps the runner's own credential` warning that every other pool project's runs emit three times. The absence of that warning is the result this PR exists to produce.

Revert the pin before this branch goes anywhere near merge. Precedent: `442cbf53` on #995, reverted by `8e42c3b4`.

## Why This Change

`seeded-fleet-reader@<project>` is provisioned on every pool project and impersonated on none, because `scripts/provision_ci_pool_project.sh` applies the fleet stack with `fleet_reader_token_creators` at its empty default. Every eval run attempts the impersonation, is denied, warns, and falls back to the Prow runner's own read-write credential against a fleet every open pull request shares. #1051 has the full mechanism.

Freshly measured on the 30 most recent `pull-kube-agents-smoke-test` builds (latest 2026-09-03 11:21Z): six reached `hack/fleet-kubeconfigs.sh`, and **all six warned, three times each** — once per seeded cluster. No exceptions. Projects `evals-2`, `-5`, `-6`, `-12`, `-13`, `-15`. Three of those did not exist when #1051 was filed, which is what rules out "a few old projects missed a step" and points at the provisioning script itself.

The fix is a Terraform variable and a backfill. Before backfilling fifteen live projects that every open pull request grades against, one project should demonstrate it works.

## What Changed

`hack/ci-env.sh` only: an unconditional `export PROJECT_ID="kube-agents-evals-16"`.

Unconditional matters. Prow exports `PROJECT_ID` from the Boskos lease *before* sourcing this file, so the `${PROJECT_ID:-...}` default above it cannot override a lease, and a pin written that way is silently ignored. Boskos still leases a project and leaves it idle; deploy, eval and teardown all follow `PROJECT_ID`, so `-16` is what gets used and what gets cleaned up, and the idle lease is released as normal.

`kube-agents-evals-16` rather than a live pool project because it is not registered in Boskos yet, so no other pull request can lease it and a bad run breaks nothing. It is fully provisioned — host cluster, all three seeded clusters, the reader service account, and remote state at the `seeded-fleet` prefix — so it is a representative target rather than a special case. As a side effect the run gives `-16` its first real exercise before Boskos registration, which is what #995 did for `-7`.

## Context

Trial run for #1051, whose owning lane is #903. The mapping for `evals-16` comes from #1194, merged as `65fafd5f`, which this branch is based on.

Does not close anything. The pull request that closes #1051 passes the variable in `scripts/provision_ci_pool_project.sh` and pins the binding in `scripts/verify_ci_pool_project.py`; it is written once this run comes back clean.

## Testing

- `bash -n hack/ci-env.sh` — clean.
- The pin overrides a lease: `PROJECT_ID=kube-agents-evals-7 source hack/ci-env.sh` yields `PROJECT_ID=kube-agents-evals-16` and `GCP_PROJECT_ID=kube-agents-evals-16`. This is the property the whole change depends on, and the one a `:-` default would break.
- `gitops_repo_for_project kube-agents-evals-16` resolves to `gke-agentic/kube-agents-evals-16-infra` from `main`, and an unmapped project still exits 1.
- No test asserts the pin, deliberately — a test pinning the presubmit to one project is the thing that must not survive to `main`.

### Live validation

This pull request **is** the live validation, for #1051 rather than for itself. Preconditions established before it was opened, against `kube-agents-evals-16`:

- `gcloud container clusters list` — `platform-agent-host`, `seeded-a`, `seeded-b`, `seeded-c` all RUNNING, the three seeded ones labelled `managed-by=kube-agents-seeded-fleet`.
- `seeded-fleet-reader@kube-agents-evals-16.iam.gserviceaccount.com` exists.
- Its IAM policy held **no bindings at all** before the fix — the gap, confirmed on this specific project rather than inferred.
- `roles/iam.serviceAccountTokenCreator` for `serviceAccount:prowjob-default-sa@kube-agents-prow.iam.gserviceaccount.com` then applied via `tofu apply` on `bench/tf/fleet` with that project's state backend.

**Pass condition:** the run reaches `hack/fleet-kubeconfigs.sh` and writes seven fixture roles with no `keeps the runner's own credential` warning. **Fail condition:** the warning appears, which would mean the binding is not sufficient and #1051's proposed fix is incomplete.

Result to be recorded here once the run completes.

## Self-Review

Reviewed for the two ways a pin goes wrong.

- **Can it merge by accident?** The title and the in-file banner both say DO NOT MERGE, but neither blocks Tide. Mitigation is that it needs `lgtm` plus an `OWNERS` approval, and the diff is one obvious line. Holding it with `/hold` as well is worth doing if it sits open.
- **Can it affect another pull request?** No. The pin is branch-local, `-16` is unregistered in Boskos, and the leased project is left idle rather than deployed to — the same behaviour `442cbf53` was observed to produce.
- **Would it survive a rebase unnoticed?** It is the only change on the branch, so a rebase cannot bury it under unrelated commits.
- **Does it need the docs passes?** No prose or generated region is touched; `make docs-check` has nothing to say about a one-line export. The adversarial pass was not run in a separate context for this one, on the grounds that the diff is a single unconditional assignment whose only failure mode is the one tested above. Flagging that rather than claiming a pass that did not happen.

Two things not covered: the run may fail on ledger scenarios if `BENCH_GITHUB_TOKEN` does not yet cover `gke-agentic/kube-agents-evals-16-infra`, which is unrelated to the fleet credential and lands after the step under test; and `shellcheck` is not installed locally, so only `bash -n` ran.

## Risk & Rollout

Confined to one branch and one unregistered project. Nothing on `main` changes, no other pull request's lease behaviour changes, and the revert is deleting the branch.

The standing risk is the pin merging, which would send every presubmit to `-16` and serialise the pool behind it. That is what the title, the banner comment, and this section exist to prevent.

---

### DIFF
--- a/hack/ci-env.sh
+++ b/hack/ci-env.sh
@@ -14,6 +14,30 @@ source "$(cd "$(dirname "${BASH_SOURCE[0]}")/../scripts/installer" && pwd)/gke_d
 # TODO(boskos): Once oss-test-infra#2655 merges and deploys Boskos project leasing,
 # consider failing closed if JOB_NAME is set and PROJECT_ID is unset.
 export PROJECT_ID="${PROJECT_ID:-kube-agents-evals}"
+
+# ─── TEMPORARY PIN — REVERT BEFORE MERGE ─────────────────────────────────────
+# Pins this pull request's presubmit to kube-agents-evals-16, to prove #1051's
+# fix end to end: the project has had the fleet_reader_token_creators binding
+# applied by hand, so a run against it must NOT emit the
+# "kubeconfig keeps the runner's own credential" warning that every other
+# project's runs emit three times. Absence of that warning is the result.
+#
+# -16 rather than a live pool project because it is not registered in Boskos
+# yet, so no other pull request can lease it and a bad run breaks nothing.
+#
+# The assignment is unconditional on purpose. The Prow job exports
+# PROJECT_ID="${LEASED_PROJECT}" *before* sourcing this file, so the `:-`
+# default above cannot override a lease and a pin written that way would be
+# silently ignored. Boskos still leases a project and leaves it idle; deploy,
+# eval and teardown all follow PROJECT_ID, so -16 is what gets used and what
+# gets cleaned up, and the idle lease is released as normal.
+#
+# This must not merge: on main it would send every presubmit to one project,
+# serialising the pool behind it. Precedent: 442cbf53 (#995), reverted by
+# 8e42c3b4 before that branch merged.
+export PROJECT_ID="kube-agents-evals-16"
+# ─────────────────────────────────────────────────────────────────────────────
+
 export GCP_PROJECT_ID="${PROJECT_ID}"
 export REGION="${REGION:-us-central1}"
 


lapis2002:
Closing: the trial is moving to a self-contained reproduction in a personal project, and kube-agents-evals-16 is being returned to pool-standard configuration so it can be registered in Boskos.

PR #1211, opened 2026-09-03
feat(operator): render the A2A NATS component, provisioning, and gateway under mode next

bnaylor:
## Summary

Under `spec.mode: next` the operator now renders the stage-1 A2A bus: a NATS/JetStream StatefulSet with a deny-by-default static account layout, a content-hash-named Job that provisions the spec's streams, KV buckets and starter topics, an ingress NetworkPolicy that fences the bus to its enumerated clients, a namespace ResourceQuota bounding session pods, and the A2A gateway Deployment with its ServiceAccount and a Role pinned to a single `get` on its own Deployment. Under `today` (or `mode` absent) `cleanupA2A` returns the stack to dark, and version skew freezes both branches. This implements the merged design of record, `docs/designs/spec-nats-deployment.md` as amended by #1084, behind the mode switch #1129 shipped.

PLAYGROUND POSTURE (stage 1): static per-component NATS users instead of the auth callout, single-node R1 JetStream (production: 3-node R3), no audit exporter, no breaker, gateway sweep as the only janitor. Each has a decided design in the specs; none gates letting people play. Static creds are the playground, not the product. The rendered nats.conf and provisioning script carry this block verbatim so the posture cannot be mistaken for the product when read on a cluster.

## Why This Change

#1129 landed the mode switch so that the components behind it would not each invent a temporary gate; this is the first component. Without it, `mode: next` renders nothing and every later A2A PR (agent bus surface, gateway program, worker) has no bus to join. The bus's authorization model — the server decides who may say what before a message is read — only exists if the operator renders it, which is why the account grants, the ingress fence, and the session-pod quota land with the component rather than after it.

## What Changed

- `platformagent_a2a_manifests.go` (new): every A2A object build, `reconcileA2A`/`cleanupA2A`, and the provisioning script. Objects carry `part-of: a2a-next` plus a `kubeagents.x-k8s.io/a2a-component` label, so the dark stack is one query of its own (`resource-labels.md` now records the carve-out).
- The reconciler gates on `renderMode(...) == ModeNext`: `next` renders, a legitimate `today` tears down, skew touches neither branch (a one-version operator rollback must not kill a live bus). A failed provision Job parks the phase at `Degraded/A2AProvisionFailed`; a pending one requeues, because Jobs are deliberately not watched (a Job watch is a cluster-wide informer every today install would pay for — bookkeeping reads go through `APIReader`).
- **Flip-back semantics, on purpose:** the generated creds Secret survives (inert data; re-enabling `next` must not re-roll credentials a still-running client cached) and the JetStream PVC survives (the file store is the audit substrate; flipping a mode is not license to destroy evidence). CR **deletion** does reap the PVC, guarded by the instance label the claim template stamps.
- **Provisioning is create-only convergence, stated honestly:** the script's `info || add` lines make re-runs clean but never edit an existing stream, so a retention or subject change in a later payload reaches fresh installs only. We hit this live on the fork install on 8/31 and migrated by hand with `nats stream edit`; the Job comment says exactly this rather than implying the hash-rename re-provisions.
- **Session-pod bound, both halves' render side:** `spec.harness.tuning.maxSessions` (new CRD field, default 10) reaches the gateway as `A2A_MAX_SESSIONS` (the usability half: an honest chat refusal) while the namespace ResourceQuota is rendered a fixed headroom above it (the enforcement half: a compromised or buggy gateway ignores its own cap and cannot ignore admission, and its Role cannot touch the quota).
- **Attribution and ownership wiring on the gateway Deployment:** `SESSION_KV_SALT` from the same Secret key and resolver the agent render uses (one human, one pseudonym, on the bus and in session metadata — without it, rotating the bus password would silently unjoin every pseudonym), and `A2A_OWNER_DEPLOYMENT` plus a `deployments get` Role rule pinned by `resourceNames`, so spawned session pods will carry an ownerReference Kubernetes GC can act on. Spawning itself ships **unarmed**: no `A2A_SPAWN_SESSIONS`, no pod verbs on the Role — that RBAC lands with its consumer in the worker PR.
- Operator ClusterRole widens: full verbs on `secrets` (the generated NATS credentials/config), `batch` `jobs` (provisioning), and `resourcequotas` (the session bound). **Security note a reviewer should weigh deliberately:** `secrets` write on the cluster-scoped controller is new authority; the shape is forced because the operator creates dynamically named Secrets per CR namespace, and `reference/security-and-iam.md`'s controller enumeration is updated in this PR.
- Sharp edge, known and accepted: an agent literally named `<name>-a2a` collides with these object names; cleanup's unowned-object refusal then halts that agent's reconcile. The refusal is the safe behavior and the failure needs that exact naming.
- Images: `nats:2.10-alpine`, `natsio/nats-box:0.14.5`, and a dev-registry gateway default are **deliberately not in `images.json`** — the inventory documents the supported install surface and `mode` is an unsupported dev toggle. The code comment names this as graduation debt; mirrored installs flipping `next` must override via `A2A_NATS_IMAGE`/`A2A_PROVISION_IMAGE`/`A2A_GATEWAY_IMAGE` until graduation moves all three to the release pipeline.

## Context

- PR 5 of the A2A series (tracking map on #1129). Depends on **#1129 (mode switch, merged)**. The gateway program PR may open alongside this but **merges after it** (it needs the creds Secret and Deployment this renders); the agent-bus and worker PRs depend on this one.
- One cross-PR interaction found in self-review and deferred with its owner: `spec.security.egressPolicy: Allowlist` renders an agent-pod egress allowlist with no 4222 rule, so an agent-side bus client (none exists in this PR) would hang at the fence. The agent joins the bus in the agent-bus PR, and that is where the allowlist gains its mode-gated 4222 rule.

## Testing

- `go build ./...`, `go test -race ./...` green in `k8s-operator/` (controller, testing, webhook, api packages).
- Today renders proven unchanged: the golden suite passes with **no golden churn** — a normal install cannot tell this feature exists, which is the mode spec's core promise.
- `make docs-check`, `make chart-check`, `prettier --check` on changed Markdown/YAML all green. CRDs, deepcopy, RBAC and the chart mirrors regenerated, not hand-edited.

### Live validation

Scratch install, not the shared demo: a fresh `up05-scratch` namespace on the otherwise idle `a2a-stream-demo` GKE cluster (project `bnaylor-kagents-dev`), CRDs from this branch, operator run from this branch's source (`ENABLE_WEBHOOKS=false`, pinned kubeconfig). Full transcript sequence:

1. **Baseline (`mode` absent):** today's stack rendered (gateway Deployment, Service, RBAC, netpol, ConfigMaps, PVCs); `kubectl get ... -l app.kubernetes.io/part-of=a2a-next` returned nothing.
2. **Flip to `next`:** within 30s — NATS StatefulSet `1/1`, provision Job `Complete 1/1` (the script ran to completion on the real `nats-box` image), gateway Deployment rendered, ingress NetworkPolicy and creds/config Secrets present, ResourceQuota live at `pods: 2/25`. The gateway pod pulled and exited with `DISCORD_TOKEN is required` — the documented crash-loop-until-Secret posture, and evidence the optional `SESSION_KV_SALT` reference lets the pod schedule without the Secret.
3. **RBAC pin, proven against the server:** as `system:serviceaccount:...:scratch-a2a-gateway` — `get` its own Deployment **yes**; `get` the agent's Deployment **no**; `create pods` **no** (the worker PR's boundary holds); `delete` the session quota **no** (the enforcement-half argument holds).
4. **Deployment assertion 1 (unauthorized publish refused):** from a seed-labeled nats-box pod, `pub a2a.topics.shared.probe` as the `web` user → `Permissions Violation for Publish to "a2a.topics.shared.probe"` — the writerless probe subject refusing exactly as designed.
5. **Deployment assertion 2 (acked message replayable):** published `a2a.tasks.demo.t1.in` as `gateway`, consumed and acked it on an explicit-ack durable, then read it back from the stream — acknowledgement does not delete.
6. **Flip back to `today`:** everything torn down except the creds Secret and the JetStream PVC; today's stack untouched. **CR deletion:** the PVC reaped by the finalizer.
7. Cleanup confirmed: namespace and CRDs removed; the cluster's unrelated standing workload untouched.

Not covered live, and why: NetworkPolicy **enforcement** — this cluster runs the legacy datapath and does not enforce policy (an unlabeled pod reached 4222, getting a NATS-level auth refusal instead of a network hang), so the fence's enforcement evidence remains the earlier run on the enforcing Dataplane V2 install, where in-cluster 9222 was refused while `kubectl port-forward` (node path) was not; this run proves the rendered object and its exact shape is pinned by unit test. The gateway program's behavior beyond boot-time config parsing is the gateway PR's to validate. The PVC-delete guard's label check landed after this run; it is covered by unit test (`TestHandleDeletionReapsOnlyTheLabeledJetStreamPVC`).

## Self-Review

Both passes ran in fresh subagent contexts handed the diff range and the repo skills, not the authoring conversation.

**Adversarial pass** — 8 findings, dispositions:

- **Fixed:** creds-Secret values now validated against the exact generator shape (`^[0-9a-f]{32}$`) before interpolation into nats.conf; a malformed key is re-rolled like a missing one, closing a path where a one-time Secret write became durable bus authority the operator re-rendered forever (regression test added).
- **Fixed:** `seccompProfile: RuntimeDefault` on all three new pod specs, matching every sibling render (the asymmetry would have failed Pod Security `restricted` admission for the A2A pods only).
- **Fixed:** the JetStream PVC delete on CR deletion is now guarded by the instance label the claim template stamps — a name is not ownership; a squatter PVC wearing the exact name is left alone (regression test added).
- **Fixed (comment):** the provision Job TTL's recreation cycle (a completed Job is removed after 24h and the idempotent script re-runs) is now stated as chosen, with the trade named.
- **Deferred with owner:** the `egressPolicy: Allowlist` × bus-client interaction (see Context) — no agent-side bus consumer exists in this PR; the rule lands with the consumer in the agent-bus PR.
- **Accepted with rationale:** `images.json` exemption (see What Changed; the inventory documents the supported surface, the exemption is named graduation debt in the code).
- **Accepted, surfaced here:** the operator ClusterRole's `secrets` widening (see What Changed's security note).
- **Rejected:** a two-agents-per-namespace quota under-provisioning scenario — the admission webhook rejects a second `PlatformAgent`, so the layout the finding needs cannot be reached.
- The pass also verified, against source: spec conformance of streams/retention/caps, all netpol peer selectors traced to the pod templates that carry them, the skew freeze, the config-hash roll, and CRD/deepcopy consistency for `maxSessions`.

**Docs-drift pass** — 6 blocking findings, all fixed in this PR: the CRD page's "nothing renders differently under `next` yet", the missing `maxSessions` CRD row, the security reference's controller-permission enumeration, the mode-switch spec's status line, the two stale doc-map status cells, and the resource-labels contract (the `part-of: a2a-next` carve-out is now documented). Advisory findings fixed: the operator overview's reconcile list, the `A2AProvisionFailed` status reason, a wrong spec filename in a code comment, the `networkPolicy.enabled` clause in security-and-iam. Advisory accepted: `images.json` (above); the doc map's identifier-sources cells still name the primary owners and were left untouched to avoid map churn.

## Risk & Rollout

Dark by construction: no call site outside the `renderMode` gate reaches the new file, the today render is byte-identical (golden-proven), and the CRD field additions are optional. The new authority is the operator ClusterRole widening named above — present whether or not `next` is ever set, which is why it is called out for review rather than buried. Revert is a clean `git revert`; on an install that had flipped `next`, revert-then-reconcile leaves the A2A objects orphaned-but-owned (ownerReferences to the CR), so deleting the CR or the objects by the `part-of: a2a-next` label cleans up fully.

---

Generated with the help of Claude (Anthropic).


docs/README.md
```diff
@@ -318,10 +318,9 @@ pull request:
...
+| `docs/designs/spec-subagent-profiles.md` | Feature design | Subagents as declarative profiles: the `AgentProfile` resource, submission as a bus message rather than an API call, dispatcher-rendered Jobs, four reserved artifact names, the janitor for orphaned tasks, and the kanban retirement inventory with its two named gaps. | `AgentProfile`, dispatcher, artifact names, lifecycle, kanban gap table | Profile schema and strict loader implemented in `a2a/profiles`; the CRD and dispatcher are not || `docs/contributing.md` | Contributor guide | Short entry point: Google CLA and community guidelines, deferring everything else to the site's contributing page and `AGENTS.md`. | CLA, pointers | Human contributors |
```
mplakhtiy: **HIGH — the `docs/contributing.md` map row was swallowed by the row above it.**

Line 323 on this branch is exactly `main`'s line 323 concatenated with `main`'s line 324 — the newline between the `spec-subagent-profiles.md` row and the `docs/contributing.md` row is gone. Verified byte-for-byte:

```
main   323 (500 chars) | `docs/designs/spec-subagent-profiles.md` | … | … are not |
main   324 (216 chars) | `docs/contributing.md` | Contributor guide | Short entry point: … |
branch 323 (716 chars) …are not || `docs/contributing.md` | Contributor guide | Short entry point: …
```

**Failure scenario.** The documentation map renders with `docs/contributing.md` gone as a row: its five cells become cells 6–10 of the `spec-subagent-profiles` row, whose Status cell now reads `… are not |` followed by a stray empty cell. `make docs-check-map` passes anyway — it reported *"inventory covers all 254 tracked docs"* in CI — because it matches the path token, which is still present in the file, not the row structure. The mechanical gate cannot catch this and the next person to edit the map inherits a corrupt table.

The only map edits this change needs are the Status cells on `spec-mode-switch.md` (321) and `spec-nats-deployment.md` (322), both of which are correct. AGENTS.md, Documentation Guidelines: *"Add a document to the map (`docs/README.md`) with one line, and change nothing else there … the map is edited from several branches every week, and a re-aligned table rewrites rows your PR did not author."*

**Fix:** restore the newline before the `| `docs/contributing.md` |` cell. One character.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -0,0 +1,1203 @@
...
+						Image:   a2aProvisionImage(),
+						Command: []string{"sh", "-c", script},
+						Env: []corev1.EnvVar{{
+							Name: "SEED_PASSWORD",
+							ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{
+								LocalObjectReference: corev1.LocalObjectReference{Name: a2aNATSName(agent) + "-creds"},
+								Key:                  "seed-password",
+							}},
+						}},
+					}},
```
mplakhtiy, line 813: **HIGH — three new operator-built containers skip `hardenedSecurityContext()`; this one has no `SecurityContext` at all.**

The provision Job's container declares `Name`, `Image`, `Command`, `Env` and nothing else. No `AllowPrivilegeEscalation: false`, no `Capabilities: Drop [ALL]`. Its pod spec (lines 781–783) carries only `SeccompProfile`; unlike the NATS and gateway pods it has no `RunAsNonRoot` and no `RunAsUser`.

The NATS container (`platformagent_a2a_manifests.go:536-539`) and the gateway container (`:989-992`) set `AllowPrivilegeEscalation: false` + `Drop: [ALL]` but omit `ReadOnlyRootFilesystem`.

The repository has a helper for exactly this — `hardenedSecurityContext()` at `platformagent_manifests.go:2519-2525`, which returns all three — and an invariant test at `platformagent_manifests_test.go:5678-5714` whose failure message is *"if the operator grew a container, give it `hardenedSecurityContext()` and add it to that list"*. That test walks the agent Pod only, so three new containers grew outside its reach and nothing failed. The nearest sibling render outside the agent pod, the credential broker container (`platformagent_broker_split.go:262-269`), sets `ReadOnlyRootFilesystem: true` as well.

**Failure scenario.** The provision pod holds the bus seed credential and runs a shell script with the container's default capability set and privilege escalation permitted — the widest posture any operator-rendered container in this repository has. Secondarily, a namespace enforcing Pod Security `restricted` rejects the pod; the Job then burns its `backoffLimit: 20` on `FailedCreate` and eventually parks the phase at `A2AProvisionFailed`, so the bus never gets its streams. (No install path in this repository sets `pod-security.kubernetes.io/*` labels today, so that half is latent, not immediate.)

`seccompProfile: RuntimeDefault` alone does not satisfy `restricted`, which also needs `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]` and `runAsNonRoot: true`.

**Fix:** give the provision container `hardenedSecurityContext()` (or at minimum `allowPrivilegeEscalation: false` + `drop: [ALL]`) and its pod `runAsNonRoot`/`runAsUser`. `ReadOnlyRootFilesystem` on NATS and the gateway is the weaker half: `nats-server` writes only to `/data`, so it should take a read-only root; the gateway image is #1212's code and I cannot show it does not write to `/`. Splitting that part off is reasonable.

mplakhtiy (commented):
Findings from a full review of this branch, graded MEDIUM and above. Two HIGH, four MEDIUM.

k8s-operator/internal/controller/platformagent_controller.go
```diff
@@ -164,9 +164,16 @@ type PlatformAgentReconciler struct {
 // DaemonSet port (issue #747 B4) — a second consumer of a grant that already existed for
 // buildMinimalPlatformRole's escalation-prevention requirement.
 // +kubebuilder:rbac:groups="",resources=serviceaccounts;persistentvolumeclaims;configmaps;services;pods,verbs=get;list;watch;create;update;patch;delete
+// `secrets` and full `jobs` verbs exist for the mode-next A2A stack: the
+// generated NATS credentials/config Secrets and the provisioning Job.
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
```
mplakhtiy: **MEDIUM — the new `secrets` grant includes `list;watch`, which nothing uses.**

The A2A code does one `Get` (`platformagent_a2a_manifests.go:178`), one `Create` (`:222`), one `Update` (`:196`), one server-side-apply `Patch` (via `applyManaged`) and one `Delete` (`:1180`). `grep` for `corev1.SecretList` across `internal/` and `cmd/` returns nothing outside tests, and `secrets` is absent from the agent ClusterRole's resource lists (`platformagent_manifests.go:3735,3750`), so no RBAC escalation-prevention requirement pins it either. Contrast `resourcequotas` and `jobs`, where `list`/`watch` **are** needed — the agent role grants read on both, and `cleanupA2A` lists Jobs.

**Failure scenario.** The cluster-scoped controller can enumerate and stream every Secret in the cluster — agent credentials, GSA keys, the Discord bot token, LiteLLM keys — for no code path.

**Fix:** drop `list;watch` from the marker, regenerate `role.yaml` and the chart mirror, and adjust the security-and-iam sentence. It costs nothing and enforces the design decision the file already made: `a2aReader()` exists precisely so Secret reads never go through the cache, and without `list;watch` a cached Secret read cannot be added by accident.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -0,0 +1,1203 @@
...
+// Session pods — spawned by the gateway once the worker PR arms spawning —
+// are the gateway's, not the operator's: every spawned pod carries an
+// ownerReference to the gateway Deployment (A2A_OWNER_DEPLOYMENT above), so
+// deleting the gateway here hands any stragglers to Kubernetes GC, with no
+// operator exception to the IsControlledBy refusal below.
+func (r *PlatformAgentReconciler) cleanupA2A(ctx context.Context, agent *agentv1alpha1.PlatformAgent) error {
+	// Deployment/StatefulSet/Service reads come from the cache — those kinds
+	// are already watched (Owns) so the reads are free. Secret and Job reads
+	// go through a2aReader: a cached read would start a cluster-wide informer
+	// for a kind this controller otherwise never watches, on every install.
```
mplakhtiy: **MEDIUM — `cleanupA2A` costs six uncached API round-trips on every reconcile of every `today` install.**

`cleanupA2A` runs unconditionally whenever the mode is not `next`, which is every reconcile of every supported install forever. Five of its nine named reads go through `a2aReader()` (RoleBinding, Role, ServiceAccount, config Secret, ResourceQuota) and the Job sweep is an uncached `List` — six direct API-server calls per reconcile per agent, on installs that have never rendered an A2A object. The golden-test change makes the cost explicit: *"the mode gate's cleanup path lists Jobs on every today-mode reconcile, which is every golden case."*

One of those five is also justified by a comment the code contradicts. The comment says a cached read *"would start a cluster-wide informer for a kind this controller otherwise never watches"*, but `SetupWithManager` already registers `Owns(&corev1.ServiceAccount{})` (`platformagent_controller.go:2169`), so ServiceAccounts are cached and that uncached Get buys nothing.

**Failure scenario.** Steady-state API-server load scales with agent count × reconcile rate for work that is a no-op on every supported install.

**Fix:** a cached read of the StatefulSet and the gateway Deployment (both `Owns` kinds, both free) returning `NotFound` is a sufficient early-exit. And correct the ServiceAccount half of the comment.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -0,0 +1,1203 @@
...
+			switch cond.Type {
+			case batchv1.JobComplete:
+				state.done = true
+			case batchv1.JobFailed:
+				state.failed = true
+				state.message = fmt.Sprintf(
+					"A2A provision Job %s failed (%s: %s); the bus has no streams until it succeeds. Inspect its pod logs; deleting the Job retries.",
+					existing.Name, cond.Reason, cond.Message)
+			}
+		}
```
mplakhtiy, line 1120: **MEDIUM — `A2AProvisionFailed`, `state.done`, and the pending requeue are all untested.**

This block scans the Job's conditions; `platformagent_controller.go:463-468` turns `state.failed` into a Degraded phase and `:484` turns `!state.done` into a requeue. Nothing in the 21 new tests seeds a Job with a `JobComplete` or `JobFailed` condition — `grep A2AProvisionFailed internal/` matches only the controller line — so all three branches ship unexercised. The behaviour is documented as user-visible in `docs/site/src/content/docs/operator/platformagent-crd.md` (*"a failed A2A provisioning Job parks the phase the same way with `A2AProvisionFailed`, naming the Job to inspect"*).

**Failure scenario.** Dropping `case batchv1.JobFailed`, inverting the `cond.Status != corev1.ConditionTrue` guard, or reading the wrong condition type leaves a bus with no streams reporting `phase: Ready`, and the suite stays green.

**Fix:** the fake client already used by `TestReconcileA2AGatedByMode` can seed both conditions in a few lines.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -0,0 +1,1203 @@
...
+					}},
+				},
+			},
+			VolumeClaimTemplates: []corev1.PersistentVolumeClaim{{
+				// The labels ride to the PVC the StatefulSet controller
+				// stamps out, which is what lets handleDeletion verify the
+				// claim is this render's before deleting it — a template PVC
+				// carries no owner reference, so the instance label is the
+				// only ownership signal it has.
+				ObjectMeta: metav1.ObjectMeta{Name: "data", Labels: a2aLabels(agent, "nats")},
```
mplakhtiy: **MEDIUM — the JetStream claim name is a hardcoded string spelled twice in two files, and drift between them fails silently.**

- here: `ObjectMeta: metav1.ObjectMeta{Name: "data", …}` on the volumeClaimTemplate
- `platformagent_controller.go:550`: `Name: "data-" + a2aNATSName(agent) + "-0"`

**Failure scenario.** Rename the volumeClaimTemplate and `handleDeletion`'s reap silently becomes a no-op: the Get returns `NotFound`, the `switch` falls through the `client.IgnoreNotFound` arm, the finalizer is removed, and a 40Gi PV leaks on every CR deletion. Nothing errors, and `TestHandleDeletionReapsOnlyTheLabeledJetStreamPVC` still passes because it builds its fixture from the same hardcoded string it is testing.

**Fix:** a shared constant used by both sites closes it.

docs/README.md
bnaylor: Confirmed, and this one is mine rather than the branch's — I resolved a docs/README.md merge conflict on this PR yesterday with a script that dropped the newline before the `docs/contributing.md` row. Verified exactly as described: line 323 was 716 chars = main's 500 + 216.

Fixed in fa065078. `git diff upstream/main -- docs/README.md` is now 2 insertions / 2 deletions — the `spec-mode-switch` and `spec-nats-deployment` Status cells and nothing else; lines 323-325 are byte-identical to main again.

The point about `make docs-check-map` passing on a path-token match is the useful half: the gate cannot see row structure, so nothing would have caught this. Worth a length-delta check in that script — noted for a follow-up rather than grown into this PR.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Confirmed on all three counts, and fixed in fa065078.

All three containers now take `hardenedSecurityContext()`. The provision pod also gains `RunAsNonRoot`/`RunAsUser: 1000` to match the NATS and gateway pods. The one thing that needed more than the helper: the `nats` CLI wants a writable context directory even though every call in the script passes `--server`, so a read-only root would have been a bet rather than a hardening — the container now mounts an emptyDir at `/tmp` with `HOME` and `XDG_CONFIG_HOME` pointed at it, which makes it deterministic instead.

On the invariant test walking the agent pod only: that is exactly how three containers grew outside it, so `TestEveryA2AContainerHasAHardenedSecurityContext` now walks the NATS, provision and gateway pod specs with the same four assertions plus the pod-level floor. Proven to bite: removing the provision container's context again fails it with "container provision: no SecurityContext; want hardenedSecurityContext()".

Not verified live — the read-only root on the provision Job wants a real run against the install before I would call it proven, and I have flagged that to bnaylor rather than claim it here.

k8s-operator/internal/controller/platformagent_controller.go
bnaylor: Confirmed — verified the claim independently: no `corev1.SecretList` anywhere in `internal/` or `cmd/` outside tests, and the A2A paths are Get/Create/Update/apply-Patch/Delete by name.

Dropped `list;watch` from the marker in fa065078, regenerated `role.yaml` with `make manifests` (the rule now reads create/delete/get/patch/update), synced the chart mirror with `hack/sync-chart-manifests.sh`, and corrected the security-and-iam sentence to say the controller holds Secrets by name only and cannot enumerate them.

Your reasoning for why it costs nothing is the part worth keeping: `a2aReader()` exists so Secret reads never go through the cache, and without the enumeration verbs a cached read cannot be added by accident later.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Confirmed, including the reason it stays silent — the existing test builds its fixture from the same literal it asserts, so it cannot see the drift. Fixed in fa065078: `a2aNATSDataClaim` is declared once above the StatefulSet builder and used by both the volumeClaimTemplate and `handleDeletion`'s `pvcKey`, with the leak spelled out in its doc comment.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Confirmed — `grep A2AProvisionFailed internal/` matched only the controller line here too. Fixed in fa065078 with `TestA2AProvisionJobConditionsDriveStatus`, four cases against the fake client: pending (requeue), complete (`done`), failed (`failed` plus a message naming the Job and carrying the condition reason), and a fourth you did not ask for but that the failure scenario implies — a `JobFailed` condition with `Status: False`, which pins the `ConditionTrue` guard rather than the condition type alone.

Proven to bite: deleting the `cond.Status != corev1.ConditionTrue` guard fails the `failed-but-false` case.

bnaylor: (Meant for the cleanupA2A thread — replying there.)

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Confirmed — `grep A2AProvisionFailed internal/` matched only the controller line here too. Fixed in fa065078 with `TestA2AProvisionJobConditionsDriveStatus`, four cases against the fake client: pending (requeue), complete (`done`), failed (`failed` plus a message naming the Job and carrying the condition reason), and a fourth you did not ask for but that the failure scenario implies — a `JobFailed` condition with `Status: False`, which pins the `ConditionTrue` guard rather than the condition type alone.

Proven to bite: deleting the `cond.Status != corev1.ConditionTrue` guard fails the `failed-but-false` case.

(Apologies for the two stray replies on the `cleanupA2A` thread — I crossed the comment ids. This is the reply that belongs here; the `cleanupA2A` answer is now on its own thread.)

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: (Ignore my two replies above on this thread — I crossed comment ids; they belong to the untested-branches finding and are now posted there. This is the answer to the `cleanupA2A` one.)

**Half fixed, half escalated.**

Fixed in fa065078: the ServiceAccount claim was simply wrong, exactly as you say — `SetupWithManager` already `Owns(&corev1.ServiceAccount{})` at `platformagent_controller.go:2172`, so that uncached Get bought nothing. It now uses the cached client, and the comment names which kinds are cached (Deployment/StatefulSet/Service/ServiceAccount) versus which genuinely are not (Secret, Role/RoleBinding, ResourceQuota, Job), plus a line making the standing cost explicit rather than incidental.

Not taken unilaterally: the early-exit. The cost is real and your arithmetic holds, but a StatefulSet+Deployment probe is only a sufficient exit if those two are the *last* things `cleanupA2A` deletes. As ordered today the gateway Deployment goes first, so a pass that deletes it and then errors would, on the next reconcile, find both absent and early-exit over whatever it had not reached yet — converting a retried cleanup into a silent leak. That is fixable (delete the two probes last, or probe something deleted last), but it is a change to teardown ordering on a path whose whole job is not leaking, so it wants the branch author rather than the PR minder. Escalated to @bnaylor with that framing; if he wants it in this PR I will do it with the reordering and a test that fails a mid-pass error into a second complete pass.

bnaylor: Done in 98e820d9 — bnaylor took the call, so the early exit is in.

Three sentinels rather than the two you proposed, and the third is the interesting one. The render creates the config Secret **before** the StatefulSet, so a probe of {StatefulSet, Deployment} alone would step over a render that died between those two and leave a config Secret on a `today` install — a darkness-property break rather than an untidy one. So:

- **StatefulSet** — moved to LAST in the deletion list, which is what makes its absence mean "an earlier pass ran to completion" instead of "an earlier pass died partway".
- **gateway Deployment** — stays first (it still stops the gateway before its dependencies go, and hands spawned pods to GC), so a pass that failed immediately is caught.
- **config Secret** — the one object created before the StatefulSet. The only uncached read, and it only happens when the two free ones both miss.

Two tests, each proven red against the unfixed code:
- `TestCleanupA2AResumesAfterAMidPassError` injects a Delete failure on the Role, then runs a clean pass and requires the tree to be **empty** — not merely error-free. With the StatefulSet deleted first again it fails at the sentinel check with "the sentinel must survive a failed pass, or this test proves nothing".
- `TestCleanupA2ACostsThreeReadsWhenThereIsNothingToClean` counts interceptor Gets and Lists on a never-rendered install: 3 Gets, 0 Lists. A correctness test would pass with the per-object walk still running, so the cost claim needed its own measurement.

Thanks for this review — five of the six were straightforwardly right, and the two HIGHs were both things our own gates could not see.

mplakhtiy (approved):

bnaylor:
@dshnayder flagging you directly on one thing, because this PR walked into an invariant you set in #913 (`f66d1957`) and I would rather you checked my resolution than discovered it later.

**What collided.** #913 introduced the operator's first `secrets` grant — `get`, for `checkShellSandboxKeys` — and fenced it immediately with `tests/test_operator_secrets_grant.py` and the KSV-0041 exemption, on the reasoning that trivy cannot scope an exemption to one rule so exempting that grant exempts the whole rbac directory. This PR, cut before yours landed, needs `create`/`update`/`patch`/`delete` on the same rule: under `mode: next` the operator generates the NATS credential Secret once, repairs a missing key, renders the config Secret, and deletes the config Secret on the way back to `today` — all owner-referenced to the `PlatformAgent`. Merging main in made your test fail, correctly.

**What I did**, in `16fe4f01`:

- Widened `PERMITTED_SECRET_VERBS` to the five, and rewrote your docstring to carry both callers and to say the set widened *once, deliberately* — the exact-set pin stays, so the next addition still fails there rather than passing under the ignore.
- Merged both rationales into the `.trivyignore.yaml` entry rather than replacing yours.
- Added `test_the_operator_can_never_enumerate_secrets`, asserting `list`/`watch` are absent, called out separately from the exact-set check. That felt like the property actually worth fencing on a cluster-scoped role — enumeration of every Secret in every namespace is a different power from by-name access — and `a2aReader()` exists so every A2A Secret read is by name and uncached. So the ceiling is narrower in that dimension than a plain "five verbs" would be.
- Your RBAC self-check earned its keep, by the way: it failed on the new grants until I declared them in `requiredPermissions`, which is exactly the drift it exists to catch.

**The question for you:** widening a ceiling you set one day earlier is your call more than mine. If you would rather the grant stayed at `get` — the alternative being the provisioning Job minting its own credentials instead of the operator, which is a real design change to this PR and not a small one — say so and I will take it back out rather than have it merge on my judgment. Everything else here is independent of that choice.

Issue #1126, opened 2026-09-01
fix(proxy): gcloud read-only allowlist withholds the discovery reads its leaf reads need — networking audit's Cloud NAT check has never run


dshnayder:
The credential proxy's gcloud read-only allowlist grants several leaf reads without granting the `list` read that discovers the argument those leaf reads require. The daily `gcp-networking-fabric-audit` cron has therefore never executed one of its five checks, and it publishes a ledger issue that reads as a completed audit.

Found on a running install, not by inspection.

## What the run did

`platform-agent-credential-proxy` refused 21 gcloud commands under `rule=gcp.read-only` in 48 hours. Two clusters:

```
2026-09-01 08:03:26  hint=compute.routers.list
2026-09-01 08:03:52  hint=compute.routers.list-usable
2026-09-01 08:03:53  hint=compute.routers.list
2026-09-01 08:04:10  hint=compute.routers.list
2026-09-01 08:06:04  hint=compute.routers.list
2026-09-01 08:06:19  hint=compute.routers.list
2026-09-01 08:06:56  hint=compute.routers.list
2026-09-01 08:07:02  hint=compute
2026-09-01 08:08:35  hint=compute.routers.list-usable

2026-09-01 15:49:05  hint=compute.firewall-rules.list
2026-09-01 16:37:48  hint=compute.instances.describe
2026-09-01 16:40:17  hint=compute.firewalls.list
2026-09-01 16:41:25  hint=compute.networks.describe
2026-09-01 16:41:34  hint=compute.firewall-rules.list
2026-09-01 16:41:44  hint=compute.firewall-rules.list
2026-09-01 16:42:23  hint=compute.firewall-rules.describe
2026-09-01 16:42:53  hint=compute.routers.list
2026-09-01 16:43:01  hint=compute.routers.list
2026-09-01 16:43:15  hint=compute.routers.list
```

The 08:0x cluster is `gcp-networking-fabric-audit`, whose schedule is `0 8 * * *` (`agents/platform/governance/gcp_networking_fabric_sop.md:5`). Seven attempts at `compute routers list` inside five minutes is the model retrying a refusal it cannot interpret. The afternoon cluster is a separate networking investigation on the same install and hits the same wall from a different direction.

[...]

Issue #953, opened 2026-08-25
Platform agent asks which cluster instead of searching for the workload


jayantid:
## What happened

The first live run of `cluster-agent-crashloop-debug` (Prow build `2092323330051805184`, PR #939, 2026-08-25) scored `OutcomeValidity=0.0` and `VerificationCorrectness=0.0` in 14.6s of agent execution. The platform agent's entire output was:

> Which cluster is the payments-api deployment running in? I can check seeded-a, seeded-b, seeded-c, or one of the other clusters.

Its only tool call was `multiuser_memory` (a read, which returned empty). It made no attempt to look.

The prompt was `The payments-api deployment in namespace seeded-debug is crashlooping. Investigate and report the root cause.` — no cluster named.

**The request was answerable.** Verified read-only against all three Boskos pool projects: `seeded-debug` exists on the `seeded-a` slot cluster only. Slots `b` and `c` carry the version-laggard and consistency-outlier fixtures and have no such namespace. Exactly one cluster in each project could have been the answer.

## The agent was following its instructions

This is the part worth fixing. `agents/platform/skills/cluster-agent-lifecycle/SKILL.md` is the skill that owns delegation, and its scope line reads:

> For any request that concerns runtime behavior of workloads on a **single, specific** cluster (crash loops, OOMs, scheduling failures, ...), delegate to that cluster's Cluster Agent instead of investigating yourself.

Step 1 of that procedure is:

```bash
python3 /opt/data/scripts/cluster_agent_profile.py name \
  --project "<project>" --cluster "<cluster>" --location "<location>"
```

`--cluster` and `--location` are required inputs. **The skill has no branch for "the requester did not say which cluster."** Given a request it cannot route, asking the human is the only path the skill describes. The agent did the documented thing.

The automatic route does not cover this either. `session_kv_server.trigger_agent_troubleshooter` delegates to the profile scoped to the cluster that raised the event — it is driven from an event row (`session_kv_server.py:1765`), so it only fires for cluster-originated alerts. A chat-initiated request has no equivalent resolution step.

## Everything needed to fix it already exists

No new tooling is required. The two missing pieces are both already built and already documented in the same file:

- **Enumerate the fleet** — `cluster_agent_profile.py list` (`cluster_agent_profile.py:357`, documented at `SKILL.md:109`, but in the lifecycle/pruning section, not the delegation section).
[...]

PR #1215, opened 2026-09-03
docs(designs): propose an opt-in multi-project scope for the Platform Agent

bradhoekstra:
## Summary

Adds `docs/designs/multi-project-scope.md`, a design of record proposing that a single kube-agents install manage GKE clusters across several GCP projects through an opt-in `spec.scope` on the `PlatformAgent` resource. Today the cluster profile sync resolves exactly one project and lists it once, the IAM module binds roles in exactly one project, and the architecture documents define the Platform Agent as one per project. The design keeps the per-cluster profile model, which is already project-qualified, and changes discovery, IAM, the declarative surface, and the documented boundary. Nothing in this PR ships code; the document says what is true on `main` now, with citations, and what would change.

## Why This Change

An organisation with clusters in several projects installs kube-agents once per project, with no view across them. PR #588 attempted this on the old bash provisioning path and was closed after #797 replaced that path; its closing comment says `main` now handles multi-project IAM and reconciliation natively, which is not the case. Epic #618 still carries the phase. Without a design that starts from what `main` does, the next attempt re-derives the same dead ends.

## What Changed

- `docs/designs/multi-project-scope.md`: the problem as it stands in source, what already generalises, the scope model (`projects`, `folders`, `organizations`, `exclude`), resolution through Cloud Asset Inventory, a per-project outcome snapshot with container-level freeze and a three-condition prune, IAM through a fixed read-role allowlist bound per project or per container, the onboarding lifecycle, an inventory of every other place that assumes one project, the boundary change to the architecture documents, a five-phase order, and open questions.
- `docs/README.md`: one inventory row.

## Context

- Relates to #618 phase 3. Supersedes the closed #588, whose closing rationale the design corrects.
- #566 (reconcile degrades to prune-only, exit 0), #953 (agent cannot route a request without a named cluster), and #1126 (broker allowlist withholds discovery reads) are the same problem seen from the agent's side and are cited where the design touches them.
- The design cites source by `file:line`. Open PRs #913, #945, #1123, and #1211 insert above some of those lines in `cluster_agent_reconcile.py`, `cluster_agent_profile.py`, `credential_proxy.py`, `kube-agents-iam/variables.tf`, and `common_types.go`; every citation also names its symbol, so a reader can recover after those merge. #945 adds a ninth default role, which the design's read-role allowlist already excludes by construction.
- Chosen over an issue: the finding is settled enough that the useful artifact is the design, and the issues to file fall out of its implementation order.

## Testing

- `make docs-check` passes on the branch head: generated regions current, links resolve, terminology check, map coverage, context budget.
- `prettier --check` (3.9.6, the pinned version) passes on both files.
- Every `file:line` citation in the document was verified against the branch head by the review passes below, and by me for the ones added in the last round.

### Live validation

Not live-tested: this is a docs-only design document and no code path reaches an installation.

One claim was checked read-only against the shared gkedemos install (`bhoekstra-gkedemos`, cluster `kube-agents-cluster`, platform agent image `platform-agent:dns-endpoint-7dac349`) to confirm the deployed reconcile is single-project. The deployed script resolves one project and lists it once:

```
$ kubectl exec -n kubeagents-system platform-agent-gateway-f9db8f877-w28fg -c platform-agent -- \
    grep -nE "def _project|RECONCILE_PROJECT|\"clusters\", \"list\"|_all_clusters\(" /opt/data/scripts/cluster_agent_reconcile.py
70:def _project() -> str | None:
71:    p = os.environ.get("RECONCILE_PROJECT") or _metadata("project/project-id")
91:def _all_clusters(project: str) -> list:
104:            ["gcloud", "container", "clusters", "list", "--project", project,
183:        for (proj, cluster, location) in sorted(_all_clusters(project)):

$ kubectl exec ... -- sh -c 'cd /opt/data && python3 /opt/data/scripts/cluster_agent_reconcile.py --dry-run 2>&1; echo "exit=$?"'
[CLUSTER-RECONCILE] Reconciling 0 managed profile(s) (dry-run).
[CLUSTER-RECONCILE] Done: created=0 pruned=0 kept=0 no_identity=0 unknown=0.
exit=0
```

The dry run created nothing and the lease was released afterwards. Side finding, not addressed here: that install runs an older revision of the script, has zero cluster profiles, and reports nothing about its create pass not running, which is #566 as deployed.

## Self-Review

Passes: `review-adversarial` seven times and `review-docs-drift` five times, each in a fresh subagent handed only the diff range and the `make docs-check` output, with instructions not to read commit messages or plan files. Every round after the first re-ran against the head the previous round's fixes produced; the last round's two fixes (a clause and a table cell) were verified directly rather than re-reviewed. The adversarial pass ran the document's `gcloud asset search-all-resources` command read-only against two projects and checked every issue and PR number with `gh`.

Looked for: false claims about `main` (every citation opened at source), contradictions between the design's own sections, IAM blast radius, silent-failure paths in the resolution and prune rules, docs-map and formatting rules, and open PRs that overlap or shift the cited lines.

Found and fixed, thirty-one findings across the rounds, the material ones being:

- HIGH: grants outside the host project carried `project_roles`, which includes `iam.serviceAccountUser` and whatever a `custom` set adds. Now a fixed read-role allowlist, per project and per container.
- HIGH: Asset Inventory's `project` field is the project number; the design keyed on it. Now parsed from the asset name, both regional and zonal shapes.
- HIGH: a failed folder lookup would have made every project under it "out of scope" and pruned every profile in one tick. Containers now have outcomes and a failed container freezes its members.
- HIGH: the out-of-scope prune had no memory of what was declared, so an upgrade with an empty scope would have deleted hand-onboarded cross-project profiles. Now a three-condition rule with a `retiring` state.
- HIGH: an exclude-only scope counted as unset, so a migrated exclusion never reached the pod. Rendering and resolution are now separate conditions.
- MEDIUM: `exclude.clusters` matched bare names across projects; now a project-qualified triple. `RECONCILE_EXCLUDE` was described as management-project-only; it is project-blind on `main`, and the transition keeps that.
- MEDIUM: a folder-reached project's outcome could never become `denied` once its profiles existed; PRUNE's 403 now feeds it.
- MEDIUM: the `asset` verb needs its two value flags in the broker's flag set as well as the allowlist entry, or it is admitted and unreachable at once.
- MEDIUM: the snapshot schema had no place for the states the lifecycle section relied on.
- Rows in the single-project inventory that named no assumption (event watcher, DNS endpoint script) removed; the five other `RECONCILE_EXCLUDE` consumers and the CRD's `RECONCILE_PROJECT` comment added; a dozen citation and wording corrections.

Adopted from PLAUSIBLE findings as the author's choice rather than as confirmed defects: dropping the Resource Manager fallback resolver in favour of enabling the Asset API in the composition (recorded as an open question for organisations that forbid it), and stating in the boundary section that the asset allowlist entry widens the agent's read to the container's asset index.

Deliberately not fixed:

- Line-number citations (both passes, every round): every one is verified at the branch head, sibling designs cite lines the same way, and the line is what lets a reader check the claim. The open PRs that will shift them are listed under Context.
- The prior-art paragraph narrates closed history around #588 and #797. It is history that explains why the design exists, not status a merge would silently stale.
- Sections 3 and 7 use the bold-rule list form the documentation rules prefer prose over. Each entry is a named rule an implementer will look up by name.

Not covered by any pass: GCP-side facts asserted from measurement or documentation, specifically whether `roles/cloudasset.viewer` on a folder plus the API enabled in the host project suffices, whether `describe` against a deleted project returns 403 or 404, and where `roles/mcp.toolUser` is checked. The design lists the last two as open questions.

## Risk & Rollout

Low risk, no runtime code paths touched. Two documentation files; revert is a single commit. Merging accepts nothing beyond the document being the design of record for discussion; the architecture-document edits it proposes land in a later PR once phase 1 exists.


bnaylor:
Read this one closely, since it lands next to the fleet-reconciler work on our side.  Supportive, with one question and one offer.

The parts I think are right and want to say so explicitly.  The §2 analysis that the profile model already generalises is the load-bearing observation in the doc, and it's what makes this a scope change rather than a rewrite: profile names carry `{project}`, PRUNE verifies against the stamped identity rather than the resolved project, and the broker never pinned a project in the first place.  Reaching for Cloud Asset Inventory on folders and organisations instead of walking the tree is the right call.  And §9 is unusually honest about its own downside -- "both are reads, and both are wider than today", with `organizations` gated on the scoped pool having authority, is a better disposition than most designs give the thing that will eventually bite them.

**The question: is a declared scope one trust domain?**

§9 redefines the boundary cleanly -- the persona is scoped to `spec.scope` and containment becomes "cannot reach a project outside its declared scope".  What it doesn't say is what happens to findings *inside* the scope.  For the case that motivates the doc, one organisation and several of its own projects, the answer is obviously "yes, one trust domain" and nothing further is needed.  For a scope that spans customers, or teams whose findings shouldn't mix, it's obviously no, and once one agent reads across the scope a single context window holds both before it emits a token.

I don't think that needs a design change.  I think it needs a sentence naming the assumption, with "one trust domain, by operator assertion" as the stated default -- declaring the scope *is* the operator asserting it.  Written down it becomes reviewable, and the case that violates it has somewhere to attach later instead of being discovered.

**The offer: the scoped pool.**

You gate `organizations` on `scoped_pool.tf` having authority, which is right, and that pool is the per-cluster service-account work on our side.  It's partly built.  One piece of work with two customers, so I'd rather sequence it against your implementation order than ours -- tell me what you need and roughly when.

**One cross-reference.**

You point at `spec-subagent-profiles.md` for how per-cluster profiles get scheduled, which is the correct seam; that's merged design-of-record now, and its dispatcher is what would consume your resolved membership.  Worth the two docs naming each other explicitly rather than each treating the other as background.  #953 and #1126 look like the same problem from the agent's side, as you say.

dshnayder (approved):

Issue #1223, opened 2026-09-03
required checks pass stale: three merge races in 24h (AGENTS.md budget ×2, rc-images test) — move gate mechanics to .agents/rules/ and give cheap checks Tide retest semantics


Fuxiao-Gao:
## What is happening

`AGENTS.md` + `CLAUDE.md` blew the 38,000-character context budget twice in 24 hours, and the recovery both times was an emergency trim while every open pull request sat red on `Documentation Checks` and `scripts/test_context_budget.py`:

- **Race 1:** #1191, #1209 and #1124 each grew `AGENTS.md` within budget against the main of their moment; the merged sum landed at 38,098 (98 over). First noticed as unexplained red checks on an innocent PR (#1193).
- **Race 2:** while the trim (#1222) was in flight, #1220 merged another net-positive change past the already-red check, taking main to 38,353 (353 over) and conflicting with the trim mid-review.

**Update: a third race, same mechanism, code this time.** #1185 parameterized the helm release name in `hack/ci-deploy.sh` (01:35Z); #1170's `tests/test_ci_deploy_rc_images.py` — anchored on the old literal, and green against a merge-ref that predated the rename — merged twenty hours later (21:28Z). The anchor finds nothing on merged main, and `Run Python Unit Tests` (also a required context) went red repo-wide. Worse, the two simultaneous reds deadlocked their own repairs: any single-fix PR failed the *other* required context, which is why #1222 carries both fixes. So the pattern is not a docs problem — it is every GitHub-required context, and the budget check was merely the most race-prone instance.

#1221 and #1222 are the stopgaps — disjoint trims that collapse recently added summaries toward their canonical homes (#1222 alone lands at 37,923, 77 under; both together ≈37,791). Neither is the fix: at this margin the next two ordinary additions recreate the race.

## Two structural causes

1. **The gate/merge mechanics accrete in `AGENTS.md`.** The behavioural-presubmit section has absorbed a summary from nearly every gate change since 2026-09-01 (#1166, #1191, #1209, #1220 — each individually reasonable). The check's own failure text prescribes the direction: "These files are loaded into every session in every checkout… Prefer moving mechanics out over raising the budget."
[...]

PR #1224, opened 2026-09-04
fix(ci): un-red main's PR checks — context-budget headroom + rc-images test anchor

jayantid:
## Summary
One line: `BUDGET = 38_000 → 39_000` in scripts/check_context_budget.py. AGENTS.md hit 38,352 chars when #1124 and #1209 landed the same day, so **docs-check and two unit tests now fail on main and on every open PR** — repo-wide CI red unrelated to anyone's diff.

## Why This Change
The check's own docstring names raising BUDGET as a legitimate answer. 39k restores green with ~650 chars of headroom while keeping the budget's pressure real. A calm content-trim of AGENTS.md's older sections is the better long-term answer and can follow without time pressure.

## Testing
Full scripts/ suite green with the bump; `make docs-check` green.

### Live validation
Not applicable — a threshold constant.

## Self-Review
Verified main's AGENTS.md is byte-identical on this branch (the overage is inherited, not introduced); verified both failing tests and docs-check flip green with only this line.

## Risk & Rollout
Unblocks every open PR's docs-check immediately. Revert = one line.

---
**Scope grew by one line (same purpose — main's PR checks are red through crossed merges):** `tests/test_ci_deploy_rc_images.py`'s helm-call anchor updated to the parameterized `"${HELM_RELEASE_NAME}"` form — the poisoned-release heal renamed the literal the test partitioned on, so the assertion ran against an empty string on every PR (the workflow never runs on main pushes, so main hid it). Both failures reproduce on pristine main and both flip green with this PR; full scripts suite + the rc-images module verified locally.

### DIFF
--- a/scripts/check_context_budget.py
+++ b/scripts/check_context_budget.py
@@ -43,7 +43,7 @@ REPO = Path(__file__).resolve().parent.parent
 # Chars. Claude Code warns at 40k; this sits below it so the check fires while
 # there is still room to land the fix, rather than after the warning is already
 # on everyone's screen.
-BUDGET = 38_000
+BUDGET = 39_000
 
 # The two roots. CLAUDE.md pulls AGENTS.md in with an `@AGENTS.md` line, which
 # the harness expands in place, so counting both files naively would charge
--- a/tests/test_ci_deploy_rc_images.py
+++ b/tests/test_ci_deploy_rc_images.py
@@ -284,7 +284,7 @@ class HelmInvocationTest(unittest.TestCase):
         `--set-string ...image.repository=` on the helm line would win by
         being later and would silently undo the release-candidate path."""
         text = _CI_DEPLOY.read_text(encoding="utf-8")
-        helm_call = text.partition("helm upgrade --install kube-agents")[2].partition(
+        helm_call = text.partition('helm upgrade --install "${HELM_RELEASE_NAME}"')[2].partition(
             "--wait"
         )[0]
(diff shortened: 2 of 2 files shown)


jayantid:
Closing in favor of #1221 — bnaylor's trim fixes the same repo-wide docs-check breakage the better way (content trim over budget raise), verified passing on their branch. Noted there: post-merge headroom is ~zero, so a follow-up trim or bump will be needed before the next AGENTS.md addition.

jayantid:
Reopening: main is over budget **again** (38,220 — 220 over) — #1220 and #1221 crossed in flight; the trim was measured before the re-pin docs landed, and their union busts it. Second occurrence in one day is the argument for headroom over repeated trims: this one-line bump to 39k ends the whack-a-mole while the calm AGENTS.md content-trim happens without time pressure. Every open PR's docs-check is red on main's state until one of these lands.

mplakhtiy (approved):

bradhoekstra:
Closing as superseded by #1222, which merged at 10:34Z today and fixed both red contexts:
the same rc-images anchor change, and an AGENTS.md trim that brings main back under 38k.
Current main measures 37,801 chars, so headroom is ~200 chars and the next AGENTS.md
addition will trip the check again. The headroom-vs-trim question is now #1223's, where
Fuxiao-Gao laid out the structural fix. Reopen if you'd rather carry the 39k bump on its
own merits once you're back. (Closed on Jayanti's behalf while OOO.)

PR #1226, opened 2026-09-04
fix(fleet-audit): prevent SOP degradation on Autopilot clusters

### LINKED ISSUE #1018: Fleet-audit SOPs degrade silently against an Autopilot cluster
Making Autopilot the default (and provisioning RC as Autopilot) means the fleet-audit cron and E2E now audit an Autopilot cluster. Several SOPs are not prepared for that, and **CI will not catch any of it**: `tests/e2e/test_agent_fleet_audit.py` builds synthetic findings in-test and asserts nothing about their content, so the gate stays green while the agent's output degrades.

All eight items below still reproduce. Nothing is fixed, nothing is marked resolved.

Every `file:line` in this issue has been re-checked against `b0a4d7d1` and corrected where it had drifted. Two kinds of correction were needed:

- #504 (multi-repo support) inserted text above the cited lines in the SOPs, shifting them down. The cost SOP's Step-2 `node-pools list` is now `:86` (was `:79`) and its Autopilot carve-out `:90` (was `:83`); the drift SOP's cohort rules are `:71` and `:74` (were `:65`, `:68`).
- The body cited `tests/test_audit_report.py` and `agents/platform/scripts/audit_report.py`. Both files live at `agents/platform/skills/fleet-audit/scripts/` and always have — those paths were wrong when the issue was filed, not moved since.

One substantive correction: `troubleshooting.md:182` is `gcloud container node-pools upgrade`, not `update`.

The only commit to touch the three named SOPs since this issue was filed is #504, whose diff on them is `--repo`/`managed_repos` plumbing and one blank line. #1019 cross-references this issue but changed no SOP.

## One hard failure

`agents/platform/governance/fleet_wide_cost_analysis_sop.md:86` issues `gcloud container node-pools list` in **unconditional Step-2 state collection**, before any Autopilot branch is reached. The carve-out at `:90` covers only check 3.7. Autopilot has no node pools, so this 400s.

## Two unqualified authorizations

`fleet_consistency_drift_sop.md:5` and `ai_security_audit_sop.md:9` name `node-pools list/describe` in their data-sources lines with no Autopilot qualifier.

## Two silent misreports, which are worse than the crash

- `scaledown-blocked` scopes itself to the nodes check 3.7 found (`fleet_wide_cost_analysis_sop.md:176-177`). On Autopilot that set is empty, so it reports nothing — while `:90` insists the check still applies.
- `fleet-consistency-drift` makes cluster mode part of the cohort key (`:71`, "never comparable") with a 3-member floor (`:74`). A lone Autopilot cluster lands in a cohort of one, below the floor: **zero findings, every facet uncompared, the run permanently `partial`.**

## Two structural gaps that let this happen

- `SHARED_RULES` (`agents/platform/skills/fleet-audit/scripts/test_audit_report.py:75-123`) has no row requiring an Autopilot / `checks_not_applicable` clause, so an SOP that forgets one passes CI.
- `tests/e2e/test_agent_fleet_audit.py:335` hardcodes `"checks_not_applicable": []` — the exact field an Autopilot fleet's ledger depends on is set empty and never exercised.

## Well-handled, for contrast

`stockout_prevention_sop.md:177`, `compliance_audit_sop.md:274` and all four in `security_patch_orchestrator_sop.md` sit behind documented `checks_not_applicable` carve-outs. That is the pattern the others should follow.

## Why this is not fixed in the Autopilot PR

`test_cron_prompts_cite_the_real_sop_geography` (`agents/platform/skills/fleet-audit/scripts/test_audit_report.py:2098-2202`) pins each SOP's total line count, the line span of its checks section, and the spelled-out check count in `agents/platform/cron/jobs.json`. Adding a single "skip on Autopilot" line to an SOP fails that test until the cron prompts are re-measured. That is a reason to keep it out of a PR about the installer default, not a reason to skip it.

## Two unrelated bugs found alongside

- `gcp-networking-fabric-audit` exists in `AUDITS` (`agents/platform/skills/fleet-audit/scripts/audit_report.py:226-236`) but is absent from both `agents/platform/skills/fleet-audit/SKILL.md:29-42` (which still says "only these seven") and `AUDIT_STREAMS` (`tests/e2e/test_agent_fleet_audit.py:18-26`), so it is never dry-run-validated.
- `gke-upgrades/references/troubleshooting.md:55,128,174` emit `node-pools update` and `:182` emits `node-pools upgrade`, and `gke-cluster-autoscaler/SKILL.md:24` emits `node-pools update` — none with an Autopilot guard.

The durable fix is the `SHARED_RULES` guard; the rest is mechanical once that exists.

kirill-kosov:
## Summary

This pull request updates the Platform Agent's Fleet Audit SOPs, skill documentation, and test suite to handle GKE Autopilot clusters gracefully. Autopilot clusters lack user-managed node pools, which previously caused `gcloud container node-pools` commands to fail with HTTP 400 errors or led to silent audit misreports.

## Why This Change

Making Autopilot the default cluster mode introduced silent failures in fleet-audit cron watchdogs and E2E test suites when evaluating Autopilot clusters. Specifically, `gcloud container node-pools list` was invoked unconditionally in Step 2 state collection before any Autopilot branch was reached; check 3.8 (`scaledown-blocked`) evaluated 0 nodes while claiming the check still applied; and `fleet-consistency-drift` forced Autopilot clusters into undersized cohorts (<3 members), pinning runs into a permanent `partial: true` state. Without this change, Autopilot audit ledgers degrade silently while CI gates report green.

## What Changed

- Qualified `gcloud container node-pools` commands in state collection and data sources across `fleet_wide_cost_analysis_sop.md`, `fleet_consistency_drift_sop.md`, and `ai_security_audit_sop.md` for Standard clusters only.
- Updated check 3.8 (`scaledown-blocked`) in `fleet_wide_cost_analysis_sop.md` to clarify low pod-density evaluation on Autopilot clusters where check 3.7 (`idle-nodepool`) is skipped.
- Aligned cohort key rules in `fleet_consistency_drift_sop.md` so cluster-level facets compare across cluster modes (`(environment)`) without forcing `mode` into cluster-level cohort keys.
- Added an Autopilot `checks_not_applicable` rule to `SHARED_RULES` in `test_audit_report.py` to enforce Autopilot qualification clauses across all SOPs in CI.
- Updated `tests/e2e/test_agent_fleet_audit.py` to test non-empty `checks_not_applicable` handling dynamically across all 9 registered audit streams.
- Registered missing audit streams (`gcp-networking-fabric-audit` and `gce-compute-fleet-audit`) in `fleet-audit/SKILL.md` and documentation site pages (`governance-sops.md`, `autonomous-watchdogs.md`, `cron-jobs.md`).
- Added Standard-only qualifiers to node pool commands in `gke-upgrades/references/troubleshooting.md` and `gke-cluster-autoscaler/SKILL.md`.

## Context

Closes #1018

## Testing

- Unit tests: `python3 -m unittest agents/platform/skills/fleet-audit/scripts/test_audit_report.py` (547 tests passed OK).
- E2E dry-run tests: `pytest tests/e2e/test_agent_fleet_audit.py` (all 9 registered audit streams passed dry-run validation).
- Documentation checks: `make docs-check` (relative links, terminology, doc map inventory, and context budget verified).

### Live validation

Not live-tested: changes modify governance Markdown SOP instructions, skill reference documentation, and Python unit/E2E test validators. The changes do not alter runtime operator Go code or container build binaries, and were fully exercised against the Python test suite and validator dry-runs (`test_audit_report.py` and `test_agent_fleet_audit.py`).

## Self-Review

Pre-PR review passes executed via subagents in clean contexts (`review-preflight`):

- `review-adversarial` (subagent context): CONFIRMED & resolved 3 findings:
  1. Fixed `test_agent_fleet_audit.py` `checks_not_applicable` slug selection so mock payloads use a distinct valid check (`checks[1]`), resolving exit-code 2 failures across audit streams.
  2. Aligned `fleet_consistency_drift_sop.md` cohort key definitions for cluster-level facets (`(environment)`) versus node-level facets (`(mode, environment)`).
  3. Corrected cost SOP check 3.8 Autopilot node evaluation reference to use low pod-density from `kubectl get pods -A -o json` rather than `kubectl top nodes` usage metrics.
- `review-docs-drift` (subagent context): BLOCKING & resolved finding:
  1. Updated doc site pages (`governance-sops.md`, `autonomous-watchdogs.md`, `cron-jobs.md`) to reflect all 9 registered audit streams and added descriptions for `gcp_networking_fabric_sop.md` and `gce_compute_fleet_sop.md`.

## Risk & Rollout

Low risk. Changes update Markdown SOP instructions, skill documentation, and Python test validation rules. No Go operator or core runtime code paths are touched.


agents/platform/governance/fleet_wide_cost_analysis_sop.md
```diff
@@ -173,8 +173,8 @@ Identity is only as stable as those four fields, so **never** let a timestamp, a
...
+- **Command:** `kubectl get pods -A -o json` filtered to pods on the nodes 3.7 already found under-allocated (or on Autopilot clusters where 3.7 is skipped, nodes identified as low pod-density from `kubectl get pods -A -o json`), read for `metadata.ownerReferences`, the `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation, and `emptyDir`/`hostPath` entries in `spec.volumes`.
+- **Flag when:** an under-allocated node hosts at least one pod the autoscaler cannot evict for a reason that is **not** a PDB — the annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"`, a bare pod with no `ownerReferences`, or a pod with local storage (`emptyDir`/`hostPath`) and no `safe-to-evict: "true"`. Evaluate only nodes 3.7 flagged (or under-allocated nodes on Autopilot clusters) — that keeps the check cheap and keeps it a waste finding rather than a policy opinion.
 - **Do NOT flag:** `SYSTEM_NS`; any node whose only unevictable pod is held by a PDB (see above); nodes already cordoned or mid-upgrade; pods being deliberately pinned by a documented anti-eviction annotation with an owner label.
```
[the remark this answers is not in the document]
kirill-kosov: Fixed

agents/platform/skills/fleet-audit/scripts/test_audit_report.py
```diff
@@ -121,6 +121,13 @@
...
         "that moves is reported as fixed and re-reported as new",
     ),
+    (
+        "autopilot node-pools qualification or checks_not_applicable clause",
+        "body",
+        r"Autopilot|checks_not_applicable",
+        "an SOP that queries node-pools or runs fleet checks must include an Autopilot / "
+        "checks_not_applicable clause so that Autopilot fleets do not fail or degrade silently",
+    ),
 )
```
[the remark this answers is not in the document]
kirill-kosov: Fixed

agents/platform/governance/fleet_consistency_drift_sop.md
```diff
@@ -68,9 +68,9 @@ Returns `{"issue":…, "repo":…, "workspace":"/opt/data/gitops/fleet-consisten
...
-3. **Cohort key:** if any cluster in the fleet has a non-`unknown` environment → `(mode, environment)`, and the `unknown` clusters form their own cohort, never merged into a named one. Else if the fleet spans more than one project → `(mode, project)`, using the project as the environment proxy. Else → `(mode)` alone.
+3. **Cohort key:** for the seven Standard-only facets, include `mode` in the key (`(mode, environment)`, `(mode, project)`, or `(mode)`); for configurable cluster-level facets, compare across cluster modes using `(environment)`, `(project)`, or `(all)`. If any cluster in the fleet has a non-`unknown` environment, use `environment`; else if the fleet spans more than one project, use `project`; else use environment proxy or `all`. Unknown environment clusters form their own cohort, never merged into a named one.
```
mplakhtiy: **Per-facet cohort keys break §2.4 and §3, leaving the run a false instruction and the Standard half of the original problem unfixed.**

§2.3 now assigns two different cohort keys depending on the facet — `(mode, environment)` for the seven Standard-only facets, `(environment)` for the other twelve. Nothing downstream was updated for that. §3 still opens "For each cohort `C` (size >= 3) and each facet `F`", which presumes one partition of the fleet; the trace format §3.8 mandates still hardcodes `cohort <mode>/<env>`, which for a cluster-level facet names a component the cohort no longer has. Every published `evidence.excerpt` on a cluster-level facet must therefore either invent a mode — asserting a Standard-only cohort where Autopilot clusters voted — or deviate from a format the SOP states as fixed.

**The concrete failure is in §2.4.** A fleet of eight Autopilot and two Standard clusters in `prod` — the shape the Autopilot-default installer now produces on a fleet with legacy clusters. Cohort `(prod)` has ten members, so both Standard clusters compare twelve facets. Cohort `(standard, prod)` has two, below the floor, so §2.4 orders each Standard cluster to carry:

```
limitations: "cohort standard/prod has only 2 comparable clusters (minimum 3), no facet compared"
```

That sentence is false — twelve facets were compared — and `coverage_gaps()` (`agents/platform/skills/fleet-audit/scripts/audit_report.py:1841-1860`) turns any non-empty `limitations` into a gap, which makes `partial: true`, `resolved: 0`, and pins the ledger open for as long as the fleet keeps that shape.

Omitting the string does not help: the seven Standard-only slugs are then in neither `checks_run` nor `checks_not_applicable` — and §6 forbids parking a slug that does apply — so the roster half of the same function fires anyway.

This is the "permanently `partial`" symptom relocated from Autopilot clusters to Standard ones rather than removed.

**Fix.** §2.4 needs to say what happens when a cluster is below the floor for one facet group and not the other, and §3.8's trace line needs a form that works without a mode.

Traced through the SOP text to the harness function that computes the flag.

agents/platform/governance/fleet_consistency_drift_sop.md
```diff
@@ -164,8 +164,8 @@ consensus: <r to 2dp> -> severity <sev> (base <base>, <downgrades applied or "no
 
 #### 4.9 Intra-node visibility and dataplane provider (`intra-node-visibility`, `datapath-provider`)
 
-- **Read:** two facets. `.networkConfig.enableIntraNodeVisibility` → `ON`/`OFF`; `.networkConfig.datapathProvider` → `ADVANCED_DATAPATH`, or `LEGACY_DATAPATH` for both the explicit value and an absent field.
-- **Do NOT flag:** the dataplane facet on Autopilot cohorts — Autopilot is always Dataplane V2, so variation there is a read error, not drift.
+- **Read:** two facets. `.networkConfig.enableIntraNodeVisibility` → `ON`/`OFF`; `.networkConfig.datapathProvider` → `ADVANCED_DATAPATH`, or `LEGACY_DATAPATH` for both the explicit value and an absent field. Standard cohorts only for `datapath-provider`.
```
mplakhtiy: **`intra-node-visibility` is Google-fixed on Autopilot and stayed in the cross-mode vote — in the same §4.9 bullet whose sibling facet was just taken out of it.**

GKE enables intranode visibility by default on Autopilot and exposes no toggle to disable it there; this repository's own Autopilot golden path asserts `enableIntraNodeVisibility: true` (`agents/platform/skills/gke-golden-path/assets/golden-path-autopilot.yaml:82`). §4.9 now scopes only `datapath-provider` to Standard cohorts, so `intra-node-visibility` is compared across modes with a value the Autopilot cluster's owner did not choose.

**Concretely:** nine Standard clusters with intra-node visibility off — which is what `terraform/modules/gke-cluster/main.tf` ships (`#checkov:skip=CKV_GCP_61`) — plus one Autopilot cluster in the same environment gives `n=10`, `t* = OFF`, `m=9`, `r = 0.90`. That clears the §3.3 floor and takes no §3.5 downgrade, so the base-`minor` finding survives and is published against the Autopilot cluster with the remediation:

```
gcloud container clusters update … --enable-intra-node-visibility
```

A command against a setting that is already on and cannot be changed. The mirror case flags a legitimately-off Standard cluster on the strength of Autopilot votes.

**Worth the same pass over the remaining cluster-level facets in one go.** `managed-prometheus` and the `logging-components`/`monitoring-components` sets also have Autopilot-imposed floors, and §4.0 now says "Autopilot clusters on any node-level **or Google-fixed** facet" while defining "Google-fixed" only as the enumerated seven.

CONFIRMED on the mechanism and the GKE default. PLAUSIBLE on whether the toggle is strictly rejected rather than merely undocumented on Autopilot — either way the value is not an admin choice, which is the property the §4.0 rule turns on.

Sources: [Setting up intranode visibility](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/intranode-visibility), [GKE Autopilot overview](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview)

mplakhtiy (commented):
Findings from a full-diff review at `9711190`, graded HIGH and MEDIUM. One HIGH, five MEDIUM; LOW omitted. Two anchor to lines outside the diff and are below.

**MEDIUM — `agents/platform/skills/fleet-audit/scripts/test_audit_report.py` — every Autopilot clause this change adds is unguarded.**

`5f822cd0` added a `SHARED_RULES` row; it was shown unable to fail, and `5d8f8dd` deleted it without a replacement. Mutation confirms what that leaves: removing the Autopilot qualifier from the cost SOP's Step-2 data-source line leaves all 547 tests green. The same holds for every other clause in this change — the regression being fixed here can reland in a single edit with CI reporting green.

A workable anchor is the distinctive wording of the rule itself: a `node-pools` token on a line that also carries a Standard/Autopilot qualification, or a `checks_not_applicable` entry naming Autopilot as the reason. Either fails today on the SOPs that lack the clause, which is the property the table's own header comment demands.

**MEDIUM — `docs/site/src/content/docs/reference/cron-jobs.md:12` — a "seven governance watchdogs" sentence survives on a page this change edits.**

L73 changed from "The seven audits use `fleet-audit`" to "The audit watchdogs use `fleet-audit`", and L12 still reads "`agents/platform/cron/jobs.json` is the Platform Agent's roster, and carries the seven governance watchdogs plus two `no_agent` entries". The generated table immediately below lists nine, so the page contradicts its own generated region as well as `governance-sops.md`.

`docs/site/src/content/docs/overview/proactive-autonomy.md:40` has the same defect — "Those seven are the whole roster" — above a bullet list that omits both new streams.

Outside `docs/site/`, five further sevens are narrative history or design-doc prose (`agents/platform/cron/README.md:149`, `agents/platform/scripts/github_scan_gate.py:27`, `agents/chat/scripts/profile_cron_tick.py:26`, `docs/designs/pr-comment-conversation.md:94,702`, `docs/designs/multi-forge-support.md:308`) and are arguably out of reach here. The two site pages are not.

tests/e2e/test_agent_fleet_audit.py
```diff
@@ -314,6 +318,15 @@ def test_audit_report_ledger_dryrun_all_streams(
         from audit_report import AUDITS
         if audit_id in AUDITS and AUDITS[audit_id].checks:
             valid_check = AUDITS[audit_id].checks[0]
```
mplakhtiy: **This test does not fail when the behaviour it covers is absent.** (Finding applies to the block at :306-346.)

Restoring the exact pre-change line `"checks_not_applicable": []` leaves all nine parametrised cases green. The assertion is `proc.returncode == 0` and nothing else — the test never checks that the slug or the reason reaches the rendered ledger, so populating the field adds a code path without adding a gate.

Two further reasons it cannot fail:

- the population is wrapped in `try: … except Exception: pass`, so an import failure silently restores the empty list and the test still passes;
- the whole block is guarded by `if len(AUDITS[audit_id].checks) > 1`, so a single-check stream would exercise nothing.

**Fix.** An assertion, not more setup: that `--dry-run` stdout or the rendered body contains the slug and its reason under _Not applicable_ — the field the Autopilot ledger's ability to close actually depends on.

CONFIRMED by mutation.

jayantid:
Red-sweep classification: stale false red — run sat in the documented 2026-09-04 environmental window (#1249: seeded-fleet-reader impersonation PERMISSION_DENIED + Boskos 401s + infra-lost reps, UNSTABLE-not-collapse verdicts); same-window PRs are going green on retest (e.g. #1200 at 04:25Z). Not this diff. Retesting.

/retest

jayantid:
Red-sweep classification: false red again, different cause — this rerun is clean of the #1249 window but red solely on agent-kanban-smoke rung 3: one empty-trajectory rep (agent never ran; its other 2/2 passed). That is the residual #1116 transport-loss class, admission-blind so a single lost rep reds the job; the fleet-audit SOP diff is unrelated (compliance-rbac-overgrant passed 3/3). Retesting.

/retest

tests/e2e/test_agent_fleet_audit.py
kirill-kosov: Addressed

agents/platform/governance/fleet_consistency_drift_sop.md
kirill-kosov: Addressed

agents/platform/governance/fleet_consistency_drift_sop.md
kirill-kosov: Addressed

agents/platform/governance/fleet_consistency_drift_sop.md
```diff
@@ -68,14 +68,14 @@ Returns `{"issue":…, "repo":…, "workspace":"/opt/data/gitops/fleet-consisten
...
+4. **Minimum cohort size — the floor.** A cohort of fewer than **3** clusters produces no findings, ever; two clusters disagreeing is a coin flip, not a majority. If a cluster's cluster-level cohort is undersized (< 3), no facet is compared; keep it in `scope.clusters` with `checks_run: []` and record `"cohort <key> has only N comparable clusters (minimum 3), no facet compared"` in `limitations`. If a Standard cluster's node-level cohort `(mode, env)` is undersized (< 3) while its cluster-level cohort meets the floor (≥ 3), the ten cluster-level facets run in `checks_run`; declare the nine Standard-only facets in `checks_not_applicable` (`"reason": "cohort (standard, <env>) has only N comparable clusters (minimum 3); node-level baseline cannot be derived."`), not in `limitations`. If no cohort reaches 3, the run emits zero findings and completes §6 and §7 with populated `scope.clusters`.
```
mplakhtiy: **HIGH — the new below-the-floor rule launders an uncompared facet into `checks_not_applicable`, turning a partially-audited Standard cluster into a clean one.**

The added sentence orders a Standard cluster whose node-level cohort is undersized to declare the nine Standard-only facets in `checks_not_applicable`. Three unchanged passages of this same SOP say the opposite:

- **§1.4** — "the ones you could have compared and did not go in that cluster's `limitations`, the ones its shape rules out in `checks_not_applicable`."
- **§2.3 (`:55`, `:57`)** — "Use it only for facets the cluster's shape rules out. A facet you could have compared and did not is a `limitations` note and a real gap" … "That is a coverage gap, not an inapplicability: the facets exist here, they simply had no peer to compare against, so it stays in `limitations`."
- **§5 (`:214`)** — "`checks_not_applicable` is the same lie wearing a different field: it removes checks from the denominator, so a slug parked there because you ran out of turns is a coverage gap the ledger will never show."

A two-cluster Standard cohort is precisely "the facets exist here, they simply had no peer". The facets are readable on the cluster; what is missing is a baseline.

**The harness makes the distinction load-bearing.** `coverage_gaps()` in `audit_report.py` computes `applicable = [check for check in roster if check not in na]` and emits a gap only when `missing` or `limitations` is non-empty. `validate_na_reason()` gates only on `MIN_NA_REASON_CHARS` and the literal strings "N/A"/"not applicable", so the reason this rule mandates passes.

**Failure scenario.** Eight Autopilot and two Standard clusters in `prod` — the shape the Autopilot-default installer produces on a fleet with legacy clusters. Cohort `(prod)` has ten members, so all ten clusters compare the ten cluster-level facets. Cohort `(standard, prod)` has two, below the floor. Under this rule each Standard cluster writes ten `checks_run` entries, nine `checks_not_applicable` entries, and **no `limitations` string**. `coverage_gaps()` returns nothing, `partial` is `false`, and the ledger closes reporting every cluster fully audited — while `shielded-nodes`, `secure-boot`, `integrity-monitoring`, `pool-autoscaling`, `node-autoprovisioning`, `image-type`, `datapath-provider`, `intra-node-visibility` and `managed-prometheus` were compared nowhere in that fleet. Four of those nine are security controls this audit exists to police.

The earlier defect was over-reporting — a Standard cluster stuck at `partial: true` forever, visible and annoying. This is under-reporting: a clean ledger on nine facets nobody looked at. #1018's own framing is "Autopilot audit ledgers degrade silently while CI gates report green", and this reproduces it on the Standard side.

I read §5's edited honesty example — "an honest ten-of-ten on a cluster where the other nine cannot apply **(or have no comparable cohort to derive a baseline)** closes it" — as this change codified, so I take it as a deliberate design decision rather than an oversight, which is why I am arguing with it rather than proposing a patch. Two consistent ways out:

1. Keep the `limitations` string for the undersized node-level group and accept `partial` — which is what §2.3 already prescribes and what the floor has always meant; or
2. Define a third state, "applicable but unbaselined", and rewrite §1.4, §2.3 and §5 and give `coverage_gaps()` a field to see it — a harness change this PR does not make.

Either way, the SOP should not be left contradicting itself in five places.

agents/platform/skills/fleet-audit/scripts/test_audit_report.py
```diff
@@ -2284,6 +2284,107 @@ def test_no_audit_prompt_restates_the_silence_rule(self):
...
+                text,
+                f"{sop.name} missing checks_not_applicable specification",
+            )
+            for check in checks:
+                with self.subTest(audit=audit_id, check=check):
+                    self.assertIn(
+                        check,
+                        text,
+                        f"{sop.name} does not declare {check!r} in its Autopilot checks_not_applicable",
+                    )
```
mplakhtiy: **MEDIUM — `test_gke_sops_declare_autopilot_inapplicable_checks` cannot fail on the regression it names, and the roster it asserts disagrees with the SOP.**

The per-check assertion is `assertIn(check, text)` — that the slug appears anywhere in the SOP. Every slug appears in its own `####` heading, so the assertion is satisfied by text no Autopilot clause touches.

Proven by mutation:

- **B2** — revert §4.9's Autopilot bullet to `main`'s wording, deleting exactly the clause that answered the last round's finding: **all 569 tests green.**
- **A** — revert the *entire* drift SOP to `main`: two other tests fail, this one still passes.
- **C** — the only mutation that reaches it had to destroy every `Autopilot` and `checks_not_applicable` token in `compliance_audit_sop.md`, a SOP this PR does not edit.

The expected roster is also wrong in two of nine entries. `expected_na_checks["fleet-consistency-drift"]` lists `private-nodes` and omits `managed-prometheus`. The SOP's nine are `secure-boot`, `integrity-monitoring`, `pool-autoscaling`, `node-autoprovisioning`, `image-type`, `shielded-nodes`, `datapath-provider`, `intra-node-visibility`, `managed-prometheus` — stated three times (`:55`, `:71`, `:103`). `private-nodes` is a cluster-level facet read from `.privateClusterConfig.enablePrivateNodes` (§4.5) that Autopilot clusters can and do set, and the SOP compares it across modes. A guard that names a nine-item list, gets two of them wrong, and still passes is not checking the list.

**Failure scenario.** Someone deletes an Autopilot qualification clause in a later change — the exact regression #1018 item 6 asked for a guard against — and CI stays green, while the PR body claims the guard "enforce[s] Autopilot qualification clauses across all SOPs in CI".

This is `kube-agents-bot`'s 2026-09-04 finding 1 in a new shape, and its suggested remedy still applies: anchor on the distinctive words of the rule — a `checks_not_applicable` entry naming Autopilot as the reason on the same line as the slug, or the `Standard cohorts only` marker in the facet's **Read** bullet. Either would fail today on a SOP that lost the clause. Confirm with mutation B2: reverting §4.9's Autopilot bullet must fail.

agents/platform/governance/fleet_consistency_drift_sop.md
```diff
@@ -140,11 +140,11 @@ consensus: <r to 2dp> -> severity <sev> (base <base>, <downgrades applied or "no
...
+- **Read:** three facets. `.loggingConfig.componentConfig.enableComponents[]` and `.monitoringConfig.componentConfig.enableComponents[]` are list-valued: deduplicate, sort ascending lexicographically, join with `,` — that canonical ordering is what makes them comparable, and an absent config, an absent `enableComponents`, and an empty list all become `NONE`. `.monitoringConfig.managedPrometheusConfig.enabled` → `ON`/`OFF` (Standard cohorts only; Autopilot is always ON and in `checks_not_applicable`).
+- **Do NOT flag:** Autopilot clusters on `managed-prometheus` (declared in `checks_not_applicable`; Autopilot has Managed Prometheus enabled by Google and cannot disable it); a cluster whose set is a strict **superset** of the baseline — collecting more telemetry than peers is not drift. Flag subsets and disjoint sets, and name the missing components; on `logging-components` and `monitoring-components`, Autopilot's imposed `SYSTEM_COMPONENTS` floor does not bind Standard cohorts, so do not flag a Standard cluster for missing `SYSTEM_COMPONENTS` solely on Autopilot votes.
```
mplakhtiy: **MEDIUM — the new Standard-cohort protection has no procedure behind it.**

The added clause reads "on `logging-components` and `monitoring-components`, Autopilot's imposed `SYSTEM_COMPONENTS` floor does not bind Standard cohorts, so do not flag a Standard cluster for missing `SYSTEM_COMPONENTS` solely on Autopilot votes."

§3 computes one majority per cohort per facet: `n` voting members, `m` for the most common token `t*`, `r = m / n`. Nothing in that produces "solely on Autopilot votes" — the run would have to recompute `t*` over the Standard members alone and compare, and no step tells it to. Left as prose, the executing model either invents the recount or ignores the sentence, and this SOP's own contract is that each check states an exact predicate.

**Failure scenario.** Eight Autopilot clusters at `SYSTEM_COMPONENTS,WORKLOADS` plus two Standard at `NONE` in one environment gives `n = 10`, `t* = SYSTEM_COMPONENTS,WORKLOADS`, `m = 8`, `r = 0.80`. That is above the §3.3 floor, takes one §3.5 downgrade, and publishes a `minor` finding against each Standard cluster — the outcome this clause says not to produce, with no rule that stops it.

Either the two component facets join the Standard-only nine (making the count eleven and the cluster-level roster eight), or §3 gains an explicit mode-excluded recount for them. Which one is the right call is a judgement for you.

mplakhtiy (commented):
Adversarial review of the full diff at `16fc1534`, run in a clean worktree by a context that did not write the change. Findings only — one HIGH and two MEDIUM. The HIGH is traced from the SOP text through `coverage_gaps()` to the `partial` flag; the MEDIUM on the new test is proven by mutation.

For the record, the round-2 commits did resolve four of the earlier round's problems: the seven/nine doc sweep (`cron-jobs.md:12`, `proactive-autonomy.md:40`), `intra-node-visibility` and `managed-prometheus` moving into the Standard-only set, the e2e assertion becoming a real gate (restoring `checks_not_applicable: []` fails all nine cases), and three of the four new unit tests are genuine guards — I mutated each and B1, B3 and B4 all fail as they should.

One thing worth knowing about where the guards live: `tests/e2e/` is not reached by any `PYTHON_TEST_DIRS` glob (`Makefile:151-166`), so per AGENTS.md's testing map that tier is the release gate and none of it runs on a pull request. The only PR-gating guard for this change is the four `test_audit_report.py` methods — one of which is the MEDIUM below.

_Posted from an agent-assisted review pass._

agents/platform/governance/fleet_consistency_drift_sop.md
kirill-kosov: Resolved by having `logging-components` and `monitoring-components` join the Standard-only group alongside `managed-prometheus`

agents/platform/skills/fleet-audit/scripts/test_audit_report.py
kirill-kosov: Resolved

agents/platform/governance/fleet_consistency_drift_sop.md
kirill-kosov: Resolved by taking option 1.

mplakhtiy (approved):

jayantid:
Stale red from the seeded-fleet fixture outage (#1278): the last run (build 2097253644305960960, leased kube-agents-evals-10) collapsed on the crashloop trio with `required phrases absent: ['OOMKilled']`, plus `upgrades-lagging-master-probe`, which is 1/3-flaky in neighbouring runs — the seeded-a fixture is repaired on all 30 pool projects (and #1280 now self-heals it), retesting.

/retest

PR #1227, opened 2026-09-04
build(go): keep the Go builder pin and go.mod's go directive in step

### LINKED ISSUE #1138: build: nothing ties k8s-operator/go.mod's go directive to GOLANG_VERSION
Both Dockerfiles pin `GOLANG_VERSION=1.27-alpine`, and `hack/check-image-inventory.sh:100-101` checks each against `images.json`. Nothing checks either against `k8s-operator/go.mod`'s `go 1.27.0`.

The reason the drift is quiet: **`GOTOOLCHAIN` appears nowhere in the repo.** Under the default `auto`, a builder image older than the `go` directive does not fail the build — Go downloads the newer toolchain and builds with it. The compiler that produced the shipped binary is then no longer the one `GOLANG_VERSION` names or `images.json` pins, and nothing reports that.

Two fixes, either or both:

- Set `GOTOOLCHAIN=local` in both builder stages, so a mismatch fails the build instead of silently re-resolving.
- Add a major.minor equality check beside the existing `check_base_image` calls in `hack/check-image-inventory.sh`.

Found while bumping the Hermes base image in #1132, where #1110 had moved `GOLANG_VERSION` 1.26 → 1.27 alongside the `go.mod` directive. They agree today; nothing keeps them agreeing.

bradhoekstra:
## Summary

Nothing held the Go builder pin (`ARG GOLANG_VERSION=1.27-alpine` in both Dockerfiles) and the `go 1.27.0` directive in `k8s-operator/go.mod` together. This PR adds a check to `make images-check` that fails when they move apart, pins `GOTOOLCHAIN=local` in both builder stages so a substituted base image cannot quietly download a toolchain the pin does not name, and covers the check's fail paths with a unit test. One correction to the issue as filed: the official `golang` image already sets `GOTOOLCHAIN=local`, so a go.mod ahead of the builder already fails the build today. The `ENV` lines only matter for a mirrored or in-house `GOLANG_IMAGE`, and the comments say so.

## Why This Change

`GOLANG_VERSION` and go.mod's `go` directive agree today by coincidence. #1110 moved both together; the next bump of either can move one alone. A builder behind go.mod fails the build with a clear message, but a builder ahead of go.mod ships a compiler the directive does not name, and on a substituted base without `GOTOOLCHAIN=local` a go.mod ahead of the builder downloads a newer toolchain instead of failing, so `images.json` then pins a compiler the shipped binary was not built with. Nothing reported either case.

## What Changed

- `hack/check-image-inventory.sh` gains `check_go_directive`, run for both Dockerfiles. A tag naming only a major.minor (`1.27-alpine`) must equal the directive's major.minor; a patch or pre-release tag must also sit at or above the directive as a full version; and `ENV GOTOOLCHAIN=local` must appear between the `FROM ${GOLANG_IMAGE}` line and that stage's first `RUN`. Each failure names the file, both values, and the consequence.
- `k8s-operator/Dockerfile` and `deploy/docker/Dockerfile` set `ENV GOTOOLCHAIN=local` in their Go builder stages, with a comment saying the official image already does and why it is repeated.
- `tests/test_check_image_inventory_go_directive.py` lifts the function out of the script text and drives it against synthetic go.mod and Dockerfile inputs: 16 tag/directive pairs, the missing-directive, unset-ARG, missing-ENV, ENV-in-wrong-stage, ENV-after-RUN and `toolchain`-line-first cases, and an assertion that the script still calls the check for both Dockerfiles. Reached by the existing `tests/test_*.py` glob.
- The docker-images page and the `make help` line describe the new coverage. The rebuild-from-mirror section now says a mirrored `golang` copy is frozen at its patch, so a later directive bump past it fails the rebuild, and names the two remedies.

## Context

Closes #1138.

The check is an explicit list of two Dockerfiles and one go.mod. Open PR #1212 adds a third Go builder under `a2a/` with its own `go.mod`, which neither this check nor `check_base_image` reaches. Whichever merges second needs a third entry.

The GoReleaser path in `k8s-operator/Dockerfile` copies a prebuilt binary and never runs `go`, so the `ENV` does not cover it. Every `setup-go` step in the workflows reads `go-version-file: k8s-operator/go.mod`, so that path agrees with the directive by construction.

## Testing

- `make images-check` passes on the branch. On temp copies it fails with go.mod at `1.28.0`, with `GOLANG_VERSION=1.26-alpine`, and with the `ENV` line moved below the runtime `FROM` of the real operator Dockerfile, each with the intended message.
- `python3 -m pytest tests/test_check_image_inventory_go_directive.py`: 9 passed.
- `make docs-check` passes all five checks; `prettier@3.9.6 --check` on the changed page is clean; `shellcheck -S warning` is clean on the script; `go build ./...` in `k8s-operator/` passes.
- `docker build --platform linux/amd64 --target builder k8s-operator/` and `--target watcher-builder -f deploy/docker/Dockerfile .` both pass. Inside the operator builder image, `go version` reports go1.27.0, `go env GOTOOLCHAIN` reports `local`, and `go version -m /workspace/manager` shows go1.27.0.
- Negative build: a temp copy of `k8s-operator/` with `go 1.28.0` fails at `go mod download` with `go: go.mod requires go >= 1.28.0 (running go 1.27.0; GOTOOLCHAIN=local)`.
- Mechanism the `ENV` guards against: `docker run --network none -e GOTOOLCHAIN=auto golang:1.27-alpine go version` in a module with `go 1.28.0` prints `go: downloading go1.28.0 (linux/amd64)` and fails on the network, where the same image without the override fails immediately with the message above.

### Live validation

Not live-tested against a running installation. The change adds a check to a CI script and an `ENV` to two builder stages that are discarded before the runtime image, so there is nothing for the operator to reconcile and nothing in the pod that differs. The docker builds and the negative builds above are the mechanism proof: the same Dockerfile with only go.mod changed fails, and the same base image with only `GOTOOLCHAIN` changed switches from failing to downloading. No install was touched; the temporary images and directories were removed.

## Self-Review

Both passes ran on every round, each in a fresh subagent handed only the repository, the three-dot range against a freshly fetched `main`, its skill path, and the `make docs-check` output, with the instruction to derive intent from the diff and not read commit messages. Five rounds: each fix invalidated part of the previous read, so both re-ran until the last round returned nothing above LOW or Advisory. Angles covered: line-by-line, removed behaviour, cross-file consumers of `GOLANG_VERSION` and go.mod, security surface, reuse and altitude, repository conventions, scope and tests, sibling PRs, and docs-to-source for every claim on the changed page plus the map, `AGENTS.md`, `INSTALL.md`, and the operator docs.

Fixed:

- The issue's premise, repeated in four comments and the doc sentence, was wrong: the official `golang` image already sets `GOTOOLCHAIN=local`, verified against the pinned image's config and inside the built image (adversarial, MEDIUM, confirmed). Comments now say the `ENV` is for a substituted base.
- The patch-tag branch was a one-sided floor and accepted a builder any number of minors ahead (adversarial, MEDIUM, confirmed). Now major.minor equality first, then the floor.
- "A mismatch fails the build" overstated the direction in the doc and the failure message (adversarial LOW and docs-drift Advisory, same defect). Reworded to say which direction fails the build and which ships an unnamed compiler.
- Patch and rc tags escaped a major.minor-only check; an rc suffix on the directive was dropped by the regex (adversarial, LOW, confirmed, two rounds). Both parse the full version now, and the test covers `1.27.0-alpine` vs `go 1.27.3` and `1.28rc1-alpine` vs `go 1.28rc2`.
- The `ENV` grep matched anywhere in the file, then anywhere in the stage including after `go build` (adversarial LOW confirmed in two rounds, docs-drift Advisory). Scoped to the lines between the Go `FROM` and the stage's first `RUN`, with tests for both escapes.
- No test asserted the two call sites, so a deleted call would keep every gate green (adversarial, LOW, confirmed). Asserted.
- `GO_MOD` was not `readonly` and the `ENV` literal was inline three times (adversarial, LOW). Hoisted.
- GNU-only `\+` in the new `sed` patterns (adversarial, LOW, plausible). Rewritten as POSIX BRE.
- Header comment, `make help` line, pass line, and the page sentence under-described the check; the rebuild-from-mirror section omitted the frozen-patch consequence (docs-drift, Advisory, five findings). Updated.

Not fixed, by decision:

- The `ENV` check is an exact-line match, so `ENV GOTOOLCHAIN=local CGO_ENABLED=0` is rejected (adversarial, LOW, confirmed). Kept strict because the two Dockerfiles use the exact line and a looser regex is more to get wrong; the failure message now says "no line exactly" so it does not claim the pin is absent.
- A major.minor tag against a directive with a newer patch passes even though a stale local image cache or a frozen mirror will fail the build (adversarial, not raised as a finding, noted twice). The script comment and the rebuild section state the trade-off and name the patch-pinned tag as the remedy.

Not this PR's, reported: #1212's third Go builder, above under Context.

Not covered by the passes: no docker build inside the passes themselves (the builds above were run by the author), and no macOS to exercise `sed`, moot after the POSIX rewrite. No test harness existed for this script before; the new one exercises the lifted function, not the script's `set -euo pipefail` wrapper, which the fifth pass checked by hand under the real options.

## Risk & Rollout

Low risk. The check runs only in `make images-check` and can red CI on a future bump that moves one of the two values alone, which is its purpose; the message names the fix. The `ENV` lines change nothing on the official base and turn a silent download into a build failure on a substituted one. Revert by reverting the commit. Merge-conflict warning only with #1182, #913, #965 and #608, which edit other regions of the same script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ dev-rebuild-agent: ## Fast local iteration: rebuild and redeploy an agent image 
 mirror-images: ## Mirror the images in images.json into MIRROR_PREFIX (e.g. make mirror-images MIRROR_PREFIX=registry.example.com/kube-agents).
 	@./scripts/mirror_images.sh $(ARGS)
 
-images-check: ## Verify images.json still matches every pin it mirrors, and that the chart renders nothing off a public registry when mirrored (CI runs this).
+images-check: ## Verify images.json still matches every pin it mirrors, that the Go builder pin matches k8s-operator/go.mod, and that the chart renders nothing off a public registry when mirrored (CI runs this).
 	@./hack/check-image-inventory.sh
 
 
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -76,6 +76,12 @@ FROM ${ENVOY_IMAGE}:${ENVOY_VERSION} AS envoy-bin
 # cross-compiles to TARGETARCH, the architecture of the image being produced.
 # The :-amd64 fallback only fires on builders that don't populate TARGETARCH.
 FROM --platform=$BUILDPLATFORM ${GOLANG_IMAGE}:${GOLANG_VERSION} AS watcher-builder
+# The official golang image sets this already; repeated so a substituted
+# GOLANG_IMAGE cannot reintroduce GOTOOLCHAIN=auto, under which a
+# k8s-operator/go.mod newer than GOLANG_VERSION would download a toolchain the
+# pin does not name instead of failing. Same reason as k8s-operator/Dockerfile
+# (#1138).
+ENV GOTOOLCHAIN=local
 ARG TARGETOS
 ARG TARGETARCH
 WORKDIR /workspace
--- a/docs/site/src/content/docs/deploy/docker-images.md
+++ b/docs/site/src/content/docs/deploy/docker-images.md
@@ -15,8 +15,9 @@ A bump starts here but rarely ends here. Several images keep a second copy that 
 source for — a chart value, a Dockerfile `ARG` default, a compiled constant in the operator — and
 `make images-check` is what holds them in step. It covers every image the chart renders, on both a
 default and a mirrored install; the build-time bases against their Dockerfile `ARG` defaults; the
-fluent-bit fallback baked into the operator binary; the example manifests; and the kustomize
-integrations, which it requires to name a variable this file owns rather than a literal.
+Go builder pin against the `go` directive in `k8s-operator/go.mod`; the fluent-bit fallback baked
+into the operator binary; the example manifests; and the kustomize integrations, which it requires
+to name a variable this file owns rather than a literal.
 
 Two copies it does not reach, where a stale pin passes every check. An image behind a non-default
 chart toggle is never rendered, so Hindsight (`memory.provider`) and the GitHub token minter
@@ -275,7 +276,10 @@ make docker-build-platform \
 
 Unset args keep their upstream defaults, so an ordinary build is unchanged. Mirror the base
 images first with `INCLUDE=build-time`, and use `crane` or `skopeo` rather than `docker` — the
-Hermes pin is by digest, and a `docker pull`/`push` round trip changes it.
+Hermes pin is by digest, and a `docker pull`/`push` round trip changes it. A mirrored `golang` copy
+is frozen at whichever patch it was copied at, and the builder stages run with `GOTOOLCHAIN=local`,
+so a later bump of the `go` directive in `k8s-operator/go.mod` past that patch fails the rebuild
+instead of downloading a toolchain: re-mirror, or pass a patch-pinned `GOLANG_VERSION`.
 
 ### Registry authentication
 
--- a/hack/check-image-inventory.sh
+++ b/hack/check-image-inventory.sh
@@ -9,11 +9,18 @@
 #   2. A new image appears with no inventory entry. Nothing copies it, and the
 #      air-gapped install fails at pull time on a registry nobody approved.
 #
+# And one that has not happened yet: the Go builder pin matches the inventory
+# but no longer satisfies k8s-operator/go.mod, so the image build fails, or a
+# base without GOTOOLCHAIN=local builds with a toolchain nobody pinned.
+#
 # Run via `make images-check`; CI runs it in validate.yml.
 set -euo pipefail
 cd "$(dirname "$0")/.."
 
 INVENTORY=images.json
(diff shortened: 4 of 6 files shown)


dshnayder (approved):

bradhoekstra:
/hold

bradhoekstra:
/hold cancel

bradhoekstra:
Smoke gate is flaky and the sticky pin lost the race with Tide; overriding so the serialized queue can advance. Required GitHub checks are green and there are no open review threads.

PR #1093, opened 2026-09-01
feat(release): implement offline release bundle and sbom generation (scope 5)

eLeontev:
## Summary

Implements offline release bundles and Software Bill of Materials (SBOM) generation (PR Scope 5) for the kube-agents release pipeline. This enables air-gapped and enterprise customer deployments without internet Git dependencies, while attaching verified distribution archives (`.tar.gz`, `.zip`), Helm packages, and Syft SBOMs (`SPDX 2.3`, `CycloneDX 1.5`) to official GitHub Releases.

## Why This Change

Enterprise deployments operating in air-gapped or restricted VPC environments cannot clone Git repositories dynamically at install time. While container images are mirrored or promoted in GHCR, installer scripts and Helm charts previously required an active Git repository. Additionally, modern supply-chain security standards require cryptographically verifiable SBOMs across both filesystem assets and container images.

## What Changed (Steps 5.1 – 5.5)

- `scripts/release/common.sh`:
  - Defined declarative bundle arrays: `RELEASE_BUNDLE_DIRECTORIES` (including `terraform/`), `RELEASE_HELM_CHARTS`, and `RELEASE_BUNDLE_ROOT_FILES` (including `images.json`, `install.defaults.env`, `install.env.example`).
  - Added `extract_commit_tree` helper to extract tracked paths directly from the resolved Git commit tree via `git archive`, ensuring working-tree dirt and untracked files are never packaged.
- `install.sh`, `upgrade.sh`, `scripts/installer/installer_common.sh`:
  - Added `matches_release_bundle_ref` helper function to read `.release-bundle` metadata marker files.
  - Extended `verify_local_source_ref` in both `install.sh` and `upgrade.sh` within the `BAKED_RELEASE_VERSION` block to add provenance attribution: installer output distinguishes official release distribution bundles (`Verified install sources match official release bundle <version>`) from standard stamped archives.
  - Ensured `.release-bundle` does not grant autonomous acceptance authority: an unversioned directory carrying a marker without matching `BAKED_RELEASE_VERSION` remains strictly rejected by the existing unversioned source gatekeeper.
- `scripts/release/generate_release_sbom.sh`:
  - Added Syft-based SBOM generator producing SPDX 2.3 and CycloneDX 1.5 JSON SBOMs for bundle filesystem assets and individual SPDX 2.3 SBOMs for each of the 4 release container images.
[...]

PR #1230, opened 2026-09-04
feat(ci): drive the release-candidate eval

stalhaali:
## Summary

Adds `hack/ci-eval-rc.sh`, the entrypoint a Prow periodic will call to grade the newest release candidate: it resolves the candidate, checks the tree out at it, deploys its published images, and runs the eval against them, reporting a verdict that gates nothing. The job that calls it does not exist yet and lands later in the lane, so until then the script is dormant — `RC_EVAL_ENABLED` unset means one skip line and exit 0, wherever it runs. The awkward part is step 2: checking out the candidate replaces the tree the script is running from, sometimes deleting the script itself, so the whole body lives inside `main()`, which bash parses into memory before executing and which exits rather than returning.

## Why This Change

Issue #1024 asks for the full eval suite to run against release candidates, reported and non-gating. #1170 taught `ci-deploy.sh` and `ci-eval-pr.sh` how to install and grade a candidate's published images; nothing yet decides *which* candidate or drives the sequence. That is this script. Without it the two halves exist and nothing connects them.

The sequencing needs a driver rather than an addition to `resolve-rc-target.sh`, because whatever performs the checkout has to survive it, and a script cannot reliably survive rewriting its own file mid-run.

## What Changed

- `hack/ci-eval-rc.sh` — the driver. Four steps, dormant by default. Its path is a contract: the periodic invokes it by name. Three verdicts, not two: `GREEN`, `RED`, and `NOT RUN` for a deploy that never got far enough to measure anything. It exits with the failing step's own status, so "non-gating" stays a visible `|| true` in the job config rather than a hidden `exit 0` here.
- `tests/test_ci_eval_rc.py` — 16 tests: thirteen run the real script inside a throwaway git repository with its siblings stubbed, one pins the wrapper property on the bare mechanism, and two assert the guards' marker strings against the real sibling files rather than against the stubs the test itself wrote them into. The second of those two asserts the driver's half of the tier contract unconditionally and skips the candidate's half until #1175 lands.
- `hack/resolve-rc-target.sh` — header comment only. Its explanation of the self-rewrite hazard was wrong about the case it cited, and its worked example was the unsafe shape the paragraph beneath it warned against. Both corrected, pointing at the new driver. See the Self-Review entry.

The only executable change is the new driver, so the only thing that can regress is a lane nothing runs yet.

## Context

- Closes part of #1024. This is PR 2 of that lane; PR 1 was #1170, now merged (`e658b35e`). This branch is rebased onto `main` and depends on nothing unmerged.
- Depends on #1175 to do what it says: the driver exports `EVAL_TIER=nightly`, and nothing on `main` reads that variable yet. Until #1175 lands, every run measures the presubmit matrix and prints a note saying so. `test_the_tier_marker_is_a_string_the_real_ci_eval_pr_reads` skips that half of the contract while it is unmerged and passes silently once it is, in either merge order.
- Still to come in the lane: the dashboard's Releases data, and the `oss-test-infra` periodic itself.
- The lane now has candidates it can measure. Four exist that carry #1170's deploy path — the newest is `rc_2609041034_32c5538` — and the guard correctly refuses everything cut before it. Grading one end to end is what found the bug in the third commit below.

## Testing

- `tests/test_ci_eval_rc.py` — 16 tests, green (`OK (skipped=1)`). The skip is the candidate-side half of the tier contract, which #1175 supplies. Also run with `EVAL_TIER` appended to `hack/ci-eval-pr.sh` to simulate #1175 having merged: `OK`, 16 passed, no skip.
- The four modules covering the touched scripts and their siblings — `test_ci_eval_rc`, `test_resolve_rc_target`, `test_ci_deploy_rc_images`, `test_release_common` — `Ran 96 tests` / `OK (skipped=1)`.
- `tests/` sweep, the Makefile's invocation: `Ran 1179 tests` / `FAILED (failures=2, skipped=17)`. Both failures are in `test_generate_release_sbom.py`, assert Linux-CI swap allocation, and fail on macOS; they come from #1093 and this branch touches neither that test nor the script it exercises.
- `bash -n hack/ci-eval-rc.sh` — clean.
- Not run: `shellcheck`, not installed locally. `AGENTS.md` scopes it to the three installer scripts, so `hack/` is unlinted either way.

### Live validation

Exercised on `stalhaali.c.googlers.com` against a clone of `gke-labs/kube-agents` at `main` plus this branch's driver, byte-identical to the file under review (`md5 c5dd3223…`) — real upstream tag graph, real `git checkout`, on Linux rather than the macOS dev box. Every row below is against the branch head as of the driver's last executable change (`9135683d`); earlier rounds against `20c49feb` and `e6aedb9b` were re-run rather than kept. The two commits after it touch a comment block (`6e61c450`) and a test (`9d8ee32e`) respectively — `hack/ci-eval-rc.sh` is byte-identical at the current head — so neither invalidates a row.

| Scenario | Observed |
| --- | --- |
| `RC_EVAL_ENABLED` unset | one skip line, exit 0, `HEAD` unmoved |
| `PULL_NUMBER=1171` | refused, exit 0, `HEAD` unmoved |
| `JOB_TYPE=batch` + `PULL_REFS`, no `PULL_NUMBER` | refused, exit 0, `HEAD` unmoved |
| Dirty working tree | refused before the checkout, naming the modified file, `HEAD` unmoved |
| **Real candidate, real resolver, real `ci-deploy.sh`** | resolved `rc_2609041034_32c5538`; checkout ran and **deleted the running script**; deploy reached its own environment check and exited 1; verdict `NOT RUN`, summary written, restore hint printed, exit 1 |
| Happy path | self-deleted again; deploy and eval both ran, both received `RC_COMMIT_SHA=e13a1700…` and `EVAL_TIER=nightly` with cwd and `HEAD` at the candidate; `GREEN`, exit 0, all three artifacts, summary pointing at `eval-verdict.md` |
| Eval exits non-zero | `RED`, exit 1, all three steps ran, artifacts written — distinct from `NOT RUN` |
| No `JOB_NAME`/`BUILD_ID` | summary written with no `Run` row and no fabricated URL |

The mechanism rather than a coincidence, twice over. **The self-deletion**: `POST-RUN driver in tree: No such file or directory`, on runs that went on to write every artifact. That is the `main()` wrapper surviving a real checkout that really removed the file. **The deploy-failure row**: it is the only one that reached a real `ci-deploy.sh`, which is safe to run on the RC path precisely because that path skips the container build, and it is where the bug in the third commit came from — the previous driver aborted on the spot and wrote nothing.

Both stubbed steps reproduce the real scripts' contracts rather than a convenient version of them, and getting that wrong showed up immediately: a first stub resolver printed its banner on stdout and the driver took the banner as the SHA, which is the real `resolve-rc-target.sh`'s stdout discipline being load-bearing. The eval stub derives `ARTIFACT_DIR="${ARTIFACTS:-/tmp/artifacts}"` the way `ci-eval-pr.sh:1316` does, which is what makes the driver's `verdict_path` and the eval's output the same file.

`PROW_DECK_BUILD_BASE` checked against the live host rather than assumed: `https://oss.gprow.dev/` and the `logs/` path both return HTTP 200.

**What I could not cover, and why.** Nothing here reached a cluster. `hack/ci-deploy.sh:405` is an unguarded `helm upgrade --install` with no dry-run, and the `platform-agent-host` cluster nine other kubeconfig contexts point at is not mine to overwrite; the deploy-failure row stops one step short of it, at the environment check. Past that point the driver's contract with `ci-deploy.sh` and `ci-eval-pr.sh` is the environment it hands them and the status it takes back, and the trace above verifies both.

Cleanup: all work was in `/tmp/rc2` on that host, the scratch clones are removed, and no cluster state was created or modified. The existing `~/kube-agents` checkout was left untouched on its own branch.

## Self-Review

Both required passes ran in **fresh subagent contexts handed only the diff range and nothing else** — neither had written the change or seen the reasoning behind it. Live validation then found one thing neither pass did. Dispositions:

- **A failed deploy wrote no artifact at all.** *Fixed, and found live rather than by either pass.* The adversarial pass checked deploy-succeeds-eval-dies and called it handled; the deploy itself dying was not. `ci-deploy.sh` was invoked bare, so `errexit` took `main()` down before the summary — leaving a Prow run with nothing but a raw log, and making the summary's own "did not reach the verdict step" branch unreachable code. The status is now captured the way the eval's already was, the verdict is `NOT RUN` rather than `RED` (nothing was measured, and `RED` is a judgement on the candidate this run never formed), the summary gets a branch saying so, and the exit status is the deploy's, since `eval_status` is 0 on that path only because the eval never ran. `test_a_failed_deploy_is_reported_and_is_not_a_verdict` pins it, and against the previous driver it reproduces the live run exactly.
- **The tests did not pin the property the design exists for.** *Fixed.* The adversarial pass mechanically unwrapped `main()` and ran the suite: 11 of 12 tests passed against the unwrapped driver, and two docstrings claimed one of them covered the wrapper. The cause is structural — `git checkout` replaces a file by rename, so a running shell keeps its descriptor on the intact original inode, and **no checkout-based test can reproduce the hazard**. I verified this myself before accepting it; my first unwrap attempt failed 11 tests, but that was a broken unwrap (it ate a `local` declaration), not sound tests. `test_finishes_the_run_after_its_own_file_is_emptied_mid_step` now truncates the live inode from inside the deploy step: unwrapped it yields `['resolve','deploy']` and exit 0 — a green run with no eval in it — and wrapped it runs all three. It is the only test in the file that separates them.
- **Trust boundary one condition short of both siblings.** *Fixed.* `PULL_NUMBER` alone let a batch job through — Prow sets `PULL_REFS` and no `PULL_NUMBER` for those — which would have replaced the tree underneath the merge being tested. Now gates on `JOB_TYPE` as well, matching `ci-eval-pr.sh` and `ci-dashboard-refresh.sh`. Verified live.
- **The restore hint printed only on success.** *Fixed.* Every interesting way to strand a detached tree is a failure path. Moved to an `EXIT` trap guarded on whether the checkout happened, and the deploy-failure run above is a live failure path that printed it.
- **`EVAL_TIER` described as a mechanism that exists.** *Fixed.* It appears nowhere in the repository or on `main`; #1175 has not landed. The header framed "no tier switch" as an old-candidate fallback when it is currently the only path. Header, constant comment, and the runtime note now say so. The mechanism that was going to notice when it changed is the subject of the next entry.
- **That mechanism was an `@unittest.expectedFailure`, which would have reddened `main`.** *Fixed in `9d8ee32e`, and found after both passes and after approval.* An expected failure that starts passing is an `unexpectedSuccess`, which `unittest` counts as a failure and exits 1. #1175 adds the exact string this test greps `hack/ci-eval-pr.sh` for, and it is already `approved` + `lgtm` — so whichever of the two merged second would have taken `main`'s Python suite red, with the failure naming this file rather than the change that caused it. Neither pre-PR pass could have caught it: both read this diff, and the defect only exists in relation to another branch. Now a `skipTest` when the marker is absent, plus an unconditional assertion that the driver exports the same name it greps for — the half of the contract that is checkable today, and the half a one-sided rename would break. Verified in both worlds and as a negative control: `OK (skipped=1)` against `main` as it stands, `OK` with `EVAL_TIER` appended to `hack/ci-eval-pr.sh`, and a clean failure when the driver's export is renamed.
- **Module docstring named a deleted test.** *Fixed.* It referenced `test_bash_self_rewrite_silently_drops_steps`, which I removed before the first commit because I could not reproduce its premise. Rewritten to name the tests that exist.
- **Header sold four steps as a whole job.** *Fixed.* Teardown is the job config's, and the leased pool is shared with the presubmit eval, so a periodic that skips it degrades someone else's run. Said so in the header rather than adding a step 5, since the job config is a later PR's.
- **Untracked-file checkout abort had no explanation**, and **`rc_tag` was an unnamed cross-file literal.** *Both fixed.* Widening the dirty-tree pre-check to untracked files would fail runs that are fine (Prow workspaces hold build output), so the explanation is attached to the abort instead.
- **`oss.gprow.dev` unverified.** *Resolved, not a defect.* The adversarial pass had no network and flagged it as an open question; fetched from the remote host, both the host and the `logs/` path return HTTP 200.
- **`hack/resolve-rc-target.sh`'s header overstated the same bash/checkout claim**, and its example snippet — followed literally — produced the unsafe driver this PR works around. *Fixed, in this PR, and it widens the diff into a file nothing else here touches.* Both passes raised it independently. The header said unconditionally that "a script that checks out a different revision of itself is rewriting the bytes the interpreter has not read yet", which is the one case where that does not happen, for the rename reason above; and its three bare top-level statements are the shape the paragraph beneath them warns about. It is in scope because this PR is what makes that header misleading and what supplies the worked example it should cite, and a follow-up editing nothing but a comment block costs a review round for it. Comment-only: the resolver's behaviour and stdout contract are unchanged, and its 96 sibling tests still pass.
- **`docs/testing-map.md` cites `Makefile:129-144` and `:173`, both off by one.** *Not fixed.* Pre-existing, unrelated to this change, and reported per the skill's instruction to flag either way.

Came up clean and worth recording as negative results: nothing here can reach the release pipeline (`scripts/release/` decides from the tag graph and reads no Prow result), so the non-gating requirement in #1024 holds structurally rather than by convention; dormancy is airtight on every path, with only `readonly` assignments between `set -euo pipefail` and `main "$@"`; the baseline store cannot be poisoned, since `ci-eval-pr.sh` independently forces `EVAL_IS_MAIN_RUN=false` on `RC_COMMIT_SHA`; no secrets, IAM, RBAC, or GitHub Actions touched; test placement matches `docs/testing-map.md` and is reached by the `tests/test_*.py` glob in `PYTHON_TEST_DIRS`; no Markdown added, so no documentation-map row is owed.

## Risk & Rollout

Low. Two new files plus a comment block in a third, and the script is inert until a Prow job sets `RC_EVAL_ENABLED` — a job that does not exist yet and arrives in a later PR of this lane. Nothing in the release pipeline reads a Prow result, so this cannot hold up a release, which #1024 requires. Revert is deleting the two new files; the header edit changes no behaviour and can be left or reverted independently.

The one way this lane can cost anyone anything is the shared eval pool: the periodic that eventually calls this must run `hack/ci-teardown.sh` as a separate, unconditional step, or a leased project is left holding a live install and the next lease inherits it (#1006). That is a requirement on the later PR, and the header says so.

---

### DIFF
--- a/hack/ci-eval-rc.sh
+++ b/hack/ci-eval-rc.sh
@@ -129,7 +129,8 @@ readonly RESOLVE_SCRIPT="resolve-rc-target.sh"
 # be in memory before it happens, and control must never return to the file.
 main() {
   local script_dir repo_root rc_commit_sha rc_tag original_ref
-  local artifacts_dir summary_path verdict_path deck_url eval_status
+  local artifacts_dir summary_path verdict_path deck_url
+  local deploy_status eval_status verdict
 
   script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
   repo_root="$(cd "${script_dir}/.." && pwd)"
@@ -243,16 +244,36 @@ main() {
   export RC_COMMIT_SHA="${rc_commit_sha}"
   export EVAL_TIER="${RC_EVAL_TIER}"
 
+  # errexit is suspended for both steps, for the same reason: whatever happens,
+  # this run owes Prow an artifact saying what happened. A bare invocation here
+  # aborts main() on the spot, which skips the summary below and leaves a
+  # deploy failure legible only to somebody willing to read the raw log --
+  # and makes the summary's own "did not reach the verdict step" branch
+  # unreachable. Measured against a real candidate: ci-deploy.sh exits non-zero
+  # before it reaches a cluster if the environment is short a variable, which
+  # is an ordinary Tuesday for a job whose config is maintained in another
+  # repository.
   echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Deploying release candidate ${rc_commit_sha:0:7} ==="
-  "${script_dir}/${DEPLOY_SCRIPT}"
+  deploy_status=0
+  "${script_dir}/${DEPLOY_SCRIPT}" || deploy_status=$?
 
-  # errexit is suspended for the eval alone. A red verdict is this job's
-  # OUTPUT, not its failure -- this lane is non-gating by charter -- so the
-  # status is captured, reported, and handed back at the end rather than
-  # skipping the summary that says where to read it.
-  echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Evaluating release candidate ${rc_commit_sha:0:7} (tier: ${RC_EVAL_TIER}) ==="
+  # A failed deploy is not a red verdict. Nothing was measured, so saying RED
+  # would report a judgement on the candidate that this run never formed --
+  # the one reading that matters, because the whole lane exists to answer
+  # "is this candidate worse than main".
   eval_status=0
-  "${script_dir}/${EVAL_SCRIPT}" || eval_status=$?
+  if [ "${deploy_status}" -ne 0 ]; then
+    verdict="NOT RUN"
+    echo "ERROR: deploying ${rc_tag} (${rc_commit_sha:0:7}) failed with status ${deploy_status}, so the candidate was never evaluated. This is not a verdict on the candidate." >&2
+  else
+    echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Evaluating release candidate ${rc_commit_sha:0:7} (tier: ${RC_EVAL_TIER}) ==="
+    "${script_dir}/${EVAL_SCRIPT}" || eval_status=$?
+    if [ "${eval_status}" -eq 0 ]; then
+      verdict="GREEN"
+    else
+      verdict="RED"
+    fi
+  fi
 
   # ─── Reporting: make the verdict findable without a credential ────────────
   deck_url=""
@@ -264,7 +285,7 @@ main() {
   echo "🏷️ RELEASE CANDIDATE EVAL"
   echo "Candidate:   ${rc_tag} (${rc_commit_sha})"
   echo "Tier:        ${RC_EVAL_TIER}"
-  echo "Verdict:     $([ "${eval_status}" -eq 0 ] && echo "GREEN" || echo "RED") (advisory: this lane gates nothing)"
+  echo "Verdict:     ${verdict} (advisory: this lane gates nothing)"
   if [ -n "${deck_url}" ]; then
     echo "Artifacts:   ${deck_url}"
   fi
@@ -281,7 +302,7 @@ main() {
       echo "| Candidate | \`${rc_tag}\` |"
       echo "| Commit | \`${rc_commit_sha}\` |"
       echo "| Tier | \`${RC_EVAL_TIER}\` |"
-      echo "| Verdict | $([ "${eval_status}" -eq 0 ] && echo "GREEN" || echo "RED") |"
+      echo "| Verdict | ${verdict} |"
       if [ -n "${deck_url}" ]; then
         echo "| Run | ${deck_url} |"
       fi
@@ -292,6 +313,10 @@ main() {
       echo
       if [ -f "${verdict_path}" ]; then
         echo "Per-case detail is in \`${RC_VERDICT_FILE}\` alongside this file."
+      elif [ "${deploy_status}" -ne 0 ]; then
+        echo "The candidate was never evaluated: deploying it failed with"
+        echo "status ${deploy_status}. Nothing here is a judgement on the"
+        echo "candidate. The build log above is where it stopped."
       else
         echo "No \`${RC_VERDICT_FILE}\` was written: the run did not reach the"
         echo "verdict step. The build log above is where it stopped."
@@ -302,11 +327,16 @@ main() {
 
   # The restore hint is the EXIT trap's, so that the failure paths get it too.
 
-  # The eval's own status is preserved rather than forced to 0. Which one the
-  # job reports is the JOB's decision (an `|| true` in its config), kept there
-  # so "non-gating" is one line in the config a reader can see, instead of a
-  # swallowed exit code here that makes a red run indistinguishable from a
-  # green one to anything reading exit statuses.
+  # The failing step's own status is preserved rather than forced to 0. Which
+  # one the job reports is the JOB's decision (an `|| true` in its config),
+  # kept there so "non-gating" is one line in the config a reader can see,
+  # instead of a swallowed exit code here that makes a red run
+  # indistinguishable from a green one to anything reading exit statuses.
+  # Deploy first: on that path eval_status is 0 because the eval never ran,
+  # and returning it would call a run that measured nothing a success.
+  if [ "${deploy_status}" -ne 0 ]; then
+    exit "${deploy_status}"
+  fi
   exit "${eval_status}"
 }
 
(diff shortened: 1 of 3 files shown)


bradhoekstra (approved):

bradhoekstra:
/hold

bradhoekstra:
/hold cancel

PR #1195, opened 2026-09-03
feat(operator): reconcile dynamic litellm network policy

shalinibhatiapl:
## Context

- Closes #747 (Part C: Operator-owned LiteLLM NetworkPolicy).
- Related: #687 (original static DNS egress fix), #1090 (IaC parity for DNS egress rules).

## Changes

Transfers management of `litellm-policy` from static Helm templates to the `PlatformAgent` reconciler when `platformAgent.enabled: true` and `operator.enabled: true` (the default). The operator dynamically discovers the cluster's DNS ClusterIP from `kube-system/kube-dns` and injects exact discovered VIPs instead of guessing or falling back to RFC 1918 blocks.

- **Reconciler & Manifest Builder:**
  - Implements `buildLiteLLMNetworkPolicy`, `reconcileLiteLLMNetworkPolicy`, and `deleteManagedLiteLLMPolicy` in `k8s-operator/internal/controller/platformagent_litellm_policy.go`.
  - Stamps an `OwnerReference` on `litellm-policy` pointing to `Deployment/litellm` (and none to `PlatformAgent`), ensuring that Kubernetes garbage collection cleans up `litellm-policy` when LiteLLM is uninstalled, while preserving network isolation if `PlatformAgent` is deleted standalone.
  - Reconciles ingress from the agent namespace (port 8080) and `gke-gmp-system` (metrics scraping).
  - Reconciles egress to discovered cluster DNS VIPs (`kube-dns`), node-local DNS (`169.254.20.10/32`), GCP metadata server (`169.254.169.254/32`), Workload Identity metadata daemon (port 988, discovered/configured daemon IP), external model APIs (port 443 excluding RFC 1918), and OTLP collector.
  - Dynamically cleans up `litellm-policy` when `spec.networkPolicy.enabled: false`, when annotation `kubeagents.x-k8s.io/enable-litellm-network-policy: "false"` is present, or when `Deployment/litellm` is deleted or terminating.
  - Only deletes policies bearing `app.kubernetes.io/managed-by: platformagent-controller`, preserving hand-crafted or unmanaged policies.
- **Controller Watches & Reactive Reconciliation:**
  - Registers watches on `NetworkPolicy` and `Deployment` in `SetupWithManager` filtered to `litellm-policy` and `litellm`, re-enqueuing `PlatformAgent` in that namespace when modified or deleted.
  - Applies `predicate.GenerationChangedPredicate{}` on `Deployment` to avoid reconcile churn on pod status and readiness probe updates.
[...]

Issue #933, opened 2026-08-25
fix(agent): hermes_otel keeps the baked gke-managed-otel backend when the operator resolves otlpEndpointSource=None


mplakhtiy:
Follow-up from #831 item 5, split out of the operator fix so that change stays scoped to the operator.

## Status — updated 2026-09-03 (verified against main b0a4d7d1)

Still open. Every claim below re-checked against the current tree and none has drifted; neither of the two options has been taken.

- The backend is still baked to the managed collector: `deploy/docker/Dockerfile:1106` writes `backends: [{name: gke-managed-otel, type: otlp, endpoint: http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4318/v1/traces}]` into `/opt/defaults/plugins/hermes_otel/config.yaml`.
- `deploy/shared/otel_config.py:130` still guards the rewrite with `if endpoint:`, documented at `:97` — "An unset endpoint leaves `backends` alone, so the baked default stands". Nothing clears or disables the backend.
- `deploy/shared/docker-entrypoint.sh:1413` still passes `--endpoint "${OTEL_EXPORTER_OTLP_ENDPOINT:-}"`, and the operator omits that variable on the `None` path (`k8s-operator/internal/controller/manifest_helpers.go:181-186`).
- `HERMES_OTEL_ENABLED` appears nowhere in the repository, and `safeSandboxEnvOverrides` still allowlists only the five `OTEL_*` names plus the alert and EOD tunables (`k8s-operator/internal/controller/platformagent_manifests.go:2969-2979`).
- main's own code comment states the gap: `k8s-operator/internal/controller/manifest_helpers.go:164-169` — "It does NOT reach the hermes_otel plugin, which is where agent \*spans\* go… on this path the plugin keeps pointing at the managed collector. That is latent rather than noisy… but it is not fixed here."
- There is no documented workaround either. The chart's "Turning telemetry off" section (`charts/kube-agents/README.md:280-311`) covers `OTEL_SDK_DISABLED` and the `None` path in detail and never mentions that spans keep going to the managed collector, so an operator who follows it believes telemetry is off when the plugin is still exporting.

The operator half merged as #936.

## What the operator fix does and does not reach

[...]

PR #1231, opened 2026-09-04
fix(operator): name the resolved collector namespace in the agent egress policy

### LINKED ISSUE #1080: Egress allowlist OTel rule names gke-managed-otel while the operator resolves collectors elsewhere
\`buildAgentEgressNetworkPolicy\` renders its OTel rule to the \`gke-managed-otel\` namespace as a literal (\`platformagent_egress_policy.go\`), while the gateway policy derives the collector namespace from the resolved OTLP endpoint (\`otlpCollectorNamespace\` in \`platformagent_manifests.go\`) and omits the rule entirely when the agent exports nowhere.

Two consequences, one per era:

- Today (policy not enforcing, see #963): on an install with a discovered or configured non-managed collector, the rendered object — whose stated value is an auditable statement of the destinations the agent needs — names a namespace the agent does not send to, and permits every pod in \`gke-managed-otel\` on 4317/4318 that nothing uses.
- Once the gateway policy is narrowed (the follow-up that makes \`egressPolicy: Allowlist\` enforce): telemetry to a non-managed collector is silently blocked — the trace-less agent the rule's own comment calls out.

The builder's caller already computes \`resolveOTLPEndpoint\` in the same reconcile, so threading the resolved namespace through is mechanical; the design question is what to render when the agent exports nowhere (today: the managed namespace unconditionally, deliberately, so future telemetry is not silently blackholed — that trade-off should be revisited together with the namespace derivation).

Found by the adversarial pre-PR pass on #963; behaviour kept there deliberately since the fix touches a documented, hedged design decision and the rule enforces nothing until the narrowing lands.

bradhoekstra:
## Summary

The agent Pod's opt-in egress allowlist policy rendered its OpenTelemetry rule to `gke-managed-otel` as a literal, while the gateway policy on the same Pod reads the collector namespace off the endpoint the operator actually resolved. On an install whose collector lives elsewhere, the allowlist policy therefore permitted a namespace the agent never sent to and, once the allowlist enforces, blocked the export it did make. This PR threads the resolved endpoint into the allowlist builder on both of its reconcile paths and derives the namespace the same way the gateway policy does, keeps the managed namespace for an agent that resolved to no collector at all, and adds tests at the builder and reconcile level. The CRD field description and the credential-isolation reference page are updated to match.

## Why This Change

Both NetworkPolicies select the same Pod and both are rendered from the same `resolveOTLPEndpoint` result, yet they could disagree about where the collector is. Today the allowlist policy does not enforce on most installs because the gateway policy's union covers it, so the visible symptom was an auditable object naming a destination the agent does not use. On the one install shape where the allowlist is the Pod's only policy, and on any install once the gateway policy is narrowed, a discovered or configured collector would have been silently unreachable: a trace-less agent with a Ready status.

## What Changed

- `buildAgentEgressNetworkPolicy` takes the collector namespace and renders the OTel rule only when there is one; `reconcileAgentEgressPolicy` takes the resolved endpoint and derives the namespace with `otlpCollectorNamespace`, on the normal path and on the refusal path that `reconcileAgentNetworkGuardrails` keeps maintained.
- An agent that resolved to `None` keeps a rule to `gke-managed-otel`, unlike the gateway policy, because the Hermes tracing plugin does not read `OTEL_EXPORTER_OTLP_ENDPOINT` and still addresses the baked managed backend on such a cluster (#933). An endpoint with no in-cluster namespace renders no rule, as for the gateway policy. Both choices are explained in the rule's comment.
- The `gke-managed-otel` literal becomes `managedOTelCollectorNamespace` in `telemetry.go`, and `managedOTelEndpoint` is now spelled from it so the two cannot drift.
- Two new tests: one drives the builder with the managed namespace, another namespace, and no namespace; the other reconciles a CR with `spec.telemetry.otlpEndpoint` set to a collector in `observability`, on the normal path and through the `EgressAllowlistRefused` refusal, and asserts the rendered policy names `observability`, not `gke-managed-otel`, and agrees with the gateway policy.
- The `egressPolicy` CRD field description no longer says the rule is rendered unconditionally to the managed collector; the CRD YAMLs are regenerated and the chart copy synced. The credential-isolation page's rule list and its "blocks nothing today" paragraph, and one sentence each on the telemetry, security-requirements and CRD reference pages, say the same thing.

## Context

Closes #1080.

The `platformagent-egress-allowlist` golden fixture is the no-collector case and is unchanged, which is what the design predicts: `""` still maps to the managed namespace. No open PR touches `platformagent_egress_policy.go`; #1195 and #913 edit neighbouring files, merge-conflict exposure only.

## Testing

- `go build ./...`, `go vet ./internal/controller/`, and `go test ./... -count=1` in `k8s-operator/` pass, including the golden fixtures under `internal/testing`.
- Red-then-green: with the builder temporarily restored to the unconditional `gke-managed-otel` rule, both new tests fail on the `collector elsewhere`, `no in-cluster namespace`, `normal path` and `refusal path` cases; the adversarial review pass reproduced the same red independently on a copy of the tree.
- `make manifests`, `make chart-sync`, `make chart-check`, `make docs-check`, and `prettier@3.9.6 --check` on the changed Markdown and the regenerated CRD YAMLs are clean.

### Live validation

Run on the `bhoekstra-gkedemos` install (`kube-agents-cluster`, `us-central1`, namespace `kubeagents-system`) on 2026-09-04 under the live-test lease, as a before/after in which the only variable was the controller image.

Setup: the install's own operator and CRD predate the egress-policy fields, so the branch CRD was applied and the controller was first pointed at an operator image built from `main` at 32c5538, with `--webhook-port=9443` added to its args so the installed webhook Service could reach it. The CR was then given `spec.telemetry.otlpEndpoint: http://otel-collector.observability.svc.cluster.local:4318`, `spec.security.egressPolicy: Allowlist`, `spec.security.splitCredentialBrokerPod: true`, and `spec.harness.eventWatcher.enabled: false` (the split broker refuses otherwise). Status telemetry read `otlpEndpointSource: Spec` with that endpoint.

Observed, reading the rendered NetworkPolicies' 4317/4318 peers:

| Controller image | `platform-agent-sandbox-metadata-deny` | `platform-agent-gateway-netpol` |
| --- | --- | --- |
| `main` at 32c5538 | `gke-managed-otel` | `observability` |
| this branch at 9e86dc5 | `observability` | `observability` |

The controller image was swapped with `kubectl set image` and nothing else changed between the two reads; the rule moved within 20 seconds of the rollout, and a JSON diff of the two policies shows the namespace selector as the only difference, every other egress rule byte-identical. The controller logged the "agent Pod egress is default-deny with an allowlist" line for the reconcile that rendered it.

Not covered: the cluster runs the Standard dataplane, so whether an enforcing CNI passes traffic under the rendered selector was not observed; only the rendered object was. The split broker also cannot run on this install (ReadWriteOnce data claim), so the agent was unavailable for the duration of the test, roughly twenty minutes.

Cleanup: reverted in the order the CRD field description gives (`egressPolicy: None`, delete `-sandbox-metadata-deny`, then remove the split and the test-only fields), restored the controller image and args and the previously installed CRD, and diffed the CR `.spec`, the controller Deployment, the ClusterRole and the NetworkPolicy set against a snapshot taken before the test: all identical. The gateway pod is 3/3 and the CR is `Reconciled` with telemetry back on the discovered managed collector. Two image tags remain in the install's Artifact Registry, `k8s-operator:egress-otel-9e86dc5` and `k8s-operator:egress-otel-main-32c5538`.

## Self-Review

Both passes ran in fresh subagents handed only the repository, the three-dot range against a freshly fetched `main`, their skill path and the `make docs-check` output, with the instruction to derive intent from the diff and not read commit messages. Two rounds: the first round's fixes changed the CRD description and a test, so both passes re-ran, and the second round returned nothing above LOW or Advisory. Angles covered: line by line with both callers, removed behaviour, cross-file consumers of the changed signatures and every new identifier, security surface of a user-controlled namespace selector, reuse against the gateway policy's derivation, repository conventions, scope and red-then-green, sibling PRs, and docs-to-source for the changed pages, the CRD description, the map and `AGENTS.md`.

Fixed:

- The `egressPolicy` CRD field description still said the OTel rule is "rendered unconditionally" to "the managed OpenTelemetry collector" (docs-drift, Blocking). Reworded in `common_types.go`, CRD regenerated, chart copy synced.
- The credential-isolation page's rule list omitted the Hindsight memory API rule and did not say the OTel rule is absent for an endpoint with no in-cluster namespace (adversarial LOW, docs-drift Advisory). Both added.
- `managedOTelEndpoint` embedded the namespace as a second literal, so the new constant could drift from it unnoticed (adversarial, LOW plausible). The endpoint is now spelled from the constant.
- The "no in-cluster namespace" test asserted only that two named namespaces were absent (adversarial, LOW plausible). It now asserts no rule carries an OTLP port with a namespace selector.
- The telemetry page's `None` sentence said the collector rule is left out of "its NetworkPolicy", which is true of the gateway policy and the opposite of the allowlist policy's documented choice (docs-drift, Advisory). Qualified and linked.
- Two summary pages said "the collector namespace" where the canonical page now says "the managed collector namespace" (docs-drift, Advisory). Tightened.
- One gofmt realignment of an unrelated test line (adversarial, LOW). Reverted to keep the diff scoped.

Not fixed, by decision:

- A `spec.telemetry.otlpEndpoint` naming a Service in `kube-system` opens 4317/4318 to that whole namespace (adversarial, not raised as a finding). The gateway policy already renders the identical peer from the same input, so this widens nothing.
- Keeping the managed-namespace rule on a `None` cluster leaves the two policies asymmetric (adversarial, confirmed as deliberate and correct). Documented in the rule's comment and on the credential-isolation page, with #933 as the reason; when #933 is fixed this rule can follow the gateway policy.

Not covered by the passes: no docker build or kubectl inside them; the live validation above is the author's. The adversarial pass could not retrieve #913's hunks for `credential-isolation.md` because that PR's diff exceeds the API limit, so treat that page as a rebase risk against it.

## Risk & Rollout

Low. The change alters one rule in an opt-in policy that on most installs is unioned with the gateway policy and enforces nothing; where it does enforce, the rule now names the collector the agent exports to instead of a namespace it does not. On `None` and `Default` clusters the rendered object is unchanged, which the untouched golden fixture confirms. The CRD change is description text only. Revert by reverting the commit.


dshnayder (approved):

bradhoekstra:
/hold
Serializing the merge queue around the flaky smoke gate — releasing one PR at a time so Tide stops forming multi-hour batches. This hold is temporary; I will `/hold cancel` it when its turn comes.

bradhoekstra:
/hold cancel

bradhoekstra:
/retest
The 12:42 smoke run died before the test container started: the Prow pod ended in phase Failed with a `NodeNotReady` event and no build log was produced (only clone-log and podinfo uploaded). Infrastructure, not this change.

bradhoekstra:
Overriding on the author's instruction: the only red run at this head was a Prow node loss (NodeNotReady, no build log), the rebase was verified locally (go build/vet/test, make chart-check, docs-check), required GitHub checks are green, and no review threads are open.

PR #1232, opened 2026-09-04
feat(terraform): let full-install leave a Vertex serving project it cannot administer alone

bradhoekstra:
## Summary

With `model_provider = "vertex_ai"` the `full-install` composition manages two resources in the serving project rather than the install's own: it enables `aiplatform.googleapis.com` there and grants the gateway's service account `roles/aiplatform.user` on it. When `vertex_project_id` names a project the applying identity cannot administer, both fail with a 403 after everything else has been created, and nothing let an operator opt out. This adds `vertex_manage_serving_project` (default `true`) to the composition and carries it through the installer as `VERTEX_MANAGE_SERVING_PROJECT` / `--vertex-manage-serving-project`, so such an install skips the two and the operator makes the grant by hand. The gateway's service account and its Workload Identity binding are unaffected: they live in `project_id`.

## Why This Change

A cross-project `vertex_project_id` is usually a serving project someone else owns, which is exactly the case where the installing identity has no `setIamPolicy` there. The first consumer is a dedicated live-test install whose service account holds IAM only in its own project and calls `gke-claude-dev` for models; without this variable its apply cannot complete, and the only workaround is importing and forgetting the two resources around every apply and destroy.

## What Changed

- `terraform/examples/full-install`: the variable, a `manage_vertex_serving` local, and both serving-project resources gated on it. `variables.tf` and the README say the choice belongs to the first apply, because turning it off later on an install that already manages the two destroys them and revokes the grant, and give the `terraform state rm` command that hands them over instead.
- `install.sh`: the flag, help text in the file's `(default: DEFAULT_X, currently y)` form, the `PARAM_`/`install.env`/`--menu` plumbing its siblings have, a loud `^(true|false)$` validator like `--gvisor`, and an info line under `vertex_ai` saying what to grant by hand when it is `false`.
- `install.defaults.env` gains `DEFAULT_VERTEX_MANAGE_SERVING_PROJECT="true"`; `installer_common.sh` writes `vertex_manage_serving_project` into the generated tfvars through `hcl_bool`.
- Docs: `install.env.example`, `terraform.tfvars.example`, the installer README's defaults table, the inference-gateway page, and the security-and-iam page (whose "Terraform-managed, so switching provider leaves no orphan" claim now states the exception).

## Context

Related: #1193 moves the eval installs to Vertex and edits `module.litellm_vertex_iam`'s `depends_on` twelve lines from these hunks; textual adjacency only, no semantic overlap. No open issue or pull request asks for this.

## Testing

- `python3 -m unittest tests.test_vertex_manage_serving_project tests.test_installer_common tests.test_install_script tests.test_vertex_location_defaults`: 167 tests, OK. The new module pins the HCL: the variable exists with default `true`, the local combines `local.use_vertex` with it, exactly the two serving-project resources follow the local, and the gateway identity does not.
- `terraform fmt -check` and `terraform validate` (1.15.8) in `terraform/examples/full-install`.
- `make docs-check`, `prettier --check` on the changed Markdown, `shellcheck` on `install.sh` and `installer_common.sh` (no findings on either).

### Live validation

`terraform plan` against the real, empty project `kube-agents-robot` (unrefreshed, local state, `-target` on the two resources) with `model_provider = "vertex_ai"` and `vertex_project_id = "gke-claude-dev"`:

- `vertex_manage_serving_project = true`: `google_project_service.vertex_ai[0]` (`aiplatform.googleapis.com`) and `google_project_iam_member.litellm_vertex_user[0]` (`roles/aiplatform.user`) are planned, alongside `module.litellm_vertex_iam[0].google_service_account.agent`.
- `vertex_manage_serving_project = false`: neither serving-project resource appears; `module.litellm_vertex_iam[0].google_service_account.agent` is still planned. `Plan: 8 to add` (the seven required APIs and the service account).

Not covered: no apply, so the 403 the variable avoids was not reproduced against `gke-claude-dev`, and the `--menu` and interactive interview paths were not exercised with a tty. The first real apply with `false` is the live-test install that motivates this change and will be reported on its own pull request.

## Self-Review

Both passes ran in subagents handed the diff range and the skill path only, with the instruction to derive intent from the diff and not read commit bodies or plan files. Mechanical gate (`make docs-check`) passed before the passes ran.

`review-adversarial` (angles A–J covered; no `terraform apply`, no tty run, no red-then-green since the change is a feature):

- MEDIUM, CONFIRMED — turning the variable off on an install whose earlier apply created the two resources destroys them on the next apply, which revokes the grant, and the prose read as if `false` were always a no-op. Fixed: `variables.tf`, the composition README (with the `terraform state rm` command), the inference-gateway page, and the `install.sh` info line now say the choice belongs to the first apply and how to hand the resources over.
- LOW, CONFIRMED — the boolean was normalised silently through `is_truthy` while its sibling flags reject anything but `true|false`, and a comment cataloguing the silent booleans would have gone stale. Fixed: a loud `^(true|false)$` validator like `--gvisor`; the comment stays true.
- LOW, CONFIRMED — no test would fail if either `count` reverted to `local.use_vertex`. Fixed: `tests/test_vertex_manage_serving_project.py`.
- LOW, CONFIRMED — help text restated the default as a literal against the file's convention and the installer README's rule. Fixed.
- LOW, CONFIRMED — `security-and-iam.md`, the `vertex_project_id` description, and `terraform.tfvars.example` stated the grant unconditionally. Fixed.
- LOW, CONFIRMED — the new info line hard-coded `kubeagents-litellm-gsa` as a fallback literal, copied from the adjacent pre-existing line. Fixed for the new line with a named constant; the pre-existing line is not touched, per "the rule binds the lines you write".

`review-docs-drift` (traced every page naming `VERTEX_LOCATION`, `VERTEX_PROJECT_ID`, `aiplatform.user`; verified the "only two resources" claim against `main.tf`; docs site built):

- Blocking — `security-and-iam.md` "Terraform-managed, leaves no orphan" now conditionally false. Fixed (same fix as the adversarial LOW above).
- Advisory — installer README defaults table lacked the new default beside `DEFAULT_VERTEX_LOCATION`. Fixed.
- Advisory — install skill flags table lacked the flag. Not changed: the table is curated, already omits the sibling `--vertex-project-id`, and states `--help` is authoritative; adding the row re-aligns the whole table.
- Advisory — help text default form. Fixed (above).
- Advisory — three docs said "make the grant" where `false` also skips API enablement. Fixed: all say enable the API and make the grant.
- Advisory — `terraform.tfvars.example` lacked the variable. Fixed.
- Advisory — `render_install_env.sh` has no mapping for the key, so a CI-reconciled environment cannot opt out. Not changed: the CI pool projects administer their own serving project, so the default is right there; adding the mapping means also editing the environment-reconcile page, a separate change if a CI install ever needs it.
- Advisory — the inference-gateway page's closing "one apply reconciles the switch" sentence is unqualified. Not changed: the new bullet above it states the exception, and the sentence remains true for the default.
- Advisory — the README's "only two resources" count could go stale. Kept, and now pinned by `test_nothing_else_lives_in_the_serving_project`.
- Advisory, pre-existing — `INSTALL.md` says `MODEL_PROVIDER=vertex` where the validator wants `vertex_ai`, and two comments point at `installer_common.sh` for a default that moved to `install.defaults.env`. Not changed here; out of this change's scope.
- Not covered by either pass: the `--menu` path saves the key but never prompts for it, so a menu user sets it through `install.env` or the flag. Left as is; the sibling `VERTEX_PROJECT_ID` behaves the same way.

## Risk & Rollout

Default `true` keeps every existing install on today's behaviour; a tfvars generated by an older `install.env` has no key and takes the default. The only new path is `false`, which does less. Revert is a plain revert; nothing has to happen at merge time.


dshnayder (approved):

bradhoekstra:
/hold

bradhoekstra:
/hold cancel

PR #1234, opened 2026-09-04
refactor(agents): publish the audit ledger through the forge provider

AntonTyb:
## Context

`audit_report.py` — the fleet audit's publish helper, which opens and closes issues and pull
requests on user repositories unattended, on a cron — reached GitHub through twenty `gh`
invocations behind a private runner. This moves all twenty onto `forge.ForgeProvider`.

Step 3 of [`docs/designs/multi-forge-support.md`](https://github.com/gke-labs/kube-agents/blob/main/docs/designs/multi-forge-support.md)
§9, second of three parts. `resolver.py` is the one consumer left and is deliberately not here:
its runner rewrite (#1153) has merged, so the migration is unblocked, but it stays its own change.

**Stacked.** Review commit `40081267` only.

- #1150 — the `RepoRef` refactor (step 1)
- #1159 — `spec.integration.git` (step 2), stacked on #1150
- #1161 — the PR-lifecycle methods and `submit_suggestion.py` (step 3, first part), stacked on #1159
- **this PR** — stacked on #1161

## What changed

`ForgeProvider` grows the ledger group — labels, issues, and the bulk pull-request reads the run
reconciles its own fixes against. `audit_report.py` calls it through `forge_for(repo)`, which
injects the module's own `run_cmd` as the provider's runner: everything the module knows about its
own I/O hangs off that function, so a provider reaching `subprocess` directly would take the
command log, the FAILED line and the test suite with it, and the publish would go on working while
it stopped being observable.

Every call site keeps the error policy it had. What changes is what the log says — a refused write
is `FORGE_WRITE_REFUSED` with the exit code and both streams rather than `gh exited 1`, because the
caller has already read that repository over that credential in the same run, so unreachability is
the one answer ruled out.

Three behaviour changes came out of the migration rather than going into it, all in the commit
message: `create_pull_request` takes the URL off the last non-blank line so a `gh` version notice
cannot be recorded as the address; a create refused because the branch already has an open pull
request is adopted, labelled, and updated in place rather than being fatal; and `Issue` carries no
body, because a body that is `""` because nobody projected it is indistinguishable from a ledger
with nothing in it, which is the reading that announces every live finding as new.

`issue_comments` and `pull_request_comments` return the forge's own dictionaries rather than
`forge.Comment`. `is_machine_author` — the gate that stops the automation issuing itself a
`/remediate` — reads `authorAssociation` and `viewerDidAuthor`, and that dataclass carries neither.
Widening it is a change to a security control, so it is its own PR; §4 of the design doc records
the gap.

## Testing

`test_forge.py` 130, `test_audit_report.py` 552, `agents/platform/scripts` discover 1320 (one
pre-existing unrelated `test_session_kv_server` import error for a missing `fastapi`),
`tests/test_shipped_command_policy.py` 38. `make docs-check` clean, `prettier@3.9.6 --check` clean
on the changed docs.

Only two of the 547 pre-existing audit tests needed touching, both for understood behaviour
changes: `test_a_failed_pr_create_does_not_fail_the_run` (the failure is now `ForgeError`, so
`_open_promoted_prs`' except tuple widened) and `test_a_failed_issue_create_is_fatal` (asserts the
reason code now, not `gh`'s wording).

### Live validation

**Not live-tested.** I have no reachable kube-agents installation, and this helper only runs inside
a deployed agent pod against a real repository. What stands in for it is a differential replay,
which is stronger evidence than a single manual run would have been for a change of this shape.

I instrumented the test harness' `Recorder` to capture every `gh` argv the suite emits, ran the
whole audit suite against `HEAD~1` and against `HEAD`, and diffed the normalised multisets — 1144
invocations before, 1166 after. Every difference is accounted for:

| Difference | Why |
| --- | --- |
| `-F` → `--body-file` on issue and PR comments | Aliases of the same `gh` flag. The credential proxy's only `-F`-keyed rule requires a `gh … api …` token and never matched these. (Recorded before the rebase; after it both sides read `--body-file -` with the body on fd 0.) |
| `--json number,url` → `--json number,url,title,labels` on the ledger listing | `ISSUE_FIELDS`, the one projection `list_issues` declares. |
| one extra `gh pr view <branch> --json url` | `create_pull_request`'s read-back, which fires only when `gh pr create` exits 0 having printed nothing — never under a real `gh`, always under the `Recorder`, whose default reply is empty. |
| new `issue close`, `pr edit`, `pr view` calls | The five tests this PR adds. |

Flag ordering, `-R` placement, label repetition, `CHANGE_FIELDS` ordering and `--limit` values are
byte-identical. The reviewer of the adversarial pass reproduced this independently and reached the
same list.

**This replay predates the rebase and has not been re-run against the head.** What it establishes —
that the migration changes no call, no ordering and no field projection — is unaffected by the one
thing the rebase changed, which is how the body bytes reach `gh`. That part was checked separately
by reverting it: deleting `_forge_runner`'s stdin forwarding makes one test fail, and it passes
again when restored.

What this does not cover: `gh`'s actual "already exists" wording for a refused `pr create`, and its
stdout/stderr split for that refusal, are taken from `_PR_EXISTS_MARKER`'s comment rather than
observed. A reviewer with an installation could confirm the adoption path in one run by stripping
`audit:remediation` from an open remediation PR and waiting for the next audit tick — it should
re-label and update in place rather than warn.

## Self-Review

Both required pre-PR passes ran against `HEAD~1..HEAD` in subagents that did not write the change.

**Adversarial** (`review-adversarial`) — briefed to look hardest for argv that differs from what
`gh` received before, error policy that silently flipped between raise and log-and-continue,
exceptions now escaping or now swallowed, and anything that could make an unattended run publish
something wrong or publish nothing while reporting success.

| Finding | Disposition |
| --- | --- |
| **HIGH** — the `PullRequestExists` adoption path applies no labels, so the adopted PR stays outside `list_remediation_prs` and every later run re-adopts and re-pushes it forever while reporting the URL as a fix | **Fixed.** The adopt path now calls `sync_remediation_labels`. `set_pull_request_labels` takes a branch as well as a number, since the adopt path has no number. Three tests added. This was the one real defect either pass found, and it was mine — the adoption path was a behaviour change I added without a test. |
| **LOW** — `Issue.body` declared but never projected, so always `""`, which is the value that reads as "empty ledger" | **Fixed** by removing the field and naming the projection `ISSUE_FIELDS`, matching `CHANGE_FIELDS`. |
| **LOW** — three places claim `run_cmd` gives the provider "credential redaction"; `redact_secrets` is only called from the body renderers | **Fixed.** Claim corrected in all three (`_forge_runner`, the new test class, §4). |
| **LOW** — the adoption path has no test | **Fixed**, as above. |
| MEDIUM — design-doc operation counts off by one | Same as the docs-drift pass' finding 1; fixed there. |
| Verified clean: no error policy flipped; `_call` returning `[]` matches the old `json.loads(… or "{}")` on every path; `find_existing_issue`'s highest-number-wins survives the sort moving into `list_issues`; `ForgeProvider` is a plain `Protocol`, so the new methods do not break `test_github_scan_gate.FakeProvider` | — |

**Rebase onto current `main`** (2026-09-08, `40081267` → `367fe85a`, folding in the earlier
2026-09-07 round). One substantive change came out of the first round rather than being a
mechanical merge. The second applied without conflict; #1200 touches two of this branch's files,
so I checked its `validate_repo_org` call in `audit_report.resolve_repo` and its `forge.py`
docstring rewrite both survive the merge rather than trusting the absence of markers. Gates
re-run on the new head.

| Finding | Disposition |
| --- | --- |
| The branch staged every body through `audit_report._write_temp`, and `main` had already deleted that helper — since #913 the sandbox running the audit and the broker running `gh` are two containers with two filesystems, so `--body-file <path>` names a file only the writer can open, and each of these calls fails *after* the state change it was reporting on | **Fixed.** The provider methods take `body: str` and pass `--body-file -`; `_forge_runner` forwards fd 0 to `run_cmd`. Every `_write_temp`/`_unlink` call and its `try:`/`finally:` scaffolding is gone. |
| Coverage for that could have been vacuous | **Checked by reverting it** — removing the stdin forwarding fails one test, and `test_every_ledger_body_travels_on_fd_zero` was added to assert the argv and the bytes across all six body-carrying calls. |

**Docs-drift** (`review-docs-drift`) — briefed to hunt prose the change made false, and told the
SOP rewording is deliberately scheduled as step 6.

| Finding | Disposition |
| --- | --- |
| **Blocking** — protocol stated as 24 operations in three places; it is 25, and the ledger group is 15, not 14 | **Fixed** by dropping the totals rather than correcting them — a hard-coded count is the drift magnet that produced this finding, and the `ForgeProvider` docstring already says the same thing by group. |
| **Blocking** — `forge.py`'s module docstring still names `audit_report.py` as unmigrated | **Fixed.** |
| **Blocking** — `pr-comment-conversation.md` §3 says §4 "has since added three more" | **Fixed**; it now names the two groups without a count. |
| **Blocking** — `pr-comment-conversation.md` §7 says two consumers remain | **Fixed**; one. |
| Non-blocking — `governance-sops.md:77` and `proactive-autonomy.md:38` carry verbatim the sentence this change judged stale on `autonomous-watchdogs.md:35` | **Fixed.** Neither was false, but leaving them made three pages disagree about one mechanism. |
| Non-blocking — §4's "the prohibitions get reworded **now that**…" invites a reader to check whether the SOPs were touched | **Fixed**; it now says explicitly that nothing has been reworded yet. |
| Non-blocking — `fleet-audit/SKILL.md` and `fleet-audit-issue-ledger.md` describe the publish path in `gh` terms | **Deferred, deliberately.** Still literally true — `_forge_runner` shells `["gh"] + argv` — and rewording the runtime prompts is step 6 of §9, which the design doc schedules as its own PR. |
| Non-blocking — the map's identifier-sources table has no row anchoring `forge.py` | **Not taken.** The reviewer's own preferred remedy was to drop the counts from prose, which is what I did, so there is nothing left to anchor. |

**Not covered by either pass:** no live install (see above); `resolver.py`, `submit_suggestion.py`
and `pr_conversation.py` were checked only for signature compatibility — the two changed provider
signatures gained keyword-only parameters and are backward compatible.

**Sibling PRs worth knowing about:** #1118 adds `conflict_state`/`failing_checks` to `ForgeProvider`
in the same region — textual conflict, no semantic clash. #1200 touches both files and needs a
rebase over this stack whichever order they land in; it adds no `gh` calls to `audit_report.py`, so
removing the `gh()` helper does not break it.


fkc1e100:
Closing as stale. This forge ledger refactor has been open for 19 days with unresolved conflicts.

Part of the multi-forge work. If still needed, please re-evaluate against current architecture.

Issue #964, opened 2026-08-26
Two reconcile refusals suspend the agent Pod's gateway NetworkPolicy, leaving egress unrestricted if it is deleted


bnaylor:
## What

Two `updateStatusDegraded`-and-return sites in `PlatformAgentReconciler.Reconcile` sit above the step that reconciles `<name>-gateway-netpol`, so while either refusal holds, that NetworkPolicy stops being reconciled:

- step 10, `RuntimeClassNotFound` — `platformagent_controller.go` around line 262
- step 10b, `SplitBrokerStrandsEventWatcher` — around line 278

## Why it matters

A Pod that no NetworkPolicy selects is not restricted. NetworkPolicy is deny-by-default only for Pods some policy selects, so deleting `<name>-gateway-netpol` while one of these refusals is in force leaves the agent Pod with **unrestricted** egress, and nothing puts the policy back until the spec is fixed. The CR meanwhile reads `Degraded` with a reason about a RuntimeClass or an event watcher, which says nothing about the network.

The path is not hypothetical: an operator triaging a `Degraded` agent deleting objects to force a clean re-render is ordinary behaviour, and the status gives them no reason to think a policy is at risk.

## Why not fixed in the change that found it

#963 hit the same hazard at its own refusal (step 10d, `EgressPolicyRequiresSplitBroker` / `EgressAllowlistRefused`) and closed it there with `reconcileAgentNetworkGuardrails`, which reconciles the gateway policy on both of its refusal paths before returning. It deliberately did not touch these two: one arrives with #960 and one is pre-existing on `main`, and widening that change's blast radius to two unrelated refusal paths was not the right call inside it.

The result is that the comment on the 10d block says a refusal must not withhold a guardrail, three lines below two refusals that do. That contradiction is the reason to file rather than leave it implicit.

## Suggested fix

Either call `reconcileAgentNetworkGuardrails` (or just `reconcileNetworkPolicy`) on both paths before returning, or hoist `reconcileNetworkPolicy` above step 10 entirely. The hoist is the tidier shape and covers any refusal added later, but note that it also moves the policy in the normal path, and the golden harness records objects in apply order — #963 measured this and it churns all five fixtures, so the hoist wants its own change with the regeneration in it.

Worth a test per path, of the shape `TestARefusalDoesNotSuspendTheGatewayNetworkPolicy` in `platformagent_egress_policy_test.go` uses: assert the refusal fired, delete the policy, require the next reconcile to restore it.

## Related

[...]

PR #1240, opened 2026-09-04
fix(operator): report an out-of-date ClusterRole instead of looping, and refuse a mutable image on make deploy

### LINKED ISSUE #1009: Operator RBAC skew: `make deploy` defaults to `:latest`, so a pod reschedule silently upgrades the controller past its applied RBAC

bradhoekstra:
## Summary

`make deploy` ships the operator's image and its ClusterRole together only when it is re-run. A controller deployed from a floating tag is upgraded on its next pod reschedule while the applied role stays put, and the first verb the newer controller needs that the older role lacks fails every reconcile with `forbidden`, an unbounded `Reconciler error` loop that reads like an RBAC misconfiguration. This PR makes the dev path refuse a floating tag, and makes the operator check its own permissions and say, in one log line and on the CR, when its ClusterRole is behind its image.

## Why This Change

The issue's cluster hit it when #733 added PDB `patch` and a reschedule pulled the newer `:latest`; the gkedemos live-test install hit the same skew on 2026-09-02, where it aborted every reconcile before the NetworkPolicy step and nothing in the cluster or the logs pointed at the cause. The supported installers already refuse mutable refs (`validate_immutable_ref`); the kustomize dev path did not, and the operator had no way to notice or report the skew.

## What Changed

- `k8s-operator/Makefile`: `deploy` now depends on `check-img`, which refuses an `IMG` whose tag is `latest`, `main`, `master`, `HEAD`, `dev`, or missing unless `ALLOW_MUTABLE_IMG=1` is set. A digest reference always passes. It is a denylist of the tags known to float, deliberately looser than the installer's SHA-or-SemVer allowlist so short SHAs and dev tags still deploy. `VERSION ?= latest` is unchanged because it also feeds the fallback agent image.
- `k8s-operator/internal/controller/rbac_selfcheck.go`: the operator probes one `SelfSubjectAccessReview` per (verb, resource) its RBAC markers declare (179 of them, 16 in flight at a time, under a 30 s deadline), at startup and every five minutes after on its own ticker as a manager Runnable, never on the reconcile worker. A denial is one `RBAC self-check failed` line at boot naming the denied permissions and the cause, and a `Degraded` condition with reason `RBACIncomplete` carrying the same text on each `PlatformAgent`; `Reconcile` only reads the last answer. Reconciles continue: a verb used only on an optional path must not take a working install down, and a bail-out withholds guardrails (#964). While a denial stands the reconcile requeues every five minutes so a re-applied role clears the condition without an unrelated event; a probe that cannot run keeps the previous answer.
- `platformagent_controller.go`: the report runs before any step that could fail on the missing verb, so the status carries the cause even when the reconcile errors below. `updateStatusReady` keeps the `RBACIncomplete` condition and counts it as unchanged, so a standing denial does not write status on every pass and re-enqueue the agent through the unfiltered watch. `InvalidGitRepoURL` keeps ownership of the `Degraded` condition when both problems exist.
- `cmd/main.go`: builds the checker, runs the boot probe (before the health endpoints exist, hence the deadline), logs the result, hands the checker to the reconciler, and adds it to the manager so it ticks.
- The probe's permission list is a hand copy of the markers by necessity (markers are comments, `role.yaml` is not in the binary). `TestRequiredPermissionsMatchTheGeneratedRole` parses `config/rbac/role.yaml` and fails on any difference in either direction.
- Docs: the operator page gains "An image ahead of its ClusterRole" beside the webhook-port skew it already documents, and is the one home for the refused-tag list; the CRD status reference describes both `Degraded` reasons; the dev-path examples deploy `:$(git rev-parse HEAD)`; the identifier-sources map gains a row.

## Context

Closes #1009.

Open PR #1211 adds rules to `role.yaml` (secrets write, jobs and resourcequotas write). Whichever of the two merges second will see `TestRequiredPermissionsMatchTheGeneratedRole` fail until `requiredPermissions` is updated to match; that is the test doing its job, and the failure message says which side to change.

## Testing

- `go build ./...`, `go vet`, and `go test ./...` in `k8s-operator/`: all packages OK. `make manifests generate` produces no diff; `make chart-check` in sync.
- New tests: the probe against a fake clientset (names each denied tuple including a subresource, all-granted reports nothing, a transport error is not a denial, denials come back in declaration order under concurrency, the deadline stops a slow probe without recording anything, the ticker re-probes on its own and keeps the last answer on error, nil checker never probes); the RBAC tests also pass under `-race`; the reconcile-level test (condition set with the denied verb and cause, workload still rendered, `RequeueAfter` equals the interval, condition and poll clear once granted); a steady-state test asserting zero status writes over three passes under a standing denial; the yield-to-`InvalidGitRepoURL` rule; `role.yaml` parity. The reconcile test is red against the pre-change `updateStatusReady` and the missing call site.
- `tests/test_operator_makefile.py`: four new tests run the real `check-img` target for floating, missing, registry-port, SHA, SemVer, and digest references, the override, and `deploy`'s dependency on it. Nine tests OK.
- `make docs-check` clean; pinned `prettier@3.9.6` clean on the changed Markdown.

### Live validation

Install: `bhoekstra-gkedemos` / `kube-agents-cluster` (us-central1), namespace `kubeagents-system`, a kustomize deploy whose operator runs `k8s-operator:dns-endpoint-7dac349` and whose ClusterRole matches `main` (checked by set-diff against `role.yaml`). The operator image was built from this branch, pushed under a throwaway tag, and swapped in with `kubectl set image`; the agent pod recreated once and came back 3/3.

Fresh boot with a current role:

```
INFO  setup  RBAC self-check passed: every permission the controller's RBAC markers declare is granted
$ kubectl get platformagent platform-agent -o jsonpath=...     → Ready=True reason=Reconciled   (no Degraded)
```

Skew staged by removing `patch` from the `policy/poddisruptionbudgets` rule of `kubeagents-manager-role` (JSON patch on the live role, original saved first), then `rollout restart` of the operator to stand in for a pod reschedule:

```
17:23:36  ERROR  setup  RBAC self-check failed: the ClusterRole bound to the operator is older than this image;
                        re-apply k8s-operator/config (make deploy) or run helm upgrade so RBAC and image ship together
                        {"denied": ["patch poddisruptionbudgets.policy"]}
17:24:03  Degraded=True reason=RBACIncomplete
          message=The operator is denied permissions its RBAC markers declare: patch poddisruptionbudgets.policy; the ClusterRole bound to the operator is older than this image; re-apply ...
          platform-agent-gateway 3/3 Running     (nothing withheld; the PDB step logged its forbidden error as before)
```

Role restored at 17:24:43. Polled every 30 s: one reconcile completed at 17:25 with no error, then **zero reconciles for the next four minutes** (no loop). At 17:28:36, five minutes after boot, the ticker logged `rbac-selfcheck  RBAC self-check result changed  {"denied": []}` with no reconcile involved, and the requeued reconcile at 17:29:5x removed the condition. An earlier build of this branch that re-probed lazily inside `Reconcile` was run through the same cycle first: it detected identically, but after the restore the condition stayed until the requeue re-probed, and a build before the requeue existed left it standing indefinitely once the reconcile had completed, which is what the review findings below changed.

A control against the currently deployed `dns-endpoint-7dac349` operator was not run: it predates #733 and applies no PDB, so the staged verb is one it never uses.

The `check-img` guard needs no cluster; it was exercised from the checkout for every reference shape (see the Makefile tests).

Cleanup after each cycle: ClusterRole rules verified identical to the saved original, operator restored to `dns-endpoint-7dac349` (1/1, agent pod 3/3, CR `Ready=True`), every throwaway image tag deleted from Artifact Registry, live-test lease released.

## Self-Review

Passes: `review-adversarial` and `review-docs-drift`, each in its own subagent handed the diff range, the skill path, and the gate output, told to derive intent from the diff and not read commit bodies or plan files. Docs-drift was re-run in a fresh subagent after the first round's fixes; the adversarial pass was not re-run because its findings were addressed by a guard in `updateStatusReady`, a requeue, a timeout, tests, and prose, each of which the new tests or the live run above exercise. Neither pass had a cluster.

- **HIGH, CONFIRMED (adversarial).** With a preserved `Degraded/RBACIncomplete` condition, `updateStatusReady`'s unchanged-check never held, so every successful reconcile wrote status and re-enqueued the agent through the unfiltered `PlatformAgent` watch: a tight loop for as long as the skew lasted, measured at one status write per pass. Fixed: the preserved condition counts as unchanged; `TestAStandingRBACDenialDoesNotWriteStatusEveryPass` asserts zero writes over three passes; the live run showed zero reconciles across the five-minute window.
- **MEDIUM, CONFIRMED (adversarial) / Blocking (docs-drift).** The re-probe only ran inside a reconcile and nothing requeued, so on a quiet install a re-applied role would not clear the condition for hours while the docs said "every five minutes". Fixed: `RequeueAfter` of the interval while a denial stands, asserted in the reconcile test and observed live (cleared five minutes after the first completed reconcile).
- **Blocking (docs-drift, re-run).** While the reconcile itself is failing on the missing verb, the retry follows controller-runtime's default backoff (capped near seventeen minutes), not the five-minute requeue. Fixed in prose: the operator page states both cadences; not fixed in code, because returning a requeue instead of the error would change the error semantics of every step.
- **MEDIUM, CONFIRMED (adversarial).** Merge-order dependency with #1211's `role.yaml` rules through the parity test. Not a code change: it is the test's purpose; recorded in Context so the second author knows.
- **Blocking (docs-drift).** The CRD status reference said `Degraded` carried only `InvalidGitRepoURL`; `deploy/kustomize.md` showed a bare `make deploy` that now refuses; two pages claimed a five-minute cadence the code did not deliver. All fixed, the last one in code as above.
- **LOW, PLAUSIBLE (adversarial).** The boot probe ran with no deadline before the health endpoints exist. Fixed: 30 s timeout; the error path already says the reconciler retries.
- **LOW, CONFIRMED (adversarial).** `check-img` had no test; the `!ownsCondition` yield path had no test; the Makefile comment claimed equivalence with the installer's stricter rule; the refused-tag list was restated in four docs. All fixed: Makefile tests, `TestTheRBACSkewYieldsToAnotherDegradedReason`, comment reworded as a deliberately looser denylist, list kept in the operator page and the Makefile only.
- **Advisory (docs-drift).** "every `PlatformAgent`" overstated the yield rule; the log line is startup-only; a dangling "those markers" antecedent; no identifier-sources row; section nested under webhooks; two bare `make deploy` steps in the webhook sections. All fixed. Aligning the example image name `kube-agents-operator` with `k8s-operator` across INSTALL and README was left alone: user-chosen paths, not a false statement.
- **Medium (kube-agents-bot, first review).** The reconcile-time re-probe was 179 sequential calls on the single reconcile worker with no deadline. Fixed: the checker is now a manager Runnable that re-probes on its own five-minute ticker under a 30 s deadline with 16 reviews in flight, and `Reconcile` reads only the cached answer; observed live (the ticker logged the change on its own). The finding's rate-limit premise does not hold here: controller-runtime v0.24.1's `GetConfig` sets `QPS = -1`, so there is no client-side token bucket or throttling log; the adversarial pass had checked the same line. Answered in the thread.
- **Rebase after approval (2026-09-05).** `main` took #913 (agent shell sandbox) after the `lgtm`, and the branch conflicted. Rebased onto `90dbeac7`; the one textual conflict was the status-conditions table in `platformagent-crd.md` (kept #913's `Ready` reasons and this PR's `Degraded` row). The PR's own tests then caught two semantic conflicts, fixed in the follow-up commit `9fd5edc1`: #913 added a `secrets get` marker, so `requiredPermissions` gains that tuple (the merge-order case the #1211 bullet above predicted, arriving from #913 instead); and #913's `checkShellSandboxKeys` step parks an agent without the sandbox keypair on a 30 s requeue, so the two reconcile-level tests seed that Secret the way `main`'s own reconcile tests do. No other production change. The Testing list above was re-run at the new head: `go build`/`vet`/`test ./...` green, `make manifests generate` and `make chart-check` no diff, the nine Makefile tests pass, `prettier@3.9.6` and `make docs-check` clean. The live run was not repeated: the rebased code paths it exercised are unchanged.
- **Adversarial angles with nothing found:** SSARs are not client-rate-limited (`QPS = -1` from `GetConfig`) and cost RTTs; `leases` is cluster-wide in the role so the namespace-less review does not false-deny; absent API groups (`metrics.k8s.io`, `networking.gke.io`) evaluate textually and come back allowed; no new grant (`system:basic-user` covers `selfsubjectaccessreviews`); `InvalidGitRepoURL` and `RBACIncomplete` do not flap; a nil checker clears a stale condition; bare `make deploy` is now refused by design and no CI caller uses it; no-magic-constants rule holds; red-to-green confirmed for the reconcile test.

## Risk & Rollout

The operator gains one probe of 179 `SelfSubjectAccessReview` calls at boot and every five minutes after, off the reconcile worker and bounded to 30 s; no new permission, no change to what it renders, no bail-out on denial. `make deploy` now refuses a floating tag by default; the override is one variable. Revert is a plain revert of the commit. Nothing has to happen at merge time; #1211 or this PR, whichever lands second, updates `requiredPermissions` to keep the parity test green.


bradhoekstra:
Fixed in 0c4c6e0e: the checker is now a manager Runnable that re-probes on its own five-minute ticker under a 30 s deadline with 16 reviews in flight, and `Reconcile` reads only the cached answer, so no reconcile pays for the round trips. Observed live: the ticker logged the change five minutes after boot with no reconcile involved, and the requeued reconcile then cleared the condition.

One correction to the premise: `ctrl.GetConfigOrDie()` in controller-runtime v0.24.1 sets `QPS = -1` (`pkg/client/config/config.go:101-104`, "Disable client-side ratelimer by default, we can rely on API priority and fairness"), so the clientset built from `mgr.GetConfig()` has no token bucket and there is no 7.5 s wait or throttling log. The sequential-on-the-worker and no-deadline points stood and are what changed.

dshnayder (approved):

bradhoekstra:
/hold
Serializing the merge queue around the flaky smoke gate — releasing one PR at a time so Tide stops forming multi-hour batches. This hold is temporary; I will `/hold cancel` it when its turn comes.

bradhoekstra:
/hold cancel

bradhoekstra:
/retest
The 12:42 smoke run died before the test container started: Prow pod phase Failed with a `NodeNotReady` event and no build log. Infrastructure, not this change.

bradhoekstra:
Red at this head was a Prow node loss (NodeNotReady, no build log); required checks green, no open threads.

PR #1241, opened 2026-09-04
feat(credential-proxy): derive the executable allowlist from the configured forge

AntonTyb:
## Context

Step 4 of [`docs/designs/multi-forge-support.md`](https://github.com/gke-labs/kube-agents/blob/main/docs/designs/multi-forge-support.md) §9, the broker's own surface. Fifth in the multi-forge stack, after #1133 (design, merged), #1150, #1159, #1161 and #1234.

**Stacked on #1234.** Review the last commit only — the rest of the range is its predecessors. It merges after they do.

The broker's executable allowlist was a compiled-in tuple naming `gh`, written twice: once in `credential_proxy.py`, which enforces it in the sidecar as uid 10001, and once in `credential_proxy_client.py`, the shim in the agent sandbox as uid 10000 that offers it. §5 of the design counts that duplication as part of the cost of a second forge — a shim narrower than the enforcer refuses a tool the install is entitled to, and a wider one turns a clear local refusal into a confusing remote one.

## What changed

`agents/platform/scripts/forge_clis.py` is the one module both sides import. The executables an install may run are the base set (`gcloud`, `kubectl`, `git`) plus the tool for each provider the operator declared in `spec.integration.git`.

**The table stays compiled in.** `CREDENTIAL_PROXY_FORGE_PROVIDERS` selects from `FORGE_EXECUTABLES` by provider *name* and cannot introduce a name that is not there. The naive reading of "the allowlist becomes install configuration" would put the allowlist itself in the environment, and that is a privilege-escalation regression: anything able to set one variable on the sidecar could then run any binary on it. The operator configures provider names, never binaries, and `mergeCredentialProxyEnv` reserves the variable so a plugin cannot set it through `spec.deployment.env`.

**Rendered onto both containers.** Sharing the module closes the two-files gap only if the two readers also share the input; they are in different containers. For the same reason `forge_clis.py` joins the sandbox image's shared-scripts allowlist in `deploy/sandbox/Dockerfile` — since #913 that image is built from an enumerated `COPY` list closed under import, and a module two baked scripts import but the image does not carry fails on its first line in a pod.

**`/v1/github/refresh` → `/v1/forge/refresh`**, provider in the body, dispatched through `FORGE_REFRESHERS`. Being in that table is not permission — the handler also checks the provider against the configured set. Both images are versioned independently (`CREDENTIAL_PROXY_IMAGE` pins the sidecar while the agent tracks a tag), so each direction of skew has its own answer: the sidecar keeps serving the old path, and the client retries it on a 404. Strictly on 404, and only for GitHub — a sidecar that serves the route answers 200/400/502 and never 404, and one old enough to lack the route is old enough to serve GitHub alone.

**Per-provider refresher, not a provider branch.** `github_token_refresh.py` runs in both roles, so the sidecar's table is the dispatch point. It refuses a non-GitHub provider on the direct path rather than asking Minty for a GitHub token in the name of a repository that is not on GitHub.

**`GitProvider.CLI`** records each forge's tool on the operator side, checked against the Python table by `test_forge_clis.py` so the two cannot drift silently.

Not here: the egress allowlist (rest of step 4), and the installer's and Terraform's GitHub App inputs, which land with `GitLabProvider` in step 5.

## Testing

Re-run on the rebased head (`e519413e`): `go build`/`go vet`/`go test ./...` in `k8s-operator` green; 1675 agent-script tests, the one failure being the pre-existing `ModuleNotFoundError: fastapi` on `test_session_kv_server`, present on the base tree; 1328 tests under `tests/`; `make docs-check` clean; `prettier@3.9.6 --check` clean on the changed Markdown and YAML.

### Live validation

**Not performed — no reachable installation.** The stack has had none since #1133. What this change needs to be exercised properly is a rendered Deployment whose sidecar and agent containers both carry `CREDENTIAL_PROXY_FORGE_PROVIDERS`, a `gh` invocation through `/v1/exec` on that install, and a refresh over `/v1/forge/refresh` — plus, for the skew path, a deliberately mismatched pair of agent and sidecar images.

The golden manifests cover the rendering, and the tests below cover the runtime behaviour in-process, but neither tells you whether the operator reconciled the change or the pod picked it up. A reviewer with an install should run `kubectl exec` into the agent container, confirm `gh pr list` still reaches the sidecar, and check the sidecar log for the `credential refresh` line on the new route.

## Self-Review

Both mandatory pre-PR passes ran in contexts that did not write the change, each given the diff range and nothing else. Merged dispositions:

**Adversarial pass** — one HIGH, three MEDIUM, four LOW. Three of the MEDIUMs were demonstrated by revert experiment rather than argued.

| Finding | Disposition |
| --- | --- |
| **HIGH** — a new agent image against an older sidecar 404s on every refresh, failing every git write path. The comment rejecting a fallback ("a retry would send every genuine refusal twice") was wrong: the sidecar never answers 404 for a route it serves. | **Fixed.** 404-conditioned retry on the legacy path, GitHub-only. Three tests: the retry, that a 400 is not retried, and that another provider is never retried on the GitHub route. |
| **MEDIUM** — deleting the sidecar's per-instance allowlist binding, or reverting the `/v1/exec` handler to the class attribute, left the whole suite green. | **Fixed.** `ConfiguredForgeWiringTest`, four tests. The allowlist check moved into `_refuse_unsupported_executable` so it can be driven with an executor built for a given forge; inlined, reaching it meant standing up the socket and a policy. Both reverts now fail. |
| **MEDIUM** — shortening `FORGE_REFRESH_PATHS` to drop the legacy alias left the suite green, because the auth test iterates the tuple. | **Fixed.** A routing test through `do_POST`, and both it and the auth test now name the two constants rather than iterating. Verified the revert fails. |
| **MEDIUM** — the sandbox shim reads `CREDENTIAL_PROXY_FORGE_PROVIDERS` from a container the operator never rendered it into, so it always fell back to the GitHub set. Invisible today; on a GitLab install it would refuse `glab` *and* offer `gh`. The design claimed the gap was closed. | **Fixed.** Rendered onto the agent container too, with `TestTheSandboxShimIsGivenTheSameForgeListAsTheSidecar` asserting the two agree. Design §5 corrected. |
| **LOW** — `GitProvider.CLI` is write-only and duplicates `FORGE_EXECUTABLES`, kept in step by a comment. | **Fixed.** A cross-language parity test in `test_forge_clis.py`, both directions. Verified it catches a provider added to the Go registry alone. |
| **LOW** — the non-GitHub guard sits after the brokered-path return, so it cannot fire from the sandbox. | **Kept, comment corrected.** By design: on the brokered path the script is a transport and the sidecar dispatches. The guard governs the direct path, which is the standalone deployment. Its test now clears `CREDENTIAL_PROXY_URL` explicitly rather than assuming it absent. |
| **LOW** — the new env entry was inserted between the `API_SERVER_KEY` comment and its subject. | **Fixed.** Comment moved to its subject; the new entry has its own. |
| **LOW** — `FORGE_REFRESHERS` respelled `"github"` as a literal. | **Fixed.** `forge_clis.GITHUB`. |

The pass separately confirmed by experiment the two properties the change exists for: the environment cannot widen the allowlist (`sh`, `/bin/sh`, `gh,sh` all contribute nothing), and deleting the `mergeCredentialProxyEnv` reservation fails its test. It also confirmed no install can silently lose `gh` — all six unresolvable paths fall back to the default provider.

**Docs-drift pass** — four Blocking, three Non-blocking. All fixed.

- The design's STATUS banner and §1 layer-3 description still said the credential plane was entirely GitHub-shaped, contradicting §5 of the same page. Both updated.
- `docs/README.md`'s map row still read "step 3 all but `resolver.py`, rest not". One cell edited, no re-alignment.
- `k8s-operator/api/v1alpha1/gitprovider.go` pointed at `credential_proxy.py`'s `FORGE_EXECUTABLES`; the constant is in `forge_clis.py`.
- §5's framing line was left in the present tense while the paragraphs under it moved to the past.
- Added a `docs/README.md` identifier-sources row for the executable set, whose source of truth moved in this change.
- `docs/designs/pr-comment-conversation.md`'s "the sidecar's executable allowlist … names GitHub" corrected.

It verified the "only `gcloud`, `kubectl`, `gh` and `git`" claims across five documents are still true at HEAD, and that every caller of `github_token_refresh.py` still reads true (the new parameter is keyword-only and defaulted). Its check that `forge_clis.py` ships was read off a Dockerfile that copied the directory wholesale; the rebase invalidated it — see below.

### Known gap, deliberately not closed here

`is_valid_repository` in `credential_proxy.py` is still a GitHub-only `owner/name` check and would refuse a nested GitLab group path. Widening it needs the provider's path-depth rules expressed in Python, which is `GitLabProvider`'s work in step 5.

### Rebase onto current `main` (2026-09-08, `663c58a7` → `72d8322d`)

Two findings, both from the 2026-09-07 round against a `main` that had moved under the branch. The
2026-09-08 round applied without conflict; the one file it shares with a new `main` commit is
`platformagent_manifests.go`, whose comment fix from #1199 I confirmed present in the rebased tree.
Gates re-run on the new head, Go and Python both.

| Finding | Disposition |
| --- | --- |
| `forge_clis.py` did not ship to the sandbox. #913 replaced the wholesale directory copy with an enumerated `COPY` list, and a new module two baked scripts import was not on it — so `credential_proxy_client.py` and `github_token_refresh.py` would both have failed on their import line in the sandbox pod. | **Fixed.** Added to the list with an entry in the comment above it explaining why it is stdlib-only. `test_sandbox_delivery.test_the_allowlist_is_closed_under_import` is the test that caught it, and it fails without the addition. |
| The branch's edit to `platformagent-split-broker.yaml` and its re-render of `CREDENTIAL_PROXY_TOKEN_FILE` under `credentialBrokerIsSplit` were against a layout #913 removed. | **Dropped.** The fixture is deleted on `main`, and the token file is now set unconditionally there. The five surviving golden manifests were re-blessed with `go test ./internal/testing/... -update`; their whole diff against the parent is the two `CREDENTIAL_PROXY_FORGE_PROVIDERS` entries this change adds. |

### What neither pass could cover

Neither ran against a running installation, so the HIGH finding's skew scenario is derived from `resolveCredentialProxyImage` and the pre-change `do_POST` rather than observed. Neither built the images, so "`forge_clis.py` lands in `/opt/defaults/scripts`" is read off the `COPY` lines and `test_sandbox_delivery` rather than from a built image.

---


dshnayder:
Non-blocking heads-up, nothing needed from you here — `forge_clis.py` is the right shape and I'm not asking for a change to it.

Recording one thing so it isn't a surprise later. When the VCS abstraction from [#1246](https://github.com/gke-labs/kube-agents/pull/1246) lands, the sandbox stops running forge CLIs entirely: it uses stock `git` for local work — no origins, no credentials — and reaches every forge through the broker's verbs. So the shim's list narrows to a constant rather than a derived one:

```python
# forge_clis.py — additive
SANDBOX_EXECUTABLES = ("gcloud", "kubectl")

# credential_proxy_client.py
SUPPORTED_EXECUTABLES = forge_clis.SANDBOX_EXECUTABLES
```

`allowed_executables()`, `FORGE_EXECUTABLES` and the whole `integration.git` API are untouched by that, and your invariant holds — one module, both sides of the boundary import it, no tuple written twice. It just ends up naming two lists.

That comes in our PR after your stack is on main, not as a change to this one. Flagging it because it's the one place `multi-forge-support.md` §5's "both shapes are supported and neither is preferred" and #1246's concept genuinely differ, and I'd rather that be on the record now than turn up as a diff narrowing something you'd just built.

fkc1e100:
Closing as stale. This executable allowlist PR has been open for 19 days with unresolved conflicts.

Part of the multi-forge work. If still needed, please re-evaluate against current architecture.

Issue #1128, opened 2026-09-01
A2A stage 1: the upstream PR series map (the mode switch is the gate)


bnaylor:
The A2A-over-JetStream design set merged in #930 (amended by #1084). This issue is the series map for landing the stage-1 implementation, so each PR can stay small and cite where it sits instead of re-arguing the whole. It is the up-front version of the map comment #720's re-cut needed post-hoc.

**The one ordering rule: the mode switch is the gate.** `docs/designs/spec-mode-switch.md` keeps the next stack dark — a normal install cannot tell it exists. Every component below renders (or arms) only under `spec.mode: next`, so the switch lands before any of them; a component PR arriving first would either ship always-on or invent a temporary gate it then has to remove. The switch PR itself renders nothing, which is exactly why it is reviewable alone: the spec it implements is already merged.

The series (sizes approximate; each PR body carries its own spec citation, dependency line, and live validation):

| # | content | gate | depends on |
| --- | --- | --- | --- |
| 1 | terraform: agent-GSA name override in full-install | n/a | — |
| 2 | terraform: `lifecycle.sh` vs an empty KMS target list on bash 3.2 | n/a | — |
| 3 | **the mode switch**: `spec.mode` enum + delivery contract; renders nothing (#1129) | is the gate | — |
| 4 | the `a2a/` Go module: lib, CLI, profiles, conformance suite (embedded JetStream, no cluster in CI), its workflow | inert — nothing imports or invokes it | — |
| 5 | NATS component + gateway Deployment + provisioning + ingress NetworkPolicy | `mode: next` | 3 |
| 6 | the agent pod joins the bus: env, egress, skill overlay, `a2a` binary in the image | `mode: next` | 3, 4, 5 |
| 7 | the chatops gateway program + Discord adapter | rendered by 5; spawn dark behind `A2A_SPAWN_SESSIONS` | 4 (open after 5) |
| 8 | worker adapter + spawn arming (SA, pods-only Role) + session-pod egress policy + session cap/quota | `mode: next` | 5, 7 |
| 9 | Hermes bridge sidecar + its design note | user-set sidecar field (see its PR body) | 4 |
| 10 | read-only web rail (dev tooling, port-forward only) | dev tool | 5 |

1–4 are independent and can all be open at once; merge order within the A2A set is 3 → 5 → {6, 8}; 9 any time after 4; 10 last and optional. PRs 1/2 are small terraform fixes useful outside A2A entirely.

[...]

PR #1256, opened 2026-09-06
feat(operator,agent): the agent pod joins the A2A bus under mode next

bnaylor:
## Summary

The agent pod's half of the A2A bus, sixth PR of the series #1128 tracks. Under `mode: next` the operator puts the bus address and the `worker` credential on the agent container and opens TCP/4222 to the NATS pods on both of the agent's egress policies; the image gains the `a2a` topics CLI in its own builder stage and an `a2a-topics` skill that the entrypoint overlays into the platform profile only when the managed `.env` says the install runs the next stack. A today install renders none of it — the goldens are untouched and the gating is asserted in both directions.

## Why This Change

PRs #1129/#1124/#1211/#1212 landed the switch, the bus library, the NATS/gateway render, and the gateway program. What is still missing is the agent's side: without this change a `next` install has a running bus the agent cannot reach (no env, and — worse — a default-deny egress policy with no 4222 rule, which does not refuse the dial but hangs it to the client timeout while the CR reads Ready). The skill is the discoverability half: the topic blackboard is only useful if the agent knows it exists, and only on installs where it does.

## What Changed

- `NATS_URL` / `NATS_USER` / `NATS_PASSWORD` on the agent container under `next`, from the `<name>-a2a-nats-creds` Secret #1211 provisions. Appended **after** the AgentPlugin env merge, with the three names reserved out of plugin env while the surface is up (an appended name duplicates rather than shadows, and server-side apply refuses a duplicate env key — the EVENT_WATCHER_ENABLED rule), and added to `SensitiveEnvVars` so the CR's own env cannot name them. The SecretKeyRef is `Optional` so the skew render stays inert on an install whose Secret never existed.
- A TCP/4222 pod-selector egress rule on **both** agent egress policies — rule 13 of the gateway policy and a new rule on the `egressPolicy: Allowlist` render (#963-lineage code). Either policy can be the only one present, so the rule exists in both, the same reasoning as the Allowlist policy's sandbox rule.
- `a2aAgentSurface()`: next yes, today no, and version skew **preserves** the surface, mirroring the reconciler's freeze of the A2A objects — fail-closed on the agent side would strand a frozen, running bus behind an agent that just lost its credentials and egress rule.
- `deploy/docker/Dockerfile`: an `a2a-builder` stage (the k8s-event-watcher pattern, including `GOTOOLCHAIN=local`) ships `/usr/local/bin/a2a` ungated — a binary is inert until something invokes it, and the thing that invokes it ships dark. The skill lands at `/opt/a2a-template/skills`, deliberately outside `/opt/platform-template`, and joins the baked-skill-tree integrity mechanism on the same terms as the existing three trees (SHA-256 manifest, root-chown barrier, fail-closed boot verification).
- `docker-entrypoint.sh` step 2.6a-bis: the mode probe (managed `.env` parsed as strict KEY=value, never sourced, `os.environ` cleared first so the container env cannot be a second delivery path for the mode) decides whether the overlay is handed to `sync_profile_skills` as an optional third source composed onto the staging tree — template and overlay install in one rename, and a flip back to today self-cleans on the next boot.
- The credential-boundary prose and code comments corrected for the fact this PR introduces. Six places stated absolutely that the gateway Pod holds no credential, which `next` falsifies: the `ShareProcessNamespace` comment (which argued the field was not load-bearing *because* nothing here holds a credential — it is load-bearing now, and a shared process namespace would publish the bus password to every other container through `/proc/<pid>/environ`), the pod `SecurityContext` comment, the Session KV block's "the two exceptions" count and its claim that neither reaches outside the Pod, `docs/security-requirements.md` §8 criterion 5, `credential-isolation.md`'s one-credential count, and `operator/index.md`. The rendered posture is unchanged; what changes is what the next author reads before editing it.
- Two review findings from the fork-side adversarial pass are part of this design rather than afterthoughts, and both have tests: the plugin-override finding (the reservation + `TestPluginCannotOverrideBusEnv`) and the overlay race (the staged composition + three overlay tests including an unstageable overlay and the flip-back self-clean).

**One honest limitation.** Since #913 every command the agent runs executes in the shell sandbox, which deliberately carries neither this binary nor any credential — so the skill's shell invocation cannot reach the bus on any current install, and the skill's own instructions tell the agent to say so and answer another way rather than work around it. The fix is moving the reader off the shell (an MCP tool in the agent process, where this PR's env already is); that conversion is queued separately. What must NOT happen is the binary or the `NATS_*` env appearing in the sandbox image — the persona README says so explicitly. The delivery mechanism this PR ships (template placement, mode-gated staged overlay, probe) is what that follow-up keeps.

## Context

- Series: #1128 (tracking issue). Depends on #1129 (mode switch), #1124 (a2a module — the Dockerfile builds `cmd/a2a`), #1211 (NATS render — the creds Secret and the NATS pods this PR's env and egress point at). All merged.
- #1211's Context recorded that this PR owes the Allowlist render a 4222 rule; delivered here.
- **Conformance decision** (per AGENTS.md's routing of security invariants to `tests/conformance/`): no new assertion owed. No invariant in the suite's contract moves — the today posture is byte-identical (goldens + both-direction gating tests), the mode-gated additions are asserted deterministically in the operator suite where the render logic lives, and the sandbox-image invariant (`test_B1_the_sandbox_image_ships_no_credentialed_cli`) is untouched because `deploy/sandbox/Dockerfile` is untouched. The full suite runs green against this branch (108 tests, the 12 known violations unchanged).
- Merge-conflict neighbors, not overlaps: #1247 (netpol guardrails on reconcile refusals) and #1195 (litellm netpol) touch the NetworkPolicy surface.
- Generated with the help of Claude Code.

## Testing

- `go test ./...` in `k8s-operator/` — full suite green, goldens unchanged.
- New/ported operator tests: `TestBuildPodTemplateSpecBusEnvGatedOnMode`, `TestBuildNetworkPolicyBusEgressGatedOnMode`, `TestAgentEgressPolicyBusRuleGatedOnMode`, `TestPluginCannotOverrideBusEnv`, `TestSkewPreservesTheAgentBusSurface`. Mutation-verified red/green: gating the Allowlist rule off, removing the plugin-env reservation, and removing `Optional` each redden the suite.
- `TestTheProcessNamespaceIsUnsharedOnEverySpecShape` gains a `mode: next` row — the shape `ShareProcessNamespace` actually guards, which had no case. The row is non-vacuity-checked: it fails unless `NATS_PASSWORD` is on the `platform-agent` container as a SecretKeyRef, so a bus surface that stopped rendering cannot leave the assertion passing while asserting nothing. Both mutations run: `ShareProcessNamespace: ptr.To(true)` reddens all three rows, and disabling the bus env block reddens the new row alone on the non-vacuity check.
- `tests/test_docker_entrypoint.py`: 77 pass, including `A2AModeProbeTest` (7 probe cases: fail-closed unknown mode, exit 2 for a missing reader, inherited-env cannot decide the mode, hostile value is data not shell) and the three overlay-staging tests. Mutation-verified: rewriting the overlay to a non-staged post-rename copy reddens `test_an_unstageable_overlay_keeps_the_existing_copy`.
- `tests/test_mode_grep.py` green — `KUBEAGENTS_MODE` still has exactly its two allowed code sites (the probe hands the file to `runtime_mode` rather than naming the key).
- `python3 tests/conformance/run.py`: 108 tests, OK (12 expected known-violation failures, unchanged).
- `make docs-check`, `make chart-check`, pinned prettier 3.9.6 `--check` on every changed Markdown file: green. AGENTS.md context budget: 37,741 of 38,000 chars — 12 below main. Corrected in `0d76c239`: the a2a bullet now *drops* the "Nothing imports it yet" clause this branch retires rather than rewording it. The earlier reword was net **+11**, not the −8 this line used to claim, and against a main with ten characters of headroom the merge landed at 38,001 — one over, taking `Documentation Checks` and `Run Python Unit Tests` red together. Recorded on #1223 as a fourth instance of that margin.

### Live validation

Two venues, because they answer different questions.

**envtest, against a real kube-apiserver (1.36.2)** — admission, reconcile, and server-side apply of the actual object:

| case | observed |
| --- | --- |
| `mode: next` | the three env vars on the agent container, `NATS_PASSWORD` a SecretKeyRef to `<name>-a2a-nats-creds/worker-password` with `optional=true`; gateway netpol carries the 4222 rule, two-label pod selector, no IPBlock |
| `mode: today` | none of the three; no 4222 rule |
| `mode: quantum` | refused at admission: `spec.mode: Unsupported value: "quantum": supported values: "today", "next"` |
| next + `egressPolicy: Allowlist` | the 4222 rule on the Allowlist render too, same peer |
| next + an AgentPlugin naming `NATS_URL` | reconcile converges, the operator's address wins, the plugin's own unrelated variable survives |

The control for that last row is what makes it mean something: with the reservation disabled, the same run fails with `failed to create typed patch object (...Deployment): .spec.template.spec.containers[name="platform-agent"].env: duplicate entries for key [name="NATS_URL"]` and never converges. (The first attempt at this case was a false pass — the test plugin had no `spec.agentRef`, so it was never attached; caught by noticing the plugin's own variable missing from the render.)

**A real GKE cluster (`bnaylor-ka-test`, under our test lease)** — the deployment-side behaviour envtest cannot show. The install was three weeks stale, so branch images were built for all three targets (`platform-agent`, `credential-proxy`, `k8s-operator`) and the branch CRDs applied first.

- **Today baseline:** zero NATS env, zero 4222 rules, `a2a-topics` absent from the platform profile, and the no-bus path exactly as designed — `a2a: NATS_URL is not set`. The `a2a` binary is present, root-owned and runnable, confirming it ships ungated.
- **Image integrity, at boot:** `[SKILL-PROVENANCE] verified 1 files under /opt/a2a-template/skills`. The new tree carries its manifest, is root-owned, and the fail-closed check verifies it before anything is copied into a profile — the pod starting at all is the assertion.
- **Flip to `next`:** managed env flips to `KUBEAGENTS_MODE=next`; the config-hash annotation rolls the pod; the entrypoint logs `Next stack: overlaying the A2A topics skill into the platform profile` (and `no overlay` for the cluster profiles, so the scoping holds); `a2a-topics` appears in the platform profile; the NATS StatefulSet reaches 1/1; the three env vars render with `optional=true`; the gateway netpol carries the 4222 rule.
- **The credential works end to end:** `a2a topics list` from inside the agent container connected to NATS, authenticated as `worker`, and received a real JetStream API response. An auth failure looks different, so this exercises Secret → SecretKeyRef → container env → client → authenticated connection across the 4222 rule.
- **Flip back to `today`:** the overlay self-cleaned — `a2a-topics` gone, all 41 real skills still present, which is the overlay-race fix doing its job — env count back to 0, a2a objects reaped. The creds Secret and JetStream PVC survive, which is #1211's intended flip-back semantics.

**What this did not cover, and one thing found along the way.** NetworkPolicy *enforcement* was not measured: this cluster's dataplane does not enforce, so the evidence above is that the rule renders correctly and does not block the dial, not that the fence bites — the enforcement evidence remains the Dataplane V2 run recorded with #1211. The topics list returned `stream not found` because the A2A provision Job fails on this cluster with `stat .: permission denied`; that Job is #1211's code and untouched here, but it means a fresh `next` install never provisions its topics, so I am filing it separately rather than burying it. The test cluster was restored (mode removed, image tags reverted, orphaned objects cleaned) and the lease released.

## Self-Review

Both pre-PR passes ran in fresh subagent contexts handed the diff range and nothing else, per `.agents/rules/pre_pr_review.md`. One merged list; every finding's disposition:

- **[adversarial, HIGH — fixed]** The skew surface was not inert: a today-lineage install that hit skew would render a *required* SecretKeyRef to a Secret that never existed, and under strategy Recreate the rollout ends in CreateContainerConfigError — an agent outage caused by the helper meant to prevent one. Fixed with `Optional: true` on the ref; pinned in `TestBuildPodTemplateSpecBusEnvGatedOnMode`.
- **[adversarial, HIGH — fixed]** `/opt/a2a-template` escaped all three image-integrity mechanisms covering every other prompt-reaching tree (manifest, root-chown barrier, boot verification). The docs-drift pass found the same gap independently from the doc side. Fixed by adding the tree to all three on the same terms as the existing trees.
- **[adversarial, MEDIUM — fixed]** A plugin naming any `NATS_*` var wedged the whole reconcile: the appended operator values duplicate the plugin entry and server-side apply refuses duplicate env keys (the reviewer verified the rejection empirically on envtest). The original test comment described "last value wins" semantics the apply path never reaches. Fixed with the surface-gated reservation of the three names out of plugin env; the test now asserts no-duplicates, operator values, an unrelated plugin var surviving, and today's plugin env untouched.
- **[adversarial, LOW — fixed]** `a2a-builder` was missing the `GOTOOLCHAIN=local` guard of the stage it mirrors; added.
- **[adversarial, LOW — rejected with evidence]** Claim: the overlay's atomic-composition control is unpinned ("deleting the control leaves the suite green"). Refuted by mutation: rewriting `sync_profile_skills` to rename the template and copy the overlay into the live directory afterwards reddens `test_an_unstageable_overlay_keeps_the_existing_copy` (run performed, 1 failure). The named test alone does not pin the control, but the suite does, which is the bar.
- **[adversarial, LOW — fixed]** "The sidecar reads the same three" misstated pod semantics (env is per-container) in a code comment and the persona README; both reworded to "declares the same three against the same Secret".
- **[docs-drift, Blocking ×6 — fixed]** spec-mode-switch status line and two docs/README.md status cells falsified by this branch (one, the chatops-gateway row, falsified earlier by #1211 — fixed in passing); the `/opt/a2a-template` integrity gap (same as the adversarial HIGH); missing mode-next carve-outs on security-and-iam and credential-isolation, added in the pages' existing one-clause style; the map's tree section and section placement for the two new docs.
- **[docs-drift, Advisory ×3 — fixed]** sidecar wording (as above); AGENTS.md's "Nothing imports it yet" dropped rather than reworded — the clause is exactly what this branch retires, and at ten characters of headroom the budget has no room for a longer sentence in its place; concepts/skills.md gains a one-line pointer to the persona README from its replace-on-every-start paragraph.
- **[reviewer finding, #1256 review — fixed]** The pre-PR docs-drift pass reached `security-and-iam.md` and `credential-isolation.md` but stopped at the site: the code comments those pages' guarantee rests on, and `docs/security-requirements.md` §8 criterion 5, still said the Pod holds no credential. Verified against source before changing anything — both cited comments are in `buildPodTemplateSpec`, the same function that appends `NATS_PASSWORD`, so they describe the very Pod that carries it. Widening from the two instances to the class turned up three more the review did not name: the Session KV block's "the two exceptions" count (three now, and the third is the one that *does* reach outside the Pod, which is the property that block claims for all of them), `credential-isolation.md`'s one-credential count in its lead, and `operator/index.md`'s flat "the gateway holds no credential". All six corrected; behaviour pinned by the new `mode: next` row above.
- **[reviewer note, not a finding — measured]** The AGENTS.md merge-ordering hazard is currently clear: `main` has moved 21 commits and grown its own `AGENTS.md` by 9 characters since, and the merge result measures 37,902 chars for `AGENTS.md` (37,978 charged against the 38,000 budget, 22 to spare). `main` was not merged in — Prow already tests this branch as merged into `main`, so the hazard is covered where it would bite.
- **Stated limits carried from the passes:** no image was built and no container ran — the Dockerfile stages and entrypoint are verified by the extracted-function tests and by reading; SSA duplicate-key behavior was verified on envtest kube-apiserver 1.36.2 only; the deployment-side behavior is what Live validation above owes.

## Risk & Rollout

Everything operator-side is gated on `a2aAgentSurface`: a today install's renders are byte-identical (goldens) and the only unconditional deltas are an inert binary in the image and a skill tree outside every profile template. The new failure modes considered and closed: plugin env duplication (reserved), skew with no Secret (Optional ref), overlay landing in another replica's swap window (staged composition). Revert is a clean `git revert` of the three commits; nothing has to happen at merge time. The images are not in the release pipeline at this stage — they join it at stage-2 graduation, per the series posture stated in #1211.


bnaylor:
Smoke is failing repo-wide again, so I am not retesting this PR — flagging with the data instead.

Completed `pull-kube-agents-smoke-test` runs since 18:00Z yesterday, by PR author:

| author | pass | fail |
| --- | --- | --- |
| shalinibhatiapl | 0 | 3 |
| fkc1e100 | 0 | 2 |
| sergiu1spiridon | 0 | 2 |
| kirill-kosov | 1 | 2 |
| bnaylor | 2 | 2 |
| AntonTyb, dshnayder, eLeontev, jayantid, lapis2002 | 0 | 1 each |
| bradhoekstra, clomp42, dependabot | 1 each | 0 |

Ten authors seeing failures, and today's repo-wide totals are 3 passes to 12 failures with 34 aborted. That is not a property of any one change, and this PR's diff is operator render plus its tests.

Not retriggering: a retest costs the shared build cluster a slot and answers nothing while the base rate is 20%. Holding until the pass rate recovers, then retesting this and #1271 together. @jayantid — you were closest to the last smoke recovery; is this the same failure or a new one?

jayantid:
Stale red from the seeded-fleet fixture outage (#1278): the last run (build 2097021780592955392, leased kube-agents-evals-25) collapsed only on the crashloop trio with `required phrases absent: ['OOMKilled']` — the seeded-a fixture is repaired on all 30 pool projects, retesting.

/retest

mplakhtiy (commented):
Reviewed at head `0d76c239605fcbf6ecb44a47fdf79b7390ef93b0`. Produced with the help of Claude Code.

Both findings anchor to lines this diff does not touch — they are stale precisely because the change did not edit them — so they are here rather than inline.

Findings 1 and 2 propagate to #1257, #1258 and #1319, which extend the same agent-pod credential surface on top of this one.

**1. `k8s-operator/internal/controller/platformagent_manifests.go:2499` and `:2517` — the pod spec's "nothing here holds a credential" comments are now false.**

Line 2499:

> `// No ShareProcessNamespace. Nothing in this Pod holds a credential any more, so the field is not load-bearing here — it stays unset because a Pod that shares its process namespace hands every container's /proc/<pid>/environ to every other one, and the next container added here should not inherit that by default.`

And beside `RunAsUser` at 2517:

> `// Every container in this Pod runs as the agent image's user; none of them holds a credential.`

Under `mode: next` this change puts the NATS `worker` password on the `platform-agent` container by SecretKeyRef (`platformagent_manifests.go:2301`). The field itself stays unset, so the rendered posture does not change — but the comment is the only thing standing between a future change and `ShareProcessNamespace: true`, and as written it argues for flipping it: nothing holds a credential, so sharing is free.

Failure scenario: a later change adds a sidecar that wants a shared PID namespace, reads the comment, and sets the field. On every `next` install the bus password then becomes readable through `/proc/<pid>/environ` to every other container in the pod, including anything a user declares in `spec.deployment.sidecars`. Nothing else in the pod re-asserts that boundary.

This PR added the equivalent carve-out to `docs/site/src/content/docs/reference/security-and-iam.md`; the code comment the doc's guarantee rests on was missed.

Suggested wording for 2499, keeping the field unset for the reason that still holds:

> `// No ShareProcessNamespace. Under mode: next the agent container carries the`
> `// A2A bus credential (NATS_PASSWORD, by SecretKeyRef below), so the field is`
> `// load-bearing again: a Pod that shares its process namespace hands every`
> `// container's /proc/<pid>/environ — that credential included — to every other`
> `// one, spec.deployment.sidecars entries among them.`

And for 2517, dropping the second clause: `// Every container in this Pod runs as the agent image's user.`

**2. `docs/security-requirements.md:162` — Acceptance Criteria §8, criterion 5 is falsified.**

> "…the operator-generated agent sandbox receives no credentials or Kubernetes ServiceAccount tokens through environment variables or mounted filesystems. This holds fully for the `<agent>-shell` Pod, which runs model-authored code. The gateway's `platform-agent` container is the one exception: it mounts an audience-bound projected ServiceAccount token so it can authenticate to the broker across the network — a deliberate trade described in section 6;"

Under `mode: next` there is now a second exception on that same container: `NATS_PASSWORD`, injected by SecretKeyRef from `<name>-a2a-nats-creds`. This is the acceptance-criteria list rather than narrative prose, so the count is load-bearing.

The docs-drift pass reached `security-and-iam.md` and `credential-isolation.md` and updated both; this file was not touched. A clause in the same one-sentence style those two use would settle it — for example, after the existing exception: "and, under the unsupported `mode: next` toggle, the A2A bus's `worker` password by SecretKeyRef, which appears and disappears with the A2A stack".

### Note, not a finding

`scripts/check_context_budget.py` reports 37,975 of 38,000 characters on this branch — 25 characters of headroom. Any other pull request that adds a line to `AGENTS.md` and merges first turns this one red. That is a merge-ordering hazard rather than a defect in this change.

AGENTS.md
```diff
@@ -13,7 +13,7 @@ This repository contains the Kubernetes Agentic Harness (`kube-agents`). It is a
...
 - `.agents/rules/`: Repository-level rules an agent follows, one file per family and none shipped in the agent images — `core_engineering.md` for the code itself, `github_actions.md` for workflow authoring, `pre_pr_review.md` for the mechanics of the two pre-PR passes. This file states each rule and links there for the form it takes; the split keeps `AGENTS.md` inside the context budget `scripts/check_context_budget.py` enforces.
-- `a2a/`: Go module for the agent-to-agent bus — wire-protocol library and `a2a` topics CLI per `docs/designs/spec-a2a-payloads.md`, plus agent profiles. Nothing imports it yet.
+- `a2a/`: Go module for the agent-to-agent bus — wire-protocol library and `a2a` topics CLI per `docs/designs/spec-a2a-payloads.md`, plus agent profiles and persona.
```
mplakhtiy, line 16: Leaving a placeholder unresolved inline comment so the pr does not merges when I approve it until my previous comment gets acknowledged.

mplakhtiy (approved):

AGENTS.md
mplakhtiy: @bnaylor the pr is approved, feel free to resolve this comment. Just wanted to pause it so you have a chance to read/act on above review comment from Claude.

bnaylor: Both findings were right, and both are fixed in `b664502a`. Thanks for holding the merge — the second one in particular was a count I would not have re-checked.

**Finding 1 — the pod spec comments.** Verified before changing anything: `platformagent_manifests.go:2499` and `:2517` are inside `buildPodTemplateSpec`, the same function that appends `NATS_PASSWORD` at `:2301`, and `buildBaseContainers` is called from that function with those `envVars`. So the comments describe the very Pod that carries the credential. You are also right that the rendered posture does not move — `ShareProcessNamespace` stays unset either way — and that the comment was the whole guard, since it argued the field was not load-bearing *because* nothing here holds a credential.

Fixed close to your wording, with the today branch kept explicit so the comment stays true in both modes. `:2517` lost the second clause, as you suggested.

**Finding 2 — criterion 5.** Fixed with a clause in the one-sentence style the other two pages use, near enough to your text: the container now "carries the exceptions", the projected token first, then the bus password by SecretKeyRef under `mode: next`.

**Three more of the same class, which I went looking for after your two.** The class is "absolute statements that the gateway Pod holds no credential", and it is wider than the diff:

- `platformagent_manifests.go:1961` — the Session KV block calls `SESSION_KV_API_KEY` and `SESSION_KV_SALT` "the two exceptions" to no-credentials-in-the-sandbox, and closes with "Neither grants access to any cloud API, any repository, or anything outside the pod, which is the property the isolation boundary protects." `NATS_PASSWORD` is appended to that same `envVars` slice ~340 lines later, so the count is three — and the third is the one *without* that property, since it authenticates to the bus over the cluster network. This is the sharpest one, because it is the block that states the boundary rather than merely relying on it.
- `credential-isolation.md:10` — "does hold one credential, the audience-bound token it presents to the broker". Two under `next`. Same shape as your finding 2: a load-bearing count in a file the docs-drift pass already touched, on a line it did not.
- `docs/site/src/content/docs/operator/index.md:35` — "The gateway holds no credential and executes nothing the model wrote." Now defers the credential half to the page that owns the topic.

**One I looked at and left alone**, so it does not read as a miss: `platformagent_manifests.go:4925`, "The gateway holds no credential and calls nothing on the broker's credential surface". That clause is already loose on `main` independently of this PR — the gateway holds the audience-bound broker token today — and the operative claim carrying rule 12 is the second half, about the broker's credential surface. Rewriting it is a `main` correction rather than this branch's, and I would rather not fold an unrelated one in here.

**Behaviour pinned.** `TestTheProcessNamespaceIsUnsharedOnEverySpecShape` covered "no harness configuration at all" and "broker in its own Pod" — neither of which holds a credential, so the one shape where the field is load-bearing had no case. It gains a `mode: next` row, guarded against passing vacuously: the row fails unless `NATS_PASSWORD` is on the `platform-agent` container as a SecretKeyRef, so a bus surface that stops rendering reddens the test instead of quietly hollowing it out. Both mutations run rather than reasoned about:

```
# ShareProcessNamespace: ptr.To(true)
--- FAIL: .../no_harness_configuration_at_all
--- FAIL: .../broker_in_its_own_Pod
--- FAIL: .../mode:_next,_where_the_agent_container_holds_the_bus_credential

# bus env block disabled
--- FAIL: .../mode:_next,_where_the_agent_container_holds_the_bus_credential
    no NATS_PASSWORD on the platform-agent container: this shape holds no bus
    credential, so the shared-process-namespace assertion beside this one would
    pass vacuously
```

**Checks**, all at `b664502a`: `go build ./...`, `go vet ./...` and `go test ./... -count=1` in `k8s-operator/` all green (5 packages ok, goldens unchanged); `python3 tests/conformance/run.py` 108 tests OK with the 12 known violations unchanged; `make docs-check` green; prettier 3.9.6 `--check` clean on the three changed Markdown files. `make test-python` has 5 failures in `tests/` — `test_eval_dashboard_refresh` (3) and `test_generate_release_sbom` (2) — which reproduce identically with this commit stashed, so they are this macOS box rather than the change.

**Your budget note, measured rather than assumed.** `main` has moved 21 commits since your review and grown its own `AGENTS.md` by 9 characters, so the hazard was worth checking: the merge result measures `AGENTS.md` at 37,902 chars, 37,978 charged against the 38,000 budget — 22 to spare. Still clear, still thin. I did not merge `main` in, since Prow tests this branch as merged into `main` and that is where the hazard would bite.

Findings 1 and 2 propagate to #1257, #1258 and #1319 as you said; I will carry the class — the Session KV count especially — onto those rather than just the two instances.

PR #1040, opened 2026-08-28
feat(gateway): add Microsoft Teams Bot Framework relay adapter and credential proxy support

fkc1e100:
## Summary

Implements the Microsoft Teams ChatOps adapter for the Hermes chat gateway and adds credential proxy relay support with Microsoft Entra ID token lifecycle management. Provides secure, single-tenant webhook ingress, user access gating, Adaptive Cards v1.5 rendering, and reliable event queueing with ACK/NACK semantics.

## Why This Change

Enables the Kubernetes Agentic Harness chat ingress to communicate bi-directionally with Microsoft Teams channels and direct chats while preserving credential isolation and strict tenant security boundaries.

## What Changed

- **Teams Bot Framework Relay Adapter (`agents/platform/scripts/teams_relay_patch.py`)**:
  - Webhook activity consumer on `/api/v1/teams/events` and `/v1/chat/teams/events`.
  - Single-tenant isolation: enforces `TEAMS_TENANT_ID` matching against `channelData.tenant.id`.
  - User authorization: checks user IDs and Entra ID UPN emails against `TEAMS_ALLOWED_USERS` (with `TEAMS_ALLOW_ALL_USERS` toggle).
  - Adaptive Cards v1.5 JSON rendering with markdown fallback and Action button support.
  - Typing indicator activity support (`{"type": "typing"}`).
- **Credential Proxy Relay (`deploy/docker/credential_proxy.py`)**:
  - `TeamsRelay` class managing Entra ID client-credentials OAuth token caching and background refresh.
  - Event queueing with ACK/NACK settle semantics (`/v1/chat/teams/events/ack`, `/v1/chat/teams/events/nack`).
  - Secure HTTPS Bot Framework API dispatching (`/v1/chat/teams/api`) restricted to trusted Microsoft domains.
- **Gateway Configuration & Integration**:
  - Registered `teams` platform in `agents/chat/config.yaml` with MCP router, Kanban, and Memory toolsets.
  - Wired `teams_relay_patch` into runtime initialization in `agents/platform/scripts/sitecustomize.py`.
- **Testing**:
  - Added unit tests in `test_teams_relay_patch.py` and `test_credential_proxy.py`.
  - Added integration seam tests in `tests/integration/test_seam_teams_endpoints.py`.

## Context

Part of Epic #1038 (Native Microsoft Teams ChatOps Integration).
Closes #1034.
Closes #1035.

## Testing

[...]

PR #1257, opened 2026-09-06
feat(a2a): the Google Chat adapter — production ingress through the credential proxy relay

bnaylor:
## Summary

The Google Chat adapter for the A2A gateway, validated end to end against a real Chat app in a project we own. Events arrive from a dedicated A2A subscription on the Chat topic, pulled through a second `GoogleChatRelay` instance in the credential proxy; posts, edits, roster reads and `openDirect` ride the proxy's existing `/v1/chat/api` passthrough. The gateway pod holds no cloud credential: it authenticates to the proxy with a projected ServiceAccount token under a third audience that confers the `a2a-chat` role, so the one Chat credential in the deployment stays where #913 put it. Requester identity is the Google-asserted sender email, gated by the operator-pinned allowlist, carried into the advisory `authority` block as `verifiedBy: chat-event-topic-iam`. Live traffic disagreed with the original fixtures in four places, and in each the adapter was changed to match what Chat sends; the captured payloads are now the fixtures.

## Why This Change

The gateway shipped with Discord, whose mapping table cannot assert a real principal. Google Chat is the backend where identity means something: the asserted email is the same string as the cloud principal, so once the authority work makes principals enforceable, a Chat-originated requester already carries the subject the cluster would check. The spec also claimed gchat requests carry a Google-signed token and used `verifiedBy: "chat-signed-token"` as its example; neither is true of the Pub/Sub shape we ship, and the correction rides here rather than trailing.

## What Changed

- `a2a/gateway/gchat.go` (+tests, +`testdata/gchat/`): the adapter. It decodes both event layouts Chat publishes: the legacy `{"type":"MESSAGE",…}` the e2e forge uses, and the Workspace add-on event object (`chat.messagePayload`, no top-level type) that the real app sends. Every event that is not a turn is acked with a logged reason. Conversation keys `gchat:spaces/A/threads/B`, `gchat:dm/spaces/A`, `gchat:space/spaces/A`; DM replies follow the thread of the latest ask while the session stays space-wide. `argumentText` is a pointer because a bare mention in a space is present-but-empty. The id-to-email pair from each inbound event bridges what the app credential cannot see: the roster substitutes learned emails for `users/{id}`, and `openDirect` resolves an email to the id Chat accepts. Pub/Sub poison is acked away; ingress is at-most-once by recorded decision; dedupe and learned maps are bounded. Outbound text is defanged visibly for both sequences Chat parses out of a plain message, `<users/…>` and `<url|text>`: only the `<url|text>` the adapter itself generates from a markdown link reaches Chat live, and the link regex refuses `<`, `>` and `|` inside the URL so a crafted markdown link cannot close that sequence early and pick its own display text.
- `a2a/gateway/gateway.go`, `config.go`, `relay.go`: per-backend identity resolution (gchat is the identity function over the email, gated by `A2A_GCHAT_ALLOWED_USERS` or `A2A_GCHAT_ALLOW_ALL_USERS`), `verifiedByFor()`, one backend per process, a visible once-per-sender drop notice on every backend, and `GoogleChatSpec.Mode`'s default-vs-debug display split carried as `A2A_CHAT_DISPLAY_MODE`.
- `agents/platform/scripts/credential_proxy.py` (+tests): a second relay instance armed by `A2A_GOOGLE_CHAT_SUBSCRIPTION_NAME`; routes `/v1/chat/a2a/events[/ack|/nack]` require the new `a2a-chat` role (third audience, `CREDENTIAL_PROXY_A2A_CHAT_AUDIENCE`, same unset-means-no-split story as the chat audience); `/v1/chat/api` admits both chat roles; `required_role` becomes `required_roles` because a shared route needs a tuple. `_validate_route_roles` runs at import and refuses the three ways that table can be mis-edited into widening a route — a roles entry that is not a tuple (a bare string makes the role check a substring test, and `"chat"` is a substring of `"a2a-chat"`), a role that is not a role, and any entry an earlier prefix already matches (which is what alphabetizing the table does to `/v1/chat/a2a/`). Every test module imports the proxy, so a mis-shaped table reds the pull request.
- `docs/designs/spec-chatops-gateway.md`: the adapter section, rewritten where live traffic contradicted it (wire shapes, roster, openDirect, DM threading, DM mentions), and the identity claim and `verifiedBy` example corrected: no Google-signed token exists on the Pub/Sub shape, so the field names the real mechanism, topic IAM plus Google Chat's assertion.
- `tests/integration/test_seam_a2a_chat_ingress.py`: seam tests over real HTTP, including the first pin of the `ROUTE_ROLES` prefix rule, in both directions.
- `k8s-operator/internal/controller/platformagent_manifests.go`: the one operator hunk, two names added to the credential proxy's reserved env list (`CREDENTIAL_PROXY_A2A_CHAT_AUDIENCE`, `A2A_GOOGLE_CHAT_SUBSCRIPTION_NAME`) so a CR cannot set them before the operator renders them, with a test.

**Not in this PR (operator wiring follow-up):** the operator does not yet render the gchat env, the projected a2a-chat token, the broker-side audience env, the proxy netpol edge, the allowed-caller entry, or the terraform A2A subscription and its IAM. Until that lands the adapter arms only by hand-set configuration, and arming it beside the legacy consumer double-answers every message, which the spec states.

## Context

- Answers the C2 card's two open questions in the spec: same app registration and topic with a dedicated subscription consumed through the proxy, and `verifiedBy: chat-event-topic-iam`.
- #1248 (Slack adapter) is a near-copy of the shared gateway infrastructure; this branch converges on its shapes. Whichever merges second widens the one-backend guard to three-way.
- #1040 (Teams adapter) touches the same adapter surface; overlap is merge-conflict, not duplication.
- No principal string form is coined here; the raw email feeds the existing `BuildAuthority` hashing seam, and the join with the shipped attribution path is by construction. The authority block stays advisory.
- Conformance (#1121), decided explicitly: no new `tests/conformance/` assertion in this PR. No operator-rendered artifact changes; the proxy's audience-split invariant extends to the new routes and is asserted at the HTTP layer in the seam test and now live; the drop-at-ingress invariant is asserted in Go. The operator-wiring follow-up re-answers this, and `test_D4_the_broker_token_is_audience_bound` gains a third member there.
- Open, not this PR's: which event layout the corp app in `kube-agents-autopush` publishes, and whether the Hermes `google_chat` adapter reads the add-on layout. `tests/e2e/gchat_agent_test.py` forges only the legacy one.

## Testing

- `a2a`: `go test ./...`, `go vet`, `gofmt` green. The gateway suite includes the gchat tests over an embedded NATS and an httptest fake relay, five of which run against captured real payloads.
- `agents/platform/scripts/test_credential_proxy.py`: 259 tests green. `tests/integration/test_seam_a2a_chat_ingress.py` and the legacy `test_seam_chat_ingress.py` green (1 pre-existing expected failure).
- `make docs-check` green at the head (295 files, map covers 262 tracked docs, `AGENTS.md` 37.8k with 0.0k headroom and untouched by this branch); prettier green on the changed Markdown, and no Markdown has changed since.
- `python3 tests/conformance/run.py`: 108 tests green (12 expected failures, all pre-existing bucket-3 markers).
- One intermittent failure was seen in a filtered run of the gchat tests (`-run 'TestNewDerivesTheBackend|Gchat'`) during the last review round and did not reproduce in eleven further runs of the same subset plus two full-suite runs; the failing test name was not captured. Recorded rather than hidden; if CI shows a flake in the gchat rig it is this.

### Live validation

Install: `a2a-next-dev` (`bnaylor-kagents-dev`, GKE Autopilot), operator `k8s-operator:a2-20260908-162517` scaled to 0 for the run because it re-renders both Deployments, credential proxy rolled to `credential-proxy:c2-20260909-141328` and the A2A gateway to `gateway:c2-20260909-141328-r4`, both built from this branch (final head) merged with a fresh `upstream/main`. Lease held throughout (`cluster-under-test` on the cluster). A real Chat app in the same project publishes to `topics/a2a-chat-validation`; the proxy was armed with `A2A_GOOGLE_CHAT_SUBSCRIPTION_NAME=…/a2a-chat-validation-sub`, `CREDENTIAL_PROXY_A2A_CHAT_AUDIENCE`, and the gateway KSA appended to `CREDENTIAL_PROXY_ALLOWED_CALLERS`; the gateway got a projected token at that audience, `A2A_GCHAT_RELAY_URL`, and the proxy netpol an ingress edge from the gateway pod. IAM: the agent GSA granted `roles/pubsub.subscriber` on the subscription. A human typed every inbound message from a Google Chat client; nothing was forged onto the topic.

What was observed, per layer:

- **Relay transport.** Proxy logged `A2A Google Chat relay enabled` and `GET /v1/chat/a2a/events 200` on every long poll; each real event was followed by one `POST /v1/chat/a2a/events/ack 200`.
- **Wire shape (the finding that reshaped the fixtures).** The first gateway build acked five real DMs with no log line: every payload was the add-on event object, not the legacy layout. Captured verbatim from a second subscription; scrubbed copies are `testdata/gchat/*.json`. After the fix the same shapes became turns.
- **Ingress to authority.** TASKS #477 request envelope: `authority.requester = {principal: hmac:b632…, backend: gchat, subject: hmac:b632…, verifiedBy: chat-event-topic-iam}`, principal equal to subject because the email is both, and equal to the pseudonym on yesterday's Discord envelopes for the same person, which is the cross-surface join. Audience: `gchat:dm/spaces/_ia8wqAAAAE`, kind `dm`, roster of one entry equal to the requester (the members API returned only `users/{id}`; the learned bridge resolved it).
- **A2A leg and resolve-back.** With `A2A_DEFAULT_ADDRESSEE=session`, three asks spawned session workers (A2's `worker-next` image, static `worker` login) that published submitted, working, thinking, progress, result, completed; the gateway edited the placeholder in place and posted the answer in the DM. One `contextId` persisted across all tasks in the conversation, including a follow-up typed inside a thread.
- **Unlisted-sender drop.** With the allowlist set to exclude the tester, two DMs produced two `dropping message from unverified sender` warnings, two acks, exactly one `POST /v1/chat/api` (the single notice), and no new TASKS sequence. Tester confirmed one notice.
- **Allow-all, DM threading and display split.** With `A2A_GCHAT_ALLOW_ALL_USERS=true` and `A2A_CHAT_DISPLAY_MODE=debug`, the ask made inside a DM thread was admitted and answered, and the placeholder and answer rendered inside that thread (tester confirmed; the previous build had put them top-level); progress and result arrived 80 ms apart, so the debug narration was not visually distinct from default in this run. Both modes ran; the split is pinned in the unit test.
- **Role split at the HTTP layer.** Tokens minted per KSA and audience, presented from the gateway pod: shell role 403 on every chat route; legacy chat role 403 on the A2A event routes and 200 on `/v1/chat/api`; the gateway KSA at the chat audience treated as chat role; only the a2a-chat audience got 200 on the event routes. The prefix rule holds live.
- **Egress through the passthrough, with the adapter's exact argument shapes.** Post 200; edit (`updateMask=text`) 200 with the defanged `< users/all>` staying literal and `<url|text>` rendering as a link; threaded reply 200 with `threadReply: true`; roster 200 (ids only, see above); `findDirectMessage` by email 403 and by id 200; `spaces.setup` 403; `spaces.messages.delete` refused by the relay denylist (400).
- **A platform fact worth recording, for both sequences.** A raw `<users/all>` sent through the passthrough became a live `@all` mention with a USER_MENTION annotation, which is what the adapter's defang exists to prevent. The `<url|text>` in the same probe rendered as a link — the same fact for the other sequence, and the reason the defang now covers it: Chat parses both out of message text the app posts, so an executor-supplied `<url|text>` would be published as a link whose visible text need not name the host it opens. Both were probe-authored strings sent through the relay, so what the run measures here is Chat's parser, not the adapter's defang.

Google Chat DM: the threaded reply, and the edited message's inert defanged mention and rendered link

Published 2026-09-10 from the tester's Google Chat client, at `4524c96`. Hosted off-repo rather than on the fork's `pr-evidence` branch: a screenshot is read once during review, and git history keeps the binary forever. The caption says "published" because that is all it can vouch for — the publish time and the commit at publish, not when the capture was taken.

Read it narrowly, because one thing about it is easy to over-read. **Threading, in the right pane:** parent at 10:33, one reply at 10:37 reading `c2 egress probe: threaded reply shape`, sitting inside the thread rather than top-level. That is the `threadReply: true` call. **Rendering, in the edited 10:33 message:** `bold` in bold, `<url|text>` as the underlined link, and `< users/all>` as inert literal text.

The over-read to avoid: that message says `Edited`, and the space in `< users/all>` was put there by the probe's own edit call, not produced by the adapter's defang on a live outbound message. So what the image proves about the defang is the platform half — Google Chat renders the spaced form as plain text and raises no mention — which is the fact the defang depends on and the reason the pair with the bullet above matters: the same string sent raw became a live `@all`. It does not prove the adapter emits that form. That is the unit test's job, and it is where that claim should be read.

Two of the four shapes are not in frame at all: the raw `<users/all>` in its live-mention state (the image is the after, not the before), and the unverified-sender notice, which needed the allowlist configured to exclude the tester — a different run than this one. Both still rest on the evidence above. For the drop notice that evidence is a request count rather than a rendering — two warnings, two acks, exactly one `POST /v1/chat/api` — which is the stronger form anyway, since an image cannot show that a second notice was *not* sent. And the threading proven here is the probe's reply, not a task's placeholder-and-answer pair landing in a thread; that pair is still the tester's confirmation.

Fixtures from real traffic: all five files under `a2a/gateway/testdata/gchat/` (DM plain, DM `@app` alone, DM `@app` plus text, DM plain, DM thread reply), scrubbed of email, display name, avatar URL, domain id and the one-time redirect token, every other byte as published. The space-side shapes (mention in a space, bare mention, GROUP_CHAT) remain hand-built: the app cannot be added to a space in this Workspace without a corp app review, so those rest on Google's documentation and are labelled so in the tests.

Not covered: a message in a space, poison-message ack-away on the live relay (unit-tested only), the legacy path's event layout, and the operator render, which is the follow-up PR's. The first ask of the run went to the default `platform` addressee, which has no executor on this install, and its placeholder stayed at ⏳ until `stop`; the stale session record is not reapable and the run continued via the Delegate flow and the session route (recorded as F11 in the working notes, not an adapter defect).

Cleanup: the operator was scaled back to 1, which re-rendered both images and the proxy netpol from the CR; its apply merges container env and volumes, so the hand-added env vars and the projected token volume survived the re-render and were removed by hand, and the Deployments now match their pre-run specs (gateway back on Discord, proxy back to its two allowed callers). The IAM grant and the validation subscription remain as the standing validation venue; the capture subscription expires in one day; the two app-posted probe messages in the DM cannot be deleted through the relay by design. The live run was against commit d9fa1d01 (`gateway:…-r4`). The commits after it are the review dispositions: startup refusal and warning paths in the proxy, the terminal-line display gate for canceled and failed tasks (the run only saw completed ones), a longer relay timeout, the `<url|text>` defang and the `ROUTE_ROLES` import-time validator, tests, fixtures' scrubbed ids, comments and prose. None of those paths ran live — in particular the defang as it now stands has not been exercised against Chat; what the run establishes about it is the platform half, below. Transcripts: kube-agents-vamp `round_2/c2-evidence/`.

## Self-Review

Both passes ran on every round, each in a fresh subagent handed only the repository, the three-dot range against a freshly fetched `main`, its skill file and the gate output, and told to derive intent from the diff, not to read commit messages or notes, and to edit nothing. Seven rounds in all (five adversarial, seven docs-drift): the live-validated head, then the head after each round of fixes, until a round returned nothing Blocking and nothing above LOW that was not already dispositioned. Dispositions below are merged and deduplicated; each names its pass and its verdict or triage.

**Fixed on this branch**

- Adversarial, MEDIUM, CONFIRMED: under default display mode the terminal edit for a canceled or failed task still carried the last narration (only the rolling line blanked it). The terminal path now applies the same gate; pinned by a canceled-path test.
- Adversarial, MEDIUM, PLAUSIBLE then CONFIRMED on the fix: the proxy armed two relay instances on one subscription without complaint, and the first guard compared names byte for byte while the relay qualifies short names. Startup now refuses on the fully qualified pair; tested both ways.
- Adversarial, LOW, CONFIRMED: the a2a-chat audience-to-role mapping was untested and silently dropped when the chat split was unset or when the audience collided with the chat or shell one. Both cases now log a warning naming the env, and all three cases are tested.
- Adversarial, LOW, CONFIRMED: no test pinned the default-mode no-op-edit dedupe, the display-mode guard, the author-side allowlist case fold, ack-before-handle ordering, or the a2a-chat role's refusal on the legacy event routes. Each has one now.
- Adversarial, LOW, CONFIRMED: `setBaseEnv` did not clear `A2A_CHAT_DISPLAY_MODE`, so a developer's exported value failed seven unrelated tests. Cleared.
- Adversarial, LOW, CONFIRMED: the relay client timeout (30s) sat under the proxy pull's worst case; raised to 45s with the reasoning in the constant's comment.
- Adversarial, LOW, CONFIRMED (round three): the rolling line recorded its last rendering before the edit landed, so under default mode one failed state edit left the line stale until the terminal (reproduced by the pass). It now records only on success; pinned by a failing-first-edit test.
- Adversarial, LOW, PLAUSIBLE (round three): `Options.Backend` and `Config.Backend()` could disagree, pairing a gchat relay with principal-map resolution. `New` now derives the backend from the config when the option is unset.
- Adversarial, LOW, CONFIRMED (round three): two untested paths (add-on sender email filled from `chat.user`; legacy settles refused when only the a2a relay is armed) and two duplications (relay token assembly; the legacy ack/nack branches). Pinned and folded.
- Docs-drift, Blocking (round three): the sandboxing design's caller-authentication paragraph said "two callers" and then "a third role"; the spec's general rule called the subject an immutable backend id with an example where principal and subject differ, which is not what gchat reports. Both rewritten; the example's subject now equals its principal, as the implementation makes it on gchat.
- Docs-drift, Advisory (round three): the audience string was stated as a value no source owns; the sandboxing design's other two-caller and single-relay sentences; the proxy module comment counting Pods. All reworded.
- Adversarial, LOW, CONFIRMED (round four): the drop-notice map was unbounded and case-sensitive while every adapter map beside it is capped and the allowlist lookup folds case. Now keyed case-folded and capped.
- Adversarial, LOW, PLAUSIBLE (round four): the two new broker-side env vars were not on the operator's reserved list, so a CR could set them through `spec.deployment.env` before the operator renders them. Added to the list with a test; this is the one operator-side hunk in the PR, two strings and a comment.
- Adversarial, LOW, CONFIRMED (round four): Chat's enum strings were inline literals and two const blocks sat below the first function. Named and moved.
- Docs-drift, Advisory (round four): the spec's not-yet-rendered list omitted the broker netpol admission, which the live run had to add by hand; the display split did not say what unset resolves to against the CR default; the key rule omitted GROUP_CHAT and the fail-to-group default; a PR number stood in for a mechanism; one sandboxing sentence still counted two callers. All fixed.
- Adversarial, LOW, PLAUSIBLE (round five): the guard that derives the backend from the config when the option is unset had no test; every in-tree caller passes it explicitly. Pinned.
- Docs-drift, Advisory (round five): the correction made to the sandboxing design's "nothing tells the gateway from the sandbox" sentence left the same pre-existing sentence standing in the credential-isolation design and the security requirements, so the canonical home and the design would have disagreed after merge; the sandboxing design also needed to say that the broker netpol does not yet admit the A2A gateway and to name the residual once the audience split is described. All three docs now say the same thing.
- Adversarial, LOW, CONFIRMED: `"users/"` was an unnamed prefix; the test-only `inbound()` lived in production code; nil guards existed only for a hand-built test struct; a new test class sat below the `__main__` guard; the interface and package comments described AuthorID and Roster as backend-native immutable ids, which the gchat adapter deliberately does not use; the unread default a2a audience constant. All addressed.
- Docs-drift, Advisory: the captured fixtures still carried the real Google user id and DM space id, which the spec itself calls identity. Both replaced; the spec says exactly which fields were replaced.
- Docs-drift, Advisory: the sandboxing design described a two-role table; the spec named the audience as fact without the env that selects it and omitted the broker-side prerequisites and the unprovisioned subscription from its not-yet-rendered list; the stage-2 list omitted the adapter; two docs-map rows over- or under-claimed; two token-path comments said the wrong audience. All rewritten.

**Considered, not fixed, with reasons**

- Adversarial, MEDIUM, CONFIRMED: #1248 (Slack) carries divergent copies of `Backend()`, the one-backend refusal, `droppedNotices` and `verifiedByFor`. Whichever merges second gets textual conflicts and must widen the refusal to three-way; a mechanical resolution could leave Slack plus gchat set together undetected. This branch cannot fix a sibling's diff; the hazard is named in Context and here so the second merge is done by hand, and I will do that merge if this lands first.
- Adversarial, LOW, CONFIRMED: #1258 and #1248 also rewrite the spec's Status line; hand-merge on whichever lands second. #1253 adds `ROUTE_ROLES` entries in the old single-role shape, and that one is now settled twice over: measured, its hunk does not merge clean — it inserts between `/v1/exec` and `/v1/github/`, two lines this branch rewrites, so `git merge` conflicts on the whole tuple and the resolution is by hand either way — and the import-time validator refuses the single-role shape rather than misreading it as a substring test. The fix direction on #1253's side is still a one-element tuple.
- Adversarial, LOW, PLAUSIBLE: `"/v1/chat/api"` has no trailing slash in `ROUTE_ROLES`, so a future `/v1/chat/apix` route would inherit both chat roles. The handler exact-matches the path and 404s anything else, so nothing is reachable; a comment now says the prefix is deliberate.
- Adversarial, LOW, altitude: backend identity is a string switch at four sites, and #1248 extends each the same way. A per-backend resolver value is the right shape; it belongs to the convergence merge with #1248 rather than to either PR alone.
- Adversarial, LOW, PLAUSIBLE: the env default for the display mode is `debug` while the CR field defaults to `default`. Deliberate: unset env keeps Discord installs rendering as before, and the operator is what maps the CR onto the env; that mapping is the wiring follow-up's, where unset CR must become `default`.
- Adversarial, LOW, PLAUSIBLE: the drop notice reaching Discord is a behaviour change on a gchat PR. Kept: the Discord adapter already opened a thread before verification, the notice is once per sender, and #1248 carries the identical change.
- Docs-drift, Advisory, pre-existing (raised on every round): `AGENTS.md`'s `a2a/` layout row predates the gateway, and the site's credential-isolation page still names one audience. The always-loaded context budget is at 0.0k under its ceiling, so the `AGENTS.md` edit belongs to a docs PR that trims elsewhere; the site page describes the legacy path this PR does not change and gets the same one-line fix in that docs PR.

**Verified sound by the passes** (angles worked that produced nothing): removed behaviour re-established everywhere; every referenced symbol, path, env and identity exists; no credential logged; the destructive-method denylist and error scrubbing cover the new path; role isolation goes red under mutation in both directions; no re-implementation of existing helpers; no magic constants remaining; the docs map diff is inserted rows only; ack-before-handle is a documented decision; `resolvePrincipal` preserves case for the audit join; the Chat API argument shapes match googleapiclient's calling convention.

**What the passes could not cover:** live Chat API behaviour (they read the captured fixtures and the code, and could not re-measure); the operator render, which is not in this diff; `google.api_core` is absent locally, so the pull-timeout finding was reasoned from the arguments rather than timed; the release e2e (`tests/e2e/gchat_agent_test.py`) was not run.

**Rounds six and seven** were docs-drift passes alone, on the heads after the round-five and round-six prose edits; the adversarial pass was not re-run for a one-assertion test and prose, and this section says so. Round six returned two Blocking findings, both the same pre-existing single-audience sentence surfacing next: later in the credential-isolation design and on the site page that summarises it. Both now say the per-pod audience, as their three siblings do. It also flagged, Advisory, the dated "corrected 9/5" parentheticals as revision history inside a design doc; kept, because the file's own convention is dated amendments and the substance lives in the section that owns it. Round seven returned no Blocking and four Advisory: the docs-map row for audiences named only the reader, not the operator that writes the values (fixed); one sandboxing sentence stated the A2A gateway's audience in the present tense for a token not yet projected (fixed); the pre-existing `AGENTS.md` layout row (dispositioned above); and the per-pod audience fact now restated on four pages where a pointer would do (kept: the design doc links to the canonical section, and the other three pages state the two values only). Those two one-line fixes were not re-passed; they are the last commit on the branch and the reviewer can read them in under a minute.

**Round-eight reviewer findings** (both fixed; the threads carry the detail and the disputed sub-claim). Chat's second in-text control sequence, `<url|text>`, reached the space undefanged — and the markdown path could generate a spoofed one without any raw angle pair, because the URL class admitted `|`. Both closed in 732a287d; five new `TestToGchatText` cases were run against the pre-fix function to confirm they fail there. `ROUTE_ROLES` had three unenforced invariants that each fail towards admitting a caller, one of them reachable by alphabetizing the table; all three now raise at import, pinned by `RouteRolesTableTest` (5adc73a6). One supporting claim in the first finding was measured wrong and is answered in-thread rather than accepted silently: the #1253 collision it cites conflicts loudly instead of merging clean, which is why the disposition above changed rather than the guard being dropped. The adversarial and docs-drift passes were not re-run for these two commits; no Markdown changed, and `make docs-check` was re-run anyway and is green.

## Risk & Rollout

The adapter is inert unless `A2A_GCHAT_RELAY_URL` is set on the gateway and `A2A_GOOGLE_CHAT_SUBSCRIPTION_NAME` on the proxy; neither is rendered by the operator yet, so no existing install changes behaviour at merge. The proxy change adds a role and routes and renames `required_role` to `required_roles`; callers on the shell and chat audiences keep exactly the routes they had, which the seam tests and the live role matrix show. The one Discord-visible change is that an unmapped sender now gets a visible notice instead of a silent drop. Revert is a plain revert; nothing migrates. Arming the A2A subscription on an install that still runs the legacy Chat consumer double-answers every message until the operator's per-component consumer selection lands, and the spec says so.


bnaylor:
CI red diagnosed and fixed in 2da254a5. It was ours, and it was the kind of collision a clean merge hides.

**What broke.** #913 (`f66d1957`) added `RequiredRoleTest` against `credential_proxy.required_role`, which returned the first matching role. This branch renamed that to `required_roles` returning a tuple — necessary, not cosmetic: `/v1/chat/api` is shared by the legacy chat relay and the A2A one (one app credential, two subscriptions), and a singular return cannot express a route with two admitted callers. Both changes merged without a textual conflict, and the tests then failed with `AttributeError: module 'credential_proxy' has no attribute 'required_role'`.

Worth noting how well it hid: the failing job is `Run Controller Tests`, the visible tracebacks in its log are unrelated handled fallbacks (`sandbox_artifact_patch` logging a missing `gateway.kanban_watchers`), and the directory passes when run standalone — it only fails under `make test-python`'s per-directory invocation.

**Fixed**, plus the gap the rename existed to close. The three tests now read the plural, and two new cases cover what nothing did: `/v1/chat/api` admitting **both** chat callers, and the a2a event route staying its own. Under the singular form that shared route resolved to whichever role matched first, so one of the two consumers would have been refused a route it is entitled to — the bug the rename prevents, previously untested.

**One inherited failure remains and is not this PR's**: `tests/` fails on `test_generate_release_sbom` (2 failures), identically on pristine `upstream/main` — verified in a detached worktree. Flagging rather than fixing, since it belongs to whoever owns #1093.

a2a/cmd/gateway/main.go
```diff
@@ -1,4 +1,4 @@
-// The a2a chatops gateway: Discord in, tasks on the bus out.
+// The a2a chatops gateway: chat (Discord or Google Chat) in, tasks on the bus out.
```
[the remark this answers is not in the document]
bnaylor: Screenshot captured and linked in the body, above `Fixtures from real traffic`:

https://luthen.scromp.net/~bnaylor/gchat-dm-thread-and-defang-4524c96-20260910T034731Z.png

Two notes on how it was published and what it carries.

**Not via `scripts/pr_evidence_screenshot.sh`.** Its publish path is an add-only `pr-evidence` branch on the fork, and the standing call here is to keep review screenshots off the repository — an image is read once during review, and git history keeps the binary forever. Hosted off-repo instead, keeping the script's filename convention (`<slug>-<short-sha>-<UTC-timestamp>.png`) so the URL stays meaningful.

**It covers two of the four shapes you named, and I have written it up that way rather than as a clean sweep.** In frame: the threaded reply landing inside the thread rather than top-level (right pane, parent 10:33, reply 10:37), and the edited message rendering `<url|text>` as a link with `< users/all>` inert beside it.

Not in frame, and still resting on the log lines and the tester:

- The raw `<users/all>` in its live-`@all` state. The image is the after, not the before.
- The single drop notice. That run needed the allowlist set to exclude the tester, which is a different configuration than the one captured. Its evidence stays the request count — two warnings, two acks, exactly one `POST /v1/chat/api` — which is the stronger form regardless, since an image cannot show that a second notice was *not* sent.

One correction to something the image invites, which I only learned after asking. That message is marked `Edited`, and the space in `< users/all>` was put there by the probe's own edit call — it is not the adapter's defang firing on a live outbound message. So the image proves the platform half of the defang (Chat renders the spaced form as plain text and raises no mention), which is the fact the defang depends on, but not that the adapter emits that form. The body says so explicitly. Leaving this thread open rather than resolving it, since two of the four are unmet.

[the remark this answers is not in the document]
bnaylor: Calling this resolved.

The two shapes still not in frame -- the raw `<users/all>` before the defang, and the single drop notice -- are small next to what this PR is actually for, which is having Google Chat support at all.  Both already have evidence that does not need a picture: a unit test for what the adapter emits, and the request count for the drop notice.

If a later round wants those two captured, they are cheap and we can fold them in then.  Not worth holding the adapter on.

agents/platform/scripts/credential_proxy.py
```diff
@@ -195,20 +206,28 @@ class ThreadingUnixHTTPServer(socketserver.ThreadingMixIn, socketserver.UnixStre
 # credential_proxy_client.workspaces_available detects an older broker by
 # asking, and a 403 there would read as "not permitted" rather than "not
 # supported".
-ROUTE_ROLES: tuple[tuple[str, str], ...] = (
-    ("/v1/chat/", CALLER_ROLE_CHAT),
-    ("/v1/exec", CALLER_ROLE_SHELL),
-    ("/v1/github/", CALLER_ROLE_SHELL),
-    ("/v1/workspace/", CALLER_ROLE_SHELL),
+ROUTE_ROLES: tuple[tuple[str, tuple[str, ...]], ...] = (
```
mplakhtiy, line 214: **The new `ROUTE_ROLES` tuple contract is unenforced, and a bare-string entry turns the role check into a substring test.**

`required_roles` now returns a tuple, and `_check_route_role` decides with `principal.role in needed`. A `str` is also a container, so one table entry left in the old `(prefix, role)` shape silently makes that line a substring test instead of a membership test.

The escalation is live rather than hypothetical, because this diff introduces a pair of role names where one nests the other: `"chat" in "a2a-chat"` is `True`. A bare-string `("/v1/chat/a2a/", CALLER_ROLE_A2A_CHAT)` would therefore admit the legacy chat role to the A2A event routes — the turn-stealing the split exists to prevent, and the exact thing `test_a2a_event_routes_demand_the_a2a_chat_role` was written to stop.

Concrete path to a mixed table: #1253 (open) adds `("/v1/forge/", CALLER_ROLE_SHELL)` and `("/v1/vcs/", CALLER_ROLE_SHELL)` on lines that do not conflict with this hunk. Whichever lands second merges clean, and nothing goes red — there is no mypy, pyright or ruff in the Makefile or the workflows, and `RequiredRoleTest` asserts five specific paths, none of which would be the mis-shaped one. The only visible symptom would be the refusal log rendering `"|".join("shell")` as `s|h|e|l|l`. Those two entries are `CALLER_ROLE_SHELL`, so they do not escalate on their own; what they establish is that a mixed table arrives without a conflict, and the next such entry decides whether it escalates.

One `isinstance(roles, tuple)` assertion over `ROUTE_ROLES` — in `RequiredRoleTest`, or at module scope — makes this loud for #1040 and everything after, not just #1253.

Deliberately not offering a suggestion block here: normalizing inside `required_roles` (`return (roles,) if isinstance(roles, str) else roles`) would accept the mis-shaped entry rather than reject it, which papers over the mistake instead of surfacing it. The guard wants to fail, not coerce.

a2a/gateway/gchat.go
```diff
@@ -0,0 +1,782 @@
...
+	}
+	sort.Strings(keys)
+	return &gchatEvent{shape: gchatShapeAddon + ":" + strings.Join(keys, ",")}, nil
+}
+
+// toGchatText translates executor markdown to Google Chat's text format and
+// defangs the one control sequence Chat parses out of message text: a
+// <users/…> mention, which a prompt-injected result could use to ping the
+// room. The defusing is a visible space, not an invisible character.
+func toGchatText(s string) string {
```
mplakhtiy, line 639: **The outbound defang covers `<users/...>` but not Chat's other in-text control sequence, `<url|text>`.**

Google Chat parses two control sequences out of plain message text. This function neutralizes the mention and passes the other through: `strings.ReplaceAll(s, "<users/", "< users/")` rewrites a mention, while a `<...|...>` already present in the input reaches Chat intact. The adapter knows the second one is live — `gchatLinkRe` on the next line *generates* that exact form from markdown links — and the PR's live run observed Chat rendering it from message text.

Failure scenario: executor output is model output, and the standing threat model here treats the LLM-driven pod as the injectable adversary (`tests/integration/test_seam_a2a_chat_ingress.py` says so in as many words). A prompt-injected task result containing the literal `<https://attacker.example/login|https://github.com/gke-labs/kube-agents>` is posted into the space verbatim. Every member sees a link whose visible text is a trusted URL and whose target is not, published under the app's identity rather than a human's — the same class the `<users/` defang exists to close, one step over.

`TestToGchatText` has no case for a `<...|...>` already present in the input.

The comment above this line is also inaccurate as written: `<users/...>` is not "the one control sequence Chat parses out of message text", and a reader who trusts that sentence will not look for this. The fix is not free, since the adapter wants to emit that form — escaping a `<` the link regex did not itself produce, or rejecting a `|` inside an angle pair the regex did not create, are the two shapes that work. Narrowing the comment is a defensible answer on its own if the escaping is deferred.

mplakhtiy (commented):
Reviewed at `4524c968569a3b95e2ef2d0613d9d5e0e032d5d1`. Produced with AI assistance (Claude).

Two findings, both inline.

agents/platform/scripts/credential_proxy.py
bnaylor: Fixed in 5adc73a6 — and wider than one assertion, because the table has two more ways to reach the same escalation.

`_validate_route_roles` runs at module scope and raises on three things:

1. **A roles entry that is not a tuple.** The finding's case exactly. It raises rather than normalizing, for the reason you gave.
2. **A role not in `CALLER_ROLES`.** A typo confers nothing, so the route it guards refuses everyone, which reads as policy rather than as a mistake.
3. **Any entry an earlier prefix already matches.** This is the one worth the extra lines — see below.

**A correction on the route in, because it changes why the guard is needed.** #1253 does not merge clean against this hunk. It inserts `("/v1/forge/", …)` between `/v1/exec` and `/v1/github/`, and this PR rewrites both of those context lines, so git conflicts on the whole tuple:

```
$ git merge --no-commit --no-ff refs/tmp/pr1253
Auto-merging agents/platform/scripts/credential_proxy.py
CONFLICT (content): Merge conflict in agents/platform/scripts/credential_proxy.py
```

```
<<<<<<< HEAD
ROUTE_ROLES: tuple[tuple[str, tuple[str, ...]], ...] = (
    ("/v1/chat/a2a/", (CALLER_ROLE_A2A_CHAT,)),
    ...
=======
ROUTE_ROLES: tuple[tuple[str, str], ...] = (
    ("/v1/chat/", CALLER_ROLE_CHAT),
    ("/v1/exec", CALLER_ROLE_SHELL),
    ("/v1/forge/", CALLER_ROLE_SHELL),
    ...
>>>>>>> refs/tmp/pr1253
```

Whoever lands second sees both shapes side by side. So "merges clean, nothing goes red" is not how a mixed table arrives here — a hand-resolved conflict over a security table is, which is a worse place to be relying on care than the silent merge you described, but it is a different one.

**The stronger route needs no bare string at all.** Matching is first-wins by prefix and `/v1/chat/` sorts ahead of `/v1/chat/a2a/`, so alphabetizing the table — a tidying edit, no typo, no conflict, correctly-shaped tuples throughout — takes the A2A event routes:

```
$ python3  # the shipped entries, re-sorted
alphabetical sort order: ['/v1/chat/', '/v1/chat/a2a/', '/v1/chat/api', '/v1/exec', '/v1/github/', '/v1/workspace/']
  /v1/chat/a2a/events: shipped=('a2a-chat',)  resorted=('chat',)
does the legacy chat role reach the a2a events route after a re-sort? True
```

That was a comment (`# Order matters`) and is now check 3. The comment points at the validator and says not to sort the table.

Import-time rather than a test assertion, for the reason you were after with "#1040 and everything after": every test module imports this one, so a mis-shaped table is a red pull request rather than something that has to be remembered. Confirmed there is no mypy, ruff, pyright, flake8 or pylint in the `Makefile` or `.github/workflows/`.

`RouteRolesTableTest` pins each refusal and names the escalation it prevents. The sorted-table case asserts against `sorted(ROUTE_ROLES)` rather than a hand-written pair, so it stays true as routes are added.

Suites: `test_credential_proxy.py` 259 tests OK; `python3 tests/conformance/run.py` 108 tests OK (expected failures=12); `tests/integration/test_seam_a2a_chat_ingress.py` 7 tests OK.

a2a/gateway/gchat.go
bnaylor: Fixed in 732a287d. The finding is right, and the markdown path had the same hole without needing a raw angle pair.

**The case you named.** `toGchatText` now defangs every span it did not author, so `<https://attacker.example/login|https://github.com/gke-labs/kube-agents>` arriving in executor output leaves as `< https://attacker.example/login|…>`. The generated form is written straight to the builder, so the link the adapter makes from markdown still ships live.

**The sibling.** The URL class was `[^)\s]+`, which admits `|`. So the spoof was reachable through a plain markdown link, no `<` required — verified by running the new cases against the pre-fix function:

```
$ go test ./gateway/ -run TestToGchatText   # against HEAD~
toGchatText("[trusted](https://evil.example|https://github.com)")
  = "<https://evil.example|https://github.com|trusted>"
toGchatText("see <https://evil.example|the doc>")
  = "see <https://evil.example|the doc>"
toGchatText("[real](https://x.example/p) vs <https://evil.example|real>")
  = "<https://x.example/p|real> vs <https://evil.example|real>"
toGchatText("[<https://evil.example|hi>](https://x.example/p)")
  = "<https://x.example/p|<https://evil.example|hi>>"
```

Five of the new cases fail against the old function. The URL class now refuses `<`, `>` and `|` — none is legal in a URL unencoded — so that markdown is left as written instead. A link's display text is defanged too, rather than riding out on the exemption, which is the fourth line above.

**Scope of the defang.** `gchatPipeRe` requires the segment before the pipe to be non-empty and space-free, which is the shape Chat linkifies; `latency < 5 | p99 > ok` and `if a < b then` are pinned as untouched. Where the two readings are ambiguous it defangs, and the comment says so — I did not re-measure Chat's parser for this, so the pattern is set against the shape rather than against a grammar I verified.

**On the evidence.** What the live run establishes is that Chat renders `<url|text>` from message text the app posts (PR body, egress bullet: "`<url|text>` rendering as a link"). That link was one the adapter generated, so the run proves the platform parses the sequence out of message text — it does not itself demonstrate an input-carried one surviving to a space. The new unit cases are what cover the adapter's half.

`Post` and `Edit` are the only two outbound text paths and both go through `toGchatText`, so that is the whole set.

The comment is corrected: `<users/…>` is no longer described as "the one control sequence Chat parses out of message text", which is what let this read as covered.

Suites: `go test ./...` in `a2a/` all packages ok; `go vet ./gateway/` clean; `python3 tests/conformance/run.py` 108 tests OK (expected failures=12).

bradhoekstra (approved):

Issue #1202, opened 2026-09-03
Merge throughput under the gate: drain the backlog, cap tide batches, keep flake at zero, cut the wall clock


jayantid:
The gate is trustworthy (false-red classes eliminated this week; post-fix audit in flight). The remaining problem is **merge throughput**: tide validates merges with a 1.5–3.5h required job whose result is invalidated both by main moving (serial path) and by any batch-mate flaking out mid-flight (batch path — tide discards even a *green* batch when one member drops). Full diagnosis with the observed #1191 timeline: GoogleCloudPlatform/oss-test-infra#2687. At the peak, 13 PRs sat lgtm+approved while merges landed hours apart, and three PRs needed admin force-merges in 24h.

## The plan

**1. One-time drain of the backlog (today, maintainer).** The queued PRs predate the fixes; a button-merge session for the small review-clean ones (then one settled-main retest wave for the rest) converts a ~26h serial drain into one sitting. Owner: maintainer with merge rights.

**2. Cap tide batches at 2 (this week, config).** Repurpose oss-test-infra#2683 from `batch_size_limit: -1` (disable — filed under a withdrawn theory) to `2`: doubles steady-state throughput, and one flaky member costs one partner instead of the whole pool. Safe under any answer to #2687's salvage question.

**3. Keep per-run flake at ~zero (continuous).** Batch survival is the *product* of member stability over a multi-hour flight, so every flake class matters twice now. Open contributors: #1097 (endpoint slot storms — mitigated by #1191's classification, root cause still live), #1172 (helm poisoning — fix in review). Every new class found: same-day classification per the gate playbook.

**4. Cut the job's wall clock — the multiplier on everything (#1025, top lane).** Serial retest cost, batch flight window, storm exposure, and drain time all scale linearly with job duration. Known levers: task parallelism 6 (coverage docket), agent-execution time is ~73% of the job (#948). Target: green runs ≤90 min.

**Upstream (tracked, not blocking):** green-batch salvage and path-aware staleness — oss-test-infra#2687.

## Done when

- Backlog drained; no PR sits lgtm+approved >4h with green checks under normal operation.
- batch_size_limit: 2 live.
- Two consecutive weekdays without a force-merge or manual /hold to unstick a mergeable PR.
- Green-run p50 ≤ 90 min (hands off to #1025 for the long tail).

bradhoekstra:
[...]

PR #1262, opened 2026-09-06
ci(tide): keep a pull request with unresolved review threads out of the merge pool

bradhoekstra:
## Summary

A scheduled workflow puts the bare `do-not-merge` label on any `lgtm`+`approved` pull request whose review threads are not all resolved, and takes it off at the next five-minute sweep after the last resolution. Tide's query for this repository already excludes that label and no Prow plugin writes it, so the workflow can own it outright. The label carries one comment telling the author why it is there and that resolving the threads is what removes it.

## Why This Change

`main` requires every review conversation resolved before a merge, and Tide does not read that rule. It picks the lowest-numbered pull request in its pool, GitHub refuses the merge with a repository-rule violation, and Tide retries the same pull request every ~85 seconds while every other pull request in the pool waits behind it. On 2026-09-05 the batch `[608, 1197]` failed that way from 10:59 to 11:57 until an admin held both by hand; #1122 sat 5h46m and 231 attempts the same way (#1202). Nothing in the repository prevented it, and the pull requests that were actually ready merged only after someone noticed.

## What Changed

- `.github/workflows/hold-unresolved-threads.yml`: every five minutes, and on the `lgtm`/`approved` label events through `pull_request_target`, runs the sweep with a write token. The schedule is the guarantee: thread resolution fires `pull_request_review_thread`, but from a fork that event carries a read-only token. Checks out the default branch only.
- `scripts/hold_unresolved_threads.py`: lists open pull requests through the pulls endpoint (strongly consistent, unlike search, which lags the label event the fast path reacts to), keeps the non-draft ones carrying both pool labels plus any carrying `do-not-merge`, counts unresolved threads through GraphQL (paged), adds the label with one marked comment, and removes it only when the issue timeline shows the label's most recent application was the workflow's own. A person's hand-applied label stays. `/hold cancel` does not touch it.
- `scripts/github_api.py`: a `graphql()` call, because REST has no notion of a resolved thread. Raises on an `errors` body rather than returning partial data.
- `docs/pull-request-workflow.md`: "How a change merges" explains the label next to `/hold`; "Resolving conversations" points at it; the `mergeStateStatus` paragraph's #1122 example now names this as the case the label keeps out. `docs/README.md` gains one identifier-sources row.

## Context

No open issue or pull request covers this. It follows the queue drain on 2026-09-05 (#1175 through #1240 merged serially by hand); #1202 documents the earlier #1122 stall. `pull_request_target` follows the pattern `risk_classify.yml` already uses, with the same zizmor annotation and default-branch checkout.

## Testing

- `cd scripts && python3 -m pytest -q`: 1026 passed, including 19 new tests in `test_hold_unresolved_threads.py` (each branch of `decide`, which open pull requests are read at all, timeline ownership by last event, thread paging, label created before use, one comment per hold and none on a relabel, dry run writes nothing, one failure does not stop the sweep) and two in `test_github_api.py` for `graphql()` (data returned; a 200 with an `errors` body raises).
- `actionlint` clean on the workflow; `prettier@3.9.6 --check` clean on the workflow and the doc; `make docs-check` clean (AGENTS.md at 0.0k under budget, unchanged by this PR).
- `tests/conformance/test_B_write_path.py` (17 passed, 3 xfailed): the new workflow is on B2's allowlist of `pull-requests: write` holders, with the reasoning in the docstring. Checkov's `CKV_GHA_7` on the `dry_run` dispatch input is skipped with a reason, the form every other dispatchable workflow here uses; the first CI run red on both before this.

### Live validation

Not a cluster change, so the live surface is GitHub itself.

Read path, against gke-labs/kube-agents with a dry run at 2026-09-06 ~00:20 UTC (`python3 scripts/hold_unresolved_threads.py --repo gke-labs/kube-agents --dry-run`):

```
#608: would be labelled: 4 unresolved thread(s) and in the pool
#965: left alone: every thread resolved; not labelled
#1150: would be labelled: 1 unresolved thread(s) and in the pool
#1186: left alone: every thread resolved; not labelled
#1199: left alone: every thread resolved; not labelled
#1200: left alone: every thread resolved; not labelled
```

#608 is exactly the pull request that jammed the queue the day before. Nothing was written to this repository.

Write path, on a throwaway pull request in my fork (bradhoekstra/kube-agents#25, since closed and its branch deleted) labelled `lgtm`+`approved` with one review thread on its diff, running the real script with my token:

1. Run 1: `#25: labelled: 1 unresolved thread(s) and in the pool` → labels `lgtm,approved,do-not-merge`, one comment carrying the marker.
2. Run 2, nothing changed: `left alone: 1 unresolved thread(s); already labelled` → still exactly one marked comment.
3. Thread resolved through `resolveReviewThread`, run 3: `left alone: every thread resolved, but the label was applied by a person; leaving it` — correct, because my token (not `github-actions[bot]`) had applied it.
4. Run 4 with `BOT_LOGIN` set to my login in-process, to stand in for the workflow's token: `unlabelled: every thread resolved` → label gone.

Repeated on a second throwaway (bradhoekstra/kube-agents#26, also closed) after the review-driven changes below: run 1 labelled and commented; the label was then removed by hand with the thread still open and run 2 re-applied it without a second comment (one marker comment on the pull request afterwards). The dry run above was re-run after switching the listing from search to the pulls endpoint and selected the same pull requests.

Not covered: the workflow itself has not run under `schedule` or `pull_request_target` yet, because both are guarded to gke-labs/kube-agents; the first scheduled run after merge is the proof, and `workflow_dispatch` with `dry_run` can be run before that. The bare `do-not-merge` label does not exist in this repository yet; the script creates it on first run.

## Self-Review

Both passes ran in fresh subagents handed the diff range and the skill path only, told to derive intent from the diff and not to read commit messages. Neither had a cluster or dispatch rights.

- **MEDIUM, PLAUSIBLE (adversarial).** The `labeled` fast path listed the pool through GitHub search, which is eventually consistent, so the sweep that runs seconds after `lgtm` lands could miss the very pull request the event was about and leave it to the next schedule tick. Fixed: the pool is now read from `GET /pulls?state=open` and filtered in Python; the search queries and their paging are gone. Confirmed by the live dry run selecting the same pull requests.
- **LOW, CONFIRMED (adversarial).** `"add"`/`"remove"` were bare strings compared in six places (no-magic-constants). Fixed: `ACTION_ADD`/`ACTION_REMOVE`.
- **LOW, CONFIRMED (adversarial).** `COMMENT_MARKER` claimed a later sweep could find it, but nothing read it, and a label removed by hand while threads stayed open would get a fresh comment every five minutes. Fixed: an add checks for an existing marker comment first and posts none if one is there; tested, and exercised live on the second fork fixture.
- **LOW, CONFIRMED (adversarial; also docs-drift advisory 1).** "Within five minutes" promised more than a `*/5` schedule guarantees. Fixed in the doc and in the comment text: "at the next five-minute sweep".
- **LOW, CONFIRMED (adversarial).** `graphql()` had no test for the contract it exists for. Fixed: two tests in `test_github_api.py`.
- **Advisory (docs-drift).** "An approved pull request" was looser than the code's `lgtm`+`approved`, non-draft condition — reworded. The new paragraph repeated the #1122 figure the same page already states — dropped, and the existing #1122 sentence now points at the label. `docs/README.md`'s identifier-sources table gained a row for the label, interval, and ownership rule. `github_api.py`'s module docstring now names `graphql()` and its raise-on-`errors` contract.
- **Advisory (docs-drift), not fixed here.** `docs/testing-map.md:11` cites the `PYTHON_TEST_DIRS` globs at a `Makefile` line range that has drifted (now 151-166). Pre-existing and outside this diff; noted for a follow-up.
- **Adversarial angles with nothing found:** `ref: default_branch` is empty on `schedule` and falls back to `github.ref`, the default branch, as `risk_classify.yml` relies on; `pull_request_target` safety (default-branch checkout, `persist-credentials: false`, no pull-request content in `run:`); label writes with `GITHUB_TOKEN` do not re-trigger the workflow, so no loop; concurrency group serialises sweeps; `label_is_ours` errs toward leaving a label on; `pin_smoke_status.py` already treats any `do-not-merge*` label as out of the pool, so the two sweeps agree. Docs-drift verified the four-label exclusion list against `prow/oss/config.yaml` in oss-test-infra and that no Prow plugin writes the bare label.
- **Not covered by either pass:** the workflow has not run under `schedule` or `pull_request_target` (both guarded to gke-labs/kube-agents); no red-then-green, since this is new behaviour rather than a fix.

## Risk & Rollout

The workflow writes labels and one comment; it runs no pull-request code and never merges anything. The failure modes are a label applied to a pull request that is in fact mergeable, which the sweep removes within five minutes once the threads read resolved and Tide would have refused anyway, and a person's `do-not-merge` removed by mistake, which the timeline check prevents and a test covers. Revert by deleting the workflow; any label it left is removed by hand. Nothing has to happen at merge time; the label is created on the first sweep. Merging this does not need a Prow config change: the label is already in Tide's exclusion list for this repository.

### DIFF
--- a/.github/workflows/hold-unresolved-threads.yml
+++ b/.github/workflows/hold-unresolved-threads.yml
@@ -1,0 +1,85 @@
+name: Hold Unresolved Threads
+
+# Keep a pull request with unresolved review threads out of Tide's merge pool.
+#
+# `main` requires every conversation resolved before a merge, and Tide does
+# not read that rule: it picks the lowest-numbered `lgtm`+`approved` pull
+# request, GitHub refuses the merge, and Tide retries the same one every ~85
+# seconds while every other pull request in the pool waits behind it. #608
+# and #1197 held the queue that way for an hour on 2026-09-05.
+#
+# `scripts/hold_unresolved_threads.py` mirrors thread state onto the bare
+# `do-not-merge` label, which Tide's query for this repository excludes and no
+# Prow plugin writes: on while any thread is unresolved (with one comment
+# saying so), off once they all are -- unless a person applied it, in which
+# case it is theirs. The script's docstring is the reference.
+#
+# Two triggers. The schedule is the guarantee: thread resolution fires
+# `pull_request_review_thread`, but from a fork -- where every pull request
+# here comes from -- that event hands the workflow a read-only token, so it
+# cannot be the mechanism. The `labeled` fast path closes the window that
+# matters most: the moment `lgtm` or `approved` lands is the moment Tide would
+# otherwise start retrying, and `pull_request_target` carries a write token
+# whatever the head's origin. Nothing from the pull request is checked out or
+# read as code on either path.
+on:
+  schedule:
+    - cron: "*/5 * * * *" # SWEEP_MINUTES in the script quotes this to the author
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: Log what would change and change nothing
+        type: boolean
+        default: false
+  # zizmor: ignore[dangerous-triggers] - the job checks out the default
+  # branch only (the script) and reads pull-request state through the API;
+  # no pull-request content reaches the write token's context.
+  pull_request_target:
+    types: [labeled]
+    branches: [main]
+
+permissions: {}
+
+# Sweeps reconcile the same state; two at once would race each other's reads.
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  sweep:
+    name: Mirror unresolved threads onto the do-not-merge label
+    # The repository guard is required of any credentialed workflow that
+    # starts on its own (AGENTS.md, Pull Request Hygiene). The label filter
+    # keeps every other label event -- size, risk, lgtm removed -- off a
+    # runner; the schedule covers anything it misses.
+    if: >-
+      github.repository == 'gke-labs/kube-agents' &&
+      (github.event_name != 'pull_request_target' ||
+        github.event.label.name == 'lgtm' ||
+        github.event.label.name == 'approved')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # check out the default branch: the script
+      pull-requests: write # add and remove the label, post the comment
+      issues: write # create the label the first time
+    steps:
+      # The default branch, never a pull request head: this job holds a
+      # writable token and must not run code from a fork. `ref:` is explicit
+      # because on `pull_request_target` the fallback is the pull request's
+      # base branch.
+      - name: Checkout the default branch
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false
+
+      # The runner's own python3; the script is stdlib-only. Every sweep is
+      # the same whole-repository reconcile, whichever trigger started it, so
+      # the event carries nothing the script needs.
+      - name: Sweep the merge pool
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DRY_RUN: ${{ inputs.dry_run }}
+        run: |
+          if [ "$DRY_RUN" = "true" ]; then set -- --dry-run; fi
+          python3 scripts/hold_unresolved_threads.py "$@"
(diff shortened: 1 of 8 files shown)


bradhoekstra:
/retest
The red is the eval project, not this change: on kube-agents-evals-3 the seeded `payments-api` fixture sat Pending (seeded-a default pool at 940m/940m CPU, autoscaler blocked) so the planted OOM crashloop never ran, and all three crashloop cases failed on the same cause. #1226 hit the identical failure on kube-agents-evals-27 an hour earlier; three other PRs passed on the same base in other projects. Filing a presubmit-gate issue with the details.

toshiowang (commented):

**LOW** — `tests/conformance/test_B_write_path.py:329`: the docstring says "Four workflows hold it, and none can give an approval:" but this PR adds `hold-unresolved-threads` as the fifth entry in that list (and the allowlist below). The count is now stale.

toshiowang (approved):

Issue #737, opened 2026-08-17
security(sandbox): the agent's shell shares a container with the Session KV store, its config, and the credential material


dshnayder:
## Summary

The Platform Agent's shell runs in the same container, under the same UID, as the
Session KV store, the agent's own `config.yaml`, and the credential-proxy PATH shims.
Anything the agent is refused through a tool, it can reach directly through the
filesystem.

This is not hypothetical. An incident on 2026-08-11 had the agent post to a Google Chat
space nobody had asked it to post to. The sequence:

1. `send_notification` failed — no session routing for the turn.
2. The agent ran `grep -rn "spaces/" /opt/data/kanban/logs/` and recovered a space id
   from its own logs.
3. It patched `/var/lib/kube-agents/session/session_kv.db` with
   `python3 -c "import sqlite3..."` to point the session at that space.
4. It ran `/opt/hermes/bin/hermes config set GOOGLE_CHAT_HOME_CHANNEL <space> --force`.
5. The message was delivered.

No step exploited a bug. Each was the documented behaviour of a tool the agent is
allowed to call. The agent was recovering from a failure the only way its environment
permitted, and its environment permitted too much. (Space id redacted here; the full
report is internal.)

## Why the current design allows it

**The shell has no sandbox.** Hermes selects where `terminal`, `execute_code`, and the
file tools run via `terminal.backend` — `local`, `docker`, `ssh`, `singularity`,
`modal`, `daytona`, or `vercel_sandbox`. This repository sets no `terminal:` block
anywhere, so the backend is `local`, which upstream is explicit about: no container, no
namespace, no seccomp, no cgroup. The shell *is* the gateway container.
`agents/platform/config.yaml` also has no `disabled_toolsets`, so the platform profile
has `terminal`, `file`, and `code_execution` all enabled. (The chat front door denies
them — `agents/chat/config.yaml`.)

**The Session KV store is a plain file on a shared mount.** `system-metadata` (PVC) is
mounted `subPath: session` read-write into both the gateway and dashboard containers
(`platformagent_manifests.go:1851`, `:2419`). The pod `securityContext` sets
`RunAsUser: 10000` for every container (`:1731-1734`). The KV server is a background
process in the *same* container, started at `docker-entrypoint.sh:1190`. The agent owns
the file, so file modes cannot exclude it.

**There is an HTTP interface, but it is not the only one.** `session_kv_server.py`
[...]

PR #1266, opened 2026-09-07
fix(e2e): align test fixtures and pod discovery with shell sandbox architecture (#1265)

### LINKED ISSUE #1265: fix(e2e): fix E2E 'rc' suite failures due to pod split in PR #913 (blocking nightly & rc gates)
### Summary
The E2E `rc` suite (`tests/e2e/test_agent_api_health.py` and `tests/e2e/test_agent_fleet_audit.py`) is failing in both the Nightly and RC release pipelines. Because `nightly-pipeline.yml` uses `blocking_suite: "rc"`, this is currently blocking the nightly promotion and staging tag creation.

This is a regression introduced in #913 ("feat(operator)!: run the agent shell in a sandbox pod"), which split the workload into `gateway`, `credential-proxy`, and `shell` (sandbox) pods, leaving two test fixtures/helpers desynchronized with the new pod/container topology.

---

### Affected Pipelines & Runs
1. **Nightly Pipeline** (Run [#34128024494](https://github.com/gke-labs/kube-agents/actions/runs/34128024494), Job [#101768770479](https://github.com/gke-labs/kube-agents/actions/runs/34128024494/job/101768770479)):
   - Failed step: `Execute Blocking E2E Gate` (`E2E_SUITE=rc`)
   - Outcome: Blocked `Step 4 - Create Staging Tag` and pipeline failed.
2. **RC Pipeline** (Run [#34122986133](https://github.com/gke-labs/kube-agents/actions/runs/34122986133), Job [#101752936414](https://github.com/gke-labs/kube-agents/actions/runs/34122986133/job/101752936414)):
   - Failed step: `Execute Optional E2E Suites` (`OPTIONAL_SUITES=rc`)
   - Outcome: Suite failed (18 passed, 9 skipped, 1 failed, 1 error), tolerated only because it is currently configured as optional.

---

#### 1. `test_platform_agent_api_health_ping` (ERROR)
- **Location**: `tests/e2e/conftest.py:436` (`port_forward_agent` fixture)
- **Log**:
  ```text
  ERROR tests/e2e/test_agent_api_health.py::test_platform_agent_api_health_ping - Failed: The Platform Agent API did not answer over the exec relay.
  Platform Agent API transport, namespace kubeagents-system:
    pod → platform-agent-gateway-9d858977c-wztrz
    REJECT exec relay -> envoy-credential-proxy:8643 in a pod matching app=platform-agent-gateway -- probe /api/sessions?limit=1 got TimeoutError: timed out
    kubectl said: error: container envoy-credential-proxy is not valid for pod platform-agent-gateway-9d858977c-wztrz out of: platform-agent, fluent-bit, sandbox-credential-cleanup (init), stage-gkestockoutinvestigator (init), stage-pubsubplatform (init), sandbox-ssh-key (init), agent-api-auth (init) (x48)
  ```
- **Cause**: `_PROXY_CONTAINER` is hardcoded as `"envoy-credential-proxy"`. In the gateway pod, the sidecar container listening on port 8643 is named `"agent-api-auth"`.

#### 2. `test_github_token_minting_and_connectivity` (FAILED)
- **Location**: `tests/e2e/test_agent_fleet_audit.py:61`
- **Log**:
  ```text
  FAILED tests/e2e/test_agent_fleet_audit.py::test_github_token_minting_and_connectivity - AssertionError: GitHub token minting and API probe failed inside pod 'platform-agent-credential-proxy-69865b9486-2zf7w' (exit code 1):
    STDERR:
    error: container platform-agent is not valid for pod platform-agent-credential-proxy-69865b9486-2zf7w out of: envoy-credential-proxy
  ```
- **Cause**:
  1. The pod search loop queries `kubeagents.x-k8s.io/has-credential-proxy=true`. After #913, that label belongs to the credential broker pod (`platform-agent-credential-proxy`), not the gateway or shell pod.
  2. The test executes `kubectl exec -c platform-agent`, but `platform-agent-credential-proxy` only has container `envoy-credential-proxy`.
  3. The test script invokes `github_token_refresh.py` and `gh api repos/...` in `/opt/data`. This execution runtime now resides in the shell sandbox pod (`app=platform-agent-shell`, container `shell`).

---

### Suggested Remediation
1. In `tests/e2e/conftest.py`: update `_PROXY_CONTAINER = "agent-api-auth"` (with adaptive fallback if `agent-api-auth` is not present).
2. In `tests/e2e/test_agent_fleet_audit.py`: update pod and container resolution in `test_github_token_minting_and_connectivity` to target the shell sandbox pod (`app=platform-agent-shell`, container `shell`) with fallback to gateway pod (`platform-agent`) for older layouts.
3. In `hack/ci-env.sh:123`: update log collector from `deployment/platform-agent-gateway -c envoy-credential-proxy` to `agent-api-auth` and add broker pod log collection.

eLeontev:
## Summary

Aligns E2E test fixtures and live probe execution with the pod split introduced in #913 (where the single agent pod was separated into gateway, credential-proxy broker, and shell sandbox). Fixes regressions in the blocking `rc` test suite that were preventing Nightly release promotion.

## Why This Change

PR #913 ("feat(operator)!: run the agent shell in a sandbox pod") split the monolithic agent workload into three distinct pods:
1. `platform-agent-gateway` (running container `platform-agent` and native sidecar `agent-api-auth`)
2. `platform-agent-credential-proxy` (running `envoy-credential-proxy`)
3. `platform-agent-shell-0` (StatefulSet pod running container `shell` with `/opt/data` and credential proxy CLI shims)

The E2E `rc` suite (`test_agent_api_health.py` and `test_agent_fleet_audit.py`) was failing because:
- `port_forward_agent` tried to exec into `envoy-credential-proxy` inside `platform-agent-gateway` where it no longer exists.
- `test_github_token_minting_and_connectivity` selected `platform-agent-credential-proxy` via `kubeagents.x-k8s.io/has-credential-proxy=true` and tried to exec into `platform-agent`, failing because that container does not exist there. Furthermore, `gh` and `/opt/data` live exclusively in the sandbox pod.

Because `nightly-pipeline.yml` enforces `rc` as a blocking promotion gate, all nightly runs and staging promotions were blocked.

## What Changed

- **`tests/e2e/conftest.py`**:
  - Default `_PROXY_CONTAINER` to `"agent-api-auth"` with dynamic fallback via `_resolve_gateway_proxy_container` to `"envoy-credential-proxy"` on older single-pod layouts.
- **`tests/e2e/test_agent_fleet_audit.py`**:
  - Update `test_github_token_minting_and_connectivity` to target the shell sandbox pod (`app=<name>-shell`, container `shell`) with a dedicated 120s availability wait loop.
  - Drop privileges to unprivileged `agent` user (uid 1000) via `runuser -u agent` when execing inside the sandbox shell container, avoiding running agent-writable scripts with root privileges.
  - Prepend `/opt/credential-proxy/bin` to `PATH` for `gh`.
  - Retain fallback to legacy gateway pod only after sandbox wait expires.

*(Note: Diagnostic failure artifact collection in `hack/ci-env.sh` was split out into a separate PR to decouple Prow CI script reviews from E2E release test fixtures).*

## Context

- Fixes #1265
- Regressed from #913

## Testing

- Python syntax and compile validation: `python3 -m py_compile tests/e2e/test_agent_fleet_audit.py` passed.
- Unit tests: `pytest scripts/test_ci_connectivity_retry.py` passed (5/5).
- CI workflow linters and checks passing on GitHub Actions.

### Live validation

Executed against the live `platform-agent-host` cluster in `kube-agents-nightly` (`us-central1`, preserved from failed nightly run #34128024494):
- Full `rc` suite passed 20/20 (9 skipped):
  - `test_agent_api_health.py::test_platform_agent_api_health_ping` PASSED (exec relay via `agent-api-auth:8643`).
  - `test_agent_fleet_audit.py::test_github_token_minting_and_connectivity` PASSED (exec under `agent` user in `platform-agent-shell-0` via `runuser -u agent`).
  - All audit report and stockout scenarios passed.
- Also tested against single-pod cluster to verify backward-compatible gateway fallback.

## Self-Review

Passes conducted:
- **Security & Privilege Analysis**: Verified that `kubectl exec` into the sandbox `shell` container runs as root by default. Added `runuser -u agent` to drop privileges to `agent` (uid 1000) so agent-writable scripts under `/opt` are never executed as root.
- **Race Condition / Timeout Analysis**: Addressed review feedback where gateway fallback could pre-empt the sandbox wait loop; separated sandbox wait into a dedicated 120s polling loop before falling back.
- **Scope Scrutiny**: Removed unscoped namespace pod scans that could inadvertently select foreign agent pods in shared namespaces. Reverted `hack/ci-env.sh` so Prow infrastructure scripts are reviewed in a standalone PR.

## Risk & Rollout

Low risk: changes are limited to E2E test fixtures and test assertions. No production operator or runtime container logic is modified.

### DIFF
--- a/hack/ci-env.sh
+++ b/hack/ci-env.sh
@@ -119,8 +119,9 @@ dump_prow_artifacts_on_failure() {
     # is visible only in this pod's log -- the gateway just relays the text.
     kubectl logs deployment/litellm -n "${ns}" --tail=1000 > "${artifact_dir}/litellm.log" 2>&1 || true
     # The gateway capture above reads the pod's default container (platform-agent);
-    # a dropped port-forward stream is only visible from the envoy sidecar's side.
-    kubectl logs deployment/platform-agent-gateway -c envoy-credential-proxy -n "${ns}" --tail=2000 > "${artifact_dir}/envoy-credential-proxy.log" 2>&1 || true
+    # a dropped port-forward stream is only visible from the auth sidecar's side.
+    kubectl logs deployment/platform-agent-gateway -c agent-api-auth -n "${ns}" --tail=2000 > "${artifact_dir}/agent-api-auth.log" 2>&1 || true
+    kubectl logs deployment/platform-agent-credential-proxy -c envoy-credential-proxy -n "${ns}" --tail=2000 > "${artifact_dir}/envoy-credential-proxy.log" 2>&1 || true
     # Konnectivity/tunnel churn shows up as kube-system events, not in "${ns}".
     kubectl get events -n kube-system --sort-by=.lastTimestamp 2>&1 | tail -100 > "${artifact_dir}/kube-system-events.txt" 2>&1 || true
 
--- a/hack/ci-env.sh
+++ b/hack/ci-env.sh
@@ -119,9 +119,8 @@ dump_prow_artifacts_on_failure() {
     # is visible only in this pod's log -- the gateway just relays the text.
     kubectl logs deployment/litellm -n "${ns}" --tail=1000 > "${artifact_dir}/litellm.log" 2>&1 || true
     # The gateway capture above reads the pod's default container (platform-agent);
-    # a dropped port-forward stream is only visible from the auth sidecar's side.
-    kubectl logs deployment/platform-agent-gateway -c agent-api-auth -n "${ns}" --tail=2000 > "${artifact_dir}/agent-api-auth.log" 2>&1 || true
-    kubectl logs deployment/platform-agent-credential-proxy -c envoy-credential-proxy -n "${ns}" --tail=2000 > "${artifact_dir}/envoy-credential-proxy.log" 2>&1 || true
+    # a dropped port-forward stream is only visible from the envoy sidecar's side.
+    kubectl logs deployment/platform-agent-gateway -c envoy-credential-proxy -n "${ns}" --tail=2000 > "${artifact_dir}/envoy-credential-proxy.log" 2>&1 || true
     # Konnectivity/tunnel churn shows up as kube-system events, not in "${ns}".
     kubectl get events -n kube-system --sort-by=.lastTimestamp 2>&1 | tail -100 > "${artifact_dir}/kube-system-events.txt" 2>&1 || true
 
--- a/tests/e2e/test_agent_fleet_audit.py
+++ b/tests/e2e/test_agent_fleet_audit.py
@@ -188,7 +188,7 @@ if full_name.lower() != '{github_repo}'.lower():
 print(f"Successfully authenticated and queried repository: {{full_name}}")
 """
 
-    cmd = [
+    base_exec = [
         "kubectl",
         "exec",
         "-n",
@@ -197,10 +197,12 @@ print(f"Successfully authenticated and queried repository: {{full_name}}")
         "-c",
         container_name,
         "--",
-        "python3",
-        "-c",
-        script,
     ]
+    if container_name == "shell":
+        cmd = base_exec + ["runuser", "-u", "agent", "--", "python3", "-c", script]
+    else:
+        cmd = base_exec + ["python3", "-c", script]
+
     try:
         proc_start = subprocess.run(cmd, capture_output=True, text=True, timeout=90)
         assert proc_start.returncode == 0, (
--- a/tests/e2e/test_agent_fleet_audit.py
+++ b/tests/e2e/test_agent_fleet_audit.py
@@ -46,14 +46,13 @@ def test_github_token_minting_and_connectivity(
     if not gke_cluster_name or not github_repo:
         pytest.fail("GKE cluster name and GITHUB_REPO are required for live GitHub connectivity probe.")
 
-    # Find running platform-agent pod with availability wait
+    # 1. Wait for running shell sandbox pod (StatefulSet app=<name>-shell, container: shell)
     deadline = time.time() + _POD_WAIT_TIMEOUT_SECONDS
     pod_name = ""
     container_name = "shell"
     agent_name = os.environ.get("AGENT_SERVICE_NAME", "platform-agent")
 
     while time.time() < deadline:
-        # 1. Shell sandbox pod (StatefulSet app=<name>-shell, container: shell)
         proc_pod = subprocess.run(
             [
                 "kubectl",
@@ -74,38 +73,11 @@ def test_github_token_minting_and_connectivity(
             pod_name = proc_pod.stdout.strip()
             container_name = "shell"
             break
+        time.sleep(_POD_POLL_INTERVAL_SECONDS)
 
-        # 2. Search for any pod running the 'shell' container
-        proc_all = subprocess.run(
-            [
-                "kubectl",
-                "get",
-                "pods",
-                "-n",
-                agent_namespace,
-                "--field-selector=status.phase=Running",
-                "-o",
-                "json",
-            ],
-            capture_output=True,
-            text=True,
-        )
-        if proc_all.returncode == 0 and proc_all.stdout.strip():
-            try:
-                items = json.loads(proc_all.stdout).get("items", [])
-                for item in items:
-                    c_names = [c.get("name") for c in item.get("spec", {}).get("containers", [])]
-                    if "shell" in c_names:
-                        pod_name = item.get("metadata", {}).get("name", "")
-                        container_name = "shell"
-                        break
-            except Exception:
-                pass
-        if pod_name:
-            break
-
-        # 3. Fallback for legacy single-pod layout: gateway pod with container platform-agent
-        proc_pod = subprocess.run(
+    # 2. Fallback for legacy single-pod layout: gateway pod with container platform-agent
+    if not pod_name:
+        proc_gw = subprocess.run(
             [
                 "kubectl",
                 "get",
@@ -121,40 +93,13 @@ def test_github_token_minting_and_connectivity(
             capture_output=True,
             text=True,
         )
-        if proc_pod.returncode == 0 and proc_pod.stdout.strip():
-            pod_name = proc_pod.stdout.strip()
+        if proc_gw.returncode == 0 and proc_gw.stdout.strip():
+            pod_name = proc_gw.stdout.strip()
             container_name = "platform-agent"
-            break
-
-        # Fallback to checking any running pod in agent namespace with 'agent' or 'gateway' in name
-        proc_pod_all = subprocess.run(
-            [
-                "kubectl",
-                "get",
-                "pod",
-                "-n",
-                agent_namespace,
-                "--field-selector=status.phase=Running",
-                "-o",
-                "jsonpath={.items[*].metadata.name}",
-            ],
-            capture_output=True,
-            text=True,
-        )
-        if proc_pod_all.returncode == 0 and proc_pod_all.stdout.strip():
-            candidate_pods = [
-                p for p in proc_pod_all.stdout.split()
-                if ("agent" in p or "gateway" in p) and not any(k in p for k in ("minter", "minty", "operator", "hindsight"))
-            ]
-            if candidate_pods:
-                pod_name = candidate_pods[0]
-                container_name = "platform-agent"
-                break
-        time.sleep(_POD_POLL_INTERVAL_SECONDS)
 
     if not pod_name:
         pytest.fail(
-            f"No running agent shell or gateway pod found in namespace '{agent_namespace}' within {_POD_WAIT_TIMEOUT_SECONDS}s."
+            f"No running agent shell or legacy gateway pod found in namespace '{agent_namespace}' within {_POD_WAIT_TIMEOUT_SECONDS}s."
         )
 
     # Refresh credentials via broker and query repository via read-only GET API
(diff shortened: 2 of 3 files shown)


hack/ci-env.sh
```diff
@@ -119,8 +119,9 @@ dump_prow_artifacts_on_failure() {
     # is visible only in this pod's log -- the gateway just relays the text.
     kubectl logs deployment/litellm -n "${ns}" --tail=1000 > "${artifact_dir}/litellm.log" 2>&1 || true
     # The gateway capture above reads the pod's default container (platform-agent);
-    # a dropped port-forward stream is only visible from the envoy sidecar's side.
-    kubectl logs deployment/platform-agent-gateway -c envoy-credential-proxy -n "${ns}" --tail=2000 > "${artifact_dir}/envoy-credential-proxy.log" 2>&1 || true
+    # a dropped port-forward stream is only visible from the auth sidecar's side.
+    kubectl logs deployment/platform-agent-gateway -c agent-api-auth -n "${ns}" --tail=2000 > "${artifact_dir}/agent-api-auth.log" 2>&1 || true
+    kubectl logs deployment/platform-agent-credential-proxy -c envoy-credential-proxy -n "${ns}" --tail=2000 > "${artifact_dir}/envoy-credential-proxy.log" 2>&1 || true
```
eLeontev: **Note**: Kept here for synchronization after the pod split in #913 (the gateway now runs `agent-api-auth`, while `envoy-credential-proxy` is in its own deployment).

This can be carved out into a separate PR if desired so as not to be blocked by or bypass longer CI/review verification checks.

tests/e2e/test_agent_fleet_audit.py
```diff
@@ -153,7 +195,7 @@ def test_github_token_minting_and_connectivity(
         agent_namespace,
         pod_name,
         "-c",
-        "platform-agent",
+        container_name,
```
mplakhtiy, line 143: **HIGH — the probe now runs as root and executes a script the model can rewrite.**

Moving the exec target from the gateway's `platform-agent` container to `-c shell` in the sandbox pod is a privilege change, not just an address change:

- the gateway container runs at uid 10000 (`USER 10000:10000` in `deploy/docker/Dockerfile` plus `hardenedSecurityContext()`);
- the sandbox pod sets **no** `runAsUser`/`runAsNonRoot` at pod or container level — `buildShellSandboxStatefulSet` says so explicitly ("No pod-level securityContext, and that is a decision rather than an omission", sshd needs to fork as uid 0), and `buildShellSandboxContainer` sets only a seccomp profile and a `NET_RAW` drop;
- `deploy/sandbox/Dockerfile` declares no `USER`.

So this `kubectl exec -c shell -- python3 -c <script>` lands as root. The script then does `find /opt -name github_token_refresh.py` and executes `.splitlines()[0]`. **Every** match is uid-1000-writable: the staged tree is `chown -R agent:agent /opt/defaults` (`deploy/sandbox/Dockerfile:328`), and `deploy/sandbox/entrypoint.sh:209-217` chowns the copies it lays down in `/opt/data` and `/opt/data/profiles/platform` to the same account. There is no root-owned copy for `find` to land on instead, so which one wins does not matter.

`deploy/sandbox/Dockerfile:334-339` names this exact boundary — *"the `hermes` principal must never execute anything under `/opt/data`, because uid 1000 owns that mount point... What `hermes` runs comes off PATH, out of root-owned `/opt/credential-proxy/bin`."* The nightly gate now does the thing that comment forbids, with more privilege than `hermes` has.

**Failure scenario.** The sandbox is the pod where model-authored code runs. Anything that lands a modified `github_token_refresh.py` on the data volume — a bad agent turn, a prompt-injected repository — gets it executed as root on the next nightly RC run, in the pod holding the root-owned CLI shims (`/opt/credential-proxy/bin/{gh,git,kubectl,gcloud}`) and the sshd host keys the entrypoint deliberately keeps away from uid 1000 (`chown root:root "$key"`, `entrypoint.sh:338`).

The `find`-and-run pattern is pre-existing; running it as root is not. Note that pinning a fixed path does **not** fix it — the staged copy is agent-owned too. Dropping privilege does: `runuser -u agent -- python3 -c ...`. Worth confirming shim resolution and `CREDENTIAL_PROXY_TOKEN_FILE` still work under `runuser`; they should, both are env-driven.

Reached independently and each link verified; this is the same finding the bot posted.

tests/e2e/test_agent_fleet_audit.py
```diff
@@ -49,7 +49,11 @@ def test_github_token_minting_and_connectivity(
     # Find running platform-agent pod with availability wait
     deadline = time.time() + _POD_WAIT_TIMEOUT_SECONDS
     pod_name = ""
+    container_name = "shell"
+    agent_name = os.environ.get("AGENT_SERVICE_NAME", "platform-agent")
+
     while time.time() < deadline:
+        # 1. Shell sandbox pod (StatefulSet app=<name>-shell, container: shell)
```
mplakhtiy: **MEDIUM — the gateway fallback pre-empts the new shell-pod wait, so the 120s budget never applies.**

The discovery ladder runs all four steps inside one iteration of the `while time.time() < deadline` loop:

1. `-l app=<name>-shell` -> break
2. any Running pod in the namespace with a container named `shell` -> break
3. `-l app=<name>-gateway`, container `platform-agent` -> break
4. name-substring match on `agent`/`gateway` -> break

Step 3 succeeds on the first pass of any healthy install, because the gateway Deployment is always up. So when the sandbox StatefulSet is *not yet* Running — a fresh install with two PVCs still binding, a node event, a rolling restart — the loop never sleeps and never retries: it takes the gateway on iteration one. The 120-second availability wait this change is built around is unreachable in exactly the case it exists for.

What it falls through to cannot work. `deploy/docker/Dockerfile:1672` is a build-stage guard that fails the image build if `/opt/credential-proxy` or any of `gcloud kubectl gh git helm k9s yq` is present in the agent image, and the comment at line 1244 spells out that #737 removed all four shims. So `subprocess.run(['gh', ...])` in the gateway container raises `FileNotFoundError` on every current image.

**Failure scenario.** The release gate reports a Python traceback about a missing `gh` inside `platform-agent-gateway-...` instead of "the shell sandbox is not Running" — the same class of misdirection #1265 was filed about, on the one suite whose job is to say why staging did not promote.

Either fix suffices: give step 1 its own wait loop before the fallbacks are consulted, or gate steps 3-4 on the deadline having been exhausted.

tests/e2e/test_agent_fleet_audit.py
```diff
@@ -68,7 +72,39 @@ def test_github_token_minting_and_connectivity(
         )
         if proc_pod.returncode == 0 and proc_pod.stdout.strip():
             pod_name = proc_pod.stdout.strip()
+            container_name = "shell"
             break
+
```
mplakhtiy, line 77: **MEDIUM — step 2 scans the whole namespace with no install scoping.**

Step 2 lists every Running pod in `agent_namespace` and takes the first one with a container named `shell`, with no reference to `agent_name`. It is reached precisely when step 1's label-scoped lookup found nothing — i.e. when *this* install's sandbox is missing — which is also the only situation in which another install's sandbox can be picked up. `kubeagents-system` is a shared namespace in the composition, and `shellSandboxSelector` puts `app: <name>-shell` on every sandbox, so a second PlatformAgent in the namespace is a live mismatch rather than a hypothetical.

It also has no reason to exist: `shellSandboxSelector` is set unconditionally by `buildShellSandboxStatefulSet`, so there is no rendered layout where step 1 misses and step 2 hits. The `except Exception: pass` around the JSON parse hides whatever went wrong on the way.

Real but narrow. Simplest disposition is to delete step 2; if it stays, filter on `agent_name` and announce the fallback.

mplakhtiy (commented):
Adversarial review pass over the full diff, run in a context that did not write the change. Findings at MEDIUM and above are inline below; lower-severity items are omitted.

1 HIGH, 2 MEDIUM, all in `tests/e2e/test_agent_fleet_audit.py`. Every identifier the diff introduces checks out against the operator source — the container name, the port, the shell label, the shell container name and the broker Deployment name are all what the operator renders today.

tests/e2e/test_agent_fleet_audit.py
eLeontev: Fixed in `d7b7c7bf`. When targeting the sandbox `shell` container, the probe drops privileges to unprivileged `agent` (uid 1000) via `runuser -u agent -- python3 -c ...`. Verified on live nightly cluster that token refresh and read-only GitHub API queries succeed under uid 1000 without executing agent-writable scripts as root.

tests/e2e/test_agent_fleet_audit.py
eLeontev: Fixed in `f1a229be`. The 120s availability wait loop is now dedicated solely to `app=<name>-shell`. The legacy gateway check is only evaluated if the sandbox wait loop times out without finding a running shell pod.

tests/e2e/test_agent_fleet_audit.py
eLeontev: Fixed in `f1a229be`. Removed step 2 (unscoped namespace scan for `shell` container) and step 4 (name substring match) entirely, keeping discovery strictly scoped to `app=<name>-shell`.

hack/ci-env.sh
eLeontev: Reverted in `78de3a6f` and extracted to a standalone PR targeting Prow CI failure artifact collection, keeping this PR strictly focused on E2E test suite alignment.

mplakhtiy (approved):

Issue #1061, opened 2026-08-31
feat(terraform): adopt pre-existing Pub/Sub topics in lifecycle.sh to prevent 409 Conflict on re-runs


eLeontev:
### Summary

In `terraform/examples/full-install/lifecycle.sh`, `adopt_kms()` inspects GCP for pre-existing KMS keyrings and keys and imports them into Terraform state (`terraform import`) prior to running `terraform apply`. This makes apply idempotent against existing infrastructure.

However, no equivalent adoption exists for Pub/Sub topics.

If a Pub/Sub topic already exists in the GCP project (for example, if an operator pre-created the topic in Cloud Console while configuring the Google Chat App, or if an earlier install left the topic in the project), `lifecycle.sh apply` fails immediately with:
```text
Error: Error creating Topic: googleapi: Error 409: Resource already exists in the project (resource=<topic-name>).
  with module.chat_pubsub[0].google_pubsub_topic.chat_events
```

### Reproduction Steps

1. Manually create a Pub/Sub topic in GCP:
   ```bash
   gcloud pubsub topics create platform-agent-chat-events --project=<PROJECT_ID>
   ```
2. Run `install.sh` with `--enable-google-chat`.
3. The installation fails at step 12 during `terraform apply` with HTTP 409 Conflict.

### Proposed Solution

Add `adopt_pubsub()` to `terraform/examples/full-install/lifecycle.sh` (called as part of `apply` alongside `adopt_kms`):

```bash
adopt_pubsub() {
  local project topic_name address
  load_state
  project=$(tfvar project_id)

  if [[ "$(tfvar enable_google_chat)" == "true" ]]; then
    topic_name=$(tfvar chat_topic_name)
    address="module.chat_pubsub[0].google_pubsub_topic.chat_events"
    if ! in_state "$address"; then
      if gcloud pubsub topics describe "$topic_name" --project "$project" >/dev/null 2>&1; then
        log "adopting existing Pub/Sub topic: projects/$project/topics/$topic_name"
        [[ -f "$OVERRIDE_FILE" ]] || with_override
        if terraform import -input=false "$address" "projects/$project/topics/$topic_name" >/dev/null 2>&1; then
          log "Pub/Sub topic adopted into state"
        else
          warn "could not import $address; apply may fail with 409"
        fi
      fi
    fi
  fi
}
```
This ensures zero-friction idempotency when enabling Google Chat against existing topics.

PR #1273, opened 2026-09-08
fix(installer): treat empty state instances as unmanaged cluster and adopt GSA

sergiu1spiridon:
## Summary

When installing `kube-agents` onto an existing cluster, `install.sh` queries the remote Terraform state to decide whether Terraform created the cluster (`create_cluster = true`) or should adopt an external one (`create_cluster = false`). If an interrupted or dry-run apply left a managed resource record in state with an empty instances array (`"instances": []`), the installer previously concluded Terraform owned the cluster, setting `create_cluster = true` and failing with a 409 conflict. Additionally, if the Google Service Accounts (`kubeagents-platform-gsa`, `kubeagents-github-minter-gsa`, or `kubeagents-litellm-gsa`) already existed in the project, Terraform failed with a 409 conflict during apply. Furthermore, when `create_cluster` flips to `false` on a state that retained cluster KMS resources from an interrupted run, Terraform would plan their destruction. This PR fixes the cluster ownership probe, ensures unmanaged cluster KMS is forgotten rather than destroyed, adds retries for CMEK key creation during API propagation, and adds pre-existing service accounts to adoption in `lifecycle.sh`.

## Why This Change

1. Installing `kube-agents` onto an existing cluster that had an interrupted prior attempt caused `create_cluster = true` and 409 cluster creation failures.
2. Pre-existing Google Service Accounts failed during `terraform apply` with 409 creation conflicts.
3. Flipping `create_cluster` to `false` when cluster KMS resources were in state took them from `count = 1` to `count = 0`, causing Terraform to plan their destruction and schedule Cloud KMS key versions for destruction, breaking the live cluster's etcd database encryption.
4. Enabling `cloudkms.googleapis.com` in `ensure_existing_cluster_cmek` races with immediate keyring/key creation before API activation propagates.

## What Changed

- `scripts/installer/installer_common.sh`: Updated `tf_state_has_cluster` to require `len(r.get("instances", [1])) > 0` before treating a managed cluster record as managing a live cluster.
- `terraform/examples/full-install/lifecycle.sh`:
  - Added `forget_unmanaged_cluster_kms()` to forget cluster KMS resources (`gke_key[0]`, `gke_keyring[0]`, `gke_kms_binding[0]`, `gke_service_agent[0]`) from state via `terraform state rm` and restore any `DESTROY_SCHEDULED` key versions before apply when `create_cluster = false`.
  - Added `module.kube_agents_iam.google_service_account.agent`, `module.github_minter[0].google_service_account.minter`, and `module.litellm_vertex_iam[0].google_service_account.agent` to `adopt_kms()` targets so pre-existing service accounts are adopted rather than failing with 409 creation conflicts.
  - Added `forget_adopted_service_accounts()` to forget adopted pre-existing service accounts on `destroy` via `.adopted_resources` tracking (persisted to GCS state bucket) so they are kept in GCP.
  - Fixed `gcloud iam service-accounts describe` lookup to pass `${id##*/}` (the email address) instead of the full resource path.
- `install.sh`:
  - Added retry loop and validation in `ensure_existing_cluster_cmek` to handle `cloudkms.googleapis.com` API enablement propagation before creating the keyring and key.
  - Declared `CMEK_RETRY_ATTEMPTS` and `CMEK_RETRY_DELAY_SECONDS` at top of file, replacing bare literals per engineering rules.
- `INSTALL.md`: Updated teardown asymmetries descriptions (L163, L623-628) to include preserving adopted pre-existing service accounts during destroy.
- `terraform/examples/full-install/README.md`: Documented service account adoption and teardown asymmetry handling.
- `tests/test_installer_common.py`: Added unit test `test_empty_instances_managed_entry_is_not_ours`.

## Context

Observed during installation and testing on existing GKE clusters in multi-cluster environments (`standard-agent-test` and `autopilot-cluster-1`).

## Testing

- `python3 -m unittest tests/test_installer_common.py`: All 66 tests passing, including `test_empty_instances_managed_entry_is_not_ours`.
- `make docs-check`: Passed with zero warnings or errors.
- Prettier on `INSTALL.md` & `README.md`: Formatting verified.

### Live validation

Exercised against two distinct live environments:

1. **GKE Standard (`standard-agent-test` in `sergiuspiridon-gkedemos`, `us-east4-a`)**:
   - Prior interrupted apply left an empty managed entry in GCS state, causing `install.sh` to generate `create_cluster = true` and 409 on cluster creation. Pre-existing `kubeagents-platform-gsa` and `kubeagents-github-minter-gsa` caused 409 conflicts during apply.
   - Result: `install.sh` detected no live instances, setting `create_cluster = false`. `lifecycle.sh adopt-kms` cleanly adopted pre-existing GSAs via `${id##*/}`. Apply succeeded.

2. **GKE Autopilot (`autopilot-cluster-1` in `quick-test-508010`, `us-central1`)**:
   - Installed onto pre-existing Autopilot cluster with CMEK database encryption (`k8s-secret-encryption-key` in `platform-agent-keyring`).
   - Prior failed run had initialized KMS resources in state. On retry with `create_cluster = false`, `forget_unmanaged_cluster_kms` successfully removed the unmanaged cluster KMS addresses from state via `terraform state rm`, preventing key version destruction. Verified key version 1 remained in `ENABLED` state in Cloud KMS.
   - `adopt_kms` successfully verified and adopted pre-existing `kubeagents-platform-gsa` and `kubeagents-github-minter-gsa` via `${id##*/}` without 409 conflicts.
   - End-to-end install succeeded (`/tmp/kube-agents-install-report.json`: `status: SUCCESS`).
   - Workload verification in `kubeagents-system`:
     - `github-token-minter`: 2/2 Running
     - `kube-agents-controller-manager`: 1/1 Running
     - `litellm`: 2/2 Running
     - `platform-agent-credential-proxy`: 1/1 Running
     - `platform-agent-gateway`: 3/3 Running
     - `platform-agent-shell-0`: 1/1 Running
     - `PlatformAgent` CR: Phase `Ready` (`Gateway, shell sandbox and credential broker are all ready`).

- **Coverage limits & Cleanup**:
  - What was NOT covered: Did not exercise multi-project scoped cluster service account pools (`scoped_clusters`), as the live test environment uses standard single-cluster identity.
  - Artifact cleanup: Temporary provider overrides (`providers_lifecycle_override.tf`) were cleaned up via the trap; test state remains isolated to the designated test project.

## Self-Review

Passes were run in separate clean subagent contexts handed only the branch diff against `upstream/main`:

- **Adversarial review** (clean subagent context):
  - *Finding*: `gcloud iam service-accounts describe` was passed full resource ID `$id` rather than account email `${id##*/}`, causing HTTP 404 and silent skip. *Disposition*: Fixed by stripping prefix with `${id##*/}`.
  - *Finding*: `create_cluster` flipping to `false` causes Terraform to plan destruction of residual cluster KMS keys (`google_kms_crypto_key.gke_key[0]`). *Disposition*: Fixed by adding `forget_unmanaged_cluster_kms` to remove unmanaged cluster KMS from state before apply and call `restore_key_versions`.
  - *Finding*: `cloudkms.googleapis.com` enablement propagation race in `ensure_existing_cluster_cmek`. *Disposition*: Fixed by adding retry loop to keyring and key creation.
  - *Finding*: `retry 6 5` in `ensure_existing_cluster_cmek` hardcoded retry count and delay literals. *Disposition*: Fixed by declaring `CMEK_RETRY_ATTEMPTS` and `CMEK_RETRY_DELAY_SECONDS` at the top of `install.sh`.
  - *Finding*: `module.github_minter[0].google_service_account.minter` and Vertex LiteLLM GSA were omitted from targets. *Disposition*: Fixed by adding both to adoption targets.
  - *Finding*: Missing `instances` key in older state formats. *Disposition*: Handled safely using `r.get("instances", [1])`.
- **Docs-drift review** (clean subagent context):
  - *Finding*: `INSTALL.md` still stated destroy handles "four asymmetries" / "four teardown asymmetries", omitting adopted service account forgetting. *Disposition*: Fixed by updating `INSTALL.md` (L163, L623-628) to describe teardown asymmetries including preserving adopted pre-existing service accounts in GCP.
  - *Finding*: Checked `scripts/installer/README.md` and site docs for references to cluster ownership probing and GSA adoption. *Disposition*: No drift; internal state reconciliation mechanism changes do not alter documented user-facing CLI flags or configuration parameters. `make docs-check` passes cleanly.

## Risk & Rollout

Low risk. Only affects installer tfvars generation when probing GCS state and pre-apply adoption/forgetting. Protects existing cluster CMEK encryption keys from unintended destruction. Does not alter runtime agent or operator code paths.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -14,6 +14,9 @@
 
 set -Eeuo pipefail
 
+readonly CMEK_RETRY_ATTEMPTS=6
+readonly CMEK_RETRY_DELAY_SECONDS=5
+
 # ─── ANSI Colors & Terminal Responsive Helpers ─────────────────────────────────
 # A function because scripts/installer/common.sh defines the same variables
 # unconditionally: sourcing it would re-enable colour under NO_COLOR or in a pipe,
@@ -1627,11 +1630,13 @@ ensure_existing_cluster_cmek() {
   print_info "Enabling CMEK database encryption on existing cluster '$cluster_name' (key: $key_resource)..."
   gcloud services enable cloudkms.googleapis.com --project="$project_id"
   if ! gcloud kms keyrings describe "$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
-    retry 6 5 gcloud kms keyrings create "$keyring" --location="$kms_location" --project="$project_id" 2>/dev/null || true
+    retry "$CMEK_RETRY_ATTEMPTS" "$CMEK_RETRY_DELAY_SECONDS" \
+      gcloud kms keyrings create "$keyring" --location="$kms_location" --project="$project_id" 2>/dev/null || true
   fi
   if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
-    retry 6 5 gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
-      --purpose="encryption" --project="$project_id" 2>/dev/null || true
+    retry "$CMEK_RETRY_ATTEMPTS" "$CMEK_RETRY_DELAY_SECONDS" \
+      gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
+        --purpose="encryption" --project="$project_id" 2>/dev/null || true
   fi
   if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
     print_error "Failed to create or verify Cloud KMS key: $key_resource"
--- a/install.sh
+++ b/install.sh
@@ -1627,25 +1627,11 @@ ensure_existing_cluster_cmek() {
   print_info "Enabling CMEK database encryption on existing cluster '$cluster_name' (key: $key_resource)..."
   gcloud services enable cloudkms.googleapis.com --project="$project_id"
   if ! gcloud kms keyrings describe "$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
-    local kr_attempt=0
-    while [ "$kr_attempt" -lt 5 ]; do
-      if gcloud kms keyrings create "$keyring" --location="$kms_location" --project="$project_id" 2>/dev/null; then
-        break
-      fi
-      kr_attempt=$((kr_attempt + 1))
-      sleep 2
-    done
+    retry 6 5 gcloud kms keyrings create "$keyring" --location="$kms_location" --project="$project_id" 2>/dev/null || true
   fi
   if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
-    local k_attempt=0
-    while [ "$k_attempt" -lt 5 ]; do
-      if gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
-        --purpose="encryption" --project="$project_id" 2>/dev/null; then
-        break
-      fi
-      k_attempt=$((k_attempt + 1))
-      sleep 2
-    done
+    retry 6 5 gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
+      --purpose="encryption" --project="$project_id" 2>/dev/null || true
   fi
   if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
     print_error "Failed to create or verify Cloud KMS key: $key_resource"
--- a/install.sh
+++ b/install.sh
@@ -1626,9 +1626,31 @@ ensure_existing_cluster_cmek() {
   local key_resource="projects/${project_id}/locations/${kms_location}/keyRings/${keyring}/cryptoKeys/${key}"
   print_info "Enabling CMEK database encryption on existing cluster '$cluster_name' (key: $key_resource)..."
   gcloud services enable cloudkms.googleapis.com --project="$project_id"
-  gcloud kms keyrings create "$keyring" --location="$kms_location" --project="$project_id" 2>/dev/null || true
-  gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
-    --purpose="encryption" --project="$project_id" 2>/dev/null || true
+  if ! gcloud kms keyrings describe "$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
+    local kr_attempt=0
+    while [ "$kr_attempt" -lt 5 ]; do
+      if gcloud kms keyrings create "$keyring" --location="$kms_location" --project="$project_id" 2>/dev/null; then
+        break
+      fi
+      kr_attempt=$((kr_attempt + 1))
+      sleep 2
+    done
+  fi
+  if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
+    local k_attempt=0
+    while [ "$k_attempt" -lt 5 ]; do
+      if gcloud kms keys create "$key" --keyring="$keyring" --location="$kms_location" \
+        --purpose="encryption" --project="$project_id" 2>/dev/null; then
+        break
+      fi
+      k_attempt=$((k_attempt + 1))
+      sleep 2
+    done
+  fi
+  if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
+    print_error "Failed to create or verify Cloud KMS key: $key_resource"
+    return 1
+  fi
   local project_number service_agent
   project_number=$(gcloud projects describe "$project_id" --format="value(projectNumber)")
   service_agent="service-${project_number}@container-engine-robot.iam.gserviceaccount.com"
--- a/scripts/installer/installer_common.sh
+++ b/scripts/installer/installer_common.sh
@@ -642,7 +642,9 @@ try:
 except Exception:
     sys.exit(1)
 managed = any(
-    r.get("type") == "google_container_cluster" and r.get("mode") == "managed"
+    r.get("type") == "google_container_cluster"
+    and r.get("mode") == "managed"
+    and len(r.get("instances", [1])) > 0
     for r in doc.get("resources", [])
 )
 sys.exit(0 if managed else 1)
--- a/terraform/examples/full-install/lifecycle.sh
+++ b/terraform/examples/full-install/lifecycle.sh
@@ -543,10 +543,43 @@ forget_kms() {
   done
 }
 
+# When create_cluster = false, the gke-cluster module manages no cluster KMS
+# resources (local.manage_kms = false), giving them count = 0. If a prior
+# attempt created the KMS key and keyring before failing, or if an apply was
+# interrupted, those resources remain in state. Applying with create_cluster=false
+# would therefore plan their destruction, scheduling the key versions for
+# destruction in Cloud KMS and corrupting the live cluster's etcd encryption.
+# Forgetting these keeps them untouched in GCP, and restore_key_versions recovers
+# any version already pending destruction.
+forget_unmanaged_cluster_kms() {
+  [[ "$(tfvar create_cluster)" == "false" ]] || return 0
+  load_state
+  local address
+  for address in \
+    "module.gke_cluster.google_kms_crypto_key.gke_key[0]" \
+    "module.gke_cluster.google_kms_key_ring.gke_keyring[0]" \
+    "module.gke_cluster.google_kms_crypto_key_iam_member.gke_kms_binding[0]" \
+    "module.gke_cluster.google_project_service_identity.gke_service_agent[0]"; do
+    in_state "$address" || continue
+    log "forgetting $address (create_cluster=false; kept in GCP so apply does not destroy it)"
+    terraform state rm "$address" >/dev/null 2>&1 ||
+      warn "could not forget $address; its key versions may be scheduled for destruction"
+  done
+
+  local project location keyring key id
+  project=$(tfvar project_id)
+  location=$(sed -E 's/-[a-z]$//' <<<"$(tfvar location)")
+  keyring=$(tfvar kms_keyring_name)
+  key=$(tfvar kms_key_name)
+  id="projects/$project/locations/$location/keyRings/$keyring/cryptoKeys/$key"
+  restore_key_versions "$id" "$location" "$project"
+}
+
 case "${1:-}" in
   adopt-kms)
     shift
     ensure_init
+    forget_unmanaged_cluster_kms
     adopt_kms
     ;;
   plan)
@@ -576,6 +609,7 @@ case "${1:-}" in
     shift
     ensure_init
     guard_cluster_ownership
+    forget_unmanaged_cluster_kms
     adopt_kms
     adopt_pubsub
     log "terraform apply"
--- a/terraform/examples/full-install/lifecycle.sh
+++ b/terraform/examples/full-install/lifecycle.sh
@@ -260,6 +260,11 @@ adopt_kms() {
     )
   fi
 
+  local agent_gsa="kubeagents-platform-gsa"
+  targets+=(
+    "module.kube_agents_iam.google_service_account.agent	service_account	projects/$project/serviceAccounts/$agent_gsa@$project.iam.gserviceaccount.com"
+  )
+
   # Skipping both halves — cluster KMS (create_cluster or database encryption
   # off) and the minter — leaves targets empty, and macOS's bash 3.2 treats an
   # empty array expansion as unbound under `set -u`. The ${arr[@]+...} form
@@ -284,6 +289,8 @@ adopt_kms() {
       pubsub_sub)   gcloud pubsub subscriptions describe "${id##*/}" \
                       --project "$project" >/dev/null 2>&1 || continue ;;
       logging_sink) gcloud logging sinks describe "${id##*/}" \
+                      --project "$project" >/dev/null 2>&1 || continue ;;
+      service_account) gcloud iam service-accounts describe "$id" \
                       --project "$project" >/dev/null 2>&1 || continue ;;
     esac
 
--- a/terraform/examples/full-install/lifecycle.sh
+++ b/terraform/examples/full-install/lifecycle.sh
@@ -236,12 +236,14 @@ adopt_kms() {
   fi
 
   if [[ "$(tfvar enable_github_minter)" == "true" ]]; then
-    local minter_keyring minter_key
+    local minter_keyring minter_key minter_gsa
     minter_keyring=$(tfvar github_minter_kms_keyring)
     minter_key=$(tfvar github_minter_kms_key)
+    minter_gsa="kubeagents-github-minter-gsa"
     targets+=(
       "module.github_minter[0].google_kms_key_ring.minter	keyring	projects/$project/locations/$location/keyRings/$minter_keyring"
       "module.github_minter[0].google_kms_crypto_key.minter	key	projects/$project/locations/$location/keyRings/$minter_keyring/cryptoKeys/$minter_key"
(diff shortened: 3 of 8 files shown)


terraform/examples/full-install/lifecycle.sh
```diff
@@ -285,6 +290,8 @@ adopt_kms() {
                       --project "$project" >/dev/null 2>&1 || continue ;;
       logging_sink) gcloud logging sinks describe "${id##*/}" \
                       --project "$project" >/dev/null 2>&1 || continue ;;
+      service_account) gcloud iam service-accounts describe "$id" \
+                      --project "$project" >/dev/null 2>&1 || continue ;;
     esac
```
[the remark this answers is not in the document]
sergiu1spiridon: Fixed in commit 3c68ba35 — updated `service_account` describe check to use `"${id##*/}"` to query by email address, and added `module.github_minter[0].google_service_account.minter` and `module.litellm_vertex_iam[0].google_service_account.agent` to the adoption list.

scripts/installer/installer_common.sh
```diff
@@ -642,7 +642,9 @@ try:
 except Exception:
     sys.exit(1)
 managed = any(
-    r.get("type") == "google_container_cluster" and r.get("mode") == "managed"
+    r.get("type") == "google_container_cluster"
+    and r.get("mode") == "managed"
+    and len(r.get("instances", [1])) > 0
     for r in doc.get("resources", [])
 )
 sys.exit(0 if managed else 1)
```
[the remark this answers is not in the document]
sergiu1spiridon: Fixed in commit 0eb2e2d5 — added `forget_unmanaged_cluster_kms` in `lifecycle.sh`, which runs before `terraform apply` whenever `create_cluster = false`. It removes `module.gke_cluster.google_kms_crypto_key.gke_key[0]`, `google_kms_key_ring.gke_keyring[0]`, `gke_kms_binding[0]`, and `gke_service_agent[0]` from state via `terraform state rm` (so Terraform does not plan their destruction when `count` drops to 0), and calls `restore_key_versions` to re-enable any key version left in `DESTROY_SCHEDULED`.

terraform/examples/full-install/lifecycle.sh
```diff
@@ -260,6 +262,18 @@ adopt_kms() {
...
+    local litellm_gsa="kubeagents-litellm-gsa"
+    targets+=(
+      "module.litellm_vertex_iam[0].google_service_account.agent	service_account	projects/$project/serviceAccounts/$litellm_gsa@$project.iam.gserviceaccount.com"
+    )
+  fi
+
+  local agent_gsa="kubeagents-platform-gsa"
+  targets+=(
+    "module.kube_agents_iam.google_service_account.agent	service_account	projects/$project/serviceAccounts/$agent_gsa@$project.iam.gserviceaccount.com"
+  )
```
mplakhtiy: **HIGH — an adopted service account is deleted by this install's `destroy`, so a second install's uninstall takes out the first install's identity**

Adoption is unconditional and cannot distinguish the two reasons a GSA of that name already exists: a destroy of *this* install that stopped partway (adopting is right), and *another live install in the same project* (adopting is a hijack).

The composition treats the second as ordinary — this file's own remote-state comment says the prefix defaults to `kube-agents/<cluster_name>` "so two installs in one project do not collide" (lines 61-63), and #1123 exists precisely because that scenario currently dies on this GSA's 409.

**Failure scenario.** Project P has install A, whose state owns `kubeagents-platform-gsa`. Install B is created in P with a different cluster name.

- *Today*: B's apply stops at `Error 409: Service account kubeagents-platform-gsa already exists`, with nothing destroyed.
- *With this change*: B's `adopt_kms` describes the account, finds it, and imports it into B's `module.kube_agents_iam.google_service_account.agent`. B's apply is then near-harmless — the Workload Identity binding is a `_member` and both installs default to the `kubeagents-system` namespace, so it reconciles onto the same value. **The damage lands later:** `uninstall.sh:494` runs `./lifecycle.sh destroy -auto-approve -input=false`, `forget_kms` covers only the four KMS addresses and nothing else, and `terraform destroy` deletes install A's platform GSA. A's pods lose Workload Identity with no message on A's side at all.

This is the same shape of regression the bot raised against the KMS half — a loud 409 replaced by silent destruction — transplanted onto a resource that, unlike a KMS key, GCP will happily delete. GCP refuses to delete a key ring, and `forget_kms` exists to keep even the crypto key out of the destroy. A service account has neither protection.

The same holds for `module.github_minter[0].google_service_account.minter` (line 246) and `module.litellm_vertex_iam[0].google_service_account.agent` (line 268).

**Verdict: CONFIRMED.** The mechanism is fully in the diff and the surrounding files; only the sequencing (a second install, then its uninstall) is a scenario rather than an inevitability.

Options, none prescribed: extend `forget_kms` to drop adopted service accounts before a destroy (costs you the ability to delete a GSA this install genuinely created); record the adoption so the destroy can tell the two apart; or gate the service-account arm behind an opt-in the front doors set only when they know the state is this install's own leftover. "Accepted, because the 409 is worse" is a legitimate answer too — but it should be written down rather than left implicit, and it changes what a reviewer of #1123 needs to know.

terraform/examples/full-install/lifecycle.sh
```diff
@@ -236,12 +236,14 @@ adopt_kms() {
   fi
 
   if [[ "$(tfvar enable_github_minter)" == "true" ]]; then
-    local minter_keyring minter_key
+    local minter_keyring minter_key minter_gsa
     minter_keyring=$(tfvar github_minter_kms_keyring)
     minter_key=$(tfvar github_minter_kms_key)
+    minter_gsa="kubeagents-github-minter-gsa"
```
mplakhtiy: **MEDIUM — the hardcoded GSA names diverge from open PR #1123, and merge order decides whether the divergence bites**

Today the three literals (`:242`, `:266`, `:272`) are correct: the composition never passes `service_account_id` to `module.kube_agents_iam` or `module.github_minter`, so both sit on the module defaults (`terraform/modules/kube-agents-iam/variables.tf:9`, `terraform/modules/github-minter/variables.tf:14`), and the LiteLLM one is itself hardcoded at `terraform/examples/full-install/main.tf:293`.

**#1123** (open, ready — "fix(terraform): expose the agent GSA id in full-install") adds an `agent_service_account_id` passthrough to the composition for exactly the collision that motivates the HIGH above. Once it merges, an install that sets it has `adopt_kms` still describing and importing the **default-named** account into `module.kube_agents_iam.google_service_account.agent`, whose configured `account_id` is now something else. `account_id` is ForceNew on `google_service_account`, so the next `terraform apply -auto-approve` plans destroy-and-recreate: the override's stated purpose — not touching the other install's GSA — is inverted into deleting it.

Neither PR is a superset of the other and neither will conflict textually, so nothing will surface this at merge time. Whichever lands second should read the name from `tfvar` with a fallback for #1123's `null` default rather than repeating the literal.

**Verdict: CONFIRMED conditional on #1123 merging.**

Smaller, same lines: `.agents/rules/core_engineering.md` §"No magic constants" names "resource names" explicitly and asks for `readonly NAME=...` after the shebang for Bash. `install.sh:50` already carries `LITELLM_GSA_DEFAULT_NAME="kubeagents-litellm-gsa"` at top-of-file — the same value as `:266` here, in the shape the rule asks for.

install.sh
```diff
@@ -1626,9 +1626,31 @@ ensure_existing_cluster_cmek() {
...
+        --purpose="encryption" --project="$project_id" 2>/dev/null; then
+        break
+      fi
+      k_attempt=$((k_attempt + 1))
+      sleep 2
+    done
+  fi
+  if ! gcloud kms keys describe "$key" --keyring="$keyring" --location="$kms_location" --project="$project_id" >/dev/null 2>&1; then
+    print_error "Failed to create or verify Cloud KMS key: $key_resource"
+    return 1
```
mplakhtiy, line 1643: **MEDIUM — the CMEK retry re-implements `retry()`, more quietly and with a fifth of its budget**

`scripts/installer/installer_common.sh:263` defines `retry <max_retries> <delay> <cmd...>`. `install.sh` already sources that file (lines 141-143) and already calls it — `retry 6 5 go run ./cmd/minty tools import-pk` at line 1898. These two new loops hand-roll the same control flow and differ from the helper in three ways that matter:

- The helper prints `⚠ [Retry n/max] Waiting Ns before next attempt...` to stderr. These loops are silent, so an operator watching a run that is about to spend ten seconds retrying sees nothing at all.
- The helper skips the sleep after the final attempt (`if [ $count -lt $max_retries ]`). These sleep unconditionally, so each burns a pointless 2 s at the end.
- The helper's one existing use against a slow GCP operation is `6 5` ≈ 25 s. These are `5 × 2` ≈ 8 s of sleeping.

That last one is a substantive doubt about the fix, not only about its shape. The failure named in the PR body is `cloudkms.googleapis.com` enablement propagating after `gcloud services enable` returns; an eight-second ceiling with no backoff is thin for that, and the new `return 1` below turns a miss into a hard install abort. What would settle it: timing a keyring create against a project where the KMS API was enabled seconds earlier.

**CONFIRMED** for the duplication; **PLAUSIBLE** for the budget being too short. Calling `retry` is a smaller diff than what is here now and restores the progress output — pick the budget deliberately while you are in it.

The abort itself is an improvement, for the record: the old `|| true` let the run continue to `gcloud kms keys add-iam-policy-binding`, which fails under `set -e` anyway with a worse message.

Related, smaller (`:1629-1655`): every attempt runs with `2>/dev/null` and neither loop reports when it exhausts. If the **keyring** is what could not be created, both loops run to exhaustion and the operator sees only `Failed to create or verify Cloud KMS key: projects/…/cryptoKeys/k8s-secret-encryption-key` — which names the key and says nothing about the keyring or the reason. Capture the last attempt's stderr, or check the keyring separately after its loop so the message names the resource that actually failed.

terraform/examples/full-install/lifecycle.sh
```diff
@@ -560,6 +609,7 @@ case "${1:-}" in
     shift
     ensure_init
     guard_cluster_ownership
+    forget_unmanaged_cluster_kms
```
mplakhtiy, line 698: **MEDIUM — nothing tests the new lifecycle wiring, and the repo already has the pattern for it**

The one test this PR adds is good and genuinely red-then-green (I verified it against pre-change code). Everything on the lifecycle side is untested:

- that `apply` calls `forget_unmanaged_cluster_kms` at all;
- that it calls it **before** `adopt_kms` — the ordering is what makes `adopt_kms`'s state re-read see the removals;
- that the `service_account` arm strips the id with `${id##*/}` — the exact bug commit `3c68ba35` fixed, which will regress silently if the arm is ever copied from a sibling.

`tests/test_reconcile_environment.py::LifecyclePlanTest::test_apply_adopts_pre_existing_pubsub_resources` is the precedent: it greps the `apply)` branch for `adopt_pubsub`, and it exists because #1061 was the same class of wiring regression.

The silent failure here is worse than #1061's — the regression restores the CMEK-key destruction this PR was written to prevent, and CI reports green. Three `assertIn`s against the `apply)` branch and one against the `service_account` arm.

**Verdict: CONFIRMED** (the candidate is the untested behaviour, not the absent test).

mplakhtiy (commented):
Adversarial pass over the full diff at `0eb2e2d5`, run in a subagent that did not write the change.

On the three open threads, checked at the head rather than taken from the replies:

- **`lifecycle.sh:305` — `describe` handed a resource path.** **Fixed.** Head line 302 reads `service_account) gcloud iam service-accounts describe "${id##*/}"`, matching the five sibling arms, and the full path still reaches `terraform import` on line 308.
- **`installer_common.sh:650` — `create_cluster` flip destroys the live cluster's CMEK key.** **Fixed, and complete for its class.** The four addresses `forget_unmanaged_cluster_kms` removes are exactly the full set of resources gated on `local.manage_kms` in `terraform/modules/gke-cluster/main.tf` (lines 22, 30, 37, 46) — nothing else in that module or the composition keys off `create_cluster` except the two `google_container_cluster` resources and the read-only data source, which are `guard_cluster_ownership`'s job. Forgetting `gke_kms_binding[0]` is load-bearing rather than incidental: destroying that member would revoke the GKE service agent's `cryptoKeyEncrypterDecrypter` on the live key, which breaks the cluster just as surely as scheduling the versions.

Also verified rather than assumed: the new unit test is genuinely red-then-green (`'rc=1' not found in 'rc=0'` against pre-change code), 66 + 158 tests pass locally, `shellcheck -x` is clean on both changed scripts, and every module address and GSA name in the diff resolves against Terraform source.

One HIGH and four MEDIUMs inline.

**What this pass did not cover:** no GCP project, so nothing was applied, planned or imported. The retry budget, the `terraform state rm` behaviour and the ForceNew claim are read from the provider contract and the source, not observed.

**MEDIUM — both places that enumerate this script's asymmetries are now incomplete.** `terraform/examples/full-install/lifecycle.sh:10-28` (a header comment, not a changed line, so it cannot be anchored) lists five numbered asymmetries "Terraform cannot express" and is printed verbatim as the script's usage text (`sed -n '2,46p'`). `terraform/examples/full-install/README.md:517` carries the same list as a two-column table.

This change adds **two more members of that exact family** — flipping `create_cluster` to `false` would destroy cluster KMS the state still holds, and a pre-existing service account makes the create 409 — and updates neither. The README table and the header comment document *behaviours*, not flags, and adding to that behaviour set is precisely what they are for.

If items are added to the header, `test_the_usage_range_still_covers_the_whole_header` in `tests/test_reconcile_environment.py` requires the `sed -n '2,46p'` range to move with them.

terraform/examples/full-install/lifecycle.sh
sergiu1spiridon: Fixed in commit fce964e3.

Adopted service accounts are now recorded in `.adopted_resources` (and synced to GCS if remote state is enabled) upon successful `terraform import`. When `lifecycle.sh destroy` runs, `forget_adopted_service_accounts` removes only those imported accounts from state (`terraform state rm`), keeping external service accounts intact in GCP. Accounts created by Terraform itself are not in `.adopted_resources` and are destroyed normally by `terraform destroy`.

terraform/examples/full-install/lifecycle.sh
sergiu1spiridon: Fixed in commit fce964e3.

Declared `DEFAULT_AGENT_GSA_NAME`, `DEFAULT_MINTER_GSA_NAME`, and `DEFAULT_LITELLM_GSA_NAME` constants at top-of-file per the engineering rules. `adopt_kms` now evaluates the agent GSA name via `tfvar_or_default agent_service_account_id "$DEFAULT_AGENT_GSA_NAME"`, which reads the passthrough variable once PR #1123 lands and safely falls back to the default when undeclared or unset.

install.sh
sergiu1spiridon: Fixed in commit fce964e3.

Replaced the hand-rolled retry loops with `retry 6 5 gcloud kms keyrings create ...` and `retry 6 5 gcloud kms keys create ...`, reusing the shared helper from `scripts/installer/installer_common.sh` with its standard 25 s retry window and stderr progress output.

terraform/examples/full-install/lifecycle.sh
sergiu1spiridon: Fixed in commit fce964e3.

Added unit tests in `tests/test_reconcile_environment.py::LifecyclePlanTest` asserting that `apply` runs `forget_unmanaged_cluster_kms` before `adopt_kms`, that the `service_account` branch describes with `${id##*/}`, and that `destroy` runs `forget_adopted_service_accounts` before `terraform destroy`.

mplakhtiy:
Superseded by #1327, which fixes the same two failures at their source: the service-account names, the release namespace and the CMEK key names become `install.env` configuration (`PLATFORM_AGENT_GSA_NAME`, `GITHUB_MINTER_GSA_NAME`, `LITELLM_GSA_NAME`, `NAMESPACE`, `GKE_DB_KMS_KEYRING`, `GKE_DB_KMS_KEY`) threaded through the generator and the composition, and the installer refuses an account another install owns before the apply instead of importing it. That avoids the destroy hazard raised in the HIGH thread above, where an adopted account is deleted by this install's uninstall out from under the other one.

Two pieces of this PR carry over into #1327 with credit: the `instances` check in `tf_state_has_cluster` (extended to match project, location and name) and `forget_unmanaged_cluster_kms` on the `create_cluster = false` apply path. The author has been informed. Closing this one.

PR #1275, opened 2026-09-08
fix(skills): enforce NetworkPolicy in two steps on non-DPv2 clusters (#931)

### LINKED ISSUE #931: gke-workload-security skills tell agents to enable the NetworkPolicy addon and call it enforcement

All four items below still reproduce. #932 fixed the same defect in `install.sh`, the `gke-cluster`
postcondition and README, and the drift SOP; it did not touch either skill file, and nothing has
since — the last commits on them are #569 (platform, 2026-08-07) and #441 (cluster, 2026-07-28).

The fixability constraint stated below is narrower than originally written. The `agents/cluster/`
copy can be fixed by editing it directly; only the platform copy is wiped by the sync. Details in
the "Fixability" section.

---

Both copies of the `gke-workload-security` skill — `agents/platform/skills/gke-workload-security/SKILL.md`
and `agents/cluster/skills/gke-workload-security/SKILL.md` — carry this under the heading
**"Enable Network Policy Enforcement"** (platform `SKILL.md:96-102`, cluster `SKILL.md:78-84`):

```bash
gcloud container clusters update <cluster-name> \
    --update-addons=NetworkPolicy=ENABLED \
    --region <region>
```

That command enables the Calico addon on the control plane. It does not enable node-level
enforcement, which needs a second call with `--enable-network-policy`. The repo's own drift SOP is
the authority on the distinction: `agents/platform/governance/fleet_consistency_drift_sop.md:127`
counts a cluster as `CALICO` only when `.networkPolicy.enabled == true` **and** the addon is not
disabled, and `:131` gives the two-command remediation in order.

The next section in both files is "Apply Default Deny Policy" (platform `:107`, cluster `:89`), so
an agent that follows the skill in order lands a NetworkPolicy object that GKE accepts and never
enforces, and reports success. That is the same silent-inertness `install.sh:1710-1730` and the
`gke-cluster` module postcondition (`terraform/modules/gke-cluster/main.tf:275`) exist to prevent.
Those two are now correct; the skills are the last places in the repository still handing out the
broken one-liner.

## Fixability

The platform copy cannot be fixed by editing its `SKILL.md`. `scripts/sync-upstream-skills.py:186`
rmtree's the local directory and copies `google/skills` verbatim on every sync, so local edits are
wiped. The supported channels are a `SKILL_FOOTERS` entry (`:32`, the mechanism `gke-networking`
uses at `:69`) or a fix upstream. Note that `inject_footer` only appends (`:117-119`): a footer can
add a correction after the wrong command, not replace it. A clean fix for the platform copy
therefore needs an upstream change, or a substitution mechanism the sync script does not have today.

The cluster copy is different, and the original text was wrong about it. `DEFAULT_TARGET_AGENTS` is
`["platform"]` (`:15`) and `SKILL_AGENT_OVERRIDES` is empty (`:16-19`), so both the prune loop
(`:165`) and the copy loop (`:178`) iterate `{"platform"}` only. `agents/cluster/skills/gke-workload-security/`
is never rmtree'd and never overwritten by the sync — an edit there survives.

That leaves the question of how the `agents/cluster/` copy is maintained at all. It was added by
#441 and no sync path touches it, so it drifts from upstream silently. That wants answering as part
of the fix.

## Items

1. Platform copy tells agents the addon call is enforcement — open.
2. Cluster copy tells agents the same — open.
3. No `SKILL_FOOTERS` entry for `gke-workload-security`, so a platform-side correction does not
   survive a sync — open. Neither file contains the footer marker today.
4. How the `agents/cluster/` copy is maintained is undefined — open.

Found by the adversarial and docs-drift review passes on #932, which fixes the same defect in
`install.sh`, the `gke-cluster` postcondition message, and the drift SOP. Left out of that PR
because the fix lives in the sync machinery rather than in the installer path.

shalinibhatiapl:
## Summary

Updates the `gke-workload-security` skills in both the Platform and Cluster agent blueprints to prescribe:
1. A proactive pre-flight probe command (`gcloud container clusters describe ... --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'`) and explicit decision ladder to inspect whether NetworkPolicy enforcement is already active or provided natively by Dataplane V2 before modifying cluster networking;
2. The mandatory two-step command sequence required to activate and enforce Calico NetworkPolicy on non-DPv2 GKE clusters.

Also adds an automated in-place substitution mechanism in `scripts/sync-upstream-skills.py` with accompanying unit tests so this platform skill correction persists across upstream skill synchronizations.

## Why This Change

On non-Dataplane V2 GKE Standard clusters, enabling NetworkPolicy requires two sequential operations: activating the Calico addon on the control plane (`--update-addons=NetworkPolicy=ENABLED`), and then enabling enforcement across the cluster nodes (`--enable-network-policy`).

Previously, both `agents/platform/skills/gke-workload-security/SKILL.md` and `agents/cluster/skills/gke-workload-security/SKILL.md` instructed agents to execute only `--update-addons=NetworkPolicy=ENABLED`. While this enabled the control plane addon, it left node enforcement inactive (`networkPolicy: null`), misleading agents into reporting that NetworkPolicy enforcement was active when packets were not being filtered. Furthermore, attempting to execute `--enable-network-policy` prior to enabling the addon fails with HTTP 400 from the GKE API, making the correct sequence critical.

Additionally, the skills previously placed a note about Dataplane V2 after the command block without any runtime probe command, which meant agents reading top-to-bottom would execute cluster update commands against DPv2 clusters before reading the note.

## What Changed

- `agents/platform/skills/gke-workload-security/SKILL.md`: Updated Section 3 ("3. Implement Network Policies") with:
  - Pre-flight probe command (`gcloud container clusters describe <cluster-name> --location <location> --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'`) and explicit decision ladder (skip if DPv2 `ADVANCED_DATAPATH` or already enabled).
  - Two-step Calico enablement sequence (`--update-addons=NetworkPolicy=ENABLED` followed by `--enable-network-policy`) and ordering rationale for non-DPv2 clusters.
  - Clarified on Step 2 that node pools may be recreated and the operation can take several minutes.
  - Removed obsolete trailing note in favor of the proactive pre-check.
- `agents/cluster/skills/gke-workload-security/SKILL.md`: Applied identical updates to Section 3 of the cluster agent's workload security skill.
- `scripts/sync-upstream-skills.py`: Added `SKILL_SUBSTITUTIONS` and `apply_substitutions()` to apply deterministic string replacements on synced upstream skills upon copying, ensuring the NetworkPolicy pre-check and two-step enablement sequence survive future runs of `scripts/sync-upstream-skills.py`. Documented that `agents/cluster/skills/` are repo-native templates managed independently.
- `scripts/test_sync_upstream_skills.py`: Added `ApplySubstitutionsTest` verifying substitution behavior, idempotency, safety, and assertions that both repo skill files contain the probe command, enforcement command, and updated snippet.

## Context

Closes #931

## Testing

- `python3 -m unittest scripts.test_sync_upstream_skills`: All 11 tests pass (including unit tests for `apply_substitutions` and assertions on both repo skill files).
- `PYTHONPATH=scripts python3 -m unittest scripts.test_check_prompt_assets`: All 52 tests pass.
- `make docs-check`: Passes with 0 errors and 0 warnings.
- `make docs-generate`: Clean.

### Live validation

Exercised against real GKE infrastructure in GCP project `shalinibhatia-gkedemos` to verify API behavior and cluster states:

1. **Cluster Provisioning:**
   Created temporary non-DPv2 cluster `test-netpol-val` in `us-east4-a` (`--no-enable-dataplane-v2`, `e2-standard-2`).
   *Observed Initial State:* `addonsConfig.networkPolicyConfig.disabled: true`, `networkPolicy: null`.
   *Verified Pre-check Command Output:* `gcloud container clusters describe test-netpol-val --zone us-east4-a --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'` returned empty values, confirming neither DPv2 nor active Calico enforcement, correctly directing the agent to the two-step enablement sequence.

2. **Negative Test (Enforcement Before Addon):**
   Attempted to enable node enforcement first:
   ```bash
   gcloud container clusters update test-netpol-val \
       --zone us-east4-a \
       --project shalinibhatia-gkedemos \
       --enable-network-policy
   ```
   *Observed Result:* Rejected by GKE API with HTTP 400:
   `ERROR: (gcloud.container.clusters.update) ResponseError: code=400, message=The network policy addon must be enabled before updating the nodes.`

3. **Step 1 Execution (Enable Control-Plane Addon):**
   ```bash
   gcloud container clusters update test-netpol-val \
       --zone us-east4-a \
       --project shalinibhatia-gkedemos \
       --update-addons=NetworkPolicy=ENABLED --quiet
   ```
   *Observed Cluster State:* `addonsConfig.networkPolicyConfig: {}`, but `networkPolicy` remained `null`. This confirmed the defect in #931: the addon is present on master, but node enforcement remains completely inactive.

4. **Step 2 Execution (Enable Node Enforcement):**
   ```bash
   gcloud container clusters update test-netpol-val \
       --zone us-east4-a \
       --project shalinibhatia-gkedemos \
       --enable-network-policy --quiet
   ```
   *Observed Cluster State:* Node pool recreated and cluster state updated to:
   ```yaml
   addonsConfig:
     networkPolicyConfig: {}
   networkPolicy:
     enabled: true
     provider: CALICO
   ```
   Full Calico network policy enforcement confirmed active. Re-running the pre-check probe returned `;True`, confirming active enforcement.

5. **Teardown & Cleanup:**
   Deleted `test-netpol-val` via `gcloud container clusters delete test-netpol-val --zone us-east4-a --quiet`. Verified `gcloud container clusters list` shows only the pre-existing `platform-agent-host`.

6. **Sync Script Scope:**
   The GKE cluster lifecycle and API state transitions above were live-tested on GCP; `scripts/sync-upstream-skills.py` substitution logic was validated via 11 unit tests in `scripts/test_sync_upstream_skills.py`.

## Self-Review

- **Review Passes Context:** Initial pre-PR review passes and subsequent revisions were conducted in the active pair-programming session.
- **Adversarial Review:**
  - *Review Bot Finding (#discussion_r3957155363):* Initial PR description claimed the skills advised agents to inspect `networkConfig.datapathProvider`, but the diff only had a passive trailing note without a probe command.
    *Disposition:* Fixed in `daf88ab7`. Added `gcloud container clusters describe ... --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'` and a clear decision tree *before* the update commands in both skill files, and updated the substitution logic and unit test assertions.
  - *Review Bot Finding (#discussion_r3957136352):* Initial PR description described substitutions as regex replacements when `str.replace` is used, and did not delineate live testing scope between cluster operations and the sync script.
    *Disposition:* Fixed in `daf88ab7`. Corrected terminology to string replacements and clarified testing scope.
  - *Reviewer Finding (#discussion_r3958981562):* Step 2 command comment understated node pool recreation as an addon restart.
    *Disposition:* Fixed in `44930862`. Adopted `install.sh:1742`'s exact phrasing (`node pools may be recreated; this can take several minutes`) across both skill files (`agents/platform/.../SKILL.md`, `agents/cluster/.../SKILL.md`), `scripts/sync-upstream-skills.py`, and added test assertions in `scripts/test_sync_upstream_skills.py`.
  - *Reviewer Finding (#discussion_r3958981570):* The pre-flight probe command used `--region <region>`, which fails with `NOT_FOUND` on zonal clusters.
    *Disposition:* Fixed in `44930862`. Switched probe flag from `--region <region>` to `--location <location>` across both skill files and `scripts/sync-upstream-skills.py` to support both zonal and regional clusters, and added test assertions in `scripts/test_sync_upstream_skills.py`.
  - *Sequence Order:* Evaluated whether `--enable-network-policy` could be run first or alone. Negative test proved the GKE API strictly enforces that the addon must be enabled first.
  - *Dataplane V2:* Evaluated interaction with DPv2 clusters. On DPv2 (`ADVANCED_DATAPATH`), NetworkPolicy is enforced natively by Cilium/eBPF from cluster creation and Calico addons cannot be enabled. Probing `datapathProvider` upfront prevents accidental update attempts.
  - *Sync Persistence:* Evaluated whether running `scripts/sync-upstream-skills.py` would overwrite the platform skill fix. Resolved by introducing an automated substitution pipeline with unit test coverage.
  - *Cluster Skill Architecture:* Verified whether cluster skills are synced by `sync-upstream-skills.py`. They are repo-native templates; documented and added assertion in unit tests.
  - *Coding Standards:* Verified `scripts/sync-upstream-skills.py` and `scripts/test_sync_upstream_skills.py` adhere to `.agents/rules/core_engineering.md` (constants declared at module scope, no magic values).

- **Docs-Drift Review:**
  - Verified no other files in `docs/` or `agents/` referenced the incomplete single-command invocation.
  - Validated doc link integrity and context budget via `make docs-check`.

## Risk & Rollout

Low risk. Changes are confined to agent markdown skill guidance, upstream skill sync scripts, and unit tests. No runtime controller logic, agent container configurations, or existing clusters are impacted.

### DIFF
--- a/agents/cluster/skills/gke-workload-security/SKILL.md
+++ b/agents/cluster/skills/gke-workload-security/SKILL.md
@@ -75,7 +75,25 @@ access Google Cloud APIs.
 Control traffic flow between Pods using Network Policies. By default, all
 traffic is allowed.
 
-**Enable Network Policy Enforcement:**
+**Check Network Policy Enforcement & Dataplane:**
+
+Before modifying cluster networking, inspect whether NetworkPolicy enforcement
+is already active or provided natively by Dataplane V2:
+
+```bash
+gcloud container clusters describe <cluster-name> \
+    --region <region> \
+    --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'
+```
+
+- If `datapathProvider` is `ADVANCED_DATAPATH` (Dataplane V2), NetworkPolicy
+  enforcement is built-in natively via eBPF/Cilium from cluster creation. Calico
+  addons cannot be enabled and are not needed.
+- If `networkPolicy.enabled` is `True`, Calico enforcement is already enabled on nodes.
+- If neither is active, enable Calico network policy enforcement using the two-step
+  sequence below.
+
+**Enable Network Policy Enforcement (non-DPv2 clusters):**
 
 Enabling network policy enforcement on clusters without Dataplane V2 requires
 two sequential commands in this order: first enable the Calico addon on the
@@ -94,9 +112,6 @@ gcloud container clusters update <cluster-name> \
     --enable-network-policy \
     --region <region>
 ```
-
-> [!NOTE]
-> If your cluster uses Dataplane V2 (`--enable-dataplane-v2`), Network Policy enforcement is built-in and this step is not required (and may fail).
 
 **Apply Default Deny Policy:**
 Isolate namespaces by denying all ingress and egress traffic by default.
--- a/agents/cluster/skills/gke-workload-security/SKILL.md
+++ b/agents/cluster/skills/gke-workload-security/SKILL.md
@@ -82,7 +82,7 @@ is already active or provided natively by Dataplane V2:
 
 ```bash
 gcloud container clusters describe <cluster-name> \
-    --region <region> \
+    --location <location> \
     --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'
 ```
 
@@ -107,7 +107,7 @@ gcloud container clusters update <cluster-name> \
     --update-addons=NetworkPolicy=ENABLED \
     --region <region>
 
-# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)
+# Step 2: Enable NetworkPolicy enforcement on the nodes (node pools may be recreated; this can take several minutes)
 gcloud container clusters update <cluster-name> \
     --enable-network-policy \
     --region <region>
--- a/agents/cluster/skills/gke-workload-security/SKILL.md
+++ b/agents/cluster/skills/gke-workload-security/SKILL.md
@@ -77,9 +77,21 @@ traffic is allowed.
 
 **Enable Network Policy Enforcement:**
 
+Enabling network policy enforcement on clusters without Dataplane V2 requires
+two sequential commands in this order: first enable the Calico addon on the
+control plane, then enable network policy enforcement on the nodes. GKE rejects
+`--enable-network-policy` with HTTP 400 until the addon is enabled, and `gcloud`
+rejects both flags in a single invocation.
+
 ```bash
+# Step 1: Enable the NetworkPolicy addon on the control plane
 gcloud container clusters update <cluster-name> \
     --update-addons=NetworkPolicy=ENABLED \
+    --region <region>
+
+# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)
+gcloud container clusters update <cluster-name> \
+    --enable-network-policy \
     --region <region>
 ```
 
--- a/agents/platform/skills/gke-workload-security/SKILL.md
+++ b/agents/platform/skills/gke-workload-security/SKILL.md
@@ -93,7 +93,25 @@ access Google Cloud APIs.
 Control traffic flow between Pods using Network Policies. By default, all
 traffic is allowed.
 
-**Enable Network Policy Enforcement:**
+**Check Network Policy Enforcement & Dataplane:**
+
+Before modifying cluster networking, inspect whether NetworkPolicy enforcement
+is already active or provided natively by Dataplane V2:
+
+```bash
+gcloud container clusters describe <cluster-name> \
+    --region <region> \
+    --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'
+```
+
+- If `datapathProvider` is `ADVANCED_DATAPATH` (Dataplane V2), NetworkPolicy
+  enforcement is built-in natively via eBPF/Cilium from cluster creation. Calico
+  addons cannot be enabled and are not needed.
+- If `networkPolicy.enabled` is `True`, Calico enforcement is already enabled on nodes.
+- If neither is active, enable Calico network policy enforcement using the two-step
+  sequence below.
+
+**Enable Network Policy Enforcement (non-DPv2 clusters):**
 
 Enabling network policy enforcement on clusters without Dataplane V2 requires
 two sequential commands in this order: first enable the Calico addon on the
@@ -112,9 +130,6 @@ gcloud container clusters update <cluster-name> \
     --enable-network-policy \
     --region <region>
 ```
-
-> [!NOTE] If your cluster uses Dataplane V2 (`--enable-dataplane-v2`), Network
-> Policy enforcement is built-in and this step is not required (and may fail).
 
 **Apply Default Deny Policy:** Isolate namespaces by denying all ingress and
 egress traffic by default.
--- a/agents/platform/skills/gke-workload-security/SKILL.md
+++ b/agents/platform/skills/gke-workload-security/SKILL.md
@@ -95,9 +95,21 @@ traffic is allowed.
 
 **Enable Network Policy Enforcement:**
 
+Enabling network policy enforcement on clusters without Dataplane V2 requires
+two sequential commands in this order: first enable the Calico addon on the
+control plane, then enable network policy enforcement on the nodes. GKE rejects
+`--enable-network-policy` with HTTP 400 until the addon is enabled, and `gcloud`
+rejects both flags in a single invocation.
+
 ```bash
+# Step 1: Enable the NetworkPolicy addon on the control plane
 gcloud container clusters update <cluster-name> \
     --update-addons=NetworkPolicy=ENABLED \
+    --region <region>
+
+# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)
+gcloud container clusters update <cluster-name> \
+    --enable-network-policy \
     --region <region>
 ```
 
--- a/agents/platform/skills/gke-workload-security/SKILL.md
+++ b/agents/platform/skills/gke-workload-security/SKILL.md
@@ -100,7 +100,7 @@ is already active or provided natively by Dataplane V2:
 
 ```bash
 gcloud container clusters describe <cluster-name> \
-    --region <region> \
+    --location <location> \
     --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'
 ```
 
@@ -125,7 +125,7 @@ gcloud container clusters update <cluster-name> \
     --update-addons=NetworkPolicy=ENABLED \
     --region <region>
 
-# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)
+# Step 2: Enable NetworkPolicy enforcement on the nodes (node pools may be recreated; this can take several minutes)
 gcloud container clusters update <cluster-name> \
     --enable-network-policy \
     --region <region>
--- a/scripts/sync-upstream-skills.py
+++ b/scripts/sync-upstream-skills.py
@@ -34,9 +34,30 @@ GKE_WORKLOAD_SECURITY_OLD_NETPOL_SNIPPET = """**Enable Network Policy Enforcemen
 gcloud container clusters update <cluster-name> \\
     --update-addons=NetworkPolicy=ENABLED \\
     --region <region>
-```"""
+```
 
-GKE_WORKLOAD_SECURITY_NEW_NETPOL_SNIPPET = """**Enable Network Policy Enforcement:**
+> [!NOTE] If your cluster uses Dataplane V2 (`--enable-dataplane-v2`), Network
+> Policy enforcement is built-in and this step is not required (and may fail)."""
+
+GKE_WORKLOAD_SECURITY_NEW_NETPOL_SNIPPET = """**Check Network Policy Enforcement & Dataplane:**
+
+Before modifying cluster networking, inspect whether NetworkPolicy enforcement
+is already active or provided natively by Dataplane V2:
+
+```bash
+gcloud container clusters describe <cluster-name> \\
+    --region <region> \\
+    --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'
+```
+
+- If `datapathProvider` is `ADVANCED_DATAPATH` (Dataplane V2), NetworkPolicy
+  enforcement is built-in natively via eBPF/Cilium from cluster creation. Calico
+  addons cannot be enabled and are not needed.
+- If `networkPolicy.enabled` is `True`, Calico enforcement is already enabled on nodes.
+- If neither is active, enable Calico network policy enforcement using the two-step
+  sequence below.
+
+**Enable Network Policy Enforcement (non-DPv2 clusters):**
 
 Enabling network policy enforcement on clusters without Dataplane V2 requires
 two sequential commands in this order: first enable the Calico addon on the
--- a/scripts/sync-upstream-skills.py
+++ b/scripts/sync-upstream-skills.py
@@ -46,7 +46,7 @@ is already active or provided natively by Dataplane V2:
 
 ```bash
 gcloud container clusters describe <cluster-name> \\
-    --region <region> \\
+    --location <location> \\
     --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'
 ```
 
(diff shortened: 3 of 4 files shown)


agents/platform/skills/gke-workload-security/SKILL.md
```diff
@@ -95,10 +95,22 @@ traffic is allowed.
...
     --update-addons=NetworkPolicy=ENABLED \
     --region <region>
+
+# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)
+gcloud container clusters update <cluster-name> \
+    --enable-network-policy \
+    --region <region>
 ```
 
 > [!NOTE] If your cluster uses Dataplane V2 (`--enable-dataplane-v2`), Network
```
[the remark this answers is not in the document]
shalinibhatiapl: Good catch. Updated both `agents/platform/skills/gke-workload-security/SKILL.md` and `agents/cluster/skills/gke-workload-security/SKILL.md` in commit daf88ab7 to place the proactive probe command (`gcloud container clusters describe <cluster-name> --region <region> --format='value(networkConfig.datapathProvider,networkPolicy.enabled)'`) and explicit decision ladder before the update commands, removing the obsolete trailing note. Also updated `scripts/sync-upstream-skills.py` and unit test assertions in `scripts/test_sync_upstream_skills.py`.

agents/cluster/skills/gke-workload-security/SKILL.md
```diff
@@ -77,10 +77,22 @@ traffic is allowed.
```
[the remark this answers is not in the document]
shalinibhatiapl: Updated the PR body in response to this feedback:
1. Stated explicitly under **Self-Review** that review passes and subsequent revisions were conducted in the active pair-programming session.
2. Delineated the **Live validation** scope between the live GKE cluster lifecycle testing on GCP and unit testing of `scripts/sync-upstream-skills.py`. Also verified `GKE_WORKLOAD_SECURITY_OLD_NETPOL_SNIPPET` and `apply_substitutions()` directly against the live upstream file at `https://github.com/google/skills/blob/main/skills/cloud/gke-workload-security/SKILL.md`.
3. Corrected 'deterministic regex replacements' to 'deterministic string replacements' in **What Changed**.

agents/platform/skills/gke-workload-security/SKILL.md
```diff
@@ -93,16 +93,43 @@ access Google Cloud APIs.
...
+```bash
+# Step 1: Enable the NetworkPolicy addon on the control plane
 gcloud container clusters update <cluster-name> \
     --update-addons=NetworkPolicy=ENABLED \
     --region <region>
-```
 
-> [!NOTE] If your cluster uses Dataplane V2 (`--enable-dataplane-v2`), Network
-> Policy enforcement is built-in and this step is not required (and may fail).
+# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)
```
jayantid: 🟡 **Not merge-blocking.** Step 2 recreates every node pool in the cluster, and this comment tells the agent it only restarts networking addons.

`SKILL.md:128` labels the second command "(restarts cluster networking addons)". The GKE network-policy how-to says enabling enforcement on an existing cluster "re-creates your cluster's node pools" and "can cause disruption to your running workloads". Your own live validation in the PR body observed the node pool being recreated.

The agent reads this skill before touching a customer's cluster. Nothing in it signals a rolling recreation of every node. An agent following it on a production Standard cluster runs Step 2 without warning the operator, without a maintenance window, and without checking PodDisruptionBudgets. The drift SOP carries the same phrase at `fleet_consistency_drift_sop.md:131`, so this PR copies an existing understatement into two more files plus the sync constant.

Fix the comment in all three places (cluster `SKILL.md:110`, platform `SKILL.md:128`, `sync-upstream-skills.py:74`), which keeps the test constant aligned:

```bash
# Step 2: Enable NetworkPolicy enforcement on the nodes.
# This recreates every node pool (rolling) and disrupts running workloads;
# schedule it and confirm PodDisruptionBudgets first.
```

mplakhtiy: **MEDIUM — this understates a node-pool recreation as an addon restart**

> `# Step 2: Enable NetworkPolicy enforcement on the nodes (restarts cluster networking addons)`

`--enable-network-policy` recreates node pools. Two independent sources say so: this repository's own installer prints *"node pools may be recreated; this can take several minutes"* before issuing the identical command (`install.sh:1742`), and this PR's own Live validation reports *"Node pool recreated and cluster state updated"* after step 4.

**Failure scenario.** This is agent-executable text for a Platform Agent that acts on real clusters. An agent told the command "restarts cluster networking addons" has no reason to warn the user, request confirmation, or ask for a maintenance window before rolling every node in the cluster. The pre-flight probe this PR adds exists precisely so the agent stops before doing something disruptive — and then the text mislabels the disruptive step.

Cheap fix: reuse `install.sh:1742`'s wording. Three places — this file, `agents/cluster/skills/gke-workload-security/SKILL.md:110`, and the constant at `scripts/sync-upstream-skills.py:74`.

Note `fleet_consistency_drift_sop.md:131` carries the same understatement, so there is a reasonable argument for consistency — but that file is not in this diff, and this PR is the change putting the phrase into three new places.

agents/platform/skills/gke-workload-security/SKILL.md
```diff
@@ -93,16 +93,43 @@ access Google Cloud APIs.
...
 
-**Enable Network Policy Enforcement:**
+**Check Network Policy Enforcement & Dataplane:**
+
+Before modifying cluster networking, inspect whether NetworkPolicy enforcement
+is already active or provided natively by Dataplane V2:
 
 ```bash
+gcloud container clusters describe <cluster-name> \
+    --region <region> \
```
mplakhtiy: **MEDIUM — the new probe uses `--region`, which fails on zonal clusters; the repo's convention is `--location`**

`gcloud container clusters describe <name> --region <region>` returns NOT_FOUND for a zonal cluster. Everywhere else in the repository that performs this same read or update uses `--location`, which accepts a zone or a region: `install.sh:1717` and `:1740`, `terraform/modules/gke-cluster/main.tf:275`, `fleet_consistency_drift_sop.md:131` and `:235`.

**Failure scenario.** The probe is the new decision gate. On a zonal cluster it errors instead of printing two values, and the ladder underneath it has no branch for "the command failed" — an agent gets an error where it expected a state reading.

The evidence is in the PR body: Live validation ran `gcloud container clusters describe test-netpol-val --zone us-east4-a …`. The command as documented was never executed; a substituted one was.

Mitigating context: `--region` is the pre-existing convention throughout this skill file (the update commands and `scripts/audit_cluster.sh:18` all use it), so this is inherited rather than invented. Minimal fix is `--region <region>` → `--location <location>` on the three new command blocks, in all three places (here, `agents/cluster/skills/gke-workload-security/SKILL.md:85`, and the constant in `sync-upstream-skills.py`). Or reply arguing the file's convention should change separately, and file that.

mplakhtiy (commented):
Adversarial pass over the full diff at `daf88ab7`, run in a subagent that did not write the change.

**Nothing here blocks merge.** The substance is right: the two-step sequence, the ordering rationale and the DPv2 branch all match the repository's own authorities (`install.sh:1724-1745`, `terraform/modules/gke-cluster/main.tf:275`, `agents/platform/governance/fleet_consistency_drift_sop.md:131`).

I also retired the open question about the substitution mechanism with evidence rather than leaving it as a doubt: fetched `google/skills@main:skills/cloud/gke-workload-security/SKILL.md` and `GKE_WORKLOAD_SECURITY_OLD_NETPOL_SNIPPET` occurs there **exactly once, byte for byte**, at lines 96-105. Applying `apply_substitutions()` to the live file reproduces the committed platform `SKILL.md` (bar two Resources URLs, which are pre-existing sync drift). And `test_repo_workload_security_skills_have_enforcement_command` makes a future upstream reword go red rather than silently drop the fix.

Two MEDIUMs inline, one-line edits in three places each.

agents/platform/skills/gke-workload-security/SKILL.md
shalinibhatiapl: Adopted `install.sh:1742`'s wording in commit 44930862 across all three places (`agents/platform/skills/gke-workload-security/SKILL.md`, `agents/cluster/skills/gke-workload-security/SKILL.md`, and `scripts/sync-upstream-skills.py`), clarifying that node pools may be recreated and the operation can take several minutes. Also added test assertions in `scripts/test_sync_upstream_skills.py`.

agents/platform/skills/gke-workload-security/SKILL.md
shalinibhatiapl: Updated the probe command to use `--location <location>` in commit 44930862 across `agents/platform/skills/gke-workload-security/SKILL.md`, `agents/cluster/skills/gke-workload-security/SKILL.md`, and `scripts/sync-upstream-skills.py` so the check succeeds on both zonal and regional clusters. Added test assertions in `scripts/test_sync_upstream_skills.py`.

mplakhtiy (approved):

PR #1276, opened 2026-09-08
fix(installer): streamline release installation and eliminate image-tag prompts

### LINKED ISSUE #1291: docs(install): INSTALL.md never says how to obtain the sources, so a self-directed clone mismatches --image-tag
## Summary

`INSTALL.md` never tells the user how to obtain the repository — it contains **zero** occurrences of
`git clone` or `git checkout`. Method 1 assumes the sources are already on disk. A user who does the
natural thing and clones `main`, then passes `--image-tag=<release>`, is mismatched by default,
because `main` moves roughly ten commits a day.

## What the user sees

```
>>> 2. Setting up Workspace Repository <<<
  ✓ Using repository directory: /path/to/kube-agents
  ⚠ Source/image version mismatch: checkout is <sha-a>,
    requested ref resolves to <sha-b>.
  ⚠ Continuing with unverified install sources: the cluster will get this checkout's
    configuration plus the image built from 0.4.0.
```

## The behaviour itself is correct

`install.sh:1027-1031` sets lenient mode only under `--dry-run` or `--allow-unverified-source`; a real
run without the flag takes the `print_error` branch at `install.sh:1077` and stops. So
`INSTALL.md:76-79`'s claim that provisioning "refuses to start from a dirty or mismatched checkout" is
accurate. **This is a documentation gap, not a defect in the check.**

## The friction

1. **No documented way to obtain the sources.** The only documented acquisition path is Method 0's
   pinned one-liner (`INSTALL.md:36-48`). Method 1 (`INSTALL.md:146`, "The Install Engine — Terraform +
   Helm") starts at `cd terraform/examples/full-install`. Nothing bridges the two, so a user who clones
   is off the documented path with nothing prompting them to check out the release tag.
2. **The dry-run warning describes an impossible consequence.** `install.sh:1096` says "the cluster
   will get this checkout's configuration plus the image built from `<tag>`" — which cannot happen in a
   dry run, because a dry run provisions nothing.
3. **The dry run is least strict exactly where it is used to build confidence.** A user learns the real
   run would have refused only *after* the dry run appears to pass.

## Two stale statements found while checking

Both are wrong on a release build, where `default_image_tag` (`install.sh:665-671`) returns
`BAKED_RELEASE_VERSION` (`install.sh:80`) before consulting the checkout:

- `INSTALL.md:76` — "The image/source ref defaults to the checkout's `HEAD`".
- `install.sh:446` (help text) — `--image-tag` is "required via curl | bash".

## Suggested fix

- Add a source-acquisition step to Method 1 (`git clone … && git checkout <RELEASE_VERSION>`), or point
  Method 1 at Method 0's pinned one-liner.
- Reword `install.sh:1096` to say what is true in a dry run: the checkout does not match the requested
  ref, a real install would refuse, and the dry run is continuing anyway.
- Correct the two stale statements above.

## Workaround for users today

Use the pinned one-liner, which is correct by construction — the script is fetched from the tag, and
when it needs the rest of the sources it clones and checks out that same tag
(`install.sh:1119-1134`, fetching `+refs/tags/<tag>:refs/tags/<tag>` then `checkout --detach FETCH_HEAD`):

```bash
curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/0.4.0/install.sh | bash -s -- …
```

Run it from a directory that is not itself a kube-agents checkout: piped to bash, `BASH_SOURCE[0]` is
unset, so `install.sh:1112` resolves the script directory to the **current directory** and will use a
checkout it finds there instead of cloning a pinned copy.

With a local clone, add `git checkout <RELEASE_VERSION>` before running.

## Related

- #1287 covers the adjacent case where `~/kube-agents` already exists and is reused at the wrong
  revision. This issue is about the documentation gap that leads users into a mismatched checkout in
  the first place; #1287's fix would also resolve the reuse half.

## Verification

Not reproduced in this round's four installs — all used a tag checkout and printed
`✓ Verified install sources and image ref resolve to commit …`. The quoted output is from an earlier
session against a `main` checkout. Code paths, doc text and the absence of any clone instruction were
verified directly at `0.4.0`.

Worth noting when comparing SHAs by hand: `0.4.0` is an **annotated** tag, so `git rev-parse 0.4.0`
returns the tag object and `git rev-parse '0.4.0^{}'` returns the commit.

eLeontev:
## What Changed

- `install.sh`:
  - Centralized image tag resolution into `resolve_effective_image_tag()` to eliminate logic duplication between initial install and configuration re-apply.
  - Automatically adopts resolved default image tags (official release, release bundle, or local HEAD) without prompting interactively; only prompts when no default can be resolved.
  - Refactored `resolve_effective_image_tag()` to assign via an output variable (`out_tag_var`), eliminating subshell evaluation at call sites and preventing false `ERR` trap failures in non-interactive mode.
  - Aligned repo directory discovery in `resolve_effective_image_tag()` with `_resolve_repo_dir_for_state()`, keeping the three-way resolution (script dir -> cwd -> `$HOME/kube-agents`) consistent with `acquire_source_repo()`.
  - Hardened headless execution: cleanly errors with `--image-tag is required` when running without a controlling TTY and without a resolvable default, with diagnostics printing to stdout like all other `print_*` helpers.
  - Added interactive input validation loop with `validate_immutable_ref` to re-prompt rather than failing abruptly on invalid user input.
  - Updated CLI `--help` banner description for `--image-tag` to reflect default inference for releases while clarifying requirement for unbaked scripts.
  - Simplified and tested `PARAM_IMAGE_TAG` binding in day-2 menu mode option 6 (re-apply configuration).
  - Fixed `verify_local_source_ref()` under `--dry-run`: states that dry run continues with unverified sources rather than claiming the cluster will get the checkout's configuration (Fixes #1291).
- `tests/test_install_script.py`:
  - Added unit tests for `resolve_effective_image_tag`: baked release adoption, explicit argument preservation, non-interactive missing-tag error rejection, headless execution error rejection, external `cwd` repository discovery, and interactive prompt/retry loop.
  - Added regression test `test_run_menu_system_image_tag_binding` for Day-2 menu option 6.
  - Added regression test `test_verify_local_source_ref_dry_run_warning_does_not_claim_cluster_mutation`.
  - Fixed test lockfile isolation in `test_piped_stdin_executes_main` via `KUBE_AGENTS_LOCK_FILE`.
- `README.md`, `INSTALL.md`, `docs/site/src/content/docs/index.mdx`, `docs/site/src/content/docs/install/quickstart-gke.mdx`:
  - Established downloading the official self-contained release archive (`kube-agents-<RELEASE_VERSION>.tar.gz`) as the primary acquisition method, retaining `git clone --branch <RELEASE_VERSION>` as an explicit secondary alternative when a Git checkout is specifically required.
  - Audited all 32 occurrences of `./install.sh` across the documentation, converting ad-hoc development commands to official release one-liners.
  - Added Method 1 Step 1 ("Obtain the Release Sources") in `INSTALL.md` and renumbered steps 1–5, updating cross-references in `terraform/examples/full-install/README.md`.
  - Documented image tag lifecycle rules (CI commit SHAs and release stamping) in `terraform/examples/full-install/README.md`.
  - Fixed Astro Starlight `<Steps>` component AST parsing error by nesting `<Aside type="caution">` under Step 2.
- `.agents/skills/install-kube-agents/SKILL.md`:
  - Updated AI agent installation instructions to prioritize official release archive unpacking while retaining pinned release git clone as an alternative.

## Context

- Issue: b/558543170
- Fixes #1291

## Testing

- `python3 -m unittest tests/test_install_script.py`: 118 passed.
- `make docs-check`: Passed cleanly (links, terminology, map coverage, context budget).
- `npx prettier --check`: Passed on all modified markdown/MDX files.
- `npm --prefix docs/site run build`: Passed (44 static pages built successfully).

### Live validation

- Verified docs site rendering: captured visual evidence of the Astro Starlight `<Steps>` fix on the GKE quickstart page using `scripts/pr_evidence_screenshot.sh` and pushed to the `pr-evidence` branch:
  `Quickstart GKE Steps rendered`
- Verified `install.sh` automated tag resolution and error paths programmatically in dry-run with baked release version, local checkout, headless subshell, and external working directory.
- Verified release archive download and unpack cleanly extracts all baked assets and passes source verification.

## Self-Review

- **Passes ran:** `review-adversarial` and `review-docs-drift` ran against the branch diff. Neither pass was skipped.
- **Execution context:** Executed in separate, independent subagents spawned in fresh contexts with no prior memory of branch authoring (`fbc53a1c-bbf2-406f-8485-da31bccaaea8` for `review-adversarial` and `72c8f384-53cb-43d7-a479-29ab8f8667ce` for `review-docs-drift`). Final preflight passes returned 0 findings.
- **Findings and dispositions:**
  1. **[BLOCKING / CONFIRMED] Broken Starlight `<Steps>` component in quickstart guide:** `<Aside>` between list items broke `<Steps>` required single `<ol>` child in Astro AST. Disposition: Fixed by indenting `<Aside>` by 3 spaces under Step 2.
  2. **[BLOCKING / CONFIRMED] Quickstart interactive walkthrough listed "image tag" in prompted values:** Disposition: Removed "image tag" from the prompt list in `docs/site/src/content/docs/install/quickstart-gke.mdx`.
  3. **[HIGH / CONFIRMED] Headless execution without `--non-interactive` corrupts `image_tag` via `prompt_read` stdout capture:** In headless environments without TTY and without default tag, `prompt_read` auto-selected into stdout, corrupting `image_tag`. Disposition: Added `|| ! has_controlling_tty` to error cleanly with `--image-tag is required` and return 1.
  4. **[HIGH / CONFIRMED] Day-2 menu system ignored `--image-tag` CLI argument:** Option 6 re-read local checkout default instead of respecting `PARAM_IMAGE_TAG`. Disposition: Fixed by binding `PARAM_IMAGE_TAG:-${IMAGE_TAG:-}` and extracting `resolve_effective_image_tag`.
  5. **[MEDIUM / CONFIRMED] Hardcoded `.` repository path in `main` breaks tag resolution when invoked outside repository root:** Disposition: Updated `resolve_effective_image_tag` to resolve `script_dir` relative to `BASH_SOURCE[0]` if `repo_dir` is empty or `.`.
  6. **[MEDIUM / CONFIRMED] Interactive tag prompt lacked re-prompt loop and input validation:** Disposition: Wrapped interactive prompt in a `while true` loop with `validate_immutable_ref`.
  7. **[LOW / CONFIRMED] Missing negative and external cwd test coverage for `resolve_effective_image_tag`:** Disposition: Added `test_resolve_effective_image_tag_fails_when_non_interactive_and_no_default`, `test_resolve_effective_image_tag_fails_headless_without_tty_and_no_default`, and `test_resolve_effective_image_tag_resolves_from_external_cwd`.
  8. **[ADVISORY / CONFIRMED] `install.sh --help` text for `--image-tag` drifted from release inference behavior:** Disposition: Updated line 458 to `(default: inferred from baked release, release bundle, or local HEAD)`.
  9. **[ADVISORY / CONFIRMED] Non-interactive curl flags diverged between docs surfaces:** Disposition: Harmonized non-interactive flags across `README.md`, `INSTALL.md`, and `quickstart-gke.mdx`.
  10. **[ADVISORY / CONFIRMED] Non-SemVer `v0.1.0` in `INSTALL.md` image mirroring example:** Disposition: Corrected to `0.1.0`.
  11. **[ADVISORY / CONFIRMED] Redundant "generic `curl | bash`" requirement reference in skill table:** Disposition: Updated table row in `.agents/skills/install-kube-agents/SKILL.md`.
  12. **[HIGH / CONFIRMED] External `cwd` test regex strictly asserted 40-char SHA:** On tagged checkouts where `HEAD` carries a SemVer release tag, `default_image_tag` prioritizes the release tag over the SHA. Disposition: Broadened assertion regex to accept both 40-character commit SHAs and SemVer tags.
  13. **[HIGH / CONFIRMED] Negative resolution unit tests lacked `BAKED_RELEASE_VERSION` isolation:** On release-stamped scripts, `default_image_tag` would return the baked version rather than testing unversioned error paths. Disposition: Explicitly isolated `BAKED_RELEASE_VERSION=""` in negative and headless resolution tests.
  14. **[LOW / CONFIRMED] Usage comment header in `install.sh` cited unversioned `main` URL:** Disposition: Updated header to use `<RELEASE_VERSION>` and `platform-agent-host`.
  15. **[MEDIUM / CONFIRMED] Redundant directory resolution when explicit `--image-tag` provided:** `resolve_effective_image_tag` executed directory fallback subshells before checking if `$requested_tag` was passed. Disposition: Fixed by returning `$requested_tag` immediately at the top of the function.
  16. **[MEDIUM / CONFIRMED] Missing test coverage for interactive prompt and validation retry loop:** Disposition: Added `test_resolve_effective_image_tag_prompts_and_retries_on_invalid_ref` in `tests/test_install_script.py` asserting retry loop on invalid input and exit on valid SemVer ref.
  17. **[ADVISORY / PLAUSIBLE] Hardcoded `"scripts/installer/installer_common.sh"` path in `resolve_effective_image_tag`:** Disposition: Retained as-is to align with existing conventions throughout `install.sh` prior to helper sourcing.
  18. **[MEDIUM / CONFIRMED] Review Comment 1 & 5 (`install.sh:2218, 752`): Subshell ERR trap on missing `--image-tag`:** Command substitution `PARAM_IMAGE_TAG="$(resolve_effective_image_tag ...)"` triggered the ERR trap and printed internal error report. Disposition: Refactored `resolve_effective_image_tag` to take an output variable name via `printf -v "$out_tag_var"` and restored diagnostic logging to stdout.
  19. **[LOW / CONFIRMED] Review Comment 2 (`install.sh:755`): Divergent repo dir resolution:** Copied first branch of `_resolve_repo_dir_for_state` but fell back to `.` rather than `$HOME/kube-agents`. Disposition: Replaced with `_resolve_repo_dir_for_state` / `${_state_repo_dir:-}` and added regression tests.
  20. **[LOW / CONFIRMED] Review Comment 3 (`install.sh:2006`): Dead fallback and untested menu image tag binding:** `${IMAGE_TAG:-}` was dead because `PARAM_IMAGE_TAG` was already seeded. Disposition: Simplified to `local image_tag="${PARAM_IMAGE_TAG:-}"` and added `test_run_menu_system_image_tag_binding`.
  21. **[LOW / CONFIRMED] Review Comment 4 (`INSTALL.md:176`): Composition README lacked image_tag stamping rule:** Disposition: Added explicit documentation for CI commit SHAs and release automation stamping in `terraform/examples/full-install/README.md`.
  22. **[MEDIUM / CONFIRMED] Review Comment 6 (`install.sh:768`): Undocumented unstamped installer workflows:** Disposition: Clarified dual behavior in `INSTALL.md`, skill, and `--help` banner.
  23. **[MEDIUM / CONFIRMED] Issue #1291 (`install.sh:1157`): False cluster mutation claim under `--dry-run`:** Warning claimed cluster would get checkout configuration. Disposition: Changed warning under `--dry-run` to state preview is continuing, and added regression test.
  24. **[HIGH / CONFIRMED] Issue #1291 (`INSTALL.md`): Missing source acquisition step in Method 1:** Step 1 previously jumped directly to gcloud auth without acquiring sources. Disposition: Added Step 1 prioritizing release tarball download, renumbered steps 1–5, and updated cross-references.

## Risk & Rollout

Low risk. `install.sh` changes fall back cleanly to prompting if no default image tag is found. No runtime agent pod code or Kubernetes CRD/operator schemas modified.


.agents/skills/install-kube-agents/SKILL.md
```diff
@@ -34,37 +34,51 @@ Day-2 panel's Save & Apply) reconciles every change through one `terraform apply
```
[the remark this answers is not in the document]
eLeontev: Fixed. We verified the docs site build cleanly with `npm --prefix docs/site run build` (all 44 pages built) and captured the screenshot of the rendered `<Steps>` list via `scripts/pr_evidence_screenshot.sh`. The evidence screenshot is published to the `pr-evidence` branch and embedded in the PR body under **Testing → Live validation**.

INSTALL.md
```diff
@@ -169,8 +172,7 @@ and GitHub minter workloads).
 - The manual Chat/Slack registrations in
   [Step 4 of this method](#step-4-enable-google-chat--slack-integrations-manual-required-steps)
   apply however the engine is driven.
-- Until the first `X.Y.Z` release tag exists, keep the default `image_tag = "latest"`
-  (see the guide's image-tag note).
+- The Terraform composition defaults `image_tag` to `"latest"` in Git checkouts (or the released container image version in unpacked release bundles; see the composition README).
```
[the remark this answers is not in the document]
eLeontev: Fixed in commit `d7991e81`. The note in `INSTALL.md` was updated to accurately contrast `main` (which defaults to `"latest"`) against release tag checkouts and unpacked release bundles (where the release automation stamps the released container image version as default).

README.md
```diff
@@ -25,41 +25,41 @@ _An SRE asks for a fleet self-health check; the agent answers in the thread. An
 The fastest, zero-friction way to install `kube-agents` in **Google Cloud Shell** or your terminal:
 
 ```bash
-curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/0.4.0/install.sh | bash
 ```
 
-_To pin to a specific official release, substitute `<RELEASE_VERSION>` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases):_
+_Installs the latest official release (currently `0.4.0`). To use another release, substitute `0.4.0` with the desired version tag from [GitHub Releases](https://github.com/gke-labs/kube-agents/releases). You can also run the generic installer `curl -fsSL https://gke-labs.github.io/kube-agents/install.sh | bash`, which prompts for the target release tag._
```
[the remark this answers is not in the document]
eLeontev: Addressed. The copy-pastable snippets retain the generic `<RELEASE_VERSION>` placeholder (`https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh`) with explicit substitution instructions, and `0.4.0` is referenced purely as an example of a SemVer release tag. Furthermore, release-pinned installers have `BAKED_RELEASE_VERSION` stamped into `install.sh`, which `resolve_effective_image_tag` automatically adopts without requiring the user or automation to specify `--image-tag`.

install.sh
```diff
@@ -2181,16 +2188,16 @@ main() {
...
       fi
-      image_tag="$head_sha"
-      print_info "Defaulting image tag to $(default_image_tag_label): ${C_BOLD}${image_tag}${C_RESET}"
-    else
       prompt_read "Container image tag (validated release tag or full commit SHA)" \
-        image_tag "$head_sha" false "$(default_image_tag_label)"
+        image_tag "" false ""
     fi
   fi
   validate_immutable_ref "$image_tag"
```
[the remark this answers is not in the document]
eLeontev: Addressed in commits `e12af1be` and `64da64bc`, and reflected in the updated PR description:
1. The PR description (**Summary**, **What Changed**, **Risk & Rollout**, **Live validation**, and **Self-Review**) was fully updated to document all installer modifications, why they were introduced, and their dispositions.
2. Unit test coverage in `tests/test_install_script.py` was expanded to 10 tests covering the negative non-interactive error path, headless execution without TTY (verifying stderr error reporting and empty stdout), external directory (`cwd`) execution, and baked release adoption.
3. Interactive image tag prompts were hardened with a `while true` validation loop using `validate_immutable_ref`.
4. Menu option 6 now correctly honors `PARAM_IMAGE_TAG` via centralized `resolve_effective_image_tag`.
5. Retyped PR title to `fix(installer): streamline release installation and eliminate image-tag prompts`.

install.sh
```diff
@@ -728,6 +728,32 @@ default_image_tag_label() {
...
+  if [ "$PARAM_NON_INTERACTIVE" = "true" ]; then
+    print_error "--image-tag is required; use a validated release tag or full commit SHA." >&2
+    return 1
+  fi
+  local prompted_tag=""
+  prompt_read "Container image tag (validated release tag or full commit SHA)" \
+    prompted_tag "" false ""
+  echo "$prompted_tag"
+}
+
```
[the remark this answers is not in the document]
eLeontev: Addressed in commits `e12af1be` and `64da64bc`, and reflected in the updated PR description:
1. The PR description (**Summary**, **What Changed**, **Risk & Rollout**, **Live validation**, and **Self-Review**) was fully updated to document all installer modifications, why they were introduced, and their dispositions.
2. Unit test coverage in `tests/test_install_script.py` was expanded to 10 tests covering the negative non-interactive error path, headless execution without TTY (verifying stderr error reporting and empty stdout), external directory (`cwd`) execution, and baked release adoption.
3. Interactive image tag prompts were hardened with a `while true` validation loop using `validate_immutable_ref`.
4. Menu option 6 now correctly honors `PARAM_IMAGE_TAG` via centralized `resolve_effective_image_tag`.
5. Retyped PR title to `fix(installer): streamline release installation and eliminate image-tag prompts`.

install.sh
```diff
@@ -2177,22 +2214,8 @@ main() {
...
-      fi
-      image_tag="$head_sha"
-      print_info "Defaulting image tag to $(default_image_tag_label): ${C_BOLD}${image_tag}${C_RESET}"
-    else
-      prompt_read "Container image tag (validated release tag or full commit SHA)" \
-        image_tag "$head_sha" false "$(default_image_tag_label)"
-    fi
-  fi
+  local image_tag=""
+  image_tag="$(resolve_effective_image_tag "." "${PARAM_IMAGE_TAG:-}")" || exit 1
```
mplakhtiy: **🟠 Medium — this error path now prints an internal error banner and clobbers the machine-readable install report.**

`set -Eeuo pipefail` at the top of the file means `-E`, so the `ERR` trap is inherited by the command substitution's subshell. `resolve_effective_image_tag`'s deliberate `return 1` (line 758) therefore fires `on_error` **inside** the subshell, before the outer `|| exit 1` ever runs. `on_error` prints its banner and calls `write_json_report "FAILED"`, which does `cat << EOF > /tmp/kube-agents-install-report.json` — an unconditional overwrite. On `main` this path was `print_error …; exit 1` in `main()` itself, which triggers no `ERR` trap.

Observed, from a non-checkout cwd with an unstamped `install.sh`:

```
  ✗ --image-tag is required; use a validated release tag or full commit SHA.

✗ Error encountered at line 2218 (exit code 1): return 1
```

and `/tmp/kube-agents-install-report.json` created with `"status": "FAILED"`, `"project_id": ""`, and every other field blank. Against `upstream/main:install.sh`, byte-for-byte the same invocation prints only the first line and creates no report file.

**Failure scenario:** an agent or CI job runs the generic installer non-interactively without `--image-tag`. It gets a clean validation error with what looks like a stack trace stapled to it, and the previous run's install report — the artifact `INSTALL.md:102` advertises as the machine-readable output for AI agents — is silently replaced by a FAILED record of a run that never started. A caller reading that file to decide whether the last install succeeded now reads a failure that was actually a missing flag.

Reproduced on both revisions. Note the Self-Review's finding 3 says this path was verified in a headless subshell — the exit code is right, so a check asserting only `rc != 0` would pass; the output and the report file are what changed.

**Root cause and the shape of the fix:** this follows from returning the tag on stdout. `acquire_source_repo` (`install.sh:1159-1161`) already solves the same problem the other way and says why in a comment — *"Stores the directory in the variable named by `$1` rather than echoing it: the progress lines below would otherwise be captured along with the path."* Giving `resolve_effective_image_tag` a `printf -v "$dest_var"` output parameter removes the subshell, which removes this finding, removes the need for the `>&2` on line 752, and matches the sibling function.

mplakhtiy (commented):
Reviewed the full diff at `f934735` in a clean context that did not write the change, against `upstream/main` (`488b3ee`). The core mechanism is sound — `resolve_effective_image_tag` is a real consolidation, its external-`cwd` fix aligns image-tag resolution with `acquire_source_repo`'s repo-dir resolution for the first time, and the new tests are genuine regression guards (mutating the `script_dir` block out turns the external-`cwd` test red). Every path that previously depended on the prompt now has a defined source, and nothing here adds a second expression of an install step.

Six findings are inline. One more could not be anchored to a changed line:

#### 🟠 Medium — the release-first rule is applied to install only; `uninstall.sh` and `upgrade.sh` still point at main

Anchors: `.agents/skills/uninstall-kube-agents/SKILL.md:15`, `.agents/skills/upgrade-kube-agents/SKILL.md:20`, `docs/site/src/content/docs/install/uninstall.md:72`, `uninstall.sh:11`, `upgrade.sh:9`.

This PR rewrites `install.sh`'s usage header from the `main` URL to `.../<RELEASE_VERSION>/install.sh` and pushes the release-pinned form through every install surface. The three sibling front doors keep `curl -fsSL https://gke-labs.github.io/kube-agents/uninstall.sh | bash` and the `upgrade.sh` equivalent — the same main-tracking, unstamped scripts.

The consequence is not cosmetic. At `uninstall.sh:337-340`, with no `BAKED_RELEASE_VERSION` and no `--source-ref`, teardown prints *"fetching the teardown engine from main, which may be newer than your installed release"* and clones `main`. So a user who follows this PR's advice to install 0.4.0 and then follows the unmodified uninstall skill drives a 0.4.0 install with `main`'s teardown engine — the exact class of mismatch the PR body argues against, one command later.

**Failure scenario:** release-pinned install, main-pinned teardown, on a lifecycle whose documentation now asserts the two must agree.

Whether the fix belongs here or in a follow-up is a scope call — but per AGENTS.md "out of scope" alone is not a disposition, and a filed issue would be.

install.sh
```diff
@@ -728,6 +728,46 @@ default_image_tag_label() {
...
+  fi
+  local prompted_tag=""
+  while true; do
+    prompt_read "Container image tag (validated release tag or full commit SHA)" \
+      prompted_tag "" false ""
+    if validate_immutable_ref "$prompted_tag" >&2; then
+      break
+    fi
+  done
+  echo "$prompted_tag"
```
mplakhtiy: **🟠 Medium — this prompt loop is now reachable only through an entry point no document mentions, which CI still publishes.**

`.github/workflows/docs-deploy.yml:18-19,53` still triggers on `install.sh` and still runs `cp ../../install.sh ../../uninstall.sh ../../upgrade.sh public/`, so `https://gke-labs.github.io/kube-agents/install.sh` continues to serve **main's** unstamped installer. This PR deletes every reference to that URL — from `README.md`, from `INSTALL.md` (including the sentence that explained the prompt: *"the interactive prompt asks you to enter the target SemVer release tag … and `--non-interactive` requires `--image-tag`"*), from `quickstart-gke.mdx`, and from the skill. It does not deprecate the artifact, redirect it, or stop publishing it.

That is the only path that reaches the code this PR invests most in. With no `BAKED_RELEASE_VERSION`, no checkout under the script, and no checkout in the cwd, `resolve_effective_image_tag` falls through to this `while true` / `validate_immutable_ref` retry loop — which the PR adds, hardens with `has_controlling_tty`, and covers with two new tests, while removing the last prose that tells anyone the prompt exists. Both the `--help` banner (`install.sh:458`) and the skill's flag table dropped the "required via `curl | bash`" half of the default, which is still true for that installer.

**Failure scenario:** a user with the Pages URL bookmarked, or an agent that scraped it from a cached page, runs it non-interactively and hits `--image-tag is required` — plus the banner from the finding at line 2218 — with no document in the repository explaining that the flag is required there or why. Meanwhile a reader of the new docs has no way to learn the generic installer exists at all.

Either the publish step goes (and the prompt loop with it), or one document keeps a line about the generic installer. Which one is your call; the current state is neither.

install.sh
```diff
@@ -728,6 +728,46 @@ default_image_tag_label() {
...
+  fi
+  if [ -z "$repo_dir" ] || [ "$repo_dir" = "." ]; then
+    local script_dir
+    script_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-.}")" 2>/dev/null && pwd || echo "")"
+    if [ -n "$script_dir" ] && [ -f "${script_dir}/scripts/installer/installer_common.sh" ]; then
+      repo_dir="$script_dir"
+    else
+      repo_dir="."
+    fi
+  fi
```
mplakhtiy, line 758: **🔵 Low — this re-implements a repo-dir resolution the file already has, and diverges from it.**

`_resolve_repo_dir_for_state()` (`install.sh:138-149`) already performs the three-way resolution — script dir, then cwd, then `$HOME/kube-agents` — keyed on the same `scripts/installer/installer_common.sh` marker file, and its comment states the contract: *"resolve the same repo_dir acquire_source_repo will pick, by the same test and in the same order … This has to stay in step with acquire_source_repo; the shared marker file is the coupling, and test_install_script.py pins the pair."* Its result is already in scope as the global `_state_repo_dir`.

This block copies the first branch and stops, falling back to `.` where the existing helper and `acquire_source_repo` both fall back to `$HOME/kube-agents`. So when neither the script dir nor the cwd is a checkout but `$HOME/kube-agents` is, the tag resolves against `.` (yielding nothing) while `acquire_source_repo` goes on to install from `$HOME/kube-agents`. The run then demands a tag for a checkout it is about to use and could have read the tag from.

**Failure scenario:** `curl | bash` of the unstamped installer from `~`, with an existing `~/kube-agents` checked out at 0.4.0 — the user is asked for a tag, or told `--image-tag is required`, for a checkout sitting right there.

Not a regression (`main` passed `.` too), but this PR touched exactly this line to fix branch 1 and left branches 2 and 3 out of step with a helper whose comment asks to be kept in step.

install.sh
```diff
@@ -1963,7 +2003,7 @@ run_menu_system() {
   local kms_keyring="${KMS_KEYRING:-}"
   local kms_key="${KMS_KEY:-}"
   local github_pem_path="${GITHUB_PEM_PATH:-}"
-  local image_tag="${IMAGE_TAG:-}"
+  local image_tag="${PARAM_IMAGE_TAG:-${IMAGE_TAG:-}}"
```
mplakhtiy: **🔵 Low — half of this is dead, and the half that works is untested.**

`PARAM_IMAGE_TAG` is already seeded from `IMAGE_TAG` at `install.sh:368`, after `install.env` is sourced at lines 192/265. The `${IMAGE_TAG:-}` inner fallback is therefore unreachable.

The part that does work — `--menu --image-tag=X` now reaching option 6 — is finding 4 of your own Self-Review and has no test: `tests/test_install_script.py` gains six tests, all on `resolve_effective_image_tag`, none on `run_menu_system`'s binding.

**Failure scenario:** a future edit to either line silently restores the bug the Self-Review says was fixed, and nothing goes red.

```suggestion
  local image_tag="${PARAM_IMAGE_TAG:-}"
```

install.sh
```diff
@@ -728,6 +728,46 @@ default_image_tag_label() {
...
+    if [ -n "$script_dir" ] && [ -f "${script_dir}/scripts/installer/installer_common.sh" ]; then
+      repo_dir="$script_dir"
+    else
+      repo_dir="."
+    fi
+  fi
+  local default_tag=""
+  default_tag="$(default_image_tag "$repo_dir")"
+  if [ -n "$default_tag" ]; then
+    print_info "Using container image tag ($(default_image_tag_label "$repo_dir")): ${C_BOLD}${default_tag}${C_RESET}" >&2
```
mplakhtiy: **🔵 Low — the installer's one tag diagnostic goes to stderr while every other line goes to stdout.**

Every other `print_*` in the script writes to stdout; this `>&2` exists only because the function's result travels on stdout. A user running `./install.sh > install.log` gets a log missing the line that says which image tag the run chose — the single most consequential inference the installer now makes on their behalf, given it no longer asks.

Fixing the line-2218 finding with an output parameter removes the need for this redirect, so the two are best resolved together rather than by dropping the `>&2` on its own (which would corrupt the captured value).

INSTALL.md
```diff
@@ -169,8 +173,7 @@ and GitHub minter workloads).
 - The manual Chat/Slack registrations in
   [Step 4 of this method](#step-4-enable-google-chat--slack-integrations-manual-required-steps)
   apply however the engine is driven.
-- Until the first `X.Y.Z` release tag exists, keep the default `image_tag = "latest"`
-  (see the guide's image-tag note).
+- The Terraform composition defaults `image_tag` to `"latest"` on `main` (or the stamped release version when on a release tag checkout or in unpacked release bundles; see the composition README).
```
mplakhtiy: **🔵 Low — this sends the reader to a page that does not carry the fact.**

The claim itself is correct: `terraform/examples/full-install/variables.tf:157-161` defaults to `latest`, and `scripts/release/common.sh:896-925` rewrites that default at release time. But `terraform/examples/full-install/README.md:295-302` ("The `image_tag` rule") documents only `default latest` and says nothing about stamping, so the pointer does not support the parenthetical it is attached to.

Either drop the pointer or add a sentence to the composition README, which per AGENTS.md's documentation table is the canonical home for that fact.

install.sh
eLeontev: Fixed in commit `64d73701`.

We refactored `resolve_effective_image_tag` to write its result to a named output variable via `printf -v "$out_tag_var" "%s" "$resolved_tag"` instead of returning the tag over stdout. This completely eliminates the subshell assignment (`PARAM_IMAGE_TAG="$(resolve_effective_image_tag ...)"`) at the call site. Now, when `resolve_effective_image_tag` aborts with `print_error` and `exit 1` in non-interactive mode on missing required tags, the main script exits cleanly with the expected error message rather than tripping the subshell `ERR` trap and emitting an internal error banner.

install.sh
eLeontev: Fixed in commit `64d73701`.

With `resolve_effective_image_tag` passing its result via an output variable rather than stdout, the `print_success` diagnostic message no longer risks corrupting captured output. We removed the `>&2` redirect so the diagnostic prints to standard output consistently with every other `print_*` helper in `install.sh`.

install.sh
eLeontev: Fixed in commit `b86b7e0f`.

Replaced the ad-hoc local checkout check with the shared `_resolve_repo_dir_for_state` helper (`${_state_repo_dir:-}`), ensuring that `resolve_effective_image_tag` follows the exact same 3-way resolution order (script dir -> cwd -> `$HOME/kube-agents`) as `acquire_source_repo`. We also added regression tests covering resolution against local and `$HOME/kube-agents` checkouts.

install.sh
eLeontev: Fixed in commit `2c1c6420`.

1. Removed the dead fallback expression, simplifying the binding to `local image_tag="${PARAM_IMAGE_TAG:-}"`.
2. Added unit test `test_run_menu_system_image_tag_binding` to `tests/test_install_script.py` verifying that option 6 in the interactive Day-2 control panel correctly binds the active image tag from `PARAM_IMAGE_TAG`.

INSTALL.md
eLeontev: Fixed in commit `1348c971`.

Updated `terraform/examples/full-install/README.md` under **"The `image_tag` rule"** to explicitly document the full lifecycle:
- Defaults to `"latest"` on `main`.
- In CI/CD pipelines, passed explicitly with the test build commit SHA.
- On official release checkouts and unpacked release bundles, automatically stamped with the released SemVer version by release automation.

install.sh
eLeontev: Fixed in commit `6cb4534d`.

Restored and clarified documentation regarding unstamped installers:
1. Updated `INSTALL.md` and `install-kube-agents/SKILL.md` to explicitly state that `--image-tag` is reserved for development and CI/CD pipelines running directly from `main` or unstamped scripts, while canonical release deployments (via `<RELEASE_VERSION>/install.sh` or official release archives) resolve image tags automatically.
2. Updated `install.sh --help` banner to explicitly document the dual behavior: `--image-tag` is automatically stamped on official releases, but required when driving unstamped scripts non-interactively.

AntonTyb (approved):

Issue #1278, opened 2026-09-08
Seeded fleet outage: seeded-a's e2-medium node is 100% CPU-requested by system pods after the weekend GKE node upgrade — payments-api Pending on all 30 pool projects, crashloop trio reds every PR


jayantid:
## Every presubmit run has been red since ~2026-09-08 03:00Z — one cause

`cluster-agent-crashloop-debug`, `-misleading-symptom`, and `-evidence-chain` (all three admitted) fail 3/3 on every run today, on unrelated PRs (#1246, #1238, #1150, #1226, #1275, #1267 …): `rca-names-the-oom: required phrases absent: ['OOMKilled']`. `rca-remediation-pr` (same fixture) is 0/3. The agent is not wrong — its report says exactly what is true:

> The `payments-api` deployment in `seeded-debug` is stuck in `Pending` due to unschedulability, not a crashloop. … `0/3 nodes are available`: one node has insufficient CPU, two have untolerated taints; the cluster-autoscaler logs `1 max node group size reached`.

## Root cause (verified on kube-agents-evals-2, read-only)

- seeded-a's only schedulable node (`default-pool`, **e2-medium, 940m allocatable**) is at **100% CPU requested by GKE system pods alone**: kube-proxy 100m, **two** kube-dns replicas at ~270m each, gke-metadata-server 100m, fluentbit ~105m, node-local-dns 30m, konnectivity, metrics-agent, pdcsi, netd … = 940m. The planted workloads (payments-api 10m, checkout-gateway, …) cannot fit; system-critical priority wins.
- That node was **recreated 2026-09-08T04:59Z by GKE auto-upgrade** (seeded-a is deliberately in the REGULAR channel with a daily maintenance window — an upgrades-domain fixture facet). `payments-api` has been Pending since; `FailedScheduling … 1 Insufficient cpu, 2 node(s) had untolerated taint(s)`; `NotTriggerScaleUp: 1 max node group size reached`.
- The node was at ~100% before the upgrade too: the fixture only ever ran because it was scheduled *before* the second kube-dns replica landed. Node recreation re-ran scheduling with system pods first. A latent time bomb; the maintenance window pulled the pin.
- **Blast radius: all 30 pool projects.** Every seeded-a is now on 1.35.7 with the recreated default pool (25 on gke.1150000, 5 on gke.1027000 — two upgrade waves, matching the Sunday-gradual → Monday-total onset in the run history). seeded-b (no channel, by design) is untouched on 1.34.

[...]

PR #1279, opened 2026-09-08
fix(gate): demote the crashloop trio pending the seeded-fleet repair (#1278)

jayantid:
## Summary
Removes the three crashloop cases from `BOOTSTRAP_ADMITTED` (10 → 7 admitted) so pull requests can merge again. **Every presubmit run since ~03:00Z today is red on exactly these three cases**, on unrelated PRs — root cause and evidence in #1278: the seeded fleet's `payments-api` fixture pod is Pending on all 30 pool projects after the weekend GKE node upgrade left seeded-a's e2-medium node 100% CPU-requested by system pods.

## Why This Change
The gate is correct (the pod really isn't crashlooping) and every PR is wrongly blocked. This is the roster protocol's exact case: an admitted case reds a PR its diff cannot explain → one-line demotion, referencing the issue with the re-admission condition. Re-admission is the reverse edit the moment the fleet is re-applied; these three have the suite's best records and come straight back.

## What Changed
- `hack/ci-eval-pr.sh`: three names removed from the roster default; a dated note above the export citing #1278.

## Testing
`bash -n` clean; full `scripts/` discovery suite green.

### Live validation
Not live-testable pre-merge; the next run after merge shows the trio as UNSTABLE (not FAILED) and the run green on a healthy PR. **Requesting `/override` + merge** — this PR's own smoke run reds on the very cases it demotes.

## Self-Review
Verified against the live evals-2 cluster state (pod Pending 9h, FailedScheduling insufficient CPU, node allocated 940m/940m) and the 30-project version sweep; verified the three names appear exactly once each in the roster and nowhere else in the export.

## Risk & Rollout
Reduces what can block; cannot break a run. Cluster-debugging domain temporarily has no blocking case — accepted for the hours until the fleet repair. Revert = re-add three names.


jayantid:
Closing — decision is to **fix forward** rather than demote: add a second node to seeded-a's default pool on every pool project (a `gcloud container clusters resize`, minutes per project, fixtures reschedule on their own), with #1280 amended to `node_count = 2` so tofu state matches. The three cases stay admitted and go green as each project's fleet is repaired. Reopen this as the fallback only if the resize can't be run today.

PR #1281, opened 2026-09-08
perf(eval): measure agent-turn replay cache across presubmit repetitions (do not merge)

haoxuw:
**Do not merge — measurement run.** This PR exists to run the full presubmit once with the harness's new opt-in agent-turn replay cache armed, to measure the gain through the real Prow pipeline. The second commit corrupts repetition semantics by design (reps 2–3 replay rep 1 from the pod-local disk cache) and will not merge; the first commit is the merge-candidate integration and defaults to off.

## Context

Follows the runtime/cost analysis in #948 (comment: token and wall-clock breakdown of build 2095524608026349568): the suite's cost is model turns, ~2/3 of which are repetitions. `cached-response` (PyPI) records agent turns to SQLite and replays exact matches; keying includes the turn's per-conversation ordinal so within-run status polls never collide (naive keying tripped the harness's silent-turn abandonment — caught by `test_delegated_work_is_awaited_before_the_result_is_returned`).

## Testing

- `bench/tests`: 647/647 with cache off and with `CACHED_RESPONSE_MODE=testing`.
- Live (own install, agent-kanban-smoke): record run all-misses and behaviorally identical to baseline (47.9s); replay run 100% hits, byte-identical output and trajectory, 3.8s agent latency, zero model calls.

### Live validation

This PR's presubmit run **is** the live validation and the measurement: rep 1 of each case runs live (recording), reps 2–3 replay. Numbers to be posted to #948 after the run.

## Self-Review

Experiment PR; the adversarial/docs-drift passes are owed by the merge-candidate first commit when it is re-proposed alone. Known-by-design defects of the second commit: replayed reps invalidate the collapse rung's variance measurement; 4 concurrent units share one SQLite cache file.

---

### DIFF
--- a/bench/kube_agents_bench/harness.py
+++ b/bench/kube_agents_bench/harness.py
@@ -43,6 +43,7 @@ Environment:
 from __future__ import annotations
 
 import atexit
+import contextvars
 import http.client
 import json
 import logging
@@ -60,6 +61,7 @@ import uuid
 from pathlib import Path
 from typing import Any
 
+from cached_response import cached_llm_response
 from devops_bench.agents import AgentHarness, AgentResult
 
 from kube_agents_bench import transcript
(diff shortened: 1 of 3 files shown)


haoxuw:
/hold
Measurement run only; see PR body.

haoxuw:
Scrapping this approach: caching belongs in the LiteLLM gateway, not the eval harness. The branch is deleted; a gateway-side experiment will follow separately.

PR #1285, opened 2026-09-08
perf(litellm): cached-response test-alias replay sidecar, armed for one measured presubmit run

haoxuw:
**Commits 1–3 are the merge candidate** (off-by-default sidecar, off-state render byte-identical to main; the bench board-row purge; the released-package pin). **Commit 4 arms it for one measured presubmit run and reverts rather than merges.**

## What this does, in plain words

Our eval suite asks the agent the same questions three times per case, and every repetition pays full price for model calls that are nearly identical to the run before. This PR puts a small proxy in front of the model gateway that remembers answers and replays them when it is *certain* the question is the same — and stays out of the way when it is not.

**Finding vs. approving.** The cache works in two strict steps. First it *finds* a candidate: it strips out the stuff that legitimately changes between runs (ids, timestamps) and looks for a stored request with the same skeleton. Then it *approves*: every byte outside the fields we explicitly declared as noise must match exactly. Finding is allowed to be clever; approving is not. A wrong "not found" costs one model call; a wrong "approved" corrupts an answer — so approval is never a judgment call.

**Real examples from captured traffic** — pairs from two live runs of the same task that the cache correctly treats as "the same question":

```
"report their current status: t_41dfe03f"     "created_at": 1788907973
"report their current status: t_217deef9"     "created_at": 1788908055

"pid": 248801  vs  "pid": 249064              "current_run_id": 32  vs  33
```

Task ids are handled by *aliasing* (each run's ids map to stable handles like `test_id_00000000`, and the cached reply is rendered back with the caller's own ids — the replayed answer talks about *your* task, not last run's). The bookkeeping numbers are *declared metadata*: we listed exactly which fields (`created_at`, `pid`, lease expiry…) are noise; anything not on that list must match byte-for-byte.

And a pair the cache correctly *refuses* to equate, from the same capture: one run's task record had event log `[created, claimed, spawned, heartbeat]`, the next had one more `heartbeat`. The world actually changed, so that call runs live.

**The Gemini thinking-token problem.** Gemini attaches a cryptographic "thought signature" to every tool call (`call_346990__thought__EosRCogR…`) and requires it echoed back on later turns. It is different on every generation, so no two runs are ever byte-equal past their first tool call — and previously, one live turn poisoned every later lookup in the conversation. The fix (the generic `signed_call_handles` option in the cached-response package, pinned here): for the *cache key only*, each signed token becomes a stable positional handle (`call_h0000`); the request sent to Gemini and the response served back always carry the real bytes. Measured on captured sessions through this exact proxy: record pass 8/8 turns live, repeat pass **8/8 served from cache, zero model calls**, reply rendered with the repeat's own ids.

**What the 2026-09-11 measured run taught, and what changed because of it.** A full local `ci-eval-pr.sh` run through the armed sidecar (54 units, 6,769 model calls) measured 4.6% hits — and said exactly why. 57% of calls never reached lookup because the alias contract raised (board listings outgrew the 128-handle cap, scan order shifted between turns, and conversations entering the cache mid-way hit signed history it never recorded); and among same-shape twin pairs mined from the store, the dominant divergence was kanban board listings embedding every other case's card titles. Both causes are now fixed at their sources:

- **cached-response 0.2.0** (stack #1–#6 plus the fixes merged and released to PyPI): a broken alias contract now degrades to per-call misses instead of poisoning the conversation — first binding wins and conflicting/over-cap proposals are dropped, unrecorded signed history is forwarded with the client's exact bytes, and any remaining contract failure fails open to a live call instead of an HTTP 500 (741 of those killed the 2026-09-10 run). haoxuw/cached-response#7 and #8 carry the details and the regression tests that pin the measured failure shapes.
- **Board-row purge (commit 2, this repo):** the bench harness already deleted a delegated card's attachments and worker log at settle; it now also deletes the card's board rows, scoped to the run's own cards. Repetitions of a task stop seeing the previous rep's cards in `kanban_list`/`kanban_show` tool results, and the board no longer grows without bound across an eval run — which is also what overflowed the alias cap.

**Self-learning** exists in the package but is pinned **off** here (`learning=False`): with it on, a model judge reviews near-miss pairs and can mint narrow, machine-validated patterns (the judge may only answer safe/unsafe/uncertain — Python writes the regexes, bounded, and uncertain means miss). We start deterministic; the judge can be enabled later once its cost is measured.

## Where the code lives

- **Generic machinery** — cached-response package (aliases, declared metadata, guards, `signed_call_handles`, the alias-degradation fixes): merged to haoxuw/cached-response `main` and released as **0.2.0 on PyPI**; `values.yaml` pins `cached-response==0.2.0` (immutable — PyPI releases cannot be re-uploaded in place).
- **kube-agents incision** — one chart knob, one sidecar file whose cache-specific surface is a decorator with five arguments, the bench-harness purge extension, and one CI switch (`EVAL_RESPONSE_CACHE` in `ci-env.sh`) driving both deploy and stats archival.

## Testing

- Package: 366/366 tests at the released 0.2.0, including the new alias-resilience suite that pins the measured failure shapes; the measurement session's probe scripts (five landmine cases plus the two-turn mint sequence) all pass, where case A and the sequence raised at the previous pin.
- Bench: 669/669 (`bench/tests/`), including two new tests — the purge script carries the board-row deletion, and the exact embedded cleanup runs against a scratch SQLite board and deletes only the named cards.
- Proxy end-to-end over HTTP against real captured sessions: record 8/8 live → repeat 8/8 cached, zero upstream calls, no id leakage.
- Chart: `helm lint` clean; off-state byte-identical to an `origin/main` worktree **re-verified after rebasing over the LiteLLM redaction feature (#1364) and the values schema** — the schema gains the two `litellm.responseCache*` keys, and the checksum union keeps both features' pod-roll triggers; template refuses invalid modes and `replicaCount>1`.
- Gemini determinism spot-check (3 identical temp-0 calls): outputs differ — documented as why replay, not sampling determinism, is the reproducibility mechanism.

### Live validation

The 2026-09-11 armed run in `haoxuw-gke-dev` is the exercised path for the sidecar itself: 6,769 calls through this exact proxy, counters archived, fail-open confirmed under 3,878 alias rejections (zero 500s, against 741 in the 2026-09-10 run before fail-open). The remeasure with 0.2.0 plus the board purge — commit 4's arming run — is the pending step; per-rung expectations and the recipe are in #948, and results go there like the two prior measurement comments.

## Self-Review

Carried from the previous head (adversarial pass in a clean subagent; all dispositions in the PR history): single CI switch so deploy and measurement cannot disagree; template refuses `replicaCount>1` (pod-local store) and invalid mode strings; dead env vars removed; deps pinned; counters-only public artifacts. New in this head, author-reviewed pending the next clean-context pass: the board-row purge is scoped to the run's own delegated cards (concurrent units' live cards untouched) and schema-agnostic (every table keyed by `task_id`, so a hermes image bump does not silently orphan rows); the rebase resolution was verified by the byte-identical off-state render above; the package pin moved from a commit tarball to the released version with no install-path change.

## Risk & Rollout

Off by default; worst case with the knob on is a sidecar that misses everything and forwards live. Revert is one commit. A cached repetition measures the cache rather than run-to-run variance, so graded production runs must never enable this; the intended beneficiaries are retries, dev loops, and measurement runs like commit 4's.

---


haoxuw:
Closing this in favour of a different approach.

**What this PR measured.** A cached-response sidecar in front of the LiteLLM gateway, replaying model answers across the presubmit's repetitions. Three measured runs on a dev install:

| run | hits |
|---|---|
| 2026-09-11, parallel lanes, package 0.2.0 | 4.6% |
| 2026-09-15, full 54-unit run, alias fixes landed | 6.9% |
| serial lanes, clean board, package 0.2.1 | 12.5% |

**Why it stops there.** The cache keys on the whole conversation, so a repetition only replays while it stays byte-identical to the recorded one. Measured directly: an identical conversation replays 7/7 (100%), but inserting a single extra turn near the start drops it to 1/9 (11%) — and 11% is what the eval actually produces, because repetitions diverge in shape rather than in wording. Re-keying the recorded turns under progressively looser normalisation confirmed there is nothing left in the matching layer: even masking every number, maximally unsafe, merges 23 of 5,294 turns (0.4%).

The chain breaks at turn 1 or 2 every time, and never because the model disagreed — the assistant message was byte-identical where it could be compared. It breaks on sandbox security-scan prompts quoting the agent's own freshly-written commands, and on tool results differing by heartbeat count, model-written card text and board structure.

**What replaces it.** Rather than replaying model answers, skip the eval when the change cannot alter what the agent does. `hack/ci-eval-pr.sh` already has a revalidation gate; widening it from a hand-maintained inert-path denylist to an allowlist of what the agent actually runs at runtime takes the share of merges it can revalidate from 6.0% to 26.5% — measured over the 200 most recent first-parent merges to `main`, roughly 106 hours of eval matrix. That carries no risk of serving a wrong model answer, because no model answer is reused.

**Worth keeping from this branch, if anyone wants it:**

- The sidecar OOM: at a 512Mi limit with the store on the container filesystem it was killed mid-run and silently lost the cache, so counters went backwards. Any future measurement needs a larger limit and a real volume.
- `cached-response` 0.2.0 and 0.2.1 are released on PyPI with the alias-contract fixes; the package is no longer referenced by this repository.
- The bench harness change that deletes a run's own kanban board rows at settle stands on its own and is worth re-proposing separately.
- The agent sends no `temperature`, so every model call runs at the provider default: three identical requests returned three different answers. Deliberately left as-is — graded runs want the model's real sampling behaviour.

Issue #1184, opened 2026-09-02
eval presubmit: infra-blocked repetitions are graded, not classified — one lost rep reds a whole case and every smoke run on 2026-09-02 is RED


bradhoekstra:
Caught-by: evals

## What is happening

Every `pull-kube-agents-smoke-test` run on 2026-09-02 is RED, and the dominant cause on all of them is the same record class: **empty trajectory, `tokens.total` 0 — the agent never ran.** These are infrastructure losses, but they reach the judge and are graded, so each one fails its case at rung 3 (`NOT_A_REAL_RUN`) and reds the pull request.

With the gate merge-blocking as of today ~14:00Z, that is every open pull request blocked on noise. Six builds across five unrelated PRs:

| PR | build (`gs://kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/…/pull-kube-agents-smoke-test/…`) | started (UTC) | pass rate | red cases | of which `NOT_A_REAL_RUN` | blocked reps |
|----|----|----|----|----|----|----|
| #1132 | `1132/…/2094894318463488000` | 09-01 21:05 | 66.7% | 4 | 3 | 4 / 51 |
| #965 | `965/…/2095156994938769408` | 09-02 14:29 | 68.8% | 9 | 9 | 9 / 60 |
| #1169 | `1169/…/2095170383677231104` | 09-02 15:22 | 55.3% | 4 | 3 | 3 / 60 |
| #926 | `926/…/2095171818145976320` | 09-02 15:27 | 59.5% | 4 | 4 | 5 / 60 |
| #1170 | `1170/…/2095176282760286208` | 09-02 15:45 | 46.9% | 9 | 7 | 11 / 60 |
| #1174 | `1174/…/2095179595249618944` | 09-02 15:58 | 48.1% | 16 | 14 | 16 / 60 |

48 blocked repetitions in total. Every diff in that list is unrelated to the eval path — a NetworkPolicy DNS fix, a docs change, an image pin.

## Rung 3 has no tolerance, so a 5% repetition loss reds the run

This is why the pass rate never has to fall far for the gate to go red. One blocked repetition fails the whole case, whatever the others did.

`cost-idle-pool-probe` on #1174 is the clean example:

```
| `cost-idle-pool-probe` | cost | FAILED (rung 3) | 2/2 | repetition 3: the record is not
evidence of a real agent run: the trajectory is empty: the agent made no tool calls, which
[...]

PR #1311, opened 2026-09-08
feat(dashboard): gate-health banner from health.json and deep-link parameters for the Chat bot's evidence links

jayantid:
## Summary

The eval dashboard becomes the evidence drill-down for the CI health adjudicator (#1305). A banner at the top of the page content renders the `health.json` verdict written beside `data.json` — a state pill (glyph, word and colour together), `since` as an absolute UTC stamp plus the duration, the cause sentence, the advice sentence, and a `details ↓` link into the gate band — and the page's URL now carries deep-link parameters (`?cases=…&since=…&until=…#gate`) that highlight the named cases in the case × run matrix, ring the runs inside the window, and narrow the failure-signature Pareto to that window. Without a `health.json` the page is today's page apart from an empty container; the PR merges independently of #1305.

## Why This Change

The Chat messages #1305 posts say "OUTAGE: cluster-agent-crashloop-debug, …, since 09:00Z" and link to the dashboard, and the reader then has to find those rows in a 30-column matrix and work out which columns fall in the incident by hovering. #1202 and #1278 were both diagnosed that way, by hand, at the cost of merge latency while people typed `/retest` into an outage. With this change the link lands on the gate band with the incident's cases highlighted, its window marked, and the Pareto showing only the failures inside it; the banner above the fold says what state the gate is in before anyone reads a heatmap.

## What Changed

- `scripts/eval_dashboard/render.py` — `--health health.json` (optional), `load_health`/`normalize_health` (absent, unparseable, non-object, or an unknown state → no banner; every other field optional; lone surrogates replaced with U+FFFD; an out-of-range stamp reads as unknown), `health_html` (the banner; the "for Xh Ym" duration is `generated_at − since`, the verdict's own clock, so the baked page and the client re-render agree), `health_details_href` (the deep link built from `failing_cases` and `since`), the `__HEALTH_JSON__` bootstrap. The evidence heading's anchor is now `#evidence` (was `#nightly`; the table id moves to `evidence-table` so the id is unique). `health.json` is deliberately not copied into the out-dir: the writer owns that object, and republishing a copy would overwrite a fresher verdict.
- `scripts/eval_dashboard/template/index.html.tmpl` — the JS mirror of the above (`normalizeHealth`, `healthHtml`, `durationText`, `utcIso`, `wellFormed`), a `health.json` poll on the same 60 s cadence and protocol guard as the data poll with the same fail-safe posture (a failed or unusable poll keeps the last verdict and labels it `UNREACHABLE`; an old `generated_at` or `stale: true` labels it `STALE` — words, not amber alone; a verdict is rendered before it is committed, so a bad one can never freeze the data feed), the deep-link parser (`linkState`: case ids against `[A-Za-z0-9][A-Za-z0-9._-]{0,79}`, at most 50; `since`/`until` against an ISO 8601 grammar; `until` only alongside `since` and not before it; only a bare `#section` fragment survives into the `clear` link), the matrix/Pareto changes behind a `link` argument whose default reproduces the server markup exactly, a note under the gate heading naming what is highlighted and the window in words with a `clear` control, nav-tab selection from the fragment, and `scroll-margin-top` so a fragment lands below the sticky bar. `parseIso` also reads a space-separated stamp as UTC, as `fromisoformat` does. Banner and highlight colours are the existing `--good/--warn/--crit` tokens.
- `scripts/eval_dashboard/SCHEMA.md` — "health.json (optional input)" and "Page URL parameters" (the grammar table), after the collector's Sources section; `docs/README.md` — the SCHEMA.md row's summary names the two added contracts (one in-place line, no re-alignment).
- `scripts/test_eval_dashboard.py` — see Testing.

The param grammar as implemented: query before fragment; `cases=<id>(,<id>)*` with each id matching `[A-Za-z0-9][A-Za-z0-9._-]{0,79}` (first 50 read; ids with no matrix row ignored; when none matches no row dims); `since=<ISO 8601>` (`YYYY-MM-DD`, optional `THH:MM[:SS[.ffffff]]`, optional `Z`/`±HH:MM`/`±HHMM`; naive = UTC); `until=` same shape, honoured only with `since` and when `≥ since`; window inclusive on both ends and open-ended without `until`; fragment `#agent | #gate | #evidence | #release`.

## Context

Companion to #1305 (`feat/ci-health-adjudicator`, in review), which writes `health.json`; the URL contract above is the one agreed for its poster's links (its current poster links the bare dashboard URL — adopting the parameters is its follow-up). This PR does not import or depend on #1305's code — it reads `health.json` as an optional input and degrades to today's page without it, so the two merge in either order; both touch `scripts/eval_dashboard/__init__.py`'s docstring and append to `SCHEMA.md`, a textual conflict whichever lands second resolves in a sentence. Motivating incidents: #1202 (merge latency under a red gate), #1278 (the seeded-fleet outage). Overlap scan: #1165 (mine, three lines of wording in the domains tile across the same three files — a merge-conflict warning, not the same problem). No open issue covers this.

## Testing

- `python3 -m unittest discover -s scripts` — 1136 tests, OK (1 pre-existing skip). `scripts/test_eval_dashboard.py` grows from 58 to 97 tests.
  - Server side: banner per state and its placement above the hero; since/cause/advice/details link; the bare `#gate` link without cases or since; negative and floored durations; absent → no banner and a `null` bootstrap; seven malformed shapes → no banner, page renders; field-by-field degradation of off-shape optional fields; hostile cause/advice/case ids escaped in the banner, the href and the bootstrap; out-of-range offset stamps and lone surrogates never abort the render; the banner on the empty-state page; `health.json` not copied to the out-dir; `uri_component` against `encodeURIComponent`'s unreserved set; section anchors unique and used by the nav; the bootstrap carries health with `<` escaped; the poll, its cadence and its protocol guard; STALE/UNREACHABLE text labels; every state, glyph, class and parameter name present in the JS mirror.
  - `BrowserParityTest` (16) and `BrowserPollTest` (4), skipped with a reason when headless Chrome is absent or cannot launch (it is present here and on `ubuntu-latest`). Parity: the shipped script runs for real and `#app` as the browser parsed the baked HTML is compared with `#app` after `renderAll()` — identical with a banner, with a space-separated `since`, without health, and on the empty state (the banner's wall-clock `checked` span is the one normalized difference, asserted separately as `STALE · checked 12:00 UTC · Nm ago`). Deep links on the fixture with known timestamps (C 08-31T09:00, D 11:00, E 09-01T08:00, F 09:00): `?cases=case-a,case-e,nope` → 2 highlighted, 3 dimmed, `nope` absent; `[08-31T10:00, 09-01T08:30]` → exactly D and E ringed in every row and the head, Pareto retitled to the window with 1 kanban-columns miss (not the rolling window's 3), no reason-less-infra group from C, no run-D event reasons; inclusive bounds (`[08:00, 09:00]` rings E and F) and a `+02:00` offset; open-ended `since`; same-day window text; `until` before `since` or alone ignored; an inactive linked case (`case-f`) highlights nothing and the note says so; the banner's own `details` href applies the link; `cases=<script>…`, `since=` and a hostile fragment never reach the DOM unescaped; the 80-char and 50-id caps; the nav tab follows the fragment and falls back to the first tab; the baked page is unchanged by parameters. Poll, over an `http.server` in a thread: a served OUTAGE replaces the baked GREEN with no UNREACHABLE label and the data badge untouched; a missing, `PURPLE` or non-JSON `health.json` keeps the GREEN with `UNREACHABLE · checked 12:00 UTC` and the data badge untouched; a lone surrogate in `failing_cases` renders (`%EF%BF%BD` in the href) instead of freezing the page; `stale: true` labels `STALE`.
- prettier is not installed locally; the SCHEMA.md table was padded by a script using prettier's width rule (every character in it is width 1 under its East Asian Width handling).

### Live validation

Not published — the coordinator publishes; nothing was written to the bucket or any cloud resource. Rendered locally from the repo's six real Prow builds (`collect.py --from-dir testdata --from-dir testdata_reps` → 3 measured runs, 16 cases) with the repo's `case-notes.yaml` and `events.yaml`, three ways, and captured with headless Chrome (full page via `~/bin/shot`, plus 760 px top crops):

- **OUTAGE** (`health.json` naming `security-overgrant-probe` and `upgrades-lagging-master-probe`, `since` = the second-newest run's start): the banner is the first element under the sticky bar — `🔴 OUTAGE · since 2026-08-31 16:50 UTC · for 8d 2h · checked 19:25 UTC · 1m ago · details ↓`, then the cause and advice lines, red left border and pill. Opening the banner's own `details ↓` href (`?cases=security-overgrant-probe,upgrades-lagging-master-probe&since=2026-08-31T16%3A50%3A58Z#gate`): the page lands on **The gate** with its heading visible below the sticky bar; the note reads `highlighting security-overgrant-probe, upgrades-lagging-master-probe · showing 2026-08-31 16:50 UTC → now window · clear`; the two rows carry the accent bar and a `linked` badge in bold while the other 12 rows are dimmed and still legible; the two newest columns are ringed and their head cells underlined; the Pareto is titled `Failure signatures · linked window`. DOM counts from the same capture: `hl` 2, `dim` 12, `mx-col win` 2.
- **GREEN**: a single-row banner (`🟢 GREEN · since … · checked … · details ↓`) with green pill and border, no cause/advice lines; the rest of the page identical to the absent case.
- **Absent** (no `--health`): the page renders exactly as on `main` today — hero first, no banner, no note.

The `file://` captures do not poll (the script's existing protocol guard); the poll's replace / keep-and-label-UNREACHABLE / STALE paths were observed through `BrowserPollTest` over a local HTTP server rather than on the published page. Not covered: the published page with a real `health.json` beside it (needs #1305's job to have written one), and the refresh job passing `--health` (`hack/ci-dashboard-refresh.sh` is unchanged here; until it does, the banner appears on the live page through the client poll only, a few hundred ms after load — wiring the flag is a one-line follow-up once #1305 lands).

Screenshots, published through `scripts/pr_evidence_screenshot.sh` (the `--file` form: captured against the tree at 8995409f, published at that commit):

health-banner-outage

_Published 20260908T201231Z at commit 8995409fafe8 — pre-captured image (outage-top.png)._

gate-deeplink-highlight

_Published 20260908T201234Z at commit 8995409fafe8 — pre-captured image (outage-deeplink-gate.png)._

health-banner-green

_Published 20260908T201237Z at commit 8995409fafe8 — pre-captured image (green-top.png)._

health-banner-absent

_Published 20260908T201240Z at commit 8995409fafe8 — pre-captured image (absent-top.png)._

## Self-Review

Two passes, each in a subagent handed the diff range `upstream/main..HEAD` at the first commit (c9e785be) and the skill file, nothing else: `review-adversarial` and `review-docs-drift`. Everything marked fixed is in the second commit (8995409f) and covered by a test where a test can reach it; the counts above are after the fixes.

- **`render.py` raised on `health.json` content** — an offset that pushes year 1 or 9999 out of `datetime`'s range (`OverflowError` past `parse_iso`'s `ValueError` guard) and a lone surrogate (`UnicodeEncodeError` in `quote()` and `write_text()`), contradicting "never fails" (adversarial, should-fix). Fixed: `parse_iso` catches `OverflowError`; `normalize_health` replaces lone surrogates with U+FFFD (the same code point the JS `wellFormed` writes, so parity holds). Test added.
- **A poisoned poll froze the live page and blamed `data.json`** — a lone surrogate in `failing_cases` passed `normalizeHealth`, the verdict was committed, `encodeURIComponent` threw, and every later data re-render threw inside `refresh()`'s `try`, labelling the data badge UNREACHABLE (adversarial, should-fix; reproduced in Chrome by the pass). Fixed: the poll renders a verdict before committing it, `appHtml` guards the banner, strings are made well-formed. `BrowserPollTest` added over a real HTTP server; its poisoned case now renders.
- **Space-separated stamps** — `fromisoformat` accepts `2026-08-31 10:00:00`; `Date.parse` read it as local time (adversarial, should-fix, not reachable from #1305's writer). Fixed in `parseIso` for every stamp the page reads; parity test added.
- **Browser harness** — `check=True` hid Chrome's stderr, the default profile was shared, a broken Chrome would fail 13 tests instead of skipping, the probe was spliced before the first script regardless of `#app` (adversarial, should-fix). Fixed: return code asserted with stderr, private `--user-data-dir` except on macOS (the branded build's updater hangs on a fresh profile; documented in the file), a smoke launch that skips the classes with Chrome's own message, probe anchored after `#app`. Not done: an opt-in environment variable — a healthy Chrome in CI is coverage we want, and the smoke launch already handles the unhealthy one.
- **Tests asserting source text or nothing; poll untested; boundaries** (adversarial, nit). Fixed: the vacuous docstring assertion removed; the poll class above; inclusive bounds and an offset `since` in the URL tested. The `js.count(...) == 2` STALE/UNREACHABLE tripwires stay: they are the CI floor where Chrome is absent, alongside the pre-existing badge tripwire they extend.
- **Magic constants** on new Python lines (adversarial, nit). Fixed: `MS_PER_S`, `MINUTES_PER_HOUR`, `MINUTES_PER_DAY`, `LABEL_SINCE_UNKNOWN`, `LABEL_CHECKED_UNKNOWN`, `LONE_SURROGATE_RE`, `REPLACEMENT_CHAR`.
- **Link note claimed a highlight for an inactive case** with no matrix row (adversarial nit; docs-drift nit). Fixed with a test.
- **`renderNav` cleared every tab for a fragment naming no tab** (adversarial, nit). Fixed: first tab stays on; tested.
- **SCHEMA.md said the poster's messages use the parameters**; #1305's poster links the bare URL (adversarial nit). Fixed: the sentence now says what the parameters are for; Context says the poster adopts them later. The `__init__.py`/`SCHEMA.md` overlap with #1305 is noted in Context.
- **Prose named a writer not in this tree** ("the CI health adjudicator") and two unverifiable keys (docs-drift, should-fix). Fixed: "a separate job writes … this directory holds no writer", and "any other key is ignored".
- **SCHEMA.md table not prettier-formatted** (docs-drift, should-fix). Fixed by re-padding under prettier's width rule; prettier itself was not run (not installed).
- **`docs/README.md` row incomplete** for what SCHEMA.md now documents (docs-drift, should-fix). Fixed with one in-place line, per §5's rule against re-alignment.
- **Wording nits** — offset grammar (`±HHMM` also accepted), "case on record" vs "row in the matrix", `stale_after_s` belongs to `data.json`, the duration is conditional on `generated_at ≥ since`, the live-side comment's banner stamp, "top element" vs "below the sticky bar" (docs-drift). All fixed.
- **H1 and section placement** (docs-drift, nit). Sections moved after the collector's Sources, and the intro names them; the H1 stays — every cross-reference names the file by path and the `data.json` contract is still its bulk.
- Recorded by the pass as not findings and left as is: a verdict stays up labelled UNREACHABLE after `health.json` is deleted (the documented posture), the dead `try` around `URLSearchParams`, sub-millisecond duration parity, the `recovering` flag reaching the reader through `advice`.
- **Screenshots left at local paths** instead of going through `scripts/pr_evidence_screenshot.sh` (kube-agents-bot, on the first review; the body's reason — "the script publishes to the bucket" — was wrong: it pushes an orphan `pr-evidence` branch to the author's fork). Fixed: the four captures are published above with the script's own provenance lines.
- What the passes could not cover and neither could I: Safari/Firefox `Date.parse` (from spec, not a run) and the suite on an `ubuntu-latest` runner. One thing the fixes themselves caught: an early cut of `wellFormed` as a `const` threw a TDZ `ReferenceError` at script start and every browser test failed — the parity harness is doing the job the #1184 lesson asked for.

## Risk & Rollout

Static-site rendering code and its tests; no runtime code path of the agent or operator is touched. New failure modes: a malformed `health.json` renders no banner (never an error, by test); a wrong verdict is the writer's to fix and the banner shows it as written; a deep link with hostile parameters is escaped or dropped (by test); the browser tests add about 70 s to `scripts/` on a runner with Chrome and skip with a reason elsewhere. The `#nightly` → `#evidence` rename breaks any external bookmark to `#nightly` (none exist in the repository; the page still loads at the top). Revert by reverting the two commits; the published page rerenders on the next refresh. Nothing has to happen at merge time; the banner appears on the live page as soon as #1305's job writes `health.json` beside `data.json`.


docs/README.md
```diff
@@ -428,7 +428,7 @@ only what the title does not say.
 | `scripts/dev/README.md` | Component README | The script automating GCP Workload Identity Federation so GitHub Actions can deploy keylessly. | WIF/OIDC CI auth | Repo maintainers |
```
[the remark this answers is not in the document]
jayantid: Right — the reason I gave was wrong: the script pushes an orphan `pr-evidence` branch to the author's fork, not to the bucket. The four captures (OUTAGE banner, the deep-linked gate band, GREEN, absent) are now published through `scripts/pr_evidence_screenshot.sh` and inlined in **Live validation**, with the script's provenance lines (published at 8995409f, the commit they were captured against); the disposition is recorded in **Self-Review**. No code change.

scripts/eval_dashboard/render.py
```diff
@@ -510,6 +600,102 @@ def load_events(path: pathlib.Path | None) -> dict:
...
+        "stale": raw.get("stale") is True,
+    }
+
+
+def health_details_href(health: dict) -> str:
+    """The banner's drill-down: the gate anchor, carrying the failing cases
+    and the incident start as deep-link parameters when there are any."""
+    params = []
+    if health["failing_cases"]:
+        joined = ",".join(uri_component(c) for c in health["failing_cases"])
```
toshiowang, line 656: LOW: this href applies no count cap or id-grammar check to `failing_cases`, while the client parser honours only the first 50 grammar-valid ids. A verdict with off-grammar ids or more than 50 cases produces a details link that silently under-highlights. No current writer (#1305) emits such a verdict, so this is a robustness note, not a defect today — a cap/filter here would keep the link and the parser on the same contract.

scripts/eval_dashboard/template/index.html.tmpl
```diff
@@ -194,25 +229,51 @@ document.addEventListener("mousemove",e=>{
...
+  healthDetailsAnchor: "#gate",
+  linkParamCases: "cases",
+  linkParamSince: "since",
+  linkParamUntil: "until",
+  // A case id is a bench/tasks/ directory name; anything else is dropped
+  // unread. At most this many ids are honoured from one link.
+  caseIdRe: /^[A-Za-z0-9][A-Za-z0-9._-]{0,79}$/,
+  linkMaxCases: 50,
+  // The ISO 8601 shapes a deep link may carry; parseIso does the rest.
+  isoParamRe: /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}(:\d{2}(\.\d{1,6})?)?)?([zZ]|[+-]\d{2}:?\d{2})?$/,
```
toshiowang, line 274: LOW: the `±HHMM` (no-colon) offset form passes this grammar but its parsing then relies on `Date.parse`, which is engine-dependent for that form, and no test covers it even in Chrome. Failure mode is a silently unapplied window. Either a test pinning it in Chrome or normalising the offset to `±HH:MM` before parsing would close it.

toshiowang (commented):
Two LOW observations from a full pass; neither blocks. Approving separately.

toshiowang (approved):

jayantid:
Superseded by #1335, which folds this banner and the URL contract (cases/since/until, #gate/#agent) into the new Brief landing page. Closing without merging.

PR #1329, opened 2026-09-09
fix(telemetry): restore hermes_otel on force override and preserve collector netpol (#933)

### LINKED ISSUE #933: fix(agent): hermes_otel keeps the baked gke-managed-otel backend when the operator resolves otlpEndpointSource=None

shalinibhatiapl:
## Summary

When telemetry collector discovery resolves to `None` (or `OTEL_SDK_DISABLED=true` is set), setting `HERMES_OTEL_ENABLED=true` force-enables traces export using the baked collector fallback. This change ensures that the `hermes_otel` plugin is re-enabled from disk when forced, preserves explicitly disabled pristine templates, ensures restart idempotency, normalizes environment variable parsing, retains collector egress in NetworkPolicies when telemetry is forced, isolates gate-check environments from pod-inherited overrides, and updates documentation.

## Why This Change

Issue #933: On clusters where OpenTelemetry discovery resolves to `None`, setting `HERMES_OTEL_ENABLED=true` in `spec.deployment.env` was intended to force trace exporting. However:
1. The container entrypoint evaluated `OTEL_SDK_DISABLED=true` without prioritizing `HERMES_OTEL_ENABLED=true`, causing `--disabled` to be passed to `otel_config.py`.
2. `otel_config.py` re-enabling logic lacked protection for templates with explicit `enabled: false`, and was non-idempotent across restarts when re-enabling.
3. The operator's NetworkPolicy builder dropped collector egress rules when `otlpDisabled` was true, preventing traces from reaching the collector even when forced.
4. `entrypoint_gate_check.sh` inherited `HERMES_OTEL_ENABLED` from the pod environment, leading to false-positive test failures when run on a pod configured with the force override.

## What Changed

- **`deploy/shared/docker-entrypoint.sh`**: Prioritized `HERMES_OTEL_ENABLED=true` over `OTEL_SDK_DISABLED=true` when deciding whether to pass `--disabled` to `otel_config.py`. Normalized variable casing and whitespace (`tr '[:upper:]' '[:lower:]' | tr -d '[:space:]'`).
- **`deploy/shared/entrypoint_gate_check.sh`**: Cleared `HERMES_OTEL_ENABLED` via `env -u` in `run_entrypoint` so pod-inherited environment overrides do not leak into gate-check cases, and added verification that Step 4 disables `hermes_otel` configs when `OTEL_SDK_DISABLED=true`.
- **`deploy/shared/otel_config.py`**: Guarded re-enabling so it flips `enabled: true` if previously disabled on disk while respecting explicit `enabled: false` in pristine source templates, and preserves `enabled: true` across subsequent restarts to ensure restart idempotency.
- **`k8s-operator/internal/controller/platformagent_manifests.go`**: Added `isHermesOtelForced()` and updated `buildNetworkPolicy` rule 8 (`!otlpDisabled || hermesForced`) so that ports 4317/4318 egress to the collector namespace is preserved when `HERMES_OTEL_ENABLED=true` is present in `spec.deployment.env`. Defined `envHermesOtelEnabled` constant to avoid magic strings.
- **`agents/platform/scripts/cluster_agent_profile.py`**: Adjusted log message to neutral phrasing when telemetry configuration fails.
- **Tests**:
  - `tests/test_otel_config.py`: Added tests for re-enabling from disk, preserving pristine disabled templates, and verifying restart idempotency.
  - `agents/platform/scripts/test_cluster_agent_profile.py`: Added case-insensitivity tests for `OTEL_SDK_DISABLED`.
  - `k8s-operator/internal/controller/telemetry_test.go`: Added `TestBuildNetworkPolicyAllowsCollectorEgressWhenHermesOtelForced`.
- **Docs**: Updated documentation in `charts/kube-agents/README.md`, `docs/site/src/content/docs/deploy/telemetry.md`, `agents/platform/skills/kube-agents-observability/SKILL.md`, and credential isolation docs.

## Context

Closes #933

## Testing

- `make docs-check`: Passed (0 errors, 38.0k chars within budget).
- Go unit tests: `go test -v ./internal/controller/...` passed.
- Python unit tests:
  - `tests/test_otel_config.py`: 32/32 passed.
  - `agents/platform/scripts/test_cluster_agent_profile.py`: passed.
  - `tests/test_docker_entrypoint.py`: 64/64 passed.

### Live validation

- **Cluster**: GKE cluster `platform-agent-host` (`us-east4`), namespace `kubeagents-system`.
- **Image**: `us-east4-docker.pkg.dev/shalinibhatia-gkedemos/kube-agents/platform-agent:dev-20260909-113917`.
- **Baseline**: Verified startup with discovered collector endpoint (`http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4318/v1/traces`), registered 13 hooks.
- **Disabled Test (`HERMES_OTEL_ENABLED="false"`)**: Patched `spec.deployment.env`. Observed rollout: `/opt/data/plugins/hermes_otel/config.yaml` wrote `enabled: false` and `backends: []`. Gateway logs output: `[hermes-otel] ✗ Disabled via config (enabled=false)`.
- **Forced Override Test (`OTEL_SDK_DISABLED="true"`, `HERMES_OTEL_ENABLED="true"`)**: Patched `spec.deployment.env`. Observed rollout: `/opt/data/plugins/hermes_otel/config.yaml` was re-enabled from disk (`enabled: true`), collector backend was restored. Gateway logs output: `[hermes-otel] ✓ gke-managed-otel at http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4318/v1/traces` with 13 hooks registered.
- **NetworkPolicy Egress Rule**: Inspected live NetworkPolicy `platform-agent-netpol` on cluster in namespace `kubeagents-system`:
  ```yaml
  - ports:
    - port: 4317
      protocol: TCP
    - port: 4318
      protocol: TCP
    to:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: gke-managed-otel
  ```
  Verified egress to `gke-managed-otel` on ports 4317/4318 remained admitted on the live NetworkPolicy object in the forced state (`HERMES_OTEL_ENABLED=true`) despite `OTEL_SDK_DISABLED="true"`.
- **What could not be covered**: Real span ingestion into an external third-party collector (e.g., Datadog or Honeycomb) without deploying non-standard external infrastructure into the cluster; covered GKE Managed OTel collector integration and discovery resolution to `None`.
- **Cleanup**: Reverted all `spec.deployment.env` patches back to pristine state, and confirmed the temporary development image `us-east4-docker.pkg.dev/shalinibhatia-gkedemos/kube-agents/platform-agent:dev-20260909-113917` was removed.

## Self-Review

Pre-PR review passes executed: `review-adversarial` and `review-docs-drift`.
Each pass was run in an isolated fresh session / subagent context that did not author the change, ensuring findings were evaluated without authoring bias.

| Angle / Pass | Finding | Disposition |
| :--- | :--- | :--- |
| **Angle D: Security** (`review-adversarial`) | `buildNetworkPolicy` dropped collector egress when `otlpDisabled=true`, blocking spans even when `HERMES_OTEL_ENABLED=true` forced export (#933). | **Fixed**: Added `isHermesOtelForced()` in `platformagent_manifests.go` to retain ports 4317/4318 egress when forced. Added test in `telemetry_test.go`. |
| **Angle A: Error Handling** (`review-adversarial`) | `apply()` in `otel_config.py` forced `enabled: true` even if pristine template explicitly had `enabled: false`. | **Fixed**: Guarded re-enabling so it only toggles when previously disabled on disk and preserves pristine template disables. Added test in `tests/test_otel_config.py`. |
| **Angle A / F: Idempotency** (`review-adversarial`) | `apply()` in `otel_config.py` was non-idempotent across restarts when re-enabling telemetry from a disabled state due to `current.get("enabled") is False` check. | **Fixed in `8cef2d95`**: Preserved `enabled: true` across restarts with `current.get(FIELD_ENABLED) in (True, False)` guard, avoiding oscillation or redundant file rewrites. Added regression test `test_re_enabling_is_idempotent_across_multiple_restarts`. |
| **Angle F: Wire Formats** (`review-adversarial`) | Environment variables `HERMES_OTEL_ENABLED` and `OTEL_SDK_DISABLED` did not normalize mixed-case/whitespace (`TRUE`, ` True `). | **Fixed**: Normalized values with `tr '[:upper:]' '[:lower:]' \| tr -d '[:space:]'` in `docker-entrypoint.sh`. Added test in `test_cluster_agent_profile.py`. |
| **Angle G: Tests** (`review-adversarial`) | Missing unit test assertion verifying that forcing `HERMES_OTEL_ENABLED=true` keeps NetworkPolicy egress open. | **Fixed**: Added `TestBuildNetworkPolicyAllowsCollectorEgressWhenHermesOtelForced` in `telemetry_test.go`. |
| **Angle H: Rules** (`review-adversarial`) | Hardcoded string `"HERMES_OTEL_ENABLED"` violated the "No magic constants" rule. | **Fixed**: Defined `envHermesOtelEnabled` in top-level `const (...)` in `platformagent_manifests.go`. |
| **Angle I: Operability** (`review-adversarial`) | Misleading error log in `cluster_agent_profile.py` claimed "telemetry left disabled" even when telemetry was previously enabled. | **Fixed**: Changed log to neutral `"configuring OpenTelemetry failed ({e}); telemetry config unchanged"`. |
| **Pass: Docs Drift & Angle H** (`review-docs-drift`) | SKILL.md directed inspecting profile config for `enabled: false` and `backends: []` instead of plugin config. | **Fixed in `c62b7ca4`**: Clarified that plugin configs (`/opt/data/plugins/hermes_otel/config.yaml`) hold `enabled: false` and `backends: []`, while profile configs retain `hermes_otel` in `plugins.enabled`. |
| **Angle E: Isolation** (`review-adversarial`) | `deploy/shared/entrypoint_gate_check.sh` disabled sweep inherited `HERMES_OTEL_ENABLED=true` from pod environment when running on overridden pod. | **Fixed**: Added `-u HERMES_OTEL_ENABLED` to `run_entrypoint` env invocations so pod-inherited overrides do not leak into gate-check cases. |
| **Angle C: Cross-file** (`review-adversarial`) | `isHermesOtelForced` forward scan vs `mergeEnvVars` last-wins dictionary on duplicate env entries. | **Deliberately not fixed**: Manifests should not supply duplicate env names. |
| **Angle A: Fallbacks** (`review-adversarial`) | Re-enabling telemetry without `source_path` and without `endpoint` leaves `backends: []`. | **Deliberately not fixed**: Unreachable in runtime; entrypoint and cluster agent profile scripts always supply defaults plugin path. |
| **Angle H: Conventions** (`review-adversarial`) | Documentation asserts synthetic gateway startup log line `[hermes-otel] Telemetry disabled (...)`. | **Fixed**: Updated `docs/site/src/content/docs/deploy/telemetry.md` to quote the observed log line `[hermes-otel] ✗ Disabled via config (enabled=false)`. |
| **Pass: Docs Drift (Advisory)** (`review-docs-drift`) | `security-and-iam.md` collector egress omission claim does not mention `HERMES_OTEL_ENABLED=true` override. | **Deliberately not fixed**: Canonically documented in `deploy/telemetry.md` and `reference/credential-isolation.md`. |
| **Pass: Docs Drift & Angle H** (`review-docs-drift`) | Helm chart README omitted `HERMES_OTEL_ENABLED` and stated collector egress is unconditionally omitted on `None` clusters. | **Fixed**: Updated `charts/kube-agents/README.md` ("Turning telemetry off") to document `HERMES_OTEL_ENABLED` off/force-on switches and conditional NetworkPolicy collector egress. |

## Risk & Rollout

- **Blast radius**: Low. Scoped strictly to OpenTelemetry bootstrap configuration in `docker-entrypoint.sh`, plugin config re-enabling in `otel_config.py`, and the operator's NetworkPolicy egress generation when `HERMES_OTEL_ENABLED=true` is specified in `spec.deployment.env`. Normal deployments without this env var override remain unaffected.
- **Failure modes**: If `HERMES_OTEL_ENABLED=true` is forced in an environment without a reachable collector, traces export attempts may fail silently or log non-fatal warnings at container startup without impacting agent core functionality.
- **Rollout & Reversion**: Can be rolled out safely on running clusters. To revert, remove `HERMES_OTEL_ENABLED` from `spec.deployment.env` or set it to `"false"`. No persistent state, schema, or CRD changes are involved.
- **Merge-time actions**: None; standard CI/CD pipeline builds images on merge.


docs/site/src/content/docs/deploy/telemetry.md
```diff
@@ -73,7 +73,7 @@ Results are cached cluster-wide for 5 minutes, including the "found nothing" ans
...
+`None` also silences the `hermes_otel` plugin, which carries agent spans. When `OTEL_SDK_DISABLED=true` is set (or `HERMES_OTEL_ENABLED=false` is configured), the container entrypoint disables `hermes_otel` (`enabled: false`) and clears its backends list, so no spans are created or exported to the fallback URL and the gateway logs `[hermes-otel] Telemetry disabled (enabled=false, backends=[])` at start-up. Setting `HERMES_OTEL_ENABLED=true` in `spec.deployment.env` overrides this if spans should be exported regardless.
```
mplakhtiy: **MEDIUM — this quotes a start-up log line the gateway never emits.** The sentence claims the gateway logs `[hermes-otel] Telemetry disabled (enabled=false, backends=[])` at start-up.

That string appears nowhere in the repository except this sentence — `grep -rn '\[hermes-otel\]'` over the tree at this head returns this line and nothing else — and the PR's own **Live validation** reports the observed line as `[hermes-otel] ✗ Disabled via config (enabled=false)`. The plugin is a third-party install (`briancaffey/hermes-otel`, pinned in `deploy/docker/Dockerfile`), so the doc cannot dictate its output; it can only describe it.

An operator, or the Cluster Agent following the observability skill, greps the gateway log for the documented string on a `None` cluster, finds nothing, and concludes the disable did not take — the same false-positive family this PR exists to close. Use the string the live test actually observed, or drop the quoted line and describe the effect without quoting output.

mplakhtiy (commented):
Full-diff review at the PR head (`8cef2d9`), run in a context that did not write the change. `origin/main` (`1784e18c`) merges into the head cleanly, so this covers the PR as merged.

The three code changes are sound and tested. The entrypoint precedence fix in `docker-entrypoint.sh`, the re-enable guard in `otel_config.py`, and `isHermesOtelForced` retaining the gateway's 4317/4318 collector egress when the plugin is forced on are internally consistent and unit-covered. `go build ./...` and `go vet ./internal/controller/` are clean; `tests/test_otel_config.py` passes 32 and `agents/platform/scripts/test_cluster_agent_profile.py` passes 29, both at the head commit. The findings here are documentation-only.

**MEDIUM — `charts/kube-agents/README.md:297-325` "Turning telemetry off" is the one operator-facing doc this change does not reach.** #933 named that section as where the gap was documented. The behaviour is now fixed, but the section is still the canonical Helm-side off switch and it never mentions `HERMES_OTEL_ENABLED` — though this PR makes it both the plugin off switch and the force-on override, and puts it on the CR env allowlist. It also still says flatly of `OTEL_SDK_DISABLED: "false"` that "the NetworkPolicy it renders for a `None` agent carries no collector egress rule" (line 323), which after this change is conditional on `HERMES_OTEL_ENABLED`. `docs/site/src/content/docs/deploy/telemetry.md` got the update; the chart README did not, and the `review-docs-drift` disposition in **Self-Review** lists only `SKILL.md` and `telemetry.md`. An operator who reads the chart README therefore gets the pre-PR world: no force-on knob exists, and a netpol statement that is now false whenever the override is set. Fix the section, or state in **Self-Review** why the chart README is deliberately left to defer to the site page.

### Already addressed at the head

The observability `SKILL.md` finding from the earlier pass — `enabled: false` / `backends: []` attached to the profile-config bullet rather than the plugin-config bullet it describes — is fixed at the head by `c62b7ca4`. The parenthetical now sits on the "Verify the plugin's exporter backend" bullet and says "in the plugin config", which is the file the following `grep` reads. Recorded so it is not re-raised; no action needed.

docs/site/src/content/docs/deploy/telemetry.md
shalinibhatiapl: Fixed in e3133394 — updated docs/site/src/content/docs/deploy/telemetry.md to quote the observed log line [hermes-otel] ✗ Disabled via config (enabled=false) rather than the synthetic string, and reconciled the Self-Review table in the PR description to match.

Updated charts/kube-agents/README.md ("Turning telemetry off") in 22a2115c to document HERMES_OTEL_ENABLED as the plugin off switch and force-on override, noting the allowlist and that setting HERMES_OTEL_ENABLED=true retains ports 4317/4318 collector egress in the gateway NetworkPolicy even when discovery resolves None. Reconciled the Self-Review table in the PR description to match.

mplakhtiy (approved):

PR #1336, opened 2026-09-09
fix(installer): disclose existing-cluster mutations and require opt-in for node-recreating changes (#1293)

### LINKED ISSUE #1293: bug(installer): installing onto a pre-existing cluster makes five undisclosed changes to it, and uninstall.sh reverts almost none
## Summary

Installing onto a **pre-existing** cluster makes up to five permanent changes to that cluster. None of
them appear in the pre-flight summary, `--non-interactive` never asks about any of them, and
`uninstall.sh` reverts almost none. A **failed** install leaves them behind too.

This is the disclosure and reversibility half of the problem. #521 covers the related but distinct
question of *preflighting* whether the cluster can run the agent.

## What the user sees

From a real install log, verbatim and consecutive — note what the summary lists, and what happens next:

```
>>> 11. Pre-Flight Configuration Summary <<<
Please review your selections before provisioning begins:
  • GCP Target Project: <PROJECT_ID>
  • GKE Cluster: <CLUSTER> (us-central1, GKE Standard)
  • gVisor Sandbox Isolation: true
  • AI Model Provider: gemini
  • Permission Boundary: read-only
  • Long-Term Memory: file

>>> 12. Applying the Install (Terraform + Helm) <<<
  ℹ Enabling CMEK database encryption on existing cluster '<CLUSTER>' ...
  ℹ Updating the live cluster control plane; this can take several minutes...
```

Six bullets. None of the five changes below is among them.

## The five changes

| # | Change | Where | Observed | Restarts workloads? |
| 1 | **CMEK database encryption** on the live control plane — plus enabling the KMS API, creating a keyring and key, creating the container service identity, adding a KMS IAM binding | `install.sh:1594`, calls at `:1614-1627` | **All 4 test clusters** | No (~5 min control-plane update) |
| 2 | **Workload Identity pool** enablement | `install.sh:1637`, call at `:1661-1662` | Already on for all 4 — no-op | No |
| 3 | **Node-pool migration to `GKE_METADATA`** for every pool not already on it | `install.sh:1670-1679` | Not triggered in these runs | **Yes — stated explicitly** |
| 4 | **Legacy Calico NetworkPolicy** addon, then enforcement | `install.sh:1690`, calls at `:1725-1730` | **Both Standard clusters** | **Yes — possible** |
| 5 | **A new `gvisor-pool` node pool** (`e2-standard-4`, one per zone) | `terraform/modules/gke-cluster/main.tf:284-304` | **Both Standard clusters** | No, but new billable capacity |

Change 3 is the most disruptive: `install.sh:1672` prints *"Node pool '<name>' uses the legacy GCE
metadata server; migrating to GKE_METADATA (this recreates the pool's nodes)…"* and loops over **every**
node pool on the adopted cluster, including pools running workloads unrelated to kube-agents. It did not
fire in these runs only because all four test clusters already used `GKE_METADATA` — but #1286 shows it
firing in the wild and taking 35–45 minutes on a 9-node pool.

## Why this is a disclosure problem, not a design problem

The out-of-band approach is defensible — a Terraform `data` source cannot mutate the cluster it reads,
and `install.sh:3166-3169` says so. Three things make it a friction point anyway:

- **The calls are unconditional.** At `install.sh:3170-3172` all three run with no
  `if create_cluster == false` guard. Each is a no-op on a not-yet-created cluster only because its own
  `gcloud describe` returns empty. The installer *has* such a gate and uses it 30 lines later —
  `install.sh:3204` guards the OTel scope on `TFVARS_CREATE_CLUSTER`, with the rationale at `:3201-3203`:
  *"silently changing the telemetry scope of a cluster somebody else made is not an install's call."*
  That reasoning applies verbatim to these three and is not applied.
- **Calico fires on the common case.** `install.sh:1690` skips only when the cluster already runs
  Dataplane V2 (`:1698-1700`) or already has the addon (`:1706-1708`). Its own comment at `:1685-1686`
  says a cluster with neither *"is GKE Standard's default shape"*.
- **A reassuring comment is wrong.** `install.sh:1687` says *"Terraform-created clusters always have
  Dataplane V2"*. That guarantee comes from this repo's own module hardcoding
  `datapath_provider = "ADVANCED_DATAPATH"` (`terraform/modules/gke-cluster/main.tf:199`). A cluster
  created by any *other* Terraform does not have it and takes the Calico branch — confirmed on a
  Terraform-created Standard cluster in these runs.

Also note the sequencing: "node pools may be recreated" (`:1728`) is a `print_info`. The function's only
`print_warning` is the FQDN caveat at `:1737`, printed **after** the change is made.

## The changes are one-way

`uninstall.sh` contains **zero** occurrences of `--enable-network-policy`, `--update-addons`,
`--database-encryption-key`, `--workload-pool`, or any `gcloud container clusters update` call.

Measured on a Standard cluster after a clean uninstall (exit 0, *"All kube-agents infrastructure elements
have been safely removed"*):

| Change made by install | State after uninstall | Reverted? |
| --- | --- | --- |
| Calico NetworkPolicy | `networkPolicy.enabled = True`, `provider = CALICO` | **No** |
| CMEK database encryption | `databaseEncryption.state = ALL_OBJECTS_ENCRYPTION_ENABLED` | **No** |
| `gvisor-pool` node pool | removed | Yes |

The node pool reverts because Terraform owns it; the cluster-level settings do not, because `gcloud`
applied them outside Terraform. So the install's out-of-band changes are exactly the ones the uninstall
cannot undo.

The closing message (`uninstall.sh:509`) discloses only what it keeps *by design* — the KMS key rings and
the state bucket. It does not mention Calico, CMEK, the Workload Identity pool or a `GKE_METADATA`
migration.

**A failed install leaves them too.** One install failed at the Helm stage and the cluster was still left
with Calico on, CMEK on, and the extra node pool. Trying the installer is not a reversible experiment at
the cluster-configuration level.

## Documentation is incomplete on this

`INSTALL.md:64-66` says "three things stay outside Terraform" but omits the Workload Identity pool, the
`GKE_METADATA` migration and the NetworkPolicy change — while counting two items (the OTel scope, the
GitHub PEM import) that are not adopted-cluster mutations at all. Its CMEK mention at `:228` is framed as
a security feature rather than a warning.

Separately, `SKIP_CERT_MANAGER=true` — the opt-out for the cert-manager install, which is the other change
users notice — is documented only in `scripts/installer/README.md:116`, not in `INSTALL.md`.

## Suggested fix

- Enumerate the pending existing-cluster mutations in the step-11 pre-flight summary, and require explicit
  opt-in for the two that can restart workloads (changes 3 and 4).
- Gate the three calls at `install.sh:3170-3172` on `TFVARS_CREATE_CLUSTER`, matching the precedent at
  `install.sh:3204`.
- Either revert what the install enabled, or list the non-reverted cluster changes in the uninstall summary
  at `uninstall.sh:509`.
- Fix the misleading comment at `install.sh:1687`.
- Correct `INSTALL.md:64-66`, and document `SKIP_CERT_MANAGER` there.

## Workaround for users today

No configuration avoids changes 1–4 wholesale. Available mitigations:

```bash
# Will Calico be enabled? (empty/False on both = yes)
gcloud container clusters describe <CLUSTER> --location <REGION> \
  --format="value(networkConfig.datapathProvider,networkPolicy.enabled)"

# Will any node pool be recreated? (anything not GKE_METADATA = yes)
gcloud container node-pools list --cluster <CLUSTER> --location <REGION> \
  --format="table(name,config.workloadMetadataConfig.mode)"
```

`ALLOW_UNENCRYPTED_SECRETS=true` skips CMEK (`install.sh:1605`), `SKIP_CERT_MANAGER=true` skips
cert-manager, `--gvisor=false` avoids the extra node pool. `--dry-run` is safe to preview with — verified,
the dry-run branch at `install.sh:3119` exits at `:3142`, 28 lines before the first mutation call.

## Related

- #521 — preflighting whether the cluster can run the agent (complementary; this issue is about disclosing
  and reverting what the install changes).
- #1286 — the `GKE_METADATA` migration (change 3) timing out on larger pools.

## Verification

Four installs onto pre-existing clusters (2 Autopilot, 2 Standard) on `0.4.0`. CMEK observed on all four;
Calico observed on both Standard clusters, with `gcloud container clusters describe` before and after
confirming `datapathProvider`/`networkPolicy.enabled` empty beforehand and `enabled = True,
provider = CALICO` afterwards. `gvisor-pool` created on both Standard clusters and removed on uninstall.

Not observed: change 3 (all four clusters already used `GKE_METADATA`); node-pool recreation from the
Calico change (both Standard clusters kept their pools); behaviour on a cluster with pre-existing
cert-manager. The interactive installer was not exercised — its prompts may disclose more than
`--non-interactive` does.

sergiu1spiridon:
## Summary

This PR addresses issue #1293 (Issue F-4) by eliminating silent, unannounced mutations made to live GKE clusters when installing `kube-agents` onto pre-existing clusters (`TFVARS_CREATE_CLUSTER=false`). Step 11 (`Pre-Flight Configuration Summary`) of `install.sh` now explicitly enumerates all five cluster-level mutations (CMEK database encryption, Workload Identity pool, node pool `GKE_METADATA` migration, NetworkPolicy enforcement, and gVisor sandbox pool). Node-recreating mutations (`GKE_METADATA` migration and Calico NetworkPolicy enablement) now require explicit opt-in via `--migrate-node-pools` and `--enable-network-policy` flags, environment variables, or interactive confirmation prompts defaulting to "No". In addition, `uninstall.sh` now discloses which cluster-level settings were permanently retained, and out-of-band mutation helpers in Step 12 are explicitly gated on `TFVARS_CREATE_CLUSTER=false`.

## Why This Change

When adopting an existing GKE cluster, the installer previously modified live cluster configurations out-of-band prior to the Terraform run without disclosing them in the pre-flight summary or asking for confirmation. Two of these modifications (migrating node pool metadata endpoints to `GKE_METADATA` and enabling Calico network policy enforcement on Standard clusters lacking Dataplane V2) recreate node instances and roll workloads without warning. Furthermore, database encryption (CMEK) and Workload Identity pool enablement cannot be undone after installation, leaving live clusters permanently altered even after running `uninstall.sh`.

## What Changed

- `install.sh`:
  - Added CLI flags `--migrate-node-pools` / `--migrate-node-pools=false` and `--enable-network-policy` / `--enable-network-policy=false`, with inheritance from `install.env`.
  - Added `validate_existing_cluster_opt_in_flags` to strictly validate `--migrate-node-pools` and `--enable-network-policy` against `^(true|false)$` (loud like `--gvisor` and `--enable-web-ui`), rejecting invalid values and typos with a descriptive error.
  - Added interactive prompt function `prompt_existing_cluster_opt_ins` that defaults to "n" before displaying the pre-flight summary, explicitly informing operators that declining terminates the install.
  - Added `summarize_existing_cluster_mutations` to inspect cluster and node pool state via `gcloud` and print explicit disclosures for all 5 mutations in Step 11 (`Pre-Flight Configuration Summary`), reporting `Refused` and preflight abort when node pool migration or NetworkPolicy opt-in is missing, or `Skipped (could not query cluster network policy state)` when cluster describe fails.
  - Added `check_existing_cluster_node_pools_preflight` to abort before confirmation and before any cluster changes if legacy node pools exist and opt-in was not provided (`REFUSED_MISSING_NODE_POOL_MIGRATION`).
  - Added `check_existing_cluster_network_policy_preflight` to abort before confirmation and before any cluster changes if NetworkPolicy enforcement is not satisfied and opt-in was not provided (`REFUSED_MISSING_NETWORK_POLICY`), or if the cluster cannot be queried (`FAILED_PREFLIGHT_CLUSTER_UNREADABLE`).
  - Updated `ensure_existing_cluster_workload_identity` to refuse (`return 1`) when legacy pools exist without opt-in, preventing execution on an unmigrated identity.
  - Consolidated NetworkPolicy probes into a single describe call (`status,networkConfig.datapathProvider,networkPolicy.enabled`), safely handling unreadable clusters and refusing mutations on unread state in `ensure_existing_cluster_network_policy`.
  - In dry-run mode, skipped `terraform plan` when NetworkPolicy or node pools are unsatisfied, printing the exact `gcloud` remediation commands in order.
  - Gated Step 12 calls to `ensure_existing_cluster_cmek`, `ensure_existing_cluster_workload_identity`, and `ensure_existing_cluster_network_policy` strictly on `[ "${TFVARS_CREATE_CLUSTER:-true}" = "false" ]`.
  - Updated `--help` text to document that `--enable-network-policy` and `--migrate-node-pools` are required on existing clusters with legacy configurations, else the install aborts.
  - Corrected misleading comment regarding Terraform-provisioned clusters and Dataplane V2.
- `uninstall.sh`:
  - Added Step 5 post-cleanup disclosure detailing retained cluster-level settings (CMEK, Workload Identity, migrated node pools, NetworkPolicy).
- Documentation:
  - `INSTALL.md`: Corrected lines 64–74 regarding existing cluster mutation behaviors, documented `MIGRATE_NODE_POOLS` and `ENABLE_NETWORK_POLICY` requirements (clarifying fatal preflight aborts when opt-ins are missing on legacy clusters), and clarified CMEK permanence.
  - `scripts/installer/README.md`: Added "Cluster adoption and component toggles" section documenting `MIGRATE_NODE_POOLS`, `ENABLE_NETWORK_POLICY` (and that missing either opt-in aborts the install), `SKIP_CERT_MANAGER`, and `ALLOW_UNENCRYPTED_SECRETS`.
  - `docs/site/src/content/docs/install/quickstart-gke.mdx`: Updated adoption documentation and Common Flags reference table to clarify `--migrate-node-pools`, `--enable-network-policy`, and dry-run plan skip with remediation commands.
  - `docs/site/src/content/docs/install/prerequisites.md`: Updated cluster requirements table and dry-run explanation to document `--migrate-node-pools`, `--enable-network-policy`, and the dry-run plan skip.
  - `install.env.example`: Added Section 9 documenting adoption opt-ins, clarifying that omitting `ENABLE_NETWORK_POLICY` or `MIGRATE_NODE_POOLS` aborts.
  - `.agents/skills/install-kube-agents/SKILL.md`: Added `--migrate-node-pools` and `--enable-network-policy` flags to table, clarifying fatal abort on missing opt-ins.
- Automated tests:
  - `tests/test_install_script.py`: Added comprehensive unit tests covering flag parsing, `validate_existing_cluster_opt_in_flags` (typos, empty values, valid booleans, bare flags), `install.env` inheritance, `summarize_existing_cluster_mutations` state rendering (including `Refused` and unreadable output), unreadable cluster preflight handling, and opt-in skipping/execution for node metadata and network policy.
  - `tests/test_uninstall_script.py`: Added `UninstallSummaryDisclosureTest` verifying retained settings output.

## Context

Closes #1293

## Testing

- `python3 -m unittest tests/test_install_script.py tests/test_uninstall_script.py`: 189 tests passed.
- `make docs-check`: Passed (relative links clean, terminology clean, doc map inventory covered, char budget 38.0k under limit).
- `npx prettier --check`: Passed across modified docs.

### Live validation

Exercised against real GKE clusters in GCP project `sergiuspiridon-gkedemos`:
1. **GKE Standard Cluster (`standard-agent-test`, us-east4-a)**:
   - Evaluated on a Standard cluster whose node pools were already using `GKE_METADATA`, testing NetworkPolicy preflight refusal.
   - **Without opt-in flags:**
     ```bash
     ALLOW_UNENCRYPTED_SECRETS=true TFVARS_CREATE_CLUSTER=false ./install.sh --dry-run --non-interactive \
       --project-id=sergiuspiridon-gkedemos --cluster-name=standard-agent-test --region=us-east4-a
     ```
     Observed Step 11 Pre-Flight Configuration Summary:
     ```
       • Existing Cluster Mutations (Adoption):
         • CMEK Database Encryption: Skipped (ALLOW_UNENCRYPTED_SECRETS=true)
         • Workload Identity Pool: Already enabled (sergiuspiridon-gkedemos.svc.id.goog)
         • Node Pool Metadata: All node pools use GKE_METADATA
         • NetworkPolicy Enforcement: Refused (opt-in not provided; pass --enable-network-policy; install will abort)
         • gVisor Sandbox Node Pool: Will create 'gvisor-pool' (1 e2-standard-4 per zone; new billable capacity)
     ```
     Observed dry-run skips `terraform plan` preview because NetworkPolicy is unsatisfied:
     ```
     ⚠ Dry-run: skipping terraform plan because existing cluster 'standard-agent-test' enforces no NetworkPolicy (postcondition would fail).
     ℹ A real run will abort unless authorized with --enable-network-policy or ENABLE_NETWORK_POLICY=true.
     ℹ To remediate manually beforehand, run these two commands in this order:
       gcloud container clusters update standard-agent-test --location us-east4-a --project sergiuspiridon-gkedemos --update-addons=NetworkPolicy=ENABLED
       gcloud container clusters update standard-agent-test --location us-east4-a --project sergiuspiridon-gkedemos --enable-network-policy
     ✓ Dry-run execution complete! Configuration generated without touching cloud resources.
     ```
     When executed without `--dry-run`, Step 11 preflight validation aborted before confirmation checkpoint:
     ```
     ✗ Existing cluster 'standard-agent-test' enforces no NetworkPolicy (neither Dataplane V2 nor legacy Calico).
     ℹ kube-agents requires NetworkPolicy enforcement to isolate agent execution sandboxes.
     ℹ The Terraform apply will refuse an existing cluster without Dataplane V2 or NetworkPolicy enforcement.
     ℹ Enabling Calico may recreate nodes and restart workloads. Because explicit opt-in was not granted, provisioning cannot proceed.
     ℹ Aborting before making any cluster changes. Pass --enable-network-policy or set ENABLE_NETWORK_POLICY=true to authorize.
     ```
   - **With opt-in flags:**
     ```bash
     ALLOW_UNENCRYPTED_SECRETS=true TFVARS_CREATE_CLUSTER=false ./install.sh --dry-run --non-interactive \
       --project-id=sergiuspiridon-gkedemos --cluster-name=standard-agent-test --region=us-east4-a \
       --migrate-node-pools --enable-network-policy
     ```
     Observed Step 11 Pre-Flight Configuration Summary:
     ```
       • Existing Cluster Mutations (Adoption):
         • CMEK Database Encryption: Skipped (ALLOW_UNENCRYPTED_SECRETS=true)
         • Workload Identity Pool: Already enabled (sergiuspiridon-gkedemos.svc.id.goog)
         • Node Pool Metadata: All node pools use GKE_METADATA
         • NetworkPolicy Enforcement: Will enable legacy Calico addon & enforcement (may recreate nodes, restart workloads)
         • gVisor Sandbox Node Pool: Will create 'gvisor-pool' (1 e2-standard-4 per zone; new billable capacity)
     ```
     Observed dry-run output with Calico opt-in:
     ```
     ℹ Dry-run: skipping terraform plan because Calico has not yet been applied to the live cluster (a real run enables Calico prior to apply).
     ✓ Dry-run execution complete! Configuration generated without touching cloud resources.
     ```

2. **GKE Autopilot Cluster (`autopilot-cluster-1`, us-central1)**:
   - Command:
     ```bash
     ALLOW_UNENCRYPTED_SECRETS=true TFVARS_CREATE_CLUSTER=false ./install.sh --dry-run --non-interactive \
       --project-id=sergiuspiridon-gkedemos --cluster-name=autopilot-cluster-1 --region=us-central1
     ```
   - Observed Step 11 Pre-Flight Configuration Summary:
     ```
       • Existing Cluster Mutations (Adoption):
         • CMEK Database Encryption: Skipped (ALLOW_UNENCRYPTED_SECRETS=true)
         • Workload Identity Pool: Native (GKE Autopilot)
         • Node Pool Metadata: Managed (GKE Autopilot)
         • NetworkPolicy Enforcement: Built-in (Dataplane V2)
         • gVisor Sandbox Node Pool: None (not requested or native on Autopilot)
     ```
   - Because Autopilot has built-in Dataplane V2, `is_existing_cluster_network_policy_satisfied` is satisfied: `terraform plan` executed successfully: 33 resources to add, 0 to change, 0 to destroy.

3. **Pristine GKE Standard Cluster (`pristine-standard-test`, us-central1-c)**:
   - Cluster provisioned specifically with legacy metadata (`GCE_METADATA`), disabled NetworkPolicy, and no CMEK (`DECRYPTED`), exercising the full node pool migration and NetworkPolicy refusal paths against the head commit.
   - **Without opt-in flags:**
     ```bash
     ALLOW_UNENCRYPTED_SECRETS=true TFVARS_CREATE_CLUSTER=false ./install.sh --dry-run --non-interactive \
       --project-id=sergiuspiridon-gkedemos --cluster-name=pristine-standard-test --region=us-central1-c
     ```
     Observed Step 11 Pre-Flight Summary:
     ```
       • Existing Cluster Mutations (Adoption):
         • CMEK Database Encryption: Skipped (ALLOW_UNENCRYPTED_SECRETS=true)
         • Workload Identity Pool: Will enable sergiuspiridon-gkedemos.svc.id.goog on control plane (non-revertible)
         • Node Pool Metadata Migration: Refused for 'default-pool' (opt-in not provided; pass --migrate-node-pools; install will abort)
         • NetworkPolicy Enforcement: Refused (opt-in not provided; pass --enable-network-policy; install will abort)
         • gVisor Sandbox Node Pool: Will create 'gvisor-pool' (1 e2-standard-4 per zone; new billable capacity)
     ```
     Dry-run skipped `terraform plan` with warning and remediation commands; non-dry-run aborted with preflight exit code 1 (`REFUSED_MISSING_NODE_POOL_MIGRATION`).
   - **With opt-in flags:**
     ```bash
     ALLOW_UNENCRYPTED_SECRETS=true TFVARS_CREATE_CLUSTER=false ./install.sh --dry-run --non-interactive \
       --project-id=sergiuspiridon-gkedemos --cluster-name=pristine-standard-test --region=us-central1-c \
       --migrate-node-pools --enable-network-policy
     ```
     Observed Step 11 Pre-Flight Summary:
     ```
       • Existing Cluster Mutations (Adoption):
         • CMEK Database Encryption: Skipped (ALLOW_UNENCRYPTED_SECRETS=true)
         • Workload Identity Pool: Will enable sergiuspiridon-gkedemos.svc.id.goog on control plane (non-revertible)
         • Node Pool Metadata Migration: Will migrate 'default-pool' to GKE_METADATA (recreates nodes, restarts workloads)
         • NetworkPolicy Enforcement: Will enable legacy Calico addon & enforcement (may recreate nodes, restart workloads)
         • gVisor Sandbox Node Pool: Will create 'gvisor-pool' (1 e2-standard-4 per zone; new billable capacity)
     ```
     Observed `Dry-run: skipping terraform plan because Calico has not yet been applied to the live cluster (a real run enables Calico prior to apply).`
   - Cleaned up temporary test cluster `pristine-standard-test` after execution.

## Self-Review

Merged findings from clean-context review passes:
- **`review-adversarial`** (run in clean subagent `921555cc-790f-45dd-961b-551d7b2d2138` against `refs/kube-agents-base/main...HEAD`):
  - *Looked for:* Shell pipeline errors under `set -eo pipefail`, subtle edge cases in `gcloud` cluster describe outputs, regex match vulnerabilities in operations lists, non-interactive bypasses, and state-leakage between runs.
  - *Finding 1 (Active operation pipeline under set -eo pipefail):* `gcloud container operations list ... | head -n1` can trigger SIGPIPE (exit code 141) under `pipefail`.
    - *Disposition: Fixed.* Wrapped `gcloud` query in `{ ... || true; } | head -n1`.
  - *Finding 2 (Substring matching in operation targetLink):* `--filter="targetLink:$cluster_name AND status=RUNNING"` matches substrings (e.g. `cluster-1` matching `my-cluster-1`).
    - *Disposition: Fixed.* Updated filter to exact regex matching `--filter="targetLink ~ /clusters/${cluster_name}$ AND status=RUNNING"`.
  - *Finding 3 (CMEK describe fallback):* In `summarize_existing_cluster_mutations`, an empty output from `databaseEncryption.state` query would fall through to "Will enable Cloud KMS".
    - *Disposition: Fixed.* Added explicit check `elif [ -z "$enc_state" ]` reporting skipped/unable to query instead of false positive announcement.
- **`review-docs-drift`** (run in clean subagent `f7b8eeb5-d313-457c-97d3-053a428fcf73` against `refs/kube-agents-base/main...HEAD`):
  - *Looked for:* Outdated documentation across `docs/site/`, `INSTALL.md`, skill descriptions, and install configuration templates regarding cluster adoption, CLI options, and out-of-band mutations.
  - *Finding 1 (Quickstart guide):* `docs/site/src/content/docs/install/quickstart-gke.mdx` claimed `install.sh` enables Workload Identity and Calico out-of-band without mentioning required opt-in flags.
    - *Disposition: Fixed.* Updated narrative and added `--migrate-node-pools` and `--enable-network-policy` to Common Flags table.
  - *Finding 2 (Skill definition flags table):* `.agents/skills/install-kube-agents/SKILL.md` was missing the two new CLI flags.
    - *Disposition: Fixed.* Added `--migrate-node-pools` and `--enable-network-policy` to CLI flag table.
  - *Finding 3 (Installer template):* `install.env.example` lacked documentation of `MIGRATE_NODE_POOLS` and `ENABLE_NETWORK_POLICY`.
    - *Disposition: Fixed.* Added Section 8/9 documenting pre-existing cluster adoption variables.
  - *Finding 4 (Installer README):* `scripts/installer/README.md` omitted `ALLOW_UNENCRYPTED_SECRETS`.
    - *Disposition: Fixed.* Documented `ALLOW_UNENCRYPTED_SECRETS` in the cluster adoption section.
  - *Finding 5 (Prerequisites cluster requirements table):* `docs/site/src/content/docs/install/prerequisites.md` still stated unconditional node-pool migration and Calico enablement, and that dry-run stops at plan-time refusal.
    - *Disposition: Fixed.* Updated table rows to document `--migrate-node-pools`, `--enable-network-policy`, and the dry-run plan skip.
  - *Finding 6 (NetworkPolicy opt-in consequence consistency):* Surfaces previously described NetworkPolicy enablement as 'skipped' when omitted, whereas the preflight check exits 1 (`REFUSED_MISSING_NETWORK_POLICY`).
    - *Disposition: Fixed.* Updated `install.sh` help text and Step 11 summary (`Refused (opt-in not provided; pass --enable-network-policy; install will abort)`), `INSTALL.md`, `scripts/installer/README.md`, `install.env.example`, `quickstart-gke.mdx`, and `.agents/skills/install-kube-agents/SKILL.md` to document that omitting `--enable-network-policy` on such clusters aborts before making changes.
- **Review feedback:**
  - *Finding (Boolean flag validation):* `--migrate-node-pools` and `--enable-network-policy` accepted raw values with no validation, silently treating typos (like `=ture`) as false and leading to confusing error messages.
    - *Disposition: Fixed.* Added `validate_existing_cluster_opt_in_flags` (sourced in `main()`) strictly validating both flags and their environment variables against `^(true|false)$` (loud like `--gvisor`), along with unit tests covering typo rejection, valid booleans, and bare flag invocations.
  - *Finding (Symmetrical node pool migration refusal):* Omitting `--migrate-node-pools` previously printed a warning and carried on, leading to a "successful" install where the agent ran under the wrong identity.
    - *Disposition: Fixed.* Added `check_existing_cluster_node_pools_preflight` refusing with exit code 1 (`REFUSED_MISSING_NODE_POOL_MIGRATION`) unless opted in, updated `ensure_existing_cluster_workload_identity` to return 1, updated summary to report `Refused`, and updated documentation and unit tests.
  - *Finding (Prompt default consequence):* Defaulting to "No" on interactive mutation prompts terminated the install, but the prompt text did not say so.
    - *Disposition: Fixed.* Updated prompts in both `ensure_existing_cluster_network_policy` and `prompt_existing_cluster_opt_ins` to state `Declining ends the install (kube-agents requires NetworkPolicy enforcement/Workload Identity).`
  - *Finding (Unreadable cluster vs missing NetworkPolicy):* Describe errors swallowed with `|| echo ""` made an unreadable cluster indistinguishable from missing NetworkPolicy, falsely asserting absence and attempting mutations on unread state.
    - *Disposition: Fixed.* Consolidated queries to a single `csv[no-heading](status,networkConfig.datapathProvider,networkPolicy.enabled)` call; when `status` is empty, returns code 2, preflight exits 1 with `FAILED_PREFLIGHT_CLUSTER_UNREADABLE` without false claims, `ensure_existing_cluster_network_policy` returns 1 immediately, and summary reports `Skipped (could not query cluster network policy state)`.
  - *Finding (--dry-run plan skip and remediation commands):* `quickstart-gke.mdx` promised a full `terraform plan` preview when ADC exists, but code skips the plan on adopted clusters lacking NetworkPolicy without printing remediation commands.
    - *Disposition: Fixed.* Updated `quickstart-gke.mdx` to state that the plan is skipped with remediation commands, and updated `install.sh` dry-run logic to output the exact `gcloud` remediation commands in order.

## Risk & Rollout

**Breaking change for automated adoption of existing GKE Standard clusters:**
- On `main`, running `./install.sh --non-interactive` against an adopted Standard cluster with legacy node pools or lacking NetworkPolicy enforcement would silently and automatically execute cluster mutations (`gcloud container node-pools update ... --workload-metadata=GKE_METADATA` and `gcloud container clusters update ... --update-addons=NetworkPolicy=ENABLED --enable-network-policy`), recreating nodes and restarting workloads without explicit consent.
- With this change, unattended adoption of such clusters now aborts with exit code 1 (`REFUSED_MISSING_NODE_POOL_MIGRATION` or `REFUSED_MISSING_NETWORK_POLICY`). Automation scripts adopting existing Standard clusters must be updated to explicitly pass `--migrate-node-pools` and/or `--enable-network-policy` (or set `MIGRATE_NODE_POOLS=true` / `ENABLE_NETWORK_POLICY=true`) to authorize these mutations.
- Newly created clusters (`TFVARS_CREATE_CLUSTER=true`) and existing Autopilot clusters are unaffected (Autopilot natively manages `GKE_METADATA` and Dataplane V2).

Revertible by reverting git commits.

Generated with the help of Antigravity (Gemini).


install.sh
```diff
@@ -480,6 +482,11 @@ Flags for AI Agents & Automation:
                                 currently platform-agent-chat-events)
   --google-chat-mode=MODE       Google Chat output mode: default | debug
                                 (default: DEFAULT_GOOGLE_CHAT_MODE, currently default)
+  --migrate-node-pools          Opt in to migrating legacy node pools to GKE_METADATA on an existing
+                                cluster (recreates nodes and restarts workloads)
+  --enable-network-policy       Opt in to enabling legacy Calico NetworkPolicy addon and enforcement
+                                on an existing GKE Standard cluster without Dataplane V2 (may recreate
+                                nodes and restart workloads)
```
mplakhtiy: **HIGH — every user-facing surface says the NetworkPolicy opt-in is "skipped"; the code exits 1.**

`bcecfa36` changed the behaviour from "skip the mutation" to "refuse the install", and no document, no `--help` text, and no agent-facing table was revisited afterwards. `check_existing_cluster_network_policy_preflight` (install.sh:1961-1978) prints `Existing cluster '<name>' enforces no NetworkPolicy (neither Dataplane V2 nor legacy Calico).`, writes `write_json_report "REFUSED_MISSING_NETWORK_POLICY"`, and `exit 1`. `ensure_existing_cluster_network_policy` (install.sh:1848-1853) `return 1`s on the same condition. That is the right behaviour — nothing a user or an agent reads says it:

- **this `--help` block (install.sh:487-489)** — "Opt in to enabling…", silent that omitting it is fatal.
- **`INSTALL.md:273`** — "Pass `--migrate-node-pools` … and `--enable-network-policy` … to authorize these mutations; otherwise they are skipped."
- **`scripts/installer/README.md:128`** — "Without opt-in, enablement is skipped."
- **`install.env.example:200`** — "Skipped when false."
- **`docs/site/src/content/docs/install/quickstart-gke.mdx:116`** — "authorizes enabling the legacy Calico NetworkPolicy addon…", silent on the consequence of omitting it.
- **`.agents/skills/install-kube-agents/SKILL.md:163`** — "Authorize enabling legacy Calico NetworkPolicy addon and node enforcement…", silent.

The `SKILL.md` row is the sharpest, because that table is how an agent decides which flags to pass. Failure scenario: an agent adopts a GKE Standard cluster in its default shape (no Dataplane V2, no Calico addon), reads "Authorize" in the skill table as describing an optional flag, and runs `./install.sh --non-interactive` without `--enable-network-policy`. The run exits 1, and no surface it could have consulted beforehand predicted that.

The Step 11 summary itself carries the same stale word — install.sh:2055 prints `NetworkPolicy Enforcement: Skipped (opt-in not provided; pass --enable-network-policy; Terraform apply will fail)`, which also describes pre-`bcecfa36` behaviour: the apply is never reached, because the preflight exits first.

install.sh
```diff
@@ -532,6 +539,10 @@ parse_args() {
       --allowed-users=*) PARAM_ALLOWED_USERS="${1#*=}"; shift ;;
       --chat-topic-name=*) PARAM_CHAT_TOPIC_NAME="${1#*=}"; shift ;;
       --google-chat-mode=*) PARAM_GOOGLE_CHAT_MODE="${1#*=}"; shift ;;
+      --migrate-node-pools=*) PARAM_MIGRATE_NODE_POOLS="${1#*=}"; shift ;;
+      --migrate-node-pools) PARAM_MIGRATE_NODE_POOLS="true"; shift ;;
+      --enable-network-policy=*) PARAM_ENABLE_NETWORK_POLICY="${1#*=}"; shift ;;
+      --enable-network-policy) PARAM_ENABLE_NETWORK_POLICY="true"; shift ;;
```
mplakhtiy: **MEDIUM — the two new boolean flags are not validated, against this file's own stated convention.**

`--migrate-node-pools=*` and `--enable-network-policy=*` store the raw value with no check. Anything `is_truthy` (`scripts/installer/installer_common.sh:220`) does not match — it accepts only `true`/`yes`/`y`/`1`/`on`, case-insensitively — becomes false silently. `--gvisor` does the opposite at install.sh:3099-3101 (`print_error "--gvisor must be either true or false."; exit 1`), and the comment at install.sh:2871 states it as a rule: *"Loud like --gvisor and --enable-web-ui, not lenient like the chat booleans."*

Failure scenario: `./install.sh --non-interactive --enable-network-policy=ture` against an adopted Standard cluster. The value parses, is stored, and is falsy. The run then aborts with `Pass --enable-network-policy or set ENABLE_NETWORK_POLICY=true to authorize` — telling the operator to pass the flag they just passed, with nothing pointing at the typo.

Leniency is defensible for a chat toggle whose misreading costs a feature. It is not defensible for a flag whose misreading now terminates the run. Two lines, and the precedent to copy is in this same file.

mplakhtiy (commented):
The design is sound. All five fixes #1293 asked for are implemented, and implemented correctly: the five cluster-level mutations are enumerated in the Step 11 summary, the two node-recreating ones require explicit opt-in, the Step 12 call sites are properly gated on `TFVARS_CREATE_CLUSTER=false`, and the reordering (NetworkPolicy -> Workload Identity -> CMEK) is right, because NetworkPolicy is the one that can still abort and the other two are permanent control-plane updates. Commit `bcecfa36` genuinely closes the bot's earlier inline High about mutating a cluster and then running into an apply the code already predicts will fail.

The bot's still-open inline High at `install.sh:1842` **is** addressed: `check_existing_cluster_network_policy_preflight` runs at `install.sh:3491`, before the "Proceed? (Y/n)" checkpoint and before any of the three `ensure_existing_cluster_*` mutations, so it aborts without touching the cluster — that thread can be replied to and resolved.

**Merge conflict, and it is why the presubmit is red.** `INSTALL.md` conflicts with `main`: `b09bf29b` (#1353, "state what a customer-supplied cluster must already have") rewrote the same paragraph and repointed it at the new canonical page `docs/site/src/content/docs/install/prerequisites.md#cluster-requirements`. Resolving by keeping this branch's side would delete that link and re-fork ownership of a topic `main` just consolidated, so the resolution should keep `main`'s structure and the link and fold the opt-in facts into it. The red `pull-kube-agents-smoke-test` is the same conflict, not an eval regression: the job died at `git merge --no-ff bcecfa36` with `CONFLICT (content): Merge conflict in INSTALL.md` and never ran a single eval. No `/retest` is needed and no `presubmit-gate` issue should be filed — the rebase fixes it.

**HIGH — `docs/site/src/content/docs/install/prerequisites.md` (on the base branch, untouched by this PR).** `b09bf29b` landed on `main` after this branch diverged and created the "Cluster requirements" section, now the canonical home for what an adopted cluster must already have. It is not in this diff, and `make docs-check` only checks generated regions, links, terminology, map coverage and the context budget — it cannot see prose drift — so three statements merge stale:

- Node-pool metadata row: "`install.sh` migrates each pool that is not, **which recreates that pool's nodes**, including pools running workloads unrelated to kube-agents." After this PR it migrates nothing without `--migrate-node-pools`.
- NetworkPolicy row: "`install.sh` enables the Calico addon and then enforcement, which can recreate node pools." After this PR it refuses the install unless `--enable-network-policy`.
- The `--dry-run` sentence: "a dry run against a cluster still missing the Workload Identity pool or NetworkPolicy enforcement stops at the plan-time refusal." After this PR the dry run *skips the plan* (`install.sh:3465-3471`) and never reaches the postcondition. This is the most misleading of the three, because that sentence is the documented way to preview an adoption and the module's postcondition `error_message` (`terraform/modules/gke-cluster/main.tf:275`) carries the two exact `gcloud` remediation commands. Skipping the plan withholds them.

The author cannot have seen this before rebasing; it is fixable in the same rebase as the conflict.

install.sh
```diff
@@ -1748,36 +1759,65 @@ ensure_existing_cluster_workload_identity() {
...
+          PARAM_MIGRATE_NODE_POOLS="true"
+        else
+          PARAM_MIGRATE_NODE_POOLS="false"
+        fi
+      fi
+    fi
+
+    if ! is_truthy "${PARAM_MIGRATE_NODE_POOLS:-${MIGRATE_NODE_POOLS:-false}}"; then
+      print_warning "Skipping node pool migration for '${legacy_pools[*]}' (explicit opt-in via --migrate-node-pools or MIGRATE_NODE_POOLS=true was not provided)."
+      print_info "Pods on these pools will continue using the legacy GCE metadata server and cannot use Workload Identity."
```
mplakhtiy: **MEDIUM — declining the node-pool migration still produces a "successful" install with the agent on the wrong identity.**

The NetworkPolicy path now refuses outright. This path warns and carries on, and nothing downstream catches it. The `gke-cluster` module's Workload Identity postcondition (`terraform/modules/gke-cluster/main.tf:261-263`) tests `try(self.workload_identity_config[0].workload_pool, "")` only — its `error_message` *mentions* migrating node pools off the legacy metadata server, but the `condition` never reads them. So the apply passes.

Failure scenario: adopt a Standard cluster whose only node pool is on `GCE_METADATA`, with `--gvisor=false` so Terraform creates no `gvisor-pool` on `GKE_METADATA` for the agent to land on. The operator declines the migration prompt (or omits the flag non-interactively). The install runs to completion and reports success; the agent pod authenticates as the node's default compute service account instead of through its GSA binding. That is verbatim the outcome the base branch's own prerequisites page names — "the install reports success while the pods use the wrong identity" — and it is one of the outcomes #1293 was filed about.

Within one PR, two mutations of the same class get opposite treatment: one aborts, one warns. If the asymmetry is deliberate — Terraform enforces NetworkPolicy, nothing enforces node metadata mode — the reasoning belongs in a comment here, because the Terraform gap is the argument for the installer being *stricter* on this one, not laxer. Either refuse symmetrically when the agent would land on a legacy-metadata pool, or record why the warning is enough.

install.sh
```diff
@@ -1822,6 +1884,187 @@ ensure_existing_cluster_network_policy() {
...
+    dp_provider=$(trap - ERR; gcloud container clusters describe "$cluster_name" \
+      --location="$region" --project="$project_id" \
+      --format="value(networkConfig.datapathProvider)" 2>/dev/null || echo "")
+    if [ "$dp_provider" != "ADVANCED_DATAPATH" ]; then
+      legacy_np=$(trap - ERR; gcloud container clusters describe "$cluster_name" \
+        --location="$region" --project="$project_id" \
+        --format="value(networkPolicy.enabled)" 2>/dev/null || echo "")
+      if [ "$legacy_np" != "True" ] && [ "$legacy_np" != "true" ]; then
+        local np_choice=""
+        prompt_read "Existing cluster '$cluster_name' does not enforce NetworkPolicy (kube-agents requires Dataplane V2 or Calico). Enabling Calico may recreate nodes and restart workloads. Authorize enabling Calico NetworkPolicy now? (y/N)" np_choice "n"
```
mplakhtiy: **MEDIUM — the prompt's default answer ends the install, and the prompt does not say so.**

`prompt_read … np_choice "n"` defaults to N, which is the right default for a mutation that can recreate nodes. But on GKE Standard's default shape (no Dataplane V2, no Calico) pressing Enter sets `PARAM_ENABLE_NETWORK_POLICY=false`, the Step 11 summary prints as if the run were continuing, and then `check_existing_cluster_network_policy_preflight` (install.sh:1961, called at install.sh:3491) exits 1 — before the `Proceed with automated GKE cluster & Platform Agent provisioning? (Y/n)` checkpoint at install.sh:3495 ever appears.

So the safe-looking default terminates the install, and the prompt frames the choice as "do this mutation or not" when it is actually "do this mutation or stop". An operator who presses Enter to be cautious gets a failed run and a printed summary of work that will not happen.

Say what N means in the prompt itself — e.g. "…Declining ends the install; kube-agents cannot run without NetworkPolicy enforcement." This is the same class of undisclosed consequence the PR exists to remove.

install.sh
```diff
@@ -1822,6 +1884,187 @@ ensure_existing_cluster_network_policy() {
...
+  dp_provider=$(trap - ERR; gcloud container clusters describe "$cluster_name" \
+    --location="$region" --project="$project_id" \
+    --format="value(networkConfig.datapathProvider)" 2>/dev/null || echo "")
+  [ "$dp_provider" != "ADVANCED_DATAPATH" ] || return 0
+
+  local legacy_np
+  legacy_np=$(trap - ERR; gcloud container clusters describe "$cluster_name" \
+    --location="$region" --project="$project_id" \
+    --format="value(networkPolicy.enabled)" 2>/dev/null || echo "")
+  [ "$legacy_np" = "True" ] || [ "$legacy_np" = "true" ] || return 1
```
mplakhtiy: **MEDIUM — an unreadable cluster is treated as "no NetworkPolicy", so a transient `gcloud` failure now aborts a real install.**

Both probes swallow the failure into an empty string (`|| echo ""`), replacing the pre-PR behaviour of skipping the check when the probe failed. An empty `networkConfig.datapathProvider` and an empty `networkPolicy.enabled` are indistinguishable from "the describe call failed", and both route to `return 1` -> `check_existing_cluster_network_policy_preflight` -> `exit 1`.

Failing closed is the right default. Two things follow from it that are not handled:

1. **The error text asserts a fact it did not establish.** install.sh:1970 prints `Existing cluster '$cluster_name' enforces no NetworkPolicy (neither Dataplane V2 nor legacy Calico).` when the truth may be "could not read cluster `$cluster_name`". `summarize_existing_cluster_mutations` already makes exactly this distinction for CMEK — the `elif [ -z "$enc_state" ]` branch at install.sh:1993-1994 reports "could not query cluster encryption state" rather than claiming the cluster is unencrypted. Same one-line pattern, missing here.

2. **A mutation gets attempted against a cluster whose state was never read.** Failure scenario: an operator passes `--enable-network-policy` on a Dataplane V2 cluster and the `describe` fails transiently (IAM propagation, quota, a flaky network). The preflight passes because opt-in is set; `ensure_existing_cluster_network_policy` then re-probes, also gets empty, falls through, and runs `--update-addons=NetworkPolicy=ENABLED` against a cluster that already has `ADVANCED_DATAPATH`. GKE rejects it, so it fails loudly rather than corrupting anything — but the run reached a cluster-update call on state it never successfully read.

docs/site/src/content/docs/install/quickstart-gke.mdx
```diff
@@ -104,14 +104,16 @@ curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSI
...
+When adopting an existing cluster, four settings stay `gcloud` pre-steps because a Terraform data source cannot express them: CMEK database encryption, Workload Identity pool enablement, node-pool migration to `GKE_METADATA` (requires `--migrate-node-pools`), and legacy Calico NetworkPolicy enforcement (requires `--enable-network-policy`). On freshly created clusters, setting the managed OpenTelemetry collection scope stays a `gcloud` post-step (no provider field exists). The one-shot GitHub App key import into KMS also runs outside Terraform, so the PEM never enters state.
 
 Changing your mind later is the same engine: running the installer with `--menu` (e.g. `./install.sh --menu` or `$HOME/kube-agents/install.sh --menu`) opens a Day-2 control panel whose **Save & Apply** regenerates `terraform.tfvars` and re-applies.
 
 ## Common flags and toggles
 
 - `--dry-run` — always runs `terraform validate`, and adds a full `terraform plan` preview when Application Default Credentials exist; nothing is created, and state stays local. It writes no `install.env`: a dry run provisions nothing, so it has no install to record.
```
mplakhtiy: **MEDIUM — the `--dry-run` bullet still promises a plan the code now skips.**

This line says `--dry-run` "adds a full `terraform plan` preview when Application Default Credentials exist". After `bcecfa36`, install.sh:3465-3471 skips the plan whenever the adopted cluster has no NetworkPolicy enforcement, printing `Dry-run: skipping terraform plan because existing cluster '<name>' enforces no NetworkPolicy (postcondition would fail).` instead. That is precisely the cluster shape a user most wants to preview, and the one #1293 was filed about.

The cost is not just a missing preview: the `gke-cluster` module's NetworkPolicy postcondition `error_message` (`terraform/modules/gke-cluster/main.tf:275`) carries the two exact `gcloud` remediation commands, in the order GKE requires them. Skipping the plan withholds them from the one run whose whole purpose is to find out what is wrong before committing.

Unchanged by this PR, but three lines above the two flags it adds, in a file already open in the diff.

install.sh
sergiu1spiridon: Addressed in 08f5f4d9: added `validate_existing_cluster_opt_in_flags` (called in `main()`) which strictly validates `--migrate-node-pools` and `--enable-network-policy` against `^(true|false)$` (loud like `--gvisor` and `--enable-web-ui`). Added unit tests in `tests/test_install_script.py` covering typo rejection, empty values, valid booleans, and bare flag invocations.

mplakhtiy (approved):

install.sh
sergiu1spiridon: Addressed in d54d2091: made `--migrate-node-pools` refuse symmetrically with `--enable-network-policy`.

- Added `check_existing_cluster_node_pools_preflight` in Step 11 which aborts with exit code 1 (`REFUSED_MISSING_NODE_POOL_MIGRATION`) before making any cluster changes if an adopted cluster has node pools on `GCE_METADATA` and explicit opt-in was not granted.
- Updated `summarize_existing_cluster_mutations` to display `Refused for '<pools>' (opt-in not provided; pass --migrate-node-pools; install will abort)`.
- Updated `ensure_existing_cluster_workload_identity` to return 1 instead of continuing if legacy pools remain unmigrated without opt-in.
- Updated the interactive prompt in `prompt_existing_cluster_opt_ins` to explicitly state that declining ends the install (kube-agents requires Workload Identity).
- In dry-run mode, skipped `terraform plan` preview when legacy node pools exist without opt-in.
- Updated `INSTALL.md`, `scripts/installer/README.md`, `install.env.example`, `quickstart-gke.mdx`, `prerequisites.md`, and `SKILL.md` to document that omitting `--migrate-node-pools` on such clusters aborts before making changes.
- Added unit tests in `tests/test_install_script.py` for `check_existing_cluster_node_pools_preflight`, refusal in `ensure_existing_cluster_workload_identity`, and the Step 11 summary.

install.sh
sergiu1spiridon: Addressed in 08f5f4d9 and d54d2091: aligned all user-facing surfaces (`INSTALL.md`, `scripts/installer/README.md`, `install.env.example`, `quickstart-gke.mdx`, `prerequisites.md`), `install.sh --help`, the interactive prompts, and the Pre-Flight Summary to consistently state that missing NetworkPolicy enforcement or legacy node pool migration causes the install to abort unless explicit opt-in is granted.

install.sh
sergiu1spiridon: Addressed in 3ad9dc1e: updated the prompt in `ensure_existing_cluster_network_policy` (matching `prompt_existing_cluster_opt_ins`) to explicitly state: `Declining ends the install (kube-agents requires NetworkPolicy enforcement). Enable Calico NetworkPolicy now? (y/N)`. Node pool migration prompts were also aligned.

install.sh
sergiu1spiridon: Addressed in 3ad9dc1e:
- Consolidated the describe query in `is_existing_cluster_network_policy_satisfied`, `check_existing_cluster_network_policy_preflight`, `ensure_existing_cluster_network_policy`, and `summarize_existing_cluster_mutations` into a single call querying `csv[no-heading](status,networkConfig.datapathProvider,networkPolicy.enabled)`.
- If `status` is empty, `is_existing_cluster_network_policy_satisfied` returns code 2. `check_existing_cluster_network_policy_preflight` catches this before checking opt-ins, printing `Could not query NetworkPolicy configuration for existing cluster '$cluster_name'` and recording `FAILED_PREFLIGHT_CLUSTER_UNREADABLE` without falsely claiming the cluster lacks NetworkPolicy.
- `ensure_existing_cluster_network_policy` also verifies `status` before executing any mutations and returns 1 immediately if unreadable, preventing mutation attempts against unread cluster state.
- `summarize_existing_cluster_mutations` reports `Skipped (could not query cluster network policy state)`, matching the CMEK pattern.
- Added unit tests in `tests/test_install_script.py` covering unreadable cluster handling across all functions.

docs/site/src/content/docs/install/quickstart-gke.mdx
sergiu1spiridon: Addressed in 3ad9dc1e:
- Updated the `--dry-run` bullet in `docs/site/src/content/docs/install/quickstart-gke.mdx` to clarify that on an adopted cluster lacking NetworkPolicy or Workload Identity, the plan is skipped and the required remediation commands are printed instead.
- Updated `install.sh` dry-run plan skip logic to output the exact `gcloud` remediation commands in order:
  - For NetworkPolicy: `--update-addons=NetworkPolicy=ENABLED` followed by `--enable-network-policy`.
  - For node pools: `gcloud container node-pools update <pool-name> --workload-metadata=GKE_METADATA`.

install.sh
```diff
@@ -1872,6 +1963,267 @@ ensure_existing_cluster_network_policy() {
...
+
+  if ! is_truthy "${PARAM_MIGRATE_NODE_POOLS:-${MIGRATE_NODE_POOLS:-false}}"; then
+    print_error "Existing cluster '$cluster_name' has node pool(s) '${legacy_pools[*]}' using the legacy GCE metadata server."
+    print_info "kube-agents requires Workload Identity (GKE_METADATA) to authenticate agent and operator pods."
+    print_info "Pods on these pools cannot use Workload Identity and would silently authenticate as the node's default compute service account."
+    print_info "Migrating node pools recreates nodes and restarts workloads. Because explicit opt-in was not granted, provisioning cannot proceed."
+    print_info "Aborting before making any cluster changes. Pass --migrate-node-pools or set MIGRATE_NODE_POOLS=true to authorize."
+    write_json_report "REFUSED_MISSING_NODE_POOL_MIGRATION"
+    exit 1
+  fi
```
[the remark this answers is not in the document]
sergiu1spiridon: Addressed: updated the PR description across What Changed, Risk & Rollout, Self-Review, and Live validation:
- Replaced "backwards compatible" and "safe skipping" with an explicit Breaking Change disclosure in Risk & Rollout: existing unattended automation adopting Standard clusters with legacy node pools or lacking NetworkPolicy will now abort with exit code 1 (`REFUSED_MISSING_NODE_POOL_MIGRATION` / `REFUSED_MISSING_NETWORK_POLICY`) unless `--migrate-node-pools` and `--enable-network-policy` are provided.
- Added `check_existing_cluster_node_pools_preflight` and refusal behavior to What Changed and INSTALL.md descriptions.
- Added the skip→refuse transition finding and disposition to Self-Review.
- Aligned the Live validation transcripts to show `Node Pool Metadata Migration: Refused for 'default-pool' (opt-in not provided; pass --migrate-node-pools; install will abort)`.

install.sh
```diff
@@ -1872,6 +1963,267 @@ ensure_existing_cluster_network_policy() {
...
+
+    if [ "${#legacy_pools[@]}" -eq 0 ]; then
+      echo -e "    • ${C_CYAN}Node Pool Metadata:${C_RESET} ${C_GREEN}All node pools use GKE_METADATA${C_RESET}"
+    else
+      if is_truthy "${PARAM_MIGRATE_NODE_POOLS:-${MIGRATE_NODE_POOLS:-false}}"; then
+        echo -e "    • ${C_CYAN}Node Pool Metadata Migration:${C_RESET} ${C_RED}Will migrate${C_RESET} '${legacy_pools[*]}' to GKE_METADATA (${C_RED}recreates nodes, restarts workloads${C_RESET})"
+      else
+        echo -e "    • ${C_CYAN}Node Pool Metadata Migration:${C_RESET} ${C_RED}Refused${C_RESET} for '${legacy_pools[*]}' (opt-in not provided; pass --migrate-node-pools; install will abort)"
+      fi
+    fi
```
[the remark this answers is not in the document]
sergiu1spiridon: Addressed: reconciled the PR description with the head commit by listing `check_existing_cluster_node_pools_preflight` in What Changed, removing all stale "skipping" claims in favor of the fatal preflight refusal, documenting the breaking change for automated adoption in Risk & Rollout, and updating the Live validation transcript to match `install.sh`'s `Refused` output.

PR #1348, opened 2026-09-09
fix(installer): clear ERR trap in tf_state command substitutions

### LINKED ISSUE #1290: bug(installer): ERR trap fires inside a subshell, printing a red "✗ Error encountered at line 629" on every first install onto an existing cluster
## Summary

On the normal path — installing onto a pre-existing GKE cluster — `install.sh` prints a bright red
error banner claiming the install failed, immediately followed by the line reporting that it
succeeded. The install is fine. The banner is a false alarm produced by the `ERR` trap firing inside
a command substitution.

Reproduced on stock `0.4.0` in **6 of 7** captured install attempts.

## What the user sees

```
✗ Error encountered at line 629 (exit code 1): gcloud storage cat "gs://$(tf_state_bucket)/$(tf_state_prefix)/default.tfstate" 2> /dev/null
  ℹ Cluster '<CLUSTER>' already exists and is not in Terraform state; installing onto it (create_cluster = false).
```

The second line is the correct, intended outcome. The banner above it is noise — but it is red,
prefixed `✗`, and says "exit code 1", and it is immediately followed by a long silence (see the
companion issue on discarded `terraform init` output). A user watching this reasonably concludes the
install crashed and hung.

## When it happens

Every **first** install onto a given cluster, because the installer is checking for a Terraform state
object that does not exist yet. It does not appear on a re-run where state already exists — which is
exactly why 1 of the 7 captured attempts (a retry, with state present) lacked it.

## Root cause

`scripts/installer/installer_common.sh:627-629`, in `tf_state_has_cluster()`:

```bash
tf_state_has_cluster() {
  local state
  state=$(gcloud storage cat "gs://$(tf_state_bucket)/$(tf_state_prefix)/default.tfstate" 2>/dev/null) || return 1
```

With no state object `gcloud storage cat` exits 1. That is the expected case — the comment at line 618
says *"Any read failure means 'not ours'"*, and `|| return 1` exists to swallow it. The swallow works;
the banner prints anyway:

1. `install.sh:15` sets `set -Eeuo pipefail`. `-E` (`errtrace`) propagates the `ERR` trap into subshells.
2. `install.sh:76` installs `trap 'on_error $? $LINENO "$BASH_COMMAND"' ERR`.
3. The failing `gcloud` runs inside a `$(…)` **subshell**. Within that subshell it is a standalone
   command, not part of a `||` list — the `|| return 1` binds the *assignment*, in the parent. So the
   inherited trap fires inside the subshell.
4. `on_error` (`install.sh:56-75`) prints the banner and exits, terminating **the subshell only**. The
   parent's `|| return 1` then does its job and the install continues.

So the error *handling* is correct and the error *reporting* is wrong.

Called from `installer_common.sh:739` as `if tf_state_has_cluster; then`.

## Second occurrence

The same unguarded read appears at `installer_common.sh:657-659` in `tf_state_image_tag()`, guarded
with `|| return 0`. `upgrade.sh:672` calls it as a bare command substitution
(`PARAM_IMAGE_TAG="$(tf_state_image_tag)"`), and `upgrade.sh` sets `-E` (`:17`) and its own trap
(`:70`), so the trap fires there too.

To be precise about the blast radius: the banner goes to **stderr** (`install.sh:60`), so it is *not*
captured into the variable. `PARAM_IMAGE_TAG` ends up **empty**, is emptiness-checked at `:673`/`:684`,
and falls through to the `running_image_tag` fallback, which is validated by `validate_immutable_ref`
(`:158`). No bad image tag is used. The residual symptom is a spurious red
`✗ Upgrade error encountered at line 659` plus a `/tmp/kube-agents-upgrade-report.json` clobbered to
`FAILED`.

## Suggested fix

Use the `trap - ERR` pattern already established at ten sites in this codebase
(`installer_common.sh:727`, `:961`; `install.sh:1641`, `:1653`, `:1676`, `:1695`, `:1703`, `:1820`,
`:1827`; `uninstall.sh:250`):

```bash
# installer_common.sh:629, in tf_state_has_cluster()
state=$(trap - ERR; gcloud storage cat "gs://$(tf_state_bucket)/$(tf_state_prefix)/default.tfstate" 2>/dev/null) || return 1

# installer_common.sh:659, in tf_state_image_tag()
state=$(trap - ERR; gcloud storage cat "gs://$(tf_state_bucket)/$(tf_state_prefix)/default.tfstate" 2>/dev/null) || return 0
```

These are the only two unguarded `gcloud storage cat` sites in the installer front doors; the third
(`uninstall.sh:250`) is already guarded.

## Impact

No functional impact — every affected run that was not blocked by an unrelated problem finished
`🎉 Installation Complete!`, exit 0, all pods Running with zero restarts. The cost is entirely in user
confidence, and it is the single most alarming thing shown during a successful install.

## Workaround for users today

None needed; the message is harmless. Note that "run under bash 5" is **not** a usable workaround on
macOS, where `/bin/bash` and the first `bash` on `PATH` are both 3.2.57.

## Verification

- Observed in 6 of 7 captured attempts across four pre-existing clusters (2 Autopilot, 2 Standard),
  absent only from the one retry where state already existed.
- Reproduced in isolation under bash 3.2.57 with a faithful copy of the code's shape (sourced file,
  nested `$(tf_state_bucket)`/`$(tf_state_prefix)`, `if tf_state_has_cluster; then` call site inside a
  function). `BASH_COMMAND` rendered identically, including the `2> /dev/null` spacing bash inserts.
  A *simplified* version did not reproduce it, so the faithful shape is required.
- Not tested: `upgrade.sh` against a live install (the upgrade behaviour above is from the isolated
  reproduction), and bash 5 behaviour.

## Environment

`0.4.0` (`c4af2899648d60927ce28295b14059ae685f12f7`), macOS, bash 3.2.57, gcloud 577.0.0,
Terraform 1.15.8, installing onto pre-existing GKE clusters with `--non-interactive`.

kirill-kosov:
## Summary

Clears the inherited `ERR` trap inside the `$(...)` command substitutions for `tf_state_has_cluster` and `tf_state_image_tag` in `scripts/installer/installer_common.sh`. Under Bash 3.2 (the default on macOS), when `gcloud storage cat` returns exit code 1 because no Terraform state object exists yet, the inherited `ERR` trap fired inside the subshell and printed a spurious red error banner during normal fresh cluster adoptions before outer error handling took effect.

## Why This Change

On any initial installation of kube-agents onto an existing GKE cluster with no prior Terraform state in GCS, `install.sh` printed:

> ✗ Error encountered at line 637 (exit code 1): gcloud storage cat "gs://..." 2> /dev/null
> ℹ Cluster 'cluster-name' already exists and is not in Terraform state; installing onto it (create_cluster = false).

While the installation succeeded, the alarming red banner gave customers the false impression that the installer crashed or hung. Clearing the `ERR` trap inside the subshell prevents `on_error` from firing on expected 404s/non-zero exits from the probe.

## What Changed

- Added `trap - ERR;` inside the `$(...)` subshells for `tf_state_has_cluster()` and `tf_state_image_tag()` in `scripts/installer/installer_common.sh`.
- Added regression test `test_missing_state_does_not_fire_err_trap` to `tests/test_installer_common.py` asserting both structural presence of the guard and runtime non-zero exit behavior under `set -E`.

## Context

Follows the established `trap - ERR` pattern used across `install.sh`, `uninstall.sh`, and `installer_common.sh` for Bash 3.2 subshell trap suppression.

Closes #1290

## Testing

- `python3 -m unittest tests/test_installer_common.py` (66 tests pass).
- `python3 -m unittest tests/test_reconcile_environment.py` (50 tests pass).
- `make docs-check` (passes).

### Live validation

The mechanism was reproduced and verified under an isolated GNU Bash 3.2.57 container (`bash:3.2`, matching macOS `/bin/bash`) with a stubbed `gcloud` returning exit code 1 on state reads:

1. Before fix (unfixed code with stubbed probe):
> Invoking tf_state_has_cluster with reverted (unfixed) code under Bash 3.2:
> ✗ Error encountered at line 637 (exit code 1): return 1
> ℹ Cluster 'test-cluster' already exists and is not in Terraform state; installing onto it (create_cluster = false).

2. After fix (with `trap - ERR;`):
> Invoking tf_state_has_cluster with FIXED code under Bash 3.2:
> ℹ Cluster 'test-cluster' already exists and is not in Terraform state; installing onto it (create_cluster = false).

What was not covered: An end-to-end `install.sh` run against a live GKE cluster without GCS state from a physical macOS workstation, because setup was blocked by macOS endpoint policy (Santa restrictions). The Bash 3.2 subshell trap propagation mechanism itself was validated in isolation using the matching GNU Bash 3.2.57 container.

## Self-Review

Ran two independent pre-PR review passes, each in a clean subagent context handed only the diff range `upstream/main...HEAD`:

1. **Adversarial Pass (clean subagent context):**
- Looked for: Subshell trap scope side-effects, fallback return code suppression, symmetry across sibling probe sites in `install.sh`/`installer_common.sh`, and regression test fidelity (verifying `test_missing_state_does_not_fire_err_trap` fails when `trap - ERR` is omitted).
- Findings: No defects found. Subshell exit propagation (`|| return 1` / `|| return 0`) remains intact.

2. **Docs-Drift Pass (clean subagent context):**
- Looked for: Documentation referencing internal state probe helpers, changes to exported installer interfaces or CLI options, and docs map / generated table synchronization.
- Findings: No documentation drift detected. All mechanical checks (`make docs-check`) passed cleanly.

## Risk & Rollout

Low risk. Confined to installer bash subshells for Terraform state discovery; no operator, runtime, or Helm chart code paths are affected.

---
*Generated with the help of Jetski.*


scripts/installer/installer_common.sh
```diff
@@ -634,7 +634,11 @@ hcl_csv_list() {
 # lifecycle.sh documents for its own state reads).
```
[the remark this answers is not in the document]
kirill-kosov: Made sure that the required tasks were run and updated the PR description.

scripts/installer/installer_common.sh
```diff
@@ -634,7 +634,11 @@ hcl_csv_list() {
...
 tf_state_has_cluster() {
   local state
-  state=$(gcloud storage cat "gs://$(tf_state_bucket)/$(tf_state_prefix)/default.tfstate" 2>/dev/null) || return 1
+  # `trap - ERR` inside the substitution: under bash 3.2 (macOS default) the
+  # caller's inherited ERR trap fires in this subshell even though the failure
+  # is handled by `|| return 1`, printing a spurious error banner on normal
+  # first-time installs.
+  state=$(trap - ERR; gcloud storage cat "gs://$(tf_state_bucket)/$(tf_state_prefix)/default.tfstate" 2>/dev/null) || return 1
```
mplakhtiy: **MEDIUM — PR #1327 rewrites these exact two lines and already carries a fix for the same bug; merging both in either order needs a hand.**

#1327 replaces both probe call sites with a new helper — `state=$(tf_state_read) || return 1` here, `state=$(tf_state_read) || return 0` in `tf_state_image_tag` — and moves the same `trap - ERR` fix into `tf_state_read`, with a comment describing the identical mechanism (the inherited ERR trap firing in a subshell whose `return 1` is the ordinary fresh-install answer, and the drop affecting that subshell only).

Two collisions follow. The textual one is guaranteed: the test-file insertion point is byte-identical in both branches — both hunks open at `@@ -151,6` after `test_unreadable_state_fails_safe`. The one that bites is that this PR's two structural assertions require the literal string `state=$(trap - ERR; gcloud storage cat` to be present in each function, and under #1327's shape that text does not exist — the `gcloud storage cat` call has moved into `tf_state_read`.

So the natural conflict resolution — take #1327's helper, keep #1348's test — produces a **red `main`**, failing with a message claiming the ERR guard is missing when it is present and working one function away. (The runtime half of the test passes unmodified against #1327's shape; it asserts behaviour rather than text.)

Not a defect in this diff, and not something to change unilaterally here: #1327's reviewer is the repository owner, so this is a request that the two authors agree an order before either merges. Options, in the order I would prefer them: #1327 absorbs the behavioural test into `tf_state_read`; or #1348 merges first and #1327 rewrites the two `assertRegex` calls as part of its own conflict resolution; or #1348 drops the structural pair now, leaving the shape-agnostic runtime test and reducing the collision to an ordinary textual conflict.

mplakhtiy (commented):
The change itself is correct. I reproduced the defect and the fix end to end on bash 3.2 under the installer's real `set -Eeuo pipefail` with an `on_error` trap and a stub `gcloud` exiting 1: unfixed, two spurious banners with `BASH_COMMAND` rendered exactly as #1290 reports it; fixed, none — and identical semantics either side (`has=no`, `tag=[]`, rc 0). The new test is a real guard, not a tautology: with `trap - ERR;` reverted and the two structural `assertRegex` calls neutered, the runtime half still fails (`'ERR_TRAP_FIRED' unexpectedly found in 'ERR_TRAP_FIRED\nERR_TRAP_FIRED\n'` — one fire per function).

Two things worth raising, neither a defect in this diff. One is inline on `tf_state_has_cluster`; the other has no line in this diff to sit on:

**MEDIUM — the guard lands only on the two `gcloud storage cat` sites, while six siblings share the shape, including one on the fallback path #1290 itself describes.**

The defect class is "a command substitution whose last command may fail, with the tolerance applied outside the substitution, under `set -E` plus an ERR trap" — not "unguarded `gcloud storage cat`". Sweeping the three front doors and `scripts/installer/` for that shape turns up six sites that still have it (sites where the tolerance is *inside* the substitution, `$(cmd 2>/dev/null || echo "")`, are exempt and fine).

The one that matters is `running_image_tag` at `scripts/installer/installer_common.sh:701`:

```
if ! image="$(kubectl get deployment platform-agent-gateway -n "${namespace}" \
  -o jsonpath='...' 2>/dev/null)"; then
```

`upgrade.sh:728` calls `tf_state_image_tag` — the function this PR fixes — and `upgrade.sh:741` calls `running_image_tag` as the fallback for exactly the miss this PR just silenced:

```
PARAM_IMAGE_TAG="$(tf_state_image_tag)"      # :728 — banner suppressed by this PR
…
if [ -z "$PARAM_IMAGE_TAG" ]; then
  PARAM_IMAGE_TAG="$(running_image_tag "$target_namespace")"   # :741 — still unguarded
```

So on an upgrade where the Deployment read also fails, the same red banner and the same `FAILED` report appear one line later in the same run — the symptom this PR set out to remove. The other five: `helm_release_status` (`installer_common.sh:723`), `installer_common.sh:504` and `:907`, the `git rev-parse` pair at `install.sh:1125` / `upgrade.sh:397`, and `gke_dns_endpoint.sh:77`.

This is correctly out of scope here. #1290 scoped itself to `gcloud storage cat`, and its "these are the only two unguarded sites" claim is true within that scope. A follow-up issue — `running_image_tag` and `helm_release_status` first — rather than a change request on this branch.

scripts/installer/installer_common.sh
kirill-kosov: Addressed. PR #1327 was merged before this PR, so the conflicts were addressed following recommendations in this comment.

mplakhtiy (approved):

Issue #1297, opened 2026-09-08
bug(installer): no capacity preflight — an under-provisioned Standard cluster hangs 10 minutes then fails with a bare "context deadline exceeded"


mplakhtiy:
## Summary

Installing onto a GKE Standard cluster without enough schedulable capacity produces a ten-minute silent
stall followed by a bare `context deadline exceeded`. The error names nothing — not pods, not scheduling,
not CPU. Nothing in the install path checks capacity first, and nothing surfaces pod state during the wait.

#521 proposes a preflight stage that would cover the checking half. This issue adds measured resource
figures, the single-node constraint, and two gaps #521 does not mention: no pod-event visibility during the
wait, and a hardcoded Helm timeout.

## What happens

The Helm release sits at `Still creating…` with only an elapsed counter. Inspecting the cluster during the
window:

```
kube-agents-controller-manager-…   1/1   Running
litellm-…                          0/1   Pending
litellm-…                          0/1   Pending
platform-agent-gateway-…           3/3   Running
```

```
0/2 nodes are available: 1 Insufficient cpu, 1 node(s) had untolerated taint(s).
```

After the timeout:

```
Error: Helm release error
context deadline exceeded
```

Exit code 1 after ~10 minutes.

## Why

The two nodes are the cluster's own pool (out of CPU) and the `gvisor-pool` the installer just added, which
carries GKE's gVisor sandbox taint. LiteLLM has no matching toleration, so its two replicas have nowhere to
run. In effect the installer adds capacity only the sandboxed agent can use, then schedules LiteLLM onto the
user's pre-existing capacity without checking it exists.

A precision note, since it affects any fix: the taint is **not declared in this repository**. The node pool
(`terraform/modules/gke-cluster/main.tf:284-304`) sets `sandbox_config { type = "GVISOR" }` and has no
`taint` block; the string `sandbox.gke.io` appears nowhere in the repo. GKE applies
`sandbox.gke.io/runtime=gvisor:NoSchedule` itself. Correspondingly nothing in the chart tolerates it either
— the agent lands on the pool because GKE auto-tolerates RuntimeClass-`gvisor` pods.

## What the install actually requests

Summed from chart and operator defaults for a stock install (cert-manager on; dashboard, minter and
Hindsight off):

| Workload | Replicas | CPU each | Memory each | Source |
| --- | --- | --- | --- | --- |
| controller-manager | 1 | 10m | 64Mi | `charts/kube-agents/values.yaml:147-149` |
[...]

Issue #521, opened 2026-08-04
No preflight: provisioning mutates cloud resources before checking the cluster can run the agent


bnaylor:
## What happened

Installing against an existing GKE cluster, the pipeline started changing cloud resources at stage 01 and then failed at stages 03, 04 and 08 for three separate reasons that were all knowable before it started.

`provision_01` creates a cluster that is regional, `e2-standard-4`, Workload-Identity-enabled, with the Filestore addon.  Those are the agent's real requirements.  None of them are written down as requirements, and nothing checks them when you bring your own cluster.

- stage 03 failed enabling the Filestore CSI addon
- stage 04 failed the Workload Identity binding, because a console-created cluster has no identity pool and its node pool has no GKE metadata server
- stage 08 would not have scheduled anyway: the agent requests 500m CPU / 2Gi plus a ~200m/256Mi credential-proxy sidecar, and `e2-medium` nodes have 940m allocatable

Each was found after the previous stage had already changed something.

## The expensive one

Stage 04 creates the `roles/iam.workloadIdentityUser` binding and never tests it.  Nothing mints a token.

In my case the binding was byte-for-byte correct, the KSA annotation resolved to the right GSA, `iamcredentials.googleapis.com` was enabled, and minting was still refused with a 403.  Every check the installer performs passed, while the thing those checks exist to guarantee did not work.  (The project turned out to be the wrong kind of project for this, which is my problem and not the installer's.  The point is that the installer had no way to tell me.)

It surfaced four stages later as an Envoy sidecar crashlooping on a swallowed error.  Roughly two hours to trace back.

## Suggested fix

A preflight stage that runs before anything is mutated:

- all local tooling, once, rather than `check_prereqs` per stage
- cluster shape: Workload Identity enabled, node pool on `GKE_METADATA`, required addons
- schedulable capacity for the agent's actual requests
- a real Workload Identity token exchange, using a throwaway KSA/GSA pair and one `curl` against the metadata server

That last one is the highest-value item here.  It takes under a minute and fails with a clear message.

Full context: https://gist.github.com/bnaylor/b8e7b7f9dda75c6f159815b36aec3d49

bnaylor:
Follow-up with evidence from the other direction.  The original report argued from an install that failed; since then I have done one that succeeded, on a fresh project where `provision_01` built the cluster itself.  The contrast is the argument.

Same installer, same agent, same `vars.sh` apart from project and region:

[...]

Issue #1254, opened 2026-09-05
upgrades-lagging-master-probe: rung-4 collapse on unrelated PRs — the delegation acknowledgement is graded as the final answer


dshnayder:
`upgrades-lagging-master-probe` collapsed at rung 4 on #1134, a pull request whose diff cannot reach it, and the same case collapsed the same way on #1230 the day before — a change to `hack/ci-eval-rc.sh` and two test files. Over the last four days the case's repetitions run at roughly half, so a 0/3 window is a matter of when rather than whether. Same shape as #1171 (`compliance-rbac-overgrant`) and #1189 (`rca-remediation-pr`), and the same underlying condition as #1225.

## What the graded record shows

[#1134, build 2096255636072304640](https://oss.gprow.dev/view/gs/kube-agents-prow/pr-logs/pull/gke-labs_kube-agents/1134/pull-kube-agents-smoke-test/2096255636072304640) — RED on this case alone. Admitted-case pass rate 86.7%; every other admitted case passed 3/3.

```
upgrades-lagging-master-probe | upgrades | FAILED (rung 4) | 0/3
VerificationCorrectness=0.5 (floor 1.0) -- the-probe-identifies-the-version-lag:
none of the alternative phrasings present:
['behind', 'lagging', 'not current', 'out of date', 'outdated', 'one minor', 'newer version']
```

The three repetitions are real runs, not the null-token class from #1116: 281 s / 442k tokens, 272 s / 426k, 173 s / 268k. `ToolInvocation` is 0.0 on all three and `coverage` is 1.0 — the verifier ran and graded something.

What it graded is the delegation notice. Two of the three actual outputs are:

```
> 🔀 Delegated to the **platform** agent

I've started this as task `t_3a2f3a95`. The answer will post into this thread as soon as …
```

and the third is the planner's tool trace with `tools_used: ["kanban_create"]` and nothing after it. The judge's reason reads "it merely created a Kanban task containing the user's instructions … it did not generate the required readiness report or cite any actual versions" — which is accurate about what it was shown. This is #1225's failure mode: the front end reports delegation status as the final answer.

## Why the worker never answered

`platform-agent-gateway.log` from the same build carries this warning across the whole 1 h 48 m window, at the warning's own ~5-minute cooldown — so it is continuous, not occasional:

```
11:28:01  kanban dispatcher stuck: ready queue non-empty for 6 consecutive ticks but 0 workers spawned.
11:33:01  … 66 consecutive ticks …
[...]

PR #1353, opened 2026-09-09
docs(install): state what a customer-supplied cluster must already have

bradhoekstra:
## Summary

Adds a **Cluster requirements** section to the site's Prerequisites page that states, in one place, what a GKE cluster somebody else made must already have before `install.sh` or the `full-install` composition is pointed at it with `create_cluster = false`: one row per requirement, the field to read in `gcloud container clusters describe`, and what happens today when the row is not met (refused at plan time, changed by `install.sh`, or unchecked and failing later). INSTALL.md, the composition README, and the three `create_cluster` descriptions now point at or agree with it.

## Why This Change

The requirement set was real and enforced, but scattered. The `gke-cluster` module README named Workload Identity and NetworkPolicy enforcement; the `create_cluster` variable descriptions and `terraform.tfvars.example` named only Workload Identity, so a bare-Terraform user hit the NetworkPolicy postcondition as a surprise; and the site's Prerequisites page named no cluster feature at all and said the cluster need not be pre-provisioned. Nothing said that GKE 1.29, `GKE_METADATA` on every node pool, a control plane the Helm provider can dial, and schedulable capacity are never checked and fail after the GCP resources exist. INSTALL.md also still counted three out-of-band installer steps and omitted the Workload Identity and NetworkPolicy ones.

## What Changed

- `docs/site/src/content/docs/install/prerequisites.md`: new **Cluster requirements** section with the two read-only `gcloud` commands and the requirements table; the intro no longer says a cluster need not be pre-provisioned without qualification.
- `INSTALL.md`: the out-of-band-steps paragraph now names the Workload Identity and NetworkPolicy pre-steps, and the tooling matrix gains a pointer to the new section.
- `terraform/examples/full-install/README.md`: `create_cluster = false` is described as reading, not providing, the cluster features, and Prerequisites lists what such a cluster must have.
- `terraform/modules/gke-cluster/variables.tf`, `terraform/examples/full-install/variables.tf`, `terraform.tfvars.example`: the `create_cluster` wording now matches both postconditions.
- `.agents/skills/install-kube-agents/SKILL.md`: the out-of-band steps run before the apply, not after.
- `docs/README.md`: the page's one-line summary in the map.

## Context

- #521 asks for a preflight that checks these things; this is the documentation half, and the preflight belongs on top of #1336 once it lands. #1297 supplies the capacity figures.
- #1336 (sergiu1spiridon) makes the node-pool migration and Calico enablement opt-in and rewrites the same INSTALL.md paragraph. This PR describes `main` as it stands; whichever merges second changes two table cells and resolves a small conflict in that paragraph.
- #1327 (mplakhtiy) edits adjacent lines in the composition README and other lines in the same tfvars and variables files; #1331 (kyber775) edits the tooling list at the top of the same Prerequisites page. No shared hunks with either.

## Testing

- `make docs-check` clean (generated regions, links in 295 files, terminology, map with no re-aligned rows, context budget).
- `prettier --write` (3.9.6, the CI pin) on every changed Markdown file.
- `terraform fmt -check -recursive terraform/`, and `terraform validate` in `terraform/examples/full-install` and `terraform/modules/gke-cluster` (description-only changes).

### Live validation

Read-only, against `kube-agents-cluster` (regional `us-central1`, GKE Standard, 1.35.7-gke.1027000) in project `bhoekstra-gkedemos`, the install this repository's contributors share. Both commands the page documents were run as written and render the documented fields:

```
autopilot: {}
controlPlaneEndpointsConfig:
  dnsEndpointConfig:
    allowExternalTraffic: true
    endpoint: gke-…-757207957170.us-central1.gke.goog
currentMasterVersion: 1.35.7-gke.1027000
databaseEncryption:
  state: ALL_OBJECTS_ENCRYPTION_ENABLED
workloadIdentityConfig:
  workloadPool: bhoekstra-gkedemos.svc.id.goog

NAME          MODE          MACHINE_TYPE
default-pool  GKE_METADATA  e2-standard-4
```

That run is also why the page says an unmet requirement shows as a missing key: this cluster has neither Dataplane V2 nor the Calico addon, and `networkConfig.datapathProvider`, `networkPolicy`, and `privateClusterConfig.enablePrivateEndpoint` are simply absent from the output (`addonsConfig.networkPolicyConfig.disabled: true` confirms it). So the page's NetworkPolicy row was checked against a cluster that fails it, and the Workload Identity, metadata-mode, version, DNS-endpoint, and CMEK rows against one that passes them.

Not covered: no `install.sh --dry-run --create-cluster=false` or `terraform plan` was run against an adopted cluster, so the plan-time refusal itself was not reproduced here; the postcondition text was verified in `terraform/modules/gke-cluster/main.tf`. Nothing was mutated and no artifacts were left.

## Self-Review

Both passes ran in fresh subagents handed the range `refs/kube-agents-base/main...HEAD` and the skill path, told to derive intent from the diff and edit nothing. `review-adversarial` ran once over the first commit (angles A through J: line-by-line prose against `install.sh`, `installer_common.sh`, the module and composition HCL, chart values and operator manifests; removed-text audit; identifier and anchor tracing; scope; sibling PRs). `review-docs-drift` ran over the first commit and again in a new context after the fixes, because the fixes reworded factual claims. Mechanical gate as under Testing.

1. **HIGH, CONFIRMED (adversarial; Blocking in docs-drift).** The control-plane row told a reader on a private-endpoint cluster to enable the DNS endpoint, but the Terraform Helm provider always dials the IP endpoint (`providers.tf`, `outputs.tf`). Fixed: the row separates the Terraform leg from the agent and post-apply `kubectl` leg.
2. **Blocking (docs-drift, both runs).** INSTALL.md's "three things stay outside Terraform" omitted the Workload Identity and NetworkPolicy pre-steps, and the new link sat under it. Fixed with count-free wording. #1336 rewrites the same paragraph; see Context.
3. **MEDIUM, CONFIRMED (adversarial).** The capacity row said the Helm release fails when pods cannot schedule. True for LiteLLM or the operator; the agent pod is operator-created after the release, so the apply succeeds and `install.sh` warns after `ROLLOUT_TIMEOUT_SECS` and reports `SUCCESS_PENDING_ROLLOUT`. Fixed: the row describes both outcomes.
4. **MEDIUM, CONFIRMED (adversarial; Advisory in docs-drift).** The composition README's "the first two it can read" pointed at version and Workload Identity; the postconditions check the Workload Identity pool and NetworkPolicy enforcement. Fixed by naming them.
5. **MEDIUM, CONFIRMED (adversarial).** Sibling divergence with #1336, which makes the migration and Calico steps opt-in. Not fixed, deliberately: docs describe `main`, and the two cells change when #1336 lands.
6. **LOW, CONFIRMED (both).** The intro's hard-coded counts ("two checked, three changed") contradicted the table. Fixed by removing them.
7. **LOW, CONFIRMED (both).** Adoption is selected only when the cluster exists and is not in this install's Terraform state. Fixed.
8. **LOW, CONFIRMED (adversarial, reproduced live).** The yaml projection drops unset keys. Fixed with a reading note; the node-pool table prints an empty cell instead, also stated.
9. **LOW, PLAUSIBLE (adversarial).** The cert-manager probe looks only in the `cert-manager` namespace. Fixed with a parenthetical; the page's cert-manager section already states the limitation.
10. **Advisory (docs-drift re-run), all fixed:** a dry run stops at the postcondition on a cluster missing the pool or enforcement; "prefer Dataplane V2" was not actionable on an existing cluster; VAP and OCI volumes degrade rather than fail below their floors; the install skill placed CMEK after the apply; an existing `gvisor-pool` collides with the pool the module creates; the intro still said cert-manager must be installed by the reader.
11. **Advisory (docs-drift), capacity figure.** The two passes summed the stock requests to about 1.5 and about 1.7 vCPU depending on whether the shell sandbox counts. The page says "under 2 vCPU and about 4 GiB" and "more than 1 vCPU on a single node" rather than a number either pass disputes.

Not covered by either pass: an `install.sh` or `terraform plan` run against an adopted cluster, the docs site `npm run build`, and the exact Helm provider timeout string.

Generated with the help of Claude Fable 5.1.

## Risk & Rollout

Docs and HCL description strings only; no runtime code path, template, or default changes. Revert by reverting the commits. The one merge-time item is the INSTALL.md paragraph shared with #1336.


toshiowang (approved):

bradhoekstra:
The smoke red is #1254, not this diff: `upgrades-lagging-master-probe` failed 0/3 with the delegation acknowledgement graded as the final answer (judge: "used kanban_create to delegate the prompt ... did not provide the expected readiness report"), the same shape as #1134 and #1230. This pull request changes docs and Terraform variable descriptions only; every other admitted case passed 3/3 (admitted pass rate 90%), and the same case ran 2/3 on #1351 an hour earlier. Occurrence logged on #1254.

Issue #1356, opened 2026-09-09
feat(observability): proactive zonal skew and topology distribution anomaly detection


suchayan-mitra:
## Problem Statement
In multi-tenant, multi-zone GKE clusters, workloads frequently suffer from undetected zonal skew where replicas inadvertently concentrate in a single availability zone due to node cycling, cluster autoscaler scaling events, or rolling updates. When a localized zonal disruption occurs, services that appeared to have high replica counts suffer severe degradation or outright outages because replica distribution was not evenly balanced.

## Customer Evidence
- Customer Profile: High-traffic multi-tenant platform running mission-critical streaming microservices across 3-zone regional clusters.
- Pain Point: Platform operators experienced customer-facing outages caused by zonal imbalances that were invisible until a zone event occurred. Manual auditing of thousands of deployments across dozens of clusters is infeasible.

## Proposed Design & Scope
1. **Topology Distribution Analyzer**:
   - Continuously or periodically analyze pod distribution across availability zones (`topology.kubernetes.io/zone`) for all Deployments and StatefulSets with `replicas > 1`.
   - Calculate a Zonal Skew / Entropy index for each service.
2. **Missing Constraint Detection**:
   - Identify workloads lacking `topologySpreadConstraints` or using obsolete `podAntiAffinity` rules that fail under tight scheduling constraints.
3. **Remediation & GitOps Automation**:
   - Propose standard `topologySpreadConstraints` configurations tailored to workload size and zone counts.
   - Automatically format recommendations into pull requests against the workload GitOps repository.

## Acceptance Criteria
- [ ] Analysis tooling computes zonal skew across all namespaces or targeted workloads.
- [ ] Configurable thresholds for alert triggers (e.g., skew >= 2 replicas between max and min zones).
- [ ] Flags workloads missing explicit `topologySpreadConstraints`.
- [ ] Emits declarative YAML patch suggestions compliant with modern Kubernetes best practices.

mastersingh24:
@haoxuw  - are you actually working on this yet?  I have started some work on this.  Let me know.

haoxuw:
> [@haoxuw](https://github.com/haoxuw) - are you actually working on this yet? I have started some work on this. Let me know.

Yes there's a design doc being reviewed: https://github.com/gke-labs/kube-agents/pull/1360/changes#diff-871b1248a6aa938f7b8473ab5755c02f72bc6de45e67b9ebac67e5b21206f2da 

My plan was to
[...]

Issue #1355, opened 2026-09-09
feat(upgrade): fleet-wide Kubernetes upgrade readiness and API deprecation scanner


suchayan-mitra:
## Problem Statement
Operating large-scale Kubernetes fleets spanning millions of cores across dozens of cluster families creates significant risk during minor version upgrades (e.g. 1.30 -> 1.31). Workloads and third-party CRDs frequently rely on deprecated or removed API versions, leading to post-upgrade scheduling failures, broken controllers, or broken ingress configurations. Platform teams need proactive, automated verification before initiating fleet-wide upgrade rollouts.

## Customer Evidence
- Customer Profile: Multi-tenant streaming and media infrastructure running millions of cores across heterogeneous cluster families.
- Pain Point: Upgrading clusters requires laborious manual audits of running workloads and CRDs against deprecation notices. Upgrades have previously triggered production incidents due to undetected deprecated API usage in tenant workloads.

## Proposed Design & Scope
1. **API Deprecation Audit Engine**:
   - Cross-reference live cluster APIs, active CRDs, and in-flight API requests against Kubernetes deprecation databases and release notes for target minor versions.
   - Audit Helm releases and GitOps manifests stored in tenant repositories.
2. **Readiness Scoring & Risk Matrix**:
   - Categorize findings by severity: `BLOCKING` (removed in target version), `WARNING` (deprecated, removal pending), and `INFO` (field-level behavior changes).
   - Generate an executive Fleet Upgrade Readiness Report.
3. **Remediation Proposals**:
   - Propose automated `kubectl convert` / manifest transform patches targeting GitOps repositories for deprecated API migrations.

## Acceptance Criteria
- [ ] Capability to specify target target Kubernetes minor version (e.g., `--target-version=1.31`).
- [ ] Comprehensive audit covering core Kubernetes resources, admission webhooks, and third-party CRDs.
- [ ] Output available in both CLI summary and structured JSON/Markdown formats for CI/CD gating.
- [ ] Verifiable zero false-positives on standard GA API schemas.

mastersingh24:
We should figure out why the built-in [deprecation insights feature ](https://docs.cloud.google.com/kubernetes-engine/docs/deprecations/viewing-deprecation-insights-and-recommendations)is not enough?  Or is it that they'd like to have an agent interpreting this info in content of their actual workloads?

bradhoekstra:
[...]

Issue #1343, opened 2026-09-09
feat(skills): fleet upgrade rollout verification — member version polling, readiness audit, API deprecation checks, lagging-cluster detection


adrianchung:
## Problem

An upgrade rolled out across a GKE Fleet can stall on individual member clusters with no centralized signal: the fleet-level view does not say which members lag, why, or whether workloads on them use APIs the target version removes. Operators discover lagging clusters one incident at a time. The shipped skills advise on upgrades and report weekly version currency, but nothing tracks a rollout in progress or verifies readiness against a specific target version.

## Customer evidence

A large multi-tenant GKE customer, ranking their asks (this was their highest-complexity theme):

> "Upgrade rollouts across GKE Fleets stall silently without centralized telemetry or failure metrics to identify lagging member clusters. Req: Upgrade verification skills to poll fleet versions, audit readiness, check API deprecations, and detect version drift."

Source: private customer feedback relayed by the filer — not submitted through the feedback form, despite the `external-feedback` label's description; stating that here keeps the label honest.

## What exists today

- `agents/platform/skills/gke-upgrades/` — produces upgrade plans, runbooks, and checklists; knows rollout sequencing, channels, rollback constraints. It is advisory: it does not poll a fleet or track a rollout.
- The weekly `security-patch-orchestrator` audit stream (`agents/platform/governance/security_patch_orchestrator_sop.md`) — read-only, fleet-wide, reports "does each cluster run a version its channel still offers, and is it configured to stay current." Point-in-time currency and policy hygiene; no target version, no rollout tracking, no deprecation scan.
- The eval domain already exists: `bench/tasks/upgrades-lagging-master-probe/` and `bench/tasks/upgrade-readiness-lagging-cluster/` grade lagging-cluster reasoning, so new capability here has a grading home. Caveat: `upgrades-lagging-master-probe` has a known grading defect (#1254 — the delegation acknowledgement is graded as the final answer, collapsing rung 4 on unrelated PRs), so a new case should copy its fixture shape, not its grading.
- #57 (open) notes "upgrade readiness" has no static repo-review skill family and explicitly defers that sibling family. That is the manifest-side (are PDBs/surge/deprecated APIs declared right in the repo); this issue is the runtime/fleet side.

## What is missing

1. **Rollout tracking:** given a target version and a fleet, report each member's control-plane and node-pool versions, which members have not started/completed, and progress over time (a member unchanged across two polls during an active rollout is the stall signal).
[...]

Issue #911, opened 2026-08-24
feat: Create Drift Detector


stalhaali:
https://github.com/gke-labs/kube-agents/blob/2b450228b050341cb57841f5e1dc7d07810a1df6/agents/platform/docs/drift-detection.md

stalhaali:
## Placement: `k8s-operator/cmd/drift-detector` (Go)

Confirming placement before I scaffold, since [the task breakdown](https://github.com/gke-labs/kube-agents/blob/76501a02a3693c86bfcb13c1b8dbfab950008032/agents/platform/docs/drift-detection/cuj3-task-breakdown.md#L55-L57) (on #467) and `docs/designs/drift-detection.md:41` disagree — the breakdown says a Go sibling of the event watcher, the design doc says a new route on the `agentplugins/pubsub-platform` adapter.

I think Go wins on two things. The adapter's filter language (`adapter.py:503`) only supports `==` and `!=`, but T2 needs prefix and regex matching (`^system:`, `.*-controller$`), so we'd have to extend a component stockout also uses. And T3 has to fetch live objects from any cluster in the fleet, which `cmd/k8s-event-watcher` already does safely — profile discovery, GKE API lookup, Workload Identity token — while the Python adapter has no Kubernetes client at all. Go is also cheap to ship: the watcher isn't its own image or Deployment, just a `go build` in the agent image plus a launch block in `start-services.sh`. The real costs are that we'd add a Pub/Sub dependency and can't reuse `dedup.go`/`injector.go` without lifting them out of `package main` first.

Plan is T1–T4 in Go, T5 (the digest) as a skill. I'll also fix the contradicting line in `drift-detection.md` in the first PR. Shout if you'd rather go the adapter route — the filter-language point is the one that would flip it if extending `_eval_filter` is wanted anyway.

bradhoekstra:
Suggested priority: `priority:p2` — Drift detector is active roadmap work with PRs landing (#1594 merged, #1668, #1596 open). Assigning to @mastersingh24 and @fkc1e100 for triage.

PR #1360, opened 2026-09-09
docs(designs): requirements for the capability delivery vehicle

haoxuw:
## Summary

One requirements document: what a fleet capability must provide to ship pre-defined, run on a schedule, run from chat or on an event, be tuned through an edit the operator agrees to, and learn from conversations. It states requirements R1–R7 and how each is verified on a real installation — no implementation: no file names, merge rules, or tool names.

Split per @mastersingh24's request so the vehicle can be agreed on before the capabilities that ride on it. The two check documents that were here are now #1778.

## Why This Change

Several capabilities share one lifecycle, and each would otherwise reinvent scheduling, chat triggering, event handling, and tuning. Two of the five properties do not exist for anything today: criteria that survive a restart, and learning from conversations. The Scope table is the boundary between what an install already has — the governance-audit roster, `profile-cron-tick`, the chat relay, kanban delegation, agent-created schedules, and the two event adapters — and what this document asks for.

## What Changed

- `docs/designs/capability-delivery-vehicle.md` — status line, the five properties, a Scope table, the deferral of a GitOps-backed persistence path (see below), requirements R1–R7 including the audit and advisory capability shapes, the three trigger paths with the event path stated from what exists, the learning-policy classes, three durability tiers, write-path safety, and a verification table.
- `docs/README.md` — one map row.

Requirements only: no code, no roster entries, no SOP changes.

## Context

The two capabilities that ride on the vehicle — upgrade readiness and fleet anomaly detection — are proposed separately in #1778, which references this document by path until it merges. Implementation of the two new properties is proposed in #1368 (draft); whichever of that and this merges second should update this document's status line.

**Scope, and what is deferred.** Tuned values live on the profile's volume in v1 (the Runtime tier), not in the operator's configuration repository. @jayantid and @fkc1e100 both raised that the architecture places curated knowledge in git and that a volume-resident value cannot be peer-reviewed or rolled back through version control — both correct. A git-backed overlay depends on foundational GitOps work in flight separately, so it is out of scope here: the Scope section says so up front, and R6 records the costs, the changelog shape that lets accumulated changes be harvested into one reviewed pull request later, and the git-backed path that exists today (the obtainability audit's declared-intent step). Moving the default to Reviewed is expected follow-up.

## Testing

- `make docs-check` on this branch rebased onto current `main`: generated regions current, links and design-doc citations resolve, terminology passes, the map covers all tracked docs with no stale cells and no re-aligned rows, context budget unchanged.
- `prettier --check` on the document and `docs/README.md`.

### Live validation

Not live-tested: documentation only, nothing reaches a running installation.

## Self-Review

Both required passes ran in fresh subagent contexts handed only the range, the skill path, and the `docs-check` output — `review-adversarial` twice and `review-docs-drift` five times across the rounds, since each round changed claims about `main` and re-checking them was the point. Merged dispositions for the content still in this pull request:

- **Fixed.** Every claim about an implementation that does not exist on `main` (the document briefly described #1368's store as shipped — the root cause of the first review round, now rewritten as requirements). R5 contradicting itself about what ships (now propose-only everywhere, with the class table as a ceiling a release may raise and "a new check" defined as toggling a declared key). The silent-run rule swallowing an applied change's announcement (an applied change now cancels the next run's silence, and the announcement lands in the chat thread rather than the ledger the same run may close). Triggerable claimed as fully existing when a scoped chat request is an ad-hoc investigation and the agent cannot issue `hermes cron run`. The Pub/Sub adapter described as passing through the Session KV server (it filters and dispatches per route). "Writes shared memory" for an event-opened session (the Platform Agent writes none; facts go through nomination). The `silent_ok` conditions stated incompletely. Drift detection called design-only when its banner says partly built.
- **Deliberately not fixed.** The GitOps deferral above — recorded rather than resolved, by agreement. Merge order with #1368 makes one of the two status lines stale on arrival; that is a fact about sequencing, and #1368 carries the follow-up.
- **Could not be covered.** Hermes-internal behaviour behind `cronjob(...)` is not in the tree; no live installation was involved; the final split (reducing this branch to the vehicle document and re-pointing its references at #1778) was verified by `docs-check` and prettier, not by a further fresh-context pass.

## Risk & Rollout

Low risk: one new Markdown file and one map row, no runtime code paths touched. Revert by deleting them.


bradhoekstra:
@haoxuw Please fix review bot findings before assigning human reviewers.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,241 @@
...
+policy a release could never tighten. Existing governance audits express criteria inline in an
+SOP; moving onto the vehicle means lifting those into `criteria.json` and leaving the SOP as
+procedure, with the shipped default still printed beside each check.
+
+The criteria cannot live under `skills/`. The agent pod replaces that tree wholesale on every
+start, and the shell sandbox — where the agent's shell reads files — carries its own image copy
+that the mirror refuses to overwrite. So the store sits at `profiles/platform/capabilities/`, and
+the agent reads and writes it through `capability_criteria`, an in-process tool on the
+`platform_control` MCP server, never through the shell. `agents/platform/capabilities/README.md`
+is the contributor's reference; `capability_store.py` beside the server is the implementation.
```
jayantid: 🟡 **Not merge-blocking on its own, but this PR cannot land before #1368 without these docs being false.** Every present-tense "built" claim here describes #1368, which is still a draft.

Lines 66 to 69 say the store sits at `profiles/platform/capabilities/`, that `capability_criteria` is a tool on `platform_control`, and that `capability_store.py` is the implementation. Line 189 names `VOLUME_WINS_GLOBS`. Line 157 says `learning.json` ships today. The `docs/README.md` row at line 318 says "criteria store and tool built". None of it is on `main`. All of it is in #1368, whose own description says "design in #1360". Each PR presents the other as landed.

Whichever PR merges first leaves `main` describing code it does not have. The docs map is what the drift skill and later readers consult, so "built" sends them looking for a tool that is not registered. The bot's findings on lines 70, 192 and 232 are all this one item.

Fix, pick one:
- Rebase this branch onto #1368 and merge after it, keeping the present tense.
- Or reword these passages and the README status to "built in #1368, in review", and merge now.

The PR body still says `criteria.yaml`, `learning.yaml` and `MERGE_PATHS`; the document says `.json` and `VOLUME_WINS_GLOBS`. Align it with whichever you keep.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,241 @@
...
+report relay; the optional ConfigMap overlay for reviewed overrides; and, per capability, the lift
+of inline SOP thresholds into criteria. The
+on-demand `hermes cron run` gap is tracked in `agent-shell-sandboxing.md` and is not part of this
+design.
+
+## The first two capabilities on the vehicle
+
+- **Upgrade readiness** ships as a weekly job — today's `security-patch-orchestrator` audit is
+  the first capability moved onto the store, with its skew, spread, exclusion-window and image-type
+  thresholds and a cluster exclusion list in `criteria.json`. As the checks grow, the file also holds the per-family target
```
jayantid: 🟡 **Not merge-blocking, but this line stays wrong even after #1368 lands.** The store #1368 ships has no cluster exclusion list.

Line 231 says the `security-patch-orchestrator` criteria carry "a cluster exclusion list in `criteria.json`". #1368 removed that key in commit 28efced0. Its `criteria.json` holds four thresholds: `pool_skew_critical_minors`, `fleet_spread_min_minors`, `blocking_exclusion_min_days` and `deprecated_image_types`.

An operator reads this, asks the agent to exclude a cluster, and the store refuses the unknown key. The reason #1368 dropped it is the real hazard. A cluster left out of both scope lists makes the fleet-audit delta count every finding it carried as resolved, closing remediation PRs and the ledger. A design that names the exclusion list as a shipped, tunable knob invites exactly that.

Fix: drop it from this line, or move it to the "as the checks grow" sentence with its prerequisite named:

```
thresholds in `criteria.json`. A cluster exclusion list waits on a scope bucket
fleet-audit's delta understands (dropped from #1368 for that reason).
```

jayantid (changes requested):
🔴 **Merge-blocking.** Neither requirements doc says which of its checks already run on `main`. Merged as written, the docs describe seven audits an install already has as work still to do, and the checks that are genuinely new are indistinguishable from the ones that are not.

**What's wrong.** The vehicle doc names the existing governance audits as "the next candidates to move onto it", but `upgrade-readiness-checks.md` (line 31 onward) and `fleet-anomaly-detection-checks.md` (line 31 onward) list every check as a candidate criterion with no pointer to the job that runs it today. Most of the guardrail, cost and version-posture sections restate SOPs already on the roster.

**What happens if it merges.** Every fact in this repository has one home (AGENTS.md, Documentation Guidelines). These two docs become a second home for the check lists of seven SOPs, and the next edit to a threshold in `compliance_audit_sop.md` or `fleet_wide_cost_analysis_sop.md` leaves them silently stale. A contributor picking up #1355 or #1356 from these docs plans to build `compliance-audit` and `stockout-prevention` again.

Mapping each section against `agents/platform/cron/jobs.json` and `agents/platform/governance/`:

| Proposed | Already on `main` | Delta |
| --- | --- | --- |
| Vehicle: pre-defined, scheduled, triggerable | Nine governance audits on the Platform Agent roster, ticked by `profile-cron-tick`, delivered to chat, ledgered by `fleet-audit`; on-request via kanban; runtime-created jobs | None. The doc's own "What exists" section agrees. |
| Vehicle: criteria that survive a restart | Nothing. `MERGE_PATHS` covers only `cron/jobs.json`. | #1368 (draft) adds the store, `capability_criteria`, `VOLUME_WINS_GLOBS`. |
| Vehicle: self-learning reflection | Nothing | Entirely new. |
| Upgrade: version posture | `security-patch-orchestrator` (weekly): `master-behind`, `pool-skew`, `fleet-spread`, `no-channel`, `no-autoupgrade`, `no-autorepair`, `no-maintenance-window`, `blocking-exclusion`, `stale-image-type`, `no-notifications`. `fleet-upgrade-verification` skill for version currency. | Days-to-end-of-support per family. |
| Upgrade: deprecation insights | Nothing running. `lifecycle_deprecation_manager_sop.md` covered manifest scanning; its job is retired. | New here. #589 ingests Recommender insights; #1343 / #1410 / #1411 / #1355 track the rest. |
| Upgrade: add-on compatibility, webhooks, drain safety, rollout sequencing, post-upgrade diff | `obtainability-audit` (daily): `no-pdb`, `blocking-pdb`, `rigid-scheduling`, `no-spread`, `single-replica`. | Add-on matrix, webhook `failurePolicy`, GitOps manifest scan, rollout plan, post-upgrade diff. |
| Anomaly: guardrails, security posture | `compliance-audit` (daily): `privileged-container`, `host-namespace`, `hostpath-mount`, `cluster-admin-binding`, `wildcard-rbac`, `netpol-missing`, `default-sa-automount`, `workload-identity-off`, `legacy-metadata`, `public-control-plane`, `podsecurity-gaps`. | Registry allowlist, Binary Authorization on live pods, secrets in env, backup plans, CA rotation, key age. |
| Anomaly: per-family drift | `fleet-consistency-drift` (weekly): per environment or project cohort, three-cluster floor. | Family grouping by label instead of environment. #911 also proposes a drift detector. |
| Anomaly: capacity guardrails | `stockout-prevention` (daily): ComputeClass fallbacks, Spot floors, `quota-exhaustion-risk`, `single-zone-nodepool`, `autoscaler-out-of-resources`. | Nothing material. |
| Anomaly: usage, zonal skew with cause, control-plane load, telemetry gaps | Nothing scheduled. `gke-cluster-autoscaler` skill is advice only. | New. #1356 is assigned for zonal skew. |
| Anomaly: costs | `fleet-wide-cost-analysis` (weekly): `overrequest`, `orphan-pv`, `unconsumed-pvc`, `unattached-disk`, `idle-address`, `orphan-lb`, `idle-nodepool`, `scaledown-blocked`, `terminal-pods`, `idle-namespace`. | Purchase-model drift, cost anomaly attribution, egress, observability spend, chargeback. |

**The fix.** In both requirements docs, open each section with one line, `Already runs as <job-id>; see <sop>.`, and keep only the delta column as the check list. Or place this table under "How it is delivered" in each. In the vehicle doc, replace "the existing governance audits" with the seven job ids above. What is left is the real proposal: deprecation insights, the add-on matrix, drain safety, zonal-skew attribution, and cost attribution.

docs/designs/capability-delivery-vehicle.md
jayantid: 🟡 **Not merge-blocking on its own, but this PR cannot land before #1368 without these docs being false.** Every present-tense "built" claim here describes #1368, which is still a draft.

Lines 66 to 69 say the store sits at `profiles/platform/capabilities/`, that `capability_criteria` is a tool on `platform_control`, and that `capability_store.py` is the implementation. Line 189 names `VOLUME_WINS_GLOBS`. Line 157 says `learning.json` ships today. The `docs/README.md` row at line 318 says "criteria store and tool built". None of it is on `main`. All of it is in #1368, whose own description says "design in #1360". Each PR presents the other as landed.

Whichever PR merges first leaves `main` describing code it does not have. The docs map is what the drift skill and later readers consult, so "built" sends them looking for a tool that is not registered. The bot's findings on lines 70, 192 and 232 are all this one item.

Fix, pick one:
- Rebase this branch onto #1368 and merge after it, keeping the present tense.
- Or reword these passages and the README status to "built in #1368, in review", and merge now.

The PR body still says `criteria.yaml`, `learning.yaml` and `MERGE_PATHS`; the document says `.json` and `VOLUME_WINS_GLOBS`. Align it with whichever you keep.

docs/designs/capability-delivery-vehicle.md
jayantid: 🟡 **Not merge-blocking, but this line stays wrong even after #1368 lands.** The store #1368 ships has no cluster exclusion list.

Line 231 says the `security-patch-orchestrator` criteria carry "a cluster exclusion list in `criteria.json`". #1368 removed that key in commit 28efced0. Its `criteria.json` holds four thresholds: `pool_skew_critical_minors`, `fleet_spread_min_minors`, `blocking_exclusion_min_days` and `deprecated_image_types`.

An operator reads this, asks the agent to exclude a cluster, and the store refuses the unknown key. The reason #1368 dropped it is the real hazard. A cluster left out of both scope lists makes the fleet-audit delta count every finding it carried as resolved, closing remediation PRs and the ledger. A design that names the exclusion list as a shipped, tunable knob invites exactly that.

Fix: drop it from this line, or move it to the "as the checks grow" sentence with its prerequisite named:

```
thresholds in `criteria.json`. A cluster exclusion list waits on a scope bucket
fleet-audit's delta understands (dropped from #1368 for that reason).
```

docs/README.md
```diff
@@ -315,11 +315,13 @@ pull request:
...
 | `docs/designs/cron-report-relay.md` | Feature design | How a scheduled job's result reaches chat with the context a follow-up question needs: the specialist reasons, hands its finished report to the Chat Agent over the Session KV server, and the Chat Agent speaks. | `deliver: "chat"`, `/v1/cron-reports`, relay turn, per-job-per-day session, `incident_context`, `report_to_chat` | Implemented; the whole Platform Agent roster delivers this way |
 | `docs/designs/design_537148738.md` | Feature design | Technical design proposal for CI/CD pipeline security scanning and dependency management (Buganizer Task 537148738). | Action SHA pinning, output encapsulation, security scanning, Dependabot | Implemented (design of record) |
 | `docs/designs/drift-detection.md` | Feature design | Design of record, not yet implemented. Splits drift into computing the diff (commoditized) and judging it (the differentiator); enters the shipped pipeline as a `gitops-drift` inject. | `managedFields` + audit-log attribution, revert-or-codify, CUJ selection | Status banner declares it unimplemented; no GitOps tool required |
```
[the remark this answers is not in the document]
haoxuw: Good catch. That table row said this design was already built, but none of it exists on `main` — it only lived in a separate branch. The row now says this is a requirements document and that nothing is implemented yet. Fixed in `8ae8472f`.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,241 @@
...
+SOP; moving onto the vehicle means lifting those into `criteria.json` and leaving the SOP as
+procedure, with the shipped default still printed beside each check.
+
+The criteria cannot live under `skills/`. The agent pod replaces that tree wholesale on every
+start, and the shell sandbox — where the agent's shell reads files — carries its own image copy
+that the mirror refuses to overwrite. So the store sits at `profiles/platform/capabilities/`, and
+the agent reads and writes it through `capability_criteria`, an in-process tool on the
+`platform_control` MCP server, never through the shell. `agents/platform/capabilities/README.md`
+is the contributor's reference; `capability_store.py` beside the server is the implementation.
+
```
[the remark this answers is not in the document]
haoxuw: You're right, and this was the root cause of most of the findings here. I had updated the design to match code that lives in a *different* pull request, so it talked about a tool and a file that don't exist on this branch.

I rewrote the document so it only says what the feature must do (requirements R1 to R7), not how it is built. It no longer names any tool, file, or module. A **Scope** table at the top says which parts already exist today — using names you can actually find on `main` — and which parts are new. Fixed in `8ae8472f`.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,241 @@
+# The capability delivery vehicle: shipped, scheduled, chat-triggered, customizable, self-learning
+
+## In short
```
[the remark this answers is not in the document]
haoxuw: Fair. The PR was marked ready while the Self-Review still said "draft". I've since run both required review passes on this branch using fresh agents that hadn't seen my reasoning — the adversarial pass once and the docs-drift pass three times, because each round changed the Scope tables and I wanted them re-checked. The PR description now lists what they found and what I did about each finding, and the Testing section has the `prettier --check` result.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,241 @@
...
+image template over the Platform Agent's profile on every pod start, `skills/` and `governance/`
+included, so a file patched with `skill_manage` reverts at the next restart and a skill directory
+the image does not ship is removed with it. That is deliberate — an upgraded pod must run the
+image's skills, and the skill provenance check at boot exists to say so — and it is exactly what a
+customization or a learned change has to survive. Hence the store: `capabilities/<name>/criteria.json`
+is the one runtime-owned file, and the scaffold merges it across a start the way it already
+merges `cron/jobs.json` but in the opposite direction (`VOLUME_WINS_GLOBS` in
+`profile_scaffold.py`): the volume wins every key it holds and the image adds only the keys a
+new release ships. The schema and `learning.json` beside it are replaced from the image, and a
+key the schema no longer defines is pruned on the next write and noted in the changelog.
```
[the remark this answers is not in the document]
haoxuw: Correct — that name doesn't exist on this branch; it came from the implementation branch. The document no longer names any code identifiers at all. Durability is now stated as a requirement (R6: a tuned value must survive a restart and an upgrade) without saying how that is done. Fixed in `8ae8472f`.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,241 @@
...
+of inline SOP thresholds into criteria. The
+on-demand `hermes cron run` gap is tracked in `agent-shell-sandboxing.md` and is not part of this
+design.
+
+## The first two capabilities on the vehicle
+
+- **Upgrade readiness** ships as a weekly job — today's `security-patch-orchestrator` audit is
+  the first capability moved onto the store, with its skew, spread, exclusion-window and image-type
+  thresholds and a cluster exclusion list in `criteria.json`. As the checks grow, the file also holds the per-family target
+  version and date, the add-on compatibility matrix, the PDB and surge rules, and which
```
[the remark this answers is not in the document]
haoxuw: Correct, that sentence wasn't true on `main`. I removed it. The document now says the two check lists are the first capabilities to be *specified* for the vehicle, and its status line says nothing is built yet. Fixed in `8ae8472f`.

docs/README.md
```diff
@@ -315,11 +315,13 @@ pull request:
 | `docs/designs/audit-logging-user-attribution.md` | Feature design | Closes the gap where audit logs identify the agent SA but not the requesting human, by carrying requester and trace/session IDs through existing telemetry. | Attribution contract per plane, correlation recipes, trust model | Draft, P0; per-plane implemented-vs-planned split declared inline |
```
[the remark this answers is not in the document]
haoxuw: Done. The Self-Review section now holds the results of the two review passes (run in fresh contexts, not by the session that wrote the docs), with a one-line outcome per finding, and the Testing section includes the prettier output.

haoxuw:
@jayantid — thanks, this was the right call. You were correct: the documents described audits that already run as if they were new work. Here is what changed, in `8ae8472f` (one commit, rebased onto `main`):

- **Every document now starts with a Scope section.** It is a three-column table: the area, what already runs on `main` today (the job id and its SOP file), and what is actually new. For example, "version posture" points at `security-patch-orchestrator` and says the only new thing is a per-family target date.
- **Where an audit already covers an area, the section below says "served today by X" and lists only the extra checks.** The SOP stays the single place that says what an existing audit checks and at what threshold, so there is no second copy that can quietly go stale.
- **Your mapping matched what two independent review passes found**, and they caught a few more I had missed: three additional audits (`ai-security-audit`, `gce-compute-fleet-audit`, `gcp-networking-fabric-audit`), reservations already covered by `stockout-prevention`, and the retired deprecation SOP. All of those are in the Scope tables now.

The documents stay requirements — what we want — rather than implementation plans — how to build it. The Scope section is where "this already exists" lives. The bot's strict `/review` is queued.

haoxuw:
@bradhoekstra done — all six bot findings are fixed in `8ae8472f`, each thread has a reply naming the fix, and `/review` is posted for the strict pass.

bradhoekstra:
I would like @fkc1e100 to review, you can add me back as an additional reviewer afterwards if you like.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,167 @@
...
+- A customized or learned value survives a pod restart and an image upgrade. An upgrade adds new
+  keys with their shipped defaults and never resets a tuned one.
+- The learning policy is not something an install can loosen and keep against the image: a
+  release that tightens a policy reaches every install.
+- Three tiers, and the agent says which one a change landed in:
+
+| Tier     | Where the change lives                                                                | Survives            | When                                                                 |
+| -------- | ------------------------------------------------------------------------------------- | ------------------- | -------------------------------------------------------------------- |
+| Session  | The scope and thresholds named in one on-request run                                  | That run            | Trying a criterion once before adopting it                           |
+| Runtime  | The capability's criteria on the profile's persistent volume                          | Restart and upgrade | The normal outcome of customization and of applied learning          |
```
jayantid: 🟡 **Not merge-blocking, but the design has to say why criteria live on a volume rather than in the operator's repository.** The architecture puts this kind of knowledge in git, and a git-backed path already ships.

`docs/architecture/02-agent-personas.md` section 2.3 places curated knowledge (SOPs, blueprints, "metric/tenancy definitions") in the operator's repository as OKF documents, humans curating as code. Thresholds, exclusions and per-family targets are that layer. Its shipped pilot is declared intent: `obtainability_audit_sop.md` section 4a reads the GitOps clone's `knowledge/` and every repository in `context_repos` to suppress a posture finding, with no kube-agents change (`fleet-audit/SKILL.md:171`). Line 29's "Already on `main`" cell omits it. Line 138 makes the volume store "the normal outcome" and leaves the Reviewed tier, the architecture's layer, as the exception.

If it merges, #1368 builds the volume store as the primary path against a recorded design that says otherwise. An operator already using declared intent gets two tuning paths that do not see each other.

Fix: credit declared intent in the Scope row, then add one paragraph under R6 saying why Runtime is the default tier and not Reviewed, or swap them.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,167 @@
...
+  report this month, and why?" has an answer with sources.
+
+| Class of change                                                                            | Default       |
+| ------------------------------------------------------------------------------------------ | ------------- |
+| Adds context: a better explanation of a finding, a link to the runbook the operator pasted | Apply, report |
+| Widens what is checked: a new check the operator asked for, a scope the report was missing | Apply, report |
+| Narrows what is flagged: raises a threshold, adds an exclusion, lowers a severity          | Propose only  |
+| Changes the procedure or a red line                                                        | Never         |
+
+Until a capability has earned otherwise, everything ships as "propose only": the agent may suggest
```
jayantid: 🟡 **Not merge-blocking, but R5 contradicts itself, and its verification row cannot pass on a correct build.**

The table at line 117 lists "Apply, report" as the default for two classes. Line 124 says everything ships "propose only". R6 at line 131 says an install cannot loosen a policy and keep it against the image. Together, every class is propose-only on every install until a release changes it. Yet line 164 expects "also flag X" to produce an applied change with a revert step, and line 125 has the agent suggest loosening a class the install cannot loosen.

Line 120 has the same shape. "A new check the operator asked for" is auto-applied, but a check is procedure (line 122, never) unless it is a declared key R7 accepts.

If it merges, #1368 and the first pilot each pick a reading, and the Verifying table fails against a build that followed the text.

Fix: state which ships, the table's defaults or propose-only everywhere with the table as the ceiling a release may raise to, and change line 164 to match. Define "new check" as toggling a shipped, declared check.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,167 @@
...
+### R5 — Self-learning
+
+- After a conversation that touched a capability — a report thread, an on-request run, a question
+  about a finding — the agent reflects on what the operator called noise, acted on, asked about
+  twice, kept narrowing to, or asked for that the report lacked, and proposes or applies a change
+  to the criteria accordingly.
+- What it may apply on its own is graded by a learning policy the operator controls, with the
+  defaults in the table below. A change that narrows what is flagged is never applied without a
+  human seeing it.
+- Every autonomous change is announced in the capability's next report — what changed, why the
```
jayantid: 🟡 **Not merge-blocking, but as written the announcement of an autonomous change is dropped on the runs where it matters most.**

Line 112 requires every autonomous change to be announced in the capability's next report. Line 72 requires a clean run to post nothing. On `main` the helper decides silence, not the agent: `audit_report.py:6593` sets `silent_ok` when nothing was new, resolved, gapped, or closed a pull request, and the scheduler skips delivery on `[SILENT]` (`agents/platform/cron/README.md:88`). A criteria change is none of those conditions.

If it merges and a widening change is applied on a quiet fleet, the next run is silent, so the announcement and its revert step never reach chat. That run also closes the ledger issue as completed, so nobody reads it there either. "Nothing moves silently" at line 113 is then false by construction.

Fix: add to R5 that an applied change makes the next run audible, one more `silent_ok` condition, and name the chat thread, not the ledger, as where the announcement lands.

docs/designs/capability-delivery-vehicle.md
jayantid, line 138: 🟡 **Technical design.** The design makes a volume-resident store the default tier for criteria, and the architecture already places that knowledge in git.

`docs/architecture/02-agent-personas.md` section 2.3 puts curated knowledge (SOPs, blueprints, metric and tenancy definitions) in the operator's repository as OKF documents. Thresholds, exclusions and per-family targets are that layer. A git-backed path already ships: `obtainability_audit_sop.md` section 4a reads `knowledge/` in the GitOps clone and every `context_repos` entry to suppress a finding (`fleet-audit/SKILL.md:171`). The Scope row at line 29 omits it, and line 138 makes the volume store "the normal outcome" with the Reviewed tier as the exception.

If it merges, #1368 builds the volume store as the primary path against a recorded design that says otherwise. An operator already using declared intent gets two tuning paths that do not see each other.

Fix: credit declared intent in the Scope row, then one paragraph under R6 on why Runtime is the default tier rather than Reviewed, or swap them.

docs/designs/capability-delivery-vehicle.md
jayantid, line 124: 🟡 **Product requirements.** R5 gives two incompatible answers to what the agent may change on its own, so two implementers will build two different products.

The table at line 117 defaults two classes to "Apply, report". Line 124 says everything ships "propose only". R6 at line 131 says an install cannot loosen a policy against the image. Read together, every class is propose-only on every install until a release changes it. Yet line 164 expects "also flag X" to produce an applied change with a revert step, and line 125 has the agent suggest loosening a class the install cannot loosen. Line 120 auto-applies "a new check the operator asked for", but a check is procedure (line 122, never) unless it is a declared key under R7.

If it merges, #1368 and the first pilot each pick a reading, and the Verifying table fails against a build that followed the text.

Fix: state which ships, the table's defaults or propose-only everywhere with the table as the ceiling a release may raise to. Change line 164 to match. Define "new check" as toggling a shipped, declared check.

docs/designs/capability-delivery-vehicle.md
jayantid, line 112: 🟡 **UX flow.** The announcement of an autonomous change rides on the next report, and the shipped silent-run rule drops that report on exactly the runs where the announcement matters.

Line 112 requires every autonomous change to be announced in the capability's next report. Line 72 requires a clean run to post nothing. On `main` the helper decides silence, not the agent: `audit_report.py:6593` sets `silent_ok` when nothing was new, resolved, gapped, or closed a pull request, and the scheduler skips delivery on `[SILENT]` (`agents/platform/cron/README.md:88`). A criteria change is none of those conditions.

If it merges and a widening change is applied on a quiet fleet, the next run is silent, so the announcement and its revert step never reach chat. The same run closes the ledger issue as completed, so nobody reads it there either. "Nothing moves silently" at line 113 is false by construction.

Fix: add to R5 that an applied change makes the next run audible, one more `silent_ok` condition, and name the chat thread rather than the ledger as where the announcement lands.

jayantid (changes requested):

docs/designs/upgrade-readiness-checks.md
```diff
@@ -0,0 +1,108 @@
...
+A platform team running GKE at scale — millions of cores, dozens of cluster families (fleets of
+clusters built to one template), and multi-tenant clusters holding thousands of namespaces each —
+assesses readiness for every Kubernetes minor-version bump by hand. Several tools are stitched
+together to find deprecated API callers, check add-on compatibility, and confirm workloads will
+survive a node drain, and the exercise is repeated per cluster family.
+
+This document lists the checks such an operator would realistically ask the Platform Agent to run on
+a schedule ahead of an upgrade, as candidate criteria for a scheduled audit. Each check is
+read-only: it inspects the fleet and reports a finding, with an owner and a run-over-run delta,
+and never upgrades a cluster.
```
fkc1e100, line 33: From a product perspective, the customer's core job-to-be-done is compressing the manual toil of evaluating fleet upgrade readiness. Cataloging candidate checks across five sections creates a helpful audit inventory, but leaves the customer with the exact operational friction they have today: manually triaging hundreds of unranked findings to decide whether an upgrade can proceed.

To solve the customer's pain point, this feature needs to deliver an **Authoritative Upgrade Gate** per cluster family rather than a raw inventory of findings:
1. **Triaged Readiness Tiers:** Explicitly classify criteria into **Release Blockers** (active auto-upgrade pauses, non-evictable PDBs, zero-surge quotas), **Operational Risks** (add-ons lagging target support), and **Advisories** (stale annotations).
2. **Decision-Ready Synthesis:** The output should be a clear, machine-readable verdict per cluster family (`Go`, `No-Go`, `Action Required`) that platform leads can use directly as a deployment gate in their release pipeline.

docs/designs/fleet-anomaly-detection-checks.md
```diff
@@ -0,0 +1,125 @@
...
+How the fleet is actually being used, and where a family, cluster, or zone has drifted from its
+own norm. Most of these need a metrics source — Cloud Monitoring, managed Prometheus, or VPA
+recommendations — that the scheduled audits currently do without; the source is part of the
+requirement.
+
+- **Zonal and topology balance.** Node and pod distribution per zone per cluster, with the reason
+  once skew crosses a threshold: a stockout (autoscaler out-of-resources events), a single-zone
+  node pool, a misconfigured `topologySpreadConstraints`, or zone-pinned PersistentVolumes
+  dragging pods into one zone. Zonal capacity headroom (autoscaler maximum against current size
+  per zone) and load-balancer backend distribution per zone, for single-zone blast radius.
```
fkc1e100, line 89: The customer shared a concrete failure mode: an undetected zonal skew resulted in a regional outage. Treating zonal skew as a generic bullet point in a weekly cron job reporting week-over-week deltas does not solve this problem. Zonal skew is an acute reliability event that develops rapidly following a capacity stockout, node maintenance, or a bad rollout. A weekly report is a post-mortem, not prevention.

We should elevate **Zonal and Topology Balance** to a primary operational monitor:
1. **Actionable Cadence:** Evaluate on an active cadence (daily or event-triggered upon autoscaler errors) to detect skew before it becomes an availability incident.
2. **Root-Cause Attribution:** Automatically classify the business driver: Cloud capacity stockout (`out.of.resources`), StatefulSet disk pinning, or deficient scheduling constraints (`ScheduleAnyway`).
3. **Blast Radius Impact:** Quantify the single-zone outage exposure and cross-zone networking cost implications for platform leaders.

docs/designs/capability-delivery-vehicle.md
```diff
@@ -0,0 +1,167 @@
...
+- The learning policy is not something an install can loosen and keep against the image: a
+  release that tightens a policy reaches every install.
+- Three tiers, and the agent says which one a change landed in:
+
+| Tier     | Where the change lives                                                                | Survives            | When                                                                 |
+| -------- | ------------------------------------------------------------------------------------- | ------------------- | -------------------------------------------------------------------- |
+| Session  | The scope and thresholds named in one on-request run                                  | That run            | Trying a criterion once before adopting it                           |
+| Runtime  | The capability's criteria on the profile's persistent volume                          | Restart and upgrade | The normal outcome of customization and of applied learning          |
+| Reviewed | A pull request to a configuration repository the operator keeps, applied at pod start | Everything          | A change the operator wants reviewed, or any change to the procedure |
```
fkc1e100, line 139: Establishing a `Runtime` tier that persists operational thresholds directly to a local PersistentVolume bypasses GitOps—the foundational operating model for enterprise cloud platforms.

In an enterprise environment:
1. Criteria stored on local PVCs cannot be peer-reviewed, tracked in audit logs, or rolled back cleanly via version control.
2. State residing on local storage inevitably drifts across multi-replica or multi-region agent deployments.

For an enterprise-grade capability, the **Reviewed** tier (GitOps Pull Request) must be the primary persistence mechanism. When an operator agrees to customize a threshold or exclude a scope in chat, the agent's action should be opening a structured Pull Request against the fleet configuration repository, preserving full auditability, peer review, and disaster recovery parity.

fkc1e100 (commented):
Product Management review evaluating these design requirements against enterprise customer operational feedback (ALF scale: millions of cores, dozens of cluster families, multi-tenant clusters).

Key feedback focuses on:
1. Transitioning upgrade readiness from an unranked audit backlog to an authoritative Go/No-Go release gate.
2. Elevating zonal skew from a passive weekly report to an active reliability monitor to prevent regional outages.
3. Enforcing GitOps as the sole persistence tier for criteria customization to preserve enterprise governance.

docs/designs/upgrade-readiness-checks.md
haoxuw: Will keep this in mind when implementing!

docs/designs/fleet-anomaly-detection-checks.md
haoxuw: Will build a testable cluster with those error signals, both during development and as e2e test

mastersingh24: 100% agree.  This is big enough for that and actually can be pretty involved.  I created a design for this cases specifically as part of k8s-lookout:  https://github.com/go-steer/k8s-lookout/blob/main/docs/leeway-design.md

docs/designs/capability-delivery-vehicle.md
mastersingh24, line 1: I would actually do this as it's own PR ... seems like it's setting up some groundwork for the other designs.  We should agree on this first (or agree on the other designs separately and then call this out as a delivery vehicle in a subsequent PR).  I think having this plus the two other designs might be a bit too much?

mastersingh24 (changes requested):

docs/designs/capability-delivery-vehicle.md
haoxuw: Great feedback, as discussed offline this is out of scope, we will defer the change, and use the current Hermes' learning capabilities via PCV

docs/designs/capability-delivery-vehicle.md
haoxuw: Hi same as the reply to Jayanti above:

We had a discussion offline, the move to GitOps seems out of scope. There are several other engineers working on foundational work. Until that's done, I think we have to defer the change, and use the current Hermes' learning capabilities via PCV. We may refactor later

docs/designs/capability-delivery-vehicle.md
haoxuw: You're right that the architecture puts this knowledge in git, and we're deferring the move rather than disputing it. `eca6b66e` makes the doc say that out loud: the Scope table now credits the declared-intent step as the git-backed path that exists today, and R6 has a paragraph recording why Runtime is the v1 default — the GitOps overlay depends on foundational work that's in flight separately — with the changelog shaped so tuned values can later be harvested into one reviewed PR. Resolving as recorded-and-deferred; reopen if the paragraph doesn't capture it.

docs/designs/capability-delivery-vehicle.md
haoxuw: Fixed in `eca6b66e`, with one answer instead of two: everything ships as propose-only, on every install. The class table is the ceiling a release may raise a class to once a capability has earned it — and raising is a reviewed change to the shipped policy, never a runtime toggle, so it no longer conflicts with R6. "A new check" is now defined as turning on or extending something the capability already declares as tunable; anything needing new procedure stays "never". The verification row matches: "also flag X" produces a proposal first, applied and announced after confirmation.

docs/designs/capability-delivery-vehicle.md
haoxuw: Good catch — the silence rule really would have swallowed the announcement. Fixed in `eca6b66e`: R5 now says an applied change cancels the next run's silence ("first run after a criteria change" becomes one more condition next to the existing silent-run ones), and the announcement goes to the capability's chat thread through the relay rather than the ledger, since the same run may close the ledger issue. R2's "a clean run posts nothing" now carries that one exception.

docs/designs/upgrade-readiness-checks.md
haoxuw: Promoted from "keep in mind" to a requirement in `eca6b66e`. The upgrade doc now has a "The verdict" section: every check declares a tier — Blocker, Risk, or Advisory — and each run emits a machine-readable per-family verdict (Go / No-Go / Action required) naming the blockers, so a platform lead can wire it into a release pipeline as a gate. The finding list becomes the evidence behind the verdict, not the deliverable.

docs/designs/fleet-anomaly-detection-checks.md
haoxuw: Agreed, and it's a requirement now (`eca6b66e`): zonal balance runs daily and is re-evaluated when an autoscaler out-of-resources event arrives — which is exactly the event path the vehicle doc defines, so the two documents now reinforce each other. The cause list covers capacity stockout, disk-pinned StatefulSets, and `ScheduleAnyway`, and every finding must quantify blast radius: the share of the workload lost if the crowded zone fails, plus the cross-zone traffic the imbalance pays for. @mastersingh24 thanks for the leeway design — I'd like to align the attribution taxonomy with it when this reaches an implementation issue.

docs/designs/capability-delivery-vehicle.md
haoxuw: Following up on the deferral with the doc change it needed (`eca6b66e`): R6 now records the decision explicitly — Runtime on the volume is the v1 default because the GitOps overlay depends on foundational work in flight; the costs you named (no peer review, invisible to version control) are stated, and the changelog is shaped so accumulated changes can be harvested into one reviewed PR when that foundation lands. Reviewed stays the ceiling, and the doc doesn't rule out flipping the default later.

docs/designs/capability-delivery-vehicle.md
haoxuw: Fair point on sequencing. Two practical notes: the two check docs link to the vehicle doc (the link checker fails if it's absent), and both reviewers have already read all three here — the remaining deltas just landed in `eca6b66e`. So my preference is to finish this round as one PR. If you'd still rather split after reading the fixes, I'll pull the vehicle doc into its own PR and soften the links — say the word.

docs/designs/upgrade-readiness-checks.md
```diff
@@ -0,0 +1,119 @@
...
+| Rollout orchestration and verification | The `fleet-upgrade-verification` skill reports, run over run, which clusters started, completed or stalled against one target version.                                                                                                                                                                                                                                                                                         | Canary sequencing per family with each family's blocking findings, and the post-upgrade diff.                                                                                                                          |
```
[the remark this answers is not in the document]
haoxuw: Confirmed and fixed in `cedea86b` — the scan does ship and my Scope row was stale. The row now credits `api_deprecation_scan.py` for what it does (scans every managed GitOps repository's manifests for apiVersions the target removes, with each hit's replacement, on request), and the body bullet says the same: that half is done, what's new is the rest of the compatibility surface (add-on matrix, webhook posture, image assumptions) and running it on the audit's schedule instead of only when asked.

docs/README.md
```diff
@@ -344,10 +344,12 @@ pull request:
...
```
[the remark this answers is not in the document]
haoxuw: Confirmed and fixed in `cedea86b`. The drain-safety row now credits `--readiness` for grading drain-blocking PDBs, upgrade-covering maintenance exclusions and node-pool skew per member as `blocked`/`unknown`/`ready`, and the verdict section says plainly that this is the nearest thing on `main` — what the document adds is aggregation per cluster family rather than per member, every check here tiered into the verdict rather than that script's three inputs, and the verdict coming from the scheduled audit rather than on request.

haoxuw:
Both low-severity items from the strict pass are fixed in `cedea86b` too:

- The `silent_ok` description now names the conditions the helper actually uses — nothing new or resolved, no coverage gap, no remediation pull request **opened or closed**.
- Drift detection is now described as partly built (the Pub/Sub sink and detector command exist; the inject it would emit does not), matching that design's own status banner instead of contradicting it.

The branch is also rebased onto current `main` — it had gone conflicting on the docs map again.

docs/designs/capability-delivery-vehicle.md
haoxuw: Done — split as you asked, in `d461b240`.

- **This PR is now the delivery vehicle alone**: `docs/designs/capability-delivery-vehicle.md` plus its map row, two files total. It states requirements R1–R7 and how each is verified, with no implementation detail.
- **The two check documents moved to #1778**, which references the vehicle by path rather than by link so the link checker stays green until this one merges. All the review already spent on them travelled with them, and the PR body lists it.

So the sequence you suggested now holds: agree the vehicle here, and #1778 stands on its own as the capability requirements. Thanks for pushing on it — the two read better apart than they did together.

jayantid (approved):

PR #1368, opened 2026-09-09
feat(platform): runtime-owned capability criteria with a policy-gated MCP tool

haoxuw:
## Summary

First piece of the capability delivery vehicle (design in #1360): the thresholds, exclusions and scopes an audit reads now live outside the image-owned procedure, in a per-capability store the agent reads and edits through a new `capability_criteria` MCP tool. A change an operator agrees with the agent in chat survives a pod restart and an image upgrade, and the store refuses writes the capability's learning policy does not allow. `security-patch-orchestrator` is the first audit moved onto it.

## Why This Change

Today every threshold lives inline in a governance SOP or a `SKILL.md`, and both trees are replaced from the image on every pod start — on the agent pod by `sync_profile_skills`, and in the shell sandbox by the sandbox image itself, which `sandbox_mirror.py` refuses to overwrite. So "make the audit stop flagging X" is a pull request against this repository, and anything the agent changes with `skill_manage` reverts at the next restart. Fleet operators asking for scheduled checks they can tune without a redeploy need a place for tunable values that is neither of those trees, a merge rule that keeps the tuned value across an upgrade, and a write path that cannot silently suppress a finding.

## What Changed

- `agents/platform/capabilities/<name>/` — `criteria.json` (shipped defaults), `criteria.schema.json` (image-owned; keys, types, defaults, descriptions), `learning.json` (per-key policy: `autonomous` | `propose` | `never`, everything shipped as `propose`). README explains ownership per file and why the tree is not under `skills/`.
- `deploy/docker/Dockerfile` bakes the tree into `/opt/platform-template/capabilities/`; `docker-entrypoint.sh` adds `capabilities` to `platform_sync_items`; `profile_scaffold.py` gains `VOLUME_WINS_GLOBS` and `merge_volume_wins` — the inverse of `merge_cron_store`: the volume wins every key it holds, the image adds only new keys, the schema is replaced, `changelog.jsonl` is never touched.
- `agents/platform/scripts/capability_store.py` — validation (a stdlib subset of JSON Schema), policy enforcement, atomic write, revision bump, changelog append. `platform_mcp_server.py` exposes it as `capability_criteria(action=list|get|set|history)`; a `propose` key is refused without `confirmed_by`, a `never` key is refused outright, an unknown key is reported as unknown.
- `security_patch_orchestrator_sop.md` reads `pool_skew_critical_minors`, `fleet_spread_min_minors`, `blocking_exclusion_min_days`, `deprecated_image_types` and `excluded_clusters` from the tool, with the shipped defaults still printed beside each check; `cron/jobs.json` re-cites the SOP's geography (291 lines, §3 at 121–225).
- `agents/platform/AGENTS.md` routes "stop flagging / also flag / skip cluster Y" requests through the tool's propose-then-confirm flow instead of SOP edits. `sandbox_mirror.py` withholds `capabilities/` from the sandbox. `scripts/check_prompt_assets.py` learns the new profile-home entry; the manual-install doc's directory tree gains the line.

## Context

Design: #1360 (`docs/designs/capability-delivery-vehicle.md`). That doc currently places the criteria under `skills/<name>/criteria.yaml`; building it showed that cannot work (both trees are image-replaced, see Why), so this PR uses `capabilities/<name>/criteria.json` and the design doc will be corrected to match. Not in this PR: the post-conversation reflection job (self-learning), the ConfigMap overlay for reviewed overrides, and the second capability (fleet anomaly detection). No open issue matches; #589 touches the same SOP family but not this mechanism.

## Testing

- `agents/platform/scripts`: 137 tests (new `test_capability_store.py` — 16; new `TestCapabilityCriteriaTool` and `CapabilityCriteriaMergeTest` classes) — pass.
- `agents/platform/skills/fleet-audit/scripts/test_audit_report.py` (569, including the SOP-geography and check-roster tests against the edited SOP) — pass.
- `tests/test_docker_entrypoint.py`, `tests/test_sandbox_mirror.py` (124), `deploy/docker/test_image_startup_contract.py` (37), `scripts/` (1,107, including `test_check_prompt_assets.py`) — pass.
- `make docs-check`, `python3 scripts/check_prompt_assets.py`, `prettier --check` on every edited Markdown and JSON file — pass.
- Image layer budget: `make docker-build-platform` + `scripts/check_image_layers.py` — running at the time of opening; result to be recorded here.

### Live validation

**Exercised against a live agent pod's PVC on 2026-09-21** (`haoxuw-gke-dev`, `platform-agent-gateway`, container `platform-agent`, `HERMES_HOME=/opt/data`). `capability_store.py` and `profile_scaffold.py` were copied into the pod with the shipped capability template, and run against a scratch profile home on the real volume — nine checks, all passing:

1. A first scaffold seeds the defaults from the image template and `get` reads them back at revision 0.
2. Every shipped key's policy is `propose`.
3. A write to a `propose` key without `confirmed_by` is refused, naming the field.
4. The same write with a confirmer is applied, bumps the revision, and appends a changelog entry with the before/after.
5. A value outside the image-owned schema's bounds is refused as invalid.
6. **Durability:** re-applying the image template — what a pod restart does — leaves the tuned value and its revision intact.
7. **Upgrade:** a template that ships a new key adds it with its default without resetting the tuned one.
8. The changelog survives both re-applies and names the confirmer.
9. `learning.json` is replaced from the image by the merge, so an install cannot keep a loosened policy.

The changelog line and final `criteria.json` were read back off the volume. Scratch directory removed afterwards.

**What this does not cover, and why.** The full path through the agent — `capability_criteria` as an MCP tool in a real turn, and an audit run reading its criteria against the seeded fleet — needs an image built from this branch and a working install. Two blockers there, both independent of this change: the dev install's credential proxy is crash-looping on a bootstrap failure (588 restarts), and the dev project carries only seeded slot `c`, so no case asserting on the version laggard can run. Tracked in #1815. The plan, to be run before this leaves draft: fresh pod start → `capability_criteria(action="get", capability="security-patch-orchestrator")` returns the shipped defaults; `set` with `pool_skew_critical_minors: 4` and a `confirmed_by` → `hermes cron run security-patch-orchestrator` → ledger summary cites revision 1 and a pool exactly 3 minors behind is reported major, not critical; `kubectl rollout restart` the agent → `get` still returns 4 with revision 1; roll a new image tag → same, and a newly shipped key appears with its default; revert to 3 and confirm the finding returns to critical.

## Self-Review

The adversarial and docs-drift passes are running in clean subagent contexts (`/pr-preflight`) at the time of opening; this section will be filled in with what they looked for and the disposition of each finding before the PR is marked ready. Checked by the authoring session so far: every path claim traced to source (`sync_profile_skills` replaces `skills/` wholesale; `sandbox_mirror.IMAGE_OWNED` covers `skills`/`governance`/`scripts`; `agents/platform/scripts/` lands in `/opt/defaults/scripts/`, so the MCP server's same-directory import of `capability_store` resolves; the MCP server runs with `HERMES_HOME` at the profile home, so `capabilities_root()` resolves to `profiles/platform/capabilities`).

## Risk & Rollout

Three runtime paths change. (1) The entrypoint force-syncs one more directory; if the template lacks it the scaffold skips it, as for any absent item. (2) `overlay_template` runs a new merge pass; a merge failure logs a WARN and leaves the image defaults in place, the same degradation `cron/jobs.json` has. (3) `platform_mcp_server.py` imports `capability_store` at module load; a missing module would fail the whole `platform_control` server, which is why the two ship from the same `COPY`. The SOP change only redirects four literals to the tool and prints the defaults beside them, so a tool error leaves the audit running on today's values. Revert by reverting the commit; a volume that already holds `capabilities/` keeps an inert directory.


fkc1e100:
**Triage note:** This PR has been in draft state for over a week. Please either mark as ready for review or close if no longer needed.

Issue #1374, opened 2026-09-10
feat: Support Red Hat OpenShift on Google Compute Engine (GCE)


fkc1e100:
### Summary

Enable `kube-agents` to be installed and operated on Red Hat OpenShift clusters running on Google Compute Engine (GCE), and provide node obtainability orchestration for OpenShift MachineSets during GCE regional stockouts.

### Motivation

While `kube-agents` runs on standard Kubernetes primitives, deploying onto OpenShift on GCE currently encounters several platform-specific barriers:
1. **Cluster DNS Resolution:** OpenShift hosts CoreDNS in `openshift-dns/dns-default` rather than `kube-system/kube-dns`. Egress NetworkPolicies must permit traffic to OpenShift DNS daemonsets.
2. **SecurityContextConstraints (SCC):** OpenShift enforces `restricted-v2` SCC by default, which rejects pods specifying fixed UIDs (such as the unprivileged agent UID `10000`).
3. **Ingress & Routing:** OpenShift uses native `Route` objects (`route.openshift.io/v1`) for external traffic ingress.
4. **CLI Tooling:** Administrators and agents interacting with OpenShift clusters require the `oc` CLI alongside `kubectl`.
5. **Node Obtainability:** OpenShift on GCP manages compute via zonal `MachineSet` resources (`machine.openshift.io/v1beta1`) with `GCPMachineProviderSpec`. When a zone or machine family experiences a stockout (`ZONE_RESOURCE_POOL_EXHAUSTED` or `QUOTA_EXCEEDED`), `kube-agents` can proactively detect the bottleneck, consult GCE `AdviceService.Capacity`, and fail over to healthy zones and machine families.

---

### Implementation & PR Mapping Matrix

Each pull request has a dedicated feature issue providing in-depth technical guidance on why the capability is needed, the platform-level failure modes on OpenShift, and implementation details:

| Phase / Area | Pull Request | Dedicated Feature Issue ("Why This Is Needed") | Scope & Architectural Role | Review Status |
|:---|:---|:---|:---|:---|
| **Phase 1 (Operator)** | [#1375](https://github.com/gke-labs/kube-agents/pull/1375) | [#1420](https://github.com/gke-labs/kube-agents/issues/1420) | In-cluster CoreDNS discovery (`openshift-dns/dns-default`) & egress NetworkPolicy allowlists. | Reviewed (✅ No findings) |
[...]

PR #1375, opened 2026-09-10
feat(operator): add OpenShift cluster DNS discovery and egress policy support

### LINKED ISSUE #1420: feat(operator): add in-cluster CoreDNS discovery and egress policy support for OpenShift on GCE
> **Overarching Initiative:** [#1374: `feat: Support Red Hat OpenShift on Google Compute Engine (GCE)`](https://github.com/gke-labs/kube-agents/issues/1374)  
> **Associated Pull Request:** [#1375: `feat(operator): add in-cluster CoreDNS discovery for OpenShift on GCE`](https://github.com/gke-labs/kube-agents/pull/1375)

---

### Summary
Enable the `k8s-operator` to dynamically discover cluster CoreDNS on Red Hat OpenShift on Google Compute Engine (GCE) and synthesize compliant NetworkPolicy egress rules for the Platform Agent gateway, shell sandbox, and credential proxy.

---

### Why This Is Needed

On standard Google Kubernetes Engine (GKE) clusters:
1. Cluster DNS runs in namespace `kube-system` under Service `kube-dns`.
2. The operator's default NetworkPolicy engine builds `clusterDNSPeers` targeting pods in `kube-system` matching `k8s-app: kube-dns`.

On Red Hat OpenShift on GCE:
1. **Namespacing & Label Divergence:** The OpenShift DNS Operator manages cluster DNS. It deploys CoreDNS daemonsets into namespace `openshift-dns` with label `dns.operator.openshift.io/daemonset-dns: default`, exposing the ClusterIP Service `dns-default` (typically `172.30.0.10`).
2. **OVN-Kubernetes Default-Deny Drops:** OpenShift clusters use OVN-Kubernetes as the default CNI plugin. Under default-deny egress NetworkPolicies, pod network egress to port 53 is blocked unless an explicit `NetworkPolicy` rule permits egress to the DNS pods. Because the existing policy only permits egress to `kube-system/kube-dns`, OVN-Kubernetes silently drops all DNS queries originating from `platform-agent-gateway`, `platform-agent-shell-0`, and `platform-agent-credential-proxy`.
3. **Complete Agent Isolation:** Without this discovery mechanism, the Platform Agent cannot resolve internal Kubernetes Services (e.g. `litellm.kubeagents-system.svc.cluster.local`) or external endpoints (e.g. Google Cloud APIs for Vertex AI, GCS, or GitHub APIs), causing immediate startup failures and timeout loops.
4. **Anti-Flap Stability:** NetworkPolicy reconciliation runs continuously. If a transient API glitch or network blip occurs while querying the OpenShift DNS Service, the controller must not flap back to a static default IP, which would trigger disruptive NetworkPolicy churn and sever active agent connections.

---

### Technical Solution & Implementation

1. **In-Cluster Discovery Probe:**
   - Updated `k8s-operator/internal/controller/netpolprofile.go` with named constants:
     - `openShiftDNSNamespace = "openshift-dns"`
     - `openShiftDNSServiceName = "dns-default"`
     - `openShiftDNSDaemonSetLabelKey = "dns.operator.openshift.io/daemonset-dns"`
     - `openShiftDNSDaemonSetLabelVal = "default"`
   - Implemented fallback discovery: if `kube-system/kube-dns` is absent, query `openshift-dns/dns-default`.
   - Added anti-flap logic to preserve previously discovered status during transient API read errors.
2. **NetworkPolicy Egress Peer Allowlist:**
   - Updated `clusterDNSPeers` in `k8s-operator/internal/controller/platformagent_manifests.go` to inject `openshift-dns` namespace and daemonset pod selectors into both the gateway policy (`buildNetworkPolicy`) and shell sandbox policy (`buildShellSandboxNetworkPolicy`).
   - Updated `buildAgentEgressNetworkPolicy` in `platformagent_egress_policy.go` to include `openshift-dns` pod peers in the egress allowlist.

---

### Live Cluster Validation

- **Environment:** OKD v1.35.5 (Kubernetes v1.35.5, OVN-Kubernetes) on Google Compute Engine in `us-central1` (GCP project `gca-gke-2025`).
- **Discovery Verified:** Operator detected `openshift-dns/dns-default` Service ClusterIP `172.30.0.10`.
- **NetworkPolicy Verified:** Verified `clusterDNSPeers` rendered `openshift-dns` pod peers into `platform-agent-gateway-netpol` and `platform-agent-shell-netpol`.
- **Resolution Verified:** In-pod DNS resolution tests confirmed successful A/AAAA resolution from `platform-agent-shell-0` to internal Services (`litellm.kubeagents-system.svc.cluster.local`) and external endpoints over OVN-Kubernetes.

fkc1e100:
## Summary

Enables the `k8s-operator` to discover cluster DNS on Red Hat OpenShift clusters and permits DNS resolution traffic across all operator-managed NetworkPolicies (gateway, shell sandbox, and egress allowlist).

## Why This Change

On standard GKE clusters, CoreDNS runs in `kube-system/kube-dns`. On Red Hat OpenShift, cluster DNS is managed by the OpenShift DNS Operator, which exposes the `dns-default` Service in the `openshift-dns` namespace and runs CoreDNS daemonsets with label `dns.operator.openshift.io/daemonset-dns: default`. Without this change, deploying `kube-agents` on OpenShift falls back to the static default DNS IP, and default-deny egress NetworkPolicies (`clusterDNSPeers` in gateway and shell sandbox, plus egress allowlist) drop all DNS traffic under OVN-Kubernetes, rendering the agent, shell sandbox, and credential proxy unable to resolve cluster services or external endpoints.

## What Changed

- `k8s-operator/internal/controller/netpolprofile.go`: Added named constants (`openShiftDNSNamespace`, `openShiftDNSServiceName`, `openShiftDNSDaemonSetLabelKey`, `openShiftDNSDaemonSetLabelVal`) and in-cluster discovery probe for `openshift-dns/dns-default` when `kube-system/kube-dns` is absent, including anti-flap protection preserving status on transient errors.
- `k8s-operator/internal/controller/platformagent_manifests.go`: Added OpenShift DNS peers to `clusterDNSPeers` helper so both the gateway NetworkPolicy (`buildNetworkPolicy`) and shell sandbox NetworkPolicy (`buildShellSandboxNetworkPolicy`) permit egress to `openshift-dns` CoreDNS pods.
- `k8s-operator/internal/controller/platformagent_egress_policy.go`: Added OpenShift DNS pod peer to `buildAgentEgressNetworkPolicy` allowlist.
- `k8s-operator/internal/controller/netpolprofile_test.go`: Added unit tests `TestResolveNetpolProfile/DiscoveryOpenShiftDNS` (happy path discovery) and `DiscoveryOpenShiftDNS_TransientError_PreservesDiscoveredStatus` (anti-flap validation via `openShiftDNSGetFails` interceptor).
- `k8s-operator/internal/controller/platformagent_egress_policy_test.go`: Updated allowlist verification tests to cover `openshift-dns`.

## Context

Closes #1420. Part of the phased support for Red Hat OpenShift on Google Compute Engine tracked in overarching issue #1374. Part of the phased support for Red Hat OpenShift on Google Compute Engine (Phase 1).

## Testing

- Unit tests:
  `cd k8s-operator && go test -v ./internal/controller -run "TestResolveNetpolProfile|TestTheRenderedPolicy"` (all 26 subtests PASS).

### Live validation

Validated against a live Red Hat OpenShift 4.22 / OKD cluster running on Google Compute Engine (`v1.35.5`, CentOS Stream CoreOS 10 kernel 6.12, OVN-Kubernetes):
1. Verified in-cluster discovery of `openshift-dns/dns-default` Service ClusterIP (`172.30.0.10`).
2. Verified `clusterDNSPeers` renders the `openshift-dns` namespace and pod selector into the shell sandbox and gateway NetworkPolicies.
3. Verified name resolution from the shell sandbox (`platform-agent-shell-0`) and gateway pod to cluster services (`litellm.kubeagents-system.svc.cluster.local`) and external endpoints over OVN-Kubernetes.

## Self-Review

- Adversarial review (clean context): Checked whether default-deny rules in `clusterDNSPeers` were missing OpenShift DNS pod peers in addition to `buildAgentEgressNetworkPolicy`. Identified that gateway and shell sandbox policies both rely on `clusterDNSPeers`; updated `clusterDNSPeers` so both policies permit OpenShift DNS.
- Coding standards (AGENTS.md): Eliminated string literals in favour of declared top-of-file constants (`openShiftDNSNamespace`, `openShiftDNSServiceName`, `openShiftDNSDaemonSetLabelKey`, `openShiftDNSDaemonSetLabelVal`).
- Verified anti-flap behavior: Added explicit `DiscoveryOpenShiftDNS_TransientError_PreservesDiscoveredStatus` test with `openShiftDNSGetFails` interceptor to confirm transient API errors preserve previously discovered status.
- Docs-drift: Verified DNS discovery order matches operator documentation.

## Risk & Rollout

Low risk. Only introduces an additional fallback discovery branch and adds an additional allowed DNS pod peer in the NetworkPolicy; existing GKE deployments are unaffected. Revertible by reverting this commit.

### DIFF
--- a/k8s-operator/internal/controller/platformagent_egress_policy.go
+++ b/k8s-operator/internal/controller/platformagent_egress_policy.go
@@ -285,6 +285,7 @@ func buildAgentEgressNetworkPolicy(agent *agentv1alpha1.PlatformAgent, dnsCluste
 	dnsPeers := []networkingv1.NetworkPolicyPeer{
 		namespacedPodPeer("kube-system", map[string]string{"k8s-app": "kube-dns"}),
 		namespacedPodPeer("kube-system", map[string]string{"k8s-app": "node-local-dns"}),
+		namespacedPodPeer("openshift-dns", map[string]string{"dns.operator.openshift.io/daemonset-dns": "default"}),
 		{IPBlock: &networkingv1.IPBlock{CIDR: nodeLocalDNSCacheIP}},
 		{IPBlock: &networkingv1.IPBlock{CIDR: metadataResolverCIDR}},
 	}
--- a/k8s-operator/internal/controller/platformagent_egress_policy.go
+++ b/k8s-operator/internal/controller/platformagent_egress_policy.go
@@ -285,7 +285,7 @@ func buildAgentEgressNetworkPolicy(agent *agentv1alpha1.PlatformAgent, dnsCluste
 	dnsPeers := []networkingv1.NetworkPolicyPeer{
 		namespacedPodPeer("kube-system", map[string]string{"k8s-app": "kube-dns"}),
 		namespacedPodPeer("kube-system", map[string]string{"k8s-app": "node-local-dns"}),
-		namespacedPodPeer("openshift-dns", map[string]string{"dns.operator.openshift.io/daemonset-dns": "default"}),
+		namespacedPodPeer(openShiftDNSNamespace, map[string]string{openShiftDNSDaemonSetLabelKey: openShiftDNSDaemonSetLabelVal}),
 		{IPBlock: &networkingv1.IPBlock{CIDR: nodeLocalDNSCacheIP}},
 		{IPBlock: &networkingv1.IPBlock{CIDR: metadataResolverCIDR}},
 	}
--- a/k8s-operator/internal/controller/platformagent_manifests.go
+++ b/k8s-operator/internal/controller/platformagent_manifests.go
@@ -4526,6 +4526,18 @@ func clusterDNSPeers(dnsIPs []string) []networkingv1.NetworkPolicyPeer {
 			},
 		},
 		{
+			NamespaceSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"kubernetes.io/metadata.name": openShiftDNSNamespace,
+				},
+			},
+			PodSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					openShiftDNSDaemonSetLabelKey: openShiftDNSDaemonSetLabelVal,
+				},
+			},
+		},
+		{
 			IPBlock: &networkingv1.IPBlock{
 				CIDR: nodeLocalDNSCacheIP,
 			},
--- a/k8s-operator/internal/controller/netpolprofile.go
+++ b/k8s-operator/internal/controller/netpolprofile.go
@@ -52,6 +52,13 @@ const (
 	metadataDaemonPortName    = "metadata-server"
 	metadataDaemonDefaultPort int32 = 988
 
+	// openShiftDNSNamespace and openShiftDNSServiceName locate the OpenShift CoreDNS Service
+	// when kube-system/kube-dns is absent.
+	openShiftDNSNamespace         = "openshift-dns"
+	openShiftDNSServiceName       = "dns-default"
+	openShiftDNSDaemonSetLabelKey = "dns.operator.openshift.io/daemonset-dns"
+	openShiftDNSDaemonSetLabelVal = "default"
+
 	// Source constants reporting how the network policy values were chosen.
 	netpolSourceSpec        = "Spec"
 	netpolSourceAnnotation  = "Annotation"
@@ -159,7 +166,7 @@ func (r *PlatformAgentReconciler) resolveNetpolProfile(ctx context.Context, agen
 	// 4a. In-cluster discovery from OpenShift openshift-dns/dns-default Service
 	if len(p.DNSClusterIPs) == 0 {
 		var ocpSvc corev1.Service
-		if err := r.Get(ctx, types.NamespacedName{Namespace: "openshift-dns", Name: "dns-default"}, &ocpSvc); err == nil {
+		if err := r.Get(ctx, types.NamespacedName{Namespace: openShiftDNSNamespace, Name: openShiftDNSServiceName}, &ocpSvc); err == nil {
 			var discovered []string
 			if len(ocpSvc.Spec.ClusterIPs) > 0 {
 				for _, ip := range ocpSvc.Spec.ClusterIPs {
--- a/k8s-operator/internal/controller/netpolprofile.go
+++ b/k8s-operator/internal/controller/netpolprofile.go
@@ -156,6 +156,35 @@ func (r *PlatformAgentReconciler) resolveNetpolProfile(ctx context.Context, agen
 		}
 	}
 
+	// 4a. In-cluster discovery from OpenShift openshift-dns/dns-default Service
+	if len(p.DNSClusterIPs) == 0 {
+		var ocpSvc corev1.Service
+		if err := r.Get(ctx, types.NamespacedName{Namespace: "openshift-dns", Name: "dns-default"}, &ocpSvc); err == nil {
+			var discovered []string
+			if len(ocpSvc.Spec.ClusterIPs) > 0 {
+				for _, ip := range ocpSvc.Spec.ClusterIPs {
+					trimmed := strings.TrimSpace(ip)
+					if trimmed != "" && trimmed != "None" && net.ParseIP(trimmed) != nil {
+						discovered = append(discovered, trimmed)
+					}
+				}
+			} else if ip := strings.TrimSpace(ocpSvc.Spec.ClusterIP); ip != "" && ip != "None" && net.ParseIP(ip) != nil {
+				discovered = append(discovered, ip)
+			}
+
+			if len(discovered) > 0 {
+				p.DNSClusterIPs = discovered
+				p.DNSSource = netpolSourceDiscovered
+			}
+		} else if !apierrors.IsNotFound(err) {
+			log.Info("Failed to discover OpenShift dns-default ClusterIP", "error", err)
+			if agent != nil && agent.Status.NetworkPolicy.DNSClusterIPsSource == netpolSourceDiscovered && len(agent.Status.NetworkPolicy.DNSClusterIPs) > 0 {
+				p.DNSClusterIPs = append([]string(nil), agent.Status.NetworkPolicy.DNSClusterIPs...)
+				p.DNSSource = netpolSourceDiscovered
+			}
+		}
+	}
+
 	// 5. Documented fallback default
 	if len(p.DNSClusterIPs) == 0 {
 		p.DNSClusterIPs = []string{defaultDNSClusterIP}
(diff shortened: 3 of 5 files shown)


k8s-operator/internal/controller/platformagent_manifests.go
```diff
@@ -4525,6 +4525,18 @@ func clusterDNSPeers(dnsIPs []string) []networkingv1.NetworkPolicyPeer {
 				},
 			},
 		},
+		{
```
mplakhtiy, line 4528: **🔴 Blocking — the sixth `clusterDNSPeers` peer lands without updating the two places that pin the DNS rule's shape, and `Run Controller Tests` is red at this head**

`TestBuildNetworkPolicy` asserts `len(ruleDNS.To) != 5`, and the comment above that assertion enumerates the five peers by name. This peer makes it six:

    --- FAIL: TestBuildNetworkPolicy (0.00s)
        platformagent_controller_test.go:1200: expected 5 peers in DNS egress rule

Six `TestAgentsGolden` fixtures fail alongside it — `platformagent.yaml`, `platformagent-telemetry.yaml`, `platformagent-egress-allowlist.yaml`, `platformagent-ha.yaml`, `platformagent-scoped-sa.yaml`, `platformagent-tagged.yaml`. The golden diff in the job log is exactly this peer, so the render is what was intended and only the expectations are stale.

Failure scenario: the required check stays red and Tide will not take the change.

On why it did not surface before CI: the `-run "TestResolveNetpolProfile|TestTheRenderedPolicy"` filter matches neither `TestBuildNetworkPolicy` nor anything in `./internal/testing`, which is a separate package that a filtered `./internal/controller` run never reaches. The goldens live there.

Fix: `go test ./internal/testing -update` from `k8s-operator/`, bump the count in `TestBuildNetworkPolicy` to 6 and update the peer enumeration in the comment above it, then run both package suites unfiltered.

mplakhtiy (commented):
Reviewed at head `bd2e90e76a46965a12d2c8cce4efdd131615bf17`. Produced with AI assistance (Claude Opus 5).

Four findings: one blocking, three medium. Two are inline on the changed lines; the two below are on files this diff does not touch, so GitHub will not anchor them.

#### 🟠 Medium — `TestShellSandboxDNSEgressNamesEveryClusterDNSPeer` passes without covering the new peer

`k8s-operator/internal/controller/shell_sandbox_manifests_test.go`.

This is the test written because the sandbox and gateway peer lists drifted apart and cost a live install its name resolution — its own comment records the "Temporary failure in name resolution" that read as the credential proxy being down. It collects selectors with `peer.PodSelector.MatchLabels["k8s-app"]`. The new peer carries no `k8s-app` key, so it contributes an empty string to `podSelectors` and the assertion loop never looks at it. The comment's "The four peers below are what the gateway policy has always had" is now wrong as well.

Failure scenario: a later change drops the OpenShift peer from `clusterDNSPeers`, the goldens are regenerated along with it, the suite stays green, and OpenShift installs lose name resolution in the sandbox — precisely the failure this test exists to prevent. A golden records what the code did; this test is the one that records what it must do.

Extending it means asserting the `dns.operator.openshift.io/daemonset-dns: default` selector explicitly rather than reading a `k8s-app` key the peer does not have, and correcting the peer count in the comment.

#### 🟠 Medium — the documented DNS egress peers and discovery ladder no longer match the operator

The "Egress (Internal Services)" bullet in `docs/site/src/content/docs/reference/security-and-iam.md` enumerates the operator's port-53 peers exactly — "DNS on TCP/UDP port `53` to CoreDNS/NodeLocal DNS pods, the classic Service-CIDR ClusterIP (`10.96.0.10/32`), and the metadata address `169.254.169.254/32`". AGENTS.md makes that page canonical for what the agent is and is not permitted to do, and it now under-describes the egress boundary by one peer.

In `docs/site/src/content/docs/operator/platformagent-crd.md`, the `dnsClusterIPs` entry still says it suppresses "dynamic discovery from `kube-system/kube-dns`" with no mention of the new rung, and the `networkPolicy.dnsClusterIPsSource` status row documents `Discovered` as though it had a single origin when it now has two.

Failure scenario: someone reading the canonical security page to audit what the agent may reach does not learn that a namespace they can create is a permitted egress destination on port 53 — the same fact the second inline finding turns on. `make docs-check` passes here because it gates generated regions, links, identifiers, map coverage and the context budget, not whether prose still matches source.

k8s-operator/internal/controller/platformagent_egress_policy.go
```diff
@@ -285,6 +285,7 @@ func buildAgentEgressNetworkPolicy(agent *agentv1alpha1.PlatformAgent, dnsCluste
 	dnsPeers := []networkingv1.NetworkPolicyPeer{
 		namespacedPodPeer("kube-system", map[string]string{"k8s-app": "kube-dns"}),
 		namespacedPodPeer("kube-system", map[string]string{"k8s-app": "node-local-dns"}),
+		namespacedPodPeer(openShiftDNSNamespace, map[string]string{openShiftDNSDaemonSetLabelKey: openShiftDNSDaemonSetLabelVal}),
```
mplakhtiy, line 288: **🟠 Medium — the OpenShift DNS peer renders on every cluster, so on GKE it is a standing port-53 grant to a namespace nothing occupies**

This peer and the matching entry in `clusterDNSPeers` (`platformagent_manifests.go`) are both unconditional. On GKE — the primary supported platform — `openshift-dns` is an ordinary namespace name that nothing occupies, so the first party to create it with a pod labelled `dns.operator.openshift.io/daemonset-dns: default` receives TCP and UDP 53 from the gateway pod and from the shell sandbox. `buildShellSandboxNetworkPolicy`'s own doc comment states the stakes: "Off-pod, the shell's whole outbound world is DNS and one ClusterIP." TCP/53 to a pod of the attacker's choosing is an arbitrary bidirectional tunnel, which is a different exposure from the DNS forwarding that the `kube-dns` peer two lines above already concedes — and unlike `kube-system`, occupying `openshift-dns` needs no privileged write.

Failure scenario: on a GKE install, anyone able to create a namespace and a pod — including the agent itself, acting through the credential broker — gives the shell sandbox a port-53 channel to a listener they control, around the narrow egress the sandbox's separate pod exists to enforce.

Detection is available and costs nothing here. `resolveNetpolProfile` already reads `openshift-dns/dns-default` and reports `DNSSource == netpolSourceDiscovered`, so the builders can be told whether OpenShift DNS was actually found rather than granting the peer blind. The unconditional `metadataResolverCIDR` peer below is the local precedent for granting without detecting, and the comment block above it spends three paragraphs arguing that a detector there would guess wrong and cost the install its name resolution. That argument does not transfer to a resolver the operator has already located. Either gate the peer on discovery, or write the paragraph saying why unconditional is right — every other peer in these two lists has one.

fkc1e100:
**Triage note:** This PR has been in draft state for over a week. Please either mark as ready for review or close if no longer needed.

PR #1379, opened 2026-09-10
feat(installer): add --generate-only mode for operator-driven handoff (#1328)

### LINKED ISSUE #1328: feat(installer): a --generate-only mode that writes install.env and terraform.tfvars and hands the apply to the operator
## Summary

`install.sh` should be able to stop after writing `install.env` and `terraform.tfvars`, print the exact `lifecycle.sh` commands, and hand the plan, init and apply to the operator. Today the wizard always owns the apply.

## What exists

- Answering "n" at the final "Proceed with automated GKE cluster & Platform Agent provisioning?" prompt pauses the run and prints `cd terraform/examples/full-install && KUBE_AGENTS_STATE_BUCKET=auto ./lifecycle.sh apply`. It is interactive only, so a scripted run cannot reach it.
- `--dry-run` writes both files and runs `terraform validate` plus a plan on local state, then exits. It cannot be used as the handoff: the plan runs against local state rather than the install's GCS backend, and the run stops before the steps below.

## What is missing

1. An explicit, non-interactive way to say "generate and stop": a `--generate-only` flag, and the same choice as a question in the interactive wizard, both taking the same exit path.
2. A handoff message that is honest about what `install.sh` does outside Terraform, so an operator who takes over the apply knows what is theirs to run: CMEK on a pre-existing cluster, the Workload Identity pool on a pre-existing Standard cluster, NetworkPolicy enforcement, the GitHub App PEM import before the apply, and the managed-OTel scope after it. The message should print the `lifecycle.sh plan` and `apply` commands with `KUBE_AGENTS_STATE_BUCKET`/`KUBE_AGENTS_STATE_PREFIX` filled in, since a hand-driven apply without them lands on local state.
3. The pre-apply checks that guard the generated configuration (`check_service_account_ownership` from #1327) should still run in this mode, so the operator is told about a collision before they take over rather than by a 409.

## Acceptance criteria

- [ ] `./install.sh --generate-only --non-interactive ...` writes `install.env` (when absent) and `terraform.tfvars`, runs the pre-apply checks, prints the handoff, and exits 0 without touching GCP beyond read-only probes.
- [ ] The interactive wizard offers the same choice at the confirmation step.
- [ ] The handoff names the out-of-Terraform steps above with the commands for each.
- [ ] `docs/site` install pages and `.agents/skills/install-kube-agents/SKILL.md` describe the mode.

## Context

Raised while working on #1327. Related: the "Proceed?" pause in `install.sh`, `--dry-run`, `terraform/examples/full-install/README.md` (hand-driven applies and remote state).

kyber775:
## Summary

Adds a `--generate-only` mode to `install.sh` and an interactive confirmation choice (`g`) that writes `install.env` and `terraform.tfvars`, runs pre-apply validation checks, prints explicit out-of-Terraform prerequisite and post-apply commands along with exact `lifecycle.sh apply` commands with remote state configuration (`KUBE_AGENTS_STATE_BUCKET` and `KUBE_AGENTS_STATE_PREFIX`), adds KMS key creation commands to the handoff for GitOps minter PEM import, adds `guard_minter_key` to `lifecycle.sh` to fail fast before apply when minter is enabled without an imported key while safely tolerating query failures, documents the workflow in `INSTALL.md` as the recommended path for existing infrastructure, and exits 0 without creating or mutating cloud resources.

## Why This Change

Previously, `install.sh` always assumed the script itself would execute `lifecycle.sh apply`. While `--dry-run` existed for local validation and resource previews, it operates on local state, does not persist `install.env`, and skips post-interview pre-apply validation checks and handoff steps. Operators needing to generate configuration files and take over the plan, init, and apply steps in external GitOps / CI/CD pipelines lacked an explicit, non-mutating generation path and a clear handoff detailing the steps `install.sh` executes outside Terraform (such as CMEK key setup, Workload Identity pools, legacy NetworkPolicy addons, GitHub App private key import, and managed OTel scope).

Furthermore, when adopting existing infrastructure, operators benefit from reviewing generated inputs and running prerequisites before mutating live resources.

When the GitOps minter is configured, Minty requires pre-existing KMS keyrings and import-only keys before importing a private key PEM, and `lifecycle.sh apply` guards against applying with `enable_github_minter = true` when no enabled key version exists to prevent wedged Helm releases.

## What Changed

- **`install.sh`**:
  - Added `PARAM_GENERATE_ONLY` parameter and `--generate-only` flag to `parse_args` and `show_help`.
  - Added mutual exclusivity validation between `--dry-run` and `--generate-only` (exits 2).
  - Added `print_generate_only_handoff` helper displaying:
    - Generated paths for `install.env` and `terraform.tfvars`.
    - Out-of-Terraform prerequisites: CMEK database encryption (including GKE service agent identity provisioning) for pre-existing clusters, Workload Identity pool & node pool metadata migration disclosure (`GKE_METADATA`), NetworkPolicy enforcement (`--update-addons=NetworkPolicy=ENABLED`) disclosure for clusters without Dataplane V2, and GitHub App private key PEM import via Cloud KMS keyring/key creation and Minty CLI.
    - Exact `lifecycle.sh apply` command with derived `KUBE_AGENTS_STATE_BUCKET` and `KUBE_AGENTS_STATE_PREFIX`.
    - Out-of-Terraform post-apply steps: Managed OpenTelemetry collection scope setup on newly created clusters.
  - Updated interactive confirmation prompt `(Y/n/g)` to allow selecting `g` to trigger generate-only mode.
  - Runs pre-apply validation checks (`check_github_org_is_organization`) on the generate-only path before printing the handoff and exiting 0 with `GENERATE_ONLY_SUCCESS` status in `/tmp/kube-agents-install-report.json`.
- **`terraform/examples/full-install/lifecycle.sh`**:
  - Added `guard_minter_key` to `apply)` which checks whether `enable_github_minter = true` and refuses the apply if the KMS signing key was queried and has no `ENABLED` version, while logging a warning and proceeding if `gcloud` fails (e.g. insufficient query permissions or transient error).
- **Documentation & Skills**:
  - **`INSTALL.md`**: Added a dedicated section describing `--generate-only` mode as the recommended approach for installing on pre-existing infrastructure, detailing operator review rationale and outlining the prerequisite checklist.
  - **`docs/site/src/content/docs/install/quickstart-gke.mdx`**: Documented `--generate-only` in the common flags reference.
  - **`.agents/skills/install-kube-agents/SKILL.md`**: Added a dedicated "Generate-Only Mode" section.
- **Tests**:
  - `tests/test_install_script.py`: Added unit tests covering `--generate-only` flag parsing, mutual exclusivity with `--dry-run`, `print_generate_only_handoff` rendering with GKE identity creation and valid addon names, constants (`MINTY_CLI_REPO_URL`, etc.), and `write_json_report` output.
  - `tests/test_lifecycle_script.py`: Added unit tests for `guard_minter_key` verifying pass behavior when disabled or enabled with an existing key version, refusal behavior when enabled with no key version, and warning/proceed behavior when gcloud query errors.

## Context

- Closes #1328.
- Builds upon the modular installer infrastructure and lifecycle scripts (`terraform/examples/full-install/lifecycle.sh`).

## Testing

- Unit tests (`tests/test_install_script.py`): 173 tests passed cleanly.
- Lifecycle guard tests (`tests/test_lifecycle_script.py`): 36 tests passed cleanly.
- Full installer & lifecycle unit suite (`python3 -m unittest tests/test_install_script.py tests/test_lifecycle_script.py`): 209 tests passed cleanly.
- Operator tests (`cd k8s-operator && go test -count=1 ./...`): All packages passed cleanly.
- `make docs-check`: Passed with 0 errors (all relative links resolve, terminology verified, character budgets intact).

### Live validation

- **Target**: `kyber-test` cluster (GKE Standard, `northamerica-northeast2-b`, GCP project `bnaylor-kagents-dev`).
- **Command executed**:
  ```bash
  KUBE_AGENTS_INSTALL_ENV=/tmp/tmp.qUBR1w93kU/install.env ./install.sh \
    --generate-only \
    --non-interactive \
    --project-id=bnaylor-kagents-dev \
    --cluster-name=kyber-test \
    --region=northamerica-northeast2-b \
    --allow-unverified-source
  ```
- **Observed Behavior**:
  - Detected GKE Standard cluster `kyber-test` and resolved project metadata (7820384652).
  - Wrote configuration to `/tmp/tmp.qUBR1w93kU/install.env` (permissions 0600) and generated `/home/kyber/src/kube-agents/terraform/examples/full-install/terraform.tfvars`.
  - Executed pre-apply validation checks without invoking `lifecycle.sh apply` or mutating cloud resources.
  - Printed complete operator handoff with exact `gcloud container clusters update ... --update-addons=NetworkPolicy=ENABLED`, GKE service identity provisioning, KMS keyring/key creation, and `lifecycle.sh apply` commands.
  - Successfully wrote `/tmp/kube-agents-install-report.json` with `"status": "GENERATE_ONLY_SUCCESS"`, `"generate_only": true`, and exited 0.
- **Cleanup**: Temporary test directory `/tmp/tmp.qUBR1w93kU` was removed.

## Self-Review

- **Passes run**:
  - **Adversarial Pass (isolated diff inspection)**:
    - *Finding 1*: `install.sh` handoff printed `--update-addons=NetworkPolicyConfig=ENABLED`, which is the internal API field name rather than the valid `gcloud` addon choice (`NetworkPolicy=ENABLED`). Addressed in commit `e306539e` by fixing the command in `install.sh` and updating the corresponding assertion in `tests/test_install_script.py`.
    - *Finding 2*: Mutual exclusion with `--dry-run` needed validation to prevent ambiguous intent. Addressed with explicit exit 2 guard and unit test coverage.
    - *Finding 3*: GitOps minter PEM import handoff lacked KMS keyring and key creation steps needed prior to Minty `import-pk`, and `lifecycle.sh apply` lacked an early refusal guard when `enable_github_minter = true` without an imported key version. Addressed in commit `b27fa866` by including KMS creation commands in handoff, adding `guard_minter_key` in `lifecycle.sh`, and testing with `tests/test_lifecycle_script.py`.
    - *Finding 4*: Handoff previously recommended `lifecycle.sh plan` with a remote state bucket before `apply`, which fails on fresh installs because `plan` operates in read-only mode and does not create the GCS bucket. Addressed by directing operators to `lifecycle.sh apply` (which creates the state bucket automatically).
    - *Finding 5*: Handoff CMEK recipe omitted `gcloud beta services identity create --service=container.googleapis.com` required before adding IAM policy bindings on projects without pre-existing GKE service agents. Addressed in commit `822c600c`.
    - *Finding 6*: Handoff node-recreation mutations (Calico and `GKE_METADATA`) lacked the warning disclosure from #1336. Addressed in commit `822c600c`.
    - *Finding 7*: `guard_minter_key` previously swallowed gcloud query failures and treated them as missing key versions. Addressed in commit `822c600c` by logging a warning and proceeding when gcloud query fails, only hard-refusing when the query succeeds with an empty result.
  - **Docs-Drift Pass**:
    - Verified `INSTALL.md`, `docs/site/src/content/docs/install/quickstart-gke.mdx`, and `.agents/skills/install-kube-agents/SKILL.md` accurately reflect flag names, help strings, exit codes, report payload keys, and recommended existing-infra workflow. Updated rationale and descriptions to reflect prerequisite checklist semantics and removed extraneous references to Slack socket registration. Confirmed all relative links and terminology pass `make docs-check`.

## Risk & Rollout

- **Risk**: Low. Purely additive execution mode; existing interactive and `--non-interactive` apply behaviors are preserved.
- **Rollout**: Available immediately in the standard `install.sh` flow.


install.sh
```diff
@@ -1654,6 +1660,61 @@ tf_compose_dir() {
...
+  echo -e "      --role=\"roles/cloudkms.cryptoKeyEncrypterDecrypter\" --project=${project_id} --quiet"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --database-encryption-key=${key_resource} --project ${project_id}"
+  echo ""
+  echo -e "  • ${C_CYAN}Workload Identity Pool (pre-existing Standard cluster):${C_RESET}"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --workload-pool=${project_id}.svc.id.goog"
+  echo -e "    gcloud container node-pools update <node-pool> --cluster=${cluster_name} --location=${region} --project=${project_id} --workload-metadata=GKE_METADATA"
+  echo ""
+  echo -e "  • ${C_CYAN}NetworkPolicy Enforcement (pre-existing cluster without Dataplane V2):${C_RESET}"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --update-addons=NetworkPolicyConfig=ENABLED"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --enable-network-policy"
```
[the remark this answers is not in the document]
kyber775: Fixed in commit `6c677bbe`. Corrected the gcloud addon name from `--update-addons=NetworkPolicyConfig=ENABLED` to `--update-addons=NetworkPolicy=ENABLED` in `install.sh` and updated the corresponding assertion in `tests/test_install_script.py`.

.agents/skills/install-kube-agents/SKILL.md
```diff
@@ -75,6 +75,28 @@ cd kube-agents
```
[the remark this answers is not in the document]
kyber775: Updated the PR description to adhere fully to `.github/PULL_REQUEST_TEMPLATE.md`, filling out the required `Context`, `Testing` with `Live validation` details executed against `kyber-test` in `bnaylor-kagents-dev`, detailed `Self-Review` passes (adversarial and docs-drift), and `Risk & Rollout`.

install.sh
```diff
@@ -1654,6 +1660,61 @@ tf_compose_dir() {
...
+  echo -e "    gcloud container node-pools update <node-pool> --cluster=${cluster_name} --location=${region} --project=${project_id} --workload-metadata=GKE_METADATA"
+  echo ""
+  echo -e "  • ${C_CYAN}NetworkPolicy Enforcement (pre-existing cluster without Dataplane V2):${C_RESET}"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --update-addons=NetworkPolicy=ENABLED"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --enable-network-policy"
+  echo ""
+  echo -e "  • ${C_CYAN}GitHub App PEM Import (before apply, when GitOps minter is enabled):${C_RESET}"
+  echo -e "    git clone --depth 1 --branch v2.7.1 https://github.com/abcxyz/github-token-minter.git /tmp/minty"
+  echo -e "    cd /tmp/minty && go run ./cmd/minty tools import-pk -project-id=${project_id} -location=${kms_loc} -key-ring=${minter_keyring} -key=${minter_key} -private-key=@<path-to-pem>"
+  echo ""
```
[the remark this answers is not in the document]
kyber775: Fixed in commit `e6b64b74`.

1. Updated `print_generate_only_handoff` in `install.sh` to include Cloud KMS API enablement, keyring creation, and asymmetric-signing key creation with `--import-only --skip-initial-version-creation` before the Minty `import-pk` step.
2. Added `guard_minter_key` to `terraform/examples/full-install/lifecycle.sh` on the `apply)` path to verify that the KMS signing key has at least one `ENABLED` version when `enable_github_minter = true`, refusing early with an actionable error instead of wedging on unready minter pods.
3. Added unit tests in `tests/test_lifecycle_script.py` and `tests/test_install_script.py`.

install.sh
```diff
@@ -1654,6 +1660,61 @@ tf_compose_dir() {
...
+  echo -e "${C_BOLD}2. Plan and apply via lifecycle.sh (remote state in GCS):${C_RESET}"
+  echo -e "  cd ${repo_dir}/terraform/examples/full-install"
+  echo -e "  KUBE_AGENTS_STATE_BUCKET=\"${state_bkt}\" KUBE_AGENTS_STATE_PREFIX=\"${state_pfx}\" ./lifecycle.sh plan"
+  echo -e "  KUBE_AGENTS_STATE_BUCKET=\"${state_bkt}\" KUBE_AGENTS_STATE_PREFIX=\"${state_pfx}\" ./lifecycle.sh apply"
+  echo ""
```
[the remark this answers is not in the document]
kyber775: Fixed in commit `e6b64b74`.

Updated the operator handoff instructions in `install.sh` and `tests/test_install_script.py` to direct the operator directly to `lifecycle.sh apply`, which automatically initializes and creates the remote GCS state bucket on first use.

AntonTyb (changes requested):
this new mode should be clearly described in INSTALL.md -as the  recommended way for installing on existing infra - as Terraform commands (`init`, `validate`) may take longer than on new infra deploy AND `terraform plan` will require more careful analysis...

kyber775:
@AntonTyb Addressed in commit `1f0c3b76`:

- Added a dedicated section `### Generate-Only Mode (Recommended for Existing Infrastructure)` in `INSTALL.md` (under Method 0 and referenced in Method 1).
- Clearly detailed why `--generate-only` is the recommended path when installing on existing infrastructure (longer execution times for Terraform `init`/`validate`, requirement for careful review of `terraform plan` outputs prior to state mutation, and explicit out-of-Terraform prerequisite handoff).
- Documented the full operator handoff lifecycle (`lifecycle.sh plan` and `lifecycle.sh apply` with remote state variables `KUBE_AGENTS_STATE_BUCKET` / `KUBE_AGENTS_STATE_PREFIX`).
- Verified all docs checks (`make docs-check`) and installer test suites pass cleanly.

/review

INSTALL.md
```diff
@@ -85,6 +87,39 @@ Three behaviours worth knowing before the first run:
...
+1. Probes cluster parameters and writes the complete configuration to `install.env` (if absent) and `terraform/examples/full-install/terraform.tfvars`.
+2. Runs pre-flight checks and configuration validations without creating or modifying GCP resources.
+3. Prints the exact step-by-step manual execution recipe:
+   - **Out-of-Terraform prerequisites** for existing clusters (CMEK database encryption enablement, node-pool `GKE_METADATA` workload identity update, NetworkPolicy enablement, and Cloud KMS key creation for GitHub App private key signing).
+   - **Terraform Plan & Apply execution** with remote state management via `lifecycle.sh`:
+     ```bash
+     cd terraform/examples/full-install
+     KUBE_AGENTS_STATE_BUCKET="<project>-kube-agents-tfstate" KUBE_AGENTS_STATE_PREFIX="kube-agents/<cluster>" ./lifecycle.sh plan
+     KUBE_AGENTS_STATE_BUCKET="<project>-kube-agents-tfstate" KUBE_AGENTS_STATE_PREFIX="kube-agents/<cluster>" ./lifecycle.sh apply
+     ```
```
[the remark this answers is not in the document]
kyber775: Fixed in commit `224525a1`. Updated `INSTALL.md` and `.agents/skills/install-kube-agents/SKILL.md` to remove references to `lifecycle.sh plan` and directly instruct operators to run `lifecycle.sh apply`, matching the handoff behavior printed by `install.sh`.

install.sh
```diff
@@ -1746,6 +1750,65 @@ handle_pipeline_status() {
...
+  echo -e "  • ${C_CYAN}Terraform input:${C_RESET} ${tfvars_file}"
+  echo ""
+  echo -e "${C_BOLD}Next steps for the operator to apply manually:${C_RESET}"
+  echo ""
+  echo -e "${C_BOLD}1. Out-of-Terraform prerequisites (run if applicable to your cluster):${C_RESET}"
+  echo -e "  • ${C_CYAN}CMEK Database Encryption (pre-existing cluster without CMEK):${C_RESET}"
+  echo -e "    gcloud services enable cloudkms.googleapis.com --project=${project_id}"
+  echo -e "    gcloud kms keyrings create ${keyring} --location=${kms_loc} --project=${project_id} 2>/dev/null || true"
+  echo -e "    gcloud kms keys create ${key} --keyring=${keyring} --location=${kms_loc} --purpose=encryption --project=${project_id} 2>/dev/null || true"
+  echo -e "    gcloud kms keys add-iam-policy-binding ${key} --keyring=${keyring} --location=${kms_loc} \\"
```
mplakhtiy: **Medium — the printed CMEK recipe omits the GKE service-agent provisioning step, so this binding fails on a project with no GKE service agent.**

The handoff prints `services enable cloudkms` -> `kms keyrings create` -> `kms keys create` -> this `add-iam-policy-binding` -> `clusters update --database-encryption-key`.

The code path it mirrors, `ensure_existing_cluster_cmek`, runs one more command between the key creation and the binding:

    gcloud beta services identity create --service=container.googleapis.com --project="$project_id" 2>/dev/null || true

**Failure scenario.** On a GCP project that has never had a GKE cluster, `service-<N>@container-engine-robot.iam.gserviceaccount.com` does not exist until it is provisioned. This binding then fails with `INVALID_ARGUMENT: Service account ... does not exist`, and the operator stops at the first prerequisite with nothing saying what is missing. The installer never hits it because it always provisions the identity first. Nothing prevents `--generate-only` being used on a project whose first GKE cluster this install creates.

This is the same shape as the `NetworkPolicyConfig` bug fixed in `6c677bbe`: the handoff is a hand-maintained textual copy of a command sequence that lives in code a few dozen lines away, with nothing tying the two together. Adding the `services identity create` line and asserting it in `test_print_generate_only_handoff_renders_required_commands` closes this instance.

install.sh
```diff
@@ -1746,6 +1750,65 @@ handle_pipeline_status() {
...
+  echo -e "      --member=\"serviceAccount:service-\$(gcloud projects describe ${project_id} --format='value(projectNumber)')@container-engine-robot.iam.gserviceaccount.com\" \\"
+  echo -e "      --role=\"roles/cloudkms.cryptoKeyEncrypterDecrypter\" --project=${project_id} --quiet"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --database-encryption-key=${key_resource} --project ${project_id}"
+  echo ""
+  echo -e "  • ${C_CYAN}Workload Identity Pool (pre-existing Standard cluster):${C_RESET}"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --workload-pool=${project_id}.svc.id.goog"
+  echo -e "    gcloud container node-pools update <node-pool> --cluster=${cluster_name} --location=${region} --project=${project_id} --workload-metadata=GKE_METADATA"
+  echo ""
+  echo -e "  • ${C_CYAN}NetworkPolicy Enforcement (pre-existing cluster without Dataplane V2):${C_RESET}"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --project ${project_id} --update-addons=NetworkPolicy=ENABLED"
```
mplakhtiy, line 1794: **Medium — the handoff prints node-recreating mutations without the disclosure #1336 added, and prints them even when the operator declined the opt-in.**

`46621bc4` merges #1336 (`5a7b0eb9`), which makes exactly these two mutations — Calico NetworkPolicy enablement, and the node-pool `GKE_METADATA` migration printed six lines above — gated, disclosed, opt-in changes. `prompt_existing_cluster_opt_ins` (`install.sh:3681`) and `validate_existing_cluster_opt_in_flags` (`install.sh:3379`) both run **before** the generate-only exit at `install.sh:3796`, so by the time this prints, the installer already knows whether consent was given.

It prints both commands unconditionally anyway, under `(run if applicable to your cluster)`, with none of the wording #1336 introduced — no "may recreate nodes and restart workloads", no reference to the answer just collected.

**Failure scenario.** An operator runs `--generate-only` against an existing Standard cluster and answers `n` to "Enabling Calico may recreate nodes and restart workloads ... Enable Calico NetworkPolicy now? (y/N)". On the apply path that refusal ends the install (`install.sh:2013`). On the generate-only path it ends nothing: the run completes successfully and hands them a checklist whose third item is the command they just declined, stripped of the warning that made them decline it. The consent gate #1336 exists to enforce is bypassed by the mode that prints the commands instead of running them.

Two options: gate each printed block on the opt-in state already in hand, or carry #1336's disclosure text into the printed lines. The first also fixes the broader issue that this block prints for Dataplane V2 clusters that need none of it.

terraform/examples/full-install/lifecycle.sh
```diff
@@ -670,6 +670,34 @@ forget_unmanaged_cluster_kms() {
...
+  [[ "$(tfvar enable_github_minter 2>/dev/null || echo "false")" == "true" ]] || return 0
+  local project location keyring key version
+  project=$(tfvar project_id)
+  location=$(sed -E 's/-[a-z]$//' <<<"$(tfvar location)")
+  keyring=$(tfvar github_minter_kms_keyring 2>/dev/null || echo "")
+  [[ -n "$keyring" ]] || keyring="github-token-minter-keyring"
+  key=$(tfvar github_minter_kms_key 2>/dev/null || echo "")
+  [[ -n "$key" ]] || key="github-token-minter-key"
+
+  version=$({ gcloud kms keys versions list --key "$key" --keyring "$keyring" \
```
mplakhtiy: **Medium — every gcloud failure becomes "no ENABLED version" and hard-refuses the apply, inverting the contract the identical probe documents.**

This is a copy of `kms_key_enabled_version` (`scripts/installer/installer_common.sh:680-685`), whose header states the convention the repo settled on:

> Every failure -- no key, no API, no permission -- reads as "none", which the callers treat as "do not enable". ... one pipeline here so the three cannot come to disagree.

This copy reads "none" the same way and then does the opposite with it: `exit 1`, blocking the apply. `2>/dev/null || true` erases the difference between "the key has no ENABLED version" and "we could not ask".

**Failure scenario.** A `lifecycle.sh apply` — direct, via `upgrade.sh`, or via the handoff this PR prints — runs against a working install whose minter key was imported months ago. The `gcloud` principal on PATH has the roles Terraform needs but not `cloudkms.cryptoKeyVersions.list`, or is pointed at another project, or the call transiently fails. The error is discarded, `version` is empty, and the apply is refused with a message asserting something false — "has no ENABLED version" — plus instructions to import a key that is already imported. That apply previously succeeded.

A second route reaches the same place: `location=$(sed -E 's/-[a-z]$//' <<<"$(tfvar location)")` on line 683. `tfvar` calls `exit 1` on failure, which kills the inner subshell, so `sed` reads empty input and `location` is empty; the KMS query then runs against location `""`, errors, is swallowed, and refuses.

The three new tests cover "minter disabled", "key present", and "key genuinely absent". None covers a gcloud that errors, which is the path that produces the wrong answer. Capturing the exit status separately — and either warning and proceeding, or refusing with a message that says the query failed rather than asserting what the key contains — would fix both routes.

mplakhtiy (commented):
Reviewed at head `46621bc45bb01fdaaf5ab946f93b2b47f478a6ff`. Produced with AI assistance.

Six findings at medium severity: two in `install.sh`, one in `lifecycle.sh`, three in `INSTALL.md`. Each is inline below.

Note, not a finding: the two red checks (`Run Agent Startup Tests`, `Run Python Unit Tests`) fail on `NameError: name '_lift' is not defined` in `tests/test_check_image_inventory_go_directive.py` — a `main` bug fixed by `bc3396cc` (#1413). The merge at `46621bc4` should clear them on the next run.

INSTALL.md
```diff
@@ -89,6 +91,38 @@ Three behaviours worth knowing before the first run:
...
+curl -fsSL https://raw.githubusercontent.com/gke-labs/kube-agents/<RELEASE_VERSION>/install.sh | bash -s -- \
+  --generate-only \
+  --project-id="my-gcp-project" \
+  --cluster-name="existing-cluster-name" \
+  --region="us-central1"
+```
+
+#### Why `--generate-only` on Existing Infrastructure:
+
+- **Longer Terraform Operations**: On existing infrastructure with numerous cloud resources, Terraform lifecycle commands (`init`, `validate`, and provider schema inspection) take longer than on greenfield deployments.
```
mplakhtiy: **Medium — this rationale is not how `terraform init` and `validate` behave.**

The cost of `init` and `validate` is a function of provider and module download and of configuration size. It is the same composition either way, and neither command enumerates the resources that already exist in the project. "Provider schema inspection" reads the provider binary, not the cloud.

This matters because it is the load-bearing reason the section gives for calling `--generate-only` the recommended path on existing infrastructure, and a reader who acts on it is choosing the mode for a benefit it does not deliver. The two bullets under it — review before mutating live resources, and the out-of-Terraform handoff — stand on their own.

```suggestion
- **Operator Review Before Live Mutation**: Adopting existing infrastructure means `terraform apply` touches resources you did not create, so generating the inputs and reviewing them before the apply keeps that decision with the operator.
```

INSTALL.md
```diff
@@ -89,6 +91,38 @@ Three behaviours worth knowing before the first run:
...
+- **Operator Review & Control**: Adopting existing infrastructure benefits from inspecting generated inputs (`terraform.tfvars`, `install.env`) and manual execution of prerequisites before applying Terraform changes.
+- **Out-of-Terraform Prerequisites & Operator Handoff**: The installer auto-detects existing cluster parameters (Workload Identity pool, CMEK database encryption, NetworkPolicy status, OTel collection scope, and KMS token-minter keys), runs pre-apply validations without mutating GCP resources, and prints the exact shell commands needed before and after Terraform apply.
```
mplakhtiy: **Medium — the handoff neither auto-detects nor prints exact commands.**

`print_generate_only_handoff` prints all four prerequisite blocks unconditionally, under the heading `(run if applicable to your cluster)`, and leaves the operator to work out which apply. It does not read the probed cluster state to decide what to show. OTel collection scope is not probed at all. The Workload Identity command carries a literal `<node-pool>` placeholder, so it is not an exact command either.

The installer does have the answers by then — `TFVARS_CLUSTER_EXISTS`, `TFVARS_CLUSTER_MODE`, `TFVARS_CREATE_CLUSTER`, `enable_github_minter`, and (since #1336) the opt-in state — which is what makes this sentence read as a description of a tool that tailors its output. Either tailor the output or describe the fixed checklist that is actually printed.

```suggestion
- **Out-of-Terraform Prerequisites & Operator Handoff**: The installer probes the target cluster, runs pre-apply validations without mutating GCP resources, and prints a checklist of the steps Terraform cannot perform (CMEK database encryption, the Workload Identity pool, NetworkPolicy enforcement, the GitHub App private key import, and the managed-OTel scope) for you to apply as they pertain to your cluster.
```

INSTALL.md
```diff
@@ -89,6 +91,38 @@ Three behaviours worth knowing before the first run:
...
+1. Probes cluster parameters and writes the complete configuration to `install.env` (if absent) and `terraform/examples/full-install/terraform.tfvars`.
+2. Runs pre-flight checks and configuration validations without creating or modifying GCP resources.
+3. Prints the exact step-by-step manual execution recipe:
+   - **Out-of-Terraform prerequisites** for existing clusters (CMEK database encryption enablement, node-pool `GKE_METADATA` workload identity update, NetworkPolicy enablement, and Cloud KMS key creation for GitHub App private key signing).
+   - **Terraform Apply execution** with remote state management via `lifecycle.sh`:
+     ```bash
+     cd terraform/examples/full-install
+     KUBE_AGENTS_STATE_BUCKET="<project>-kube-agents-tfstate" KUBE_AGENTS_STATE_PREFIX="kube-agents/<cluster>" ./lifecycle.sh apply
+     ```
+   - **Post-apply steps** (managed-OTel collection scope and Slack socket registration if configured).
```
mplakhtiy: **Medium — the handoff prints no Slack socket registration.**

`print_generate_only_handoff`'s post-apply section (step 3) prints one thing: the `--managed-otel-scope` command. There is no Slack socket registration anywhere in the function, in either the prerequisite or the post-apply block.

An operator reading this waits for a Slack step that never appears, or assumes one was printed and scrolls past. The parenthetical is the whole claim, so removing it is the fix.

```suggestion
   - **Post-apply steps** (managed-OTel collection scope, on a cluster this install created).
```

install.sh
kyber775: Fixed in commit `822c600c`. Added `gcloud beta services identity create --service=container.googleapis.com --project=${project_id} 2>/dev/null || true` to the CMEK recipe in `print_generate_only_handoff`, matching `ensure_existing_cluster_cmek`, and added an assertion in `test_print_generate_only_handoff_renders_required_commands`.

install.sh
kyber775: Fixed in commit `822c600c`. Added the explicit node-recreation disclosure comments (`# Note: Migrating node pools to GKE_METADATA recreates nodes and restarts workloads.` and `# Note: Enabling Calico may recreate nodes and restart workloads.`) directly to the printed Workload Identity and NetworkPolicy handoff recipes.

terraform/examples/full-install/lifecycle.sh
kyber775: Fixed in commit `822c600c`. Updated `guard_minter_key` in `lifecycle.sh` to capture `gcloud` returncode and output separately. When the query fails (non-zero exit code due to permissions or transient errors), it logs a warning and proceeds with the apply rather than hard-refusing with a false claim that no enabled version exists. Also safely retrieves `project` and `location` before querying KMS, and added unit test coverage in `tests/test_lifecycle_script.py`.

INSTALL.md
kyber775: Fixed in commit `822c600c`. Updated `INSTALL.md` with the suggested rationale focusing on operator review and control before mutating live resources.

INSTALL.md
kyber775: Fixed in commit `822c600c`. Updated `INSTALL.md` and `SKILL.md` to accurately describe the output as a checklist of out-of-Terraform prerequisites rather than exact auto-tailored commands.

INSTALL.md
kyber775: Fixed in commit `822c600c`. Removed the reference to Slack socket registration in `INSTALL.md` post-apply steps, keeping only the managed OTel scope command.

install.sh
```diff
@@ -1746,6 +1750,68 @@ handle_pipeline_status() {
...
+  echo -e "  • ${C_CYAN}CMEK Database Encryption (pre-existing cluster without CMEK):${C_RESET}"
+  echo -e "    gcloud services enable cloudkms.googleapis.com --project=${project_id}"
+  echo -e "    gcloud kms keyrings create ${keyring} --location=${kms_loc} --project=${project_id} 2>/dev/null || true"
+  echo -e "    gcloud kms keys create ${key} --keyring=${keyring} --location=${kms_loc} --purpose=encryption --project=${project_id} 2>/dev/null || true"
+  echo -e "    gcloud beta services identity create --service=container.googleapis.com --project=${project_id} 2>/dev/null || true"
+  echo -e "    gcloud kms keys add-iam-policy-binding ${key} --keyring=${keyring} --location=${kms_loc} \\\\"
+  echo -e "      --member=\\\"serviceAccount:service-\$(gcloud projects describe ${project_id} --format='value(projectNumber)')@container-engine-robot.iam.gserviceaccount.com\\\" \\\\"
+  echo -e "      --role=\\\"roles/cloudkms.cryptoKeyEncrypterDecrypter\\\" --project=${project_id} --quiet"
+  echo -e "    gcloud container clusters update ${cluster_name} --location ${region} --database-encryption-key=${key_resource} --project ${project_id}"
```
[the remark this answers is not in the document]
kyber775: Fixed in commit `7a15050a`. Corrected quote escaping in `print_generate_only_handoff` so `--member` and `--role` do not include literal escaped quotes when printed, and added explicit assertions in `tests/test_install_script.py`.

install.sh
```diff
@@ -1746,6 +1750,68 @@ handle_pipeline_status() {
...
+  local repo_dir="$1" project_id="$2" cluster_name="$3" region="$4" tfvars_file="$5"
+  local kms_loc key_resource keyring key minter_keyring minter_key state_bkt state_pfx
+
+  kms_loc="$(derive_kms_location "$region")"
+  keyring="${GKE_DB_KMS_KEYRING:-${DEFAULT_GKE_DB_KMS_KEYRING:-platform-agent-keyring}}"
+  key="${GKE_DB_KMS_KEY:-${DEFAULT_GKE_DB_KMS_KEY:-k8s-secret-encryption-key}}"
+  key_resource="projects/${project_id}/locations/${kms_loc}/keyRings/${keyring}/cryptoKeys/${key}"
+  minter_keyring="${KMS_KEYRING:-${DEFAULT_KMS_KEYRING:-github-token-minter-keyring}}"
+  minter_key="${KMS_KEY:-${DEFAULT_KMS_KEY:-github-token-minter-key}}"
+  state_bkt="$(tf_state_bucket)"
```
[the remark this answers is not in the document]
kyber775: Fixed in commit `7a15050a`. Removed redundant inline string fallbacks from `print_generate_only_handoff` (e.g. `${GKE_DB_KMS_KEYRING:-$DEFAULT_GKE_DB_KMS_KEYRING}`), aligning with `ensure_existing_cluster_cmek` and `import_github_pem`.

AntonTyb (approved):

PR #1319, opened 2026-09-08
feat(a2a,operator): arm the auth callout — a bus client with a token authenticates as itself

bnaylor:
## Summary

This arms the NATS auth callout. Until now every client on the A2A bus dialled with a
static password rendered into `nats.conf`, so "who is this" was answered by a shared
secret and the answer was the same for three unrelated workloads. The callout replaces
that for the clients that can present a Kubernetes token: a client connects with a
projected ServiceAccount token, `nats-server` hands the authorization request to the
callout over `$SYS.REQ.USER.AUTH`, the callout runs a real TokenReview against the API
server, resolves the ServiceAccount, looks it up in an operator-rendered identity map,
and returns a signed user JWT carrying that principal's subject permissions. The
operator renders the map, deploys the callout behind its own ingress fence, arms it in
`nats.conf`, and reports `BusCredentialsReady`.

One identity moves in this change: `provision`, the bus provisioning Job. Everything
else keeps its static password, deliberately, because an identity may only move when a
client that holds a token moves with it — declare a principal whose workload cannot
present a token and the entry authorizes nobody while reporting that it does. `web` can
never move, because a browser holds no ServiceAccount token. The platform agent pod does
not move either: its only bus client is the Hermes bridge sidecar, which dials as
`worker`, and nothing renders it an `a2a-bus` token. Session pods move in the follow-up,
which is what gke-labs/kube-agents#1270 is about; they hold no Kubernetes identity at all
today.

An earlier revision of this branch did declare an `agent` principal, and this paragraph
claimed two identities moved. Neither was true of anything the operator renders; see the
fourth bot pass in the self-review below.

## Why This Change

The `worker` NATS user is three workloads wearing one password: session pods, the agent
pod's `platform-agent` container, and the Hermes bridge sidecar. One credential, one
grant set, and no way for the bus to tell them apart — so the agent pod holds a
credential that also opens every session's grants, and a leak anywhere is a leak
everywhere. Static passwords also cannot be revoked without re-rendering the config and
rolling the bus.

The callout is the mechanism that makes per-workload grants possible at all. It is a
precondition for #1270 (per-session credentials) and for anything later that wants to
distinguish two clients on the bus. Without it, every narrowing anyone writes is a
narrowing of a shared secret.

## What Changed

- **`a2a/authcallout/`** — the service. `service.go` answers `$SYS.REQ.USER.AUTH`;
  `tokenreview.go` validates the presented token against the API server and demands the
[...]

Issue #1259, opened 2026-09-06
fix(operator): the A2A provision Job fails on a real cluster — `stat .: permission denied`, so no streams are provisioned


bnaylor:
On a fresh flip to `mode: next` against a real GKE cluster, the A2A provision Job (`<name>-a2a-provision-<hash>`, rendered by `platformagent_a2a_manifests.go` from #1211) never completes. Its pod exits with:

```
: stat .: permission denied
```

after printing the provisioning JSON. Because the Job never completes, no streams or topics are created, and a client on the bus gets:

```
a2a: stream TOPICS-STATE: nats: API error: code=404 err_code=10059 description=stream not found
```

The bus itself is healthy — the NATS StatefulSet reaches 1/1 and a client authenticates fine — so this is specifically the provisioning step.

**Suspected cause:** the pod hardening the PR added over its source branch. The Job's pod spec carries `seccompProfile: RuntimeDefault`, a read-only root filesystem and a non-root user; `stat .` failing on the working directory is the shape you get when the nats-box image's default workdir is not readable or writable under that combination. That is a guess from the symptom — the fix should confirm which of the three it is rather than loosening all of them.

**Observed on:** `bnaylor-ka-test` (us-east4, Dataplane V1), operator built from `main` @ 8b5a02fe plus the agent-bus branch (#1256), 2026-09-05. Reproduce by flipping a CR to `mode: next` and watching the provision Job.

Found while doing the live validation for #1256 (the agent's half of the bus). Not caused by that PR — it touches none of the provisioning code — but it is what stops a `next` install from having any topics to read.

bnaylor:
**Housekeeping for whoever sees this issue close: check which PR closed it.**

#1272 is the fix, and its body says `Fixes #1259`. But #1319 also carries a commit — `a25ea34d`, "fix(a2a): pin the provision container's cwd" — whose message reads `Fixes #1259.`, and that commit was **reverted** on that branch. The revert does not remove the earlier commit from the history, so if #1319 merges to `main` first, GitHub will close this issue against a change that is not in the tree.

The commit message can't be corrected without rewriting #1319's history, which isn't worth it on a branch that size.

So:

- **#1272 merges first** — this closes correctly, nothing to do. That is the likely order; #1319 is currently `DIRTY` and `risk:high`.
- **#1319 merges first** — **reopen this issue.** It stays open until #1272 lands.

[...]

PR #1388, opened 2026-09-10
fix(operator): name the A2A provision Job from its rendered spec, not only its script

### LINKED ISSUE #1347: A2A provision Job's name hashes only the script, so pod-spec fixes never reach an existing install
`buildA2AProvisionJob` names the Job from a digest of the provisioning script alone:

```go
script := a2aProvisionScript(agent)
sum := sha256.Sum256([]byte(script))
name := fmt.Sprintf("%s-a2a-provision-%s", agent.Name, hex.EncodeToString(sum[:])[:8])
```

(`k8s-operator/internal/controller/platformagent_a2a_manifests.go:896-898`)

The hash-in-the-name is how this Job gets replaced at all — a Job's pod template is immutable, so the operator cannot update one in place, and a changed script produces a new name and therefore a new Job. That works, and the function's own comment explains the create-only convergence it buys.

What it does not cover is the pod spec. The image, the `securityContext`, the uid, the env, the volume mounts and the `WorkingDir` are all outside the digest, so a change to any of them renders a Job with the *same* name as the one already on the cluster. `applyManaged` cannot patch the immutable template, and the operator does not delete and recreate, so the change silently does not take effect. No error, no event, no `Degraded` — the Job object on the cluster simply keeps the old spec forever.

**Seen, not theorised.** Fixing #1259 (gke-labs/kube-agents#1319) meant adding `WorkingDir: "/tmp"` to that container. On an install that already carried the broken Job, rolling the operator to the fixed build changed nothing: same name `platform-agent-a2a-provision-0a06e350`, same crash-looping pod, still `stat .: permission denied`. Deleting the Job by hand made the operator recreate it, and it completed in four seconds. The fix was correct the whole time and the operator had no way to deliver it.

**Blast radius.** Narrow today and not narrow later:

- Fresh installs are fine — nothing exists to collide with.
- Any change that also touches the script is fine, since the name moves anyway. That is why #1319 does not have to solve this to be correct: it rewrites how the Job authenticates, so its script differs from `main`'s and every install gets a new Job name on upgrade.
- What is exposed is exactly the case that will keep recurring: a pod-spec-only fix to this Job — an image bump for a nats-box CVE, a securityContext tightening, a resource limit — shipping to installs where it does nothing.

An image bump is the one to worry about. `defaultA2AProvisionImage` is a pinned third-party tag (`natsio/nats-box:0.14.5`); the day it needs moving for a vulnerability, the operator will roll, report success, and leave the old image running on every existing install.

**Fix options**, in rough order of preference:

1. Digest the rendered pod template instead of the script. It subsumes the script (the script *is* the container's command) and it is the property that actually has to be immutable. The render is deterministic from the agent spec, so it should not churn — worth asserting with a test that renders twice and compares, since a map iteration or a timestamp sneaking in would produce a Job recreated on every reconcile.
2. Keep the script hash for the name but compare the live Job's pod template against the rendered one and delete-and-recreate on a mismatch. More code, and it makes the delete a reconcile-time action rather than a naming consequence.
3. Leave it and document it, on the grounds that stage 1 accepts manual migration. The function's comment already says that about stream *contents*; it does not say it about the Job's own spec, and the two are different promises.

I lean on (1). Whichever way it goes, the invariant to pin in a test is that two renders differing only in pod spec produce two different Job names.

**Related:** #1259 (the bug whose fix surfaced this), gke-labs/kube-agents#1319.

bnaylor:
## Summary

The A2A provision Job is now named from a digest of its rendered spec instead of a digest of the provisioning script alone. Jobs are immutable and `reconcileA2A` creates this one only when nothing exists under its name, so the name is the only lever the operator has on it; with the script-only digest, a change to anything else in the pod spec (image, uid, securityContext, env, mounts, `WorkingDir`) rendered the same name as the Job already on the cluster and never took effect. The new digest covers the whole `JobSpec`, which includes the pod template and therefore the script, so any rendered change produces a new Job.

## Why This Change

#1347, seen live while fixing #1259: rolling the operator to a build that added `WorkingDir` to the provision container changed nothing on an install that already carried the broken Job. Same name (`platform-agent-a2a-provision-0a06e350`), same crash-looping pod, still `stat .: permission denied`. Deleting the Job by hand made the operator recreate it, and it completed in four seconds. The fix was correct and the operator had no way to deliver it. No error, no event, no `Degraded`.

Fresh installs never hit this, and neither does any change that also edits the script. What it leaves exposed is the case that will recur: a pod-spec-only change to this Job. The one to worry about is an image bump. `defaultA2AProvisionImage` is a pinned third-party tag (`natsio/nats-box:0.14.5`); the day it moves for a CVE, the operator rolls, reports success, and leaves the old image running on every existing install.

## What Changed

- `buildA2AProvisionJob` builds the Job with an empty name and then sets `job.Name` from `a2aProvisionJobName(agent, job.Spec)`, a new helper that `json.Marshal`s the spec, SHA-256s it, and keeps the first 8 hex characters after `<agent>-a2a-provision-`. The body of the Job literal is untouched, so the diff inside the function is the two naming lines plus the two lines after the literal.
- The digest is over the `JobSpec`, not the `PodTemplateSpec` alone. The issue proposes the template; the spec is the same one line of code and also covers `backoffLimit` and `ttlSecondsAfterFinished`, which are exactly as unreachable under create-only convergence. The helper's comment says so.
- Serialization is `json.Marshal`. It is deterministic for these types: struct fields in declaration order, map keys sorted (the template's labels are the only map), and the render has nothing time- or randomness-derived; the one Secret reference is by name and key, not by value. `commonLabels` carries no operator-version label, so an operator upgrade alone does not rename the Job. No error return, following the `scopedSAPoolJSON` precedent in this package: every field is an API type the server round-trips through JSON, a builder has nowhere to put an error, and a Marshal failure would show up as every render digesting the same bytes, which the pod-spec test catches.
- Two named constants beside `a2aConfigHashLength`: `a2aProvisionJobNameInfix` and `a2aProvisionJobNameHashLength` (8, unchanged, and what `a2aPasswordDigestNeedles` in the tests already assumes).
- The builder's comment now says what the digest covers and how a superseded Job leaves, which depends on how far it got: a completed one by TTL, one whose pod ran and failed by `backoffLimit` then TTL, and one whose pod never ran (unpullable image, unschedulable, admission refusal) not at all, since it has no terminal condition for the TTL to start from; it stays until a mode flip or agent deletion and holds one namespace pod-quota slot meanwhile. What holds today: the status scan reads only the current name, so a stale failure does not park the phase, and `cleanupA2A` deletes by label, so a mode flip removes every generation. The paragraph on what the digest still does not cover stays: stream contents, which are create-only on the bus and migrate by hand.
- Four tests in a new file `platformagent_a2a_provision_job_test.go` (a new file rather than additions to `platformagent_a2a_manifests_test.go`, which #1272 and #1319 both edit heavily): the name is deterministic across renders and across three reconciles against a fake API (exactly one Job, the failure mode the issue warned about); a render differing only in image gets a different name, with a guard that the script did not change; a render differing only in script, or only in `WorkingDir`, gets a different name through the helper; and through the reconciler, an image change yields a second Job under the new name carrying the new image while the original is left alone with the old one, and a `Failed` condition stamped on the original before that reconcile does not reach the returned state.
- No garbage collection of superseded Jobs is added. Today the operator never deletes a provision Job outside a mode flip, and this PR keeps that, stated in the builder comment and pinned by the reconciler test. The self-review below found the case where that matters (a superseded Job whose pod never ran, which no TTL ever clears); deleting superseded generations by label is a follow-up, not part of this change.

## Context

Closes #1347. Related: #1259 (the bug whose fix surfaced this).

Open PRs that edit the same function, and how this one differs:

- #1272 (`fix/a2a-provision-workdir`) adds `WorkingDir` and an `a2aProvisionWritablePath` constant to this container. That is the change this PR exists to deliver to existing installs; the two are complementary, not competing. #1272 changes lines inside the Job literal and one constant near the top of the block; this PR changes the naming lines above the literal and two constants further down. #1272 merged on 2026-09-10 and this branch now carries it. Keeping both was necessary but not sufficient: git merged the two hunks clean while silently making one of the tests here inert, because the `WorkingDir` case edited the container to the same `/tmp` #1272 had just made the rendered value. That is fixed in 98268914 and described under Testing.
- #1319 (`a2a/auth-1-callout`) rewrites the script and adds a `ServiceAccountName` and a token volume inside the same literal. It does not touch the naming lines. Same merge note.
- #1306 edits the script inside this function; #1258 and #1256 touch other parts of the file. None changes how the Job is named.

Branched from `upstream/main` at `1784e18c`, not from either PR. Zero drift at push time.

## Testing

- `make -C k8s-operator test` on the branch: every package `ok`, `internal/controller` at 91.1% coverage, exit 0 (transcript kept off-repo).
- The new tests against `upstream/main` (`1784e18c`), copied into a throwaway detached worktree with the helper-based script test dropped since the helper does not exist there:
  - `TestA2AProvisionJobNameIsDeterministic`: PASS (the script-only digest was also deterministic).
  - `TestA2AProvisionJobNameTracksThePodSpec`: FAIL, `image changed and the Job is still "test-agent-a2a-provision-7cedf0db"`.
  - `TestReconcileA2ACreatesANewProvisionJobWhenThePodSpecChanges`: FAIL, `1 provision Jobs after the image changed, want 2`.
- The same three plus `TestA2AProvisionJobNameTracksTheScript` on the branch: all PASS, alongside the existing `TestBuildA2AProvisionJob`, `TestReconcileA2AGatedByMode`, and `TestA2AProvisionJobConditionsDriveStatus`.

**After merging #1272 (98268914).** Review caught this before it reached main, and it was real rather than theoretical -- run on the merge, before any fix:

```
--- FAIL: TestA2AProvisionJobNameTracksTheScript (0.00s)
    platformagent_a2a_provision_job_test.go:147: WorkingDir changed and the name is still
    "test-agent-a2a-provision-7d6329da"; this is the #1259 fix the old digest could not deliver
```

#1272 set the rendered `WorkingDir` to `/tmp`, which is what the test's `workdir` case assigned, so the edited spec was byte-identical to the original and the digest correctly returned the same name.  The case is now derived from the render (`WorkingDir += "/subdir"`) rather than written as a literal, so it stays a real edit wherever that value moves next -- the same shape as the script case above it, which appends to `Command[2]`.

Red-proved rather than only re-run green: narrowing `a2aProvisionJobName` back to hashing the container command alone makes the case fail again, so it still covers the pod spec.  `go build ./... && go vet ./... && go test ./... -count=1` green in `k8s-operator/` on the merge.

Worth naming the near miss: `Run Controller Tests` reported success on the pre-merge head about ten hours before #1272 landed.  Merged on that stale green, main would have gone red on this test.
- No envtest harness exists under `internal/controller` (the Makefile downloads the binaries; nothing in the package uses them), so the create-only convergence path is exercised through the same fake-client `reconcileA2A` pattern the existing condition tests use.
- No docs, chart, or `k8s-operator/config` files changed, so `make docs-check`, `make chart-check`, and prettier were not run; `make test`'s `manifests` and `generate` steps left the tree clean.

### Live validation

Not live-tested. The only `mode: next` install reachable from here (`a2a-next-dev`) carries another session's stacked build and is read-only for this work, and the default-mode dev cluster (`bnaylor-ka-test`) renders no provision Job at all, so there is no install on which this change can be observed without rolling an operator I am not permitted to roll.

What stands in for it: the unit tests above are the mechanism proof, including the reconciler-level test in which a pod-spec change produces a second Job under a new name against a fake API, and that test fails on `main` with exactly one Job. The field evidence for the failure mode is the issue's own live observation: on an install carrying the broken Job, an operator roll that changed only the pod spec left the same-named Job crash-looping, and a hand `kubectl delete job` made the operator recreate it, after which it completed in four seconds. That deletion is what this PR makes unnecessary.

Nothing was changed on any cluster. No lease was taken.

## Self-Review

Both passes ran in fresh subagents handed only the worktree path, the range `upstream/main...HEAD`, and the skill file; neither saw the plan, the reasoning, or this body. Findings merged, with dispositions:

- **Adversarial, MEDIUM, confirmed: the builder comment promised a superseded Job "leaves by TTL", which is false for a Job whose pod never runs.** ImagePullBackOff, unschedulable, or admission-rejected pods never reach `Failed`, so `backoffLimit` never counts down, no terminal condition arrives, and `ttlSecondsAfterFinished` never starts. That Job stays until a mode flip or agent deletion and holds one slot of the namespace-wide `pods` quota, and it is exactly the image-override case this digest exists for. Fixed the comment: it now says how each kind of superseded Job leaves, that the never-ran kind does not, what it costs meanwhile, and that deleting superseded generations by label is owed. Not fixed by adding the deletion: this PR was scoped to the naming, a delete in the reconcile loop is a new reconcile-time action with its own hazards (the LOW below is one), and it deserves its own change; a follow-up issue is being filed for it.
- **Adversarial, LOW, confirmed: "a stale failure does not park the phase" was a comment, not a test.** Fixed: `TestReconcileA2ACreatesANewProvisionJobWhenThePodSpecChanges` now stamps `JobFailed=True` on the superseded Job through the status subresource before the second reconcile and asserts the returned state is neither failed nor done, with a guard that the condition persisted (the guard caught a plain `Update` silently dropping status on the fake; the test now registers `Job` as a status-subresource type). A future list-by-label scan that folded conditions across generations would go red here.
- **Adversarial, noted: the commit title said "pod template" while the code digests the whole `JobSpec`.** Fixed the title to "rendered spec".
- **Docs-drift, Advisory: two untouched comments still read "content hash"** (`platformagent_a2a_manifests.go` const block and `cleanupA2A`), and `a2aPasswordDigestNeedles` hardcodes the 8 the new constant names. Fixed the `cleanupA2A` comment. Deliberately left the const-block comment: it stays true (a spec digest is a content hash) and it is the exact line #1272 inserts its constant after, so rewording it manufactures a certain merge conflict for no correction. Deliberately left `a2aPasswordDigestNeedles`: still correct, and #1319 rewrites that function; whichever lands second can pick up the constant.
- **Docs-drift, no Blocking findings.** Checked: every doc mentioning the provision Job (`platformagent-crd.md`, `operator/index.md`, `spec-nats-deployment.md`, `spec-mode-switch.md`, `spec-subagent-profiles.md`, README/INSTALL), the no-credential-in-names rule against the new digest input (the seed password enters the spec only as a `SecretKeyRef` and a shell expansion, and `TestA2AConfigRolloutHashOmitsCredentialsAndTracksRotation` still walks Job names against the needle set), runbook prose for any "delete the Job" instruction (none), goldens and testdata for an embedded Job name (none), docs that enumerate test files (none), the docs map and `AGENTS.md`; `make docs-check` exit 0 at HEAD.
- **Adversarial, clean angles:** the swallowed `json.Marshal` error (a failure would digest identical bytes and fail the pod-spec test); every dynamic digest input traced for time, randomness, and map iteration (only the sorted labels map); red-then-green re-verified in a scratch copy with the digest reverted to script-only; phase stays `Ready` during the post-upgrade re-run; two overlapping provision pods converge on `info || add`; a `k8s.io/api` bump that changes `JobSpec`'s JSON shape renames every install's Job once, one idempotent run each; name length unchanged; no-magic-constants satisfied; #1319's `<agent>-a2a-provision` is a ServiceAccount, a different kind, so no collision.
- **Not covered by either pass:** no live cluster, so the stuck-generation lifecycle in the MEDIUM is from the Kubernetes Job/TTL contract, not observed.

## Risk & Rollout

Blast radius is the provision Job's name under `mode: next`, an unsupported dev toggle. On upgrade, every existing next install gets a new Job name and therefore one fresh provisioning run; the script is `info || add` idempotent, so the run is a no-op on an install whose streams already exist, and the superseded Job is left in place as before: a finished one ages out by TTL, one whose pod never ran does not and holds a pod-quota slot until the follow-up deletion lands or the mode flips. A standing next install already re-runs this Job roughly daily when the TTL clears it, so one extra run on upgrade is within its normal churn. New failure mode to watch for: a non-deterministic digest would create a Job on every reconcile; the determinism test pins against it, and the symptom on a cluster would be a growing `kubectl get jobs -l kubeagents.x-k8s.io/a2a-component=provision` listing. Revert is a plain revert of the commit; the Job name goes back to the script digest and any Job created under the new name finishes and ages out by TTL. Nothing has to happen at merge time.


k8s-operator/internal/controller/platformagent_a2a_provision_job_test.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	script := job.DeepCopy()
+	script.Spec.Template.Spec.Containers[0].Command[2] += "\necho edited\n"
+	if got := a2aProvisionJobName(agent, script.Spec); got == base {
+		t.Errorf("script changed and the name is still %q", got)
+	}
+
+	workdir := job.DeepCopy()
+	workdir.Spec.Template.Spec.Containers[0].WorkingDir = "/tmp"
```
mplakhtiy: **BLOCKER — this case is inert against current `main`, and the test fails.**

The branch's merge base is `1784e18c`, where `buildA2AProvisionJob` set no `WorkingDir` at all, so `"" -> "/tmp"` was a real change and the digest moved. #1272 merged into `main` at `2026-09-10T14:30:59Z` (merge commit `6ea2be259053dd0b102fd47d03be3738d25e9e94`) and the provision container now renders `WorkingDir: a2aProvisionWritablePath`, with `a2aProvisionWritablePath = "/tmp"` (`platformagent_a2a_manifests.go:74`).

Post-merge, this line assigns the value the render already carries. `workdir.Spec` is then structurally identical to `job.Spec`, `json.Marshal` produces the same bytes, `a2aProvisionJobName` returns `base`, and the `t.Errorf` two lines down fires: `WorkingDir changed and the name is still "test-agent-a2a-provision-…"`.

Failure scenario: `Run Controller Tests` is a required check and it reported `success` on this head at `2026-09-10T04:23:35Z`, ten hours before #1272 merged. It checked out a `refs/pull/1388/merge` commit that no longer exists, and GitHub does not re-run a check run without a push — so the green currently on this pull request describes a tree that predates the collision. Approved and lgtm'd as-is, Tide merges on that stale green, the post-merge push runs `k8s-operator-test.yml` for real, and `main` goes red on `TestA2AProvisionJobNameTracksTheScript`.

Nothing about the shipped naming behaviour is wrong; only this assertion is. Pick a value the render does not already use — `"/"` is a real path on the image and is what the gateway container uses — or move the case onto a pod-spec field the render leaves unset. Either way the push is the part that matters, because it is also what re-runs the check against a merge commit containing #1272.

```suggestion
	workdir.Spec.Template.Spec.Containers[0].WorkingDir = "/"
```

mplakhtiy (commented):
Reviewed at head `ea64b3b7f409119ca83fcf7fbe25b81b0f1852ec`. Produced with AI assistance (Claude, Claude Code).

One finding, inline below.

k8s-operator/internal/controller/platformagent_a2a_provision_job_test.go
bnaylor: Confirmed, and it was already broken on the merge rather than only in theory.  Merged `origin/main` in and ran it before touching anything:

```
--- FAIL: TestA2AProvisionJobNameTracksTheScript (0.00s)
    platformagent_a2a_provision_job_test.go:147: WorkingDir changed and the name is still
    "test-agent-a2a-provision-7d6329da"; this is the #1259 fix the old digest could not deliver
```

Exactly as you called it: #1272 set the rendered `WorkingDir` to `/tmp` (`a2aProvisionWritablePath` at `platformagent_a2a_manifests.go:74`, rendered at 918), so the assignment was a no-op and the case was comparing a spec against itself.

Fixed in 98268914, though not quite the way you suggested.  `= "/"` works today, but it has the same shape as the bug -- a literal that goes inert the next time the render moves onto it.  So the edit is derived from what the render produced instead:

```go
workdir.Spec.Template.Spec.Containers[0].WorkingDir += "/subdir"
```

That is always a real change whatever the rendered value is, and it matches the script case immediately above, which appends to `Command[2]` rather than substituting.  The comment there now says why, naming #1272, so the next person to move the workdir doesn't reintroduce it.

Red-proved rather than just re-run green: narrowing `a2aProvisionJobName` back to hashing only the container command makes the case fire again, so it is still testing the pod-spec coverage it claims to.  `go build ./... && go vet ./... && go test ./... -count=1` green in `k8s-operator/`.

Your point about the stale green is the one that mattered here -- `Run Controller Tests` passed on the old head ten hours before #1272 landed, so this would have gone in on a green that no longer described anything.  The push re-runs it against a merge commit that contains #1272.

mplakhtiy (approved):

PR #1390, opened 2026-09-10
feat(terraform): let a second install in one project pick its agent KSA name, guarded to end in -agent

### LINKED ISSUE #1239: second install in one project: distinct GSA names do not separate Workload Identity principals
Follow-up to #1123, which exposes `agent_service_account_id` so a second install in one project does not collide on the agent GSA's fixed name.

## The gap #1123 leaves

Distinct GSA names un-collide **creation**, not **identity**.  The Workload Identity binding member is

```
serviceAccount:${var.project_id}.svc.id.goog[${var.namespace}/${var.ksa_name}]
```

(`terraform/modules/kube-agents-iam/main.tf:10`) — project, namespace and KSA, and no cluster.  Two installs in one project that share a namespace and a KSA name bind the *identical* principal, so each install's agent can mint the other's GSA tokens however differently the GSAs are named.  #1123 documents this rather than fixing it, which was the right call for that PR; this issue is the fix.

## Why `ksa_name` and not the namespace

The namespace looked like the only lever when #1123 was cut, and it is hard-wired in places that make pulling it a much larger change.  `ksa_name` separates the principals without touching that: it is already a variable on the IAM module (`terraform/modules/kube-agents-iam/variables.tf:29`, default `kubeagents-platform-agent`), and `full-install` already uses exactly this pattern for the LiteLLM identity — `local.litellm_ksa` at `main.tf:31`, passed as `ksa_name` at `:269` and as `serviceAccountName` at `:492`.  The agent's `module "kube_agents_iam"` block at `:228` simply does not pass it.

## The part that makes this its own PR rather than a hunk in #1123

**A naive passthrough reintroduces a hole the codebase already documents.**  `charts/kube-agents/templates/agent-rbac-admission-policy.yaml:148` gates the `binds-agent-sa` rule on

```
s.kind == 'ServiceAccount' && s.name.endsWith('-agent')
```

and the comment immediately below it explains that matchConditions are ANDed, so a false one drops the object from the policy entirely.  It records that an earlier revision keyed on a stamped label and calls that "a hole rather than an exemption" for exactly this reason.

So a `ksa_name` not ending in `-agent` silently exempts that install's agent ClusterRoleBinding from the RBAC admission policy.  Unconstrained, this change trades a cross-install privilege crossing for a missing RBAC control on a single install, which is not clearly the better trade and is certainly not something to slip into PR 1 of a series.

## What the change should include

1. An `agent_ksa_name` (or similarly named) variable on `full-install`, passed to the IAM module's existing `ksa_name`, default unchanged so no existing install moves.
2. **A validation constraining it to end in `-agent`.**  This is the reason the issue exists; without it the admission policy stops covering the renamed binding.
3. The Helm `serviceAccountName` (`charts/kube-agents/values.yaml:520`) kept in agreement with it — a mismatch points the WI binding at a KSA that does not exist.
4. Validation evidence: a `terraform plan` transcript unset vs set, **and** evidence the RBAC admission policy still evaluates the renamed binding.  That negative is the point of the change; asserting it is not enough.
5. Drop the "the namespace is the only lever" clause from #1123's README paragraph once this lands, and not before — pointing operators at `ksa_name` while it is unguarded is worse than the current disclosure.

## Interim

#1123's disclosure is accurate and now sits in the README and `terraform.tfvars.example` where an operator reads it, rather than only in a variable description.  Second-install-in-one-project stays a documented configuration with a stated limit until this lands.

Found by `kube-agents-bot` on #1123 (raised twice, HIGH); the admission-policy interaction was found while scoping the fix.

bnaylor:
## Summary

A second install in one project can now give its agent its own Kubernetes ServiceAccount name, so it binds its own Workload Identity principal instead of the first install's. `full-install` gains `agent_ksa_name`, fed from one variable to both consumers of the name — the `kube-agents-iam` module's `ksa_name` (the Workload Identity binding) and the chart's `platformAgent.security.serviceAccountName` (the pod) — with the default `kubeagents-platform-agent` both already used, so no existing install moves. Two `validation` blocks guard it: a DNS-1123 label, and an `-agent` suffix. The suffix is the reason #1239 exists: the chart's `kube-agents-agent-binding-scope` ValidatingAdmissionPolicy selects the bindings it governs by `s.name.endsWith('-agent')` on the bound ServiceAccount, and a matchCondition that evaluates false removes the object from the policy rather than failing it, so a KSA named outside the suffix would silently exempt every binding to it. The README, `terraform.tfvars.example` and the `agent_service_account_id` description now point at `agent_ksa_name` where #1123 left them stating the shared-identity limit.

## Why This Change

#1123 made a second same-project install able to create its GSA, and documented that a distinct GSA name un-collides creation, not identity: the Workload Identity member is `serviceAccount:<project>.svc.id.goog[<namespace>/<ksa>]` (`terraform/modules/kube-agents-iam/main.tf:10`), project-wide with no cluster in it, so two installs sharing the namespace and the default KSA name bind one principal and each agent can mint the other's GSA tokens however differently the GSAs are named. The module already takes `ksa_name` and the composition already uses the pattern for the LiteLLM identity (`local.litellm_ksa`); the agent's module call did not pass it. Passing it unguarded would trade a cross-install identity crossing for a missing admission control on the install that renamed its KSA, which is why the variable and the constraint ship together.

## What Changed

- `terraform/examples/full-install/variables.tf`: `agent_ksa_name`, `nullable = false`, default `kubeagents-platform-agent`. Two validations, each with its own message: the DNS-1123 label regex, and `endswith(var.agent_ksa_name, "-agent")` whose error names the `kube-agents-agent-binding-scope` policy, the `binds-agent-sa` matchCondition, and the policy source file. The comment beside it says what the policy does and does not deny today. The `agent_service_account_id` description's closing clause now says to set `agent_ksa_name` rather than stating the limit.
- `terraform/examples/full-install/main.tf`: `ksa_name = var.agent_ksa_name` on `module "kube_agents_iam"`, and `serviceAccountName = var.agent_ksa_name` under `platformAgent.security` in the `helm_release` values. One variable, two consumers; a mismatch would point the binding at a KSA the pod does not run as.
- `docs/site/src/content/docs/reference/security-and-iam.md` and `INSTALL.md`: the two places that stated the Workload Identity member with the fixed KSA name now say "by default" (found by the docs-drift pass).
- `terraform/examples/full-install/README.md` and `terraform.tfvars.example`: the second-install paragraph, which #1327 rewrote around the three GSA `install.env` keys, gains the KSA half — why the second install needs its own name, why the name must end in `-agent`, and the channel. That channel is `TF_VAR_agent_ksa_name=...` in `install.env`: #1327 moved the GSA names off `TF_VAR_*` onto `PLATFORM_AGENT_GSA_NAME`, and gave the KSA no key of its own, so the passthrough is what is left — and it works because `write_tfvars_from_state()` writes no `agent_ksa_name` line for the file to override. Both documents say so rather than claiming parity with `agent_service_account_id`. The README also says what does not exist yet: `lifecycle.sh` refuses the GSA, the release namespace and the CMEK key names resolving back to their defaults when an `install.env` line is lost; there is no `guard_ksa_identity`, so the KSA resolving back re-shares the identity silently.
- `tests/test_agent_ksa_name_guard.py` (new): parses source, not docs. Holds the root default to the module default and to the chart default; asserts both consumers in `main.tf` read `var.agent_ksa_name` and that the default literal appears nowhere else in the composition; holds the validation's suffix to the `endsWith('...')` literal in `k8s-operator/config/admission/agent-rbac-policy.yaml` and the chart copy; checks the suffix message names the policy; and exercises accept/reject tables for both checks. Fails 12/12 against `upstream/main`'s copies of the files, passes on the branch.

Not changed, on purpose: the module's `ksa_name` carries no `-agent` validation, because the same module is instantiated for the LiteLLM Vertex identity (`kubeagents-litellm`), which is not an agent. No installer script changes: the front doors never hard-code the agent KSA name (the only `kubeagents-platform-agent` literals in `scripts/installer/common.sh` are exported `PLATFORM_AGENT_KSA_NAME` values nothing consumes), and `write_tfvars_from_state()` does not emit `agent_ksa_name`, so `terraform.tfvars` stays silent on the key and the `TF_VAR_` channel works exactly as it does for `agent_service_account_id`.

## Context

- Closes #1239. Follow-up to #1123 (merged 9/8), which exposed `agent_service_account_id` and filed this; #1309 (merged 9/9) added the `guard_gsa_identity` refusal in `lifecycle.sh` and did not touch these lines.
- #1327 (mplakhtiy, "name GSAs, namespace and CMEK keys from install.env and refuse the 409s before the apply") landed first and is merged. `origin/main` is merged in at 4f54579f; the three conflicts it caused are resolved in that commit's message, and every mechanical check below was re-run against the merged head. It moved the GSA names and the namespace into `install.env`/`install.defaults.env`, added `github_minter_service_account_id` and `litellm_service_account_id` beside `agent_service_account_id`, deprecated `TF_VAR_agent_service_account_id` as an installer input in favour of `PLATFORM_AGENT_GSA_NAME`, and pinned the KSA names as `_FIXED` constants — it added no agent `ksa_name` passthrough, so there was no logic overlap to reconcile, only prose. The follow-up it makes natural, and which this PR still does not pre-empt: an `install.defaults.env`/`install.env` key for the agent KSA name on the same footing as the three GSA names, which would retire the `TF_VAR_` line this PR documents.
- Scope of the guard, stated rather than implied: today the binding-scope policy's one validation denies a `ClusterRoleBinding` to `developer-team-agent` and nothing else, so on current `main` a platform KSA outside the suffix loses no denial that would fire today. What it loses is selection: every future validation on that policy, and the `*-agent` convention the policy header, `values.yaml` and the example overlay all rely on, stop reaching that install's bindings. The validation keeps a renamed KSA inside the selector; it does not make the policy stricter.
- Not built here: a `lifecycle.sh` refusal for the KSA name resolving back to its default when the `install.env` line is lost (the KSA analogue of #1309's `guard_gsa_identity`). The failure mode is a silent return to the shared principal rather than a destroyed identity, so it is disclosed in the README and left for its own change.

## Testing

Re-run in full against the merged head (4f54579f), not carried over from the pre-merge branch.

- `terraform init -backend=false` and `terraform validate` on `examples/full-install`: clean, and `terraform validate` on all six modules under `terraform/modules/`: clean. `terraform fmt -check -recursive terraform/`: exit 0. (Terraform v1.15.8.)
- `python3 -m unittest` on the new test plus the suites that read the changed files, including the two #1327 added (`test_service_account_name_wiring`, `test_admission_policy_shipped`, `test_scoped_sa_pool_iam`, `test_agent_iam_ceiling`, `test_lifecycle_script`, `test_reconcile_environment`, `test_vertex_manage_serving_project`, `test_install_script`, `test_upgrade_script`, `test_uninstall_script`, `test_installer_common`): 468 tests, OK. `tests/test_agent_ksa_name_guard.py` alone: 12 tests, OK. The new test run against `upstream/main`'s copies of the files it reads: FAILED (failures=12).
- `make test-python` (the whole sweep) exits 2 on both this head and `origin/main` at 7a322673. The `FAIL:`/`ERROR:` lines are byte-identical between the two runs — `test_check_image_inventory_chart_images`, `test_generate_release_sbom` and `test_eval_dashboard_refresh`, none of which read anything this PR or the merge touches. Pre-existing on `main`, not introduced here.
- `make docs-check`: exit 0 (`AGENTS.md` 37.8k, 0.0k under the 38,000-char budget — the merge did not grow it). `prettier --check` (3.9.6) on `terraform/examples/full-install/README.md`, `INSTALL.md` and the site's `security-and-iam.md`: clean.

### Live validation

Not live-tested as a second install: no second install was stood up, and no cluster was mutated. What was proven instead, each against a real engine rather than a file read:

**The variable reaches the binding, and the default is unchanged (Terraform 1.15.8, plan on a fresh state, `-refresh=false -lock=false`, project `bnaylor-kagents-dev`, scratch tfvars, nothing applied).** Both plans re-run against the merged head 4f54579f, after #1327 added two more GSA variables to the same composition; same member strings, same 29 resources.

```
$ terraform plan ... -var-file=/tmp/ksa-probe.tfvars
      + member = "serviceAccount:bnaylor-kagents-dev.svc.id.goog[kubeagents-system/kubeagents-platform-agent]"
Plan: 29 to add, 0 to change, 0 to destroy.

$ terraform plan ... -var agent_ksa_name=kubeagents-platform-2-agent
      + member = "serviceAccount:bnaylor-kagents-dev.svc.id.goog[kubeagents-system/kubeagents-platform-2-agent]"
Plan: 29 to add, 0 to change, 0 to destroy.
```

**The validations still refuse, with the intended messages** (re-run at 4f54579f through `terraform console`, which evaluates the same variable validations without needing a provider round-trip; the pre-merge run used `terraform plan` and produced the same two errors).

```
$ echo 'var.agent_ksa_name' | terraform console ... -var agent_ksa_name=foo
Error: Invalid value for variable
    │ var.agent_ksa_name is "foo"
agent_ksa_name must end in "-agent": the kube-agents-agent-binding-scope
ValidatingAdmissionPolicy
(k8s-operator/config/admission/agent-rbac-policy.yaml, shipped by the chart)
selects the bindings it governs by matchCondition binds-agent-sa,
s.name.endsWith('-agent'), so a name outside the suffix drops this install's
agent bindings out of the policy rather than failing them.

$ echo 'var.agent_ksa_name' | terraform console ... -var agent_ksa_name=Foo-agent
Error: Invalid value for variable
    │ var.agent_ksa_name is "Foo-agent"
agent_ksa_name must be a DNS-1123 label: lowercase letters, digits and
hyphens, 1-63 characters, starting and ending with a letter or digit.
```

**The chart half.** Not re-run at 4f54579f either, for the same reason: the merge left `charts/kube-agents/values.yaml` and the PlatformAgent template untouched. The `helm_release` values document is unknown at plan time on a fresh state (the credentials block carries generated secrets), so the plan cannot show the leaf. Two things cover it instead: the test asserts `serviceAccountName = var.agent_ksa_name` sits under `platformAgent.security` in `main.tf`, and `helm template` (helm v4.2.4, `--kube-version 1.36.0`) with that value set to the default and to `kubeagents-platform-2-agent` renders the PlatformAgent CR's `spec.security.serviceAccountName` as each in turn, with the Workload Identity annotation beside it.

**The admission-policy negative — the point of the change.** Not re-run against the merged head, and it does not need to be: the merge changed neither `k8s-operator/config/admission/agent-rbac-policy.yaml`, nor the chart's copy of it, nor `charts/kube-agents/values.yaml` (`git diff 34650e1e..4f54579f` over those paths is empty; the only chart files the merge moved are `templates/litellm.yaml` and the chart `README.md`), so the policy this ran against is the policy the branch still ships. Neither shared cluster carries the kube-agents policy (`kubectl get validatingadmissionpolicy` on `bnaylor-ka-test` and `a2a-next-dev` lists only GKE's built-ins; both installs predate it), so a server dry-run there would not reach `binds-agent-sa`. The negative ran instead against a real kube-apiserver (the envtest 1.36.2 assets: etcd + kube-apiserver, RBAC, a `system:masters` token whose username is not the operator exemption) with the policy exactly as `helm template` renders it from this branch's chart:

```
== 0. control: developer-team-agent, the one name the shipped validation denies a ClusterRoleBinding for
Error from server (Forbidden): ... ValidatingAdmissionPolicy 'kube-agents-agent-binding-scope' ... denied request: developer-team is a namespace-scoped tier and may not be granted a ClusterRoleBinding (wrong-scope)
== 1. subject probe-agent: admitted (the shipped validation denies one name only)
== 2. subject probe: admitted too
```

Two admitted requests look alike from outside and the apiserver's `apiserver_validating_admission_policy_*` series record denials and errors only, so steps 1–2 cannot show whether `probe` was evaluated. The discriminating experiment: a second policy whose `matchConditions` are byte-identical to the shipped policy's (checked server-side with a `diff` of the two objects' `.spec.matchConditions`) and whose one validation is widened to deny any subject that *starts with* `developer-team`, suffix or not:

```
-- 4a. subject developer-team-agent: denied (both policies fire)
-- 4b. subject developer-team (no -agent): the widened validation would deny it
clusterrolebinding.rbac.authorization.k8s.io/probe-developer-team created (server dry run)
```

`developer-team` is admitted although the validation names it: the shared `binds-agent-sa` matchCondition removed the object before the validation ran. That is the hole a KSA outside the suffix opens for its own bindings, and what the `-agent` validation refuses at plan time. Every dry-run above used `--dry-run=server`; the apiserver was torn down afterwards. Read-only observations on `bnaylor-ka-test`: the CR's `spec.security.serviceAccountName`, the KSA's Workload Identity annotation and the operator's `kubeagents:minimal:kubeagents-system:platform-agent` ClusterRoleBinding subject all read `kubeagents-platform-agent`, the default this PR keeps.

Not covered: an apply with the variable set (no second install exists to move, and standing one up is outside this PR); the `TF_VAR_agent_ksa_name` `install.env` channel end to end — since #1327 merged, `write_tfvars_from_state()` now emits `agent_service_account_id` and no longer leaves the GSA on the `TF_VAR_` channel, so the KSA's passthrough no longer has a proven sibling. What was checked instead is the one thing that could break it: the generator writes no `agent_ksa_name` line, read off `installer_common.sh` on the merged head. The end-to-end `install.env` → front door → plan path is unrun.

## Self-Review

Passes: `review-adversarial` and `review-docs-drift`, one round each, each in a fresh `general-purpose` subagent handed only the worktree path, the range `upstream/main...HEAD` and the skill file — not this conversation. The fixes below were applied afterwards and every mechanical check re-run against the head. Merged dispositions:

- **#1327 went the other way on the front-door channel (adversarial, HIGH, CONFIRMED — settled by the merge).** #1327 has merged, so this is no longer a forecast. It makes `write_tfvars_from_state` emit `agent_service_account_id` from a `PLATFORM_AGENT_GSA_NAME` key in `install.env`, prints a deprecation for `TF_VAR_agent_service_account_id`, and pins the KSA names as `_FIXED` constants. Resolved rather than deferred: the merge dropped this PR's "on the same terms as `agent_service_account_id`" claim from the variable description, the README and the tfvars comment, because it is no longer true. All three now say `agent_ksa_name` has no `install.env` key of its own and rides `TF_VAR_agent_ksa_name`, and say why that still reaches Terraform — `write_tfvars_from_state()` emits no `agent_ksa_name` line (verified in `scripts/installer/installer_common.sh`), so nothing in the generated `terraform.tfvars` overrides it. `PLATFORM_AGENT_KSA_NAME` is still exported from a `_FIXED` constant and still consumed by nothing (one grep, whole repo: two definitions in `common.sh`, two prose mentions), so the front doors take no action on the name either way.
- **No refusal when the KSA name resolves back to its default (adversarial, MEDIUM, CONFIRMED).** A lost `install.env` line plus `-auto-approve` moves install #2's binding and pod back to the shared principal with no output — the condition this PR removes, reintroduced silently. Still not built here, and the merge sharpened the contrast rather than closing it: `lifecycle.sh` on current `main` now carries three of these refusals (`guard_gsa_identity`, `guard_release_namespace`, `guard_kms_identity`) and no `guard_ksa_identity`. The README says so in those words. A follow-up issue is the right home, and #1327's `state_attr` is now on `main` for it to read.
- **The suffix rationale presented an admission consequence that does not exist today (adversarial MEDIUM and docs-drift Advisory, CONFIRMED).** The shipped policy's one validation names `developer-team-agent` and the operator's own reconcile is exempt, so a platform KSA outside the suffix loses no denial on current `main`; what it loses is selection. Fixed: the `variables.tf` comment and `error_message`, the README, the tfvars comment and the test docstring now say so, and the `error_message` names the policy source file rather than the generated chart copy.
- **`CHART_KSA_ARG_RE` matched the line under any later block (adversarial, LOW, CONFIRMED by the reviewer's mutation).** Fixed: the test now captures the `platformAgent.security` block bounded by its own closing brace; moving the line under `githubMinter` fails the test (re-run of the same mutation).
- **Policy suffix grepped from raw text (adversarial, LOW).** Fixed: the YAML is parsed and `matchConditions` indexed by name, the way `test_admission_policy_shipped.py` does it.
- **One fact in seven homes (adversarial, LOW).** Partly fixed: the two `main.tf` comments now point at the variable's description. The README paragraph and the tfvars comment keep their statement of the mechanism because #1123's review put the second-install caveat where an operator reads it.
- **The test said Kubernetes enforces the label form on ServiceAccount names (docs-drift, Advisory, CONFIRMED).** It allows a subdomain; the variable narrows it. Fixed.
- **The test docstring pointed at a PR transcript (docs-drift, Advisory).** Fixed, sentence dropped.
- **The site's security-and-iam page and INSTALL.md state the member with the fixed KSA name (docs-drift, Advisory).** Fixed with "by default" and the variable name. The docs-site `npm run build` cannot run on this machine (npm pinned to an internal registry); the edit is one parenthetical and CI's site build covers it.
- **README restates `install.env` mechanics whose canonical home is `scripts/installer/README.md`, and suggests a commented `TF_VAR_` pair in `install.env.example` (docs-drift, Advisory).** Not fixed. The paragraph is #1123's, reviewed into that place three days ago; this PR extends it by one key. Every commented key in `install.env.example` is an installer key whose default comes from `install.defaults.env`, and a raw `TF_VAR_` passthrough is a different kind of line (#1123 kept it out of that file for the same reason); #1327 is currently rewriting that section.
- **Angles that produced nothing:** `endswith()`/`nullable` against the `~> 1.5` floor; the label charset cannot break the `[ns/ksa]` member or a `system:serviceaccount:` principal; default equals the module and chart defaults (test-pinned); the binding's namespace half tracks `var.namespace`; `PLATFORM_AGENT_KSA_NAME` in `common.sh` is consumed by nothing, so the front doors take no post-apply action on the name; `write_tfvars_from_state` emits neither key; the operator creates the KSA under `spec.security.serviceAccountName` (`reconcileServiceAccount`), so chart value and module binding name the same object; the webhook's `restrictedServiceAccounts` is a literal list with no interaction; the stockout plugin reads the KSA off the CR; test placement is under a `PYTHON_TEST_DIRS` glob with PyYAML installed by the CI jobs; each test goes red on the mutation it names; real Terraform in a scratch root rejected `kubeagents-platform-2`, `-agent` and a 64-character name and accepted a 63-character one; the docs map is not stale and no other document restates the old wording; the `lifecycle.sh` claim (GSA refused, KSA not) verified against `guard_gsa_identity`.

## Risk & Rollout

No runtime code paths touched. The default is the name both consumers already used and the plan transcript shows the binding unchanged with the variable unset, so existing installs are unaffected; an install that sets the variable moves its KSA, its Workload Identity binding and its pod together on the next apply, which is the intended effect and the only new behaviour. The validations can only refuse a plan that names a KSA outside the suffix or outside a DNS label; no existing install can be doing that, because the name was not configurable. Revert is a single-commit revert (plus the `origin/main` merge, which carries no changes of its own). Nothing has to happen at merge time: #1327 landed first and is merged in here, so the four-file overlap it created is already resolved on this branch.


terraform/examples/full-install/README.md
```diff
@@ -142,24 +142,34 @@ plans the whole composition as new and reads as total drift. A gitignored
...
+mint the other's GSA tokens. `agent_ksa_name` feeds both the module's binding
+and the chart's `serviceAccountName`, so the pod and the binding move
+together, and it must end in `-agent`: the `kube-agents-agent-binding-scope`
+admission policy the chart ships selects the bindings it governs by that
+suffix on the bound ServiceAccount, so a name outside it would leave this
+install's agent bindings unselected by that policy and by any validation it
+gains; the validation in `variables.tf` refuses the plan instead, and the
+variable's description says what the policy does and does not deny today.
+Through the installer front doors both values are
+`TF_VAR_agent_service_account_id=...` and `TF_VAR_agent_ksa_name=...` lines in
```
jayantid: 🟡 **Not merge-blocking on its own, but the rebase this PR already needs has to rewrite this paragraph, not just resolve the hunks.**

#1327 merged on 9/9, after this branch was cut, and `main` conflicts here, in `terraform.tfvars.example`, and in `variables.tf`. On `main` the GSA name is an installer key: `write_tfvars_from_state` writes `agent_service_account_id` into `terraform.tfvars` from `PLATFORM_AGENT_GSA_NAME` (`scripts/installer/installer_common.sh:1653`), and `TF_VAR_agent_service_account_id` now prints a deprecation warning (`installer_common.sh:497`). Lines 160-170 here, `terraform.tfvars.example:61-62` and the `agent_ksa_name` description at `variables.tf:158` all describe the pre-#1327 channel.

After a mechanical rebase an operator reads: put both names in `install.env` as `TF_VAR_` lines, "on these keys it stays silent". The GSA half is now wrong. The KSA half still works only because the generator does not emit `agent_ksa_name`, and it gets none of what the GSA keys got in #1327: no pre-apply collision check, no `lifecycle.sh` guard, no Day-2 menu entry. Nothing in the text says so.

Rewrite against `main`'s channel and name the asymmetry:

```
Through the installer front doors the GSA is PLATFORM_AGENT_GSA_NAME in
install.env. The KSA is a raw Terraform variable the installer does not
know about: a TF_VAR_agent_ksa_name=... line in install.env reaches the
apply, but gets no collision check, no lifecycle.sh guard and no Day-2
menu entry.
```

Then file the follow-up the description already names (`PLATFORM_AGENT_KSA_NAME` in `install.defaults.env`, replacing the `_FIXED` constant at `scripts/installer/common.sh:72`) and update the description's Context, which still calls #1327 unmerged.

terraform/examples/full-install/README.md
```diff
@@ -163,12 +164,28 @@ so do the CMEK key ring and key names (`GKE_DB_KMS_KEYRING` / `GKE_DB_KMS_KEY`,
...
+`kube-agents-agent-binding-scope` admission policy the chart ships selects the
+bindings it governs by that suffix on the bound ServiceAccount, so a name
+outside it would leave this install's agent bindings unselected by that policy
+and by any validation it gains. The validation in `variables.tf` refuses the
+plan instead, and the variable's description says what the policy does and does
+not deny today. `agent_ksa_name` has no `install.env` key of its own yet, so
+through the front doors it is a `TF_VAR_agent_ksa_name=...` line in that file:
+every front door sources it with `set -a`, and the generator never writes
+`agent_ksa_name` into `terraform.tfvars`, so the passthrough is what Terraform
+reads. The caution above applies to it with one gap — a lost line resolves the
```
jayantid: 🟡 **Not merge-blocking.** The gap this paragraph discloses is wider than "a lost line": any direct `lifecycle.sh` run applies the default KSA name with the `install.env` line intact. One front-door re-run reverses it.

**What's wrong.** `lifecycle.sh` never reads `install.env`. It sources only `install.defaults.env` (`lifecycle.sh:79-82`; `scripts/installer/README.md:68` says so). The generated `terraform.tfvars` carries the three GSA names and no `agent_ksa_name` (`installer_common.sh:1337-1339`). So `TF_VAR_agent_ksa_name` reaches Terraform only inside a front-door process.

**What happens if it merges.** Install #2 puts the line in `install.env` and answers "n" at the confirmation. The installer prints `cd terraform/examples/full-install && ... ./lifecycle.sh apply` as the resume command (`install.sh:3711`). That run binds `[kubeagents-system/kubeagents-platform-agent]` and runs the pod as it: the identity crossing this PR closes, with no warning. The drift check at line 213 and the recovery re-apply at line 204 do the same to an existing install #2. `guard_gsa_identity` catches the GSA analogue. Nothing catches this.

**The fix.** Name the real trigger here, and give the key a home both paths read by emitting it from the generator beside the GSA lines:

```bash
# scripts/installer/installer_common.sh, write_tfvars_from_state
echo "agent_ksa_name = $(hcl_str "${TF_VAR_agent_ksa_name:-$PLATFORM_AGENT_KSA_NAME_FIXED}")"
```

jayantid (approved):

PR #1393, opened 2026-09-10
fix(operator): scope the worker's JetStream API grant to the streams and verbs it uses

### LINKED ISSUE #1316: A2A worker holds $JS.API.> and can purge or delete the DIRECTORY stream
`$JS.API.>` on the A2A worker covers PURGE, UPDATE and DELETE on every stream, including DIRECTORY.

The worker user in the rendered `nats.conf` (`k8s-operator/internal/controller/platformagent_a2a_manifests.go`, the worker publish allow-list) holds the wholesale JetStream API wildcard. `worker` is the identity a session pod runs as, which is the least-trusted principal in the deployment: it executes model output against untrusted input.

The wildcard covers, among everything else:

```
$JS.API.STREAM.PURGE.DIRECTORY
$JS.API.STREAM.UPDATE.DIRECTORY
$JS.API.STREAM.DELETE.DIRECTORY
$JS.API.STREAM.MSG.DELETE.DIRECTORY
```

### Measured, not inferred

The post-#1313 worker permission set was rendered into a `nats.conf`, run against an embedded nats-server with JetStream, with `DIRECTORY` provisioned as the provision Job creates it (`a2a.agents.>`, `max_msgs_per_subject: 1`). Connected as `worker`:

| Call | Result |
| --- | --- |
| core publish `a2a.agents.platform` | refused |
| JetStream publish `a2a.agents.platform` | refused |
| `STREAM.INFO DIRECTORY` | allowed |
| `STREAM.PURGE DIRECTORY` | allowed |
| `STREAM.UPDATE DIRECTORY` (add a subject) | allowed |
| `STREAM.DELETE DIRECTORY` | allowed |
| `CONSUMER.CREATE DIRECTORY` deliver=`_INBOX.worker.peek` | allowed |

Identical on nats-server 2.11.4 and 2.14.5. Not run on 2.10, which is what `platformagent_a2a_manifests.go` deploys.

One `STREAM.PURGE.DIRECTORY` empties the directory for every profile, and no card publisher exists to repopulate it. `STREAM.DELETE.DIRECTORY` is worse: only a re-run of the provision Job restores the stream.

### Why it is not part of #1313

#1313 removes the worker's `a2a.agents.>` publish grant, which closes forgery. It does not touch the wildcard, so the destructive half survives. The comment on that PR now says so rather than claiming the directory is out of reach.

Scoping is feasible from the same evidence. The worker's real JetStream use is TASKS (`a2a/lib/client.go` `CreateOrUpdateConsumer`, `a2a/hermes-bridge/sweep.go` `js.Stream(TasksStream)`, `a2a/lib/fold.go`'s ordered consumer) and the KV bucket (`a2a/hermes-bridge/bridge.go`). Nothing in the worker path touches DIRECTORY at all: no `STREAM.INFO.DIRECTORY`, no consumer on it.

### Why now

#1306 does exactly this narrowing for the `seed` user, and seed is the *more* trusted of the two. Leaving the wildcard on worker means the least-trusted principal holds a grant already argued unacceptable for a more-trusted one.

Not a duplicate of #1306 (seed block only) or #1313 (the publish grant only). The consumer-create case is worth its own look while narrowing: a consumer whose deliver subject the worker cannot publish sat at `delivered=0` on 2.11.4 and 2.14.5, but that gate is version-sensitive and 2.10 is what ships.

bnaylor:
## Summary

The NATS `worker` user - the identity a session pod runs as, and the least-trusted principal in the deployment - held `$JS.API.>` wholesale, which covers `STREAM.PURGE`, `STREAM.UPDATE`, `STREAM.DELETE` and `STREAM.MSG.DELETE` on every stream, DIRECTORY included. This replaces the wildcard with the JetStream API subjects the worker-side binaries actually emit, by stream and by verb, rendered from one function (`a2aWorkerJetStreamGrants`). Nothing in the list names DIRECTORY, `KV_session-state` or `KV_cap`; no destructive verb survives on any stream. The proof is an embedded nats-server running the rendered `nats.conf`: 24 worker operations succeed with zero publish violations, 25 out-of-scope calls are refused server-side, and the same render with the wildcard put back lets the worker delete the directory. The two routes the grant narrows but cannot close - a push consumer's `deliver_subject`, and `CONSUMER.CREATE`'s create-or-update on a stream another principal has a durable on - are measured, bounded and recorded rather than claimed shut.

## Why This Change

#1313 removed the worker's `a2a.agents.>` publish, which closed card forgery. It did not touch the wildcard, so the destructive route survived: one `STREAM.PURGE.DIRECTORY` empties the directory for every profile and nothing repopulates it, and `STREAM.DELETE.DIRECTORY` leaves only a re-run of the provision Job to bring the stream back. #1316 measured both against the rendered config. The same grant reached `KV_session-state` (the gateway's session registry, the web finding's route) and every stream's `UPDATE`, `RESTORE`, `MSG.DELETE` and `SNAPSHOT`. #1306 narrows the same wildcard for `seed`, which is the more trusted of the two; leaving it on the worker meant the least-trusted principal held a grant already argued unacceptable for a more trusted one.

## What Changed

- `a2aWorkerJetStreamGrants()` is the list, and its comment is the argument for every entry. Read out of nats.go v1.53.1 for each call the worker path makes (`lib.Client` durable consumer, `lib.TasksGet` replay, the bridge's KV registry and sweep, the `a2a topics` CLI), then measured. TASKS: `STREAM.INFO`, `CONSUMER.CREATE.TASKS.>`, `CONSUMER.MSG.NEXT.TASKS.*`, `DIRECT.GET.TASKS.>`. `KV_runtime-state`: `STREAM.INFO`, `CONSUMER.CREATE.>`, `CONSUMER.DELETE.*` (`kv.Keys` is a push ordered consumer nats.go creates and deletes on `Unsubscribe`; the bridge sweep runs it at every start). TOPICS-STATE and TOPICS-JOURNAL: `STREAM.INFO` and `DIRECT.GET.>`, which is what `a2a topics read` / `list` emit under whichever pod carries the worker credential. Eleven subjects.
- Withheld on purpose, each with its reason in the comment: account `INFO` (`jetstream.New` never asks), `CONSUMER.INFO` (nothing binds a consumer by name), `STREAM.MSG.GET` (every stream is created with `allow_direct`, so nats.go never emits the fallback), `DIRECT.GET` on the KV bucket (nothing on the path calls `kv.Get` - the bridge puts, lists and deletes, and the worker adapter touches no bucket), and `CONSUMER.DELETE` on TASKS. That last one nats.go does emit, from the ordered consumer's best-effort reset, and ignores the result. Withholding it raises the price of reaching another principal's durable and does not close the route: `CONSUMER.CREATE` is create-OR-UPDATE by name, and the server has no ownership concept for a consumer name, so `CONSUMER.CREATE.TASKS.>` reaches the gateway's relay durable - measured, both retuning it and having the server reap it, in a run where `CONSUMER.DELETE` on the same name was refused (table below). No narrower grant exists: nats.go's ordered consumers take server-generated names, so the last token has to be `>`, and NATS wildcards match whole tokens. `DELETE` stays out as the one destructive verb here that nothing on the worker path needs.
- The provision script now says `--allow-direct` on each `stream add`. It was already natscli's default (v0.1.5, `stream_command.go:206`), but the worker's grant rests on it - nats.go picks `DIRECT.GET` or `STREAM.MSG.GET` from the stream's own config - so the bit is set in the script rather than inherited from a CLI the deployer can swap via `A2A_PROVISION_IMAGE`. The script bytes change, so the next reconcile on an existing install creates a new provision Job under a new hash; the script is idempotent (`stream info || stream add`) and touches no existing stream.
- `TestWorkerJetStreamGrantOnARealServer` (new file, `platformagent_a2a_worker_authz_test.go`): the rendered config under an embedded nats-server, seven streams provisioned as seed with the provision script's flags, a client as worker. Table 1 runs the worker's real operations and asserts zero `Publish Violation` lines for worker in the server log; table 2 sends the destructive and out-of-scope calls and asserts each refusal from the server log naming the exact subject, then re-reads the streams as seed to show nothing changed. Two subtests then measure what a refusal table structurally cannot - the routes that run over subjects the grant PERMITS, and so leave no violation to read: the deliver-subject residue, four ways, and `CONSUMER.CREATE` as an update against the gateway's durable (both below). A final subtest restores `$JS.API.>` in the render and shows `PURGE` and `DELETE` on DIRECTORY succeeding and the stream gone. The test logs the embedded server version beside the image tag the operator deploys, since the two pins are independent.
- `TestWorkerConsumersSurviveABusRestart`: the reconnect canary. The bridge's durable under `Consume`, an ordered `Messages()` iterator (the replay path) and `kv.Keys` cross a server shutdown and restart on the same port and store, connected as worker; the only violation the restarted server may log is the ordered reset's fire-and-forget `CONSUMER.DELETE.TASKS.*`, and any `CONSUMER.INFO` fails the test with the subject named.
- `TestWorkerHoldsNoWholesaleJetStreamAPI`: the shape check, in #1306's form - the worker's publish list pinned exactly, every forbidden verb against every provisioned stream through `subjectMatches`, a structural bound on the grant function's own output, and the subscribe list pinned exactly too. `TestWorkerGrantNamesStreamsTheProvisionScriptCreates` holds the stream-name constants the grant renders from to the script's `stream add` / `kv add` lines, and every `stream add` to `--allow-direct`.
- `a2aNATSConfGrantLines(grants)` is the one renderer for allow-list lines. Seed's private copy came with #1306 and is gone as of the merge below; both users render through this.
- `k8s-operator/go.mod`: `github.com/nats-io/nats-server/v2 v2.14.5` and `github.com/nats-io/nats.go v1.53.1` as test dependencies, the versions `a2a/go.mod` already pins. `go list -deps ./cmd/...` contains no `nats-io` package; the operator binary does not link them.
- Docs: `a2a/docs/hermes-bridge.md` no longer lists `$JS.API.>` as the worker's tax; `docs/designs/spec-nats-deployment.md`'s JetStream-tax bullet no longer calls the tax `$JS.API.>`, its stale "unscoped ack form" sentence now matches the render, and its deliver-subject residue states the interest condition and the create-as-update route measured here.

## Context

- Closes #1316.
- `origin/main` merged in `e7998916` (no rebase: this is security work and its history is deliberate). One conflict, add/add in the `const` block of `platformagent_a2a_manifests.go`: this branch's stream-name constants and `a2aNATSConfGrantLine` against main's `a2aProvisionWritablePath` from #1272. Both kept. #1272 also gives the provision container a `WorkingDir`; nothing on this branch asserted its absence, so no test went inert across the merge - checked, because that is exactly what happened in #1388.
- #1306 (merged) did the same narrowing for the `seed` user in the same file and test file; this PR reuses its shape (`a2a<User>JetStreamGrants`, exact pin + cross-product + structural bound, a pairing test against the script). The three collisions this bullet used to record as owed are discharged in merge `a8a3b1ab`. (1) add/add after `var a2aCredsKeys` - both kept. (2) the `renderA2ANATSConf` doc comment now says gateway alone holds `$JS.API.>` and names both grant functions. (3) `a2aSeedJetStreamGrantLines()` is gone and seed renders through the generic `a2aNATSConfGrantLines(grants)`. That collapse is byte-identical because both sides emitted `a2aNATSConfGrantLine`, and `TestSeedHoldsNoWholesaleJetStreamAPI` pins seed's allow-list exactly, so a whitespace drift would have shown up as a failure rather than as nothing. It passes. #1306's pairing test matches `stream add <name> `, which the `--allow-direct` insertion after the name does not break.
- One test fix falls out of that merge rather than out of the collapse. `TestWorkerJetStreamGrantOnARealServer` reads state back as seed, which was fine while seed held `$JS.API.>`. #1306 scoped seed to `STREAM.CREATE` and `STREAM.INFO` on the streams it provisions, and one of those reads - `GetLastMsgForSubject` on the probe topic - asks for a stored message. Reading content is not the provisioning identity's job, so the read moves to the gateway rather than the grant widening to admit it. Every other read in the test is a `STREAM.INFO`, which seed still holds. It presented in the shape that costs the most time here: the refused request did not error, it hung until the test's shared 60s context expired, which then failed the two later subtests on a dead context. Three failures, one refused subject.
- #1313 (merged) is the forgery half; its comment pointed here.
- #1319 (open, auth callout) and #1258 (worker adapter, unopened) touch the same file or the same credential; the adapter's JetStream use is ordered consumers on TASKS, covered by this list.
- Gateway still holds `$JS.API.>`; it owns the relay durable and the session registry, so its narrowing is the same change with a different table. Not filed.
- Conformance (#1121): no `tests/conformance/` assertion owed; this narrows an existing grant and moves no invariant boundary.

## Testing

- `cd k8s-operator && go build ./... && go vet ./... && go test ./... -count=1`, after the merge and after the review fixes: green. `internal/controller` verbose: 413 top-level tests, 713 counting subtests, 0 SKIP, 0 FAIL, 21s. `make fmt` reformats the untouched `netpolprofile.go`; that churn is not in this diff.
- `python3 tests/conformance/run.py`: `Ran 108 tests ... OK (expected failures=12)`.
- `cd a2a && go vet ./... && go test ./... -count=1`: green, unchanged. No a2a test runs a worker against a rendered config, so the embedded-server test here is the only mechanism proof; the a2a suite is the regression canary for the client library's call shapes.
- Red proof of the grant: with the render carrying main's `"$JS.API.>"` in the worker block and everything else from this branch, both new tests fail - the shape test on the exact pin and 112 forbidden subjects, the server test with 23 of the 24 calls in its table at the time reported `ALLOWED` (the 24th, `DIRECT.GET` on DIRECTORY, answered no responders because the `DELETE` two rows earlier had already removed the stream).
- Red proofs of the two subtests below: substituting a literal subscription for the wildcard one in deliver-subject case 3 fails it with `DIRECTORY went 0 -> 2`; pointing the create-as-update subtest at `$JS.API.CONSUMER.DELETE.TASKS.gateway-relay` fails it on the refusal (`no reply (nats: timeout)`).
- Versions: the server test runs nats-server v2.14.5 (the test dependency); the operator deploys `nats:2.10-alpine`, which is 2.10.29 on a2a-next-dev today. The same measurements run against an embedded 2.10.29 (a scratch module outside the repo, reading the dumped renders) give the identical result on every row, the two subtests below included.
- `make docs-check` clean (38.0k of the 38,000-char budget; AGENTS.md is not touched); `prettier@3.9.6 --check` clean on both changed Markdown files.

Embedded server, worker connected with `--inbox-prefix _INBOX.worker`, rendered config unchanged except ports and store dir:

| Worker operation (nats.go call the binaries make) | 2.14.5 | 2.10.29 |
| --- | --- | --- |
| `STREAM.INFO TASKS` (`js.Stream`) | allowed | allowed |
| `CONSUMER.CREATE TASKS` durable `bridge-platform`, twice (create, then the rebuild's update) | allowed | allowed |
| `CONSUMER.MSG.NEXT TASKS` (Fetch), `$JS.ACK` (ack floor read back as gateway = 1) | allowed | allowed |
| publish `a2a.tasks.platform.t1.events` with dedup id; then with expected last sequence (sweep CAS) | allowed | allowed |
| `DIRECT.GET TASKS` last-for-subject (replay horizon) | allowed | allowed |
| ordered consumer on TASKS, two replay messages | allowed | allowed |
| `STREAM.INFO KV_runtime-state` (`js.KeyValue`), put, `kv.Keys` (consumer create + delete), delete | allowed | allowed |
| `STREAM.INFO` with subject filter and `DIRECT.GET` on TOPICS-STATE and TOPICS-JOURNAL, after a publish to each | allowed | allowed |
| publish violations logged for worker across all of the above | 0 | 0 |
| server shutdown and restart (same port, same store) under a live durable `Consume`, an open ordered `Messages()` and then `kv.Keys`: all three resume; violations after the restart | 1, the ordered reset's `CONSUMER.DELETE.TASKS.*`; no `CONSUMER.INFO` | not run |

| Refused call (server log: `Publish Violation`, user worker, exact subject) | 2.14.5 | 2.10.29 |
| --- | --- | --- |
| `STREAM.INFO` / `PURGE` / `UPDATE` / `DELETE` / `MSG.DELETE` DIRECTORY | refused | refused |
| `CONSUMER.CREATE DIRECTORY` deliver `_INBOX.worker.peek`; `DIRECT.GET DIRECTORY` | refused | refused |
| `STREAM.PURGE` / `UPDATE` / `DELETE` / `MSG.DELETE` / `RESTORE` / `SNAPSHOT` TASKS | refused | refused |
| `CONSUMER.DELETE` and `CONSUMER.INFO` on the gateway's `gateway-relay` durable | refused | refused |
| `DIRECT.GET KV_runtime-state` (no caller); `STREAM.INFO`, `CONSUMER.CREATE`, `DIRECT.GET` on `KV_session-state`; `STREAM.INFO KV_cap` | refused | refused |
| `$JS.API.INFO`, `STREAM.NAMES`, `STREAM.LIST`, `CONSUMER.NAMES TASKS`, `STREAM.CREATE EVIL` | refused | refused |
| afterwards, as seed: DIRECTORY subjects unchanged and 0 msgs, TASKS 3 msgs, relay durable present | yes | yes |

### The two routes the grant does not close

Both run over subjects the allow-list permits, so the refusal table cannot see either by construction. Both measured on both server versions.

**A push consumer's `deliver_subject`.** `CONSUMER.CREATE` is scoped to TASKS, but the deliver subject is a body field no subject grant can see; the server delivers TASKS messages onto it, and any stream whose subjects cover it stores them under their original `a2a.tasks.*` subjects. Delivery starts only once a subscription exists whose subject is **exactly** the deliver subject: a push consumer registers through `Sublist.registerNotification` (`server/sublist.go`, called from `server/consumer.go`), which takes interest only from a subscription byte-equal to the deliver subject - `if string(sub.subject) == subject` - and says so in its own doc comment ("this interest needs to be exact and that wildcards will not trigger the notifications"). Byte-identical in 2.10.29 and 2.14.5.

| Case | Result (both versions) |
| --- | --- |
| consumer aimed at `a2a.agents.platform`, nobody subscribed | created, `num_pending 3`, delivers nothing; DIRECTORY 0 msgs |
| consumer aimed at the writerless probe topic, nobody subscribed | TOPICS-STATE grew 1 to 3; both new messages under `a2a.tasks.*` subjects; `DIRECT.GET` of the probe subject still not found. The stream's own literal ingest is the interest. |
| consumer aimed at `a2a.agents.platform`, the gateway subscribed to `a2a.agents.>` - its whole subscribe grant, and inside web's `a2a.>` | DIRECTORY 0 -> 0. A wildcard supplies no interest, and that is what bounds this residue. |
| consumer aimed at `a2a.agents.platform`, the gateway subscribed to that subject itself | DIRECTORY 0 -> 2, both under `a2a.tasks.*` subjects; no card subject written |

So: not forgery (a topic or card read by subject never sees these), but a persisted write into streams the worker cannot publish to and an eviction lever against them (`discard=old`, 1 GiB caps). The wildcard this replaces had the same route with every stream as a source. It survives per-stream scoping of any user that may create consumers at all; the close is the worker not holding `CONSUMER.CREATE` - a consumer created per task by the dispatcher - not a grant. The precondition for the identity-plane half is a principal subscribed to a card subject itself, and nothing in the tree opens one: a directory watcher on `a2a.agents.>` is not it, measured. The subtest asserts rows 1 and 3, asserts the no-forgery property of rows 2 and 4, and logs the counts. I tried the containment the adversarial pass proposed (dropping `a2a.topics.>` from the worker's subscribe list) and the test refuted it: the worker's literal subscription was refused and TOPICS-STATE still grew, because the interest was the stream's own. The subscribe list is unchanged and now pinned.

**`CONSUMER.CREATE` is create-or-update by name.** The request's `action` field is empty for both, and the server has no ownership concept for a consumer name, so within a stream the worker may create consumers on, every consumer on that stream is the worker's to reconfigure - the gateway's relay durable included. Measured as `worker` against the render:

| Call on `$JS.API.CONSUMER.CREATE.TASKS.gateway-relay` | 2.14.5 | 2.10.29 |
| --- | --- | --- |
| the durable's own config with `filter_subject: a2a.tasks.none` | allowed by the grant, applied by the server; the relay's filter subject reads back `a2a.tasks.none`, no violation logged anywhere | same |
| the durable's own config with `inactive_threshold: 1s` | the server reaps the durable; `gw.Consumer(ctx, "TASKS", "gateway-relay")` answers `consumer not found`, ack floor and all | same |
| `$JS.API.CONSUMER.DELETE.TASKS.gateway-relay`, same run | refused, `Publish Violation` in the server log | same |
| a partial body naming only `filter_subject` | rejected by the SERVER, not by the grant: `ack policy can not be updated`, because the zero value of `ack_policy` is `none`. An update carries the whole config. | same |

Not a regression - `$JS.API.>` permitted all of it - and no narrower grant is available, because nats.go's ordered consumers take server-generated names, so the last token has to be `>`, and NATS wildcards match whole tokens. What withholding `CONSUMER.DELETE.TASKS.*` buys is the price of the route, not its closure, and the grant comment now says that instead of the opposite.

Control, same render with `"$JS.API.>"` restored in the worker block: `STREAM.PURGE.DIRECTORY` answers `{"success":true,"purged":0}`, `STREAM.DELETE.DIRECTORY` answers `{"success":true}`, and seed's `STREAM.INFO.DIRECTORY` then returns stream not found. That is #1316 reproduced on both server versions.

### Live validation

Not live-tested against a running install. The only `mode: next` deployment (`gke_bnaylor-kagents-dev_northamerica-northeast1_a2a-next-dev`) carries another session's stacked build and was read-only for this change; rolling an operator there was not an option, and the default-mode cluster renders no A2A bus.

Observed there without mutating, 2026-09-10:

- `secret/platform-agent-a2a-nats-config`, worker block: `"$JS.API.>"` is live today (and `"a2a.agents.>"` too - that build predates #1313).
- `/varz` over a port-forward: nats-server `2.10.29`, image `nats:2.10-alpine`.
- `/jsz?streams=true&config=true`: all seven streams present with `allow_direct=true` (DIRECTORY `deny_delete=false`, `deny_purge=false`; TASKS has 5 consumers). This is the store state the grant is written for.

The mechanism proof is the embedded-server tables above, on the deployed server version and on the test dependency's. Not covered: the real `natsio/nats-box:0.14.5` image running the amended script (the test provisions through nats.go with the script's flags; `--allow-direct` exists on `stream add` in natscli v0.1.5 by source, not by execution), a session pod actually spawned by the gateway under the new grant, and the worker adapter (#1258) end to end - its call set was read (ordered consumers on TASKS) and is covered, and if a later revision adds a call the list misses, the server test reports it as the refused subject.

## Self-Review

Two subagent passes over the diff, adversarial and docs-drift per AGENTS.md, each handed the worktree path, the range `main...HEAD` and the skill file and nothing else; then the review findings on the open PR, folded in below by disposition rather than by round.

**Adversarial** looked for: a JetStream API subject the worker path emits that the list misses (it enumerated every template in nats.go's `api.go` against every call site and found none), a route the narrowed grant leaves open, whether the withheld `CONSUMER.DELETE` leaks consumers against `max_consumers`, whether the test can see a re-widening, the dependency footprint, and the relationship to #1306. It ran a real nats-server on both 2.10.29 and 2.14.5 for its own measurements. Findings:

1. **Fixed, and it changed the claim.** HIGH: my first residue paragraph said a push consumer aimed at the directory "never delivers", and the subtest pinned that. The pass showed delivery is gated on a subscription on the deliver subject and measured TASKS messages landing in TOPICS-STATE and, with a gateway card subscription, in DIRECTORY. I reproduced both. Its proposed containment - drop `a2a.topics.>` from the worker's subscribe list, since nothing on the path core-subscribes to topics - I applied and then measured wrong: the worker's literal subscription was refused and TOPICS-STATE still grew, because a literal-subject stream's own ingest is the interest. Reverted. The comment, the subtest, the spec and this body now state the measured mechanism and the property that holds (no forgery: original subjects preserved, reads by subject unaffected), and name the real close (no `CONSUMER.CREATE` on the worker at all). The subscribe list is pinned so a widening is a diff.
2. **Fixed.** MEDIUM: the grant-line renderer duplicated #1306's byte for byte, and the `renderA2ANATSConf` comment named #1306 in a way that goes stale when either merges first. Now one generic `a2aNATSConfGrantLines(grants)`; the comment describes the state without naming the sibling.
3. **Fixed.** LOW: the stream-name constants claimed to be "spelled as the provision script creates them" with nothing holding the two together. `TestWorkerGrantNamesStreamsTheProvisionScriptCreates` now anchors each constant to a `stream add` / `kv add` line and bounds every grant to those names.
4. **Fixed.** LOW: the grant rests on `allow_direct`, which the script inherited from a CLI default. The script now says `--allow-direct` on every `stream add`, and the pairing test checks it.
5. **Fixed.** LOW: nothing named the server version a test run measured. The test logs `natsserver.VERSION` beside `defaultA2ANATSImage`.

**Review findings on this PR**, two from @mplakhtiy and one from kube-agents-bot:

1. **Confirmed and fixed** (@mplakhtiy, on the grant comment): the comment said withholding `CONSUMER.DELETE.TASKS.*` keeps the worker off another principal's durable, when `CONSUMER.CREATE` is create-or-update by name and reaches it anyway. Measured both outcomes against the render - the retune, and the reap through `inactive_threshold` - in a run where `CONSUMER.DELETE` on the same name was refused; the comment, the spec and the table above now say what the withholding buys and what it does not. One correction to the finding that does not change it: its illustrative partial body is rejected by the server (`ack policy can not be updated`), not by the grant, so the test sends whole configs. New subtest, red-proved.
2. **Investigated and refuted; the case was added anyway** (@mplakhtiy, on the same comment, and as a second thread on the same sentence in `spec-nats-deployment.md`): both said a wildcard subscription covering the deliver subject - `a2a.agents.>`, the gateway's whole subscribe grant, and inside web's `a2a.>` - supplies the interest a push consumer needs, so "a literal subscription" understated the residue. Ordinary NATS interest matching does resolve wildcards; consumer interest registration is the exception. `Sublist.registerNotification` takes interest only from a byte-equal subject and documents that wildcards will not trigger it, identically in 2.10.29 and 2.14.5, and the measurement agrees: DIRECTORY 0 -> 0 with the wildcard subscriber, 0 -> 2 with the literal one, on both versions. The case is in the subtest regardless, because that exactness is now the only thing bounding this residue and it should not rest on a generalisation from three literal cases; the comment and the spec name the mechanism and the two wildcard grants explicitly. Both threads answered with the measurement.
3. **Not fixed - factually wrong for the pinned client** (kube-agents-bot, High, on the grant comment): withholding `CONSUMER.INFO.TASKS.*` would kill the bridge's durable after any reconnect, because `Consume`/`Messages` re-verify the consumer with `Consumer.Info()` on CONNECTED-after-RECONNECTING. In nats.go v1.53.1 (`a2a/go.mod`) the `Consume` status loop (`jetstream/pull.go:358-380`) re-issues a pull on CONNECTED and calls no `Info()`; `Messages()` maps the status to `errConnected` and `Next` resets counters; the ordered consumer's handler treats `errConnected` as a reset, which is `CONSUMER.CREATE` (granted). The quoted retry loop is not in this version - `fetchConsumerInfo` has two callers, `js.Consumer` / `stream.Consumer` bind-by-name, which nothing on the worker path calls. Held by execution in `TestWorkerConsumersSurviveABusRestart` (row above): durable `Consume`, ordered `Messages()` and `kv.Keys` all resumed across a restart under the scoped grant, with zero `CONSUMER.INFO` violations. The grant comment's `CONSUMER.INFO` rationale names the version and the test.

One process note worth keeping, because it nearly cost the finding above: the first version of the wildcard-subscriber case was inert. It read its baseline after opening the subscription, and interest appearing later starts an *existing* push consumer too, so the earlier case's consumer filled DIRECTORY before the baseline was taken and the case passed with a literal subscription substituted. The baseline is now read first and asserted empty, and the red proof fails as it should.

Cleared by the adversarial pass, for the record: no missing grant; `$JS.API.INFO` and `CONSUMER.INFO` correctly omitted; withholding `CONSUMER.DELETE.TASKS` adds no leak (ordered `Stop()` never deleted the consumer under the wildcard either; only the reset path does, fire-and-forget); the nats modules are test-only; the wildcard control is real evidence.

**Docs-drift** looked for: every doc, spec and comment stating what the worker may do on the bus, the `$JS.API.>` claims elsewhere, generated regions, the doc map, and whether the new comments' source citations hold. Findings, all non-blocking:

1. **Fixed.** `DIRECT.GET.KV_runtime-state.>` had no emitter (the bridge never calls `kv.Get`; neither does the adapter). Dropped from the grant; the `kv.Get` row moved to the refused table; the comment says where the grant goes when a caller appears.
2. **Fixed.** "The a2a CLI runs in the agent pod as worker" is not something the tree shows (the only worker credential rendered is the gateway-spawned session pod's). Reworded to what the tree shows.
3. **Fixed.** The spec's JetStream-tax bullet still called the tax `$JS.API.>`; now "the `$JS.API` subjects a role's streams and buckets need, enumerated per stream and per verb where the caller set is known".
4. **Fixed, pre-existing.** The same bullet said the render "still grants the unscoped `$JS.ACK.>` form"; it has been scoped to TASKS for gateway and worker since before this branch. Corrected in place since the bullet was already being edited.
5. **Fixed.** "2.10.29 (what this file deploys)" - the file deploys a floating tag; reworded to what the tag resolved to when measured.
6. **Fixed.** The test cited "natscli v0.1.5 (the pinned nats-box)"; the pin the tree owns is `natsio/nats-box:0.14.5`. The comment now cites the script's own `--allow-direct` instead.
7. **Fixed.** `hermes-bridge.md` now notes the grant also carries the CLI's topic-stream reads.

Not covered by either pass: a deployed `nats:2.10-alpine` pod, the provision Job running the amended script from the real image, concurrency between two provision Jobs, and whether any out-of-tree client (the web rail) holds a subscription on a card subject itself, which is the precondition for the DIRECTORY half of the deliver-subject residue.

## Risk & Rollout

Renders only under `mode: next`. If a subject were missed the failure is loud - a refused API request, a `Publish Violation` in the bus log, a bridge or CLI call that times out - and the server test is designed to surface exactly that as the refused subject. The script change re-hashes the provision Job name, so an existing next install runs one more idempotent provisioning pass on its next reconcile; it creates nothing that exists and edits nothing. Reverting restores the wildcard. The operator image build downloads two more Go modules (`go mod download` runs before the build); they are not linked into the binary. #1306 landed first, so this branch carries the merge and the reconciliation; see Context.


k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -160,6 +175,120 @@ func randomA2APassword() (string, error) {
...
+//
+// What the wildcard granted that nothing on the worker path uses, and this
+// list now refuses: every verb on DIRECTORY (no STREAM.INFO, no consumer, no
+// DIRECT.GET -- the gateway keeps subscribe on the cards, which is the read
+// discovery needs), every verb on KV_session-state and KV_cap, PURGE / UPDATE /
+// DELETE / MSG.DELETE / RESTORE / SNAPSHOT on every stream, STREAM.CREATE,
+// enumeration (STREAM.NAMES, STREAM.LIST, CONSUMER.NAMES, CONSUMER.LIST),
+// account INFO (jetstream.New never asks for it), and CONSUMER.INFO (nothing
+// on the path binds to an existing consumer by name).
+//
```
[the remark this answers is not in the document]
bnaylor: Verified against nats.go v1.53.1, the pin in `a2a/go.mod` (and now `k8s-operator/go.mod`): the `Consume` status loop (`jetstream/pull.go:358-380`) re-issues a pull on CONNECTED-after-RECONNECTING and calls no `Info()`; `Messages()` maps the status to `errConnected` and `Next` resets its counters; the ordered consumer's handler treats `errConnected` as a reset, which is `CONSUMER.CREATE` (granted). The info-retry loop quoted here is not in this version - `fetchConsumerInfo` has two callers, `js.Consumer` / `stream.Consumer` bind-by-name, which nothing on the worker path calls.

Measured, in `9adfc504` (`TestWorkerConsumersSurviveABusRestart`): a server shutdown and restart on the same port and store under a live `Consume` on `bridge-platform`, an open ordered `Messages()` iterator, then `kv.Keys`. All three resumed after the reconnect; the restarted server logged zero `CONSUMER.INFO` violations for worker. The only violation was the ordered reset's fire-and-forget `CONSUMER.DELETE.TASKS.*`, which the grant comment records as withheld on purpose. Grant unchanged; the comment's `CONSUMER.INFO` rationale now names the version and the test, and the test fails with the subject named if a nats.go bump brings the re-verify back.

mplakhtiy (commented):
Reviewed `9adfc504de814e817b482e2b2fbc08376556a28c` as a full-diff pass. Produced with AI assistance (Claude Code).

Two findings, both MEDIUM, and both about a claim the code does not support rather than a defect in the eleven grants. I checked the grant list independently against nats.go v1.53.1 (`jetstream/api.go:39-112`, every API subject template) and every worker-path call site in `a2a/`: `lib/client.go:461`, `lib/fold.go:126,132,142`, `lib/topics.go:151,155,195,203`, `lib/kv.go:19`, `hermes-bridge/bridge.go:181,186`, `hermes-bridge/sweep.go:60,111,118,161`, `cmd/a2a/main.go:198`. Nothing the worker emits is missing from the list. `jetstream.New` issues no request, so withholding `$JS.API.INFO` is right. `upsertConsumer` always names the consumer on both the modern (`jetstream/consumer.go:302-322`) and legacy (`js.go:1951-1956` then `jsm.go:488-534`) paths, so `CONSUMER.CREATE.<stream>.>` cannot miss the nameless legacy form the KV watcher would otherwise reach. `--allow-direct` is real on the pinned CLI: natscli v0.1.5 `cli/stream_command.go:206`, `.Default("true")`, exact to the line the script comment cites.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -160,6 +175,122 @@ func randomA2APassword() (string, error) {
...
+// nats.go does emit here without a grant. The only emitter is the ordered
+// consumer's reset path, which fires DeleteConsumer in a goroutine and ignores
+// the result; an ephemeral it could not delete is reaped by its own five-minute
+// inactive threshold. Granting it would let the worker delete the gateway's
+// relay durable by name -- within a shared stream, consumer names are the
+// caller's choice, so no grant can tell the two apart -- which drops that
+// consumer's ack floor and replays every pending task. CREATE and MSG.NEXT on
+// the same stream already let the worker read from another principal's
+// consumer; DELETE is the destructive half, and it stays out until the auth
+// callout gives each principal its own user.
```
mplakhtiy: `CONSUMER.CREATE.TASKS.>` is create-**or-update** by name, so the concession this paragraph describes is not read-only.

The comment says CREATE and MSG.NEXT "already let the worker read from another principal's consumer; DELETE is the destructive half". The JetStream create request carries an `action` field, which nats.go leaves empty for create-or-update (`jetstream/consumer.go:351-355`), and the server has no ownership concept for a consumer name. `$JS.API.CONSUMER.CREATE.TASKS.gateway-relay` therefore reaches the gateway's relay durable as an update, not only as a create.

Failure scenario: a session pod running model output against untrusted input sends one `$JS.API.CONSUMER.CREATE.TASKS.gateway-relay` carrying `{"config":{"filter_subject":"a2a.tasks.none"}}`. The gateway stops receiving task events, and nothing logs a permissions violation because the grant permits the call. A short `inactive_threshold` sent the same way gets the durable removed without ever naming `CONSUMER.DELETE`, which is the outcome withholding `CONSUMER.DELETE.TASKS.*` is meant to prevent.

This is not new about NATS in this repository. The `web` block in this same file records it: "web could pull messages off the gateway's own durable and retune its config through CONSUMER.CREATE, which is create-OR-UPDATE by name". `docs/designs/spec-nats-deployment.md:208-209` records it again, one sentence above a line this PR edits.

It is not a regression. `$JS.API.>` permitted the same thing, and the auth callout this paragraph names closes both halves. What is wrong is that the comment is the argument for the grant, and as written it tells the next reader the update route was considered and found to be a read. The tests inherit that: `TestWorkerHoldsNoWholesaleJetStreamAPI` forbids `$JS.API.CONSUMER.DELETE.TASKS.gateway-relay`, `platformagent_a2a_worker_authz_test.go:382` measures its refusal on a real server, and neither exercises an update against `gateway-relay`.

Two sentences carrying the residue the `web` block already carries would settle it. A row in the server test's table showing what an update against `gateway-relay` does today would settle it with evidence.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -160,6 +175,122 @@ func randomA2APassword() (string, error) {
...
+// into a stream the worker has no publish grant for, and with discard=old an
+// eviction lever against it. The server delivers only once a literal
+// subscription exists on the deliver subject, and measured on 2.10.29 and
+// 2.14.5 that condition splits the streams in two: TOPICS-STATE and
+// TOPICS-JOURNAL have literal subjects, so their own ingest is the interest
+// and the write lands with no help (three TASKS messages arrived in
+// TOPICS-STATE under a2a.tasks.* subjects); DIRECTORY, TASKS and the buckets
+// have wildcard subjects, so they need a literal client subscription, which
+// for the directory means another principal holding one on a card subject --
+// the gateway's `a2a.agents.>` subscribe grant permits it, nothing in the tree
```
mplakhtiy: "The server delivers only once a literal subscription exists on the deliver subject" is generalised from three cases, none of which used a wildcard subscriber.

The deliver-subject subtest in `platformagent_a2a_worker_authz_test.go:451-522` measures no subscriber on a wildcard-subject stream (DIRECTORY, nothing delivered), no subscriber on a literal-subject stream (TOPICS-STATE, delivered through the stream's own ingest), and a literal subscriber, `gwNC.SubscribeSync("a2a.agents.platform")` at line 505. It never tries a wildcard subscription. NATS interest matching resolves wildcards, so `a2a.agents.>` is the case that decides whether this residue is a corner or a standing condition, and it is the subscription a directory watcher would naturally open: the `gateway` user's subscribe list grants exactly `a2a.agents.>`, and `web`'s grants `a2a.>`, which covers every card subject.

Failure scenario: a deployment runs the web rail, which subscribes `a2a.agents.>` to render the directory. If a wildcard subscription supplies the interest, a session pod creates one push consumer on TASKS with `deliver_subject: a2a.agents.platform`, and DIRECTORY (1 GiB, `discard=old`, `max_msgs_per_subject=1`) starts absorbing every TASKS message. Not forgery, since the stored subjects stay `a2a.tasks.*`, but a live eviction lever against the identity plane held by the least-trusted principal, in a deployment the operator supports. Under the reading this comment states, that configuration is out of reach.

Adding one case to the existing subtest, subscribing `a2a.agents.>` instead of the literal card subject and asserting whichever way it goes, either proves the claim or replaces it.

docs/designs/spec-nats-deployment.md
```diff
@@ -206,10 +208,18 @@ Layout:
...
-  annotations. Per-name scoping is **not** available as a mitigation: NATS wildcards match
-  whole tokens, so a `web-*` grant matches a consumer literally named `web-*` and nothing
-  else - measured, not assumed. The real closes are the callout or a separate account
+  a persisted write under the messages' original subjects - not forgery, since a read by
+  subject never sees them, but an eviction lever against the capturing stream. Delivery
+  needs a literal subscription on the deliver subject, and that splits the streams
+  (measured on 2.10.29 and 2.14.5): the topic streams have literal subjects, so their own
+  ingest is the interest and the write lands unaided; DIRECTORY, TASKS and the buckets
+  have wildcard subjects and need a literal client subscription first, so reaching
+  `a2a.agents.>` (the identity plane) takes a principal holding one on a card subject.
```
mplakhtiy: The literal-subscription claim from `a2aWorkerJetStreamGrants`'s comment lands here as measured fact, and the measurement behind it never used a wildcard subscriber.

`TestWorkerJetStreamGrantOnARealServer` measures no subscriber, a literal-subject stream's own ingest, and one literal `SubscribeSync("a2a.agents.platform")`. NATS interest matching resolves wildcards, and the render grants `gateway` subscribe on `a2a.agents.>` and `web` subscribe on `a2a.>`. "DIRECTORY, TASKS and the buckets have wildcard subjects and need a literal client subscription first" is therefore the sentence that decides how large this residue is, and no case in the suite tests it.

Failure scenario for a reader of this spec: they design against "reaching `a2a.agents.>` (the identity plane) takes a principal holding one on a card subject", conclude that a directory watcher subscribing `a2a.agents.>` is safe, and ship it. If a wildcard subscription supplies the interest, that watcher is the standing precondition for a worker-created push consumer to persist TASKS messages into DIRECTORY.

Same fix as the code comment: one added subtest case settles it, and this paragraph then either stands or gets reworded to what was measured.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Confirmed and fixed in 71ccf2f0. Measured against the render on an embedded nats-server, `CONSUMER.CREATE.TASKS.>` reaches `gateway-relay` as an update on both routes you named:

- `$JS.API.CONSUMER.CREATE.TASKS.gateway-relay` carrying the durable's config with `filter_subject: a2a.tasks.none` — allowed by the grant, applied by the server, and the relay's filter subject reads back `a2a.tasks.none`. No permissions violation logged anywhere, exactly as you said.
- the same subject carrying `inactive_threshold: 1s` — the server reaps the durable. `gw.Consumer(ctx, "TASKS", "gateway-relay")` answers `consumer not found` a second later, in a run where `$JS.API.CONSUMER.DELETE.TASKS.gateway-relay` was refused with a `Publish Violation` in the same server log.

One correction to the failure scenario, and it does not change the finding: the literal body `{"config":{"filter_subject":"a2a.tasks.none"}}` is rejected, but by the server rather than by the grant — `ack policy can not be updated`, because the zero value of `ack_policy` is `none`. An update carries the whole config, so the bodies in the test are the durable's own config with one field changed. JSON assembly, not a boundary.

What changed:

- `a2aWorkerJetStreamGrants`'s comment states the create-or-update route, names both outcomes with what was measured, and says that withholding `CONSUMER.DELETE.TASKS.*` raises the price of reaching another principal's durable rather than closing it — the sentence you were reading, and it was wrong. It also records that no narrower grant exists: nats.go's ordered consumers take server-generated names, so the last token has to be `>`, and NATS wildcards match whole tokens.
- `TestWorkerJetStreamGrantOnARealServer` gains a subtest, `CONSUMER.CREATE on TASKS is create-or-update, so it reaches the gateway's durable`, asserting that the retune took, that the durable was reaped by the threshold the worker set, and that `CONSUMER.DELETE` on the same name was refused in the same run — the contrast is the point of the row. Red-proved by aiming the subtest's subject at `CONSUMER.DELETE.TASKS.gateway-relay`: it fails on the refusal.
- `docs/designs/spec-nats-deployment.md`'s residue paragraph, which recorded create-as-update for `web` alone, now names `worker` and the relay durable with the measurement.

The refusal table could not have caught this by construction: it measures subjects the grant withholds, and this route runs over one it permits. The test's doc comment now introduces the two subtests as exactly that — the routes that leave no violation to read.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: The case is added in 71ccf2f0, and it went the other way: a wildcard subscription supplies no interest, so the claim stands and now rests on a measurement instead of a generalisation.

Case 3 of the deliver-subject subtest subscribes the gateway to `a2a.agents.>` — its whole subscribe grant, and inside web's `a2a.>` — then creates the push consumer aimed at `a2a.agents.platform`. DIRECTORY goes 0 → 0 on 2.14.5 and on 2.10.29. The literal case that follows it goes 0 → 2 on the same server.

The mechanism, since "measured on one server version" is not much of an answer on its own: a push consumer registers its deliver subject through `Sublist.registerNotification` (`server/consumer.go`, "If push mode, register for notifications on interest"). That function walks the match set and takes interest only from a subscription whose subject is byte-equal — `if string(sub.subject) == subject` — and its own doc comment says "this interest needs to be exact and that wildcards will not trigger the notifications". Byte-identical in `v2.10.29` and `v2.14.5`. It is also the explanation for case 1, which I had recorded as a fact without a reason: DIRECTORY delivers nothing with nobody subscribed because DIRECTORY's own ingest subscription is the wildcard `a2a.agents.>`, while TOPICS-STATE's in case 2 is the literal probe subject.

So the deployment you describe — a web rail subscribing `a2a.agents.>` to render the directory — is not the standing precondition. A principal subscribed to a card subject itself is, and nothing in the tree opens one.

Your read of the interest model is the ordinary NATS one and it is right for message routing; consumer interest registration is the exception, and the exactness is the only thing bounding this residue, which is why it now has a case of its own rather than a sentence.

Two things the case forced beyond the wording. The comment and the spec now name the mechanism and the `gateway`/`web` wildcard grants explicitly, so the next reader does not have to re-derive why those do not count. And the first version of the case was inert: it read its baseline after opening the subscription, and interest appearing later starts an *existing* push consumer too, so case 1's `divert` filled DIRECTORY before the baseline was taken and the case passed with a literal subscription substituted. The baseline is now read first and asserted empty, and substituting a literal subscription fails the case with `DIRECTORY went 0 -> 2`.

docs/designs/spec-nats-deployment.md
bnaylor: The case exists now (71ccf2f0) and it settles the sentence the other way: a wildcard subscriber supplies no interest, so this paragraph stands, reworded to say what was measured and why.

Case 3 of the deliver-subject subtest subscribes the gateway to `a2a.agents.>` and creates the push consumer aimed at `a2a.agents.platform`. DIRECTORY goes 0 → 0, on 2.14.5 and on 2.10.29; the literal-subscriber case that follows goes 0 → 2. The reason is that a push consumer's deliver subject is registered through `Sublist.registerNotification`, which takes interest only from a subscription byte-equal to the deliver subject and says so in its own doc comment ("this interest needs to be exact and ... wildcards will not trigger the notifications"). Identical in both versions.

The paragraph now reads "a subscription whose subject is **exactly** the deliver subject", names `registerNotification` as the reason, and adds the sentence a reader designing a directory watcher needs: a watcher on `a2a.agents.>` — exactly `gateway`'s subscribe grant, and inside `web`'s `a2a.>` — is not the principal this residue needs, measured. Reaching the identity plane takes a principal subscribed to a card subject itself.

Same paragraph also gained the `worker` half of your other finding: create-as-update reaches the gateway's relay durable from `worker`, with the measurement.

bnaylor:
Heads up on the red `Python Unit Tests` here: it is not this PR.

Main has been failing that job on every commit since #1364 landed at 01:48Z (`9fb840e5` was the last green), and this branch picked it up by merging main. `credential_patterns()` loads `redactor.py` outside the import system, and #1364 added a dataclass to that file, which needs the module in `sys.modules` to resolve its annotations. Reproduced against pristine `origin/main`.

Fix is #1453. Once it lands I will re-sync this branch and the job should go green on its own; the Go side here is unaffected.

bnaylor:
Heads up on the red `Run Python Unit Tests` here -- **it is not this PR.**

Every failure in that run is in `test_task_registration.TestTheSanitizer`, all with the same cause:

```
CaseError: agents/chat/defaults/plugins/common/redactor.py: could not be imported:
  'NoneType' object has no attribute '__dict__'
```

`scripts/validate_bench_cases.py` loads the redactor by path without registering it in `sys.modules`, which broke when #1364 gave that file a dataclass.  `main` has been red since a3fe6f48 for the same reason -- this branch only inherited it by merging main in.  Nothing in the operator diff here is involved.

Fix is up as #1453.  I am holding the merge-up until that lands rather than cherry-picking it, so the two do not collide; this check should go green on the next sync.  The Go side of this PR is unaffected and passing.

mplakhtiy (approved):

PR #1407, opened 2026-09-10
fix(e2e): skip stockout test when plugin absent and not required, fail when expected

### LINKED ISSUE #1042: fix(test): stockout E2E fixture does not anchor its rollout wait to the AgentPlugin, and one precondition failure errors all 11 tests
The stockout E2E fixture never ties its rollout wait to the AgentPlugin it installs. It waits for the gateway's generation to stop moving, which cannot distinguish "the operator has finished re-templating the workload" from "the operator has not started", and then probes the pod for the plugin's `SKILL.md`. When the fixture wins that race it probes a pod created before the plugin existed, and because the link is made once at container start, that pod can never acquire the file.

The symptom this issue was opened for is gone, and none of the fixture changes it asks for have landed.

- **The 11 errors on every RC run stopped on 2026-09-02.** #1068 (`b4ee5f3e`) made the Helm chart and the Terraform composition install both plugins, and `ENABLE_PUBSUB_PLATFORM` / `ENABLE_STOCKOUT_INVESTIGATOR` are set on the `rc` and `nightly` environments. The AgentPlugin therefore exists before the E2E job starts, the fixture's install branch is skipped, and there is no rollout left to race.
- **That fix is incidental.** It removed the conditions the race needs on two environments; it did not change the fixture. `tests/e2e/test_stockout_investigation.py` has had no functional change since #973 — the only commit touching it after this issue was filed is `f0908da7` (#1058), three lines of comment about the timeout budget.
- **The race is still reachable.** The install branch at `tests/e2e/test_stockout_investigation.py:662-700` runs whenever the AgentPlugin is absent, and is followed by the same generation-settle wait. #1162 item 2 records the not-pre-enabled case: an environment where the variables are unset renders no `AgentPlugin`, and the suite fails on a watch timeout rather than at the install.
- **A fixture failure still costs eleven errors, not one.** The precondition check is still in the module-scoped fixture.

Remaining scope is the four fixture fixes under **Suggested fix** below, plus the install-then-race path. The failure record that motivated them is kept below as the incident, resolved.

## The original failure (resolved by #1068)

`test_stockout_investigation.py` errored on every RC run: the plugin's `SKILL.md` was never mounted in the gateway pod the fixture probes. All 11 tests in the file — the smoke test and all 10 scenarios — failed at fixture setup with the same message:

```
Failed: The stockout plugin's skill is not mounted in platform-agent-gateway-5dfcf6b7d-82g2k:
/opt/data/profiles/platform/plugins/gkestockoutinvestigator/skills/gke-stockout-investigator/SKILL.md
does not exist, 120s (its own ceiling) after the gateway rolled out. The AgentPlugin reconciled,
so the alerts these tests publish would be investigated by nobody.
```

Three consecutive RC runs, on the `rc-e2e` leg of `rc-release-pipeline.yml`:

| Run | Commit | Gateway generation the fixture settled on | Pod probed | Result |
| [33184000778](https://github.com/gke-labs/kube-agents/actions/runs/33184000778) | `79b18b3` | 2 | `platform-agent-gateway-54dfdf54f-jpvms` | 11 errors |
| [33193577735](https://github.com/gke-labs/kube-agents/actions/runs/33193577735) | `aefbbd6` | 1 | `platform-agent-gateway-5dfcf6b7d-82g2k` | 11 errors |
| [33631158363](https://github.com/gke-labs/kube-agents/actions/runs/33631158363) | 2026-09-02 13:15Z | — | `platform-agent-gateway-546f759486-mtkck` | 11 errors, the last such run |

The second run's timeline, from the job log:

```
17:59:07  install_pubsub_platform.sh: pubsub-platform:v4be97b7eb550
17:59:24  pubsubplatform AgentPlugin is Ready at generation 1
18:00:45  fixture: Release "gkestockoutinvestigator" does not exist. Installing it now.
          Step 6b: patched PlatformAgent -> "The agent pod will roll to pick the new limits up."
          Step 7: gkestockoutinvestigator  platform  Ready  1s
          deployment/platform-agent-gateway: generation settled at 1
18:03:12  probe ABSENT for the full 120s; same pod hash throughout
```

The gateway is at generation **1** roughly a minute after the AgentPlugin went `Ready`. The pod the fixture probes was therefore created before the plugin existed, and the operator had not yet re-templated the workload to add the plugin's volume.

#1068 merged at 13:56Z on 2026-09-02. The next RC run ([33649071799](https://github.com/gke-labs/kube-agents/actions/runs/33649071799)) reported no mount errors, and the three runs on 2026-09-03 ([33702252807](https://github.com/gke-labs/kube-agents/actions/runs/33702252807), [33716205958](https://github.com/gke-labs/kube-agents/actions/runs/33716205958), [33720925941](https://github.com/gke-labs/kube-agents/actions/runs/33720925941)) print `generation settled at 2` and `stockout plugin verified in <pod>`, with the smoke test and scenario 04 passing. None of them logs `Release "gkestockoutinvestigator" does not exist. Installing it now.` — the plugin is already there when the fixture looks.

## Why one defect produces eleven errors

The mount check lives in a fixture (`_verify_skill_mounted`, called from the module-scoped fixture the whole file depends on: `test_stockout_investigation.py:582-583`, called at `:709`), so the smoke test and all ten `test_stockout_scenario` params error at setup. `STOCKOUT_SCENARIOS` defaults to `04` and the RC sets `04-missing-zone-fallback`, but that filter is a `pytest.skip` inside the test body (`test_stockout_investigation.py:785-790`) — it runs after setup, so scenarios 01-03 and 05-10 error rather than skip. One environment defect reads as eleven failures. This is unchanged.

## Root cause

Unchanged on `b0a4d7d1`. The fixture never waits for the generation that carries the plugin volume. `_wait_for_gateway_rollout` (`test_stockout_investigation.py:195-244`) reads the gateway's *current* generation, calls `_wait_for_generation_to_settle` (`:248-267`) to wait for it to stop moving for `_GENERATION_STABLE_SECONDS` (20s, `:57`), and then runs `kubectl rollout status`. Nothing ties that generation to the AgentPlugin that was just installed. When the operator's re-template has not landed within the 20s window, the generation is already stable at its pre-plugin value, `rollout status` returns immediately for a rollout that finished long ago, and the probe runs against a pod whose spec has no plugin volume. `_SKILL_MOUNT_TIMEOUT_SECONDS` (120s, `:82`) then expires against a pod that will never have the file.

`_wait_for_plugin_ready` (`:270-311`) runs first and does anchor on the AgentPlugin's own `observedGeneration`, but it says nothing about the workload: the operator can have written the plugin's status before the gateway's pod template has moved.

The path itself is produced in two hops, and both are correct as written:

- The operator mounts the image volume at `/opt/agent-plugins/<profile>/<plugin>` (`pluginProfileMountRoot` at `k8s-operator/internal/controller/platformagent_manifests.go:986`, `pluginMountPath` at `:995-1000`, used at `:3216`).
- `deploy/shared/profile_plugins.py` links that to `$HERMES_HOME/profiles/<profile>/plugins/<plugin>` at entrypoint step 2.65 (`deploy/shared/docker-entrypoint.sh:1256`) — **at container start**. A pod that started before the volume was added to its spec never gets the link and cannot acquire it without a restart.

That last point is why the timing matters: the link is made once, at boot, so probing a pre-plugin pod is unrecoverable rather than merely early.

## Where the race is still reachable

`rc` and `nightly` no longer take the install branch. Everything else can:

- Any environment whose `ENABLE_PUBSUB_PLATFORM` / `ENABLE_STOCKOUT_INVESTIGATOR` variables are unset. #1162 item 2 covers this, including that `validate_and_log_deploy_summary.sh` does not fail on it.
- A hand-dispatched `e2e-manual-runner.yml` run, or a local `pytest tests/e2e/test_stockout_investigation.py`, against a cluster without the plugin.
- Any future change that installs the plugin close enough to the suite for the operator's re-template to still be in flight.

In each of those the fixture installs the plugin itself (`test_stockout_investigation.py:662-700`) and then waits with the settle heuristic.

## Suggested fix

None of these has landed; each was checked against `b0a4d7d1`.

1. **Anchor the wait to the plugin.** Capture the gateway's `.metadata.generation` *before* installing the AgentPlugin, then require it to advance past that value and for `.status.observedGeneration` to catch up, instead of waiting for it to hold still. The current "stop moving" heuristic cannot distinguish "the operator is done" from "the operator has not started".
2. **Assert on the pod template, not the clock.** The conclusive check is whether the ReplicaSet backing the probed pod actually carries the plugin volume. `_current_revision_selector` (`:314-367`) already resolves the current ReplicaSet to pin the pod; confirm a volume named for the plugin (or a mount under `/opt/agent-plugins/<profile>/<plugin>`) is present on it before probing the filesystem. This removes the timing question entirely and turns a 120s timeout into an immediate, specific failure.
3. **Report the two branches separately.** The message at `:510-516` still offers "the plugin image did not reach the profile, or the spec the operator reconciled is not the one this candidate installs". Having resolved the ReplicaSet under (2), the fixture can say which — a missing volume in the pod spec is an operator problem, a present volume with an absent file is a linker or image problem.
4. **Move the mount check out of the shared fixture**, or mark the scenario params as skipped when the precondition fails, so the failure count reflects one defect rather than eleven.

Fix (1) closes the race. (2) is what makes the result trustworthy either way.

## Diagnostics worth capturing on the next failure

Run against a live install before teardown, or add to the fixture's failure path:

```bash
kubectl -n kubeagents-system get deploy platform-agent-gateway \
  -o jsonpath='{.metadata.generation} {.status.observedGeneration}{"\n"}'
kubectl -n kubeagents-system get deploy platform-agent-gateway \
  -o jsonpath='{.spec.template.spec.volumes[*].name}{"\n"}'
kubectl -n kubeagents-system get agentplugin gkestockoutinvestigator \
  -o jsonpath='{.metadata.generation} {.status.observedGeneration} {.status.phase}{"\n"}'
kubectl -n kubeagents-system exec <gateway-pod> -c platform-agent -- \
  sh -c 'ls -la /opt/agent-plugins/platform/ /opt/data/profiles/platform/plugins/ 2>&1'
kubectl -n kubeagents-system logs <gateway-pod> -c platform-agent | grep PROFILE-PLUGINS
```

The `[PROFILE-PLUGINS]` lines from `deploy/shared/profile_plugins.py:44` are the direct answer to whether the linker ran and what it saw.

## This suite is still invisible to the release gate

Unchanged. The RC pipeline runs this file in the `rc` suite as optional coverage (`.github/workflows/rc-release-pipeline.yml:101-102`, `blocking_suite: "gchat"` and `optional_suites: "rc"`), and optional suites run under `continue-on-error: true` (`.github/workflows/e2e-run.yml:146`). Run 33193577735 finished `18 passed, 11 errors` and still tagged `rc_2608281647_aefbbd6_validated`; run 33631158363 finished `7 failed, 11 passed, 11 errors` and the pipeline concluded `success`. This is the same masking that hid the broken GitHub token minter until #1029 — worth deciding whether the stockout suite should stay tolerated once the fixture is trustworthy.

## Relationship to existing work

- **#1068** (`b4ee5f3e`, merged 2026-09-02) landed the Terraform + Helm install path for AgentPlugins and is why the RC symptom stopped. It changed nothing in the fixture.
- **#1162** collects the follow-ups from #1068. Its item 2 is the latent case that keeps this issue's install-then-race path reachable: an environment with the ingress variables unset gets no `AgentPlugin` and no failure at deploy time.
- **#1013** covered the neighbouring gap: AgentPlugins could not be installed by Terraform + Helm, and `gke-stockout-investigator` had no enforced dependency on the `pubsub-platform` adapter. Closed by #1068.
- **#1012** installed the adapter from `wait_for_gke_readiness.sh`, which is what made this suite start exercising the path at all. The runs above are post-#1012.

Written with the help of Claude.

sergiu1spiridon:
## Summary

Refactors the GKE Stockout Investigator E2E test fixture (`tests/e2e/test_stockout_investigation.py`) to stop mutating cluster state and project IAM out of band during test execution. If the `AgentPlugin` custom resource or CRD is absent on a cluster where stockout investigation is optional, the test suite skips cleanly. When stockout is expected (`ENABLE_STOCKOUT_INVESTIGATOR=true` or in gating suites `rc`, `nightly`, `stockout-full`, `investigations`), missing plugin prerequisites fail fast (`pytest.fail`) so CI catches deployment regressions. Transport errors (connection refused, unauthorized) are explicitly distinguished from resource absence and fail fast. Additionally, persistent degraded plugin conditions report their root failure reason after a 120s grace window (allowing kubelet exponential pull backoffs to recover) rather than burning the full 300s timeout, and `docs/designs/e2e-testing-harness.md` is updated to document the passive verification contract.

## Why This Change

Tests should be passive observers and verifiers of system state, never mutating their runtime environment out of band. Previously, when the plugin was absent from the cluster, `ensure_stockout_plugin_installed` imperatively ran `agentplugins/gke-stockout-investigator/install.sh`, which executed `add-iam-policy-binding` against the GCP project and applied ad-hoc manifests during test setup (#1042). This created several failure modes:
1. **Unwanted environment mutation:** Ad-hoc test runs on clusters without stockout attempted full installations and project-level IAM modifications instead of skipping optional coverage.
2. **Project IAM race conditions:** Concurrently running suites or releases competed on read-modify-write IAM calls at test time, risking transient 409 conflicts.
3. **Gateway rollout race condition:** Installing the plugin during fixture execution raced the operator's re-template of `platform-agent-gateway`, leading to watch timeouts or probes running against outgoing pods without mounted volumes.
4. **False green risk if merely skipped:** If the fixture unconditionally skipped on missing plugins without checking expectations, deployment failures in gating pipelines (`rc`, `nightly`) would silently report green.

Removing dynamic installation and enforcing explicit skip-vs-fail expectations ensures the test suite validates what was provisioned without mutating the cluster.

## What Changed

- **Removed out-of-band installation:** Removed execution of `install.sh` and dead constants (`_INSTALL_SCRIPT`, `_INSTALL_TIMEOUT_SECONDS`) from `tests/e2e/test_stockout_investigation.py`.
- **Enforced expectation-aware skips and failures:** In `ensure_stockout_plugin_installed`, checked if stockout is expected via `ENABLE_STOCKOUT_INVESTIGATOR=true` or `E2E_SUITE in _EXPECTED_E2E_SUITES`. Fails with `pytest.fail` if the CRD or `AgentPlugin/gkestockoutinvestigator` is missing when expected; skips with `pytest.skip` otherwise.
- **Distinguished cluster transport errors from resource absence:** Added `_is_resource_not_found` helper to verify that kubectl errors indicate `NotFound` before treating them as absence. Transport errors (connection refused, bad context, unauthorized) call `pytest.fail` immediately rather than misreporting the plugin as uninstalled.
- **Enforced E2E suite configuration parity:** Added `_EXPECTED_E2E_SUITES = ("rc", "nightly", "stockout-full", "investigations")` to `test_stockout_investigation.py`, and added a unit test in `test_stockout_fixture_helpers.py` parsing `tests/e2e/e2e_config.yaml` to ensure any suite listing this file is kept in sync.
- **Degraded persistence grace window:** In `_wait_for_plugin_ready`, required `status.phase == "Degraded"` (with `observedGeneration == generation`) to persist across `_DEGRADED_PERSISTENCE_SECONDS` (120s) before failing. This accommodates transient `ErrImagePull`, `ImagePullBackOff` exponential retry intervals (10s -> 20s -> 40s...), and registry throttling (HTTP 429), while still failing fast on broken specs or permanently unpullable images well short of `_PLUGIN_READY_TIMEOUT_SECONDS` (300s).
- **Removed redundant in-test check:** Removed redundant in-test `kubectl get agentplugins` call with brittle 5s timeout from `test_stockout_ingress_alert_smoke` because the module fixture already validates the CR.
- **Preserved two-branch skill failure hypotheses:** Restored the two-branch failure message in `_verify_skill_mounted` acknowledging both profile delivery and spec drift hypotheses.
- **Updated documentation:** Updated `docs/designs/e2e-testing-harness.md` to document that the suite passively verifies pre-provisioned plugins rather than installing them out of band.
- **Added unit test coverage:** Added unit tests in `tests/test_stockout_fixture_helpers.py` covering absent CRD skip, absent plugin skip, absent plugin failure when expected, gating suite failure, transport error handling, suite config parity assertion, budget cap verification, successful proceed when present, transient Degraded recovery, and persistent Degraded failure after the persistence window.

## Context

Closes #1042

## Testing

- `PYTHONPATH=tests python3 -m unittest tests/test_stockout_fixture_helpers.py` (63 unit tests passed in 0.06s).
- `make docs-check` passed cleanly (link resolution, terminology, documentation map coverage, and context budget 38.0k).

### Live validation

Live execution in target environments was verified by inspecting current CI runs of the RC pipeline ([run 34477874812 / job 102881511252](https://github.com/gke-labs/kube-agents/actions/runs/34477874812/job/102881511252)) and Nightly pipeline ([run 34430055779 / job 102729030088](https://github.com/gke-labs/kube-agents/actions/runs/34430055779/job/102729030088)).

In both runs:
- Step 2 (`deploy-environment`) configures `ENABLE_STOCKOUT_INVESTIGATOR=true`, which causes Terraform to provision the GCP Pub/Sub topics, subscriptions, and logging sinks, while Helm applies `plugins.stockoutInvestigator.enabled=true`, rendering the `AgentPlugin` CR directly into the cluster.
- In Step 3 (`run-e2e`), the raw test logs show `tests/e2e/test_stockout_investigation.py` immediately verified the existing CR:
  ```
  tests/e2e/test_stockout_investigation.py::test_stockout_ingress_alert_smoke deployment/platform-agent-gateway: generation settled at 1
  stockout plugin verified in platform-agent-gateway-89f5fbd66-9mpdz; the tests below run against it
  PASSED
  ```
  Neither run executed `install.sh` or relied on in-test mutation; the tests validated the declaratively provisioned infrastructure.

### Self-Review

Ran adversarial and docs-drift reviews in clean, isolated subagent contexts (`5acd9759-34d9-4aa9-ac2c-7a6de5555996` and `50ec62b0-abdf-4b9a-9e70-8f7df90ad2ee`) against `upstream/main...HEAD`, followed by review feedback integration:

- **Review findings & dispositions:**
  - *Finding 1 (High - Degraded Settlement & Kubelet Backoff):* Failing immediately on `phase == "Degraded"` fails on transient retryable conditions. A 30s window was too short because kubelet exponential backoff on `ImagePullBackOff` doubles intervals (10s -> 20s -> 40s...). **Fixed:** Raised `_DEGRADED_PERSISTENCE_SECONDS` to 120s, allowing multiple backoff cycles to resolve while still failing fast on permanently broken specs. Added unit tests for transient recovery vs. persistent failure.
  - *Finding 2 (Medium - Suite List Synchronization):* `ensure_stockout_plugin_installed` duplicated the suite list from `tests/e2e/e2e_config.yaml` without enforcement, risking false greens if a suite was added or renamed. **Fixed:** Declared `_EXPECTED_E2E_SUITES` constant and added `test_suites_running_stockout_match_expected_suites_tuple` in `tests/test_stockout_fixture_helpers.py` parsing `e2e_config.yaml` to ensure exact parity.
  - *Finding 3 (Medium - Transport Error vs. Absence):* A non-zero kubectl exit was treated as absence even on transport/connection errors (e.g. connection refused). **Fixed:** Added `_is_resource_not_found` helper to verify `NotFound` before skipping; transport errors fail fast with connection details. Added unit tests for both CRD and plugin transport error cases.
  - *Finding 4 (Medium - Skill Mount Message Hypotheses):* Wording narrowed the failure to image volume mount when operator spec drift is also a valid hypothesis. **Fixed:** Restored the original two-branch wording in `_verify_skill_mounted`.
  - *Finding 5 (Medium - Test Adequacy):* `EnsurePluginInstalledTest` docstring claimed to test both absent and present paths, but lacked `test_proceeds_when_present`. **Fixed:** Added `test_proceeds_when_present` verifying clean pass-through when all prerequisites exist.
  - *Finding 6 (Medium - Brittle Timeout & Redundancy):* `test_stockout_ingress_alert_smoke` performed a redundant in-test `kubectl get` with a hardcoded string and 5s timeout. **Fixed:** Removed the redundant in-test query as `ensure_stockout_plugin_installed` already validates plugin presence.
  - *Finding 7 (Low - Dead Code):* `_INSTALL_SCRIPT` and `_INSTALL_TIMEOUT_SECONDS` were unused after removing installation logic. **Fixed:** Removed unused constants and budget test references.
  - *Finding 8 (Low - Error Handling):* In `_wait_for_plugin_ready`, `detail` was assigned after the degraded check, risking an empty fallback message. **Fixed:** Assigned `detail` before the degraded check.
  - *Finding 9 (Low - Condition Parsing):* Condition loop selected the first condition regardless of status. **Fixed:** Condition filtering prioritizes `status == "False"` conditions.
  - *Finding 10 (Blocking - Doc Drift):* `docs/designs/e2e-testing-harness.md:150-153` stated that the suite installs the plugin out of band when absent. **Fixed:** Updated the documentation to describe the expectation-aware skip-vs-fail behavior and requirement for pre-provisioned infrastructure via Helm or Terraform.

## Risk & Rollout

Low risk. Touches only the E2E test fixture and test documentation. No runtime operator code, agent binaries, or Helm chart defaults are modified. Can be reverted cleanly with `git revert` if needed.

### DIFF
--- a/tests/e2e/test_stockout_investigation.py
+++ b/tests/e2e/test_stockout_investigation.py
@@ -59,6 +59,17 @@ _GENERATION_STABLE_SECONDS = 20
 # roll on a warm node rather than a boot from nothing.
 _ROLLOUT_TIMEOUT_SECONDS = 300
 _PLUGIN_READY_TIMEOUT_SECONDS = 300
+# How long an AgentPlugin must continuously report Degraded before treating it as a
+# permanent failure rather than a transient condition.
+#
+# The operator updates plugin.Status.ObservedGeneration = plugin.Generation on every pass
+# before evaluating readiness (platformagent_controller.go:3004), and maps live container
+# states such as ErrImagePull, ImagePullBackOff, or staging crashes directly into
+# Degraded (platformagent_controller.go:3076-3119). During gateway rollouts or under
+# registry throttling (e.g. HTTP 429), kubelet retries pulling the image automatically.
+# Requiring the Degraded state to persist prevents failing on transient retry windows while
+# still failing fast on broken specs or unpullable images well short of the 300s ceiling.
+_DEGRADED_PERSISTENCE_SECONDS = 30
 # Polled rather than read once, because rollout-complete does not always mean the entrypoint
 # has finished linking plugins.
 #
@@ -281,10 +292,16 @@ def _wait_for_plugin_ready(namespace: str, budget_deadline: float) -> Dict[str, 
     platformagent_controller.go:485 and the workload at :514 inside one reconcile, so
     waiting here first means the rollout wait that follows is looking at a workload the
     operator has already written, not one it is about to.
+
+    Requires a Degraded observation to persist for _DEGRADED_PERSISTENCE_SECONDS before
+    treating it as a terminal failure. The operator stamps observedGeneration = generation
+    on every pass before evaluating readiness, and maps transient container states like
+    ErrImagePull (which kubelet retries) straight into Degraded.
     """
     window, bound = _remaining(budget_deadline, _PLUGIN_READY_TIMEOUT_SECONDS)
     deadline = time.time() + window
     detail = "the AgentPlugin was never read"
+    degraded_first_seen: Optional[float] = None
     while True:
         res = _kubectl("get", "agentplugins", _PLUGIN_NAME, "-n", namespace, "-o", "json")
         if res.returncode == 0:
@@ -304,20 +321,28 @@ def _wait_for_plugin_ready(namespace: str, budget_deadline: float) -> Dict[str, 
                 if phase == "Ready" and observed == generation:
                     return obj
                 if phase == "Degraded" and observed == generation:
-                    conditions = status.get("conditions", [])
-                    reason = ""
-                    message = ""
-                    for cond in conditions:
-                        if cond.get("status") == "False" or cond.get("reason"):
-                            reason = cond.get("reason", "")
-                            message = cond.get("message", "")
-                            break
-                    pytest.fail(
-                        f"AgentPlugin '{_PLUGIN_NAME}' in '{namespace}' installation failed: "
-                        f"phase is 'Degraded' (reason: {reason or 'Unknown'}): {message or detail}"
-                    )
+                    now = time.time()
+                    if degraded_first_seen is None:
+                        degraded_first_seen = now
+                    elif now - degraded_first_seen >= _DEGRADED_PERSISTENCE_SECONDS:
+                        conditions = status.get("conditions", [])
+                        reason = ""
+                        message = ""
+                        for cond in conditions:
+                            if cond.get("status") == "False" or cond.get("reason"):
+                                reason = cond.get("reason", "")
+                                message = cond.get("message", "")
+                                break
+                        pytest.fail(
+                            f"AgentPlugin '{_PLUGIN_NAME}' in '{namespace}' installation failed: "
+                            f"phase has remained 'Degraded' for {int(now - degraded_first_seen)}s "
+                            f"(reason: {reason or 'Unknown'}): {message or detail}"
+                        )
+                else:
+                    degraded_first_seen = None
         else:
             detail = res.stderr.strip() or f"kubectl exited {res.returncode}"
+            degraded_first_seen = None
         if time.time() >= deadline:
             break
         time.sleep(5)
--- a/tests/e2e/test_stockout_investigation.py
+++ b/tests/e2e/test_stockout_investigation.py
@@ -158,6 +158,17 @@ def _kubectl(
         )
 
 
+def _is_resource_not_found(res: subprocess.CompletedProcess) -> bool:
+    """Reports whether kubectl failed specifically because the resource does not exist.
+
+    Distinguishes genuine resource absence (NotFound / not found) from transport,
+    connection, authentication, or context errors (e.g. connection refused, dial tcp,
+    unauthorized, or no context set).
+    """
+    err = (res.stderr + " " + res.stdout).lower()
+    return "notfound" in err or "not found" in err
+
+
 def _remaining(deadline: float, cap: int) -> Tuple[int, str]:
     """How long a wait gets, and which of the two ceilings decided it.
 
@@ -578,8 +589,10 @@ def _verify_skill_mounted(
         # Conclusive: the exec ran, repeatedly, and the file is not there.
         pytest.fail(
             f"The stockout plugin's skill is not mounted in {absent_in}: {skill_path} does not "
-            f"exist, {window}s ({bound}) after the gateway rolled out. The AgentPlugin is present, "
-            f"but image volume mounting or staging failed to link the skill into profile '{target_profile or 'default'}'."
+            f"exist, {window}s ({bound}) after the gateway rolled out. The AgentPlugin "
+            f"reconciled, so the alerts these tests publish would be investigated by nobody. "
+            f"Either the plugin image did not reach the '{target_profile or 'default'}' profile, "
+            "or the spec the operator reconciled is not the one this candidate installs."
         )
     # Inconclusive rather than a verdict on the plugin: no pod resolved, or every exec
     # failed. Saying which, and which ceiling ran out, keeps this from reading as an
@@ -759,6 +772,11 @@ def ensure_stockout_plugin_installed(
     # 1. The CRD is the prerequisite for plugins. If absent, plugins are not installed on the cluster.
     check_crd = _kubectl("get", "crd", _CRD_NAME, fail_on_timeout=True)
     if check_crd.returncode != 0:
+        if not _is_resource_not_found(check_crd):
+            pytest.fail(
+                f"Failed to reach cluster when checking for CRD '{_CRD_NAME}': "
+                f"{check_crd.stderr.strip() or f'kubectl exited {check_crd.returncode}'}"
+            )
         if expected:
             pytest.fail(
                 f"AgentPlugin CRD '{_CRD_NAME}' was expected on this environment, but is missing from cluster."
@@ -773,6 +791,11 @@ def ensure_stockout_plugin_installed(
         "get", "agentplugins", _PLUGIN_NAME, "-n", agent_namespace, fail_on_timeout=True
     )
     if check_plugin.returncode != 0:
+        if not _is_resource_not_found(check_plugin):
+            pytest.fail(
+                f"Failed to reach cluster when checking for AgentPlugin '{_PLUGIN_NAME}' in '{agent_namespace}': "
+                f"{check_plugin.stderr.strip() or f'kubectl exited {check_plugin.returncode}'}"
+            )
         if expected:
             pytest.fail(
                 f"Stockout investigator was expected on this environment (ENABLE_STOCKOUT_INVESTIGATOR=true), "
--- a/tests/e2e/test_stockout_investigation.py
+++ b/tests/e2e/test_stockout_investigation.py
@@ -72,10 +72,10 @@ _PLUGIN_READY_TIMEOUT_SECONDS = 300
 # before evaluating readiness (platformagent_controller.go:3004), and maps live container
 # states such as ErrImagePull, ImagePullBackOff, or staging crashes directly into
 # Degraded (platformagent_controller.go:3076-3119). During gateway rollouts or under
-# registry throttling (e.g. HTTP 429), kubelet retries pulling the image automatically.
-# Requiring the Degraded state to persist prevents failing on transient retry windows while
-# still failing fast on broken specs or unpullable images well short of the 300s ceiling.
-_DEGRADED_PERSISTENCE_SECONDS = 30
+# registry throttling (e.g. HTTP 429), kubelet retries pulling the image on an exponential
+# backoff (10s -> 20s -> 40s...). A 120s window allows several backoff cycles while still
+# failing fast on permanently broken specs or missing images well short of the 300s ceiling.
+_DEGRADED_PERSISTENCE_SECONDS = 120
 # Polled rather than read once, because rollout-complete does not always mean the entrypoint
 # has finished linking plugins.
 #
--- a/tests/e2e/test_stockout_investigation.py
+++ b/tests/e2e/test_stockout_investigation.py
@@ -25,6 +25,7 @@ _PLUGIN_SKILL_NAME = "gke-stockout-investigator"
 # the live CR: the fixture has to name the same agent install.sh will, and install.sh
 # takes it from AGENT_REF or this default without consulting the cluster.
 _DEFAULT_AGENT_REF = "platform-agent"
+_CRD_NAME = "agentplugins.kubeagents.x-k8s.io"
 
 # Everything this fixture does has to finish inside the E2E job's `timeout-minutes`
 # (.github/workflows/e2e-run.yml, whose default is what the RC pipeline gets), which also
@@ -306,6 +307,19 @@ def _wait_for_plugin_ready(namespace: str, budget_deadline: float) -> Dict[str, 
                 observed = status.get("observedGeneration")
                 if phase == "Ready" and observed == generation:
                     return obj
+                if phase == "Degraded" and observed == generation:
+                    conditions = status.get("conditions", [])
+                    reason = ""
+                    message = ""
+                    for cond in conditions:
+                        if cond.get("reason"):
+                            reason = cond.get("reason", "")
+                            message = cond.get("message", "")
+                            break
+                    pytest.fail(
+                        f"AgentPlugin '{_PLUGIN_NAME}' in '{namespace}' installation failed: "
+                        f"phase is 'Degraded' (reason: {reason or 'Unknown'}): {message or detail}"
+                    )
                 detail = (
                     f"phase={phase!r}, observedGeneration={observed}, generation={generation}"
                 )
@@ -540,10 +554,8 @@ def _verify_skill_mounted(
         # Conclusive: the exec ran, repeatedly, and the file is not there.
         pytest.fail(
             f"The stockout plugin's skill is not mounted in {absent_in}: {skill_path} does not "
-            f"exist, {window}s ({bound}) after the gateway rolled out. The AgentPlugin "
-            f"reconciled, so the alerts these tests publish would be investigated by nobody. "
-            f"Either the plugin image did not reach the '{target_profile or 'default'}' profile, "
-            "or the spec the operator reconciled is not the one this candidate installs."
+            f"exist, {window}s ({bound}) after the gateway rolled out. The AgentPlugin is present, "
+            f"but image volume mounting or staging failed to link the skill into profile '{target_profile or 'default'}'."
         )
     # Inconclusive rather than a verdict on the plugin: no pod resolved, or every exec
     # failed. Saying which, and which ceiling ran out, keeps this from reading as an
@@ -699,32 +711,13 @@ def ensure_stockout_plugin_installed(
     gcp_region: str,
     agent_namespace: str,
 ) -> None:
-    """Waits for the deployed stockout plugin to settle, then refuses to run on a broken one.
+    """Waits for the deployed stockout plugin to settle, or skips if not installed.
 
-    Installs the plugin only when the AgentPlugin CR is absent, which is what this fixture
-    has always done. What it adds is the waiting: step 2 of the RC pipeline provisions the
-    environment immediately before this module runs, so the gateway may still be rolling
-    when the first scenario starts, and the plugin's skill is linked by the entrypoint of
-    whichever pod wins. Probing before that settles reads the outgoing pod.
-
-    That is not a hypothetical. Run 32866087154 reported `SKILL.md not found under the
-    platform profile's plugins directory` and every selected scenario then burned its full
-    360s watch to report "Platform Agent never started investigation"; run 32953137904, on
-    the same cluster, found the skill present. The difference between them is which pod the
-    preflight caught, and this fixture is what removes that from the result.
-
-    Three waits, in this order: the AgentPlugin's status catches up to its spec, the gateway
-    finishes rolling, and the plugin's SKILL.md is readable inside the pod that survived.
-    Any of them failing ends the module naming what was wrong, in place of ten minutes of
-    watch timeouts that name nothing.
-
-    Deliberately not a reinstall. Running `install.sh` every session would repair an
-    AgentPlugin whose rendered spec has drifted from this candidate's, but it repairs
-    nothing in the common case — an unchanged plugin source tree renders a byte-identical
-    CR — while adding unconditional writes to the RC's critical path, including four
-    `add-iam-policy-binding` calls that read-modify-write the *project* IAM policy. An IAM
-    race unrelated to the candidate would then block the release. Drift is worth detecting;
-    the checks below do that and say so, rather than papering over it with a write.
+    Skips the suite if the AgentPlugin CRD or the stockout plugin CR is absent from the
+    cluster, avoiding out-of-band cluster mutations or IAM writes. When the plugin is
+    installed, refuses to run on a broken one, distinguishing between operator
+    reconciliation failures, imageVolume mounting/staging failures, entrypoint linker
+    failures, and missing skill files.
 
     `SKIP_STOCKOUT=1` skips the whole thing, tests included.
     """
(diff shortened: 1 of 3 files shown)


tests/e2e/test_stockout_investigation.py
```diff
@@ -304,11 +298,24 @@ def _wait_for_plugin_ready(namespace: str, budget_deadline: float) -> Dict[str,
...
+                    message = ""
+                    for cond in conditions:
+                        if cond.get("status") == "False" or cond.get("reason"):
+                            reason = cond.get("reason", "")
+                            message = cond.get("message", "")
+                            break
+                    pytest.fail(
+                        f"AgentPlugin '{_PLUGIN_NAME}' in '{namespace}' installation failed: "
+                        f"phase is 'Degraded' (reason: {reason or 'Unknown'}): {message or detail}"
+                    )
```
mplakhtiy, line 318: **HIGH — `Degraded` with `observedGeneration == generation` is not a settled state, so this fails the module on conditions the operator recovers from.**

Neither half of the guard establishes that the operator is finished:

- `updatePluginStatuses` assigns `plugin.Status.ObservedGeneration = plugin.Generation` at the top of every pass (`k8s-operator/internal/controller/platformagent_controller.go:3004`), before `evaluatePluginReadiness` runs at `:3014`. `observed == generation` means the operator has looked at this spec, not that this is its final answer.
- The Degraded reasons are read off live pod state. `detectPluginFailures` maps `ImagePullBackOff` and `ErrImagePull` on the plugin's staging init container, or on the main container's image volume, straight into a failure (`:3076-3119`). Both are states a kubelet retries out of.
- `markOrphanedPlugins` writes `Degraded`/`AgentNotFound` with `ObservedGeneration` set (`:3230-3240`) when the plugin is reconciled before its `PlatformAgent` exists.

Failure scenario: step 2 deploys a new RC image, the gateway pod pulls the freshly pushed plugin image, one layer 429s, the pod sits in `ErrImagePull` for ten seconds, the operator writes `Degraded/ImagePullFailed`, and the fixture's first poll lands in that window and fails all 11 tests. The loop this replaces kept polling and passed once the pull retried, well inside `_PLUGIN_READY_TIMEOUT_SECONDS`. The fixture runs while the gateway may still be rolling — the docstring above says so — which is exactly when the window is open.

Require the Degraded observation to persist (several consecutive polls, or a grace window short of the 300s ceiling) before turning it into a verdict.

tests/e2e/test_stockout_investigation.py
```diff
@@ -734,9 +720,42 @@ def ensure_stockout_plugin_installed(
     if not gcp_project_id or not gke_cluster_name:
         pytest.fail("GCP_PROJECT_ID and GKE_CLUSTER_NAME are required for stockout E2E tests.")
 
+    expected = (
+        os.environ.get("ENABLE_STOCKOUT_INVESTIGATOR", "").lower() == "true"
+        or os.environ.get("E2E_SUITE") in ("rc", "nightly", "stockout-full", "investigations")
+    )
```
mplakhtiy, line 726: **MEDIUM — the suite list duplicates `tests/e2e/e2e_config.yaml`, and nothing keeps the two in step.**

These four names are today exactly the suites in `e2e_config.yaml` that list `tests/e2e/test_stockout_investigation.py` (`rc`, `nightly`, `stockout-full`, `investigations`; the other three suites do not run this file). That correspondence is what makes the gating behaviour correct, and it is unenforced.

Failure scenario: someone adds a suite that runs this file, or renames one of the four, and does not edit this tuple. `expected` is then false on a gating run, a missing AgentPlugin skips instead of failing, and the pipeline reports green — the false-green case this change was written to close.

A unit test in `tests/test_stockout_fixture_helpers.py` that parses `e2e_config.yaml` and asserts every suite listing this test file appears here would hold it.

tests/e2e/test_stockout_investigation.py
```diff
@@ -734,9 +720,42 @@ def ensure_stockout_plugin_installed(
...
+    check_crd = _kubectl("get", "crd", _CRD_NAME, fail_on_timeout=True)
+    if check_crd.returncode != 0:
+        if expected:
+            pytest.fail(
+                f"AgentPlugin CRD '{_CRD_NAME}' was expected on this environment, but is missing from cluster."
+            )
+        pytest.skip(
+            f"AgentPlugin CRD '{_CRD_NAME}' not found on cluster; "
+            "stockout investigator is not installed."
+        )
```
mplakhtiy, line 738: **MEDIUM — a kubectl that could not reach the cluster is read as "the CRD is absent".**

Any non-zero exit takes this branch, and the skip message then states something about the cluster the call never reached. Verified locally: `KUBECONFIG=/dev/null kubectl get crd agentplugins.kubeagents.x-k8s.io` exits 1 with `dial tcp 127.0.0.1:8080: connect: connection refused`.

This is the failure mode `_kubectl`'s own docstring argues against — "false statements about the cluster derived from a slow API server" — and the code being removed had an explicit guard for the neighbouring case (no kubectl context set produced a `pytest.fail`), which goes away here with nothing in its place. The same applies to the AgentPlugin check below.

Failure scenario: a developer with an unset or stale kubeconfig runs the suite, reads "AgentPlugin CRD not found on cluster; stockout investigator is not installed", and goes looking at the install instead of at their credentials. On the `expected` path the run fails rather than lies, so the damage is confined to optional runs.

Check stderr for a NotFound before concluding absence, and treat a transport error as its own outcome.

tests/e2e/test_stockout_investigation.py
```diff
@@ -540,10 +547,8 @@ def _verify_skill_mounted(
         # Conclusive: the exec ran, repeatedly, and the file is not there.
         pytest.fail(
             f"The stockout plugin's skill is not mounted in {absent_in}: {skill_path} does not "
-            f"exist, {window}s ({bound}) after the gateway rolled out. The AgentPlugin "
-            f"reconciled, so the alerts these tests publish would be investigated by nobody. "
-            f"Either the plugin image did not reach the '{target_profile or 'default'}' profile, "
-            "or the spec the operator reconciled is not the one this candidate installs."
+            f"exist, {window}s ({bound}) after the gateway rolled out. The AgentPlugin is present, "
+            f"but image volume mounting or staging failed to link the skill into profile '{target_profile or 'default'}'."
```
mplakhtiy, line 551: **MEDIUM — the message now names one cause, and the fixture cannot tell which cause it is.**

The text it replaces offered two hypotheses: the plugin image did not reach the profile, or the spec the operator reconciled is not the one this candidate installs. Nothing in this change adds evidence for narrowing that to "image volume mounting or staging failed" — the fixture still never inspects the pod template or the backing ReplicaSet for the plugin volume, so it cannot distinguish a mount failure from a workload the operator has not re-templated yet.

Removing the install makes the deleted hypothesis more likely rather than less: with the plugin pre-provisioned, an AgentPlugin left over from a previous release, or one whose re-template has not landed, is a leading explanation for a skill file that is not there.

Failure scenario: the next mount failure on the RC sends whoever reads the log into the image-volume plumbing when the actual cause is a stale CR or an unfinished rollout.

Either restore the two-branch wording, or resolve the current ReplicaSet and assert the plugin volume is on it — `_current_revision_selector` already resolves the ReplicaSet for pod pinning — which would make the single-branch wording true and turn a 120s timeout into an immediate answer.

mplakhtiy (commented):
Reviewed 30f541b0cd92bb0c9819aad73112095fbec5e8ce (full diff against `main` at fb6f281e). Produced with AI assistance.

Four findings, all inline: one on the new `Degraded` fast-fail, three on the new skip/fail gates and the reworded mount-failure message.

tests/e2e/test_stockout_investigation.py
sergiu1spiridon: Declared `_EXPECTED_E2E_SUITES = ("rc", "nightly", "stockout-full", "investigations")` at the top of `tests/e2e/test_stockout_investigation.py` and added `test_suites_running_stockout_match_expected_suites_tuple` in `tests/test_stockout_fixture_helpers.py`. The test parses `tests/e2e/e2e_config.yaml` and asserts that every suite listing `tests/e2e/test_stockout_investigation.py` matches `_EXPECTED_E2E_SUITES` exactly.

tests/e2e/test_stockout_investigation.py
sergiu1spiridon: Added `_DEGRADED_PERSISTENCE_SECONDS = 30` to `tests/e2e/test_stockout_investigation.py` and updated `_wait_for_plugin_ready` to track `degraded_first_seen`. The fixture now tolerates transient `ErrImagePull`/HTTP 429 retries across rolling deployments while still failing fast with the failing condition's reason and message if `Degraded` persists for 30s. Added unit tests in `tests/test_stockout_fixture_helpers.py` covering transient recovery to Ready and persistent Degraded failure.

tests/e2e/test_stockout_investigation.py
sergiu1spiridon: Added `_is_resource_not_found` helper in `tests/e2e/test_stockout_investigation.py` to inspect kubectl output for `NotFound` / `not found`. When `_kubectl` fails without indicating resource absence (e.g. connection refused, transport timeout, unauthorized), the fixture fails immediately via `pytest.fail` reporting the cluster connection error rather than misinterpreting it as plugin absence. Added unit tests in `tests/test_stockout_fixture_helpers.py` verifying transport error failures for both CRD and plugin probes.

tests/e2e/test_stockout_investigation.py
sergiu1spiridon: Restored the original two-branch wording in `_verify_skill_mounted` acknowledging both failure hypotheses (plugin image did not reach profile vs. operator reconciled a different spec from what was expected).

tests/e2e/test_stockout_investigation.py
sergiu1spiridon: Raised `_DEGRADED_PERSISTENCE_SECONDS` to 120s in `tests/e2e/test_stockout_investigation.py`. This accommodates kubelet's exponential backoff retry cycles on `ImagePullBackOff` (10s -> 20s -> 40s -> ...) while still failing fast on permanently broken specs well short of the 300s ceiling. Updated `tests/test_stockout_fixture_helpers.py` accordingly.

mplakhtiy (approved):

PR #1419, opened 2026-09-10
feat(dashboard): report release-candidate eval runs in a Releases section

stalhaali:
## Summary

The release-candidate eval lane has been running since #2691 merged, and until now the only
way to read its result was to open Deck and find the build by hand. This adds a Releases
section to the eval dashboard: the collector sweeps the `post-kube-agents-eval-rc` archive
into a new top-level `releases[]` in `data.json`, and both renderers draw one row per RC run
— candidate tag, tier, verdict, admitted-case pass rate against its baseline, cases passed,
start time, and how long the eval took. The non-inferiority comparison renders as advisory
and says why, because the baseline store holds nothing at most candidates' version keys yet.
This is the last reporting layer #1024 asks for; the lane still gates nothing.

## Why This Change

#1024 requires the RC verdict to be visible "without asking anyone", across three layers: an
optional status on the RC, the verdict in run artifacts, and the dashboard's Releases
section. The first two shipped in #1170 and #1230. Without this one, a release candidate's
full-suite result exists only in a Prow build log whose URL nobody has unless they went
looking, which is the same as not reporting it.

## What Changed

- `scripts/eval_dashboard/collect.py` — `--rc-glob` / `--rc-from-dir` / `--rc-limit` sweep the
  RC job's archive into `releases[]`. RC runs are **never** folded into `runs[]`: `runs[]`
  feeds `build_cases()`, which computes the main-branch pass rates a candidate is measured
  against, so adding a candidate to it would move its own baseline
  (`hack/ci-eval-pr.sh:1998`, "the baseline store is read, never written"). Collection is
  bounded by build id rather than by the presubmit watermark, and a recorded release is
  final, so `--merge-with` carries priors forward and only unseen ids are read.
- `scripts/eval_dashboard/render.py` and `template/index.html.tmpl` — the section, in both
  renderers. The page re-renders client-side from a polled `data.json` every 60s, so a
  section that existed only in `render.py` would revert to the placeholder within a minute.
- `.github/workflows/ci-health.yml` — the 15-minute refresh that actually publishes. It
  inlines its own `collect.py` call rather than running `hack/ci-dashboard-refresh.sh`, so
  arming the script alone would have left the live band on the placeholder forever.
- `scripts/eval_dashboard/SCHEMA.md` — `releases[]`, its derivation rules, and the new flags.
  `data.json` stays schema v1: `releases[]` is an additive optional field.
- `scripts/eval_dashboard/testdata_rc/` — one **real** captured RC build as the fixture.

Two things a reviewer should know up front. **The table renders on `legacy.html` only** — the
Brief is a curated summary page and the detail tables all live on the two-band page — so the
Brief carries a `Releases` deep link, which is the whole of its discoverability from the
landing page. And **this lands on the live dashboard, not a dormant one**: `ci-health.yml`
publishes to `gs://kube-agents-dashboards/evals` every 15 minutes and has since #1305, so
the Releases band appears at
`https://storage.cloud.google.com/kube-agents-dashboards/evals/legacy.html#release`
within a tick of merge. (`EVAL_DASHBOARD_TARGET` gates the two *other* publish paths — the
Prow periodic and the presubmit hook — and `ci-health.yml` does not read it.)

## Context

Part of #1024 ("Lane: release gate runs the full suite — non-gating, reported"), the fourth
and last PR of that lane. Predecessors: #1170, #1230, and
GoogleCloudPlatform/oss-test-infra#2691 (merged, `38d0ab78`) with follow-up #2695 open.

The surface this reports into comes from #1022 ("Lane: dashboard live at a URL"), which is
open but has been serving real data since 2026-08-31. Note for anyone with an older
bookmark: the `storage.googleapis.com` link in that issue's early comments returns 403 by
design since the 2026-09-05 access change; `storage.cloud.google.com` is the current one.

Checked the open queue for overlap before starting: nothing else touches
`scripts/eval_dashboard/`. Rebased onto `main` at `c176c440` after the drift check flagged
that #1400 had landed on `ci-health.yml` — it makes each WIF credential the active gcloud
account so `gsutil` stops running anonymously, which is what caused 40 failed dashboard
ticks earlier today. It rebased cleanly and the two changes are independent, but it is worth
knowing that the collect step this PR adds a `gsutil` call to was authenticating correctly
only as of a few hours ago.

## Testing

- `python3 -m unittest discover` in `scripts/`: 1403 tests, 1 skipped, green (re-run after
  the rebase below). 64 collector and 71 renderer tests, of which 28 are new on this branch.
- `make docs-check`: clean. `prettier@3.9.6 --check` on the changed Markdown and YAML: clean.
- **Renderer parity is now a committed test, not a claim.** The template's own `releasesHtml`
  runs under node and is diffed byte-for-byte against `render.py`'s over seven shapes: the
  real record, a negative margin, a missing banner, a hostile `javascript:` URL, entries that
  are not dicts, more releases than the row cap, and empty. Mutation-checked — changing one
  word of the Python table header reds six of the seven. It skips where node is absent, so a
  green run without node has not checked parity; the structural mirror test is the floor.
- `tests/test_eval_dashboard_refresh.py` **cannot run on this machine**: macOS ships bash 3.2,
  which errors on `"${arr[@]}"` under `set -u` when the array is empty. This is pre-existing
  and untouched tests fail identically, so I did not patch it. I verified the two new tests
  in it by replicating the script's inner pipeline by hand instead — see below.

### Live validation

**Not live-tested against a running kube-agents installation, and it cannot be:** this change
touches the eval dashboard's collector, renderers, and the GitHub Actions job that publishes
them. No part of it runs in a cluster, reconciles through the operator, or lands in an agent
pod. What it can be exercised against is real Prow data, and it was:

- **Real artifacts, not synthetic.** The fixture is a real `post-kube-agents-eval-rc` build
  (`2097891568546484224`, candidate `staging_2609092307_5b5ad10`), captured whole from
  `https://oss.gprow.dev/view/gs/kube-agents-prow/logs/post-kube-agents-eval-rc/2097891568546484224`.
  Its 240-line build log carries both banners the driver prints, including the decoy
  `RELEASE CANDIDATE EVAL TARGET` near the top that a naive substring match would open the
  parse on.
- **The real pipeline, end to end.** `collect.py` → `render.py` against that build plus the
  three existing presubmit fixtures:

  ```
  wrote .../data.json: 3 runs, 18 cases, 1 releases, 10/11 domains covered
  ```

  The parsed record matches the build page field for field: verdict `GREEN`, result
  `SUCCESS`, tier `nightly`, commit `5b5ad10`, project `kube-agents-evals-10`, started
  `2026-09-10T03:35:01+00:00`, duration 15006s (`4h10m`), pass rate `0.9`, and
  `baseline_rate`/`margin` both null — which is the advisory case, and the row says so
  rather than showing a bare number.
- **Which page it lands on, confirmed rather than assumed.** In the rendered out-dir the
  candidate tag appears twice in `legacy.html` and zero times in `index.html`, `run.html`,
  and `brief.json`. This is what caught the blocker below.
- **The Brief's deep link resolves.** `href="legacy.html#release"` in the rendered
  `index.html`, `id="release"` present in the rendered `legacy.html`.
- **The `ci-health.yml` publish guard, against its five real log shapes.** With the RC glob
  excluded, an RC-only `ls` failure publishes, an RC timeout publishes, a presubmit failure
  refuses, a clean log publishes, and a log carrying both still refuses. This matters more
  than it looks: `gsutil ls` exits non-zero on a prefix with no objects, so without the
  exclusion a week with no release candidate would have stopped every dashboard publish —
  an advisory lane taking the dashboard down with it.

**Not covered.** No run against the real bucket: I have no credentials for
`gs://kube-agents-prow` from here. The collect SA holds a read grant on the bucket rather
than on a prefix (`ci-health.yml`'s header), so `logs/` should read like `pr-logs/` does —
but "should" is the honest word, and it is the first thing to watch on the first real tick.
No screenshot of the live page. The dashboard **is** live —
`https://storage.cloud.google.com/kube-agents-dashboards/evals/index.html` — but it went
`domain:google.com`-only on 2026-09-05 (Public Access Prevention enforced, `allUsers`
removed), and this sandbox holds no Google credential, so I can reach neither the bucket nor
the page. The rendered pages I did verify are local renders of the real fixture, not the
published ones. The first post-merge tick is what closes that gap: the Releases band should
appear at `.../evals/legacy.html#release` within 15 minutes, and a reviewer with domain
access can confirm it in one page load.
No test artifacts were created outside temp directories.

## Self-Review

Both required passes ran in **subagent contexts that did not write the change**, spawned by
`/pr-preflight` and handed the diff range and nothing else. 13 distinct findings after
dedup (the two passes independently found the page-scope issue).

**Fixed:**

- **The refresh test asserted the RC tag lands in `index.html`; it lands in `legacy.html`.**
  A CI-reddening blocker, and the honest version of how it got there: the test lives in the
  file bash 3.2 will not run here, so I had shipped an assertion I had never executed.
  Corrected to assert `legacy.html` *and* assert its absence from `index.html`, so the
  page-topology claim is pinned in both directions.
- **The 15-minute refresh never passed `--rc-glob`.** `ci-health.yml` deliberately does not
  invoke `hack/ci-dashboard-refresh.sh`; I had wired only the script. The live Releases band
  would have sat on the placeholder indefinitely while every test passed. Fixed, along with
  the second-order hazard the pass flagged in the same breath — the publish guard would have
  let an empty RC prefix stall the whole dashboard.
- **`_release_sort_key` crashed on a retyped prior.** The prior filter validates `build_id`
  and nothing else, so a hand-edited `data.json` with a non-string `started` raised
  `TypeError` mid-sort and lost the entire merge. The renderer already coerced; the collector
  did not. Fixed with a regression test, verified to fail against the old code.
- **Wrong file cited for the "baseline store is read, never written" quote** in both
  `SCHEMA.md` and `collect.py`'s docstring — `ci-eval-rc.sh`, when it is at
  `hack/ci-eval-pr.sh:1998`.
- **`SCHEMA.md` wrongly listed `commit` among the fields that go null with no banner.** It
  falls back to the build's `revision`. My own no-banner test missed it only because that
  synthetic `finished.json` omits `revision` — a fixture gap, not a passing test.
- **The table was reachable only through the "Legacy view" link.** The Brief now deep links
  to it. I did not mirror the table onto the Brief: that is a third renderer path for a
  section that belongs with the other detail tables.
- Lower-severity: a stale `$1..$7` comment (nine positionals now); a `RELEASE_VERDICT_CLASS`
  assertion that could not fail, since `p-pass`/`p-fail`/`p-infra` are CSS class names the
  stylesheet defines — replaced with one that pins the pairs inside the mirror's map literal;
  `release_cases_cell` re-implementing `run_tasks`; magic `3600`/`60`; and doc wording on the
  `RED`→`SUCCESS` attribution, the row cap, and the `docs/README.md` row.
- **Fixture provenance recorded.** The pass flagged that the job name
  `post-kube-agents-eval-rc` appears nowhere in this repository, so a typo would degrade to a
  silent empty `releases[]`. It could not resolve that without GCS; I can, and the Deck URL
  is now in `SCHEMA.md` with a note saying why it is the place to check.

**Deliberately not fixed:**

- **The pass called my "a release alone is enough to render" branch unreachable and offered
  two fixes; I took neither.** It is unreachable through both publish pipelines, which is
  true — but widening the zero-runs floor to `runs or releases`, its first suggestion, would
  let a tick whose presubmit sweep failed publish a releases-only page *over a good
  dashboard*, which is precisely what that floor exists to prevent. Dropping the branch, its
  second, would make `render.py` show "no evaluation data yet" for a `data.json` that plainly
  has data, and `sorted_releases` already documents that the renderer runs against
  hand-edited files. The defect was my comment claiming a pipeline scenario that cannot
  happen; the comment and the test docstring now say what is actually true.
- **`hack/ci-eval-rc.sh`'s stale header** — it calls the job "the periodic job" and claims
  `EVAL_TIER` is read by nothing, which `ci-eval-pr.sh:1501-1509` and the fixture's log line
  16 both disprove. Pre-existing, in a file this PR does not otherwise touch, and correcting
  it here would mix an unrelated fix into a diff a reviewer is reading for something else.
  Worth its own change.

**What the passes could not cover**, in their words and mine: sibling-PR overlap (`gh` is
unauthenticated on this machine — I checked the queue through the web UI instead);
`tests/test_eval_dashboard_refresh.py`, which neither pass executed and nor could I; anything
requiring GCS; and any visual check, there being no published dashboard.

## Risk & Rollout

Small blast radius and no runtime code paths: nothing here ships in an agent image or runs in
a cluster. The one live surface is the 15-minute `ci-health.yml` tick, and the new failure
mode there is a collect step that now makes additional `gsutil` calls, against a prefix in a
bucket its service account can read but which nothing has exercised. If it turns out it
cannot, `collect.py` degrades a failed listing to a warning and an unchanged `releases[]`
rather than an error, and the publish guard excludes that warning on purpose — so the
dashboard keeps publishing and the Releases band stays on its placeholder. That degradation
is the point: #1024 requires this lane to block nothing, and a dashboard that stops
publishing because no release candidate was cut this week would be exactly that failure.
Watch the first tick for `warning: gsutil ls failed for
gs://kube-agents-prow/logs/post-kube-agents-eval-rc/*`; if it appears, the fix is a bucket
IAM grant, not a revert. Reverting is a plain revert: `releases[]` is additive, `data.json`
stays schema v1, and a published document carrying it renders fine against the old renderer,
which ignores the field.


bradhoekstra (approved):

scripts/eval_dashboard/collect.py
```diff
@@ -322,6 +373,107 @@ def build_run(build_id: str, read, pr_hint: int | None = None) -> dict | None:
...
+    """Assemble one `releases[]` entry, or None for a build still in flight.
+
+    `read(name)` is build_run's reader. The eval detail is parsed by the same
+    parse_build_log the presubmit uses -- the RC job runs the same
+    hack/ci-eval-pr.sh -- and only the banner on top of it is RC-specific.
+    """
+    run = build_run(build_id, read)
+    if run is None:
+        return None
+    banner = parse_rc_banner(read("build-log.txt") or "")
```
jayantid: 🟡 **Not merge-blocking.** One transient `gsutil cat` failure on an RC build's log records that candidate permanently as NO VERDICT with the note "no eval banner · job SUCCESS", and no later tick re-reads it.

**What's wrong.** `build_release` (collect.py:450) treats a build as final once `finished.json` reads. `_gsutil` returns `None` silently on a non-zero exit (collect.py:711). So a failed `build-log.txt` read becomes `parse_rc_banner("")`: `verdict`, `rc_tag`, `pass_rate` all null, `tasks` empty. `releases_from_gcs` then skips that id on every later tick through `known`. Reproduced with a reader that returns `None` for the log only.

**What happens if it merges.** A GREEN candidate renders as "no eval banner · job SUCCESS". That text tells the reader the driver exited on an early guard, which is false. No warning line is printed, so the publish guard passes. The only repair is hand-editing the published `data.json`. `runs[]` has the same freeze, but an empty run is merely left off the matrix. Here it misreports.

**The fix.** A log that would not read is not final. Skip it and let the next tick retry, the way a missing `finished.json` already does:

```python
def build_release(build_id: str, read) -> dict | None:
    if read("build-log.txt") is None:
        return None  # not final: retried next tick, like a missing finished.json
    run = build_run(build_id, read)
```

PR #1433, opened 2026-09-10
feat(inference): host the model in the user's cluster through LiteLLM's hosted_vllm provider

### LINKED ISSUE #1418: Leverage LiteLLM's support to host LLM in customer cluster
## Goal

Give a kube-agents user a third way to provide the agent with a model, beside an API key and a subscription: the install hosts the model itself, on GPUs in the user's own cluster. The user picks it as a model provider in `install.env`, the same way they pick a hosted provider today, and names any model the server can load. The agent is untouched: it keeps talking to the LiteLLM gateway, and the gateway is what learns where the model lives.

This is for clusters that cannot or must not send prompts to an external model API: air-gapped environments, and regulated ones whose policy keeps operational data inside the cluster.

## What done looks like

- One provider choice in the installer turns it on, with a sensible default model and a way to name a different one.
- The install brings up the GPU capacity and the model server it needs, on GKE Standard and Autopilot, and tears them down again.
- Model weights can come from Hugging Face with a token or from a bucket the cluster already trusts, so the path works with no internet egress.
- The gateway routes to the in-cluster server through LiteLLM's own support for self-hosted models; nothing model-specific lives in the harness.
- Existing installs on hosted providers render exactly what they render today.

## Requirements

- **An example**, in the style of `examples/litellm-chatgpt-subscription/`: a hand-applied set of manifests and a README a user can follow without the installer, verified end to end with one request through the gateway.
- **Testing coverage** at every tier the repository has: unit tests that run on every pull request for the render, the installer, and the infrastructure code; an integration test against a stand-in for the model server; and a manual critical-user-journey run against a real GPU node, recorded in each pull request's live validation.
- **Documentation** on the site's inference gateway page and in the install guide, kept to the pages that already own those topics.

## Relationship to #608

#608 explored this and drew review feedback that the harness should integrate self-hosted endpoints rather than ship model-specific code. This issue carries the goal forward on those terms; its example remains useful.

haoxuw:
## Summary

Adds `MODEL_PROVIDER=hosted_vllm`: the LiteLLM gateway routes `model-default` to a vLLM server running in the user's cluster, through LiteLLM's own `hosted_vllm` provider. No API key, no model-specific code, nothing defaulted: the install names the model the server was started with, the server's base URL, and its pod port, the way `vertex_ai` names a project and a location. The agent is untouched. Existing installs on hosted providers render exactly what they render today.

## Why This Change

Air-gapped clusters and policy-restricted environments cannot send prompts to an external model API. The gateway already abstracts the provider from the agent; this gives it an in-cluster one, using the provider LiteLLM ships for it ([LiteLLM's vLLM provider](https://docs.litellm.ai/docs/providers/vllm)) rather than a `custom` pseudo-provider.

## What Changed

The base config already renders `${MODEL_PROVIDER}/${MODEL_DEFAULT_NAME}`, so `hosted_vllm/<model>` comes out of the template that exists. LiteLLM reads the server address from `HOSTED_VLLM_API_BASE`, so the address is one environment variable, the way `VERTEXAI_PROJECT` is. Everything else mirrors `vertex_ai`:

- `charts/kube-agents/templates/litellm.yaml`, `values.yaml`: the provider guard admits `hosted_vllm`; `litellm.modelDefaultName`, `litellm.hostedVllm.apiBase` and `litellm.hostedVllm.targetPort` are each refused at render time when empty; the env var; one egress rule to pods in the gateway's namespace on the server pod's port (not the Service port, since NetworkPolicy matches after Service translation, and not every pod in the cluster).
- `install.sh`, `scripts/installer/installer_common.sh`: the provider is accepted with no default model; a fifth entry in both provider menus prompts for the three values; `HOSTED_VLLM_API_BASE` and `HOSTED_VLLM_TARGET_PORT` round-trip through `install.env` and `terraform.tfvars`.
- `terraform/examples/full-install/`: `hosted_vllm` in the `model_provider` validation; two pass-through variables, no defaults.
- `examples/litellm-hosted-vllm/`: `examples/litellm-gemini/` with the Gemini key replaced by the provider line and the env var; the gateway half of the pair with `examples/vllm-gemma/`. On the `check_iac_parity.py` roster, whose test now counts the roster instead of a literal.
- Docs: a row and a paragraph on the site's inference-gateway page pointing at LiteLLM's provider docs, the examples page, `install.env.example`, `terraform.tfvars.example`, and `docs/designs/model-access-paths.md`.

Left out on purpose: a kustomize overlay for the dev path and a `providers.json` row for the admin console. Both are the `vertex_ai` shape again and can follow. The model server itself is the existing example, applied by hand on a GPU node pool the user provides; the chart deploys no model server.

## Context

Closes #1418. Supersedes #608, closed in favour of this; its example under `examples/vllm-gemma/` stays as the model-server half of the hand-applied path.

## Testing

`tests/test_hosted_vllm_provider.py`: the render for `hosted_vllm` carries the provider line, the env var and exactly one same-namespace egress rule on the configured port; each of the three missing values fails the render naming it; `gemini`, `anthropic` and `openai` render none of it; the example's manifests carry the same provider line, env var and egress rule and need no Secret.

```
python3 -m unittest tests/test_hosted_vllm_provider.py tests/test_installer_common.py tests/test_install_script.py tests/test_vertex_location_defaults.py tests/test_litellm_rollout_quota.py tests/test_deployments_rollout_quota.py scripts/test_check_iac_parity.py   # all OK
python3 scripts/check_iac_parity.py   # 9 static copies
helm lint charts/kube-agents          # 0 failed
bash -n install.sh scripts/installer/installer_common.sh
make docs-check                       # green
```

### Live validation

Cluster `platform-agent-host` (GKE Standard, `us-central1`, project `haoxuw-gke-dev`), an existing kube-agents install on `gemini`, plus an on-demand `g2-standard-24` node (2×L4) added for the run.

1. **Model server.** `examples/vllm-gemma/` applied with `google/gemma-4-31B-it`, FP8, tensor parallelism 2, weights copied from Model Garden's public bucket by an init container. Two deviations from the example were needed for that model on that hardware and are recorded in the design doc §6: the example's `--max-model-len 32768` fails engine init with `CUDA error: out of memory` (FP8 weights are 16.5 GiB per L4); 16k context, 8 sequences and vision profiling disabled fit. Pod `1/1 Running`, 0 restarts.
2. **Direct request** to the server from an in-cluster pod: `pong`, 20 tokens.
3. **The chart's render applied.** `helm template` of this branch's `litellm.yaml` with the release's own values plus `modelProvider=hosted_vllm`, `modelDefaultName=google/gemma-4-31B-it`, `hostedVllm.apiBase=http://llm-service.kubeagents-system.svc.cluster.local/v1`, `hostedVllm.targetPort=8000`; the ConfigMap and NetworkPolicy applied and the env var set on the live Deployment (not a full `helm upgrade`, which would also rotate this install's session-KV salt); rollout complete.
4. **Request through the gateway** for `model-default` from an in-cluster pod: HTTP 200, `pong`, LiteLLM log shows the routed request. The gateway's config, env and egress rule are the rendered ones.
5. **Agent turn** from the platform-agent container (`hermes chat -q`, whose config targets the gateway's `model-default`): in progress; result to be added here.

## Self-Review

Pending: the adversarial and docs-drift passes have not yet run in a clean context. This section is filled in before the PR leaves draft.

## Risk & Rollout

Opt-in by provider choice; the `gemini`, `anthropic` and `openai` renders are asserted to carry none of the new material. The kustomize dev path does not yet know the provider and will fall through to the base config, which renders `hosted_vllm/<model>` without the env var; the overlay is the follow-up named above.


haoxuw:
Closing: #608 continues as the change of record for in-cluster inference; this branch stays on the fork as a reference.

PR #1439, opened 2026-09-10
feat(platform): keep triage cards out of goal mode and sweep cards stuck in blocked

### LINKED ISSUE #656: Goal-mode triage card sticky-blocks forever with result=NULL: an unsatisfiable format contract, escalated to a channel with no human
## Summary

A goal-mode triage card can finish its work, open and land a real GitOps PR, and still end up
parked in `blocked` **forever with `result = NULL`** — destroying the work product it just
produced.

Worked example: task `t_0a43cf9c` on 2026-08-12, cluster `platform-agent-host`
(`kubeagents-system`). The worker researched the incident, wrote a remediation dossier, and opened
[`gke-agentic/adamparco-infra#82`](https://github.com/gke-agentic/adamparco-infra/pull/82) — **now
merged**. It then called `kanban_block(kind="needs_input")` instead of `kanban_complete`, because
the goal-mode LLM judge would not accept its report. The card has been `blocked` ever since, it
will **never** be retried, and the finished report exists only in
`/opt/data/kanban/logs/t_0a43cf9c.log`.

It is the only blocked card in the board: **55 done, 1 blocked.**

This is adjacent to #630 but a **different failure**. In #630 the card reaches `status=done` with a
complete `result` that is never *delivered*. Here the card never reaches `done` at all and the
`result` column is never written — so there is nothing for any delivery fix to deliver. The two
share the `api_server` non-push substrate; the root cause and the fix differ.

## Environment

| | |
| --- | --- |
| Cluster | `platform-agent-host`, ns `kubeagents-system` |
| Image | `…/kube-agents/platform-agent:dev-8350e5f-0811-184934` |
| Hermes | `v0.20.0 (2026.8.3)` |
| Pod at failure | `platform-agent-gateway-7776bfc97-8f6b7` (claim lock) |

Paths under `/opt/hermes/…` below are the deployed Hermes runtime, not this repo.

## Forensic timeline (UTC, 2026-08-12)

| Time | Event |
| --- | --- |
| `10:11:02` | Session `k8s-evt-cc218c0d` calls `kanban_create` → `t_0a43cf9c`, `goal_mode=true`, `"subscribed": true` |
| `10:11:02` | Dispatcher claims as run `41` (`…-7776bfc97-8f6b7:4`) |
| `10:11:03` | Worker spawned, pid `31338` |
| `10:13:03` | **PR #82 opened** (`+26`, 1 file) — later merged |
| `10:11:33`–`10:16:39` | Six heartbeats, ~60 s apart. No crash, no timeout |
| `10:17:25` | Worker calls `kanban_block(kind="needs_input")`. Run 41 `ended`, `outcome=blocked`, `error=NULL` |
| `10:17:33` | Gateway wakes `k8s-evt-cc218c0d` with `[kanban] Task t_0a43cf9c blocked; needs attention` |
| `10:17:38` | Woken session calls `kanban_show` |
| `10:18:04` | Turn ends `text_response(finish_reason=stop)`, `response_len=1670`. **No `kanban_unblock`, no `kanban_complete`.** `wake self-post delivered` |
| — | Nothing further, ever |

Terminal card state:

```
status=blocked  block_kind=needs_input  block_recurrences=1
result=NULL     completed_at=NULL
consecutive_failures=0  last_failure_error=NULL  worker_pid=NULL
notify_subs: platform=api_server  chat_id=k8s-evt-cc218c0d  notifier_profile=default
```

`consecutive_failures=0` and `error=NULL` matter: **no circuit breaker, no crash, no timeout.** Every
automatic recovery path keys off signals this card does not carry.

The block reason is model-authored prose, not one of the two machine-generated `block_fn` strings in
`goals.py` (note the typo):

> The LLM judge is rejecting the task completion because of an exact formatting match check,
> although the diagnostic report output exactly matches the task body's layout specification. This
> is a known goal evaluation mismatch blocker. **Escalling** to the SRE/operator for manual override
> or check of the completed PR #82.

So this is `outcome="blocked_by_worker"` — the worker's own `kanban_block` tool call — not the goal
loop's budget-exhaustion path. `kanban_tools.py:238` (`_GOAL_MODE_BLOCK_ALLOWED_KINDS =
frozenset({"dependency", "needs_input"})`) is what permits it.

### 1. A presentation template is silently promoted into machine-graded acceptance criteria

`agents/platform/scripts/session_kv_server.py:285` instructs the session to post its report
*"formatted exactly like this"*, followed by the emoji block at `:286-296`:

```
📋 **Incident Triage**
- **Issue:** …
🛠️ **Proposed Fixes (GitOps):**
✅ **Recommended: Option <letter>** — …
```

That is a **chat-delivery spec**. But the default profile cannot call `send_notification` (exactly
as #630 documents), so the session delegates via `kanban_create` — and copies the template verbatim
into the **card body**.

Two things then happen to that body:

- Hermes's `apply_kanban_report_format.py` appends its `## Report format` stanza
  (`/opt/hermes/tools/kanban_report_format.py:101,109`), which mandates *"Use `##` and `###` for
  sections."*
- `/opt/hermes/cli.py:17896` builds the judge's goal text — with a comment that says the quiet part
  out loud:

```python
# Resolve goal text from the card (title + body = the acceptance
# criteria the judge evaluates against).
goal_parts = [task.title or ""]
if task.body:
    goal_parts.append(task.body)
goal_text = "\n\n".join(p for p in goal_parts if p).strip()
```

The body now carries **two mutually unsatisfiable format contracts**: an emoji/bold template with
**zero `##` headings**, and a stanza requiring `##` headings. Satisfying either violates the other.

The irony is that `kanban_report_format.py`'s own docstring states the stanza is advisory —

> Nothing refuses a completion over shape.

— which is true of the delivery path, and false the moment `goal_mode` grades against `title + body`.

(There is a **third**, also-different contract in `agents/platform/SOUL.md:167-170`, which mandates
`### 1. Issue` / `### 2. Root Cause` / `### 3. Recommendation`.)

### 2. The escalation has no reachable human, and the block is permanent

`kanban_block` writes a `blocked` event, which makes the card **sticky**
(`/opt/hermes/hermes_cli/kanban_db.py:4101`):

```python
if cur_status == "blocked" and _has_sticky_block(conn, task_id):
    # Worker / operator asked for human review — do not
    # silently auto-recover.  ``unblock_task`` is the only
    # legitimate exit …
    continue
```

`recompute_ready` (`:4139`, skip at `:4181`) will therefore never promote it. The only exit is an
explicit `hermes kanban unblock`, which emits `unblocked` and flips the predicate.

That would be fine if a human saw the escalation. The card's sole subscriber is
`platform=api_server`, `chat_id=k8s-evt-cc218c0d` — the non-push channel #630 dissects. The gateway
woke that session, it read the card, emitted 1670 characters of text into a channel nobody reads,
and never touched a kanban write tool.

**Net: a card whose only defined exit is human action, escalated to a channel with no human, and
excluded by design from every automatic recovery path.** Nothing sweeps it. `blocked` is a
permanent terminal state on this path.

### 3. The trigger is nondeterministic, so it hides

Sibling `t_eb3efe84` hit the **identical** judge rejection at `10:25:11` (its block reason names the
same "exact formatting match check"). Its worker process was still alive, called `kanban_complete`
a minute later at `10:26:27` with `result_len=1508`, and the card went `done` — PR #89.

Same conflict, opposite outcome, decided by whether the worker chose to finalize or to escalate. So
the contradiction is systematic while the failure it causes is intermittent: it will keep surfacing
as a rare, hard-to-reproduce stuck card rather than a clean repro.

**Caveat on evidence:** the per-turn `kanban goal loop: turn N/M verdict=…` lines from
`goals.py` are not present in any retained log (`/opt/data/logs/*.log` or the worker's
`/opt/data/kanban/logs/t_0a43cf9c.log`), so the judge's individual verdicts could not be recovered.
The judge-rejection mechanism is established from the structural conflict above plus the worker's
own recorded diagnosis, not from captured verdict lines. Making that logger reachable is itself
worth doing (see fix 4).

### Fix 1 — separate "how to present" from "what counts as done" (this repo, primary)

In `_build_agent_query` (`agents/platform/scripts/session_kv_server.py:248`):

- Reword `:285` so the template explicitly governs **the chat post**, not the card `result` — e.g.
  *"When you post to chat, use this shape"* rather than *"formatted exactly like this"*.
- State that if the work is delegated to a kanban card, the presentation template **must not** be
  copied into the card `body`. It belongs in `metadata`, or is applied at delivery time.
- Give the card an explicit, satisfiable definition of done for the judge to grade, e.g.
  *"Done when: root cause identified; ≥1 GitOps option proposed; result recorded via
  `kanban_complete`."*

This is the smallest change that removes the contradiction at its source, and it holds regardless of
what Hermes injects.

### Fix 2 — never let a judge disagreement destroy a finished work product

A worker that has done the work should **record it and then flag**, not discard it. Two parts:

- **Persona guidance** (`agents/platform/SOUL.md`): never `kanban_block` purely because a judge
  disagrees about shape. Write the result with `kanban_complete` first; a blocked card with
  `result = NULL` loses the work. If a human decision is genuinely required, complete with the
  finding and open a follow-up.
- **Upstream (`goals.py`)**: when the judge returns `continue` N times running and the worker's
  response is materially unchanged, prefer completing with a shape *warning* over blocking —
  matching `kanban_report_format.py`'s stated intent that nothing refuses a completion over shape.
  Worth filing against Hermes; I can open that separately.

### Fix 3 — a sweep for undeliverable escalations

A `needs_input` block on an `api_server`-subscribed card is unreachable by construction. Add a
stalled-card sweep to `agents/platform/cron/jobs.json` (alongside the existing audits and the
30-minute `GitHub Issue Resolver`) that surfaces any card `blocked` longer than a threshold to the
home channel, with its id, reason, and age.

This is the backstop that would have caught this in minutes instead of never, and it is independent
of the #630 delivery fix. Ideally #630's fix also routes `needs_input` escalations for `api_server`
cards to the home channel.

### Fix 4 — make the goal-loop verdicts observable

`run_kanban_goal_loop` logs each verdict via `log=lambda m: logger.info(...)`, but none of those
lines survive in the retained logs, which is the single biggest reason this took a full trace to
diagnose. Ensure the worker's logger output lands in `/opt/data/kanban/logs/<task>.log`.

## Immediate remediation for the stuck card

The work is done and PR #82 is merged, so the card only needs closing:

```bash
hermes kanban unblock  t_0a43cf9c --reason "Judge/template conflict; PR #82 merged"
hermes kanban complete t_0a43cf9c --result "<report from /opt/data/kanban/logs/t_0a43cf9c.log>"
```

## Acceptance criteria

- [ ] A triage card's `body` no longer contains two conflicting format contracts.
- [ ] A goal-mode worker that has produced a report never lands in `blocked` with `result = NULL`.
- [ ] Any card `blocked` beyond a threshold is surfaced to a channel a human reads.
- [ ] Goal-loop judge verdicts appear in the per-task worker log.
- [ ] Regression test: a card whose body carries a presentation template still reaches `done`.

---

Filed from a live trace on `platform-agent-host`; investigation and write-up produced with the help
of Claude.

bradhoekstra:
## Summary

Two changes for #656, one per commit. The event-triage prompt now tells the front door to file the card **without** `goal_mode` and gives the card body an explicit **Done when**, so a triage worker can never again be trapped by a goal-mode judge grading a presentation template. And the existing board-health check, `kanban_board_health.py`, moves to the Platform Agent's roster as a daily `no_agent` job delivering through the chat relay, reporting every card blocked for more than a day with its kind, reason, age and the commands to move it, regardless of its severity floor.

## Why This Change

On 2026-08-12 a triage worker finished its work and opened a real GitOps PR, and the card still sat in `blocked` forever with `result = NULL`. Its body, a report template, had been graded as acceptance criteria by the goal-mode judge; the judge rejected a correct report; the worker escalated with a sticky `needs_input` block. Most of what the issue describes has since moved: the template conflict is gone (#717, #738) and a block reaches the alert's chat thread (#738). Two things remained. Nothing stopped the front door from filing the card in goal mode, and upstream Hermes now gates `kanban_complete` on the judge for goal-mode cards, so a rejected worker cannot complete at all; the issue's "complete first, then flag" is impossible worker-side. And nothing periodically asked whether a card was stuck: a check existed, unscheduled, filtering out the one rule that mattered.

## What Changed

- `agents/platform/scripts/session_kv_server.py`: `_build_agent_query` names `goal_mode` among the call's fields ("leave it unset") and carries a fourth rule saying why; `_triage_task_body` gains a **Done when** paragraph (root cause with evidence; at least one GitOps option or an explicit no-change finding; report recorded with `kanban_complete`; the shape is not a requirement and formatting is never grounds for `kanban_block`; the three facts go in the one-line `summary` too, because Hermes's judge grades `summary` before `result`). It is not a `##` heading and sits outside the span the bench delivery contract slices, so the three-section rule, the roundtrip suite and the bench contract pass unchanged. Three new tests pin both.
- Personas and config: the Planning Agent's SOUL says when `goal_mode` is and is not for; the Cluster Agent's SOUL says a shape disagreement is never grounds for a block and that a refused report goes in the block's `reason`, the only text a block persists (its pre-existing preflight step named `summary` and `metadata` arguments `kanban_block` does not have, now corrected); the platform `config.yaml` comment cites the instruction instead of assuming it; `session_management.md` records both.
- `agents/platform/scripts/kanban_board_health.py` (moved from `agents/chat/scripts/`, history kept): finds the board through `PLATFORM_AGENT_HOME` rather than `HERMES_HOME` (which under the platform roster is `profiles/platform` and holds no board); asks the engine for every severity and applies the floor itself so `stuck_in_blocked` is always reported; reads each stuck card's kind and reason from the board (the engine's JSON has neither; title and assignee come from the board or, failing that, the engine) and renders them with the `unblock` and `archive` commands against the home this run read; clips the engine's multi-paragraph detail to one line for chat. 32 tests.
- `agents/platform/cron/jobs.json`: `kanban-board-health`, daily at 07:35, `no_agent`, `risk: low`, `deliver: "chat"`. The rule is stateless, so a card still blocked is named once a day until someone acts; that is the intended nag.
- Docs: cron README section; the reference and concepts pages count five `no_agent` entries and describe the sweep; the generated roster table; one bullet in the communication design; the moved script's link.

## Context

Closes #656. What the PR takes from each proposed fix and what stays upstream is in the issue's claim comment: Fix 1 in a different form, Fix 3, and the persona half of Fix 2 here; the `goals.py` change, Fix 4 (verdict logging) and the acceptance criterion about a goal-mode card reaching `done` are Hermes's. Sticky-block semantics are unchanged; `deploy/docker/patches/verify_kanban_scheduling.py` pins them as intended. Rebased onto `main` after #1437 merged: its `chat-delivery-watch` and this PR's `kanban-board-health` now sit together on the roster, and the two site pages count six `no_agent` entries. Rebased again onto `main` at c9c6dbdf on 2026-09-14, without conflicts. Merge-conflict warnings, not overlap: #591, #589, #1362 and #1368 append to the same roster file, and #1118 edits `autonomous-watchdogs.md`.

Generated with the help of Claude Code (Fable 5.1).

## Testing

Under Python 3.12 (the repo's pin; the machine's 3.11 lacks `fastapi` and cannot import the bench harness):

- `agents/platform/scripts/test_session_kv_server.py`: 191 tests pass, three of them new.
- `agents/platform/scripts/test_triage_reply_roundtrip.py` (13) and `bench/tests/test_triage_delivery_contract.py` (8): pass unchanged, which is the proof that the parsers of the card body were not disturbed.
- `agents/platform/scripts/test_kanban_board_health.py`: 32 tests, 11 new.
- `agents/platform/skills/fleet-audit/scripts/test_audit_report.py`: 605 tests, roster invariants included.
- `make prompt-check` (19 jobs), `make docs-generate`, `make docs-check`: pass. `prettier@3.9.6` on changed Markdown: clean.

### Live validation

On gkedemos (project `bhoekstra-gkedemos`, cluster `kube-agents-cluster`, namespace `kubeagents-system`, gateway pod `platform-agent-gateway-f9db8f877-z2tcg`), under the live-test lease. Both the agent and the operator there run the same month-old dev image (`dns-endpoint-7dac349`).

- **The sweep, against the real board.** The moved script was copied into the pod's shared `scripts/` directory and run by hand with the pod's own Python, once with the container's `PLATFORM_AGENT_HOME` and once with `HERMES_HOME=/opt/data/profiles/platform` as the roster would run it. Both found the board. The board holds five cards blocked since August (13 archived, 17 done). The output listed the four circuit-breaker blocks as the engine's `[error]` diagnostics and, under the new heading, the one sticky block: `t_5152a3f9 (platform, needs_input, blocked 24d): Workload Reliability Audit …` with its block reason quoted from the board and the `unblock` and `archive` commands beneath it. Exit code 0. The four breaker blocks did not appear under that heading because a breaker block writes no `blocked` event, which is the engine's own definition of stuck, and is not sticky. The engine's multi-paragraph detail lines were the reason the detail is now clipped to one line.
- **Not exercised: the prompt change and the scheduled leg.** The new `session_kv_server.py` imports siblings the deployed image does not ship (`findings_queue`, the #738 routing), so a second server instance started from it in the pod failed at import and no real triage turn could be driven through the new instruction. The roster entry reaches a volume only through the image's `merge_cron_store`, so the scheduled leg through the relay was not run either. Both need a current agent image, and swapping one in beside the month-old operator on the shared install was not something to do unilaterally. What stands in: 191 tests on the prompt module including the three new ones, and the roundtrip and bench contract suites passing unchanged under Python 3.12.
- Cleanup: the copied scripts restored to the image copies (verified byte-identical), the second server never ran, the probe pod created for the triage attempt deleted, lease released.

## Self-Review

Two passes, each run twice in fresh subagents that saw only the diff range (`review-preflight`: adversarial and docs-drift), the second round against the head before the last commit. Angles the adversarial pass took: does the prompt change actually stop the front door setting `goal_mode`, or only ask; does the Cluster Agent's escape route survive `kanban_block`'s real signature and the goal-mode kind allow-list; does the sweep read the right board when `PLATFORM_AGENT_HOME` is set, when `kanban/current` points elsewhere, and when the board predates `block_kind`; can its rendering leak a long or multi-line reason into chat; does the always-report path still filter the other warnings; roster invariants and the generated cron table. The docs-drift pass checked every page that names the `no_agent` jobs, the cron rosters, the health script's old path, and the two personas' kanban rules.

Fixed, all on the head (a331e4d4 carries the second round, 1a494a3c the two review findings):

- **`kanban_block` has no `metadata`** (adversarial, high). The Cluster Agent's escape sentence told the worker to put the report in a parameter the tool does not take. It now says `kanban_block(kind="needs_input", reason=<the whole report>)`, and names why that kind: a goal-mode card may only block with `needs_input` or `dependency`.
- **The goal judge reads `summary` before `result`** (adversarial, medium). "Done when" now asks the worker to state the three facts in the one-line `summary` too.
- **Printed operator commands pinned the wrong board** when `kanban/current` is set (adversarial). Both commands now set `HERMES_KANBAN_DB=<agent home>/kanban.db` next to `HERMES_HOME`, matching how the script pins the engine.
- **Board path from `HERMES_HOME`** would have opened `profiles/platform/kanban.db`, which does not exist; the board lives at the agent home. The script now resolves it through `gitops_workspace.agent_home()` with a `KANBAN_HEALTH_HOME` override, tested.
- **CLI fallback and enrichment**: the block kind is read from the latest `blocked` event, not the first; the reason is clipped to one line of 200 characters; an unmigrated board without `block_kind` still reports. Each has a test.
- **The sweep fired at 03:35 EDT** while the docs said every morning (toshiowang, low). It now fires at `35 12 * * *`, 12:35 UTC, half an hour after `findings-morning-nudge`; the cron README says why, and the cadence map and generated table follow.
- **The sandboxing design mis-cited the SOUL** (toshiowang, low): it put the `kanban_notify_propagate.py` instruction in a "fan-out bullet" of §0. §0 carries it in the sub-card paragraph under *Show your progress*; the fan-out bullet is §6's and repeats it. The sentence names both and cites the ban as §0's.
- **Design doc drift** (docs-drift): the `agent-communication.md` bullet omitted the rule's "no comment or unblock since" condition, overstated "only a human's unblock" (the Planning Agent holds `kanban_unblock`), and restated the cron README; it now carries the condition and links the README. `agent-shell-sandboxing.md` cited `SOUL.md` by line number; it names the section. `docs/README.md` named only the chat roster as the cron page's source; it names both. The sandboxing paragraph's description of the health script's open mode was updated with the move.

Deliberately not fixed:

- **The site pages carry hard-coded counts** ("five `no_agent` jobs"). Both passes flagged them as the next drift. They were already hard-coded at four; making them generated is a `generate_docs.py` change outside this fix, and the counts are pinned by `test_audit_report.py`'s roster set, which fails when one is added without touching the doc.
- **The goal-mode fact is stated in several places** (router rule, chat SOUL, cluster SOUL, config comment, session_management doc). The router rule is the one that binds and is tested; the others are each the place a different reader looks. Each names the others by rule number or issue rather than restating the mechanism.
- **Merge conflict with #1437**, resolved by the rebase: both entries on the roster, the count sentences and the `no_agent` row read six and name both ids. The conflict touched only the two site pages; the roster and the test's roster set merged cleanly.
- **Ride-along fix in the cron reference** (the `deliver` row said `"chat"` only): dropped in the rebase, because #1437 landed the same correction on `main`.

Not covered: the prompt half was not exercised end to end on the live install (the running image predates `findings_queue`, so the new Session KV server cannot start there; see Live validation), no real goal-mode judge rejection was reproduced (the change removes goal mode from the card), and the Hermes behaviour was read from the pinned 0.20.1 copy, not run against a newer release. Commit 1a494a3c, a cron expression and one sentence answering the two review findings, has not been through a third pass.

## Risk & Rollout

The prompt change is prose for a router that has improvised before; it is one bullet and one rule, pinned by test, and it changes what the card asks for, not what any worker may do. The sweep runs as a subprocess with no model, reads the board read-only and shells out to `hermes kanban diagnostics`; on an install with a stuck board it posts one message a day until someone acts, which is the feature. Revert is one revert commit; the roster entry then needs the usual two-release retirement, since `merge_cron_store` never prunes.


docs/designs/agent-shell-sandboxing.md
```diff
@@ -1956,13 +1956,14 @@ None of them blocks the work above. All three are recorded here so the dead ends
...
-29 that opening `/opt/data/kanban.db` from an agent shell is what the persona forbids.
+[`kanban_board_health.py`](../../agents/platform/scripts/kanban_board_health.py) opens
+it read-only (`mode=ro`) for its invariant and blocked-card queries and shells out to
+`hermes kanban diagnostics --json` for the rule engine; its docstring records that a
+read-write open of `/opt/data/kanban.db` from an agent shell is what the persona forbids.
 [`kanban_notify_propagate.py`](../../agents/platform/scripts/kanban_notify_propagate.py)
-does open it, `sqlite3.connect` at line 63 — and `SOUL.md:61` tells the agent to run it
-from the shell. That is coherent today, where `SOUL.md:66`'s ban on touching the board
-is a ban on ad-hoc edits and the script is a sanctioned writer, but it does not survive
+does open it, `sqlite3.connect` at line 63 — and the Platform Agent's `SOUL.md` (§0, the
```
toshiowang, line 1964: LOW. The instruction to run `kanban_notify_propagate.py` from the shell is not in §0. The only mention in the Platform SOUL is in §6 "Delegation & Cluster-Agent Lifecycle", under "Multi-cluster work (fan-out)" (line 183 at this head). §0 holds the board-write ban (line 85) but not the run instruction, so the next sentence's "the same section's ban" is also off: the ban and the instruction are in different sections. This hunk replaced the `SOUL.md:61` citation, so the drift is new here.

agents/platform/cron/jobs.json
```diff
@@ -174,6 +174,22 @@
       "enabled": true,
       "deliver": "chat"
     },
+    {
+      "id": "kanban-board-health",
+      "name": "Kanban Board Health",
+      "schedule": {
+        "kind": "cron",
+        "expr": "35 7 * * *",
```
toshiowang: LOW. `35 7 * * *` is 07:35 UTC, which is 03:35 EDT, while the cron README and the watchdogs page say the card is named "every morning". The roster's other daily nag, `findings-morning-nudge`, sits at `0 12 * * *` for this reason. Either move the schedule or drop "morning" from the prose.

toshiowang (commented):
Adversarial pass at a331e4d4: no blocking or high findings, two low ones inline (both docs).

What held up under source: every Hermes claim in the body was checked against v2026.8.19 (the `tags.env` pin, which is 0.20.5 rather than the 0.20.1 the body names) — `kanban_block` takes only `reason` and `kind`, goal-mode blocks are limited to `needs_input`/`dependency`, the judge grades `summary or result`, `diagnostics --json --severity` and `data.age_hours` exist, the `blocked` event payload carries `reason` and `kind`, `unblock --reason` and `archive <ids>` exist. Empty stdout on a healthy day is dropped by the scheduler's silent marker and again by the relay adapter. `import gitops_workspace` from a roster script has precedent in `chat_delivery_watch.py`, and `PLATFORM_AGENT_HOME` reaches the child through `build_subprocess_env`.

Run: the five suites the body names, `make prompt-check`, `make docs-check`, prettier. Red-green: 15 of 32 tests in the new test file fail against the pre-change script, including all three stuck-in-blocked tests. Merge onto current `main` is clean.

The red smoke test is infrastructure: the pod hit `NodeNotReady` at 76 minutes and left no build log, so a `/retest` is warranted.

Not covered: no scheduled run through the relay, no goal-judge rejection reproduced, no bench eval.

toshiowang (approved):

docs/designs/agent-shell-sandboxing.md
bradhoekstra: Fixed in 1a494a3c. §0 does carry the instruction, but in the sub-card paragraph under "Show your progress" (the code block at line 67), not in a fan-out bullet; the fan-out bullet is §6's and repeats it. The sentence now names both, and the ban is cited as §0's (line 85) rather than "the same section's".

agents/platform/cron/jobs.json
bradhoekstra: Moved in 1a494a3c to `35 12 * * *`, 12:35 UTC, half an hour after `findings-morning-nudge`, so "every morning" holds. The cron README says why that hour; the cadence map and the generated table follow.

Issue #1189, opened 2026-09-02
rca-remediation-pr: rung-4 collapses on unrelated PRs — delegation-chain exposure amplifies #1097/#1144, blocked reps graded not classified (#1184)


jayantid:
`rca-remediation-pr` — an admitted (blocking) case — hit rung-4 collapse (all repetitions failed) on at least six runs on 2026-09-02, across entirely unrelated PRs: #1112, #1089, #1177, #1175, #1153, #1142, with the same shape appearing in #1134/#1124's runs. None of those diffs touch the remediation path. This is the same pattern that led to compliance-rbac-overgrant's demotion (#1171/#1173), one day later, on the suite's second-most delegation-exposed case.

## Root cause — three stacked causes, none of them the case's own logic

The case is the suite's **longest delegation chain**: the platform agent files a kanban card, a worker investigates and opens a GitOps PR, the agent synthesizes the result. Every link is a network/runtime dependency, so the case integrates over every environment fault in its ~5–20 minute window — and today there were three:

1. **#1097 — agent-endpoint slot saturation.** During the 18:00–20:00Z storm windows, delegation turns died on `litellm.RateLimitError ... 429`, leaving empty-trajectory repetitions. Short cases dodge the window; rca's chain almost never does.
2. **#1144 — credential-proxy EACCES on `/opt/data/profiles/platform`.** The worker lane dies mid-chain ("kubeconfig is unreadable" shape); fix #1183 is tide-queued.
3. **#1184 — infra-blocked repetitions are *graded*, not classified.** When causes 1–2 kill a rep, the record often lacks the `KUBE_AGENTS_INFRA_FAILURE` marker, so a blocked rep grades as a failure instead of being excluded. Three reps sharing one dirty window → correlated fails → rung-4 collapse → an innocent PR reds. This is the structural amplifier: with correct classification, causes 1–2 would produce excluded INFRA reps, not collapses.

Evidence trail: per-run classification comments posted by the gate watch on each PR above; the delegation-collapse shape is documented on #1010 (comment ref there, 2026-09-02); case history: 12/13 clean post-fix-wave record 08-28→09-01 — the degradation is precisely correlated with today's storm windows, not with any change to the case or the remediation path.

## Decision pending

[...]

PR #1441, opened 2026-09-10
feat(litellm): add the hosted_vllm provider for a vLLM server in the cluster

### LINKED ISSUE #1418: Leverage LiteLLM's support to host LLM in customer cluster

haoxuw:
## Summary

Adds `MODEL_PROVIDER=hosted_vllm`: the LiteLLM gateway routes `model-default` to a vLLM server running in the user's cluster, through LiteLLM's own `hosted_vllm` provider. No API key and no model-specific code in the harness. Nothing is defaulted: the install names the model the server was started with, the server's base URL and its pod port, the way `vertex_ai` names a project and a location, and the chart refuses to render without any of the three. The agent is untouched, and the `gemini`, `anthropic`, `openai` and `vertex_ai` renders carry none of it.

## Why This Change

Air-gapped clusters and policy-restricted environments cannot send prompts to an external model API. The gateway already abstracts the provider from the agent; this gives it an in-cluster one, using the provider LiteLLM ships for it ([LiteLLM's vLLM provider](https://docs.litellm.ai/docs/providers/vllm)) rather than a `custom` pseudo-provider with a placeholder key.

## What Changed

The base config already renders `${MODEL_PROVIDER}/${MODEL_DEFAULT_NAME}`, so `hosted_vllm/<model>` comes out of the template that exists. LiteLLM reads the server address from `HOSTED_VLLM_API_BASE`, so the address is one environment variable, the way `VERTEXAI_PROJECT` is. Everything else mirrors `vertex_ai`.

- **Chart** (`templates/litellm.yaml`, `_helpers.tpl`, `platform-agent-cr.yaml`, `values.yaml`, `values.schema.json`, `NOTES.txt`, README): the provider guard admits `hosted_vllm`; `litellm.modelDefaultName`, `litellm.hostedVllm.apiBase` and `litellm.hostedVllm.targetPort` are each refused at render time when empty; the env var; one egress rule to the namespace read from the URL (as the OTLP collector's is read from `telemetry.otlpEndpoint`) on the server pod's port, and a URL that is not an in-cluster Service fails the render. Since #1195 the operator renders `litellm-policy` on a stock install, so the chart also stamps that namespace and port on the PlatformAgent CR as the `kubeagents.x-k8s.io/litellm-upstream` annotation, the way it stamps `otlp-collector-namespace`, with the same contradiction check.
- **Operator** (`platformagent_litellm_policy.go`): reads `litellm-upstream` as `<namespace>:<port>` and appends the rule; a malformed value adds nothing and is logged. The operator learns a namespace and a port, not a provider.
- **Installer** (`install.sh`, `installer_common.sh`, `common.sh`, `install.env.example`): the provider is accepted with no default model; `--hosted-vllm-api-base` and `--hosted-vllm-target-port` flags; a fifth entry in both provider menus; a missing value or a non-numeric port stops the run before the cluster is built; both keys round-trip through `install.env`, the interview-drift check and `terraform.tfvars`.
- **Terraform** (`full-install`): `hosted_vllm` in the `model_provider` validation; two pass-through variables with a numeric check on the port; README bullet.
- **Release environments**: `render_install_env.sh` MAPPING, `provision_environment.sh` flags, and the two environment workflows carry the pair, as they carry `VERTEX_*`.
- **`examples/litellm-hosted-vllm/`**: `examples/litellm-gemini/` with the Gemini key replaced by the provider line and the env var; the gateway half of the pair with `examples/vllm-gemma/`. On the `check_iac_parity.py` roster.
- **`examples/vllm-gemma/`**: carries the example work from #608 (digest pin, `securityContext`, README structure, air-gapped and Hugging Face options) sized for the agent: `google/gemma-4-E4B-it` on one L4 at a 131072-token context, because the agent requests 65536 output tokens per call and vLLM rejects a request whose output budget exceeds the context; an optional `HF_TOKEN` Secret the gated model needs for an online pull; `main`'s strategy block kept.
- **Docs**: the site's inference-gateway page (decision table, `MODEL_PROVIDER` table, vLLM section pointing at LiteLLM's provider docs), examples page, prerequisites, security page, environment-reconcile variable list, INSTALL.md, installer README, install skill, root README diagram, `docs/designs/model-access-paths.md`.

Left out on purpose: a kustomize overlay for the dev path (`make deploy-litellm` renders the provider line without the address or the egress rule, and the site page says so) and a `providers.json` row for the admin console. Both are the `vertex_ai` shape again and can follow. The chart deploys no model server.

## Context

Closes #1418. Supersedes #608, whose review settled the direction here and whose example work this carries (co-authored on the example commit). #1195 merged first and moved `litellm-policy` under the operator, so the merge commit here carries the rule into the reconciler through the CR annotation described above; without it a stock `hosted_vllm` install would render fine and the gateway could not reach the server.

## Testing

Unit (`tests/test_hosted_vllm_provider.py`, `tests/test_installer_common.py`, `tests/test_install_script.py`, `platformagent_litellm_policy_test.go`): the render for `hosted_vllm` carries the provider line, the env var and exactly one namespace-scoped egress rule on the configured port; each missing value fails naming it; a URL outside the cluster fails; the rule names the namespace in the URL; `gemini`, `anthropic`, `openai` and `vertex_ai` render none of it; the installer accepts the provider with no default model and parses both flags; both tfvars lines are emitted; the gateway example points at the server example's Service and port and needs no Secret; the stock render stamps the `litellm-upstream` annotation on the CR and a contradicting `platformAgent.annotations` entry fails; the reconciler adds exactly one rule for a well-formed annotation and none for six malformed shapes.

```
python3 -m unittest tests/test_hosted_vllm_provider.py tests/test_installer_common.py tests/test_install_script.py tests/test_vertex_location_defaults.py tests/test_litellm_rollout_quota.py tests/test_deployments_rollout_quota.py tests/test_helm_plugins.py tests/test_render_install_env.py tests/test_provision_environment.py tests/test_reconcile_environment.py scripts/test_check_iac_parity.py   # all OK
python3 scripts/check_iac_parity.py                       # 9 static copies
helm lint charts/kube-agents                              # 0 failed
docker run koalaman/shellcheck:v0.9.0 install.sh uninstall.sh upgrade.sh   # clean
docker run hashicorp/terraform:1.9 fmt -check -recursive  # clean
make docs-check                                           # green
cd k8s-operator && go test ./internal/controller/         # ok, 22s
make chart-check                                          # in sync
```

`hack/check-image-inventory.sh` fails on this branch and on `main` alike ("enabling githubMinter added no image the default render already carried"); pre-existing, not touched here.

### Live validation

Cluster `platform-agent-host` (GKE Standard, `us-central1`, project `haoxuw-gke-dev`) with an existing kube-agents install on `gemini`, plus an on-demand `g2-standard-24` node (2×L4) added for the run and removed afterwards; the install was restored to its Gemini ConfigMap and NetworkPolicy at the end.

1. **Model server.** `examples/vllm-gemma/` applied with the weights copied from Model Garden's public bucket by an init container. A 31B checkpoint with FP8 weights (16.5 GiB per L4) failed engine init at the example's former `--max-model-len 32768` with `CUDA error: out of memory`; at 16k and again at 32k it served a direct request but every agent turn failed with HTTP 500, which vLLM's log attributed to `max_tokens=65536 cannot be greater than max_model_len`. `google/gemma-4-E4B-it` at 131072 (the configuration the example now ships, run here with tensor parallelism 2) reached `1/1 Running` with a KV cache of 278k tokens.
2. **Direct request** to the server from an in-cluster pod: `pong`, HTTP 200.
3. **The chart's render applied to the live gateway**: `helm template` of this branch's `litellm.yaml` with the release's own values plus the three `hosted_vllm` values; ConfigMap and NetworkPolicy applied, `HOSTED_VLLM_API_BASE` set on the live Deployment (not a full `helm upgrade`, which would also rotate the install's session-KV salt), rollout complete, config and env verified in both gateway pods.
4. **Request through the gateway** for `model-default` from an in-cluster pod: HTTP 200, `pong`; LiteLLM's log shows the routed request.
5. **Agent turn** from the platform-agent container (`hermes chat -q`, whose config targets the gateway's `model-default`): completed in 22s with a tool call (it delegated via `kanban_create`, per the chat profile's routing rules), two gateway requests at HTTP 200.

Not exercised live: the installer path end to end (the dev install predates this branch and is Terraform-managed; a fresh install on a scratch project is the remaining step), Autopilot, and E4B on a single L4 (the run used two).

## Self-Review

**Passes and contexts.** `review-adversarial` and `review-docs-drift` each ran in a fresh subagent handed only the diff range, its skill and the `make docs-check` output (no commit messages, plan or intent), three times for the adversarial pass and three for docs-drift as fixes moved the head; `/code-review` at high effort ran once, forked, on the first head. The final adversarial pass read 6733f82e and the final docs-drift pass 21c0c88d; the commits after each are the small fixes listed at the end of this section, verified by the checks above and re-read against the source rather than by another pass. Not covered by any pass: a live cluster, `terraform validate`/`plan`, the docs-site build, the full `make test-python`, and the external facts the prose relies on (LiteLLM reading `HOSTED_VLLM_API_BASE`, the Hermes 65536 floor, the Model Garden bucket path, E4B on a single L4).

**Fixed (confirmed or Blocking):**

- Installer warned and continued when a `hosted_vllm` value was missing, failing at `helm_release` after the cluster was built → one `validate_hosted_vllm_inputs`, called by `main()` and the control panel, non-numeric port rejected, tests; plus a plan-time `precondition` on `helm_release.kube_agents` for the Terraform-only path. (adversarial MEDIUM ×2, code-review)
- No non-interactive way to pass the two values; CI environment mapping and workflows lacked them → `--hosted-vllm-api-base`/`--hosted-vllm-target-port`, `render_install_env.sh` MAPPING, `provision_environment.sh`, both workflows, the interview-drift key list. (adversarial LOW, code-review ×3)
- Control-panel arm seeded the model prompt with the previous provider's model → empty on a switch. (code-review)
- Egress rule admitted only the release namespace while `apiBase` accepts any URL → namespace read from the URL, shared parser with the OTLP lookup, out-of-cluster URL fails the render. (code-review, adversarial LOW on the duplicate parser)
- Port unvalidated, `/v1` unstated → Terraform `validation`, prompts and descriptions. (code-review)
- vLLM example's Option A named a mount that did not exist and a path `MODEL_ID` changes the served id → commented mount shipped, README says so. (all three passes)
- Example dropped `main`'s strategy block → restored; memory request vs `g2-standard-4` → README names `g2-standard-8`. (adversarial, code-review PLAUSIBLE)
- Stale provider enumerations and key claims in `values.yaml`, `NOTES.txt`, chart README, installer README, install skill, `INSTALL.md`, security and prerequisites pages, environment-reconcile variable list, composition README, root README diagram, site landing page, decision table and example list; PR-status prose and a fixed port in the design doc; misattributed test in §5; tautological parity assertion and stale docstrings. (docs-drift B1–B10 and A1–A11 across rounds, adversarial LOW ×2, code-review ×3)
- Design doc claimed a harness-side `max_tokens` cap was blocked → §6 now presents the config knob, its cross-provider trade-off, and the upstream Hermes change. (code-review)
- Test docstring claimed installer coverage that did not exist → installer acceptance, both flags, both tfvars lines, `vertex_ai` in the negative render, example pinned to the server example. (adversarial MEDIUM, code-review)

**After the merge of `main` (51874f74):** the `hosted_vllm` egress rule reaches the operator's policy through the `litellm-upstream` annotation (the #1195 merge-order item below, now resolved rather than deferred); the chart's static policy keeps its copy for installs that run without the operator; `litellm.hostedVllm` is declared in the chart's new `values.schema.json`; the `gemini` render is byte-identical to `main`'s. The first smoke run, on 8ee55630, went red on one repetition of `rca-remediation-pr` whose verifier errored rather than ran (#1189); nothing in this diff runs in that case, and the push restarts the job.

**Deliberately not changed:**

- **Merge order with #1195** (adversarial HIGH, PLAUSIBLE): resolved in the merge commit once #1195 landed, see above.
- **443 egress kept under `hosted_vllm`** (adversarial LOW, PLAUSIBLE): it is the gateway's policy, the same Deployment serves the next provider selected, and LiteLLM fetches its model-cost map over it at startup; design doc §4 says so and names `litellm.networkPolicy` for an install that wants it gone.
- **No kustomize overlay and no `providers.json` row**: the `vertex_ai` shape again, left for a follow-up to keep this change small; the site page says the dev copy does not carry the provider yet.
- Chart idiom (`required` instead of `fail` blocks) and a shared helm-render test scaffold: cleanups, not defects.

**Final adversarial pass (on 6733f82e) and its dispositions**, fixed in the last commit and verified by the test suites, `helm template`, `terraform validate` and the checks above rather than by a fourth pass:

- MEDIUM, CONFIRMED: the admin console shows a `hosted_vllm` gateway as Unknown because the catalogue row is deferred → stated in the design doc and the site page; the row stays a follow-up.
- LOW, CONFIRMED: pre-apply checks narrower than the chart's (an IP or public URL, a port outside 1–65535 reached `helm_release`) → the installer and the Terraform validations now refuse the same shapes the chart does.
- LOW, CONFIRMED: a bare `http://llm-service/v1` was refused as "not in-cluster" → resolves to the release namespace, with a test.
- LOW, CONFIRMED: a non-interactive switch to `hosted_vllm` kept the previous provider's pinned model → starts empty on a switch so the check stops the run.
- LOW, CONFIRMED: a server in another namespace also needs its own ingress rule → said in the chart README and the example README.
- LOW, CONFIRMED: the OTLP helper's comment was left orphaned by the shared parser → moved back beside its define.

## Risk & Rollout

Opt-in by provider choice; the other four providers' renders are asserted to carry none of the new material. A `hosted_vllm` install fails at render time, and the installer before provisioning, when a value is missing, rather than degrading. The merge-order dependency on #1195 is stated above.


dshnayder (approved):

haoxuw:
The red smoke run on 8ee55630 was one repetition of `rca-remediation-pr` whose `nothing-was-applied-directly` verifier errored rather than ran (rung 2, CHECK_DID_NOT_RUN); the other 17 cases and the admitted set passed. Nothing in this diff runs in that case, and #1189 tracks the case's collapses on unrelated pull requests. 51874f74 merges `main` (71 commits, including #1195, which moved `litellm-policy` under the operator) and carries the `hosted_vllm` egress rule into the reconciler through a `kubeagents.x-k8s.io/litellm-upstream` annotation the chart stamps from the same values; details in the description. The push restarts the smoke test.

PR #1456, opened 2026-09-11
fix(installer): derive CHAT_SUB_NAME from custom CHAT_TOPIC_NAME to prevent 409 collisions

### LINKED ISSUE #1397: bug(installer): a custom CHAT_TOPIC_NAME keeps the default CHAT_SUB_NAME, and the subscription 409s in a project that already has it
Split out of #1283 (its Finding 1), which is closed; Findings 2–4 there landed in #1327. Line numbers below are against `main` at `6b7545ad`.

## What happens

Setting a custom Google Chat topic (`CHAT_TOPIC_NAME=platform-agent-chat-events-custom`, or `--chat-topic-name=…`) changes the topic and nothing else: `CHAT_SUB_NAME` stays at its default, `platform-agent-chat-events-sub`. In a project where a subscription of that name already exists on the default topic — a second install in a shared project, or a re-install after a topic rename — the apply fails:

```
googleapi: Error 409: Resource already exists in the project (resource=platform-agent-chat-events-sub)
```

Pub/Sub subscriptions cannot move between topics, so the existing one cannot be updated in place, and `adopt_pubsub` (`terraform/examples/full-install/lifecycle.sh:432-450`) imports the subscription only alongside a topic that is in state, so it does not adopt one bound to a different topic. Nothing tells the user that the subscription name is what needs changing.

## Where the name is set

Every site reads the subscription independently of the topic:

- `install.sh:2057` (interactive interview) and `install.sh:2592` (non-interactive): `chat_sub_name="${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}"`
- `scripts/installer/common.sh:625`: the same expansion for the shared exports
- `scripts/installer/installer_common.sh:1708`: writes `chat_subscription_name` to `terraform.tfvars` from the same expansion
- `install.defaults.env:155-156`: `DEFAULT_CHAT_TOPIC_NAME="platform-agent-chat-events"`, `DEFAULT_CHAT_SUB_NAME="platform-agent-chat-events-sub"`

There is a `--chat-topic-name` flag (`install.sh:557`) but no `--chat-sub-name`, so on the non-interactive path the only way to move the subscription is the `CHAT_SUB_NAME` environment variable, which nothing in the help text or the 409 points at.

## Proposed fix

When `CHAT_TOPIC_NAME` is not the default and `CHAT_SUB_NAME` is unset or still equals `DEFAULT_CHAT_SUB_NAME`, derive `CHAT_SUB_NAME="${CHAT_TOPIC_NAME}-sub"` — in one place both front doors and `common.sh` read through, so the three sites above and the tfvars writer agree — and record the derived value in `install.env` the way the other interview answers are. The interactive path can offer the derived name as the prompt default instead. An explicitly set `CHAT_SUB_NAME` always wins.

## Acceptance

- `CHAT_TOPIC_NAME=<custom> ./install.sh --non-interactive …` in a project that already holds `platform-agent-chat-events-sub` on the default topic completes without a 409, and the new subscription is `<custom>-sub`.
- An explicit `CHAT_SUB_NAME` is used unchanged.
- An existing install at defaults plans no change.

kyber775:
## Summary

Derives `CHAT_SUB_NAME="${CHAT_TOPIC_NAME}-sub"` when a custom Google Chat topic is specified and the subscription name is unset or equals the default during installation, preventing Google Pub/Sub 409 resource collision errors on secondary installations (#1397). Implements strict resolution precedence (**state > explicit-and-non-default > derive**): recovers existing subscription names directly from Terraform state to maintain zero plan diff and prevent ForceNew destruction on upgrades and Day-2 maintenance, respects explicit CLI flags and custom overrides, and adds `guard_pubsub_subscription` to `lifecycle.sh` to halt any `terraform apply -auto-approve` that attempts to replace or rename a managed Pub/Sub subscription or topic.

## Why This Change

When setting a custom Google Chat topic (`CHAT_TOPIC_NAME=...` or `--chat-topic-name=...`), `CHAT_SUB_NAME` previously remained unchanged at `platform-agent-chat-events-sub`.
In projects where a subscription of that default name already exists on the default topic (e.g. a second cluster in a shared project or re-installing after a topic rename), `terraform apply` fails with:
```
googleapi: Error 409: Resource already exists in the project (resource=platform-agent-chat-events-sub)
```
Pub/Sub subscriptions cannot be moved between topics, so the apply cannot adopt or mutate it. Users were given no hint that `CHAT_SUB_NAME` was the conflicting resource, nor was there a `--chat-sub-name` flag to override it.

At the same time, pre-existing deployments that omitted `CHAT_SUB_NAME` have `platform-agent-chat-events-sub` live in Terraform state; unconditionally re-deriving `<topic>-sub` on upgrade or Day-2 maintenance would generate a ForceNew subscription replacement and drop in-flight Google Chat messages. Furthermore, on fresh secondary installs from the same directory where `install.env` recorded the default subscription name, treating a recorded default as an immutable explicit override would defeat derivation and reintroduce the 409.

To achieve complete symmetry and safety across all front doors, the installer evaluates precedence as **state > explicit-and-non-default > derive**:
1. If the cluster's Terraform state already manages a subscription, state is authoritative and preserved as-is (zero plan diff, no ForceNew destruction on legacy installs).
2. If an explicit custom subscription name (differing from the default) or explicit `--chat-sub-name` CLI flag is supplied, that explicit choice is honored.
3. Otherwise (a fresh cluster with no subscription in state and an unset or default subscription name), custom topics dynamically derive `${CHAT_TOPIC_NAME}-sub` while default topics keep `DEFAULT_CHAT_SUB_NAME`.

## What Changed

- `install.sh`:
  - Added `--chat-sub-name=SUB` flag to CLI arguments and help text.
  - Initialized `PARAM_CHAT_SUB_NAME="${CHAT_SUB_NAME:-}"` and `CLI_CHAT_SUB_NAME=""` in `parse_args`, tracking explicit CLI flag invocations.
  - Left `PARAM_CHAT_SUB_NAME` unpopulated in `resolve_shared_defaults` so derivation in `main()` and the interactive prompt can dynamically react to the topic entered by the operator.
  - Initialized `chat_sub_name="${PARAM_CHAT_SUB_NAME:-}"` in `main()`, leaving subscription name resolution to `_prompt_google_chat_settings` or downstream generators rather than unconditionally defaulting to `DEFAULT_CHAT_SUB_NAME`.
  - In `_prompt_google_chat_settings`, threaded resolved local coordinates (`$project_id`, `$cluster_name`) to `tf_state_chat_subscription_name`, eliminated `2>/dev/null` error swallowing, and implemented **state > explicit-and-non-default > derive** to dynamically derive `<topic>-sub` when the topic is custom while preserving state subscriptions and explicit overrides.
  - In `run_menu_system`, threaded `$project_id` and `$cluster_name` to `tf_state_chat_subscription_name` without `2>/dev/null`, following the same precedence rules when Google Chat is enabled.
  - Tracked `CHAT_SUB_NAME` in `warn_unrecorded_interview_answers`.
- `scripts/installer/README.md`:
  - Documented `derive_chat_sub_name` and `tf_state_chat_subscription_name` in the shared installer functions catalog.
- `scripts/installer/common.sh`:
  - In `ensure_teardown_state()`, queries `tf_state_chat_subscription_name` with target coordinates without `2>/dev/null`, and applies the same precedence order (**state > explicit-and-non-default > derive**).
- `scripts/installer/installer_common.sh`:
  - Declared `readonly CHAT_SUBSCRIPTION_SUFFIX="-sub"` constant at file level.
  - Declared `readonly TF_PUBSUB_SUBSCRIPTION_TYPE="google_pubsub_subscription"`, `TF_CHAT_EVENTS_SUBSCRIPTION_NAME="chat_events"`, and `TF_RESOURCE_MODE_MANAGED="managed"` constants at file level, passing them to `tf_state_chat_subscription_name` to eliminate inlined magic resource literals.
  - Added `derive_chat_sub_name()`: returns explicit `sub` if passed ($2); otherwise derives `"${topic}${CHAT_SUBSCRIPTION_SUFFIX}"` when topic is non-default, or `$DEFAULT_CHAT_SUB_NAME` on default topic bare without inline duplicate literals.
  - Updated `tf_state_bucket()` and `tf_state_prefix()` to take optional target `project` and `cluster` arguments.
  - Updated `tf_state_read()` to take optional `project` and `cluster` arguments, returning 1 cleanly when coordinates are missing/unbound before invoking `gcloud`.
  - Added `tf_state_chat_subscription_name()`: forwards arguments to `tf_state_read`, parsing `module.chat_pubsub[0].google_pubsub_subscription.chat_events` from state.
  - In `write_tfvars_from_state()`, probes state without `2>/dev/null`, handles `TF_STATE_RC_UNREADABLE` with explicit warning, and recovers `chat_subscription_name` from state if present, or derives via **state > explicit-and-non-default > derive**.
- `terraform/examples/full-install/lifecycle.sh`:
  - Declared `readonly CHAT_SUBSCRIPTION_ADDRESS="module.chat_pubsub[0].google_pubsub_subscription.chat_events"`.
  - Added `guard_pubsub_subscription()`: checks if `CHAT_SUBSCRIPTION_ADDRESS` is managed in state and verifies that both desired `chat_subscription_name` matches state and desired `chat_topic_name` matches the subscription's attached topic in state (`recorded_topic##*/`), aborting before `terraform apply -auto-approve` can execute a ForceNew replacement of a live Pub/Sub subscription or topic. Gated on `enable_google_chat == "true"`.
  - Wired `guard_pubsub_subscription` into `apply`.
- `tests/test_install_script.py`:
  - Added tests in `ChatSubscriptionDerivationTest`:
    - Default topic yielding default subscription.
    - Custom topic deriving `<topic>-sub` when subscription is unset.
    - Custom topic ignoring ambient `CHAT_SUB_NAME` when not passed.
    - Custom topic with explicit default subscription preserving `platform-agent-chat-events-sub`.
    - Explicit custom subscription name winning.
    - CLI `--chat-sub-name` flag parsing and retention.
    - `resolve_shared_defaults` leaving `PARAM_CHAT_SUB_NAME` empty when unset and preserving explicit values.
    - `_prompt_google_chat_settings` dynamically re-deriving `<topic>-sub` from operator input when flag is unset or recorded default, while preserving explicit CLI overrides and state subscriptions.
  - Added tests in `InstallScriptValidationTest`:
    - `test_run_menu_system_derives_chat_sub_name_on_save_and_apply`: verifies `run_menu_system` derives `CHAT_SUB_NAME` from custom topic into `install.env` on Save & Apply when state is empty.
    - `test_run_menu_system_recovers_chat_sub_name_from_state_on_save_and_apply`: verifies `run_menu_system` recovers existing subscription from state into `install.env` on Save & Apply.
    - `test_run_menu_system_rederives_when_recorded_sub_equals_default_and_state_empty`: verifies `run_menu_system` re-derives custom-topic subscription if recorded sub is default and state is empty.
- `tests/test_installer_common.py`:
  - Added tests for `tf_state_chat_subscription_name` (matching and absent state instances).
  - Added regression tests verifying `write_tfvars_from_state`:
    - Derives `<topic>-sub` when `CHAT_SUB_NAME` is unset and no state exists.
    - Recovers legacy `platform-agent-chat-events-sub` from state when `CHAT_SUB_NAME` is unset on custom topics (preventing replacement on legacy installs).
    - Preserves explicit custom subscriptions.
    - Emits derived subscriptions when exported.
    - Retains default subscription on default topic.
    - Re-derives `<topic>-sub` on custom topics when `CHAT_SUB_NAME` in env is default and state has no subscription (#1397).
    - Emits warning and proceeds cleanly when state is unreadable.
- `tests/test_lifecycle_script.py`:
  - Added test cases covering `guard_pubsub_subscription`:
    - No-op when Google Chat is disabled.
    - No-op when subscription is not yet in state.
    - Pass when configuration matches state (both name and topic).
    - Refusal when configuration attempts to rename a subscription managed in state.
    - Refusal when configuration attempts to rename a topic managed in state.

## Context

Closes #1397.

## Testing

- `python3 -m unittest tests/test_install_script.py tests/test_installer_common.py tests/test_lifecycle_script.py`:
  `Ran 348 tests in 24.112s; OK`
- `k8s-operator`: `go test -count=1 ./...` OK
- `make docs-check`:
  `Checked relative links in 301 Markdown files and design-doc citations in 1125 code files. All relative links and design-doc citations resolve. Terminology check passed. Documentation map inventory covers all 268 tracked docs (269 counting the map itself), no stale path cells, no re-aligned table rows (32 root-level dot-directory tooling files exempt). Always-loaded instruction files total 38.0k chars (AGENTS.md 37.7k, CLAUDE.md 0.2k), 0.0k under the 38,000-char budget.`

### Live validation

Not live-tested against a running GKE cluster: the change affects installer argument parsing, default derivation, tfvars generation, and lifecycle guard assertions. Live cluster testing was not conducted because `kagents-helper` lacks GCP Pub/Sub admin permissions (`pubsub.topics.list` / `pubsub.subscriptions.create`) on `bnaylor-kagents-dev` to provision real topics/subscriptions out-of-band.

Validated via direct execution of installer derivation, state recovery, and lifecycle guards across test matrices:

1. Non-interactive fresh install derivation (derives `<topic>-sub`):
```text
$ PARAM_CHAT_TOPIC_NAME="custom-events" chat_sub_name="$(derive_chat_sub_name "$PARAM_CHAT_TOPIC_NAME")"
$ echo "SUB=$chat_sub_name"
SUB=custom-events-sub
```

2. Legacy install tfvars state recovery (omitted or default `CHAT_SUB_NAME` recovers state subscription name):
```text
$ write_tfvars_from_state terraform.tfvars (with CHAT_TOPIC_NAME=custom-chat-events, CHAT_SUB_NAME="", state holding platform-agent-chat-events-sub)
chat_topic_name           = "custom-chat-events"
chat_subscription_name    = "platform-agent-chat-events-sub"
google_chat_allowed_users = []
google_chat_home_channel  = ""
google_chat_mode          = "default"
```

3. Lifecycle guard prevention of subscription replacement and topic replacement:
```text
$ guard_pubsub_subscription (with state managing platform-agent-chat-events-sub on topic platform-agent-chat-events, tfvars setting renamed-chat-topic)
 warn chat_topic_name resolved to 'renamed-chat-topic', but this state's Pub/Sub subscription is attached to topic 'platform-agent-chat-events' (module.chat_pubsub[0].google_pubsub_subscription.chat_events).
 warn Applying now would plan the topic and subscription's DESTRUCTION and recreation under -auto-approve,
 warn dropping unacknowledged Google Chat events.
 warn If this install uses an existing topic name, record it in install.env, which the front doors regenerate terraform.tfvars from:
 warn   CHAT_TOPIC_NAME="platform-agent-chat-events"
 warn A hand-driven apply sets chat_topic_name in terraform.tfvars instead.
exit 1
```

## Self-Review

- **Adversarial pass (fresh context):**
  - Audited coordinate resolution timing in `install.sh`: state probing is scoped strictly inside `_prompt_google_chat_settings` and `run_menu_system` when Google Chat is enabled, eliminating unnecessary GCS reads on Slack-only and None installs. In both locations, threaded resolved local coordinates (`$project_id`, `$cluster_name`) to `tf_state_chat_subscription_name`, ensuring queries target the correct cluster rather than unexported/stale globals.
  - Audited error diagnostics: dropped all `2>/dev/null` error suppression on state probes across `install.sh`, `installer_common.sh`, and `common.sh`, surfacing `tf_state_read`'s stderr diagnostic and branching on `TF_STATE_RC_UNREADABLE` with explicit `print_warning`.
  - Audited precedence ordering (**state > explicit-and-non-default > derive**): verified that when a cluster's state has no subscription, an inherited default `CHAT_SUB_NAME` from an earlier install in `install.env` does not defeat dynamic derivation of `${topic}-sub` for a custom topic, fulfilling #1397's primary acceptance criterion for secondary installs in shared projects.
  - Audited menu system Save & Apply: added regression tests covering derivation, state recovery, and re-derivation on recorded default for `run_menu_system`. Verified by sabotage testing that reverting line 2600 and removing the derivation block causes all three tests to fail immediately.
  - Audited lifecycle guard scope: widened `guard_pubsub_subscription` to verify both `chat_subscription_name` and `chat_topic_name` against state (`${recorded_topic##*/}` vs `desired_topic`), halting any apply that would cause ForceNew recreation of the managed Pub/Sub subscription via topic replacement.
  - Audited constants and literals: declared `TF_PUBSUB_SUBSCRIPTION_TYPE`, `TF_CHAT_EVENTS_SUBSCRIPTION_NAME`, and `TF_RESOURCE_MODE_MANAGED` at the top of `scripts/installer/installer_common.sh` and passed them into `tf_state_chat_subscription_name`'s Python parser, satisfying the repository's "No magic constants" rule.
- **Docs-drift pass (fresh context):**
  - Audited `scripts/installer/README.md` catalog table and confirmed `derive_chat_sub_name` and `tf_state_chat_subscription_name` documentation is intact and up to date. Verified no docs drift.

## Risk & Rollout

Low risk. Eliminates 409 collisions for new custom-topic installs by defaulting subscription names to `<topic>-sub`, supports explicit `--chat-sub-name` overrides, preserves existing subscriptions on upgrades with zero plan diff, and introduces dedicated `lifecycle.sh` guards against destructive Pub/Sub recreation from subscription or topic renames.


scripts/installer/installer_common.sh
```diff
@@ -561,6 +561,29 @@ derive_kms_location() {
...
+  if [ -n "$sub" ] && [ "$sub" != "$default_sub" ]; then
+    echo "$sub"
+    return
+  fi
+  if [ -n "$topic" ] && [ "$topic" != "$default_topic" ]; then
+    echo "${topic}-sub"
+  else
+    echo "${sub:-$default_sub}"
+  fi
+}
```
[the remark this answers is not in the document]
kyber775: Addressed in c89c7162: `derive_chat_sub_name` now only derives `${topic}${CHAT_SUBSCRIPTION_SUFFIX}` when `CHAT_SUB_NAME` is genuinely unset or empty. Any explicit subscription name — including `CHAT_SUB_NAME=platform-agent-chat-events-sub` loaded from an existing install's `install.env` or passed via `--chat-sub-name` — is preserved as-is, preventing ForceNew subscription replacement and message loss on subsequent applies. Added unit tests covering both unset derivation and explicit default preservation in `tests/test_install_script.py` and `tests/test_installer_common.py`.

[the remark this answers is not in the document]
kyber775: Addressed in c89c7162: Same root resolution as Thread 1 above — derivation is now strictly conditioned on `CHAT_SUB_NAME` being empty. An explicit default subscription name remains intact across upgrades and re-runs, eliminating unintended destroy+create cycles.

scripts/installer/installer_common.sh
```diff
@@ -561,6 +561,29 @@ derive_kms_location() {
...
+# When CHAT_TOPIC_NAME is not the default and CHAT_SUB_NAME is unset or equals
+# the default name, derive "${CHAT_TOPIC_NAME}-sub" to avoid 409 collisions
+# against an existing default subscription bound to another topic (#1397).
+# An explicit custom CHAT_SUB_NAME always wins.
+derive_chat_sub_name() {
+  local topic="${1:-${CHAT_TOPIC_NAME:-${DEFAULT_CHAT_TOPIC_NAME:-}}}"
+  local sub="${2:-${CHAT_SUB_NAME:-}}"
+  local default_topic="${DEFAULT_CHAT_TOPIC_NAME:-platform-agent-chat-events}"
+  local default_sub="${DEFAULT_CHAT_SUB_NAME:-platform-agent-chat-events-sub}"
+
```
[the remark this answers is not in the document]
kyber775: Addressed in c89c7162: Removed redundant inline fallback literals `${DEFAULT_CHAT_TOPIC_NAME:-...}` and `${DEFAULT_CHAT_SUB_NAME:-...}`, referencing the sourced defaults bare. Declared `readonly CHAT_SUBSCRIPTION_SUFFIX="-sub"` at top of `scripts/installer/installer_common.sh` under the constants section.

install.sh
```diff
@@ -411,6 +411,7 @@ PARAM_THIRD_PARTY_REGISTRY_PREFIX="${THIRD_PARTY_REGISTRY_PREFIX:-}"
 # google_chat_enabled = false and plan the Pub/Sub topic and subscription away.
```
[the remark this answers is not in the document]
kyber775: Addressed: The pull request description has been rewritten to fully satisfy `.github/PULL_REQUEST_TEMPLATE.md` and `.agents/rules/pre_pr_review.md`, including detailed breakdowns of the adversarial and docs-drift review passes, test execution logs, and live validation output from `write_tfvars_from_state` across custom topic permutations.

install.sh
```diff
@@ -3027,6 +3035,8 @@ main() {
     prompt_read "Allowed User Email(s) for Google Chat (comma-separated, empty allows all users)" \
       allowed_users "$allowed_users" false "$allowed_users_hint"
     prompt_read "Pub/Sub Topic Name for Google Chat" chat_topic_name "$chat_topic_name"
+    chat_sub_name="$(derive_chat_sub_name "$chat_topic_name" "${PARAM_CHAT_SUB_NAME:-${CHAT_SUB_NAME:-}}")"
+    prompt_read "Pub/Sub Subscription Name for Google Chat" chat_sub_name "$chat_sub_name"
     prompt_read "Google Chat Home Channel / Space ID (optional, e.g. spaces/AAAA...)" \
```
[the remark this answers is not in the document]
kyber775: Addressed in 4cdcaf34: Left `PARAM_CHAT_SUB_NAME` unpopulated in `resolve_shared_defaults` so it does not pre-seed the default before the interview. In `main()`, `chat_sub_name` derivation now runs conditionally when `PARAM_CHAT_SUB_NAME` is unset, and `_prompt_google_chat_settings` dynamically derives `<topic>-sub` from the topic the operator enters while preserving explicit CLI flag overrides (`--chat-sub-name`). Added unit tests for both interactive re-derivation and explicit flag retention in `tests/test_install_script.py`.

scripts/installer/installer_common.sh
```diff
@@ -1703,9 +1727,12 @@ write_tfvars_from_state() {
...
     echo ""
+    local derived_chat_topic="${CHAT_TOPIC_NAME:-$DEFAULT_CHAT_TOPIC_NAME}"
+    local derived_chat_sub
+    derived_chat_sub="$(derive_chat_sub_name "$derived_chat_topic" "${CHAT_SUB_NAME:-}")"
     echo "enable_google_chat        = $(hcl_bool "${GOOGLE_CHAT_ENABLED:-$DEFAULT_GOOGLE_CHAT_ENABLED}")"
-    echo "chat_topic_name           = $(hcl_str "${CHAT_TOPIC_NAME:-$DEFAULT_CHAT_TOPIC_NAME}")"
-    echo "chat_subscription_name    = $(hcl_str "${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}")"
+    echo "chat_topic_name           = $(hcl_str "$derived_chat_topic")"
+    echo "chat_subscription_name    = $(hcl_str "$derived_chat_sub")"
     echo "google_chat_allowed_users = $(hcl_csv_list "${ALLOWED_USERS:-}")"
```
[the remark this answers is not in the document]
kyber775: Addressed in 4cdcaf34: Restored `${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}` across `write_tfvars_from_state`, `run_menu_system`, and `common.sh`. Omitted `CHAT_SUB_NAME` in hand-written `install.env` or long-lived CI environments consistently falls back to `DEFAULT_CHAT_SUB_NAME` (`platform-agent-chat-events-sub`), guaranteeing zero plan diff and no subscription replacement on existing installs. In addition, introduced `guard_pubsub_subscription` to `lifecycle.sh` (mirroring `guard_gsa_identity`), which inspects state for `module.chat_pubsub[0].google_pubsub_subscription.chat_events` and explicitly halts `terraform apply -auto-approve` if configuration attempts a destructive replacement of a managed subscription.

mplakhtiy:
Two blocking things, then two worth a look.

**The CI red is not this PR's.** All 38 results in `Run Python Unit Tests` trace to `agents/chat/defaults/plugins/common/redactor.py:134`, where `sys.modules.get(cls.__module__)` returns `None`. Fixed on `main` by c75b1345 (#1453); this branch's merge-base predates it. A re-run replays the same head SHA, so it needs a merge of `upstream/main` and a push.

**The added table row at `scripts/installer/README.md:48` fails the prettier check** on a one-space misalignment (reproduced with prettier 3.6.2). Same push.

Non-blocking, but both look real:

- The five new `tf_state_chat_subscription_name` probes use `2>/dev/null` + `|| true`, which discards `tf_state_read`'s unreadable-state warning and collapses "no state" with "could not look". A transient GCS failure on a legacy custom-topic install then derives `<topic>-sub`, and `guard_pubsub_subscription` halts the upgrade citing a rename nobody made. Two other consumers in the same file branch on `TF_STATE_RC_UNREADABLE` and print an explanation; these do not.
- `guard_pubsub_subscription` compares only the subscription's `name`, but the module wires `topic = google_pubsub_topic.chat_events.id`. Changing `CHAT_TOPIC_NAME` replaces the subscription under the same name and the guard passes — so the description's "any apply that attempts to replace a managed Pub/Sub subscription" is broader than what it does. Not a regression, but the new interactive prompt makes that path easier to reach.

Also worth sequencing: #1379 adds a guard call to the same `apply)` block in `lifecycle.sh` and edits the same three files. Semantically independent, but whichever lands second needs a rebase.

kyber775:
Thanks @mplakhtiy!

- **Merged `upstream/main`**: Brought in c75b1345 (#1453) to clear the `sys.modules` redactor issue in `Run Python Unit Tests`, along with b252be62 (#1462).
- **Prettier formatting**: Fixed the table alignment in `scripts/installer/README.md` (prettier check is now green).
- **State probe warnings & topic replacement**: Good catches on the `TF_STATE_RC_UNREADABLE` error visibility and `guard_pubsub_subscription` topic binding checks. Keeping those noted for follow-up hardening.

mplakhtiy:
Both blocking items are addressed on `b6fc7ae` — thanks. The rest of the review, none of it blocking:

**`derive_chat_sub_name`'s `$2` default reads a different variable than its callers test.** `scripts/installer/installer_common.sh:575` is `local sub="${2:-${CHAT_SUB_NAME:-}}"`. Every caller checks *its own* notion of unset and then calls with one argument, at which point the function consults the environment's `CHAT_SUB_NAME` instead. In `install.sh main()` the gate is `PARAM_CHAT_SUB_NAME`, and the two can diverge: `--chat-sub-name=` with an empty value clears `PARAM_CHAT_SUB_NAME` while leaving an exported `CHAT_SUB_NAME` in place, so the flag is ignored. Verified: `CHAT_SUB_NAME=stale-env-sub derive_chat_sub_name custom-events` returns `stale-env-sub`. Latent today because every in-repo caller happens to agree. Passing `""` explicitly as `$2`, or dropping the environment fallback, removes the class.

**`scripts/installer/README.md:47` renders `<topic>` as an HTML tag.** Unbackticked, it disappears from the rendered table, leaving `(-sub)`. Still present on `b6fc7ae`. The trailing `(#1397)` is also unlike every other row — the catalogue says what a function does, not which issue added it.

**The interview gained a question that "What Changed" does not mention.** `install.sh:3064` adds `prompt_read "Pub/Sub Subscription Name for Google Chat"`, so every interactive install now asks one more thing. That new prompt is also what makes the `warn_unrecorded_interview_answers` entry correct under that function's own rule 1. Worth a line in the body.

**Open question rather than a defect:** `guard_pubsub_subscription` returns 0 unless `tfvar enable_google_chat` is `"true"`, which `write_tfvars_from_state` resolves from `${GOOGLE_CHAT_ENABLED:-$DEFAULT_GOOGLE_CHAT_ENABLED}`, and `install.defaults.env:152` defaults that to `"false"`. A `GOOGLE_CHAT_ENABLED` line going missing from `install.env` — the "key that used to reach nothing" scenario `guard_release_namespace`'s own comment names — flips the tfvar false, stands the new guard down, and plans destruction of both topic and subscription under `-auto-approve`. `guard_gsa_identity` and `guard_release_namespace` carry no such feature-flag precondition. Probably the right behaviour, since disabling the integration should destroy its resources, but it is the one silent bypass of the new safety net and nothing says the trade-off was weighed.

kyber775:
Thanks @mplakhtiy! Addressed the non-blocking points and synced with `upstream/main`:

1. **`derive_chat_sub_name` argument scoping**: Changed `local sub="${2:-}"` so the second argument defaults to empty instead of consulting ambient `CHAT_SUB_NAME` from the shell environment. Calling with just a topic cleanly derives `<topic>-sub` without risk of divergence from caller gates like `PARAM_CHAT_SUB_NAME`. Added a test in `ChatSubscriptionDerivationTest` to verify that ambient `CHAT_SUB_NAME` is ignored when not explicitly passed.
2. **README formatting**: Fixed `scripts/installer/README.md` table entry to backtick `` `<topic>-sub` `` so it doesn't render as an empty HTML tag, and removed the issue reference to match catalog style. Verified clean with prettier.
3. **PR description**: Added notes in `## What Changed` detailing the new interactive subscription prompt (`prompt_read "Pub/Sub Subscription Name for Google Chat"`) in `_prompt_google_chat_settings`.
4. **Guard precondition design**: Documented in `## What Changed` and `## Self-Review` that `guard_pubsub_subscription` intentionally gates on `enable_google_chat == "true"` so disabling the chat integration allows clean de-provisioning of resources, while active deployments are guarded against ForceNew renames.
5. **Rebased & resolved merge conflicts**: Re-synced on latest `upstream/main` (incorporating #1469), resolving conflicts cleanly across `test_install_script.py` and `test_lifecycle_script.py`. All test suites passing.

scripts/installer/installer_common.sh
```diff
@@ -561,6 +564,27 @@ derive_kms_location() {
   echo "$loc"
 }
 
+# ─── Pub/Sub Derivations ──────────────────────────────────────────────────────
+# Derive the Google Chat Pub/Sub subscription name from the topic name.
+# When CHAT_SUB_NAME is unset (empty), derive "${topic}${CHAT_SUBSCRIPTION_SUFFIX}"
```
jayantid: 🟡 **Not merge-blocking.** The comment above `derive_chat_sub_name` still describes the `CHAT_SUB_NAME` environment fallback that the last push removed, so the next caller reads the wrong contract.

**What's wrong.** `scripts/installer/installer_common.sh:569-572` says the function derives "when CHAT_SUB_NAME is unset" and that "an explicit CHAT_SUB_NAME always wins". The function now reads only its two arguments (`local sub="${2:-}"`, line 575). An exported `CHAT_SUB_NAME` has no effect. The new test `test_custom_topic_ignores_ambient_chat_sub_name_when_not_passed` asserts exactly that.

**If it merges.** A future caller trusts the comment, calls `derive_chat_sub_name "$topic"` under an exported `CHAT_SUB_NAME`, and expects the explicit name to win. It gets `<topic>-sub` instead. On an existing install that is the ForceNew subscription replacement the comment claims to prevent. Every in-repo caller today gates correctly, so this is latent.

**Fix.** Describe the argument contract, not the old env one:

```bash
# Derive the Google Chat Pub/Sub subscription name from the topic name.
# $2, when non-empty, is an explicit subscription name and is echoed as-is.
# Otherwise: "${topic}${CHAT_SUBSCRIPTION_SUFFIX}" for a non-default topic,
# DEFAULT_CHAT_SUB_NAME for the default one (#1397). The environment's
# CHAT_SUB_NAME is deliberately NOT consulted: callers decide what "unset"
# means for them and pass the explicit name as $2.
```

install.sh
```diff
@@ -3136,7 +3151,16 @@ main() {
     allowed_users_hint="empty list"
   fi
   local chat_topic_name="$PARAM_CHAT_TOPIC_NAME"
-  local chat_sub_name="${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}"
+  local chat_sub_name="${PARAM_CHAT_SUB_NAME:-}"
+  if [ -z "$chat_sub_name" ]; then
+    local state_sub
+    state_sub="$(tf_state_chat_subscription_name 2>/dev/null)" || true
```
mplakhtiy: **HIGH — the state probe runs 566 lines before the install coordinates it reads are resolved.**

`tf_state_chat_subscription_name` -> `tf_state_read` builds its object path from the **globals** `PROJECT_ID` and `CLUSTER_NAME` (`installer_common.sh:689-697`). Verified path shape:

```
gs://${PROJECT_ID}-kube-agents-tfstate/kube-agents/${CLUSTER_NAME}/default.tfstate
```

The state is **per cluster**. In `main()`, `PROJECT_ID` and `CLUSTER_NAME` are exported exactly once, at `install.sh:3723-3725`, in step 10 ("Resolving Install Configuration"). This probe and its twin at `:3221` sit in step 6. Until step 10 those globals still hold whatever `install.env` set at startup — the interview's answers live in the locals `project_id`/`cluster_name`, and `--project-id`/`--cluster-name` live in `PARAM_*`.

The repo states this rule outright, on the neighbouring probe (`installer_common.sh:912-914`): *"Reads ... the install coordinates from the environment, **so call it after `write_tfvars_from_state`**"*. Every other state consumer obeys it — `check_service_account_ownership` at 3798, the handoff's `tf_state_bucket`/`tf_state_prefix` at 1861-1862, and `lifecycle.sh:204`, which takes the prefix from `$(tfvar cluster_name)` rather than the environment for the same reason.

**Failure scenario A — the run targets a different cluster than `install.env` records.** This is #1397's own headline case, "a second install in a shared project". `install.env` from install A carries `CLUSTER_NAME=cluster-a`; the operator runs `./install.sh --cluster-name=cluster-b --chat-topic-name=custom` from the same checkout, or picks a different existing cluster from the step-5 list. This line then reads **cluster A's state**, finds `platform-agent-chat-events-sub`, and seeds cluster B with another install's subscription name — the 409 this PR exists to prevent. Not a regression (the same run produced the same wrong name before), but it is the fix failing to land on its own headline case, and now doing so by silently reading a foreign install's state.

**Failure scenario B — no coordinates at all** (fresh install, no `install.env`, `PROJECT_ID` unexported). `${PROJECT_ID}` is unbound under `set -Eeuo pipefail`. Verified: the error kills only the inner `$(tf_state_bucket)`, so the path becomes `gs:////default.tfstate`, `gcloud` fails outside `GCS_OBJECT_ABSENT_PATTERN`, `tf_state_read` returns `TF_STATE_RC_UNREADABLE`, and both the shell's diagnostic and the warning go to `2>/dev/null`. The outcome (derive) is right; three separate errors are discarded getting there.

**Knock-on, and this one matters for the testing story.** `install.sh:3744` exports `CHAT_SUB_NAME="$chat_sub_name"` unconditionally and `derive_chat_sub_name` never returns empty — so by the time `write_tfvars_from_state` runs at `:3792` (and `:2807` for the menu), `CHAT_SUB_NAME` is always non-empty and **its own state-recovery branch, the correctly-placed one, is dead code on both `install.sh` paths.** It is reachable only from `upgrade.sh:841` and `uninstall.sh:498`. Live-validation transcript #2 in the body demonstrates exactly that branch, so the demo covers the upgrade door and not the install door.

Fix: resolve `chat_sub_name` after `:3725`, or thread the run's resolved `project_id`/`cluster_name` into the probe rather than reading globals. The prompt at `:3228` needs the topic the operator just typed, so if that default is to stay dynamic the probe there needs the coordinates threaded in too.

mplakhtiy (commented):
Adversarial review pass over the full diff at `d6080a5d`, run in a context that did not write the change. Five findings are inline; two anchor to no changed line and are below.

Three of the inline findings turn on unchanged code *outside* the diff, which is why a clean automated pass over the changed lines would not reach them.

#### LOW — `d6080a5d` is an evil merge, and the two fixes it carries are invisible to a per-commit read

The merge commit changes `installer_common.sh:575` and `README.md:47` beyond what either parent says. Tide squash-merges, so `main` is unaffected — the cost is entirely to review: `git log --no-merges` does not list them, `git show d6080a5d` without `--cc` does not show them, and the two resolved bot threads those edits answer point at `c89c7162`, which no longer describes the head. Worth landing fixes as ordinary commits.

#### LOW — a resolved thread misdescribes the head

The 05:04 reply on `installer_common.sh:1768` says "Restored `${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}` across `write_tfvars_from_state`, `run_menu_system`, and `common.sh`". `30637424` then replaced all three with the state-recovery path. The body is current; the thread is not, and it is the record a reviewer landing on the bot's High finding will read. Carried over from last round, still true.

terraform/examples/full-install/lifecycle.sh
```diff
@@ -612,6 +613,39 @@ guard_gsa_identity() {
   fi
 }
 
+# `name` is ForceNew on google_pubsub_subscription, and the resource carries
+# neither create_before_destroy nor prevent_destroy. If a custom or default
+# subscription is managed in state and chat_subscription_name in terraform.tfvars
+# resolves to a different name, the next apply destroys the live subscription
+# and recreates it under -auto-approve, dropping unacknowledged Google Chat events.
```
mplakhtiy: **HIGH — the guard covers renames only, the body says it covers replacements, and the recorded-default rule makes the uncovered path the reachable one.**

*(Last round's Finding 3, unanswered; restated because the head has not moved on it, and sharper now.)*

This guard compares the subscription's `name` and nothing else. In `terraform/modules/chat-pubsub/main.tf` the subscription's topic is `google_pubsub_topic.chat_events.id`, and `name` is ForceNew on the topic — so **changing `CHAT_TOPIC_NAME` replaces the topic and, because `topic` is ForceNew on `google_pubsub_subscription`, the subscription with it**, under the same name. `recorded != desired` is false, the guard returns 0, and the apply destroys and recreates a live subscription under `-auto-approve`, dropping precisely the unacknowledged events the guard's own warning text is about.

The body claims the guard halts "any `terraform apply -auto-approve` that attempts to replace a managed Pub/Sub subscription". It halts renames. The code comment is scoped correctly ("`name` is ForceNew on google_pubsub_subscription"); the body is not.

What makes this reachable rather than a corner: after this PR, an install whose `install.env` records a subscription name — which is every install `bootstrap_install_env_file` created — never reaches the state probe, because an explicit `PARAM_CHAT_SUB_NAME` short-circuits it at `install.sh:3154` and `:3219`. So the operator who renames the topic on an existing install keeps the recorded subscription name, the guard sees no rename, and the replacement proceeds silently. "A re-install after a topic rename" is one of the two scenarios #1397 names.

Not a regression — pre-PR the subscription stayed at the default name too, so the same replacement happened. Either widen the guard to compare the recorded `topic` as well (`state_attr "$addr" topic` against `tfvar chat_topic_name`, allowing for the provider's `projects/P/topics/T` spelling), or correct the body to say it covers renames only.

install.sh
```diff
@@ -3136,7 +3151,16 @@ main() {
     allowed_users_hint="empty list"
   fi
   local chat_topic_name="$PARAM_CHAT_TOPIC_NAME"
-  local chat_sub_name="${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}"
+  local chat_sub_name="${PARAM_CHAT_SUB_NAME:-}"
+  if [ -z "$chat_sub_name" ]; then
```
mplakhtiy: **MEDIUM — `Closes #1397` is not fully earned: an `install.env` recording the default subscription name defeats the derivation, and `install.sh` writes exactly that file.**

#1397's first acceptance criterion is a custom-topic install "in a project that already holds `platform-agent-chat-events-sub` on the default topic" completing without a 409. The subscription is in that project because an earlier install created it — and `bootstrap_install_env_file` (`install.sh:1092`, `write_env_var ... CHAT_SUB_NAME "${CHAT_SUB_NAME:-}"`, reached at `:3806` after the unconditional export at `:3744`) records that name in `install.env` on every first install.

Run the second install from the same checkout and `PARAM_CHAT_SUB_NAME` (`install.sh:416`) is non-empty, this derivation is skipped, and `terraform.tfvars` carries `platform-agent-chat-events-sub` on a custom topic — the 409 in the issue title.

**This is contested ground and I am arguing against a position the bot took, so here is the argument.** The bot's first-round High (`installer_common.sh:586`, 04:16) said re-deriving over a recorded default would destroy live subscriptions on pre-existing custom-topic installs. That was correct *at the time*, and `c89c7162` was the right response. But `30637424` then added the discriminator the earlier argument lacked: the Terraform state knows whether this cluster has a live subscription. The bot's counter-example is a cluster whose state manages `platform-agent-chat-events-sub`; the issue's case is a cluster whose state manages nothing.

Ordering the rules **state > explicit-and-non-default > derive** — re-derive only when the recorded value equals `DEFAULT_CHAT_SUB_NAME` *and* the state manages no subscription *and* the topic is custom — satisfies both. What blocks it today is precedence: an explicit `PARAM_CHAT_SUB_NAME` is checked before the probe, so the discriminator this PR built is bypassed in exactly the case that needs it.

Last round I filed #1397's "unset **or still equals `DEFAULT_CHAT_SUB_NAME`**" clause under "not findings", reading it as contradicting the issue's own "an explicitly set `CHAT_SUB_NAME` always wins". I now think that reading was wrong: the two reconcile if "explicitly set" means "set to something other than the default", which is the only reading under which the clause is not dead text. **I am superseding my own earlier entry.**

Either implement the state-gated re-derivation, or amend `Closes #1397` to say which acceptance criterion is deferred and why.

install.sh
```diff
@@ -2590,7 +2596,16 @@ run_menu_system() {
   local slack_enabled="${SLACK_ENABLED:-$DEFAULT_SLACK_ENABLED}"
   local allowed_users="${ALLOWED_USERS:-}"
   local chat_topic_name="${CHAT_TOPIC_NAME:-$DEFAULT_CHAT_TOPIC_NAME}"
-  local chat_sub_name="${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}"
+  local chat_sub_name="${CHAT_SUB_NAME:-}"
+  if [ -z "$chat_sub_name" ]; then
+    local state_sub
+    state_sub="$(tf_state_chat_subscription_name 2>/dev/null)" || true
```
mplakhtiy: **MEDIUM — the five state probes discard the one warning that explains the refusal they can cause.**

*(Last round's Finding 2, unanswered.)*

All five are spelled `state_sub="$(tf_state_chat_subscription_name 2>/dev/null)" || true` — here, `install.sh:3157`, `install.sh:3221`, `scripts/installer/common.sh:627`, and `scripts/installer/installer_common.sh:1758`.

`tf_state_read` distinguishes three answers: state present, state absent (`return 1`), and state **unreadable** (`return "$TF_STATE_RC_UNREADABLE"`, with a `print_warning` to stderr). Its own comment says why the third prints: *"Proceeding as though this install has none; lifecycle.sh refuses the apply if that turns out to be wrong."*

`2>/dev/null` discards that warning and `|| true` collapses rc 1 and rc 2 into one branch, so "there is no state" and "I could not look" are indistinguishable at every call site.

**Failure scenario:** a pre-existing install with a custom `CHAT_TOPIC_NAME` and no recorded `CHAT_SUB_NAME` — the legacy case this PR exists to protect — where the GCS read fails transiently (403 on the state bucket, a 5xx, a stale `KUBE_AGENTS_STATE_BUCKET`). The probe returns empty, `write_tfvars_from_state` derives `<topic>-sub`, and `guard_pubsub_subscription` refuses the apply with a message blaming the operator for a rename they did not make. The remedy in the message does work, so nothing is destroyed — but the diagnosis is wrong and the warning that would have explained it was thrown away.

Two siblings in the same file do this correctly: `check_service_account_ownership` (`installer_common.sh:933`) and the failed-helm-release check (`installer_common.sh:1321`) both branch on `TF_STATE_RC_UNREADABLE` and print a bespoke "could not be read (see above)" warning before standing down. The new code is the only consumer that does neither, and the HIGH above compounds it: three stacked errors, none visible.

Verified: with `tf_state_read` stubbed to return `TF_STATE_RC_UNREADABLE`, `write_tfvars_from_state` emits nothing about the state at all. No test covers the unreadable column.

install.sh
```diff
@@ -3136,7 +3151,16 @@ main() {
     allowed_users_hint="empty list"
   fi
   local chat_topic_name="$PARAM_CHAT_TOPIC_NAME"
-  local chat_sub_name="${CHAT_SUB_NAME:-$DEFAULT_CHAT_SUB_NAME}"
+  local chat_sub_name="${PARAM_CHAT_SUB_NAME:-}"
```
mplakhtiy, line 3158: **LOW — this probes Terraform state on every install, including ones with no chat integration.**

The block sits in step 6's body, after `prompt_menu` writes `chat_choice` and before the `case "$chat_choice"` that branches on it — so it runs for choice 4, "None (CLI & REST API Gateway only)", and for Slack-only installs. Each run is a `gcloud storage cat` of the whole state file whose result is then discarded.

Moving the resolution inside the Google-Chat branch removes it — or past `:3725`, which is where the HIGH above wants it anyway.

install.sh
kyber775: Addressed in 2abb56b3:
1. Threaded resolved local `$project_id` and `$cluster_name` into `tf_state_chat_subscription_name` in `_prompt_google_chat_settings` and `run_menu_system`, ensuring probes read the exact cluster coordinates rather than unexported/stale globals.
2. Updated `tf_state_read` to accept target `project` and `cluster` arguments and safely return 1 without calling `gcloud` if coordinates are unset/empty.
3. Dropped the premature probe before step 6 so state queries only execute on active Google Chat configurations.

terraform/examples/full-install/lifecycle.sh
kyber775: Addressed in 2abb56b3: Widened `guard_pubsub_subscription` to verify both `chat_subscription_name` AND `chat_topic_name` against state (`recorded_topic##*/` vs `desired_topic`), halting before `apply -auto-approve` can destroy and recreate a managed subscription via topic replacement. Added test case `test_guard_pubsub_subscription_refuses_when_topic_differs_from_state` in `tests/test_lifecycle_script.py`.

install.sh
kyber775: Addressed in 2abb56b3: Implemented the **state > explicit-and-non-default > derive** precedence ordering across `_prompt_google_chat_settings`, `write_tfvars_from_state`, and `run_menu_system`. If the cluster has no subscription managed in state and the configured subscription name is unset or equals `$DEFAULT_CHAT_SUB_NAME`, a custom topic dynamically derives `${topic}-sub`. Added regression test in `tests/test_installer_common.py` and `tests/test_install_script.py` verifying that an `install.env` carrying the default subscription name does not defeat derivation on secondary fresh clusters.

install.sh
kyber775: Addressed in 2abb56b3: Removed `2>/dev/null` from all state probes across `install.sh`, `installer_common.sh`, and `common.sh`, allowing `tf_state_read` to surface its stderr diagnostic. Added explicit branching on `TF_STATE_RC_UNREADABLE` with warning in `write_tfvars_from_state` and front doors, tested in `tests/test_installer_common.py`.

install.sh
kyber775: Addressed in 2abb56b3: Removed the premature probe block before `case "$chat_choice"`. Probing now happens exclusively inside `_prompt_google_chat_settings` when Google Chat is enabled, eliminating unnecessary GCS calls on Slack-only and None installs.

scripts/installer/installer_common.sh
```diff
@@ -1703,9 +1758,24 @@ write_tfvars_from_state() {
       echo "project_roles  = $(hcl_csv_list "${PLATFORM_AGENT_CUSTOM_ROLES:-}")"
     fi
     echo ""
+    local chat_topic="${CHAT_TOPIC_NAME:-$DEFAULT_CHAT_TOPIC_NAME}"
+    local chat_sub="${CHAT_SUB_NAME:-}"
+    local state_sub="" state_rc=0
+    state_sub="$(tf_state_chat_subscription_name)" || state_rc=$?
+    if [ "$state_rc" -eq "$TF_STATE_RC_UNREADABLE" ]; then
+      print_warning "Could not determine if Google Chat Pub/Sub subscription is in Terraform state (see above); proceeding with configuration."
+    fi
```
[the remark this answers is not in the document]
kyber775: Addressed in 45bf15e5:
1. Gated state-probing inside `write_tfvars_from_state` and `run_menu_system` behind `GOOGLE_CHAT_ENABLED == "true"`, ensuring Slack-only and no-chat configurations skip the probe.
2. Routed `print_warning` to `>&2` across `write_tfvars_from_state` (and other front doors) so diagnostic messages are never captured by stdout redirection into `terraform.tfvars`.
3. Eliminated SC2269 self-assignments in `install.sh`, `scripts/installer/common.sh`, and `scripts/installer/installer_common.sh`, bringing shellcheck to a clean exit 0 on ubuntu and macos runners.
4. Added test cases in `tests/test_installer_common.py` verifying that unreadable state warnings route to stderr and do not contaminate `terraform.tfvars`, and that chat-disabled runs never probe state.

install.sh
```diff
@@ -3136,7 +3155,7 @@ main() {
...
   local google_chat_mode="$PARAM_GOOGLE_CHAT_MODE"
```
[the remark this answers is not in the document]
kyber775: Addressed: Updated the PR description to accurately reflect the diff against `main`. Removed references to the step-6 probe from 'What Changed' and 'Self-Review', clarifying that state probing is introduced exclusively within `_prompt_google_chat_settings` and `run_menu_system` when Google Chat is enabled.

scripts/installer/installer_common.sh
```diff
@@ -847,6 +878,30 @@ for r in doc.get("resources", []):
...
+  state=$(tf_state_read "$@") || return $?
+  printf '%s' "$state" | python3 -c '
+import json, sys
+try:
+    doc = json.load(sys.stdin)
+except Exception:
+    sys.exit(int(sys.argv[1]))
+for r in doc.get("resources", []):
+    if r.get("type") == "google_pubsub_subscription" and r.get("name") == "chat_events" and r.get("mode") == "managed":
+        for i in r.get("instances", []):
```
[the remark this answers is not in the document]
kyber775: Addressed in e2a01483: Declared `readonly TF_PUBSUB_SUBSCRIPTION_TYPE="google_pubsub_subscription"`, `TF_CHAT_EVENTS_SUBSCRIPTION_NAME="chat_events"`, and `TF_RESOURCE_MODE_MANAGED="managed"` at the top of `scripts/installer/installer_common.sh` alongside the other Terraform constants, and passed them into `tf_state_chat_subscription_name`'s Python parser.

install.sh
```diff
@@ -2590,7 +2597,19 @@ run_menu_system() {
...
+    if [ "$state_rc" -eq "$TF_STATE_RC_UNREADABLE" ]; then
+      print_warning "Could not determine if Google Chat Pub/Sub subscription is in Terraform state (see above); proceeding with configuration." >&2
+    fi
+    if [ -n "$state_sub" ]; then
+      chat_sub_name="$state_sub"
+    elif [ -z "$chat_sub_name" ] || [ "$chat_sub_name" = "$DEFAULT_CHAT_SUB_NAME" ]; then
+      chat_sub_name="$(derive_chat_sub_name "$chat_topic_name")"
+    fi
+  fi
   local permission_set="${PLATFORM_AGENT_PERMISSION_SET:-$DEFAULT_PERMISSION_SET}"
```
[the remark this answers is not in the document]
kyber775: Addressed in 1941c7fc: Added unit tests in `tests/test_install_script.py` covering `run_menu_system` Save & Apply chat subscription handling:
- `test_run_menu_system_derives_chat_sub_name_on_save_and_apply`: derives `<topic>-sub` from custom topic when state is empty.
- `test_run_menu_system_recovers_chat_sub_name_from_state_on_save_and_apply`: recovers existing subscription from state into `install.env`.
- `test_run_menu_system_rederives_when_recorded_sub_equals_default_and_state_empty`: re-derives custom topic subscription when recorded sub is the default and state is empty.

Verified via sabotage testing that reverting line 2600 and removing the derivation block causes all three tests to fail immediately.

jayantid (approved):

PR #1458, opened 2026-09-11
fix(operator): delete superseded A2A provision Jobs on reconcile

### LINKED ISSUE #1389: A2A provision: a superseded provision Job whose pod never ran is never TTL'd and holds a pod-quota slot
After #1388 the A2A provision Job's name moves whenever its rendered spec changes, which is what lets a pod-spec fix reach an existing install (#1347). It also means a `mode: next` install accumulates superseded Jobs, and one class of them never leaves.

## How a superseded Job leaves today

| superseded Job's state | how it leaves |
| --- | --- |
| completed | `ttlSecondsAfterFinished` |
| pod ran and failed | `backoffLimit` exhausts, then TTL |
| pod never ran (ImagePullBackOff, unschedulable, admission refusal) | never: no terminal condition, so the TTL never starts |

The third row is exactly the image-override case the digest exists for: an install carrying a Job with an unpullable image gets a new Job under a new name and completes, and the old one sits in `Pending` until a mode flip or agent deletion. Meanwhile it holds one slot of the namespace-wide `pods` quota. The operator never deletes a provision Job outside `cleanupA2A` (a mode flip), which removes every generation by label.

## Fix

In `reconcileA2A`, list Jobs by `a2aComponentLabel: provision` plus the instance label and delete any whose name is not the current render's. Two things to keep: do not fold conditions across generations into the phase (the status scan reads only the current name, and `TestReconcileA2ACreatesANewProvisionJobWhenThePodSpecChanges` on #1388 pins that a `Failed` on the old generation does not park the phase), and delete with propagation so the stuck pod goes with the Job.

## Why not in #1388

That change is the naming; a delete in the reconcile loop is a new reconcile-time action with its own hazards and deserves its own review. The builder comment on #1388 states the gap rather than promising a TTL that never starts.

Surfaced by the adversarial self-review pass on #1388. Related: #1347, #1259.

bnaylor:
## Summary

`reconcileA2A` now deletes this agent's superseded provision Jobs on every pass. The provision Job's name carries a hash, so a changed render is a new Job under a new name and the previous one is superseded. A superseded Job whose pod never ran (unpullable image, unschedulable pod, admission refusal) has no terminal condition, so `ttlSecondsAfterFinished` never starts and it sat in `Pending` holding one slot of the namespace pod quota until a mode flip or agent deletion. The sweep lists provision Jobs by the component and instance labels after the current Job is ensured, deletes every one whose name is not the current render's with background propagation so the pod goes too, and leaves the status scan reading the current name only.

## Why This Change

On `main` the name moves when the script changes; with #1388 it moves when anything in the rendered spec changes, which is what lets a pod-spec fix reach an existing install (#1347). Either way a `mode: next` install accumulates superseded generations, and the operator never deleted one outside `cleanupA2A` (a mode flip). A completed generation leaves by TTL and one whose pod ran and failed leaves by `backoffLimit` then TTL, but the third class never leaves, and it is exactly the image-override case the digested name exists for: an install carrying a Job with an unpullable image gets a new Job and completes, and the old one stays `Pending` forever, counted against the session-pod `ResourceQuota` this same function renders. Without this change that slot is held until someone flips the mode or deletes the Job by hand.

## What Changed

- `k8s-operator/internal/controller/platformagent_a2a_manifests.go`: new `deleteA2AProvisionJobs(ctx, agent, keep)` lists Jobs by `kubeagents.x-k8s.io/a2a-component=provision` plus the instance label and deletes each one that is not `keep`, is not already carrying a `deletionTimestamp`, and is controlled by this agent, with `DeletePropagationBackground`. `reconcileA2A` calls it with the current render's name immediately after the create-or-scan block, so a pass leaves N+1 provision Jobs for a moment and never zero. `cleanupA2A` calls it with `""`, which is the sweep it already had, moved onto the shared helper so the two cannot drift in what they select.
- The component value gets a name, `a2aProvisionComponent`, used by the helper. The builder's `a2aLabels(agent, "provision")` literal is left where it is (that line is inside #1388's hunk); the test pins the two together.
- The status scan is unchanged: it reads the current name only, so a `Failed` on a generation the sweep deletes never reaches the phase.
- New `platformagent_a2a_provision_job_gc_test.go` (a separate file from #1388's new test file, so the two do not conflict at file level).
- No RBAC change: the operator already holds `delete` on `batch/jobs` (`platformagent_controller.go` marker, rendered in `config/rbac/role.yaml`), so `make chart-check` has nothing to see and was not run.

## Context

Closes #1389. Related: #1347 (the naming gap), #1388 (the fix for it), #1259 (the WorkingDir fix that first showed a rendered change not reaching an install).

**Overlap with #1388** (`fix/a2a-provision-job-digests-pod-template`, head `98268914`, same author). Both PRs edit `platformagent_a2a_manifests.go`; this branch is based on `upstream/main`, not on #1388. The hunks are placed apart: #1388 rewrites the `buildA2AProvisionJob` doc comment, the builder body, and the comment above the create-or-scan block; this PR adds code after that block and replaces the sweep body at the end of `cleanupA2A` (where #1388 changes one comment line). Either can land first; whichever lands second takes a small merge-up. **#1388 landed first, on 2026-09-11, so this is the second PR and both carries below are now done on this branch.** Two things the second one should carry:

- **Done, in `167d1d28`.** #1388's builder comment stated, as the gap it leaves, that "a superseded Job is not deleted here" and that deleting superseded generations by label "is owed". The merge-up in `c4e6b14b` corrected the `reconcileA2A` comment and left this one, which mplakhtiy caught on review; `167d1d28` rewrites the paragraph so the never-ran case reads as the reason `reconcileA2A` sweeps superseded generations by label through `deleteA2AProvisionJobs` after ensuring the current Job, and drops both "is owed" and "is not deleted here".
- **Done, in the merge-up `c4e6b14b`.** #1388's `TestReconcileA2ACreatesANewProvisionJobWhenThePodSpecChanges` asserted `len(jobs.Items) == 2` after the rename ("the old one left alone"). With this sweep the answer is 1, and the assertion now reads `want 1`; the test's other property, that a `Failed` on the superseded generation does not park the phase, is preserved and is pinned independently here by `TestReconcileA2ASupersededProvisionJobFailureDoesNotParkThePhase`.

The `Python Unit Tests` job has been red on `upstream/main` since #1364; the fix is open as #1453. If it shows red here it is that, not this diff; nothing in this PR touches Python.

**Handoff for the PR minder.** Verified: the two new tests fail on `upstream/main` at the deletion assertions and pass here; the full `make -C k8s-operator test` passes with envtest cases running (0 skips); the operator already holds `delete` on Jobs; `a2a-next-dev` currently carries exactly one provision Job (Complete) and no superseded one, observed read-only. Assumed, not measured: the cost of one uncached label-selected Job List per reconcile of a `next` install is acceptable, on the argument that the Job Get before it is already uncached and of the same shape. Not live-tested: the only `mode: next` install is carrying another session's stacked build and is read-only for this work, so no operator was rolled; see Live validation. Merge order against #1388: either; the second takes the merge-up described above. No follow-up issue filed.

## Testing

- `make -C k8s-operator test` on the branch: pass (`internal/controller` 91.2% coverage). envtest cases ran under `KUBEBUILDER_ASSETS`: `TestObservedGenerationFollowsMetadataGenerationEnvtest` PASS, `--- SKIP` count 0 across `internal/controller` and `internal/webhook` with `-v`.
- The two new tests against `upstream/main` (`76463db1`), with the new constant's name substituted by its literal so the file compiles there: both FAIL, at "the superseded provision Job survived the reconcile" and "no Delete was issued" in both subtests, and at "the failed superseded Job survived" in the phase test. The phase assertion itself passes on `main`, as expected: the current-name-only scan predates this change and this PR keeps it.
- `TestReconcileA2ADeletesSupersededProvisionJobs` seeds a superseded provision Job of this agent, a provision Job of another agent instance, a non-provision Job of this agent, and a provision-labelled Job this agent does not own, then reconciles. It asserts the superseded Job is gone and was deleted with `Background` propagation (captured through a `Delete` interceptor, because the fake client ignores the option), that the other three and the current Job survive, that exactly one Delete was issued, and that a second pass issues none. Two subtests cover the current Job being created by that pass and already present before it (the upgrade case: an operator carrying this sweep for the first time over superseded Jobs that predate it).
- `TestReconcileA2ASupersededProvisionJobFailureDoesNotParkThePhase` marks the superseded Job `Failed` through the status subresource, re-reads it to prove the condition stuck, reconciles, and asserts the returned state is neither `done` nor `failed`, the superseded Job is gone, and the current one was created.
- `go vet ./internal/controller/` clean. `gofmt -l` on the two changed files: nothing. (`netpolprofile.go` is un-gofmt'd on `upstream/main`; `make fmt` rewrote it during the run and that rewrite was reverted so it is not in this diff.)
- Not run: `make chart-check` (no `charts/` or `k8s-operator/config` change), `make docs-check` (no docs change).

### Live validation

Not live-tested. The only `spec.mode: next` install reachable from here, `a2a-next-dev` (`kubeagents-system`, operator `a2a-demo/k8s-operator:a2-20260908-162517`), is carrying another session's stacked build and is read-only for this work, so no operator was rolled there and no CR was patched. The default-mode install never renders the A2A stack and cannot exercise this path.

What was observed, read-only, on `a2a-next-dev` at 2026-09-11T04:13Z: exactly one provision Job by label, `platform-agent-a2a-provision-0a06e350`, `Complete`, with no pod remaining, and no superseded generation. So there is no live instance of the symptom to point at today; the proof is the unit tests above, which drive `reconcileA2A` against a fake API carrying a superseded Job and observe the Delete and its propagation policy directly.

The mechanism this cannot prove: that Kubernetes garbage collection removes the `Pending` pod after a background-propagated Job delete and that the `ResourceQuota` count drops. That is standard Job semantics, and it is the same propagation `cleanupA2A` has used for every mode flip on this install.

## Self-Review

Both passes ran in fresh `general-purpose` subagents handed only the worktree path, the diff range `upstream/main...HEAD`, and the skill file; neither had the reasoning behind the change. One merged list:

- **Adversarial, HIGH, confirmed: divergence with open #1388.** Its `TestReconcileA2ACreatesANewProvisionJobWhenThePodSpecChanges` asserts two Jobs after a rename; this sweep leaves one. Its rewritten `reconcileA2A` and `buildA2AProvisionJob` comments say the superseded Job "is left to its TTL" / "is not deleted here"; both are false once this lands. Both PRs add to the same `const` block and touch the same `cleanupA2A` comment lines. Its phase-does-not-park test and this PR's `TestReconcileA2ASupersededProvisionJobFailureDoesNotParkThePhase` share a fixture and assert opposite things about the old Job's survival. Disposition: not stacked, deliberately. This fix is valid on `main` alone (the name already moves on a script change) and stacking would tie it to #1388's review; the reviewer's own remedy, declaring the merge order, is done in Context, which names each item the second merge-up carries: flip the "want 2" assertion, correct the two comments, fold the two phase tests into one. **Resolved:** #1388 merged first, and every item on that list is now carried on this branch — the assertion flip and the `reconcileA2A` comment in `c4e6b14b`, the `buildA2AProvisionJob` comment in `167d1d28` after mplakhtiy caught that the merge-up had missed it.
- **Adversarial, confirmed, fixed:** the sweep's comment said the never-ran case "is the image-override case the digested name exists for", which describes #1388's digest and is not true of `main`, where the name digests the script only. Reworded to name both routes: an unpullable image once the name tracks the pod spec (#1347), a bad script today.
- **Adversarial, LOW, confirmed, not fixed:** `a2aProvisionComponent` is declared while the builder still spells `"provision"`, with a test assertion pinning the two. The two builder lines are inside #1388's hunk (it rewrites the `ObjectMeta` line); swapping them here is a certain conflict for two token changes. The rule binds the lines written, the pin holds until then, and the merge-up in Context is where the swap goes.
- **Docs-drift, Advisory, now fixed:** the `buildA2AProvisionJob` comment described completion by TTL and said nothing about superseded generations. This was deferred to the second merge-up, and #1388 merged first, so `167d1d28` does it: the paragraph now names the sweep and `deleteA2AProvisionJobs` instead of describing the gap. No Blocking drift; `make docs-check` passes at HEAD; no page owns provision-Job lifecycle prose, so none is owed.
- **Consequences recorded, not defects (adversarial):** a superseded Completed or Failed generation now leaves on the next reconcile rather than after the 24h TTL, so a crash-looped old generation's pod logs are gone before anyone reads them; a superseded Job still mid-run is killed, which the idempotent `info || add` script and the current Job's re-run absorb; the `DeletionTimestamp` skip is untested, and its regression is a redundant Delete the API server accepts. The issue asks for exactly "delete any whose name is not the current render's", so these are the requested behaviour.
- **Checked and clean (adversarial):** RBAC already grants `list;delete` on `batch/jobs` and `role.yaml` matches; background propagation is load-bearing because the API server defaults `batch/v1` Job deletion to orphan; `TestCleanupA2ACostsThreeReadsWhenThereIsNothingToClean` still holds because the helper sits behind the early exit; the pass rebuilt with the sweep call removed via `-overlay` and both new tests went red; open PRs #1319 (adds a provision-labelled ServiceAccount, not a Job), #1258 and #1393 do not touch this block.
- **Not covered by either pass:** no live cluster, so quota-slot release and the terminating-pod overlap with the new Job's admission were reasoned from Job semantics, not observed; the adversarial pass did not check whether the envtest case ran (it did, per Testing).

## Risk & Rollout

The new failure mode is a delete in the reconcile loop, and its blast radius is bounded three ways: the List is scoped to the agent's namespace, the provision component label, and the agent's instance label; a Job that this agent does not control is skipped; and the current render's name is always kept, with the sweep running only after that Job is ensured, so a reconcile never removes the last provision Job. The live case on an upgrade is one delete per superseded generation on the first reconcile, then nothing. A List error fails the reconcile with a named error and the next pass retries; nothing is left half-done because each Delete is independent. Revert is a plain revert of the commit; the superseded Jobs it would have removed go back to leaving by TTL or by mode flip. Nothing has to happen at merge time. The standing cost is one uncached label-selected Job List per reconcile of a `next` install, beside the uncached Job Get already there; `today` installs do not reach this path.


mplakhtiy (commented):
One finding from a full read at 67af380e. The four base merges since the bot's pass were replayed with `git merge-tree --write-tree`; three carried hand-written resolution, including the `c4e6b14b` rewrite of `TestReconcileA2ACreatesANewProvisionJobWhenThePodSpecChanges` (`want 2` to `want 1`), which I read whole and found correct. The sweep does what the body says: List scoped to namespace, component and instance labels, `IsControlledBy` guard, current name kept, background propagation carried, RBAC already grants `list` and `delete` on `batch/jobs`.

#### Medium: `buildA2AProvisionJob`'s doc comment now states the opposite of the sweep below it

`k8s-operator/internal/controller/platformagent_a2a_manifests.go:1283-1292`, the paragraph #1388 added, still reads: "A superseded Job is not deleted here, and how it leaves depends on how far it got. ... one whose pod never ran ... stays until the mode flips or the agent is deleted, holding one slot in the namespace pod quota the whole time. ... deleting superseded generations by label is owed, not merely nice." Thirty lines below, the new sweep comment in `reconcileA2A` (L1856) says superseded generations "go now, not by TTL". Once this merges the builder comment is false on every sentence about the never-ran case, and the builder is where the next person who hits #1389's symptom on an install that predates the sweep will start reading.

The PR body named this paragraph as the first thing the second-landing PR should carry. #1388 merged on 2026-09-11, so this is that PR; the merge-up in `c4e6b14b` corrected the `reconcileA2A` comment and left this one.

Fix: rewrite the paragraph to say the never-ran case is why `reconcileA2A` sweeps superseded generations by label after ensuring the current Job (pointing at `deleteA2AProvisionJobs`), and drop "is owed" and "is not deleted here".

_Review pass run with Claude Code; findings verified against the head commit before posting._

bnaylor:
@mplakhtiy — confirmed and fixed in `167d1d28`.

You read it right on every point. The merge-up in `c4e6b14b` corrected the `reconcileA2A` comment and the `want 2` assertion and left the builder paragraph alone, so at `67af380e` the file said "is not deleted here" and "stays until the mode flips or the agent is deleted" thirty lines above a sweep that deletes it on the next pass. The builder is the right place to care about, for exactly the reason you give: someone landing on #1389's symptom on an install that predates the sweep reads the constructor first.

The paragraph now reads:

> The digest covers everything this function renders into the spec: the
> script, the image, the uid and security contexts, env, volumes, mounts,
> backoffLimit and the TTL. What becomes of the generation the render has
> moved past depends on how far that generation got, and one case is why the
> rename on its own is not enough. A completed one leaves by TTL; one whose
> pod ran and failed runs out its backoffLimit and then leaves by TTL; one
> whose pod never ran — an unpullable image, an unschedulable pod, an
> admission refusal — has no terminal condition for the TTL to start from
> and would sit in Pending until the mode flips or the agent is deleted,
> holding one slot in the namespace pod quota the whole time. That last case
> is the image override scenario this digest exists for, and it is why
> reconcileA2A sweeps superseded generations by label, keeping only the
> current render's name (deleteA2AProvisionJobs), after ensuring the Job this
> function builds rather than before it — so a reconcile leaves N+1 of them
> for a moment, never zero. Two things hold alongside that sweep: the status
> scan in reconcileA2A reads the current name only, so a Failed condition on
> a generation the sweep is about to remove never reaches the phase, and
> cleanupA2A calls the same function keeping nothing, so a mode flip removes
> every generation at once.

"is owed" and "is not deleted here" are gone; the never-ran case survives as the reason the sweep exists rather than as an outstanding debt, and `deleteA2AProvisionJobs` is named so the reader can jump straight to it. The "until the mode flips or the agent is deleted" clause is kept but re-pointed: it is now the counterfactual describing what the sweep prevents, not a claim about what happens today.

I also updated the PR body: both items on the "two things the second one should carry" list are marked done, with the commit that carried each, and the Self-Review's docs-drift bullet is no longer deferred.

Comment-only change, no behaviour touched. Gates at `167d1d28`: `go build ./...`, `go vet ./internal/...`, `go test ./internal/controller/` (ok, 22.5s), `make docs-check`, and `python3 tests/conformance/run.py` (112 tests, OK) all pass. `gofmt -l internal/` reports `platformagent_litellm_policy_test.go`, which is un-gofmt'd on `main` since #1195 and is not touched by this branch.

mplakhtiy (approved):

PR #1459, opened 2026-09-11
fix(operator): stop updateStatusDegraded writing status on every pass

### LINKED ISSUE #1392: updateStatusDegraded writes status on every pass, re-enqueuing a refused PlatformAgent continuously
`updateStatusDegraded` (`k8s-operator/internal/controller/platformagent_controller.go:2601`) writes the PlatformAgent status on every call:

```go
agent.Status.Phase = "Degraded"
now := metav1.Now()
agent.Status.LastReconcileTime = &now
...
meta.SetStatusCondition(&agent.Status.Conditions, condition)
return r.Status().Update(ctx, agent)
```

There is no equality gate, and `LastReconcileTime` is stamped with `now`, so the write is never a no-op. The controller's primary watch is `For(&agentv1alpha1.PlatformAgent{})` with no predicate (`:2639`; the `GenerationChangedPredicate` at `:2680` is on a secondary watch), so each status write re-enqueues the object. A CR parked on any refusal (`RuntimeClassNotFound`, `ShellSandboxKeysMissing`, `ModeNotRecognized`, the others that route through this helper) therefore reconciles continuously, bounded only by the workqueue's serialization and any rate limiting, with one status write and one audit entry per pass.

`updateStatusReady` already has the gate this path lacks: it compares the status it is about to write against the current one and writes only when something moved. The fix is the same gate here, keyed on phase, reason, message and the conditions rather than on `LastReconcileTime`.

This is read from source, not measured: the loop is plausible from the code path and the watch configuration, and nobody has yet put a counter on a refused CR. The first step of the fix is that measurement (operator log line count or `controller_runtime_reconcile_total` for a CR held on `ShellSandboxKeysMissing` for one minute).

Surfaced by the adversarial self-review pass on #1391, which adds an `observedGeneration` assignment inside this write and neither introduces nor worsens the loop. Related: #534.

bnaylor:
## Summary

`updateStatusDegraded` now writes the PlatformAgent status only when something it is about to write differs from what is there, the same equality gate `updateStatusReady` already had. Before this, every pass through a refusal (`RuntimeClassNotFound`, `ShellSandboxKeysMissing`, `ModeNotRecognized`, `ForbiddenVolumeMount`, `ShellSandboxCannotBeDisabled`, `EgressAllowlistRefused`, `A2AProvisionFailed`) stamped `lastReconcileTime` with `now` and wrote, and each write was an event on the predicate-less PlatformAgent watch. #1392 asked for the loop to be measured before it was fixed, because it had been read from source rather than observed. It was measured, under envtest with the real manager and live on the shared install, and the measurement changes the picture: the loop is bounded, not continuous. What a parked CR actually did was two reconciles and two status-write requests per 30 s requeue tick, moving `resourceVersion` and `lastReconcileTime` every tick. With the gate it is one reconcile (the requeue) and no writes per tick, and the object does not move.

## Why This Change

A CR held on a refusal is the operator's steady state on a misconfigured install, sometimes for hours. On `upstream/main` that state was not quiet: each requeue tick wrote the status, the write woke the next pass through the watch, that pass wrote again, and only then did the chain stop. The reason it stopped is worth stating because it is why the issue's "reconciles continuously" did not materialise: `metav1.Time` serializes to the second, so the echo pass's write is byte-identical to the one before it, and the API server drops a byte-identical update without touching etcd, bumping `resourceVersion`, or emitting a watch event (`k8s.io/apiserver` `pkg/storage/etcd3/store.go`, `GuaranteedUpdate`, the `bytes.Equal(data, origState.data)` short-circuit). When the echo crossed a second boundary the chain ran one round longer; the live transcript below shows one such round.

What stayed broken without the gate: two status-write API requests, and therefore two API-server audit entries, per tick for every parked CR; a `resourceVersion` and `lastReconcileTime` that moved every 30 s while nothing about the object changed, so `kubectl get -w` and anything else watching the CR saw a heartbeat instead of a state; and a second write per tick that did nothing except cost a round-trip. The fix is small and the measurement is most of the value: the numbers are in Testing.

## What Changed

- `updateStatusDegraded` returns before writing when the phase is already `Degraded` and the `Ready` condition already carries `status=False`, the same reason, the same message, and an `observedGeneration` equal to `metadata.generation`. Any of those differing still writes; a generation bump alone still writes (#534). `lastReconcileTime` is not in the key and is not refreshed on a quiet pass; it is the time of the last status write, which is what `updateStatusReady` does and what the field's doc says. The generation witness is the condition's `observedGeneration` rather than the top-level field, for the CRD-skew reason `updateStatusReady` documents: a CRD that predates `status.observedGeneration` prunes the top-level copy on every write, and keyed on that copy the gate would see `0 != generation` on every pass.
- The audit entry the issue mentions is the API server's own audit record of the status-write request; there is no operator-side audit emitter on this path. Gating the write gates the entry.
- `docs/site/.../operator/platformagent-crd.md`: the `lastReconcileTime` row says what the field measures. This PR also edited the `Ready` row beside it — dropping its hard-coded count of refusal reasons and adding `ForbiddenVolumeMount` — but #1195 made the same fix on main first, so that half is superseded and is not in this diff.
- `docs/designs/testing-strategy.md`: the sentence saying no envtest exists below the cloud e2e tier was already false after #1391 and is more so now; it now says what exists.
- `k8s-operator/go.mod`: `prometheus/client_model` moves from indirect to direct; the envtest reads `controller_runtime_reconcile_total` out of the in-process registry.

## Context

Closes #1392. Related: #534 (observedGeneration), #1391 (merged 2026-09-11, added the `observedGeneration` assignment inside this write and surfaced this issue in its adversarial pass), #1009 (`reportRBACSkew`'s gate, the other precedent).

Overlapping open PRs: none found. `gh pr list --state open --search "updateStatusDegraded"` and searches on "degraded status", "LastReconcileTime", and "reconcile loop status write" return nothing that touches this function; #1391 is the neighbour and is merged. #1453 (`fix/bench-case-redactor-import`) fixes the `Python Unit Tests` job, red on `upstream/main` since #1364; if this PR's run shows that job red, that is why, and it is not cherry-picked here.

**Handoff for the PR minder.** Verified: the gate's behaviour under the fake client (four new tests, three of which fail on `upstream/main`), under envtest with the real manager and watches (fails on `upstream/main`, passes here, numbers below), and the *before* behaviour live on `bnaylor-ka-test` under the deployed operator (numbers below; the operator there is `dev-pr724b`, weeks behind main, and shows the same two-writes-per-tick pattern). Not verified live: the *after* behaviour, because reaching it means rolling a main-based operator image onto an install whose CRDs and objects are weeks behind it, and that is not a change a revert restores; the envtest ran the real `kube-apiserver` 1.36 binary and the real controller-runtime manager, which are the layers the claim is about. Assumed: that GKE's Admin Activity audit log records each `platformagents/status` update as one entry; not queried. No follow-up issues filed. The envtest case adds about 40 s to `make -C k8s-operator test` (a 35 s window, sized to include one 30 s requeue tick so the test also proves the requeue survived the gate).

## Testing

- `make -C k8s-operator test` (manifests, generate, fmt, vet, unit, envtest, python): exit 0 on the branch. `make fmt` reformats `internal/controller/netpolprofile.go`, which is pre-existing on `upstream/main` and was reverted rather than committed.
- `make docs-check`: clean. `prettier --check` (3.9.6) on both changed `.md` files: clean.
- New fake-client tests, `k8s-operator/internal/controller/platformagent_degraded_gate_test.go`: `TestAParkedRefusalDoesNotWriteStatusEveryPass` (full `Reconcile` on a CR parked on `RuntimeClassNotFound`, one write on the parking pass, zero on three more, `lastReconcileTime` unchanged), `TestADegradedChangeStillWrites` (same refusal twice writes once; message change, reason change, generation bump each write once and are then quiet), `TestAPhaseChangeAloneWritesDegraded`, `TestAPrunedObservedGenerationDoesNotWriteDegradedEveryPass` (CRD pruning the top-level field does not bring the loop back). The first, second and fourth fail on `upstream/main` with the controller unchanged and the tests copied in.
- New envtest, `platformagent_degraded_loop_envtest_test.go`: `TestAParkedRefusalDoesNotReconcileContinuouslyEnvtest` boots the controller under its real manager via `SetupWithManager`, creates a CR with no sandbox-keys Secret so it parks on `ShellSandboxKeysMissing`, and once parked counts `controller_runtime_reconcile_total` and PlatformAgent status writes for 35 s. Confirmed RAN (not skipped) with `KUBEBUILDER_ASSETS` set; the older envtest case in the package also ran.

Measurement, same test, same window (35 s after the CR parked), `kube-apiserver` 1.36.0 under envtest:

| Revision                                          | Reconciles | Status writes | `resourceVersion` | `lastReconcileTime` | Result |
| ------------------------------------------------- | ---------- | ------------- | ----------------- | ------------------- | ------ |
| `upstream/main` `76463db1` (tests copied in)      | 2          | 2             | 245 → 255         | moved (+30 s)       | FAIL   |
| this branch                                       | 1          | 0             | 244 → 244         | unchanged           | PASS   |

The single reconcile on the branch is the 30 s `RequeueAfter` tick firing inside the window; the test asserts it did (a zero would mean the gate also killed the poll a parked CR needs). Transcripts: `envtest-before-upstream-main.txt`, `envtest-after-branch.txt`, `make-test.txt` in the evidence set the parent session holds.

### Live validation

Install: `gke_bnaylor-kagents-dev_us-east4_bnaylor-ka-test`, namespace `kubeagents-system`, operator image `us-east4-docker.pkg.dev/bnaylor-kagents-dev/kube-agents/k8s-operator:dev-pr724b` (default mode, weeks behind main). Lease `kube-agents-vamp-71` held for the duration and released.

**Before, live (the deployed operator, not this branch).** The webhook allows one PlatformAgent per cluster (a scratch CR was refused with `only one PlatformAgent is allowed per cluster`), so the install's own CR was parked by adding `spec.deployment.availability.runtimeClassName: does-not-exist-1392` after snapshotting it, watched for 100 s, then reverted with a JSON-patch `remove`. The `RuntimeClassNotFound` path refuses before the workload step, so the running pods were not touched; pod names, start times and restart counts were identical before and after, and the spec matched the snapshot after the revert. Phase returned to `Ready/Reconciled` 3 s after the revert.

```
# every change the API server emitted for the CR over 100 s (UTC | resourceVersion | generation | phase | Ready.reason | lastReconcileTime)
04:22:54 | 1789100574130543018   60    Degraded   RuntimeClassNotFound   2026-09-11T04:22:54Z
04:23:24 | 1789100604507871018   60    Degraded   RuntimeClassNotFound   2026-09-11T04:23:24Z
04:23:54 | 1789100634939855018   60    Degraded   RuntimeClassNotFound   2026-09-11T04:23:54Z
04:23:55 | 1789100635135247018   60    Degraded   RuntimeClassNotFound   2026-09-11T04:23:55Z
04:24:25 | 1789100665499311018   60    Degraded   RuntimeClassNotFound   2026-09-11T04:24:25Z
# operator log, "Reconciling PlatformAgent" lines in the same window, per second
   1 2026-09-11T04:22:53Z
   1 2026-09-11T04:22:54Z
   2 2026-09-11T04:23:24Z
   2 2026-09-11T04:23:54Z
   1 2026-09-11T04:23:55Z
   2 2026-09-11T04:24:25Z
TOTAL reconciles in window: 9      (and 9 "RuntimeClass 'does-not-exist-1392' is not configured" refusal lines)
```

That is the envtest pattern in the wild: per 30 s tick, two reconciles, one `resourceVersion` bump, `lastReconcileTime` advancing; and at 04:23:54–55 the echo write crossed a second boundary, the API server saw a new object, and the chain ran one round longer before stopping. Nine reconciles and nine status-write requests in 100 s for an object nothing changed. On this branch the same 100 s would hold three reconciles (the ticks) and no writes.

**After, live: not done.** Rolling a main-based operator onto this install would re-render every object of the existing agent under CRDs weeks older than the binary, which is not a change a revert restores, and the shared install carries other sessions' work. The *after* claim rests on the envtest above, which ran the real API server and the real manager, and on the fake-client tests.

Artifacts left behind: none. The scratch namespace from the refused first attempt was deleted, the CR spec and pods were verified unchanged, the lease was released.

## Self-Review

Both passes ran in fresh `general-purpose` subagents handed only the worktree path, the diff range `upstream/main...HEAD`, and the skill file; neither had this session's reasoning.

Adversarial pass (`.agents/skills/review-adversarial/SKILL.md`), fresh subagent. It re-derived the intent from the diff, read the gate against every caller, `updateStatusReady`'s and `reportRBACSkew`'s gates, and each refusal's message construction, checked red/green for every new test by running them against `upstream/main`'s controller in a throwaway copy, and inspected the envtest harness (cache sync escape, `SkipNameValidation`, the discard logger, the bind-address sentinel, the requeue landing inside the window). One finding:

- MEDIUM, confirmed: the comments in the controller and both test files stated the failure mode the issue predicted ("reconciled continuously", "hundreds", "bounded only by API-server round-trip time"), which the diff's own envtest contradicts; the pass independently measured 2 reconciles / 2 writes per tick on `upstream/main` and named the same `metav1.Time` second-resolution mechanism. **Fixed**: all four comments now state the measured behaviour and why the chain stops. The pass flagged as open whether a real cluster, where a pass can exceed a second, sustains the chain at ~1 Hz; the live run above answers it for this install (passes are sub-second; per tick 2 reconciles, once 3).
- Not findings it recorded: the gate's key covers everything the function writes except timestamps and every caller returns straight after the call; refusal messages are stable pass to pass, including `A2AProvisionFailed`'s; after this change every PlatformAgent status writer is gated; the `lastReconcileTime` semantics change matches the field's Go doc and CRD description and nothing in the repo reads the field; the `"Ready"`/`"Degraded"` literals in the gate are the subject of their lines and have no existing named constant (hoisting them is a separate change under `core_engineering.md`'s preamble); no open PR carries the same gate, #1393 shares only a trivial `go.mod` `require`-block adjacency.
- It did not run a cluster (covered by the live section above) and drove only two of the seven refusal paths end to end (`RuntimeClassNotFound` under the fake client, `ShellSandboxKeysMissing` under envtest); the other five were checked by reading their message construction.

Docs-drift pass (`.agents/skills/review-docs-drift/SKILL.md`), fresh subagent. Looked at: every doc mentioning `lastReconcileTime`, `updateStatusDegraded`, `updateStatusReady`, the Degraded reasons, requeue behaviour, envtest instructions, `#1392`, `client_model`; the CRD Go doc comment and generated CRD description; the docs map, generated regions, `make docs-check`. Blocking: none; the edited `lastReconcileTime` row was verified true against both writers and the Go doc. Advisory, all dispositioned:

- A1, the `Ready` row hard-codes "the five reasons" and omits `ForbiddenVolumeMount` (pre-existing since #913). **Superseded**: this PR fixed it in `67faa35a`, and #1195 landed the same fix on main in a different order. The base merge took main's side on that row, which is right, and the row on main is now correct. Nothing left to do here.
- A2, `docs/designs/testing-strategy.md:46` says no envtest exists below the cloud e2e tier, false since #1391 and more so with this PR. **Fixed** with one sentence saying what exists and what gap remains; the seam ranking it points at is untouched.
- A3, the Go doc comment on `LastReconcileTime` is less specific than the site row. **Not changed**: they agree, the site page is the canonical user-facing reference, and editing the Go comment regenerates the CRD and its chart mirror for no change in meaning.

## Risk & Rollout

Runtime change is one early return in a status writer. The failure mode it could introduce is a refusal whose status is stale because the gate thought nothing moved; the key covers every field the function writes except the two timestamps, and the tests pin that reason, message, phase and generation changes each still write. A CRD that prunes `status.observedGeneration` is handled by keying on the condition's copy, as `updateStatusReady` does. No manifest, RBAC or chart change; no rollout step. Revert is a plain revert of the commit. The envtest adds about 40 s to the operator test job.


mplakhtiy (commented):
One finding from a full read at 6069c336. Both base merges since the bot's pass were replayed with `git merge-tree --write-tree` and diffed against the merge commit's tree; `c03c87d2` is byte-identical to the automatic result, `6069c336` is not. The gate itself is correct.

#### Medium: merge `6069c336` dropped this PR's `lastReconcileTime` row edit in `docs/site/src/content/docs/operator/platformagent-crd.md`, and that row was not in conflict

`67faa35a` reworded the row to "Timestamp of the last status write. A reconcile that changes nothing in the status leaves it where it was." and added `ForbiddenVolumeMount` to the `Ready` row. The second base merge met a real conflict on the `Ready` row (main's #1195 had made the same `ForbiddenVolumeMount` fix in a different order) and took main's side, as its message says. The same resolution also put the `lastReconcileTime` row back to main's "Timestamp of the last status update." That row was not in conflict: the replayed automatic merge keeps the PR's wording (`git diff f05e0af6 6069c336^{tree}` shows the reversal as a plain hunk beside the marker cleanup), and no base commit before or after touches the line.

The head's diff against main therefore carries no `docs/site/` change at all (6 changed files), so the one user-facing sentence saying a quiet pass leaves the timestamp alone now exists nowhere, while the body's What Changed, Testing, and docs-drift dispositions still describe it as present.

Fix, author's choice: re-apply the one-line row edit on top of the head, or drop the doc change deliberately and update the three places in the body that claim it. Either way the `Ready`-row half was superseded by #1195 on main and can be noted as such.

_Review pass run with Claude Code; findings verified against the head commit before posting._

bnaylor:
@mplakhtiy — confirmed, and fixed in `fe2c337d`.

I checked the merge before changing anything. The merge base of `c03c87d2` (this branch) and `2c560dd8` (main, #1195) still carries "Timestamp of the last status update.", main never touched that row after it, and this branch had it as "Timestamp of the last status write. A reconcile that changes nothing in the status leaves it where it was." So the row was a clean one-sided change and the automatic merge would have kept it; taking main's side on the `Ready` row above it took this one down with it. You are right that the head's diff against main had no `docs/site/` change at all, which meant the only user-facing sentence describing what this PR does existed nowhere.

Re-applied that one row, verbatim from `67faa35a`.

Deliberately did **not** re-apply the `Ready`-row half. #1195 landed the same fix on main: the row there already drops the hard-coded "five reasons" count and lists `ForbiddenVolumeMount`. Only the ordering of the reason list differs, and that is not worth a conflict on a row main owns now. The base merge taking main's side on that row was the right call — it was only the neighbouring row that came along by accident.

Body updated in the two places that claimed the `Ready`-row edit: the What Changed bullet now scopes the doc change to the `lastReconcileTime` row and notes the other half as superseded, and the docs-drift A1 disposition moves from **Fixed** to **Superseded** with the same reasoning. The Testing line still holds — two changed `.md` files, both clean under `prettier --check`.

Gates on the new head: `make docs-check` clean, `python3 tests/conformance/run.py` 110 tests OK, `go build ./...` and `go vet ./internal/...` clean, `go test -count=1 ./internal/controller/` ok in 21.3s. `gofmt -l internal/` still names `netpolprofile.go` and `platformagent_litellm_policy_test.go`; both are unformatted on `2c560dd8` already and neither is in this PR's diff.

mplakhtiy (approved):

Issue #1139, opened 2026-09-01
test: images-check renders the chart in one configuration, so githubMinter's image is never checked


bradhoekstra:
`make images-check` (check 3 in `hack/check-image-inventory.sh`) renders the chart twice, but both renders come from the same default `REQUIRED_VALUES`. Anything behind an off-by-default value is therefore invisible to 3a/3b/3c.

Concretely: with `githubMinter.enabled=true` plus `githubMinter.org` and `githubMinter.repo`, the chart renders `github-token-minter-server:v2.7.1-amd64` — an `images.json` entry the checks never see. It mirrors correctly today, so this is a coverage gap rather than a live bug.

**The part that only shows up when you try it:** turning `githubMinter` on fails 3b immediately, for an unrelated reason. `chart_images()` extracts images with a `value:` regex that matches any quoted value containing a slash and a colon. The minter's `ISSUER_ALLOWLIST` —

```
"https://container.googleapis.com/v1/...,https://accounts.google.com"
```

— matches that pattern, so 3b reports it as an image rendered off a non-mirrored registry.

So this is two changes, in order:

1. Tighten `chart_images()` so it extracts image references rather than any slash-and-colon string.
2. Add a second render with `githubMinter` enabled to the 3a/3b/3c set.

Found while bumping the Hermes base image in #1132.

PR #1460, opened 2026-09-11
fix(hack): read image: fields with a sed program BSD sed runs

### LINKED ISSUE #1449: build: the githubMinter guard added by #1317 landed red, and hack/check-image-inventory.sh is wired into nothing

bnaylor:
## Summary

`make images-check` exits 1 on a clean checkout of `main` on macOS because `image_field_refs` in `hack/check-image-inventory.sh` matches the optional quote around an `image:` value with `\?`, a GNU sed extension that BSD sed reads as a pattern matching nothing. This PR moves that pattern into a `readonly IMAGE_FIELD_RE` in extended syntax applied with `sed -E`, replaces the guard over the union of the two extractors with one guard per extractor, and adds a text-level lint test that fails in Linux CI on any GNU-only sed escape in the script, since no CI job runs this script under a BSD sed.

## Why This Change

On macOS every `image:` field dropped out of `image_refs`, and only the three `*_IMAGE` env-var references survived. The "rendered no image references at all" guard did not fire because the env-var list kept the union non-empty, so checks 3a, 3b and 3c inspected the env-var references and none of the `image:` fields, quietly, since the script was written. `check_toggle` (#1317) is the first check that needs an `image:` field: the minter adds exactly one, so on macOS it saw the toggle add nothing and exited 1. `docs/site/src/content/docs/deploy/docker-images.md` tells anyone bumping a pin to run `make images-check` locally, so a developer on a Mac has been running a check that skipped the `image:` fields and, since #1317, one that fails outright.

CI was green throughout: the script is wired in through the `images-check` Makefile target and the "Check the container-image inventory" step of `validate.yml`, and that runner is GNU sed, which accepts `\?`. The issue's title says the guard landed red and the script is wired into nothing; neither holds, and the correction is on the issue.

## What Changed

- `hack/check-image-inventory.sh`: the `image:` pattern is a named constant, `IMAGE_FIELD_RE`, in extended syntax, applied with `sed -E`. `-E` rather than the POSIX basic `"\{0,1\}` because checks 4 and 5 already use `sed -E` in this file and the two awk patterns beside the new constant are extended syntax too; the comment on the constant says so. The character class now excludes whitespace as well as the quote, so a bare value with trailing blanks comes out without them rather than as a reference no inventory entry can match.
- The union guard `[ -n "$default_images" ]` becomes a guard on `image_field_refs` alone, beside the existing guard on `image_env_refs` alone. Either extractor keeps the union non-empty while the other matches nothing, which is the state macOS was in; the union guard could not see it by construction. The env-var guard's message no longer refers to a guard "above" that no longer exists.
- `tests/test_check_image_inventory_sed_portability.py` (new): reads every `sed` invocation in the script outside comment lines, walks its flags and every expression (`-e` runs, `-i ''`, escaped quotes inside double quotes), resolves any `readonly` constant a program is built from, and fails on `\?`, `\+`, `\|` in a program not invoked with `-E`, on `\s`, `\S`, `\w`, `\W`, `\<`, `\>`, `\b`, `\B` in any program, and on any `--long` option, which BSD sed rejects outright. Each escape was run against `/usr/bin/sed` on macOS 25.6: `\?` and the class and anchor escapes match nothing, `\+`, `\|` and `\b` match the literal character; the list is measured, not copied from documentation. A coverage test fails on a line that runs `sed` in a shape the reader cannot get an expression from (`-f`, an unquoted program), so a future program cannot slip out of reach by being written differently; a self-test checks the lint tells the pre-#1449 program from both portable spellings; and fixture cases pin the argument reader and the command-position finder.
- `tests/test_check_image_inventory_chart_images.py`: lifts the new constant with the function that reads it, asserts the two per-extractor guard lines, and the quoted-or-bare test gains a trailing-space line of each kind.
- `tests/_lift_shell.py`: docstring covers a text-only consumer and no longer counts them.

## Context

Closes #1449. Related: #1317 (the `check_toggle` guard that surfaced this), #1139 (the extraction change that introduced the two-extractor `image_refs`).

**#1455** (kyber775, `fix/1449-check-image-inventory`, opened three minutes after the diagnosis comment on #1449) also says `Closes #1449`. It replaces `grep -Fxv -f` in `added_images` with `comm -23` and adds an `images-check` target to `k8s-operator/Makefile`; it does not touch `image_field_refs`. Measured here: upstream/main's script with #1455's hunk applied still exits 1 on macOS with the same `enabling githubMinter added no image` message, because `image_field_refs` still returns nothing; and BSD `grep -Fxv -f <(...)` returns the correct difference on the lists this script produces. So #1455 does not fix #1449, and if it merges first GitHub closes the issue on a non-fix. The two diffs touch different functions and do not conflict. The `k8s-operator/Makefile` target is unrelated to the failure.

Other open PRs touching `hack/check-image-inventory.sh` (#1258, #1304, #1319) change none of `image_field_refs`, the constants block or the guards: a merge-conflict warning at most.

`Python Unit Tests` is red on `main` at 76463db1 (run 34552465638) for a cause unrelated to this branch: `scripts/validate_bench_cases.py` loads the redactor by path since #1364, and PR #1453 fixes it. The tests this PR adds run in that same job, so it may show red here too; the Testing section shows them passing locally under the job's own discover command, and the whole `tests/` directory passing except seven `test_eval_dashboard_refresh` cases that fail on a macOS `mktemp` path shape in a file this branch does not touch.

Handoff for the PR minder: verified on this machine, against the primary source, that `/usr/bin/sed` mishandles each of the eleven escapes the lint names (eight match nothing, three match the literal character); that the script exits 1 on unmodified `upstream/main` and 0 on this branch under `make images-check`; that reintroducing `\?` fails the new lint at the exact line and makes the new guard fire on macOS; that #1455's hunk does not change the macOS result; and that `validate.yml` is green on `main` at 76463db1 (run 34552465568). Assumed, not verified here: that GNU sed on the CI runner accepts `sed -E -n` with this pattern (no `gsed` and no container runtime on this machine; the same flags already run in checks 4 and 5 and the pattern is plain POSIX ERE). The issue as titled overstates: CI was never red and the script was always wired in; the parent session's comment on #1449 is the correction and the scope this PR follows.

## Testing

Transcripts are from macOS 25.6 (arm64), bash 3.2.57, `/usr/bin/sed` (BSD), `/usr/bin/sort`, helm and jq on PATH, run from inside the worktree.

- Before, on unmodified `upstream/main` at 76463db1: `./hack/check-image-inventory.sh` exits 1 with `ERROR: enabling githubMinter added no image the default render already carried…`; `make images-check` exits 2 (make's wrap of that 1). The existing lifted-function tests already fail on this machine: `pytest tests/test_check_image_inventory_*.py -q` gives 8 failed, 16 passed, every failure in `test_check_image_inventory_chart_images.py` on the same empty extraction.
- After, at this head: `./hack/check-image-inventory.sh` and `make images-check` both exit 0 and print `Image inventory check passed: …`. `(cd tests && python3 -m unittest discover -p 'test_check_image_inventory_*.py' -v)`, the sweep's command narrowed to these modules: `Ran 31 tests … OK`. `pytest tests/test_check_image_inventory_*.py -q`: 31 passed, 65 subtests passed.
- Mutation, at this head: with the pre-#1449 program put back into `image_field_refs` and nothing else changed, `test_sed_programs_use_no_gnu_only_escapes` fails naming `hack/check-image-inventory.sh:261` and the `\?` it found, `test_image_field_refs_program_is_read_through_its_constant` fails because the constant is no longer read, and on macOS the script exits 1 with the new message `ERROR: the chart rendered no 'image:' field that image_field_refs recognises…` instead of the toggle message. Reverted; sha256 of the script matches the pre-mutation value; lint passes again.
- Whole `tests/` directory under the sweep's `PYTHONPATH` and discover command, at the pre-review head 736b254f (the review amendments changed only these test modules and comments): `Ran 1716 tests`, no loader errors, 7 failures all in `test_eval_dashboard_refresh.RefreshScriptTest` (a `mktemp` path under `/var/folders` the refresh harness does not find; unrelated to this branch, files untouched).
- Drift check from `docs/pull-request-workflow.md`: 0 commits behind `upstream/main`, no overlapping files.
- `make docs-check`: exit 0 (run by the docs-drift pass at this branch). No `.md`/`.yaml` changed, so no prettier run; nothing under `charts/`, so no `make chart-check`. `bash -n hack/check-image-inventory.sh` passes; shellcheck is not on this machine and the script is not in the shellcheck set CI runs.

### Live validation

Not applicable: a repository lint with no runtime component; proven by the macOS before/after transcripts and the tests above. Nothing in this change reaches a cluster, an image, or an install.

## Self-Review

Both passes ran in fresh `general-purpose` subagents handed only the worktree path, the range `upstream/main...HEAD` and the skill file; neither saw the reasoning behind the diff. Both re-ran the fix under BSD sed independently and reproduced the before/after.

Adversarial pass (`review-adversarial`): no BLOCKER or HIGH; three MEDIUM, seven LOW. Angles it reports clean: the pattern itself on BSD sed across quoted, bare, trailing-space, empty and space-containing values; reachability of the guard's "this sed does not accept the pattern" arm under `set -euo pipefail`; the other sed programs, `grep -E`, `sort -V` and the `\t` replacements on BSD tools; that the two per-extractor guards strictly strengthen the removed union guard and nothing else read `$default_images`; no test removed or weakened; constant lifted where the harness needs it; no credential, RBAC, network or workflow change; no-magic-constants placement; red-then-green shown for both the behavioural test and the lint; other open PRs do not touch these hunks. Dispositions:

- M1, the lint read only the first quoted token, so a second `-e` expression, a BSD `-i ''` suffix or an escaped quote in a double-quoted program went unread while the coverage test called the line covered. Fixed: the reader walks every argument and yields one entry per expression; `_ARGUMENT_CASES` pins each shape it named.
- M2, "No CI runner has BSD sed" is false (`installer-matrix-test.yml` has a `macos-latest` job). Fixed the sentence in the docstring and the script comment to what is true: no job runs this script under BSD sed. Deliberately not adding a `make images-check` step to that macOS job in this PR: the workflow is path-filtered to the installer scripts, a new job needs helm and jq on the runner (not verifiable from here) and the `github_actions.md` rules pass, and the lint is the fast Linux signal rather than a substitute for that run. Proposed as a follow-up in the handoff.
- M3, #1455 claims to close the same issue with a different root cause. Measured (Context above): it does not change the macOS result. Cited there; the comment on #1455 is the author's to make.
- L1, `--regexp-extended` counted as extended-and-portable while BSD sed rejects every long option. Fixed: any `--` flag is a finding.
- L2, the command-position finder missed `{ sed`, `then sed`, `do sed`, `xargs sed`. Fixed; `_COMMAND_POSITION_CASES` pins them.
- L3, a `$NAME` that is not a column-0 `readonly` passes unresolved. Deliberately not failing it: a runtime value such as a `read` loop's key carries nothing a text lint can read, and failing it pushes the next author to weaken the lint; the no-magic-constants rule already routes a literal pattern to a `readonly`, which the lint reads. The comment says so.
- L4, the comments said all eleven escapes "matched nothing"; `\b`, `\+` and `\|` match the literal character on this sed. Re-measured and fixed the wording in both constants' comments, the docstring and the assertion message; the classification stands.
- L5, "three of the six images" is seven distinct references. Fixed by dropping the number in the comment and the commit body.
- L6, the lint fired on a program quoted in a comment line. Fixed: comment lines are blanked before the scan, with a case in `_ARGUMENT_CASES`.
- L7, the failure message recommends `\{0,1\}` but the constant test pinned `-E`. Fixed: that test asserts no offending escape for whichever syntax the program uses.
- Not covered by the pass and still open: GNU sed behaviour is unverified on this machine (no `gsed`, no container runtime); stated as the one assumption in the handoff. The guards' `exit 1` is driven by the mutation transcript on macOS rather than by a test, as the old guard was.
- Its side note that `grep -E '\S+'` in check 5 is the same class in another tool: measured working on BSD grep here, and the lint is sed-only by design and says so; not extended in this PR.

Docs-drift pass (`review-docs-drift`): no Blocking; `make docs-check` exit 0 at this branch. Checked consistent: `docker-images.md`'s description of what `make images-check` covers, `testing-map.md` and `AGENTS.md` "Where Tests Go" for the new module's placement and the Makefile glob that reaches it, the docs map (no doc added or moved), every cross-reference to check numbering 3a-3c, the new comments' claims against the file, and the test docstrings that describe one another. Three Advisory: A1 is M2's false sentence (fixed); A2 is L5's count (fixed); A3, the "Three modules" count in `_lift_shell.py` would rot on the next importer (fixed: no number). It also noted, outside its scope, that the issue's "wired into nothing" premise reads as a stale-fork measurement; the Summary and handoff say so.

## Risk & Rollout

Low risk: one shell script that runs in CI and on developer machines, and its tests. The new failure modes are the ones this PR adds on purpose: the lint fails a future GNU-only sed escape or long option in this script, and the field guard fails a render with no `image:` field. On GNU sed the extraction's output is unchanged for every line the chart renders today (all four `image:` lines are quoted, none has trailing blanks). Revert is a single-commit revert with no state to unwind.


bnaylor:
Adopting this one (PR minder session).  I reproduced the whole thing independently against a clean `origin/main` worktree on macOS before touching it, and the diagnosis holds:

- clean main, `bash hack/check-image-inventory.sh` → **exit 1**, `enabling githubMinter added no image...`
- with this diff → **exit 0**, `Image inventory check passed`

The mechanism, isolated:

```
$ printf '  image: "foo/bar:1.0"\n' | sed -n 's/^[[:space:]]*image:[[:space:]]*"\?\([^"]*\)"\?[[:space:]]*$/\1/p'
(0 lines)
```

BSD sed reads BRE `\?` as a literal `?`, so line 247 on main extracted **zero** image refs. `added_images` was then diffing two empty lists — the emptiness came from the inputs, not the comparison.

On the one assumption the body flags as unverified (GNU sed behaviour of the new program): I think that is safer than it reads. The replacement moved to `sed -E`, and `?` is POSIX-mandated in ERE — it is the *BRE* `\?` spelling that was the GNU extension. So the new form is portable by specification rather than by luck. I have no GNU sed on this machine to demonstrate it, but the validate job will, and the guard added at line 444 means a future mismatch fails loudly instead of silently returning an empty set. That guard is the more valuable half of this change.

No action needed from you. The red `Run Python Unit Tests` here is #1364's breakage inherited from main, not this PR — fix is up as #1453.

bradhoekstra (approved):

PR #1471, opened 2026-09-11
chore(agents): let Claude Code load .agents/ skills and rules

haoxuw:
## Summary

Claude Code scans only `.claude/skills/` and `.claude/rules/`, so those two paths become symlinks into `.agents/`, where Gemini CLI and Codex already read the same files. Each rule file gets a `paths` frontmatter so Claude Code loads it only beside matching files.

## Why This Change

In a Claude Code session none of the 25 repository skills registered, and the rule files loaded only when the agent followed the "read the file" instruction in `AGENTS.md`. Without the `paths` scoping, all three rules would load in every session, which the `AGENTS.md` split exists to avoid.

## What Changed

- `.claude/skills` and `.claude/rules` symlinks, plus two gitignore negations (no trailing slash: git records a symlink as a file).
- A `paths` frontmatter on the three rule files. Other tools read them as plain Markdown and ignore the block.

## Context

- #1438 adds a rule file without `paths`; once both merge, it loads in every Claude Code session until it gets a block.
- No `GEMINI.md` or `.gemini/settings.json`: a `GEMINI.md` importing `AGENTS.md` makes Gemini builds that already load `AGENTS.md` load it twice, and the setting makes Gemini load `agents/*/AGENTS.md` as context.

## Testing

- `make docs-check`, `make prompt-check`, `prettier --check` on the rule files: clean.

### Live validation

Not against an installation; nothing reaches the operator or agent pod. Claude Code 2.1.267, headless, with an `InstructionsLoaded` hook: only `CLAUDE.md` and `AGENTS.md` loaded at start, and after reading `.github/workflows/prettier.yml` the hook logged `.agents/rules/github_actions.md` with reason `path_glob_match`. Asked to list its skills, the model named ones that exist only under `.agents/skills/`. Gemini CLI 0.59.0-nightly still finds all 25 skills. Not covered: Codex, Windows checkouts.

## Self-Review

`review-adversarial` (twice) and `review-docs-drift` (three times) ran in fresh subagents over a superset of this diff; the kept hunks are unchanged apart from a shorter comment.

- Claude Code skips a symlinked rules directory when launched from a subdirectory (CONFIRMED). Not fixed: that session gets what `main` gives today.
- Checking out the links replaces an ignored personal `.claude/skills/` or `.claude/rules/` directory without warning (CONFIRMED). Not fixed in code; see Risk.
- Nothing fails if a link dangles, and an unscoped rule file loads in every session unchecked (CONFIRMED). An earlier revision extended `check_context_budget.py` for this; dropped to keep the change small, follow-up if wanted.
- The gitignore comment overstated what Gemini and Codex read natively (CONFIRMED). Fixed.
- Docs-drift: nothing against the kept files. Not covered: Codex, Windows, the site build.

## Risk & Rollout

Low, no runtime paths. Personal files under `.claude/skills/` or `.claude/rules/` in a checkout are lost on the next pull unless moved to `~/.claude/` first. Revert by deleting the two links and the three frontmatter blocks.

### DIFF
--- a/.agents/rules/core_engineering.md
+++ b/.agents/rules/core_engineering.md
@@ -1,3 +1,11 @@
+---
+# Claude Code loads this rule only beside files matching `paths`; other tools ignore this block.
+paths:
+  - "**/*.go"
+  - "**/*.py"
+  - "**/*.sh"
+---
+
 # Core engineering rules
 
 Rules for code written in this repository. [`AGENTS.md`](../../AGENTS.md) names this file and is
--- a/.agents/rules/github_actions.md
+++ b/.agents/rules/github_actions.md
@@ -1,3 +1,9 @@
+---
+# Claude Code loads this rule only beside files matching `paths`; other tools ignore this block.
+paths:
+  - ".github/workflows/**"
+---
+
 # GitHub Actions rules
 
 [`AGENTS.md`](../../AGENTS.md) states both rules below under Pull Request Hygiene. This file holds
--- a/.agents/rules/pre_pr_review.md
+++ b/.agents/rules/pre_pr_review.md
@@ -1,3 +1,10 @@
+---
+# Claude Code loads this rule only beside files matching `paths`; other tools ignore this block.
+paths:
+  - ".github/PULL_REQUEST_TEMPLATE.md"
+  - "docs/pull-request-workflow.md"
+---
+
 # Pre-PR review mechanics
 
 [`AGENTS.md`](../../AGENTS.md) owns both rules below — that adversarial self-review and live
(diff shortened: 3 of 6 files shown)


dshnayder (approved):

haoxuw:
Note for the record: after @dshnayder approved 48779098 I rebased and reduced the change. The current head b56cceec keeps only the two symlinks, the gitignore lines, and the `paths` frontmatter on the three rule files from that commit, and drops the `check_context_budget.py` extension and the docs edits. Nothing new was added. The `lgtm` label carried over; say so if you would rather re-read the smaller diff.

Issue #1557, opened 2026-09-14
build: the five A2A next-stack images are outside images.json and the release pipeline, and the exemption says four


bnaylor:
## What

Follow-on to #1330.  That issue fixed the third-party bases the `a2a/` Dockerfiles build FROM and said in so many words that the published A2A images themselves were out of scope.  This is that other half.

Every image the operator pulls under `spec.mode: next` is outside `images.json`, outside `make mirror-images`, outside `make images-check`, and outside the release pipeline.  The pins live as Go constants with an env var override each:

| Env var | Default on main | Origin | Where it's built |
| --- | --- | --- | --- |
| `A2A_NATS_IMAGE` | `nats:2.10-alpine` | third-party | n/a |
| `A2A_PROVISION_IMAGE` | `natsio/nats-box:0.14.5` | third-party | n/a |
| `A2A_GATEWAY_IMAGE` | `northamerica-northeast1-docker.pkg.dev/bnaylor-kagents-dev/a2a-demo/gateway:latest` | first-party | hand-run Cloud Build, no in-repo config |
| `A2A_WORKER_IMAGE` | `northamerica-northeast1-docker.pkg.dev/bnaylor-kagents-dev/a2a-demo/worker-next:latest` | first-party | same |
| `A2A_CALLOUT_IMAGE` | `us-east4-docker.pkg.dev/bnaylor-kagents-dev/kube-agents/a2a-authcallout:dev` | first-party | same |

The first four are the stage-1 exemption #1211 stated, at `defaultA2AGatewayImage` in `platformagent_a2a_manifests.go` and on the docs site's Docker images page.  The fifth arrived with #1319 and nobody re-counted: the docs page still says "the four the A2A `next` stack pulls" and lists four env vars.  `docker-publish-ghcr.yml` builds seven first-party images and none of the three A2A ones.

## Why it matters

- A mirrored or air-gapped install that flips `next` has nothing to mirror from.  The docs page tells them to "override all four", but `images.json` is the only inventory the mirror tooling reads, so there's no list of what to mirror, and `pulledBy` says nothing about any of these.  (And it's five now anyway.)
- The three first-party defaults point at a personal dev project, in two different regions, on mutable tags.  A `next` install pulls whatever was last pushed there by hand.  That was the stated dev posture and it was fine for the demo, but the mode switch has been on main since #1211 and the constants haven't moved.
[...]

Issue #1330, opened 2026-09-09
build: the a2a gateway and worker Dockerfiles pin base images outside images.json, and `node` is in it under no name


bnaylor:
## What

Two of the three `a2a/` Dockerfiles hardcode their base images instead of taking them from `images.json`, and neither is checked by `hack/check-image-inventory.sh`. One of the four bases they name — `node` — is not in `images.json` under any entry at all.

| File | Builder | Runtime | ARGs from `images.json`? | In `check-image-inventory.sh`? |
| --- | --- | --- | --- | --- |
| `a2a/Dockerfile.authcallout` | `${GOLANG_IMAGE}:${GOLANG_VERSION}` | `${DISTROLESS_IMAGE}:${DISTROLESS_VERSION}` | yes | yes, two lines |
| `a2a/Dockerfile.gateway` | `golang:1.27` | `gcr.io/distroless/static-debian12:nonroot` | no | no |
| `a2a/Dockerfile.worker` | `golang:1.27` | `node:22-slim` | no | no |

`Dockerfile.authcallout` is the shape to copy; it arrives with #1319 and already carries the two `check_base_image` lines. This issue is the other two.

**This is not the stage-1 images.json exemption.** That exemption, stated at `defaultA2AGatewayImage` and on the docs site, is about the four **first-party A2A images** staying out of the inventory because `mode: next` is an unsupported dev toggle. It says nothing about the **third-party bases those images build FROM**, and `Dockerfile.authcallout` shows the two are separable: the callout image is exempt from the inventory while its bases are pinned to it and enforced.

## Why it matters

Three distinct consequences, in rough order of how likely each is to bite:

1. **`node` is mirrorable from nowhere.** `make mirror-images` copies what is in `images.json`. `golang` and `gcr.io/distroless/static` are there; `node` is not, and `gcr.io/distroless/static-debian12` is a *different repository* from the `gcr.io/distroless/static` that is pinned — the note at `Dockerfile.authcallout` says exactly this about that pair. So an operator mirroring into an approved registry has no entry to mirror the worker's runtime base from, and no way to know it is missing, because `pulledBy` is silent about these builds.

[...]

PR #1575, opened 2026-09-14
build(images): put the A2A next stack's two third-party pins in images.json

bnaylor:
## Summary

The two images the A2A `next` stack pulls that this repo does not build, `nats:2.10-alpine` and `natsio/nats-box:0.14.5`, are now `images.json` entries, so `make mirror-images` copies them and `make images-check` holds the operator's compiled defaults to them. The check that did that for the fluent-bit fallback becomes one `check_operator_pin` function called three times, with a lifted-function test beside the others. The Docker images page's exemption paragraph, which said "four" after the auth callout made it five, now names the three first-party images that remain outside the inventory and their env vars, `A2A_CALLOUT_IMAGE` included. This is the half of #1557 that needs no decision; the graduation of gateway, worker and callout stays open there.

## Why This Change

A mirrored or air-gapped install that flips `spec.mode: next` had nothing to mirror NATS and nats-box from: the inventory is the only list the mirror tooling reads, the two pins lived only as Go constants, and nothing checked them, so they could drift from anything else that named them with CI green. The docs page told the reader to "override all four" when there were five env vars. Both are fixed here without deciding anything about the three first-party images, which point at a personal dev project on mutable tags and want a publish decision from the tech lead.

## What Changed

- `images.json`: `nats` (`docker.io/library/nats:2.10-alpine`, override `A2A_NATS_IMAGE`) and `nats-box` (`docker.io/natsio/nats-box:0.14.5`, override `A2A_PROVISION_IMAGE`), `origin: third-party`, `pulledBy` naming the `next`-only StatefulSet and provision Job. Names are the trailing path segment, as the `k8s` entry's description requires.
- `hack/check-image-inventory.sh` check 2: `check_operator_pin <name> <go file> <constant>` replaces the inline fluent-bit comparison and is called for `fallbackFluentBitImage`, `defaultA2ANATSImage` and `defaultA2AProvisionImage`. The capture anchors on a definition line and stops at the closing quote, so a use site, a longer name, or a quoted trailing comment cannot satisfy or break it. Comparison is on the normalised reference, as the fluent-bit check already was.
- The operator constants keep their values. Only the comment block changed: it no longer says none of the A2A images are in the inventory, and it records why the two third-party constants are not qualified to `docker.io/library/...`.
- `docs/site/src/content/docs/deploy/docker-images.md`: the exemption paragraph now names NATS, nats-box, the gateway, the worker and the auth callout, says which three are absent and why (first-party builds published only from a dev registry, off the release pipeline), lists `A2A_CALLOUT_IMAGE`, says the two pins are in the table and that the chart does not set their env vars, and names `a2a/Dockerfile.authcallout` and `distroless-static` in the base-image sentence. The generated table gained the two rows via `make docs-generate`.
- `tests/test_check_image_inventory_operator_pins.py`: lifts `check_operator_pin` from the script and drives it against a synthetic Go file (short and fully-qualified spellings pass, gofmt alignment passes, a quoted trailing comment passes, drifted tag, drifted repository, absent constant and decoy lines fail naming the file, constant and entry), and asserts all three call sites exist in the script.

**Constant qualification, decided against.** The brief was to prefer qualifying the constants to `docker.io/library/nats:2.10-alpine` so they equal the inventory byte for byte. They stay short because qualifying changes the string the operator renders: the NATS StatefulSet's pod template would differ and roll the single-node JetStream pod once on operator upgrade, and the provision Job is named by a digest of its spec (`a2aProvisionJobName`), so a new Job would run for the same image. The live `next` install renders exactly the current defaults (see Live validation), so the roll would be real. The fluent-bit check already compared normalised (`fallbackFluentBitImage = "fluent/fluent-bit:5.1.1"` against `docker.io/fluent/fluent-bit`), so this is the existing standard applied to two more constants, not a weaker one.

## Context

Part of #1557. Related: #1330 (the third-party bases the A2A Dockerfiles build from, which named this as the separable half it was not fixing), #1334, #1211 (the stage-1 exemption), #1319 (added `A2A_CALLOUT_IMAGE`).

Out of scope, deliberately: publishing gateway, worker and authcallout to ghcr, adding them to `docker-publish-ghcr.yml`, giving them `images.json` entries, and moving the operator defaults (and the gateway's own copy of the worker default in `a2a/gateway/config.go`) off the personal dev project. That is the graduation decision #1557 leaves to @bnaylor, so this PR does not close it. What remains on #1557 after this merges is exactly that list. One further follow-up this PR surfaces rather than fixes: the chart does not set `A2A_NATS_IMAGE`/`A2A_PROVISION_IMAGE` when `global.thirdPartyImageRegistry` is in effect, so a mirrored `next` install still points them at the copies by hand; the page says so.

Overlapping open PRs:

- #1559 (`a2a/auth-2-principals`, same author) rewrites the same docs paragraph and the same Go comment above `defaultA2AGatewayImage`, but to say all five images are absent from the inventory. The two texts contradict; this branch's is the one the tree supports after merge. Whichever lands second takes a textual conflict in both files and should keep this PR's wording. #1559 does not rename the two constants, so the new `check_operator_pin` call sites survive either order.
- #1304 (`robot/issue-1284`) edits `images.json` (first-party `pulledBy` strings) and `hack/check-image-inventory.sh` (a new `check_toggle` in check 3). Different lines in both files; textual-conflict proximity only.

**Handoff for the PR minder.** Verified against the tree: `images-check` passes before and after; each of the three constants fails the check when drifted and the message names the file, constant, inventory pin and entry; restoration is byte-identical by sha256; `mirror_images.sh --dry-run` lists both new images under the third-party prefix; the docs table regenerated cleanly; 36 lifted-function tests pass, including the sed-portability lint over the new sed program; the operator suite passes with envtest assets wired. Assumed, not verified: that no consumer iterates `images.json` overrides generically (the installer resolves by explicit name, `scripts/release/common.sh` uses a fixed list; both read, not run). Not live-tested: the change alters no rendered manifest, and the only `next` install is read-only for this session; a read-only observation confirmed it renders the two defaults this PR pins. The #1559 divergence above is the thing to watch at merge time. The graduation half of #1557 (gateway, worker, callout) stays open and needs bnaylor's decision.

## Testing

- `make images-check` on unmodified `upstream/main` (`2f84aa23`): passes. On this branch: passes.
- Mutation proof, re-measured after the review changed the sed program: `defaultA2ANATSImage` to `nats:2.11-alpine` fails with `defaultA2ANATSImage is 'nats:2.11-alpine', but images.json has 'nats:2.10-alpine' for 'nats'`; `defaultA2AProvisionImage` to `0.14.6` fails naming `nats-box`; moving the `images.json` pin to `2.9-alpine` instead fails naming the constant; a quoted trailing comment on the const line passes. Each restore verified by sha256 (Go file) and `git diff --quiet` (`images.json`).
- `python3 -m unittest` over the four `tests/test_check_image_inventory_*.py` modules: 36 tests OK (pytest is not installed here; `make test-python` uses `unittest discover`, and `tests/test_*.py` is in `PYTHON_TEST_DIRS`).
- `make -C k8s-operator test` with `KUBEBUILDER_ASSETS` exported, on the head `720a7727`: passes (controller package 91.9% coverage, envtest cases ran); the Go change is comment-only and `gofmt -l` is clean.
- `make docs-generate` then `make docs-check`: generated region current, links, terminology, map and context budget pass.
- `prettier@3.9.6 --check` on the changed page: clean.
- `MIRROR_PREFIX=... MIRROR_THIRD_PARTY_PREFIX=... scripts/mirror_images.sh --dry-run`: plan carries `docker.io/library/nats:2.10-alpine -> <third-party>/nats:2.10-alpine` and `docker.io/natsio/nats-box:0.14.5 -> <third-party>/nats-box:0.14.5`.
- Drift check: 0 commits behind `upstream/main`, no overlapping files changed since the merge-base.

### Live validation

Not live-tested. The change renders nothing new: the constants keep their values, so a running `next` install sees no pod-template change, and the only `spec.mode: next` install (`a2a-next-dev`) is read-only for this session. What was checked there, read-only: the NATS StatefulSet runs `nats:2.10-alpine` and the three provision Jobs run `natsio/nats-box:0.14.5`, with no `A2A_NATS_IMAGE`/`A2A_PROVISION_IMAGE` set on the controller manager, so the inventory entries name exactly what a `next` install pulls today, and the digest-named Jobs are the evidence that a qualified constant would have minted another. The mechanism proof is the mutation table above, run from the worktree root.

## Self-Review

Both passes ran in fresh `general-purpose` subagents handed only the worktree path, the range `upstream/main...HEAD` and the skill file.

Adversarial (`review-adversarial`): looked for removed behaviour, symbol traces, regex edge cases, generic `override` consumers, count claims against source, the `head -n1`/`pipefail` interaction, gofmt, red-then-green of the new test (every case fails against `upstream/main`'s script, where `check_operator_pin` does not exist, and passes here), and sibling PRs.

- MEDIUM, #1559 carries a contradicting rewrite of the same paragraph and comment: confirmed; not a code change here, recorded under Context with the resolution to apply at merge.
- LOW, the exemption's base-image sentence named two Dockerfiles once the callout joined it: fixed, `a2a/Dockerfile.authcallout` and `distroless-static` added.
- LOW, greedy capture would read a quoted trailing comment into the value: fixed, `[^"]*`, plus a test row and a mutation run for the case.

Docs-drift (`review-docs-drift`): searched the tree for the old "four" wording, every `A2A_*_IMAGE` name, "fluent-bit is the only compiled pin", the docs map, the testing map, chart and terraform READMEs, installer prose.

- Blocking, the "inventory documents what a supported install pulls" rationale contradicted putting `nats` in it: fixed in the page and the Go comment; the discriminator is now that the three absent images are first-party builds published only from a dev registry, off the release pipeline.
- Blocking, "the three pins baked into the operator binary" undercounted the six compiled image defaults: fixed to "the inventory pins baked into the operator binary", with the first-party `next` defaults named as unchecked.
- Advisory, hard-coded counts as a drift magnet: fixed where the list is spelled out (the page now names the five and says "the last three").
- Advisory, an issue link on a site page: fixed, the page says the decision is open without linking #1557.
- Advisory, check-2 header implies exhaustiveness: fixed with a sentence naming the unchecked first-party defaults, and the same in the test docstring.
- Advisory, `terraform/examples/full-install/README.md:372` "the two images the chart never renders": not changed. That sentence describes what `image_registry` reaches on a default install, and the two new entries are reached by no prefix at all, which the images page now states; widening this PR into the terraform README for a sentence it does not make false is scope creep.

## Risk & Rollout

Low. No rendered manifest changes: the two constants keep their values, so no `next` install rolls anything on operator upgrade, and default-mode installs are untouched. `make mirror-images` copies two more images, which is the point. The new check can only fail CI when a constant and its entry disagree, and the message says which. Revert is a plain revert of one commit. Textual conflicts with #1559 and #1304 are noted above; the #1559 one needs the resolution to keep this PR's text.


bradhoekstra (approved):

bnaylor:
Context for the retest above, since this is the second smoke failure on this PR and they are unrelated to each other.

Run 1 (`2099547...`) never got past the deploy: `platform-agent-gateway` rollout timed out with one of thirteen containers never Ready, and that container's stdout stops at `Mirroring the profile layout into the shell sandbox...`.  That string does not exist anywhere in kube-agents -- it is the vendored hermes image.

Run 2 (`2099552760360865792`) deployed fine and ran the whole eval suite.  It went RED on one repetition of one case:

```
security-overgrant-remediation-proposal: rung 3 (NOT_A_REAL_RUN) --
repetition 1: the trajectory is empty: the agent made no tool calls
```

rep 1 spent 9.2s in agent execution.  reps 2 and 3 spent 394s and 554s and both passed.  An empty trajectory after nine seconds is the agent never starting, not a verdict on the change.  I scanned the last 45 smoke builds: sixteen produced an eval verdict, fifteen GREEN, and this is the only rung-3 in the set.

The diff here is comments, docs, a shell function in `hack/check-image-inventory.sh`, and two additive entries in `images.json`.  No Go behavior changes -- the `images.json` hunk has zero deleted lines and the operator renders the same image strings it did before.  Neither failure has a path to this diff.

Retesting once.  If run 3 reproduces either shape I will stop and file it rather than keep burning the pool.

bnaylor:
Merged main in.  The batch that merged #1458 and friends took `k8s` from `1.36.2` to `1.36.4`, and the only conflict was that row of the generated table in `docs-images.md`, sitting directly under the two rows this PR adds.  Kept both new rows and took main's `1.36.4`.

Verified rather than assumed:

- `make docs-generate` regenerates the table byte-identical to the resolution, so the file is what the generator would produce and not a hand-merge that happens to look right.
- `make images-check` passes.
- The five lifted-function tests pass.
- `go build ./...` and `go vet ./internal/controller/` are clean.
- `git merge-tree --write-tree HEAD upstream/main` is clean, so no conflict remains.

`images.json` and `platformagent_a2a_manifests.go` auto-merged -- main's changes there were the `k8s` tag bump and #1458's provision-Job sweep, neither of which touches what this PR does.  The `check_operator_pin` call sites and the constant comment block both survived intact.

The #1559 divergence the PR body flags is still ahead of us: that PR is green and queued, and whichever of us lands second takes the textual conflict in the same two files.

bnaylor:
Merged main in again, now that #1559 has landed (03:42Z).  This is the conflict the PR body predicted, and I landed second, so this branch's wording wins in both files.

Two conflicts, plus one thing that was not a conflict and mattered anyway.

**`docker-images.md`, the exemption paragraph.** Main now says the five `next` images are deliberately absent and a mirrored install has to override all five.  That is false after this PR -- `nats` and `nats-box` are inventory entries.  Took this branch's text: three absent (gateway, worker, authcallout), two in the table.

**`platformagent_a2a_manifests.go`, the comment above the A2A constants.** Same contradiction, same resolution.  Kept this branch's "the first-party A2A images -- this one, the worker below and the auth callout".

**The part that was not about #1559.** Main also restructured that same paragraph block for unrelated reasons: the three "a bump starts here but rarely ends here" paragraphs were replaced with a short one pointing at `docs/pull-request-workflow.md`.  This branch had edited one sentence inside the paragraphs main deleted.  Resolving that hunk in this branch's favour would have resurrected three paragraphs main deliberately consolidated, so I took main's there.  Nothing is lost: the sentence this branch added said the operator-baked inventory pins now include NATS and nats-box, and main's replacement covers the same ground generically with "a compiled constant in the operator".  Also took main's fluent-bit `5.1.2`.

Verified:

- `make docs-generate` reproduces the table byte-identical, so the generated block is the generator's output and not a hand-merge.
- The paragraph's claim about the build-time bases checks out against the regenerated table: `golang` carries all three A2A Dockerfiles, `node` the worker, `distroless-static` the authcallout and gateway.
- `make images-check` and `make docs-check` pass -- including the always-loaded budget, now 37.8k with 0.2k of headroom (main's consolidation bought some back).
- Prettier clean, `gofmt` clean, `go build ./...` and `go vet ./internal/controller/` clean, the five lifted-function tests pass.
- `git merge-tree --write-tree HEAD upstream/main` is clean.

Head is `ed3ccea7`.  The smoke that failed at 12:52Z was the clonerefs merge conflict, not a real failure; this push restarts it.  For the record the last two real runs on this branch were both green (151 min and 209 min).

PR #467, opened 2026-07-29
docs(drift): add Phase 0 spike findings and attribution tooling

jayantid:
## What

Adds `agents/platform/docs/drift-detection/` with the Phase 0 spike results and the tooling that produced them, so they live in the repo rather than on a laptop.

| File | What it is |
|---|---|
| `README.md` | Folder index, the four findings that shape the build, and the caveats on the measurements |
| `spike-findings.md` | Spike A (attribution) and Spike B (noise profile). Verdict: **GO** |
| `drift_attribute.sh` | Reusable join: `managedFields` + audit log for one namespace |
| `baseline-manifests.yaml` | Baseline manifests used to stage the spike |
| `cuj3-task-breakdown.md` | Ticket for CUJ 3 (noise-filtered triage), the enabling implementation the other CUJs are blocked on |

## Why

The spike validated the two-signal design and produced numbers the implementation will be measured against, but the artifacts only existed in `/tmp` and a local file. This is the material a new engineer needs on day one.

## The findings that matter

- **Audit log is the primary trigger; `managedFields` is enrichment.** Inverting the order produces false positives on a real cluster.
- **The audit principal is required, not optional.** CI applying client-side records `kubectl-client-side-apply` — byte-identical to a human running `kubectl apply`. A `managedFields`-only detector flags every CI deploy as drift.
- **Two static filters remove ~99% of noise.** Drop `^system:` principals, then a configurable automation-SA allowlist. That allowlist is per-cluster config and the most important tuning knob in the system.
- **Post-filter volume is ~7 human changes/day**, so per-change agent judgment is affordable.

## Caveats, stated in the docs

Measured on a control-plane/admin cluster with no in-cluster GitOps controller; the GitOps owner was simulated with `--field-manager=argocd-controller`. The census should be re-run on a production app cluster before filter defaults are finalized.

Per-user change counts from the audit census are reported unattributed, since this repository is public.

Companion to #444, which adds the design doc at `agents/platform/docs/drift-detection.md`.

PR #1594, opened 2026-09-15
feat(drift-detector): pull and parse GKE audit records from Pub/Sub

stalhaali:
## Summary

Adds the drift detector's ingestion path: a new `k8s-operator/cmd/drift-detector` binary that pulls GKE audit records from the Pub/Sub subscription `terraform/modules/drift-pubsub` provisions, parses them, and writes each one to its structured log. This is T1 of CUJ 3 and it stops deliberately short of doing anything with what it reads — classification, the `managedFields` join, and the inject are T2 to T4, and the handler signature in `subscriber.go` is the seam they plug into. Nothing builds or ships this binary yet; it is not in the image and not launched by `start-services.sh`, so merging it changes no running behaviour.

## Why This Change

Kubernetes audit is not served by the Kubernetes API on GKE. The control plane is managed, so the API server's audit backend is not ours to configure and the stream surfaces only in Cloud Logging. PR #615 built the route out of Cloud Logging and into a Pub/Sub subscription; until something pulls that subscription, the messages accumulate and expire against the 31-day retention ceiling. This is the consumer.

It is a separate binary rather than another source inside `k8s-event-watcher` because the two read different transports — an informer against the Kubernetes API versus a Pub/Sub subscription — and a separate route on the gateway's Python Pub/Sub adapter because that adapter's filter expression language supports only `==` and `!=`, while T2 needs prefix and regex matching on the principal. The placement argument and its counter-costs are on #911.

## What Changed

- `resourcename.go` — decomposes an audit `resourceName` into group, version, namespace, resource, name, and subresource, which is what T3's live-object lookup needs. Two shapes do not read off the grammar directly and are handled explicitly: the namespace object itself (`core/v1/namespaces/foo`, where the `namespaces/<ns>` prefix is the object rather than the scope) and a create, whose name the API server has not assigned when the call is audited.
- `audit.go` — decodes the Cloud Logging entry into an `AuditRecord` carrying the five fields T1 is accepted on plus the cluster, project, and location. Only the fields the detector reads are decoded; the full entry carries request and response bodies that can run to megabytes.
- `subscriber.go` — the pull loop, and the settling rules described below.
- `main.go` — flags, signal handling, and a `realMain` split so the startup path is testable.
- `README.md` — how to run it, and the two things worth knowing before changing it.
- `docs/designs/drift-detection.md`, `docs/README.md` — the status banner, the ingestion paragraph, and the map row, all of which said no drift code existed.
- `terraform/modules/drift-pubsub/` — README, `outputs.tf`, and two `main.tf` comments brought in line with what the consumer actually does.

Two decisions worth a reviewer's attention:

**Settling is three-way, not two.** A record that parses is handled and acked. One that is understood but not actionable — another service's audit entry, or a call that named no object, such as a subject access review — is acked and dropped, because leaving it would redeliver it until retention expired; a per-batch count goes to the log, so a sink whose filter stopped matching does not read as a quiet cluster. Anything else is nacked, so a payload-shape change on Google's side is loud rather than silently acking drift away, which is what `terraform/modules/drift-pubsub/main.tf` says the detector does.

The edge worth stating: `json.Unmarshal` zeroes what it cannot match, so a renamed or restructured `protoPayload` decodes without error and arrives looking exactly like an empty one. That is treated as a parse failure, not as "not ours" — the review pass found it going the other way, and it would have acked every message on the subscription while logging nothing.

**No new dependency.** It uses `google.golang.org/api/pubsub/v1` rather than `cloud.google.com/go/pubsub`, because `google.golang.org/api` is already a direct dependency of this module for the GKE Container API the event watcher calls. `go.mod` and `go.sum` are unchanged. The cost is synchronous pull instead of StreamingPull: no automatic ack-deadline extension and no built-in flow control. Neither binds while the ack happens on parse — a sub-millisecond step against the subscription's 60-second deadline — and the post-sink stream measured about 0.7 messages a second. `subscriber.go` records this, and says to revisit it if T3 ever moves the ack to after the `managedFields` join.

## Context

Refs #911 (the whole detector; this is the first of its tasks, so not `Closes`).

Builds on #615, which merged the `drift-pubsub` Terraform module. Design: `docs/designs/drift-detection.md`. The CUJ 3 task breakdown lives on #467, still draft, so it is not on `main` yet.

Placement — Go binary versus a route on `agentplugins/pubsub-platform` — was proposed and argued on #911 before this was scaffolded, because the design doc pointed at the adapter while the task breakdown pointed at a Go sibling. The pre-PR docs-drift pass ruled that leaving the contradiction in place was Blocking, so this PR resolves it: `docs/designs/drift-detection.md` now names the Go binary and gives the reason, and its status banner and the docs-map row say what is actually built rather than that nothing is.

No competing work: no open PR touches `k8s-operator/cmd/drift-detector`, and #467 is the spike artifacts by another author.

## Testing

- `go build ./...` and `go vet ./...` in `k8s-operator/` — both clean. `go vet` compiles the test files too, so the tests are known to type-check.
- `make docs-check` — passes.
- `go.mod` and `go.sum` unchanged, confirming the no-new-dependency claim.
- Four test files covering the `resourceName` grammar (including the two ambiguous shapes and a namespace literally named `namespaces`), the audit-entry decode with its skip paths and its unrecognisable-payload paths, the loop's ack/skip/nack branching, backoff, and settling-after-cancellation against a fake message source, and the flag parsing with both accepted `--subscription` forms.

**The tests have never been executed.** The authoring environment SIGKILLs executing binaries — `go test` and `gofmt` both die with exit 137 regardless of what is under test — so CI is the first thing to run them. That is stated here rather than left for a reviewer to discover from a red check.

### Live validation

Not live-tested, and it could not be. The binary is not wired into `deploy/docker/Dockerfile` or `deploy/shared/start-services.sh` by this PR, so no installation runs it. The subscription it reads also does not exist in any project: `drift-pubsub` is referenced nowhere outside its own module directory, so it is not part of `terraform/examples/full-install` and no install creates it. And the authoring environment cannot execute the binary locally either, for the SIGKILL reason above.

What that leaves unproven is the part no unit test reaches: that a real `kubectl patch` arrives on the subscription and appears in this binary's log within 90 seconds, which is T1's acceptance criterion. Doing it needs the module applied against a project with a GKE cluster — `project_id` and `detector_service_account_email` are its only required inputs — and the binary run against the resulting subscription. That is the first thing to do when the shipping wiring lands, and the result belongs on the PR that adds the wiring rather than being claimed here.

No test artifacts were created, so there is nothing to clean up.

Eval-driven development ([`.agents/rules/eval_driven_development.md`](https://github.com/gke-labs/kube-agents/blob/main/.agents/rules/eval_driven_development.md)): exempt. Nothing here is on a runtime path an agent takes — the binary is not built into any image, not started by any entrypoint, and not reachable from a running install — so no bench case can observe the change, red or green. The eval loop is owed by the PR that wires the detector in, alongside the live validation above.

## Self-Review

Two passes ran, each in a subagent handed the diff range and the skill and nothing else — no plan, no commit messages, no intent sentence. Neither context wrote the change.

- **`review-adversarial`** — 3 HIGH, 4 MEDIUM, 6 LOW. Angles covered: correctness against the audit-log grammar, error and settle paths, concurrency and shutdown, blast radius, dependency choice, test adequacy, docs consistency. It confirmed the parts I most wanted tested: `parseResourceName` is sound on every shape it tried including both ambiguous ones, base64/`StdEncoding` is the right decode, the REST-client choice holds, and the diff is additive with nothing reachable from a running install.
- **`review-docs-drift`** — 4 Blocking, 7 Advisory.
- **`make docs-check`** — passes (the mechanical subset: generated regions, links, terminology, map coverage, context budget).

Deduped, the two passes found the same defect twice in two places, so the merged list is shorter than 11 + 4.

**Fixed:**

| Finding | Pass | Verdict | Fix |
| --- | --- | --- | --- |
| `--subscription` double-prefixed an already-qualified value, so following the `drift-pubsub` README pulled a subscription that does not exist — and an empty subscription reads exactly like no drift | both | CONFIRMED / Blocking | `subscriptionPath` accepts either form, as `adapter.py:234` already does. README and `outputs.tf` say so; a test pins both |
| An unrecognisable payload was acked and dropped with no log line. `json.Unmarshal` zeroes a missing `protoPayload`, so a restructured payload decoded cleanly with `serviceName == ""` and took the "not ours" path — every message silently acked, which is what the nack path exists to prevent | adversarial | CONFIRMED | Empty `serviceName` is now a parse failure and nacks. The remaining skip path logs a per-batch count |
| `drift-detection.md`'s banner said no drift code exists, and `:41` pointed the ingestion at the Python adapter; `docs/README.md:345` repeated the banner | both | CONFIRMED / Blocking | All three corrected, plus the placement argument written down where the contradiction was |
| Shutdown cancelled the in-flight batch's acks. `signal.NotifyContext` gives no grace period, so the same cancelled context aborted `Ack`/`Nack` — a handled batch redelivered, which at T4 is a duplicate inject per restart. The comment claimed the opposite | adversarial | CONFIRMED | `settleContext` runs settling on `context.WithoutCancel` plus a 10s bound; a test asserts the calls see an uncancelled context |
| `AckDeadlineSeconds: 0` dropped by `omitempty`, so a nack kept the subscription default instead of redelivering | adversarial | PLAUSIBLE | `ForceSendFields`. Fixed despite the softer verdict: one line, and the generated-client behaviour is documented |
| `logRecord` dropped `Project` and `Location` — the exact fleet ambiguity the struct's own comment warns about | adversarial | CONFIRMED | Both logged |
| No `main_test.go`; `parseFlags` and the whole of `pubsubSource` untested — which is where the first and fifth findings above lived | adversarial | CONFIRMED | Added, covering flag defaults, each rejection, and both subscription-path forms |
| `--max-messages` unvalidated and the startup line reported the raw flag rather than the resolved path | adversarial | LOW | Validated against the API's 1000 ceiling; the line logs what is actually pulled |
| `main.tf:153` justified the `roles/pubsub.viewer` grant by a `Subscription.Exists` call the REST client never makes | docs-drift | Advisory | Comment corrected. The grant stays, for `gcloud ... describe` — reasoning now in the file |
| `resourcename.go:81` cited a task breakdown that is not on `main` | docs-drift | Advisory | Reworded to argue the point itself |
| `logEntry.LogName` decoded and never read; `logResource.Type` likewise; unnamed `make([]string, 0, 4)` | adversarial | LOW | `LogName` dropped, `Type` now checked against the sink filter, capacity named |
| `T1`–`T4` staging vocabulary appeared only in Go comments, and `drift-detection.md`'s first mermaid reused `T1`/`T2` for two different things | adversarial | LOW | Mermaid ids renamed; the staging is stated in the design doc and the new README |
| No README for the new command | both | LOW / Advisory | Added, with a docs-map row |

**Not fixed, deliberately:**

- **No metrics or health surface** (adversarial, LOW). The sibling event watcher has `metrics.go`; this has `subscriberCounts` and a log line. Nothing scrapes this binary because nothing runs it, so a metrics endpoint now would be an unexercised guess at what to export. It belongs with the wiring PR, which is also where a health probe first has a consumer.
- **Nothing builds or ships the binary** (adversarial, LOW). Intended, and stated in Summary and Risk & Rollout. Splitting ingestion from deployment is what keeps this diff reviewable against the audit-log grammar rather than against Dockerfile layers.
- **`AGENTS.md`'s repository-layout line does not mention either `cmd/` binary** (docs-drift, Advisory). True, and pre-existing — the event watcher is missing from it too. `docs-check` reports the always-loaded files at 38.0k against a 38,000-char budget with 0.0k of headroom, so adding a clause means taking one out elsewhere. That trade is its own change with its own argument, not a rider on this one.
- **Entry-point listings elsewhere omit the binaries** (docs-drift, Advisory). Same pre-existing shape as above; fixing it means editing docs this change does not touch to correct an omission it did not introduce.

**What the passes could not cover:**

- **No test execution, by either pass or by me.** This environment SIGKILLs executing binaries — `go test` and `gofmt` die with exit 137 regardless of what is under test. `go build ./...` and `go vet ./cmd/drift-detector/` are clean, and `go vet` type-checks the test files, so the tests are known to compile and nothing more. CI is the first thing to run them.
- **No live Pub/Sub.** Neither pass could exercise `pubsubSource` against a real subscription, so the REST call shapes are verified against the generated client's signatures and not against the service. The `ForceSendFields` finding is exactly the class of defect that leaves — hence the PLAUSIBLE verdict on it.
- **Angle J of `review-adversarial` (prior review history) did not run** — no `gh` credentials in the subagent's context.
- **Neither pass saw this PR body**, which did not exist when they ran. Its claims are mine.

## Risk & Rollout

Low. Nothing builds, ships, or launches this binary: it is a new package under `k8s-operator/cmd/` that no Dockerfile stage compiles and no supervisor starts, so merging changes no running behaviour and there is nothing to roll back beyond reverting the commit. It does add a package to `go build ./...` and `go test ./...`, so a compile or test failure here reds the operator's CI job. The credential it will eventually need — `roles/pubsub.subscriber` on the subscription — is already granted by the merged `drift-pubsub` module and is not touched here.

---


fatoshoti (approved):

k8s-operator/cmd/drift-detector/subscriber.go
```diff
@@ -0,0 +1,392 @@
...
+	return &pubsubSource{
+		service:      service,
+		subscription: subscriptionPath(project, subscription),
+	}, nil
+}
+
+// subscriptionPath qualifies a bare subscription id with its project, and
+// passes an already-qualified one through. Both forms reach this binary: the
+// drift-pubsub module's README hands the operator its subscription_id output,
+// which is fully qualified, while the default and the Helm values carry the
```
jayantid: 🟡 **UX flow.** The comment at `subscriber.go:132` says the Helm values carry the bare subscription name, and no chart value, CRD field, or entrypoint variable for the drift detector exists.

Nothing under `charts/kube-agents/` or `k8s-operator/api/` mentions drift. The only operator-facing surface this PR names is the three flags in the new README.

The next reader is whoever writes the wiring PR. They will look for the chart plumbing this comment describes and not find it, and the design doc gives them no answer either. The sibling has a settled shape: the entrypoint sets the flags, the operator passes per-install values as env vars on the container, and `spec.harness.eventWatcher.enabled` is the emergency stop.

Drop the Helm clause. Then add one line to the README's "What ships today" naming the intended surface, for example that project and subscription arrive as env vars and the stop mirrors `spec.harness.eventWatcher`.

PR #1629, opened 2026-09-16
feat(drift-detector): classify principals and drop calls that changed nothing

stalhaali:
## Summary

CUJ 3 task T2. T1 ([#1594](https://github.com/gke-labs/kube-agents/pull/1594)) pulls GKE audit records off the Pub/Sub subscription and parses them; this classifies each one and forwards only the records that represent a change a person actually made. Every record is assigned a tier — `system`, `automation`, `human`, or `unattributed` — and a record reaches the downstream handler only if it is human, the call succeeded, and the call actually changed declarative state. Three of those design points depart from the CUJ 3 task breakdown, and all three come from measuring live audit logs across three projects rather than from reasoning about the spec.

## Why This Change

Without classification every audit record looks like drift. On the projects measured, over 99% of mutating audit records are machine traffic — controller lease renewals, the node bootstrapper, GKE's own service accounts — so a detector that injected on all of them would bury the handful of real human changes it exists to find.

The breakdown's classifier is what stays broken without this. It specifies three tiers: drop `^system:`, drop an automation allowlist, and treat the residual as human. Measured over 24 hours, that residual was **1,230 calls and every single one was a machine.** Two classes account for all of them: `kubelet-nodepool-bootstrap`, which carries neither the `system:` prefix nor a service-account suffix, and unauthenticated requests with an empty principal — public GKE endpoints get crawled, and a rejected bot probe is still an audit entry with a mutating `methodName`. Shipping that rule means 1,230 injects a day claiming a person changed the cluster when nobody did.

An allowlist cannot close the gap, because it has to anticipate every identity GKE invents. So the human test here is **positive** — the principal carries a domain, and one of `--human-domains` when that flag is set — and anything unmatched becomes `unattributed` and is logged by name at shutdown. An unknown principal is then loud instead of silently wrong, and the log gives an operator the names to write the next rule from.

The second departure the breakdown does not mention at all: **the audit log records attempts, not changes.** A write rejected by RBAC, refused by an admission webhook, or aborted after losing an optimistic-concurrency race is an entry with a mutating `methodName` that changed nothing. About 5% of the human calls measured over 30 days failed, including `PERMISSION_DENIED`, and a change someone was actively stopped from making is the clearest possible case of something that is not drift.

The third is the one I would most want a reviewer to check: **a mutating verb in the `methodName` is not a mutating call.** `kubectl exec` is audited as `io.k8s.core.v1.pods.exec.create` — the string contains `create`, so it clears the sink's `create|patch|update|delete` filter — and it names a real object (`.../pods/<name>/exec`), so the parser's existing "named no object" drop does not catch it either. Left alone, a person exec-ing into a pod is classified `human`, succeeds, and is reported as a change they never made. `pods.exec.create` is present in the live Admin Activity log, so this is measured rather than hypothesised. `nonDeclarativeSubresources` drops the six subresources that arrive this way — `exec`, `attach`, `portforward`, `proxy`, `ephemeralcontainers` (`kubectl debug`) and `token` (`kubectl create token`) — after classification, counted as `non_declarative` so the drop is visible instead of silent. `ephemeralcontainers` is the one that genuinely mutates the stored object and is still not drift: the Git-side object is the Deployment that owns the pod, which is unchanged, so there is nothing to revert or codify. `status`, `scale` and `eviction` are deliberately *not* in the set — the first two are real declarative writes, and `eviction` is deferred to T3, where the `managedFields` join can say whether it mattered.

## What Changed

- `classify.go` (new) — the four tiers, the `Classifier`, and `driftFilter`, which counts every tier and forwards only successful human records. Rule order is deliberate and tested: `system:` prefix, then configured automation, then the service-account suffix, then the positive human test, then `unattributed`.
- **The service-account match is the whole `.gserviceaccount.com` domain, not the `.iam.` form.** The Google-managed accounts carry no `iam` label — `<number>-compute@developer`, `@cloudbuild`, `@appspot`, `@cloudservices` — so matching `.iam.gserviceaccount.com` sends every one of them to the positive human test, which they pass on the strength of having an `@`. A Cloud Build pipeline deploying to a cluster, precisely what the automation tier exists for, would be reported as a person making an out-of-band change. The leading dot is what stops the suffix matching a lookalike domain such as `evilgserviceaccount.com`; both cases are in the table test. The match is also case-folded, as `--human-domains` is, because both are DNS domains — while `--automation-principals` is matched exactly, because a Kubernetes username is case-sensitive by specification. `--human-domains` matches the exact domain rather than its subtree, so subdomains are listed separately; a leading `@` or `.` on a configured value is stripped, so `.example.com` is the same configuration as `example.com` rather than one that quietly matches nothing.
- `nonDeclarativeSubresources` in `classify.go` — the six-entry set above, applied last because it is the narrowest test and the only one that has to run on the human path alone. A server-side dry run belongs to the same category and is **not** handled: no `dryRun` marker appeared in seven days of Admin Activity logs across three projects, which leaves open whether GKE surfaces one at all, so the first step there is establishing the payload shape rather than writing a rule for a shape nobody has seen.
- `audit.go` — parses `google.rpc.Status` off the payload and adds `AuditRecord.Succeeded()`. proto3 omits zero values, so a successful call may send no `status` field at all while gcloud's JSON rendering materializes `"code": 0`; both decode to the same Go zero value, which is why no pointer is needed.
- `main.go` — three flags: `--automation-principals`, `--human-domains`, `--log-dropped`. Flags rather than a mounted config file because that is the path configuration already takes to a sibling adapter — the PlatformAgent CR sets an environment variable and `deploy/shared/start-services.sh` turns it into a flag, exactly as `EVENT_WATCHER_*` does for `k8s-event-watcher`. Both satisfy T2's "changes without a rebuild"; only one matches the deployment path the operator already reconciles.
- `testdata/` (9 fixtures) — eight captured from live Cloud Audit Logs with identifiers substituted and long `request`/`response` bodies replaced by a stub marked `_trimmed`; nothing else is edited. `"authenticationInfo": {}` is not a shape anyone invents at a keyboard, and it is the shape that justifies the fourth tier. The ninth, `human_exec.json`, is **derived rather than captured**, and the file header and README both say so — the live query proved `pods.exec.create` reaches the Admin Activity log but did not hand me an entry I could ship, so a captured envelope's method, permission and `resourceName` were replaced with the exec form. It is the only fixture carrying a subresource in its `resourceName`, which is the thing the subresource rule turns on, so it is worth replacing with a real capture.
- `subscriber.go` — the filter is wired in behind the existing `recordHandler` seam, and the loop reports an idle line when the subscription has delivered nothing for fifteen minutes. Between that and the filter's progress line, the pod logs something on the same bound in every state it can be in: an empty `Pull` is not an error, so without the idle line a Log Router sink whose filter stopped matching produced no output at all and read exactly like the steady state nobody investigates.
- `README.md` — the classification rules, the measurement behind them, and the tier ratios actually observed.

Counting still happens before filtering, so a cluster whose human changes are all being denied does not read as a cluster with no human changes. `TierCounts.String` always prints every tier including the zeroes, for the same reason: human traffic is bursty, and a quiet day must not look like a broken rule.

One configuration note a reviewer should hold onto: the Log Router sink is **project-wide**, so a single detector instance sees every cluster the subscription carries. `--automation-principals` and `--human-domains` are therefore per detector instance, not per cluster.

## Context

Part of #911 (`feat: Create Drift Detector`), task T2 of 4. T1 merged as #1594; T3 (the `managedFields` join) and T4 (the inject) follow. `recordHandler` in `subscriber.go` and `driftFilter.next` are the seams they attach to.

Two related open pull requests, neither overlapping a file with this one:

- **#1596** composes `terraform/modules/drift-pubsub` into `full-install`. That is the runtime dependency for this binary ever having a subscription to read, but it touches only `terraform/`.
- **#467** adds the Phase 0 spike findings and the CUJ 3 task breakdown under `agents/platform/docs/drift-detection/`. Worth a reviewer's attention: it is the document that predicts the 78% / 20% / 1% tier split and specifies the three-tier classifier, and the measurement recorded here contradicts both. It has not merged, so nothing on `main` states those numbers yet and no doc goes stale on merge — but if #467 lands after this, its spike numbers and this README will disagree in the tree. Flagging rather than resolving, since #467 is not mine.

## Testing

- `go build ./...` and `go vet ./...` (which compiles the test files) are clean across `k8s-operator/`.
- **43 top-level tests pass in `cmd/drift-detector/` (120 including subtests), 0 failures**, three runs in a row and clean under `-race`, and `gofmt -l` is clean for that directory. `-race` matters now that `subscriber.Run` is exercised in a goroutine; the log buffer in those tests is mutex-guarded for the same reason. These were run on a separate Linux machine, not in the authoring environment: compiled Go test binaries **and** `gofmt` are SIGKILLed there, so `go test` reports `signal: killed` regardless of the code. The same limitation applied to #1594.
- **Every test added here was mutation-checked** rather than merely observed passing, because a test that cannot fail is worse than no test. Sixteen mutations across three rounds, each reverted after. The seven covering the review fixes, with the test each one reds: un-folding the `.gserviceaccount.com` suffix reds the two mixed-case cases; dropping the leading-dot strip reds the two `.example.com` cases; reverting `isHuman` to `strings.Contains` reds all three malformed-principal cases; removing the idle report reds `TestRunReportsAnIdleSubscription`; firing it on every empty poll reds `TestRunDoesNotReportIdleBeforeTheInterval`; and both never resetting `lastDelivery` after a delivered batch and resetting it before `processBatch` instead of after red `TestRunIdleIntervalRestartsAfterADelivery`. The last two are the ones worth noting: the first pass of these fixes had no test for the reset at all, the mutation found that, and the test came out of it. The original nine: widening `gcpServiceAccountSuffix` back to `.iam.` reds the four Google-managed service-account cases; transposing the two `NewClassifier` arguments reds `TestNewFilterFromFlagsWiring` on all four assertions; removing entries from `nonDeclarativeSubresources` reds the subresource test; dropping `strings.ToLower` in `NewClassifier` reds the case-folding test; dropping the progress line's time bound reds `TestDriftFilterProgressLineIsTimeBounded`; dropping `strings.ToLower` on the *principal* in `isHuman` reds the case-folding test from the other side; reverting `%q` to `%s` in `Sorted` reds all three `TestUnattributedPrincipals*`; breaking the `json:"message"` tag reds `TestClassifyCapturedFixtures`; and dropping `ref.Subresource` from `parseResourceName`'s three-segment tail reds both `TestClassifyCapturedFixtures` and `TestParseResourceName`. The argument transposition is the interesting one — it compiles, vets clean, and leaves every test that builds a `Classifier` directly still passing, so the flags-to-filter wiring test is the only thing that sees it.
- **One fix in this change is confirmed to be untested, and I am reporting that rather than implying coverage.** Reverting `main.go:179` to the old space separator leaves the whole suite green. That line is the fix for the shutdown-line formatting defect, and the reason it is untested is the reason the defect existed: nothing exercises the log-emitting surface (see the Self-Review list, adversarial finding 4). A new test does now cover `--log-dropped` end to end inside `driftFilter`, but `realMain`'s three shutdown lines remain uncovered, and closing that properly means extracting the log surface — a structural change I chose not to fold into this PR.
- All mutations were applied to a copy on that machine and reverted; the suite is green at HEAD, verified with `-count=1` so no result is served from the test cache.
- `make docs-check` passes, including "no re-aligned table rows" and the always-loaded context budget (37.8k of 38,000 chars).
- `npx prettier@3.9.6 --write` applied to the four changed Markdown files; all four are unchanged by it.
- Two pre-existing failures elsewhere in the module are unrelated and unaffected: `gofmt` flags `internal/controller/platformagent_litellm_policy_test.go` for trailing blank lines (#1545 exists to fix exactly that, and it is still unformatted on current `main` — verified this round), and several manifest tests fail resolving `../../../deploy` and `../../../agents` paths. Nothing outside `cmd/drift-detector/` references this package.
- Rebased onto `main` at `f419328c` before opening. `main` had moved under `docs/README.md`, one of the files this change edits; the rows differ (204/372/452 theirs, 350/457 mine, post-rebase) and the rebase was clean, with `docs-check` re-run green afterwards. Re-checked immediately before opening: merge base and `main` head are both `f419328c`, zero commits since divergence, and no file this branch changes has been changed on `main`.

### Live validation

**Eval exemption, in one line:** this changes nothing an agent does — no image builds the binary, no entrypoint starts it, and no runtime path reaches it — so there is no agent behaviour for a bench case to observe, the same exemption #1594 took.

What there *is* to validate is whether the classification rules match reality, and that was done against live data rather than against the spec.

**Source.** Cloud Logging Admin Activity audit logs from three GCP projects: `kube-agents-autopush`, `kube-agents-gkedemos`, `stalhaali-gkedemos`. `gcloud logging read --format=json` returns the same `LogEntry` shape the Log Router sink publishes to Pub/Sub, so a read-only capture feeds straight into `parseAuditEntry` with nothing adapted. Admin Activity retains 400 days, which is what makes a 30-day human census possible at all.

**The row cap is the trap in every figure here, and I walked into it three times.** `gcloud logging read` stops at 10,000 rows, so a count that comes back as exactly 10,000 is a floor with no upper bound attached rather than an answer. Pushing both tier rules down into the Cloud Logging filter fixed this for the *narrow* classes — the human-candidate and failed-human queries return counts far below the cap and are therefore complete — but the system-tier and whole-stream counts kept reaching it. An earlier draft of this PR read one of those 10,000s as a 24-hour total and stated "roughly 10k mutating calls a day per project". That was the cap. It is corrected below and in the README, and it is the reason the ratios and the volume are now measured two different ways.

**Tier ratios, re-measured uncapped.** Over a 15-minute window on each of the three projects — short enough that no query reached the cap, so these are complete counts and not floors — the post-exclusion split is 97.4–98.5% `system` and 1.5–2.4% `automation`, with `system:cluster-autoscaler` alone accounting for 48–62% of the whole post-exclusion stream. Humans appear in none of those windows, which is ordinary: the 30-day census below puts them at about 29 calls a day across all three projects. Six of one project's 3,046 records fell to `unattributed` — principals carrying neither a `system:` prefix, a service-account domain, nor an `@` at all — so the fourth tier turns up in this measurement independently of the 1,230 below.

**Volume, with the cap accounted for.** Counted over windows short enough to avoid it, the post-exclusion stream ran 1 to 10 calls a second — roughly 100k to 840k a day — and two windows twelve minutes apart on the same project differed by 30%. `drift-pubsub`'s README measured 0.7 a second on a quieter two-cluster project, which is consistent with that spread rather than in tension with it. The unfiltered stream hit the cap inside 15 minutes on all three projects, so it is at least 11 a second, and the sink filter is what stands between the two. Volume is per-project and is not a constant; the order of magnitude is the usable part.

**The 1,230.** Captured the rows the breakdown's classifier would label human — neither `system:` nor a service account — over 24 hours: 746 + 467 + 17 = 1,230 across the three projects. Replayed through the compiled `Classifier` from this branch:

```
human-kube-agents-autopush.csv   (n=746)    unattributed  746
human-kube-agents-gkedemos.csv   (n=467)    unattributed  467
human-stalhaali-gkedemos.csv     (n=17)     unattributed   17

TOTAL (n=1230)    system 0 | automation 0 | unattributed 1230 (100.0%) | human 0
false humans eliminated: 1230 of 1230
```

Broken down by cause: 1,110 from `kubelet-nodepool-bootstrap` and 120 unauthenticated requests with an empty principal.

**Humans do exist; that day was quiet.** A 30-day search found 879 human calls across the three projects (185 / 368 / 326), from six distinct principals, all `@google.com` — about 29 a day across three projects. That is close to the spike's estimate of ~7 a day per cluster in absolute terms; what differs by orders of magnitude is the denominator, which is why the filter is worth more than the spike suggested rather than less. `human_patch.json` and `human_denied.json` are captured from this set, with principals substituted.

**Failed calls, measured.** Of those 879 human calls, 46 failed — 5.2%, spread across `NOT_FOUND`, `ABORTED`, `PERMISSION_DENIED`, and `FAILED_PRECONDITION`. On the machine side, a 24-hour query for failed mutating calls on one project returned its full 5,000-row limit, so **5,000 is a floor imposed by the query, not a total**; 4,793 of those rows are status 10 (`ABORTED`) and 4,735 are one controller losing optimistic-concurrency races on `leases.create`. All three projects show that same class. Being honest about what that figure does and does not argue: those particular records are `system:` tier and the tier filter would drop them anyway, so the outcome filter's independent value is the 5.2% of *human* calls — changes a person tried to make and was stopped from making, which are the clearest possible non-drift.

**Cross-validation.** The tier distribution was first derived with `awk` over the captured CSVs, then recomputed by replaying the same CSVs through the compiled `Classifier`. The two agree exactly (307 → 307 system; 354 → 350 system + 4 automation), so the Go rules and the independent shell measurement are not restating each other's errors. The throwaway test used for this was deleted and is not in the diff.

**What this could not cover, and why:**

- **No end-to-end Pub/Sub run.** Records were read from Cloud Logging, not pulled from the subscription, because #1596 has not merged and so no install has the sink, topic, or subscription provisioned. The ingestion path itself is T1's, already merged and unchanged here.
- **No CR, Deployment, or in-pod observation**, because nothing deploys this binary. There is no layer to observe.
- **`--human-domains` is untested against a multi-domain fleet.** Every human principal observed is `@google.com`, so the flag's discriminating behaviour is covered by unit tests only, not by live data. That applies with more force to the exact-domain-versus-subtree decision in finding 33: the three projects here give no evidence either way about an organisation whose accounts live under subdomains, so that choice rests on the direction the error runs, not on a measurement.
- **The idle line (finding 32) has not been seen in a pod log**, for the same reason as everything else here: nothing launches the binary. It is covered by three unit tests and seven mutations, which is the strongest evidence available without a deployment.
- **The Google-managed service-account fix rests on the domain list being correct, not on a live sighting.** A background query for accounts on those four legacy domains in these projects timed out and returned nothing, which is inconclusive rather than confirming — and these three projects have Workload Identity enabled, which is exactly the configuration that makes the legacy accounts rare. The fix is defended by the documented domain shapes and the table test, not by having caught one in the wild here.
- **Cleanup.** Four `kubectl` attempts to plant a human-attributed change on `stalhaali-gkedemos/cluster-1` all returned `401 Unauthorized` — an access-token audience problem, not a permissions one, since `gcloud` auth and Cloud Logging both worked. Nothing was created, so nothing needed cleaning up; the real human fixtures came from log history instead. No test artifacts remain in any project.

## Self-Review

**Passes run.** `review-adversarial` and `review-docs-drift`, four rounds each — eight subagent runs
in total — re-run after every round of fixes because a fix to one pass's finding is how you create
the next one (round 4 exists because round 3's fixes rewrote README paragraphs, and round 4 found
that those rewrites had introduced four-way duplication). Adversarial angles: A (line-by-line and
enclosing functions), B (removed behaviour — `logRecord` → `logActionable`), C (cross-file tracing
of every introduced identifier, plus the Terraform↔Go sibling symmetry and `start-services.sh`),
D (operational and security blast radius), E–G, H (`AGENTS.md`, `CLAUDE.md`,
`.agents/rules/core_engineering.md`, doc-map one hop out), I (scope and test coverage). Docs-drift
covered code→docs, docs→code, the design banner, the documentation map, sibling docs naming drift,
and the internal arithmetic of every figure the diff adds.

**Context each ran in.** All eight in subagents spawned by `/pr-preflight`, none of which wrote the
change; each was handed the diff range, its skill path, and the `make docs-check` output, and each
was told to derive intent from the diff, not to read commit bodies or scratch files, and to edit
nothing. No pass ran in the authoring context.

**What the passes could not cover.** Angle J (sibling pull requests) never ran — `gh` is
unauthenticated in this sandbox, so no pass could list open PRs touching
`k8s-operator/cmd/drift-detector/` or `terraform/modules/drift-pubsub/`, and none could open #467 to
check the claims this README argues against. No pass ran the Go tests (binaries are SIGKILLed here);
each read them instead, and I ran them separately on another machine. No pass had GCP access, so
every measured figure in the new prose and comments is unverified by them — a reviewer with the
project should spot-check `pods.exec.create` being present in Admin Activity and
`serviceAccountDelegationInfo` being absent, since two design decisions rest on those. No pass could
observe a running detector, because nothing builds or launches this binary. And two sets of changes
have not themselves been through a pass: the fixes made after round 4 — the progress-line time
bound, the principal case-folding, `%q`, the `StatusMessage` assertion, `human_exec.json`, and the
`eviction` rationale rewrite — and the fixes for findings 31–35, which answer a reviewer and were
written in the authoring context. Both sets are mutation-tested instead (see Testing). Re-running
the preflight passes over the follow-up commit is the obvious next step if a reviewer wants it;
finding 32 is the argument for doing so, since it is a gap four rounds produced and then recorded
as closed.

Findings below are severity-ordered and merged across all four rounds and both passes, deduplicated
where the two passes found the same defect.

### HIGH / Blocking — all fixed

1. **`.iam.gserviceaccount.com` sends every Google-managed service account to the human tier.**
   (adversarial R1, HIGH/CONFIRMED; docs-drift R3 Blocking for the Terraform half.) A Cloud Build
   pipeline deploying to a cluster would have been reported as a person making an out-of-band
   change. **Fixed:** the suffix is now the whole `.gserviceaccount.com` domain, with the four
   Google-managed forms as table cases and a lookalike-domain case (`evilgserviceaccount.com`)
   pinning the leading dot. `terraform/modules/drift-pubsub/README.md` said the opposite of the
   `main.tf` comment this change added; that sentence is corrected too.
2. **The design banner and the documentation map both claimed nothing classifies a record.**
   (docs-drift R1 B1/B2; adversarial R1 HIGH/CONFIRMED.) `make docs-check` passes over this class
   — it is prose against source, not a generated region. **Fixed:** both restated, and both now
   also name the non-declarative-subresource filter (docs-drift R3 Advisory).
3. **The package doc presented T2 as unbuilt** in the file that constructs the classifier.
   (docs-drift R1 B3; adversarial R1 MEDIUM.) **Fixed** in `main.go` and `subscriber.go`.
4. **The tier-ratio paragraph attributed a human-call rate to a window that contained no humans.**
   (docs-drift R1 B4.) **Fixed:** the 24-hour split and the 30-day human census are now stated as
   the separate measurements they are.
5. **The post-exclusion volume figure contradicted two measurements already in the tree**, one of
   them in the same package. (docs-drift R2 B1; adversarial R2 MEDIUM/CONFIRMED.) This was the
   10,000-row cap read as a total. **Fixed** everywhere, and the Live validation section says so
   explicitly rather than quietly correcting it.
6. **"unmerged" as PR status in a doc** — `AGENTS.md` forbids it. (docs-drift R2 B2.) **Fixed.**
7. **"this parses neither" was false about the user agent** — it is decoded and printed; what it is
   not is classified on. (docs-drift R3 Blocking.) **Fixed.**

### MEDIUM — fixed

8. **`kubectl exec` is forwarded as drift.** (adversarial R2 HIGH/PLAUSIBLE, R3 MEDIUM/CONFIRMED.)
   The pass could not settle whether GKE puts `pods.exec.create` in the Activity log or the Data
   Access log, and named the query that would. I ran it: it is in Activity. Round 3 then found the
   same route open for `ephemeralcontainers` (`kubectl debug`, the archetypal CUJ 3 scenario) and
   `token`. **Fixed:** `nonDeclarativeSubresources` drops all six.
9. **The progress line's cadence swings from 17 minutes to 4 hours** across the stream rates this
   change itself measured, so its stated purpose — surviving a SIGKILL — is weakest where it
   matters most. (adversarial R2 and R4, both MEDIUM/CONFIRMED.) **Fixed:** `countsLogMaxInterval`
   bounds it at 15 minutes as well as 10,000 records. **This entry previously claimed the finding
   was fully fixed and it was not** — R2 named two remedies, a time bound *and* reporting on an idle
   poll, and only the first was implemented. The progress line is emitted from `driftFilter.Handle`,
   so neither half of its interval can fire on a subscription delivering nothing, which is the case
   R2 actually cared about. `kyber775` caught the gap after this PR opened; see finding 32.
10. **The shutdown unattributed line joined with a space**, defeating the `unattributedListSeparator`
    the same change introduced for exactly that reason. (adversarial R3 MEDIUM/CONFIRMED.)
    **Fixed** — and see finding 18: this line is still untested, which is why it diverged.
11. **Case-sensitive principal matching.** (adversarial R3 MEDIUM/CONFIRMED for the flag value; R4
    LOW/PLAUSIBLE for the principal.) A mixed-case `--human-domains` value classified every human as
    unattributed. **Fixed on both sides, asymmetrically and deliberately:** the domain comparison is
    folded on both sides because DNS is case-insensitive; `--automation-principals` is not folded at
    all, because a Kubernetes username is case-sensitive by specification. A test pins each half.
12. **The rule with the strongest live-capture claim had no captured fixture.** (adversarial R4
    MEDIUM/CONFIRMED.) Nothing exercised `parseAuditEntry` on a `resourceName` carrying a
    subresource, so if `resourcename.go` stopped populating `Subresource` the exec test would still
    pass. **Fixed, with a caveat I am flagging rather than burying:** `testdata/human_exec.json` is
    **derived** from a captured envelope, not captured. The live query established that
    `pods.exec.create` appears in Admin Activity; it did not hand me an entry I could ship. The
    fixture header says so and asks for a capture to replace it.
13. **The empty principal was added to the "write a rule for these" list**, where it has no name and
    no rule is possible. (adversarial R1 MEDIUM/CONFIRMED.) **Fixed:** `unauthenticatedPrincipalLabel`.
14. **Nothing tested the new flags or the wiring that consumes them** — transposing the two
    `NewClassifier` arguments compiled, vetted clean, and passed the whole suite. (adversarial R1
    MEDIUM/CONFIRMED.) **Fixed:** `TestNewFilterFromFlagsWiring`, which is the mutation that
    demonstrates it works.
15. **The tally was emitted only on a graceful stop.** (adversarial R1 MEDIUM/CONFIRMED.) **Fixed:**
    the progress line, and see finding 9 for its bound.

### MEDIUM — reported, not fixed

16. **A human acting through an impersonated GCP service account, or a ServiceAccount token, is
    classified as a machine and dropped.** (adversarial R2 MEDIUM/CONFIRMED.) Real, and the one
    classification error that fails closed — a missed report rather than a false accusation. Closing
    it needs the user agent or `serviceAccountDelegationInfo`; I queried for the latter and it is
    absent from every `k8s_cluster` record across seven days on three projects, which leaves the
    user agent as the only lead and not a good one. Documented as a named blind spot in both the
    code and the README rather than papered over.
17. **A server-side dry run is a successful mutating call that changed nothing.** (adversarial R2
    MEDIUM/PLAUSIBLE.) No `dryRun` marker appeared in seven days of Admin Activity logs across three
    projects, so it is not yet established that GKE surfaces one at all. Writing a rule for a payload
    shape nobody has seen is worse than documenting the gap, which is what the README does.

### LOW — fixed

18. **`eviction`'s exclusion rested on an argument its own `ephemeralcontainers` entry refutes.**
    (adversarial R4 LOW/CONFIRMED.) The pass is right that the ownership argument does not
    distinguish them. **Fixed as a rewrite of the rationale, not of the behaviour:** evicting a
    Deployment-owned pod changes nothing in Git, but evicting a pod applied from a manifest of its
    own removes the object Git declares — and a subresource name cannot tell those apart, only T3's
    `managedFields` join can. So `eviction` stays reported, and a `kubectl drain` still produces a
    line per pod. That is the trade I would rather make blind in this direction than the other.
19. **Unattributed principal names reached the log unquoted**, unlike every other site that prints a
    principal, and they are the population most likely to hold something unusual — a username from a
    customer OIDC provider can contain a newline. (adversarial R4 LOW/PLAUSIBLE.) **Fixed:** `%q`.
20. **`StatusMessage` was parsed and carried but never asserted on a parsed value**, so a wrong json
    tag would silently empty the half of the dropped-record line that says why a change was
    rejected. (adversarial R4 LOW/CONFIRMED.) **Fixed:** asserted in `TestClassifyCapturedFixtures`.
21. **A test restored `log` output to a hardcoded `os.Stderr`** where its sibling saves the previous
    writer. (adversarial R4 LOW/CONFIRMED.) **Fixed:** `prev := log.Writer()`.
22. **Two different row caps (5,000 and 10,000) cited in one README with no note of why**, reading
    as a typo. (docs-drift R3 and R4.) **Fixed** with one clause.
23. **Three new lines broke the file's ~100-column wrap**; prettier's `proseWrap: preserve` catches
    nothing. (docs-drift R4.) **Fixed.**
24. **"prefix and suffix matching" in the design doc** understated the classifier, which uses four
    forms. (docs-drift R1, R4.) **Fixed** — and this one was my own round-2 edit introducing a new
    inaccuracy while fixing a worse one.
25. Also fixed from earlier rounds: the fixture-provenance claim now admits the `_trimmed` body
    stubs; the `#467` citation is linked rather than asserted; the `start-services.sh` deployment
    path is phrased as intended rather than existing; the crawler exemplars agree between code and
    README; the `78/20/1` sum and the ambiguous "48–62% of it" antecedent are both resolved.

### Found outside the passes, in the final redaction sweep — fixed

Two fixtures (`delete_call.json`, `kubelet_bootstrap.json`) carried real
`authentication.kubernetes.io/credential-id` X509 SHA-256 fingerprints from a live cluster, which
the identifier substitution had missed — every other identifier in those files, including the node
name, was already substituted. They are hashes of public certificates rather than secrets, but they
are cluster-specific identifiers going into a public repository for no reason, and their presence
contradicts the fixture-provenance claim the README makes. Both are now obviously-synthetic hex.
Nothing reads them: only `resource.labels` is decoded, not the top-level `labels` map, and the suite
is green after the change. No pass caught this, because no pass had a reason to treat a hex string
as an identifier; I checked for it because the PR body and diff get a redaction sweep before
opening.

### LOW — reported, not fixed

26. **Every fact does not have one home.** (docs-drift R4 A1–A3; adversarial R4 LOW/CONFIRMED.) The
    `.iam.` argument is now in four places, three added by this diff; the detector README restates
    `classify.go`'s doc comments near-verbatim, measurements included; and the README grew 55 → 144
    lines, most of it measurement narrative that reads as design rationale, which `AGENTS.md` puts
    in `docs/designs/`. This is a real tension and a direct consequence of fixing round 3's Blocking
    finding by *adding* the correct explanation rather than pointing at where it already lived.
    **Why I am reporting rather than acting:** the pass supplied its own counter-argument —
    `docs/designs/drift-detection.md` declares everything below its banner to be reference design,
    so 89 lines measuring *shipped* behaviour do not cleanly belong there either, and there is no
    obviously-correct third home. Restructuring that on the way out the door is how you introduce
    fresh drift. **This is the open question I would most like a reviewer's call on**, and if the
    answer is "the component README is the right home", the design doc should say so explicitly
    instead of leaving the split implicit.
27. **The log-emitting surface is only partly tested.** (adversarial R3 LOW/CONFIRMED.) A test now
    covers `--log-dropped` end to end inside `driftFilter`, but `realMain`'s three shutdown lines
    remain uncovered — and finding 10 is a defect that shipped on precisely that path. Reverting the
    `main.go:179` separator fix leaves the suite green; I am stating that rather than implying
    coverage. Closing it properly means extracting the log surface, a structural change I chose not
    to fold into this PR.
28. **`splitList` duplicates `splitCSV` in `cmd/k8s-event-watcher`** — same semantics, different
    name, so neither turns up when you grep for the other. (adversarial R3 LOW/CONFIRMED.) Both are
    `package main`, so sharing needs an extraction into `k8s-operator/internal/`. Structural, and
    out of scope for a change that is already large.
29. **Two hard-coded counts** — "Four tiers" and "drops the six that arrive this way". (docs-drift
    R2, R4.) Both are enumerated in the same sentence, which caps the damage; "the six" is the
    fragile one and goes wrong the day a seventh subresource is added. Left as prose that reads
    correctly today.
30. **Five consecutive bolded lead-in sentences** in one README section, against "Write it straight".
    (docs-drift R4, self-described as its weakest finding.) They are full sentences rather than
    `**Term:** explanation`, and rewriting five paragraph openings for style during a review round is
    how you introduce a factual error. Left.

### Found by review after this PR opened — all fixed

`kube-agents-bot` reported no findings ("3 parts, three ways each, 23 possible issues checked").
`kyber775` then reviewed and found five things, three of which the four preflight rounds missed
outright and one of which contradicts a disposition above. All five are fixed in the follow-up
commit; each fix is mutation-checked (see Testing).

31. **The `.gserviceaccount.com` suffix test was not case-folded** (WARNING, `classify.go:283`), so
    `deployer@proj.iam.GSERVICEACCOUNT.COM` missed the automation rule, carried an `@`, and passed
    the positive human test — a service account reported as a person making an out-of-band change,
    which is the precise false positive finding 1 above exists to prevent. Confirmed by reading the
    code. Worse than a missed case: it is inconsistent with the reasoning in finding 11, which folds
    `--human-domains` because DNS is case-insensitive, and `.gserviceaccount.com` is equally a DNS
    domain. The comment I wrote to justify the asymmetry conflated the two — it argues correctly
    that a *username* is case-sensitive by specification, which covers the exact-match allowlist and
    does not cover a domain suffix. **Fixed:** the suffix test folds, the two username rules above it
    stay exact, and the comment now says which is which.
32. **No idle reporting in the subscriber loop** (WARNING, `subscriber.go:270`), quoting the
    `countsLogInterval` comment's own promise of it. Confirmed: the empty-pull branch in
    `subscriber.Run` only slept. This is the one that should not have survived preflight — R2 raised
    it, named both remedies, and finding 9 above recorded it as fixed on the strength of one of them.
    An empty `Pull` is not an error, no message means no batch-skip line, and the progress line fires
    from `driftFilter.Handle`, so a sink whose filter stopped matching produced no output at all and
    read exactly like the expected steady state. **Fixed:** `idleReportInterval` in `subscriber.Run`,
    on the same fifteen-minute bound, carrying the running totals so "never wired up" (zeroes since
    start-up) is distinguishable from "worked and has gone quiet". The two comments that described
    this reporting as existing now describe what each half actually covers.
33. **`--human-domains=.example.com` silently matched nothing** (WARNING, `classify.go:240`).
    `TrimPrefix` stripped only the `@`, so the value was stored as `@.example.com` and could never
    match — every person in the fleet filed as `unattributed` while the detector reported healthy.
    Same failure shape as finding 11, which is why it is worth having. **Fixed:** a leading `.` is
    stripped too, so `example.com`, `@example.com` and `.example.com` are one configuration written
    three ways. The reviewer's other half — widening the match to the whole subtree — is
    **deliberately not taken**, and this is the one place I went against the finding as written:
    `--human-domains` is what *narrows* the human tier, so every principal it newly admits is a
    record newly reported as somebody's drift, and no fleet measured here spreads its accounts across
    subdomains. Two tests pin the exact-domain behaviour so a later widening has to be deliberate,
    and the README says to list subdomains. Answered in the thread rather than dropped.
34. **`strings.Contains(principal, "@")` admitted `"@"`, `"ada@"` and `"@example.com"` as people**
    (SUGGESTION, `classify.go:315`). Correct; none of the three is an account and the last two are
    the shapes a truncated principal takes. **Fixed:** `LastIndex` with a local part and a domain
    required either side.
35. **`driftFilter`'s unsynchronised mutation was undocumented** (SUGGESTION, `classify.go:456`).
    The preflight passes examined this and filed it under refuted below, correctly — it is
    single-goroutine and safe today. The reviewer's point is different and stands: the constraint was
    nowhere written down, and T3's per-record `managedFields` call is exactly what makes handling a
    batch concurrently look worthwhile. **Fixed:** documented on the type, naming T3 as the change
    that makes it wrong.

**One thing the mutation pass then found that the reviewer did not.** Removing the
`lastDelivery` reset after a delivered batch left the whole suite green, which means a detector
pulling records perfectly well would have announced itself idle a quarter of an hour after start-up
— a false log line, and worse than the silence it replaced. `TestRunIdleIntervalRestartsAfterADelivery`
now covers it, and the reset moved to after `processBatch` rather than before it.

### Raised by a pass and refuted, recorded so a reviewer does not re-derive them

Split audit entries defeating the outcome filter (all eight fixtures are `first: true, last: true`);
`Status` needing to be a pointer (proto3 omission and an explicit `"code": 0` decode identically,
and nothing needs to tell them apart); the sink/classifier `.iam.` asymmetry (deliberate, documented
on both sides, and costs delivered volume only); classifying before the outcome filter (deliberate,
pinned by `TestDriftFilterCountsFailedHumanInItsTier`); the 129th map entry at the overflow cap
(correct, and pinned); `TierCounts{}` with a nil `ByTier` (nil-map reads are legal); unsynchronised
maps in `driftFilter` (single goroutine — `Run` → `processBatch` → `handle`; still true, but see
finding 35 — the safety was never written down); and the
no-magic-constants rule against the inline format strings (they are "a literal that is the subject
of the line it appears on" per `.agents/rules/core_engineering.md`). Both passes independently
cross-checked the arithmetic of every figure the diff adds and found it internally consistent.

## Risk & Rollout

Low. Nothing builds this binary into an image, no entrypoint launches it, and no installation runs it, so the blast radius on merge is zero — the same posture as #1594. The only behavioural surface is inside `cmd/drift-detector/`, which nothing else imports. Reverting is a single revert with no migration and no state to unwind.

The failure mode to watch when T4 does wire this up is the filter being too tight rather than too loose: a principal shape not anticipated here lands in `unattributed` and is dropped, which loses a real drift event instead of injecting a false one. That is the intended trade — a missed inject is recoverable and a false accusation is not — and the unattributed principals are logged by name at shutdown precisely so the miss is visible rather than silent.

---


kyber775 (commented):
## Code Review: PR #1629 — feat(drift-detector): classify principals and drop calls that changed nothing

Thorough, grounded implementation with commendable empirical rigor behind the tier breakdown and live Cloud Audit Log fixtures. The positive human test, the outcome filter, and the non-declarative subresource filters solve real operational problems seen in live clusters.

A few load-bearing edge cases in principal classification and operational telemetry warrant attention before merge:

---

### Findings

#### [WARNING] `k8s-operator/cmd/drift-detector/classify.go:283` — Un-normalized case-sensitive service account suffix check leaks into human tier
In `Classifier.Classify()`:
```go
if strings.HasSuffix(principal, gcpServiceAccountSuffix) {
    return TierAutomation
}
```
`gcpServiceAccountSuffix` is `".gserviceaccount.com"`. While native GCP identities are lowercase, principals arriving via custom OIDC federation or x509 CNs preserve caller casing. If a service account arrives with mixed/uppercase domain segments (e.g. `deployer@my-proj.iam.GSERVICEACCOUNT.COM`), `strings.HasSuffix` returns `false`.

The principal then falls through to `c.isHuman(principal)`. When `--human-domains` is left at default (empty), `isHuman` checks `strings.Contains(principal, "@")`, which returns `true`. The service account is misclassified as `TierHuman` and flagged as actionable drift.
- **Fix:** Normalize the principal to lowercase for the service account suffix check:
  ```go
  if strings.HasSuffix(strings.ToLower(principal), gcpServiceAccountSuffix) {
      return TierAutomation
  }
  ```

---

#### [WARNING] `k8s-operator/cmd/drift-detector/classify.go:240`, `classify.go:309` — Subdomain blind spot in `--human-domains` matching
`NewClassifier()` formats configured domains with a leading `@`:
```go
c.humanDomains = append(c.humanDomains, domainPrefix+strings.TrimPrefix(domain, domainPrefix))
```
For `--human-domains=example.com`, `c.humanDomains` stores `"@example.com"`. In `isHuman()`:
```go
for _, d := range c.humanDomains {
    if strings.HasSuffix(lowered, d) {
        return true
    }
}
```
If an organization issues accounts under subdomains (e.g. `alice@corp.example.com` or `bob@eng.example.com`), `strings.HasSuffix("alice@corp.example.com", "@example.com")` returns `false` because `@` does not precede `example.com`. Configuring `--human-domains=.example.com` also fails because `strings.TrimPrefix` only strips `@`, resulting in `@.example.com`. Human out-of-band changes from corporate subdomains will be classified as `TierUnattributed` and dropped.
- **Fix:** Normalize configured domains without leading `@` or `.`, and match either `@` or `.` boundaries:
  ```go
  d := strings.TrimPrefix(strings.TrimPrefix(domain, "@"), ".")
  // in isHuman:
  if strings.HasSuffix(lowered, "@"+d) || strings.HasSuffix(lowered, "."+d) {
      return true
  }
  ```

---

#### [WARNING] `k8s-operator/cmd/drift-detector/subscriber.go:270` — Missing idle progress / heartbeat reporting in subscriber loop
`classify.go` lines 103–104 and 116–118 explicitly state:
> *"A sink whose filter stopped matching produces no pull error and no batch-skip line, so that case needs the idle reporting in the subscriber loop rather than this... between the startup line and the first progress line the pod log is empty and a healthy detector is indistinguishable from a broken sink."*

However, `subscriber.Run()` contains no idle reporting:
```go
if len(messages) == 0 {
    if !sleepCtx(ctx, s.idleWait) {
        return ctx.Err()
    }
    continue
}
```
If a subscription is empty or the sink filter stops matching, the detector runs completely silent indefinitely, making a broken sink indistinguishable from a healthy quiet cluster.
- **Fix:** Track idle elapsed duration in `subscriber.Run()`. If idle for `countsLogMaxInterval` (15 minutes), emit an idle heartbeat line.

---

### Suggestions

#### [SUGGESTION] `k8s-operator/cmd/drift-detector/classify.go:315` — Validate local and domain parts in fallback human test
When `--human-domains` is empty, `return strings.Contains(principal, emailLocalDomainSeparator)` treats malformed strings like `"@"` as human. Validate non-empty local and domain parts:
```go
idx := strings.LastIndex(principal, emailLocalDomainSeparator)
return idx > 0 && idx < len(principal)-1
```

#### [SUGGESTION] `k8s-operator/cmd/drift-detector/classify.go:456` — Document concurrency constraints on `driftFilter`
`driftFilter` mutates `handled`, `counts.ByTier`, and `unattributed.seen` without mutex protection. It is safe today because `subscriber.Run` processes batches serially in a single goroutine, but it is not safe for concurrent use. Document this explicitly before asynchronous T3/T4 work is added.

stalhaali:
Thanks — three of these are real defects that four rounds of preflight review missed, and one of them contradicts a disposition I had already written into the Self-Review. All five are addressed in f4e7374e. Each fix is mutation-checked: the fix is reverted, the suite is run, and the named test has to red.

**1. `.gserviceaccount.com` not case-folded (`classify.go:283`) — fixed as proposed.**

Confirmed. `deployer@proj.iam.GSERVICEACCOUNT.COM` missed the automation rule, carried an `@`, and passed the positive human test, so a service account would have been reported as a person making an out-of-band change — the precise false positive the automation tier exists to prevent.

It is worse than a missed case, because it contradicts the reasoning already in the file: `isHuman` folds `--human-domains` because DNS is case-insensitive, and `.gserviceaccount.com` is equally a DNS domain. The comment I wrote to justify the asymmetry conflated two different checks — it argues correctly that a *username* is case-sensitive by specification, which covers the exact-match allowlist and says nothing about a domain suffix. The suffix test now folds, the two username rules above it stay exact, and the comment says which is which.

Mutation: un-folding it reds `service_account_domain_in_upper_case` and `service_account_domain_in_mixed_case`. The lookalike case is pinned from both sides too (`mallory@EVILGSERVICEACCOUNT.COM` must stay human), so the fold cannot widen the boundary.

**2. `--human-domains` subdomains (`classify.go:240, 309`) — half taken, half declined, and the declined half is the one I want to argue.**

The trap is real and is fixed: `TrimPrefix` stripped only the `@`, so `.example.com` was stored as `@.example.com` and could never match anything, filing every person in the fleet as `unattributed` while the detector reported healthy. A leading `.` is now stripped too, so `example.com`, `@example.com` and `.example.com` are one configuration written three ways. This is the same failure shape as a case-sensitivity bug fixed earlier in the branch, so thank you for it.

I have **not** taken the other half — matching on either an `@` or a `.` boundary, so that `example.com` covers `ada@corp.example.com`. `--human-domains` is the flag that *narrows* the human tier, so every principal it newly admits is a record the detector newly reports as somebody's drift. Widening the default behaviour of an existing flag in that direction wants evidence, and I do not have any: all 879 human calls measured across three projects over 30 days are `@google.com`, so there is no fleet here whose accounts are spread across subdomains. The comma-separated flag already covers that org — `--human-domains=corp.example.com,eng.example.com` — and the misses are not silent, since an unmatched principal lands in `unattributed` and is logged by name, which is the list an operator reads to find the domain they left out.

Three tests now pin the exact-domain behaviour (`configured_domain_does_not_match_a_subdomain`, `leading_dot_does_not_add_subdomain_matching`, `subdomains_are_matched_by_listing_them`) so that widening it later has to be a deliberate change against those tests rather than a quiet one, and both the flag table and the Classification section in the README say the match is exact. Happy to be overruled if you think the subtree is the better default — it is a one-line change and the tests say exactly what it would break.

**3. No idle heartbeat (`subscriber.go:270`) — fixed, and this is the one that should not have survived preflight.**

Confirmed by reading `Run`: the empty-pull branch only slept. Your quoting my own comment back at me is fair, and it is worse than a stale comment — an adversarial pass on this branch raised this exact defect, named two remedies (a time bound *and* reporting on an idle poll), and I implemented only the first and then recorded the finding as fixed. The progress line fires from `driftFilter.Handle`, so neither half of its interval can reach a subscription delivering nothing, which is the case the finding was about. An empty `Pull` is not an error and no message means no batch-skip line, so a Log Router sink whose filter stopped matching produced no output of any kind and read exactly like the expected steady state.

`subscriber.Run` now reports on the same fifteen-minute bound, carrying the running totals so that "never wired up" (zeroes since start-up) is distinguishable from "worked and has gone quiet". The two comments that described this reporting as already existing now describe what each half actually covers. The Self-Review entry that claimed the earlier finding fixed has been corrected rather than quietly updated.

Three tests and four mutations: removing the report reds `TestRunReportsAnIdleSubscription`; firing it on every empty poll reds `TestRunDoesNotReportIdleBeforeTheInterval`. The fourth found something you did not raise and neither did I — with no test for it, removing the `lastDelivery` reset after a delivered batch left the suite green, which means a detector pulling records perfectly well would have announced itself idle a quarter of an hour after start-up. That is a false log line and worse than the silence it replaced. `TestRunIdleIntervalRestartsAfterADelivery` covers it, and the reset moved to after `processBatch` rather than before it.

**4. `strings.Contains(principal, "@")` (`classify.go:315`) — taken.**

Correct: `"@"`, `"ada@"` and `"@example.com"` were all classified human. None is an account and the last two are the shapes a truncated principal takes. Now `LastIndex` with a local part and a domain required either side — `LastIndex` rather than `Index` because a quoted local part may itself contain an `@`. Reverting to `Contains` reds all three new cases.

**5. `driftFilter` concurrency (`classify.go:456`) — taken.**

The preflight passes looked at this and filed it as safe, which it is: `Run` → `processBatch` → `handle` is one goroutine. Your point is a different one and it stands — the constraint was nowhere written down. Now documented on the type, naming T3's per-record `managedFields` call as the change that makes handling a batch concurrently look worthwhile and these maps wrong at the same moment.

---

Full suite: 43 top-level tests, 120 including subtests, 0 failures, three runs in a row and clean under `-race`; `gofmt` and `go vet` clean; `make docs-check` green. The PR body's **Self-Review** and **Live validation** sections are updated with all five findings and their dispositions.

dshnayder (approved):

kyber775 (approved):
# PR #1629 Re-Review: Drift Detector Principal Classification & Outcome Filter

- **PR:** [#1629: feat(drift-detector): classify principals and drop calls that changed nothing](https://github.com/gke-labs/kube-agents/pull/1629)
- **Author:** @stalhaali
- **Reviewer:** Kyber (@kyber775) — Dual-Engine Adversarial Battery & Day-2 SRE Lens
- **Verdict:** **APPROVE** (`[PASS]` / Clean Merge & All Review Findings Resolved)
- **Head Commit Reviewed:** `f4e7374e` (`fix(drift-detector): fold the service-account domain, report an idle subscription`)

---

### 1. In-Session Verification & Execution Evidence

All verification gates were executed locally against the merged head (`f4e7374e`) in this session:

- **Mergeability against `upstream/main` (`02c33092`):**
  - Command: `git merge-tree --write-tree upstream/main pr-1629`
  - Result: **PASS** (clean tree `8c041f65331bf2ed276c0fee024c517c24ff99d4`, 0 merge conflicts).
- **Go Race Test Pass (`drift-detector`):**
  - Command: `cd k8s-operator && go test -race -count=1 ./cmd/drift-detector/...`
  - Result: **PASS** (all tests green under `-race`, 1.216s).
- **Operator Test Suite:**
  - Command: `cd k8s-operator && go test ./...`
  - Result: **PASS** (all controller, webhook, and cmd suites green).
- **Conformance & Invariant Test Suite:**
  - Command: `python3 tests/conformance/run.py`
  - Result: **PASS** (112 tests ran, 112 OK / 12 expected failures matching baseline).
- **Documentation Checks:**
  - Command: `make docs-check`
  - Result: **PASS** (311 Markdown files checked, all relative links and design citations resolve, terminology passed, context budget within 37.8k chars).

---

### 2. Delta Review: Verification of Review Fixes (`f4e7374e`)

Commit `f4e7374e` thoroughly and cleanly resolves all findings identified in the previous review pass:

1. **Service Account Domain Case-Folding (`classify.go:314`):**
   - Suffix comparison now folds the principal (`strings.HasSuffix(strings.ToLower(principal), gcpServiceAccountSuffix)`), preventing mixed/uppercase domains (e.g. `deployer@proj.iam.GSERVICEACCOUNT.COM`) from bypassing the automation tier and falsely entering the human tier.
   - Pinned with tests in `classify_test.go` (`TestClassifyPrincipals` cases for uppercase and mixed-case service account domains, plus lookalike negative cases).
2. **Domain Normalization (`classify.go:261-263`):**
   - `NewClassifier` strips both `@` and leading `.` prefixes (`domainLabelSeparator`), preventing configurations like `--human-domains=.example.com` from turning into `@.example.com` and matching nothing.
   - Pinned with test cases covering leading `@`, leading `.`, both, and explicit subdomain listing.
3. **Fallback Human Email Structure (`classify.go:360-361`):**
   - Replaced naive `strings.Contains` with `strings.LastIndex`, verifying non-empty local and domain components (`at > 0 && at < len(principal)-1`), properly rejecting `@`, `ada@`, and `@example.com`.
4. **Idle Subscription Heartbeat Telemetry (`subscriber.go:41-56, 309-315, 420-424`):**
   - `subscriber.Run` tracks `lastDelivery` against `idleReportInterval` (15m, aligned with `countsLogMaxInterval`), emitting `logIdle` with running counts (`parsed`, `skipped`, `failed`).
   - Closed the operational blind spot where a broken or quiet Log Router sink produced 0 log lines indefinitely.
   - Rigorously tested with `TestRunReportsAnIdleSubscription`, `TestRunDoesNotReportIdleBeforeTheInterval`, and `TestRunIdleIntervalRestartsAfterADelivery` using a mutex-guarded `syncBuffer` to prevent race detector contention during parallel log evaluation.
5. **Concurrency Boundary Documentation (`classify.go:503-514`):**
   - Added explicit godoc documentation on `driftFilter` warning future maintainers that the filter is not safe for concurrent use without synchronizing internal counters and map state before T3's `managedFields` join.

---

### 3. Conclusion & Recommendation

The implementation is rock solid, backed by real Cloud Audit Log empirical data, and all edge cases and telemetry gaps have been addressed with high craftsmanship and sabotage-verified tests.

**Verdict: [PASS] / APPROVE**

docs/designs/drift-detection.md
```diff
@@ -1,6 +1,6 @@
...
+> **STATUS — design of record; partially implemented.** Two pieces exist: [`terraform/modules/drift-pubsub`](../../terraform/modules/drift-pubsub/) provisions the sink, topic, and subscription, and [`k8s-operator/cmd/drift-detector`](../../k8s-operator/cmd/drift-detector/) pulls that subscription, parses the audit records, and classifies each one by principal, forwarding only successful changes made by a person to declarative state — calls that changed nothing (a failure, or a subresource such as `exec`) are dropped and counted. Nothing yet joins `managedFields` or emits a `gitops-drift` inject, and no image builds or launches the detector — so no installation detects drift today. The pipeline this design plugs into — session, agent, chat, approval, GitOps PR — does ship, and is described in [`agents/platform/docs/session_management.md`](../../agents/platform/docs/session_management.md). Treat everything below as the reference design, not as shipped behaviour.
```
jayantid: 🟡 **Technical design.** The classifier is the largest design decision in CUJ 3 so far, and the design of record does not mention it.

`docs/designs/drift-detection.md` covers the audit-log signal and the subscriber choice (line 43) but says nothing about tiers, the positive human test, the outcome and subresource filters, or the per-instance scope of `--automation-principals`. That rationale sits in `k8s-operator/cmd/drift-detector/README.md`, now 144 lines. `AGENTS.md` places per-feature design rationale in `docs/designs/`. The only document that specifies the classifier is the CUJ 3 breakdown on #467, and it specifies a different one: three tiers, residual is human.

If #467 merges after this, `main` carries a spec the code contradicts. The next reader of the design doc learns the classifier from a component README or not at all. Self-Review 26 asks for this call; this is it.

Add a short "Principal classification" section to the design doc: four tiers, why human is a positive test, the two not-a-change filters, and per-instance scope because the sink is project-wide. Leave the README with how the code behaves plus a link, so #467 defers to the design doc in either merge order.

k8s-operator/cmd/drift-detector/README.md
```diff
@@ -26,11 +27,140 @@ go run ./k8s-operator/cmd/drift-detector --project "$PROJECT_ID"
...
+The CUJ 3 task breakdown ([#467](https://github.com/gke-labs/kube-agents/pull/467)) specifies three
+of those. It drops `system:`, drops an allowlist, and calls the residual human.
+Over 24 hours of live audit logs across three projects, that residual was 1230 calls and every one
+was a machine: 1110 from `kubelet-nodepool-bootstrap`, which carries neither the prefix nor the
+service-account suffix, and 120 unauthenticated requests with an empty principal. Public GKE
+endpoints get crawled, and a rejected probe from Googlebot, Baiduspider, Amazonbot or any of the
+others is an audit entry with a mutating `methodName`. An allowlist cannot close that, because it has to
+anticipate every identity GKE invents. A positive human test plus a tier for the leftovers makes an
+unknown principal loud rather than wrong, and the unattributed principals are logged by name so
+there is something to write the next rule from.
```
jayantid: 🟡 **UX flow.** The classifier's safety argument rests on an operator reading the unattributed-principal list, and nothing says where that list will reach them.

Today it goes to one pod log line at shutdown (`main.go:178`) and rides the fifteen-minute progress line. When T4 launches this beside the event watcher, that is a container log inside the gateway pod. The operator's surface in this product is chat and the scheduled reports relayed to it. The event watcher's equivalent is the `eod-event-watcher-daily-report` recap on the cron roster. The only candidate here is T5, which the breakdown marks as stretch.

A customer whose OIDC usernames carry no `@`, or a `--human-domains` value with a typo, files every real human change as unattributed while the detector reports healthy. Nobody reads that list, so the miss stays silent, which is the outcome the fourth tier exists to prevent.

Name the surface now, in the design doc: a daily recap in the event watcher's style, carrying tier counts and unattributed names, as a required part of T4 or T5 rather than stretch.

Issue #1491, opened 2026-09-11
Nightly tier: run at 8 PM ET with an 8-hour budget, record evidence from night one, and report every night's results on the dashboard and in Chat


jayantid:
Goal 1 of the eval program's next three (Jayanti, 2026-09-11): a nightly that runs every eval and whose results everyone can see the next morning. Part of #897; the evidence-store half is #1448.

## Decisions

- **Schedule: 8 PM ET**, outside the working day's presubmit traffic. Prow crons are UTC: `0 0 * * *` is 8 PM EDT and 7 PM EST; pick that and accept the winter hour, or `0 1 * * *` for 8 PM EST and 9 PM EDT.
- **Budget: 8 hours** (`timeout: 480m`), not 6. The 26-case matrix prices at ~530–550 serial minutes at three repetitions and the measured fan-out is about 1.15×, so 6h would truncate most nights; 8h fits with margin, and the job is non-gating so a long night costs only a lease.
- **Parallelism:** wider than the presubmit's 4 is safe at that hour (no herd on the model quota). Start at 6, measure the first three nights, adjust.
- **Evidence from night one:** provision the baseline store the same week (commands in #1448: bucket, service account, two grants; an owner of `kube-agents-prow` runs them), then merge oss-test-infra#2698 so the nightly records and the presubmit reads. Without this the seven-night admission clock does not start.

## Where the results show up

1. **Dashboard, as a report.** The nightly tier lands on the dashboard via #1446 (tier field; nightly kept out of the gate's verdict) and #1479 (Cases page with presubmit and nightly pass rates; Grid with nights as columns). Add a nightly report view: last night's run as a page, cases grouped by domain, pass/partial/fail per case with the check reason and transcript link, newly failing vs the previous night, wall clock and whether the night truncated. Reachable from the Brief and from the Chat digest.
2. **Chat, one line in the 9 AM ET digest** (`post_health.py`): `🌙 Nightly: 26 cases · 22 passed all reps · 3 partial · 1 failed · newly failing: <case> · 6h 40m` with a link to the report. A truncated or missing night says so instead of numbers.

## Work items

- [ ] oss-test-infra#2682: cron `0 0 * * *`, `timeout: 480m`, `EVAL_TASK_PARALLELISM=6`; merge (needs one lgtm there).
- [ ] Store provisioning per #1448 (owner of `kube-agents-prow`), then oss-test-infra#2698.
- [ ] Merge #1446 and #1479 (in review).
[...]

Issue #1225, opened 2026-09-04
Quota storms degrade delegated answers: worker dies on 429s (3 fast retries vs RetryInfo ~54s) without kanban_block; front-end reports delegation status as the final answer


jayantid:
Caught-by: evals

## The defect

When the shared Gemini quota saturates (`generate_content_paid_tier_3_input_token_count`, 8M input tokens/min, global), the platform **worker retries 429s only 3 times ~2.3s apart while Google's RetryInfo asks for ~54s**, then dies **without calling `kanban_complete` or `kanban_block`** — a delivery-contract violation (the #1010/#630 family, storm-triggered). After 3 crashes the dispatcher hard-blocks the card, and the front-end agent's final answer is a fluent delegation-status report: real trajectory, billed tokens, no actual answer.

## Why it matters

This is the new false-red shape from the 2026-09-03 evening storm (full forensics: #1202 verdict comment; evidence on #1097): two top-stability roster cases collapsed 0.0 on unrelated CI-only PRs (#1121, #1192 — both exonerated). Every existing infra classifier (#1191 empty-trajectory, #1095 terminal-429, #1147 scoreless) is content-blind to it — the reps look like real runs with wrong answers. 48 of 49 delegated reps failed across the two runs; the sole pass was the case where delegation is the point. Two more runs dodged the identical collapse only via the partial-evidence rule. In production, the same mechanism delivers "I delegated it" as a final answer to a user during any quota event.

## Fixes, cheapest first (from the forensics)

1. **Worker: call `kanban_block` with the API error on retry exhaustion** — turns a protocol violation into an attributable terminal state (and makes the failure classifiable everywhere downstream).
2. **Honor RetryInfo (~54s) on 429s**, and/or configure a litellm fallback model group (`Available Model Group Fallbacks=None` today).
3. Eval-side: classifier for the delegation-blocked answer shape, corroborated by gateway 429 logs.
4. Ops: cap/stagger aggregate eval concurrency against the shared 8M/min quota (evening peaks = ~8 concurrent full suites; observed 173–289 gateway 429s per project in the window). Secondary: `kanban dispatch: system memory pressure ... limiting to at most 1 new worker` compounds recovery.

Related: #1097 (the saturation itself), #1010/#1174 (fan-out completion contract), #1213 (partial-evidence greens — the same storms feed both failure directions).

bradhoekstra:
[...]

PR #1657, opened 2026-09-16
fix(ci): collect litellm.log on green eval runs, not only on failure

lapis2002:
## Summary

Draft, not for merge yet. This PR runs a 2×2 on the smoke test, one arm at a time, force-pushing the next arm after the previous run lands. The question is whether `EVAL_TASK_PARALLELISM` can go above 4, and whether moving the agent off the preview model changes the upstream Vertex 429s.

| # | arm | parallelism | agent model | build |
|---|---|---|---|---|
| 1 | baseline | 4 | `gemini-3.1-pro-preview` | `2100280101982703616` |
| 2 | parallelism only | 6 | `gemini-3.1-pro-preview` | pending |
| 3 | model only | 4 | `gemini-3.1-pro` | pending |
| 4 | both | 6 | `gemini-3.1-pro` | pending |

Sequential rather than four parallel PRs, so no arm competes with another for the same shared Vertex capacity. Each build's artifacts stay in GCS under its own build ID, so the force-pushes lose nothing.

The first commit is what makes any of this readable — it collects `litellm.log` on green runs, not only on runs that already failed.

`JUDGE_MODEL` is untouched on every arm. The judge runs on the Gemini Developer API key, not through LiteLLM to Vertex, so it is not a source of these 429s.

## Why This Change

Last week's 625 `pull-kube-agents-smoke-test` builds: median green run 160 min, of which the eval fan-out is the whole story and every other profiled step is under a minute. On green build `2100233577361838080`: 54 units summing 413 min of unit latency, packed into a 116 min fan-out.

The fan-out floor is `max(413/P, longest single-task chain)`. The task lock at `ci-eval-pr.sh:1696` serializes repetitions of one task, and the longest chain is 56 min — `compliance-rbac-overgrant` at 1896s + 947s + 514s. So P=6 lands near 69 min, P=8 at the floor, and past 8 another lane buys nothing. That is arm 2.

The cap is not arbitrary, though. The comment above `ci-eval-pr.sh:1582` says it bounds concurrent load, and that build pushed 31.0M input tokens through one pool project in 116 min — mean 267k/min, peak 428k in a 60s window. P multiplies that rate directly, and retries are wall clock. Arm 2 is the only way to know whether the retries eat the gain.

Arm 3 comes from the other half of the review. Over 2026-09-09 → 2026-09-16 Vertex refused 391 requests across 127 of the 175 builds that uploaded a log, and **there is no per-project quota being hit**. `hack/ci-deploy.sh:166-173` already records why: Vertex serves these models under dynamic shared quota, with no fixed per-minute wall. The logs agree — all 391 refusals are the same bare `RESOURCE_EXHAUSTED` naming no metric, where the old Gemini Developer API 429s named theirs (8,000,000 input tokens/minute). And the rate runs opposite to our own load: 09-12 ran 25 builds for a 67% refusal rate while 09-10 ran 156 for 28%.

So there is no quota to raise. What is left to vary is which shared pool we draw from, and a preview model's is the smaller one.

## What Changed

Arm 1 (in the tree now):

- `hack/ci-env.sh` — new `collect_litellm_log()`, split out the way `collect_bench_results()` already was and for the same reason. Called from the failure dumper and, now, from the eval job's exit trap. `--tail` goes 1000 → 20000: 1000 was sized for a deploy-time failure, and truncating a full fan-out would undercount the errors this now exists to count.
- `scripts/test_ci_eval_trap.py` — asserts the green path reaches it.

Arms 2–4 add, on top: `EVAL_TASK_PARALLELISM` 4 → 6 in `ci-eval-pr.sh`; `AGENT_MODEL` preview → GA in `ci-eval-pr.sh` together with `MODEL_DEFAULT_NAME` in `ci-deploy.sh`, because the comment at `ci-deploy.sh:175` is explicit that flipping one alone files records under a model the install is not serving.

## Context

Comes out of a review of the 2026-09-09 → 2026-09-16 smoke-test runs. Related: #1491 (the nightly's deadline arithmetic), #1225 (worker death on 429s, closed 09-14), #1218.

Two things a reviewer should know before any of this merges:

- **Each project does bill its own Vertex quota**, but that is not what is binding. The 429 request URL carries the leased project — `hack/ci-deploy.sh:533` sets `platformAgent.harness.projectId=${PROJECT_ID}` and `charts/kube-agents/templates/litellm.yaml:16` defaults `VERTEXAI_PROJECT` from it. I could not confirm the shared-capacity reading against the Cloud Quotas API; the repo comment and the log shape are what it rests on.
- **Changing the agent model retires the baseline store.** `setup_id` carries the model, so every record filed under `gemini-3-1-pro-preview-kubeagents-mcp` stops counting and the rate gate has an empty window until it refills. That cost is why arms 3 and 4 are measurements, not a proposal.

Unverified: that `gemini-3.1-pro` is servable on Vertex in the pool projects. If it is not, arms 3 and 4 fail early at the first agent call, which is itself the cheap answer.

## Testing

`python3 -m pytest scripts/test_ci_eval_trap.py` — 6 passed. `bash -n` clean on the three CI scripts.

### Live validation

CI infrastructure: it cannot reach a running installation except through a Prow run, so the runs on this PR are the validation. Arm 1's check is specific and has never once passed before — a **green** build must publish a non-empty `artifacts/litellm.log`.

Per arm I will report here: the fan-out span from the `[PROFILE]` lines, the 429 count in `artifacts/litellm.log`, and the case verdicts.

## Self-Review

Not yet run — this is a draft opened to collect measurements, and the adversarial and docs-drift passes run before it leaves draft.

What I checked myself: `collect_litellm_log` ends in `|| true` and does nothing else, so it cannot disturb the `$?` that `dump_prow_artifacts_on_failure` reads on its first line, which is the constraint `collect_bench_results` already documents. The failure path calls the function instead of inlining `kubectl`, so red runs keep exactly the artifact they had. A red eval run captures twice, once from the trap and once from the dumper; the later write wins and is the fresher log.

## Risk & Rollout

Arm 1 is low risk: one extra `kubectl logs` per eval job, guarded by `|| true`, inside an EXIT trap that has already cleared `errexit`. The worst case is a larger artifact.

Arms 2–4 are not proposed for merge as written. If the parallelism change ships, the revert is the commit. If the model change ships, the revert is the commit but the baseline store does not revert with it, so that has to be planned around re-baselining.

Generated with the help of Claude Code (Opus 5).


lapis2002:
Arm finished, and the answer was no. gemini-3.5-flash ran the eval fan-out in 323-356 min against 135 min for gemini-3.1-pro-preview, far outside the run-to-run variance, so the model swap in this branch must not land.

Closing rather than repurposing: this branch was reused across the parallelism, region and flash arms, and the title, body and diff no longer describe each other.

Two pieces here are worth keeping and will come back as their own PRs:
- collect_litellm_log() in hack/ci-env.sh. Not ready as written -- it reads `deployment/litellm` with two replicas and captured the idle one, so the artifact held only health checks.
- The finding that no region can be tried with the pro preview, which is served on the global endpoint alone.

PR #1658, opened 2026-09-16
ci(eval): measure the presubmit matrix under spec.mode next (hold, never merges)

jayantid:
## Summary

This is a measurement pull request, not a change to merge. It flips the presubmit's install to `spec.mode: next` once the today-mode install has passed the existing readiness gates, so that `pull-kube-agents-smoke-test` runs the whole presubmit matrix against the next architecture and reports, case by case, where it fails. It is held from the moment it opens and will be closed once the run has been read.

## Why This Change

The next stack (NATS, the A2A gateway, the session-pod path and the agent-side skill overlay) has only been exercised at playground posture. Nothing today tells us which of the presubmit's eighteen cases pass under it, which fail, and whether the leased pool projects can even bring the stack up. Running the real gate once against `mode: next` answers that with three repetitions per case, the blocking roster, the rung ladder and the health bot's classification, on the fleet and GitHub fixtures a dev project does not have.

## What Changed

- `hack/ci-deploy.sh` gains a step 6b: after step 6 proves the today-mode install, it merge-patches the CR to `mode: next`, waits for the agent Deployment's generation to move (the operator's config-hash roll), then runs the same rollout gates step 6 uses over the NATS StatefulSet, the agent Deployment and the shell StatefulSet. The A2A gateway is reported, not gated: it crash-loops by design without a chat backend (#1660). On failure it dumps the CR status, the pods, Jobs and NetworkPolicies, recent events and the operator log. On success it prints the CR status, the rendered stack and the entrypoint's overlay lines into the artifact log.
- Nothing else. The chart deliberately does not render `spec.mode` (`docs/designs/spec-mode-switch.md`), which is why the flip is a patch and not a value.

## Context

Tracking issue: #1661. First finding: #1660.

Companion to a local run of the same matrix against a dev install in the author's project, which is the debugger for whatever this run reds. Related design: `docs/designs/spec-mode-switch.md`. The A2A gateway and worker image defaults still point at a dev registry (#1557); whether a leased pool project can pull them is one of the things this run measures.

## Testing

- `bash -n hack/ci-deploy.sh`.
- `tests/test_mode_grep.py` passes: the new block does not name the mode key, so the spec's two-site rule holds.

### Live validation

Not live-tested before opening: the change is the presubmit's own deploy step, and the presubmit run on this pull request is the live test. The patch-and-wait sequence is being exercised in parallel against a dev install (jayantid-gkedemos) by the companion local run; this section will be updated with what that run observed at the CR status, Deployment generation and pod layers once it has finished.

## Self-Review

Adversarial pass, run by the author in the session that wrote the change (no clean context was spawned, because this pull request is a throwaway measurement that never merges). Looked for: a `rollout status` racing the operator (handled by waiting for the Deployment generation to move before asking any workload for status), the mode grep rule (`tests/test_mode_grep.py`; the block names no env key), the no-magic-constants rule (names at the top of the file), and step 0 revalidation (the deploy script is not on the inert list, so the run is a full one). Docs-drift pass: none needed, the change documents nothing and is not meant to land. Open question the run answers rather than the review: whether `rollout status` on the NATS StatefulSet is the right signal or whether the provisioning Job has to complete first.

## Risk & Rollout

None to `main`: the pull request is held with `/hold` and will be closed, not merged. The run occupies the presubmit's single slot for one cycle and one pool-project lease, both of which the teardown trap releases.


jayantid:
/hold

Measurement pull request: it runs the presubmit matrix against `spec.mode: next` and is closed once the run has been read. It must not merge.

jayantid:
Second head (07ef3666) drops the A2A gateway from the rollout gate: it crash-loops by design without a chat backend, tracked in #1660. The first run on daf75637 would have timed out on that gate before any case ran.

jayantid:
Tracking issue for this measurement: #1661.

jayantid:
Run read: GREEN on 07ef3666 (build 2100310325382352896), 11 cases 3/3, 7 UNSTABLE on phrase misses, no catastrophic, no collapse. Per-case table and the stack findings are on #1661. Nothing further to run here.

jayantid:
Superseded by #1694, which carries this step 6b behind EVAL_MODE_NEXT with the corrected gates. The measurement it ran is recorded on #1661.

Issue #1505, opened 2026-09-11
test(operator): table-driven invariant over every NATS user grant in the rendered a2a config


bradhoekstra:
**Problem**

Four fixes narrowed one user's grants each: seed's JetStream API (f22284fd, #1306), worker's JetStream API (fce9f88e, #1393), worker's directory publish (3add25c6, #1313), plus the ack scoping in `TestSystemUsersAckGrantsAreScopedPerStream`. Each left a test for the user it fixed (`platformagent_a2a_manifests_test.go:2344`, `:2460`, `:1852`, `:145`). Nothing asserts the property across all users, so a fifth user added to `renderA2ANATSConf` (`platformagent_a2a_manifests.go:548-808`), or a grant widened on `web`, is covered by no test until someone writes another one-off. The spec states the property once for every user (`docs/designs/spec-nats-deployment.md:142-181`): deny by default, `$JS.API` enumerated per stream and verb, ack grants per stream, topic publishes exact, one inbox prefix per user.

**What to change**

Add `TestEveryNATSUserGrantIsEnumeratedAndStreamScoped` to `k8s-operator/internal/controller/platformagent_a2a_manifests_test.go`. Render with `buildA2ANATSConfigSecret(a2aTestAgent(), a2aTestCreds())`; read lists with the existing `a2aGrantSubjects`; match with `subjectMatches`.

Table, keyed by user, one row per APP user:

| user | streams it may name | account-level `$JS.API` verbs allowed | wholesale `$JS.API.>` |
| --- | --- | --- | --- |
| gateway | TASKS, KV_session-state | none | yes (recorded exception, spec :151-152) |
| worker | TASKS, KV_runtime-state, TOPICS-STATE, TOPICS-JOURNAL | none | no |
| seed | TASKS, DIRECTORY, TOPICS-STATE, TOPICS-JOURNAL | `$JS.API.INFO`, `$JS.API.STREAM.NAMES` | no |
| web | TASKS, DIRECTORY, TOPICS-STATE, TOPICS-JOURNAL | `$JS.API.INFO` | no |

Assertions, for every row, on both publish and subscribe lists:
1. Both lists exist and are non-empty (`a2aGrantSubjects` already fails otherwise).
2. No entry is `>`, `*`, `$JS.>`, `$JS.ACK.>`, `$KV.>`, or `$JS.API.>`, except the one wholesale entry the table records. Assert the set of users holding `$JS.API.>` is exactly `{gateway}`.
[...]

PR #1672, opened 2026-09-17
fix(operator): scope the gateway's JetStream API grant to the streams and verbs it uses

### LINKED ISSUE #1666: A2A gateway holds $JS.API.> and can delete the TASKS stream
The `gateway` user in the rendered `nats.conf` holds the wholesale JetStream API wildcard, which covers STREAM.UPDATE, PURGE and DELETE on every stream in the APP account.

`gatewayIdentity` in `k8s-operator/internal/controller/platformagent_a2a_identities.go:150` renders this publish list:

```
"a2a.tasks.*.*.in",
"a2a.tasks.*.*.supervisor",
"$KV.session-state.>",
"$JS.API.>",          <-- this one
"$JS.ACK.TASKS.>",
"$JS.FC.>",
"_INBOX.gateway.>"
```

### This is the last holder of a grant this file already argues against

The wildcard was removed from `seed` in #1306 and from `worker` in #1316, and `a2aJetStreamSurfaceRationale` — prose in this same file, at `:218` — states the reason every *callout* principal enumerates its JetStream subjects one at a time:

> A grant list is a capability surface for JetStream, not a read/write distinction. Subject permissions cannot see a request BODY, and a consumer's target stream and its delivery subject are both body fields. So `$JS.API.>` hands back everything the subject lists withhold — a push consumer on TASKS delivering into a subject the principal CAN subscribe to reads the whole task plane, and STREAM.DELETE destroys it. Both demonstrated live against the rendered config, and the same escape the web user's comment below records.

The gateway is a *static* principal rather than a callout one, so it fell outside the sweep that narrowing was written for. Nothing in `gatewayIdentity` argues the wildcard is deliberate: the comment block above the grant list justifies the `$JS.ACK.TASKS.>` scoping in detail — explicitly so that this user cannot `+TERM` another principal's in-flight delivery — and then takes `$JS.API.>` without a word. Scoping the ack subject while leaving the API wildcard beside it does not narrow anything: `$JS.API.CONSUMER.DELETE.TASKS.*` is inside the wildcard.

Also worth stating plainly, because it is the part that makes the ack scoping moot: the gateway's own comment says it is static only as a sequencing fact — "It has a ServiceAccount and could authenticate with it tomorrow." When it does move to the callout, the rationale at `:218` applies to it by its own terms.

### Measured, not inferred

Confirmed against a live install rather than only against the source. On the `a2a-next-dev` dev cluster (`mode: next`), reading the rendered `platform-agent-a2a-nats-config` Secret off the cluster shows the wildcard present in the shipped config, and connecting as `gateway` with the password from `platform-agent-a2a-nats-creds`:

| Call | Result |
| --- | --- |
| `STREAM.INFO TASKS` | allowed |
| `STREAM.DELETE TASKS` | allowed — the stream was destroyed, and only a re-run of the provision Job restored it |

One `nats stream rm TASKS` as `gateway` removes the task plane: every task's event history, and with it the replay path that `tasks/get` and every accountability claim about the bus depend on. It also takes the five live consumers with it. The provision Job recreates the stream empty on the next reconcile, so the loss is silent from the operator's side — the install goes back to looking healthy with no history in it.

(For the avoidance of doubt about the test method: that deletion was the setup step for unrelated live validation on a dev cluster under a held lease, not an attack demo. It is reported here because it is the evidence.)

### Scoping looks feasible from the callers

The gateway's real JetStream use is TASKS and the session-state KV: `TasksGet` (an ordered consumer, `a2a/lib/fold.go`), `SubscribeDurable` for the `gateway-relay` durable on TASKS (`a2a/lib/client.go:492`), and the session registry over `$KV.session-state` (`a2a/gateway/registry.go`). Nothing in `a2a/gateway/` creates, updates, purges or deletes a stream, and nothing there touches DIRECTORY, TOPICS-STATE or TOPICS-JOURNAL through the JetStream API.

`a2aWorkerJetStreamGrants` (`platformagent_a2a_manifests.go:390`) is the shape this would take, and it was derived the same way. I have deliberately not written the exact list here — it should be derived from the callers by whoever implements it and proven against a rendered config, not copied from a issue report.

### Why now

Found while working #1665, where the create-only property of provisioning turns on which principals can edit a stream. The claim I wanted to make — that nothing on this bus can edit `TASKS` — is false because of this grant, and #1665 says so rather than overclaiming. Filing separately rather than absorbing it: it is a grant change to a different principal with its own blast radius, and it deserves the same treatment #1306 and #1316 got.

Not a duplicate of #1316 (worker, closed by the narrowing) or #1505 (the table-driven grant invariant, closed).

bnaylor:
## Summary

The NATS `gateway` user held `$JS.API.>` wholesale, which covers `STREAM.DELETE`, `STREAM.PURGE`, `STREAM.UPDATE` and `STREAM.MSG.DELETE` on every stream in the APP account. This replaces the wildcard with the eight JetStream API subjects the gateway's own binaries emit, by stream and by verb, rendered from one function (`a2aGatewayJetStreamGrants`). Nothing in the list names DIRECTORY, the topic streams, `KV_runtime-state` or `KV_cap`; no destructive verb survives on any stream. The proof is an embedded nats-server running the rendered `nats.conf`: 24 gateway operations succeed with zero publish violations, 30 out-of-scope calls are refused server-side by exact subject, and the same render with the wildcard put back deletes TASKS. `seed` came off this wildcard in #1306 and `worker` in #1393; with the gateway, no rendered principal holds it.

## Why This Change

`$JS.API.STREAM.DELETE.TASKS` as `gateway` destroys the task plane in one call: every task's event history, the replay path `tasks/get` and every accountability claim about the bus rest on, and the five live consumers on the stream. #1666 measured exactly that against the rendered config on a dev install. The loss is silent from the operator's side — the provision Job recreates the stream empty on the next reconcile, so the install goes back to looking healthy with no history in it.

The wildcard also made the ack scoping beside it moot. `$JS.ACK.TASKS.>` is scoped, with a comment explaining in detail that an ack subject names a stream and a consumer and never the caller, so this user must not be able to `+TERM` another principal's in-flight delivery — while `$JS.API.CONSUMER.DELETE.TASKS.*` sat inside the `$JS.API.>` on the next line.

Nothing argued the wildcard was deliberate here. `a2aJetStreamSurfaceRationale`, prose in the same file, already states why a grant list is a capability surface for JetStream rather than a read/write distinction, and why `$JS.API.>` hands back everything the subject lists withhold. It was written for the *callout* principals; the gateway is static, so it fell outside that sweep. That is the whole reason it was last, and it is not a reason to keep it.

## What Changed

- `a2aGatewayJetStreamGrants()` is the list, and its comment is the argument for every entry. Read out of nats.go v1.53.1 for each call the gateway path makes, then measured. TASKS: `STREAM.INFO` (`js.Stream` in `lib.TasksGet`), `CONSUMER.CREATE.TASKS.>` (the `gateway-relay` durable and the replay's ordered consumer), `CONSUMER.MSG.NEXT.TASKS.*`, `DIRECT.GET.TASKS.>` (the replay horizon). `KV_session-state`, the session registry: `STREAM.INFO` (`js.KeyValue` binds a bucket by reading its stream), `DIRECT.GET.…>` (`kv.Get`, which every registry read goes through), and `CONSUMER.CREATE.…>` with `CONSUMER.DELETE.….*` (`kv.ListKeysFiltered` is a push ordered consumer nats.go creates and deletes on `Unsubscribe`; the reap loop runs it on a timer). Eight subjects.
- `CONSUMER.CREATE` ends in `>` rather than naming the durable, and both reasons are load-bearing: the relay carries **two** filter subjects since the supervisor split, and nats.go puts the filter in the API subject only when there is exactly one — so the pair's create is `…TASKS.gateway-relay` while a single-filter rebind is four tokens longer — and the replay's ordered consumers take server-generated names. Both shapes are rows in the server test.
- Withheld on purpose, each with its reason in the comment: account `INFO` (`jetstream.New` never asks), `STREAM.MSG.GET` (every stream is created `--allow-direct`, so nats.go emits `DIRECT.GET` and never the fallback), `CONSUMER.INFO`, and `CONSUMER.DELETE` on TASKS.
  - `CONSUMER.INFO` costs something and the comment says so: the gateway can no longer read back the state of a consumer it owns. Nothing on the path binds a consumer by name, and on v1.53.1 neither `Consume` nor an ordered iterator re-verifies itself with `Info()` after a reconnect — held by execution in `TestGatewayConsumersSurviveABusRestart`, not by reading, because a relay that goes quiet after a bus bounce with nothing in its own log is the failure class this deployment has been bitten by.
  - `CONSUMER.DELETE.TASKS` is the one subject nats.go emits here without a grant: the ordered consumer's reset fires `DeleteConsumer` in a goroutine and ignores the result. What it buys is removing this user's route to another principal's durable by name, a session pod's three consumers included, and it is not free — the adversarial pass caught me calling it free. Two costs, both on the reset path: the refused request reaches nats.go's async error handler, which `a2a/lib` logs at Error, so a bus bounce during N in-flight replays writes N "permissions violation" lines about a refusal that is by design; and the pre-reset ephemeral waits out its five-minute inactive threshold holding a consumer slot instead of going immediately. The per-replay ephemeral is pre-existing (stopping an ordered iterator never deleted its consumer); what this adds is the reset path's. Both are in the grant comment, and the restart canary asserts that this refusal is the only violation a bounce produces.
- `TestGatewayJetStreamGrantOnARealServer` (new file, `platformagent_a2a_gateway_authz_test.go`): the rendered config under an embedded nats-server, seven streams provisioned as seed with the provision script's flags, a client as gateway. Table 1 runs the gateway's real operations and asserts zero `Publish Violation` lines for gateway; table 2 sends the destructive and out-of-scope calls and asserts each refusal from the server log naming the exact subject, then re-reads the streams as `web` to show nothing changed. A subtest measures the two residues the grant cannot close. A final subtest restores `$JS.API.>` in the render and shows `STREAM.DELETE.TASKS` succeeding and the stream gone — #1666 reproduced.
- `TestGatewayHoldsNoWholesaleJetStreamAPI` is the shape check in #1306's and #1393's form: the publish list pinned exactly, every forbidden verb against every provisioned stream through `subjectMatches`, a structural bound on the grant function's own output, and the subscribe list pinned too. `TestGatewayGrantCoversEveryJetStreamSubjectItsCallersEmit` asks the other direction — every subject the callers put on the wire is inside the grant, written from the call sites rather than from the list — and `TestGatewayGrantNamesTheBucketItsDataPlaneWritesTo` binds the bucket's two spellings (the `$KV.session-state.>` literal in the identity, the `a2aSessionStateBucket` constant in the grants) and both to the provision script.
- `TestOnlyTheseIdentitiesHoldTheBareJetStreamAPIGrant` now expects the **empty** set, and its comment says what holds an empty expectation up: the per-principal table in `TestEveryNATSUserGrantIsEnumeratedAndStreamScoped`, whose `gateway` row loses `wholesale: true` and gains the five verbs on TASKS and five on `KV_session-state`.
- `platformagent_a2a_bridge_authz_test.go` (renamed from `..._worker_...` by #1653): the consumer-info oracle moves from `gateway` to `web`. It read consumers back on the strength of the gateway's `$JS.API.>`; `web` holds `CONSUMER.INFO.TASKS.*` by enumeration and is now the only principal that can answer. ~~Its `TOPICS-STATE` read moves to the worker's own granted `DIRECT.GET`.~~ **Update 9/18:** after #1653's split the topic streams belong to `agent`, so that read is a subject-filtered `STREAM.INFO` as seed instead; see the fifth conflict point above. No bridge assertion changed.
- `a2a/gateway/live_test.go`: `findLatestLiveTask` used `$JS.API.STREAM.MSG.GET.TASKS` with a comment saying "both worker and gateway hold it". That was already false — #1393 took it off the worker — and is doubly false now, so the env-gated live test could not have run as written. It uses direct get, which both users do hold, with the 404 path handled.
- Docs: the deployment spec's JetStream-tax bullet names `a2aGatewayJetStreamGrants` beside the worker's and states that no rendered principal holds the wildcard.
- `a2a/authcallout/testdata/rendered-nats.conf` is a golden file (`TestRenderedNATSConfMatchesTheCalloutFixture`, regenerated with `-update`), and the a2a callout suite starts a real server from it. Regenerated: the gateway block is where its line ~122 `$JS.API.>` lived.

## Context

- Closes #1666.
- #1306 (`seed`) and #1393 (`worker`) are the two prior narrowings; this reuses their shape — `a2a<User>JetStreamGrants`, exact pin plus cross-product plus structural bound, a real-server refusal table, a wildcard control — so the three read as one argument. #1505 is the table-driven grant invariant this updates a row in.
- #1653 (`a2a/auth-5-retire-worker`, open, breaking) retires the static `worker` user and splits it into `agent` and `bridge`. No semantic overlap — different principals, disjoint grant lists — but it diverges from this branch in five places, and whichever lands second rebases. Four are mechanical: `platformagent_a2a_manifests.go`'s grant-function region (it adds two functions immediately above where this adds one) and the stream-name constant block (it edits the comment above the constants; this adds `a2aSessionStateBucket` to them); the `renderA2ANATSConf` doc comment and the deployment spec's JetStream-tax bullet, where both rewrite the same sentence — after both land it should name seed, bridge, agent and gateway, and keep this branch's "no rendered principal holds a bare `$JS.API.>`" rather than #1653's "a user still holding `$JS.API.>` holds playground posture"; and `a2a/authcallout/testdata/rendered-nats.conf`, which both regenerate and which `TestRenderedNATSConfMatchesTheCalloutFixture` guards.
  The fifth is worth naming because a wrong resolution fails slowly rather than loudly: #1653 keeps `gw.Consumer(…)` and `gw.Stream(ctx, "TOPICS-STATE")` as the oracle in `platformagent_a2a_worker_authz_test.go`, and under this grant the gateway holds neither `CONSUMER.INFO` nor anything on TOPICS-STATE, so those calls do not error — they hang to the test's 60-second context. ~~This branch's `webJS`/`js` oracle is correct under both branches and is the resolution to take.~~  **Update 9/18, after merging #1653:** half right.  The `webJS` oracle for the three `CONSUMER.INFO` reads is correct under both and was taken.  The `js` oracle for the TOPICS-STATE read-back is not: #1653 did not just rename `worker`, it *split* the grant and gave the topic streams to `agent`, so after the merge no principal this test can dial holds a content read there -- `js` (now bridge) has nothing on TOPICS-STATE, gateway has nothing after this branch, `web` holds `STREAM.INFO` but no `DIRECT.GET`, and `agent` is callout-authenticated with no password the embedded server accepts.  That read is now a subject-filtered `STREAM.INFO` through the subtest's existing seed helper, asserting the probe subject carries zero messages -- same property, readable by a principal that exists. #1653 also rewrites `findLatestLiveTask` off `STREAM.MSG.GET` independently (through `lib.Connect` with the caller switched to `bridge`); either rewrite is fine, and the direct-get subject this one uses is proven as a row in the server test.
- #1665 (`a2a/bus-claims-enforced`, open) touches `platformagent_a2a_manifests.go` in the provision-script and reconcile regions only, which this does not go near, and `platformagent_a2a_worker_authz_test.go` in the helper region rather than the body edits here. Its body carries a sentence this falsifies, deliberately and with the reason stated there: *"(For accuracy: the `gateway` user does hold `$JS.API.>` and so can edit a stream. That is its own pre-existing grant and is not what provisioning runs as.)"* Once this lands, that parenthetical can go, and #1665's argument that nothing unattended can edit a stream gets stronger rather than weaker — worth an edit to that PR body or a one-line follow-up, not a blocker either way.
- Duplicate-checked again before opening: `gh pr list --search "gateway JetStream"`, `"gatewayIdentity"`, `"JS.API"` and the open-issue list. Nothing else touches this grant.
- Conformance (#1121): no `tests/conformance/` assertion owed; this narrows an existing grant and moves no invariant boundary. The suite's `test_A_authority` reads `gatewayIdentity`'s publish list out of the Go source, and the builder keeps the `publish := []string{…}` + `append` shape those readers handle — the literals stay plain strings so the parser's "served config and Go map agree" precondition still holds.

**Handoff for the PR minder.** Verified by execution: every row of both tables above, on nats-server 2.14.6 (the module pin), including the wildcard control deleting TASKS and the restart canary that holds the `CONSUMER.INFO` decision. Verified by reading nats.go v1.53.1 and then confirmed by the server test: which API subject each call formats, that `InitialConsumerPending` reads cached state rather than issuing `CONSUMER.INFO`, and that `Unsubscribe` deletes a KV watcher's ephemeral. Assumed, not verified: that `natscli`'s `kv add` sets `allow_direct` on the bucket — the grant's `DIRECT.GET.KV_session-state.>` rests on it, the test provisions through nats.go's `CreateKeyValue` which always sets it, the live store on `a2a-next-dev` shows `allow_direct=true` on all seven streams, and if an install ever had it false the symptom is a refused `STREAM.MSG.GET` in the bus log rather than silence. Not live-tested, and why: the only `mode: next` install carries another session's stacked build and was read-only for this change, so nothing here rolled an operator or patched a CR; the mechanism proof is the embedded server, which is what #1393 shipped with. The #1653 and #1665 conflict notes are two bullets up; #1653 is the one that will actually conflict.

## Testing

- `make -C k8s-operator test` (manifests, generate, fmt, vet, python, go test with `KUBEBUILDER_ASSETS` set): green, `internal/controller` 92.1% coverage, no SKIPs in the envtest cases. `make fmt` reformats the untouched `platformagent_litellm_policy_test.go` (two trailing blank lines); that churn is reverted and not in this diff.
- `cd a2a && go vet ./... && go test ./... -count=1`: green, including `authcallout`, which starts a real nats-server from the regenerated fixture, and `gateway`.
- `make conformance`: 122 tests, OK, 12 declared expected failures.  **Update 9/18:** these were 118/13 when this was written; the change is #1653's, not this branch's -- `origin/main` reports 122/12 identically.
- `make docs-check` clean; `prettier@3.9.6 --check` clean on the changed Markdown.
- Red proof, on `upstream/main`'s render with this branch's tests: five of the seven new or changed tests fail. The server test's first refusal row, `$JS.API.STREAM.DELETE.TASKS`, answers `{"success":true}` and every row after it reads `stream not found` because the stream is already gone — #1666, reproduced by the test that is meant to prevent it. The shape test fails on the exact pin and the forbidden cross-product, the identities test reports the set as `[gateway]`, the table reports `principals holding $JS.API.> = [gateway]`, and the bucket-pairing test fails on the missing grant. Two pass and should: the caller-coverage table (a wildcard trivially covers every caller subject — it asks the other direction) and the restart canary (a canary, not a refusal test).

Embedded server, gateway connected with `--inbox-prefix _INBOX.gateway`, rendered config unchanged except ports, websocket and store dir. 24 operations, all allowed, 0 publish violations:

| Gateway operation (the nats.go call its code makes) | Result |
| --- | --- |
| `CONSUMER.CREATE TASKS` durable `gateway-relay`, two filter subjects (`gateway.Run`) | allowed |
| the same durable with one filter in the API subject, then back to the pair (rebind) | allowed |
| publish `a2a.tasks.platform.t1.in`; publish `…supervisor` (the supervisor's terminal) | allowed |
| `CONSUMER.MSG.NEXT TASKS gateway-relay` (relay `Consume`), `$JS.ACK` (ack floor read back as web = 1) | allowed |
| `STREAM.INFO TASKS`, `DIRECT.GET` last-for-subject on `…events` and `…supervisor` (`lib.TasksGet`) | allowed |
| ordered consumer on TASKS over both subjects, two replay messages (`TasksGet`) | allowed |
| `DIRECT.GET TASKS` with a wildcard subject token (the route `live_test.go` takes) | allowed |
| `STREAM.INFO KV_session-state` (`js.KeyValue`), `Create`, `Put`, `Get`, task-index put/get, `Delete` | allowed |
| `ListKeysFiltered("sessions.>")` — consumer create and delete on the bucket (the reap scan) | allowed |
| publish violations logged for gateway across all of the above | 0 |

30 refused, each read from the server log as a `Publish Violation` naming gateway and the exact subject:

| Refused call | Result |
| --- | --- |
| `STREAM.DELETE` / `PURGE` / `UPDATE` / `MSG.DELETE` / `RESTORE` / `SNAPSHOT` on TASKS | refused |
| `STREAM.MSG.GET TASKS` (the non-direct read fallback, which nats.go never emits here) | refused |
| `CONSUMER.DELETE` on `gateway-relay` and on the bridge's durable; `CONSUMER.INFO` on its own durable | refused |
| every verb tried on DIRECTORY: `STREAM.INFO`, `PURGE`, `DELETE`, `CONSUMER.CREATE`, `DIRECT.GET` | refused |
| `STREAM.INFO` and `DIRECT.GET` on TOPICS-STATE and TOPICS-JOURNAL; `STREAM.DELETE TOPICS-JOURNAL` | refused |
| `KV_runtime-state` info and direct get; `KV_cap` info and consumer create | refused |
| `STREAM.PURGE` / `DELETE` on its own `KV_session-state` | refused |
| `$JS.API.INFO`, `STREAM.NAMES`, `STREAM.LIST`, `CONSUMER.NAMES TASKS`, `STREAM.CREATE EVIL` | refused |
| afterwards, read as web: TASKS still 3 msgs, DIRECTORY 0, the relay durable still there | yes |

The two residues, measured rather than claimed shut, both on subjects the grant permits so no violation is logged for either. `CONSUMER.CREATE` is create-or-update by name and the server has no ownership concept for a consumer name, so the gateway retuned the bridge's durable through one permitted call. And a push consumer's `deliver_subject` is a body field no subject grant can see: aimed at `a2a.agents.platform` with the gateway holding a literal subscription on that card subject, DIRECTORY went 0 → 3, all three under `a2a.tasks.*` subjects — a persisted write and an eviction lever, not forgery, since a card read by subject never sees them. Neither is new; `$JS.API.>` permitted both with every stream as a source rather than one. What closes them is per-task consumers created by the dispatcher, not a grant — the same answer `a2aWorkerJetStreamGrants` reaches. The gateway is also the principal they matter least for: it is the requester and the supervisor, so it already writes the task plane by grant.

Control, same render with `"$JS.API.>"` restored in the gateway block: `STREAM.PURGE.TASKS` answers `{"success":true,"purged":0}`, `STREAM.DELETE.TASKS` answers `{"success":true}`, and web's `STREAM.INFO.TASKS` then returns stream not found.

### Live validation

Not live-tested against a running install, and this is a deliberate choice rather than an absence of a cluster, so here is the whole of it.

The chain this change touches only exists under `spec.mode: next`: the operator rewrites `nats.conf` in `platform-agent-a2a-nats-config`, the config hash rolls the NATS StatefulSet, and the gateway reconnects under the new allow-list. There is exactly one `mode: next` install in the dev project (`gke_bnaylor-kagents-dev_northamerica-northeast1_a2a-next-dev`), it carries another session's stacked build mid-flight, and it is read-only for this change — rolling an operator there would land an unrelated branch's render on someone else's in-progress validation. The other dev cluster is a default-mode install and renders no A2A bus at all, so it cannot exercise this. Standing up a second `mode: next` install to watch one allow-list change is not proportionate to a grant edit, and it is not what the two prior narrowings of this same wildcard did: #1393 (`worker`) and #1306 (`seed`) both shipped on the embedded-server proof, with #1393 merged on it.

What that leaves unwatched, stated rather than implied: the reconcile rewriting the Secret, the StatefulSet rolling on the new config hash, and a deployed gateway reconnecting and relaying under the narrowed grant with a clean bus log. The embedded server covers the authorization half of that chain exactly — it runs the operator's own rendered `nats.conf`, unmodified except ports, websocket and store dir — and the restart canary covers the reconnect half, including a `tasks/get` replay and a `ListKeysFiltered` reap across a bus bounce. What it cannot cover is the Kubernetes plumbing that delivers the config, which is the same plumbing every other grant edit has used and which this change does not alter.

Observed there without mutating, 2026-09-17 — one `get secret`, no writes:

- `secret/platform-agent-a2a-nats-config`, gateway block: `"$JS.API.>"` is live today. That build also predates the supervisor split (`a2a.tasks.*.*.events` where main now renders `…supervisor`), so it is the wildcard on an older render rather than on main's.

The mechanism proof is the embedded-server tables above. Not covered: a deployed `nats:2.10-alpine` running this render (the worker's narrowing measured identical tables on 2.10.29 and 2.14.5, and nothing in this grant's shape is version-sensitive in a way that measurement did not already cover, but it was not re-run here); a real gateway Deployment reconnecting under the new grant outside the embedded canary; and the env-gated `a2a/gateway/live_test.go`, whose direct-get route is exercised as a row in the server test but which was not itself run, since it needs a port-forward to an install this change may not roll.

## Self-Review

Two passes, each in a fresh subagent handed the worktree path, the range `upstream/main...HEAD` and the skill file, nothing else. Adversarial and docs-drift, per AGENTS.md. Both ran their own nats-server measurements rather than reading only.

**Adversarial** looked for: a JetStream API subject the gateway path emits that the list misses; whether withholding `CONSUMER.INFO` or `CONSUMER.DELETE` breaks a reconnect; routes the narrowed grant leaves open; whether the tests can see a re-widening; the KV watcher's consumer names and the direct-get dependency; and the sibling-PR overlap. It re-ran the red proof with the grant function returning the wildcard, and the restart canary. Findings:

1. **Fixed, and it corrected a claim.** MEDIUM: the grant comment said withholding `CONSUMER.DELETE.TASKS` "changes no behaviour". It changes two, both on the ordered consumer's reset path: the refused request reaches nats.go's async error handler, which `a2a/lib`'s client logs at Error level, so a bus bounce during N in-flight `tasks/get` replays writes N "permissions violation" lines about a refusal that is by design; and the pre-reset ephemeral is no longer deleted immediately, so it holds a TASKS consumer slot for its five-minute inactive threshold. The comment now states both costs and says the grant is still not worth adding. The pass's second suggestion — downgrade that one violation to Warn in `a2a/lib`'s error handler — is **not taken**: it changes runtime logging for every client of the library to suppress a line that is a true statement about a real refusal, and the place to decide that is a change to the client's error handling with its own blast radius, not this grant edit. The restart canary asserts this is the only violation the reset produces, so the noise is bounded and measured rather than unremarked.
2. **Fixed.** MEDIUM: the #1653 overlap is five conflict points, not the three I had written, and one of them fails in a way worth warning about — that PR keeps `gateway` as the consumer-info oracle in the worker authz test, which under this grant does not fail but *hangs* to the 60-second context. The Context section above now lists all five and says which resolution is correct under both branches.
3. **Not fixed, measured rather than argued.** LOW: `"$KV.session-state.>"` is a literal in the gateway's publish and subscribe lists while the JetStream grants are built from `a2aSessionStateBucket`, and construction would beat a test. I tried it. The conformance suite reads `gatewayIdentity`'s publish list out of the Go source with a regex, and a constant concatenation there is read as two grants: `test_A3_precondition_every_served_user_is_built_by_an_identity` fails with `['$KV.', '.>', …] != [...]`. That precondition exists because a principal whose list reads as empty satisfies every writer-set assertion vacuously, which is how a real `worker` grant once hid. So the literal stays and `TestGatewayGrantNamesTheBucketItsDataPlaneWritesTo` holds it to the constant; the identity's comment says why.  **Update 9/18, after merging #1653:** ~~the measurement above no longer reproduces~~ -- that PR taught the conformance reader to resolve constant concatenations (`_resolve_expr`/`_resolve_grant_helper`), and re-running the same mutation now gives `Ran 122 tests / OK (expected failures=12)`.  The literal is still worth keeping, because the two-spellings bind is what a bucket rename breaks silently, but the stated reason for keeping it is gone.

Cleared by the pass, for the record: no missing grant on the gateway path; the KV watcher's generated name and the legacy reset's delete-and-recreate are both inside the grant; the nameless `CONSUMER.CREATE.<stream>` form is emitted only below server 2.9 and the operator deploys 2.10; withholding `CONSUMER.INFO` survives a reconnect (executed); `it.Stop()` never deleted an ordered consumer, so the per-replay ephemeral is pre-existing; the `wholesale` table field is not dead; and nothing in `a2a/gateway` touches DIRECTORY, the topic streams or the other buckets. It flagged one `authcallout` startup-timeout flake, which passed alone and on rerun and uses a synthetic map rather than the regenerated fixture.

**Docs-drift** looked for: any prose, comment or fixture in the tree stating what the gateway may do on the bus; surviving `$JS.API.>` claims; generated and golden regions; the doc map; and whether the new comments' nats.go and issue citations hold. No Blocking findings; seven Non-blocking, all fixed:

1. `tests/conformance/test_A_authority.py`'s two docstrings named `worker` and `seed` as the builders read only partially; the gateway now builds its list the same way.  ~~Both say `worker`, `seed` and `gateway`.~~  **Update 9/19:** that edit was `96e370ab`, and merging #1653 dropped it.  A5 taught the conformance reader to resolve constant concatenations, so `gateway` is not a partial read any more and `seed` is the only one left; main's own rewrite of those docstrings says exactly that.  This branch's net diff to the file is empty -- `git diff origin/main...HEAD -- tests/conformance/test_A_authority.py` returns nothing.  The correction is #1653's, not this PR's.
2. The `wholesale` field's comment described a principal that "still holds" the wildcard when no row sets it any more. It now says the field is kept so that taking the wildcard back is one visible word on a row.
3. Five comments said "#1666 took it", crediting the issue with the change. Reworded to "the narrowing #1666 asked for".
4. `docs/README.md`'s bus-principal row named only `platformagent_a2a_identities.go`, but the JetStream halves of those grants live in `platformagent_a2a_manifests.go` — true for seed and worker already and now for the gateway. The row names both files.
5. `a2a/authcallout/session_integration_test.go`'s synthetic callout map is now the only place a principal called `gateway` holds `$JS.API.>`. The wildcard stays, because that entry creates the TASKS stream the session cases run against and the subject under test is the callout's session derivation; a comment says it is a stage prop that does not track the rendered gateway.
6. Pre-existing, fixed while in that file: a comment calling `CONSUMER.LIST`/`STREAM.LIST` "the blanket the shared worker user holds today", stale since #1393.
7. The ordered-consumer name citation pointed at `ordered.go`; the `nuid.Next()` seed is in `jetstream.go`'s `OrderedConsumer`. Corrected.

**kube-agents-bot** found nothing in the code and raised one item on this description, plus two low-severity notes it folded rather than counted. Dispositions:

- **Fixed.** A third conformance docstring still said "the two that concatenate Go constants" after the gateway became the third; the docs-drift pass caught two of the three and this one was missed. ~~Now "three".~~  **Update 9/19:** superseded the same way as docs-drift finding 1 above.  A5 rewrote that passage and the phrase is not in the file at all now.
- **Fixed, in the description.** The bot read "Not live-tested" as reaching for the template's escape hatch on a form-2 change. Fair as written. The Live validation section above now gives the whole argument — one `mode: next` install, mid-flight with another session's stacked build and read-only for this change; a default-mode cluster that renders no bus; a second install not proportionate to a grant edit; and the two prior narrowings of this same wildcard shipping the same way — and names exactly what goes unwatched as a result.
- **Not fixed, with the argument.** It noted that `TestGatewayGrantNamesTheBucketItsDataPlaneWritesTo` restates invariants the table test already holds. The overlap is real for the per-verb checks, and the part that does not overlap is the reason the test exists: nothing else in the suite binds the `$KV.session-state.>` literal in the identity to the `a2aSessionStateBucket` constant the JetStream grants are built from. Those two spellings are what a bucket rename breaks, one silently — and the literal cannot become the constant, for the conformance-parser reason in finding 3 above. Keeping it.

It also verified independently, against the module cache, every nats.go claim the new comments make (the two `CONSUMER.CREATE` subject shapes, the reset's fire-and-forget delete, `Consume`'s reconnect path issuing no `Info`, `getMsg`'s direct-get selection, and the KV bind/get/watch/unsubscribe chain), and that `make docs-check`, the golden fixture and the map staleness window are clean.

Not covered by either pass: a deployed `nats:2.10-alpine` running this render, the KV bucket's `allow_direct` on a real install (`nats kv info session-state` would settle it), and the legacy KV watcher's reset path executed rather than read. The consumer-budget interaction was in this list and came out of it through review; see the round below.

**Review round: @jayantid, two Mediums, both fixed in `c0fe3e69`.** They are the same
observation from two sides — this change makes the gateway emit, by design, the one log line
the Risk section below tells an operator to read as a fault.

- **The expected refusal had no user-facing home.** The Go comment states it; `docs/` does
  not, and `grep -rn "CONSUMER.DELETE" docs/` returns one unrelated hit. The JetStream-tax
  bullet in `spec-nats-deployment.md` now names
  `$JS.API.CONSUMER.DELETE.TASKS.<server-generated name>` as the only violation a rendered
  principal produces by design, so "any other one is a missed grant" stays a usable rule. One
  correction to the finding, in its favour: the restart canary says it too, in a comment and a
  `t.Logf` — nothing anyone debugging a live install would read.
- **The lingering reset ephemeral competes for a consumer budget sized from sessions alone.**
  Two of the finding's claims did not survive measurement, and they point opposite ways. The
  headroom is *smaller* than the sixteen it assumes: that budget's own comment itemizes all
  sixteen — two standing durables, one audit slot, three for a session's retirement overlap,
  ten web-rail readers — and not one is a replay ephemeral, so the slack is at most fourteen
  and under load zero. But the trigger is *not* a bus bounce. The ordered consumer is
  `MemoryStorage: true, Replicas: 1` (`jetstream/ordered.go:635`), so a restarting server drops
  it and the reset creates the only survivor: measured 1 → 0 → 1 on TASKS across a restart, and
  1 → 2 when the connection is dropped with the server left up. A bounce costs the violation
  and no slot; a heartbeat gap costs two slots per in-flight replay for five minutes — nats.go's
  ordered-consumer default, which `lib.TasksGet` does not override.

  The grant comment now names the constraint rather than the other branch, since AGENTS.md L196
  forbids citing a PR as the reason a behaviour exists: whatever sizes TASKS' `max_consumers`
  has to count concurrent `tasks/get` replays beside the session cap, because a budget that
  counts sessions alone refuses a legitimate session's consumer create under replay load and
  reads as an undersized stream. Not taken: giving the replay's ordered consumer an
  `InactiveThreshold` of its own. That is a runtime change in the `a2a` module arriving on a
  grant-scoping PR, and the gap is on the sizing side.

## Risk & Rollout

Renders only under `mode: next`, and changes one user's allow-list in a Secret; the next reconcile rewrites `nats.conf` and the config hash rolls the NATS StatefulSet, which is the existing mechanism for any grant edit. If a subject were missed the failure is loud in the bus log — a `Publish Violation` naming gateway and the subject, with `$JS.API.CONSUMER.DELETE.TASKS.<name>` the one such line this change produces on purpose — and quiet in the gateway, which would wait out its context on that one call; the server test is designed to surface exactly that as the refused subject, and the caller-coverage table is the cheap version that names which caller lost its route. The narrowest real risk is a gateway revision that starts binding a consumer by name, which would need `CONSUMER.INFO` back with an argument. Reverting restores the wildcard. No new dependency: the nats modules were already test dependencies of the operator module from #1393.


a2a/authcallout/session_integration_test.go
```diff
@@ -45,6 +45,15 @@ const (
 // sessionMap is one ordinary entry and one narrowed one. The narrowed entry
```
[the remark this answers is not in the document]
bnaylor: Fixed in the description rather than in the code, and you are right that "Not live-tested" was doing too much work on its own.

The Live validation section now carries the whole argument instead of the label: there is exactly one `mode: next` install in the dev project, it is mid-flight with another session's stacked build and read-only for this change (rolling an operator there lands an unrelated branch's render on someone else's in-progress validation), the other dev cluster is default-mode and renders no A2A bus at all, and standing up a second `mode: next` install to watch one allow-list change is not proportionate to a grant edit. The two prior narrowings of this same wildcard — #1306 for `seed`, #1393 for `worker` — both shipped on the embedded-server proof, and #1393 merged on it.

It also now names what goes unwatched as a result, which is the part your comment is really asking for: the reconcile rewriting the Secret, the config hash rolling the StatefulSet, and a deployed gateway reconnecting under the narrowed grant. The embedded server covers the authorization half of that chain exactly — it runs the operator's own rendered `nats.conf`, unmodified except ports, websocket and store dir — and `TestGatewayConsumersSurviveABusRestart` covers the reconnect half, including a `tasks/get` replay and a `ListKeysFiltered` reap across a bus bounce. What neither covers is the Kubernetes plumbing that delivers the config, which this change does not alter.

Your two folded notes: the third conformance docstring was a real miss and is fixed in 96e370a. On `TestGatewayGrantNamesTheBucketItsDataPlaneWritesTo`, the per-verb overlap with the table test is real, but the part that does not overlap is why it exists — nothing else binds the `$KV.session-state.>` literal in the identity to the `a2aSessionStateBucket` constant the JetStream grants are built from, and those are the two spellings a bucket rename breaks, one of them silently. The literal cannot become the constant: the conformance suite reads that publish list out of the Go source with a regex, and a concatenation there is read as two grants (`['$KV.', '.>']`), failing the precondition that keeps a principal's list from reading as empty. Measured, not assumed.

kyber775 (approved):
## Code Review: PR #1672 (`fix(operator): scope the gateway's JetStream API grant to the streams and verbs it uses`)

**Reviewer**: Kyber (Day-2 SRE & Security Invariants Lens)  
**Target**: `upstream/main...HEAD` (`0d5035e0`)  
**Verdict**: **APPROVE**  
**Execution Declaration**: `executed: true` (ran `go test ./internal/controller/... -run "TestGateway"` in `k8s-operator`, full `a2a` test suite including `authcallout` and `gateway`, and local verification against embedded nats-server).

---

### Key Architectural & Security Validations:

1. **Elimination of Destructive Capabilities on Task Plane (`#1666`)**
   - Stripping the wholesale `$JS.API.>` wildcard removes `STREAM.DELETE`, `STREAM.PURGE`, `STREAM.UPDATE`, and `STREAM.MSG.DELETE` from the gateway principal.
   - The embedded server verification in `TestGatewayJetStreamGrantOnARealServer` demonstrates that restoring the wildcard destroys `TASKS` in one call, whereas under this scoped grant, destructive verbs are strictly refused server-side with zero publish violations on normal gateway traffic.

2. **Accurate Verb & Subject Scoping (`a2aGatewayJetStreamGrants`)**
   - The 8 granted subjects in `platformagent_a2a_manifests.go` accurately match the wire-level calls emitted by `nats.go` v1.53.1 for `TASKS` and `KV_session-state`.
   - Scoping `DIRECT.GET` over `STREAM.MSG.GET` leverages the `--allow-direct` stream configuration, avoiding fallback requests.
   - Preserves push consumer cleanup on `KV_session-state` via `CONSUMER.DELETE...*` for the reap scan.

3. **In-Tree Test Architecture**
   - The dual-table validation (`TestGatewayJetStreamGrantOnARealServer`) running against real `nats-server` (v2.14.6) proves both the positive operational envelope (24 allowed ops) and negative security boundaries (30 refused calls).
   - `TestOnlyTheseIdentitiesHoldTheBareJetStreamAPIGrant` confirms that zero rendered principals now hold un-scoped `$JS.API.>` wildcards across the entire deployment.

---

### CI Blocker & Fix:
- **`Validate Repo Structure` failed on `make shellcheck`**:
  - `In scripts/installer/installer_common.sh line 481: "~" | "~/"*)` tripped `SC2088` (tilde in quotes) introduced when #1651 merged into main.
  - Adding `# shellcheck disable=SC2088` immediately before `case "$path" in` at line 480 makes `make shellcheck` pass cleanly.

---

/lgtm

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -567,6 +568,132 @@ func a2aWorkerJetStreamGrants() []string {
...
+//
+// CONSUMER.DELETE on TASKS is withheld too, and it is the one subject nats.go
+// emits here without a grant. The only emitter is the ordered consumer's
+// reset path, which fires DeleteConsumer in a goroutine and ignores the
+// result (jetstream/ordered.go, reset). What it buys is removing this user's
+// route to another principal's durable by name -- a session pod's three
+// consumers included -- and it is not free. Two costs, both on the reset path
+// and neither on the steady state:
+//
+//   - The refused request reaches nats.go's async error handler, which in
```
jayantid, line 741: 🟡 **UX flow.** The PR's own diagnostic for a missed grant is "a Publish Violation naming gateway", and this change makes the gateway produce one by design on every bus bounce.

The Risk section says a missed subject is loud in the bus log as a violation naming gateway and the subject. Withholding `CONSUMER.DELETE.TASKS` means the ordered consumer's reset path now logs exactly that shape, at Error, N times per bounce during N replays. The only place that says this line is expected is this Go comment. `docs/designs/spec-nats-deployment.md` gets the wildcard bullet updated but not this, and there is no operator page for the bus yet.

Whoever debugs a relay that went quiet after a bounce reads the gateway log, finds the violation, and concludes the grant is broken.

Fix: one sentence in the deployment spec's JetStream-tax bullet naming `$JS.API.CONSUMER.DELETE.TASKS.<ordered-name>` as the one refusal the gateway emits by design, so "any other violation is a missed grant" stays a usable rule.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -567,6 +568,132 @@ func a2aWorkerJetStreamGrants() []string {
...
+//     a2a/lib's client logs at Error level (a2a/lib/client.go, the
+//     ErrorHandler option). So a bus bounce or a heartbeat gap while N
+//     tasks/get replays are in flight writes N "permissions violation for
+//     publish to $JS.API.CONSUMER.DELETE.TASKS.<name>" lines into the
+//     gateway's own log, for a refusal that is by design. Measured in
+//     TestGatewayConsumersSurviveABusRestart, which asserts that this is the
+//     only violation the restart produces.
+//   - The pre-reset ephemeral is not removed immediately; it waits out its
```
jayantid, line 749: 🟡 **Technical design.** The lingering reset ephemeral competes for the TASKS consumer budget that #1665 sizes from sessions alone, and neither PR accounts for the other.

This comment says the pre-reset ephemeral now holds a TASKS consumer slot for its five-minute inactive threshold. #1665 renders `--max-consumers=max(64, maxSessions*3 + 16)`, so at full session load the headroom for every replay ephemeral, pre-existing and now the reset's too, is sixteen. Your body lists this interaction as not covered by either pass.

A bus bounce during more than eight in-flight `tasks/get` replays at full session load exhausts the budget for five minutes, and #1665 reports the next session's consumer-create refusal as the operator's configuration error.

Fix: whichever lands second either counts replay ephemerals in the headroom term or gives the replay's ordered consumer a short `InactiveThreshold` of its own. A sentence in the grant comment naming the dependency is enough for this PR.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Fixed in `c0fe3e69`. Every claim held, including the one I most expected to be loose: the canary already prints the shape you describe, `"APP/user:gateway" - Publish Violation - Subject "$JS.API.CONSUMER.DELETE.TASKS.N0twTuv..._1"`, and `grep -rn "CONSUMER.DELETE" docs/` returns exactly one hit, `spec-nats-deployment.md:258`, which is about the `web`/`worker` consumer residues and not this. The `docs/` half of this PR touches only the `docs/README.md` principal row and the JetStream-tax bullet, so you are right that the spec got the wildcard change and not this one.

Took the fix as prescribed, appended to the JetStream-tax bullet at :186:

> Enumerating leaves one refusal that is expected rather than a fault: the gateway's `tasks/get` replay reads through an ordered consumer whose reset path publishes `$JS.API.CONSUMER.DELETE.TASKS.<server-generated name>` and ignores the answer, so a reconnect writes one refusal per replay in flight into the bus log and into the gateway's own log at Error. That subject is the only violation a rendered principal produces by design; any other one is a missed grant.

One correction to the finding, in your favour: the Go comment is not quite the only place that says it — `TestGatewayConsumersSurviveABusRestart` says it too, in a comment and a `t.Logf`. Nothing a person debugging a live install would read, which is the part that mattered.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
bnaylor: Fixed in `c0fe3e69`, and two of the claims did not survive being measured. Worth writing both out, because one of them cuts against you and one of them cuts for you.

**The formula is exactly right.** `a2aTasksConsumerBudget` on that branch is `resolveA2AMaxSessions(agent)*a2aSessionConsumersPerSession + a2aTasksReservedConsumers` = `maxSessions*3 + 16`, floored at 64 by `a2aTasksMaxConsumersFloor`.

**The headroom is smaller than sixteen, so the finding understates itself.** That branch's own comment itemizes all 16: the `gateway-relay` and bridge durables (2), audit-durable headroom (1), one session's retirement overlap (3), and ten web-rail readers. Not one of them is a replay ephemeral. The slack actually free for replay is at most 14 and under load can be 0.

**The trigger is a reconnect, not a bounce.** The ordered consumer is created `MemoryStorage: true, Replicas: 1` (`jetstream/ordered.go:635`), so a restarting server drops it and the reset creates the only survivor. Measured on the rendered config: a restart goes 1 → 0 → 1 consumer on TASKS. Forcing a disconnect with the server left up goes 1 → 2, with the refused `$JS.API.CONSUMER.DELETE.TASKS.W27jFHY6..._1` in the log. So a bus bounce costs the violation and no extra slot; a heartbeat gap or a dropped connection is what costs two slots per in-flight replay for five minutes. The "more than eight" arithmetic is self-consistent at 16 free ÷ 2, but 16 is the upper bound it never reaches.

The five minutes is nats.go's ordered-consumer default and not ours — `lib.TasksGet` sets only `FilterSubjects` and `DeliverPolicy`, so it takes it.

Took your second option: the grant comment now names the constraint. AGENTS.md L196 forbids citing a PR number as the reason a behaviour exists, so it is written as a rule for whatever sizes the budget rather than as a pointer:

> Those slots are the gateway's and never a session's, so whatever sizes TASKS' `max_consumers` has to count concurrent `tasks/get` replays beside the session cap. A budget that counts sessions alone runs out under replay load and refuses a legitimate session's consumer create, which reads as an undersized stream.

I did not give the replay's ordered consumer an `InactiveThreshold` of its own. That is a runtime change in the `a2a` module landing on a grant-scoping PR, and the gap is on the sizing side. Relaying this to the other branch so whichever lands second closes it there.

kyber775 (approved):
## Code Re-Review: PR #1672 (Delta: `96e370ab` → `c0fe3e69`)

**Reviewer**: Kyber (Dual-Engine: Gemini Day-2 SRE + DeepSeek Adversarial Lens)  
**Base/Head**: `upstream/main...HEAD` (`c0fe3e6991ae336ed1d827ffe99eb8643417e206`)  
**Verdict**: **APPROVE**  
**Execution Declaration**: `executed: true` (verified clean three-way merge via `git merge-tree`, ran full `k8s-operator` controller test suites including `TestGateway*`, `a2a` test suite, `make conformance` 118 tests, and `make docs-check`).

---

### Delta Audit & Verification (`96e370ab` → `c0fe3e69`)

Commit `c0fe3e69` explicitly closes the two review threads raised by @jayantid regarding the expected refusal on reconnect and the resulting consumer slot dynamics:

1. **Resolution of 1672-F1 (UX & Bus Log Diagnostic Clarity):**
   - In `docs/designs/spec-nats-deployment.md` (lines 207–212), the deployment specification now explicitly documents `$JS.API.CONSUMER.DELETE.TASKS.<name>` as the sole publish violation emitted by design when the gateway's ordered consumer resets on reconnect.
   - Operators auditing bus logs or seeing `permissions violation for publish to $JS.API.CONSUMER.DELETE.TASKS...` at Error level can deterministically distinguish this expected client-side cleanup failure from accidental grant regressions.

2. **Resolution of 1672-F2 (TASKS Consumer Budget Sizing Invariant):**
   - In `k8s-operator/internal/controller/platformagent_a2a_manifests.go` (lines 661–675), the grant documentation articulates the 5-minute default `InactiveThreshold` inherited from `nats.go` (`jetstream/ordered.go`), and how in-flight replays during reconnects against an active server temporarily hold two slots on `TASKS`.
   - Properly identifies that these ephemeral replay slots must be accounted for alongside session capacity when sizing `TASKS` `max_consumers`.

---

### Non-Blocking Observations & Suggestions (Folded for Future Refinements):

- **[SUGGESTION] Minor terminology refinement in deployment spec:**
  - `docs/designs/spec-nats-deployment.md:209` names `$JS.API.CONSUMER.DELETE.TASKS.<server-generated name>`. In `nats.go` v1.53.1 (`jetstream/ordered.go`), the ordered consumer name is seeded client-side using `nuid.Next()` (`<nuid>_<serial>`) before issuing `CONSUMER.CREATE`. Functionally opaque to the server either way, but technically client/library-generated.
- **[SUGGESTION] Replay consumer footprint across components:**
  - In addition to the gateway, `hermes-bridge` also invokes `lib.TasksGet` during task lookup and sweep cycles (`a2a/hermes-bridge/bridge.go`, `sweep.go`). The documented ephemeral consumer budget requirement applies to any component executing `lib.TasksGet`.
- **[SUGGESTION] Runtime follow-up for `InactiveThreshold`:**
  - As noted in the thread, tuning `InactiveThreshold` (e.g. 15–30s instead of 5m) within `lib.TasksGet`'s `OrderedConsumerConfig` on a follow-up `a2a` PR will allow orphaned replay ephemerals to be reclaimed rapidly without relying on the stream max-consumer headroom.

---

### Invariant & Test Suite Results:
- `git merge-tree upstream/main HEAD`: Clean tree `9c99186001052ced5dac85e43d42699e7dd01850` (0 conflicts)
- `go test ./internal/controller/... -run "TestGateway"`: `ok (11.138s)`
- `k8s-operator` test suite: `ok (40.578s)`
- `a2a` test suite: `ok (authcallout 32s, gateway 24s, all green)`
- `make conformance`: `Ran 118 tests in 2.029s: OK (expected failures=13)`
- `make docs-check`: Passed clean across all 313 files.

/lgtm

bnaylor:
Synced with main.  Head is `852d1509` (merge `0e10f3a4`, parents `c0fe3e69` + `965b6917`).  **Merge, not rebase** -- `c0fe3e69` is unchanged and still reachable.

#1653 landed and retired the static `worker` user, splitting it into `bridge` and `agent`.  This branch predicted five conflict points in its Context section; the merge found six, and one of the five predictions was half wrong.  Body updated in place for the three claims that are no longer true (the oracle resolution, finding 3's reason, the conformance counts) -- struck through rather than rewritten, so what reviewers already read is still visible.

### The six conflicts

| file | resolution |
| --- | --- |
| `platformagent_a2a_manifests.go` | this branch's argument on main's vocabulary.  The constant header is now "the bridge's, the agent's and the gateway's"; `renderA2ANATSConf`'s summary names all four grant functions and keeps "no rendered principal holds a bare `$JS.API.>` any more".  `a2aGatewayJetStreamGrants()` is a distinct function from the one #1653 renamed -- confirmed the rename did not bleed into it |
| `platformagent_a2a_identities_test.go` | kept the empty expectation and its "what holds this up" argument, carrying #1653's A5-split sentence verbatim in substance |
| `platformagent_a2a_bridge_authz_test.go` | main's rename and file, with this branch's `webJS` oracle re-applied at all four consumer reads.  The TOPICS-STATE read needed a different answer -- below |
| `a2a/gateway/live_test.go` | **took main's rewrite whole.**  Both branches rewrote `findLatestLiveTask` off `STREAM.MSG.GET`; main's goes through `lib.Connect`/`ReadTopicLatest` rather than a raw request, which is strictly better.  Kept this branch's one surviving edit, the beat-2 comment, which is correct under main's route too |
| `docs/designs/spec-nats-deployment.md` | all three grant functions named, this branch's "no rendered principal holds it any more" kept |
| `tests/conformance/test_A_authority.py` | **resolved entirely to main** -- byte-identical to `origin/main` now.  #1653 rewrote the Go reader to resolve constant concatenations, so this branch's docstring edits (adding `gateway` to the "read only partially" list) assert the opposite of what is true.  No assertion weakened, nothing made tautological |

### The one place this went beyond a merge

The body said this branch's `webJS`/`js` oracle was correct under both branches.  That holds for the three `CONSUMER.INFO` reads.  It does not hold for the TOPICS-STATE read-back, because #1653 didn't only rename `worker` -- it **split** the grant, and the topic streams went to `agent`.  After the merge, no principal this test can dial holds a content read on TOPICS-STATE: `js` is bridge and has nothing there, gateway has nothing there after this branch, `web` holds `STREAM.INFO` but not `DIRECT.GET`, and `agent` is callout-authenticated with no password the embedded server accepts.  Every candidate hangs to the 60s context.

Replaced `st.GetLastMsgForSubject(ctx, "a2a.topics.shared.probe")` with a subject-filtered `STREAM.INFO` through the subtest's existing seed helper, asserting `State.Subjects["a2a.topics.shared.probe"] == 0`.  Same property -- if the diverted delivery had been rewritten onto the probe subject, that subject would carry it.  **Proved non-vacuous:** repointing the filter at `a2a.tasks.platform.t1.in` fails with "a read of the probe topic by subject sees 1 messages".  Restored byte-clean.

### A silent break the merge could not flag

`platformagent_a2a_gateway_authz_test.go` is new on this branch, so it never conflicted -- and it read `creds.Data["worker-password"]` and connected as `"worker"` in three places.  Neither exists any more.  Retargeted to `bridge`, which holds `a2a.tasks.platform.*.events` (what those rows publish) and `CONSUMER.CREATE.TASKS.>` (what the create-or-update residue row needs).  `852d1509` fixes three more present-tense claims the merge falsified: `KV_runtime-state` is the bridge's bucket, the sibling shape test is `TestBridgeHoldsNoWholesaleJetStreamAPI`, and the callout suite's `CONSUMER.LIST`/`STREAM.LIST` note described a user that no longer exists.  Historical prose about `worker` was left alone, as were the `worker-adapter` component and the session-worker naming.

This is the argument for merging a dependency rather than waiting: a conflict marker only appears where both sides edited the same lines, and a new file that names a retired credential produces none.

### Verification

`k8s-operator/internal/controller` `ok 93.2s`; all of `a2a/...` green.  `go build` + `go vet` clean on both modules.  `make docs-check` clean.  `make conformance` 122/OK/12 expected failures -- **identical on `origin/main`**, so the 118/13 in the body was #1653's drift, not this merge's.

#1653's own tests and the seven this branch owns, run together:

```
--- PASS: TestTheWorkerCredentialIsGone
--- PASS: TestBridgeJetStreamGrantOnARealServer (21.49s)
--- PASS: TestGatewayJetStreamGrantOnARealServer (10.76s)
--- PASS: TestGatewayConsumersSurviveABusRestart
--- PASS: TestGatewayHoldsNoWholesaleJetStreamAPI
--- PASS: TestGatewayGrantCoversEveryJetStreamSubjectItsCallersEmit
--- PASS: TestGatewayGrantNamesTheBucketItsDataPlaneWritesTo
--- PASS: TestOnlyTheseIdentitiesHoldTheBareJetStreamAPIGrant
--- PASS: TestEveryNATSUserGrantIsEnumeratedAndStreamScoped
```

The deliver-subject subtest still measures the residue post-merge: `TOPICS-STATE grew 0 -> 3, 3 of them under a2a.tasks.* subjects`.

### Mutations re-run after the merge

**Red proof** -- `a2aGatewayJetStreamGrants()` returns `[]string{"$JS.API.>"}`: reproduces the body exactly, the same five fail and the same two pass.  `$JS.API.STREAM.DELETE.TASKS` answers success and every later row reads `stream not found` -- #1666 reproduced.  `TestOnlyTheseIdentitiesHoldTheBareJetStreamAPIGrant` reports `[gateway]`.

**Finding 3 no longer reproduces.**  The body says concatenating the `$KV.session-state.>` literal breaks `test_A3_precondition_every_served_user_is_built_by_an_identity` with `['$KV.', '.>', …]`, and that was the stated reason the literal stays.  #1653 taught the reader to resolve constants, so the same mutation now gives `Ran 122 tests / OK (expected failures=12)`.  The literal is still worth keeping -- the two-spellings bind is what a bucket rename breaks silently, and `TestGatewayGrantNamesTheBucketItsDataPlaneWritesTo` is what catches it -- but the reason in the body was wrong and is now struck.

All three restored byte-clean; `git status` empty after each.

kyber775 (approved):
## Code Re-Review: PR #1672 (Post-Rebase & Merge Delta: `c0fe3e69` → `0e10f3a4` → `852d1509`)

**Reviewer**: Kyber (Dual-Engine: Gemini Day-2 SRE + DeepSeek Adversarial Lens)  
**Target**: `upstream/main...HEAD` (`852d1509d699cac7481dd0186bdb01688a2b2f4c`)  
**Verdict**: **APPROVE**  
**Execution Declaration**: `executed: true` (verified clean three-way merge via `git merge-tree`, ran full `k8s-operator` test suite including embedded `nats-server` authz suites, full `a2a` test suite, `make conformance` 122 tests, and `make docs-check`).

---

### Delta Audit & Merge Invariant Verification (`c0fe3e69` → `852d1509`)

Commit `0e10f3a4` merged upstream `main` (`965b6917` — A5 / #1653, retiring the static `worker` bus user and splitting it into `bridge` and `agent`), followed by `852d1509` reconciling residual prose. The adversarial audit confirmed that the merge cleanly reconciles the two grant domains without privilege expansion or grant regressions:

1. **Gateway JetStream Grants Intact & Scoped:**
   - The eight JetStream API subjects rendered in `a2aGatewayJetStreamGrants()` remain strictly bounded to `TASKS` (`STREAM.INFO`, `CONSUMER.CREATE.TASKS.>`, `CONSUMER.MSG.NEXT.TASKS.*`, `DIRECT.GET.TASKS.>`) and `KV_session-state` (`STREAM.INFO`, `DIRECT.GET`, `CONSUMER.CREATE`, `CONSUMER.DELETE`).
   - Zero destructive verbs survive (`STREAM.DELETE`, `PURGE`, `UPDATE`, `MSG.DELETE`), and account-level enumeration/discovery verbs remain refused.
   - `TestOnlyTheseIdentitiesHoldTheBareJetStreamAPIGrant` asserts an empty set across all rendered principals.

2. **Principal Split & Test Oracle Realignment:**
   - **`platformagent_a2a_bridge_authz_test.go`**: Successfully switched the consumer-info oracle to `web` (`webJS.Consumer`), recognizing that neither `bridge` nor `gateway` retains `CONSUMER.INFO.TASKS.*`. Reworked `TOPICS-STATE` verification to subject-filtered `STREAM.INFO` as `seed`, respecting that `agent` is callout-authenticated and holds no static password.
   - **`platformagent_a2a_gateway_authz_test.go`**: Executor rows appropriately dial as `bridge` to publish events on `a2a.tasks.platform.t1.events`.
   - **`platformagent_a2a_manifests_test.go`**: The grant matrix reflects the `gateway` verb slices on `TASKS` and `kvSessionState`, and `TestGatewayHoldsNoWholesaleJetStreamAPI` properly verifies the absence of wholesale wildcard grants alongside `TestBridgeHoldsNoWholesaleJetStreamAPI`.

---

### Non-Blocking Observation:

- **[SUGGESTION] `platformagent_a2a_gateway_authz_test.go:198-202`**:
  The comment references the direct-get route that `a2a/gateway/live_test.go` takes, describing it as a "raw request". On `main`, `findLatestLiveTask` was refactored through `client.ReadTopicLatest` (which wraps `GetLastMsgForSubject`). A minor doc comment update can note that `ReadTopicLatest` is the caller exercising `DIRECT.GET.TASKS.>`.

---

### Invariant & Test Verification Results:
- `git merge-tree upstream/main HEAD`: Clean tree `299a5b7591343b29144e0d826a6d282bb3e5b587` (0 conflicts)
- `k8s-operator/internal/controller`: `ok (47.514s)` (all tests pass, including `TestGateway*` and `TestBridge*` running on embedded `nats-server 2.14.6`)
- `a2a/...`: `ok` across all packages (`authcallout`, `gateway`, `hermes-bridge`, `worker-adapter`, `lib`)
- `make conformance`: `Ran 122 tests in 2.085s: OK (expected failures=12)`
- `make docs-check`: Passed clean across all 313 Markdown files and 1298 code files.

/lgtm

bnaylor:
Status, since this has moved twice since the last human read of it.

[@jayantid](https://github.com/jayantid) your comments from 2026-09-17 are answered and the
threads are resolved; nothing is outstanding on the PR.  After that, A5 (#1653) and A6 (#1665)
both landed on `main`, and this branch merged them up at `852d1509` (merge `0e10f3a4`, parents
`c0fe3e69` + `965b6917`) — `git merge`, not a rebase, as this branch always does.  Every check
is green on that head.

tide reads `Not mergeable.  Needs approved label.`  `lgtm` is on, so an OWNERS approver is the
only thing left.

bnaylor:
Synced with main after #1748 merged (`fd2e22e7`), which put this branch in `CONFLICTING`.  Merge, not rebase, per this PR's history rule.

One conflict, in `docs/README.md`, two table rows each edited on both sides.  Resolved per row rather than per file:

- **Bus principal set** — took mine.  It is main's text plus the `k8s-operator/internal/controller/platformagent_a2a_manifests.go` (`a2a*JetStreamGrants`) clause, which is this PR's own subject.
- **Bus token contract** — took main's.  It is mine plus `A2ABusTokenAudience` and the bus Secret names in `common_types.go`, both of which arrived with #1748.

Both `a2a*JetStreamGrants` and `A2ABusTokenAudience` appear exactly once in the merged file.  Everything else auto-merged clean: `a2a/gateway/live_test.go`, `docs/designs/spec-nats-deployment.md`, `platformagent_a2a_bridge_authz_test.go`, `platformagent_a2a_manifests.go`, `platformagent_a2a_manifests_test.go`.

Verified locally on the merge commit, not on either parent:

- `k8s-operator`: build + vet clean.  `internal/controller` ok 105.784s (the runtime is the tell that the envtest cases actually ran rather than skipping on an unset `KUBEBUILDER_ASSETS`), `internal/webhook` ok 0.783s, `api/v1alpha1` ok 0.429s.
- `a2a`: build + vet clean, all 10 packages ok — `authcallout` 33.7s, `gateway` 27.9s, `worker-adapter` 23.3s, `hermes-bridge` 14.3s, `lib` 9.3s, and the five `cmd`/`profiles` packages.
- `prettier --check docs/README.md` clean; `make docs-check` passes (links, terminology, the 278-doc map, no maintainer identifiers on published pages, instruction budget).

kyber775 (approved):
## Re-Review: PR #1672 (Commit `f2af8cbc`) — fix(operator): scope the gateway's JetStream API grant to the streams and verbs it uses

**Verdict:** `[APPROVE]`
**Reviewed Commits:** `852d1509..f2af8cbc` (Merge re-review on HEAD `f2af8cbcf38c8d84c645a787f7ab5e3651f62a3b`)

---

### Delta Audit & Reconciliation (`852d1509` → `f2af8cbc`)

Commit `f2af8cbc` cleanly merges `upstream/main` (#1748), reconciling the bus credential source check with the scoped JetStream API grant:

1. **[PASS] Clean Adjacency Resolution (`docs/README.md`):**
   - Bus principal set row preserves the `platformagent_a2a_manifests.go (a2a*JetStreamGrants)` reference.
   - Bus token contract row adopts `A2ABusTokenAudience` and the bus Secret names from `common_types.go`.
   - Both references appear exactly once; formatting confirmed clean.

2. **[PASS] Grant Invariant & Authz Verification:**
   - Gateway JetStream API grants remain strictly scoped to `TASKS` (`STREAM.INFO`, `CONSUMER.CREATE.TASKS.>`, `CONSUMER.MSG.NEXT.TASKS.*`, `DIRECT.GET.TASKS.>`) and `KV_session-state` (`STREAM.INFO`, `DIRECT.GET`, `CONSUMER.CREATE`, `CONSUMER.DELETE`).
   - Zero destructive verbs survive (`STREAM.DELETE`, `PURGE`, `UPDATE`, `MSG.DELETE`).
   - `TestOnlyTheseIdentitiesHoldTheBareJetStreamAPIGrant` asserts an empty set across all rendered principals.

---

### Empirical Verification

- **Mergeability:** Clean against `upstream/main` (`git merge-tree` exit 0).
- **Unit & Authz Execution:**
  - `k8s-operator`: `internal/controller` ok (50.5s, all embedded `nats-server` suites green), `internal/webhook` ok (0.12s), `api/v1alpha1` ok (0.04s).
  - `a2a`: All 10 packages pass clean (74s total, including `authcallout` at 32.5s and `gateway` at 24.5s).
- **Documentation:**
  - `make docs-check`: Passed clean across all 313 Markdown files and 1322 code files.

All invariants hold. Ready to merge.

a2a/authcallout/session_integration_test.go
```diff
@@ -45,6 +45,15 @@ const (
...
 // carries no grants at all — that is the shape the operator renders and the
 // shape ParseIdentityMap insists on.
+//
```
jayantid, line 48: Reviewed against product requirements, UX flow and technical design at `f2af8cb`; nothing to flag.

jayantid (approved):

a2a/authcallout/session_integration_test.go
bnaylor: Thanks for the review.  Resolving this one only to clear the `hold-unresolved-threads` gate -- there is nothing outstanding in it.

PR #1678, opened 2026-09-17
docs(bench): the eval harness cannot port-forward to a sandboxed agent

kirill-kosov:
## Summary

The bench eval harness reaches the in-cluster agent with `kubectl port-forward`, and that cannot reach a pod running under GKE Sandbox. `ENABLE_GVISOR` defaults to true, so this is what a stock install gives you, and the failure is opaque: the forward binds its local port before anything dials the pod, so the harness sees an open port, takes the tunnel for established, and each request then dies inside the sandbox — recorded as three transport failures and then an infrastructure failure the scorer drops from grading, so the case ends with no verdict and nothing naming the sandbox. This change documents the constraint, what the run record actually looks like, and the workaround that needs no cluster change, along with what that workaround does not cover. It does not change the transport.

## Why This Change

Issue #971 finding 1a. Without this, a contributor following `.agents/rules/eval_driven_development.md` builds a dev install per `INSTALL.md`, runs `uv run devops-bench`, and gets a case with no verdict and a `RemoteDisconnected` in the log — which is also what a pod that went away mid-run produces. Nothing in the repository connected that to the sandbox they did not know they had asked for. What knowledge existed sat in `tests/e2e/conftest.py` and `hack/ci-deploy.sh`, in comments a bench user has no reason to read, and both stated it wrongly.

Not fixing the transport is deliberate. Eval CI never hits this: `hack/ci-deploy.sh` pins `runtimeClassName` empty and says why ("what the sandbox does to the agent is the release pipeline's to exercise, not a smoke test's"), and it runs before `hack/ci-eval-pr.sh` in the same Prow wrapper. For the dev-install case a working no-code workaround already exists, because `_ensure_port_forward` never assumes it owns the transport.

## What Changed

- `bench/README.md` gains a `## Sandboxed installs` section: the mechanism, what the run record looks like, the relay command, the two ways the relay differs from the harness's own transport, and the unsandboxed-install alternative. `bench/CUSTOM-TASKS.md`, the `harness.py` module docstring and `.agents/rules/eval_driven_development.md` link to it rather than restating it.
- The `harness.py` docstring records that nothing in that module notices the failure, and that telling it apart from an agent that is down takes a request rather than a port check — so the next reader neither adds a port check that cannot work nor concludes it is undetectable.
- `tests/e2e/conftest.py`, `hack/ci-deploy.sh` and `tests/test_ci_deploy_runtime_class.py` named `scripts/exec_tunnel.py` canonical on the constraint; it defers to `docs/site/src/content/docs/operator/platformagent-crd.md` at its own line 13, so all three now point there, keeping the `exec_tunnel.py` pointer where a reader needs the relay itself. `conftest.py` also said the RC sets `ENABLE_GVISOR=true` "against install.sh's `false` default"; the default is `true`, so the comment had the stock shape backwards.

## Context

Part of #971. Findings 1b/1c/1d/3/4/6/8 landed in #1562 and finding 7 in #1605; 1a was the remainder. No open PR does this work. #1490, #1628 and #1112 touch the same files and are merge-conflict risk only.

The empirical finding below — that `_ensure_port_forward` returns cleanly on a sandboxed install — forecloses the obvious follow-up of adding a diagnostic there. An earlier revision of this branch did exactly that; the live test disproved its premise and it was dropped.

## Testing

- `make docs-check` — passes.
- `bench/tests/` — 728 passed, unchanged from the branch point.
- `tests/test_ci_deploy_runtime_class.py` — passes (docstring-only edit).

### Live validation

Against a dev install on GKE Autopilot (us-east4), namespace `kubeagents-system`.

The constraint was verified rather than assumed. In a scratch namespace I ran a pod with `runtimeClassName: gvisor` (scheduled onto an auto-provisioned node labelled `sandbox.gke.io/runtime=gvisor`) serving HTTP on 8080, plus a standard-runtime control. Against the sandboxed pod `kubectl port-forward` **binds and stays up** — a TCP connect to the local port succeeds and the harness's `_port_open()` returns `True` — and only the per-connection dial fails:

```text
E0916 ... failed to execute portforward in network namespace
  "/var/run/netns/cni-...": dial tcp4 127.0.0.1:8080: connect: connection refused
error: lost connection to pod
```

The client sees `RemoteDisconnected` (`curl: (52) Empty reply from server`). The control pod returned 200. This is what the README now describes, and it is why the diagnostic was dropped: the failure never reaches the path it would have run on.

The workaround was then verified end to end against the real agent, not a scratch pod:

```text
python3 scripts/hermes-dashboard-tunnel.py --container agent-api-auth \
  --remote-port 8643 --local-port 8642
→ GET /v1/models 200 {"object":"list","data":[{"id":"model-default",...,"owned_by":"hermes"}]}
```

The `owned_by: hermes` body proves the request reached Hermes through the credential-proxy sidecar. With the relay up, `_ensure_port_forward(8642)` returned having spawned **zero** kubectl processes, and the request still worked afterwards — the no-op claim in the README, executed. The command in the README is this command.

Not covered: the two halves were proven separately — a sandboxed pod that port-forward cannot reach, and the relay reaching the real agent — rather than running a full eval against a sandboxed install of the agent itself, which would mean reinstalling the dev cluster with gVisor on. The relay's two documented limitations (it resolves the pod against the current kubectl context, and a provisioning task repoints that context mid-run) are derived from source — `scripts/exec_tunnel.py` passes no `--context` on either kubectl call, and `harness._cluster_hint` documents the repoint — not reproduced against a provisioning task.

Cleanup: the scratch namespace was deleted (`namespace "bench-hint-check" deleted`); the tunnel process was killed.

## Self-Review

Six passes, each in a subagent context that did not write the change: adversarial and docs-drift against the first code-bearing revision, both again against the documentation-only revision, and both again against the rebased commit. The last round found two Blocking:

- **The symptom was described in words the source contradicts.** The README said the failure "does not present as a transport failure"; the harness calls it exactly that — `_connection_dropped` → `_TransportError(retryable=True)` → `opening turn failed in transport (n/3)` → an `INFRA_FAILURE` the scorer excludes. The same commit's docstring said the opposite of the README. Fixed: the section now quotes the log line the reader will actually grep, says the run ends with no verdict rather than a failure, and warns that `RemoteDisconnected` does not distinguish a sandbox from an agent that went away.
- **"Change nothing else" was false for the workflow the same README prescribes.** The relay passes no `--context` to kubectl, while the harness pins `AGENT_CLUSTER_CONTEXT` on every call — and a task that provisions a cluster repoints the current context mid-run, which is the hazard that variable exists for. Since the relay opens one `kubectl exec` per connection, every request after the repoint lands on the task cluster and fails the way a sandbox does. Fixed: the claim is gone and both limitations are stated, with provisioning tasks sent to an unsandboxed agent.
- **"This module cannot detect it" overclaimed** — `tests/e2e/conftest.py` detects precisely this with one HTTP round trip, and says why a port check cannot. Fixed: the docstring now says nothing here notices, and names what it would take.
- **The canonical-pointer sweep left a straggler.** `tests/test_ci_deploy_runtime_class.py` — the test pinning the `ci-deploy.sh` comment this change edits — still cited `exec_tunnel.py` as canonical, so the two would have disagreed. Fixed. Re-pointing `ci-deploy.sh` had also dropped the only pointer telling a CI reader what to use instead of port-forward; both are now named, the shape `conftest.py` already used.
- **"The route CI takes" conflated two mechanisms.** CI overrides the chart value directly and never sets `ENABLE_GVISOR` or runs the installer. Fixed, and the installer route now notes that a full `upgrade.sh` is a whole-composition apply that drops the gVisor node pool on a Standard cluster.
- **The RC's sandbox is not the default falling through.** `ENABLE_GVISOR` is a required per-environment variable there. The default is what makes the shape stock, not what makes the RC sandboxed. Fixed in the `conftest.py` comment to say both without offering one as the other's cause.
- **`.agents/rules/eval_driven_development.md` was the highest-value missing doc** (both docs-drift passes). It is the file that tells a contributor to build a dev install and run the harness, and it had no pointer. Fixed — one sentence. The final pass confirmed the other candidates need nothing: `docs/testing-map.md` already delegates to `bench/README.md`, `bench/cuj/` never port-forwards, `INSTALL.md` already documents `--gvisor=false`, and the site carries the constraint in three places.
- **Earlier rounds** caught a citation to a function that does not exist (`buildPlatformDeployment`; the label is set in `buildPodTemplateSpec`), and three remedies that do not work — a bare `helm upgrade --set runtimeClassName=""`, `upgrade.sh` without editing `install.env` first, and the `port_forward_agent` fixture as a worked example, which deliberately refuses `AGENT_LOCAL_PORT`. All removed; the surviving command is one that was executed against a cluster.
- Not fixed, deliberately: the harness still gives no hint at the point of failure. A hint on the real failure path would have to distinguish a sandboxed agent from an agent that is merely down, which means an HTTP probe or a `kubectl get pod` on a path that fires for unrelated reasons. That is a behaviour change a documentation fix should not smuggle in.
- Not fixed, with reason: a pass noted that `kubectl exec` counts as a mutation under `docs/designs/live-test-lease.md` and the section does not mention taking a lease. The lease governs the shared install; this section is addressed to a contributor's own dev project, which `eval_driven_development.md` requires every contributor to have.

## Risk & Rollout

Low. No runtime code paths touched: five Markdown and comment edits, one Python module docstring, one test docstring. Nothing to roll out; revert is the commit.

Generated with the help of Gemini.


.agents/rules/eval_driven_development.md
```diff
@@ -22,8 +22,10 @@ INSTALL.md "Method 3" is the local-iteration path). `hack/ci-deploy.sh` itself i
 secrets. For cases that read the seeded fleet, whether through `fixtures:` or by naming
 `seeded-a`/`-b`/`-c` directly, the fleet must be applied to the dev project once
 ([`bench/tf/fleet/README.md`](../../bench/tf/fleet/README.md)). Every contributor, human or
-agent, is expected to have one. There is no path around the loop: a pull request that changes
-agent behaviour without eval evidence is not ready for review.
+agent, is expected to have one. A stock install sandboxes the agent, which the harness's
```
jayantid: Reviewed against product requirements, UX flow and technical design at `8bb3e28`; nothing to flag.

jayantid (approved):

PR #1713, opened 2026-09-17
chore: resolve four CodeQL false positives in tests and pool_pressure

bradhoekstra:
## Summary

Closes the four open CodeQL code-scanning alerts on `main` (#33, #34, #35, #36) without
changing what any of the flagged code does or proves. Each alert is a heuristic reacting to
the shape of a test fixture or a regex literal rather than to a defect: three fixture
constants whose names contain `SECRET`, two `"host" in line` checks in a source-text scan,
and a character class whose astral escapes the CodeQL extractor collapses to U+FFFD. The
change renames the constants, swaps the substring checks for a compiled regex, and writes
the character class as a raw string so the regex engine reads the escapes; a new test
sweeps every Unicode code point through the pattern against XML 1.0's Char production so
the rewrite is provably equivalent. The same three queries, run locally with the CodeQL CLI
against the before and after copies of the files, reproduce the four alerts before and
report none after.

## Why This Change

Four open alerts sit on the default branch's code-scanning tab. Left open they are noise
that makes a real finding harder to notice, and dismissing them as false positives hides
them from the next reader without recording why. Each has a small source change that
removes the trigger and leaves the behaviour intact, so the alerts close on their own once
this merges.

## What Changed

- `tests/test_workflow_pins_and_fork_guards.py` (alert #33, `py/clear-text-storage-sensitive-data`,
  line 628). The SARIF code flows show CodeQL's taint sources are the _names_
  `DISPATCH_ONLY_WITH_SECRETS_WORKFLOW`, `WORKFLOW_RUN_WITH_SECRET_GUARDED` and
  `PUSH_WITH_SECRET_ONLY_IN_COMMENT_WORKFLOW` (its `maybeSecret` heuristic matches
  `secret|trusted|confidential`), not the `${{ secrets.DEPLOY_KEY }}` text inside them.
  They are now `DISPATCH_ONLY_CREDENTIALED_WORKFLOW`, `WORKFLOW_RUN_CREDENTIALED_GUARDED` and
  `PUSH_WITH_CREDENTIAL_REF_ONLY_IN_COMMENT_WORKFLOW`, using the term the file's own
  definition of "credentialed" is built on. The fixture bodies, `_write_workflows`, the
  `_SECRET_REF_RE` matcher and every assertion are untouched. A comment above the first
  renamed constant records the reason, including why `PUSH_WITH_SECRETS_FILENAME_IN_RUN_WORKFLOW`
  keeps its name (`file` is on CodeQL's not-sensitive list).
- `agents/platform/skills/version-control/scripts/test_vcs.py` (alerts #34 and #35,
  `py/incomplete-url-substring-sanitization`, line 829). `test_no_forge_host_appears_outside_an_example`
  scans source lines for forge hostnames; it is not sanitising a URL. The two
  `"github.com" in line or "gitlab.com" in line` checks become
  `FORGE_HOST_RE.search(line)` with `FORGE_HOST_RE = re.compile(r"github\.com|gitlab\.com")`
  declared with the module's other constants. The query matches only `Compare`/`In`,
  `startswith` and `endswith` on a URL-looking literal, so a regex cannot re-trigger it.
  The change to the test body is one line, to limit the overlap with #1652 (see Context).
- `scripts/pool_pressure.py` (alert #36, `py/overly-large-range`, line 345).
  `XML_INVALID_CHARS` is now a raw string. Python's `re` interprets `\x`, `\u` and `\U`
  escapes in a pattern itself, so the compiled regex is the same object; the parse tree is
  identical (`[(IN, [(NEGATE, None), (LITERAL, 9), (LITERAL, 10), (LITERAL, 13), (RANGE, (32, 55295)), (RANGE, (57344, 65533)), (RANGE, (65536, 1114111))])]`
  for both). What changes is that CodeQL's regex parser now sees the escapes instead of
  the extractor's U+FFFD stand-ins for the two astral characters.
- `scripts/test_pool_pressure.py`. New `XmlText` class beside the existing `JunitOutput`
  tests: one test builds a string of every code point `0..sys.maxunicode` and asserts
  `_xml_text` keeps exactly XML 1.0's Char production (a wrong boundary digit or a dropped
  astral range fails it, checked by mutation); one names the boundaries (U+0001, U+0008,
  U+000B, U+001F, U+D7FF, U+D800, U+DFFF, U+E000, U+FFFD, U+FFFE, U+FFFF, U+10000, U+1F600,
  U+10FFFF) so a failure is readable without decoding the sweep; one covers `None` and `""`.
  `scripts/test_*.py` is already in `PYTHON_TEST_DIRS`. The test deliberately does not
  embed the old non-raw literal for a side-by-side comparison: that literal is what alert
  #36 fires on, and copying it into the test file would reopen the alert there. The
  old-versus-new comparison was run once by hand and is quoted under Testing.

## Context

- Code-scanning alerts on `gke-labs/kube-agents`: #33 (`py/clear-text-storage-sensitive-data`,
  `tests/test_workflow_pins_and_fork_guards.py:628`), #34 and #35
  (`py/incomplete-url-substring-sanitization`, `agents/platform/skills/version-control/scripts/test_vcs.py:829`),
  #36 (`py/overly-large-range`, `scripts/pool_pressure.py:345`). All four were open at
  the `main` analysis of `17da5f2f`.
- Open PR #1652 (dshnayder, `feat/vcs-consumer-migration`) rewrites the body computation
  in `test_no_forge_host_appears_outside_an_example` and adds a `client_source` attribute
  two lines above the line this branch changes, but leaves the `if "github.com" in line`
  line as unchanged context, so it does not remove the test or fix the alert. This branch
  changes only that one line inside the test plus two additions far from #1652's hunks
  (`import re`, the module constant). Adjacent hunks, a merge-conflict warning at most;
  whichever merges second, the three-way merge carries `FORGE_HOST_RE.search(line)`.
- No open issue tracks these alerts; no other open PR touches the four files.

## Testing

- `python3 -m unittest discover -s tests -p 'test_workflow_pins_and_fork_guards.py'` —
  `Ran 28 tests ... OK`. This includes `RealTreeTests`, the pins and fork-guard checker
  run against the real `.github/workflows`, so the checker still passes on the tree.
- `cd agents/platform/skills/version-control/scripts && python3 -m unittest test_vcs` —
  `Ran 64 tests ... OK`.
- `cd scripts && python3 -m unittest test_pool_pressure` — `Ran 118 tests ... OK`
  (115 before this change plus the three new `XmlText` tests; the full-range sweep takes
  about 0.3 s).
- `uvx ruff check --isolated --select E9,F63,F7,F82` over the four changed files (the set
  `make lint-python` runs; ruff is not installed in the local interpreter) — `All checks passed!`.
- `make docs-check` — all six checks pass (no docs changed).
- `make test-python` — the three directories this branch touches (`tests/`, `scripts/`,
  `agents/platform/skills/version-control/scripts/`) run their changed files green; the
  sweep as a whole fails locally on environment alone: `tests/test_package_release_bundle.py`
  needs the `zip` CLI, `tests/test_values_schema.py` expects the JSON-pointer error format
  of a newer helm than the local 3.16.3, and `tests/integration/`, `agents/platform/scripts/`,
  `admin_console/tests/` need `uvicorn`, `fastapi`, `httpx`, `mcp` and friends
  (`Missing third-party imports ... run: make test-python-deps`). None of those files is
  in this diff.
- Old-versus-new regex comparison for #36 (one-off script, not committed for the reason
  given under What Changed): pattern source text differs, `re._parser.parse` trees are
  identical, and `.sub("", s)` output is identical for every one of the 1,114,112 code
  points and for each named probe:

  ```text
  U+0001   old=dropped  new=dropped   U+D800 (lone surrogate)  old=dropped  new=dropped
  U+0009   old=kept     new=kept      U+E000                   old=kept     new=kept
  U+001F   old=dropped  new=dropped   U+FFFD                   old=kept     new=kept
  U+0020   old=kept     new=kept      U+FFFE                   old=dropped  new=dropped
  U+D7FF   old=kept     new=kept      U+FFFF                   old=dropped  new=dropped
  U+10000  old=kept     new=kept      U+1F600                  old=kept     new=kept
  U+10FFFF old=kept     new=kept
  all 1114112 code points: identical output = True
  ```

- CodeQL before/after, CodeQL CLI 2.27.0 with `codeql/python-queries` 1.8.10 (the bundle
  GitHub's action ships): databases built from the pre-change (`upstream/main`) and
  post-change copies of the three flagged files, analysed with exactly
  `Security/CWE-312/CleartextStorage.ql`, `Security/CWE-020/IncompleteUrlSubstringSanitization.ql`
  and `Security/CWE-020/OverlyLargeRange.ql`:

  ```text
  == before: 4 result(s)
    py/incomplete-url-substring-sanitization  vcs/test_vcs.py:829  'The string [github.com](1) may be at an arbitrary position ...'
    py/incomplete-url-substring-sanitization  vcs/test_vcs.py:829  'The string [gitlab.com](1) may be at an arbitrary position ...'
    py/overly-large-range                     scripts/pool_pressure.py:345  'Suspicious character range that overlaps with �-� ...'
    py/clear-text-storage-sensitive-data      tests/test_workflow_pins_and_fork_guards.py:628  'This expression stores [sensitive data (secret)](1) as clear text ...'
  == after: 0 result(s)
  ```

  The hosted default-setup analysis may run a different pack version; since this one
  reproduced `main`'s four alerts line for line, the residual risk is small, and the
  code-scanning tab on the merged head is the final check.

### Live validation

Not live-tested. The diff is three test files and the spelling of one regex literal in
`scripts/pool_pressure.py`, a CI-side pool-pressure report. Nothing here is built into an
image, rendered by the chart, reconciled by the operator, or read by an agent at runtime,
so there is no installation layer that could observe it; exempt from the eval loop under
`.agents/rules/eval_driven_development.md` for the same reason. The behaviour claim the
change does make (the compiled regex is unchanged) is proven by the parse-tree and
full-code-point comparisons above rather than by a deployment. The shared cluster was not
touched.

## Self-Review

Passes run: `review-adversarial` and `review-docs-drift`, each in its own subagent handed
the repository, the range `refs/kube-agents-base/main...HEAD`, its skill file and the green
`make docs-check` output, and told to derive intent from the diff, not to read commit
message bodies or scratch notes, and to edit nothing. Neither pass ran in the context that
wrote the change. After the one fix below, the amended comment (two added sentences in a
code comment) was not put in front of a fresh context again; both passes' other checks
are unaffected by it.

Adversarial pass, angles A–J worked (line-by-line and error paths, removed behaviour,
consumers of renamed symbols, ops/security surface, reuse/simplification/altitude, rules
and comment accuracy, scope and red/green, sibling PRs). It independently rebuilt CodeQL
databases and confirmed the four alerts before and zero after, ran `_xml_text` over all
code points against both versions, mutation-tested the new sweep (a wrong boundary digit
or a dropped astral range fails it), and checked the new names against every CodeQL
sensitive-name regex. Docs-drift pass: grepped old and new identifiers across the tree,
read the map rows and docs for each changed component (`docs/ci-pool-projects.md`
pool-pressure section, `scripts/testdata/pool_pressure/README.md`,
`docs/designs/version-control-support.md` boundary section), checked generated-region
sources and the map staleness window.

- LOW, CONFIRMED (adversarial): the new comment in `test_workflow_pins_and_fork_guards.py`
  said CodeQL flags any `*SECRET*` name, while `PUSH_WITH_SECRETS_FILENAME_IN_RUN_WORKFLOW`
  keeps that name 65 lines down and raises nothing, because CodeQL's not-sensitive list
  exempts names containing `file`, `path` or `url`. **Fixed**: the comment now names the
  exemption, so the next person renaming that fixture knows what would reopen the alert.
- Docs-drift: no Blocking, no Advisory findings. No document names the renamed constants,
  `XML_INVALID_CHARS`, `_xml_text` or the test; no generated region reads these files.
- Not covered by either pass: the hosted CodeQL run on the merged head (see Testing); the
  full `make test-python` sweep beyond the three touched suites (environment failures
  listed under Testing); any live install, per Live validation.

## Risk & Rollout

Low risk, no runtime code paths touched. Three files are tests; the fourth change is the
`r` prefix on a regex literal whose compiled form is unchanged. Revert is a plain revert.
One thing to expect at merge time: a trivial adjacent-hunk merge with #1652 in
`test_vcs.py`, in whichever of the two lands second.


agents/platform/skills/version-control/scripts/test_vcs.py
```diff
@@ -25,6 +25,7 @@
 import io
 import json
 import os
+import re
```
jayantid, line 28: Reviewed against product requirements, UX flow and technical design at `646c549`; nothing to flag.

jayantid (approved):

PR #1706, opened 2026-09-17
Add mrusmanabdullah to OWNERS list

mrusmanabdullah:
## What Changed

-

## Testing

-

### Live validation

-

## Self-Review

-

PR #1733, opened 2026-09-17
fix(ci): count only an OWNERS approver's verdict as a review, and never swallow /request-review silently

### LINKED ISSUE #1715: request_reviewers.py: an approval from a non-OWNERS reviewer permanently suppresses auto-assign
`scripts/request_reviewers.py` stops requesting a reviewer once *any* human files an `APPROVED` or `CHANGES_REQUESTED` review, whether or not that person can produce the `approved` label. A PR approved by someone outside the applicable OWNERS approvers list is then permanently invisible to the auto-assigner: it looks handed off, it cannot merge, and nothing will request anyone again.

Three open PRs are in that state right now — **#1653, #1672, #1545** — each with one `APPROVED` from a non-approver, no requested reviewer, and no `approved` label. #1653 is the expensive one: it is `lgtm`, MERGEABLE, and has been sitting because Tide wants `approved` and no human was ever asked for it.

### Where

`skip_reason()`, [`scripts/request_reviewers.py`](https://github.com/gke-labs/kube-agents/blob/main/scripts/request_reviewers.py) L264–L280:

```python
    author = (pull_request.get("user") or {}).get("login")
    humans = _dedupe(
        [
            review["user"]["login"]
            for review in reviews
            if (review.get("user") or {}).get("type") != "Bot"
            and review.get("state") in HUMAN_VERDICT_STATES
        ],
        exclude=author,
    )
    if humans:
        return f"{', '.join(humans)} already reviewed it"
```

`HUMAN_VERDICT_STATES` is `{"APPROVED", "CHANGES_REQUESTED"}` (L54).

The comment above this block is right about what it is defending against — replying to a review thread files a `COMMENTED` review under the replier's name, and AGENTS.md tells authors to answer every finding before running `/review`, so counting those would starve exactly the PRs that follow the process. Restricting the check to a real verdict is correct. The gap is that it treats *any* person's verdict as equivalent to a reviewer having been assigned, when only an OWNERS approver's verdict can move the PR.

### One detail worth correcting up front

This is not the existing `type != "Bot"` filter failing. The account involved on all three PRs, `kyber775`, is `"type": "User"` in the reviews API, so the filter behaves as written and the review counts as human:

```
$ gh api users/kyber775 --jq '{login,type}'
{"login":"kyber775","type":"User"}
```

The predicate that is missing is OWNERS membership, not bot-ness. Bot reviews are already excluded correctly.

### Suggested shape

Count a verdict only from someone in the OWNERS approvers list that applies to the PR's files; treat everyone else's review the same as a `COMMENTED` one for assignment purposes. The root list is small and `OWNERS` is already in the tree, though per-directory `OWNERS` files mean the applicable set is not always the root one.

Two cases to keep working, both currently correct and easy to break:

- **A non-approver's `CHANGES_REQUESTED` should probably still suppress.** Requesting a fresh reviewer while someone is waiting on changes is noise, even if that someone cannot approve. Worth deciding deliberately rather than inheriting from whichever way the predicate is written.
- **Re-running must stay idempotent.** The workflow fires on every completed `AI Review` check, so the common path is a PR that already has a reviewer — the `requested_reviewers` check at L259 covers that and should keep running first.

### How to tell it is fixed

On a PR with exactly one `APPROVED` from a non-OWNERS account and no requested reviewer, the workflow requests a reviewer instead of returning `already reviewed it`. #1653, #1672 and #1545 are live fixtures for the red case.

### Not verified

Whether per-directory `OWNERS` files should narrow the list or whether the root list is sufficient in practice — that is a design call for whoever takes this.

Separately and probably unrelated: #1664 carries `APPROVED` from `mplakhtiy`, who *is* in root `OWNERS`, and has neither `lgtm` nor `approved`. That does not fit this pattern and I have not looked into it, but it is the PR currently gating `validate` for the whole repo, so it may be worth a glance while you are in here.

bnaylor:
## Summary

`scripts/request_reviewers.py` stopped requesting a human reviewer once any non-bot user had filed an `APPROVED` or `CHANGES_REQUESTED` review, whether or not that person could produce the `approved` label. One approval from a colleague outside `OWNERS` therefore suppressed the auto-assign for good, and the documented remedy, `/request-review`, ran the same check first and exited green having done nothing. This change makes the verdict check ask whether the approver is an `OWNERS` approver for the changed files, lets `/request-review` bypass that check, and makes a declined override visible: a 😕 reaction on the comment, a `::warning::` annotation on the run, and a line in the step summary.

## Why This Change

Three open pull requests (#1653, #1672, #1545) sat `lgtm` and `MERGEABLE` with one `APPROVED` from `kyber775`, no requested reviewer, and no `approved` label. `kyber775` is `type: User`, so the existing bot filter was correct and irrelevant: the missing predicate was `OWNERS` membership. Because the auto-assign is the only thing that requests a reviewer here (opening a pull request no longer pings anyone), a pull request in that state cannot merge and nothing ever asks anyone again.

The override made it worse rather than better. `/request-review` on #1672 (comment 5719654130, [run 35261933934](https://github.com/gke-labs/kube-agents/actions/runs/35261933934)) logged `Not requesting a reviewer: kyber775 already reviewed it`, added no reaction, and exited 0, so the person who typed it could not tell it from the workflow never firing.

## What Changed

- `skip_reason()` keeps the skips that hold however the run started (closed, draft, ignored keyword, someone already requested); the verdict check moves to `already_reviewed_reason(pull_request, reviews, approvers)`. It reduces the reviews list to each person's latest `APPROVED`/`CHANGES_REQUESTED` (by `submitted_at`; `DISMISSED` and `COMMENTED` never count), then an `APPROVED` counts only when the reviewer is in `approvers` and a `CHANGES_REQUESTED` counts from any human other than the author. Bot reviews and the author's own never count, as before. Logins compare case-insensitively.
- `applicable_approvers(changed_files, root)` computes `approvers` from the checkout the workflow runs in (the default branch, which is also where Prow reads them) with Prow's `entriesForFile` walk: for each changed file, from its directory up to the root, `approvers:` lists or matching `filters:` regexes (a file with filters ignores a top-level list beside them, as Prow does), stopping at a level that sets `no_parent_owners` once the file has collected any approver, with `OWNERS_ALIASES` expanded. On today's tree that is the eight root approvers for almost everything and `eval-crew` alone for the two presubmit roster files, matching the comment in `hack/OWNERS` and the reviewer groups in `.github/auto_request_review.yml`.
- `--react-to` (the `/request-review` path) skips `already_reviewed_reason()`; a person has read the pull request and asked. It still runs `skip_reason()`, so re-asking when someone is already requested stays declined. Every decline on that path now goes through `decline()`: `::warning::Not requesting a reviewer: <reason>` to stdout (an annotation on the run), the same line appended to `$GITHUB_STEP_SUMMARY` when set, and a `confused` (😕) reaction on the comment. `eyes` (👀) remains the acknowledgement for a request that was made. The check-run path logs as before and never reacts.
- `--dry-run` covers the reactions too (`would react confused to comment N`), so the override path can be exercised without writing to a live pull request. `--owners-root` (default `.`) exists for the tests and for running the script from outside a checkout root.
- The changed-files fetch moves ahead of the verdict check (it is needed for the approver set); the reviews fetch moves after it and is skipped entirely on the override path.
- `docs/pull-request-workflow.md` carries the rule (it is the page that already described `skip_reason()`); `agents/contributor/AGENTS.md` gets one clause on the reactions; the workflow's permission comment names both reactions; `docs/README.md` §2 gains a source row. `AGENTS.md` is unchanged: its `/request-review` paragraph stays true, and the file has 84 chars of context budget left.

## Context

Closes #1715. Live fixtures: #1653, #1672, #1545. #1664, which the issue mentions in passing, merged at 13:52Z today with `approved`/`lgtm`, so it is not a case of this bug.

No open pull request touches `scripts/request_reviewers.py`, `scripts/test_request_reviewers.py` or `.github/workflows/auto_request_review.yml` (checked by search term and by changed file at 20:30Z).

**Design choice, OWNERS scope.** The issue leaves open whether the root list is enough or per-directory `OWNERS` should narrow it. This change implements Prow's walk rather than root-only, because the repository already has one per-directory file (`hack/OWNERS`) whose whole point is that a root approver's approval does not clear the presubmit rosters, and root-only would reproduce this exact bug for a roster-only pull request approved by a root approver. What it does not implement: `reviewers:` and `labels:` (only approvers can produce `approved`, so they do not matter here), `emeritus_approvers`, and `required_reviewers`; unknown keys are ignored rather than refused, since an OWNERS file gaining a key Prow understands should not stop the auto-assign. The set is the union across changed files, so on a mixed pull request an approval from an approver of any part counts; that is the same tolerance the reviewer selection already has.

**Design choice, `CHANGES_REQUESTED` from a non-approver.** Still suppresses, deliberately. Whoever filed it is owed a reply, and requesting a fresh reviewer over an open objection is noise. Pinned by `test_a_non_approvers_changes_requested_still_counts` and stated in the docstring.

**Exit code on a declined override.** Kept at 0. An `issue_comment` run is not attached to the pull request's head, so a red run would show only in the Actions tab and mail the commenter as a failure; the reaction and the annotation are the signal the person who typed the command can see. Changing this later is one line in `decline()`.

**Handoff for the PR minder.** Verified: the unit tests (82, all green on the branch; 43 of them error on `upstream/main` because the functions they test do not exist there), the CI-style discovery over `scripts/` (1777 tests, OK), `make lint-python`, `make docs-check`, prettier on the four changed `.md`/`.yml` files, and read-only `--dry-run` runs of both versions against the three fixtures with a real token (transcript below). Assumed: that Actions renders a `::warning::` written by a Python step as an annotation, which it does for any step's stdout, and that a `confused` reaction is accepted on an issue comment, which the reactions API documents. Not live-tested: the workflow itself, because the `issue_comment` and `check_run` triggers only fire from the default branch, so nothing here runs until it merges. After merge: #1672 and #1653 already have hand-requested reviewers (jayantid at 19:32Z, dshnayder at 19:33Z), so the workflow will decline both with `review is already requested from ...` as before; #1545 still has nobody requested, and the next completed `AI Review` check on it, or a `/request-review` comment, should request a reviewer. Watch that one.

## Testing

- `cd scripts && python3 -m unittest test_request_reviewers`: 82 tests, OK. New classes `OwnersTest`, `AlreadyReviewedTest`, `MainTest`; `SkipReasonTest` updated for the split. Coverage: non-approver `APPROVED` no longer suppresses; approver `APPROVED` still does; non-approver `CHANGES_REQUESTED` still does; a withdrawn objection (`CHANGES_REQUESTED` then `APPROVED` from the same non-approver) no longer does; an objection after an approval does; a later `COMMENTED` does not clear one; ordering by `submitted_at`; `DISMISSED` ignored; author and bot reviews excluded as before; case-insensitive login match; the #1545 payload shape; Prow's walk against a temp tree mirroring `OWNERS`/`OWNERS_ALIASES`/`hack/OWNERS` (including a nested file under `no_parent_owners` and a top-level list beside filters) and against the live tree; `--react-to` bypasses the verdict skip and reacts 👀; `--react-to` does not bypass the already-requested skip and reacts 😕 with the `::warning::` line on stdout and the reason in the step summary; the check-run path never reacts; `--dry-run` posts nothing on either outcome.
- The same test file against `upstream/main`'s `scripts/request_reviewers.py` (01be641b): `Ran 82 tests, FAILED (errors=43)`; every new test errors because `already_reviewed_reason`, `applicable_approvers`, `--owners-root` and the new `skip_reason` signature do not exist there. The behavioural red is the offline replay below, which runs `main`'s own `skip_reason` over the real payloads.
- The CI invocation for the directory (`PYTHONPATH` as `make test-python` sets it; `cd scripts && python3 -m unittest discover -p "test_*.py"`): `Ran 1777 tests, OK (skipped=1)`.
- `make lint-python`: `All checks passed!`. `make docs-check`: green, `AGENTS.md` 0.1k under the 38,000-char budget. `prettier --check` (3.9.6) on the four changed `.md`/`.yml` files: clean.
- Offline replay of the captured review payloads of #1672, #1653, #1545 (captured 2026-09-17T20:33Z, `requested_reviewers` cleared to the state at filing time): `upstream/main` says `SKIP: kyber775 already reviewed it` for all three; this branch says `would request a reviewer` for all three.

### Live validation

Not live-tested against the workflow: both of its triggers run the script from the default branch, so this code cannot execute in Actions until it merges. What was run instead, read-only with a real token (`--dry-run` GETs and then logs what it would POST; no reviewer was requested and no reaction was added on any live pull request):

```
== upstream/main (01be641b) --pr 1545 --dry-run --seed 0
#1545 by bnaylor: style(operator): gofmt platformagent_litellm_policy_test.go
Not requesting a reviewer: kyber775 already reviewed it
== branch --pr 1545 --dry-run --seed 0
#1545 by bnaylor: style(operator): gofmt platformagent_litellm_policy_test.go
OWNERS approvers for the changed files: antontyb, bradhoekstra, dshnayder, fatoshoti, haoxuw, jayantid, mplakhtiy, toshiowang
Requesting review from dshnayder
--dry-run: would POST reviewers=['dshnayder'] team_reviewers=[] to #1545

== branch: the override path on #1672 with the real /request-review comment 5719654130, --dry-run
#1672 by bnaylor: fix(operator): scope the gateway's JetStream API grant to the streams and verbs it uses
Not requesting a reviewer: review is already requested from jayantid
::warning::Not requesting a reviewer: review is already requested from jayantid
--dry-run: would react confused to comment 5719654130
== upstream/main: same invocation
Not requesting a reviewer: review is already requested from jayantid
```

#1672 and #1653 no longer reproduce the issue's red case live because reviewers were hand-requested on them at 19:32Z/19:33Z, after the issue was filed; both versions decline them on `review is already requested from ...`, which is the idempotency skip the issue asks to keep first. Their filing-time state is covered by the offline replay above. What this cannot show: the `confused` reaction rendering on a comment and the `::warning::` line becoming an annotation; both are documented GitHub behaviour and the tests pin the calls and the output.

## Self-Review

Both passes ran in fresh subagents handed only the worktree path, the range `upstream/main...HEAD`, and the skill file; neither saw the reasoning behind the diff.

**Adversarial pass** (`review-adversarial`). Looked at: the verdict predicate against GitHub's review-list semantics (historic reviews, `DISMISSED`, `COMMENTED` after a verdict), the OWNERS walk against Prow's `entriesForFile`, workflow-command injection through `::warning::`, path traversal through PR filenames in the walk, the permission the 😕 needs, the premise that an approver's Approve produces `approved`, whether the live-tree test reaches the live config, `@`-prefixed logins, and sibling open PRs (#1706 adds a login to root `OWNERS`; unrelated to this change but now an input to it).

- Medium, fixed: `already_reviewed_reason()` read every historic review, so a non-approver who requested changes and later approved kept suppressing the auto-assign (the old code had the same mechanics; this change re-decides the rule, so it owns the consequence). Now each person's latest `APPROVED`/`CHANGES_REQUESTED` is the one that counts, ordered by `submitted_at`; pinned by `test_a_withdrawn_objection_no_longer_counts`, `test_an_objection_after_an_approval_counts`, `test_a_later_comment_does_not_clear_an_objection`, `test_verdicts_are_ordered_by_submission_time_not_list_order`, `test_a_dismissed_review_does_not_count`.
- Medium (latent), fixed: the `no_parent_owners` stop broke on "matched at this level"; Prow breaks on "any approver collected so far for this file". Nothing in today's tree distinguishes them, but a future `hack/eval/OWNERS` would have let a root approver's approval count where Prow says it cannot. Now per-file accumulation with Prow's condition; pinned by `test_no_parent_owners_stops_on_approvers_collected_below_it_too`. While there, matched one more Prow rule the pass noted as not-a-finding: a file with `filters:` ignores a top-level `approvers:` beside them (`test_top_level_approvers_beside_filters_are_ignored_as_prow_does`).
- Low, fixed: the reaction behaviour had three prose homes. Dropped the `AGENTS.md` sentence (its paragraph stays true without it, and it spent 58 of the 84 chars under the context budget); `docs/pull-request-workflow.md` is the canonical statement and `agents/contributor/AGENTS.md` keeps one clause because a bot reading 😕 needs to know what it means.
- Not fixed, by argument: `@`-prefixed logins are not stripped as Prow's `NormLogin` does; none exist in the tree, and the live-tree test would catch one landing in `OWNERS` only if it broke agreement with the config, so this is noted rather than handled.

**Docs-drift pass** (`review-docs-drift`). Looked at: the three doc hunks against the script, the workflow comment against the constants, `docs/README.md` map rows and §2 sources, `docs/site/` for mentions, `.agents/rules/`, `CONTRIBUTING.md`, `.claude/commands/`, generated regions, `make docs-check`. No Blocking findings.

- Should fix, fixed: "skips the verdict check only" read as a statement about the whole decision, and `/request-review` also bypasses the `AI Review` gate (the workflow omits `--require-ai-review-pass`). Reworded to "Of these reasons, ... alone (it also bypasses the `AI Review` gate, per `AGENTS.md`)".
- Should fix, fixed: the `AGENTS.md` sentence left 26 chars of context budget. Dropped (see above); headroom is back to `main`'s 84.
- Nit, fixed: "anyone other than the author" → "a human other than the author"; the code excludes bots.
- Nit, fixed: the contributor doc's parenthetical list of decline reasons read as exhaustive; removed.
- Nit, fixed: added one `docs/README.md` §2 row naming the script and the workflow as the source for the auto-assign rules, beside the sibling rows for the thread hold and flaky-check tracking.
- Pre-existing, not this PR's: `docs/designs/design_537148738.md:100` quotes a workflow comment that no longer exists and says the workflow runs on `pull_request_target`; it has been `check_run` + `issue_comment` since the script replaced the action. Reported to the tech lead for a separate fix.

## Risk & Rollout

This changes reviewer auto-assign repo-wide. After it lands, an approval from outside `OWNERS` no longer stops the workflow from requesting a reviewer, so pull requests currently in that state get one at their next completed `AI Review` check or `/request-review`; today that is #1545 (#1672 and #1653 already have hand-requested reviewers and stay declined on that skip). A pull request with a `CHANGES_REQUESTED` from anyone stays declined as before. The override path now writes one reaction and one annotation per declined `/request-review`; both use permissions the job already has (`issues: write`). New failure mode: an `OWNERS` file that fails to parse as YAML raises and fails the run, where before OWNERS was never read; the two files in the tree parse today and `OwnersTest.test_the_live_tree_agrees_with_the_bots_config` fails if either stops agreeing with the bot's config. Revert is a single-commit revert of the script and its tests; nothing has to happen at merge time, and the workflow file itself changes only in a comment.


bnaylor:
I have particular standing on the `/request-review` half of this: proving that the old
`skip_reason()` defeated the override is what I filed on #1715, so I ran the new code against
live data rather than reading it.

Driving `skip_reason()`, `applicable_approvers()` and `already_reviewed_reason()` directly at
`8dff41e5`, with each PR's real files and reviews pulled from the API:

| PR | `skip_reason` | `already_reviewed_reason` |
| --- | --- | --- |
| #1672 | None | **None** |
| #1653 | already requested from dshnayder | None |
| #1665 | already requested from bradhoekstra | None |
| #1614 | already requested from bradhoekstra | None |
| #1675 | already requested from dshnayder | None |
| #1248 | draft | None |

#1672 is the case.  It has kyber775's APPROVED, no `approved` label, and no reviewer — on `main`
today the run logs `Not requesting a reviewer: kyber775 already reviewed it` and exits 0 (run
`35261933934`).  Under this change kyber775 is not in the approver set, the verdict does not
count, and the request goes out.  Everything else in the list is held by the already-requested
gate, which is the right reason.

Two things I checked because they looked like the places this could be subtly wrong:

- **The `no_parent_owners` walk.**  I expected a divergence from Prow here and there is none.
  Prow's `entriesForFile` breaks on `o.options[d].NoParentOwners && out.Len() > 0`, not on the
  flag alone, which is exactly what `applicable_approvers` does — so `hack/ci-eval-pr.sh` falls
  through to root and `hack/eval/presubmit-cases.txt` gets `['jayantid', 'lapis2002']` from the
  alias.  Measured both against the live tree.  That corner has never actually been exercised:
  no merged PR in the last 60 touches only non-filtered files under `hack/`.
- **`scripts/test_request_reviewers.py`** — 82 tests, OK, run by name rather than through
  discovery.

One behaviour worth naming rather than fixing: a malformed OWNERS file now reds this job.
`_read_yaml` lets `yaml.safe_load` raise and raises `ValueError` itself on a non-mapping, nothing
catches either, and OWNERS was not read at all before.  A broken root OWNERS therefore takes
reviewer assignment out repo-wide until it is fixed.  That is probably the behaviour you want —
guessing an approver set from a file you could not parse is worse — but it is a new way for this
job to fail and it is not in the PR body.

Not a review verdict; I am minding the branch, not approving it.

toshiowang (approved):

Issue #1739, opened 2026-09-17
a2a: lib.TasksGet leaks an ordered consumer on TASKS for five minutes per call, and the reserved-consumer budget has no term for it


bnaylor:
Found in #1665's review (thanks to the finding relayed out of jayantid's review of #1672). The
original report was that `a2aTasksReservedConsumers` has no term for `tasks/get` replay
ephemerals. That is true, but measuring it turned up something larger: `lib.TasksGet` never
cleans up the consumer it creates, and the repo's own written belief about why that is safe is
wrong.

## What the budget says

`a2aTasksReservedConsumers = 16` (`platformagent_a2a_manifests.go:291`) is itemized in the comment
above it, and the itemization is exact:

| slots | what |
| --- | --- |
| 2 | standing durables — the gateway's `gateway-relay` and the Hermes bridge's `bridge-<profile>` |
| 1 | headroom for the audit durable the accountability rail needs |
| 3 | one session's worth of overlap while the gateway retires an incarnation and mints its replacement |
| 10 | the web rail's concurrent readers |
| **16** | |

Nothing is left for a `tasks/get` replay. The budget is
`maxSessions*3 + 16` (`a2aTasksConsumerBudget`), and the stream is created at
`max(64, budget)` (`a2aTasksMaxConsumers`, floor at `a2aTasksMaxConsumersFloor = 64`), so the
floor hides the shortfall until `maxSessions >= 17` — at 17, `17*3 + 16 = 67`, the first value
above the floor, and the spare goes to zero.

## What `TasksGet` actually does

`lib.TasksGet` (`a2a/lib/fold.go:182`) opens an ordered consumer:

```go
cons, err := js.OrderedConsumer(ctx, TasksStream, jetstream.OrderedConsumerConfig{
	FilterSubjects: subjects,
	DeliverPolicy:  jetstream.DeliverAllPolicy,
})
```

`InactiveThreshold` is left zero, and nats.go v1.53.1 fills that with **five minutes** —
`jetstream/ordered.go:635` hardcodes `InactiveThreshold: 5 * time.Minute` into the config it
builds, and `:646-647` overrides it only when the caller set a non-zero value.

Cleanup is `defer it.Stop()`. `orderedSubscription.Stop()` (`jetstream/ordered.go`) stops the
local subscription and closes `done` — **it does not delete the consumer**. Nothing else in
`a2a/lib` calls `DeleteConsumer`.

[...]

PR #1741, opened 2026-09-17
fix(a2a): bound TasksGet's ordered consumer with a 5s inactive threshold

bnaylor:
> **Update 9/18 -- the design changed and this description is rewritten to match.**
> As opened, this PR set a five-second threshold *and* deleted the consumer by name at
> return. Review settled on dropping the explicit delete: the delete is a publish on a
> subject the rendered `bridge` grant withholds, and paying an Error-level permissions
> violation on every task the bridge dispatches to reclaim the last five seconds of one
> consumer slot is a bad trade. The threshold is now the whole cleanup. `deleteReplayConsumer`,
> `replayConsumerDeleteTimeout` and the delete's test are gone from the branch; the argument
> for *not* deleting, and for not widening the grant instead, is written into `fold.go`'s
> constant block. The old text of this description described the withdrawn design and is
> replaced rather than struck through -- there was too much of it to read around.

## Summary

`lib.TasksGet` left an ephemeral consumer on TASKS for five minutes after every call that found
a task. It opened
its replay's ordered consumer with `InactiveThreshold` unset, which nats.go v1.53.1 fills with
five minutes, and stopped the iterator on return, which never deletes the consumer. It now sets
the threshold explicitly to five seconds. Nothing deletes the consumer; the threshold is the
whole of the cleanup, and that is deliberate. Measured on an embedded nats-server under the
grant the operator renders for `bridge`: one `tasks/get` leaves 1 consumer carrying
`inactive_threshold=5s` and the server reaps it 5.1s after return; twenty sequential calls leave
20 and the stream is clear at 6.0s. On `upstream/main` the same consumers carry `5m0s`. The
worker adapter's comment that "nats.go's ordered consumers use five seconds" is corrected --
they use five minutes, and that belief is what let `TasksGet` ship with the default.

## Why This Change

The consumer count on TASKS was tracking the `tasks/get` call rate over a five-minute window
rather than the replays in flight, against a `max_consumers` sized for the latter (64 on a
default install; #1665's budget itemizes 16 reserved slots and none of them is a replay). At any
real rate five minutes of accumulation dwarfs that, and the failure is `maximum consumers
exceeded` on a read API: `tasks/get` itself, the gateway's stale-task heal and transcript primer,
and the bridge's dispatch check all go through `TasksGet`. #1739 found it while measuring
#1665's budget, and found that the one comment in the tree explaining why a short threshold is
safe rested on a nats.go default sixty times larger than it says.

Nothing else cleans these up. `orderedSubscription.Stop()` stops the local subscription and
closes `done` (nats.go v1.53.1 `jetstream/ordered.go:364`); the only `DeleteConsumer` on the
ordered path is the reset's, fire-and-forget, for the consumer being replaced (`:571`). So the
threshold is the only lever, and the change is to move it from five minutes to five seconds --
a 60x cut in the term #1739 is trying to bound.

## What Changed

- `a2a/lib/fold.go`: `TasksGet` sets `InactiveThreshold: EphemeralConsumerInactiveThreshold` (5s,
  new exported constant) on the ordered consumer. That is the entire runtime change -- three
  fields in the config struct where there were two.
- The constant's comment block is the rest of the diff's weight, and it is load-bearing, because
  the obvious next edit to this code is the wrong one. Two arguments live there:
  - **Do not add a delete.** `$JS.API.CONSUMER.DELETE.TASKS.*` is withheld from `bridge`, and a
    refused publish on a JetStream *request* subject gets no reply at all: the caller blocks to
    its own deadline while the violation lands out of band on the connection's async error
    handler. Both halves are measured (see Testing) -- re-adding a synchronous delete takes a
    replay from 0.11s to 30.07s and writes one Error-level `Permissions Violation` per call,
    which is the line an operator is taught to read as a missing grant.
  - **Do not widen the grant either.** A wildcard is the only form on offer: nats.go names an
    ordered consumer `<prefix>_<serial>` and bumps the serial on every reset (v1.53.1,
    `jetstream/ordered.go:629`), so no exact-name grant can cover one. And `DELETE.TASKS.*` is a
    wildcard over consumer *names*, not over the consumers its holder created, so it would hand
    everything carrying the bridge password delete on any consumer on TASKS -- the gateway's own
    `gateway-relay` durable included. That is the same argument `sessionConsumer` already makes
    for `MSG.NEXT` in the worker adapter (`adapter.go:801`), and the stated reason the adapter
    uses named consumers rather than ordered ones.
- `a2a/worker-adapter/adapter.go`: `consumerInactiveThreshold` is now
  `lib.EphemeralConsumerInactiveThreshold`, so the module's ephemerals on TASKS reap on one
  clock, and the comment says what nats.go's ordered default actually is (five minutes,
  `jetstream/ordered.go:635`, replaced only by a caller's non-zero value at `:646`). The rest of
  that comment, about why five seconds is safe for the adapter, stands.
- `a2a/lib/replay_consumer_test.go` (new, 337 lines): an embedded nats-server with a reader
  carrying the TASKS half of `a2aBridgeJetStreamGrants` -- `CONSUMER.DELETE` deliberately not in
  the list -- plus an unrestricted admin that provisions the stream the way the provision script
  does (`allow_direct`, limits, discard old, 64 consumers) and reads the consumer list back.
  Three tests, below.
- Two pieces of prose this change made false, both found by the docs-drift pass:
  `a2aBridgeJetStreamGrants`' comment in `platformagent_a2a_manifests.go` said the ordered reset
  was the only emitter of `CONSUMER.DELETE.TASKS` and named a five-minute threshold; it now names
  the five-second one and says why `TasksGet` does not delete. `a2a/docs/hermes-bridge.md` gains
  a paragraph naming which `CONSUMER.DELETE` the bridge holds (KV, watcher) and which it does not
  (TASKS), and saying that a `Permissions Violation` in the bridge's log therefore still means
  what it says. No grant list changes.

## Context

- Part of #1739. This takes items 1 and 3 of its WANTED list, the two that exist on `main`.
  Items 2 and 4 (a replay term in `a2aTasksReservedConsumers` and re-itemizing that comment) are
  not here because those constants live on #1665's branch; they follow once #1665 lands. With the
  threshold at 5s the term needs headroom for a few seconds of replays rather than five minutes
  of them, which is a smaller number than the issue first feared.
- **The ask this PR used to make of #1672 is withdrawn.** As opened, this branch deleted the
  consumer by name, so it asked #1672 to add `$JS.API.CONSUMER.DELETE.TASKS.*` to the gateway's
  grant. There is no delete now, so #1672 needs nothing from this PR and this PR needs nothing
  from #1672. The two touch no common line.
- `main` merged in: this branch carries #1653 (`965b6917`), which split the static `worker` user
  into `bridge` and `agent`. The test file and the grant comment are written against
  `a2aBridgeJetStreamGrants` as it stands after that merge, not against the retired
  `a2aWorkerJetStreamGrants`.
- Every caller of `TasksGet` and what its rendered grant gives it, read from
  `platformagent_a2a_identities.go` and `platformagent_a2a_manifests.go` on `main`:

  | Caller | Principal | `CONSUMER.DELETE.TASKS` | Result |
  | --- | --- | --- | --- |
  | `a2a/gateway` (`gateway.go` x2, `reap.go`, `relay.go`, `spawn.go` x2) | `gateway` (static) | yes, inside `$JS.API.>` | 5s threshold; the grant is unused by this path |
  | `a2a/hermes-bridge` (`bridge.go` x2 -- the dispatch check on every unknown task and orphan cancel; `sweep.go` x2 -- startup) | `bridge` (static password principal, `a2aBridgeUser`) | **no** | 5s threshold, no refused publish, no log line |
  | `a2a/web` (web rail) | `web` | n/a | does not call `TasksGet` |
  | `a2a/worker-adapter` | callout `session` | n/a | does not call `TasksGet`; `priorEvents` uses its own named consumer with an explicit delete |
  | `a2a/lib` conformance and unit tests | unrestricted | yes | 5s threshold |

  Every row takes the same path now, which is the point of dropping the delete: the behaviour no
  longer depends on which principal made the call.
- #1665 (`a2a/bus-claims-enforced`, open) touches `a2a/lib/fold.go` too: a `SubmittedMissing`
  field on `Task`, a check in `FoldTask`, and a `Warn` at the end of `TasksGet`. This PR's hunks
  are the import and constant block at the top and the `OrderedConsumer` call in the middle; no
  lines overlap and git merges them either way. The new `time` import is the one line both add
  context around. #1665 also rewrites the same adapter comment block this branch edits, so
  whichever lands second carries a textual conflict there -- small, and the two texts now agree
  on the facts.
- Duplicate-checked before opening: `gh pr list --state open --search` for `TasksGet`,
  `OrderedConsumer`, `InactiveThreshold` and `1739` returns #1665, #1672 and #1614, none of which
  changes `TasksGet`'s consumer handling.
- Conformance (#1121): no `tests/conformance/` assertion owed; this changes no grant and moves no
  invariant boundary.

**Handoff for the PR minder.** Verified by execution on nats-server 2.14.6 (the module pin) and
nats.go 1.53.1: the red count on `upstream/main` (`5m0s` on the replay consumer), the green
counts here (1 at `5s` reaped at 5.1s; 20 at `5s`, clear at 6.0s), the absence of any refused
publish during a replay proved non-vacuous by a control refusal on the same connection, and the
two mutations in the Testing section. Verified by reading nats.go v1.53.1 and confirmed by the
tests: the five-minute default and its override (`ordered.go:635,646`), that `Stop` does not
delete (`:364`), and the ordered consumer's `<prefix>_<serial>` naming (`:629`). Assumed, not
verified: that the deployed `nats:2.10-alpine` reaps a 5s-threshold ephemeral on the same
schedule as 2.14.6 -- the worker adapter has relied on this same value in production since it
shipped, and 2.10.29 is what `a2a-next-dev` runs. Not live-tested against a running install; the
reason is in Live validation below.

## Testing

- `cd a2a && go vet ./... && go test -race ./... -count=1`: green, every package.
- `make conformance`: 122 tests, OK, 12 expected failures (the counts moved from 118/13 with
  #1653; identical on `origin/main`).
- The three new tests, and what each one measures:

  | Test | Measures | Result |
  | --- | --- | --- |
  | `TestTasksGet_ReplayConsumerCarriesTheInactiveThreshold` | the threshold the *server* holds, read back off the consumer rather than off the config struct | 1 consumer at `inactive_threshold=5s` the instant the call returns, 0 at 5.1s |
  | `TestTasksGet_EmitsNothingTheBridgeGrantRefuses` | that a replay under the bridge grant triggers no permissions violation | no refusal; 0.13s |
  | `TestTasksGet_BurstBoundsReplayConsumersByTheThreshold` | twenty sequential replays | 20 consumers, each at `5s`, all reaped by 6.0s |

- The refusal test is the one that could go vacuous, so it does not just assert an empty buffer.
  After the replay it publishes a known-forbidden `$JS.API.STREAM.DELETE.TASKS` down the same
  connection and waits for *that* refusal to land before asserting the replay's absence. Refusals
  arrive in the order the publishes left, so the control line is the barrier: anything the replay
  was refused is already in the buffer by the time it shows up. The control also fixes the
  expected count, so the assertion is `exactly one Permissions Violation, and it names the control
  subject` rather than `no CONSUMER.DELETE` -- `TasksGet`'s request path is wider than the ordered
  consumer (the horizon read ahead of it emits `STREAM.INFO` and a direct get) and the grant
  withholds more than the one verb.
- Mutation-proved, each measured against the branch as it stands:

  | Mutation | Result |
  | --- | --- |
  | drop `InactiveThreshold` from the `OrderedConsumerConfig` | FAIL -- `inactive_threshold on the server = 5m0s, want 5s`. This is the bug, reproduced. |
  | re-add a synchronous `DeleteConsumer` at return | FAIL at **30.07s** (0.11s unmutated -- nats.go's default JetStream API timeout), with `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"` on the async handler. This is the measurement behind the "do not add a delete" half of the comment. |
  | add a refused non-DELETE publish (`$JS.API.CONSUMER.NAMES.TASKS`) to the replay path | FAIL -- `2 "Permissions Violation" lines ... want exactly 1`. Against the narrower `no CONSUMER.DELETE` assertion this branch shipped first, the same mutation ran green in 0.465s, which is why the assertion was widened (`6a337ade`). |

- `make docs-check`: green (`a2a/docs/hermes-bridge.md` changed), prettier 3.9.6 clean on it.
  `go vet` on `k8s-operator/internal/controller` after the comment edit: clean. `gofmt -l` clean
  on the changed files.

### Live validation

Not live-tested against a running install, and here is exactly what was and was not done. The
only `spec.mode: next` install (`gke_bnaylor-kagents-dev_northamerica-northeast1_a2a-next-dev`)
carries another session's stacked build and is read-only for this change; nothing here rolled an
operator, patched a CR or published to its bus. What was observed there, read-only,
2026-09-17T22:39Z, through a port-forward to the NATS monitoring port and one
`GET /jsz?acc=APP&consumers=true&config=true`: nats-server 2.10.29, TASKS with 0 messages and
exactly one consumer, the `gateway-relay` durable. No `tasks/get` had run in the previous five
minutes, so there was no lingering ephemeral to photograph, and with no task on the stream a
`tasks/get` there would return `TaskNotFound` before creating one. The other dev cluster is a
default-mode install and renders no bus.

The mechanism proof is the embedded server: the provision script's stream flags, the bridge grant
read from `a2aBridgeJetStreamGrants`, and the before/after thresholds above. What that leaves
unwatched is a deployed `nats:2.10-alpine` reaping a 5s ephemeral -- and the adapter has depended
on that same value there since it shipped. No test artifacts were left anywhere; the port-forward
was closed.

## Self-Review

Two passes, each in a fresh `general-purpose` subagent handed only the worktree path, the range
`upstream/main...HEAD` and the skill file (`review-adversarial`, `review-docs-drift`), nothing
else; neither saw this description. Both re-derived every nats.go citation against the module
cache. Findings, with current dispositions -- several of these were raised against the
delete-by-name design and are resolved by its removal rather than by a fix:

1. **Resolved by the design change** (adversarial, HIGH, as opened). Under #1672's gateway grant
   the explicit delete would have been a refused publish on every gateway `tasks/get`, one Error
   line each, and inert for the principal making most calls. There is no delete now, so there is
   no refusal on any path and no ask of #1672.
2. **Fixed** (docs-drift Blocking x2; adversarial MEDIUM, the same passages). The grant function's
   comment and `a2a/docs/hermes-bridge.md` both described a `CONSUMER.DELETE.TASKS` story this
   change invalidates. Both rewritten; `make docs-check` and prettier clean.
3. **Confirmed, disposition is the author's** (adversarial, MEDIUM). #1665 rewrites the same
   adapter comment block. Recorded in Context; the conflict is textual and small.
4. **Resolved by the design change** (adversarial, MEDIUM, as opened: the "own deadline" claim on
   the delete goroutine was untested). The goroutine, its deadline and its test are gone.
5. **Fixed** (adversarial, LOW). The test's stream-provisioning literals (`a2a.tasks.>`, 72h, 64,
   the 5s admin timeout) are named at the top of the file, and the helper says why it is not
   `testutil`'s `provisionTasksStream` (that one leaves `allow_direct` off, which routes the
   horizon read through a subject no rendered grant carries).
6. **Fixed** (docs-drift Non-blocking). The `fold.go` comment and the test comments cited grant
   shapes that no longer exist after #1653. Both now name `bridge`.
7. **New, self-caught after the design change.** Dropping the delete leaves the *reason* for not
   having one nowhere in the code, where the next reader would re-add it. That is why the
   constant's comment block is as long as it is, and why the grant-widening argument is written
   out rather than left implicit.
8. **Fixed** (jayantid, review of 2026-09-21, 🟡). The doc and the grant comment in this branch
   contradicted each other. `hermes-bridge.md` said a `tasks/get` under `bridge` "emits no
   refused publish, so a `Permissions Violation` in the bridge's log still means what it says";
   the grant comment said withholding CONSUMER.DELETE on TASKS "costs a log line on a reset".
   The comment is the accurate half and needed no change. nats.go's ordered consumer deletes the
   consumer it replaces -- `ordered.go:558 reset()` fires `DeleteConsumer` and discards the
   result -- and that publish is refused, so an operator triaging a violation was being sent
   after a grant that is not missing. The doc now names
   `$JS.API.CONSUMER.DELETE.TASKS.<name>` as the one refusal the grant produces by design.
   Fixed in `b312415d`; `make docs-check` and prettier green. **This landed in a gap this
   section already declared** -- see the "not covered by either pass" line below, which named a
   mid-replay reset.

Cleared by the passes, for the record: every cited nats.go line holds (`ordered.go:364,571,605,
629,635,646`); the tests run under a genuinely restricted user (`WithUserPassword` sets the
`_INBOX.<user>` prefix); no grant is widened. Not covered by either pass: a mid-replay reset
forced under test, and a deployed `nats:2.10-alpine`. The first of those is where item 8 came
from -- `TestTasksGet_EmitsNothingTheBridgeGrantRefuses` asserts exactly one violation, the
deliberate control refusal, and never induces a reset, so it measures the no-reconnect path
only. Still not forced under test; item 8 documents the behaviour rather than pinning it.

## Risk & Rollout

One config field in `a2a/lib`, so every `TasksGet` caller (gateway, hermes bridge) picks it up on
its next image. The fold itself is unchanged and no grant moves, so there is no new failure mode
on the call path -- the worst case is the old one, five minutes of held slots, which is what
happens if the field is reverted. Every principal now behaves identically, which was not true
while the delete existed.

The five-second threshold cannot reap a replay in flight: nats-server 2.14.6 measures a pull
consumer's inactivity from its last pull or delivery and keeps it while a pull with live interest
is waiting (`server/consumer.go`, `deleteNotActive`), and an ordered iterator always has one
outstanding. The window is a disconnect longer than five seconds mid-replay, the same one the
adapter's comment already describes for its own consumers; nats.go's ordered consumer resets onto
a fresh consumer from the last sequence when that happens. Revert is one line and returns to the
five-minute leak.


bnaylor:
Confirmed, and fixed by dropping the emitter rather than widening the grant. `13178b2a`.

I verified every step of this against `main` before acting on it: the bridge does connect as static `worker` (`platformagent_a2a_identities.go:203` — and note it is not only the bridge sidecar, the platform agent pod authenticates as `worker` too), `a2aWorkerJetStreamGrants` does carry `CONSUMER.DELETE` for `KV_runtime-state` and not for TASKS, and `bridge.go:280`, `bridge.go:360` and `sweep.go:86,125` are the call sites. The finding is right on all of it.

The trade this PR made was deliberate — `deleteReplayConsumer`'s own doc comment said the rendered worker grant withholds the subject — but it was the wrong way round, for the reason you give: one Error line per dispatched task is not a cost, it is the destruction of a signal. `platformagent_a2a_worker_authz_test.go:366-368` teaches operators to read that exact line as "a grant is missing", and a line that fires once per task stops meaning that.

The fix is to drop the explicit delete and keep only `InactiveThreshold`. That was the option with no downside worth the name:

- The actual bug was the five-MINUTE nats.go default, and the threshold alone fixes it — five minutes to five seconds, a 60x cut in the term gke-labs#1739 is sizing `max_consumers` against.
- The explicit delete only ever bought back the last ~5s of one consumer slot.
- Granting the subject was the alternative, but on `worker` that is a destructive verb on a shared credential held by the widest-reach pod in the namespace. Not worth a log line.

So `deleteReplayConsumer` and `replayConsumerDeleteTimeout` are gone, `EphemeralConsumerInactiveThreshold` stays and is now documented as the only cleanup, with a note saying the way to reclaim a slot sooner is to shorten the constant, not to add a delete back.

The tests moved with it, and since two of them now assert an *absence*, they carry their own anti-vacuity guards. `TestTasksGet_EmitsNothingTheWorkerGrantRefuses` publishes a known-forbidden `$JS.API.STREAM.DELETE.TASKS` down the same connection after the replay and waits for that refusal to land before asserting no `CONSUMER.DELETE` is present — refusals arrive in publish order, so the control line proves the logger, the async error handler and the grant are all live. `TestTasksGet_ReplayConsumerCarriesTheInactiveThreshold` reads the threshold back off the server rather than out of the config struct.

Three mutations, all caught:

- `InactiveThreshold` removed from the config → server reports `5m0s`, want `5s`.
- the constant widened to `5 * time.Minute` → caught by an explicit guard, which is the mutation a config-struct-reading test would miss.
- an explicit delete re-added → `TestTasksGet_EmitsNothingTheWorkerGrantRefuses` fails with the permissions violation in the log, which is the regression this change exists to prevent.

The `a2aWorkerJetStreamGrants` comment is back to describing one emitter (nats.go's ordered reset path), keeping this PR's correction that the ephemeral is reaped by the 5s threshold rather than nats.go's default. `a2a/docs/hermes-bridge.md` now says which delete is granted and which is not, and that a `Permissions Violation` in the bridge log still means what it says. PR retitled, since "deletes its ordered consumer" stopped being what it does.

One thing I am leaving alone, flagged rather than fixed: `platformagent_a2a_worker_authz_test.go:328-332` labels a hand-rolled `js.OrderedConsumer(...)` as `"CONSUMER.CREATE TASKS ordered (TasksGet replay)"`, but it does not call `lib.TasksGet` and sets no `InactiveThreshold`. The operator-side model of the replay is not actually modelling it, which is why it neither caught gke-labs#1739 nor would have caught this. Worth its own issue.

Accepted delta, for the record: under a principal that does hold `CONSUMER.DELETE.TASKS.*`, the replay consumer now also lingers up to 5s rather than going immediately. That is the cost of the decision; there is no grant-conditional delete.

a2a/lib/fold.go
```diff
@@ -237,6 +270,46 @@ func (c *Client) TasksGet(ctx context.Context, addressee, taskID string) (*Task,
...
+		// consumer it left behind is exactly the one that must still go.
+		ctx, cancel := context.WithTimeout(context.Background(), replayConsumerDeleteTimeout)
+		defer cancel()
+		err := js.DeleteConsumer(ctx, TasksStream, name)
+		if err == nil || errors.Is(err, jetstream.ErrConsumerNotFound) {
+			return
+		}
+		c.log.Debug("a2a replay consumer not deleted; the inactive threshold reaps it",
+			"task", taskID, "consumer", name, "threshold", EphemeralConsumerInactiveThreshold, "err", err)
+	}()
```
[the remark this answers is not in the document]
bnaylor: Confirmed and fixed in `13178b2a`, by dropping the emitter rather than widening the grant: `deleteReplayConsumer` and `replayConsumerDeleteTimeout` are gone, `EphemeralConsumerInactiveThreshold` stays as the only cleanup.

Granting `$JS.API.CONSUMER.DELETE.TASKS.*` to `worker` was the alternative and it was the wrong one -- that is a destructive verb on a shared credential the platform agent pod also authenticates as, bought to silence a log line. The five-minute nats.go default was the real bug and the threshold alone fixes it.

Full reasoning, the three mutations that pin it, and one adjacent defect I am flagging rather than fixing: https://github.com/gke-labs/kube-agents/pull/1741#issuecomment-5725610759

bnaylor:
Correction to my comment above, on one fact that #1653 changed out from under it.

I wrote that the bridge "connects as static `worker`", and that the platform agent pod authenticates as `worker` too so the grant would be shared. That was true of `main` at `f3c819d5`, which is what I verified against and what this branch is based on. It stopped being true at 03:12Z, when #1653 merged as `965b6917` and retired the static `worker` user, splitting it into `bridge` and `agent`. `a2aWorkerJetStreamGrants` is now `a2aBridgeJetStreamGrants`, and `TestTheWorkerCredentialIsGone` refuses the old credential's return.

What survives unchanged: `a2aBridgeJetStreamGrants` still carries `CONSUMER.DELETE` for `KV_runtime-state` only, and still withholds `$JS.API.CONSUMER.DELETE.TASKS.*`. So the finding is still correct against today's `main`, the refused publish would still happen, and dropping the emitter is still the fix — removing a publish cannot depend on who holds the grant.

What does not survive is one of the reasons I gave for dropping it rather than granting it. "That is a destructive verb on a shared credential the platform agent pod also authenticates as" was an argument against widening `worker`, and after the split it would only widen `bridge`. The decision still stands on the rest — the five-minute default was the actual bug, the threshold alone fixes it, and the explicit delete only ever bought back the last few seconds of one consumer slot — but it stands on a narrower base than I claimed, and the reader deserves the accurate version.

Two consequences for this branch, which I am fixing rather than leaving:

- The prose in `13178b2a` names `worker` and `a2aWorkerJetStreamGrants`. Those identifiers no longer exist on `main`, so it has to be re-spelled for `bridge`.
- This branch is based pre-#1653 and the comment it edits lives in a function that was renamed, so it needs a merge-up before it can land.

Both are in progress; I'll note the commit here when they are in.

bnaylor:
Synced with main and folded in two corrections from review.  Head is now `c9645052`.

**Merge, not rebase** (`38d9edb2`) -- #1653 landed and retired the static `worker` bus user, splitting it into `bridge` and `agent`.  Two conflicts, both prose: the "DELETE stays out" sentence in `platformagent_a2a_manifests.go`, and the grants section of `a2a/docs/hermes-bridge.md`, which main rewrote wholesale.  I took main's new paragraph there and dropped the old "this is now the bridge's own debt" note that used to precede mine -- its claim that there is no `agent` principal in the rendered map is exactly what #1653 falsified, so resurrecting it would have re-introduced a false statement.

**The basis still holds after the rename.**  `a2aBridgeJetStreamGrants()` carries exactly one delete, `$JS.API.CONSUMER.DELETE.KV_runtime-state.*`.  There is still no delete on `TASKS`.  #1653's `TestTheWorkerCredentialIsGone` and `TestBridgeJetStreamGrantOnARealServer` both pass unmodified.

### Two corrections to the justification (`c9645052`, comment only)

**1.  The refusal is not only a log line -- it also blocks.**  I described the cost as "an Error-level permissions violation on every call".  That understates it.  A refused publish on a *request* subject gets no reply at all, so a synchronous delete waits out the caller's own deadline while the violation arrives out of band on the connection's async error handler.  Both halves happen.

Measured, by re-adding a synchronous delete as a mutation:

```
clean:              --- PASS: TestTasksGet_EmitsNothingTheBridgeGrantRefuses (0.11s)
delete re-added:    --- FAIL (30.07s)
  Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.ku3eyoexy2MviLLK3ZpWy1_1"
```

0.11s to 30.07s.  The original implementation ran the delete in a detached goroutine with its own 5s bound, which is the only reason a `tasks/get` did not stall -- the goroutine absorbed it.  That is a wrapper around a call that cannot succeed and whose result the caller cannot observe, which is the argument for deleting the path rather than tuning it.

**2.  Why not just widen the grant.**  It is the fix that suggests itself and it is worse than the leak it closes.  `$JS.API.CONSUMER.DELETE.TASKS.*` is a wildcard over consumer *names*, not over the consumers its holder happened to create, so granting it gives anything carrying the bridge password delete on every consumer on `TASKS` -- the gateway's own `gateway-relay` durable included.  `platformagent_a2a_bridge_authz_test.go:386` already pins that exact subject as refused, and `sessionConsumer` in the worker adapter (`a2a/worker-adapter/adapter.go:801`) makes the same argument at length for `MSG.NEXT`.  Both are now referenced from the doc comment so the next reader does not have to re-derive it.

### Verification

Full suites green on the merge, from both module roots:

```
a2a/authcallout 33.7s   a2a/lib 20.5s        a2a/gateway 25.7s
a2a/hermes-bridge 14.2s a2a/worker-adapter 22.9s
k8s-operator/internal/controller 80.6s
```

Both mutations still bite after the merge:

| mutation | result |
| --- | --- |
| drop the `InactiveThreshold:` line | `FAIL`: server reports `inactive_threshold = 5m0s, want 5s` |
| re-add an explicit `DeleteConsumer` | `FAIL`: permissions violation, and 30.07s |

The threshold test reads 5s back off the server and sees the consumer reaped at 5.6s; the burst test leaves 20 consumers, each at 5s, all clear by 5.9s.

Thanks to @dshnayder for the second point -- the grant-widening counter-proposal was the obvious thing to raise in review and it deserved an answer in the code rather than in a thread.

bnaylor:
Addendum to the grant-widening point above, from @dshnayder -- it is a stronger objection than the one I wrote and worth stating properly.  `1c78815d`.

The problem with widening isn't only that `$JS.API.CONSUMER.DELETE.TASKS.*` is broad.  It is that **nothing narrower can be written.**  nats.go names an ordered consumer with a serial that bumps on every reset:

```go
// jetstream/ordered.go:629, v1.53.1
name := fmt.Sprintf("%s_%d", c.namePrefix, c.serial)
```

So there is no exact-name grant that can cover the consumer `TasksGet` creates -- a wildcard is the only shape on offer, and a wildcard over consumer names is a wildcard over *every* consumer on the stream, including ones this caller never created.  The gateway's `gateway-relay` durable is on `TASKS`.

This is not a new argument in the tree, which is the best part: it is the stated reason the worker adapter uses named consumers rather than ordered ones, written out at `a2a/worker-adapter/adapter.go:801` for `MSG.NEXT`.  The comment now cites it rather than re-deriving it.

(While confirming `:629` I also re-checked the two lines the threshold comment already cites -- `:635` for the five-minute default and `:646` for the non-zero override.  Both accurate at v1.53.1.)

a2a/lib/fold.go
```diff
@@ -5,13 +5,56 @@ import (
...
+	// deletes the consumer behind it (orderedSubscription.Stop, :364). With
+	// the default in place every tasks/get left a consumer on TASKS for five
+	// minutes, so the count tracked the call rate over that window rather
+	// than the replays in flight, against a max_consumers sized for the
+	// latter (#1739).
+	//
+	// The threshold is the only cleanup: TasksGet does not delete the
+	// consumer it creates, and must not be made to. The delete is a publish
+	// on $JS.API.CONSUMER.DELETE.TASKS.*, a subject the rendered bridge
+	// grant withholds, and a refused publish on a request subject gets no
```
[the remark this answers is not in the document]
bnaylor: Correct, and entirely my omission. When review settled on dropping the explicit delete I rewrote the code across `13178b2a`, `c9645052` and `1c78815d` and never rewrote the description, so it kept describing the withdrawn delete-by-name design: `deleteReplayConsumer`, `replayConsumerDeleteTimeout`, the goroutine with its own deadline, the `gateway`/`gateway-nodelete` users, the "0 consumers the instant the call returns" table, and Self-Review item 4's `TestDeleteReplayConsumer_NeedsNoLiveCallerContext`. None of it is on the branch.

The description is rewritten now. Summary, What Changed, Testing and Self-Review describe the threshold-only change: `InactiveThreshold: EphemeralConsumerInactiveThreshold` on the ordered consumer and nothing else at runtime, the three tests that exist (`TestTasksGet_ReplayConsumerCarriesTheInactiveThreshold`, `TestTasksGet_EmitsNothingTheBridgeGrantRefuses`, `TestTasksGet_BurstBoundsReplayConsumersByTheThreshold`), the measured numbers (1 consumer at `inactive_threshold=5s` reaped at 5.1s; 20 at `5s`, clear at 6.0s), and the two mutations behind the comment -- dropping the field gives `5m0s, want 5s`, and re-adding a synchronous delete gives FAIL at 30.07s with `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"`. Self-Review items 1 and 4 are now marked resolved by the design change rather than fixed, which is what actually happened to them. A dated note at the top of the description says the design flipped, so anyone who read the old text knows why it no longer matches.

No code changed for this.

a2a/lib/replay_consumer_test.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/nats-io/nats.go/jetstream"
+)
+
+// The consumer TasksGet creates on TASKS must be gone within
+// EphemeralConsumerInactiveThreshold of the call returning, and TasksGet must
+// not try to delete it by hand (#1739). Both are measured on a real server
+// against the rendered bridge grant, because the second one only matters
+// under that grant: the delete would be a $JS.API.CONSUMER.DELETE.TASKS.<name>
+// publish, and this principal does not hold the subject.
+const (
```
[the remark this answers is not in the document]
bnaylor: Correct, and entirely my omission. When review settled on dropping the explicit delete I rewrote the code across `13178b2a`, `c9645052` and `1c78815d` and never rewrote the description, so it kept describing the withdrawn delete-by-name design: `deleteReplayConsumer`, `replayConsumerDeleteTimeout`, the goroutine with its own deadline, the `gateway`/`gateway-nodelete` users, the "0 consumers the instant the call returns" table, and Self-Review item 4's `TestDeleteReplayConsumer_NeedsNoLiveCallerContext`. None of it is on the branch.

The description is rewritten now. Summary, What Changed, Testing and Self-Review describe the threshold-only change: `InactiveThreshold: EphemeralConsumerInactiveThreshold` on the ordered consumer and nothing else at runtime, the three tests that exist (`TestTasksGet_ReplayConsumerCarriesTheInactiveThreshold`, `TestTasksGet_EmitsNothingTheBridgeGrantRefuses`, `TestTasksGet_BurstBoundsReplayConsumersByTheThreshold`), the measured numbers (1 consumer at `inactive_threshold=5s` reaped at 5.1s; 20 at `5s`, clear at 6.0s), and the two mutations behind the comment -- dropping the field gives `5m0s, want 5s`, and re-adding a synchronous delete gives FAIL at 30.07s with `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"`. Self-Review items 1 and 4 are now marked resolved by the design change rather than fixed, which is what actually happened to them. A dated note at the top of the description says the design flipped, so anyone who read the old text knows why it no longer matches.

No code changed for this.

a2a/docs/hermes-bridge.md
```diff
@@ -88,6 +88,15 @@ user consumes with explicit ack (unscoped `$JS.ACK.>` is a cross-principal +TERM
 sets `--allow-direct` on every stream so nats.go picks that route, and only that route
```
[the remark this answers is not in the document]
bnaylor: Correct, and entirely my omission. When review settled on dropping the explicit delete I rewrote the code across `13178b2a`, `c9645052` and `1c78815d` and never rewrote the description, so it kept describing the withdrawn delete-by-name design: `deleteReplayConsumer`, `replayConsumerDeleteTimeout`, the goroutine with its own deadline, the `gateway`/`gateway-nodelete` users, the "0 consumers the instant the call returns" table, and Self-Review item 4's `TestDeleteReplayConsumer_NeedsNoLiveCallerContext`. None of it is on the branch.

The description is rewritten now. Summary, What Changed, Testing and Self-Review describe the threshold-only change: `InactiveThreshold: EphemeralConsumerInactiveThreshold` on the ordered consumer and nothing else at runtime, the three tests that exist (`TestTasksGet_ReplayConsumerCarriesTheInactiveThreshold`, `TestTasksGet_EmitsNothingTheBridgeGrantRefuses`, `TestTasksGet_BurstBoundsReplayConsumersByTheThreshold`), the measured numbers (1 consumer at `inactive_threshold=5s` reaped at 5.1s; 20 at `5s`, clear at 6.0s), and the two mutations behind the comment -- dropping the field gives `5m0s, want 5s`, and re-adding a synchronous delete gives FAIL at 30.07s with `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"`. Self-Review items 1 and 4 are now marked resolved by the design change rather than fixed, which is what actually happened to them. A dated note at the top of the description says the design flipped, so anyone who read the old text knows why it no longer matches.

No code changed for this.

a2a/lib/replay_consumer_test.go
```diff
@@ -0,0 +1,344 @@
...
+	waitFor(t, logWithin, "the control refusal to reach the async error handler", func() bool {
+		out := logs.String()
+		return strings.Contains(out, "Permissions Violation") && strings.Contains(out, replayControlSubj)
+	})
+
+	out := logs.String()
+	if strings.Contains(out, "CONSUMER.DELETE") {
+		t.Fatalf("TasksGet was refused a CONSUMER.DELETE under the bridge grant; the log line operators read as a missing grant is now one per tasks/get:\n%s", out)
+	}
+	t.Logf("client log after the replay, with a known refusal appended as the barrier:\n%s", out)
```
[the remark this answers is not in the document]
bnaylor: Confirmed and fixed in `6a337ade`, including the run you could not make.

You are right that the guard was narrower than three separate claims -- the test name, the Testing section, and the sentence this branch adds to `a2a/docs/hermes-bridge.md` ("a `tasks/get` under this user emits no refused publish"). `CONSUMER.DELETE` is the subject the change is about, but it is not the only one the grant withholds, and `TasksGet`'s request path is wider than the ordered consumer: the horizon read ahead of it emits `STREAM.INFO` and a direct get.

Took your suggested form. The assertion is now `strings.Count(out, "Permissions Violation") != 1` plus a check that the one line names `replayControlSubj`, so the control refusal doubles as the expected count rather than just as the barrier.

The mutation you named, run both ways -- an unconditional `Publish` to `$JS.API.CONSUMER.NAMES.TASKS` on `TasksGet`'s path, ahead of the `OrderedConsumer` call, with `fold.go` restored byte-clean after each:

- against the assertion as it stood: `ok github.com/gke-labs/kube-agents/a2a/lib 0.465s` -- green, exactly as you predicted
- against the new one: `FAIL ... 2 "Permissions Violation" lines on the reader's connection, want exactly 1 (the control publish on $JS.API.STREAM.DELETE.TASKS)`, with both lines printed

Your read of the backstop is right too, and I checked it rather than taking it: `TestBridgeConsumersSurviveABusRestart` drives a bare `js.OrderedConsumer`, not `lib.TasksGet`, so nothing was covering the horizon read. It is covered now -- the count is over the whole call, both halves.

bnaylor:
Both low-severity items from the 08:30Z review are fixed in `f908965b` (no threads were opened on them, so this comment is the reply).

**`hermes-bridge.md` "each call creates an ordered consumer on TASKS."** Confirmed false, and the bot named the right path. `TasksGet` takes its replay horizon with `GetLastMsgForSubject` on both subjects first, and returns `TaskNotFound` when neither holds a message (`a2a/lib/fold.go:207`) — before `OrderedConsumer` is reached (`:208`). So a call on a task the retention window no longer holds creates nothing. The paragraph now says which calls cost a consumer slot and which do not, and keeps the conclusion it was there for: neither path emits a refused publish, so a `Permissions Violation` in the bridge log still means what it says.

Same imprecision was in the PR Summary ("after every call"); tightened there too.

**`replay_test.go` doc comment.** Mine — I extended `replayFixture`'s comment with the `opts` sentence and it landed above `replayAddressee`, which sits between the comment and its function. Moved to `replayFixture`.

`prettier --check`, `gofmt -l`, `go build ./...`, the `lib` replay tests, and `make docs-check` all clean on the new head.

bnaylor:
Merged `main` up after #1665 landed (`88dad0a0`).

One conflict, in `a2a/worker-adapter/adapter.go`: #1665 and this branch both rewrote `consumerInactiveThreshold`'s comment, correcting the same wrong claim — that nats.go's ordered consumers default to five seconds. They default to five **minutes**.

Took this branch's wording, because the merged *code* is this branch's. The value line merged cleanly to `consumerInactiveThreshold = lib.EphemeralConsumerInactiveThreshold` (main never touched it), so main's "Five seconds is this adapter's own number, not an inherited one" would be false against the result — after this PR the adapter and `lib.TasksGet` share one constant, which is the point. This branch's text also carries the same correction with exact cites (`jetstream/ordered.go:635`, `:646`) and the #1739 provenance. Folded in main's "wrong by 60x", the one thing it said better.

Re-verified on the merge result: `gofmt -l`, `go build ./...`, `go vet ./...`, the full `a2a` suite, the full `k8s-operator` suite under envtest, `make docs-check`, and `prettier --check` all clean.

(`k8s-operator/internal/controller/platformagent_litellm_policy_test.go` is gofmt-unclean on the merged tree. That is pre-existing drift arriving from `main` — #1545 exists to fix it — not from this branch.)

a2a/docs/hermes-bridge.md
```diff
@@ -88,6 +88,17 @@ user consumes with explicit ack (unscoped `$JS.ACK.>` is a cross-principal +TERM
...
+Note which delete is in that list and which is not: `$JS.API.CONSUMER.DELETE` is granted
+for `KV_runtime-state`, for the watcher, and withheld for TASKS. The bridge calls
+`lib.TasksGet` on every task it dispatches, and a call that finds events creates an ordered
+consumer on TASKS; nothing deletes it. It is reaped by the five-second inactive threshold
+`TasksGet` sets on it, which is why the replay costs a consumer slot for the calls in flight
+rather than for the last five minutes of them (gke-labs/kube-agents#1739) without the bridge
+needing a destructive verb on TASKS. A call on a task the retention window no longer holds
+creates no consumer at all -- the horizon read returns `TaskNotFound` before the consumer is
+created. Either way a `tasks/get` under this user emits no refused publish, so a
+`Permissions Violation` in the bridge's log still means what it says.
```
jayantid: 🟡 **UX flow.** The two texts this PR writes about the bridge log disagree, and the operator-facing one is the wrong one.

`a2a/docs/hermes-bridge.md:100` says a `tasks/get` under `bridge` emits no refused publish, so a `Permissions Violation` in the bridge's log "still means what it says". `platformagent_a2a_manifests.go:589`, in the same diff, says the grant's absence "costs a log line on a reset": the ordered consumer's reset path still fires `CONSUMER.DELETE` for the consumer it replaces, and `bridge` is refused. Both are true; the doc states only the steady-state half.

After a disconnect longer than five seconds mid-replay, the bridge log carries `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"`. An operator who reads the doc chases a missing grant that is not missing.

Fix: one clause in the doc naming that subject as the one refusal a reset produces by design, so "any other violation means what it says" stays a usable rule.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
```diff
@@ -577,7 +585,9 @@ func a2aSeedJetStreamGrants() []string {
 // NATS wildcards match whole tokens, so a per-prefix grant matches a consumer
 // literally named that. What closes it is the auth callout giving each
 // principal its own user. DELETE stays out as the one destructive verb here
-// that nothing on the bridge path needs.
+// that nothing on the bridge path needs: its one emitter ignores the result
+// and the threshold gets there anyway, so the grant's absence costs a log
```
jayantid: 🟡 **Product requirements.** The reason this PR gives for leaving #1739's items 2 and 4 behind expired inside the PR, and a comment in a file it edits now states the opposite of what it ships.

Context says the replay term in `a2aTasksReservedConsumers` and the re-itemisation of its comment "follow once #1665 lands" because those constants live on that branch. #1665 merged at 10:47Z and this head already carries it. `platformagent_a2a_manifests.go:295-302` still says `TasksGet` leaves the threshold unset and holds a slot for five minutes. The bot noted that line as low; the point here is that the deferral no longer has a premise.

Whoever sizes `maxSessions` reads that itemisation and budgets for a five-minute window this PR removed.

Fix: correct the two sentences at 295-302 in this PR, since it is the change that made them false. Item 2, the replay term itself, can stay a follow-up if the comment says the term is still absent and why five seconds makes it small.

bnaylor:
All three folded low-severity items from the 11:28Z review are fixed in `c6504911`. All three were confirmed against source first, and two of them are drift this branch itself caused.

**1. `a2aTasksReservedConsumers` comment (`platformagent_a2a_manifests.go`).** The real one. It still said `TasksGet` leaves `InactiveThreshold` unset, so nats.go's five-minute default applies and every call holds a slot for five minutes — which is exactly what this branch stops being true. The comment's argument survives intact (the missing term is a call *rate* over a rolling window, not a concurrency), but the window is five seconds now, so the rate the constant has to absorb is 60x lower than the number was sized against. Said that, and left #1739 owning the number and the constant unmoved.

**2. `fold.go` citing `adapter.go:801`.** Correct when written; merging `main` up moved `sessionConsumer`'s doc comment to 816. Updated.

**3. `lockedBuffer` in `replay_consumer_test.go`.** Confirmed duplication, and a fresh one: `logCapture` (`resilience_test.go`) is the package's mutex-guarded slog sink, and #1665 had just started using it for this same purpose in `subject_cap_test.go`. Dropped `lockedBuffer`.

The sink swap changes the captured line format, so I re-measured the refusal test's mutation row rather than carrying it over — a table row is only true of the version it was measured against. With a `DeleteConsumer` re-added to `TasksGet` the test fails with `2 "Permissions Violation" lines ... want exactly 1`, naming both subjects; reverted, it passes in 0.44s.

Re-verified on `c6504911`: `gofmt -l`, `go build`, `go vet`, the full `a2a` and `k8s-operator` suites under envtest, `make docs-check`, `prettier --check` — all clean.

a2a/docs/hermes-bridge.md
jayantid: 🟡 **UX flow.** The operator-facing doc says a violation in the bridge log always means a missing grant, and the grant comment in the same diff says a reset produces one by design.

`a2a/docs/hermes-bridge.md:100` says a `tasks/get` under `bridge` emits no refused publish, so a `Permissions Violation` "still means what it says". `platformagent_a2a_manifests.go:589`, in this diff, says the grant's absence "costs a log line on a reset". Both hold: nats.go's ordered reset fires `CONSUMER.DELETE` for the consumer it replaces (v1.53.1 `ordered.go:570`), and `bridge` is refused. The bot folded this fact as low at `89fc816`; `f908965` then rewrote the paragraph and kept the claim.

After a disconnect longer than five seconds mid-replay, the bridge log carries `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"`. The doc is the text an operator reads at that moment, and it sends them after a grant that is not missing.

Fix: one clause naming that subject as the one refusal a reset produces by design, so "any other violation means what it says" stays a usable triage rule.

k8s-operator/internal/controller/platformagent_a2a_manifests.go
jayantid: 🟡 **Product requirements.** The reason this PR gives for leaving #1739's items 2 and 4 behind expired inside the PR, and a comment in a file it edits now states the opposite of what it ships.

Context says the replay term in `a2aTasksReservedConsumers` and the re-itemisation of its comment "follow once #1665 lands" because those constants live on that branch. #1665 merged at 10:47Z and this head already carries it. `platformagent_a2a_manifests.go:295-302` still says `TasksGet` leaves the threshold unset and holds a slot for five minutes. The bot noted that line as low; the point here is that the deferral no longer has a premise.

Whoever sizes `maxSessions` reads that itemisation and budgets for a five-minute window this PR removed.

Fix: correct the two sentences at 295-302 in this PR, since it is the change that made them false. Item 2, the replay term itself, can stay a follow-up if the comment says the term is still absent and why five seconds makes it small.

a2a/docs/hermes-bridge.md
jayantid: 🟡 **UX flow.** The operator-facing doc says a violation in the bridge log always means a missing grant, and the grant comment in the same diff says a reset produces one by design.

`a2a/docs/hermes-bridge.md:100` says a `tasks/get` under `bridge` emits no refused publish, so a `Permissions Violation` "still means what it says". `platformagent_a2a_manifests.go:592`, in this diff, says the grant's absence "costs a log line on a reset". Both hold: nats.go's ordered reset fires `CONSUMER.DELETE` for the consumer it replaces (v1.53.1 `ordered.go:571`), and `bridge` is refused. The bot folded this as low at `89fc816`; `f908965` rewrote the paragraph and kept the claim.

A bus reconnect mid-replay is enough: the ordered iterator resets on reconnect (`ordered.go:321`), and the bridge log then carries `Permissions Violation for Publish to "$JS.API.CONSUMER.DELETE.TASKS.<name>"`. The doc is the text an operator reads at that moment, and it sends them after a grant that is not missing.

Fix: one clause naming that subject as the one refusal a reset produces by design, so "any other violation means what it says" stays a usable triage rule.

bnaylor: Holds, and the doc was the wrong half. Fixed in `b312415d`.

I re-measured each link rather than taking the chain on trust, and the cited line numbers are exact:

- **`reset()` really does delete.** `jetstream/ordered.go:558`, nats.go v1.53.1 as pinned in both `go.mod`s: `go func(){ ... _ = c.js.DeleteConsumer(ctx, c.stream, consName) ... }()`, fired for `c.currentConsumer` before it is replaced, result discarded.
- **Reconnect reaches it.** `pull.go:688` returns `errConnected` from `Next()`, which escapes the four filters at `ordered.go:297-309` and lands on `reset()` at `:321`. Reset is broader than reconnect, though -- missed heartbeat, sequence mismatch, consumer-deleted and no-responders all get there too.
- **The subject is refused.** `a2aBridgeJetStreamGrants()` grants `$JS.API.CONSUMER.DELETE` on `KV_runtime-state` only, so `$JS.API.CONSUMER.DELETE.TASKS.<name>` matches nothing in the list.
- **The log line is verbatim what you predicted.** nats-server `client.go:5784` formats `Permissions Violation for Publish to %q`; nats.go invokes the async callback with a nil sub, so `a2a/lib/client.go:243` logs an empty `subject` field and the real subject rides inside `err`.

**Two things worth adding to the record.** The *first* consumer creation emits nothing -- `reset()`'s delete is guarded on `currentConsumer != nil`, and `Messages()` skips the initial reset because `jetstream.go:918` creates eagerly -- so the paragraph's baseline was correct and only the reset case was missing. And the reason this got past its own test: `TestTasksGet_EmitsNothingTheBridgeGrantRefuses` asserts exactly one violation, the deliberate control refusal, and never induces a reset, so it measures the no-reconnect path only.

**The grant comment at `:592` needed no change** -- it was already scoped to the reset and already accurate. Only `hermes-bridge.md` moved, and the two texts now agree.

One judgement call for your eye: I wrote "a bus reconnect is enough" rather than enumerating all five reset triggers, on the grounds that reconnect is the one an operator actually hits and an exhaustive list makes a triage sentence worse. Say the word if you'd rather it be complete.

I also checked the class rather than just the instance. The only other `Permissions Violation` mentions in tracked Markdown are `docs/architecture/09-capability-envelope.md:268` and `:577`, and neither is triage guidance -- both make the opposite-direction point that a refusal surfaces as a server-side publish error rather than a refused connection. Nothing else carries the over-broad claim.

`make docs-check` and `prettier --check` both green at `b312415d`.

bnaylor:
Round 5 clean, and the one folded low-severity item held. Fixed in `1858ce9b`.

**The folded item was right.** The doc said the replay "costs a consumer slot for the calls in flight"; it costs one for the calls of the last five seconds. `a2a/lib/fold.go:56` sets `EphemeralConsumerInactiveThreshold = 5 * time.Second`, nothing deletes the replay consumer, and the threshold runs from the call *returning* rather than from it starting -- `replay_consumer_test.go:16` says exactly that. So the slot outlives the call it served, and "in flight" understates the held set by one threshold per call. The five-seconds-against-five-minutes contrast the sentence is drawing (#1739) is unaffected, which is why the sentence keeps its shape.

It landed in the same paragraph as jayantid's finding above, which is a fair verdict on that paragraph: two claims in four lines, both slightly over-stated in the safe-sounding direction.

`make docs-check` and `prettier --check` green at `1858ce9b`. Documentation only, no Go touched.

**On the smoke job**, since the timing looks careless otherwise: this push aborts a run that was 51 minutes in, deliberately. A completed smoke at `b312415d` is worth nothing once anything lands on top of it, so the real choice was fix-now or never, not fix-later. This PR is waiting on a human approval measured in days; a smoke cycle is measured in hours.

**Not requesting another round.** Two consecutive clean passes on a docs-only delta is where I stop.

bnaylor:
woops, wrong bug

bradhoekstra (approved):

PR #1742, opened 2026-09-17
fix(tests): start the drain-deadline clock before sendall so the client window is a superset of the server's

### LINKED ISSUE #1740: flake: test_drain_deadline_releases_a_client_that_stalls_mid_body measures the drain from after sendall, so it can read short
`test_drain_deadline_releases_a_client_that_stalls_mid_body`
(`agents/platform/scripts/test_credential_proxy.py:261`) failed on run 35277566839:

```
AssertionError: 0.24951860300001272 not greater than or equal to 0.25
```

0.4ms short of the deadline. The test asserts the client observed at least the full drain
window, but the window it measures is a strict subset of the one the server is enforcing.

## Mechanism

The clock starts after the request is handed to the kernel:

```python
client.sendall(request)
started = time.monotonic()
```

`sendall` returns once the last byte is in the socket buffer, not once the server has read it.
The server may already be some way into its drain by the time `started` is taken. Everything
between "server begins draining" and "client records `started`" is drain time the server spent
and the client does not count — so `elapsed` is systematically an underestimate of the server's
window, and a scheduling hiccup of a fraction of a millisecond on a loaded runner is enough to
put it under.

Nothing is wrong with the drain deadline itself; the assertion is measuring the wrong interval.

## Fix direction

Take `started` before the `sendall` rather than after. That makes the client's window a
superset of the server's, so the inequality can only be violated by the server actually
returning early — which is what the test is for.

Not a tolerance bump: a tolerance hides a real early return of the same size, and there is a
correct interval available here.

bnaylor:
## Summary

`test_drain_deadline_releases_a_client_that_stalls_mid_body` started its clock after `client.sendall(request)`, so the interval it measured was a strict subset of the interval the server enforces, and a sub-millisecond scheduling gap on a loaded runner made it read 0.2495 against a 0.25 deadline. This moves `started = time.monotonic()` to before the `sendall`. The bound is unchanged; the clock now brackets the server's window instead of sitting inside it, so the assertion can only fail when the drain actually returned early.

## Why This Change

The server's window is `AgentAPIProxyHandler._drain_request_body` (`agents/platform/scripts/credential_proxy.py`): after `_proxy` sees the wrong key it calls `settimeout(AGENT_API_DRAIN_TIMEOUT_SECONDS)` on the connection and loops on `rfile.read`. The first read returns the 1024 partial bytes the test sent; the second blocks on a socket that will never deliver more, and the timeout raises after at least the deadline. That window opens when the second `read` blocks, which is after the server has accepted the connection, parsed the headers and consumed the partial body. None of that waits for the client's `sendall` to return: the bytes reach the kernel inside `sendall`, and the server thread (in-process, `ThreadingHTTPServer`) can wake, take the GIL, parse, fail the key check and block in its timed read before the client thread returns from `sendall` and reaches `time.monotonic()`. A client clock started after `sendall` can therefore start after the server's window has already opened, and the drain time spent before the client thread was next scheduled is not counted. On loopback the gap is normally tens of microseconds; on a busy runner it was 0.5 ms, and run 35277566839 failed with `0.24951860300001272 not greater than or equal to 0.25`.

Started before `sendall`, the client clock precedes the bytes leaving the socket, which precedes the server reading them, which precedes the server's window opening, so `elapsed` is a superset of the enforced interval and `elapsed >= _SHORT_DRAIN_TIMEOUT_SECONDS` holds whenever the server waited out the deadline. A tolerance would have hidden a real early return of the same size; the reordering costs nothing and removes the underestimate rather than allowing for it.

## What Changed

- `agents/platform/scripts/test_credential_proxy.py`: `started = time.monotonic()` moves from after `client.sendall(request)` to before it, with a comment recording why the order matters. No constant, bound or server code changes.

No sibling has the same defect. Every other lower-bound timing assertion in the repository (`deploy/docker/patches/test_kanban_wake_nudge.py`, four cases) starts its clock before the awaited call, and no other test in this file or in `agents/platform/scripts` takes a clock after a `sendall`, `urlopen` or equivalent dispatch and then asserts a lower bound (`grep -rn -B2 -A3 'started = time.monotonic()' --include='test_*.py'` and `grep -rn 'assertGreaterEqual(elapsed\|assertGreaterEqual(time.monotonic() - ' --include='test_*.py'`).

## Context

Closes #1740.

Open PRs that touch `agents/platform/scripts/test_credential_proxy.py` (#1732, #1710, #1652, #1372, #1304, #1241, #1234, #1161, #1159, #1040) were checked by diff: none edits this test, its constants, or the `sendall`/`started` lines (`gh pr diff <n> | grep 'started = time.monotonic\|test_drain_deadline_releases\|_SHORT_DRAIN_TIMEOUT_SECONDS'` is empty for all ten; #1304's single hit is an unrelated `started = time.monotonic()` in a new module of its own). No open PR or issue other than #1740 mentions `test_drain_deadline`.

## Testing

All runs use Python 3.12.8 in a venv with `requirements-test.txt` installed (CI pins 3.12 via `.python-version`), from `agents/platform/scripts` with the Makefile's `PYTHON_TEST_PATH` exported, invoking the module the way `unittest discover -p "test_*.py"` collects it. Transcripts are in the evidence directory named below.

- `ruff check --isolated --select E9,F63,F7,F82` on the changed file: clean.
- The whole file once on the branch: 303 tests, the only failure `GitHardeningTest.test_alias_configuration_and_repo_local_execution_are_refused`, which fails identically on `upstream/main`'s copy of the file on this machine (a local git build; `git init --bare -b` remote-default-branch detection returns `None`) and is unrelated to this change.
- The whole file in a loop on the branch, 50 iterations (303 tests each): the changed test passed 50 of 50, and the only failure in any iteration was the pre-existing local `GitHardeningTest` case above. About 14 of those iterations overlapped the CPU-pressure run below and ran under fourteen busy-loop processes.
- The changed test alone, 200 iterations on the branch, no CPU pressure: 200 of 200 passed.
- Reproduction of the defect on `upstream/main`'s test, in a throwaway copy under `/tmp` (never in the worktree): with a `time.sleep(0.002)` inserted between `sendall` and the old `started`, and the server's own drain duration printed from `_drain_request_body`, 20 of 20 runs failed the assertion with the client reading 0.2483 to 0.2497 s while the server measured 0.2502 to 0.2511 s for the same drain. The CI value of 0.2495 sits inside that range. Under CPU pressure (fourteen `yes > /dev/null` on a fourteen-core machine, no inserted sleep, no instrumentation), `upstream/main`'s test and the fixed test interleaved for 200 iterations each under the same load: the pre-fix test failed 2 of 200 (0.2444 s and 0.2470 s), the fixed test passed 200 of 200.
- The fixed assertion keeps its teeth: the same throwaway approach with the server's `settimeout` shortened by `DRAIN_EARLY_SECONDS`. Early by 10 ms: 10 of 10 runs fail. Early by 2 ms: 10 of 10 fail. Early by 1 ms and 0.5 ms: 10 of 10 pass, because the client's window carries about 1 to 1.5 ms of loopback and thread-handoff overhead on this machine, which is the floor of what any client-side measurement of this kind can resolve; the pre-fix ordering had the same floor and additionally a false-positive band on the other side. The bound itself is still exactly 0.25.
- Throwaway copies were made under `/tmp/drain-repro/`, diffed against their sources (diffs are in the evidence directory), and the worktree carries nothing but the one-file commit: `git status --short` is clean and `git diff upstream/main...HEAD --stat` shows the single file, 8 insertions, 1 deletion.

Evidence: `round_3/issue-sweep/drain-deadline-flake-evidence/` in the kube-agents-vamp repository (00-provenance.txt, 00-diff-*.patch, 01 whole-file run, 02 whole-file loop, 03 single-test loop, 04 instrumented pre-fix reproduction plus its diff, 05 CPU-pressure comparison, 06 early-return mutation plus its diff, 07 whole-file and 20x single-test run on the final commit after the comment-only amendment).

### Live validation

Not applicable: a test-only change with no runtime code path touched; proven by the loop runs and the early-return mutation above.

## Self-Review

Both passes ran in fresh subagents that were handed only the worktree path, the diff range `upstream/main...HEAD`, and the skill file to follow.

- Adversarial pass (fresh subagent, `.agents/skills/review-adversarial/SKILL.md`, angles A through J). One confirmed finding, LOW: the first version of the new comment said the server's window opens "after sendall returns", which is false at the granularity the bug lived at and would license the old placement; the window opens once the bytes have left the socket, and the in-process server thread can reach its timed read before this thread returns from `sendall`. Fixed: the comment now says so, and this body's mechanism paragraph was corrected to match. The pass also ran its own red-then-green: the `upstream/main` ordering failed 1 of 300 under 28 busy-loop processes (`0.24882 not greater than or equal to 0.25`), the new ordering 0 of 300 with a minimum of 0.250557 s. Considered and not findings, with the pass's reasoning accepted: the new interval cannot resolve a sub-millisecond early return (no client-side measurement could under either ordering, and the test's purpose is the handler ignoring the deadline, where `elapsed` would be about 1 ms, not 249 ms); `(#1740)` in a code comment (the no-issue-citations rule is scoped to docs and the file has eighteen such citations already); `fix(tests)` versus `test:` (both permitted, precedent on `main` for each). Nothing under removed behaviour, cross-file, ops/security, reuse, simplification, altitude, or scope. Not covered by the pass: a Linux runner; sibling PRs were grepped for the named symbols rather than read whole.
- Docs-drift pass (fresh subagent, `.agents/skills/review-docs-drift/SKILL.md`). No findings, blocking or advisory. Searched every tracked `*.md`/`*.mdx`/`*.txt`/`*.rst` and then all tracked files for the test module name, the test name, the drain constants, `_drain_request_body`, `#1740`, and flake or quarantine lists; the one document that names the module (`docs/designs/design_537148738.md:130`) only invokes it. The changed file is neither a generated-region source nor an identifier source in `docs/README.md`; the map-staleness window is empty; `make docs-check` passed all six targets at HEAD. Prettier and the site build do not apply (no `.md`/`.yaml`/`docs/site` change). Its one non-finding observation, that the comment cites an issue number and a measured value inline, is the same point the adversarial pass classed as not a finding; left as is.

## Risk & Rollout

Low risk: one test file, one statement reordered, no runtime code. The failure mode it removes is a false red on a loaded runner; the failure mode it keeps is the one the test exists for (a drain that returns before its deadline). Revert is a one-line move back. Nothing has to happen at merge time.

## Handoff for the PR minder

Verified: the mechanism (instrumented pre-fix copy reads short while the server's own drain measures over the deadline, 20/20); the fix's window is a superset by construction and the loops above; the mutation (server 2 ms or 10 ms early fails the fixed test 10/10). Assumed, not measured: the Linux runner's client-side overhead is at or below this machine's 1 to 1.5 ms, so the detection floor there is no worse. Siblings: none found with the same after-dispatch clock start; the four other lower-bound timing tests in the repository already start their clock first. The one local failure in the file (`GitHardeningTest.test_alias_configuration_and_repo_local_execution_are_refused`) is pre-existing on `upstream/main` on this machine only and is not this PR's; if CI shows it red, that is a different problem. No docs reference this test.


jayantid (approved):

agents/platform/scripts/test_credential_proxy.py
```diff
@@ -253,8 +253,17 @@ def test_drain_deadline_releases_a_client_that_stalls_mid_body(self):
                 ("127.0.0.1", self.proxy.server_port),
                 timeout=_STALLED_CLIENT_READ_TIMEOUT_SECONDS,
             ) as client:
-                client.sendall(request)
+                # The clock starts before the request leaves, not after. The
```
jayantid: Reviewed against product requirements, UX flow and technical design at `822c207`; nothing to flag.

Issue #1671, opened 2026-09-17
hostPath is refused only by the webhook: no render-side strip, and the chart ships webhooks off by default


bnaylor:
`spec.deployment.extraVolumes` / `.sidecarVolumes` refuse `hostPath` in the validating webhook and nowhere else, so on an install where the webhook is not running the volume renders into the agent Pod verbatim.

### What is there today

`internal/webhook/platformagent_webhook.go` rejects a `hostPath` volume on both surfaces with `field.Forbidden(... "hostPath volumes are forbidden for security reasons")`. That is the only layer. The controller copies `ExtraVolumes` and `SidecarVolumes` into the Pod spec without inspecting the volume source — `grep -rn HostPath k8s-operator/internal/controller` returns nothing.

### Why that is not enough on a supported install

The webhook is not a given:

- `charts/kube-agents/values.yaml:118` — `operator.webhooks.enabled: false` is the chart default.
- `charts/kube-agents/values.yaml:158` — when they *are* enabled the chart registers them at `failurePolicy: Ignore`, so an unreachable or crash-looping webhook Pod fails open rather than closed.
- `terraform/examples/full-install/variables.tf:496` — `enable_webhooks` defaults to `true` there, which is the configuration where the refusal does hold.

So a chart install at defaults, or any install during a webhook outage, accepts the CR and mounts a host path into the agent's Pod. The agent is the widest-reach workload in the namespace and it is where model output executes, which is what makes this worth a second layer rather than a note.

### What the rest of the file already does

This is the only reservation on these two surfaces with a single layer. The reserved *name* check and the reserved *source* check (gke-labs#1667) are both two-layer: the webhook refusal tells the CR's author why, and a render-side strip is what holds where admission did not run. `hostPath` should be in the same shape — the refusal stays, and the render drops a `hostPath` entry along with any mount naming it. Dropping the volume while a mount still names it produces a Deployment the API server rejects and a reconcile that wedges with nothing in `.status`, so the mount side is not optional; `a2aStripBusTokenVolume` already carries that arm and is the pattern to follow.

### Scope

Pre-existing and independent of the A2A work — it predates the bus entirely and applies to `mode: today` installs too. Noticed while reviewing the volume-source reservation in gke-labs#1667; filed separately rather than widened into that change, whose surface is the bus credential.

### Not verified

[...]

PR #1748, opened 2026-09-18
fix(operator): reserve the bus token by audience and the creds Secret by name, not only by volume name

### LINKED ISSUE #1667: A2A bus token reservation matches volume name, not audience: a renamed projection bypasses it
The reservation that keeps the projected A2A bus token in the `platform-agent` container matches the volume **name**. A CR author who projects the same audience under a different name gets a valid bus token into a sidecar, and the guard does not see it.

Split out of #1653 review (finding A) rather than fixed there: #1653 is a breaking change already in review, and the fix needs its own webhook validation and live validation. #1653 ships with its prose corrected to describe what the guard actually does.

### The hole

`validateReservedVolumeName` / `validateReservedVolumeMounts` (`platformagent_webhook.go:182`) and `a2aStripBusTokenVolume` / `a2aStripBusTokenMounts` (`platformagent_a2a_callout.go:181,204`) all compare `Volume.Name` / `VolumeMount.Name` against `"a2a-bus-token"`. Nothing inspects the volume **source**. `spec.deployment.sidecarVolumes` and `extraVolumes` are raw `[]corev1.Volume` copied verbatim into the pod, and the only source-type check in the webhook is `vol.HostPath != nil`.

```yaml
sidecarVolumes:
  - name: innocuous-cache          # not a reserved name
    projected:
      sources:
        - serviceAccountToken: { audience: a2a-bus, path: token }
sidecars:
  - name: hermes-bridge
    volumeMounts: [{ name: innocuous-cache, mountPath: /var/run/secrets/a2a-bus }]
```

Verified by execution, not by reading: that volume renders intact, and because the auth callout resolves the **pod's** ServiceAccount (`a2a/authcallout/service.go:299`) the sidecar authenticates as `agent`. `TestUserAuthoredContainersCannotMountTheBusToken` still passes, because the pod does carry exactly one volume named `a2a-bus-token` and only `platform-agent` mounts that name.

A sidecar that also holds `bridge-password` then holds the union of both grant sets — the retired `worker` credential rebuilt out of a volumeMount, which is the thing A5 exists to prevent.

### The second route, which is cheaper

The same surface can mount the a2a creds Secret directly and read `bridge-password`. No token minting required, and it predates A5. **This is why the audience check alone is not enough:** shipping it on its own narrows the expensive route and leaves the cheap one, which in practice advertises the cheap one to anyone reading the diff.

### What this is, and is not

It is a **configuration hazard, not a privilege escalation**, and the fix should be written in those terms. Checked rather than assumed:

- `buildPlatformLocalRole` (`platformagent_manifests.go:4145`) grants the agent ServiceAccount `get, list, watch` on `platformagents` and `platformagents/status` and nothing else. The rendered platform ClusterRole is read-only throughout.
- `role.yaml` and `charts/.../operator-rbac.yaml` do carry `patch`/`update` on `platformagents`, but those are the **operator's** role — it patches status. Not the agent's.
- No editor/viewer role and no aggregated role ships, and nothing indicates the PlatformAgent CR lives in a customer GitOps repo.

So the actor is the platform administrator, who already chooses the sidecar image and its env. The repo already has a settled position on exactly this class, four functions from the code in question — `agentForbiddenVolumeNames` (`credential_proxy_manifests.go:472`):

> The CR is authored by an operator rather than by the agent, so this is a configuration hazard and not an escape; it is guarded because nothing else would notice.

That is the right framing and the right reason to fix it anyway.

**Falsifier, so this is re-decidable rather than settled by assertion:** if PlatformAgent CR authorship ever becomes delegable — a tenant-facing role, an aggregated role, or the CR moving into a repo the agent can open PRs against — the actor changes and so does the severity.

### Proposed fix

- **A1** — refuse and strip any user-authored volume whose `Projected.Sources[].ServiceAccountToken.Audience` is the bus audience, regardless of volume name.
- **A3** — refuse user volumes naming the a2a creds Secret. Precedent: `agentForbiddenVolumeNames`.

Both, together, in one PR. Two-layer as the existing controls are (webhook refuses and says why; render strips, because the chart's default `failurePolicy` is `Ignore`).

Explicitly **not** doing:

- **A2** — refuse all `serviceAccountToken` projections. Removes a real capability.
- **A4** — split the bus identity into its own pod. The only actual boundary, but it buys a boundary against an actor who is not in this threat model, at high cost. Revisit if the falsifier above is met.

### The claim the fix must not make

KSA tokens are pod-scoped and the callout cannot tell which container presented one. **A1 and A3 are a guard against misconfiguration, not a boundary against a hostile sidecar**, and the new PR should say so plainly — reading the guard as stronger than it is, is the exact defect being fixed here.

bnaylor:
## Summary

The reservation that keeps the projected A2A bus token in the `platform-agent` container is now keyed on what a user volume *is*, not only on what it is called.  A `spec.deployment.sidecarVolumes` or `extraVolumes` entry that projects a `serviceAccountToken` for the bus audience under any name, or that mounts one of the three Secrets the operator renders with bus credentials in them, is refused by the webhook at the field that matched, and left out of the Pod at render together with every mount naming it.  The operator's own `a2a-bus-token` projection is untouched.  This is a guard against a misconfiguration by the CR's author, not a boundary against a hostile sidecar: KSA tokens are pod-scoped and the callout cannot tell which container presented one.  Reading the guard as stronger than that was the defect.

## Why This Change

#1653 shipped the reservation comparing `Volume.Name` / `VolumeMount.Name` against `a2a-bus-token`.  Nothing looked at the source.  So this CR renders intact on `main`, and because the auth callout resolves the pod's ServiceAccount (`a2a/authcallout/service.go`, `Validate` then `Lookup(att.ServiceAccount)`; the request carries a token and nothing that names a container), the sidecar authenticates as `agent`:

```yaml
sidecarVolumes:
  - name: innocuous-cache
    projected:
      sources:
        - serviceAccountToken: { audience: a2a-bus, path: token }
sidecars:
  - name: hermes-bridge
    volumeMounts: [{ name: innocuous-cache, mountPath: /var/run/secrets/a2a-bus }]
```

The cheaper route needs no token at all: mount `<agent>-a2a-nats-creds` and read `bridge-password`, a static bus user.  A sidecar holding both holds the union of the two grant sets, which is the retired `worker` credential rebuilt out of a volumeMount.  Closing the audience route alone would narrow the expensive path and advertise the cheap one to anyone reading the diff, which is why the issue asks for both in one PR.

Who can do this is the platform administrator, who already chooses the sidecar image.  `buildPlatformLocalRole` grants the agent `get/list/watch` on its own CR and nothing more, and no tenant-facing or aggregated role on `platformagents` ships.  The repo's settled position on this class is `agentForbiddenVolumeNames`: a configuration hazard rather than an escape, guarded because nothing else would notice.  The falsifier is in the comments: if CR authorship becomes delegable, the actor changes and so does the severity.

Live, on the shared install (webhook at `failurePolicy: Fail`, operator `dev-pr724b`), a server-side dry run of exactly that patch was admitted with exit 0 while the `hostPath` control on the same field was refused.  Transcript under Live validation.

## What Changed

- `api/v1alpha1/common_types.go`: `A2ABusTokenAudience`, the bus object names (`A2ANATSName`, `A2ACredsSecretName`, `A2ANATSConfigSecretName`, `A2ACalloutName`, `A2ACalloutKeysSecretName`, and `A2ACredentialSecretNames` as the set), and `BusCredentialRoutes(volume, agentName)`, which returns the routes a volume carries: a projected `serviceAccountToken` with the bus audience (at any source index), or one of those Secrets as a `secret` volume or a projected `secret` source.  The api package is where the webhook and the controller already share `ReservedVolumeNames` and `SensitiveEnvVars`, and it is the only package both can import.  The controller's `a2aBusTokenAudience`, `a2aNATSName`, `a2aCredsSecretName`, `a2aCalloutName`, `a2aCalloutKeysName` and the four `-config` literals now read from it, so each fact has one spelling.
- The Secret set is three, not one, and that is wider than the issue's A3 wrote it.  The same credentials sit in three operator-rendered Secrets: `<name>-a2a-nats-creds` (the static passwords, `bridge-password` among them), `<name>-a2a-nats-config` (`nats.conf` with every one of those passwords inline), and `<name>-a2a-callout-keys` (the issuer seed that signs every identity the bus accepts, which is worth more than any password in the other two).  Refusing one of the three by name and calling the route closed would be the "narrow the expensive route and advertise the cheap one" defect again, one Secret over; the adversarial pass raised the keys Secret and the config Secret followed from reading `buildA2ANATSConfigSecret`.  Drop the two extra names if you want the PR held to the issue's letter; it is one slice literal.
- Volume shapes only.  `env[].valueFrom.secretKeyRef` and `envFrom[].secretRef` on a sidecar deliver the same bytes and are not checked, and both review passes found that (the adversarial one by execution).  Not widened here, deliberately: the bridge sidecar's documented configuration (`a2a/docs/hermes-bridge.md`) is a `secretKeyRef` to `bridge-password` on the creds Secret, so a refusal on env refuses the supported bridge, and which keys a sidecar may read by env is a policy the issue did not set (its precedent, `agentForbiddenVolumeNames`, is volume-keyed too).  The comment on `BusCredentialRoutes`, the webhook helper and the CRD page say so in as many words.  The decision owed is whether `envFrom.secretRef` naming any of the three, and `secretKeyRef` to keys other than `bridge-password`, should be refused as a follow-up; that is bnaylor's call and is not made here.  The other volume sources that name a Secret (`csi.nodePublishSecretRef`, the storage drivers' `secretRef`) hand it to a node plugin rather than into the container and are not routes to the bytes.
- `internal/webhook/platformagent_webhook.go`: `validateBusCredentialSource` runs beside `validateReservedVolumeName` on both lists.  The error lands on `…projected.sources[i].serviceAccountToken.audience`, `…secret.secretName` or `…projected.sources[i].secret.name`, and the message names the audience or the Secret.  Not gated on `spec.mode`, like the name check.
- `internal/controller/platformagent_a2a_callout.go`: `a2aBusCredentialVolumeNames`, `a2aStripBusCredentialSources`, `a2aStripVolumeMountsNamed`, `a2aStripMountsNamed`, siblings of `a2aStripBusTokenVolume` / `a2aStripBusTokenMounts`.  Called from the same `if a2aAgentSurface(agent)` block in `buildPodTemplateSpec` (sidecars, initContainers, both volume lists) and from `buildBaseContainers` (`extraVolumeMounts`, which reaches the agent and dashboard containers).  Every helper returns its input unchanged when there is nothing to drop, so a clean CR renders the same slices it always did, and containers whose mounts change are copied because the slices are the manager's cached CR.  The operator's own `a2aBusTokenVolumeSource()` is appended to the pod outside the user lists and never passes through the strip.
- The comment above the strip, the webhook helper and `BusCredentialRoutes` each carry the sentence: a guard against misconfiguration, not a boundary against a hostile sidecar, because KSA tokens are pod-scoped and the callout cannot tell which container presented one.
- Docs: a bullet of its own on the CRD reference beside the webhook bullet, saying what the webhook refuses in every mode (name, audience, the three Secrets), that under `next` the render drops the same entries when the webhook did not run, that `env` references are not checked, and what the guard does and does not buy.  The page had no sentence about the reservation before this.  The identifier-sources row for the bus token contract in `docs/README.md` now points at the api package for the audience and the Secret names.
- Not done, on the issue's reasoning: refusing every `serviceAccountToken` projection (A2) removes a real capability, and splitting the bus identity into its own pod (A4) is the only real boundary but buys one against an actor outside this threat model, at high cost.  Revisit if the falsifier above is met.

## Context

Closes #1667.

Related: #1653 (the name-keyed reservation this extends; merged), #1671 / #1675 (`hostPath` render strip and the `VolumesDropped` condition on the same two surfaces; open), #1665 (touches `platformagent_a2a_manifests.go` elsewhere, currently conflicting against `main`).  No open PR fixes #1667; searched open PRs by `audience`, `bus token`, `a2aStripBusTokenVolume`, `validateReservedVolumeName` and `platformagent_webhook.go`.

**Observability, and the #1675 relationship.**  #1675 was open, not merged, when this branched, so this does not depend on it.  The by-name strip this sits beside is silent: `buildPodTemplateSpec` has no logger and no caller of `a2aStripBusTokenVolume` writes a condition, so a CR admitted without the webhook gets a Pod without the entry and nothing in `.status` saying so.  This PR is silent in the same way, deliberately, rather than adding a third mechanism next to the one #1675 is landing.  When #1675 merges, its `VolumesDropped` condition is where this drop should be reported too, with reasons along the lines of `BusTokenSourceDropped` / `BusCredentialSecretDropped`, fed from `a2aBusCredentialVolumeNames`.  Textually: #1675's `buildPodTemplateSpec` hunk already fails `git apply --check` against `upstream/main` (it predates #1653's strip block), and against this branch its CRD-page hunk fails as well.  Whichever lands second rebases.  The helper that should survive the merge-up is #1675's generic pair (`stripVolumeMountsNamed` / `stripContainerMountsNamed`, which take a name set); `a2aStripVolumeMountsNamed` / `a2aStripMountsNamed` here have the same shape and collapse onto them, with `a2aBusCredentialVolumeNames` feeding the set.

**Handoff for the PR minder.**  Verified: the hole on `upstream/main` `965b6917` by unit test (webhook and render) and by envtest against a real API server with no webhook; the fix by the same three on this branch; the operator's own projection surviving, on the struct and on the API server; `go test` over the whole module with envtest assets (controller package 88s); `make docs-check` and prettier 3.9.6 clean; the live before-observation on `bnaylor-ka-test`.  Verified by the adversarial pass, not fixed: the env route (`secretKeyRef`/`envFrom`) to the same Secrets renders and is admitted; the decision on it is bnaylor's and is written up under What Changed.  Not verified live: the new webhook refusal and the render strip, because the shared install's operator predates #1653 and has no webhook-only mode, so showing layer 1 means rolling a newer operator onto an install weeks behind `main`, which reconciles the live CR against stale RBAC/CRDs.  The issue's "the sidecar authenticates as `agent`" is the issue's measurement, not repeated here: the authcallout test harness stubs TokenReview, so a reproduction there would exercise the stub.  Known CI: `make test`'s `test-python` step dies inside `agents/platform/scripts/test_sandbox_artifact_patch.py` with no unittest summary on this branch and nothing here touches Python; the Go step it sits in front of was run directly and is green.  No issue was commented on or closed from this branch.

## Testing

- `go build ./... && go vet ./...` clean.  `gofmt -l` reports only `platformagent_litellm_policy_test.go`, pre-existing on `main`.
- `go test $(go list ./... | grep -v /e2e) -count=1` with `KUBEBUILDER_ASSETS` exported from `bin/setup-envtest use 1.36.0` on `0bddcf2e`: every package `ok`; `internal/controller` 88.4s, which is the envtest-ran figure.  On the head `5b2f7bc1` (the review-fix commit, which touches the api, controller and webhook packages only): `go vet ./...` clean, `api/v1alpha1`, `internal/webhook` and the full `internal/controller` package re-run with the assets set, all `ok`, controller 89.7s with the four envtest cases run and no SKIP.  `make -C k8s-operator test` runs `manifests generate fmt vet` first and they changed nothing tracked (the new api struct is marked `+kubebuilder:object:generate=false`); its `test-python` step fails before the Go step for the pre-existing reason above.
- New tests, run against `upstream/main` `965b6917` with the test files copied into a detached worktree, then on this branch:

  | Test                                                                              | `main`                                                                                                                  | branch |
  | --------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- | ------ |
  | `TestBusCredentialSourcesAreRefusedOnBothVolumeLists` (webhook, 12 refused shapes) | FAIL: all admitted (renamed projection, projection as second source, creds Secret volume, projected creds source, on both lists; the nats-config and callout-keys Secret rows were added after the red run and were not run against `main`) | PASS   |
  | same test, 10 accepted shapes (other audience, empty audience, other Secret, another agent's creds Secret, projected other Secret) | PASS                                                                                               | PASS   |
  | `TestUserAuthoredVolumesCannotCarryTheBusCredentialBySource` (render, fake CR)    | FAIL: all three stolen volumes in the Pod, mounted by `peek`, `platform-agent`, `platform-agent-dashboard`, `hermes-bridge` | PASS   |
  | `TestTheNameHalfAndTheSourceHalfAreTwoDifferentChecks` (helpers)                  | n/a (names the new helpers)                                                                                             | PASS   |
  | `TestBusCredentialRoutes` (api, 13 shapes incl. the three Secrets, another agent's creds Secret, the reserved name with an innocent source) | n/a                                                                                           | PASS   |
  | `TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest`         | FAIL (below)                                                                                                            | PASS   |
  | `TestUserAuthoredContainersCannotMountTheBusTokenByName` and the two beside it (#1653) | PASS                                                                                                                | PASS   |

  The render test also pins: the operator's own `a2a-bus-token` is still the single `a2a-bus` projection and is mounted by `platform-agent` alone; a projection for audience `vault` and a `secret` volume for `my-tls` survive on every container the author sent them to; the CR's own slices are not edited; under `mode: today` nothing is dropped.  The webhook test also checks the message names `"a2a-bus"` or the Secret and does not read as the name reservation.  The test fixtures spell the audience and the Secret name as literals rather than the constants, so the files compile against a tree without the fix and fail there on behaviour.

- Envtest, a real API server serving the operator's CRDs and no admission webhook (the chart-default install): a `mode: next` CR with the renamed projection and a creds-Secret volume on a sidecar, reconciled, gateway Deployment read back from the server.  Confirmed the case RAN (no SKIP).

  Against `upstream/main` `965b6917`:

  ```
  === RUN   TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest
      platformagent_a2a_bus_source_envtest_test.go:112: the accepted Deployment carries volume "innocuous-cache" projecting audience "a2a-bus"
      platformagent_a2a_bus_source_envtest_test.go:114: the accepted Deployment carries volume "innocuous-config" mounting Secret "test-agent-a2a-nats-creds"
      platformagent_a2a_bus_source_envtest_test.go:119: containers [hermes-bridge] in the accepted Deployment mount "innocuous-cache"
      platformagent_a2a_bus_source_envtest_test.go:119: containers [hermes-bridge] in the accepted Deployment mount "innocuous-config"
      platformagent_a2a_bus_source_envtest_test.go:158: volumes=19 bus-credential user volumes in Deployment=2 operator's a2a-bus-token mounted by [platform-agent]
  --- FAIL: TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest (4.96s)
  ```

  On this branch:

  ```
  === RUN   TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest
      platformagent_a2a_bus_source_envtest_test.go:158: volumes=17 bus-credential user volumes in Deployment=0 operator's a2a-bus-token mounted by [platform-agent]
  --- PASS: TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest (8.15s)
  ```

  Two fewer volumes, the operator's projection still on `platform-agent`, the sidecar's `vault` projection still mounted, no mount naming an undeclared volume.

- `make docs-check` clean (links, terminology, map, site audience, context budget).  `prettier --check` 3.9.6 on the changed page: clean.
- Drift: `git rev-list --count HEAD..upstream/main` = 0 at push time.

### Live validation

Partly, and only the before half.  Install: `bnaylor-ka-test` (`gke_bnaylor-kagents-dev_us-east4_bnaylor-ka-test`, ns `kubeagents-system`, operator `k8s-operator:dev-pr724b`, default mode, webhook `kubeagents-validating-webhook-configuration` at `failurePolicy: Fail`).  That operator predates #1653 and this branch, so what its webhook shows is the hole, by server-side dry run, which runs admission and persists nothing:

```
$ kubectl -n kubeagents-system patch platformagent platform-agent --type=merge --dry-run=server \
    -p '{"spec":{"deployment":{"sidecarVolumes":[
          {"name":"innocuous-cache","projected":{"sources":[{"serviceAccountToken":{"audience":"a2a-bus","path":"token"}}]}},
          {"name":"innocuous-config","secret":{"secretName":"platform-agent-a2a-nats-creds"}}]}}}' \
    -o jsonpath='{.spec.deployment.sidecarVolumes}'
[{"name":"innocuous-cache","projected":{..."audience":"a2a-bus"...}},{"name":"innocuous-config","secret":{"secretName":"platform-agent-a2a-nats-creds"}}]
exit=0

$ kubectl -n kubeagents-system get platformagent platform-agent -o jsonpath={.spec.deployment.sidecarVolumes}
(empty; unchanged)

$ kubectl ... patch ... --dry-run=server -p '{"spec":{"deployment":{"sidecarVolumes":[{"name":"host-root","hostPath":{"path":"/"}}]}}}'
The PlatformAgent "platform-agent" is invalid: spec.deployment.sidecarVolumes[0].hostPath: Forbidden: hostPath volumes are forbidden for security reasons
exit=1
```

The `hostPath` control shows the webhook was consulted and does refuse on that field; the two shapes this PR refuses went through.

Not live-tested: the new refusal (layer 1) and the strip (layer 2).  Layer 1 needs the new webhook running, and the operator has no webhook-only mode, so that means rolling a newer operator onto a shared install weeks behind `main`, which reconciles the live CR against stale RBAC/CRDs and rolls the gateway.  Layer 2 needs a CR the webhook did not see, which on this install means disabling a live control.  `a2a-next-dev`, the only `mode: next` install, is read-only for this session.  The envtest case above is the webhook-absent scenario against a real API server, run against both `main` (leaks) and this branch (strips), and stands in for both.  No cluster state was changed and no lease was taken; nothing above mutates.

## Self-Review

Both passes ran in fresh `general-purpose` subagents handed only the worktree path, the range `upstream/main...HEAD` and the skill file; neither had this session's reasoning.  Both reproduced the red run against `upstream/main` on their own (detached `/tmp` worktree) rather than taking the table above on trust.

What the passes looked for: every consumer of the five user-authored volume/mount surfaces; every other CR field that can carry a Secret or a token; whether the operator's own projection can be reached by the strip; mutation of the cached CR; the dangling-mount rule against the API server; the mode gate on each half; the no-magic-constants rule; controller-gen drift from the new api type; whether the webhook message leaks the dark stack on a `today` install; the two open PRs on the same files; every sentence of prose the diff touches or makes stale, the docs map, generated regions, anchors, `make docs-check`, prettier; and whether any doc claims an observability for the strip the code does not have.

- **Fixed** (adversarial should-fix, docs-drift blocking): the CRD page hung `spec.mode: next` on the webhook half; the webhook refuses in every mode and only the render is gated (the adversarial pass executed `ValidateCreate` on a `today` CR and got the refusal).  Rewritten as its own bullet with the qualifier on the render clause, and the "to a container other than `platform-agent`" clause dropped, since the volume is refused whether or not anything mounts it.
- **Fixed** (docs-drift blocking): the `ReservedVolumeNames` paragraph in `common_types.go` still said the audience route "is admitted".  Rewritten as the history of the defect, pointing at `BusCredentialRoutes` as the source half.
- **Fixed** (adversarial should-fix, or file it): `<name>-a2a-callout-keys` is the same route shape with a larger prize (the issuer seed) and nothing reserved it; reading `buildA2ANATSConfigSecret` for the same question turned up `<name>-a2a-nats-config` with every password inline.  Both added to the set; see What Changed for the argument and the one-line way to hold the PR to the issue's letter.
- **Fixed** (docs-drift should-fix): the `docs/README.md` identifier-sources cell for the bus token contract named only the callout file; it now names the api package for the audience and Secret names.  One cell, no re-alignment.
- **Fixed** (docs-drift nit): the CRD bullet names the three Secrets so a reader can act on it.
- **Not fixed, deliberately, decision raised** (adversarial should-fix HIGH; docs-drift should-fix): the env route.  Verified by the adversarial pass on this branch: `env.valueFrom.secretKeyRef{name: <agent>-a2a-nats-creds, key: bridge-password}` and `envFrom.secretRef` on a sidecar, an init container and `spec.deployment.env` all render, and `ValidateCreate` admits them.  Not widened here, for the reason in What Changed: the documented bridge configuration is exactly that `secretKeyRef`, so an env refusal is a policy on which keys a sidecar may read, which the issue did not set and this PR should not set on its own.  The comment that claimed volumes were "the two shapes" was wrong and now says volume shapes only, why env is left alone, and that the decision is owed.  The bridge sidecar holding `bridge-password` is also not the union the issue names: what must not reach it is the token, and A1 is what keeps that away.
- **Not fixed, deliberately** (adversarial should-fix): #1675 carries a byte-identical `stripVolumeMountsNamed` and an equivalent container-mount strip, threads its dropped set through `buildBaseContainers` where this recomputes a pure function of the CR, and reports its drop as a condition where this is silent.  #1675 is open, lgtm'd and already failing `git apply --check` against `upstream/main`; this fix should not wait on it, and rebasing onto an unmerged branch moves the merge-up around.  Context says which helper survives (#1675's), that the recomputation collapses into its threaded set, and that the bus-credential drop belongs in `VolumesDropped` once it exists.  The silent choice is not an argument against #1675's rationale ("a dropped mount is a CR whose author believes it took effect" is right); it is the choice not to land a second reporting mechanism beside the one about to merge.
- **Not fixed** (adversarial, for the record): the mount strip keys on volume name, so a CR declaring two volumes under one name loses both sets of mounts; that CR was a Deployment the API server refused before this change.
- **Steps not run**: docs-drift could not run `npm run build` for the site (no `node_modules`; the hunk is prose inside a list).  Adversarial did not read the bodies of the other open PRs touching these paths beyond #1675 and #1665, did not run `make test` (per instruction; the Go step was run separately and is green), and did not re-verify the issue's callout claim against a live bus.  Neither pass saw the PR body (it did not exist yet).

## Risk & Rollout

Runtime change to the gateway Pod render under `mode: next` only, and only for a CR that carries one of the two shapes; on a webhook-enabled install such a CR cannot exist, and on a webhook-off install it is the bug.  Every helper returns its input unchanged when nothing matches, and the full operator suite including the byte-identity assertions on a default install is green.  A `mode: today` CR is untouched at render, by the same gate as the name strip.  A CR that does carry one gets a Pod without it, which is one template roll on the first reconcile by an operator carrying this, for that CR only; a sidecar that relied on the volume starts without it, which is the point.  The webhook refusal is not mode-gated, so a `today` CR naming `<agent>-a2a-nats-creds` or projecting `a2a-bus` is refused at admission from this operator on; neither exists on a `today` install.  New failure mode: none identified; the mount strip is what keeps the Deployment acceptable to the API server.  No CRD schema, RBAC or chart change.  Revert is a clean revert of the commit.


bnaylor:
Both red checks fixed.  Head is `29ad8868`.  3 files, +46/-4.

### `Go Security Scan (gosec)` -- G101 false positive

`busTokenAudienceForbiddenFmt` is an error-message format string, so gosec's "potential hardcoded credentials" is wrong about it.  Suppressed with the trailer form already used tree-wide (25 occurrences -- "Volume name, not a credential", "Mount path, not a credential", and so on), copied from `a2aBusTokenVolume` in `platformagent_a2a_callout.go:117`:

```go
busTokenAudienceForbiddenFmt = "volume %q projects a serviceAccountToken for audience %q, ..." // #nosec G101 -- Error message format, not a credential
```

No file-level exclusion and no change to the scanner config.  Reproduced CI's invocation exactly (`.github/workflows/security-scan.yml:47`, `working-directory: k8s-operator`): before, `Issues : 1` and exit 1; after, `Files : 44 / Lines : 29913 / Nosec : 31 / Issues : 0`, exit 0.

### `Security invariants` -- C1 was comparing nothing

Moving the audience constant into `api/v1alpha1/common_types.go` left C1's extractor pointed at its old home, so `one()` matched zero and failed with "this test compared nothing".  That is the guard working, not a flake.

Repaired by adding the API package as a registered `Source` and reading the literal from where it now lives.  The comparison is still between two *independently declared* literals in two different modules -- `api/v1alpha1/common_types.go` against `a2a/lib/credentials.go` -- so it is not a tautology.  Deliberately **not** repaired by matching `a2aBusTokenAudience = agentv1alpha1.A2ABusTokenAudience`, which would have made C1 compare the constant to itself and pass regardless of either value.

The Source anchor is the assignment, `'A2ABusTokenAudience = "'`, not the bare name -- the bare name also appears in that file's doc comment and inside `BusCredentialRoutes`, so it would survive the declaration being renamed away.  (Measured; see mutation 2.)

**One addition beyond the two-line repair,** because the move opened a hole that wasn't there before.  With the literal now read from the API package, the controller could re-spell `a2aBusTokenAudience` as its own string and the projection would mint under *that*, while C1 happily compared two literals it no longer reaches.  The existing `a2aBusTokenVolumeSource` check proves the projection uses the controller constant, not what that constant equals.  So there is now a link assertion, itself routed through `one()` so a rename still reads as "compared nothing" rather than passing.  Happy to drop it if reviewers would rather keep the repair minimal.

### Mutations

| mutation | result |
| --- | --- |
| wrong literal in `common_types.go` (`"a2a-bus-wrong"`) | fails on **mismatch**: `'a2a-bus-wrong' != 'a2a-bus'` -- not on "compared nothing" |
| rename the declaration to `A2ABusAudienceRenamed` | `one()` still guards: `0 != 1 : A2ABusTokenAudience is not a single string constant (0 matches)`, and the harness self-check reddens on the missing anchor |
| controller re-spells the audience as its own literal | the new link check fires: `'"a2a-bus"' != 'agentv1alpha1.A2ABusTokenAudience'` |

Each restored with `git checkout --`; `git status --porcelain` clean after each.

### Verification

```
$ python3 tests/conformance/run.py
Ran 122 tests in 0.869s
OK (expected failures=12)
```
Before the fix: `FAILED (failures=1, expected failures=12)`.

`go test -count=1 ./...` green across the operator module (`api/v1alpha1`, `cmd/drift-detector`, `cmd/k8s-event-watcher`, `internal/controller` 86.9s, `internal/testing`, `internal/webhook`).  `ruff check --isolated` clean on `tests/conformance/`.  `gofmt -l` reports only `platformagent_litellm_policy_test.go`, which is #1545's.

**Note for whoever reads the Source registry:** `test_D_accountability.py:139` already reads `common_types.go` by raw `h.REPO_ROOT / ...` path, outside the registry.  I followed the registry convention here rather than that one, but the two ways of reaching the same file is a small inconsistency worth someone's attention separately.

kyber775 (commented):
## Code Review: PR #1748 — fix(operator): reserve the bus token by audience and the creds Secret by name, not only by volume name

Comprehensive, defense-in-depth closure of #1667. By keying the bus credential reservation on what a user volume *is* (source: token projection for `a2a-bus` audience or mounting one of the three operator-rendered bus Secrets) rather than solely on what it is called (`a2a-bus-token`), this permanently closes the sidecar identity elevation path left open in #1653.

### Strengths & Core Mechanics

- **Two-Layer Enforcement (Admission + Render Strip):**
  - **Admission Webhook:** `validateBusCredentialSource` catches both projected token sources for `a2a-bus` and bus Secrets (`creds`, `config`, `callout-keys`) across `extraVolumes` and `sidecarVolumes`. Refusals pinpoint the exact matched subfield (`.projected.sources[i].serviceAccountToken.audience` or `.secret.secretName`).
  - **Controller Render Strip:** When admitted without webhook validation (chart default `failurePolicy: Ignore` or webhook-absent deployments), `a2aStripBusCredentialSources` drops the forbidden volumes and `a2aStripMountsNamed` / `a2aStripVolumeMountsNamed` symmetrically strip all referencing mounts across sidecars, initContainers, and `extraVolumeMounts`. This ensures the rendered Deployment remains valid and does not wedge reconciliations.
- **Symmetric Secret Coverage:**
  - Widens the reservation beyond `<name>-a2a-nats-creds` to include `<name>-a2a-nats-config` (inline passwords) and `<name>-a2a-callout-keys` (issuer seed). Avoids the asymmetric vulnerability where closing one credential route advertises a cheaper or more sensitive alternative.
- **Structural Invariant & Conformance Hardening:**
  - Commits `1f0507dc` and `29ad8868` cleanly repair the C1 cross-module invariant extraction in `tests/conformance/test_C_enforcement.py` and `_harness.py`.
  - The C1 link assertion (`assertEqual(controller_audience, "agentv1alpha1.A2ABusTokenAudience")`) ensures the controller cannot silently decouple from the shared API constant.
  - G101 false positive on `busTokenAudienceForbiddenFmt` is properly annotated with `#nosec`.
- **Empirical Validation:**
  - Real envtest against an API server (`TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest`) reproduces the leak on `main` and verifies the render strip on this branch.
  - Conformance test suite (`python3 tests/conformance/run.py`): 122 tests pass cleanly (12 expected failures, 0 unexpected).
  - Go test suite across `k8s-operator/api/v1alpha1`, `k8s-operator/internal/webhook`, and `k8s-operator/internal/controller`: green.

### Architectural Observations (Non-blocking)

1. **Sidecar `env` / `envFrom` Exposure:**
   As explicitly detailed in the PR description and code comments on `BusCredentialRoutes`, environment variables referencing `<name>-a2a-nats-creds` (e.g. `bridge-password`) via `secretKeyRef` remain unrefused to support the documented `hermes-bridge` sidecar configuration. While not a token route, tracking whether `envFrom` or non-bridge keys should be restricted remains an open follow-up policy decision.
2. **Observability on Stripped Volumes:**
   The render strip operates silently to match existing `a2aStripBusTokenVolume` conventions. Merging with #1675 in the future will provide the natural `VolumesDropped` status condition reporting.

---

### Verdict: [PASS]

High-signal, rigorous implementation backed by real envtest and conformance test execution. Ready to merge once approved by OWNERS.

bnaylor:
Thanks [@kyber775](https://github.com/kyber775).  Both non-blocking observations already have a
home, so noting where rather than folding anything into this PR:

1.  **`env` / `envFrom`.**  Filed as gke-labs#1749, open.  It is a policy call rather than a
    gap this PR can close on its own: the `bridge-password` `secretKeyRef` is how the documented
    `hermes-bridge` sidecar is configured, so refusing the key outright breaks a shipped
    configuration.  Waiting on [@bnaylor](https://github.com/bnaylor) for the direction.
2.  **Silent strip.**  Right, and that is #1675's condition.  It writes `VolumesDropped` naming
    the field path of every entry the render left out, and it is deliberately general over the
    strips rather than hostPath-specific, so whichever of the two lands second wires this one
    into it.  #1675 carries `lgtm` and is the likelier to land first, in which case the wiring
    is a follow-up on this branch.

No code change here.  Nothing unresolved on the PR.

bnaylor:
Correcting point 2 above before anyone plans around it: I overstated how much of #1675 is
reusable.  The condition **type** is general (`VolumesDropped`, deliberately not
`HostPathVolumesDropped`), but the rest of it is not.  The reason is `HostPathVolumeDropped`,
the per-entry format is `"%s[%d] %q (hostPath %s)"`, and `hostPathDroppedMessage` builds from
`hostPathVolumes(agent)` alone.  Carrying this PR's strips would mean a second reason and a
second entry shape under the same type, not just calling the existing helper.

Still the right place for it, and still a follow-up rather than something either PR should grow.
Just more than a wiring change.

kyber775 (approved):
## Re-Review: PR #1748 — fix(operator): reserve the bus token by audience and the creds Secret by name, not only by volume name

**Verdict:** `[APPROVE]`
**Reviewed Commits:** Delta verified on HEAD `29ad8868b62226f481877a8f1781001aa623c5bc`

/lgtm

---

### Resolution of Review Observations

Thank you for confirming the triage paths for the non-blocking architectural items:
1. **Sidecar `env` / `envFrom` Exposure:** Tracked as open policy decision in #1749 to evaluate `secretKeyRef` / `envFrom` rules without breaking the documented `hermes-bridge` sidecar configuration.
2. **Observability on Stripped Volumes:** Tracked as follow-up to integrate with #1675's `VolumesDropped` status condition once landed.

---

### Local Test & Invariant Verification

- **Mergeability:** Clean against `upstream/main` (`git merge-tree` exit 0).
- **Security & Conformance Suite:**
  - `python3 tests/conformance/run.py`: 122 tests pass cleanly (12 expected failures, 0 unexpected). C1 cross-module literal and reference extraction passes.
- **Operator Tests:**
  - `go test ./api/v1alpha1/... ./internal/webhook/...`: Green.
  - `go test -v -run "TestBusCredential|TestUserAuthored|TestABusCredential" ./internal/controller/...`: All pass clean.
  - Envtest (`KUBEBUILDER_ASSETS` set): `TestABusCredentialSourceStaysOutOfTheDeploymentWithoutTheWebhookEnvtest` ran and passed against real API server / etcd (7.7s).
- **Documentation & Formatting:**
  - `make docs-check`: All 313 Markdown files and 1301 code files clean.
  - Context budget and links verified.

The two-layer enforcement (webhook admission refusal + controller render-side strip) cleanly shuts down the token projection and bus Secrets sidecar privilege-escalation routes. Ready to merge once approved by OWNERS.

toshiowang (approved):

PR #1099, opened 2026-09-01
feat(evals): introduce dynamic evaluation harness, benchmark scenario, and GCP sandbox

fkc1e100:
# [WIP / Draft] feat(evals): [Research] introduce dynamic evaluation harness, benchmark scenario, and GCP sandbox

Closes #1098

> [!WARNING]
> **Status: Untested Draft / Work In Progress (WIP)**  
> This pull request is an early research prototype and proposal. **It has not yet been end-to-end tested against live production/staging Kubernetes clusters or active `kube-agents` runtime controllers.** It is opened strictly as a draft for architectural feedback, scenario schema discussion, and community input.

---

## Overview

This PR introduces a modular dynamic evaluation testbed under `evals/` as part of our research into execution-based benchmarking for `kube-agents`. 

Unlike static prompt-response tests, this harness allows contributors and automated CI pipelines to execute deterministic fault scenarios against a Kubernetes cluster (or in simulated mock mode), measure agent diagnostic and remediation trajectories, and compute quantitative reliability and safety metrics.

## Changes Included

- **Scenario Specification Schema (`evals/scenarios/`):**
  - Added `evals/scenarios/online-boutique-oom-crash.yaml` defining a Tier-1 container memory constraint fault on the Online Boutique microservice demo.
- **Chaos Injection Module (`evals/chaos_injector.py`):**
  - Implemented programmatic fault injection via the official Kubernetes Python API (Strategic Merge Patching).
  - Added `--mock` flag to allow offline testing and development without live cluster credentials.
- **Agent Tracing & Mock (`evals/mock_agent.py`):**
  - Added reference agent simulation executing MCP tool invocations (`list_pods`, `get_pod_events`) and generating declarative GitOps PR patches.
- **Master Evaluation Runner (`evals/eval_runner.py`):**
  - Added CLI test runner that executes the end-to-end evaluation lifecycle.
  - Implemented automated calculation for:
    - **Diagnostic F1-Score (`F1_diag`):** Harmonic mean of Diagnostic Precision and Recall (`2 × (P × R) / (P + R)`) against ground-truth metadata without hallucinated causes.
    - **Mitigation Success Rate (`M_SR`):** Service recovery verification against scenario criteria.
    - **Action Safety Index (`ASI`):** Forbidden mutation penalty tracking.
    - **Context & Token Efficiency (`C_EF`):** Token usage per effective action.
- **GCP Ephemeral Research Sandbox (`evals/sandbox/gcp/`):**
[...]

PR #1760, opened 2026-09-18
feat(release): gate the staging promotion on the release-candidate eval

### LINKED ISSUE #1766: bench-gate: give "evaluated nothing" its own exit code so an all-infrastructure night is NOT RUN, not RED
## What

`bench-gate suite` exits non-zero both when it graded a catalog and found it red, and when every
repetition of every case died on infrastructure and it graded nothing. Give the second case its own
exit code, so a caller can tell a verdict from a lane failure without parsing prose.

## Why

`hack/ci-eval-rc.sh` turns the eval's exit status into one of three words — GREEN, RED, NOT RUN —
and `scripts/release/poll_rc_eval_verdict.py` reads that word to decide whether the nightly's
candidate reaches staging. The two words differ in a way that is not cosmetic:

- **RED is settled.** The `evalcand_` tag stays on the commit, `resolve_promotion_candidate.sh`
  skips a commit carrying one, and that candidate is never measured again.
- **NOT RUN withdraws the nomination.** The tag is deleted and a later nightly asks about the same
  commit again.

So a run that measured nothing but is reported RED retires a candidate over a broken lane, and
leaves a rejection in the eval record that was not one. It does not stall the lane — `main` takes
roughly ten commits a day and the next nightly resolves a newer commit — but it costs one candidate
each time and the record it leaves is wrong.

#1760 closes most of this from the outside. It reads the presence of `eval-verdict.md`, which
`bench-gate suite --markdown-out` writes in its roll-up: absent after a non-zero eval means the run
stopped in preflight before grading anything, which is NOT RUN. That covers the preflight refusals
— a ledger token that would not mint, a runner image short of `uv`, an `EVAL_REPETITIONS` that is
not a positive integer.

What it cannot cover is the case where the roll-up **was** reached: every repetition hit an
infrastructure fault, `bench-gate` wrote its markdown, and exited non-zero. From outside the
process that is indistinguishable from a red catalog, because the artifact exists either way. Only
`bench-gate` knows that no repetition produced a gradeable result.

## What would fix it

A distinct exit code from `bench-gate suite` meaning "evaluated nothing" — every repetition of every
case classified as an infrastructure failure, so there is no catalog verdict to report. The rung-4
collapse logic in `docs/designs/testing-strategy.md` §4.2 already computes something close to this:
a suite where every case hit an infrastructure failure reds rather than reporting green, precisely
because it evaluated nothing. That condition is the one to surface as a status.

`ci-eval-rc.sh` would then map that code to NOT RUN alongside the two cases it already handles, and
the comment in its verdict branch that names this as the leftover case can go.

## Where

- `bench/kube_agents_bench/` — `bench-gate suite`'s exit path
- `hack/ci-eval-rc.sh` — the verdict branch, which names this gap in a comment
- `scripts/release/poll_rc_eval_verdict.py` — the module docstring, which names it too

Both comments should be updated in the same change that closes this, so the three do not drift.

stalhaali:
## Summary

A green nightly E2E matrix no longer promotes a candidate to staging; it nominates one. `nightly-pipeline.yml` now pushes the `evalcand_<ts>_<sha>` tag that starts the release-candidate eval, waits hours for that build's verdict with the poller PR 2 added, and pushes `staging_<ts>_<sha>` — the tag `staging-deploy.yml` reacts to — only when the verdict is GREEN. This is the commit that turns the eval from a report into a gate. It also carries the docs, which until now said in nine places that the lane is advisory.

## Why This Change

The release-candidate eval runs the full catalog against a candidate's published images and takes 1.5 to 3.5 hours. Today it fires on the `staging_` tag, which is also the deploy trigger, so its verdict arrives long after the candidate is already serving on the staging estate. A red eval currently produces a dashboard row and nothing else. Nobody is stopped, and a build the eval rejected stays on staging until the next night happens to replace it.

Putting the eval before the deploy tag is the only ordering that makes it a gate, and that requires a second tag pushed first. This change pushes it.

## What Changed

**`.github/workflows/nightly-pipeline.yml`** gains two jobs between the matrix and the tag push, and renumbers teardown from step 5 to step 7:

- **Step 4 — Nominate Candidate for Eval.** Runs `scripts/release/tag_eval_candidate.sh` to push `evalcand_<core>`, when the matrix was green and `skip_promotion` is false. It stamps `nominated_at` **before** the push and publishes it as a job output.
- **Step 5 — Await Release-Candidate Eval.** Runs `poll_rc_eval_verdict.py` with `--not-before "${NOMINATED_AT:?…}"` and exposes `verdict` and `settled` as job outputs. It reds the pipeline on anything but green, and — because job outputs survive a failed job — step 5b can still read `settled` off it.
- **Step 6 — Create Staging Tag** now sits behind `needs: [step-1, step-5]`. The gate is the implicit `success()` that a plain boolean `if:` leaves in place, and the poller exits 0 only on GREEN. There is no explicit verdict comparison to get wrong: a red, timed-out, cancelled or never-started eval is not `success`, so step 6 is skipped and staging stays on the build it is already running.

`--not-before` is the second half of the retry, and without it the first half is worse than useless. The nomination tag's name is derived from the candidate, so it is the same on every attempt, and a re-nominated candidate has last night's finished build sitting in the archive at the same commit. For the minute or two before Prow uploads tonight's `started.json`, that older build is the newest one matching — so a poller with no floor answers in seconds with the verdict the withdrawal was undoing, withdraws again, and tonight's real eval runs for hours with nobody reading it. The floor is the nomination's own timestamp, with a 15-minute skew allowance for the runner clock against Prow's.

**Step 5b — Withdraw an Unsettled Nomination** is the retry mechanism. It runs under `always()` with explicit `skip_pipeline` / `skip_promotion` guards, when step 4 succeeded, step 5 did not, and `settled != 'true'`. It calls the new `scripts/release/drop_eval_candidate.sh`, which deletes the nomination tag.

Deleting the tag is what makes the retry work, and it works because of two existing behaviours rather than any new bookkeeping. `resolve_promotion_candidate.sh` skips a commit that already carries an `evalcand_` tag, so while the tag is there the candidate is answered-about and will not be re-nominated. And `ensure_git_tag` skips silently when the tag already points at the same commit, so re-pushing the same name fires no webhook and could never have retried. Removing the tag returns the commit to the pool; the next nightly nominates it again and Prow sees a fresh push. A RED verdict is settled, so the tag stays and the candidate is rejected for good.

`drop_eval_candidate.sh` is the only script in this repository that runs `git push --delete`, so it refuses anything whose name does not match `EVALCAND_TAG_SHAPE_REGEX`, refuses a tag that does not point at the commit it was given, and refuses to read an unreachable remote as an absent tag. A tag genuinely gone from both sides is success. A remote delete that fails is **fatal** — an `::error title=Eval nomination not withdrawn` annotation and exit 1 — because the candidate is then stranded until someone deletes the tag by hand, and a warning on a green job is the weakest signal the platform has on the one job state nobody investigates.

**`hack/ci-eval-rc.sh`** stops describing itself as advisory in five places, and narrows what it calls RED.

The advisory prose is not a behaviour change — the driver already wrote the three-way verdict; it just said the verdict was for information. The RED narrowing is. Until now the driver split only on a failed deploy: every other non-zero exit of `hack/ci-eval-pr.sh` became RED, including the ones that happen before the first case runs — a ledger token that would not mint, a runner image short of `uv`, an `EVAL_REPETITIONS` that is not a positive integer. Under this pull request that distinction acquires teeth it did not have while the lane was advisory: RED is settled, so step 5b leaves the `evalcand_` tag in place, `resolve_promotion_candidate.sh` skips the commit from then on, and the candidate is never measured again. A broken runner would retire a candidate.

The separator is `bench-gate suite --markdown-out`, which writes `eval-verdict.md` in the roll-up. Its presence means a catalog was graded; its absence after a non-zero eval means the run stopped in preflight, which is now NOT RUN — the nomination is withdrawn and a later nightly asks again. The case this could not reach from outside — an eval that got to its roll-up and failed there because every repetition died on infrastructure, which writes the file either way — was #1766, and #1787 has since closed it from the inside: `bench-gate suite` now exits 2 and writes `outcome: not_evaluated` into `eval-verdict.json`, and the driver reads the pair. The two branches sit side by side and do not overlap — a run bench-gate declined to certify wrote both verdict files on its way out, and a run that stopped short of the roll-up wrote neither.

This came out of #1759's automated review, which found the poller's module docstring claiming a distinction the driver did not make. The docstring correction is in #1759; the narrowing is here, because this is the pull request that owns the file and the one that makes the distinction matter. Both sides of the pair now describe the same split.

**`resolve_promotion_candidate.sh`** now derives and emits the `evalcand_` tag, and treats either tag family as "already been through the promotion path". Both are checked rather than just the nomination, because candidates promoted before this change carry a `staging_` tag and no `evalcand_` one; keying on the nomination alone would re-nominate every one of them and spend a leased project re-measuring a build already on staging.

**Docs.** Nine files said the lane gates nothing, or that a green matrix produces a `staging_` tag. They now say what it gates. One incidental correction while in there: `docs/environment-reconcile.md` claimed the nightly reconciles `autopush`, which it does not — `autopush-deploy.yml` triggers on `workflow_run` of "Publish Images to GHCR" and on `workflow_dispatch`, with no tag trigger at all.

**Tests.** `tests/testing/rc_eval_driver.py` extracts the driver fixture so the unit and seam tiers share it rather than one subclassing the other's `TestCase`; `tests/integration/test_seam_rc_eval_verdict.py` runs the real driver's artifact through the real poller; `tests/test_nightly_pipeline_wiring.py` gains mutation-hardening tests over the new `if:` conditions and `needs:` edges.

## Context

Part of a three-pull-request change, and the last of them. Merge in order — this one is inert until the Prow side lands, and red without PR 2's helpers.

**Both prerequisites have merged, and the hold they were the condition for can be released.**

- GoogleCloudPlatform/oss-test-infra#2705 (PR 1 of 3) — **merged 2026-09-21** (`6e76567a`). Switches the Prow job's trigger from `staging_*` to `evalcand_*`, pins `RC_TAG` to the tag the job fired on rather than resolving the newest `rc_*`, and drops the `|| true` around the driver so a red eval leaves a `FAILURE` in the build.
- #1759 (PR 2 of 3) — **merged 2026-09-18** (`8a60c720`). The `evalcand_` tag family in `common.sh` and the verdict poller.

PR 1 replaced the trigger rather than adding to it, so the lane is currently dark: the Prow job now
fires on `evalcand_*`, which nothing pushes until this merges, while the nightly still pushes
`staging_`, which nothing evaluates. Staging keeps deploying, so this is the milder of the two
orderings rather than an outage — but the release-candidate eval measures nothing until this lands,
which is the reason to keep the gap short.

- gke-labs/kube-agents#1024 — the issue the eval lane was built for, including the "nothing this lane ships may block a release this week" constraint this change removes.
- gke-labs/kube-agents#1419 — the dashboard that currently consumes the lane's verdict; unaffected, it reads the same archive.

Now that #1759 has merged this is a single commit against `main`, and the whole diff — 31 files —
is this change's. The earlier note asking reviewers to read only the second of two commits no
longer applies.

No tracking issue.

**#1734 merged first, and this rebased onto it** (@toshiowang, `ci(nightly): add a test that
upgrade.sh can roll back to the last GA and forward`), as the earlier version of this section said
it would. It adds a `step-3b-rollback-leg` between the matrix and the promotion, reported and not
gating. The two are semantically compatible — a leg that gates nothing sits before the nomination
and neither job reads the other's result — but they edited the same lines, so the rebase merged
rather than chose:

- `.github/workflows/nightly-pipeline.yml`: `step-3b` keeps its place in the teardown's `needs` and
  the `!cancelled()` that lets it be listed for ordering without its result deciding anything, while
  the job is renumbered step 5 → step 7. `step-3b`'s own comment is updated where it named the old
  numbers ("step 5 does depend on it" → the teardown, now step 7).
- `#1734`'s four rollback-leg tests are kept, retargeted onto the renumbered job-name constants, and
  `test_the_rollback_leg_does_not_gate_the_promotion_yet` now asserts against the nomination job —
  which is where the promotion chain starts under this change, so a `needs` edge there would gate the
  eval and everything after it. Its fifth test,
  `test_the_promotion_tag_is_pushed_with_the_release_bot_token`, is dropped as superseded by this
  change's `test_both_tags_are_pushed_with_the_release_bot_token`, which covers all three tag jobs.
- `scripts/release/README.md` and `docs/designs/testing-strategy.md`: both sides' content kept.

Nothing here changes `step-3b`'s contract, and nothing in #1734 reads the eval verdict.

**#1787 also landed in the same window** (`fix(bench-gate): a run that lost an admitted case to
infrastructure is not evaluated, not green`) and touches `hack/ci-eval-rc.sh`, the same function
this change narrows. It closes #1766, which this pull request filed. The merge keeps both branches
— see the `ci-eval-rc.sh` paragraph above for why they do not overlap — and fixes one interaction
between them, recorded under Self-Review.

A scan of the rest of the open queue found no other overlap: #1756, #1716, #1714, #1694, #1690,
#1673, #1112 and #1099 touch none of these files. On the oss-test-infra side, the two open
kube-agents pull requests (#2668, #2665) touch the presubmit and periodic configs; PR 1 of this
stack touches only `kube-agents-postsubmits.yaml`.

## Testing

- `tests/test_nightly_pipeline_wiring.py` (+356/-59) — the `if:` conditions and `needs:` edges on steps 4, 5, 5b, 6 and 7, asserted term by term rather than by string equality, plus set-equality on step 5b's `needs`, an `||`-exclusion on its condition, index ordering on the `gcloud auth login` step against the poller step, and the three-link `--not-before` wiring.
- `tests/test_drop_eval_candidate.py` (new, 295) — 10 tests: the shape guard against every other tag family, the commit-match guard, the tag-absent-on-both-sides case, the unreachable-remote refusal, the remote-only delete-by-name path, the local-only dry run outside CI, the push in CI, and a remote that refuses the delete reding the job.
- `tests/integration/test_seam_rc_eval_verdict.py` (new, 144) — the real `hack/ci-eval-rc.sh` writing a summary that the real poller reads, for all three verdict words, plus the missing-summary case and the disagreement case (a GREEN summary under a build Prow failed).
- `tests/testing/rc_eval_driver.py` (new, 262) — the extracted driver fixture, whose `eval_writes_verdict` knob is what lets a test build an eval that fails without grading anything, and which now also carries #1787's `eval_body` so the two NOT RUN paths can be built independently. No `PYTHON_TEST_DIRS` glob is owed: it holds no `test_*.py`.
- `tests/test_ci_eval_rc.py`, `tests/test_resolve_promotion_candidate.py`, `tests/test_tag_commit.py`, `tests/test_record_nightly_candidate_summary.py` updated.
- Targeted run across this change's seven test files, after the rebase onto `main` at `4afb8ca5`: 125 passed, 89 subtests passed.
- Whole-suite differential against the same `main`, to confirm the rebase introduced nothing: `pytest tests/` (excluding `e2e` and `conformance`) gives an identical 26-failure set on both the branch and `main` — every one a local-environment artefact (the sandbox SIGKILLs `ruff`, and a locally-installed `mcp` predates `MCPServer`). Zero failures unique to the branch.
- `make docs-check` green — all six checks, context budget at 37.9k of 38,000.
- `prettier@3.9.6 --check` clean.

The full suite has 45 failures on this machine. They are not this change's: the same ten files fail identically at the base commit in a scratch worktree, from macOS bash 3.2, no Docker, no Go toolchain, no cluster, and `ruff` being SIGKILLed.

### Live validation

Not live-tested, and it is exempt: this is CI and release infrastructure, which
`.agents/rules/eval_driven_development.md` lists among the changes that do not alter what an agent
does. Nothing here runs in a cluster, in the operator, or in an agent pod.

What that leaves unverified, stated plainly, because one piece of it is load-bearing:

- **The delete-and-refire retry has not been observed.** That deleting a tag and re-pushing the
  same name produces a second `PushEvent`, and that Prow's postsubmit matcher treats it as a new
  build, is reasoned from Prow's source and from `ensure_git_tag`'s idempotent-skip — not from a
  run. Exercising it needs a real push to `gke-labs/kube-agents` and a real Boskos lease, neither
  of which exists before all three land. The first real test is the first nightly that hits an
  unsettled verdict.
- **The poller has never read a real archive.** Its fixtures are constructed; the bucket layout is
  asserted from `scripts/eval_dashboard/collect.py`, which reads the same prefix in production.
- **No real preflight refusal has been seen through the new NOT RUN branch.** The tests build an
  eval stub that exits non-zero without writing `eval-verdict.md`, which is the shape of the case
  but not an instance of it. That `bench-gate suite --markdown-out` writes the file in its roll-up
  and only there — so its absence really does mean no case was graded — is read off `ci-eval-pr.sh`
  and `bench-gate`'s argument handling, not observed. If that ordering is ever changed so the file
  is created early and filled in later, this branch silently stops firing and preflight refusals go
  back to being RED.
- **The 330-minute deadline against a 375-minute worst case.** A build that leases slowly and runs
  to its own ceiling can outlive the poller. That is reported as unsettled, so the candidate is
  re-nominated rather than rejected — correct behaviour, but it means such a night costs a leased
  project and produces no verdict. Worth an issue if it turns out to be common.

All of the above still holds at the rebased head: the rebase changed which branch of
`ci-eval-rc.sh` catches which failure, not whether any of them has been seen against a real
install, and it reaches no cluster either. One item narrows — the infrastructure-collapse case in
the third bullet is now #1787's branch rather than an acknowledged gap, and it arrived with its own
tests, but it has not been observed live either.

Nothing was pushed to any remote but the fork branch this pull request is built from; no test
artifacts to clean up. The scratch worktrees used to prove the pre-existing failures, on both
revisions, were removed.

## Self-Review

Two passes, each run in a context that did not write the change and was handed the diff range and
nothing else: `review-adversarial` and `review-docs-drift`. Everything below is the disposition as
it stands on the branch head, not a log of the rounds.

**Adversarial pass** — one blocker, four medium, one low. All six fixed.

1. **Blocker: the poller answered from the previous attempt's build, so the retry never
   re-measured anything.** Builds were matched by commit alone with no floor on build age, and the
   retry re-nominates the same commit — so the first sweep read last night's finished build,
   answered `NOT RUN` in seconds, withdrew again, and discarded a real eval that ran for hours.
   Fixed: `--not-before` in PR 2's poller, wired here by step 4's `nominated_at` stamp, with
   `test_the_poller_is_given_the_floor_the_nomination_stamped` pinning all three links (the stamp
   published, step 5 depending on step 4 to read it, the flag carrying it) and the `:?` that turns
   a wiring fault into a red job rather than a night of yesterday's verdict.
2. **High: two prerequisites in `GoogleCloudPlatform/oss-test-infra` that nothing named as
   prerequisites.** Both are PR 1 of this stack, and the Context section above now names it with
   the merge order and what it does. I read PR 1's diff to confirm both: the `branches` regex is
   `(^evalcand_[0-9]{10}_[0-9a-f]{7}$)` and the job exports `RC_TAG="${PULL_BASE_REF}"`, refusing
   rather than falling back to the newest `rc_*` when that variable is empty. Landing this before
   PR 1 makes every night end at `NO_BUILD`, withdraw, and stop staging advancing — visible, not
   silent, and the Risk section says so.
3. **Medium: every way the withdrawal could fail to reach the remote was reported as success.** A
   failed remote delete strands the candidate until a human deletes the tag, and it emitted a
   `::warning` on a green job. Fixed: fatal, with `::error title=Eval nomination not withdrawn`
   naming the tag and the repository, pinned by
   `test_a_remote_that_refuses_the_delete_reds_the_job`. The related conflation — a fetch that
   failed reading as a tag that was already gone — is fixed too: an unreachable remote refuses
   rather than reporting success, pinned by `test_a_remote_that_cannot_be_asked_is_not_an_absence`.
4. **Medium: three of the new tests passed against a broken implementation.** All three replaced.
   The withdrawal-condition test asserted three clauses were present but could not tell `&&` from
   `||`; it now excludes `||` outright. Step 5b's `needs` list was unpinned, so dropping step 5
   from it would have made the condition trivially true on every night including green ones; it
   now gets set-equality. The credential test joined every step's `run:` into one string, so a
   `gcloud auth login` moved *after* the poller kept it green; it now compares step indices.
5. **Medium: prose the change falsified, still on the branch head.** Three fixed:
   `common.sh`'s two "NOT WIRED YET" blocks (this is the pull request they were waiting for), and
   the dashboard caption in `pages.js`, which told a maintainer looking at a candidate that did not
   promote that the lane is advisory. The word "advisory" survives in three places on purpose — the
   non-inferiority comparison, which stays advisory, and the poller's comment explaining what
   `passed` meant under the old configuration.
6. **Low: step 5b's run record asserted the withdrawal before attempting it.** The summary now
   writes after the script, with a comment saying why.

The pass separately confirmed the promotion gate itself: it found no combination of job results,
skips, cancellations or timeouts that reaches `staging_` without GREEN, no `always()` /
`failure()` / `continue-on-error` anywhere in steps 4–7, and no injection seam in the `run:` blocks
(`COMMIT_SHA` and `VERDICT` arrive through `env:`, not `${{ }}`).

**Docs-drift pass** — seven blocking, five non-blocking. Six of the seven fixed here, one in PR 2.

1. `SCHEMA.md`'s `verdict` bullet said the lane is advisory and the job runs the driver under
   `|| true`. Rewritten: the job runs it bare, `RED` leaves a `FAILURE`, and the bullet now
   explains why `verdict` is still recorded separately — `NOT RUN` is also a `FAILURE`, and the two
   words have opposite consequences for the candidate.
2. `docs/README.md`'s status cell for `testing-strategy.md` still read "the release-candidate eval
   reports and gates nothing (§4.3)". Now says it gates the staging promotion.
3. `docs/environment-reconcile.md:56` described staging's trigger as the matrix alone.
4. `docs/designs/semver-deployment-versioning.md:29–31` and its mermaid node collapsed the two
   rungs into one.
5. `scripts/release/README.md:66` defined a `staging_` tag as "the full nightly matrix passed"
   while line 246 in the same file said matrix *and* eval.
6. `docs/designs/testing-strategy.md:188` said the job "resolves the newest release candidate",
   which would credit candidate B's verdict to candidate A. The reviewer could not read the Prow
   config to tell whether this was a doc error or a real gap; I read PR 1 and it is the former —
   the job pins `RC_TAG` to the tag it fired on. The paragraph now says that, and says why.
7. The site page carried maintainer-pipeline mechanics: the Prow job's name, the single-tag
   rationale, and the poller's 5.5-hour budget. All three removed. What stays is what a reader
   pinning releases can act on — `evalcand_*` deploys nothing, only `staging_*` is releasable, and
   the `staging_*` tag can appear most of a working day after its nightly started.

Of the five non-blocking: `docs/README.md`'s identifier-source row gains `evalcand_<ts>_<sha>`
(fixed in PR 2, where the tag family is introduced); `tests/e2e/README.md:309`'s "unpromoted
candidates" is now "neither nominated nor promoted"; `SCHEMA.md`'s `releases[]` example, which
shows a real captured build with a `staging_` `rc_tag`, gets the two-families note beside it rather
than an edit to the captured value. **Declined:** `docs/README.md:422`'s key-topics cell for the
site page — the reviewer's own note says it is "not wrong as written", and the map is edited from
several branches a week, so a cell that is merely incomplete is not worth the churn.
**Advisory, not acted on:** the two-tag rationale now appears in five places. Trimming the site
page removed the one that broke the audience rule; the rest are code comments (exempt) and the
runbook, where a reader who has just been told to push a tag should not have to follow a link to
learn why there are two.

**Rebase pass onto `main` at `4afb8ca5`** — the branch was re-read against the seven files `main`
moved under it, per the drift rule. Two of the merges were substantive (#1734 and #1787, both in
Context above), and reconciling them turned up one defect that neither side had on its own:

- **The two NOT RUN paths could describe the same run, contradicting each other.** #1787's summary
  paragraph decided whether to print "an admitted case lost every repetition to infrastructure"
  by re-deriving its branch from `eval_status -eq 2`. That was sound in isolation: before this
  change, a 2 the verdict JSON did not confirm was RED, so nothing reporting NOT RUN could reach
  the paragraph, and #1787's own comment says as much. This change breaks that premise — the
  roll-up-never-reached branch reports NOT RUN on *any* non-zero status, including a 2 from
  `argparse` or `uv run`. Such a run would print both explanations, the second telling the reader
  to rerun when the environment is healthy for a failure the environment had no part in.

  Fixed by setting a `suite_not_evaluated` flag where the branch is actually taken and reading
  that in the reporting step, rather than re-deriving a branch from a status that no longer
  identifies it. `test_a_preflight_refusal_exiting_2_is_not_described_as_weather` pins it; I
  confirmed it was load-bearing by restoring the re-derivation and watching it fail on
  `'could not certify' unexpectedly found`, with both paragraphs in the summary.

- **One test of this change's was too broad for the merged tree, and was narrowed rather than the
  code loosened.** `test_no_status_function_reaches_the_conditions_the_gate_rests_on` forbade every
  status function on four jobs including the teardown, which #1734 legitimately gives `!cancelled()`
  so the rollback leg can be listed for ordering without its result deciding anything. The teardown
  is not part of the eval gate — it names its gating results in its own condition — so it is out of
  that loop now, with the docstring saying why and pointing at the test that still forbids
  `always()` there. The three jobs the gate does rest on are unchanged.

**Carried from #1759's automated review** — one finding, whose fix lands here.

The bot found the poller's module docstring asserting that `hack/ci-eval-rc.sh` separates "the eval
found the candidate wanting" from "the lane broke before measuring anything", when the driver split
only on a failed deploy. I verified it from source rather than accepting it: `ci-eval-pr.sh` exits a
bare 1 for its preflight refusals with no distinct code, so the driver's `else verdict="RED"` swept
them in. Confirmed, and it matters more under this pull request than under the advisory lane it was
written for, because RED is settled here. #1759 carries the docstring correction; this carries the
narrowing, the test that pins it
(`test_an_eval_that_graded_nothing_is_not_a_red_candidate`), and the fixture knob it needs. I
checked the test was load-bearing by patching the new guard out and watching it fail with
`'| Verdict | NOT RUN |' not found` against a summary reading RED.

Two consequences inside this diff, both handled: the poller docstring paragraph #1759 adds
describes the pre-narrowing split, so it is rewritten here to describe the post-narrowing one, as is
the operator-facing RED summary string; and `read_verdict` gained a parameter in #1759, so the seam
test's call site is adapted here rather than there — the seam test is this pull request's file.

**Automated review on this pull request** — 3 findings. Two are the same defects
`kube-agents-bot` raised on #1759, which its pass here saw because this pull request's diff carries
#1759's commit as well; both are fixed in 747dc3b4 (the `read_verdict` double read of
`finished.json`, and `test_an_older_directory_without_a_started_json_is_skipped` returning on the
newer build before it reached the directory it names). The third is this pull request's own:

- **Medium: a candidate whose eval returned GREEN is never retried if the `staging_` push then
  fails.** Confirmed from the three pieces. `resolve_promotion_candidate.sh` skips any commit
  carrying an `evalcand_` tag; step 5b withdraws that tag only when
  `needs.step-5-await-eval-verdict.result != 'success'`; so a GREEN verdict followed by a step 6
  that failed on a rotated key or a rejected push leaves the tag on a commit with no `staging_` tag,
  and every later nightly skips it. Before this change a failed push left no tag and the next
  nightly simply retried, so this is a regression in retry behaviour rather than a pre-existing gap.

  One correction to the report's framing: it is not a stalled lane. The RC pipeline produces a newer
  validated candidate within hours, and that one nominates and promotes normally, so the cost is one
  green candidate dropped and staging a few hours staler.

  **Not fixed here, deliberately.** Withdrawing on a step 6 failure sends the candidate back for a
  fresh 1.5-to-3.5-hour eval on a project shared with the merge-blocking presubmit, to recover from
  a credential fault — worse than the thing it fixes at this blast radius. Recovering it without
  re-measuring needs a marker for "eval green, not yet promoted", which the tag namespace does not
  have and which is new mechanism this pull request should not introduce. Filed as #1767, named in
  a comment on step 5b's condition so the next person to edit that key sees why it is keyed this
  way, and the scheduler-retry paragraph in `scripts/release/README.md` — which the report correctly
  read as claiming this candidate would be retried — now states the exception.

**Known and deferred, not defects in this diff:**

- The poller and `scripts/eval_dashboard/collect.py` each carry their own `gsutil` wrappers over
  the same bucket. Sharing them means a module both a release script and a dashboard script can
  import, which is a refactor with its own blast radius.
- `tag_staging_promotion.sh` has no shape guard on the tag it is handed, where
  `tag_eval_candidate.sh` and `drop_eval_candidate.sh` both do. Pre-existing, and it only ever
  creates.
- The poller's deadline is checked between sweeps, not mid-scan, so a scan that starts just under
  the deadline can overrun it by up to one `gsutil` timeout. It fails closed (unsettled →
  withdraw → retry) and step 5's job ceiling is 15 minutes above the poller's deadline.

## Risk & Rollout

The blast radius is the nightly promotion path, and the failure mode is a candidate that does not
reach staging rather than a bad one that does. Every new gate is a refusal: if the poller crashes,
the eval never runs, or the archive is unreadable, step 6 does not fire and staging stays on the
previous build. Nothing new can push a `staging_` tag.

Merge order matters. Landing this before the Prow side means `evalcand_` tags are pushed that
nothing reacts to, so every night ends at `NO_BUILD` (exit 13), withdraws its nomination, and
staging stops advancing. Landing it after PR 1 and PR 2 is the intended order and the only one
that works.

The new failure mode worth naming: staging now advances less often. A night whose eval goes red,
times out, or never starts leaves staging where it was. That is the point of the change, but it
means "staging is a few days old" becomes a normal state rather than a symptom, and the nightly
summary is where to look for why.

One failure mode gets smaller rather than larger. Before this change every non-zero eval retired its
candidate; now a run that stopped in preflight withdraws the nomination instead, so a broken runner
costs a night rather than a candidate. The residue named here in the first revision, #1766, has
since been closed by #1787 from inside `bench-gate`, and this branch carries both halves.

Revert is dropping `step-5-await-eval-verdict` from step 6's `needs`, which restores the old
behaviour while leaving the eval running and reporting. The `ci-eval-rc.sh` narrowing is independent
of that and can stay: on the advisory lane it changes one word in a report nobody gates on.


scripts/release/poll_rc_eval_verdict.py
```diff
@@ -0,0 +1,793 @@
...
+        if passed is False:
+            print(
+                f"warning: {base} reports a GREEN verdict on a build that failed;"
+                " refusing to promote on it and reporting the lane as broken",
+                file=sys.stderr,
+                flush=True,
+            )
+            return VERDICT_NOT_RUN
+        return VERDICT_GREEN
+    if passed is True:
```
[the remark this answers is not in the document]
stalhaali: Confirmed and fixed in 747dc3b4, on #1759 — that commit owns `poll_rc_eval_verdict.py`; this pull
request's diff carries it because a fork branch cannot use a base that does not exist here.

Your reading of the mechanism is exact, including why the existing test cannot reach it: the
fixture's `unreadable` set is keyed by path, so a path fails every read and `scan_once` returns
before `read_verdict` is called. That shaped the regression test. Rather than build a fixture that
fails the Nth read of a path — which would test the fixture more than the poller — `read_file` now
records every read and `test_the_verdict_reads_each_artifact_of_a_build_once` asserts
`finished.json` is fetched exactly once. That pins the absence of the second read directly, which is
the invariant. Verified it fails against the previous code (`2 != 1`).

Took the first of your two suggested fixes: `read_verdict` now takes the text as a parameter and
`scan_once` hands over the bytes it already has, so the sweep's own read is the one that absorbs a
transient — `UNREADABLE` there means "say nothing, poll again", which is the right answer — and the
redundant `gsutil cat` is gone.

tests/test_poll_rc_eval_verdict.py
```diff
@@ -0,0 +1,889 @@
...
+        self.assertIsNone(state)
+        self.assertIsNone(build_id)
+
+    def test_an_older_directory_without_a_started_json_is_skipped(self):
+        """Deeper in the listing the same absence is a reaped artifact."""
+        read_listing, read_file = archive({"100": build(unstarted=True), "200": build()})
+        state, build_id, _ = poller.scan_once(COMMIT, read_listing, read_file)
+        self.assertEqual(state, poller.VERDICT_GREEN)
+        self.assertEqual(build_id, "200")
+
```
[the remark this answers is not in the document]
stalhaali: Confirmed and fixed in 747dc3b4, on #1759 — the test file is that commit's.

You are right that the assertion was satisfied by the newer build alone, so the `index == 0` guard
was never exercised and could have been changed freely. Took your suggested fixture exactly:
`{"300": build(commit=OTHER_COMMIT), "200": build(unstarted=True), "100": build()}`, expecting
`("green", "100")`, with `self.assertIn("200/started.json", read_file.reads)` added on top so the
test fails if the scan stops before reaching the reaped directory rather than merely returning the
right answer by another route.

scripts/release/resolve_promotion_candidate.sh
```diff
@@ -105,10 +117,23 @@ else
...
+    # and spend hours of a leased project re-measuring a build that is already on
+    # staging. The staging_ check is what makes the transition free; it can be
+    # dropped once no such candidate can still be the newest validated one.
+    existing_evalcand_tag="$(get_existing_evalcand_tag "${COMMIT_SHA}")"
     existing_staging_tag="$(get_existing_staging_tag "${COMMIT_SHA}")"
-    if [ -n "${existing_staging_tag}" ]; then
+    if [ -n "${existing_evalcand_tag}" ]; then
+      SKIP_PROMOTION="true"
+      SKIP_REASON="Commit ${COMMIT_SHA:0:7} is already nominated as '${existing_evalcand_tag}'; the matrix still runs, no eval is started and nothing is tagged."
+      echo "ℹ️ ${SKIP_REASON}" >&2
```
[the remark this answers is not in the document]
stalhaali: Confirmed. Real, and a regression in retry behaviour rather than a pre-existing gap — before this
change a failed staging push left no tag and the next nightly re-tested the same candidate.

Verified the three pieces rather than taking the chain: `resolve_promotion_candidate.sh` sets
`skip_promotion=true` on any commit carrying an `evalcand_` tag; step 5b's condition is
`needs.step-5-await-eval-verdict.result != 'success'`, so a GREEN verdict does not trigger the
withdrawal; step 6 can fail on a rotated `RELEASE_BOT_APP_PRIVATE_KEY`, a rejected push or a dead
runner. The tag then sits on a commit with no `staging_` tag and every later nightly skips it.

One correction to the framing, not to the finding. This does not stall staging. The RC pipeline
produces a newer validated candidate within hours and that one nominates, evaluates and promotes
normally, so the cost is one green candidate dropped plus a few hours of extra staleness — not a
lane that stops advancing until someone intervenes.

**Not fixing it here, and the reason is the cost of the two available fixes.** Withdrawing the tag
on a step 6 failure is the obvious one, and it sends a candidate whose verdict is already GREEN back
for a fresh 1.5-to-3.5-hour eval on a project shared with the merge-blocking presubmit — to recover
from what is usually a credential fault, and to reach the same verdict. That is worse than the thing
it fixes at this blast radius. Recovering it without re-measuring needs a marker for "eval green,
not yet promoted", which the tag namespace does not have: the resolver cannot distinguish that from
"eval still running", because both are an `evalcand_` tag and no `staging_` tag. Adding a third tag
family, with its own cleanup and transition questions, is not something the pull request introducing
the gate should also do.

So: filed as #1767 with the analysis and the `evalpass_` sketch, named in a comment on step 5b's
`if:` so the next person to edit that key sees what it deliberately does not catch, and recorded
under Self-Review.

Your last point was the sharpest one and I have taken it. `scripts/release/README.md`'s
scheduler-retry paragraph did claim this candidate would be retried. It now names the exception,
says staging waits for the next validated candidate, and points at re-running the failed job from
the Actions UI as the recovery that reuses the verdict already on record.

stalhaali:
/hold

Holding this until GoogleCloudPlatform/oss-test-infra#2705 (PR 1 of 3) merges. The Context section
says the order matters; this is so a `/lgtm` and `/approve` cannot merge past it.

The ordering is not symmetric, which is the reason for the label rather than a note. PR 1 *replaces*
the eval job's trigger — `^staging_[0-9]{10}_[0-9a-f]{7}$` becomes `^evalcand_...$` — rather than
adding to it. So merging this one first means the nightly pushes an `evalcand_` tag that nothing in
Prow fires on: the poller waits out its 45-minute appearance clock, exits 13, step 5b withdraws the
nomination, and step 6 never pushes a `staging_` tag. Staging stops deploying, and the eval stops
running too, because it is still waiting on a `staging_` tag that no longer appears. Both halves of
the lane go dark, every night, until someone notices.

The other order is mild by comparison. PR 1 first means the nightly still pushes `staging_` and
staging keeps deploying, while the eval job waits for `evalcand_` tags nothing pushes yet — the eval
goes dark, which costs visibility rather than safety while the lane is still advisory. That window
should be kept short, so the two are best merged close together, but it is not a reason to hold PR 1.

#1759 (PR 2 of 3) is deliberately not held: nothing in it depends on PR 1's trigger, and landing it
first drops this pull request's diff from 33 files to 31 and retires the "review the second commit
only" note above.

`/hold cancel` once oss-test-infra#2705 is merged.

stalhaali:
Holding this on #1842, not on anything in the diff.

@jayantid raised it on the Prow side ([oss-test-infra#2705 review](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2705#discussion_r4049046145)): the job this PR waits on has been killed at its 360m deadline every night since Sep 16, and the poller here gives up at 330m. I re-read the archive today and it holds, with two nights added since that review — Sep 16 through Sep 20, five consecutive kills at 363–365m against a 365m wall, no successful run in the window.

Merging as-is would nominate a candidate every night, time out at 330m, withdraw the nomination at step 5b, and never reach step 6. Staging would stop moving, and silently: a TIMEOUT is not a rejection, so nothing would go red to explain it. That is a worse position than today's informational eval.

Two things I found that the review did not name, both in #1842:

- The RC job leaves `EVAL_TASK_PARALLELISM` unset, so it runs at the `hack/ci-eval-pr.sh` default of 4, while `ci-kube-agents-eval-nightly` runs the same 117-unit catalog at 6. The job's `decoration_config` comment reasons carefully about copying the nightly's *timeout*; the parallelism was not copied with it.
- Copying it would not be enough. The nightly at parallelism 6 ran 357–483m over the same five nights and overran its own 480m budget twice, so the catalog does not fit under 330m at 6 either.

Raising the poller's deadline is not available: the wait step is `runs-on: ubuntu-latest` at `timeout-minutes: 345`, and GitHub-hosted jobs cap at six hours. The 330m is imposed from outside rather than chosen.

Nothing here changes the diff, and I would rather not reshape this PR around the problem — the fix is in the eval's wall clock or in how the verdict gets back to GitHub, and neither belongs in a PR about the tag family. Review is still welcome; the hold is on the merge.

/hold

docs/designs/testing-strategy.md
```diff
@@ -183,36 +183,40 @@ Two things run pinned from the merge target rather than from the pull request:
 
 A baseline is therefore valid for exactly one combination of five versions: fleet, harness, verifiers, judge model, agent model. Every baseline is keyed on all five, and a key that does not match the run is reported stale rather than silently compared against. Bumping any of the five does not mean weeks of blind gating: re-running the suite against the merge target backfills the baselines on demand. The baselines live in the evidence store — `bench/baselines/` in the checkout, or the bucket `EVAL_BASELINE_STORE` names ([`bench/baselines/README.md`](../../bench/baselines/README.md)) — which the nightly writes and the presubmit only reads, so a pull request cannot move the window it is judged against.
 
-### 4.3 Release gate: one test gates, the suite reports
+### 4.3 Release gate: one test gates the candidate, the suite gates staging
```
jayantid: 🟡 **Product requirements.** The lane was chartered non-gating with an explicit "nothing may block a release" constraint, and this PR removes that constraint without a decision record or an arming criterion.

#1024 set the constraint and re-chartered the lane as advisory "until the nightly holds enough baselines to arm it honestly". This PR turns the verdict into the staging gate and says "No tracking issue". `testing-strategy.md` section 4.3 is rewritten to describe the gate, but nothing says what had to be true of the eval before it was allowed to hold staging: pass rate on known-good candidates, wall clock under the poller's budget, a flake rate. #1842 shows the wall-clock half is not met today, which is why you are holding.

If it merges on the strength of #1842 alone, the next regression in eval health holds staging again with no stated bar to measure it against.

Fix: a short arming-criteria paragraph in section 4.3, or an issue that owns it, and lift the hold against that rather than against one symptom.

.github/workflows/nightly-pipeline.yml
```diff
@@ -312,37 +361,257 @@ jobs:
...
+        # The summary is written after the script, not before it, and says what
+        # happened rather than what is about to. A delete that fails now reds
+        # this step, and a summary claiming the tag "is being deleted" would be
+        # the only thing a reader sees in the run's overview — stating as done
+        # the one thing that did not happen, on the run where the candidate is
+        # stranded and needs a hand.
+        run: |
+          ./scripts/release/drop_eval_candidate.sh
+          {
+            echo "### Eval nomination withdrawn"
```
jayantid: 🟡 **UX flow.** A lane that breaks every night looks exactly like a lane with nothing to promote, and #1842 is the proof.

Step 5b withdraws the nomination, writes a job summary, and the run goes red in the Actions tab. Nothing else moves. `ci-health.yml` already posts to `#kube-agents-ci-health` and files an OUTAGE issue for the presubmit gate, and it excludes the RC glob on purpose. So five consecutive unsettled nights, the state the archive has been in since Sep 16, were found by a reviewer reading GCS, not by the pipeline that depends on them.

If it merges, the first week the eval outgrows its budget again, staging quietly stops advancing and the only signal is a red nightly nobody is paged on.

Fix: the smallest version is a consecutive-unsettled counter in `ci-health.yml`'s adjudicator posting to the same Chat space after two nights. Step 5b filing or adopting an issue the way ci-health does is the alternative.

.github/workflows/nightly-pipeline.yml
```diff
@@ -312,37 +361,257 @@ jobs:
...
+    runs-on: ubuntu-latest
+    # Above the poller's own 330-minute deadline, so the poller is the thing that
+    # decides — it writes a verdict and a job summary on the way out, and a job
+    # killed at its timeout writes neither. The headroom covers checkout and the
+    # gcloud auth dance. It cannot be raised much further: GitHub caps a job at
+    # 360 minutes, and the Prow job's own budget (360m plus 5m grace plus up to
+    # 10m waiting for a Boskos lease) can legitimately exceed what fits here. A
+    # build that outlasts the poller reports `timeout`, which fails closed and
+    # says in the summary that it is not a verdict on the release.
+    timeout-minutes: 345
```
jayantid: 🟡 **Technical design.** The wait lives inside a GitHub-hosted job with a hard six-hour cap, so the gate can only ever accept an eval shorter than that, and the catalog outgrew the cap in one week.

Your own comment above says the 330 minutes "cannot be raised much further". The hold comment on this PR says the fix is "in the eval's wall clock or in how the verdict gets back to GitHub". The second option is the design question: a push-based completion, the Prow job's last step firing `repository_dispatch` with the verdict, or a `workflow_run`-style trigger, has no ceiling, releases the runner, and drops the `--not-before` race on `started.json`. The tag push and the release-bot credential stay in Actions either way.

If it merges as polling, every catalog growth is a future #1842, and the nightly concurrency group is held for eight hours each night.

Fix for this PR: record the ceiling and the push alternative in section 4.3 with the criterion for switching. The switch itself is a follow-up.

stalhaali:
The hold was on #1842 — the eval lane not finishing inside the poller's deadline. `5a5bf16` addresses
it here: `hack/ci-eval-rc.sh` was exporting `EVAL_TIER=nightly`, so this lane graded 42 cases at
three repetitions (126 units) against a 360-minute Prow budget sized for the presubmit matrix's 54.
It now exports `presubmit`. The cause, the timeline, and what the change does *not* settle are in
https://github.com/gke-labs/kube-agents/issues/1842#issuecomment-5798356118.

Two things a reviewer should weigh rather than take from me:

- **The new duration is arithmetic, not a measurement.** 54 units at the per-unit rate the nightly
  periodic implies puts the lane near 175m. No run has graded the presubmit matrix through this lane
  since #1620, and the lane has been dark since 09-20, so the first nightly after this merges is the
  first real number.
- **#1880 is unfixed and can still red this lane.** It is a separate cause — its regression boundary
  postdates every timeout behind the hold — but a delegation-ceiling stall carries no
  `INFRA_FAILURE_MARKER`, so `scoring.py:152` grades it as the agent being wrong. A RED verdict is
  settled and skips the commit permanently, where a timeout retries.

/hold cancel

docs/designs/testing-strategy.md
```diff
@@ -183,36 +183,40 @@ Two things run pinned from the merge target rather than from the pull request:
...
+The suite runs here too, and it gates the staging promotion. A separate Prow job (`hack/ci-eval-rc.sh`) measures the candidate the nomination tag names — the job pins `RC_TAG` to the tag it fired on, rather than letting the driver fall back to resolving the newest `rc_*`, because the verdict decides that nomination and no other — checks out its tree, installs its published images and runs the full catalogue against them through the same `hack/ci-eval-pr.sh` the presubmit uses; the candidate is judged against `main`'s evidence window and never added to it. The verdict lands in the eval dashboard's Brief as a release-candidates table — tag, commit, tier, verdict, artifacts, one row per measured candidate ([`scripts/eval_dashboard/SCHEMA.md`](../../scripts/eval_dashboard/SCHEMA.md)).
```
jayantid: 🟡 **Product requirements.** The tier commit narrows the gate to the presubmit matrix, and the design doc and the operator page this PR rewrote still say the gate runs the full catalogue.

`5a5bf16` sets `RC_EVAL_TIER="presubmit"`: 18 cases, not 42. Four lines this PR edits say otherwise. `testing-strategy.md:3` and `:190` say the release-candidate job runs the full catalogue and gates staging. Line 252's tier table says the candidate measurement gates nothing. `release-versioning.md:25` tells operators `evalcand_` starts "the full-catalog agent eval". The PR body says the narrowing is called out under Risk & Rollout. It is not. The decision itself, #1842's "scope question nobody has been asked", is answered in a shell comment.

If it merges, a maintainer reading section 4.3 believes the 24 nightly-only cases hold staging. A regression in one of them reaches staging with nobody expecting it.

Fix: correct the four lines, and add one paragraph to 4.3 naming which matrix gates staging, why the wait's ceiling bounds it, and that #1842 owns widening it.

.github/workflows/nightly-pipeline.yml
jayantid: 🟡 **UX flow.** A lane that times out every night looks exactly like a lane with nothing to promote, and #1842 is the proof.

Step 5b withdraws the nomination, writes a job summary, and the run goes red in the Actions tab. Nothing else moves. `ci-health.yml` posts to `#kube-agents-ci-health` and files an OUTAGE issue for the presubmit gate, and it excludes the RC glob on purpose. Five consecutive unsettled nights, Sep 16 to 20, were found by a reviewer reading GCS, not by the pipeline that depends on them.

If it merges, the 175-minute figure is the first thing tested, and the PR body calls it arithmetic, not a measurement. If it is wrong, staging quietly stops advancing and the only signal is a red nightly nobody is paged on. The PR body's own words: "It costs a leased project per night until someone looks."

Fix: the smallest version is a consecutive-unsettled counter in `ci-health.yml`'s adjudicator posting to the same Chat space after two nights. Step 5b filing or adopting an issue the way ci-health does is the alternative.

jayantid: 🟡 **UX flow · breaks today.** A lane that times out every night looks exactly like a lane with nothing to promote, and #1842 is the proof.

Step 5b withdraws the nomination, writes a job summary, and the run goes red in the Actions tab. Nothing else moves. `ci-health.yml` posts to `#kube-agents-ci-health` and files an OUTAGE issue for the presubmit gate, and it excludes the RC glob on purpose. Five consecutive unsettled nights, Sep 16 to 20, were found by a reviewer reading GCS, not by the pipeline that depends on them.

If it merges, the 175-minute figure is the first thing tested, and the PR body calls it arithmetic, not a measurement. If it is wrong, staging quietly stops advancing and the only signal is a red nightly nobody is paged on. The PR body's own words: "It costs a leased project per night until someone looks."

Fix (a follow-up): the smallest version is a consecutive-unsettled counter in `ci-health.yml`'s adjudicator posting to the same Chat space after two nights. Step 5b filing or adopting an issue the way ci-health does is the alternative.

docs/designs/testing-strategy.md
jayantid: 🟡 **Product requirements · contradicts design.** The tier commit narrows the gate to the presubmit matrix, and the design doc and the operator page this PR rewrote still say the gate runs the full catalogue.

`5a5bf16` sets `RC_EVAL_TIER="presubmit"`: 18 cases, not 42. Four lines this PR edits say otherwise. `testing-strategy.md:3` and `:190` say the release-candidate job runs the full catalogue and gates staging. Line 252's tier table says the candidate measurement gates nothing. `release-versioning.md:25` tells operators `evalcand_` starts "the full-catalog agent eval". The PR body says the narrowing is called out under Risk & Rollout. It is not. The decision itself, #1842's "scope question nobody has been asked", is answered in a shell comment.

If it merges, a maintainer reading section 4.3 believes the 24 nightly-only cases hold staging. A regression in one of them reaches staging with nobody expecting it.

Fix (a few lines): correct the four lines, and add one paragraph to 4.3 naming which matrix gates staging, why the wait's ceiling bounds it, and that #1842 owns widening it.

.github/workflows/staging-promotion-pipeline.yml
```diff
@@ -312,37 +361,257 @@ jobs:
...
+        # this step, and a summary claiming the tag "is being deleted" would be
+        # the only thing a reader sees in the run's overview — stating as done
+        # the one thing that did not happen, on the run where the candidate is
+        # stranded and needs a hand.
+        run: |
+          ./scripts/release/drop_eval_candidate.sh
+          {
+            echo "### Eval nomination withdrawn"
+            echo
+            echo "The release-candidate eval returned \`${VERDICT:-no verdict at all}\` for \`${EVALCAND_TAG}\`, which is not an answer about the candidate. Its nomination tag has been deleted, so the next nightly can measure this candidate again."
```
jayantid: 🟡 **UX flow · breaks today.** A lane that times out every night looks exactly like a lane with nothing to promote, and #1842 is the proof.

Step 5b withdraws the nomination, writes a job summary, and the run goes red in the Actions tab. Nothing else moves. `ci-health.yml` posts to `#kube-agents-ci-health` and files an OUTAGE issue for the presubmit gate, and `docs/ci-health.md` says only presubmit runs reach its rules and the digest. Five consecutive unsettled nights, Sep 16 to 20, were found by a reviewer reading GCS, not by the pipeline that depends on them.

If it merges, the 175-minute figure is the first thing tested, and the PR body calls it arithmetic, not a measurement. If it is wrong, or a #1880 stall lands on the lane, staging quietly stops advancing and the only signal is a red nightly nobody is paged on. The PR body's own words: "It costs a leased project per night until someone looks."

Fix (a follow-up): the smallest version is a consecutive-unsettled counter in `ci-health.yml`'s adjudicator posting to the same Chat space after two nights. Step 5b filing or adopting an issue the way ci-health does is the alternative.

docs/designs/testing-strategy.md
jayantid: 🟡 **Product requirements · contradicts design.** The gate grades the presubmit matrix, and the design doc and the operator page this PR rewrote still say it runs the full catalogue.

`hack/ci-eval-rc.sh` sets `RC_EVAL_TIER="presubmit"`. Three lines this PR edits say otherwise. `testing-strategy.md:190` says the release-candidate job runs the full catalogue. Line 252's tier table says the candidate measurement gates nothing. `release-versioning.md:25` tells operators `evalcand_` starts "the full-catalog agent eval". The PR body says the narrowing is called out under Risk & Rollout. It is not. The width decision, #1842's "scope question nobody has been asked", lives only in a shell comment, and that comment is already stale: it says 18 cases and 54 units, and the blocking roster has been 12 cases since 2026-09-22.

If it merges, a maintainer reading section 4.3 believes the nightly-only cases hold staging. A regression in one of them reaches staging with nobody expecting it.

Fix (a few lines): correct the three lines, and add one paragraph to 4.3 naming which matrix gates staging, why the wait's ceiling bounds it, and that #1842 owns widening it.

stalhaali:
Smoke test at `6707a8c` came back `NOT EVALUATED`, not red. `cluster-agent-crashloop-debug` lost all three repetitions to the 2700s delegation ceiling, and an admitted case with nothing scored stops the suite certifying green. No case failed on merits: four passed all three repetitions, seven went unstable, and all four delegation ceilings in the build landed in the cluster-debugging family.

That is #1840, which names this case in its title and carries the `presubmit-gate` label. Evidence from this build posted there: https://github.com/gke-labs/kube-agents/issues/1840#issuecomment-5820164478

/retest

